PR #337, opened 2023-09-28
Client binaries builds

drogus:
# Description of Changes

This PR adds builds for client binaries for all of the platforms that we support. The binaries are uploaded to a DO Spaces bucket.

# API and ABI


jdetter (approved):
This is awesome, thank you Piotr!

PR #290, opened 2023-09-13
Store the current module hash in a system table

kim:
# Description of Changes

Adds a new system table `st_module` which stores the hash of the stdb module currently associated with the database. The hash is updated _transactionally_ when the database is initialized or updated (that is, the table does not have any rows until the `init` procedure ran successfully).

`st_module` also associates an "epoch" number with the row, which is used to store a _fencing token_. As sketched out in the changes to standalone, this allows to use an external (distributed) locking service to safely order lifecycle operations on the database. We can thus guarantee _at-most-once_ semantics for `init` and `update` reducers.

# API and ABI

 - [ ] This is a breaking change to the module ABI
 - [ ] This is a breaking change to the module API
 - [ ] This is a breaking change to the ClientAPI
 - [ ] This is a breaking change to the SDK API
 - [X] This is a breaking change to the database internals

Because a new system table is introduced, compatibility with existing databases cannot be maintained. The patch is against `cloud-next`, which contains breaking changes anyways.

PR #289, opened 2023-09-12
Wrangle benchmarks

kazimuth:
# Description of Changes

Rewrite benchmarks to allow direct comparison between:
- `spacetime_module`: Spacetime via a module
- `spacetime_raw`: Spacetime via the RelationalDB struct
- `sqlite`

And add some targeted serialization benchmarks.

PR #267, opened 2023-09-06
Refactoring bootstrapping and adding full support for constraints & multi-column indexes

mamcx:
# Description of Changes

### NOT REVIEW IT YET!

Note: This is a multi-step set of PRs to bring constraints + multi-column indexes

Step:
- [x] [Add system tables](https://github.com/clockworklabs/SpacetimeDB/pull/225)
- [x] [Add multi-column indexes](https://github.com/clockworklabs/SpacetimeDB/pull/243)
- [x] [Refactoring IDs for system objects](https://github.com/clockworklabs/SpacetimeDB/pull/439)
- [ ] Rework schema definition
- [x] Rework bootstrapping

# API and ABI

 - [x] This is a breaking change to the module ABI
 - [x] This is a breaking change to the module API
 - [x] This is a breaking change to the ClientAPI
 - [x] This is a breaking change to the SDK API

*If the API is breaking, please state below what will break*

PR #342, opened 2023-09-29
Fix bootstrapping of ST_MODULE

kim:
# Description of Changes

System tables are inserted into the committed state lazily, but since ST_MODULE (introduced in #290) is not referenced by other system tables, it won't. Interestingly, this only surfaces in tests written for #289.

This is mostly a quick fix, hoping #267 will generalize bootstrapping enough to cover this case.

# API and ABI


gefjon (approved):
Fixes tests in #289 .

cloutiertyler (approved):
Approved as a temporary fix until this is properly fixed in #267

PR #344, opened 2023-09-29
Print the elapsed time & # of returned rows in the interactive repl

mamcx:
# Description of Changes

To help do interactive timing of queries, now it print on the repl:

```bash
└──────────────────────────────────────────────────────────┘
🪐bitcraft>select * from st_table where table_id = 1
 table_id | table_name | table_type | table_access
----------+------------+------------+--------------
 1        | st_columns | system     | public

[Elapsed  ]: 14.62ms.
[Result  0]: 1 rows.
```

# API and ABI

### DIFF
--- a/crates/cli/src/subcommands/repl.rs
+++ b/crates/cli/src/subcommands/repl.rs
@@ -41,7 +41,7 @@ sort by
 pub async fn exec(con: Connection) -> Result<(), anyhow::Error> {
     let database = con.database.clone();
     let mut rl = Editor::<ReplHelper, DefaultHistory>::new().unwrap();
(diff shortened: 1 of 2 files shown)


gefjon (approved):
Neat!

cloutiertyler (changes requested):
Please fix the smoketests that this PR breaks. I would also recommend using the same format as `psql` for timing:

```
# \timing on
Timing is on.

# select 1234;        
 ?column? 
----------
     1234
(1 row)

Time: 0.203 ms
```

cloutiertyler (approved):
LGTM! very cool

PR #345, opened 2023-09-29
Benchmarks: report tx/s as elements/s

gefjon:
# Description of Changes

Based on my limited understanding of Criterion, I think we can get add transactions per second reporting to benchmarks using Criterion's Throughput::Elements tracking, by treating transactions as elements.

This is not ideal, as the output says e.g. `thrpt:  [38.763 Kelem/s 38.781 Kelem/s 38.799 Kelem/s]`, where we'd like `Ktx/s`, but it's not clear whether Criterion allows us that much customization.

Each of the benchmarks currently does one tx per iteration, so pass `Throughput::Elements(1)` for each of them. Do this as close to the test as possible, despite redundancy, in case we ever add benchmarks to the same group which do multiple transactions per iteration.

# API and ABI


cloutiertyler:
benchmarks please

cloutiertyler (approved):
LGTM

PR #346, opened 2023-10-02
Commit test clients' module_bindings

gefjon:
# Description of Changes

Having discussed with the team, I've come around to it being correct to commit the `module_bindings` for the SDK tests' two test clients.

This way, people not working on codegen can run `cargo test --workspace` &c, and open the test clients with rust-analyzer without any additional rigamarole.

The downside is that any time codegen changes,
we'll see very large commits touching all of these files, but consensus is we don't really care.

An additional upside is that now we can run clippy against the bindings in CI, which will incentivize ensuring that the bindings generate lint-free code.

# API and ABI


gefjon:
> Maybe worth adding `linguist-generated=true` for those folders?

Oh, that's pretty slick!

RReverser:
Actually, can you maybe add this to top-level `.gitattributes`? That way it will cover any `module_bindings`, including those for existing projects in `examples`.

cloutiertyler (approved):
LGTM, still has warnings, but we can fix later.

PR #348, opened 2023-10-02
SDK tests use a tempdir as STDB_PATH, rather than ~/.spacetime

gefjon:
# Description of Changes

We've had continual issues with test isolation when developing breaking changes. This commit doesn't fully address those, but is a step in the right direction: the SDK tests now create a tempdir as their `STDB_PATH`, rather than using `~/.spacetime`.

# API and ABI

### DIFF
--- a/crates/testing/src/sdk.rs
+++ b/crates/testing/src/sdk.rs
@@ -11,10 +11,14 @@ use tempfile::TempDir;
 
 pub fn ensure_standalone_process() {
     lazy_static! {
-        // TODO: this leaks the tempdir, as dtors for lazy_statics never run.
-        static ref STANDALONE_PATH: TempDir = TempDir::with_prefix("stdb-sdk-test").expect("Failed to create tempdir");
         static ref JOIN_HANDLE: Mutex<Option<JoinHandle<()>>> = {
-            std::env::set_var("STDB_PATH", STANDALONE_PATH.path());
+            let stdb_path = TempDir::with_prefix("stdb-sdk-test")
+                .expect("Failed to create tempdir")
+                // TODO: This leaks the tempdir.
+                //       We need the tempdir to live for the duration of the process,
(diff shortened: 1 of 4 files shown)


crates/testing/src/sdk.rs
```diff
@@ -7,11 +7,16 @@ use std::{collections::HashSet, fs::create_dir_all, sync::Mutex};
...
 use crate::modules::{module_path, CompiledModule};
 use std::path::Path;
+use tempfile::TempDir;
 
 pub fn ensure_standalone_process() {
     lazy_static! {
-        static ref JOIN_HANDLE: Mutex<Option<JoinHandle<()>>> =
-            Mutex::new(Some(std::thread::spawn(|| invoke_cli(&["start"]))));
+        // TODO: this leaks the tempdir, as dtors for lazy_statics never run.
+        static ref STANDALONE_PATH: TempDir = TempDir::with_prefix("stdb-sdk-test").expect("Failed to create tempdir");
```
RReverser, line 15: I'd inline this into initializer below - it doesn't need to be its own lazy static I think?

gefjon: Done.

RReverser: Alternatively, wrap it into `ManuallyDrop` to make the intentional leak clearer.

kulakowski: Yes, as written this will drop the TempDir at the end of the block.

gefjon: I don't think that's true; I believe `into_path` leaks the tempdir.

kulakowski: Ah gotcha, I think this is new in the tempdir to tempfile update. too sensitive to this after too many footgun bugs around the tempdir crate, sorry. lgtm

RReverser (approved):

PR #351, opened 2023-10-02
Preliminary module-side interface for timing execution

gefjon:
# Description of Changes

At a high level, define a struct `Span` in `bindings` which manages a timing span, with methods `start` and `end` to begin and end timing, respectively. `Span::start` records the `Instant` of its call; `Span::end` computes the elapsed `Duration` and prints it to the module logs.

At a low level, add a slab of timing spans to WasmInstanceEnv, with ABI functions `_start_span` and `_end_span` to implement `Span::start` and `Span::end`.

# API and ABI


cloutiertyler (approved):
This looks fantastic to me

PR #354, opened 2023-10-03
New Upgrade Version Util

jdetter:
# Description of Changes

 - New upgrade-version tool written in rust that actually works properly

If you're in the `tools/upgrade-version` directory, you can run the command via:
```bash
cargo run -- 0.7.0
```

# API and ABI


cloutiertyler (approved):
LGTM

PR #356, opened 2023-10-03
Change base64 crate to simple-base64 crate

drogus:
# Description of Changes

Standard base64 encode/decode functions have been deprecated and it's clear that the maintainer doesn't care about ergonomics of using the crate[1], so I've forked the crate to be able to get back to a simple format of calling the decode/encode functions.

# API and ABI


crates/sdk/Cargo.toml
```diff
@@ -14,7 +14,7 @@ spacetimedb-client-api-messages = { path = "../client-api-messages", version = "
 
 anyhow.workspace = true
 anymap.workspace = true
-base64.workspace = true
+simple-base64.workspace = true
```
kulakowski, line 17: Reorder to alphabetical please.

crates/core/Cargo.toml
```diff
@@ -26,7 +26,7 @@ spacetimedb-client-api-messages = { path = "../client-api-messages", version = "
 anyhow.workspace = true
 async-trait.workspace = true
 backtrace.workspace = true
-base64.workspace = true
+simple-base64.workspace = true
```
kulakowski, line 29: Ditto.

kulakowski (approved):
Up to you when to rebase and land this

kim (changes requested):
I disagree with this. The only reason to move away from the base64 crate seems to be a convenience function, which we could easily define in lib or wherever. I am fairly sure that we would depend on the original base64 crate regardless, even if transitively.

kulakowski:
> I disagree with this. The only reason to move away from the base64 crate seems to be a convenience function, which we could easily define in lib or wherever. I am fairly sure that we would depend on the original base64 crate regardless, even if transitively.

I don't feel strongly. I'll clarify, though, that simple-base64 is a fork of the original base64 and not a wrapper. So there would not be a transitive dependency on it.

kim:
> simple-base64 is a fork of the original base64 and not a wrapper.

Well, that's my point. Because base64 is the obvious choice for the purpose, it is quite likely to be re-introduced either by us or transitively -- which will just make a future audit of our dependency footprint harder.

There are surely reasons to avoid certain crates even if "obvious choice", but then we should ensure that we avoid them entirely (e.g. via `cargo deny`) and provide ample documentation about what to use instead and why. 

I just don't think this is a good reason, because we could solve the perceived nuisance with a wrapper.

kulakowski:
> > simple-base64 is a fork of the original base64 and not a wrapper.
> 
> Well, that's my point. Because base64 is the obvious choice for the purpose, it is quite likely to be re-introduced either by us or transitively -- which will just make a future audit of our dependency footprint harder.
> 
> There are surely reasons to avoid certain crates even if "obvious choice", but then we should ensure that we avoid them entirely (e.g. via `cargo deny`) and provide ample documentation about what to use instead and why.
> 
> I just don't think this is a good reason, because we could solve the perceived nuisance with a wrapper.

Ahh I understand now, I misunderstood and thought you had the specific objection that the new crate *itself* transitively depended on it.

Yeah.  I could see adopting this if other people do, given that? I don't feel as against it as Kim does but certainly see the line of reasoning.

kulakowski:
@drogus any thoughts about the above discussion?

PR #364, opened 2023-10-04
Typo Fix

BiswajyotiRay:
# Description of Changes
Typo Fix

### DIFF
--- a/.github/workflows/linux-build.yml
+++ b/.github/workflows/linux-build.yml
@@ -18,7 +18,7 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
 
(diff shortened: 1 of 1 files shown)


.github/workflows/linux-build.yml
```diff
@@ -18,7 +18,7 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
 
-      - name: Compile Aarch64 Docker
```
jdetter, line 21: The only actual typo here I think is actually this should be `AArch64` instead of `Aarch64`, but this is minor enough where I would leave it. AArch64 and ARM64 are mostly interchangable so in this case I would probably leave this as is. Thank you for the thought though ❤️

PR #367, opened 2023-10-05
C# SDK - Fix Genererated FilterByX for Option Columns

jdetter:
# Description of Changes

 - This fixes the generation issues related to `FilterByX` where X is a column that is of type `Option<Y>`.
 - This has been tested and I was able to get into the BitCraft world with no compilation errors or warnings

# API and ABI


crates/cli/src/subcommands/generate/csharp.rs
```diff
@@ -1629,6 +1629,7 @@ pub fn autogen_csharp_globals(items: &[GenItem], namespace: &str) -> Vec<Vec<(St
         output.indent(1);
     }
 
+    writeln!(output, "[ReducerClass]").unwrap();
```
jdetter, line 1632: This fixes another unrelated issue to the `FilterByX` functions

cloutiertyler (approved):
LGTM

PR #369, opened 2023-10-05
Fix deadlock

coolreader18:
# Description of Changes

Ok I guess it was kinda a deadlock in the semaphore. I guess that's why they say FuturesUnordered is a footgun, huh. Basically, it was given the permit but not waking up because we weren't polling `handle_queue` anymore at that point. See [Status quo of an AWS engineer: Solving a deadlock](https://rust-lang.github.io/wg-async/vision/submitted_stories/status_quo/aws_engineer/solving_a_deadlock.html).

# API and ABI

### DIFF
--- a/crates/client-api/src/routes/subscribe.rs
+++ b/crates/client-api/src/routes/subscribe.rs
@@ -274,6 +274,10 @@ async fn ws_client_actor(client: ClientConnection, mut ws: WebSocketStream, mut 
     log::debug!("Client connection ended");
     sendrx.close();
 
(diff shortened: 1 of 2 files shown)


jdetter (approved):
I've verified that this fixes both my repro and fixes BitCraft - thank you Noa 👍

kurtismullins:
> I've verified that this fixes both my repro and fixes BitCraft - thank you Noa 👍

👍 Confirmed that it also fixed BitCraft for me locally as well.

PR #370, opened 2023-10-05
Bench for SQL scan / where

mamcx:
# Description of Changes

- Bench for SQL queries: SELECT * FROM table /  SELECT * FROM table WHERE pk = value
- Instrument calls for tracy
- Add tracy support in run_standalone_temp.sh

# API and ABI

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -190,7 +190,7 @@ wasmtime = { version = "7", default-features = false, features = ["cranelift"] }
 # We use the "ondemand" feature to allow connecting after the start,
 # and reconnecting, from the tracy client to the database.
 # TODO(George): Need to be able to remove "broadcast" in some build configurations.
-tracing-tracy = { version = "0.10.2", features = [
+tracing-tracy = { version = "0.10.3", features = [
   "enable",
   "system-tracing",
   "context-switch-tracing",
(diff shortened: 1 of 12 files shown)


joshua-spacetime (approved):
LGTM!

kazimuth:
This PR broke the module benchmarks, FWIW. I would have appreciated if you had waited for my review.

mamcx:
So, what should I do instead?

kazimuth:
Rather than adding trait methods to `BenchDatabase`, it would be best to add a separate file `spacetime_sql.rs` with a new struct `SpacetimeSQL : BenchDatabase` that measures database access via the SQL API. This new file could use `TableSchema` for `TableId`.

This means that each generic benchmark file is measuring a separate access path to the database. Times of the SQL api will be clearly separate from times of reducer calls.

The specific thing that broke here is that a `PascalCase` name has been used instead of a `snake_case` due to the refactoring in this branch, here: https://github.com/clockworklabs/SpacetimeDB/pull/370/files#diff-6179b2533c85dd1dc6d18d56e1bff1a01fb6c4dfcd2861a08d7ea565ad1f02fbR192

I'm working on a branch that reverts this, so that benchmarks will run on master. Then I'll start a new PR to add the spacetime-SQL benches, we can collab on that :)

(We'll need to have no-ops for insertion benchmarks there, in some way. I guess we'll need to add an `insertion_supported() -> bool` method to `BenchDatabase`.)

PR #349, opened 2023-10-02
chore: Add trace logging to all datastore iterators

joshua-spacetime:
Fixes #347.

# Description of Changes

# API and ABI

 - [ ] This is a breaking change to the module ABI
 - [ ] This is a breaking change to the module API
 - [ ] This is a breaking change to the ClientAPI
 - [ ] This is a breaking change to the SDK API

*If the API is breaking, please state below what will break*

PR #371, opened 2023-10-05
Disable Tracing in master for Release Mode

jdetter:
# Description of Changes

 - A PR was merged earlier this week that had a negative impact on performance: https://github.com/clockworklabs/SpacetimeDB/pull/349
 - This PR just disables tracing by default, which should mostly bring us back to where we were in terms of performance
 - This does so only for the release profile, not for the debug profile.

# API and ABI


joshua-spacetime (approved):

kulakowski:
I'd just clarify in the PR description that it does so for the release profile, not also for the debug profile.

PR #375, opened 2023-10-05
Build linux binary on bare metal

drogus:
# Description of Changes

Build linux binary on bare metal
# API and ABI


jdetter (approved):
Seems fine

PR #378, opened 2023-10-06
Add some comments to `FutureQueue` and the `subscribe` route

gefjon:
# Description of Changes

Written while I was backtracking the deadlock fix from yesterday.

# API and ABI


crates/client-api/src/routes/subscribe.rs
```diff
@@ -180,7 +184,13 @@ async fn ws_client_actor(client: ClientConnection, mut ws: WebSocketStream, mut
         let message = tokio::select! {
             // NOTE: all of the futures for these branches **must** be cancel safe. do not
             //       change this if you don't know what that means.
+
+            // If we have a result from handling a past message to report,
+            // grap it to handle in the next `match`.
             Some(res) = handle_queue.next() => Item::HandleResult(res),
+
+            // If we've received an incoming message,
+            // grap it to handle in the next `match`.
```
kulakowski, line 193: spelling: grap?

coolreader18 (approved):

PR #379, opened 2023-10-06
SATS: Flatten AlgebraicType (Nix BuiltinType)

Centril:
# Description of Changes

Flattens `BuiltinType` into `AlgebraicType`.
See note below on BSATN ABI breakage.

# API and ABI

BSATN does not change for `AlgebraicValue` but it does for `AlgebraicType` as this includes tags for each variant so the indirection of `BuiltinType` was present in the codec but is no longer.


crates/sats/src/algebraic_type.rs
```diff
@@ -342,7 +394,7 @@ mod tests {
     fn algebraic_type() {
         let algebraic_type = AlgebraicType::meta_type();
         assert_eq!(
-            "(sum: (variants: Array<(name: (some: String | none: ()), algebraic_type: &0)>) | product: (elements: Array<(name: (some: String | none: ()), algebraic_type: &0)>) | builtin: (bool: () | i8: () | u8: () | i16: () | u16: () | i32: () | u32: () | i64: () | u64: () | i128: () | u128: () | f32: () | f64: () | string: () | array: &0 | map: (key_ty: &0, ty: &0)) | ref: U32)",
+            "(ref: U32 | sum: (variants: Array<(name: (some: String | none: ()), algebraic_type: &0)>) | product: (elements: Array<(name: (some: String | none: ()), algebraic_type: &0)>) | array: &0 | map: (key_ty: &0, ty: &0) | bool: () | i8: () | u8: () | i16: () | u16: () | i32: () | u32: () | i64: () | u64: () | i128: () | u128: () | f32: () | f64: () | string: ())",
```
mamcx, line 397: Question: Because the type is structural is not the order significant?

mamcx (approved):
LGTM.

Note: We don't have a way to assert the changes in the bindings are correct

cloutiertyler:
Closing in favor of: https://github.com/clockworklabs/SpacetimeDB/pull/1559

PR #156, opened 2023-08-08
Shrink SATS & friends

Centril:
# Description of Changes

Shrinks various types in SATS and friends, primarily by storing `(len: u32, ptr: NonNull<T>)` for owned slices of `T`, i.e. `Box<[T]>` but with `u32` length, rather than using an expensive `Vec<T>`. As `len: u32` now holds, larger errors will result in errors.

Some other improvements such as moving boxing out of `MapType` to `AlgebraicType` is done as well. Two chosen regressions (because they are assumed to be statistically uncommon) are introduced boxing for `u128` and `MapValue` as this is required to shrink `AlgebraicValue`. The `BuiltinType` and `BuiltinValue` types are also done away with, as they were in the way of shrinkage.

Examples:
- `AlgebraicValue` goes from 32 bytes to 16.
- `ProductValue` goes from 24 bytes to 12.
- `AlgebraicType` goes to 16 bytes.

Other changes in this PR are mostly a consequence of the above, with some drive-by improvements to reduce cloning and whatnot.

After this PR, we could elect to do some small string and vector optimizations for e.g., `SatsString`, `SatsVec<u8/u16/...>`.

# API

 - [ ] This is a breaking change to the module API
 - [ ] This is a breaking change to the ClientAPI
 - [ ] This is a breaking change to the SDK API

*If the API is breaking, please state below what will break*

PR #380, opened 2023-10-07
`build_csharp`: ensure project path exists 

Centril:
# Description of Changes

This makes `spacetime generate --lang rust --project-path modules/sdk-test-connect-disconnect-cs` work better so that you don't get:
> `Error: Failed to build project. dotnet not found in path. Please install the .NET Core SDK.`

Written by @jdetter as part of #156.

# API and ABI

### DIFF
--- a/crates/cli/src/tasks/csharp.rs
+++ b/crates/cli/src/tasks/csharp.rs
@@ -1,4 +1,6 @@
+use anyhow::Context;
 use duct::cmd;
+use std::fs;
 use std::path::{Path, PathBuf};
 
 pub(crate) fn build_csharp(project_path: &Path, _build_debug: bool) -> anyhow::Result<PathBuf> {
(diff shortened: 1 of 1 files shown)


crates/cli/src/tasks/csharp.rs
```diff
@@ -11,6 +12,14 @@ pub(crate) fn build_csharp(project_path: &Path, _build_debug: bool) -> anyhow::R
         std::fs::remove_file(&output_path)?;
     }
 
+    // Ensure the project path exists
+    if fs::metadata(project_path).is_err() {
```
cloutiertyler, line 16: Why are we not returning the actual error? Are we sure there's only one error that can be returned?

cloutiertyler (changes requested):
Question about the error.

cloutiertyler (approved):
LGTM

PR #384, opened 2023-10-08
Reduce cloning of `Header` & co

Centril:
# Description of Changes

The primary purpose of this PR is to nix some unnecessary `.clone()`s in `Header`, `MemTable` and friends by making cloning more explicit by not cloning inside constructor functions like `::new`.

# API and ABI

### DIFF
--- a/crates/lib/src/relation.rs
+++ b/crates/lib/src/relation.rs
@@ -89,10 +89,8 @@ impl FieldName {
     }
 
     pub fn table(&self) -> &str {
-        match self {
-            FieldName::Name { table, .. } => table,
-            FieldName::Pos { table, .. } => table,
-        }
+        let (FieldName::Name { table, .. } | FieldName::Pos { table, .. }) = self;
+        table
     }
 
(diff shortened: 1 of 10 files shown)


cloutiertyler:
benchmarks please

crates/lib/src/relation.rs
```diff
@@ -182,44 +187,41 @@ pub struct Header {
...
-                },
-            };
-            cols.push(Column::new(name, f.algebraic_type));
-        }
+    pub fn new(table_name: String, fields: Vec<Column>) -> Self {
+        Self { table_name, fields }
+    }
+
+    pub fn from_product_type(table_name: String, fields: ProductType) -> Self {
+        let cols = fields
```
cloutiertyler, line 205: I wonder if these iter chains hurt debug build performance more than the loop form on the left. Do you happen to have any knowledge of how that works @Centril?

cloutiertyler (approved):
This seems like a reasonable and positive change. Just please see the one question/comment.

PR #385, opened 2023-10-09
Simplify the `encoding_roundtrip` tests

Centril:
# Description of Changes

Simplifies the `encoding_roundtrip` tests.
Extracted from the sats shrinking effort to facilitate review.

# API and ABI


kim (approved):
LGTM

PR #386, opened 2023-10-09
Create multi-column indexes defined in modules

### LINKED ISSUE #368: Module support for multi-column index definitions
For example
```Rust
#[spacetimedb(table)]
#[spacetimedb(index(btree, name = "multi", x, y))]
pub struct Point {
  x: i64,
  y: i64,
}
```

kim:
# Description of Changes

We ignored them before.

Fixes #368 

# API and ABI

### DIFF
--- a/crates/core/src/host/db.rs
+++ b/crates/core/src/host/db.rs
@@ -8,10 +8,17 @@ use nonempty::NonEmpty;
 use spacetimedb_lib::IndexType;
 use spacetimedb_sats::{ProductType, Typespace};
 
-use crate::db::{
-    datastore::{self, locking_tx_datastore::MutTxId, traits::IndexId},
-    relational_db::RelationalDB,
-};
+use crate::db::datastore::{self, traits::IndexId};
+
+/// Magic table id.
+///
+/// When a [`datastore::traits::TableDef`] is constructed, the corresponding
+/// table does not yet exist (typically) and thus no table id is known.
+///
+/// Nevertheless, [`datastore::traits::IndexDef`] requires to specify a table
+/// id. The magic value can be used as a placeholder, which the datastore will
+/// replace with the actual value upon creation of the table.
+const AUTO_TABLE_ID: u32 = 0;
 
 /// Schema information for a single table, as extracted from an STDB module.
 pub struct ModuleTableSchema<'a> {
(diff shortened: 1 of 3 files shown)


crates/core/src/host/wasm_common/module_host_actor.rs
```diff
@@ -761,6 +764,21 @@ impl<T: WasmInstance> WasmModuleInstance<T> {
             }
         }
 
+        // Multi-column indexes cannot be unique (yet), so just add them.
```
kim, line 767: Actually, we might be able to support it with an additional attribute on the macro. I can add it if we want it right now.

kim: Should that PR be split into smaller pieces maybe?

crates/core/src/host/wasm_common/module_host_actor.rs
```diff
@@ -761,6 +764,21 @@ impl<T: WasmInstance> WasmModuleInstance<T> {
...
+        let multi_col_indexes = table.indexes.iter().filter_map(|index| {
+            if index.col_ids.len() > 1 {
+                Some(IndexDef {
+                    table_id: 0,
```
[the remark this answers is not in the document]
cloutiertyler: It will be an autoinc value (there should be a comment here though)

cloutiertyler, line 771:
```suggestion
                    table_id: 0, // Autogenerated by autoinc
```

kim: https://github.com/clockworklabs/SpacetimeDB/pull/386/commits/001fcf163c153f0853ca748ea25f6052dfb81e8b#diff-840af797a96dc965f1b75269b1c343356ecb6a0b4e28b84c611a0db514e9d3faR21

kim:
> Could we add a module test that adds a multi-column index to a table definition?

I don't know how to do this end-to-end, but I split those helper into a separate module to allow for unit-level tests (001fcf1).

kim:
> Although we're still missing the rust macro bindings I believe for actually being able to define a multi-column index.

This already works (`#[spacetimedb(index(btree, name = "le_boeuf", col_a, col_b))]`), but doesn't allow to say that the index should be unique. I believe #267 adds support for that.

crates/core/src/host/db.rs
```diff
@@ -0,0 +1,648 @@
...
+    let mut indexes_to_create = Vec::new();
+    let mut indexes_to_drop = Vec::new();
+
+    let mut known_tables: BTreeMap<String, datastore::traits::TableSchema> = current
+        .into_iter()
+        .map(|schema| (schema.table_name.clone(), schema))
+        .collect();
+
+    for proposed_schema_def in proposed {
+        let proposed_schema_def = proposed_schema_def?;
```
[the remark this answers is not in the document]
kim: One of those Rustisms: to extract the `Result`s, I'd need to allocate a new `Vec` at the call site, and then will be asked why allocate :shrug: 

I'd be happy to pass two `Vec`s instead.

kim:
Closing in favour of #267 and #417

PR #388, opened 2023-10-09
Move the two `Box`es in `MapType` to one in `BuiltinType`

Centril:
# Description of Changes

Instead of having one `Box` around the `key_ty` and one around the (value) `ty`, we have one around the entire ´MapType`.
Why? Because both values tend to be accessed together so having one Box is more cache friendly. Moreover, this enables shrinking `AlgebraicType` later on, as `Box<MapType>` is 8 bytes vs. the old `MapType` being 16 bytes.

Extracted from https://github.com/clockworklabs/SpacetimeDB/pull/156.

# API and ABI


mamcx (approved):
LGTM

PR #389, opened 2023-10-09
Flatten `BuiltinValue` into `AlgebraicValue`

Centril:
# Description of Changes

Flattens `BuiltinValue` into `AlgebraicValue` to facilitate #156.
This does not flatten the type structure.
Therefore there is no ABI break as `BuiltinValue` was never noticeable in BSATN.

# API and ABI


mamcx (approved):
LGTM

drogus:
benchmarks please

PR #390, opened 2023-10-09
fix: README.md wording should be DB + App Server

### LINKED ISSUE #201: Readme wording should be DB + App server
I noticed this line at the top of the readme:
"You can think of SpacetimeDB as both a database and server combined into one."

This is confusing to me, I think it should read:
"You can think of SpacetimeDB as both a database and application server combined into one."

A server could be anything! Initially I thought you wanted to replace the OS as well.

JoyelJohny:
# Description of Changes

Readme word reading as DB + App Server
Closes #201 

# API and ABI

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@
 
 ## What is [SpacetimeDB](https://spacetimedb.com)?
 
-You can think of SpacetimeDB as both a database and server combined into one.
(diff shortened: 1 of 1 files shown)


JoyelJohny:
Please can you add hacktoberfest and hacktoberfest-accepted labels

PR #391, opened 2023-10-09
Refactor system_tables + locking_tx_datastore tests with functions

Centril:
# Description of Changes

Adds some helper functions like `column_row` that help out in `system_tables` and in `locking_tx_datastore` tests.
This helps out with SATS shrinking by dealing with strings in a more central location. (first commit)

Also simplifies some test setup code (second commit).

# API and ABI

### DIFF
--- a/crates/core/src/db/datastore/locking_tx_datastore/mod.rs
+++ b/crates/core/src/db/datastore/locking_tx_datastore/mod.rs
@@ -2151,9 +2151,9 @@ impl traits::MutProgrammable for Locking {
 
 #[cfg(test)]
 mod tests {
-    use super::{ColId, Locking, StTableRow};
+    use super::{ColId, Locking, MutTxId, StTableRow};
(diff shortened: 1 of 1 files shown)


crates/core/src/db/datastore/system_tables.rs
```diff
@@ -248,36 +260,13 @@ pub fn st_table_schema() -> TableSchema {
                 is_unique: true,
             },
         ],
-        columns: vec![
```
cloutiertyler, line 251: I'm going to veto these changes. This is some of the most critical code in the whole system, and the changes here make it more difficult to read and less obvious what is really going on. Particularly with the field names not inline like they were.

Additionally this change seems unrelated to Sats, no?

cloutiertyler (changes requested):
Please revert the changes to the `systems_table.rs` file. See my comment inline for informatino.

PR #396, opened 2023-10-10
Benchmark modules in release mode

RReverser:
# Description of Changes

Previously module benchmarks were running against modules compiled in debug mode, which made them much slower and not represent real-world performance.

This PR fixes that and adds explicit `CompilationMode` enum that needs to be passed to `CompiledModule::compile` to prevent accidental debug builds in other places.

# API and ABI


kazimuth (approved):
Damn, good catch. LGTM.

PR #400, opened 2023-10-10
Collect timing information about WASM execution in reducers

gefjon:
# Description of Changes

This commit augments our logging of reducer execution time by logging the time spent specifically executing WASM code, in addition to the overall elapsed time.

This is accomplished by adding two time-related fields to `WasmInstanceEnv`, which we interact with in four contexts:
- before calling into a reducer.
- after returning from a reducer.
- after entering a host function called by a reducer.
- before returning from a host function called by a reducer.

The timings are probably not terribly precise,
as we necessarily capture a small amount of overhead during calls and returns, but they're good enough for our purposes,
i.e. knowing whether slow reducers are slow
because they spend a lot of time in the datastore, or because they spend a lot of time doing WASM computations.

EDIT: Converting to draft, as `perf` profiling seems to show this has non-negligible overhead when running BitCraft - slightly over 1% of time spent in a certain reducer we're concerned about. We'll need to figure out whether to feature-gate, whether the info is valuable enough to make the overhead worthwhile, or whether we can collect the same info in a more efficient way.

# API and ABI

### DIFF
--- a/crates/core/src/host/wasmer/wasm_instance_env.rs
+++ b/crates/core/src/host/wasmer/wasm_instance_env.rs
@@ -1,6 +1,7 @@
 #![allow(clippy::too_many_arguments)]
 
 use crate::database_logger::{BacktraceFrame, BacktraceProvider, ModuleBacktrace, Record};
+use crate::host::instance_env::InstanceEnv;
 use crate::host::scheduler::{ScheduleError, ScheduledReducerId};
 use crate::host::timestamp::Timestamp;
 use crate::host::wasm_common::{
@@ -9,9 +10,8 @@ use crate::host::wasm_common::{
 };
 use bytes::Bytes;
 use itertools::Itertools;
+use std::time::{Duration, Instant};
 use wasmer::{FunctionEnvMut, MemoryAccessError, RuntimeError, ValueType, WasmPtr};
-
-use crate::host::instance_env::InstanceEnv;
 
 use super::{Mem, WasmError};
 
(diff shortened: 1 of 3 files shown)


gefjon:
benchmarks please

crates/core/src/host/wasmer/wasm_instance_env.rs
```diff
@@ -35,9 +47,79 @@ fn mem_err(err: MemoryAccessError) -> RuntimeError {
...
+/// This version mutably borrows `caller` once before `body`, and again after,
+/// but does not hold the borrow during `body`.
+macro_rules! with_timing_paused {
+    ($caller:ident -> { $($body:tt)* }) => {{
+        $caller.data_mut().timing_after_exit_wasm();
+
+        // Improvised try block: errors bubbled with `?` are captured in `res`,
+        // and the `timing_before_enter_wasm` runs regardless.
+        //
+        // Clippy will warn about this if `body` doesn't include a `?` or `return`,
```
kulakowski, line 74: I'm no expert at the interactions of clippy and macros but this seems fine to me.

kulakowski (approved):

crates/core/src/host/wasmer/wasmer_module.rs
```diff
@@ -377,10 +389,23 @@ impl WasmerInstance {
...
+
+        // Close the `execution_duration` span and collect the overall execution time.
+        // Unlike `wasm_duration`, this includes time spent in host functions,
+        // e.g. evaluating queries.
+        let duration = start.elapsed();
+
+        // Clear buffers and timing spans to avoid leaking data across reducer runs.
         self.env.as_mut(store).buffers.clear();
+        self.env.as_mut(store).iters.clear();
+        self.env.as_mut(store).timing_spans.clear();
```
kulakowski, line 404: I'm now wondering why you are doing this here but not other places that were calling `self.env.as_mut(store).buffers.clear()`, specifically in `call_describer`. I guess that we just don't care about the timing in that path right now?

Also, why `self.env.as_mut(store).iters.clear()` is now being called here now.

IMO the lack of encapsulation between wasm_instance_env and instance_env makes this challenging to understand.

gefjon: Ah, I didn't mean to commit this change. At some point we should investigate what state we do and don't persist between calls, as I think we're probably leaking some of it that we don't want to, but I think that's out of scope.

gefjon:
Converting to draft, as perf profiling seems to show this has non-negligible overhead when running BitCraft - slightly over 1% of time spent in a certain reducer we're concerned about. We'll need to figure out whether to feature-gate, whether the info is valuable enough to make the overhead worthwhile, or whether we can collect the same info in a more efficient way.

PR #406, opened 2023-10-10
Allow running `crates/sdk` test suite on windows

kazimuth:
Had some `ncurses` builds fail today on my Windows machine. Apparently the approved solution of the `ncurses` crate is to not use it on windows:

https://github.com/jeaye/ncurses-rs/issues/127#issuecomment-506444431

The `pancurses` crate supports Linux and Windows -- and in theory OS X, if anybody's using that, it just falls back to the `ncurses` crate on any Unix -- so I switched `cursive`'s backend to use that.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1085,7 +1085,7 @@ dependencies = [
  "libc",
  "log",
  "maplit",
- "ncurses",
+ "pancurses",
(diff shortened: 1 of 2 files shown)


kazimuth:
Closed in favor of #412

PR #407, opened 2023-10-10
Add a section to the PR template requiring an estimate of complexity and risk

gefjon:
# Description of Changes

As discussed in our 0.7.0 retrospective / post-mortem, add a section to the PR template requesting an estimate of complexity and risk in the changes. The template includes an explicit and somewhat verbose explanation of what is expected.

# API and ABI

1; this is not a code change.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,7 +1,6 @@
 # Description of Changes
 
 
-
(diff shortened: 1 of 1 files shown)


cloutiertyler (approved):
LGTM

PR #408, opened 2023-10-10
Use newtypes `ColId`, `TableId`, `IndexId`, `SequenceId` everywhere (*)

Centril:
# Description of Changes

All cases of e.g., `col_id/table_id/index_id/sequence_id: u32` are done away with and the newtypes are used instead for clearer code and more type safety. Moreover, the change in using `ColId` means we have to do `.clone()` and `.map(..)` less as we can stay in `ColId`-land until we actually have to transform from/into `AlgebraicValue`. I hope to eventually do away with all cases of reallocation due to the u32/ColId type difference as the representations are the same so a "transmute" should be possible.

The `St*Fields` enums now also use a macro for their definition to reduce boilerplate and to keep the entire definition in one place (strings, indices, and variant names).

# API and ABI

 - [x] This is a breaking change to the module API

The module API breaks because the newtypes is used in the `bindings*` crates now.

I'd rate this PR a 2.5; each change is rather trivial in itself, but there are a lot of them so there's some risk in numbers.
Moreover, there are some nixed reallocations and new allocations.


crates/sdk/tests/test-client/src/module_bindings/mod.rs
```diff
@@ -861,147 +861,147 @@ impl SpacetimeModule for Module {
...
-            "update_unique_u16" => _reducer_callbacks.handle_event_of_type::<update_unique_u_16_reducer::UpdateUniqueU16Args, ReducerEvent>(event, _state, ReducerEvent::UpdateUniqueU16),
-            "update_unique_u32" => _reducer_callbacks.handle_event_of_type::<update_unique_u_32_reducer::UpdateUniqueU32Args, ReducerEvent>(event, _state, ReducerEvent::UpdateUniqueU32),
-            "update_unique_u64" => _reducer_callbacks.handle_event_of_type::<update_unique_u_64_reducer::UpdateUniqueU64Args, ReducerEvent>(event, _state, ReducerEvent::UpdateUniqueU64),
-            "update_unique_u8" => _reducer_callbacks.handle_event_of_type::<update_unique_u_8_reducer::UpdateUniqueU8Args, ReducerEvent>(event, _state, ReducerEvent::UpdateUniqueU8),
-            unknown => { spacetimedb_sdk::log::error!("Event on an unknown reducer: {:?}", unknown); None }
-        }
+match &function_call.reducer[..] {
+						"delete_pk_address" => _reducer_callbacks.handle_event_of_type::<delete_pk_address_reducer::DeletePkAddressArgs, ReducerEvent>(event, _state, ReducerEvent::DeletePkAddress),
```
kulakowski, line 865: I don't get why this changed? Not a big deal.

crates/sats/src/convert.rs
```diff
@@ -48,3 +49,27 @@ built_in_into!(f32, F32);
 built_in_into!(f64, F64);
 built_in_into!(&str, String);
 built_in_into!(&[u8], Bytes);
+
```
kulakowski, line 52: Here and some places it looks like `ConstraintId` isn't handled? I'm assuming `ConstraitId` is a type that we expect to use a lot more of shortly? And so it's named but not always plumbed all the way through etc

If that take is correct, I'm indifferent between you adding the `impl` for it here now, or later when it's need.

crates/core/src/vm.rs
```diff
@@ -746,10 +748,10 @@ pub(crate) mod tests {
             p,
             ST_SEQUENCES_NAME,
             (&StSequenceRow {
-                sequence_id: 1,
+                sequence_id: SequenceId(1),
```
kulakowski, line 751: Mostly my curiosity, I don't demand this sort of consistency: How come sometimes you write 1.into() and sometimes SequenceId(1)?

crates/bindings-sys/Cargo.toml
```diff
@@ -11,3 +11,4 @@ bench = false
 
 [dependencies]
 getrandom = {workspace = true, optional = true}
+spacetimedb-primitives = { path = "../primitives" }
```
kulakowski, line 14: Again for my edification: Why does this one not need to name the 0.7.1 tag? Is it because bindings-sys isn't published or something?

kulakowski (approved):
I know there's some outstanding big chunks of work at least @mamcx needs to get in, I trust you to coordinate about patch ordering.

kulakowski:
> I know there's some outstanding big chunks of work at least @mamcx needs to get in, I trust you to coordinate about patch ordering.

Ah I see now that mario pushed to this :)

crates/sdk/tests/test-client/src/module_bindings/mod.rs
mamcx: Maybe a change in `fmt`.

crates/sats/src/convert.rs
mamcx: Good catch. I have that in my PRs but here this move to another crate so I miss it.

crates/core/src/vm.rs
mamcx: I do it pure by convenience, but should we enforce one way or another?

crates/bindings-sys/Cargo.toml
mamcx: I left it as was before. I also wonder if we require the tag.

Issue #401, opened 2023-10-10
Figure out why rerunning benchmarks causes clean rebuild


RReverser:

kazimuth:
This is definitely caused by the benchmarks module compilation. If I don't run the module benchmarks, the issue doesn't arise.

In fact, it's caused by compiling the benchmarks module in release mode. If you compile the module in debug, the issue doesn't arise.

kazimuth:
Interestingly, the same issue doesn't arise when running `cargo test` in `crates/testing` -- that uses the same infra to compile some modules in debug mode, but it doesn't need to recompile anything in debug after running the tests. So something particular is happening with release mode.

RReverser:
> So something particular is happening with release mode.

FWIW just to save time - one more thing I tried after our conversation is removing custom `profile.bench` from `Cargo.toml` in case two different profiles collected into same directory were causing an issue, but that didn't fix it either.

RReverser:
Ok, at least now I know it _is_ some issue with workspace interaction when building modules in release mode.

It seems that when build scripts (of all those dependencies we have) are invoked due to module compilation, they are being executed in same folder as build scripts invoked due to SpacetimeDB's own compilation, so next time Cargo considers them all fresh.

I don't understand why they're being put in the same folder instead of respecting `--target` and being put into Wasm-specific dir, or why it's not an issue when building in development mode, but at least I can confirm that splitting out target dirs helps.

For now I'm adding a workaround by temporarily setting `CARGO_TARGET_DIR` when building benchmarks, but longer-term I think we'll have to create a 2nd workspace for all the Wasm modules to properly split them out in Cargo's view, so that it automatically puts them into different target folders.

RReverser:
Fixed by #411.

PR #411, opened 2023-10-11
Compile benchmarks in separate folder

RReverser:
# Description of Changes

Imperfect workaround for #401 that caused clean rebuild of SpacetimeDB with all 3rd-party deps each time you do `cargo bench`.

# API and ABI


kazimuth (approved):
LGTM.

PR #412, opened 2023-10-11
Switch cursive to cross-platform backend

RReverser:
# Description of Changes

By default cursive uses ncurses, which is Unix-only library. This led to build failures on Windows due to dependency on cursive in the workspace itself.

As a simple fix I'm switching it to a pure Rust cross-platform backend (crossterm).

# API and ABI

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1062,6 +1062,31 @@ dependencies = [
 ]
 
 [[package]]
+name = "crossterm"
+version = "0.25.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


kulakowski:
I think @kazimuth was also looking at this via https://github.com/clockworklabs/SpacetimeDB/pull/406 ?

RReverser:
Ah lol missed that. I saw pancurses backend too, but it seemed better to switch to pure-Rust dependency as these tend to be more cross-platform. I don't mind either one though.

gefjon (approved):

PR #415, opened 2023-10-11
test(414): multi-column index scan

### LINKED ISSUE #414: Add test for multi-column index scan
Add a test for calling `iter_by_col_eq` with multiple columns, and ensure the index is scanned when there is an applicable multi-column index present.

joshua-spacetime:
Closes #414.

Adds a test for calling iter_by_col_eq with multiple columns. Ensures the index is scanned if an applicable multi-column index is present. To write this test, iter_by_col_eq was updated to take multiple columns.

# Description of Changes

# API and ABI

### DIFF
--- a/crates/core/src/db/datastore/locking_tx_datastore/mod.rs
+++ b/crates/core/src/db/datastore/locking_tx_datastore/mod.rs
@@ -207,7 +207,7 @@ impl CommittedState {
     pub fn index_seek<'a>(
         &'a self,
         table_id: &TableId,
-        cols: NonEmpty<ColId>,
+        cols: &NonEmpty<ColId>,
         range: &impl RangeBounds<AlgebraicValue>,
     ) -> Option<BTreeIndexRangeIter<'a>> {
         if let Some(table) = self.tables.get(table_id) {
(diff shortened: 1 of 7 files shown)


crates/core/src/db/datastore/locking_tx_datastore/mod.rs
```diff
@@ -1502,31 +1502,26 @@ impl Inner {
...
-                .committed_state
-                .index_seek(table_id, NonEmpty::new(*col_id), &range)
-            {
+            match self.committed_state.index_seek(table_id, &cols, &range) {
                 //If we don't have `self.tx_state` yet is likely we are running the bootstrap process
                 Some(committed_rows) => match self.tx_state.as_ref() {
                     None => Ok(IterByColRange::Scan(ScanIterByColRange {
                         range,
-                        col_id: *col_id,
+                        cols: NonEmpty::collect(cols.into_iter().map(|col| col.0)).unwrap(),
```
kim, line 1524: I think you can just map directly over cols

kim (approved):
:+1:

PR #417, opened 2023-10-11
Simplify/deabstract module_host_actor and wasmer_module

coolreader18:
# Description of Changes

# API and ABI

### DIFF
--- a/crates/core/src/db/datastore/traits.rs
+++ b/crates/core/src/db/datastore/traits.rs
@@ -1,12 +1,13 @@
 use crate::db::relational_db::ST_TABLES_ID;
 use crate::execution_context::ExecutionContext;
+use anyhow::Context;
 use nonempty::NonEmpty;
 use spacetimedb_lib::auth::{StAccess, StTableType};
 use spacetimedb_lib::relation::{DbTable, FieldName, FieldOnly, Header, TableField};
 use spacetimedb_lib::{ColumnIndexAttribute, DataKey, Hash};
 use spacetimedb_primitives::{ColId, IndexId, SequenceId, TableId};
 use spacetimedb_sats::product_value::InvalidFieldError;
-use spacetimedb_sats::{AlgebraicType, AlgebraicValue, ProductType, ProductTypeElement, ProductValue};
+use spacetimedb_sats::{AlgebraicType, AlgebraicValue, ProductType, ProductTypeElement, ProductValue, WithTypespace};
 use spacetimedb_vm::expr::SourceExpr;
 use std::{borrow::Cow, ops::RangeBounds, sync::Arc};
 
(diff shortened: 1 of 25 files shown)


kim:
This will conflict with #267

kulakowski:
Not to jump the gun, but I took a peek and think this is going to be helpful to people clicking around in and reading this code.

kim:
Not saying this is bad, but it will make @mamcx’s life harder to get #267 over the line. Also as a heads up: with the changes in #267 much of the init/update code becomes obsolete (for one because we can no longer update index changes). Please let me know when this becomes ready, I think I’m mostly the one familiar with that code.

coolreader18:
More than happy to wait on #267 to be merged first.

coolreader18:
What's the status with #267 re:being split into multiple parts? I can't tell whether the chunks that are relevant to wasmer_module are in master already or not.

coolreader18:
benchmarks please

coolreader18:
lol. well

coolreader18:
I guess that confirms that works @kurtismullins

coolreader18:
benchmarks please

coolreader18:
benchmarks please

mamcx:
> What's the status with #267 re:being split into multiple parts? I can't tell whether the chunks that are relevant to wasmer_module are in master already or not.

Is still waiting for approving, and it will take some time because was split in several other PRs...

kim:
Alright, then let’s merge this?

kim:
@coolreader18 Maybe you could add this from #386: https://github.com/clockworklabs/SpacetimeDB/pull/386/files#diff-840af797a96dc965f1b75269b1c343356ecb6a0b4e28b84c611a0db514e9d3faR113-R127 ?

I think it is still better meanwhile to create multi-column indexes than to ignore them.

coolreader18:
Maybe as a separate PR based on this one :grimacing: ? unless I missed that when I was rebasing and it's in master already. Just cause it's kinda its own feature rather than a refactor like this.

coolreader18:
benchmarks please

coolreader18:
cool it works :))))

crates/core/src/db/relational_db.rs
```diff
@@ -370,11 +369,7 @@ impl RelationalDB {
 
     /// Roll back transaction `tx` if `res` is `Err`, otherwise return it
     /// alongside the `Ok` value.
-    #[tracing::instrument(skip_all)]
```
kulakowski, line 373: I'd keep the `instrument` bit.

crates/core/src/db/datastore/traits.rs
```diff
@@ -292,6 +293,73 @@ pub struct TableDef {
     pub(crate) table_access: StAccess,
 }
 
+impl TableDef {
```
kulakowski, line 296: I don't love this code being here. But I don't think there's a great place to put it when we are in this state of defining a pile of concrete types, and traits (that aren't really used, relationaldb directly makes a Locking or whatever). Curious how you feel about it.

crates/core/src/db/update.rs
```diff
@@ -0,0 +1,176 @@
...
+
+#[derive(thiserror::Error, Debug)]
+pub enum UpdateDatabaseError {
+    #[error("incompatible schema changes for: {tables:?}")]
+    IncompatibleSchema { tables: Vec<String> },
+    #[error(transparent)]
+    Database(#[from] DBError),
+}
+
+pub fn update_database(
```
kulakowski, line 23: I didn't scrutinize this code, am trusting that it's functionally unchanged.

kulakowski (commented):
Ok I think this is good and ready. It does a bit of unrelated cleanups but I like them all.

I'm not hitting approve just to make sure that we are all squared up on which PR lands before whichever other PR, but I'll hit it as as soon as someone acknowledges that the plan is ready.

crates/core/src/db/relational_db.rs
kulakowski: This is the only thing to fix before landing, then, I think.

kulakowski (approved):

crates/core/src/db/relational_db.rs
coolreader18: IMO it's really unnecessary - it's a utility function that we have no need to track and only serves to clutter up logs/graphs. If you feel strongly about it I can add it back, but I really see no value in it.

crates/core/src/db/datastore/traits.rs
coolreader18: Yeah, I think there's better ways to go about this, but ideally that would come with a general reorganizing of all these types.

coolreader18:
Merged. @joshua-spacetime if you'd like help fixing #486 to work with this I'd be happy to help

crates/core/src/db/datastore/traits.rs
cloutiertyler: I also don't love this code being here FWIW

crates/core/src/db/relational_db.rs
kulakowski: It's not that I do or don't see the value in it. It's that it is unrelated to the stated point of the PR.

PR #420, opened 2023-10-12
[ABI] Remove the special first element of iterator

RReverser:
# Description of Changes

I'm not sure why we're sending encoded schema as this special first element of iteration.

There is some commented out code that suggests it might have been used in the past, but at least nowadays modules (both Rust and C#) are strongly typed and already know the schema of the tables they asked to iterate over, and ignore the first element returned from the iteration.

As a result, we've been unnecessarily reading schema, encoding it into binary, taking care to allocate it as individual small blob, stored in list of buffers and sent it between Wasm and Rust just for it to be ignored.

Removing it simplifies code and if we ever want to get table schema, we can do that via a dedicated method rather than send it automatically on beginning of each iteration.

This is an ABI breaking change, but likely worth it.

# API and ABI

### DIFF
--- a/crates/bindings-csharp/Runtime/Runtime.cs
+++ b/crates/bindings-csharp/Runtime/Runtime.cs
@@ -140,20 +140,18 @@ public static class Runtime
 
     public class RawTableIter : IEnumerable<byte[]>
     {
-        public readonly byte[] Schema;
-
-        private readonly IEnumerator<byte[]> iter;
+        private readonly uint tableId;
+        private readonly byte[]? filterBytes;
 
         public RawTableIter(uint tableId, byte[]? filterBytes = null)
         {
-            iter = new BufferIter(tableId, filterBytes);
-            iter.MoveNext();
-            Schema = iter.Current;
+            this.tableId = tableId;
+            this.filterBytes = filterBytes;
         }
 
         public IEnumerator<byte[]> GetEnumerator()
         {
-            return iter;
+            return new BufferIter(tableId, filterBytes);
         }
 
         IEnumerator IEnumerable.GetEnumerator()
--- a/crates/bindings-csharp/Runtime/bindings.c
+++ b/crates/bindings-csharp/Runtime/bindings.c
@@ -100,7 +100,7 @@ static MonoArray* stdb_buffer_consume(Buffer buf);
 //   return out;
 // }
 
-#define STDB_IMPORT_MODULE_MINOR(minor) "spacetime_6." #minor
+#define STDB_IMPORT_MODULE_MINOR(minor) "spacetime_7." #minor
 #define STDB_IMPORT_MODULE STDB_IMPORT_MODULE_MINOR(0)
 
 __attribute__((import_module(STDB_IMPORT_MODULE),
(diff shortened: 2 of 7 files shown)


RReverser:
@coolreader18 I rebased and this broke due to ABI check. I think I updated all the places we have ABI defined after the recent overhaul, but maybe not? Can you point to which one I missed?

kulakowski:
> # Description of Changes
> 
> I'm not sure why we're sending encoded schema as this special first element of iteration.
> 
> There is some commented out code that suggests it might have been used in the past, but at least nowadays modules (both Rust and C#) are strongly typed and already know the schema of the tables they asked to iterate over, and ignore the first element returned from the iteration.
> 
> As a result, we've been unnecessarily reading schema, encoding it into binary, taking care to allocate it as individual small blob, stored in list of buffers and sent it between Wasm and Rust just for it to be ignored.
> 
> Removing it simplifies code and if we ever want to get table schema, we can do that via a dedicated method rather than send it automatically on beginning of each iteration.

Yes, we used to expose out of the Rust bindings a non-strongly-typed iteration. I don't know if it was ever actually used, but I think for Chesteron's fence type reasons, someone should speak to why it was ok to remove but to leave the code commented out in the bindings: https://github.com/clockworklabs/SpacetimeDB/blob/master/crates/bindings/src/lib.rs#L272 and so on (and all the commented out code which used that, like the delete_filter etc functions).

RReverser:
Rebased & fixed conflicts.

coolreader18:
IIRC it was like this because `fn __iter__` took an arbitrary table id without any type information, and it was designed to return an iterator over `AlgebraicValue` without needing any generics or `TableType` or anything. (It also had FromSats and ToSats that took/returned an AlgebraicValue directly, rather than the Serialize/Deserialize traits we have now, if that makes that design make more sense.) When I made the changes away, I considered removing the now-vestigial row type bit, but didn't because I was unsure if maybe there was a vision for allowing more dynamic db operations, but clearly that's not come out to much. Feel free to remove it, it's totally vestigial at this point.

cloutiertyler:
I just synced with Noa as well. TL;DR this is approx how it used to work in psuedo code:

```rust
let algebraic_type = AlgebraicType::decode_from_bytes(algebraic_type_bytes);
let algebraic_value = AlgebraicValue::decode_from_bytes(bytes, algebraic_type);
let my_table: MyTable = MyTable::from_algebraic_value(algebraic_value);
```

So we actually used to use both the static and dynamic type information. @coolreader18 switched it to be:

```rust
let my_table: MyTable = MyTable::from_bytes(bytes);
```

but left the dynamically send AlgebraicType bytes in there in case it was used eventually. Ultimately if we were to have a dynamic API though, you probably wouldn't want it to be mandatory with every iter call, you'd want to have a separate API/ABI sys call for getting the `AlgebraicType` of a table. So this is good to be merged and Chesterson can rest easy.

RReverser:
> So this is good to be merged and Chesterson can rest easy.

Glad to hear :) Can one of you set an approval tick please?

coolreader18 (approved):

PR #422, opened 2023-10-12
Simplify and document WasmInstanceEnv's interface

kulakowski:
This abstracts out all the internals of WasmInstanceEnv, and provides a slightly higher level API. In particular, we expect to potentially add more state associated to a single reducer invocation, and it will be nice to interact with that in general rather than per type of state.

This also documents WasmInstanceEnv and its members.

# Description of Changes

# API and ABI

1: This has no functional change, and adds a number of comments.

### DIFF
--- a/crates/core/src/host/wasmer/wasm_instance_env.rs
+++ b/crates/core/src/host/wasmer/wasm_instance_env.rs
@@ -86,7 +86,8 @@ impl WasmInstanceEnv {
         self.mem.clone().expect("Initialized memory")
     }
 
-    /// Return a reference to the `InstanceEnv`.
+    /// Return a reference to the `InstanceEnv`,
+    /// which is responsible for DB instance and associated state.
     pub fn instance_env(&self) -> &InstanceEnv {
         &self.instance_env
     }
--- a/crates/core/src/host/wasmer/wasm_instance_env.rs
+++ b/crates/core/src/host/wasmer/wasm_instance_env.rs
@@ -15,12 +15,39 @@ use crate::host::instance_env::InstanceEnv;
 
 use super::{Mem, WasmError};
 
+/// A `WasmInstanceEnv` provides the connection between a module
+/// and the database.
+///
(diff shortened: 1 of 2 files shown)


kulakowski:
Happy for any thoughts about naming, and about the doc comments. I certainly sometimes struggle with doc comments when they feel a bit like "the FooAndBarManager manages a Foo and a Bar".

crates/core/src/host/wasmer/wasm_instance_env.rs
```diff
@@ -15,12 +15,39 @@ use crate::host::instance_env::InstanceEnv;
 
 use super::{Mem, WasmError};
 
+/// A `WasmInstanceEnv` provides the connection between a module
```
cloutiertyler, line 18: 🙏

crates/core/src/host/wasmer/wasm_instance_env.rs
```diff
@@ -38,11 +65,50 @@ fn mem_err(err: MemoryAccessError) -> RuntimeError {
...
+
+    /// Take ownership of the given `data` and give back a `BufferIdx`
+    /// as a handle to that data.
+    pub fn insert_buffer(&mut self, data: bytes::Bytes) -> BufferIdx {
+        self.buffers.insert(data)
+    }
+
+    /// Reset all of the state associated to a single reducer call.
+    pub fn clear_reducer_state(&mut self) {
+        // For the moment, we only explicitly clear the set of buffers.
```
gefjon, line 108:
```suggestion
        // For the moment, we only explicitly clear the set of buffers.
        // TODO: should we be clearing `iters` and/or `timing_spans`?
```

gefjon (approved):
Excellent. Minor proposed changes.

Out-of-scope thought: I wonder if we should consider renaming `InstanceEnv` -> `DbInstanceEnv`, to more clearly distinguish it from `WasmInstanceEnv`.

PR #423, opened 2023-10-12
Optimise binaries with wasm-opt

RReverser:
# Description of Changes

Now that our ABI checks can handle binaries produced by wasm-opt, we can use it (if installed) to optimise produced Wasm binaries.

For now it's added as an optional step until we figure out a decent deployment strategy to automatically get it for the user. (There are some wrappers on crates.io, but they tend to lag behind modern versions and take a while to compile, whereas Binaryen provides official prebuilt binaries on Github.)

This gives an easy 8-10% performance boost in iteration benchmarks, much more for an empty table but that's less interesting.

# API and ABI

1

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -72,7 +72,7 @@ jobs:
 
       - name: Install latest wasm-opt for module optimisations
         run: |
-          curl https://github.com/WebAssembly/binaryen/releases/download/version_116/binaryen-version_116-x86_64-linux.tar.gz -L | tar xz -C /usr/local --strip-components=1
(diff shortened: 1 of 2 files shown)


kulakowski:
I'm really curious what optimizations are affecting our modules most.

RReverser:
> I'm really curious what optimizations are affecting our modules most.

That would need to be much deeper investigation and likely out of scope of this PR.

gefjon (approved):
Looks good. I appreciate not depending on `wasm-opt`, but encouraging users to install it. Long-term I wonder if we can link against binaryen rather than invoking it as a subprocess, but this is good for now.

PR #424, opened 2023-10-12
🐮 Optimize `decode_column` and friends using `Cow` 🐮

Centril:
# Description of Changes

The methods `schema_for_column`, `row_schema_for_table`, and `schema_for_table` are optimized by returning `Cow<'_, T>`.

# API and ABI

1

### DIFF
--- a/crates/core/src/host/wasm_common/module_host_actor.rs
+++ b/crates/core/src/host/wasm_common/module_host_actor.rs
@@ -275,12 +275,12 @@ impl<T: WasmModule> Module for WasmModuleHostActor<T> {
         let db = &*self.worker_database_instance.relational_db;
         db.with_auto_commit(|tx| {
             let tables = db.get_all_tables(tx)?;
-            let clear_ids = tables
+            // We currently have unique table names,
+            // so we can assume there's only one table to clear.
+            if let Some(table_id) = tables
                 .iter()
-                .filter(|t| t.table_name == table_name)
-                .map(|t| t.table_id)
-                .collect::<Vec<_>>();
-            for table_id in clear_ids {
+                .find_map(|t| (t.table_name == table_name).then_some(t.table_id))
+            {
                 db.clear_table(tx, table_id)?;
             }
             Ok(())


crates/core/src/db/datastore/locking_tx_datastore/mod.rs
```diff
@@ -1635,7 +1636,7 @@ impl Locking {
         for write in &transaction.writes {
             let table_id = TableId(write.set_id);
             let schema = inner.schema_for_table(table_id)?;
-            let row_type = inner.row_type_for_table(table_id)?;
+            let row_type = inner.row_type_for_table(table_id)?.into_owned();
             let table = inner.committed_state.tables.entry(table_id).or_insert(Table {
                 row_type: row_type.clone(),
```
RReverser, line 1641: It seems if we move `table` into the `match`, we could get rid of this clone as well. It's only used in `insert` operation and in a way where `table` can be constructed after it's used.

crates/core/src/db/datastore/traits.rs
```diff
@@ -459,7 +459,11 @@ pub trait TxDatastore: DataRow + Tx {
 pub trait MutTxDatastore: TxDatastore + MutTx {
     // Tables
     fn create_table_mut_tx(&self, tx: &mut Self::MutTxId, schema: TableDef) -> Result<TableId>;
-    fn row_type_for_table_mut_tx(&self, tx: &Self::MutTxId, table_id: TableId) -> Result<ProductType>;
+    fn row_type_for_table_mut_tx<'tx>(
```
kulakowski, line 462: I think a comment describing the lifetime is justified, why it's that of the transaction and not self.

Why the change from `Result` to `super::Result`?

Also, not exclusively about this PR, but I think this is some evidence about the lack of utility of this trait. Low level representational details leak through for performance reasons. It's not that plausible to me that a different datastore would want to make the same set of choices.

crates/core/src/db/relational_db.rs
```diff
@@ -214,13 +215,28 @@ impl RelationalDB {
...
+    ) -> Result<Cow<'tx, AlgebraicType>, DBError> {
+        Ok(match self.row_schema_for_table(tx, table_id)? {
+            Cow::Borrowed(schema) => Cow::Borrowed(
+                &schema
+                    .elements
+                    .get(col_id as usize)
+                    .ok_or(TableError::ColumnNotFound(col_id))?
+                    .algebraic_type,
+            ),
+            Cow::Owned(mut schema) => {
```
kulakowski, line 232: Mostly echoing what I have perceived to be Tyler's feedback around this part of the codebase: I would prefer comments aimed at a person who is not expert at Rust. Right here, I would take that to mean someone who is unfamiliar with the `Cow` interface, for instance. It's unclear why this path requires a manual bounds check!

crates/core/src/db/datastore/locking_tx_datastore/mod.rs
```diff
@@ -564,7 +565,7 @@ impl Inner {
             let table_row = StTableRow::try_from(&row)?;
             let table_id = TableId(table_row.table_id);
             let schema = self.schema_for_table(table_id)?;
-            let row_type = self.row_type_for_table(table_id)?;
+            let row_type = self.row_type_for_table(table_id)?.into_owned();
```
kulakowski, line 568: I'd echo Ingvar's comments. I think eventually we'll converge on the right patterns uniformally throughout this file, though. I don't think we need to do all of it at once atomically.

kulakowski (commented):
Part of the reason this code is subtle is because of the transactional semantics we want to apply to migrations and schema changes. I don't see any particular reason to worry about this change with respect to that, but think an acknowledgement that this is still correct in the case that the schema changes. cf the comment in `locking_tx_datastore/mod.rs:796`

kulakowski:
I hate to be no fun but just 🐄 in the list of PRs is pretty hard to understand.

crates/core/src/host/wasm_common/module_host_actor.rs
```diff
@@ -275,12 +275,13 @@ impl<T: WasmModule> Module for WasmModuleHostActor<T> {
...
-                if table.table_name != table_name {
-                    continue;
-                }
-
-                db.clear_table(tx, table.table_id)?;
+            let clear_ids = tables
+                .iter()
+                .filter(|t| t.table_name == table_name)
+                .map(|t| t.table_id)
+                .collect::<Vec<_>>();
```
kulakowski, line 282: Is there a stateful reason to collect these? ie, if you called `clear_table` on the first one, it would in some way invalidate things later on?

kulakowski (approved):

PR #425, opened 2023-10-12
Incremental Evaluation of JOIN queries

kim:
# Description of Changes

WIP correctness of incremental join query evaluation.

@joshua-spacetime Feel free to add changes 

# API and ABI

### DIFF
--- a/crates/core/src/subscription/query.rs
+++ b/crates/core/src/subscription/query.rs
@@ -205,6 +205,42 @@ mod tests {
         Ok(db.create_table(tx, schema)?)
     }
 
+    fn insert_op(table_id: u32, table_name: &str, row: ProductValue) -> DatabaseTableUpdate {
+        let row_pk = row.to_data_key().to_bytes();
+        DatabaseTableUpdate {
+            table_id,
+            table_name: table_name.to_string(),
+            ops: vec![TableOp {
+                op_type: 1,
+                row,
+                row_pk,
+            }],
+        }
+    }
+
+    fn delete_op(table_id: u32, table_name: &str, row: ProductValue) -> DatabaseTableUpdate {
+        let row_pk = row.to_data_key().to_bytes();
+        DatabaseTableUpdate {
+            table_id,
+            table_name: table_name.to_string(),
+            ops: vec![TableOp {
+                op_type: 0,
+                row,
+                row_pk,
+            }],
+        }
+    }
+
+    fn insert_row(db: &RelationalDB, tx: &mut MutTxId, table_id: u32, row: ProductValue) -> ResultTest<()> {
+        db.insert(tx, table_id, row)?;
+        Ok(())
+    }
+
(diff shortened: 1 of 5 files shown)


crates/core/src/subscription/query.rs
```diff
@@ -137,12 +119,76 @@ pub fn compile_read_only_query(
...
     }
 }
 
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum Supported {
+    Scan,
+    Semijoin,
+}
+
+pub fn classify(expr: &QueryExpr) -> Option<Supported> {
```
kim, line 134: This is probably not very robust. I expected to look for IndexJoin, being exactly the kind of join we support. But the examples from the test always yield JoinInner.

crates/core/src/subscription/query.rs
```diff
@@ -679,44 +710,7 @@ mod tests {
...
-    //```
-    //SELECT * FROM table1
-    //SELECT * FROM table2
-    // =
-    //SELECT * FROM table2
-    //SELECT * FROM table1
-    //```
-    // return just one row irrespective of the order of the queries
-    #[test]
-    fn test_subscribe_commutative() -> ResultTest<()> {
```
kim, line 697: It’s a set now, so always this test doesn’t make sense anymore.

crates/core/src/subscription/subscription.rs
```diff
@@ -93,48 +118,38 @@ impl QuerySet {
...
-                        let pos_op_type = result.head.find_pos_by_name(OP_TYPE_FIELD_NAME).unwrap_or_else(|| {
-                            panic!(
-                                "failed to locate `{OP_TYPE_FIELD_NAME}` on `{}`. fields: {:?}",
-                                result.head.table_name,
-                                result.head.fields.iter().map(|x| &x.field).collect::<Vec<_>>()
-                            )
+        for expr in self {
+            use query::Supported::*;
+            match query::classify(expr) {
+                None => log::warn!("invalid query expression for incremental evaluation"),
```
kim, line 124: It is probably preferable to classify the queries when subscribing, and return an error if there are unsupported ones. Threading the tag through would make the diff much larger though, so I punted on that for now.

crates/core/src/subscription/subscription.rs
```diff
@@ -160,41 +175,301 @@ impl QuerySet {
...
+            if let Some(t) = q.source.get_db_table() {
+                for table in run_query(relational_db, tx, q, auth)? {
+                    {
+                        let mut table_row_operations = Vec::new();
+
+                        for row in table.data {
+                            let row_pk = pk_for_row(&row);
+
+                            //Skip rows that are already resolved in a previous subscription...
+                            if seen.contains(&(t.table_id, row_pk)) {
```
kim, line 188: Ah I think this could also just check the return value of insert, so we don’t copy the pk twice.

PR #431, opened 2023-10-16
Fix authorization check for CrudCode

kim:
# Description of Changes

Make all mutating SQL operations require the caller to be the database owner.

# API and ABI

1


cloutiertyler (approved):
LGTM

PR #434, opened 2023-10-16
`DataRef`: borrow for great perf

Centril:
# Description of Changes

Get rid of the old `DataRef` so that it instead borrows the `DataKey` and `ProductValue`.
Then this is used throughout so that `iter`, `iter_by_col_eq`, `table_name_from_id`, and friends, do not `.clone()` and instead just returns or works off of borrowed data.

# API and ABI

2

### DIFF
--- a/crates/core/src/db/datastore/locking_tx_datastore/mod.rs
+++ b/crates/core/src/db/datastore/locking_tx_datastore/mod.rs
@@ -353,7 +353,6 @@ impl Inner {
 
     fn bootstrap_system_table(&mut self, schema: TableSchema) -> Result<(), DBError> {
         let table_id = schema.table_id;
-        let table_name = &schema.table_name;
 
         // Insert the table row into st_tables, creating st_tables if it's missing
         let st_tables = self
@@ -361,24 +360,27 @@ impl Inner {
             .get_or_create_table(ST_TABLES_ID, &ST_TABLE_ROW_TYPE, &st_table_schema());
         let row = StTableRow {
             table_id,
-            table_name,
+            table_name: schema.table_name,
             table_type: StTableType::System,
             table_access: StAccess::Public,
         };
-        let row: ProductValue = (&row).into();
+        let row: ProductValue = row.into();
         let data_key = row.to_data_key();
         st_tables.rows.insert(RowId(data_key), row);
 
         // Insert the columns into st_columns
-        for (i, col) in schema.columns.iter().enumerate() {
+        let first_col_id = schema.columns.first().unwrap().col_id;
+        for (i, col) in schema.columns.into_iter().enumerate() {
+            let col_name_for_autoinc = col.is_autoinc.then(|| col.col_name.clone());
+
             let row = StColumnRow {
                 table_id,
                 col_id: i.into(),
-                col_name: &col.col_name,
-                col_type: col.col_type.clone(),
+                col_name: col.col_name,
+                col_type: col.col_type,
                 is_autoinc: col.is_autoinc,
             };
-            let row = ProductValue::from(&row);
+            let row = ProductValue::from(row);
             let data_key = row.to_data_key();
             {
                 let st_columns =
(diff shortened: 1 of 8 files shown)


crates/core/src/db/datastore/locking_tx_datastore/mod.rs
```diff
@@ -669,15 +652,16 @@ impl Inner {
...
+        // TODO(centril): `result` is already owned, so pass that it.
+        let sequence_row = StSequenceRow::try_from(&result)?.to_owned();
         let sequence_id = SequenceId(sequence_row.sequence_id);
 
-        let schema = (&sequence_row).into();
+        let schema = sequence_row.into();
         self.sequence_state.sequences.insert(sequence_id, Sequence::new(schema));
 
-        log::trace!("SEQUENCE CREATED: {}", seq.sequence_name);
+        log::trace!("SEQUENCE CREATED: id = {}", sequence_id);
```
kulakowski, line 664: Why did this change from giving a name to an ID?

crates/core/src/db/datastore/locking_tx_datastore/mod.rs
```diff
@@ -2097,12 +2082,12 @@ impl MutTxDatastore for Locking {
...
 
     fn get_mut_tx<'a>(
-        &'a self,
+        &self,
         tx: &'a Self::MutTxId,
         table_id: TableId,
-        row_id: Self::RowId,
-    ) -> super::Result<Option<Self::DataRef>> {
-        tx.lock.get(&table_id, &row_id)
+        row_id: &'a Self::RowId,
```
kulakowski, line 2088: This sort of change in lifetimes here is really the heart of this change, I think. In a mem-arch'd future, is this the one you want?

gefjon:
> Would you be able to upload a before and after flamegraph to show the improvements?

I have these flamegraphs, but I don't have a good way to upload them - the `perf` flamegraph script generates HTML, which GitHub doesn't allow as an attachment.

crates/core/src/db/datastore/locking_tx_datastore/mod.rs
```diff
@@ -81,45 +80,32 @@ pub enum SequenceError {
...
 /// A `DataRef` represents a row stored in a table.
 ///
 /// A table row always has a [`DataKey`] associated with it.
 /// This is in contrast to rows that are materialized during query execution
 /// which may or may not have an associated `DataKey`.
-#[derive(Clone)]
-pub struct DataRef {
-    id: DataKey,
-    data: ProductValue,
+#[derive(Copy, Clone)]
```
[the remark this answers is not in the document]
kulakowski: Right, but I said Clone not Copy.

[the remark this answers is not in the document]
kulakowski: Yeah I realized it was a brain fart as soon as I typed it, sorry.

crates/core/src/db/datastore/locking_tx_datastore/mod.rs
kulakowski: I also just now clicked the show-me-a-few-more-lines-please github button and saw that the corresponding opening log has the id and name, so I don't think this is a big regression. 👍

crates/core/src/db/datastore/locking_tx_datastore/mod.rs
kulakowski: Yeah, this comment wasn't intended to be blocking at all and I certainly think iterating fast is the way to go now.

kulakowski (approved):

kulakowski (approved):

crates/core/src/host/instance_env.rs
```diff
@@ -145,24 +146,22 @@ impl InstanceEnv {
...
 
         // Delete them and count how many we deleted and error if none.
-        let count = stdb
-            .delete_by_rel(tx, table_id, seek)
-            .inspect_err_(|e| log::error!("delete_by_col_eq(table_id: {table_id}): {e}"))?
-            .ok_or(NodesError::ColumnValueNotFound)?;
-
-        Ok(count)
+        let count = stdb.delete(tx, table_id, rows_to_delete);
+        NonZeroU32::new(count).ok_or(NodesError::ColumnValueNotFound)
```
kulakowski, line 164: I believe this introduced the new log lines we are discussing.

PR #435, opened 2023-10-16
Reuse Wasm-side buffer while reading iterators

RReverser:
# Description of Changes

Reuse buffer when reading new chunks from the iterator, instead of freeing and allocating a new one each time.

This will be mainly useful on large iterations that span over our 64KiB chunks.

# API and ABI

### DIFF
--- a/crates/bindings/src/lib.rs
+++ b/crates/bindings/src/lib.rs
@@ -345,10 +345,8 @@ struct RawTableIter<De> {
     /// The underlying source of our `Buffer`s.
     inner: BufferIter,
 
-    /// The current position in the current buffer,
-    /// from which `deserializer` can read.
-    /// A value of `None` indicates that we need to pull another `Buffer` from `inner`.
-    reader: Option<Cursor<Box<[u8]>>>,
+    /// The current position in the buffer, from which `deserializer` can read.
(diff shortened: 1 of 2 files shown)


RReverser:
Woops, gonna fix in a sec.

kulakowski:
Conceptually makes sense. This doesn't change the buffer sharing protocol for all the iteration stuff, correct, so it's not an ABI break? The wasm side was and is responsible.

RReverser:
> This doesn't change the buffer sharing protocol for all the iteration stuff, correct, so it's not an ABI break? The wasm side was and is responsible.

Correct, this is purely Wasm-side optimisation with no ABI change.

crates/bindings/src/lib.rs
```diff
@@ -263,7 +263,11 @@ fn buffer_table_iter(
     let mut iter = sys::iter(table_id, filter.as_deref())?;
 
     // First item is an encoded schema.
-    let schema_raw = iter.next().expect("Missing schema").expect("Failed to get schema");
+    let schema_raw = iter
```
RReverser, line 266: FWIW this will go away if/when #420 is merged, so will also become simpler.

kulakowski (approved):

PR #436, opened 2023-10-17
Optimize `drop_table`, `rename_table`, `delete_by_col_eq`, and `clear_table`

Centril:
# Description of Changes

These functions and methods now clone less and collect keys/ids to rename/drop/delete rather than storing a cloned `ProductValue` that is then used to do the work.

# API and ABI

2

### DIFF
--- a/crates/core/src/db/relational_db.rs
+++ b/crates/core/src/db/relational_db.rs
@@ -20,7 +20,7 @@ use fs2::FileExt;
 use nonempty::NonEmpty;
 use prometheus::HistogramVec;
 use spacetimedb_lib::ColumnIndexAttribute;
-use spacetimedb_lib::{data_key::ToDataKey, DataKey, PrimaryKey};
+use spacetimedb_lib::{data_key::ToDataKey, PrimaryKey};
 use spacetimedb_sats::{AlgebraicType, AlgebraicValue, ProductType, ProductValue};
 use std::borrow::Cow;
 use std::fs::{create_dir_all, File};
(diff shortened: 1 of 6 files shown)


crates/core/src/host/instance_env.rs
```diff
@@ -144,23 +145,24 @@ impl InstanceEnv {
     ///
     /// Returns an error if no columns were deleted or if the column wasn't found.
     #[tracing::instrument(skip(self, value))]
-    pub fn delete_by_col_eq(&self, table_id: u32, col_id: u32, value: &[u8]) -> Result<u32, NodesError> {
+    pub fn delete_by_col_eq(&self, table_id: u32, col_id: u32, value: &[u8]) -> Result<NonZeroU32, NodesError> {
```
kulakowski, line 148: I don't understand what removing clones has to do with the type of this function.

crates/core/src/host/instance_env.rs
```diff
@@ -144,23 +145,24 @@ impl InstanceEnv {
...
+            .map(|x| *x.id())
+            .collect::<Vec<_>>();
 
         // Delete them and count how many we deleted and error if none.
         let count = stdb
-            .delete_by_rel(tx, table_id, seek)
-            .inspect_err_(|e| log::error!("delete_by_col_eq(table_id: {table_id}): {e}"))?
-            .ok_or(NodesError::ColumnValueNotFound)?;
-
+            .delete(tx, table_id, rows_to_delete)
```
kulakowski, line 163: This seems like it will now make a call into LockingTxDatastore per row to delete. And therefore in particular take the lock over and over. Was that intentional?

crates/core/src/db/relational_db.rs
```diff
@@ -545,12 +543,21 @@ impl RelationalDB {
...
         tx: &mut MutTxId,
         table_id: u32,
-        relation: R,
-    ) -> Result<Option<u32>, DBError> {
+        relation: impl IntoIterator<Item = DataKey>,
+    ) -> Result<u32, DBError> {
+        let table_id = TableId(table_id);
+        relation
+            .into_iter()
+            .map(|dk| self.inner.delete_mut_tx(tx, table_id, RowId(dk)).map(|b| b as u32))
```
kulakowski, line 555: Here's where I see the multiple calls into the datastore.

crates/core/src/host/instance_env.rs
kulakowski: I'm fine with this change for now. I'm surprised at these semantics, though, but I agree that the code and the docs all agreeing right now is an improvement.

kulakowski (approved):

PR #243, opened 2023-08-25
Adding multi-column indexes to the DB

mamcx:
# Description of Changes

Note: This is a multi-step set of PRs to bring constraints + multi-column indexes

Step:
- [x] [Add system tables](https://github.com/clockworklabs/SpacetimeDB/pull/225)
- [x] Add multi-column indexes
- [ ] [Rework schema definition](https://github.com/clockworklabs/SpacetimeDB/pull/267)
- [ ] [Rework bootstrapping](https://github.com/clockworklabs/SpacetimeDB/pull/267)

IMPORTANT:

The use of `multi-column` indexes changes what is stored in the `index` according to how many fields are defined:

- 1 Field: Single `AlgebraicValue`
- 2 or more: `AlgebraicValue::Product`

*If the API is breaking, please state below what will break*

 - [ ] This is a breaking change to the module ABI
 - [ ] This is a breaking change to the module API
 - [ ] This is a breaking change to the ClientAPI
 - [ ] This is a breaking change to the SDK API
 - [x] This is a breaking change to the database system tables

PR #439, opened 2023-10-17
Refactoring IDs for system objects

mamcx:
# Description of Changes

Note: This is a multi-step set of PRs to bring constraints + multi-column indexes

Step:
- [x] [Add system tables](https://github.com/clockworklabs/SpacetimeDB/pull/225)
- [x] [Add multi-column indexes](https://github.com/clockworklabs/SpacetimeDB/pull/243)
- [x] [Rework schema definition](https://github.com/clockworklabs/SpacetimeDB/pull/267)
- [ ] Refactoring IDs for system objects <-- THIS PR

This PR is to simplify the review of https://github.com/clockworklabs/SpacetimeDB/pull/267. It only have the movement of schema objects and the use of `newtypes` for system objects.

# API and ABI

 - [x] This is a breaking change to the module API
 - [x] This is a breaking change to the ClientAPI
 - [x] This is a breaking change to the SDK API


Centril (commented):
Some more ideas on how to further split this PR.

crates/sats/src/db/mod.rs
```diff
@@ -0,0 +1,29 @@
...
+}
+
+impl<'de> de::Deserialize<'de> for ColumnAttribute {
+    fn deserialize<D: de::Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        Self::from_bits(deserializer.deserialize_u8()?)
+            .ok_or_else(|| de::Error::custom("invalid bitflags for ColumnIndexAttribute"))
+    }
+}
+
+impl ser::Serialize for ColumnAttribute {
```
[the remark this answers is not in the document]
mamcx: I can't see how, it demands the use of generics.

crates/core/src/vm.rs
```diff
@@ -37,10 +39,14 @@ pub fn build_query<'a>(
         result = match op {
             Query::IndexScan(IndexScan {
                 table,
-                col_id,
+                columns,
                 lower_bound,
                 upper_bound,
-            }) if db_table => iter_by_col_range(stdb, tx, table, col_id, (lower_bound, upper_bound))?,
+            }) if db_table => {
+                assert_eq!(columns.len(), 1, "Only support single column IndexScan");
```
kulakowski, line 46: Apologies if this has been covered before by Mazdak: are we ok with panicking in places like this that we don't yet support, instead of returning an Err?

crates/core/src/vm.rs
```diff
@@ -373,6 +380,9 @@ impl<'db, 'tx> DbProgram<'db, 'tx> {
                     self.db.drop_sequence(self.tx, id)?;
                 }
             }
+            DbType::Constraint => {
+                todo!()
```
kulakowski, line 384: A comment at least please!

crates/bindings-macro/src/lib.rs
```diff
@@ -24,6 +24,9 @@ use syn::{
     parse_quote, BinOp, Expr, ExprBinary, ExprLit, ExprUnary, FnArg, Ident, ItemFn, ItemStruct, Member, Token, Type,
     UnOp,
 };
+#[path = "../../sats/src/db/attr.rs"]
```
kulakowski, line 27: This is a pretty rare feature to need, so I think a comment about it please.

crates/vm/src/expr.rs
```diff
@@ -713,44 +728,44 @@ impl QueryExpr {
...
+                    rhs: rhs.with_index_lower_bound(table, columns, value, inclusive),
                     col_lhs,
                     col_rhs,
                 }));
                 self
             }
             // merge with a preceding upper bounded index scan (inclusive)
             Query::IndexScan(IndexScan {
-                col_id: lhs_col_id,
+                columns: lhs_col_id,
```
kulakowski, line 746: I'm guessing that future work will extend index scans to multiple columns?

Centril (changes requested):

Centril (approved):

Cargo.lock
```diff
@@ -1585,6 +1585,12 @@ dependencies = [
  "winapi",
 ]
 
+[[package]]
+name = "fuchsia-cprng"
```
kulakowski, line 1589: Hilarious (to me)

crates/bench/src/sqlite.rs
```diff
@@ -12,7 +12,7 @@ use std::{
     hint::black_box,
     sync::{Arc, RwLock},
 };
-use tempfile::TempDir;
+use tempdir::TempDir;
```
kulakowski, line 15: Don't change it at this point, but why did this change in this PR?

crates/core/src/db/commit_log.rs
```diff
@@ -248,7 +245,7 @@ impl CommitLogView {
     /// the transactions in a [`Commit`].
     ///
     /// The iterator attempts to read each large object in turn, yielding an
-    /// [`io::Error`] with kind [`io::ErrorKind::NotFound`] if the object was
+    /// [`io::Error`] with constraints [`io::ErrorKind::NotFound`] if the object was
```
kulakowski, line 248: Was this a stray find-and-replace change? It reads oddly now.

crates/core/src/db/datastore/traits.rs
```diff
@@ -223,7 +223,7 @@ pub trait MutTxDatastore: TxDatastore + MutTx {
 
 /// Describes a programmable [`TxDatastore`].
 ///
-/// A programmable datastore is one which has a program of some kind associated
+/// A programmable datastore is one which has a program of some constraints associated
```
kulakowski, line 226: Ditto on the find-and-replace in a comment.

crates/core/src/db/relational_db.rs
```diff
@@ -477,7 +467,7 @@ impl RelationalDB {
     ///
     /// NOTE: It loads the data from the table into it before returning
     #[tracing::instrument(skip(self, tx, index), fields(index=index.name))]
-    pub fn create_index(&self, tx: &mut MutTxId, index: IndexDef) -> Result<IndexId, DBError> {
+    pub fn create_index(&self, tx: &mut MutTxId, table_id: TableId, index: IndexDef) -> Result<IndexId, DBError> {
```
kulakowski, line 470: Why is `table_id` plumbed through but not used? Does it get used in #267 ?

crates/core/src/db/datastore/system_tables.rs
```diff
@@ -169,14 +167,14 @@ st_fields_enum!(
...
-    "kind", Kind = 2,
+    "constraints", Constraints = 2,
     "table_id", TableId = 3,
     "columns", Columns = 4,
 });
 // WARNING: For a stable schema, don't change the field names and discriminants.
 st_fields_enum!(enum StModuleFields {
     "program_hash", ProgramHash = 0,
-    "kind", Kind = 1,
+    "constraints", Kind = 1,
```
kulakowski, line 177: It's a bit odd that this one did not change both, but the equivalent in `StConstraintsFields` did. But maybe that's an artifact of the split from #267?

crates/core/src/db/datastore/system_tables.rs
```diff
@@ -945,7 +941,7 @@ pub struct ModuleKind(u8);
 
 /// The [`ModuleKind`] of WASM-based modules.
 ///
-/// This is currently the only known kind.
+/// This is currently the only known constraints.
```
kulakowski, line 944: Another constraint-kind doc comment change?

crates/core/src/db/datastore/system_tables.rs
```diff
@@ -928,14 +924,14 @@ impl From<StConstraintRow<String>> for ProductValue {
...
-            x.kind.bits(),
+            x.constraints.bits(),
             x.table_id,
             ArrayValue::from(cols)
         ]
     }
 }
 
-/// Indicates the kind of module the `program_hash` of a [`StModuleRow`]
+/// Indicates the constraints of module the `program_hash` of a [`StModuleRow`]
```
kulakowski, line 934: Ditto. I'm a bit concerned about these.

crates/core/src/db/datastore/locking_tx_datastore/mod.rs
```diff
@@ -503,7 +505,7 @@ impl Inner {
                     is_unique: false,
                     index_type: IndexType::BTree,
                 },
-                x => panic!("Adding constraint of kind `{x:?}` is not supported yet."),
+                x => panic!("Adding constraint of constraints `{x:?}` is not supported yet."),
```
kulakowski, line 508: Yet another.

kulakowski (changes requested):
I'm worried about the number of seemingly stray kind-constraint find and replace changes.

crates/vm/src/expr.rs
mamcx: Yes, this come in the next PR.

crates/core/src/db/commit_log.rs
mamcx: You are right. I make a wrong refactor.

crates/core/src/db/relational_db.rs
mamcx: Correct. I will remove this change because if is not used is pointless here.

crates/sats/src/db/def.rs
```diff
@@ -343,24 +332,118 @@ impl From<&ColumnSchema> for ProductTypeElement {
 #[derive(Clone)]
 pub struct ColumnDefMeta {
     pub column: ProductTypeElement,
-    pub attr: ColumnIndexAttribute,
+    pub attr: Constraints,
```
kulakowski, line 335: Nit: should this field be named `constraints` now?

crates/sats/src/db/def.rs
```diff
@@ -343,24 +332,118 @@ impl From<&ColumnSchema> for ProductTypeElement {
...
 
-impl From<&ColumnSchema> for ColumnDefMeta {
-    fn from(value: &ColumnSchema) -> Self {
+/// Describe the columns + meta attributes
+/// TODO(cloutiertyler): This type should be deprecated and replaced with
+/// ColumnDef or ColumnSchema where appropriate
+#[derive(Debug, Clone, Eq, PartialEq, PartialOrd, Ord)]
+pub struct ProductTypeMeta {
+    pub columns: ProductType,
+    pub attr: Vec<Constraints>,
```
kulakowski, line 345: Nit: should this field be named `constraints` now?

kulakowski (approved):

kulakowski:
Also, my approval is assuming the chat discussion about autoinc sematics etc is agreed to be good

crates/sats/src/db/def.rs
mamcx: This will be gone it the last PR, and for now `attr` is more correct because is for the column, not the table.

PR #440, opened 2023-10-17
[DO NOT MERGE] Make rust 1.71 happy

Centril:
Pacify a probable bug in Rust 1.71 that does not reproduce in later compilers.

Do not merge into master; only into the 0.7.1 release.


jdetter:
We got this into `v0.7.1-beta-hotfix1`, thank you @Centril !

PR #445, opened 2023-10-18
Fix "identity not yet recieved" intermittent failure in tests

gefjon:
# Description of Changes

This error turns out to have been a legitimate, albeit rare, race condition, which could have (but apparently didn't) manifested in user code.

The cause, ultimately, was holding locks for too-short scopes while initializing the connection. It was possible for a new connection to receive and completely process its `IdentityToken` message before calling `maybe_set_credentials`, which would then clobber the received identity.

This is now fixed (hopefully) by holding the `CredentialStore` lock for the duration of `BackgroundDbConnection::connect`.
Two other locks, on `ClientCache` and `ReducerCallbacks`, are also newly held for the duration to avoid similar issues.

In addition, this commit adds a call to `disconnect` to the beginning of `connect`. This will prevent pathological user code
which connects twice in a row without an intervening disconnect from encountering bizarre errors caused by the first connection's background workers staying active.

This commit also bumps the `RUST_LOG` env variable for the SDK test clients to `trace`, and adds several trace-level logs to the SDK, which were instrumental in pinning down the failure. This does mean that log output of failed SDK tests will now be much noisier.

Note that # API and ABI

3; this PR non-trivially alters the SDK's connection logic, changing edge-case interactions with other parts of the SDK not directly edited in the PR.

### DIFF
--- a/crates/sdk/src/callbacks.rs
+++ b/crates/sdk/src/callbacks.rs
@@ -863,7 +863,10 @@ impl CredentialStore {
     /// Do not invoke on-connect callbacks.
     pub(crate) fn maybe_set_credentials(&mut self, credentials: Option<Credentials>) {
         if let Some(creds) = &credentials {
-            log::trace!("maybe_set_credentials: Setting stored credentials to Some({:?})", creds.identity);
+            log::trace!(
+                "maybe_set_credentials: Setting stored credentials to Some({:?})",
+                creds.identity,
+            );
         } else {
             log::trace!("maybe_set_credentials: Setting stored credentials to None");
         }
--- a/crates/sdk/src/callbacks.rs
+++ b/crates/sdk/src/callbacks.rs
@@ -862,6 +862,12 @@ impl CredentialStore {
     ///
     /// Do not invoke on-connect callbacks.
     pub(crate) fn maybe_set_credentials(&mut self, credentials: Option<Credentials>) {
+        if let Some(creds) = &credentials {
+            log::trace!("maybe_set_credentials: Setting stored credentials to Some({:?})", creds.identity);
+        } else {
+            log::trace!("maybe_set_credentials: Setting stored credentials to None");
+        }
+
         self.credentials = credentials;
     }
 
--- a/crates/sdk/src/callbacks.rs
+++ b/crates/sdk/src/callbacks.rs
@@ -287,6 +287,10 @@ impl<Args: OwnedArgs> CallbackMap<Args> {
                     // Enter a dynamic context where `CURRENT_STATE`
                     // is the state which caused this callback invocation.
                     let _guard = CurrentStateGuard::with_current_state(db_state);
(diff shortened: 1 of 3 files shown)


crates/sdk/src/callbacks.rs
```diff
@@ -930,8 +939,6 @@ impl CredentialStore {
             );
         }
 
-        self.callbacks.invoke((creds.clone(), address), state);
```
kulakowski, line 933: Confirming that this just moves this call to after some log tracing?

kulakowski (approved):

crates/sdk/src/callbacks.rs
gefjon: Oh, yeah, this was an instinctive edit to remove half a clone. Moving the callbacks.invoke to the end means you don't need to clone at all if you already have stored creds.

gefjon:
Well that's weird as hell. We see:

```
[2023-10-18T17:11:27Z TRACE spacetimedb_sdk::callbacks] handle_identity_token: received address Address(43a73635274fb1532303119c059f21e5)
[2023-10-18T17:11:27Z TRACE spacetimedb_sdk::callbacks] handle_identity_token: storing new credentials
[2023-10-18T17:11:27Z TRACE spacetimedb_sdk::callbacks] CallbackMap: invoking for args type (spacetimedb_sdk::identity::Credentials, spacetimedb_lib::address::Address)
[2023-10-18T17:11:27Z TRACE spacetimedb_sdk::callbacks] CallbackMap worker: received invoke message for args (spacetimedb_sdk::identity::Credentials, spacetimedb_lib::address::Address) and got state guard
```

and then much later in the logs:

```
[2023-10-18T17:12:27Z TRACE spacetimedb_sdk::callbacks] CallbackMap: invoking for args type (test_client::module_bindings::one_u_8::OneU8, core::option::Option<alloc::sync::Arc<dyn core::any::Any + core::marker::Send + core::marker::Sync>>)
[2023-10-18T17:12:27Z TRACE spacetimedb_sdk::callbacks] CallbackMap worker: received invoke message for args (spacetimedb_sdk::identity::Identity, core::option::Option<spacetimedb_lib::address::Address>, spacetimedb_sdk::reducer::Status, test_client::module_bindings::insert_one_u_8_reducer::InsertOneU8Args) and got state guard
thread 'spacetimedb-background-connection' panicked at 'called `Result::unwrap()` on an `Err` value: Identity not yet received', crates/sdk/tests/test-client/src/main.rs:621:41
```

gefjon:
Eureka! I've solved it! This PR is now a fix! I've updated the PR description to document the bug and the fix.

kim (approved):
👌

kulakowski (approved):

RReverser:
Ohh finally! Those spurious failures were annoying. Well done finding the root cause!

PR #446, opened 2023-10-18
Benchmark: deltas + html

kazimuth:
# Description of Changes

Adds html output + deltas to the custom benchmark reporters.

I spent some time looking at how Criterion does statistically significance comparison; they actually are doing some nontrivial statistical stuff (bootstrapping + best fit against iters:time), which I didn't want to reimplement, so I compute deltas in a dumb interval arithmetic way here.

Because I'm not confident in performing statistical tests here, I don't report largest / most significant changes; I think to do that properly you'd need to do a multiple-comparisons correction of some sort..?

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -121,9 +121,9 @@ jobs:
         working-directory: crates/bench/
         run: |
           if [ -e target/criterion/$NORMALIZED_BRANCH_NAME.json ]; then
-            cargo run --bin summarize markdown-report branch.json $NORMALIZED_BRANCH_NAME.json --report-name report
+            cargo run --bin summarize report branch.json $NORMALIZED_BRANCH_NAME.json --report-name report --markdown
           else
-            cargo run --bin summarize markdown-report branch.json --report-name report
+            cargo run --bin summarize report branch.json --report-name report --markdown
           fi
 
       # this will work for both PR and master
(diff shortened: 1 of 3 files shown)


kazimuth:
benchmarks please

RReverser:
Perhaps a stupid question, but why can't we use Criterion's own comparison and HTML reports?

As in, download previous baseline and then just run our benchmarks via

```
cargo bench -- --baseline previous
```

so it would compare with the downloaded one?

kazimuth:
@RReverser I was writing that script because long-term Nate wants to make a dashboard with benchmark results over time. I wanted to be able to feed him those. It's also nice because the script knows about the schema for our benchmarks and can generate tables that group related benchmarks together. (For various reasons it's difficult to do this with criterion's existing groups due to control flow issues, since we would want to have a single group dispatching to multiple generic benchmark implementations.)

kulakowski:
What's the status of this? Is it blocked on anything I can help with?

drogus:
@kazimuth I only now noticed that you added the changes to the existing `summarize.rs` file. When I was changing the setup I moved the functionality (apart from generating files) to https://github.com/clockworklabs/benchmarks-viewer. If you want to still keep it in the project I guess we could extract it to a crate, but either way it should be in one place (so either extract or move to the web viewer)

kazimuth:
This is no longer relevant here because of @drogus' comment, the relevant functionality belongs on the other project (https://github.com/clockworklabs/benchmarks-viewer).

PR #449, opened 2023-10-19
Working on making -s flag global

jdetter:
# Description of Changes

# API and ABI


Centril (commented):
I'm assuming there are some more changes necessary to tests and whatnot, but the diff so far looks good though.

jdetter:
I've decided against this, we'll just add the `-s` flag to the commands where it makes sense.

The reason why I've decided against it is because you end up with something like this:

```bash
spacetime -s testnet server add ...
```

Where you can specify a server to a command which *shouldn't* be able to take a server and this could be confusing to users. You should only be able to supply `-s <server>` to commands which actually require a server.

PR #451, opened 2023-10-19
vm: Make UPDATE actually update

kim:
# Description of Changes

Previously, an UPDATE statement would be the identity function. Fix this by threading through the assignments (SET clause) and transforming the rows matching the query.

# API and ABI

2


kim:
Yes I realize that the original idea was to avoid an "imperative" evaluation strategy, but it's quite non-obvious how the header information would have to be transformed to make the projection actually work. I suppose the language is currently lacking vocabulary to make those things less error-prone.

PR #266, opened 2023-09-05
Update/upgrade dependencies

coolreader18:
# Description of Changes

<pre><font color="#FF5555"><b>    Removing</b></font> addr2line v0.19.0
<font color="#FF5555"><b>    Removing</b></font> addr2line v0.20.0
<font color="#55FFFF"><b>      Adding</b></font> addr2line v0.21.0
<font color="#55FF55"><b>    Updating</b></font> aho-corasick v1.0.2 -&gt; v1.1.2
<font color="#55FF55"><b>    Updating</b></font> allocator-api2 v0.2.15 -&gt; v0.2.16
<font color="#55FF55"><b>    Updating</b></font> anstream v0.3.2 -&gt; v0.6.4
<font color="#55FF55"><b>    Updating</b></font> anstyle v1.0.1 -&gt; v1.0.4
<font color="#55FF55"><b>    Updating</b></font> anstyle-parse v0.2.1 -&gt; v0.2.2
<font color="#55FF55"><b>    Updating</b></font> anstyle-wincon v1.0.1 -&gt; v3.0.1
<font color="#55FF55"><b>    Updating</b></font> anyhow v1.0.71 -&gt; v1.0.75
<font color="#55FFFF"><b>      Adding</b></font> arbitrary v1.3.1
<font color="#55FF55"><b>    Updating</b></font> async-trait v0.1.68 -&gt; v0.1.74
<font color="#FF5555"><b>    Removing</b></font> atty v0.2.14
<font color="#55FF55"><b>    Updating</b></font> axum v0.6.18 -&gt; v0.6.20
<font color="#55FF55"><b>    Updating</b></font> backtrace v0.3.68 -&gt; v0.3.69
<font color="#55FF55"><b>    Updating</b></font> base64 v0.21.2 -&gt; v0.21.4
<font color="#55FF55"><b>    Updating</b></font> bindgen v0.64.0 -&gt; v0.65.1
<font color="#55FF55"><b>    Updating</b></font> bitflags v2.3.3 -&gt; v2.4.1
<font color="#55FF55"><b>    Updating</b></font> bumpalo v3.13.0 -&gt; v3.14.0
<font color="#55FF55"><b>    Updating</b></font> bytecount v0.6.3 -&gt; v0.6.4
<font color="#55FF55"><b>    Updating</b></font> byteorder v1.4.3 -&gt; v1.5.0
<font color="#55FF55"><b>    Updating</b></font> bytes v1.4.0 -&gt; v1.5.0
[...]

PR #457, opened 2023-10-20
Switch from wasmer to wasmtime

coolreader18:
# Description of Changes

A draft PR to run benchmarks on wasmtime vs wasmer

# API and ABI

 - [x] This is a breaking change to the ClientAPI

The host_type parameter to `/publish` is renamed from wasmer to wasmtime, but I don't think that it was even being checked anyway.


coolreader18:
benchmarks please

cloutiertyler:
> There seem to be quite a few refactorings that are independent of the move to wasmtime; could these be extracted and reviewed separately? That could also make the benchmark more reliable.

If we do end up merging this (and I will not do so without performance data that demonstrates we understand the impact), then we should break this up and make sure we document it well.

Right now it's just a draft to evaluate performance I believe.

coolreader18:
@Centril this is like 3 separate PRs that are in one branch b/c they're not merged yet. once #266 and #417 merge this'd just be the last commit

coolreader18:
hmmm :/ idk what's up with this. how old are the old benchmark numbers?

RReverser:
They should be from master. That said, this PR does quite a lot, including upgrade of a bunch of dependencies, so it's hard to attribute the performance difference to just the engine change. It would be better to split them out and keep this change isolated.

coolreader18:
@RReverser well, approve #266 and we can do that :))

RReverser:
> @RReverser well, approve #266 and we can do that :))

I started benchmarks on that one for now :) (also I only reread conversation above which had the same points raised already, only after leaving my comment lol)

coolreader18:
hey

coolreader18:
test

crates/core/src/host/wasmer/mod.rs
```diff
@@ -1,22 +1,40 @@
...
 use super::wasm_common::{abi, module_host_actor::WasmModuleHostActor, ModuleCreationError};
-use super::{EnergyMonitor, EnergyQuanta};
+use super::EnergyMonitor;
+
+static ENGINE: Lazy<Engine> = Lazy::new(|| {
+    let mut config = wasmtime::Config::new();
+    config
+        .cranelift_opt_level(wasmtime::OptLevel::Speed)
+        .consume_fuel(true)
+        .wasm_backtrace_details(wasmtime::WasmBacktraceDetails::Enable);
```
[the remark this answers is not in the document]
coolreader18: I assume you know more on the matter than I do, but are you sure it comes with runtime overhead? From what I can tell of looking at the source code, it'd only have a cost at JIT time, as well as whenever a backtrace is captured, which shouldn't happen frequently.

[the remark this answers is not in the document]
RReverser: Definitely don't assume that 😅

I think wasmtime also had to add extra logic to wasm<>host calls to correctly unwind stacks between JIT and native code, but it seems at least not anymore?

If so, yeah, you're right the overhead would be only for when a panic actually occurs. Might be worth measuring on one of the iteration benchmarks as those do a lot of host<>wasm calls just to be sure.

crates/client-api/src/routes/energy.rs
```diff
@@ -46,11 +46,10 @@ pub async fn add_energy<S: ControlStateDelegate>(
     let mut balance = ctx
         .get_energy_balance(&auth.identity)
         .map_err(log_and_500)?
-        .map(|quanta| quanta.0)
-        .unwrap_or(0);
+        .map_or(0, |quanta| quanta.get());
```
gefjon, line 49: Are the changes in this file relevant to this PR? If not, I'd prefer to see them split into a separate PR.

crates/core/Cargo.toml
```diff
@@ -85,10 +85,11 @@ tracing.workspace = true
...
 urlencoding.workspace = true
 uuid.workspace = true
-wasmer-middlewares.workspace = true
-wasmer-types.workspace = true
-wasmer-vm.workspace = true
-wasmer.workspace = true
+# wasmer-middlewares.workspace = true
+# wasmer-types.workspace = true
+# wasmer-vm.workspace = true
+# wasmer.workspace = true
```
gefjon, line 91: Please remove these. Also, if we no longer depend on Wasmer at all, please remove the deps from the workspace. If we do still depend on Wasmer somehow, I'd like a comment in the workspace Cargo.toml describing what we use it for.

crates/core/src/control_db.rs
```diff
@@ -549,7 +549,7 @@ impl ControlDb {
                 return Err(Error::DecodingError(bsatn::DecodeError::BufferLength));
             };
             let balance = i128::from_be_bytes(arr);
-            Ok(Some(EnergyQuanta(balance)))
+            Ok(Some(EnergyQuanta::new(balance)))
```
gefjon, line 552: Same question as above: is this relevant, or an unrelated refactoring?

crates/core/src/host/host_controller.rs
```diff
@@ -108,10 +118,19 @@ impl EnergyQuanta {
         self.0 as u64
     }
 
+    const WASMTIME_MULTIPLIER: u64 = 1_000;
+    pub fn as_wasmtime_points(&self) -> u64 {
+        self.as_wasmer_points() / Self::WASMTIME_MULTIPLIER
+    }
```
gefjon, line 124: Since `WASMTIME_MULTIPLIER` is used as both a product and a divisor, could you include the units in the name? I.e. `FOO_TO_WASMTIME_MULTIPLIER` or `WASMTIME_TO_FOO_MULTIPLIER` as appropriate, where `FOO` is... whatever `wasmer_points` gets changed to, I hope. Alternatively, add a docstring to the const which communicates that information.

crates/core/src/host/host_controller.rs
```diff
@@ -90,16 +90,26 @@ impl fmt::Display for DescribedEntityType {
...
+
+    #[inline]
+    pub fn get(&self) -> i128 {
+        self.0
+    }
+
     /// A conversion function to convert from the canonical unit to points used
     /// by Wasmer to track energy usage.
-    pub fn as_points(&self) -> u64 {
+    pub fn as_wasmer_points(&self) -> u64 {
```
gefjon, line 112: Since we're no longer using Wasmer, can this be renamed?

crates/core/src/host/instance_env.rs
```diff
@@ -194,13 +194,13 @@ impl InstanceEnv {
...
     #[tracing::instrument(skip_all)]
-    pub fn get_table_id(&self, table_name: String) -> Result<TableId, NodesError> {
+    pub fn get_table_id(&self, table_name: &str) -> Result<TableId, NodesError> {
         let stdb = &*self.dbic.relational_db;
         let tx = &mut *self.get_tx()?;
 
         // Query the table id from the name.
         let table_id = stdb
-            .table_id_from_name(tx, &table_name)?
+            .table_id_from_name(tx, table_name)?
```
gefjon, line 203: This seems like a solid change, but appears not to be relevant to this PR.

crates/core/src/host/instance_env.rs
```diff
@@ -238,7 +238,7 @@ impl InstanceEnv {
             _ => return Err(NodesError::BadIndexType(index_type)),
         };
 
-        let cols = NonEmpty::from_slice(&col_ids)
+        let cols = NonEmpty::from_vec(col_ids)
```
gefjon, line 241: Ditto.

crates/core/src/host/wasm_common.rs
```diff
@@ -220,7 +221,7 @@ pub trait ResourceIndex {
 
 macro_rules! decl_index {
     ($name:ident => $resource:ty) => {
-        #[derive(Copy, Clone, wasmer::ValueType)]
+        #[derive(Copy, Clone)] // , wasmer::ValueType
```
gefjon, line 224:
```suggestion
        #[derive(Copy, Clone)]
```

crates/core/src/host/wasmer/mod.rs
```diff
@@ -1,22 +1,40 @@
...
+use super::wasm_common::module_host_actor::InitializationError;
 use super::wasm_common::{abi, module_host_actor::WasmModuleHostActor, ModuleCreationError};
-use super::{EnergyMonitor, EnergyQuanta};
+use super::EnergyMonitor;
+
+static ENGINE: Lazy<Engine> = Lazy::new(|| {
+    let mut config = wasmtime::Config::new();
+    config
+        .cranelift_opt_level(wasmtime::OptLevel::Speed)
+        .consume_fuel(true)
```
gefjon, line 28: In the future, I think we should investigate whether epoch interruption is sufficient for our needs. Within this PR, fuel seems like the right thing to use, as it's the closest translation of Wasmer metering.

crates/core/src/host/wasmer/mod.rs
```diff
@@ -62,52 +67,90 @@ pub fn make_actor(
...
-        data: &[u8],
-    ) -> Result<(), MemoryAccessError> {
-        ptr.slice(&self.view(store), len)?.write_slice(data)
+impl MemView {
+    fn from_slice_mut(v: &mut [u8]) -> &mut Self {
+        unsafe { &mut *(v as *mut [u8] as *mut MemView) }
+    }
+    fn from_slice(v: &[u8]) -> &Self {
+        unsafe { &*(v as *const [u8] as *const MemView) }
+    }
```
gefjon, line 129: Please add `SAFETY` comments which mention that `MemView` is a transparent newtype around `[u8]`.

crates/core/src/host/wasmer/mod.rs
```diff
@@ -62,52 +67,90 @@ pub fn make_actor(
...
+            .ok_or_else(|| WasmError::Wasm(wasmtime::Trap::MemoryOutOfBounds.into()))
+    }
+    fn deref_str(&self, offset: WasmPtr<u8>, len: u32) -> Result<&str, WasmError> {
+        let b = self.deref_slice(offset, len)?;
+        std::str::from_utf8(b).map_err(|e| WasmError::Wasm(e.into()))
+    }
+    fn deref_str_lossy(&self, offset: WasmPtr<u8>, len: u32) -> Result<Cow<str>, WasmError> {
+        self.deref_slice(offset, len).map(String::from_utf8_lossy)
+    }
+    fn opt_deref_str_lossy(&self, offset: WasmPtr<u8>, len: u32) -> Result<Option<Cow<str>>, WasmError> {
```
gefjon, line 143:
```suggestion
    /// Convert `self[offset..offset+len]` to a string,
    /// replacing invalid UTF-8 bytes with U+FFFD REPLACEMENT CHARACTER.
    /// Returns `None` if `offset` is 0, the null pointer.
    fn try_deref_str_lossy(&self, offset: WasmPtr<u8>, len: u32) -> Result<Option<Cow<str>>, WasmError> {
```

gefjon (commented):
This looks great, thanks Noa! Before we merge, we should:
1. Rename the wasmer files to wasmtime.
2. Run a BitCraft playtest with this branch to see how our performance metrics change.

crates/client-api/src/routes/energy.rs
coolreader18: Yeah, the change from .0 to .get() is part of foolproofing fuel<->quanta conversion. So that EnergyQuanta is more opaque.

crates/core/src/host/host_controller.rs
coolreader18: Before I saw this comment, was actually working on a different way of going about it that makes it more clear and moves it to `host::wasmtime`

crates/core/src/host/instance_env.rs
coolreader18: It is relevant, since with wasmtime we actually do get a `str` reference directly to memory instead of being forced to clone it out, so without this change there'd be a new explicit `.to_owned()` in wasm_instance_env. If you do feel strongly about it I can do that and leave this as a follow-up?

gefjon: Nah, as long as it's connected to this PR I'm glad to have it in. Just trying to be diligent about not merging in unrelated changes.

crates/core/src/host/wasmer/mod.rs
RReverser: Ideally I'd use something like [`bytemuck::TransparentWrapper`](https://docs.rs/bytemuck/latest/bytemuck/derive.TransparentWrapper.html) derive for those for clarity + it makes it harder to make a mistake.

crates/core/src/host/wasmer/mod.rs
coolreader18: Hmmm, idk about the try naming - I'm not a huge fan of `opt_` as a prefix either, but try implies the deref is a fallible operation, when it's really just treating null as a sentinel value. I could do a follow up fixing this, maybe it could be one set of functions without the `opt_` prefix but that return `Option<>`, and then a extension trait with a `fn non_nullable(self: Result<Option<T>, WasmError>) -> Result<T, WasmError>;` so that the calls look like `mem.deref_str(buf, buf_len).non_nullable()?`

coolreader18: Oh, or I guess I could do it here. I forgot that I added those functions in this branch, lol

crates/core/src/host/wasmer/mod.rs
coolreader18: Maybe I'll go through in a separate PR and switch that manual stuff to a derive - there's at least a good few of these throughout the codebase (mostly written by me)

kulakowski (commented):
I think the code as such is fine and pretty thoroughly reviewed now so I'll say what I think we need to do to land this.

edit: added the need for comments

- [ ] appropriate levels of doc comments
- [x] ensure no perf regressions
- [ ] ensure energy monitoring is sensible when deployed
- [ ] understand what is needed to deploy this for production
  - [ ] is the "wasmer" -> "wasmtime" name change going to break any deployment scripts?
  - [ ] is that same name change going to affect control db instances?
- [ ] test on all supported OSes
  - [x] linux
  - [ ] macos
  - [ ] windows

gefjon:
> ensure no perf regressions

As discussed elsewhere, internal testing has shown mild to moderate performance improvements compared to v0.7.2-beta on one of BitCraft's hot loops. That testing was not particularly rigorous, and it's possible there are edge-case-ey regressions we haven't caught - e.g. Wasmtime might make module compilation/instantiation more expensive - but I feel comfortable saying that the part we care about, module execution, is at least no worse.

crates/core/src/host/wasmtime/mod.rs
```diff
@@ -0,0 +1,228 @@
...
+
+impl WasmtimeFuel {
+    /// 1000 energy quanta == 1 wasmtime fuel unit
+    const QUANTA_MULTIPLIER: i128 = 1_000;
+
+    /// Convert from EnergyQuanta to wasmtime fuel. The second element of the tuple is the remainder
+    /// of quanta in the case that `eq` isn't in u64 range. It can be safely ignored if the supplied
+    /// budget isn't being bookkept.
+    fn from_energy_quanta(eq: EnergyQuanta) -> (Self, EnergyQuanta) {
+        if eq.get() < 0 {
```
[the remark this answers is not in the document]
coolreader18: Negative fuel doesn't exist. My reading of the EnergyQuanta type is that the only reason it supports negative values is because it represents an energy balance (i.e. belonging to an account, that can go into the negative rather than hard cut-off). I was considering making a separate unsigned ReducerBudget type, which I might do as a followup cause I feel like adding negative-value handling everywhere just bc EnergyQuanta is signed is more effort than specifically using a signed integer in the case where we care.

crates/core/src/host/wasmtime/wasmtime_module.rs
```diff
@@ -0,0 +1,277 @@
...
+        WasmtimeModule { module }
+    }
+
+    pub const IMPLEMENTED_ABI: abi::VersionTuple = abi::VersionTuple::new(7, 0);
+
+    pub(super) fn link_imports(linker: &mut Linker<WasmInstanceEnv>) -> anyhow::Result<()> {
+        #[allow(clippy::assertions_on_constants)]
+        const _: () = assert!(WasmtimeModule::IMPLEMENTED_ABI.major == spacetimedb_lib::MODULE_ABI_MAJOR_VERSION);
+        linker
+            .func_wrap("spacetime_7.0", "_schedule_reducer", WasmInstanceEnv::schedule_reducer)?
```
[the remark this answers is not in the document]
coolreader18: Do we? I think that might be for the crate version, not the abi version. I can't see anything in `tools/`that edits abi versions.

crates/core/src/host/wasmtime/mod.rs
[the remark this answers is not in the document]
coolreader18: My thinking with the remainder was that logically, the energy quanta used for a budget would be like, subtracted from the balance. so:

```
budget = 1000
balance -= budget
budget, remainder = from_energy_quanta(budget)
balance += remainder
```
So therefore if the budget is somehow negative, we wouldn't want to zero out the negative balance, so this behavior ensures everything still works even with the same above pseudocode.

crates/core/src/host/wasmtime/wasmtime_module.rs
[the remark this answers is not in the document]
kulakowski: Looks like no, but wanted to double check in case @RReverser had made something since I last looked.

crates/core/src/host/wasmtime/mod.rs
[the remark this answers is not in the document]
coolreader18: I mean if you ignore the second tuple element this is just like, the standard way you'd convert between these 2 units. I feel like by forcing you to ignore it, it makes it explicit that it's lossy and there's a remainder that's not paid attention to. And in any case I think I do wanna do a follow up that splits EnergyQuanta into ReducerBudget & EnergyBalance. u64 is more than enough for a budget for a single reducer.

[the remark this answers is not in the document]
coolreader18: Or even if it doesn't narrow the integer width for ReducerBudget, I wanna make it unsigned.

crates/core/src/host/wasmer/mod.rs
kulakowski: Would you mind filing a ticket about this?

crates/core/src/host/wasmtime/mod.rs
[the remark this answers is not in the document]
kulakowski: I don't agree it's standard because of the special casing negative values.

coolreader18:
@kazimuth just pushed a commit that lets wasmtime cache compilation artifacts to the filesystem - I think you were looking for something like this for benchmarking, right?

coolreader18:
(if someone feels that's too tangential/confounding to this PR, fair, I'm happy to split it off to a follow-up)

crates/core/src/host/wasmtime/wasmtime_module.rs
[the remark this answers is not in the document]
RReverser: No, this new ABI scheme is a very recent @coolreader18's addition, I don't think we have automated it yet.

crates/core/src/host/wasmtime/wasm_instance_env.rs
```diff
@@ -396,37 +385,23 @@ impl WasmInstanceEnv {
...
-            new_row.encode(&mut row_buffer);
-            assert_eq!(
-                row_buffer.len(),
-                row_len as usize,
-                "autoinc'd row is different encoded size from original row"
-            );
-            mem.set_bytes(&caller, row, row_len, &row_buffer)?;
+            let ctx = env.reducer_context();
+            let new_row = env.instance_env.insert(&ctx, table_id.into(), row_buffer)?;
+            new_row.encode(&mut { row_buffer });
```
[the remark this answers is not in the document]
coolreader18: So that it copies `row_buffer` as a temporary instead of mutating the local. It's like the weirdness that comes with `impl io::Read for &[u8]`, basically. Obviously in this scenario we don't care about accessing the original value again anyway, but I still feel this conveys intent better.

crates/core/src/host/wasmtime/mod.rs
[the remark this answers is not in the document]
coolreader18: But it's not special casing - it's impossible to have negative fuel (like, it's literally a u64), so this is the only real way to do it. The `-1e9 eq -> (-1e6 fuel, 0 eq)` alternative you mentioned is impossible.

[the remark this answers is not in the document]
kulakowski: (To be clear I don't think this is a big deal at all)

crates/core/src/host/wasmtime/wasm_instance_env.rs
```diff
@@ -169,16 +179,14 @@ impl WasmInstanceEnv {
...
-        mut caller: FunctionEnvMut<'_, Self>,
-        func: &'static str,
-        call: AbiCall,
-        f: impl FnOnce(FunctionEnvMut<'_, Self>, &Mem) -> WasmResult<()>,
-    ) -> RtResult<u16> {
-        let span_start = span::CallSpanStart::new(call);
+        mut caller: Caller<'_, Self>,
+        func: AbiCall,
+        f: impl FnOnce(&mut Caller<'_, Self>) -> WasmResult<()>,
+    ) -> RtResult<u32> {
```
kulakowski, line 185: re the u16 -> u32 change throughout this file: is this ultimately stemming from a wasmtime api? like does it only let you have fn types whose return type impls some trait goop, and it bottoms out in only u32s etc because wasm only has i32s and not i16s?

I'm assuming the same for the anyhow::Result change in RtResult.

kulakowski (approved):

crates/core/src/host/wasmtime/wasm_instance_env.rs
coolreader18: Yeah, which was the same for wasmer and the difference is that wasmtime only lets you use `{i,u,f}{32,64}` for those, while wasmer was happy to truncate/zero-extend with other integer types.

coolreader18:
wahoo!

PR #466, opened 2023-10-25
Alterations to noa/wasmtime for running perf

gefjon:
# Description of Changes

- `inline(never)` typed wrappers around host functions in the module, so the types appear in backtraces.
- Pass `-g` to `wasm-opt` to preserve symbols in WASM blobs. I've been running without `wasm-opt` anyway since it inlines
  without respecting `inline(never).`
- Pass `ProfilingStrategy::PerfMap` to wasmtime so `perf` has names for JITted symbols.

# API and ABI

Not up for review, but 1: only affects debug symbols, inlining and JIT compilation for WASM code; changes observable only under `perf` or a debugger.


coolreader18:
@gefjon is this still something you wanna merge? You could probably put the `ProfilingStrategy::PerfMap` behind an env var check (`STDB_WASM_PERF=1` or something) but otherwise the rest looks like it'd be fine to merge as-is

PR #478, opened 2023-10-26
perf(468): Add metric counters for committed transactions and rollbacks

### LINKED ISSUE #468: Record metric for number of transactions
Label metric with transaction type, database id, and reducer id.

joshua-spacetime:
Closes #468.

# Description of Changes

# API and ABI

### DIFF
--- a/crates/core/src/db/datastore/traits.rs
+++ b/crates/core/src/db/datastore/traits.rs
@@ -446,15 +446,21 @@ pub trait Tx {
     type TxId;
 
     fn begin_tx(&self) -> Self::TxId;
-    fn release_tx(&self, tx: Self::TxId);
+    fn release_tx(&self, ctx: &ExecutionContext, tx: Self::TxId);
 }
 
 pub trait MutTx {
     type MutTxId;
 
     fn begin_mut_tx(&self) -> Self::MutTxId;
-    fn rollback_mut_tx(&self, tx: Self::MutTxId);
-    fn commit_mut_tx(&self, tx: Self::MutTxId) -> Result<Option<TxData>>;
+    fn commit_mut_tx(&self, ctx: &ExecutionContext, tx: Self::MutTxId) -> Result<Option<TxData>>;
+    fn rollback_mut_tx(&self, ctx: &ExecutionContext, tx: Self::MutTxId);
+
+    #[cfg(test)]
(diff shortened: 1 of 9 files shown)


cloutiertyler (approved):
I'll approve this, but I wonder if it would be a better design to put the `database_id` in the `Datastore` or the trait, and to put the transaction type and transaction info inside the Tx itself. Not sure if that's better tbh though, but food for thought.

Also why use the `database_id` rather than the database's `Address`? Please note that `database_id` should never be surfaced to SpacetimeDB users. It's purely internal. `Address` is the identifier for a database for users.

cloutiertyler (changes requested):
Please replace `database_id` with `address` in the exposed metrics.

crates/core/src/db/datastore/traits.rs
```diff
@@ -378,14 +378,20 @@ pub trait Tx {
...
     fn begin_tx(&self) -> Self::TxId;
-    fn release_tx(&self, tx: Self::TxId);
+    fn release_tx(&self, ctx: &ExecutionContext, tx: Self::TxId);
 }
 
 pub trait MutTx {
     type MutTxId;
 
     fn begin_mut_tx(&self) -> Self::MutTxId;
+    fn rollback_mut_tx_with_context(&self, ctx: &ExecutionContext, tx: Self::MutTxId);
```
kulakowski, line 388: Should this just stay `rollback_mut_tx`, and the test-only `rollback_mut_tx` get some worse name? Lots of things are going to plump around an ExecutionContext to the point that it doesn't seem exceptional or worth calling out.

crates/core/src/db/datastore/traits.rs
```diff
@@ -378,14 +378,20 @@ pub trait Tx {
     type TxId;
 
     fn begin_tx(&self) -> Self::TxId;
-    fn release_tx(&self, tx: Self::TxId);
+    fn release_tx(&self, ctx: &ExecutionContext, tx: Self::TxId);
```
kulakowski, line 381: eg it is not `release_tx_with_context`

kulakowski (approved):
Just a note about naming.

crates/core/src/db/datastore/traits.rs
kulakowski: Yeah. Like I said, I'm happy to stamp another PR that changes em later if that's easier.

cloutiertyler (approved):
LGTM

PR #481, opened 2023-10-26
Fix benchmarks please 100%

coolreader18:
# Description of Changes

# API and ABI

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -33,10 +33,12 @@ jobs:
       - name: Check membership
         if: ${{ github.event_name == 'issue_comment' }}
         env:
+          CONTRIB_ORG: clockworklabs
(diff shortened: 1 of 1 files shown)


coolreader18:
I'm actually not necessarily 100% on this but I'm kinda confused. the error is "Message: Resource not accessible by personal access token, Locations: [{Line:1 Column:49}]", which looking at the github cli source code seems to indicate that it's a graphql error. but this isn't using graphql? so idk what the deal is

coolreader18:
ohhh ok now I think I get it

coolreader18:
though that would still mean that the check is failing :/

coolreader18:
ok maybe I'll just try something real quick

coolreader18:
for faster iteration

coolreader18:
Ok, NOW I'm confident this'll work.

kurtismullins (approved):
Nice! LGTM

PR #482, opened 2023-10-26
Fix wasm-opt for C# modules

RReverser:
# Description of Changes

C# generates Wasm code that uses some of the newer performance features (notably, bulk memory ops) and wasm-opt rejects any such instructions by default (there were talks about changing this behavious upstream but that never went anywhere).

Activate all features so that wasm-opt works on anything produced by our compilers, as well as make any future failures a non-critical error just in case.

# API and ABI

### DIFF
--- a/crates/cli/src/tasks/mod.rs
+++ b/crates/cli/src/tasks/mod.rs
@@ -14,17 +14,24 @@ pub fn build(project_path: &Path, skip_clippy: bool, build_debug: bool) -> anyho
         ModuleLanguage::Csharp => build_csharp(project_path, build_debug),
     }?;
     if !build_debug {
+        eprintln!("Optimising module with wasm-opt...");
         let wasm_path_opt = wasm_path.with_extension("opt.wasm");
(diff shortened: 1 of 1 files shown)


RReverser:
@kulakowski Changed the label as I'd rather land this as soon as possible to unbreak C# users who have wasm-opt installed.

kulakowski:
> @kulakowski Changed the label as I'd rather land this as soon as possible to unbreak C# users who have wasm-opt installed.

kulakowski (approved):
Seems like a good tradeoff for right now. It might be that we want some sort of option to fail if not present so that we don't accidentally regress e.g. production deployments.

RReverser:
> It might be that we want some sort of option to fail if not present so that we don't accidentally regress e.g. production deployments.

I agree that would be good, but only when we figure some auto-packaging so that it's automatically available for users. I could probably just integrate https://github.com/brson/wasm-opt-rs but it takes a non-negligible amount of time to build, and generally I'm hesitant to add even more C++ deps to the build for a tool that has native binaries. So for now it remains a TODO for later...

PR #483, opened 2023-10-26
perf(470): Add metrics for transaction cpu and elapsed time

### LINKED ISSUE #470: Record metric for transaction elapsed time
That is, the total execution time of a transaction, not including time spent waiting in the queue.
Label each recording with transaction type, database id, and reducer id.

joshua-spacetime:
Closes #470.

# Description of Changes

# API and ABI

### DIFF
--- a/crates/core/src/db/db_metrics/mod.rs
+++ b/crates/core/src/db/db_metrics/mod.rs
@@ -75,6 +75,16 @@ metrics_group!(
         #[help = "The cumulative number of rolled back transactions"]
         #[labels(txn_type: TransactionType, db: Address, reducer: str)]
         pub rdb_num_txns_rolledback: IntCounterVec,
+
+        #[name = spacetime_txn_elapsed_time_ns]
+        #[help = "The total elapsed (wall) time of a transaction (nanoseconds)"]
+        #[labels(txn_type: TransactionType, db: Address, reducer: str)]
+        pub rdb_txn_elapsed_time_ns: HistogramVec,
+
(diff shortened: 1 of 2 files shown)


crates/core/src/db/relational_db.rs
```diff
@@ -268,6 +268,12 @@ impl RelationalDB {
         self.inner.rollback_mut_tx_with_context(ctx, tx)
     }
 
+    #[tracing::instrument(skip_all)]
+    pub fn rollback_tx_with_context(&self, ctx: &ExecutionContext, tx: MutTxId) {
```
kulakowski, line 272: Same naming question as in the other PR, maybe just `rollback_tx` for the non-test case.

crates/core/src/db/db_metrics/mod.rs
```diff
@@ -75,6 +75,16 @@ metrics_group!(
         #[help = "The cumulative number of rolled back transactions"]
         #[labels(txn_type: TransactionType, db: Address, reducer: str)]
         pub rdb_num_txns_rolledback: IntCounterVec,
+
+        #[name = spacetime_txn_elapsed_time_us]
+        #[help = "The total elapsed (wall) time of a transaction (microseconds)"]
```
kulakowski, line 80: As transactions get faster, I do wonder if we will wish this was nanoseconds. Curious what others think.

kulakowski (approved):
lgtm modulo the unit question and the naming question

I'm happy if the internal names get tweaked later, too, if that's easier than rebasing or whatnot, if we do decide to change.

PR #484, opened 2023-10-26
Move optimizations from sql to vm

mamcx:
# Description of Changes

To improve the performance of the `query` macro, move the query optimization from the `sql` compilation into the `vm` execution engine instead.

# API and ABI

1

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4461,6 +4461,7 @@ version = "0.7.1"
 dependencies = [
  "anyhow",
  "derive_more",
+ "nonempty",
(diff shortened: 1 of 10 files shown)


joshua-spacetime (commented):
Nice! Is it possible to add a test for `query!` now?

mamcx:
> Nice! Is it possible to add a test for `query!` now?

In this PR or make another?

joshua-spacetime (approved):
LGTM! I'm not sure where a test for `query!` would go, so it's probably best to figure that out in a follow up patch.

PR #486, opened 2023-10-27
refactor(485): Label metrics with database address and reducer name

### LINKED ISSUE #485: Label metrics with database address instead of id

joshua-spacetime:
Fixes #485.

Also label metrics with reducer name instead of internal reducer id.

# API and ABI

### DIFF
--- a/crates/core/src/db/commit_log.rs
+++ b/crates/core/src/db/commit_log.rs
@@ -122,8 +122,8 @@ impl CommitLog {
                     // Increment rows inserted metric
                     let metric = rows_inserted.with_label_values(
                         &ctx.txn_type(),
-                        &ctx.database_id(),
-                        &ctx.reducer_id().unwrap_or_default(),
+                        &ctx.database(),
+                        ctx.reducer_name().unwrap_or_default(),
                         &table_id,
                     );
                     metric.inc();
@@ -133,8 +133,8 @@ impl CommitLog {
                     // Increment rows deleted metric
                     let metric = rows_deleted.with_label_values(
                         &ctx.txn_type(),
-                        &ctx.database_id(),
-                        &ctx.reducer_id().unwrap_or_default(),
+                        &ctx.database(),
+                        ctx.reducer_name().unwrap_or_default(),
                         &table_id,
                     );
                     metric.inc();
--- a/crates/core/src/host/wasmer/wasm_instance_env.rs
+++ b/crates/core/src/host/wasmer/wasm_instance_env.rs
@@ -13,7 +13,6 @@ use crate::host::wasm_common::{
     instrumentation::{Call, CallTimes},
     AbiRuntimeError, BufferIdx, BufferIterIdx, BufferIters, Buffers, TimingSpan, TimingSpanIdx, TimingSpanSet,
 };
-use crate::host::ReducerId;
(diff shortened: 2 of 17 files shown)


crates/lib/src/address.rs
```diff
@@ -23,6 +23,12 @@ pub struct Address {
 
 impl_st!([] Address, _ts => AlgebraicType::product([("__address_bytes", AlgebraicType::bytes())]));
 
+impl Default for Address {
```
kulakowski, line 26: Maybe a comment about what, if any, semantics that an all-zeroes address is imbued with. This might be documented at the point that Address is defined.

crates/core/src/host/wasmer/wasm_instance_env.rs
```diff
@@ -154,13 +154,16 @@ impl WasmInstanceEnv {
     }
 
     /// Provide this environment with information about the current reducer being executed.
-    pub fn set_reducer_id(&mut self, id: u64) {
-        self.reducer_id = id;
+    pub fn set_reducer_name(&mut self, name: String) {
```
kulakowski, line 157: This is going to collide with Noa's refactor in https://github.com/clockworklabs/SpacetimeDB/pull/417 which moves the setting of this into the place where reducers are called. It's a change I approve of, so I'd like to see this updated to do that. Noa's PR is approved with one tiny change request so hopefully soon.

crates/core/src/db/commit_log.rs
```diff
@@ -122,8 +122,8 @@ impl CommitLog {
                     // Increment rows inserted metric
                     let metric = rows_inserted.with_label_values(
                         &ctx.txn_type(),
-                        &ctx.database_id(),
-                        &ctx.reducer_id().unwrap_or(0),
+                        &ctx.database(),
+                        ctx.reducer_name().unwrap_or(""),
```
kulakowski, line 126: I'm fine if this also changes metrics to track reducer names rather than ids. But if so, please include that in the commit message, and assure me that it's done uniformly for all metrics.

crates/core/src/host/wasmer/wasm_instance_env.rs
```diff
@@ -96,7 +96,7 @@ impl WasmInstanceEnv {
             timing_spans: Default::default(),
             reducer_start,
             call_times: CallTimes::new(),
-            reducer_id: 0,
+            reducer_name: None,
```
kulakowski, line 99: I don't hugely care: I'm curious if just String is ok, given that accessors to it do map_or("", ...). Like, if there is no semantic difference between an empty name and None, we could just have the one state.

crates/core/src/host/wasmer/wasmer_module.rs
```diff
@@ -240,29 +239,25 @@ impl module_host_actor::WasmInstance for WasmerInstance {
 
     fn call_reducer(
         &mut self,
-        reducer_id: usize,
+        op: ReducerOp<'_>,
```
kulakowski, line 242: Nice.

kulakowski (approved):
My actionable feedback is

- [x] please try to rebase past Noa's refactor once that lands (I'll ping about it soon)
- [x] mention the reducer name vs id change
- [x] is it possible that access to the db id got plumbed through somewhere that is no longer used?

PR #487, opened 2023-10-27
fix benchmarks run on master

coolreader18:
# Description of Changes

# API and ABI

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -64,6 +64,10 @@ jobs:
         uses: actions/checkout@v3
         with:
           ref: ${{ env.PR_REF || github.ref }}
+          # if we're on master we want to know what the sha of HEAD~1 is so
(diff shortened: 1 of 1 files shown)


kurtismullins:
@coolreader18 Can you update this Pull Request's description and add a little comment in the Workflow to explain what is going on for those who are less familiar? 

I assume it is fixing this issue:
Screenshot 2023-10-27 at 11 30 31 AM

Also, not everyone is probably going to be familiar with `git fetch deepen` so it may be worth dropping a link in there to the docs with a quick mention of what it does or why its being used: 
https://git-scm.com/docs/git-fetch#Documentation/git-fetch.txt---deepenltdepthgt

kurtismullins (changes requested):
I left a comment that should have been a review. Marking this accordingly.

kurtismullins (approved):
LGTM

PR #488, opened 2023-10-27
perf(469): Add metric for delayed transaction execution

### LINKED ISSUE #469: Record metric for the amount of time a transaction is blocked
I.e the amount of time a transaction spends waiting in the queue before being executed.
Label each recording with transaction type, database id, and reducer id.

joshua-spacetime:
Closes #469.

This metric tracks the amount of time a reducer spends in the wait queue,
past its scheduled execution time.

# API and ABI

### DIFF
--- a/crates/core/src/host/scheduler.rs
+++ b/crates/core/src/host/scheduler.rs
@@ -6,7 +6,10 @@ use sled::transaction::{ConflictableTransactionError::Abort as TxAbort, Transact
 use spacetimedb_lib::bsatn;
 use spacetimedb_lib::bsatn::ser::BsatnError;
 use tokio::sync::mpsc;
+use tokio_util::time::delay_queue::Expired;
 use tokio_util::time::{delay_queue, DelayQueue};
+
+use crate::db::db_metrics::DB_METRICS;
 
 use super::module_host::WeakModuleHost;
 use super::{ModuleHost, ReducerArgs, ReducerCallError, Timestamp};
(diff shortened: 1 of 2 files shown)


crates/core/src/host/scheduler.rs
```diff
@@ -258,14 +259,23 @@ impl SchedulerActor {
...
         let Some(scheduled) = self.db.get(id.0.to_le_bytes()).unwrap() else {
             return;
         };
         let scheduled: ScheduledReducer = bsatn::from_slice(&scheduled).unwrap();
+        // Note, we are only tracking the time a reducer spends delayed in the queue.
+        // This does not account for any time the executing thread spends blocked by the os.
+        DB_METRICS
+            .scheduled_reducer_delay_ns
+            .with_label_values(&module_host.info().address, &scheduled.reducer)
+            .observe(delay.as_nanos() as f64);
```
kulakowski, line 278: I know this got talked about, and see that you're using nanos, but I lost the logic for the final conclusion.

It seems to me there are a couple ways to represent this.

One is that you just take the integral number of nanos, stick that in an f64, and move on with your life. So 10 billion nanos would be (10 * 1000 * 1000 * 1000) as f64, and 1 nano would be 1 as f64

The other is that you take treat the integral number of nanos as a fixed point number of seconds, and then convert that to a floating point number of seconds. So 10 billion nanos would be 10 as f64, and 1 nano would be .000000001 as f64.

The former seems much easier (just an `as` away)

Thinking through the fidelity of the f64 number of nanoseconds representation. 2^52 nanos is 52 days. So any single latency we measure is very unlikely to be impacted, but aggregates are (for long enough running modules).

I'm assuming then that this is fine, but wanted to type out the thought.

kulakowski (approved):

PR #491, opened 2023-10-28
spacetimedb-data-structures: new crate + nstr!

Centril:
# Description of Changes

Adds a new crate `spacetimedb-data-structures` where any general purpose data structures we invent can live.
The first data structure to be added to this is a simple `NStr<N>` type which is the UTF-8 equivalent of `[u8; N]`.
This type can be constructed via the `nstr!("foo")` macro which accepts a string literal and produces `NStr<3>`.

# API and ABI

No breakage.

1

### DIFF
--- a/crates/data-structures/src/nstr.rs
+++ b/crates/data-structures/src/nstr.rs
@@ -8,6 +8,9 @@ use core::{fmt, ops::Deref, str};
 
 /// A UTF-8 string of known length `N`.
 ///
+/// The notion of length is that of the standard library's `String` type.
+/// That is, the length is in bytes, not chars or graphemes.
+///
 /// It holds that `size_of::<NStr<N>>() == N`
 /// but `&NStr<N>` is always word sized.
 #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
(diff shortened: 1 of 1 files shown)


crates/data-structures/src/nstr.rs
```diff
@@ -0,0 +1,77 @@
+//! Provides UTF-8 strings of compile-time-known lengths.
+//!
+//! The strings can be constructed with `nstr!(string_literal)`
+//! producing a type `NStr<N>` where `const N: usize`.
+//! An example would be `nstr!("spacetime"): NStr<9>`.
+
+use core::{fmt, ops::Deref, str};
+
+/// A UTF-8 string of known length `N`.
```
kulakowski, line 9: I'd want to be precise about what notion of length this is in the docs. I'd probably try to use language similar to what rust std uses when clarifying, eg in https://doc.rust-lang.org/std/string/struct.String.html#method.len

crates/data-structures/src/nstr.rs
```diff
@@ -0,0 +1,77 @@
...
+///
+/// # Example
+///
+/// ```
+/// use spacetimedb_data_structures::{nstr, nstr::NStr};
+/// let s: NStr<3> = nstr!("foo");
+/// assert_eq!(&*s, "foo");
+/// assert_eq!(s.len(), 3);
+/// assert_eq!(format!("{s}"), "foo");
+/// assert_eq!(format!("{s:?}"), "foo");
```
kulakowski, line 48: I think you're missing the closing ```.

I'll also wonder if in the future we want more surface area for creating these a la the const_format crate. I can imagine wanting to stamp out a bunch of nstrs for system tables or something that vary only in some integer constant, say, and getting to write `nconcat!("something {i}")` or whatever could be nice.

kulakowski (approved):

PR #495, opened 2023-10-30
Refactor commit encoding / decoding

kim:
# Description of Changes

Use `sats::{BufReader, BufWriter}` for decoding / encoding of `Commit`
and associated types. This makes `decode` fallible (which is quite
desirable, instead of panicking).

As the `DecodeError` from sats is fairly sparse, also add some context
about where exactly decoding failed.

Also add documentation and property test.

# API and ABI

1

### DIFF
--- a/crates/core/src/db/messages/commit.rs
+++ b/crates/core/src/db/messages/commit.rs
@@ -40,13 +40,17 @@ pub struct Commit {
 mod arbitrary {
     use super::*;
 
+    // [`Hash`] is defined in `lib`, so we can't have an [`Arbitrary`] impl for
+    // it just yet due to orphan rules.
     pub fn parent_commit_hash() -> impl Strategy<Value = Option<Hash>> {
         any::<Option<[u8; 32]>>().prop_map(|maybe_hash| maybe_hash.map(|data| Hash { data }))
     }
 
+    // Custom strategy to apply an upper bound on the number of [`Transaction`]s
+    // generated.
+    //
+    // We only ever commit a single transaction in practice.
     pub fn transactions() -> impl Strategy<Value = Vec<Arc<Transaction>>> {
-        // We only ever commit a single transaction, but for the sake of test
-        // coverage let's generate like two handful.
         prop::collection::vec(any::<Arc<Transaction>>(), 1..8)
     }
 }
--- a/crates/core/src/db/messages/commit.rs
+++ b/crates/core/src/db/messages/commit.rs
@@ -111,9 +111,7 @@ impl Commit {
     }
 
     pub fn encoded_len(&self) -> usize {
-        let mut count = 0;
-
-        count += 1; // tag for option
+        let mut count = 1; // tag for option
         if let Some(hash) = self.parent_commit_hash {
             count += hash.data.len();
         }
--- a/crates/core/src/db/messages/commit.rs
+++ b/crates/core/src/db/messages/commit.rs
@@ -152,15 +152,27 @@ mod tests {
     use super::*;
 
     proptest! {
-        // Generating arbitrary commits is quite slow, so limit this test to
-        // just a few cases.
+        // Generating arbitrary commits is quite slow, so limit to just a few
+        // cases.
+        //
+        // Note that this config applies to all `#[test]`s within the enclosing
+        // `proptest!`.
         #![proptest_config(ProptestConfig::with_cases(64))]
+
+
         #[test]
         fn prop_commit_encoding_roundtrip(commit in any::<Commit>()) {
             let mut buf = Vec::new();
             commit.encode(&mut buf);
             let decoded = Commit::decode(&mut buf.as_slice()).unwrap();
             prop_assert_eq!(commit, decoded)
+        }
+
+        #[test]
+        fn prop_encoded_len_is_encoded_len(commit in any::<Commit>()) {
+            let mut buf = Vec::new();
+            commit.encode(&mut buf);
+            prop_assert_eq!(buf.len(), commit.encoded_len())
         }
     }
 }
--- a/crates/core/src/db/messages/commit.rs
+++ b/crates/core/src/db/messages/commit.rs
@@ -5,6 +5,11 @@ use std::{fmt, sync::Arc};
 use super::transaction::Transaction;
 use crate::hash::Hash;
 
+#[cfg(test)]
+use proptest::prelude::*;
+#[cfg(test)]
+use proptest_derive::Arbitrary;
+
 /// A commit is one record in the write-ahead log.
 ///
 /// Encoding:
@@ -12,10 +17,12 @@ use crate::hash::Hash;
 /// ```text
 /// [0u8 | 1u8<hash(32)>]<commit_offset(8)><min_tx_offset<8>[<transaction>...]
 /// ```
-#[derive(Debug, Default)]
+#[derive(Debug, Default, PartialEq)]
+#[cfg_attr(test, derive(Arbitrary))]
 pub struct Commit {
     /// The [`Hash`] over the encoded bytes of the previous commit, or `None` if
     /// it is the very first commit.
+    #[cfg_attr(test, proptest(strategy = "arbitrary::parent_commit_hash()"))]
     pub parent_commit_hash: Option<Hash>,
     /// Counter of all commits in a log.
     pub commit_offset: u64,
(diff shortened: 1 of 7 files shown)


crates/core/src/db/messages/commit.rs
```diff
@@ -91,21 +127,36 @@ impl Commit {
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    proptest! {
+        // Generating arbitrary commits is quite slow, so limit this test to
+        // just a few cases.
+        #![proptest_config(ProptestConfig::with_cases(64))]
```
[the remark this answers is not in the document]
kulakowski: @Centril might know

crates/core/src/db/messages/write.rs
```diff
@@ -1,13 +1,52 @@
...
 }
 
-#[derive(Debug, Copy, Clone)]
+#[cfg(test)]
+mod arbitrary {
+    use super::*;
+
+    pub fn datakey() -> impl Strategy<Value = DataKey> {
+        prop_oneof![
+            prop::collection::vec(any::<u8>(), 0..31).prop_map(DataKey::from_data),
```
kulakowski, line 34: I'd comment on the reason for the two ranges.

crates/core/src/db/messages/write.rs
```diff
@@ -28,60 +67,66 @@ impl Operation {
...
+
+    pub fn encode(&self, writer: &mut impl BufWriter) {
+        let mut flags = 0u8;
+        flags = if self.to_u8() != 0 { flags | 0b1000_0000 } else { flags };
+        writer.put_u8(flags);
+    }
+}
+
+/// Error context for [`Write::decode`].
+enum Context {
```
kulakowski, line 90: It's slightly weird to me that this name is so generic and doesn't mention errors in any way. If that is very idiomatic, I'm fine with it though.

crates/core/src/db/messages/write.rs
```diff
@@ -1,13 +1,52 @@
...
+/// ```text
+/// <flags(1)><set_id(4)><value(1-33)>
+/// ```
+#[derive(Debug, Copy, Clone, PartialEq)]
+#[cfg_attr(test, derive(Arbitrary))]
 pub struct Write {
     pub operation: Operation,
-    pub set_id: u32,
+    pub set_id: u32, // aka table id
+    #[cfg_attr(test, proptest(strategy = "arbitrary::datakey()"))]
```
kulakowski, line 24: I'd also comment about directing this to a hand-written strategy, or in general anywhere that the derive macro doesn't do the right thing on its own.

crates/core/src/db/messages/commit.rs
```diff
@@ -1,81 +1,117 @@
...
-use std::sync::Arc;
 
-// aka "Block" from blockchain, aka RecordBatch, aka TxBatch
-#[derive(Debug)]
+#[cfg(test)]
+use proptest::prelude::*;
+#[cfg(test)]
+use proptest_derive::Arbitrary;
+
+/// A commit is one record in the write-ahead log.
```
kulakowski, line 13: A drive by thought: similarly to how we have a bunch of code that says `set_id` but commented `"this is the table id"`, I would love to settle on saying "write ahead log" everywhere, including data type names etc. Or some other name. But let's pick one!

kulakowski (approved):
Great.

I left a few minor sorts of comments about comments and names.

I'd also like to verify this isn't a performance regression. I don't think this code is a bottleneck for us anymore, but it sure used to be.

Is the restart smoke test failure related?

crates/core/src/db/messages/write.rs
kim: ea62cf8c779c1860207c645a8ff189c14c1b0f9b

crates/core/src/db/messages/write.rs
kim: Well I don't know what's idiomatic, it's just a private type to make it convenient to exploit `anyhow::Context` for providing, well, context. My idea is to add another layer of context with the byte offset in #496.

If you prefer `ErrorContext`, `ErrCtx` or whatever, I'm happy to paint it like this.

crates/core/src/db/messages/write.rs
kim: Is it fine to have those on the `mod arbitrary` intentionally kept close the to type definition? See ea62cf8c779c1860207c645a8ff189c14c1b0f9b

crates/core/src/db/messages/commit.rs
kim: Yeah, I'd love to settle on WAL / write ahead log, maybe even collapsing `CommitLog` and `MessageLog` into a single type. Possibly not for this PR, though?

kim:
> I'd also like to verify this isn't a performance regression. 

Hm, would this entail landing a separate patch with a benchmark before this, so we can track the difference?

crates/core/src/db/messages/commit.rs
kulakowski: Yeah I'd love to have us try to take things that way. I agree on the naming, I think the commit log and message log names are implementation details and that "write ahead log" is the name that reflects the semantics the datastore needs to care about. But yeah not for this PR.

kulakowski:
> > I'd also like to verify this isn't a performance regression.
> 
> Hm, would this entail landing a separate patch with a benchmark before this, so we can track the difference?

I don't think so. I meant regression at the level of in benchmarks / game performance, not in any new functionality or anything. Like if worldgen goes from N seconds to 1.1N seconds, I'd want to look at this again, but if it stays at N seconds, send it.

kim:
benchmarks please

kim:
Let's try this :)

kim:
> Is the restart smoke test failure related?

Looks like it was a flake.

crates/core/src/db/messages/write.rs
```diff
@@ -1,13 +1,57 @@
...
     pub data_key: DataKey,
 }
 
-#[derive(Debug, Copy, Clone)]
+#[cfg(test)]
+mod arbitrary {
+    use super::*;
+
+    // [`DataKey`] is defined in `lib`, so we can't have an [`Arbitrary`] impl
+    // for it just yet due to orphan rules.
```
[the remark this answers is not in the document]
kim: Sure. I tried to keep changes local for this one, but happy to send a follow-up defining `Arbitrary` for `Hash` and `DataKey`. Perhaps that'd even increase probability for folks to write property tests ;)

crates/core/src/db/messages/commit.rs
cloutiertyler: Just adding my two cents here. I agree on coalescing on the name. I avoided WAL because historically that's referred to a log that is eventually compressed/deleted, but I think on balance WAL is probably the best name. We can include a comment about the fact that the WAL should never be deleted.

We should also note that it will include information which is not normally in a WAL, including reducer call event info and that it forms a Merkle DAG between commits.

cloutiertyler: The term `MessageLog` was borrowed from Kafka.

crates/core/src/db/messages/write.rs
```diff
@@ -28,60 +72,66 @@ impl Operation {
...
+            set_id,
+            data_key,
+        })
     }
 
     pub fn encoded_len(&self) -> usize {
-        // 1 for flags, 4 for set_id
-        let mut count = 1 + 4;
+        let mut count = self.operation.encoded_len();
+        count += 4; // set_id
```
cloutiertyler, line 127: This seems wrong. Shouldn't it be 5 based on the operation encoding?

crates/core/src/db/messages/commit.rs
```diff
@@ -91,21 +131,36 @@ impl Commit {
         count
     }
 
-    pub fn encode(&self, bytes: &mut Vec<u8>) {
-        bytes.reserve(self.encoded_len());
```
cloutiertyler, line 95: It's a little unfortunate that we're losing this optimization, but maybe we can improve things by not using a Vec at all but a buffer pool or something in the future.

cloutiertyler (dismissed):
I think there's a bug in `encoded_len` for `Write`.

crates/core/src/db/messages/write.rs
kim: It’s operation.len + 4 = 1+4 = 5, like before no?

crates/core/src/db/messages/commit.rs
kim: Good point. I meant to make sure the caller takes care of this, will add.

crates/core/src/db/messages/commit.rs
kim: Noted to include all of these in the upcoming formalization / design changes doc.

crates/core/src/db/messages/commit.rs
kim: e05f58eedb0466de78537f295594a386fa571420

crates/core/src/db/messages/commit.rs
```diff
@@ -1,81 +1,121 @@
...
 
     pub fn encoded_len(&self) -> usize {
         let mut count = 0;
 
-        if self.parent_commit_hash.is_none() {
-            count += 1;
-        } else {
-            count += 1;
-            count += self.parent_commit_hash.unwrap().data.len();
+        count += 1; // tag for option
```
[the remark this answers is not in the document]
kim: 6b66febcfac7deef09e24729e0a8451481291513

PR #496, opened 2023-10-30
core: Track the number of bytes read when iterating over the WAL

kim:
# Description of Changes

~~Prerequisite for being able to truncate log segments without seeking, e.g. while replaying up until a corrupted record.~~

Above did not turn out to be feasible, but I think the additional error context is worthwhile.

# API and ABI

1

### DIFF
--- a/crates/core/src/db/message_log.rs
+++ b/crates/core/src/db/message_log.rs
@@ -335,9 +335,10 @@ impl TryFrom<SegmentView> for IterSegment {
     type Error = io::Error;
 
     fn try_from(view: SegmentView) -> Result<Self, Self::Error> {
+        let segment = view.offset();
         File::try_from(view)
             .map(BufReader::new)
-            .map(|file| IterSegment { file })
+            .map(|file| IterSegment { segment, read: 0, file })
     }
 }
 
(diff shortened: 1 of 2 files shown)


kim:
Setting to draft until #495 lands. I intend to use this for error context as well.

crates/core/src/db/message_log.rs
```diff
@@ -381,6 +397,8 @@ impl Iterator for IterSegment {
         if let Err(e) = self.read_exact_or_none(&mut buf)? {
             return Some(Err(e));
         }
+        self.read += HEADER_SIZE as u64;
```
kulakowski, line 400: Should this add happen after the first read but before the second? Is failure in the middle visible?

kulakowski (approved):

crates/core/src/db/message_log.rs
kim: Hm right, this doesn't make much sense anymore.

For context: my original idea was to piggyback on this number to reset the log when replaying hits an un-decodable commit -- which almost (but only almost) certainly is the last message in the log. So in this case the byte offset should always fall on a message boundary.

I discarded the idea because it would require iteration to be synchronized with mutation of the log, which is undesirable if we are using those iterators for replication purposes. 

Now, appending to the log while reading concurrently is fine, but truncating is not. In practice, truncation would happen _before_ any readers could possibly obtain a handle (i.e. replaying, new leader term), but I'll need to noodle on it a bit more to find a way to statically ensure that's the case. Thoughts and ideas welcome!

PR #497, opened 2023-10-30
Misc refactoring preparing for multi col indices

Centril:
# Description of Changes

Extracted from https://github.com/clockworklabs/SpacetimeDB/pull/439.

- Some code refactoring
- Some field/method renames and additions.

# API and ABI

2

### DIFF
--- a/crates/core/src/db/datastore/locking_tx_datastore/mod.rs
+++ b/crates/core/src/db/datastore/locking_tx_datastore/mod.rs
@@ -2715,7 +2715,7 @@ mod tests {
             ColRow { table: 3, pos: 2, name: "index_name", ty: AlgebraicType::String, autoinc: false },
             ColRow { table: 3, pos: 3, name: "columns", ty: AlgebraicType::array(AlgebraicType::U32), autoinc: false },
             ColRow { table: 3, pos: 4, name: "is_unique", ty: AlgebraicType::Bool, autoinc: false },
-            ColRow { table: 3, pos: 5, name: "index_type", ty: AlgebraicType::String, autoinc: false },
+            ColRow { table: 3, pos: 5, name: "index_type", ty: AlgebraicType::U8, autoinc: false },
 
             ColRow { table: 4, pos: 0, name: "constraint_id", ty: AlgebraicType::U32, autoinc: true },
             ColRow { table: 4, pos: 1, name: "constraint_name", ty: AlgebraicType::String, autoinc: false },
--- a/crates/core/src/db/datastore/system_tables.rs
+++ b/crates/core/src/db/datastore/system_tables.rs
@@ -305,7 +305,7 @@ pub static ST_COLUMNS_ROW_TYPE: Lazy<ProductType> =
 ///
 /// | index_id | table_id | index_name  | columns | is_unique | index_type |
 /// |----------|----------|-------------|---------|-----------|------------|
-/// | 1        |          | "ix_sample" | [1]     | false     | "Btree"    |
+/// | 1        |          | "ix_sample" | [1]     | false     | "btree"    |
 pub fn st_indexes_schema() -> TableSchema {
     TableSchema {
         table_id: ST_INDEXES_ID,
@@ -350,7 +350,7 @@ pub fn st_indexes_schema() -> TableSchema {
                 table_id: ST_INDEXES_ID,
                 col_id: StIndexFields::IndexType.col_id(),
                 col_name: StIndexFields::IndexType.col_name(),
-                col_type: AlgebraicType::String,
+                col_type: AlgebraicType::U8,
                 is_autoinc: false,
             },
         ],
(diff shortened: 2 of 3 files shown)


crates/lib/src/lib.rs
```diff
@@ -195,16 +196,38 @@ pub struct TypeAlias {
...
+            1 => Ok(IndexType::Hash),
+            _ => Err(()),
+        }
+    }
+}
+
+impl TryFrom<&str> for IndexType {
+    type Error = ();
+    fn try_from(v: &str) -> Result<Self, Self::Error> {
+        match v {
```
kulakowski, line 223: I'm concerned about committing to the exact spelling here. Where did BTree vs btree come from?

crates/core/src/db/datastore/system_tables.rs
```diff
@@ -109,10 +110,15 @@ macro_rules! st_fields_enum {
...
+
+        impl From<$ty_name> for NonEmpty<ColId> {
+            fn from(value: $ty_name) -> Self {
+                NonEmpty::new(value.col_id())
+            }
+        }
     }
 }
 
-// WARNING: In order to keep a stable schema, don't change the discriminant of the fields
```
kulakowski, line 115: Why did we lose these comments?

crates/core/src/db/datastore/locking_tx_datastore/mod.rs
```diff
@@ -2486,50 +2566,38 @@ mod tests {
...
-            .sorted_by_key(|x| x.constraint_id)
-            .collect::<Vec<_>>();
-
+        let constraints_rows = query.scan_st_constraints()?;
         #[rustfmt::skip]
         assert_eq!(
             constraints_rows,
             vec![
-                StConstraintRow{ constraint_id: 5.into(), constraint_name: "ct_columns_table_id".to_string(), kind:  ColumnIndexAttribute::INDEXED, table_id: 1.into(), columns: vec![0.into()] },
+                constraint_row(5, "ct_columns_table_id", 1, ColumnIndexAttribute::INDEXED, &[0]),
```
kulakowski, line 2600: I mildly don't like that this loses some clarity at call sites, 5 and 1 are like, just numbers now.

crates/core/src/db/datastore/locking_tx_datastore/mod.rs
```diff
@@ -339,6 +334,85 @@ impl SequencesState {
     }
 }
 
+/// Utility to query the system tables and return their concrete table row
+pub struct SystemTableQuery<'a> {
+    db: &'a Inner,
+    ctx: &'a ExecutionContext<'a>,
+}
+
+impl<'a> SystemTableQuery<'a> {
```
kulakowski, line 343: Love these.

cloutiertyler, line 343: I think these should come with a comment about their performance. Each of these functions requires materializing the full table into a vec.

crates/core/src/db/datastore/locking_tx_datastore/mod.rs
cloutiertyler: I provided the same feedback when @Centril proposes this in another PR. This explicitness is by design. IMO we should revert this change and the other changes like it.

crates/lib/src/lib.rs
mamcx: This is for `StIndexRow` where we need to parse the index type.

crates/core/src/db/datastore/system_tables.rs
mamcx: I see them inside the macro `st_fields_enum`

crates/core/src/db/datastore/locking_tx_datastore/mod.rs
mamcx: So far this is only for simplify the testing. Probably add a `todo` to change them when we have a better mem-arch?

crates/core/src/db/datastore/locking_tx_datastore/mod.rs
```diff
@@ -2503,165 +2632,99 @@ mod tests {
...
-                column_row(5, 1, "kind", AlgebraicType::U8, false),
-                column_row(5, 2, "epoch", AlgebraicType::U128, false),
-            ]
-        );
-        let index_rows = datastore
-            .iter_mut_tx(&ExecutionContext::default(), &tx, ST_INDEXES_ID)?
-            .map(|x| StIndexRow::try_from(x.view()).unwrap().to_owned())
-            .sorted_by_key(|x| x.index_id)
-            .collect::<Vec<_>>();
+        assert_eq!(query.scan_st_columns()?, map_array([
```
kulakowski, line 2647: It seems like there's two kinds of changes to all this test code.

One is fairly mechanical changes due to the renames etc that make up the rest of this PR.

The other is a substantial change to how the datastore exposes system tables to test code.

Is there any other kind of change that happens in this test code?

mamcx (commented):
LGTM

mamcx (approved):
LGTM

kulakowski (approved):

PR #499, opened 2023-10-30
perf(471): Add duration metrics for wasm abi calls

### LINKED ISSUE #471: Record duration metric for wasm abi calls
This should include the time spent serializing rows from the database and copying into wasm memory.
Label each recording with transaction type, database address, reducer name, and syscall.

joshua-spacetime:
Closes #471.

# Description of Changes

# API and ABI

### DIFF
--- a/crates/core/src/host/mod.rs
+++ b/crates/core/src/host/mod.rs
@@ -4,6 +4,7 @@ use anyhow::Context;
 use bytes::Bytes;
 use bytestring::ByteString;
 use derive_more::Display;
+use enum_map::Enum;
 use spacetimedb_lib::de::serde::SeedWrapper;
 use spacetimedb_lib::de::DeserializeSeed;
 use spacetimedb_lib::{bsatn, Hash, Identity};
@@ -168,11 +169,19 @@ impl EnergyMonitor for NullEnergyMonitor {
     }
 }
 
-#[derive(Debug, Display)]
-pub enum SysCall {
-    Insert,
+/// Tags for each call that a `WasmInstanceEnv` can make.
+#[derive(Debug, Display, Enum)]
+pub enum AbiCall {
+    CancelReducer,
+    ConsoleLog,
+    CreateIndex,
     DeleteByColEq,
+    GetTableId,
+    Insert,
     IterByColEq,
-    IterChunks,
-    IterChunksFiltered,
+    IterDrop,
+    IterNext,
+    IterStart,
+    IterStartFiltered,
+    ScheduleReducer,
 }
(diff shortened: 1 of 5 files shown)


crates/core/src/host/mod.rs
```diff
@@ -166,3 +167,12 @@ impl EnergyMonitor for NullEnergyMonitor {
     ) {
     }
 }
+
+#[derive(Debug, Display)]
+pub enum SysCall {
```
kulakowski, line 172: I'd love to reconcile this with https://github.com/clockworklabs/SpacetimeDB/blob/c566dbae3dc759b3f712cd0fc4a5f172d32306ed/crates/core/src/host/wasm_common/instrumentation.rs#L84

crates/core/src/util/typed_prometheus.rs
```diff
@@ -121,6 +121,7 @@ impl_prometheusvalue_string!(
     Identity,
     Address,
     TransactionType,
+    SysCall,
```
kulakowski, line 124: We've waffled a lot on calling this syscalls vs abicalls vs... what. And even in this PR you say SysCall here but wasm_abi_call elsewhere.

Let's pick one.

kulakowski (commented):
The rest lgtm tho

crates/core/src/host/wasmer/wasm_instance_env.rs
```diff
@@ -436,15 +460,36 @@ impl WasmInstanceEnv {
...
-            let count = caller
-                .data()
-                .instance_env
-                .delete_by_col_eq(&ctx, table_id.into(), col_id.into(), &value)?;
-            Ok(count.get())
-        })
+        let ctx = caller.data().reducer_context();
+        let txn_type = &ctx.txn_type();
+        let db = &ctx.database();
+        let reducer = &ctx.reducer_name().unwrap_or_default().to_owned();
```
kulakowski, line 466: The reducer name copies are a bit unfortunate. Are they necessary? Is this what the static metric stuff is about?

https://crates.io/crates/prometheus-static-metric

This is definitely addressable via a followup if so.

kulakowski (approved):

crates/core/src/host/wasmer/wasm_instance_env.rs
kulakowski: Ah right, I was thinking about just the abi call names. Yeah, sounds good.

crates/core/src/util/typed_prometheus.rs
cloutiertyler: I would argue it's really a syscall where SpacetimeDB is the OS. Happy to resolve the bikeshedding in a separate PR though.

PR #500, opened 2023-10-30
Added multi column index usage to the rust-wasm-test module

cloutiertyler:
# Description of Changes
Just adds usage of the multi column index macro feature

# API and ABI

`


kulakowski:
Did this tickle the very brittle standalone integration test? It has a comment about needing to run things in order, and the failing assertion is that the number of lines in some log output is off.

kim:
Yes I think the test is failing because creating the index writes to the module log.

kulakowski (approved):
The introduction of the `Point` table itself seems fine, so approving, because the bug fix seems trivial and unrelated.

PR #501, opened 2023-10-30
perf(472): Add metric for index seeks and keys scanned

### LINKED ISSUE #472: Record metric for the number of index keys scanned
Label each recording with transaction type, database address, reducer name, and table id.

joshua-spacetime:
Closes #472.

# Description of Changes

# API and ABI

### DIFF
--- a/crates/core/src/db/datastore/locking_tx_datastore/mod.rs
+++ b/crates/core/src/db/datastore/locking_tx_datastore/mod.rs
@@ -1501,7 +1501,7 @@ impl Inner {
                 inserted_rows,
                 committed_rows: self.committed_state.index_seek(table_id, &cols, &range),
                 committed_state: &self.committed_state,
-                committed_rows_fetched: 0,
+                num_committed_rows_fetched: 0,
             }))
         } else {
             // Either the current transaction has not modified this table, or the table is not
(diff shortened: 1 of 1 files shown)


crates/core/src/db/datastore/locking_tx_datastore/btree_index.rs
```diff
@@ -62,14 +62,26 @@ impl Iterator for BTreeIndexIter<'_> {
 /// [BTreeIndex]
 pub struct BTreeIndexRangeIter<'a> {
     range_iter: btree_set::Range<'a, IndexKey>,
+    keys_scanned: u64,
```
cloutiertyler, line 65: Nit: I would do `num_keys_scanned` or `keys_scanned_count`, so that it's clear that it's a count and not some identifier.

cloutiertyler (approved):
Approved with comment.

crates/core/src/db/datastore/locking_tx_datastore/mod.rs
```diff
@@ -1495,11 +1495,13 @@ impl Inner {
             // The current transaction has modified this table, and the table is indexed.
             let tx_state = self.tx_state.as_ref().unwrap();
             Ok(IterByColRange::Index(IndexSeekIterInner {
+                ctx,
                 table_id: *table_id,
                 tx_state,
                 inserted_rows,
                 committed_rows: self.committed_state.index_seek(table_id, &cols, &range),
                 committed_state: &self.committed_state,
+                committed_rows_fetched: 0,
```
cloutiertyler, line 1504: Nit: similar num fetched comment.

PR #504, opened 2023-10-31
Add `delete_by_{non-unique column}` method to `#[spacetimedb(table)]` macro

gefjon:
Macro generation of `filter_by_column` on non-unique columns is currently broken; this will stay in draft until that's fixed, as it depends on the same code path.

# Description of Changes

- Refactor `bindings::query::delete_by_field` to require `FilterableValue` and return `u32` count of deleted rows.
- Add `bindings::query::delete_by_unique_field` with the previous behavior of `delete_by_field`, i.e. requiring `UniqueValue` and returning `bool` with true = deleted. This is implemented in terms of `delete_by_field`.
- Macro-generated unique filter funcs call `delete_by_unique_field` rather than `delete_by_field`.
- The macro now generates non-unique delete functions alongside the existing non-unique filter functions for non-unique fields with primitive types.

Note that this PR does not include any host-side changes; the diff is localized to the `bindings` and `bindings-macro` crates.

# API and ABI

This is an additive-only change to the module API, and leaves the module ABI untouched.

2 - a reviewer should be prepared to verify that the host-side implementation of `delete_by_col_eq` correctly handles the existence of multiple matching rows.

### DIFF
--- a/crates/bindings-macro/src/lib.rs
+++ b/crates/bindings-macro/src/lib.rs
@@ -654,18 +654,22 @@ fn spacetimedb_tabletype_impl(item: syn::DeriveInput) -> syn::Result<TokenStream
 
         unique_delete_funcs.push(quote! {
             #vis fn #delete_func_ident(#column_ident: &#column_type) -> bool {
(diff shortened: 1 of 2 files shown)


gefjon:
Superseded by #859

PR #515, opened 2023-11-01
Expose `delete_by_rel`

Centril:
# Description of Changes

Fixes https://github.com/orgs/clockworklabs/projects/22/views/9?pane=issue&itemId=43239806.

# API and ABI

A `delete_by_rel` function is added to the ABI/API and also exposed via `TableType::delete`.
This is optimized for deleting a single row.

2

### DIFF
--- a/crates/core/src/host/instance_env.rs
+++ b/crates/core/src/host/instance_env.rs
@@ -182,6 +182,8 @@ impl InstanceEnv {
 
         // Find the row schema using it to decode a vector of product values.
         let row_ty = stdb.row_schema_for_table(tx, table_id)?;
+        // `TableType::delete` cares about a single element
+        // so in that case we can avoid the allocation by using `smallvec`.
         let relation = ProductValue::decode_smallvec(&row_ty, &mut &*relation).map_err(NodesError::DecodeRow)?;
 
         // Delete them and return how many we deleted.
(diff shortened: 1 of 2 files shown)


gefjon (commented):
Could you document your rationale for your usage of SmallVec? I'm not opposed to it, but I'd like it to be explained in the code.

gefjon (approved):

crates/core/src/host/instance_env.rs
```diff
@@ -170,6 +170,24 @@ impl InstanceEnv {
         NonZeroU32::new(count).ok_or(NodesError::ColumnValueNotFound)
     }
 
+    /// Deletes all rows in the table identified by `table_id`
+    /// where the rows match one in `relation`
+    /// which is a bsatn encoding of `Vec<ProductValue>`.
+    ///
+    /// Returns an error if no rows were deleted.
+    #[tracing::instrument(skip(self, relation))]
+    pub fn delete_by_rel(&self, table_id: TableId, relation: &[u8]) -> Result<u32, NodesError> {
```
kulakowski, line 179: If we keep it, I think the optimization favoring 1 row should be documented at least here.

kulakowski: I'm fine with this in doc comments for now. My thought process in the long run is that I would hope a query optimizer could be written against a documented interface. Like, I have that target audience for documentation in mind when making that original comment.

PR #516, opened 2023-11-01
Add OptIn serialization attribute for C# generated types

aasoni:
# Description of Changes

Generate the OptIn Json.NET attribute (https://www.newtonsoft.com/json/help/html/JsonObjectAttributeOptIn.htm) for autogenerated C# types.
This fixes a bug where user generated properties in partial classes get serialized to json as well. This caused a deserialization issue on the server, when a user generated property was the last one, resulting in the following error:
image

# API and ABI

1


cloutiertyler (approved):
LGTM

jdetter:
Yeah this is a known issue with json serialization, this should be a good fix for now 👍

PR #517, opened 2023-11-01
lib: Provide `Arbitrary` impls for `Hash` and `DataKey`

kim:
# Description of Changes

Allows to remove custom strategies for the commit log types (in core), and may become useful for future property testing efforts.

**This is stacked on top of #495. Do not merge until after #495 is landed.**

# API and ABI

1

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4388,6 +4388,7 @@ dependencies = [
  "insta",
  "itertools 0.11.0",
  "proptest",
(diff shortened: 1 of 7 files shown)


kim:
Turns out `#[cfg_attr(any(test, feature = "proptest"), ..)]` is needed to allow using the impls in _this_ crate's tests as well as in dependants.

Issue #506, opened 2023-10-31
Identities should be printed in hexadecimal in SQL


joshua-spacetime:
If you have a table like this:

```rust
pub struct MyTable {
  id: Identity
}
```

If you do a SQL query on this, the table will be printed like:

```rust
id 
----------------------------------------------------------------------------------------------------------------------------------------------------
[143, 144, 204, 148, 169, 150, 141, 246, 210, 151, 186, 166, 101, 3, 168, 147, 177, 195, 75, 11, 0, 38, 138, 81, 72, 152, 221, 149, 75, 13, 162, 11]
[143, 144, 204, 148, 169, 150, 141, 246, 210, 151, 186, 166, 101, 3, 168, 147, 177, 195, 75, 11, 0, 38, 138, 81, 72, 152, 221, 149, 75, 13, 162, 11]
[143, 144, 204, 148, 169, 150, 141, 246, 210, 151, 186, 166, 101, 3, 168, 147, 177, 195, 75, 11, 0, 38, 138, 81, 72, 152, 221, 149, 75, 13, 162, 11]
```

We just want these to be printed in hex (or maybe base64?) instead of as a byte array.

PR #521, opened 2023-11-03
Format binary data as hex in sql/satn

mamcx:
# Description of Changes

# API and ABI

1


crates/lib/tests/serde.rs
```diff
@@ -26,14 +56,16 @@ fn test_json_mappings() {
...
     ]);
     let data = r#"
 {
     "foo": 42,
     "bar": "404040FFFF0A48656C6C6F",
     "baz": ["heyyyyyy", "hooo"],
     "quux": { "Hash": "54a3e6d2b0959deaacf102292b1cbd6fcbb8cf237f73306e27ed82c3153878aa" },
-    "and_peggy": { "some": 3.141592653589793238426 }
+    "and_peggy": { "some": 3.141592653589793238426 },
+    "identity": ["0000000000000000000000000000000000000000000000000000000000000000"]
```
cloutiertyler, line 68: Why is this an array of strings and not just a string?

crates/lib/tests/snapshots/serde__json_mappings.snap
```diff
@@ -1,18 +1,21 @@
...
     foo = 42,
-    bar = [64, 64, 64, 255, 255, 10, 72, 101, 108, 108, 111],
+    bar = 404040ffff0a48656c6c6f,
     baz = [
         "heyyyyyy",
         "hooo",
     ],
     quux = (
-        Hash = [84, 163, 230, 210, 176, 149, 157, 234, 172, 241, 2, 41, 43, 28, 189, 111, 203, 184, 207, 35, 127, 115, 48, 110, 39, 237, 130, 195, 21, 56, 120, 170],
+        Hash = 54a3e6d2b0959deaacf102292b1cbd6fcbb8cf237f73306e27ed82c3153878aa,
```
cloutiertyler, line 13: We probably want to prefix hexadecimal values with `0x` in SATN.

crates/sats/src/satn.rs
```diff
@@ -290,7 +290,7 @@ impl<'a, 'b> ser::Serializer for SatnFormatter<'a, 'b> {
     }
 
     fn serialize_bytes(mut self, v: &[u8]) -> Result<Self::Ok, Self::Error> {
-        write!(self, "{:?}", v)
+        write!(self, "{}", hex::encode(v))
```
cloutiertyler, line 293: We probably want to prefix hexadecimal byte arrays with `0x` I think.

crates/core/src/control_db/tests.rs
```diff
@@ -82,3 +85,46 @@ fn test_domain() -> anyhow::Result<()> {
...
+        program_bytes_address: Hash::ZERO,
+        publisher_address: Some(Address::zero()),
+    };
+
+    let id = cdb.insert_database(db.clone())?;
+
+    let dbs = cdb.get_databases()?;
+
+    assert_eq!(dbs.len(), 1);
+    assert_eq!(dbs[0].id, id);
```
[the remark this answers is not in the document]
cloutiertyler: I agree with Josh, not sure what is going on here.

cloutiertyler (dismissed):

crates/lib/tests/serde.rs
mamcx: Because the internal representation of identity is `AlgebraicType::product([("__identity_bytes", AlgebraicType::bytes())])`

Keeping the same structure here means avoids having a discrepancy between `bsatn, json, satn`.

joshua-spacetime (approved):

Centril (approved):

kulakowski:
@cloutiertyler I believe your requested change is in, if you don't mind poking it along

mamcx:
This should close https://github.com/clockworklabs/SpacetimeDB/issues/506

PR #526, opened 2023-11-06
core: Add truncation to message log

kim:
# Description of Changes

Adds a method `reset_to(offset)` to `MessageLog`, which allows to truncate the log to the given message offset.

Note that this currently requires to iterate over the last segment until the message offset is found -- we may in the future maintain an offset index (a la Kafka) to be able to seek closer to the desired message offset.

`reset_to` subsumes the unused `reset_hard` method, and is itself not yet used (this will be added in a later patch).

**Note** The first commit is submitted separately in #496 

# API and ABI

3

### DIFF
--- a/crates/core/src/db/message_log.rs
+++ b/crates/core/src/db/message_log.rs
@@ -1,6 +1,6 @@
 use std::{
     fs::{self, File},
-    io::{self, BufReader, BufWriter, Read, Write},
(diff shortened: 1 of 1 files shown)


kulakowski (commented):
If there's no rush I'd like to finish the review after #496 lands

PR #527, opened 2023-11-06
core: Replay database prefix

kim:
# Description of Changes

When replaying the log, stop if a commit is encountered which cannot be decoded. As this implies a corrupted log, truncate the log to the known-good prefix.

**Depends on #526**

# API and ABI

3 -- the risk of truncating the commit log wrongly is data loss.

### DIFF
--- a/crates/core/src/db/relational_db.rs
+++ b/crates/core/src/db/relational_db.rs
@@ -5,23 +5,24 @@ use super::message_log::MessageLog;
 use super::ostorage::memory_object_db::MemoryObjectDB;
 use super::relational_operators::Relation;
 use crate::address::Address;
-use crate::db::commit_log;
 use crate::db::db_metrics::DB_METRICS;
 use crate::db::messages::commit::Commit;
 use crate::db::ostorage::hashmap_object_db::HashMapObjectDB;
 use crate::db::ostorage::ObjectDB;
 use crate::error::{DBError, DatabaseError, IndexError, TableError};
 use crate::execution_context::ExecutionContext;
+use crate::hash::Hash;
+use anyhow::anyhow;
 use fs2::FileExt;
 use nonempty::NonEmpty;
 use spacetimedb_lib::PrimaryKey;
 use spacetimedb_primitives::{ColId, ColumnIndexAttribute, IndexId, SequenceId, TableId};
 use spacetimedb_sats::data_key::ToDataKey;
 use spacetimedb_sats::db::def::{IndexDef, SequenceDef, TableDef, TableSchema};
-use spacetimedb_sats::hash::Hash;
 use spacetimedb_sats::{AlgebraicType, AlgebraicValue, ProductType, ProductValue};
 use std::borrow::Cow;
 use std::fs::{create_dir_all, File};
+use std::io;
 use std::ops::RangeBounds;
 use std::path::Path;
 use std::sync::{Arc, Mutex};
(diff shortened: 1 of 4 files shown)


crates/core/src/db/relational_db.rs
```diff
@@ -95,10 +96,18 @@ impl RelationalDB {
...
+                let mut message_log = message_log.lock().unwrap();
                 let max_offset = message_log.open_segment_max_offset;
-                for commit in commit_log::Iter::from(message_log.segments()) {
-                    let commit = commit?;
+                'commits: for commit in commit_log::Iter::from(message_log.segments()) {
+                    let commit = match commit {
+                        Ok(commit) => commit,
+                        Err(e) if e.kind() == io::ErrorKind::InvalidData => {
+                            log::warn!("Corrupt commit after offset {last_commit_offset:?}");
+                            message_log.reset_to(last_commit_offset.unwrap_or(0))?;
```
kim, line 106: Since this is destructive, I'm wondering if we want to add more safeguards against data loss here. And if so, which ones.

We could, for example, only reset the log if the offset falls into the last segment, and/or is within a certain distance from `open_segment_max_offset` (for some definition of "certain distance").

crates/core/src/db/relational_db.rs
```diff
@@ -95,33 +95,56 @@ impl RelationalDB {
...
-                            log::debug!(
-                                "[{}] Loaded {}% ({}/{})",
-                                address,
-                                percentage,
-                                transaction_offset,
-                                max_offset
-                            );
+                let total_segments = message_log.total_segments();
+
+                'replay: for (segment_offset, segment) in message_log.segments().enumerate() {
```
kim, line 102: Note that I intend to reduce the rightward drift here and move the replay logic elsewhere.
I left it here for this patch to preserve the side-by-side diff for review.

crates/core/src/db/relational_db.rs
```diff
@@ -1449,4 +1473,61 @@ mod tests {
 
     //     Ok(())
     // }
+
+    #[test]
+    fn test_replay_corrupted_log() -> ResultTest<()> {
```
kim, line 1478: It would be nice to also test the "safeguard" condition (log corrupted in the middle), but the default segment size in 1GiB. Punting on making this configurable at this point.

crates/core/src/db/relational_db.rs
```diff
@@ -1449,4 +1473,61 @@ mod tests {
...
+            Ok(balance)
+        }
+
+        // Deposit some coins, creating 10 transactions.
+        for _ in 0..10 {
+            db.with_auto_commit(&ctx, |tx| db.insert(tx, table_id, product![AlgebraicValue::U64(0)]))?;
+        }
+        assert_eq!(55, balance(&ctx, &db, table_id)?);
+
+        // Close the db an pop a byte from the end of the message log.
```
kim, line 1514: It would be nice to also test the "safeguard" condition (log corrupted in the middle), but the default segment size in 1GiB. Punting on making this configurable at this point.

kim: Actually, that’s not true but obscured by the various test helpers. Will add a test before landing.

crates/core/src/db/relational_db.rs
```diff
@@ -95,33 +95,56 @@ impl RelationalDB {
...
+                let total_segments = message_log.total_segments();
+
+                'replay: for (segment_offset, segment) in message_log.segments().enumerate() {
+                    for commit in segment.try_into_iter_commits()? {
+                        let commit = match commit {
+                            Ok(commit) => commit,
+                            Err(e) if e.kind() == io::ErrorKind::InvalidData => {
+                                log::warn!("Corrupt commit after offset {last_commit_offset:?}");
+
+                                // We expect that partial writes can occur at
```
kulakowski, line 109: Maybe for later: it feels like the RelationalDB should not be the place that knows this fact.

kulakowski (approved):
lgtm pending Joshua's feedback.

crates/core/src/db/relational_db.rs
kim: I agree -- I have more patches on the stack to move most of the replaying into `CommitLog`. But wanted to keep the diff small for this one.

crates/core/src/db/relational_db.rs
```diff
@@ -95,33 +95,56 @@ impl RelationalDB {
...
+                                // allows to continue appending to the log
+                                // without hitting above condition.
+                                message_log.reset_to(last_commit_offset.unwrap_or(0))?;
+                                break 'replay;
+                            }
+                            Err(e) => return Err(e.into()),
+                        };
+                        segment_index += 1;
+                        last_hash = commit.parent_commit_hash;
+                        last_commit_offset = Some(commit.commit_offset);
```
kim, line 127: Turns out we should've checked that this is contiguous.

crates/core/src/db/relational_db.rs
```diff
@@ -95,33 +95,56 @@ impl RelationalDB {
...
-                                transaction_offset,
-                                max_offset
-                            );
+                let total_segments = message_log.total_segments();
+
+                'replay: for (segment_offset, segment) in message_log.segments().enumerate() {
+                    for commit in segment.try_into_iter_commits()? {
+                        let commit = match commit {
+                            Ok(commit) => commit,
+                            Err(e) if e.kind() == io::ErrorKind::InvalidData => {
```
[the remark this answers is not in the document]
kim: I did consider this, but it turns out we need to handle more cases here:

1. The `Commit` envelope decodes successfully, but is actually invalid
2. The `Commit` envelope fails to decode
3. The file is EOF _before_ commit decoding even starts

It's not obvious how to unify this into a single error type (at least not without some major refactoring), and I'm not even sure it would be of value -- we would not gain any exhaustiveness checking.

I hope to have improved the upstream error reporting (i.e. what comes out of `RelationalDB::open`) in the recent commits, though.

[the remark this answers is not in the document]
kulakowski: This line of thought ok with you Joshua?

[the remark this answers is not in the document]
kim: I have formulated this differently in #554, which is a large enough diff to submit separately:

https://github.com/clockworklabs/SpacetimeDB/pull/554/files#diff-37c945f95a02769137efa2981278629b8e8aa25d24504f0c4210624b552dda93R473

This keeps enough state to keep handling of the different cases close to each other. It may also become general iteration at some point, but I’m not sure.

PR #528, opened 2023-11-06
Add the `slim_slice` data-structures

Centril:
# Description of Changes

Adds the data structures, with constraints `len <= u32::MAX` to `spacetimedb_data_structures::slim_slice`:
- `SlimSliceBox`
- `SlimSmallSliceBox`
- `SlimStrBox`
- `SlimSlice`
- `SlimSliceMut`
- `SlimStr`
- `SlimStrMut`
- `SlimNonEmpty`

The tests have been run through miri.

# API and ABI

Not API/ABI breaking. Entirely new code.

3

The code is not deep-reaching nor complex, but it involves a lot of code that is unsafe in shallow ways.

### DIFF
--- a/crates/data-structures/src/slim_slice.rs
+++ b/crates/data-structures/src/slim_slice.rs
@@ -68,7 +68,7 @@ use core::{
     mem,
     mem::ManuallyDrop,
     ops::{Deref, DerefMut},
-    ptr::{read, slice_from_raw_parts_mut, NonNull},
+    ptr::{slice_from_raw_parts_mut, NonNull},
     slice,
     str::{from_utf8_unchecked, from_utf8_unchecked_mut},
 };


kulakowski:
Is this going to be used on the host, the module, or both?

crates/data-structures/src/slim_slice.rs
```diff
@@ -0,0 +1,1801 @@
...
+    }
+}
+impl PartialEq<str> for SlimStr<'_> {
+    #[inline]
+    fn eq(&self, other: &str) -> bool {
+        self.deref() == other
+    }
+}
+
+impl Ord for SlimStr<'_> {
```
kulakowski, line 1264: It's hard to say without knowing how all this is used, but the provided set of impls feels a bit ad hoc. I suppose it can change later.

eg why not `impl Ord for SlimSlice<T> where T: Ord` etc.

kulakowski (commented):
Ok. I have minor organizational qualms, but nothing that's a breaking change in the future.

All the unsafe stuff looks good to me but Noa should probably double check.

My biggest concern, though, is one design choice, and that's introducing a bunch of places that can panic. It's part of the reason I asked about whether this would be used in modules, as we've seen overflows there, so I don't think it's outside the realm of plausibility. I've had this sort of thing in the back of my head since noticing that deserializing SATS trusts the provided size hints.

crates/data-structures/Cargo.toml
```diff
@@ -5,4 +5,10 @@ edition = "2021"
 license-file = "LICENSE"
```
cloutiertyler, line 5: Could you please add an empty license file.

crates/data-structures/src/slim_slice.rs
```diff
@@ -0,0 +1,1934 @@
...
+/// The safety requirements must also be the same.
+pub unsafe trait SafelyExchangeable<T> {}
+
+#[inline]
+fn transmute_safely_exchangeable<T, U: SafelyExchangeable<T>>(x: T) -> U {
+    let x = ManuallyDrop::new(x);
+    // SAFETY: Caller has proven that `x` is valid for `U`.
+    // The reference is also properly aligned
+    // and valid for reads by virtue of `&*x`.
+    unsafe { read(&*x as *const T as *const U) }
```
coolreader18, line 169: Any reason not to use transmute_copy? Seems more in-line with the operation being done here.

crates/data-structures/src/slim_slice.rs
```diff
@@ -0,0 +1,1934 @@
...
+        Vec::from(self).into_iter()
+    }
+}
+
+/// A wrapper to achieve `FromIterator<T> for Result<SlimSliceBox<T>, Vec<T>>`.
+///
+/// We cannot do this directly due to orphan rules.
+pub struct SlimSliceBoxCollected<T> {
+    /// The result of `from_iter`.
+    pub inner: Result<SlimSliceBox<T>, LenTooLong<Vec<T>>>,
```
coolreader18, line 567: This feels... inelegant. Maybe we could have a `try_from_iter() -> Result<Self, LenTooLong<Vec<T>>>` function on SlimSliceBox + a `try_collect_slim_slice_box(self)` extension trait method? Idk. Not blocking this on a better solution, this is fine, but just feels like it could be better.

crates/data-structures/src/slim_slice.rs
```diff
@@ -0,0 +1,1934 @@
...
+        // SAFETY: ^-- satisfies `len <= u32::MAX`.
+        Ok(unsafe { Self::from_slice_unchecked(slice) })
+    }
+}
+
+/// Converts `&[T]` into the slim limited version.
+///
+/// Panics when `slice.len() > u32::MAX`.
+#[inline]
+pub fn slice<T>(s: &[T]) -> SlimSlice<'_, T> {
```
coolreader18, line 1154: Maybe `from_slice`? And then it gets used as `slim_slice::from_slice`

coolreader18 (approved):
Some nits, but nothing blocking; overall lgtm

kulakowski (changes requested):
Hitting the request change box just for the LICENSE thing

crates/data-structures/src/slim_slice.rs
gefjon: Why is this not just `mem::transmute(x)`?

kulakowski:
> @kulakowski Addressed your comments + added a basic "slim small slice box" type that is a cross of `SmallVec` and `SlimSliceBox`. This new type does not have tests yet, but I'll add them in a follow-up PR if you don't mind.

Otherwise looks good to me, plus Noa's comments.

PR #529, opened 2023-11-06
perf(507): Add disk and memory usage metrics

### LINKED ISSUE #507: Add metrics for memory and disk usage
Disk usage comes from one of 3 things: the message log, object storage, and logs. We should have a metric for each of these as well as complete free/used disk space on the node coming directly from the OS. The latter metric will be a cluster level metric whereas the former ones will be tied to a specific database. Note also that these metrics will be in the form of a single statistic.

Similarly we want to track a node's free and used memory as reported by the OS. Again, this will be reported as a single statistic for the cluster.

joshua-spacetime:
Closes #507.

# Description of Changes

# API and ABI

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -157,6 +157,7 @@ sqllogictest = "0.17"
 strum = { version = "0.25.0", features = ["derive"] }
 syn = { version = "2", features = ["full", "extra-traits"] }
 syntect = { version = "5.0.0", default-features = false, features = ["default-fancy"] }
+sys-info = "0.9.1"
 tabled = "0.14.0"
 tar = "0.4"
 tempdir = "0.3.7"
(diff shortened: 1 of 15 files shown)


Cargo.toml
```diff
@@ -157,6 +157,7 @@ sqllogictest = "0.17"
 strum = { version = "0.25.0", features = ["derive"] }
 syn = { version = "2", features = ["full", "extra-traits"] }
 syntect = { version = "5.0.0", default-features = false, features = ["default-fancy"] }
+sys-info = "0.9.1"
```
kulakowski, line 160: I'm pretty sure this one is going to be fine because it's a pretty trivial dependency but worth mentioning:

Any time we land something that has an OS dependency (like this does), I'd like the person landing it to be a little heads up in case there's a problem on whichever of linux/macos/windows they don't run on.

crates/standalone/src/lib.rs
```diff
@@ -163,6 +163,8 @@ fn get_key_path(env: &str) -> Option<PathBuf> {
 #[async_trait]
 impl spacetimedb_client_api::NodeDelegate for StandaloneEnv {
     fn gather_metrics(&self) -> Vec<prometheus::proto::MetricFamily> {
+        // Note, we update certain metrics such as disk usage on demand.
+        self.db_inst_ctx_controller.update_metrics();
```
kulakowski, line 167: I'm guessing this thing is called slowly enough that `stat`ing a bunch of stuff is not going to be a big deal, but I'd like to make sure that's an ok assumption.

kulakowski (approved):

PR #532, opened 2023-11-07
refactor(531): Label transaction count metric with boolean flag

### LINKED ISSUE #531: Label transaction count metric with boolean "committed" flag
Instead of having two distinct metrics for commits and rollbacks, we should have a single metric and label each recording with a boolean flag representing whether that transaction was committed or rolled back.

joshua-spacetime:
Fixes #531.

Instead of having two distinct metrics for commits and rollbacks, this patch replaces them with a single metric plus a boolean label representing whether the transaction was committed or rolled back.

# Description of Changes

# API and ABI

### DIFF
--- a/crates/core/src/util/typed_prometheus.rs
+++ b/crates/core/src/util/typed_prometheus.rs
@@ -122,6 +122,7 @@ impl_prometheusvalue_string!(
     Address,
     TransactionType,
     AbiCall,
+    bool,
     u8,
     u16,
     u32,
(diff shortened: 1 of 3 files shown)


crates/core/src/util/typed_prometheus.rs
```diff
@@ -122,6 +122,7 @@ impl_prometheusvalue_string!(
     Address,
     TransactionType,
     AbiCall,
+    bool,
```
kulakowski, line 125: Out of curiosity, suppose in the future we wanted to track something with 3 or more states, which in rust you'd naturally represent as `enum X { A, B, C }`. How we would do that with prometheus?

I'm asking partly just out of curiosity, but also wanting to understand if we'd regret this change when we have a third variant pop up.

kulakowski (approved):

PR #534, opened 2023-11-07
Capture logs in integration tests

RReverser:
# Description of Changes

This helps debug failures.

# API and ABI

1


crates/testing/tests/standalone_integration_test.rs
```diff
@@ -2,10 +2,21 @@ use serde_json::Value;
 use serial_test::serial;
 use spacetimedb_testing::modules::{CompilationMode, CompiledModule, DEFAULT_CONFIG};
 
+fn init() {
+    let _ = env_logger::builder()
+        .parse_filters(
+            "spacetimedb=trace,spacetimedb_client_api=trace,spacetimedb_lib=trace,spacetimedb_standalone=trace",
+        )
+        .is_test(true)
+        .try_init();
```
gefjon, line 11:
```suggestion
        // `try_init` and ignore failures to continue if a logger is already registered.
        // This allows us to call `init` at the start of every test without a `once_cell` or similar.
        .try_init();
```

gefjon (commented):
Complete the PR template and I'll approve.

RReverser:
> Complete the PR template and I'll approve.

I wonder - why did we add that field? It seems kinda useless in majority of cases as most PRs are simple changes.

gefjon:
> I wonder - why did we add that field? It seems kinda useless in majority of cases as most PRs are simple changes (and if they're not, they should be split up into smaller ones anyway).

Maybe you weren't here for that meeting, but we added it in the hopes of making review easier by encouraging people to think about the complexity of their PRs. We had an endemic problem with people creating huge PRs that were difficult to review because they touched many parts of the codebase. That made reviewing more difficult even for the easy PRs, because it was hard to tell what you were getting into when you sat down to do a review.

We also had problems with PRs that appeared simple, but actually had non-obvious interactions with parts of the codebase that hadn't been directly edited. The PR authors were generally aware of those interactions, but the reviewers often were not, so we got some bugs that weren't caught by reviewers who didn't know where to look.

I pushed for the new template field because I figure on easy PRs it takes a few seconds to type a 1 or 2 in there, and on more complex PRs it's valuable. Adoption has been slow, so I'm trying to bully people into building the habit by refusing to approve any PRs that don't have it filled in, even for trivial PRs like this one.

RReverser:
(Anyway, I did finish template so just waiting for the review.)

PR #538, opened 2023-11-08
Add integration test for the query macro

mamcx:
# Description of Changes

The integration test for `rust-wasm-test` contains the use of the `query!` macro but was not called anywhere.

This later will be used to verify the use of `multi-column` indexes on this macro when https://github.com/clockworklabs/SpacetimeDB/pull/267 lands.

# API and ABI

1

### DIFF
--- a/modules/rust-wasm-test/src/lib.rs
+++ b/modules/rust-wasm-test/src/lib.rs
@@ -127,7 +127,7 @@ pub fn test(ctx: ReducerContext, arg: TestAlias, arg2: TestB, arg3: TestC) -> an
         });
     }
 
-    let multi_row_count = query!(|row: Point| row.x >= 0 && row.x <= 200).count();
+    let multi_row_count = query!(|row: Point| row.x >= 0 && row.y <= 200).count();
 
     log::info!("Row count filtered by multi-column condition: {:?}", multi_row_count);
 
--- a/modules/rust-wasm-test/src/lib.rs
+++ b/modules/rust-wasm-test/src/lib.rs
@@ -1,6 +1,7 @@
 #![allow(clippy::disallowed_names)]
 use spacetimedb::{
-    delete_by_col_eq, query, spacetimedb, AlgebraicValue, Deserialize, ReducerContext, SpacetimeType, Timestamp,
+    delete_by_col_eq, query, spacetimedb, AlgebraicValue, Deserialize, ReducerContext, SpacetimeType, TableType,
+    Timestamp,
 };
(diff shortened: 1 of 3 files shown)


modules/rust-wasm-test/src/lib.rs
```diff
@@ -117,10 +118,37 @@ pub fn test(ctx: ReducerContext, arg: TestAlias, arg2: TestB, arg3: TestC) -> an
...
+
+    log::info!("Row count filtered by multi-column condition: {:?}", multi_row_count);
+
     log::info!("END");
     Ok(())
 }
 
+#[spacetimedb(reducer)]
+pub fn test_query(ctx: ReducerContext) -> anyhow::Result<()> {
+    test(
```
RReverser, line 140: Do you need both to be reducers? If you want to just invoke this one, then maybe `test` itself shouldn't have `spacetimedb(reducer)`. Alternatively, just invoke `test` itself from standalone_integration_test.

mamcx: I don't understand the purpose of having `test` and not using it elsewhere, I think we speculate it just was to confirm it compiles. If that is the case having both makes some sense.

modules/rust-wasm-test/src/lib.rs
```diff
@@ -117,10 +118,37 @@ pub fn test(ctx: ReducerContext, arg: TestAlias, arg2: TestB, arg3: TestC) -> an
...
+    log::info!("MultiColumn");
+
+    for i in 0i64..1000 {
+        Point::insert(Point {
+            x: i + arg.x as i64,
+            y: i + arg.y as i64,
+        });
+    }
+
+    let multi_row_count = query!(|row: Point| row.x >= 0 && row.x <= 200).count();
```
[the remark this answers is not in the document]
mamcx: Oh yeah, good catch

modules/rust-wasm-test/src/lib.rs
RReverser: Yeah but then you can probably just call `test` directly from your test runner, just pass non-empty `args` maybe?

mamcx: Done

kulakowski: You happy with this @RReverser ?

RReverser: Yup, thanks.

kulakowski (approved):

PR #543, opened 2023-11-08
CI for Publish Crates.io Dry Run

jdetter:
# Description of Changes

 - Adds CI action for doing a dry-run publish to crates.io to restore a bit of my sanity. During the release of 0.7.3 we were not able to publish to crates.io because of several issues, these issues are still present in master and they are described here: https://github.com/clockworklabs/SpacetimeDB/actions/runs/6805346253/job/18504668802?pr=543
  - After this PR is approved we should make this a required CI check so that we don't get into a state where its impossible to publish.
 
corresponding GitHub label.


jdetter:
Follow up item after approval: Make this CI test required in order for PRs to merge

joshua-spacetime (approved):

kim (approved):

jdetter:
This doesn't work because we do a publish dry-run on each of our packages. If any of our packages has an API change then it will report an error because the version which is on crates.io doesn't have the new changes (the changes only exist locally).

Maybe there's another way to do this, but closing this for now 🤷

PR #544, opened 2023-11-09
Desparse Tar Files When Building

jdetter:
# Description of Changes

 - When we are creating tar files right now we are creating tar files which contain a sparse binary file (just meaning that it has holes in it). This is not supported by the `tar` library that we're using in the `spacetime upgrade` command and I couldn't find another library that was better (plus we just sent out a new release and I don't want to change it).

Github Issue: https://github.com/alexcrichton/tar-rs/issues/295

**Note: This is required for the next release otherwise spacetime upgrade will not work**

You can test this PR by running `spacetime upgrade`, the upgrade from 0.7.2 to 0.7.3 works now because of this PR. You can also check the CLI builds here (the .tar.gz files should not be sparse): <https://cloud.digitalocean.com/spaces/spacetimedb-client-binaries?path=release%2Fsparse-test%2F>

You can test a client binary with this program (generated with chatgpt):

First, you'll need to include the necessary dependencies in your `Cargo.toml` file:

```toml
[dependencies]
tar = "0.4.38"
flate2 = "1.0.22"
```

Now, here's a simple Rust program that accomplishes this:

```rust
use std::env;
use std::fs::File;
use std::path::Path;
use tar::Archive;
use flate2::read::GzDecoder;

fn main() -> Result<(), std::io::Error> {
    // Get the command-line arguments
    let args: Vec<String> = env::args().collect();

    // Check for the required argument
    if args.len() != 2 {
        eprintln!("Usage: {} <TAR_FILE>", args[0]);
        std::process::exit(1);
    }

    // The second argument is the file path for the .tar.gz file
    let tar_gz_path = &args[1];

    // Open the file in read-only mode
    let tar_gz = File::open(Path::new(tar_gz_path))?;

    // Create a decoder that can handle .gz files
    let tar = GzDecoder::new(tar_gz);

    // Create a new archive with the decoded contents
    let mut archive = Archive::new(tar);

    // Unpack the contents to the current directory
    archive.unpack(".")?;

    Ok(())
}
```

To use this application:

1. Include the dependencies in your `Cargo.toml`.
2. Save the Rust code to a file, for example, `extract_tar.rs`.
3. Compile the program with `cargo build --release`.
4. Run the program using `cargo run --release -- <path_to_your_tar_file>`.

corresponding GitHub label.

### DIFF
--- a/.github/workflows/linux-build.yml
+++ b/.github/workflows/linux-build.yml
@@ -21,7 +21,7 @@ jobs:
           export PATH=$HOME/.cargo/bin:$PATH
           cargo build --release -p spacetimedb-cli
           mkdir build
-          cp -Sv target/release/spacetime build/spacetime
+          cp --sparse=never target/release/spacetime build/spacetime
           cd build && tar -czf spacetime.linux-amd64.tar.gz spacetime
           rm spacetime
 
@@ -52,7 +52,7 @@ jobs:
           export PATH=$HOME/.cargo/bin:$PATH
           cargo build --release -p spacetimedb-cli
           mkdir build
-          cp -Sv target/release/spacetime build/spacetime
+          cp --sparse=never target/release/spacetime build/spacetime
           cd build && tar -czf spacetime.linux-arm64.tar.gz spacetime
           rm spacetime
 
--- a/.github/workflows/linux-build.yml
+++ b/.github/workflows/linux-build.yml
@@ -21,7 +21,7 @@ jobs:
           export PATH=$HOME/.cargo/bin:$PATH
           cargo build --release -p spacetimedb-cli
           mkdir build
-          mv target/release/spacetime build/spacetime
+          cp -Sv target/release/spacetime build/spacetime
           cd build && tar -czf spacetime.linux-amd64.tar.gz spacetime
           rm spacetime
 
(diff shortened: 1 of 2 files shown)


.github/workflows/linux-build.yml
```diff
@@ -21,7 +21,7 @@ jobs:
           export PATH=$HOME/.cargo/bin:$PATH
           cargo build --release -p spacetimedb-cli
           mkdir build
-          mv target/release/spacetime build/spacetime
+          cp -Sv target/release/spacetime build/spacetime
```
jdetter, line 24: copy, but prevent the output file from being a sparse file

kulakowski:
Does anything have to happen for the windows build? And could we add comments to the `cp`/`dd` lines saying why?

Otherwise this works locally for me and makes sense, awesome that you were able to track this down!

jdetter:
> Does anything have to happen for the windows build? And could we add comments to the `cp`/`dd` lines saying why?
> 
> Otherwise this works locally for me and makes sense, awesome that you were able to track this down!

No, the reason being that we don't actually zip the windows build (which maybe we should because its quite large). We only gzip the mac and linux builds.

kulakowski (approved):
Sounds good, it would be nice to have those comments saying why.

PR #546, opened 2023-11-09
Disable latest docker tag

drogus:
# Description of Changes

This PR disables tagging with `latest` tag

corresponding GitHub label.


jdetter (approved):
Ok let's give this a try 👍

PR #547, opened 2023-11-09
delete_by_col_eq: optimize with smallvec for 1 row case

Centril:
# Description of Changes

I noticed that `delete_by_col_eq` is mostly called by `delete_by_field` which has `debug_assert!(count <= 1);`.
This suggests that we do an unnecessary allocation in most cases.
We can get rid of that by using a `SmallVec<[RowId; 1]>` so that in the `count = 1` case, we don't need to allocate.

None.

1

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4342,6 +4342,7 @@ dependencies = [
  "sha1",
  "slab",
  "sled",
+ "smallvec",
  "spacetimedb-client-api-messages",
  "spacetimedb-lib",
  "spacetimedb-primitives",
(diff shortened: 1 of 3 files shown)


kulakowski:
benchmarks please

gefjon (approved):
This looks reasonable, and I appreciate the comment explaining why `SmallVec` is useful here. We don't seem to have benchmarks for deletes, so I can't say whether this is a performance win or how much of one it is, but intuition says avoiding an allocation in the common path is a win.

kulakowski:
> This looks reasonable, and I appreciate the comment explaining why `SmallVec` is useful here. We don't seem to have benchmarks for deletes, so I can't say whether this is a performance win or how much of one it is, but intuition says avoiding an allocation in the common path is a win.

I suspect as well but would like to include the "...at the expense of branches" in the train of thought.

kulakowski (approved):
just needs `cargo fmt`

PR #556, opened 2023-11-13
Fix out of order subscription updates

coolreader18:
# Description of Changes

corresponding GitHub label.


kulakowski (commented):
Is there any way to write a test for this?

jdetter:
> Is there any way to write a test for this?

Yes, but it would require an SDK testsuite. We could technically use the rust sdk to write the test, although it is pretty hard to exercise this issue. This really only happens when the server is overloaded (typically during playtests)

kulakowski:
Gotcha. I'd like to see this approved by you (or someone) in any case before picking it to the 0.7.4-beta branch.

crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -21,73 +21,56 @@ use spacetimedb_lib::Identity;
 use tokio::sync::mpsc;
 
 #[derive(Debug)]
-enum ModuleSubscriptionCommand {
+enum Command {
```
kulakowski, line 24: I think a comment on `Command` (or on the mpsc queue of them) describing the ordering invariants we have to maintain would be appropriate.

kulakowski (approved):
LGTM but please put the requested comment in.

crates/core/src/subscription/module_subscription_actor.rs
coolreader18: There's not really any ordering invariants, they were just getting processed out-of-order of being sent because I had put 2 queues for some reason. Like, if you did `for _ in 0..100 { broadcast_commit_event(); }; update_subscription()` the actor could process the subscription update before getting through the whole backlog of commit events.

kulakowski: Yes, the fact that they have to be one queue is exactly what you should describe in a comment in this code please.

kulakowski: Thanks!

kulakowski (approved):

PR #558, opened 2023-11-14
Fixes client API breaks introduced by #460

cloutiertyler:
# Description of Changes

This fixes API breaking changes introduced by #460

corresponding GitHub label.

1


kulakowski (approved):

kim (approved):
Quite unfortunate

PR #561, opened 2023-11-14
Move 'constraints' markers to primitive and fix incompatibility with deploy

mamcx:
# Description of Changes

Refactor the definition of `constraints` into the `primitives` crate, removing duplications of logic and fix bug when

```rust
#[spacetimedb(table)]
struct Foo {
    #[primarykey]
    id: u32,
    #[unique]
    #[autoinc]
    other: u32
}
```

gives error `Multiple primary columns defined for table`, which shows up in the deployment of `spacetimedb.com`

This changes the `bits` stored in the `st_constraints` system table.

1


crates/primitives/src/attr.rs
```diff
@@ -1,5 +1,59 @@
...
+//!
+//! ### UNIQUE:
+//!
+//!  One or many columns with enforced uniqueness using an auto-generate index.
+//!
+//! ### IDENTITY:
+//!
+//! A column that is made up of values generated by the database using a sequence.
+//!
+//!  Differs from `PRIMARY_KEY` in that its values are managed by the database and usually cannot be modified.
```
mamcx, line 15: NOTE: Currently `IDENTITY` implies `UNIQUE` but that is wrong: https://en.wikipedia.org/wiki/Identity_column.

However, to change this need we review how others are using this.

modules/rust-wasm-test/src/lib.rs
```diff
@@ -52,6 +52,15 @@ pub struct Point {
     y: i64,
 }
 
+// Test we can compile multiple constraints
+#[spacetimedb(table)]
```
mamcx, line 56: This is added to check it compiles.

crates/primitives/src/attr.rs
```diff
@@ -1,5 +1,59 @@
...
+//! Enforce uniqueness using an auto-generate index.
+//!
+//! Can only be one per-table.
+//!
+//! ## For Columns
+//!
+//! Additionally, is possible to add markers to columns as:
+//!
+//! - AUTO_INC: Auto-generate a sequence
+//! - INDEXED: Auto-generate a non-unique index
```
mamcx, line 30: Note: `AUTO_INC` right now is semantically ambiguous.  it clashes with the purpose of `IDENTITY` (and is possible to have many of it) which is the recommended one when asked for auto-generated numbers. 

I think we should use instead `SEQUENCE` and deprecate the use of this tag.

modules/rust-wasm-test/src/lib.rs
kulakowski: This looks good. You might have to change an integration smoke test which reads specific log lines because another table will add a log line and through the numbers off by one.

kulakowski (commented):
The code change itself looks good to me. I want to make sure we discuss in chat how we are going agree on the semantics you bring up here, before approving. Thanks for the pointers to documentation about it, Mario.

kim (approved):
Thanks! <3 the docs

crates/primitives/src/attr.rs
```diff
@@ -1,5 +1,59 @@
...
+//! - PRIMARY_KEY_AUTO: Make it a `PRIMARY_KEY` + `AUTO_INC`
+//! - PRIMARY_KEY_IDENTITY: Make it a `PRIMARY_KEY` + `IDENTITY`
+//!
+//! NOTE: We have [ConstraintKind] and [AttributeKind] intentionally semi-duplicated because
+//! the first is for the [Constrains] that are per-table and the second is for markers of the column.
 
+/// The assigned constraint for a `Table`
+#[allow(non_camel_case_types)]
+#[derive(Debug, Clone, Copy, Eq, PartialEq, PartialOrd, Ord)]
```
cloutiertyler, line 39: IMO these are all confused. We should just have attributes which are separate from constraints:

IMO these shouldn't even be on the same axis.

We have the following column attributes:
- auto_inc

We should have the following table attributes:
- primary_key e.g. `(col_a, col_b)`
- indexed e.g. `(col_a, col_b)`

We should have the following table constraints:
- unique e.g. `(col_a, col_b)`

These are all mutually exclusive and should not be combined into a single "kind"

Also constraints != attributes!

mamcx: This will be addressed in a separate PR, this one is to solve the immediate issue with deploy.

PR #565, opened 2023-11-15
Simplify contributor check

coolreader18:
Left over change from when I was doing the benchmarks-please stuff, but it was only tangential to that so I stashed it for later.


coolreader18:
Oh wait may as well test the b*nchmarks please org check too, I don't think I ever got around to that.

coolreader18:
Awesome :))

PR #568, opened 2023-11-16
core: Refine panic message when decoding fails during replay

kim:
# Description of Changes

It is significant for forensics to differentiate the source of the row data being decoded. Change the error messages to that end.

1


kim:
I am also tempted to make those regular `Err`s instead of panicking -- the method is already fallible and it would help to be able to determine the commit/transaction offset.

PR #570, opened 2023-11-16
Move lib::{name,recovery} to client-api-messages

coolreader18:
This way, `spacetimedb-lib` is for stuff that's specifically shared between host and module (that's how I've always interpreted it, at least), and less a catch all for anything that might need to be accessed from 2 different crates. lib is a dependency of bindings, and we should want to keep that as small as is reasonable, so that it builds fast and `spacetime init && spacetime build` as an onboarding demo is snappy.

2, moves some files around so could conflict with other PRs but it's not very
likely nor are there huge changes to the moved files otherwise.


kulakowski:
@coolreader18 I'd love to do this in general after 0.8 lands.

kazimuth (approved):
LGTM, I think this all makes sense.

PR #573, opened 2023-11-17
test: Create indexes for restarting tests

kim:
# Description of Changes

Restarting a database with indexes + sequences is more interesting, in particular it could have caught #569

1


kim:
Meh, seems like this doesn't quite work because we'd need to fsync the log.

kim:
Turns out I just can't write bash.

jdetter (approved):
LGTM 👍

PR #576, opened 2023-11-20
refactor(571): Track module queue length with histogram

### LINKED ISSUE #571: Make reducer queue length into a histogram metric
Right now it's a gauge which means we only know the length of the queue at the time prometheus samples it. If it were a histogram, we would know the min and max values in the interval between samplings, as well as the distribution of values among the buckets.

Note that for this metric, we should **not** use the default bucket configuration.

joshua-spacetime:
# Description of Changes

Closes #571.

Before this patch we could only observe queue length values when prometheus sampled the metric.
This is because we were using a gauge to track the metric.
Now we are using a histogram and therefore we can observe the metric whenever it changes, regardless of sampling rate.

corresponding GitHub label.

### DIFF
--- a/crates/core/src/util/lending_pool.rs
+++ b/crates/core/src/util/lending_pool.rs
@@ -9,31 +9,28 @@ use std::sync::Arc;
 use std::task::{Context, Poll};
 
 use parking_lot::Mutex;
-use prometheus::IntGauge;
+use spacetimedb_lib::{Address, Hash, Identity};
 use tokio::sync::{OwnedSemaphorePermit, Semaphore};
 
-use crate::util::prometheus_handle::IntGaugeExt;
+use crate::worker_metrics::WORKER_METRICS;
 
 use super::notify_once::{NotifiedOnce, NotifyOnce};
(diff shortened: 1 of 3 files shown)


crates/core/src/util/lending_pool.rs
```diff
@@ -52,50 +49,38 @@ struct PoolVec<T> {
...
     pub fn request_with_context(
         &self,
-        context: G::Context<'_>,
+        (identity, module_hash, database_address, reducer_symbol): (Identity, Hash, Address, &str),
     ) -> impl Future<Output = Result<LentResource<T>, PoolClosed>> {
         let acq = self.sem.clone().acquire_owned();
         let pool_inner = self.inner.clone();
-        let waiter_guard = self.waiter_gauge.inc(context);
+
+        let queue_len = WORKER_METRICS.instance_queue_length.with_label_values(
```
coolreader18, line 64: this is fine, I guess. it just makes it less easy to pull out/appraise as its own data structure. I might take a look at making this much simpler while still keeping LendingPool generic.

coolreader18 (approved):

PR #577, opened 2023-11-20
Better module backtraces for panics and whatnot

coolreader18:
# Description of Changes

image

If these symbols aren't in the stack, it does no processing

1 - it's pretty self-contained, and backwards-compatible with the existing logs data format

### DIFF
--- a/crates/cli/src/subcommands/logs.rs
+++ b/crates/cli/src/subcommands/logs.rs
@@ -158,6 +158,32 @@ pub struct BacktraceFrame<'a> {
     pub module_name: Option<Cow<'a, str>>,
     #[serde(borrow)]
     pub func_name: Option<Cow<'a, str>>,
+    #[serde(borrow)]
+    pub file: Option<Cow<'a, str>>,
+    pub line: Option<u32>,
+    pub column: Option<u32>,
+    #[serde(default)]
+    pub symbols: Vec<BacktraceFrameSymbol<'a>>,
+    #[serde(default)]
+    pub kind: BacktraceFrameKind,
+}
+
+#[derive(serde::Deserialize, Default)]
+#[serde(rename_all = "lowercase")]
+pub enum BacktraceFrameKind {
+    #[default]
+    Wasm,
+    Js,
+}
+
+#[derive(serde::Deserialize)]
+pub struct BacktraceFrameSymbol<'a> {
+    #[serde(borrow)]
+    pub name: Option<Cow<'a, str>>,
+    #[serde(borrow)]
+    pub file: Option<Cow<'a, str>>,
+    pub line: Option<u32>,
+    pub column: Option<u32>,
 }
 
 #[derive(serde::Serialize)]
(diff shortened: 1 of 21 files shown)


RReverser:
I'd prefer we used something like [color-eyre](https://github.com/eyre-rs/color-eyre) for formatting which already handles panics instead of rolling something custom we'd have to maintain.

kulakowski:
I just want to double check if the parsing and color formatting is happening in module code or not.

crates/cli/src/tasks/mod.rs
```diff
@@ -16,7 +16,7 @@ pub fn build(project_path: &Path, skip_clippy: bool, build_debug: bool) -> anyho
     if !build_debug {
         eprintln!("Optimising module with wasm-opt...");
         let wasm_path_opt = wasm_path.with_extension("opt.wasm");
-        match cmd!("wasm-opt", "-all", "-O2", &wasm_path, "-o", &wasm_path_opt).run() {
+        match cmd!("wasm-opt", "-all", "-O2", &wasm_path, "-g", "-o", &wasm_path_opt).run() {
```
kulakowski, line 19: I'd maybe comment about the `-g` for the debug info being relatively important.

kulakowski (commented):
Also, what's up with the smoketest

crates/cli/src/subcommands/logs.rs
```diff
@@ -165,24 +177,90 @@ pub async fn exec(mut config: Config, args: &ArgMatches) -> Result<(), anyhow::E
...
+            fmt_backtrace(&mut out, trace)?;
         }
 
         line.clear();
     }
 
     Ok(())
 }
+
+// based on fmt::Display impl for wasmtime::WasmBacktrace
```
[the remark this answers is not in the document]
coolreader18: It didn't have color, and it's not serializable or anything so we can't create it here.

coolreader18:
@RReverser does color-eyre panic stuff work in wasm? From what I can tell it doesn't expose the standalone backtrace formatting functionality, and even if it did it only supports the actual native `backtrace::Backtrace` type. (color-backtrace does expose it, but has the same thing with `Backtrace`). We're in a weird situation here cause wasm doesn't have a means to walk the call stack or even access the DWARF symbols itself, so we have to depend on wasmtime to do it for us. So, we can get a `WasmBacktrace`, which no crates like color-eyre can actually consume themselves, and we also have to serialize it to store it in the logs anyway. I chose to do the filtering and formatting on the client side rather than the server side so that there's less overhead in the console_log syscall, and so that possibly important information (e.g. for a weird bug somehow happening in omitted frames) doesn't get lost permanently and can just be configured to be shown in the client.

crates/cli/src/subcommands/logs.rs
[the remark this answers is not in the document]
kulakowski: I meant as a comment checked in

RReverser:
@coolreader18 Does Wasmtime attaching symbols not help with backtrace? If it helps with gdb/lldb, there should be a way to make Rust backtraces use the same symbol resolution so that all these backtrace formatters/tools "just work". 

Another reason I'd really prefer not to do this on module side is that it further diverges support between module languages. With AOT we'll be able to get symbol names for C# functions in Wasm too, which will work correctly work with wasm2wat or debuggers, but still won't provide readable backtraces with this approach because it's doing something Rust-specific only in Rust modules.

Ideally Wasm modules should be as opaque to the host as possible, and treated the same regardless of the language they were compiled in, especially if we add more langs down the line, otherwise we'll end up manually replicating same functionality for each language individually.

coolreader18:
No, wasm has no mechanism to access the call stack or the DWARF section. `backtrace-rs` on wasm32 unconditionally selects a noop implementation. And this isn't rust-specific at all? This would work with any language that generates debug symbols, we'd just need to possibly add a tiny bit of support to the cli to make it display nicer.

RReverser:
> No, wasm has no mechanism to access the call stack or the DWARF section.

Wasm doesn't, but the native stack walker likely will. I can try to look into this myself later.

> And this isn't rust-specific at all?

The changes in `bindings` crate - e.g. `__rust_{begin,end}_short_backtrace` - are by definition Rust-specific, as only Rust modules use it. And then in log formatter you're looking for those `__rust_{begin,end}_short_backtrace` which suggests it would behave differently for Wasm that doesn't have those functions.

coolreader18:
Bringin this one back, I guess. @RReverser you're correct that these changes are rust-specific, but the bindings crate changes are rust-specific *because* the bindings crate is rust-specific. If csharp has similar frames to `__rust_{begin,end}_short_backtrace`, it's a two line change to add those in `cli`.

RReverser:
Yeah I can't say I care much anymore, go ahead.

templates/basic-rs/spacetimedb/Cargo.toml
```diff
@@ -8,6 +8,9 @@ edition = "2024"
 [lib]
 crate-type = ["cdylib"]
 
+[profile.release]
+debug = 1 # include some location information for backtraces
+
```
cloutiertyler, line 13: Shouldn't we do this in all templates? And or make sure that `cargo init` has it? I assume if you don't do this, you get the same backtrace as before (or similar)?

crates/cli/src/subcommands/logs.rs
```diff
@@ -352,6 +366,69 @@ pub async fn exec(mut config: Config, args: &ArgMatches) -> Result<(), anyhow::E
     Ok(())
 }
 
+// based on fmt::Display impl for wasmtime::WasmBacktrace
+// modified to print in color and to skip irrelevant frames
+fn fmt_backtrace<W: WriteColor>(out: &mut W, trace: &[BacktraceFrame<'_>]) -> anyhow::Result<()> {
```
cloutiertyler, line 371: Why do we have this custom thing? Out of curiosity, if this changes the log format, will the website be impacted?

cloutiertyler (approved):
It's a little odd with the magic words, but I dunno, seems fine given that it's backwards compatible. We should follow up with C# and C++ as well.

coolreader18:
I actually disagree with changing the rust name, since `__rust_begin_short_backtrace` is also found in the rust standard library (and so is `__rust_end_short_backtrace`, which we definitely can't change). I think it's reasonable to specialize for specific languages for debugging purposes, since many compiled languages will have symbols akin `__rust_begin_short_backtrace`, and we can't always pick the name.

cloutiertyler:
@coolreader18 That's fair enough. I will have the agent address your concern.

PR #580, opened 2023-11-20
Error for no abi version

coolreader18:
# Description of Changes

helps with the 0.7 -> 0.8 transition

### DIFF
--- a/crates/core/src/host/wasm_common/abi.rs
+++ b/crates/core/src/host/wasm_common/abi.rs
@@ -5,7 +5,7 @@ const MODULE_PREFIX: &str = "spacetime_";
 pub fn determine_spacetime_abi<I>(
     imports: impl IntoIterator<Item = I>,
     get_module: impl Fn(&I) -> &str,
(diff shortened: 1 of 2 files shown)


crates/core/src/host/wasm_common/abi.rs
```diff
@@ -54,4 +48,7 @@ pub enum AbiVersionError {
         got: VersionTuple,
         implements: VersionTuple,
     },
+    // TODO: (by 1.0, maybe) remove the parenthetical from this error message
+    #[error("unable to determine ABI of module (may be on an spacetime version < 0.8)")]
```
kulakowski, line 52: spelling: `a spacetime version`

kulakowski (approved):

crates/core/src/host/wasm_common/abi.rs
coolreader18: oops, had it on auto-merge. well, it'll get fixed before 1.0 :P

PR #587, opened 2023-11-21
Move to .NET 8 WASI support

RReverser:
# Description of Changes

.NET 8 shipped last week with built-in WASI support. It's still marked as experimental because some things - like better import/export handling that we could use - are still being worked on, but it's already better than using the `Wasi.Sdk` package which is now soft-deprecated and has a number of bugs that nobody's going to fix anymore.

This change is pretty big because it tries to move from custom Mono C APIs to native C# marshalling as much as possible, because it's now supported, it's safer, and it's more future-proof. We should be able to eventually remove the `bindings.c` layer altogether once `UnmanagedCallersOnly` and `DllImport` are handled automatically and generate Wasm imports/exports by .NET's PInvoke, which is already available in .NET alpha channel. It also provides better foundation for NativeAOT support where Mono APIs don't exist and where otherwise we'd have to maintain two very different paths for AOT and non-AOT bindings.

The performance and instruction cost of the resulting bundle is more or less the same as before, because it's still using an interpreter bundled together with .NET bytecode, but thanks to various trimming options the Wasm itself is now ~3 times smaller.

I don't think we have the right category for this, but after this change users will need to install .NET 8 to build new SpacetimeDB projects, but we'll automate the rest, e.g. installing the WASI SDK (Clang compiler & C stdlib) as well as Wasi workload for .NET itself.

I'll tentatively mark it as ABI change to draw attention to breakage even though ABI of the generated module is still compatible.

3

### DIFF
--- a/crates/bindings-csharp/Runtime/RawBindings.cs
+++ b/crates/bindings-csharp/Runtime/RawBindings.cs
@@ -10,8 +10,16 @@ using static System.Text.Encoding;
 
 public static partial class RawBindings
 {
+    // For now this must match the name of the `.c` file (`bindings.c`).
+    // In the future C# will allow to specify Wasm import namespace in
+    // `LibraryImport` directly.
     const string StdbNamespace = "bindings";
 
+    // This custom marshaller takes care of checking the status code
+    // returned from the host and throwing an exception if it's not 0.
+    // The only reason it doesn't return `void` is because the C# compiler
+    // doesn't treat `void` as a real type and doesn't allow it to be returned
+    // from custom marshallers, so we resort to an empty struct instead.
     [CustomMarshaller(typeof(CheckedStatus), MarshalMode.ManagedToUnmanagedOut, typeof(StatusMarshaller))]
     static class StatusMarshaller
     {
@@ -68,6 +76,10 @@ public static partial class RawBindings
         private readonly ulong schedule_token;
     }
 
+    // We need custom marshaller for `Buffer` because we return it by value
+    // instead of passing an `out` reference, and C# currently doesn't match
+    // the common Wasm C ABI in that a struct with a single field is supposed
+    // to have the same ABI as the field itself.
     [CustomMarshaller(typeof(Buffer), MarshalMode.Default, typeof(BufferMarshaller))]
     static class BufferMarshaller
     {
@@ -109,10 +121,10 @@ public static partial class RawBindings
         public static bool operator !=(BufferIter left, BufferIter right) => !(left == right);
     }
 
-    [LibraryImport("bindings")]
+    [LibraryImport(StdbNamespace)]
     public static partial CheckedStatus _get_table_id([In] byte[] name, uint name_len, out TableId out_);
 
-    [LibraryImport("bindings")]
+    [LibraryImport(StdbNamespace)]
     public static partial CheckedStatus _create_index(
         [In] byte[] index_name,
         uint index_name_len,
@@ -122,7 +134,7 @@ public static partial class RawBindings
         uint col_len
     );
 
-    [LibraryImport("bindings")]
+    [LibraryImport(StdbNamespace)]
     public static partial CheckedStatus _iter_by_col_eq(
         TableId table_id,
         ColId col_id,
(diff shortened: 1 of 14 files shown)


crates/bindings-csharp/Runtime/bindings.c
```diff
@@ -1,781 +1,231 @@
...
+       (uint32_t id, Buffer caller_identity, Buffer caller_address,
+        uint64_t timestamp, Buffer args),
+       &id, &caller_identity, &caller_address, &timestamp, &args);
 
 // Shims to avoid dependency on WASI in the generated Wasm file.
 
 #include <stdlib.h>
 #include <wasi/api.h>
 
+#pragma clang diagnostic ignored "-Wc2x-extensions"
```
kulakowski, line 120: Maybe comment what c20 feature you're using.

crates/bindings-csharp/Runtime/bindings.c
```diff
@@ -1,781 +1,231 @@
...
-                                                   int32_t arg2,
-                                                   int64_t arg3,
-                                                   int32_t arg4) {
+// For `fd_write`, we need to at least collect and report sum of sizes.
+// If we report size 0, the caller will assume that the write failed and will
+// try again, which will result in an infinite loop.
+int32_t WASI_NAME(fd_write)(__wasi_fd_t fd, const __wasi_ciovec_t* iovs,
+                            size_t iovs_len, __wasi_size_t* retptr0) {
+  for (size_t i = 0; i < iovs_len; i++) {
+    // Note: this will produce ugly broken output, but there's not much we can
```
kulakowski, line 206: Maybe a p2 ticket tracking this?

crates/bindings-csharp/Runtime/Runtime.cs
```diff
@@ -4,131 +4,74 @@ namespace SpacetimeDB;
...
         }
 
         public void Dispose()
         {
-            BufferIterDrop(ref handle);
+            if (!handle.Equals(RawBindings.BufferIter.INVALID))
+            {
+                RawBindings._iter_drop(handle);
+                handle = RawBindings.BufferIter.INVALID;
+                GC.SuppressFinalize(this);
```
kulakowski, line 67: Worth a comment I think.

crates/bindings-csharp/Runtime/RawBindings.cs
```diff
@@ -0,0 +1,216 @@
+namespace SpacetimeDB;
+
+using System;
+// using System.Runtime.CompilerServices;
```
kulakowski, line 4: Can this be deleted?

kulakowski (commented):
Small code comments but looks really great!

The only thing I want is some approval from I guess @jdetter or @cloutiertyler that the download strategy is ok from a product perspective.

crates/bindings-csharp/Runtime/bindings.c
RReverser: We already have one + implementation in a branch I should complete, but I've been waiting for migration to Wasmtime to happen first as it's much easier to do custom WASI on top of it.

RReverser:
> that the download strategy is ok from a product perspective.

FWIW this is the same download strategy the previously used Wasi.Sdk package uses. I'm open to better suggestions, just pointing out that functionally it's not something new.

kulakowski:
@RReverser Let me know if there's anything I can do to help test or land this.

RReverser:
@jdetter wanted to do some more testing after my fixes - we ran into some issues, but they were all with unclear error messages e.g. due to .NET 8 not being installed, .NET requiring sudo on macOS to install workloads and so on.

I fixed those so in principle nothing prevents this from getting merged now, but I'd prefer to wait for @jdetter to double-check that it's looking good.

RReverser:
The main problem is actually that I don't have macOS to test that sudo issue is gone now. @kulakowski if you want to test this, you can try to check out, do `cargo run -- build -p .\modules\spacetimedb-quickstart-cs` locally and see if error messages you get on each step make sense.

kulakowski:
> `cargo run -- build -p .\modules\spacetimedb-quickstart-cs`

Trying now.

RReverser:
> Trying now.

@kulakowski Any luck?

kulakowski:
Xcode has a couple levels, the lowest is essentially just a toolchain, and the new dotnet needed a newer version of the toolchain.

PR #588, opened 2023-11-21
fix(CommitLog): Do not truncate the log on out of order errors

joshua-spacetime:
# Description of Changes

See the main code comment.
I think until we have a better log format, or at the very least, better diagnostic tooling, we shouldn't ignore OutOfOrder errors.
It's not apparent that we'll see such errors in practice due to the the more lenient fsync policy.

corresponding GitHub label.

### DIFF
--- a/crates/core/src/db/commit_log.rs
+++ b/crates/core/src/db/commit_log.rs
@@ -94,16 +94,22 @@ impl CommitLog {
                 match commit {
                     Ok((commit, objects)) => f(commit, &objects)?,
                     Err(ReplayError::Other { source }) => return Err(source.into()),
-
-                    // We expect that partial writes can occur at the end of a
-                    // segment. Trimming the log is, however, only safe if we're
-                    // at the end of the _log_.
+                    // Note that currently the commit offset is just a u64.
+                    // It does not have any additional structure or semantic validation.
+                    // Hence an OutOfOrder error does not imply an out of order write.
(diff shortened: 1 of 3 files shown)


kim (changes requested):
Sorry, I think you misread the code. We cannot actually distinguish between an out-of-order commit and corrupted data -- given the log is not appended to via other means than `append_tx` (which is properly synchronized), all occurrences of this error case will actually be a corrupt segment.

By not truncating, we will render the database unusable for those unfortunate enough.

kim:
I think we can do the following:

1. Check that `Commit::decode` completely consumed its input. If it didn’t, it’s a “false positive”, and we can treat it as data corruption.
2. If the commit offset is not contiguous, also verify the parent hash. This does not prove the integrity of the current commit, but if verification fails it _does_ prove that the current commit is faulty (i.e. treat as data corruption).
3. Check that hash keys are in the object db. If an object is missing, treat as corruption.
4. The remaining case is a non-contiguous commit, while everything else looks fine. Treat as application-level bug.

kim:
Well, 2. Is still ambiguous if we assume data can be corrupted for reasons other than a partial write. We should maybe consider introducing record-level checksums for 0.8, which is considered breaking.

kim:
1. and 3. are implemented in #593 and #594 respectively

kim (approved):

PR #592, opened 2023-11-22
Rename HostType::Wasmtime to just Wasm

coolreader18:
# Description of Changes

Per @kim's comment:

> I’d prefer if we just named the host type Wasm — it seems very unlikely that we’ll support multiple runtime implementations ever. That is, if we gain another variant it will be Csharp or whatever.

That does make sense to me - the decision to name the host type Wasmer/Wasmtime was made about a year ago, in case we wanted to have both at the same time, and we never took advantage of that and I don't think we ever will (and I think we've pretty conclusively settled on wasmtime as our runtime). I was more or less content having it be like that but if other people feel it's worth changing I'm happy for it.

2 - in a similar way to #457 this is sorta theoretically an http API breaking change to the `/publish` route, but that doesn't actually ever get exercised anyway. And it'd be in the same release cycle.

### DIFF
--- a/crates/cli/src/subcommands/publish.rs
+++ b/crates/cli/src/subcommands/publish.rs
@@ -19,8 +19,8 @@ pub fn cli() -> clap::Command {
             Arg::new("host_type")
                 .long("host-type")
                 .short('t')
-                .value_parser(["wasmtime"])
-                .default_value("wasmtime")
+                .value_parser(["wasm"])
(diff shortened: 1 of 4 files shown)


kim (approved):
:+1: from me, obviously

kulakowski (approved):
lgtm. I'm fine with the theoretical risk to the publish route, so good to land now :)

PR #593, opened 2023-11-22
core: Fully consume input when decoding `Commit`

kim:
# Description of Changes

Ensures that message- and commit log agree on record boundaries, minimizing "false positive" decodings.

nada 

1


kulakowski (commented):
lgtm. I'm leaving the approval to Joshua so we can be sure you are aligned on this

PR #594, opened 2023-11-22
core: Load large objects eagerly during log replay

kim:
# Description of Changes

Instead of deferring object storage loads to when they're needed, load them in the `Replay` iterator. This allows to handle log trimming logic uniformly for the situation when a commit happened to be flushed to disk, but object storage didn't.

Nein.

3

Trims the log if the missing object is within the last segment. Previously, missing objects manifested as a panic.

### DIFF
--- a/crates/core/src/db/commit_log.rs
+++ b/crates/core/src/db/commit_log.rs
@@ -2,7 +2,7 @@ use super::{
     datastore::traits::{MutTxDatastore, TxData},
     message_log::{self, MessageLog},
     messages::commit::Commit,
-    ostorage::ObjectDB,
+    ostorage::{memory_object_db::MemoryObjectDB, ObjectDB},
     FsyncPolicy,
 };
 use crate::{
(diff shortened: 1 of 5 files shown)


crates/core/src/db/datastore/locking_tx_datastore/mod.rs
```diff
@@ -1632,11 +1632,7 @@ impl Locking {
             .rows
     }
 
-    pub fn replay_transaction(
-        &self,
-        transaction: &Transaction,
-        odb: Arc<std::sync::Mutex<Box<dyn ObjectDB + Send>>>,
-    ) -> Result<(), DBError> {
+    pub fn replay_transaction(&self, transaction: &Transaction, odb: &dyn ObjectDB) -> Result<(), DBError> {
```
kulakowski, line 1635: I like the removal of the Send bound as a side effect.

crates/core/src/db/commit_log.rs
```diff
@@ -63,8 +72,7 @@ impl CommitLog {
     /// and can thus safely be written to via the resulting [`CommitLogMut`].
     pub fn replay<F>(&self, mut f: F) -> Result<CommitLogMut, DBError>
     where
-        // TODO(kim): `&dyn ObjectDB` should suffice
```
kulakowski, line 66: Ah, so do you :)

kulakowski (approved):

PR #595, opened 2023-11-22
EXPERIMENTAL: Add a crc32 checksum to message log records

kim:
# Description of Changes

Cost is > 0, but would improve confidence in suffix trimming not deleting data inadvertently.

It seems somewhat unlikely that the low-level message log format is subject to change (soon), so perhaps after quantifying the performance impact this could be deployed until it does.

Changes the on-disk format in an incompatible way.

2

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4141,6 +4141,7 @@ dependencies = [
  "bytes",
  "bytestring",
  "clap",
+ "crc32fast",
  "criterion",
  "crossbeam-channel",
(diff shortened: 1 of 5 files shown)


kim:
My (crude) measurements show ~15% degradation in write and ~20% in read throughput. 

As the read path is much slower anyway, and probably more critical, we may want to consider delaying checksum verification until something else indicates a potential corruption issue (e.g. commit ordering appears off).

PR #597, opened 2023-11-24
Fix id semantics of `DatabaseInstance`

kim:
# Description of Changes

The `id` field of `DatabaseInstance` was used both as an (auto-inc) identity column and a revision of the corresponding `Database`. The latter mainly to support publishing with the `--clear` flag while preserving the stable address of the logical database.

However, with a replica count > 1 this made it impossible to determine which replicas are supposed to hold the same prefix, assuming scheduling is asynchronous.

Thus, we introduce new fields on both `DatabaseInstance` and `Database` which designate the _revision_ of either entity. The database revision is incremented on every update (regardless of `--clear` flag), and the instances point to the revision they where scheduled with.

As a side effect, this allows to retire the (distributed) locking scheme used to order schema/module updates.

Also add some commentary and remove unused methods from the `ControlStateReadAccess` trait.

Breaks the internal `client-api` (traits) and control db "schema".

2

### DIFF
--- a/crates/client-api/src/lib.rs
+++ b/crates/client-api/src/lib.rs
@@ -105,9 +105,10 @@ pub trait ControlStateReadAccess {
     fn get_databases(&self) -> spacetimedb::control_db::Result<Vec<Database>>;
 
     // Database instances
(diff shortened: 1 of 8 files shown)


kim:
This makes more sense in a different context, will reopen later

PR #603, opened 2023-11-28
Improve TypeScript generated code (fixes and improvements)

drogus:
# Description of Changes

This is a set of changes improving various parts of TypeScript autogen:

* properly map u64 and i64 values to BigInt
* change the way we handle client db in classes, which is needed for allowing to run multiple SpacetimeDB clients in TypeScript. The new API will allow to do sth like `User.with(client).onInsert(...)`
* change in how reducers and tables are registered - instead of doing it automagically on import, we will now require explicit call to `SpacetimeDBClient.registerReducers` or `.registerTables`. The problem with doing it automagically is that it will sometimes fail. Node will filter out any imports that are unused, so if you import a table without actually using the class constant, but you do a subscribe, it would result in an error. With explicit include it's no longer a problem
* now we will also export constants and/or functions for enum variants. So far using enum variants was cumbersome, cause given a reducer like `foo(user_type: UserType)` where `UserType` is an enum, in javascript you would have to do sth like `FooReducer.call({ tag: "Admin", value: undefined })`, which requires you to know how sum types are constructed in TypeScript. With the changes from the PR you can now do `FooReducer.call(UserType.Admin)`. For an enum variant with value it would be sth like `AnEnum.VariantA("a value")` which will return `{ tag: "VariantA", value: "a value" }`
* I fixed a case where we converted `__identity_bytes` to camel case, which resulted in an error
* changed the API for how reducer callbacks are defined. instead of providing an `args: any[]` array we now list all of the values with proper types. So for example given an `add(u64, u64)` reducer the callback signature for a callback would be sth like `(reducerEvent: ReducerEvent, a: BigInt, b: BigInt)`

These changes will break the TypeScript SDK, so a new release will be needed when this is released.

I think it's a 2, it's not overly complex and it will actually lower some of the complexity in the SDK (for example by being more explicit with registering reducers and tables)


gefjon (commented):
I like all these changes. I'd like to see a corresponding docs PR to update the SDK ref (and quickstart, if necessary) to reflect the new interface. I'd especially like the docs on `Table.with` to be very clear about whether that mutates global state (I believe it doesn't), with an example that describes the behavior of:

```ts
MyTable.onInsert(someCallback)

MyTable.with(conn).onInsert(otherCallback)

MyTable.onInsert(thirdCallback)
```

I also notice that the corresponding TypeScript SDK PR doesn't update `examples/quickstart`. The main way I'll be convinced that this is ready for the big time is when the quickstart client builds and runs.

gefjon (approved):

PR #604, opened 2023-11-28
Hotfix for log replay bug, fix type for constraints column

mamcx:
# Description of Changes

Fix wrong type definition as identified by @kim that cause the smoke test to sometimes fail

1


kulakowski (commented):
Is it possible to write a test which checks that these two different definitions explicitly line up?

kim (approved):
I agree we should have roundtrip tests as a followup.

PR #613, opened 2023-11-29
perf(610): More granular histogram buckets for timing metrics

### LINKED ISSUE #610: Add custom buckets for histogram metrics

joshua-spacetime:
Closes #610.

# Description of Changes

corresponding GitHub label.


crates/core/src/db/db_metrics/mod.rs
```diff
@@ -91,11 +91,17 @@ metrics_group!(
         #[name = spacetime_txn_elapsed_time_sec]
         #[help = "The total elapsed (wall) time of a transaction (in seconds)"]
         #[labels(txn_type: WorkloadType, db: Address, reducer: str)]
+        #[buckets(
```
kulakowski, line 94: I wouldn't mind a future refactor that has something like a

```
fn power_of_tens_buckets(...) -> Vec<f64> {
   ...
}
```

or something to get rid of this duplication.

kulakowski (approved):

PR #627, opened 2023-12-04
Try to fsync message log on drop

kim:
Not great, but potentially TERM will let finalizers run usually.

Seems to fix the zz-* smoketests.

### DIFF
--- a/crates/core/src/db/message_log.rs
+++ b/crates/core/src/db/message_log.rs
@@ -195,6 +195,14 @@ impl std::fmt::Debug for MessageLog {
     }
 }
 
+impl Drop for MessageLog {
+    fn drop(&mut self) {
+        if let Err(e) = self.sync_all() {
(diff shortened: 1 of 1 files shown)


joshua-spacetime (approved):

crates/core/src/db/message_log.rs
```diff
@@ -195,6 +195,14 @@ impl std::fmt::Debug for MessageLog {
     }
 }
 
+impl Drop for MessageLog {
+    fn drop(&mut self) {
+        if let Err(e) = self.sync_all() {
```
joshua-spacetime, line 200: Do we need to do the same with the objectDB?

kim: Maybe -- but sled will try to flush its pagecache on drop already, so it's unclear if it would buy us much.

PR #632, opened 2023-12-06
ci: Tame the doge

### LINKED ISSUE #630: Fix disabled smoketests
As of https://github.com/clockworklabs/SpacetimeDB/pull/629, the following smoke tests have been disabled due to flakiness
```
zz_docker-restart-module
zz_docker-restart-repeating-reducer
zz_docker-restart-sql
```
Each one involves restarting docker and therefore SpacetimeDB as well. Note that the `restart_docker` utility invokes `sleep` as a way of ensuring the necessary database components are ready. This should be replaced with a more principled way of determining when SpacetimeDB is "up".

kim:
# Description of Changes

Incantation to restart containers in a controlled way.

If you use `docker-compose`, you shouldn't. Use `docker compose`.

1

Fixes #630

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,8 +16,10 @@ jobs:
       - name: Start containers
         run: docker compose up -d
       - name: Run smoketests
-        # TODO: Fix flaky tests https://github.com/clockworklabs/SpacetimeDB/issues/630
         run: test/run-smoke-tests.sh --parallel -x bitcraftmini-pretest zz_docker-restart-module zz_docker-restart-repeating-reducer zz_docker-restart-sql
+      # These cannot run in parallel, even though the script tries to handle it
+      - name: Run restarting smoketests
+        run: test/run-smoke-tests.sh zz_docker-restart-module zz_docker-restart-repeating-reducer zz_docker-restart-sql
       - name: Stop containers
         if: always()
         run: docker compose down
--- a/test/lib.include
+++ b/test/lib.include
@@ -74,9 +74,45 @@ fsed() {
 }
 
 restart_docker() {
-	docker-compose stop node
-	docker-compose start node
-	sleep 10
+	# Behold!
+	#
+	# You thought stop/start restarts? How wrong. Restart restarts.
+	docker compose restart
+	# The suspense!
+	#
+	# Wait until compose believes the health probe succeeds.
+	#
+	# The container may decide to recompile, or grab a coffee at crates.io, or
+	# whatever. In any case, restart doesn't mean the server is up yet.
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -16,8 +16,10 @@ jobs:
       - name: Start containers
         run: docker compose up -d
       - name: Run smoketests
-        # TODO: Fix flaky tests https://github.com/clockworklabs/SpacetimeDB/issues/630
         run: test/run-smoke-tests.sh --parallel -x bitcraftmini-pretest zz_docker-restart-module zz_docker-restart-repeating-reducer zz_docker-restart-sql
+      # These cannot run in parallel, obviously
+      - name: Run restarting smoketests
+        run: test/run-smoke-tests.sh zz_docker-restart-module zz_docker-restart-repeating-reducer zz_docker-restart-sql
```
joshua-spacetime, line 22: Were parallel restarts the main issue?

test/lib.include
```diff
@@ -74,9 +74,38 @@ fsed() {
...
+	#
+	# Wait until compose believes the health probe succeeds.
+	#
+	# The container may decide to recompile, or grab a coffee at crates.io, or
+	# whatever. In any case, restart doesn't mean the server is up yet.
+	docker compose up --no-recreate --detach --wait-timeout 60
+	# Belts and suspenders!
+	#
+	# The health probe runs inside the container, but that doesn't mean we can
+	# reach it from outside. Ping until we get through.
```
joshua-spacetime, line 91: How useful is the health check then?

test/lib.include
```diff
@@ -74,9 +74,38 @@ fsed() {
...
+	ping
```
joshua-spacetime, line 92: Am I correct in asserting that the minimal set of changes needed to remove the flakiness from those tests were
1. Not running them in parallel
2. This ping utility

.github/workflows/ci.yml
kim: No, I just fixed that as a drive-by

test/lib.include
kim: Because _sometimes_, potentially related to the phase of the moon, it will be reachable immediately. I should move the `sleep` to the end of the loop.

test/lib.include
kim: Also `restart`. `stop` followed by `start` works only some of the time.

.github/workflows/ci.yml
kim: Well, turns out turns out we need to run them as a separate task, for whatever reason.

PR #636, opened 2023-12-06
refactor(metrics): Move table size metric into lib

joshua-spacetime:
Query optimization will soon need this metric.
Therefore it must be moved out of core and into lib. Eventually we might want to move all metrics into lib.

# Description of Changes

corresponding GitHub label.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4215,6 +4215,8 @@ dependencies = [
  "hex",
  "insta",
  "itertools 0.11.0",
+ "once_cell",
+ "prometheus",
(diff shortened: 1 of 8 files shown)


kulakowski (approved):
Re crate layout, I know @coolreader18 has expressed wanting to move some code around too, between crates. I don't know the best way exactly to set that up, but I think it's about time for us to actually design it out (like, write down that things with X property or needed for Y have to go in Z crate, and the code within it is subject to C constraint because...) and then commit to it within a release cycle.

PR #637, opened 2023-12-06
chore(vm): Pass database address to the compiler

joshua-spacetime:
The database address will be needed to optimize joins. In particular it will be used to look up table size metrics.

# Description of Changes

corresponding GitHub label.


crates/core/src/sql/compiler.rs
```diff
@@ -248,7 +248,7 @@ fn compile_drop(name: String, kind: DbType, table_access: StAccess) -> Result<Cr
 }
 
 /// Compiles a `SQL` clause
-fn compile_statement(statement: SqlAst) -> Result<CrudExpr, PlanError> {
+fn compile_statement(db: &RelationalDB, statement: SqlAst) -> Result<CrudExpr, PlanError> {
```
cloutiertyler, line 251: Are we sure we'd like to pass the entire DB here if we're just using the address? Is the idea that it will eventually be needed when the metrics are in the DB instead of prometheus?

joshua-spacetime: We can definitely get away with just the address for now, but yes we will need the db once we start storing a full set of statistics for query optimization.

PR #639, opened 2023-12-07
refactor(subscriptions): Support returning either side of a join

joshua-spacetime:
Before this patch we made reference to the "lhs" and the "rhs" of an index join, in the context of incremental evaluation.
This patch further generalizes the incremental evaluation of index joins, by replacing such references with either the probe side or the index side. This is in preparation for adding more types of incremental evaluation plans.

# Description of Changes

corresponding GitHub label.

### DIFF
--- a/crates/core/src/subscription/subscription.rs
+++ b/crates/core/src/subscription/subscription.rs
@@ -36,7 +36,7 @@ use crate::db::db_metrics::{DB_METRICS, MAX_QUERY_CPU_TIME};
 use crate::error::{DBError, SubscriptionError};
 use crate::execution_context::{ExecutionContext, WorkloadType};
 use crate::sql::query_debug_info::QueryDebugInfo;
-use crate::subscription::query::{run_query, OP_TYPE_FIELD_NAME};
+use crate::subscription::query::{run_query, to_mem_table_with_op_type, OP_TYPE_FIELD_NAME};
 use crate::{
     client::{ClientActorId, ClientConnectionSender},
     db::relational_db::RelationalDB,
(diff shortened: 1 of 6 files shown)


crates/core/src/subscription/subscription.rs
```diff
@@ -395,64 +470,12 @@ impl From<Op> for TableOp {
...
 }
 
-/// Incremental evaluation of the supplied [`QueryExpr`].
-///
-/// The expression is assumed to project a single virtual table consisting
-/// of [`DatabaseTableUpdate`]s (see [`query::to_mem_table`]). That is,
-/// the `op_type` of the resulting [`Op`]s will be extracted from the virtual
-/// column injected by [`query::to_mem_table`], and the virtual column will be
-/// removed from the `row`.
-fn eval_incremental(
```
joshua-spacetime, line 405: This has been moved/renamed to `evaluator_for_primary_updates`

crates/core/src/subscription/subscription.rs
```diff
@@ -587,66 +644,69 @@ impl<'a> IncrementalJoin<'a> {
...
-            let b = run_query(ctx, db, tx, &rhs_virt, *auth)?
-                .into_iter()
-                .filter(|result| !result.data.is_empty())
-                .flat_map(|result| {
-                    result.data.into_iter().map(move |row| {
-                        Op {
-                            op_type: 1, // Insert
-                            row_pk: pk_for_row(&row),
-                            row: row.data,
-                        }
```
joshua-spacetime, line 609: This transformation is now handled in `evaluator_for_secondary_updates`

crates/core/src/subscription/subscription.rs
```diff
@@ -587,66 +644,69 @@ impl<'a> IncrementalJoin<'a> {
...
-                tx,
-                auth,
-                &query::to_mem_table(rhs_virt, &self.lhs.deletes()),
-                self.info,
-            )?;
+            let c = eval(&join_c, true)?;
             // {A- join B} U {A join B-} U {A- join B-}
-            let mut set = a.map(|op| (op.row_pk, op)).collect::<HashMap<PrimaryKey, Op>>();
-            set.extend(b.map(|op| (op.row_pk, op)));
-            set.extend(c.map(|op| (op.row_pk, op)));
```
joshua-spacetime, line 649: These transformations are no longer needed because `eval` returns a HashMap

mamcx (approved):
LGTM

PR #640, opened 2023-12-07
[CI] Merge docker images from multiple platforms

drogus:
# Description of Changes

Because we build our docker images in more than one job, they are also pushed as separate images. This commit adds a script that will merge the images into a multiplatform manifest

### DIFF
--- a/tools/merge-docker-images.sh
+++ b/tools/merge-docker-images.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+set -e
+
+# Shorten the first argument (commit sha) to 7 chars
+SHORT_SHA=${1:0:7}
+TAG="sha-$SHORT_SHA"
+
+# Check if images for both amd64 and arm64 exist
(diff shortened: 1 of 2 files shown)


tools/merge-docker-images.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+set -e
+
+# Shorten the first argument (commit sha) to 7 chars
+SHORT_SHA=${1:0:7}
+TAG="sha-$SHORT_SHA"
+
+# Check if images for both amd64 and arm64 exist
+if docker pull clockworklabs/spacetimedb:$TAG-amd64 --platform amd64 >/dev/null 2>&1 && docker pull clockworklabs/spacetimedb:$TAG-arm64 --platform arm64 >/dev/null 2>&1; then
```
jdetter, line 9: This seems like it could be race-condition-ey if the images are built in parallel (you could have this script executed twice with both images available). Is this not the case or is it not a problem?

drogus: It could potentially happen that it's executed twice, but it's not a problem - the one that will run later will just overwrite the existing manifest. This situation is unlikely, though, the window is probably like a second after the image is uploaded before the script runs.

jdetter (approved):
Ok seems fine

PR #641, opened 2023-12-07
Add timestamp to module logs

coolreader18:
# Description of Changes

Unsure about the TODO comment; walltime now breaks the illusion that all of a reducer happens at the same time, albeit not in a way observable to wasm. However, if we go with logical now it's more likely that the order of timestamps in the log would be non-monotonic, and we lose the information of how long between 2 logs internal to the same reducer.

1


coolreader18:
(the tests needed updating because they separated the log prefix from the message by the 4th colon, and the timestamp has 2 colons in it)

cloutiertyler (approved):
This looks good to me. Initially my thought was to go with reducer time, but actually I think wall-time is the better solution to go with here at least for now. My reasoning is that  a random user is not going to understand why all of their logs in the same reducer are running at the same exact millisecond.

Ultimately I would recommend stuffing both into the `Record` type.

coolreader18:
oh lol those weren't even flukes, I just missed them when going through & updating the tests

PR #645, opened 2023-12-08
Update database_logger.rs

cloutiertyler:
# Description of Changes

I'm moving the logs to the same directory that the database instances are in.

corresponding GitHub label.

1

### DIFF
--- a/crates/core/src/database_logger.rs
+++ b/crates/core/src/database_logger.rs
@@ -125,7 +125,7 @@ impl DatabaseLogger {
     // }
 
     pub fn filepath(address: &Address, instance_id: u64) -> PathBuf {
-        let root = crate::stdb_path("worker_node/database_instances");
+        let root = crate::stdb_path("database_instances");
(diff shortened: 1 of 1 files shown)


kim:
You’ll need to pass in the db root from db instance context, unfortunately, as the logs are still in the same place in standalone.

kim:
Standalone still has the control/worker dichotomy. Because paths are generally hardcoded (and in several places), we have the logs in the desired place in standalone, but not in cloud.

We should instead have a function which computes the root database path based on database address and instance id, and pass that around to derive the sub-paths.

cloutiertyler:
> Standalone still has the control/worker dichotomy. Because paths are generally hardcoded (and in several places), we have the logs in the desired place in standalone, but not in cloud.
> 
> We should instead have a function which computes the root database path based on database address and instance id, and pass that around to derive the sub-paths.

I'm going to close this PR in favor of creating a ticket to track this larger issue which we'll deal with before 1.0

PR #596, opened 2023-11-22
Refactoring bootstrapping and adding full support for constraints & multi-column indexes

mamcx:
# Description of Changes

Note: This is a multi-step set of PRs to bring constraints + multi-column indexes

Step:
- [x] [Add system tables](https://github.com/clockworklabs/SpacetimeDB/pull/225)
- [x] [Add multi-column indexes](https://github.com/clockworklabs/SpacetimeDB/pull/243)
- [x] [Refactoring IDs for system objects](https://github.com/clockworklabs/SpacetimeDB/pull/439)
- [x] Rework schema definition
- [x] Rework bootstrapping

# API and ABI

 - [x] This is a breaking change to the module ABI
 - [x] This is a breaking change to the module API
 - [x] This is a breaking change to the ClientAPI
 - [x] This is a breaking change to the SDK API

*If the API is breaking, please state below what will break*

PR #647, opened 2023-12-08
Fixes for performance drop related to missing index selection

mamcx:
# Description of Changes

This PR fixes the issues that cause the query optimizations to miss the opportunity to use indexes.

This is caused by several problems introduced in https://github.com/clockworklabs/SpacetimeDB/pull/596:

- The `constraints` information was not read correctly
- When a module passes the `TableDef` information it generates `Constraints::unset()` for all the columns without attributes. This causes the index information to skip them despite having indexes available

This requires a database wipe

1


cloutiertyler (approved):
This LGTM assuming tests pass.

PR #654, opened 2023-12-11
Fixes for SpacetimeDB.Runtime package

RReverser:
# Description of Changes

As usual, NuGet / MSBuild behaving differently for local projects vs published ones causes build issues that are not revealed by testing...

This is published to NuGet now and verified to work on an empty project.

Also changed the way `SpacetimeDB.Runtime.{props,targets}` are referenced locally by test projects by moving them to `Directory.Build.{props,targets}` at the top level of `modules/` directory.

This should simulate behaviour of published NuGet packages a bit more closely - in particular, I verified that it can reproduce the bug in question before the fix, while the previous `<Import />` approach succeeded regardless - and also makes it easier to share the same configs between multiple projects.

corresponding GitHub label.

1


jdetter (approved):
LGTM, thanks for the fix Ingvar 👍

PR #664, opened 2023-12-12
core: Initialize the database context only once

kim:
Access to the DatabaseInstanceContextController is racy, as another `get` may occur while a new DatabaseInstanceContext is initializing. Because the database acquires filesystem locks, this will lead to errors.

To fix this, the context value is wrapped in a OnceCell.

### DIFF
--- a/crates/core/src/database_instance_context_controller.rs
+++ b/crates/core/src/database_instance_context_controller.rs
@@ -1,14 +1,34 @@
 use std::sync::Arc;
 use std::{collections::HashMap, sync::Mutex};
 
+use once_cell::sync::OnceCell;
+
 use crate::db::db_metrics::DB_METRICS;
 use crate::host::scheduler::Scheduler;
 
 use super::database_instance_context::DatabaseInstanceContext;
 
+/// The database state managed by [`DatabaseInstanceContextController`].
+///
+/// Morally, this holds a pair of [`DatabaseInstanceContext`] and [`Scheduler`].
+/// The former holds metadata and state of a running database instance, while
+/// the latter manages scheduled reducers.
+///
+/// Operationally, this pair is wrapped in a [`OnceCell`], which ensures that
+/// the same physical database instance is not initialized multiple times due
+/// to concurrent access to the controller. Note that this is not primarily for
+/// safety -- databases acquire filesystem locks which ensure singleton access --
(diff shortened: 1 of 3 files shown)


crates/core/src/database_instance_context_controller.rs
```diff
@@ -1,63 +1,139 @@
...
-        contexts.remove(&database_instance_id)
+        let arc = contexts.remove(&database_instance_id)?;
+        match Arc::try_unwrap(arc) {
+            Ok(cell) => cell.into_inner(),
+            Err(arc) => {
+                // If the `Arc`'s refcount is > 1, another thread is currently
+                // executing the `get_or_try_init` closure. Wait until it
+                // completes (instead of returning `None`), as callers rely on
+                // calling `scheduler.clear()`.
+                let (dbic, scheduler) = arc.wait();
```
kim, line 106: This is a bit finicky: ideally, calling `remove` would give us ownership of the values, because what we want most of the time is to release all resources.

I'm wondering if we'd want to unify the concepts of `HostController` and `DatabaseInstanceContextContoller`, because the former is actually what controls database lifecycle.

crates/standalone/src/lib.rs
```diff
@@ -655,26 +655,23 @@ impl StandaloneEnv {
...
-                    let database = database.clone();
-                    let path = root_db_path.clone();
-                    let config = self.config;
-                    move || -> anyhow::Result<_> {
-                        let dbic = DatabaseInstanceContext::from_database(config, &database, instance_id, path)?;
-                        let sched = Scheduler::open(dbic.scheduler_db_path(root_db_path))?;
-                        Ok((dbic, sched))
-                    }
+        let (dbic, (scheduler, scheduler_starter)) = {
+            let (dbic, scheduler) = tokio::task::block_in_place(|| {
```
kim, line 659: Reverting to `spawn_blocking` is probably better, as this can literally take hours.

crates/core/src/database_instance_context_controller.rs
kim: To be clear: `Arc::try_unwrap` _should_ usually succeed, because current callers remove the instance context from `HostController` before calling `remove()`.

gefjon (approved):

PR #672, opened 2023-12-13
test: Build modules in release mode

kim:
# Description of Changes

Debug now creates wasm artifacts which are bigger than what we can support currently.

1

### DIFF
--- a/test/tests/autoinc1.sh
+++ b/test/tests/autoinc1.sh
@@ -40,7 +40,7 @@ EOF
 
   fsed "s/REPLACE_VALUE/$1/g" "${PROJECT_PATH}/src/lib.rs"
 
-  run_test cargo run publish --project-path "$PROJECT_PATH" --clear-database -d -S
+  run_test cargo run publish --project-path "$PROJECT_PATH" --clear-database --skip_clippy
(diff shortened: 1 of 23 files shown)


test/tests/autoinc1.sh
```diff
@@ -40,7 +40,7 @@ EOF
 
   fsed "s/REPLACE_VALUE/$1/g" "${PROJECT_PATH}/src/lib.rs"
 
-  run_test cargo run publish --project-path "$PROJECT_PATH" --clear-database -d -S
+  run_test cargo run publish --project-path "$PROJECT_PATH" --clear-database --skip_clippy
```
jdetter, line 43: It would probably be better to use an env variable like `PUBLISH_ARGS` or something instead of just blanketing `--skip_clippy` here, but I'm fine with this

jdetter (approved):
LGTM, I left a suggestion but this can merge as is

gefjon (approved):
Can you think of a reasonable place to document why we can't build these modules in debug mode? I don't fully understand the issue, so I'm not sure if it's something we can put in this repo, or if it's liable to reveal the Cloud secret sauce. If the former, could you create a `README.md` (or `NOTES` or whatever) in the `test` dir that describes why we have to build in release?

PR #677, opened 2023-12-14
perf(623): Use row counts in query optimization

### LINKED ISSUE #623: Join ordering based on table size metrics
Currently sql join optimizations only consider syntactic heuristics. We now have table size metrics that we can use to determine which side should be the index/build side and which one should be the probe side when optimizing joins.

joshua-spacetime:
Closes #623.

Before this patch query optimization was entirely syntax driven. Now that we keep table size metrics we can be somewhat data driven.

This patch improves index joins,
by using row counts to determine the index side and the probe side.

# Description of Changes

corresponding GitHub label.

### DIFF
--- a/crates/core/src/db/relational_db.rs
+++ b/crates/core/src/db/relational_db.rs
@@ -36,6 +36,7 @@ pub struct RelationalDB {
     commit_log: Option<CommitLogMut>,
     _lock: Arc<File>,
     address: Address,
+    row_count_fn: Arc<dyn Fn(TableId) -> i64 + Send + Sync>,
 }
 
 impl DataRow for RelationalDB {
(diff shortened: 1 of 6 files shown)


crates/core/src/subscription/query.rs
```diff
@@ -544,6 +546,12 @@ mod tests {
     #[test]
     fn test_eval_incr_for_index_join() -> ResultTest<()> {
         let (db, _) = make_test_db()?;
+        run_eval_incr_for_index_join(&db)?;
+        run_eval_incr_for_index_join(&db.with_row_count(Arc::new(|_| 5)))?;
```
joshua-spacetime, line 550: This new test case will run our incremental evaluation suite under the new query plan.

crates/vm/src/expr.rs
```diff
@@ -454,7 +454,10 @@ impl IndexJoin {
         // during construction of the index join.
         let probe_column = self.probe_side.source.head().column(&self.probe_field).unwrap().col_id;
         match self.index_side {
-            Table::DbTable(_) => self,
+            // If the size of the indexed table is sufficiently large, do not reorder.
+            Table::DbTable(DbTable { table_id, .. }) if row_count(table_id) > 1000 => self,
+            // If this is a delta table, we must reorder.
+            // If this is a sufficiently small physical table, we should reorder.
```
joshua-spacetime, line 460: Sufficiently small means 1000 rows or less.

crates/core/src/db/relational_db.rs
```diff
@@ -139,12 +140,29 @@ impl RelationalDB {
...
+                    .with_label_values(&db_address, &table.into())
+                    .get()
+            }),
         };
 
         log::info!("[{}] DATABASE: OPENED", address);
         Ok(db)
     }
 
+    /// Returns an approximate row count for a particular table.
```
mamcx, line 155: There is already a `RowCount` object:

```rust
#[derive(Debug, Copy, Clone, PartialOrd, Ord, PartialEq, Eq)]
pub struct RowCount {
    pub min: usize,
    pub max: Option<usize>,
}
```

on `sats -> Relation`

mamcx (approved):
LGTM

crates/core/src/db/relational_db.rs
joshua-spacetime: I will add a TODO like you mentioned.

I wonder if we should move `Relation` out of `sats`, since it will need access to database statistics. Because right now `RowCount` only works on `MemTable`s.

PR #678, opened 2023-12-14
Use associate type for tx

Shubham8287:
# Description of Changes
1. moving from concrete type to use of associate type for tx, this wil help to adapt different type of transaction later
  i.e to use   `relational_db::MutTx` instead of MutTxId, and `realtional_db::Tx` for `compiler.rs` as that is suppose to work with read only tx.
2. changed `traits::MutTxId` -> `traits::MutTx`, similarly for `traits::TxId` as none of them mean to refer to any id.

1


gefjon (approved):
Seems trivial. Tests pass; code is unobjectionable.

PR #679, opened 2023-12-15
read tx implemented in datastore

Shubham8287:
# Description of Changes
1. Modify the current Iter to work with Option<TxState>. This is necessary to enable the use of the iterator with read-only transactions and during bootstrapping.
2. Create a Tx struct and implement the required methods.
3. Add a basic test to ensure shared access by read-only transactions.
No


crates/core/src/db/datastore/locking_tx_datastore/mod.rs
```diff
@@ -3551,6 +3667,26 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn test_read_only_tx_shared_lock() -> ResultTest<()> {
```
gefjon, line 3671: I was about to leave a comment asking for tests, but you beat me to it! I feel good about this test.
```suggestion
    /// Test that two read-only TXes can operate concurrently without deadlock or blocking,
    /// and that both observe correct results for a simple table scan.
    fn test_read_only_tx_shared_lock() -> ResultTest<()> {
```

crates/core/src/db/datastore/locking_tx_datastore/mod.rs
```diff
@@ -3551,6 +3667,26 @@ mod tests {
...
+        let row2 = u32_str_u32(2, "Bar", 20);
+        datastore.insert_mut_tx(&mut tx, table_id, row2.clone())?;
+        datastore.commit_mut_tx_for_test(tx)?;
+
+        // create multiple read only tx, and use them together.
+        let read_tx_1 = begin_tx(&datastore);
+        let read_tx_2 = begin_tx(&datastore);
+        let rows = vec![row1, row2];
+        assert_eq!(all_rows_tx(&read_tx_2, table_id), rows.clone());
+        assert_eq!(all_rows_tx(&read_tx_1, table_id), rows);
```
gefjon, line 3684:
```suggestion
        let rows = &[row1, row2];
        assert_eq!(&all_rows_tx(&read_tx_2, table_id), rows);
        assert_eq!(&all_rows_tx(&read_tx_1, table_id), rows);
```

Haven't tested this change, but my gut tells me it should work.

crates/core/src/db/datastore/locking_tx_datastore/mod.rs
```diff
@@ -120,6 +123,99 @@ impl<'a> DataRef<'a> {
...
+                NonEmpty::new(StTableFields::TableName.col_id()),
+                AlgebraicValue::String(table_name.to_owned()),
+            )
+            .map(|mut iter| {
+                iter.next()
+                    .map(|row| TableId(*row.view().elements[0].as_u32().unwrap()))
+            })
+    }
+
+    fn iter<'a>(&'a self, ctx: &'a ExecutionContext, table_id: &TableId) -> super::Result<Iter<'a>> {
```
gefjon, line 151: I'm unclear on the relationship between a `TxId` and an `ExecutionContext`. What happens if I perform multiple operations on a single `TxId` with different `ExecutionContext`s? Is it possible that `TxId` should hold an `&ExecutionContext`? I assume you've thought about this and come up with a good solution, but I'd like to see comments which explain your reasoning added to the code.

Shubham8287: I'm not entirely sure if I understood correctly, but it seems like there might be a need to refactor the usage of `ExecutionContext` in the datastore. Currently, it doesn't mostly consider the `WorkloadType` (sql/reducer), often defaulting to Internal.
So for example, If you call `Locking::table_id_from_name_mut_tx` it doesn't take `context` as param and uses `Internal` but `iters` do.

Currently, I've made `Tx` implementation consistent with `MutTxId`, but it might be worth to revisit it as whole as a separate task .

> What happens if I perform multiple operations on a single TxId with different ExecutionContexts

This may mess up with labels on metrics for `WorkLoadType` but will still provide meaningfull values as they are local states.

crates/core/src/db/datastore/locking_tx_datastore/mod.rs
```diff
@@ -120,6 +123,215 @@ impl<'a> DataRef<'a> {
...
+        &'a self,
+        ctx: &'a ExecutionContext,
+        table_id: &TableId,
+        cols: NonEmpty<ColId>,
+        value: AlgebraicValue,
+    ) -> super::Result<IterByColEq<'_>> {
+        self.iter_by_col_range(ctx, table_id, cols, value)
+    }
+
+    fn schema_for_table<'tx>(&'tx self, table_id: TableId) -> super::Result<Cow<'tx, TableSchema>> {
```
Shubham8287, line 193: This has been mostly copied from `CommittedState::schema_for_table`, as `iter_by_col_eq()` method implementation differs for `TxId` and `CommittedState`.
I am experimenting if we can avoid duplication by some way in next PR as it's dead-code right now.

crates/core/src/db/datastore/traits.rs
```diff
@@ -90,7 +90,7 @@ pub trait TxDatastore: DataRow + Tx {
         ctx: &'a ExecutionContext,
         tx: &'a Self::Tx,
         table_id: TableId,
-        cols: NonEmpty<ColId>,
+        cols: impl Into<NonEmpty<ColId>>,
```
cloutiertyler, line 93: What's the reason for this change? Seems unrelated, no?

crates/core/src/db/datastore/traits.rs
```diff
@@ -121,7 +123,6 @@ pub trait MutTxDatastore: TxDatastore + MutTx {
     fn schema_for_table_mut_tx<'tx>(&self, tx: &'tx Self::MutTx, table_id: TableId) -> Result<Cow<'tx, TableSchema>>;
     fn drop_table_mut_tx(&self, tx: &mut Self::MutTx, table_id: TableId) -> Result<()>;
     fn rename_table_mut_tx(&self, tx: &mut Self::MutTx, table_id: TableId, new_name: &str) -> Result<()>;
-    fn table_id_exists(&self, tx: &Self::MutTx, table_id: &TableId) -> bool;
```
cloutiertyler, line 124: I guess this change was already merged into `master` somehow?

cloutiertyler (commented):
In general I think this PR makes a lot of sense. It's definitely what I was looking for. I'd like to chat with you about the comments I left, but then I'm okay to approve this with a few small changes.

crates/core/src/db/datastore/locking_tx_datastore/mod.rs
Shubham8287: Yes, method arguments are consistent between `MutTxId` and `TxId`. So, If some method does not takes an `ExecutionContext`, it's `TxId` counter-part will also not and vice versa.

crates/core/src/db/datastore/traits.rs
Shubham8287: no, this has been part of `TxId`, not required with `MutTxIxd`

crates/core/src/db/datastore/locking_tx_datastore/mod.rs
```diff
@@ -120,6 +123,215 @@ impl<'a> DataRef<'a> {
...
+        &'a self,
+        ctx: &'a ExecutionContext,
+        table_id: &TableId,
+        cols: NonEmpty<ColId>,
+        value: AlgebraicValue,
+    ) -> super::Result<IterByColEq<'_>> {
+        self.iter_by_col_range(ctx, table_id, cols, value)
+    }
+
+    fn schema_for_table(&self, table_id: TableId) -> super::Result<Cow<'_, TableSchema>> {
```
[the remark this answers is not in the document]
Shubham8287: Reduntation here it due to `self.iter_by_col_eq()` has different implementation for `CommittedState` and both transactions.
`schema_for_table()/iter_by_col_eq()` in `CommittedState` doesn't consider indexes and thus implemented only to be used during  bootstrapping.
for `TxId` `Iter_by_col_eq` try to use index as well, so not sure if there is any straight-forward way to remove this duplicacy.

crates/core/src/db/datastore/traits.rs
Shubham8287: This is to keep them consistent with fn `iter_by_col_range_mut_tx()` in `MutTxDataStore`.

crates/core/src/db/datastore/traits.rs
cloutiertyler: Well, I would assume you might still need to check to see if a table id exists in the context of a mutable transaction, no?

Shubham8287: Not for this case, method is only required during query compiling to `ast`, now whole compiling stage works with read-tx only.

crates/core/src/db/datastore/locking_tx_datastore/mod.rs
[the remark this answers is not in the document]
cloutiertyler: Food for thought, could we implement this as a default implemenation inside of the Datastore trait?

```
fn schema_for_table(self: &TxDatastore, table_id: TableId) -> super::Result<Cow<'_, TableSchema>> {
    self.iter_by_col_eq()
}
```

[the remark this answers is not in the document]
Shubham8287: That would not be any beneficial for code duplicacy as it will still require different implementation for `MutTxDatastore` and bootstrapping code is part of `CommittedState` which can't really use datastore and expects `self.schema_for_table()`.
rather, I tried by creating internal `View` trait, keeping it in locking module not in `trait.rs` as it is specific to locking datastore.
https://github.com/clockworklabs/SpacetimeDB/pull/679/commits/495f21c75a46a9a8f9dcf4c7626adefafcc907b1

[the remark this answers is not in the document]
cloutiertyler: Yeah this strategy works really well. I think the name is good as well. I'm okay with this.

crates/core/src/db/datastore/traits.rs
cloutiertyler: Okay, but for symmetry's sake it would make sense for this to be a thing you might want to do during a mutable transaction even if it's unused. I'm fine skipping it for now, and adding it later though.

cloutiertyler (approved):
This now LGTM.

PR #685, opened 2023-12-18
feat: subscription to use read type tx

Shubham8287:
# subscription actor to use read tx.
1. In datastore, made assigned `traits::Tx::Tx` to point to`TxId` rather `MutTxId`.
2. Done required changes in `relational_db.rs` to support 1.
3. started using`Tx` in subscription actor.
4. refactored `vm.rs` to work with both types of tx.
5. refactored `execute.rs`

broken test will be fixed by https://github.com/clockworklabs/SpacetimeDB/pull/695


crates/core/src/db/relational_db.rs
```diff
@@ -256,16 +272,29 @@ impl RelationalDB {
     /// state. See also [`Self::with_auto_commit`].
     #[tracing::instrument(skip_all)]
     pub fn begin_tx(&self) -> MutTx {
-        log::trace!("BEGIN TX");
+        log::trace!("BEGIN Mut TX");
         self.inner.begin_mut_tx()
     }
 
+    // TODO(shub): name this as per decided convention (mut_tx/ tx).
```
cloutiertyler, line 279: 👍

crates/core/src/db/relational_db.rs
```diff
@@ -256,16 +272,29 @@ impl RelationalDB {
...
+    #[tracing::instrument(skip_all)]
+    pub fn begin_read_tx(&self) -> Tx {
+        log::trace!("BEGIN TX");
+        self.inner.begin_tx()
+    }
+
     #[tracing::instrument(skip_all)]
     pub fn rollback_tx(&self, ctx: &ExecutionContext, tx: MutTx) {
-        log::trace!("ROLLBACK TX");
+        log::trace!("ROLLBACK Mut TX");
```
cloutiertyler, line 288:
```suggestion
        log::trace!("ROLLBACK MUT TX");
```

Just for consistency

crates/core/src/db/relational_db.rs
```diff
@@ -256,16 +272,29 @@ impl RelationalDB {
     /// state. See also [`Self::with_auto_commit`].
     #[tracing::instrument(skip_all)]
     pub fn begin_tx(&self) -> MutTx {
-        log::trace!("BEGIN TX");
+        log::trace!("BEGIN Mut TX");
```
cloutiertyler, line 275:
```suggestion
        log::trace!("BEGIN MUT TX");
```

Just for consistency

crates/core/src/db/relational_db.rs
```diff
@@ -191,10 +193,19 @@ impl RelationalDB {
...
+    pub fn schema_for_table_mut<'tx>(
+        &self,
+        tx: &'tx MutTx,
+        table_id: TableId,
+    ) -> Result<Cow<'tx, TableSchema>, DBError> {
         self.inner.schema_for_table_mut_tx(tx, table_id)
     }
 
+    #[tracing::instrument(skip_all)]
+    pub fn schema_for_table<'tx>(&self, tx: &'tx Tx, table_id: TableId) -> Result<Cow<'tx, TableSchema>, DBError> {
```
cloutiertyler, line 205: I suppose we could have this take a `TxMode` or something of a similar concept and not duplicate all the functions, but I dunno how I feel about that. I think what you have now is probably better, but thought I'd mention it.

crates/core/src/sql/execute.rs
```diff
@@ -79,26 +75,38 @@ pub fn execute_single_sql(
...
     auth: AuthCtx,
 ) -> Result<Vec<MemTable>, DBError> {
     let total = ast.len();
     let ctx = ExecutionContext::sql(db.address(), query_debug_info);
-    let p = &mut DbProgram::new(&ctx, db, tx, auth);
-    let q = Expr::Block(ast.into_iter().map(|x| Expr::Crud(Box::new(x))).collect());
-
     let mut result = Vec::with_capacity(total);
-    collect_result(&mut result, run_ast(p, q).into())?;
+    match ast.iter().all(|expr| matches!(expr, CrudExpr::Query(_))) {
```
cloutiertyler, line 85: Probably want to check with Mario or Joshua on this one if you haven't already, simply because I'm just not familiar.

crates/core/src/vm.rs
```diff
@@ -328,26 +354,32 @@ impl<'db, 'tx> DbProgram<'db, 'tx, MutTx> {
...
+        match self.tx {
+            TxMode::MutTx(tx) => match table {
+                // TODO: How do we deal with mutating values?
+                Table::MemTable(_) => Err(ErrorVm::Other(anyhow::anyhow!("How deal with mutating values?"))),
+                Table::DbTable(t) => {
+                    let count = self.db.delete_by_rel(tx, t.table_id, rows);
+                    Ok(Code::Value(count.into()))
+                }
+            },
+            TxMode::Tx(_) => Err(ErrorVm::Unsupported("Read tx".to_owned())),
```
cloutiertyler, line 382: Is this an error or is this actually unreachable? Like is it possible to have a TxMode::Tx if you've arrived at this function?

crates/standalone/src/lib.rs
```diff
@@ -572,18 +572,19 @@ impl StandaloneEnv {
...
-                stdb.rollback_tx(&cx, tx);
+                stdb.rollback_read_tx(&cx, tx);
                 Err(e.into())
             }
 
             Ok(maybe_hash) => {
                 // Release tx due to locking semantics and acquire a control db
                 // lock instead.
-                stdb.commit_tx(&cx, tx)?;
+                // Verify this
```
cloutiertyler, line 586:
```suggestion
```

You are correct, this should be a release here not a commit.

cloutiertyler (commented):
Left some small comments but overall this is looking really really good. I would be okay to merge this once the one it's dependent on merges.

crates/core/src/db/relational_db.rs
Shubham8287: I thought about that as well but then all callers have to keep wrapping tx objects in`TxMode` to pass as param, Right now, `TxMode` is local to `vm.rs` only.
I felt bit of wrapper methods duplicacy is fine here to avoid using extra enum everywhere.

crates/core/src/sql/execute.rs
Shubham8287: Let me add them as reviewer, specially on this snippet @joshua-spacetime , @mamcx

crates/core/src/vm.rs
Shubham8287: This should be unreachable code.

crates/core/src/sql/execute.rs
mamcx: Yes this looks correct. `CrudExpr::Query(_)` only takes `SELECT..` expressions without mutate effects.

I suggest to put this logic in `CrudExpres` (similar to `CrudExpr::is_reads()`) instead of here. The idea is that here is an upper layer and all decisions should be inside the `vm` or `query` layer so is more visible to the people that do optimizations.

crates/core/src/vm.rs
```diff
@@ -328,26 +354,32 @@ impl<'db, 'tx> DbProgram<'db, 'tx, MutTx> {
...
+                Table::DbTable(x) => {
+                    for row in rows {
+                        self.db.insert(tx, x.table_id, row)?;
+                    }
+                    Ok(Code::Pass)
                 }
-                Ok(Code::Pass)
-            }
+            },
+            TxMode::Tx(_) => unreachable!("mutable operation is invalid with read tx"),
```
mamcx, line 368: If here is this then the upper layer should prove this is not possible. If can't it should be a normal `Err` instead (is better to avoid a crash)

crates/core/src/sql/execute.rs
Shubham8287: Agree, that makes sense

crates/core/src/vm.rs
Shubham8287: Reaching to this branch anytime implies a bug in code and not to do anything with Users interaction.
I added a comment on `VmProgram::eval_query()` (only caller) to specify safetu gurantee.

crates/core/src/db/relational_db.rs
```diff
@@ -492,13 +531,19 @@ impl RelationalDB {
         self.inner.iter_mut_tx(ctx, tx, table_id)
     }
 
+    #[tracing::instrument(skip(self, ctx, tx))]
+    pub fn iter<'a>(&'a self, ctx: &'a ExecutionContext, tx: &'a Tx, table_id: TableId) -> Result<Iter<'a>, DBError> {
+        let _guard = DB_METRICS.rdb_iter_time.with_label_values(&table_id.0).start_timer();
```
joshua-spacetime, line 536: There's no need to add this metric since it doesn't have the desired behavior.

crates/core/src/sql/execute.rs
```diff
@@ -79,26 +75,38 @@ pub fn execute_single_sql(
...
 /// Run the `SQL` string using the `auth` credentials
 #[tracing::instrument(skip_all)]
-pub fn run(db: &RelationalDB, tx: &mut MutTx, sql_text: &str, auth: AuthCtx) -> Result<Vec<MemTable>, DBError> {
-    let ast = compile_sql(db, tx, sql_text)?;
-    execute_sql(db, tx, ast, Some(&QueryDebugInfo::from_source(sql_text)), auth)
+pub fn run(db: &RelationalDB, sql_text: &str, auth: AuthCtx) -> Result<Vec<MemTable>, DBError> {
+    let query_info = &QueryDebugInfo::from_source(&sql_text);
+    let ctx = &ExecutionContext::sql(db.address(), Some(query_info));
+    let ast = db.with_read_only(&ctx, |tx| compile_sql(db, tx, sql_text))?;
+    execute_sql(db, ast, Some(&QueryDebugInfo::from_source(sql_text)), auth)
```
joshua-spacetime, line 109: Are compilation and execution in two separate transactions now?

crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -164,14 +167,15 @@ impl ModuleSubscriptionActor {
...
-        let mut tx = self.relational_db.begin_tx();
+        let mut tx = self.relational_db.begin_read_tx();
 
         let result = self._add_subscription(sender, subscription, &mut tx).await;
 
         // Note: the missing QueryDebugInfo here is only used for finishing the transaction;
         // all of the relevant queries already executed, with debug info, in _add_subscription
         let ctx = ExecutionContext::subscribe(self.relational_db.address(), None);
-        self.relational_db.finish_tx(&ctx, tx, result)
+        self.relational_db.rollback_read_tx(&ctx, tx);
```
joshua-spacetime, line 177: What does it mean to rollback a read transaction?

crates/core/src/sql/execute.rs
Shubham8287: yes, compilation works with read-only tx, then execution could be read-tx or write-tx based on compilation result (ast).

crates/core/src/subscription/module_subscription_actor.rs
Shubham8287: I guess naming could be better here, `finish_tx` looks better. It's basically to drop the guard.

crates/core/src/sql/execute.rs
joshua-spacetime: I think that's fine. Can you just add a comment to that effect since it might be unexpected?

crates/core/src/subscription/module_subscription_actor.rs
joshua-spacetime: Yeah I agree, `finish` instead of `rollback` is clearer.

joshua-spacetime (approved):
LGTM pending some minor nits and tests passing.

PR #686, opened 2023-12-19
core: Remove index update when updating a database

kim:
The `TableDef` type is nominally the same for the database schema and
the schema description from a module, but only the former contains
information about indexes and sequences. This would cause all indexes to
be dropped when updating a database, because none of the existing
indexes would ever be present in the module's schema definition.

It is unclear if the functionality could be recovered, but it does not
seem to be robust without a dedicated schema manipulation language
either way.

We still want to allow module updates which do not change the schema at
all, or only introduce new tables, as otherwise users may be forced to
re-run expensive initialisation steps even though they only introduced
debugging code or somesuch.

To make this work, a somewhat dubious equivalence comparison is made,
comparing the constraint definitions in a way deemed correct in post-hoc
analysis.

# Description of Changes

corresponding GitHub label.

### DIFF
--- a/crates/core/src/db/update.rs
+++ b/crates/core/src/db/update.rs
@@ -3,6 +3,7 @@ use std::borrow::Cow;
 use std::collections::{BTreeMap, HashMap};
 
 use anyhow::Context;
+use spacetimedb_primitives::Constraints;
 
 use crate::database_logger::SystemLogger;
 use crate::error::DBError;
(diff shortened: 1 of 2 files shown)


mamcx (approved):
LGTM. I think we need to document that we remove temporary support of migrating tables.

kim:
We never supported migrating tables. We supported migrating _indices_ (because it was both obviously supportable and convenient for my own work at the time). We still support adding new tables after this patch.

kim:
What is implemented here, [was documented](https://github.com/clockworklabs/SpacetimeDB/pull/686/files#diff-bb54ccb2df43ca366b13e60aa5db0610ca8d7a187dc5a1aacffd7a71275ce9c9L25-L29) btw

PR #688, opened 2023-12-21
`BTreeIndex`: optimize with `BTreeMap<AV, SmallVec<[RowId; 1]>>`

Centril:
# Description of Changes

Optimizes `BTreeIndex` using `BTreeMap<AV, SmallVec<[RowId; 1]>>` as the backing implementation where the `SmallVec` is sorted.
This was found to improve performance of inserts, deletes, and most importantly, seeks, in the mem-arch-prototype.
Numbers can be found in https://github.com/clockworklabs/SpacetimeDBPrivate/pull/671#issue-2051086035 morally does the same for the prototype. However, it is likely that the performance improvements are even greater here, as `IndexKey` only shrunk from 40 to 32 bytes in the prototype, but shrinks from 72 to 32 bytes here. Moreover, this implementation here uses a more direct and inlined implementation than using `flat_map`, avoiding paying for overhead associated with `DoubleEndedIterator` that the latter has.

None

2, limited to this module and fairly easy to review.

### DIFF
--- a/crates/core/src/db/datastore/locking_tx_datastore/btree_index.rs
+++ b/crates/core/src/db/datastore/locking_tx_datastore/btree_index.rs
@@ -1,12 +1,12 @@
 use super::RowId;
 use crate::error::DBError;
+use core::ops::RangeBounds;
+use core::slice;
+use smallvec::SmallVec;
 use spacetimedb_primitives::*;
 use spacetimedb_sats::data_key::ToDataKey;
-use spacetimedb_sats::{AlgebraicValue, DataKey, ProductValue};
-use std::{
-    collections::{btree_set, BTreeSet},
-    ops::{Bound, RangeBounds},
-};
+use spacetimedb_sats::{AlgebraicValue, ProductValue};
+use std::collections::btree_map::{BTreeMap, Range};
 
 /// ## Index Key Composition
 ///
@@ -28,39 +28,19 @@ use std::{
 /// [AlgebraicValue::I32(0)] = Row(ProductValue(...))
 /// [AlgebraicValue::Product(AlgebraicValue::I32(0), AlgebraicValue::I32(1))] = Row(ProductValue(...))
 /// ```
-#[derive(Clone, Eq, PartialEq, Ord, PartialOrd)]
-struct IndexKey {
-    value: AlgebraicValue,
-    row_id: RowId,
-}
+type IndexKey = AlgebraicValue;
 
-impl IndexKey {
-    #[tracing::instrument(skip_all)]
-    pub(crate) fn from_row(value: &AlgebraicValue, row_id: DataKey) -> Self {
-        Self {
-            value: value.clone(),
-            row_id: RowId(row_id),
-        }
-    }
-}
-
-pub struct BTreeIndexIter<'a> {
-    iter: btree_set::Iter<'a, IndexKey>,
-}
-
-impl Iterator for BTreeIndexIter<'_> {
-    type Item = RowId;
-
-    #[tracing::instrument(skip_all)]
-    fn next(&mut self) -> Option<Self::Item> {
-        self.iter.next().map(|key| key.row_id)
-    }
-}
+/// A sorted collection of row ids.
+///
+/// This is `Vec<_>` but we allow a single element to be stored inline
+/// to improve performance for the common case of one element.
+type RowIds = SmallVec<[RowId; 1]>;
 
 /// An iterator for the rows that match a value [AlgebraicValue] on the
 /// [BTreeIndex]
 pub struct BTreeIndexRangeIter<'a> {
-    range_iter: btree_set::Range<'a, IndexKey>,
+    outer: Range<'a, AlgebraicValue, RowIds>,
+    inner: Option<slice::Iter<'a, RowId>>,
     num_keys_scanned: u64,
 }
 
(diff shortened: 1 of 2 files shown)


crates/core/src/db/datastore/locking_tx_datastore/btree_index.rs
```diff
@@ -70,11 +50,17 @@ impl<'a> Iterator for BTreeIndexRangeIter<'a> {
...
+            if let Some(inner) = self.inner.as_mut() {
+                if let Some(ptr) = inner.next() {
+                    self.num_keys_scanned += 1;
+                    return Some(ptr);
+                }
+            }
+
+            self.inner = None;
+            let (_, next) = self.outer.next()?;
+            self.inner = Some(next.iter());
```
gefjon, line 63: I find this hard to follow. Could you add comments describing the flow of execution here? Or maybe a comment showing what this would look like as a chain of iterator combinators instead of in the imperative style?

crates/core/src/db/datastore/locking_tx_datastore/btree_index.rs
```diff
@@ -29,39 +29,19 @@ use std::{
...
-
-    #[tracing::instrument(skip_all)]
-    fn next(&mut self) -> Option<Self::Item> {
-        self.iter.next().map(|key| key.row_id)
-    }
-}
+/// A sorted collection of row ids.
+///
+/// This is `Vec<_>` but we allow a single element to be stored inline
+/// to improve performance for the common case of one element.
```
gefjon, line 37: Extend docstring justifying why this needs to be sorted, and who maintains that.

crates/core/src/db/datastore/locking_tx_datastore/btree_index.rs
```diff
@@ -155,35 +155,27 @@ impl BTreeIndex {
...
-        let end = map(range.end_bound(), DataKey::max_datakey());
         BTreeIndexRangeIter {
-            range_iter: self.idx.range((start, end)),
+            outer: self.idx.range((range.start_bound(), range.end_bound())),
+            inner: None,
             num_keys_scanned: 0,
         }
     }
 
     /// Construct the [BTreeIndex] from the rows.
```
gefjon, line 169: Extend docstring to describe the added `bool` in the return type.

crates/core/src/db/datastore/locking_tx_datastore/table.rs
```diff
@@ -40,7 +40,7 @@ impl Table {
         let row = self.rows.remove(row_id)?;
         for (cols, index) in self.indexes.iter_mut() {
             let col_value = row.project_not_empty(cols).unwrap();
-            index.delete(&col_value, row_id)
+            index.delete(&col_value, row_id);
```
gefjon, line 43:
```suggestion
            let _removed = index.delete(&col_value, row_id);
            debug_assert!(_removed, "Deleted a row which wasn't present in the index!");
```

gefjon (commented):
Minor changes requested. I'd really like to get @cloutiertyler 's eyes on this one before it merges.

gefjon:
A question: we've demonstrated that this performs better for integer- and string-keyed indexes, where duplicates are permitted but relatively rare. I wonder if the speedup still holds for key types where duplicates are expected to be common. E.g. if I write:

```rust
#[spacetimedb(table)]
#[spacetimedb(index(btree, name = "logged_in_users", logged_in))]
struct User {
  name: String,
  logged_in: bool,
}
```

and have 50k users with ~10k logged in, does the `remove` operation in the `SmallVec` become prohibitively slow?

This leads me to wonder, would it be better to define `BTreeIndex` as `BTreeMap<AlgebraicValue, BTreeSet<RowId>>`? Is there a `SmallBTree` type we could use? This would give us `O(log(num_matching))` deletions, where num_matching is the number of rows with that particular value in the index, where sorted `SmallVec` gives us `O(num_matching)`.

mamcx:
> This leads me to wonder, would it be better to define `BTreeIndex` as `BTreeMap<AlgebraicValue, BTreeSet<RowId>>`? Is there a `SmallBTree` type we could use? This would give us `O(log(num_matching))` deletions, where num_matching is the number of rows with that particular value in the index, where sorted `SmallVec` gives us `O(num_matching)`.

I think that was close to my original implementation, long ago.
 
> EDIT: Follow-up question: do we care if indexes where duplicates are common are slow? And more specifically, do we care in the short term (potentially bad to merge this) or the long term (merge this and fix it later)?

Yes (certainly long-term). Idiomatically, what most users know about DB is "if is slow, add indexes", but from the `SQL`/`User` viewpoint they haven't knowledge or any way to tweak them. People expect that adding an index works for any kind of datatype, even if adding it is unnecessary. I expect most DB have too few indexes OR too many. Most people never bother to find what is the proper number/application of indexes.

I think, long term, we should make a trait for indexes instead of writing special implementations for common cases and picking the most appropriate for the datatype and the cardinality of the data.

mamcx:
> The implementation in the PR is faster for duplicates in the seek case so it's only the write heavy tables where index value duplicates are common that are the concern.

A common case that is not necessarily `write heavy` is the use of indexes on `booleans`. For example, this happens:

```sql
CREATE TABLE customer (
    customer_code TEXT PRIMARY KEY CHECK (not_empty(customer_code)),
    name         TEXT NOT NULL CHECK (not_empty(name)),

    is_active    BOOLEAN  NOT NULL DEFAULT TRUE,    
);

CREATE INDEX idx_customer_active ON customer (is_active);
```

This kind of index will have a lot of `true` values and few `false`.  Ideally, a DB should have for fields with low cardinality something like `BtreeMap<bool, BtreeMap<Pk, Row>>` but that kind of control is impossible. (How wonderful could be if we can supply our own internal storage and just implement a trait!)

mamcx:
> For the purposes of this PR, it's out of scope to add other index kinds for specialized cases like autoinc u64 or a bool.

yes, right, I just trying to point out a case that could be useful for test the perf

cloutiertyler:
Oh, so the 20% improvement is not for this PR, but rather for the PR to the mem-arch prototype? What are the numbers for this PR. We should be able to get those, right?

benchmarks please

cloutiertyler:
benchmarks please

kazimuth:
Ah, @Centril I did look at this last month, I just didn't have any particular comments. LGTM.

callgrind please

PR #691, opened 2023-12-21
Datastore: `Cow<'tx, ProductValue>` rather than `&'tx ProductValue`.

gefjon:
# Description of Changes

Our current datastore stores `ProductValue`s, and thus can return references to them, but the prototype new mem-arch stores a new flattened representation, and so cannot return `&'tx ProductValue`.

In preparation for the new storage format, this commit changes the datastore traits to return `Cow<'tx, ProductValue>`, so that the new implementation will be able to perform conversion and return Cow::Owned(ProductValue::from(internal_repr))`.

None; this is a strictly internal change.

1; this is a tiny diff which does not change the datastore's semantics. It may (or may not) introduce some performance regressions, as `Cow<T>` is larger than `&T`, but I expect the LLVM to inline and eliminate the `Cow::Owned` branches in most cases.

### DIFF
--- a/crates/core/src/db/datastore/traits.rs
+++ b/crates/core/src/db/datastore/traits.rs
@@ -38,7 +38,7 @@ pub struct TxData {
 }
 
 pub trait Data: Into<ProductValue> {
-    fn view(&self) -> &ProductValue;
+    fn view(&self) -> Cow<'_, ProductValue>;
 }
 
 pub trait DataRow: Send + Sync {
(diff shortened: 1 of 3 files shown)


crates/core/src/db/datastore/traits.rs
```diff
@@ -38,15 +38,22 @@ pub struct TxData {
...
+    fn view(&self) -> Cow<'_, ProductValue>;
 }
 
 pub trait DataRow: Send + Sync {
     type RowId: Copy;
 
     type DataRef<'a>;
 
-    fn view_product_value<'a>(&self, data_ref: Self::DataRef<'a>) -> &'a ProductValue;
+    /// Return a `Cow`, which currently will always be `Borrowed`,
```
kim, line 49: Is this relevant for the user or implementor of the trait? Is there anything I’d need to do differently depending on whether the cow is borrowed or owned? How does borrowing hide any facts about the stored format?

gefjon: The core interface of `Cow` is that it implements `Deref`. `Deref` gives a ref with a lifetime bounded by the `Cow`, not by (in our case) the TX. This is in contrast to our current `view_product_value`, which gives a ref with a lifetime bounded by the TX, which may outlive the `DataRef`. The punchline here is that nowhere in our code do we depend on that longer lifetime, so `Cow` (with deref coercion) is a drop-in change.

Returning `Cow` here hides the stored format in the sense that the current datastore can return `Cow::Borrowed`, whereas the new datastore will return `Cow::Owned`, and consumers (at least in our codebase as it exists now) are blissfully ignorant of the difference.

kim: Well, I was asking to clarify this in the commentary.

PR #693, opened 2023-12-22
Delete Files Flag for Generate Command

jdetter:
# Description of Changes

- This adds a flag `--delete-files` to the `spacetime generate` command which, when supplied, shows the user which files were not generated by us and prompts them if they would like to delete the files. We also check to see if the file starts with our auto-gen tag so that we only prompt the user to delete files that were created by us. This means they can add in their own files and we won't delete them.

When running with the new flag we get something like this:
```
boppy@boppy-macbook game % bash ./generate-client-files.sh 
Generating C# files
    Checking <project> v0.1.0 (/Users/boppy/clockwork/<project>/packages/game)
    Finished release [optimized] target(s) in 14.40s
    Finished release [optimized] target(s) in 0.10s
compilation took 1.433383083s
The following files were not generated by this command and will be deleted:
  ../../../<project>/Assets/_Project/autogen/File1.cs
  ../../../<project>/Assets/_Project/autogen/File2.cs
  ../../../<project>/Assets/_Project/autogen/File3.cs
Are you sure you want to delete these files? [y/N]
```

If you enter y or yes to this prompt it will remove the files for you. You can add `--force` onto the end as well if you would like to execute the removal of the files without a prompt.

corresponding GitHub label.

- No API or ABI break

0


crates/cli/src/subcommands/generate/mod.rs
```diff
@@ -120,7 +128,47 @@ Failed to compile module {:?}. See cargo errors above for more details.",
...
+        }
+        if !files_to_delete.is_empty() {
+            println!("The following files were not generated by this command and will be deleted:");
+            for path in &files_to_delete {
+                println!("  {}", path.to_str().unwrap());
+            }
+            println!("Are you sure you want to delete these files? [y/N]");
+            let mut input = String::new();
+            std::io::stdin().read_line(&mut input)?;
+            if input.trim().to_lowercase() == "y" || input.trim().to_lowercase() == "yes" {
```
gefjon, line 162: I'd like to see a `--force` flag or similar which skips this check, so that `generate --delete-files --force` can be called in scripts.

jdetter: > I'd like to see a `--force` flag or similar which skips this check, so that `generate --delete-files --force` can be called in scripts.

Added 👍

crates/cli/src/subcommands/generate/mod.rs
```diff
@@ -66,17 +67,30 @@ pub fn cli() -> clap::Command {
...
+                .long("delete-files")
+                .action(SetTrue)
+                .help("Prompts to delete files in the output directory that were not generated by this command"),
+        )
+        .arg(
+            Arg::new("force")
+                .long("force")
+                .action(SetTrue)
+                .requires("delete_files")
+                .help("If delete-files is set, this flag will not prompt before deleting files"),
```
gefjon, line 87:
```suggestion
                .help("delete-files without prompting first. Useful for scripts."),
```

crates/cli/src/subcommands/generate/mod.rs
```diff
@@ -66,17 +67,30 @@ pub fn cli() -> clap::Command {
...
                 .action(SetTrue)
                 .env("SPACETIME_SKIP_CLIPPY")
                 .value_parser(clap::builder::FalseyValueParser::new())
                 .help("Skips running clippy on the module before generating (intended to speed up local iteration, not recommended for CI)"),
         )
+        .arg(
+            Arg::new("delete_files")
+                .long("delete-files")
+                .action(SetTrue)
+                .help("Prompts to delete files in the output directory that were not generated by this command"),
```
gefjon, line 80:
```suggestion
                .help("Delete outdated generated files whose definitions have been removed from the module. Prompts before deleting unless --force is supplied."),
```

gefjon (approved):
Suggestions for phrasing in flag help messages. Code looks good; is clear to me.

Issue #662, opened 2023-12-12
Ensure that the Rust `query!` macro and the equivalent method in C# use multi column indexes


jdetter:
The following macro should work in Rust and use the associated multi-column index on `x` and `z` if there is one:

```rust
query!(|position: Position| position.x == 5 && position.z == 6)
```

- https://github.com/clockworklabs/SpacetimeDB/pull/596
- https://github.com/clockworklabs/SpacetimeDB/pull/647
- https://github.com/clockworklabs/SpacetimeDB/pull/694

PR #694, opened 2023-12-22
Adding an index selector that take in account multi-column indexes (and improve the `query!` macro)

mamcx:
# Description of Changes

[Adding the ability to match predicates](https://github.com/clockworklabs/SpacetimeDB/issues/662) like `a = 1 AND b = 2` to their (multi)column index.

2

All the test pass but the lack of a proper `query optimizer` means the logic is all over that place and we haven't exercised the full combinatory possibilities that arise on real-world queries. The `sqltest` suite can help with that, but we need to add a proper `query optimizer API to handle things In a more idiomatic & correct way.


crates/sats/src/relation.rs
```diff
@@ -170,6 +172,58 @@ impl Column {
     }
 }
 
+/// TODO: This is duplicate from `crates/lib/src/operator.rs`
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
```
mamcx, line 176: How avoid this? I tried moving `OpCmp` this but failed by cyclic dep

bfops:
@mamcx I just want to double-check - is this change backwards-compatible?

mamcx:
> @mamcx I just want to double-check - is this change backwards-compatible?

Yes is purely internal. It could manifest runtime regressions (the most distinct possibility when changing how queries are optimized).

crates/core/src/sql/compiler.rs
```diff
@@ -423,13 +425,48 @@ mod tests {
...
+        let mut tx = db.begin_mut_tx(IsolationLevel::Serializable);
+
+        // Create table [test] with index on [b]
+        let schema = &[
+            ("a", AlgebraicType::U64),
+            ("b", AlgebraicType::U64),
+            ("c", AlgebraicType::U64),
+            ("d", AlgebraicType::U64),
+        ];
+        let indexes = &[(0.into(), "a"), (1.into(), "b")];
```
joshua-spacetime, line 447: How many columns do we support? Also can we take advantage of a multi-column index in a join? It's fine if the answer is no, we just probably want to add another test if the answer is yes.

joshua-spacetime (commented):
@mamcx do you think it's possible to simplify the logic here at all? If we can make the algorithm simpler by requiring your predicate to be ordered, I say we do it.

If it's too much trouble, I'm also fine to merge it as is, since it's a big performance win. I'm just trying to keep the complexity of this code from getting out of hand.

crates/vm/src/expr.rs
```diff
@@ -181,6 +252,92 @@ impl ColumnOp {
             op => vec![op],
         }
     }
+
+    /// Flatten `Self` into a list of [ColumnOpFlat], so we can collect together all the `AND` expressions,
+    /// ie: `a = 1 AND b = 2 AND c = 3` to be `[a = 1, b = 2, c = 3]`
+    ///
+    /// It also split the kind of `queries` that *could* be answered by a `index`,
+    /// from the ones that need to be executed with a `scan`.
+    pub fn flatten_op(self) -> Vec<ColumnOpFlat> {
```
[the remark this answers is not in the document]
mamcx: > Note, ColumnOp already has a `to_vec` method which does the flattening you're describing. Can we reuse it? If not, we should delete it in favor of this one probably.

Is different. `to_vec` generate a sequence of `ColumnOp`. `flatten_op` instead collects all the `and` in a single `vec`.

crates/sats/src/relation.rs
```diff
@@ -283,6 +338,110 @@ impl Header {
             .unwrap_or(false)
     }
 
+    /// Pick the best index that matches the *permutation* of the supplied `fields`.
+    ///
+    /// This function is designed to handle complex scenarios when selecting the optimal index for a query. The scenarios include:
+    ///
+    /// - Combination of multi and single column indexes that could refer to the same field,
+    ///   so, if we could have `indexes`: `[a], [a, b]` and being asked: `a = 1 AND b = 2 AND a = 3`.
```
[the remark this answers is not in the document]
mamcx: It means do all the combinations for all the columns? it could be big in some cases...

crates/core/src/sql/compiler.rs
mamcx: > How many columns do we support?

We support up to 4 columns now. 

> We take advantage of a multi-column index in a join?

Internally, the `SQL` parser chokes with `on (a,b) = (d,c)` (not supported) and `ambiguous field` if used in a `where clause`.

crates/vm/src/expr.rs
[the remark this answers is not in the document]
mamcx: I make this to make it easier to pick at once the full chain of `AND`, so it is for representation. Can I see how you done it?

crates/bench/benches/subscription.rs
```diff
@@ -18,10 +18,27 @@ fn create_table_location(db: &RelationalDB) -> Result<TableId, DBError> {
...
+        (0.into(), "entity_id"),
+        (1.into(), "chunk_index"),
+        (2.into(), "x"),
+        (3.into(), "z"),
+        (4.into(), "dimension"),
+    ];
     db.create_table_for_test("location", schema, indexes)
 }
 
+fn create_table_location_multi_column(db: &RelationalDB) -> Result<TableId, DBError> {
```
joshua-spacetime, line 31: Instead of creating a new table, can you use the existing location table?

crates/bench/benches/subscription.rs
```diff
@@ -174,6 +198,24 @@ fn eval(c: &mut Criterion) {
...
+        "select * from location WHERE x = 0 AND z = 10000 AND dimension = 0",
+    );
+
+    // To profile this benchmark for 30s
+    // samply record -r 10000000 cargo bench --bench=subscription --profile=profiling -- query-indexes-multi --exact --profile-time=30
+    // Iterate 1M rows.
+    bench_eval(
+        c,
+        "query-indexes-multi",
+        "select * from location_multi WHERE x = 0 AND z = 10000 AND dimension = 0",
```
joshua-spacetime, line 217: We don't need both of these benchmarks if we only have the one location table.

joshua-spacetime (changes requested):
I have a small ask in regards to the newly added benchmarks.

crates/bench/benches/subscription.rs
mamcx: Done

joshua-spacetime (approved):
@mamcx LGTM

PR #695, opened 2024-01-02
fix tests for read-tx-in-subs

Shubham8287:
# Description of Changes

-  Intention of this PR is to keep diff of main PR smaller - https://github.com/clockworklabs/SpacetimeDB/pull/685 by separating trivial unit test related changes from main logic.
- naming changes `begin_tx()` -> `begin_mut_tx()` and `begin_read_tx()` -> `begin_tx()`

1


crates/core/src/sql/compiler.rs
```diff
@@ -371,15 +374,17 @@ mod tests {
 
     #[test]
     fn compile_index_eq() -> ResultTest<()> {
-        let (db, _) = make_test_db()?;
-        let mut tx = db.begin_tx();
+        let (db, _tmp) = make_test_db()?;
```
mamcx, line 377: Any reason for adding `_tmp`?

crates/core/src/sql/execute.rs
```diff
@@ -227,12 +225,11 @@ pub(crate) mod tests {
     #[test]
     fn test_select_catalog() -> ResultTest<()> {
         let (db, _, _tmp_dir) = create_data(1)?;
-        let mut tx = db.begin_tx();
+        let tx = db.begin_tx();
         let schema = db.schema_for_table(&tx, ST_TABLES_ID).unwrap().into_owned();
-
+        db.rollback_read_tx(&ExecutionContext::internal(db.address()), tx);
```
mamcx, line 230: If feels weird why this context is needed and the chance the `db.address()` is different.

crates/core/src/sql/compiler.rs
Shubham8287: `make_test_db()` creates a temporary dir (for db) which lives until its binding is in the scope.
`db.commit_tx()` below writes to this temporary dir, so `_tmp` is require to bind dir scope to current test.
Why this was not a problem before? because we were using the single `tx` in all the tests and never writting anything back to disk.  Now, we need to swtich between type of tx.

crates/core/src/sql/execute.rs
Shubham8287: Sorry, I don't think I understood this comment well, can you please eloborate more?
 
though ctx is there only for the sake that `rollback_read_tx` expects it? Can it be done in better way?

mamcx: I mean that the context of the rollback looks like it can be different from the `db.begin_tx()`. Maybe I'm misreading the API but why is it not just `rollback_read_tx()`?

Shubham8287: Agree, It feels weird and may be 'ctx' should be binded to `tx` lifecycle but right now, code is following https://github.com/clockworklabs/SpacetimeDB/blob/master/crates/core/src/db/datastore/traits.rs#L69-#L71 current implmented trait. I feel usage of ctx needs to be fixed for datastore in general as there are incosistencies between methods expecting ctx.
I think most of it is out of scope of this PR and should be done separately.

mamcx (approved):
LGTM

crates/core/src/sql/compiler.rs
cloutiertyler: Note, there is a semantic difference between `_` and `_tmp` for lifetimes

cloutiertyler (approved):
This PR also looks good to me. Thanks for making this uniform.

PR #696, opened 2024-01-02
Inline typerefs

RReverser:
# Description of Changes

Database is currently unable to handle algebraic type references in table fields - that especially covers recursive types, but breaks even simple cases like type aliases that are stored in the typespace and referenced from multiple other types.

This was breaking C# SDK tests (upcoming in a separate PR to keep things simple to review), and it's relatively easy to make a repro that would break Rust modules too.

A proper support for arbitrary references and mutual recursion would be complex and out of scope, so for now we talked with @gefjon and agreed on a simpler workaround where all type references in tables are recursively inlined as their destination types during init. This allows to gracefully report errors on any recursion and support type references otherwise.

corresponding GitHub label.

2 - there is a risk I might've missed some types, but otherwise it's similar to what Rust modules already do in their type descriptions.

### DIFF
--- a/crates/sats/src/typespace.rs
+++ b/crates/sats/src/typespace.rs
@@ -3,8 +3,19 @@ use std::ops::{Index, IndexMut};
 
 use crate::algebraic_type::AlgebraicType;
 use crate::algebraic_type_ref::AlgebraicTypeRef;
-use crate::WithTypespace;
 use crate::{de::Deserialize, ser::Serialize};
+use crate::{BuiltinType, WithTypespace};
+
+#[derive(thiserror::Error, Debug)]
+pub enum TypeRefError {
+    // TODO: ideally this should give some useful type name or path.
+    // Figure out if we can provide that even though it's not encoded in SATS.
+    #[error("Found recursive type reference {0}")]
+    RecursiveTypeRef(AlgebraicTypeRef),
+
+    #[error("Type reference {0} out of bounds")]
+    InvalidTypeRef(AlgebraicTypeRef),
+}
 
 /// A `Typespace` represents the typing context in SATS.
 ///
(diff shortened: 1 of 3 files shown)


crates/sats/src/typespace.rs
```diff
@@ -86,6 +102,73 @@ impl Typespace {
...
+        Ok(())
+    }
+
+    /// Inlines all nested references behind the current [`AlgebraicTypeRef`] recursively using the current typeset.
+    ///
+    /// Returns the fully-resolved type or an error if the type reference is invalid or self-referential.
+    fn inline_typerefs_in_ref(&mut self, r: AlgebraicTypeRef) -> Result<&AlgebraicType, TypeRefError> {
+        let resolved_ty = match self.get_mut(r) {
+            None => return Err(TypeRefError::InvalidTypeRef(r)),
+            Some(AlgebraicType::Ref(_)) => return Err(TypeRefError::RecursiveTypeRef(r)),
```
gefjon, line 143: Should this be checking that the `Ref` you extracted is equal to `r`? Is there ever a case where we'll have two non-equal refs `A` and `B` such that `A` refers to `B`?

gefjon:
```suggestion
            // If `r` refers to itself, that means an enclosing call to `inline_typerefs_in_ref(r)` 
            // swapped its definition out,
            // i.e. the type referred to by `r` is recursive.
            // Our database can't handle recursive types, so return an error.
            // TODO: support recursive types in the future.
            Some(AlgebraicType::Ref(_)) => return Err(TypeRefError::RecursiveTypeRef(r)),
```

gefjon (approved):
One question, and one suggestion of a place where I felt a comment was warranted. (Coincidentally on the same line.) Otherwise looks good. Ideally I'd like to see an integration test added which exercises this behavior - perhaps a `cs-wasm-test` which stores a `SpacetimeType` struct/class inside a table - but I'm assuming you've tested manually and verified that you can now run such a module.

crates/sats/src/typespace.rs
RReverser: Added a variation of your comment that explains any of the parent calls might've swapped it out.

gefjon: No, I mean, could we have a non-recursive typespace like:

```
0 -> 1
1 -> U32
```

PR #697, opened 2024-01-02
Update to axum 0.7

coolreader18:
# Description of Changes

Had this on the backburner for a bit, whoops. This brings us to depending on Hyper 1.0.
Hyper 0.14 is still in the dependency tree, but once reqwest updates we'll be
fully moved over.

2

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -179,19 +179,19 @@ checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
 
 [[package]]
 name = "axum"
-version = "0.6.20"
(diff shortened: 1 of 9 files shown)


mamcx:
Code-wise the PR looks fine, but what is a "2" risk? Does this need special testing?

coolreader18:
No, not really, I'm confident that there's no issues with this. I just figured it's a core dependency and Tyler said that he's hesitant about updating stuff cause it can introduce bugs, but really these updates should be *more* stable so this should probably be more of a 1.

mamcx (approved):
LGTM.

I think to mitigage any risk, merge it with enough time before a release.

PR #699, opened 2024-01-03
spacetime local clear command impl

jdetter:
# Description of Changes

- Added a `spacetime local clear [--force]` which allows you to clear your local spacetimedb database without having to manually `rm -rf ~/.spacetime`
- Moved `control_db` and `worker_node` into a base `~/.spacetime/data/` directory

corresponding GitHub label.

- This is API breaking because of filesystem changes

1


crates/cli/src/subcommands/local.rs
```diff
@@ -0,0 +1,69 @@
...
+}
+
+pub async fn exec(config: Config, args: &ArgMatches) -> Result<(), anyhow::Error> {
+    let (cmd, subcommand_args) = args.subcommand().expect("Subcommand required");
+    exec_subcommand(config, cmd, subcommand_args).await
+}
+
+fn get_subcommands() -> Vec<Command> {
+    vec![Command::new("clear")
+        .about("Deletes all data from the local database")
```
gefjon, line 24:
```suggestion
        .about("Deletes all data from all local databases")
```

Existing language is unclear; does it delete a single module-and-db or all of them?

gefjon (changes requested):
The new `local clear` command seems great.

I'm curious why you made the change to the local data path. In general it seems like a major breakage without much benefit, and if we want to make that change, it seems important enough to require its own PR with separate review.

I'd strongly prefer to split this into two PRs, one for `local clear` and the other changing the storage path.

bfops (changes requested):
bunch of nitpicks, otherwise looks good as far as I can tell :heavy_check_mark:

crates/standalone/src/lib.rs
```diff
@@ -653,7 +653,7 @@ impl StandaloneEnv {
             .context("failed to load module program")?
             .ok_or_else(|| anyhow!("missing object: {}", database.program_bytes_address.to_hex()))?;
 
-        let root_db_path = stdb_path("worker_node/database_instances");
+        let root_db_path = stdb_path("data/worker_node/database_instances");
```
cloutiertyler, line 656: This is a breaking change, yes? Could you please add to the PR description an exhaustive list of all the files that you expect to be in the `.spacetime` directory. 

Might need to get Kim's buy in because this could also break cloud maybe?

cloutiertyler (changes requested):
See single comment.

jdetter:
closing in favor of https://github.com/clockworklabs/SpacetimeDB/pull/702

crates/standalone/src/lib.rs
jdetter: Will make a proposal for this 👍

PR #701, opened 2024-01-04
Print logs to stdout instead of stderr

jdetter:
# Description of Changes

- We've been printing the output of `spacetime logs ...` to stderr. This isn't correct because the purpose of the `spacetime logs ...` command is to print logs to the console, so the logs themselves are considered normal/expected output which shouldn't go to stderr. Also printing the logs to stderr is terrible UX because then if you want to search for something in the logs using grep then you have to pipe stderr into stdout which is a huge pain.

corresponding GitHub label.

- This is technically an API break because people may have assumed up to this point that we're printing the logs to stderr so this change could technically break CI scripts and such.

1


bfops (approved):
LGTM, but I don't know enough to be worried about unintended side effects, e.g. pipelines that depend on current behavior :shrug:

PR #702, opened 2024-01-05
spacetime local clear subcommand

jdetter:
# Description of Changes

- I separated out the functionality from https://github.com/clockworklabs/SpacetimeDB/pull/699 where you can clear the local database. This PR does *not* break the ABI or API.
- Added a `spacetime local clear [--force]` which allows you to clear your local spacetimedb database without having to manually `rm -rf ~/.spacetime`

corresponding GitHub label.

- This PR does not break the ABI or API

1


gefjon (approved):
Excellent!

crates/cli/src/subcommands/local.rs
```diff
@@ -0,0 +1,104 @@
...
+            println!("Control node database: {}", control_node_dir.to_str().unwrap());
+        } else {
+            println!("Control node database: <not found>");
+        }
+        if worker_node_dir.exists() {
+            println!("Worker node database: {}", worker_node_dir.to_str().unwrap());
+        } else {
+            println!("Worker node database: <not found>");
+        }
+    }
```
bfops, line 102: this seems to be the `else` case for `control_node_dir.exists() || worker_node_dir.exists()` above? If so, wouldn't that mean that neither of the `exists` checks should pass..?

(also, is this overall `if` layer even needed? if neither directory exists, nothing really breaks, it just doesn't do much.)

crates/cli/src/subcommands/local.rs
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        if control_node_dir.exists() {
+            std::fs::remove_dir_all(&control_node_dir)?;
+            println!("Deleted control node database: {}", control_node_dir.to_str().unwrap());
+        }
+        if worker_node_dir.exists() {
+            std::fs::remove_dir_all(&worker_node_dir)?;
+            println!("Deleted worker node database: {}", worker_node_dir.to_str().unwrap());
+        }
```
bfops, line 90: Similar comment to the one I made on https://github.com/clockworklabs/SpacetimeDB/pull/699#discussion_r1442135585. I would personally move to consolidate the "go" logic.

bfops (changes requested):

crates/cli/src/subcommands/local.rs
jdetter: I looked at the suggestion on the other PR and I think I've done what you're suggesting - whether you supply `--force` or not this is the codepath which removes the files. Could you give me an example of what you mean here?

crates/cli/src/subcommands/local.rs
jdetter: Good catch, fixed 👍

crates/cli/src/subcommands/local.rs
bfops: Ah, sorry, maybe just a bit left over - we still call `remove_dir_all` further up. I would replace the code on lines 68-78 with:
```rust
            let yes = input.trim().to_lowercase() == "y" || input.trim().to_lowercase() == "yes"
            if ! yes {
                println!("Aborting");
                return Ok(());
            }
```

bfops (approved):
LGTM!

PR #705, opened 2024-01-08
Create output directory in `generate` if it doesn't exist

RReverser:
# Description of Changes

This is a minor change but has bothered me for as long as I remember - I believe there is no reason to error out if directory doesn't exist; most users will expect the tool to create the output dir, so we can do just that.

corresponding GitHub label.

1


jdetter:
This also really bothers me, this is a good change 👍

jdetter (approved):
LGTM

PR #706, opened 2024-01-08
C# SDK tests

RReverser:
# Description of Changes

Ports SDK tests to C#.

In the process had to modify the test runner to cache compilation results and not just binding generation results - this was necessary because .NET doesn't expect same project to be built in parallel via same commands, but as a side benefit speeds up both Rust and C# tests as it doesn't even have to try to build same project again and instead reuses prebuilt Wasm for all other stages.

Note that this also modifies the `publish` command so that if project path is a file that ends in `.wasm` then it uses it as-is instead of building it as a dir containing a project. (cc @cloutiertyler can you confirm if it's okay?)

corresponding GitHub label.

2


crates/cli/src/subcommands/publish.rs
```diff
@@ -137,7 +137,11 @@ pub async fn exec(mut config: Config, args: &ArgMatches) -> Result<(), anyhow::E
         query_params.push(("trace_log", "true"));
     }
 
-    let path_to_wasm = crate::tasks::build(path_to_project, skip_clippy, build_debug)?;
+    let path_to_wasm = if !path_to_project.is_dir() && path_to_project.extension().map_or(false, |ext| ext == "wasm") {
+        path_to_project.clone()
+    } else {
+        crate::tasks::build(path_to_project, skip_clippy, build_debug)?
+    };
```
gefjon, line 144: Could you extend the `.help` message (or add a `.long_help`) to the `-p` flag which describes this behavior?

crates/testing/src/sdk.rs
```diff
@@ -34,41 +35,9 @@ pub fn ensure_standalone_process() {
...
-    // That would be more complicated, as we'd need to re-write dependencies
-    // on the client language's SpacetimeDB SDK to use a local absolute path.
-    // Doing so portably across all our SDK languages seemed infeasible.
-    static ref BINDINGS_GENERATED: Mutex<HashSet<String>> = Mutex::new(HashSet::new());
-}
-
 pub struct Test {
     /// A human-readable name for this test.
+    #[allow(dead_code)] // TODO: should we just remove this now that it's unused?
     name: String,
```
gefjon, line 41: Do we not print it when a test fails?

gefjon (approved):
Perhaps you should hold off merging until @cloutiertyler weighs in on the new extension to `spacetime publish`, but it seems reasonable to me. The tests look good, and I like how you've memoized client compilation. Thanks for doing this; I'm glad we're getting tests in to verify that C# and Rust modules are interchangeable.

crates/testing/src/sdk.rs
RReverser: No, Cargo uses test names from corresponding functions. Before this PR this field was used in compilation messages, but now that they're memoized and shared between tests, it seemed misleading to use arbitrary test name (whichever happens first) in that context, so I changed them to just messages like "(compiling)".

Now this field is completely unused, but it's assigned in a bunch of spaces and I wasn't sure whether it's worth removing it altogether from all those places or whether we want to use it again in the future.

cloutiertyler:
> Note that this also modifies the `publish` command so that if project path is a file that ends in `.wasm` then it uses it as-is instead of building it as a dir containing a project. (cc @cloutiertyler can you confirm if it's okay?)

I don't see a problem with this per se, but I also don't see why we couldn't just have a different parameter for --wasm-module instead of --project-path.

crates/cli/src/subcommands/publish.rs
cloutiertyler: See my comment above.

RReverser:
> I don't see a problem with this per se, but I also don't see why we couldn't just have a different parameter for --wasm-module instead of --project-path.

We could, but mainly not sure of the value of having separate parameters, given that they're mutually exclusive and mean more or less the same thing (path to the project). We already auto-detect language of the project (Rust vs C#), so detecting precompiled Wasm seems in line with that as just another language.

cloutiertyler:
> > I don't see a problem with this per se, but I also don't see why we couldn't just have a different parameter for --wasm-module instead of --project-path.
> 
> We could, but mainly not sure of the value of having separate parameters, given that they're mutually exclusive and mean more or less the same thing (path to the project). We already auto-detect language of the project (Rust vs C#), so detecting precompiled Wasm seems in line with that as just another language.

In that case, I would leave it to your discretion. Keep in mind that it's a decision we'll have to live with for a long time.

PR #708, opened 2024-01-08
fix(707): Collect table size metrics again

### LINKED ISSUE #707: Table size metrics are broken
Prometheus is no longer collecting table size metrics.

joshua-spacetime:
Fixes #707.

Previously the table size metric definition was moved to a different crate. This was to facilitate its use in query optimization. However as part of that change it was never added back into the prometheus registry. This resulted in the metric no longer being collected by prometheus. This patch reintroduces it back into the registry, forcing its collection, again.

# Description of Changes

corresponding GitHub label.


jdetter (approved):
LGTM 👍

PR #709, opened 2024-01-09
fix(684): Reset max value metrics

### LINKED ISSUE #684: Max cpu time is not resetting
Image

The maximum transaction cpu time is not resetting. Note that the maximum queue length is resetting which is the correct behavior.

Image

joshua-spacetime:
Fixes #684.

Prometheus does not reset metrics.
It only updates them with new values that it collects. If it does not receive a new value for a metric, it will keep the old value.

Therefore it is important to periodically reset certain metrics. Namely gauge metrics that only receive updates occasionally.

Maximum reducer duration is one such example,
since some reducers run much more infrequently than others.

# Description of Changes

corresponding GitHub label.

### DIFF
--- a/crates/core/src/db/datastore/locking_tx_datastore/mod.rs
+++ b/crates/core/src/db/datastore/locking_tx_datastore/mod.rs
@@ -11,8 +11,6 @@ use self::{
     sequence::Sequence,
     table::Table,
 };
-use std::collections::hash_map::DefaultHasher;
-use std::hash::Hasher;
 use std::ops::Deref;
 use std::time::{Duration, Instant};
 use std::{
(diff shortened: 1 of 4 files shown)


crates/core/src/db/datastore/locking_tx_datastore/mod.rs
```diff
@@ -2399,18 +2397,9 @@ impl traits::MutTx for Locking {
...
-            a.hash(&mut hasher);
-            b.hash(&mut hasher);
-            c.hash(&mut hasher);
-            hasher.finish()
-        }
-
         let mut guard = MAX_TX_CPU_TIME.lock().unwrap();
         let max_cpu_time = *guard
-            .entry(hash(workload, db, reducer))
+            .entry((*db, *workload, reducer.to_owned()))
```
joshua-spacetime, line 2402: This was an optimization to avoid extra allocations when recording metrics. But I'd rather take the allocations than deal with a potential hash collision when debugging a performance issue.

cloutiertyler (approved):
LGTM

PR #714, opened 2024-01-10
Add small 'realistic' workload benchmark integration test

mamcx:
# Description of Changes

A small integration test that benchmarks based on "realistic" workloads we are focusing on improving.

1


joshua-spacetime:
Is there a reason you put these in their own module rather than in `modules/benchmarks`? Also can you add one for updates?

cloutiertyler:
> Is there a reason you put these in their own module rather than in `modules/benchmarks`? Also can you add one for updates?

@mamcx I agree with Joshua here. This should not require a new module I don't think.

mamcx:
> Is there a reason you put these in their own module rather than in `modules/benchmarks`? Also can you add one for updates?

The original is tied into the `criterion` benchmarks in `crates/bench`.

kazimuth:
benchmarks please :)

crates/bench/benches/special.rs
```diff
@@ -48,6 +51,31 @@ fn custom_module_benchmarks(c: &mut Criterion) {
     }
 }
 
+fn custom_db_benchmarks(c: &mut Criterion) {
+    let runtime = start_runtime();
+
+    let config = Config {
+        storage: Storage::Memory,
```
kazimuth, line 58: Might be worth running these on disk?

You could also split them out:
```rust
for storage in [Storage::Memory, Storage::OnDisk] {
   let config = /* ... */
   // ...
   let storage_label = match storage {
     Storage::Memory => "in_memory",
     Storage::Disk => "on_disk"
   };
   group.bench_function(&format!("db_game/{storage_label}/load={n}"), |b| { /* ... */ });
}
```

the nice thing about this is that you can still ignore some of these settings when running criterion:
`cargo bench --bench special -- 'db_game/in_memory/*'` will only run the in-memory ones.

kazimuth (approved):
This looks great! No complaints from me aside from one comment.

crates/bench/benches/special.rs
kazimuth: On the other hand, this is gonna be pretty slow -- 8 minutes for (load=10, load=100) x (on_disk, in_memory) -- so it may be worth turning the measurement time down...

crates/bench/benches/special.rs
```diff
@@ -48,6 +51,31 @@ fn custom_module_benchmarks(c: &mut Criterion) {
...
+    let module = runtime.block_on(async { BENCHMARKS_MODULE_DB.load_module(config).await });
+    let mut group = c.benchmark_group("special");
+    // This bench take long, so adjust for it
+    group.measurement_time(Duration::from_secs(60 * 2));
+    group.sample_size(10);
+
+    for n in [10, 100] {
+        let args = sats::product![n];
+        group.bench_function(&format!("db_game/load={n}"), |b| {
+            b.iter_batched(
```
kazimuth, line 70: Instead of doing `iter_batched`, try using `iter_custom` here. Then, you can:
- initialize the database
- start a timer
- run the `ai_agent_loop` `n` times (`n` will be passed as an argument to your `iter_custom` closure), advancing the current timestamp passed to the benchmark as you go.
- end the timer and return

Similar to how you were doing before.
In particular, I think right now this benchmark is mainly measuring initialization time. Adding more loops of the ai agent loop will allow us to measure that instead.

modules/benchmarks_db/src/lib.rs
```diff
@@ -0,0 +1,529 @@
...
+    println!("ENEMY IA LOOP PLAYERS: {players}, processed: {count}");
+}
+
+#[spacetimedb(reducer)]
+pub fn game_benchmark(initial_load: u8) {
+    let initial_load = initial_load as u32;
+    let small_table = initial_load * 10;
+    let num_players = (initial_load * 100) as u64;
+    let big_table = initial_load * 10_000;
+    let biggest_table = initial_load * 20_000;
```
kazimuth, line 513: These numbers mean that the main thing we're benchmarking here are operations on big_table and biggest_table.

Do we know what the sizes of the corresponding tables look like in bitcraft?

kazimuth (changes requested):
Actually, wait, I was running this locally and I thought of a few things that might be helpful.

Cargo.toml
```diff
@@ -20,6 +20,7 @@ members = [
   "crates/primitives",
   "modules/rust-wasm-test",
   "modules/benchmarks",
+  "modules/benchmarks_db",
```
cloutiertyler, line 23: I don't see why this can't just go in the `benchmarks` module. At the very least you should create a README which explains what the difference between `benchmarks` and `benchmarks_db` is because it's not at all obvious based on the name.

cloutiertyler (changes requested):

crates/testing/tests/standalone_integration_test.rs
```diff
@@ -139,3 +140,53 @@ fn test_call_query_macro() {
...
+
+    module.send(json).await.unwrap();
+
+    now.elapsed()
+}
+
+#[test]
+#[serial]
+#[allow(clippy::disallowed_macros)]
+fn test_calling_bench_db() {
```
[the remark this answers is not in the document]
mamcx: One of the reason of the large numbers is because this reduces the impact of noise, and makes bigger the evidence of perf improvement/regresions.

Micro-benchmarks are less useful for checking our db engine, where is not kilobytes, but mega/giga data sizes we should start checking for.

It is not what criterion is for, but that is the bench suite we have.

crates/bench/benches/special.rs
mamcx: I have done as suggested.

Cargo.toml
mamcx: Ok, I have refactor the bench to be in the same module

mamcx: benchmarks please :)

crates/testing/tests/standalone_integration_test.rs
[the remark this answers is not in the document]
kazimuth: That's totally fair. I think this test is why the test suite is crashing below, though.

cloutiertyler (approved):
This LGTM now!

modules/benchmarks/src/synthetic.rs
```diff
@@ -0,0 +1,448 @@
+//! STDB module used for benchmarks.
```
kazimuth, line 1: if you're renaming this file you also need to adjust the config used in crates/bench/src/spacetime_module

kazimuth: also you'll need to rebase, everything in here has been renamed

kazimuth (changes requested):
In general I still think it still to use numbers higher than ten thousand in a criterion benchmark

crates/bench/benches/special.rs
mamcx: I fix them now so the test finishes earlier...

kazimuth:
benchmarks please

kazimuth (approved):
👍 Sorry for the delay on review I missed the re-up.

PR #715, opened 2024-01-10
Fix benchmark actions

kazimuth:
# Description of Changes

This fixes two things:

### Let other people use the actions runner after we're done with it
Specifically, we chmod the git repo directory to a+rw after we touch it in the callgrind benchmarks action.
This is needed because the callgrind benchmarks run as root in a docker container. Github Actions tries to reuse the resulting directory (and I guess stores it on the host?) afterwards, but now the git repo is owned by root and can't be modified, which causes other benchmarks to break.
This issue was pointed out by @coolreader18.

(Note: we can't just chown the git repo, because we're inside a docker image, so who would we chown to?)

### Remove residual references to moved code in the benchmarks & callgrind benchmarks actions.
I pulled out the `markdown-report[-callgrind]` commands from the `benches/summarize` bin in my last callgrind PR -- they've moved to another repo -- but there were residual references to them in these ymls, despite the resulting reports not being used. I'm not sure why that didn't cause errors for that PR, but the residual references are gone now.

This issue was pointed out by @cloutiertyler.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -127,12 +127,6 @@ jobs:
       #        })
       #      }
 
-      - name: PR; compare benchmarks
-        if: ${{ env.PR_NUMBER }}
-        working-directory: crates/bench/
-        run: |
-          cargo run --bin summarize markdown-report branch.json --report-name report
-
       # this will work for both PR and master
       - name: Upload criterion results to DO spaces
         uses: shallwefootball/s3-upload-action@master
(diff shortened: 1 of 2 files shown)


kazimuth:
callgrind please

kazimuth:
callgrind please

kazimuth:
benchmarks please

kazimuth:
Wait, damn, it's using the `master` actions yaml so it's not seeing the changes. I'll try running manually

kazimuth:
Here's the manual runs with the ymls from this branch:
Callgrind: https://github.com/clockworklabs/SpacetimeDB/actions/runs/7492793066
Regular: https://github.com/clockworklabs/SpacetimeDB/actions/runs/7492802811

coolreader18 (approved):
LGTM!

PR #717, opened 2024-01-12
chore: Remove unused metrics

joshua-spacetime:
# Description of Changes

corresponding GitHub label.

### DIFF
--- a/crates/core/src/db/relational_db.rs
+++ b/crates/core/src/db/relational_db.rs
@@ -494,7 +494,6 @@ impl RelationalDB {
         tx: &'a MutTx,
         table_id: TableId,
     ) -> Result<Iter<'a>, DBError> {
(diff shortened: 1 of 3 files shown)


crates/core/src/db/relational_db.rs
```diff
@@ -494,7 +494,6 @@ impl RelationalDB {
         tx: &'a MutTx,
         table_id: TableId,
     ) -> Result<Iter<'a>, DBError> {
-        let _guard = DB_METRICS.rdb_iter_time.with_label_values(&table_id.0).start_timer();
```
joshua-spacetime, line 497: This metric was was used, but it did not measure actual iteration time, just the time to open an iterator.

jdetter (approved):
LGTM 👍

PR #724, opened 2024-01-16
Implement disk usage metering

coolreader18:
# Description of Changes

Apologies about the energy types refactor - I can try and pull it out if we'd prefer to make it as easy as possible to merge for Mario's PR.

2


crates/core/src/host/wasm_common/module_host_actor.rs
```diff
@@ -585,7 +583,7 @@ impl<T: WasmInstance> WasmModuleInstance<T> {
                 // discard this instance
                 self.trapped = true;
 
-                if energy.remaining == EnergyQuanta::ZERO {
+                if energy.remaining.get() == 0 {
```
joshua-spacetime, line 586: Should this be `<= 0`?

crates/core/src/host/wasmtime/mod.rs
```diff
@@ -91,34 +91,24 @@ struct WasmtimeFuel(u64);
 
 impl WasmtimeFuel {
     /// 1000 energy quanta == 1 wasmtime fuel unit
-    const QUANTA_MULTIPLIER: i128 = 1_000;
-
-    /// Convert from EnergyQuanta to wasmtime fuel. The second element of the tuple is the remainder
-    /// of quanta in the case that `eq` isn't in u64 range. It can be safely ignored if the supplied
-    /// budget isn't being bookkept.
-    fn from_energy_quanta(eq: EnergyQuanta) -> (Self, EnergyQuanta) {
```
joshua-spacetime, line 99: It doesn't look like the below logic has been moved anywhere else. Why?

crates/standalone/src/energy_monitor.rs
```diff
@@ -48,15 +38,33 @@ impl EnergyMonitor for StandaloneEnergyMonitor {
                 .upgrade()
                 .expect("Worker env was dropped.")
         };
-        tokio::spawn(async move {
-            standalone_env
-                .withdraw_energy(&module_identity, energy_used.as_quanta())
-                .await
-                .unwrap();
-        });
+        tokio::spawn(async move { standalone_env.withdraw_energy(&identity, amount).await.unwrap() });
```
joshua-spacetime, line 41: Why do we need to spawn a task in order to withdraw energy?

crates/standalone/src/energy_monitor.rs
```diff
@@ -48,15 +38,33 @@ impl EnergyMonitor for StandaloneEnergyMonitor {
...
+        energy_used: EnergyQuanta,
+        _execution_duration: Duration,
+    ) {
+        self.withdraw_energy(fingerprint.module_identity, energy_used)
+    }
+
+    fn record_disk_usage(&self, database: &Database, _instance_id: u64, disk_usage: u64, period: Duration) {
+        let amount = EnergyQuanta::from_disk_usage(disk_usage, period);
+        self.withdraw_energy(database.identity, amount)
+    }
```
joshua-spacetime, line 63: Note we need to be able to identify somewhere that this energy came from disk usage. See here for context. This could be recorded in `EnergyQuanta` itself or as an extra parameter to `withdraw_energy`.

crates/core/src/database_instance_context.rs
```diff
@@ -79,47 +45,80 @@ impl DatabaseInstanceContext {
...
+
+    /// Obtain an array which can be summed to obtain the total disk usage.
+    ///
+    /// Some sources of size-on-disk may error, in which case the corresponding array element will be None.
+    pub fn total_disk_usage(&self) -> TotalDiskUsage {
+        TotalDiskUsage([
+            Some(self.message_log_size_on_disk()),
+            // the errors get logged by the functions, we're not discarding them here without logging
+            self.object_db_size_on_disk().ok(),
+            self.log_file_size().ok(),
```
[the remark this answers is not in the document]
coolreader18: It sorta makes the most sense this way I feel. Plus `tracing::instrument(err)` wouldn't work if they returned options.

crates/core/src/database_instance_context.rs
```diff
@@ -79,47 +45,80 @@ impl DatabaseInstanceContext {
...
+impl Deref for DatabaseInstanceContext {
+    type Target = Database;
+
+    fn deref(&self) -> &Self::Target {
+        &self.database
+    }
+}
+
+#[derive(Copy, Clone, Default)]
+pub struct TotalDiskUsage(pub [Option<u64>; 3]);
```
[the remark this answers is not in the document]
coolreader18: Oh, yeah, that makes sense

crates/core/src/energy.rs
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    #[inline]
+    pub fn get(&self) -> i128 {
+        self.0
+    }
+
+    pub fn from_disk_usage(bytes_stored: u64, storage_period: Duration) -> Self {
+        // TODO: this line is lossy if bytes_stored is >1.6 PiB. do we ever care about that case?
+        let energy = bytes_stored as f64 * storage_period.as_secs_f64();
```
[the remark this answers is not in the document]
coolreader18: Can't multiply them otherwise. We can't just discard the fractional component of the duration, since with a large enough bytes_stored a couple of milliseconds more storage time is a non-negligible amount of energy. And, afaict, there's no way to multiply a i128 and a f64 natively/with high precision. Oh, I guess I could just do it manually - `bytes_stored / 1_000_000_000 * ns + bytes_stored * s`. (Though also to note - the potential inaccuracy of the float op would only be about 1eV, since float accuracy is in stages of powers of 2)

crates/core/src/host/wasm_common/module_host_actor.rs
coolreader18: It's an unsigned integer

crates/core/src/energy.rs
```diff
@@ -0,0 +1,127 @@
...
+        execution_duration: Duration,
+    );
+    fn record_disk_usage(&self, database: &Database, instance_id: u64, disk_usage: u64, period: Duration);
+}
+
+// what would the module do with this information?
+// pub enum EnergyRecordResult {
+//     Continue,
+//     Exhausted { quanta_over_budget: u64 },
+// }
```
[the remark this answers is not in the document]
coolreader18: :shrug: I just copied it from host.rs

crates/core/src/host/wasmtime/mod.rs
coolreader18: with ReducerBudget, this conversion logic is unnecessary

crates/standalone/src/energy_monitor.rs
coolreader18: Right - I figured that's something that could be done as a follow-up; I wasn't sure how to do that since right now it seems the sole place energy usage is being stored is the controldb as just an integer of the total balance, and also since this is standalone it shouldn't really matter too much.

crates/core/src/energy.rs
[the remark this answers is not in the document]
joshua-spacetime: I say let's remove it then :)

crates/standalone/src/energy_monitor.rs
joshua-spacetime: I'm fine with that as long as this lands before @mamcx's.

crates/core/src/energy.rs
```diff
@@ -0,0 +1,127 @@
...
+
+use crate::messages::control_db::Database;
+
+/// [EnergyQuanta] represents an amount of energy in a canonical unit.
+/// It represents the smallest unit of energy that can be used to pay for
+/// a reducer invocation. We will likely refer to this unit as an "eV".
+///
+/// NOTE: This is represented by a signed integer, because it is possible
+/// for a user's balance to go negative. This is allowable
+/// for reasons of eventual consistency motivated by performance.
```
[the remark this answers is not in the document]
joshua-spacetime: > may as well just not run the reducer at all

This is true. But if ReducerBudget corresponds to an EnergyBalance, then it _could_ be signed since EnergyBalance is signed.

But the bigger question is why is EnergyQuanta signed? If EnergyQuanta is how we measure energy consumption, it will never be negative.

Although I don't know how much it will expand the diff. But I figure if you're already making the corresponding change to ReducerBudget, then you might as well do the same thing for EnergyQuanta. WDYT?

joshua-spacetime (commented):
This looks good to go. Just one remaining question on the type of EnergyQuanta.

crates/core/src/energy.rs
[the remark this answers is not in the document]
coolreader18: Well, there's the comment that explains why it's signed, but yeah, I figure EnergyQuanta being unsigned would make a ton of things much simpler, and maybe we could have an EnergyBalance that's signed. I was hesitant to do that in this PR, cause I didn't want to cause too much churn, but if you also think it'd be good I can definitely do it.

joshua-spacetime (approved):

PR #725, opened 2024-01-16
fix: Reset max queue length and reducer wait time metrics

joshua-spacetime:
Max value metrics are currently reset to zero after each collection period. However this won't be reflected in the metrics store immediately.

Once the database records new values for said metrics, only then will the reset be reflected in the metrics store.

This subtlety is imperceptible for metrics that are frequently updated. But if a metric is not frequently updated,
or for periods of reduced load in general,
the metrics store may present stale max value metrics.

This anomaly was thought to have been fixed by #709. However it only fixed it for DB_METRICS.
This patch removes the anomaly for WORKER_METRICS as well.

# Description of Changes

corresponding GitHub label.

### DIFF
--- a/crates/core/src/host/scheduler.rs
+++ b/crates/core/src/host/scheduler.rs
@@ -1,12 +1,10 @@
-use std::collections::hash_map::DefaultHasher;
-use std::hash::Hasher;
 use std::path::Path;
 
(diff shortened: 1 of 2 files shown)


jdetter (approved):
Thanks for this Josh 👍

PR #727, opened 2024-01-17
core: Sort constraint definitions before comparing

kim:
When updating a database, we may falsely report schemas as not
compatible due to the `ConstraintDef`s coming out in different orders in
the proposed and existing schemas.

Not sorting them was an oversight in 9bb2b215.

1

### DIFF
--- a/crates/core/src/db/update.rs
+++ b/crates/core/src/db/update.rs
@@ -1,6 +1,6 @@
 use core::fmt;
 use std::borrow::Cow;
-use std::collections::{BTreeMap, HashMap};
+use std::collections::{BTreeMap, BTreeSet, HashMap};
 
 use anyhow::Context;
 use spacetimedb_primitives::Constraints;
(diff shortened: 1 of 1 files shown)


kurtismullins (approved):
@kim I am going to stamp this but won't pretend to understand the difference :) May I steal some of your time soon to get a walk-through of the code that gives some context? I've been meaning to ask you, anyways :)

crates/core/src/db/update.rs
```diff
@@ -182,14 +182,21 @@ fn equiv(a: &TableDef, b: &TableDef) -> bool {
...
+
+    // Constraints may not be in the same order, depending on whether the
+    // `TableDef` was loaded from the db catalog or the module.
+    fn as_set(constraints: &[ConstraintDef]) -> BTreeSet<&ConstraintDef> {
+        constraints
+            .iter()
+            .filter(|c| c.constraints != Constraints::unset())
+            .collect()
+    }
+
```
mamcx, line 194: Could be better to do this change inside `TableDef`?:
```rust
pub struct TableDef {
    pub table_name: String,

    pub constraints: BtreeSet<ConstraintDef>,
}
```

cloutiertyler: Do we actually want to store them like that? We might just want this function as a way to compare them. I think I'm going to propose that we merge this and consider that for future work.

cloutiertyler (approved):
This LGTM

crates/core/src/db/update.rs
kim: It threw me off a bit to realize that there can be unset constraints in the table def, so we would still need an element-wise comparison if it was a set already.

PR #728, opened 2024-01-17
Expand callgrind benches

kazimuth:
# Description of Changes

When we're not fighting sled, the callgrind benches run extremely quickly, so I've added more. In also made the short names visible in the terminal more readable & used the same load curve (1,4,16,64) for all benchmarks.

0


cloutiertyler (approved):
Seems fine to me!

PR #730, opened 2024-01-18
Add `ColList`: a compact replacement of `NonEmpty<ColId>`, 8 vs 32 bytes

Centril:
# Description of Changes

Adds `ColList`, a compact 8 byte replacement for `NonEmpty<ColId>` which was 32 bytes.
The former takes advantage of the fact that `ColId`s tend to be very low numbers for any practical table.
This allows us to represent `[ColId(2)]` as `0b1001` where the first bit is a sentinel to signal that we use a bitset and 1 marks the 2-index bit. For `ColList`s of up to the 62 columns, we can represent this inline. In the unlikely even that we need to represent `ColId(63)` or beyond, we fall back to a version of a `TinyVec` which is 8 byte sized.

The benefit of this is that we avoid more heap allocations, reduce sizes + memory footprint, and in my local version of mem-arch-prototype, this turned out to be a win for insertions into a table with an index.

This also gets rid of `nonempty` as a dependency and the unnecessary slim_slice version of it.

Proptests are provided which have been run under miri.

None; this does not cross from/to the WASM boundary.

2 - the `ColList` changes contain some unsafe; the rest is localized or mechanical changes.

### DIFF
--- a/crates/primitives/src/col_list.rs
+++ b/crates/primitives/src/col_list.rs
@@ -1,0 +1,554 @@
+#![forbid(unsafe_op_in_unsafe_fn)]
+
+extern crate alloc;
+
+use crate::ColId;
+use alloc::alloc::{alloc, dealloc, handle_alloc_error, realloc};
+use core::{
+    alloc::Layout,
+    cmp::Ordering,
+    fmt,
+    hash::{Hash, Hasher},
+    iter,
+    mem::{size_of, ManuallyDrop},
+    ops::Deref,
+    ptr::NonNull,
+    slice::from_raw_parts,
+};
+use either::Either;
+
+/// Constructs a `ColList` like so `col_list![0, 2]`.
+///
+/// A head element is required.
+/// Mostly provided for testing.
+#[macro_export]
+macro_rules! col_list {
+    ($head:expr $(, $elem:expr)* $(,)?) => {{
+        let mut list = $crate::ColList::new($head.into());
+        $(list.push($elem.into());)*
+        list
+    }};
+}
+
+/// An error signalling that a `ColList` was empty.
+#[derive(Debug)]
+pub struct EmptyColListError;
+
+/// A builder for a [`ColList`] making sure a non-empty one is built.
+pub struct ColListBuilder {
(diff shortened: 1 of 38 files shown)


crates/primitives/src/col_list.rs
```diff
@@ -0,0 +1,554 @@
...
+        let mut value = self.undo_mark();
+        iter::from_fn(move || {
+            if value == 0 {
+                // No set bits; quit!
+                None
+            } else {
+                // Count trailing zeros and then zero out the first set bit.
+                // For e.g., `0b11001`, this would yield `[0, 3, 4]` as expected.
+                let id = ColId(value.trailing_zeros());
+                value &= value.wrapping_sub(1);
```
kazimuth, line 306: Huh, I don't remember this trick.

It works because any nonzero bitstring ends with `'1' + n'0'`, and subtracting 1 takes this to `'0' + n'1'`.

`('1' + n'0') & ('0' + n'1') = ('1' & '0') + (n'0' & n'1') = (n+1)'0'`. 👍

kazimuth (approved):
Fun bitbashing, LGTM.

Any particular reason you didn't use the `thin_vec` crate instead of `ColListVec`? I don't mind having `ColListVec`, we may find other uses for it eventually.

Unsafe audit:
- [x] `alloc::{alloc, dealloc, realloc}`
- [x] `core::ptr::NonNull::{set_capacity, set_len}`
- [x] `(* _)::{deref, add}`

crates/primitives/src/col_list.rs
```diff
@@ -0,0 +1,554 @@
...
+        }
+    }
+
+    /// Push a [`ColId`] to the list.
+    pub fn push(&mut self, col: ColId) {
+        self.list.push(col);
+    }
+
+    /// Build the [`ColList`] or error if it would have been empty.
+    pub fn build(self) -> Result<ColList, EmptyColListError> {
```
kazimuth, line 63: Oh, and this is why a separate `ColListBuilder` is needed, forgot to comment on this. 👍

PR #731, opened 2024-01-18
Allows to convert into/from micros_since_epoch:u64 for TimeStamp

mamcx:
# Description of Changes

To make it possible to integrate `spacetimedb::Timestamp` with `chrono`, I add a way to  convert into/from micros_since_epoch:u64 

1

### DIFF
--- a/crates/bindings/src/timestamp.rs
+++ b/crates/bindings/src/timestamp.rs
@@ -69,6 +69,16 @@ impl Timestamp {
         let micros_since_epoch = self.micros_since_epoch.checked_sub(micros)?;
         Some(Self { micros_since_epoch })
     }
+
+    /// Converts the timestamp into the number of microseconds since the UNIX epoch.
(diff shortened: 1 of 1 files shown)


joshua-spacetime (approved):
Any reason not to prefer `From<Timestamp> for u64`?

mamcx:
> Any reason not to prefer `From<Timestamp> for u64`?

Mainly for the semantics (so is clear wich unit is used here)

gefjon:
> Any reason not to prefer `From<Timestamp> for u64`?

Agreed w/ Mario; `Timestamp::from(n)` could do any number of different things, e.g. treat as seconds, millis, micros or nanos, and doesn't necessarily refer to an offset from the UNIX epoch (e.g. Rust's `std::time::Instant` is not), but `Timestamp::from_micros_since_epoch` says exactly what it does.

PR #734, opened 2024-01-19
core: Ignore `indexed`-only constraints when updating

kim:
As explained in the commentary.

This will allow updating a module which introduces an index via `#[spacetimedb(index, ..)]`, but it will _not_ create that index.

### DIFF
--- a/crates/core/src/db/update.rs
+++ b/crates/core/src/db/update.rs
@@ -1,9 +1,8 @@
 use core::fmt;
 use std::borrow::Cow;
-use std::collections::{BTreeMap, BTreeSet, HashMap};
+use std::collections::{BTreeMap, HashMap};
 
 use anyhow::Context;
-use spacetimedb_primitives::Constraints;
 
 use crate::database_logger::SystemLogger;
 use crate::error::DBError;
@@ -11,7 +10,7 @@ use crate::execution_context::ExecutionContext;
 
 use super::datastore::locking_tx_datastore::MutTxId;
 use super::relational_db::RelationalDB;
-use spacetimedb_sats::db::def::{ConstraintDef, TableDef, TableSchema};
+use spacetimedb_sats::db::def::{TableDef, TableSchema};
 use spacetimedb_sats::hash::Hash;
 
 #[derive(thiserror::Error, Debug)]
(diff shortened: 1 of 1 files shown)


crates/core/src/db/update.rs
```diff
@@ -167,11 +170,30 @@ pub fn schema_updates(
...
-/// compared sans any unset attributes.
+///   - The number of `constraints` obtained from the module will always be the
+///     same as the number of columns, emitting [`Constraints::unset()`] if no
+///     specific constraint is defined.
+///
+///   - Indexes defined using `#[spacetimedb(index, ..)]` will not have a
+///     corresponding [`Constraints::indexed()`] constraint (it will be unset).
+///     The schema obtained from the database **will** have one such constraint,
+///     however.
+///
```
mamcx, line 183: This is because the conversion into `TableSchema` fills that info. 

`TableDef` is meant to be a partially defined schema that is resolved/qualified when converted to `TableSchema`. See `TableSchema::from_def`

mamcx: Yes, is better.

cloutiertyler: The problem with this is that what the module produces is not a `TableSchema`. The module only specifies a `TableDef`, so by some mechanism we need to compare that to the existing schema (or at least the old def) to find out what changed. You'd have to create a fake `TableSchema` for the new module to compare against.

cloutiertyler: The real problem is that we're not being very careful about what is actual database *schema* (the thing which requires data migration if it changes) and our various attributes (autoinc) or optimizations (indexes).

Indexes are not really part of the schema of a table in the precise meaning of the term schema, neither are attributes.

cloutiertyler (approved):
This looks good to me, but please read the comments I posted inline.

We will have to have a proper proposal on how we should handle this.

crates/core/src/db/update.rs
```diff
@@ -322,22 +275,155 @@ mod tests {
...
+                table_id: TableId(42),
+                col_pos: ColId(0),
+                col_name: "name".into(),
+                col_type: AlgebraicType::String,
+            }],
+            vec![IndexSchema {
+                index_id: IndexId(68),
+                table_id: TableId(42),
+                index_type: IndexType::BTree,
+                index_name: "bobson_dugnutt".into(),
```
cloutiertyler, line 332: lol

cloutiertyler (approved):
Well it certainly does simply things and I appreciate the tests.

PR #735, opened 2024-01-19
Fix inconsistent auth/identity creation

coolreader18:
# Description of Changes

Now, in most cases, it'll create an identity if the user doesn't pass one.

## Testing
```bash
spacetime init my_new_project --lang rust
cd my_new_project
spacetime publish my_module

# Make sure we can call a reducer with no identity
curl -X POST http://localhost:3000/database/my_module/say_hello
```


bfops (changes requested):
I think this mostly makes sense to me! I left some questions and confusions.

Also I have no idea if there's any automated test coverage we could/should be updating or anything :shrug: 

(I also left some notes for other reviewers based on my read-through; I'm not looking for a response from you on those! :slightly_smiling_face:)

bfops:
Summary of changes, I think:
- Restructuring `SpacetimeCreds` datatype
- Factoring out helper functions, especially in `SpacetimeCreds` + `SpacetimeAuthHeader`
- Add `SpacetimeAuthRequired`
- Plumb a `ctx` parameter through our routes
- Add `auth_middleware` to create identities "on the fly" (and return them if appropriate)
- Wrap a bunch of routes in `auth_middleware`

crates/standalone/src/routes/mod.rs
```diff
@@ -7,14 +7,17 @@ use spacetimedb_client_api::{
...
     let router = axum::Router::new()
-        .nest("/database", database::control_routes().merge(database::worker_routes()))
-        .nest("/identity", identity::router())
-        .nest("/energy", energy::router())
+        .nest(
+            "/database",
+            database::control_routes(ctx.clone()).merge(database::worker_routes(ctx.clone())),
+        )
+        .nest("/identity", identity::router(ctx.clone()))
+        .nest("/energy", energy::router(ctx.clone()))
```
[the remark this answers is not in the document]
coolreader18: Limitations of axum, sorta - the middleware needs access to ctx, but axum doesn't let middleware access the state from the router, so you have to double up on passing the state.

crates/client-api/src/auth.rs
```diff
@@ -260,3 +243,15 @@ impl headers::Header for SpacetimeExecutionDurationMicros {
...
+pub async fn auth_middleware<S: ControlStateDelegate + NodeDelegate>(
+    State(worker_ctx): State<S>,
+    auth: SpacetimeAuthHeader,
+    mut req: Request,
+    next: Next,
+) -> axum::response::Result<impl IntoResponse> {
+    let auth = auth.get_or_create(&worker_ctx).await?;
+    req.extensions_mut().insert(auth.clone());
+    let resp = next.run(req).await;
+    Ok((auth.into_headers(), resp))
```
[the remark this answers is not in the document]
coolreader18: Nope, they're equivalent.

crates/client-api/src/routes/identity.rs
```diff
@@ -206,39 +201,39 @@ pub struct ValidateTokenParams {
...
 pub async fn get_public_key<S: NodeDelegate>(State(ctx): State<S>) -> axum::response::Result<impl IntoResponse> {
-    let res = Response::builder()
-        .header("Content-Type", "application/pem-certificate-chain")
-        .body(())
-        .map_err(log_and_500)?;
-    Ok((res, ctx.public_key_bytes().to_owned()))
+    Ok((
+        [(CONTENT_TYPE, "application/pem-certificate-chain")],
+        ctx.public_key_bytes().to_owned(),
+    ))
```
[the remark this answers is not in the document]
coolreader18: Same behavior, yeah, just more idiomatic.

crates/client-api/src/auth.rs
```diff
@@ -140,43 +138,16 @@ impl IntoResponse for AuthorizationRejection {
...
         }
     }
 }
 
-#[derive(Debug)]
-enum AuthorizationRejectionReason {
-    Jwt(JwtErrorKind),
-    Header(TypedHeaderRejection),
-    MalformedTokenQueryString,
-    CantDecodeAuthorizationToken,
```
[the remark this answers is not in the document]
coolreader18: That wasn't an error that could ever actually occur; we basically encoded the token to base64 and then decoded it from base64 and if the decoding failed we returned that error. It could've just been an unwrap() instead of a variant.

crates/client-api/src/auth.rs
```diff
@@ -37,95 +39,91 @@ impl authorization::Credentials for SpacetimeCreds {
...
-                TypedHeaderRejectionReason::Missing => Ok(Self { auth: None }),
-                _ => Err(AuthorizationRejection {
-                    reason: AuthorizationRejectionReason::Header(e),
-                }),
-            },
-        }
+        let Some(creds) = SpacetimeCreds::from_request_parts(parts)? else {
+            return Ok(Self { auth: None });
+        };
+        let claims = creds.decode_token(state.public_key())?;
```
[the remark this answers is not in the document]
coolreader18: Yeah, the error handling before was unnecessarily complicated, this is equivalent.

crates/client-api/src/auth.rs
```diff
@@ -37,95 +39,91 @@ impl authorization::Credentials for SpacetimeCreds {
...
+    fn from_request_parts(parts: &request::Parts) -> Result<Option<Self>, headers::Error> {
+        let res = match parts.headers.typed_try_get::<headers::Authorization<Self>>() {
+            Ok(Some(headers::Authorization(creds))) => return Ok(Some(creds)),
+            Ok(None) => Ok(()),
+            Err(e) => Err(e),
+        };
+        if let Ok(Query(creds)) = Query::<Self>::try_from_uri(&parts.uri) {
+            return Ok(Some(creds));
+        }
+        res.map(|()| None)
```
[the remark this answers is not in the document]
coolreader18: Yeah, I guess, but then it would be a None that would never be Some. it doesn't much matter I suppose, it could be `Ok(None)`, the logic is just kinda convoluted and I was trying to preserve it while making it idiomatic.

bfops (approved):
Okay, I think this looks good to me - thank you for explaining everything!

bfops:
Note: `SpacetimeDBPrivate` will need an update! I think at least the `cloud-next` crate seems to call into some of these.

crates/client-api/src/auth.rs
```diff
@@ -260,3 +243,15 @@ impl headers::Header for SpacetimeExecutionDurationMicros {
         values.extend([(self.0.as_micros() as u64).into()])
     }
 }
+
+pub async fn auth_middleware<S: ControlStateDelegate + NodeDelegate>(
```
bfops, line 247: what do you think about renaming this to something more descriptive? e.g. `anonymous_auth_middleware`?

crates/client-api/src/auth.rs
```diff
@@ -1,11 +1,10 @@
-use std::fmt::Write;
 use std::time::Duration;
 
-use axum::extract::Query;
+use axum::extract::{Query, Request, State};
```
bfops, line 3: on re-reading, I have a question - could most of the refactoring changes in this file be made in a separate PR? Aside from adding `auth_middleware`, it seems like these changes are backwards-compatible, largely independent, and I think they're maybe easier to review if they're separate from the auth functionality changes.

crates/standalone/src/routes/mod.rs
```diff
@@ -7,14 +7,17 @@ use spacetimedb_client_api::{
 };
 
 #[allow(clippy::let_and_return)]
-pub fn router<S>() -> axum::Router<S>
+pub fn router<S>(ctx: S) -> axum::Router<()>
```
bfops, line 10: I think this (and maybe the new parameters to other similar functions) cause breakage in `SpacetimeDBPrivate`

bfops (changes requested):
I did another review pass, and had a few more comments

crates/client-api/src/auth.rs
coolreader18: Not fully opposed, but I think this is descriptive enough as is - it's a feature of our auth system that you don't actually *need* to be auth'd, that we'll generate an identity for you. So this does do standard authentication, not just anonymous auth.

coolreader18: The only places we don't generate it for you is where it doesn't make any sense to - but there's no reason to disallow it, just that it doesn't make much sense.

crates/client-api/src/auth.rs
coolreader18: I could, if you feel strongly about it, though one could consider them all part and parcel of making the auth code generally more straightforward/readable.

coolreader18: (It might be difficult to split them at this point, but yeah if you feel strongly I could try)

crates/client-api/src/auth.rs
```diff
@@ -37,95 +39,91 @@ impl authorization::Credentials for SpacetimeCreds {
...
+/// this to be present.
+#[derive(Clone)]
 pub struct SpacetimeAuth {
     pub creds: SpacetimeCreds,
     pub identity: Identity,
 }
 
-pub struct SpacetimeAuthHeader {
-    pub auth: Option<SpacetimeAuth>,
+impl SpacetimeAuth {
```
bfops, line 85: this block was just moved from below, unchanged, so it's next to the `struct` definition.

bfops:
Our reviewers are a bit bottlenecked, and I don't think we should necessarily trust me as solo-reviewer yet :sweat_smile: - I moved just the `auth.rs` changes into [their own refactor PR](https://github.com/clockworklabs/SpacetimeDB/pull/871), on the hope that might make it easier to get review on everything. What do you think?

crates/client-api/src/routes/energy.rs
```diff
@@ -130,13 +124,17 @@ pub async fn set_energy_balance<S: ControlStateDelegate>(
...
 {
     use axum::routing::{get, post, put};
+    let auth_middleware = axum::middleware::from_fn_with_state(ctx, auth_middleware::<S>);
     axum::Router::new()
         .route("/:identity", get(get_energy_balance::<S>))
-        .route("/:identity", post(set_energy_balance::<S>))
-        .route("/:identity", put(add_energy::<S>))
+        .route(
+            "/:identity",
+            post(set_energy_balance::<S>).route_layer(auth_middleware.clone()),
```
bfops, line 137: wait, doesn't `auth_middleware` create an identity if it's not explicitly provided? Didn't we add `SpacetimeAuthRequired` above? Aren't these.. trying to do opposite things? Am I misunderstanding?

crates/client-api/src/routes/database.rs
```diff
@@ -409,27 +403,17 @@ pub struct LogsQuery {
     follow: bool,
 }
 
-fn auth_or_unauth(auth: SpacetimeAuthHeader) -> axum::response::Result<SpacetimeAuth> {
-    auth.get()
-        .ok_or((StatusCode::UNAUTHORIZED, "Invalid credentials").into())
-}
```
bfops, line 415: I can't tell if I'm reading this correctly. Did this use to effectively require a pre-existing auth? If so it feels like the usages below should switch to `Required`?

crates/client-api/src/routes/identity.rs
```diff
@@ -111,10 +115,9 @@ pub async fn set_email<S: ControlStateWriteAccess>(
     State(ctx): State<S>,
     Path(SetEmailParams { identity }): Path<SetEmailParams>,
     Query(SetEmailQueryParams { email }): Query<SetEmailQueryParams>,
-    auth: SpacetimeAuthHeader,
+    Extension(auth): Extension<SpacetimeAuth>,
 ) -> axum::response::Result<impl IntoResponse> {
     let identity = identity.into();
-    let auth = auth.get().ok_or(StatusCode::BAD_REQUEST)?;
```
bfops, line 117: it feels like this one should still be `SpacetimeAuthRequired`?

crates/client-api/src/routes/identity.rs
```diff
@@ -199,39 +197,39 @@ pub struct ValidateTokenParams {
 
 pub async fn validate_token(
     Path(ValidateTokenParams { identity }): Path<ValidateTokenParams>,
-    auth: SpacetimeAuthHeader,
+    SpacetimeAuthRequired(auth): SpacetimeAuthRequired,
```
bfops, line 200: nit: this one ends up doing a `auth.identity != identity` check anyway, so would anonymous auth be fine like the others?

crates/core/src/auth/identity.rs
```diff
@@ -27,21 +29,17 @@ pub fn encode_token(private_key: &EncodingKey, identity: Identity) -> Result<Str
...
+    expiry: Option<Duration>,
 ) -> Result<String, JwtError> {
     let header = Header::new(jsonwebtoken::Algorithm::ES256);
 
-    let expiry = expiry.map(|seconds| {
-        let mut timer = SystemTime::now();
-        timer += Duration::from_secs(seconds);
-        // SAFETY: duration_since will panic if an argument is later than the time
-        // used for the duration calculation. In case of UNIX_EPOCH it can't be the case
-        timer.duration_since(UNIX_EPOCH).unwrap().as_secs()
```
bfops, line 39: why did this get to be removed?

coolreader18: Tidying up auth stuff, I guess. serde_with can do this conversion for us.

crates/client-api/src/routes/database.rs
```diff
@@ -148,8 +146,6 @@ pub async fn call<S: ControlStateDelegate + NodeDelegate>(
             let (status, body) = reducer_outcome_response(&identity, &reducer, result.outcome);
             Ok((
                 status,
-                TypedHeader(SpacetimeIdentity(caller_identity)),
-                TypedHeader(SpacetimeIdentityToken(caller_identity_token)),
```
bfops, line 152: to double-check: are these always added by a different layer of the stack right now? even for things that don't use the `auth_middleware`?

crates/core/src/auth/identity.rs
bfops: Ah I see the new `serde_with` line. Seems reasonable.

crates/client-api/src/routes/database.rs
```diff
@@ -409,27 +403,17 @@ pub struct LogsQuery {
...
 where
     S: ControlStateDelegate + NodeDelegate,
 {
     // You should not be able to read the logs from a database that you do not own
     // so, unless you are the owner, this will fail.
-    // TODO: This returns `UNAUTHORIZED` on failure,
-    //       while everywhere else we return `BAD_REQUEST`.
-    //       Is this special in some way? Should this change?
-    //       Should all the others change?
-    let auth = auth_or_unauth(auth)?;
```
bfops, line 432: (I think this used to fail if auth wasn't present, but it's fine because we check for a specific identity below)

crates/client-api/src/routes/database.rs
```diff
@@ -907,10 +886,8 @@ pub struct SetNameQueryParams {
 pub async fn set_name<S: ControlStateDelegate>(
     State(ctx): State<S>,
     Query(SetNameQueryParams { domain, address }): Query<SetNameQueryParams>,
-    auth: SpacetimeAuthHeader,
+    Extension(auth): Extension<SpacetimeAuth>,
 ) -> axum::response::Result<impl IntoResponse> {
-    let auth = auth_or_unauth(auth)?;
```
bfops, line 912: (review note: I think this used to fail if auth wasn't present, but it's fine because we check for a specific identity below)

crates/client-api/src/routes/database.rs
```diff
@@ -885,10 +866,8 @@ pub struct DeleteDatabaseParams {
 pub async fn delete_database<S: ControlStateDelegate>(
     State(ctx): State<S>,
     Path(DeleteDatabaseParams { address }): Path<DeleteDatabaseParams>,
-    auth: SpacetimeAuthHeader,
+    Extension(auth): Extension<SpacetimeAuth>,
 ) -> axum::response::Result<impl IntoResponse> {
-    let auth = auth_or_unauth(auth)?;
```
bfops, line 890: (review note: I think this used to fail if auth wasn't present, but it's fine because `delete_database` checks the identity)

crates/client-api/src/routes/database.rs
```diff
@@ -798,7 +780,6 @@ pub async fn publish<S: NodeDelegate + ControlStateDelegate>(
 
     // You should not be able to publish to a database that you do not own
     // so, unless you are the owner, this will fail.
-    let auth = auth_or_unauth(auth)?;
```
bfops, line 801: (review note: `create_dns_record` and/or `delete_database` will fail if you're not the right identity anyway)

crates/client-api/src/routes/database.rs
```diff
@@ -523,15 +507,14 @@ pub async fn sql<S>(
...
-    auth: SpacetimeAuthHeader,
+    Extension(auth): Extension<SpacetimeAuth>,
     body: String,
 ) -> axum::response::Result<impl IntoResponse>
 where
     S: NodeDelegate + ControlStateDelegate,
 {
     // Anyone is authorized to execute SQL queries. The SQL engine will determine
     // which queries this identity is allowed to execute against the database.
-    let auth = auth.get().ok_or((StatusCode::UNAUTHORIZED, "Invalid credentials."))?;
```
bfops, line 534: (review note: I believe this used to fail if there was no auth, but it's fine because I think `AuthCtx` does identity checking?)

crates/client-api/src/routes/database.rs
coolreader18: No, only for `anon_auth_middleware` routes. However, the only routes that added the headers before now have `anon_auth_middleware`, so it's fine.

crates/client-api/src/routes/database.rs
coolreader18: `AuthCtx` doesn't do identity checking, but it actually just didn't make sense to check this before, given the comment right above it. We don't need to require credentials for this.

crates/client-api/src/routes/identity.rs
coolreader18: I think given we're validating the token - it makes sense to require a token to validate against. If you don't pass a token to validate_token it was probably a mistake.

crates/client-api/src/routes/database.rs
cloutiertyler: We do check. What's happening here is that instead of returning an error if you are unauthenticated (not unauthorized), we will create an anon authenticated user for you (which will presumably not be authorized).

cloutiertyler (approved):
On a high level this looks good to me. Your review comments are very helpful @bfops

bfops (approved):
PR is already approved, but adding my own approval notes - I will vouch for the fact that all the individual route changes (e.g. which ones are now `SpacetimeAuthRequired` etc.) are faithful translations of the previous logic.

I left review notes where the logic changed a bit, but is effectively the same.

I did not have the capacity to verify the small refactorings, because I'm unfamiliar with the underlying libs.

PR #737, opened 2024-01-20
Fix for reducer queue metric

jdetter:
# Description of Changes

- Fixes an issue where the reducer queue metric wouldn't get decremented, which would leave a baseline count which would never reach 0 even if the queue was empty.

corresponding GitHub label.

Not ABI/API breaking

1

### DIFF
--- a/crates/core/src/util/lending_pool.rs
+++ b/crates/core/src/util/lending_pool.rs
@@ -80,9 +80,10 @@ impl<T> LendingPool<T> {
         queue_len_max.set(max_queue_len);
 
         async move {
-            let permit = acq.await.map_err(|_| PoolClosed)?;
+            let permit_result = acq.await.map_err(|_| PoolClosed);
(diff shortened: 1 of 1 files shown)


crates/core/src/util/lending_pool.rs
```diff
@@ -80,9 +80,10 @@ impl<T> LendingPool<T> {
         queue_len_max.set(max_queue_len);
 
         async move {
-            let permit = acq.await.map_err(|_| PoolClosed)?;
+            let permit_result = acq.await.map_err(|_| PoolClosed);
             queue_len.dec();
```
joshua-spacetime, line 84: I might wrap this in a scope guard to future proof. At the very least I would add a comment about decrementing in the case of a semaphore closure.

joshua-spacetime (approved):
Thanks for the quick fix!

PR #738, opened 2024-01-20
Use `ColList` more and improve it

Centril:
# Description of Changes

Improve `ColList`'s interface and use the type a bit more.

None

1

### DIFF
--- a/crates/primitives/src/col_list.rs
+++ b/crates/primitives/src/col_list.rs
@@ -49,9 +49,14 @@ impl Default for ColListBuilder {
 impl ColListBuilder {
     /// Returns an empty builder.
     pub fn new() -> Self {
-        Self {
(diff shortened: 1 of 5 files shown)


crates/primitives/src/col_list.rs
```diff
@@ -319,6 +349,17 @@ impl ColListInline {
     fn undo_mark(&self) -> u64 {
         self.0 >> 1
     }
+
+    /// Returns an equivalent list in heap form instead of inline, and adds `col` to it.
+    /// The capacity of the vec will be `2 * (self.len() + 1)`
```
kazimuth, line 354: We could choose to round this to a power of two, but this also seems like a reasonable choice.

kazimuth (approved):
Straightforward enough, LGTM.

PR #741, opened 2024-01-22
made subscription broadcast multithreaded

Shubham8287:
# Description of Changes

Made `broadcast_commit_event` multithreaded to leverage read locks.

### DIFF
--- a/crates/core/src/subscription/module_subscription_actor.rs
+++ b/crates/core/src/subscription/module_subscription_actor.rs
@@ -19,9 +19,10 @@ use crate::{
     host::module_host::{EventStatus, ModuleEvent},
 };
 use futures::{stream::FuturesUnordered, FutureExt, StreamExt};
+use parking_lot::RwLock;
 use spacetimedb_lib::identity::AuthCtx;
 use spacetimedb_lib::Identity;
-use tokio::sync::mpsc;
+use tokio::{sync::mpsc, task::JoinHandle};
 
 /// All of these commands (and likely any future ones) should all be in the same enum with the
 /// same queue so that e.g. db updates from commit events and db updates from subscription
@@ -98,9 +99,11 @@ impl ModuleSubscriptionManager {
     }
 }
 
+type SubscriptionRw = Arc<RwLock<Subscription>>;
+
 struct ModuleSubscriptionActor {
     relational_db: Arc<RelationalDB>,
-    subscriptions: Vec<Subscription>,
+    subscriptions: Vec<SubscriptionRw>,
     owner_identity: Identity,
 }
 
@@ -130,27 +133,28 @@ impl ModuleSubscriptionActor {
     ) -> Result<(), DBError> {
         self.remove_subscriber(sender.id);
         let auth = AuthCtx::new(self.owner_identity, sender.id.identity);
-
         let mut queries = QuerySet::new();
         for sql in subscription.query_strings {
             let qset = compile_read_only_query(&self.relational_db, tx, &auth, &sql)?;
             queries.extend(qset);
         }
 
-        let sub = match self.subscriptions.iter_mut().find(|s| s.queries == queries) {
+        let sub = match self.subscriptions.iter_mut().find(|s| s.read().queries == queries) {
             Some(sub) => {
-                sub.add_subscriber(sender);
+                sub.write().add_subscriber(sender);
                 sub
             }
             None => {
-                self.subscriptions.push(Subscription::new(queries, sender));
+                self.subscriptions
+                    .push(Arc::new(Subscription::new(queries, sender).into()));
                 self.subscriptions.last_mut().unwrap()
             }
         };
 
-        let database_update = sub.queries.eval(&self.relational_db, tx, auth)?;
+        let subscription = sub.read_arc();
+        let database_update = subscription.queries.eval(&self.relational_db, tx, auth)?;
 
-        let sender = sub.subscribers().last().unwrap();
+        let sender = subscription.subscribers().last().unwrap();
 
         // NOTE: It is important to send the state in this thread because if you spawn a new
         // thread it's possible for messages to get sent to the client out of order. If you do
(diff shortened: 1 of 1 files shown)


crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -179,51 +183,52 @@ impl ModuleSubscriptionActor {
...
+                    }
+                    Err(e) => log::error!("error occured in broadcasting commit event {:?}", e),
+                }
             }
-        }
-
-        futures.collect::<()>().await;
+            let ctx = ExecutionContext::incremental_update(relational_db.address());
+            relational_db.release_tx(&ctx, tx);
+            futures.collect::<()>().await;
```
coolreader18, line 229: The order of the futures.collect() and the incremental_update call has been switched now.

crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -130,21 +134,21 @@ impl ModuleSubscriptionActor {
     ) -> Result<(), DBError> {
         self.remove_subscriber(sender.id);
         let auth = AuthCtx::new(self.owner_identity, sender.id.identity);
-
+        let mut subscriptions = self.subscriptions.write_arc();
```
coolreader18, line 137: is the write_arc() necessary here over just write()? and in general I'd say probably drop the lock as soon as possible barring interactions with external things that need to logically happen under the mutex. So, maybe before the send_message() call at the end? though maybe that was happening under the lock before anyway.

Shubham8287: Good point. Also, There was no lock here before.

coolreader18:
I think right now, I mostly don't see the point of this change? Is there a plan to start calling these methods from other threads in the future? Because with the current changeset, I don't see any real operational difference. Is the goal to run multiple `ModuleSubscriptionActor` commands in parallel? I'm not sure if this is a good idea as implemented here, since theoretically commands could run out-of-order from the order they were sent, which is bad. Oh also - I think I just realized, the deadlock we're getting is possibly write-starving; the read() locks collectively holding on to the rwlock constantly with the write() waiter stuck forever.

joshua-spacetime (commented):
I just want to verify that this change doesn't affect consistency at all.

If we have two transactions `A` and `B`, and `A` runs(commits) before `B`, then we need to ensure that `broadcast_commit_event` is called on `A` before it is called on `B`.

I believe this will still be the case because inserting a `ModuleEvent` into the `ModuleSubscriptionManager`'s channel is still synchronized [here](https://github.com/clockworklabs/SpacetimeDB/blob/602f1bac26b6c303c2edf2d2d3bbae8ccd6264ec/crates/core/src/host/wasm_common/module_host_actor.rs#L626).

Is this correct @Shubham8287?

crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -179,51 +183,52 @@ impl ModuleSubscriptionActor {
...
+        subscriptions.retain_mut(|sub| {
             sub.remove_subscriber(client_id);
             !sub.subscribers().is_empty()
         })
     }
 
-    async fn _broadcast_commit_event(&mut self, mut event: ModuleEvent, tx: &mut Tx) -> Result<(), DBError> {
-        let futures = FuturesUnordered::new();
+    async fn broadcast_commit_event(&mut self, mut event: ModuleEvent) -> Result<(), DBError> {
```
joshua-spacetime, line 193: This function really shouldn't be async. The reason being is that subscription evaluations need to have a serial equivalent ordering to the original transactions.

If txn `A` inserts a row, and then txn `B` deletes that same row, it would be incorrect to evaluate the effects of `B` before the effects of `A`.

Note this only applies to incremental evaluation `eval_incr`. It does not apply to the initial subscribe call `eval` because of the absence of deletes.

joshua-spacetime (changes requested):
Requesting changes because I think we can only multithread the initial subscription calls, not the incremental evaluations unfortunately. At least not under the current concurrency model.

Shubham8287:
> I don't see any real operational difference. Is the goal to run multiple ModuleSubscriptionActor commands in parallel? begin_tx() in broadcast_commit_event takes a lock on the datastore anyway, so it would still run in serial.

`begin_tx()` holds only read lock, so we can have parallel `broadcast_commit_event` executions.

> the deadlock we're getting is possibly write-starving; the read() locks collectively holding on to the rwlock constantly, with the write() waiter stuck forever

this could be true.

crates/core/src/subscription/module_subscription_actor.rs
Shubham8287: I got you, This would require some more work I guess. Thanks for pointing.

joshua-spacetime: @Shubham8287 just reiterating the solution that we discussed earlier:

If each reducer holds an exclusive lock on the `ModuleSubscriptionManager`, we can evaluate all of the subscription queries for **single** transaction concurrently, and at the same time preserve the ordering across transactions.

That is, reducers are still run sequentially and do not complete until all of the required subscription queries have been evaluated. However within said reducer, queries can be run in parallel.

Shubham8287: Thanks for summarizing.
I have done changed the we intented to achive parrallelism, so that now instead of trying to run `broadcast_commit_event` parallelly. The method will spawn threads to evaluate each Subsription `QuerySet`s
and keeping subscription actor itself single threaded.

crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -98,9 +99,11 @@ impl ModuleSubscriptionManager {
     }
 }
 
+type SubscriptionRw = Arc<RwLock<Subscription>>;
```
joshua-spacetime, line 102: I actually don't think a lock is necessary anymore, the way you have things written.

It's technically not necessary for incremental evaluation, because `broadcast_event_blocking` will block each reducer until `broadcast_commit_event` completes.

It's also technically not necessary for the initial subscription, because `add_subscriber` acquires an exclusive lock on the datastore.

Of course if any of those assumptions become invalid, then we would need it. However, for now I would remove it and just make sure to comment things sufficiently.

Shubham8287: primary motivation here is not to prevent race condition but to have a `Sync` so that I can share the reference of `Subscription` between threads.
I am new to rust concurrency, would love to know better ways to do it.

jdetter (approved):
I've tested this thoroughly using the bots and I'm not seeing the deadlock that we had before. The performance here is *really* good so let's try to get this in asap 👍

crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -180,23 +184,31 @@ impl ModuleSubscriptionActor {
...
+            !subscription.subscribers().is_empty()
         })
     }
 
-    async fn _broadcast_commit_event(&mut self, mut event: ModuleEvent, tx: &mut Tx) -> Result<(), DBError> {
-        let futures = FuturesUnordered::new();
-        let auth = AuthCtx::new(self.owner_identity, event.caller_identity);
-
-        for subscription in &mut self.subscriptions {
+    async fn broadcast_commit_event(&mut self, event: ModuleEvent) -> Result<(), DBError> {
```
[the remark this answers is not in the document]
joshua-spacetime: Resolving this comment since what I said is wrong. It is indeed run in an async manner, which means we already have a consistency issue, but we can solve that in a separate patch.

joshua-spacetime (approved):

joshua-spacetime:
Feel free to merge @Shubham8287. The consistency issue can be handled separately #758.

PR #742, opened 2024-01-22
Unify benchmark runs (to some extent)

kazimuth:
# Description of Changes

Run callgrind and regular benchmark suites together.

You'll still get separate comments, that way if one of them fails you still get the other one.

0


gefjon (approved):
I don't understand Github Actions enough to meaningfully review this, but I sure would like it to be merged.

kazimuth:
callgrind please
benchmarks please

PR #744, opened 2024-01-23
Cleanup SpacetimeDB name usages

jdetter:
# Description of Changes

- Fixed any instances where SpacetimeDB is not spelled correctly

corresponding GitHub label.

- This may have some small API breakages in tests and things, and we changed the name of a file on disk although I'm not sure that actually impacts the API at all. This *probably* breaks the API somehow but not in a meaningful way.

0


mamcx (approved):
LGTM. I check the `syntax coloring` running the `repl` and looks fine.

PR #745, opened 2024-01-23
Fix row count for the 'sql' command

mamcx:
# Description of Changes

The `tabled` library add the header as part of the number of rows...

corresponding GitHub label.

1

### DIFF
--- a/crates/cli/src/subcommands/sql.rs
+++ b/crates/cli/src/subcommands/sql.rs
@@ -112,7 +112,8 @@ pub(crate) async fn run_sql(builder: RequestBuilder, sql: &str, with_stats: bool
         .map(|stmt_result| {
             let mut table = stmt_result_to_table(stmt_result)?;
             if with_stats {
-                let row_count = print_row_count(table.count_rows());
+                // The `tabled::count_rows` add the header as a row, so subtract it.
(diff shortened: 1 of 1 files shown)


crates/cli/src/subcommands/sql.rs
```diff
@@ -112,7 +112,8 @@ pub(crate) async fn run_sql(builder: RequestBuilder, sql: &str, with_stats: bool
         .map(|stmt_result| {
             let mut table = stmt_result_to_table(stmt_result)?;
             if with_stats {
-                let row_count = print_row_count(table.count_rows());
+                // The `tabled::count_rows` add the header as a row, so subtract it.
+                let row_count = print_row_count(table.count_rows().wrapping_sub(1));
```
joshua-spacetime, line 116: Are you able to update count_rows instead?

joshua-spacetime (approved):

crates/cli/src/subcommands/sql.rs
mamcx: No, that is from a external crate `tabled`

jdetter: I agree that this is super jank but if this is how the external crate determines this value then I think this is the best we can do 🤷‍♂️

jdetter (approved):
Looks fine 👍

PR #746, opened 2024-01-23
Static feature flag for generating perfmaps

gefjon:
# Description of Changes

Add a feature/config flag, `perfmap`, to `core` and its dependents, which passes `ProfilingStrategy::PerfMap` to our `Engine`.

As noted, it would be better to configure this dynamically via a command-line flag to `spacetime start`, but for now, doing `cargo build -p spacetimedb-standalone --release --features perfmap` is sufficient.

See the internal docs and https://perf.wiki.kernel.org/index.php/Tutorial and https://docs.wasmtime.dev/examples-profiling-perf.html for a description.

Note that, now that we've split our binaries, it's a little inconvenient to use this. Well, it was always a little inconvenient, and now it's somewhat more inconvenient. You have to build the `spacetimedb-standalone` binary with the feature enabled, then manually install that binary into wherever your CLI looks for it.

It's not.

2, in the sense that this is an obviously imperfect design, and IDK if we want to merge it or put in the work to do something better. But it's not going to break anything.


jdetter (approved):
Me and Ryan tested this doing the bot tests and it works, thanks Phoebe 👍

joshua-spacetime:
@jdetter Moving this to `DevEx`. Feel free to kick it back if you need anything else.

PR #749, opened 2024-01-24
feat(748): Add compile time feature flag for table metrics

### LINKED ISSUE #748: Add feature flag for metrics collection
While metrics collection is still done synchronously, we should, at the very least, have a feature flag for disabling the metrics that contribute most to latency.

joshua-spacetime:
Closes #748.

This patch adds a temporary feature flag for enabling table metrics.

Table metrics are potentially recorded many times within a transaction. This recording is synchronous and therefore can have a noticable impact on latency.

This new flag will be enabled by default.
With it, table metrics will be recorded; without it, they won't.

Note this flag is temporary.
It will be removed once metrics are recorded async.

# Description of Changes

corresponding GitHub label.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,7 @@ jobs:
           sudo chmod 777 /stdb
 
       - name: Run cargo test
-        run: cargo test --all --features odb_rocksdb,odb_sled
+        run: cargo test --all --features odb_rocksdb,odb_sled,metrics
 
   lints:
     name: Lints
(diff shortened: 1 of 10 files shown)


joshua-spacetime:
@Centril these were the only ones I saw in the profile, apart from query duration metrics. I didn't want to remove all of them, since @jdetter probably still wants at least some observability when perf testing. These ones will remove the added latency to reducer calls.

I can also put the query duration metrics behind the flag, but perhaps @jdetter should weigh in if he thinks they're still important for perf testing.

jdetter:
I'm watching this one closely, if we can cover as many metrics as possible with the temporary `metrics` flag that would be awesome. Let me know when this is ready for testing again 👍

jdetter (approved):
Excellent, this looks great Josh. I'll include this in the benchmark this weekend 👍

PR #750, opened 2024-01-24
SATS: add support needed for BFLATN => BSATN directly

Centril:
Also some associated refactoring

# Description of Changes

Extends `spacetimedb_sats::ser::Serializer` to support:
- serializing from the BSATN encoded in granules and in blobs
- serializing from a string encoded in granules

Adds `spacetimedb_sats::bsatn::to_len` which computes the length of a bsatn encoding without storing the encoding.

None

2, Some unsafe code, and this extends the `Serializer` trait.

### DIFF
--- a/crates/sats/src/satn.rs
+++ b/crates/sats/src/satn.rs
@@ -357,7 +357,7 @@ impl<'a, 'b> ser::Serializer for SatnFormatter<'a, 'b> {
         value.serialize(self)
     }
 
-    unsafe fn serialize_bsatn_in_chunks<'c, I: Iterator<Item = &'c [u8]>>(
+    unsafe fn serialize_bsatn_in_chunks<'c, I: Clone + Iterator<Item = &'c [u8]>>(
         self,
         ty: &crate::AlgebraicType,
         total_bsatn_len: usize,
@@ -376,7 +376,7 @@ impl<'a, 'b> ser::Serializer for SatnFormatter<'a, 'b> {
         value.serialize(self)
     }
 
-    unsafe fn serialize_str_in_chunks<'c, I: Iterator<Item = &'c [u8]>>(
+    unsafe fn serialize_str_in_chunks<'c, I: Clone + Iterator<Item = &'c [u8]>>(
         self,
         total_len: usize,
         string: I,
@@ -591,7 +591,7 @@ impl<'a, 'b> ser::Serializer for PsqlFormatter<'a, 'b> {
         unsafe { self.0.serialize_bsatn(ty, bsatn) }
     }
 
-    unsafe fn serialize_bsatn_in_chunks<'c, I: Iterator<Item = &'c [u8]>>(
+    unsafe fn serialize_bsatn_in_chunks<'c, I: Clone + Iterator<Item = &'c [u8]>>(
         self,
         ty: &crate::AlgebraicType,
         total_bsatn_len: usize,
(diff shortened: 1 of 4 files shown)


crates/sats/src/satn.rs
```diff
@@ -340,6 +343,53 @@ impl<'a, 'b> ser::Serializer for SatnFormatter<'a, 'b> {
         })?;
         write!(self, ")")
     }
+
+    unsafe fn serialize_bsatn(self, ty: &crate::AlgebraicType, bsatn: &[u8]) -> Result<Self::Ok, Self::Error> {
```
gefjon, line 347: Can default implementations of supertrait methods meaningfully have doc comments? If so, I'd like to see a comment on this method (and the other impls here) which describe how they're implemented, and when implementors should overwrite them. If not, I'd put that comment on `trait Satn` instead.

crates/sats/src/ser.rs
```diff
@@ -124,10 +124,82 @@ pub trait Serializer: Sized {
...
+    /// ```
+    ///
+    /// This method is a concession to performance,
+    /// allowing some implementations to write the buffer directly.
+    ///
+    /// # Safety
+    ///
+    /// - `total_bsatn_len == bsatn.map(|c| c.len()).sum() <= isize::MAX`
+    /// - Let `buf` be defined as above, i.e., the bytes of `bsatn` concatenated.
+    ///   Then `AlgebraicValue::decode(ty, &mut bytes).is_ok()`
```
gefjon, line 159:
```suggestion
    /// - Let `buf` be defined as above, i.e., the bytes of `bsatn` concatenated.
    ///   Then `AlgebraicValue::decode(ty, &mut buf).is_ok()`.
    ///   That is, `buf` encodes a valid element of `ty`.
```

crates/sats/src/ser.rs
```diff
@@ -124,10 +124,82 @@ pub trait Serializer: Sized {
...
+    /// This method is a concession to performance,
+    /// allowing some implementations to write the buffer directly.
+    ///
+    /// # Safety
+    ///
+    /// - `total_len == string.map(|c| c.len()).sum() <= isize::MAX`
+    /// - Let `buf` be the bytes of `string` concatenated.
+    ///   Then `core::str::from_utf8(&buf).is_ok()`.
+    ///   That is, `buf` is valid UTF-8.
+    ///   Note however that individual chunks need not be valid UTF-8.
```
gefjon, line 191:
```suggestion
    ///   Note however that individual chunks need not be valid UTF-8,
    ///   as multi-byte characters may be split across chunk boundaries.
```

crates/sats/src/ser.rs
```diff
@@ -124,10 +124,82 @@ pub trait Serializer: Sized {
...
+        self,
+        ty: &AlgebraicType,
+        total_bsatn_len: usize,
+        bsatn: I,
+    ) -> Result<Self::Ok, Self::Error>;
+
+    /// Serialize the given `string`.
+    ///
+    /// The string is provided as an iterator of chunks, at arbitrary boundaries,
+    /// with a total concatenated length of `total_len` which callers can assume.
```
gefjon, line 170:
```suggestion
    /// with a total concatenated length of `total_len` which callers can trust.
```

crates/sats/src/ser.rs
```diff
@@ -124,10 +124,82 @@ pub trait Serializer: Sized {
...
     ) -> Result<Self::Ok, Self::Error>;
+
+    /// Serialize the given `bsatn` encoded data of type `ty`.
+    ///
+    /// This is a concession to performance,
+    /// allowing some implementations to write the buffer directly.
+    ///
+    /// # Safety
+    ///
+    /// - `AlgebraicValue::decode(ty, &mut bytes).is_ok()`
```
gefjon, line 135:
```suggestion
    /// - `AlgebraicValue::decode(ty, &mut bytes).is_ok()`.
    ///   That is, `bytes` encodes a valid element of `ty`.
```

gefjon (approved):
Comment-related suggestions. Code looks good.

crates/sats/src/satn.rs
gefjon: I misread (I blame the GitHub UI) and thought these were part of the above `trait Satn`.

PR #756, opened 2024-01-25
Add workflow to tag latest docker image on release

drogus:
# Description of Changes

This PR adds a workflow to tag a docker image linked to the release as `latest`

Low, just a workflow change

### DIFF
--- a/.github/workflows/tag-release.yml
+++ b/.github/workflows/tag-release.yml
@@ -1,0 +1,22 @@
+on:
+  release:
+    types: [published]
+
+jobs:
+  on-release:
+    name: Re-tag latest
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
+      - name: Login to DockerHub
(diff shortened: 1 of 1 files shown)


kurtismullins:
@drogus  -- Looks great from simply reading it but it will be impractical to test until our next release. 

We should confirm with @cloutiertyler and @jdetter that we want to auto-tag `spacetimedb:latest` because I _vaguely_ remember someone mentioning they want it to be a manual step. I could be wrong!

drogus:
> Looks great from simply reading it but it will be impractical to test until our next release.

Yeah, that's the biggest problem I think 😅 fwiw I've tested it on a test repo and it worked correctly - the only thing that I chnaged bringing it back here is the docker image name

> We should confirm with @cloutiertyler and @jdetter that we want to auto-tag spacetimedb:latest because I vaguely remember someone mentioning they want it to be a manual step. I could be wrong!

👍

kurtismullins (approved):
I saw the verbal approval from John that this is the release behavior we want. Ship it!

PR #761, opened 2024-01-26
Fixed broken quickstart links in readme

jdetter:
# Description of Changes

- User Rory in the discord reported this as an issue. All of the quickstart links were returning 404, so I've updated them to be the correct links.

corresponding GitHub label.

No ABI/API break

0


NateTheDev1 (approved):

jdetter:
This is just a readme change, bypassing protections to save the CI from having to run on this

PR #766, opened 2024-01-26
Energy metering for persistent memory usage

coolreader18:
# Description of Changes

As on the tin. I think this is the idea, right? Or am I missing another part of the database that stores user data in-memory.

2

### DIFF
--- a/crates/standalone/src/energy_monitor.rs
+++ b/crates/standalone/src/energy_monitor.rs
@@ -40,6 +40,11 @@ impl EnergyMonitor for StandaloneEnergyMonitor {
         let amount = EnergyQuanta::from_disk_usage(disk_usage, period);
         self.withdraw_energy(database.owner_identity, amount)
     }
+
+    fn record_memory_usage(&self, database: &Database, _instance_id: u64, mem_usage: u64, period: Duration) {
+        let amount = EnergyQuanta::from_memory_usage(mem_usage, period);
+        self.withdraw_energy(database.owner_identity, amount)
+    }
 }
 
 impl StandaloneEnergyMonitor {
--- a/crates/table/src/blob_store.rs
+++ b/crates/table/src/blob_store.rs
@@ -15,6 +15,8 @@ use blake3::hash;
 use spacetimedb_data_structures::map::{Entry, HashMap};
 use spacetimedb_lib::{de::Deserialize, ser::Serialize};
 
+use crate::MemoryUsage;
+
 /// The content address of a blob-stored object.
 #[derive(Eq, PartialEq, PartialOrd, Ord, Clone, Copy, Hash, Debug, Serialize, Deserialize)]
 pub struct BlobHash {
@@ -23,6 +25,8 @@ pub struct BlobHash {
     /// Uses BLAKE3 which fits in 32 bytes.
     pub data: [u8; BlobHash::SIZE],
 }
+
+impl MemoryUsage for BlobHash {}
 
 impl BlobHash {
     /// The size of the hash function's output in bytes.
@@ -142,12 +146,26 @@ pub struct HashMapBlobStore {
     map: HashMap<BlobHash, BlobObject>,
 }
 
+impl MemoryUsage for HashMapBlobStore {
+    fn heap_usage(&self) -> usize {
+        let Self { map } = self;
+        map.heap_usage()
+    }
+}
+
 /// A blob object including a reference count and the data.
 struct BlobObject {
     /// Reference count of the blob.
     uses: usize,
     /// The blob data.
     blob: Box<[u8]>,
+}
+
+impl MemoryUsage for BlobObject {
+    fn heap_usage(&self) -> usize {
+        let Self { uses, blob } = self;
+        uses.heap_usage() + blob.heap_usage()
+    }
 }
 
 impl BlobStore for HashMapBlobStore {
--- a/crates/table/src/memory_usage.rs
+++ b/crates/table/src/memory_usage.rs
@@ -1,0 +1,247 @@
+use std::hash::{BuildHasher, Hash};
+use std::mem;
+
+use spacetimedb_sats::{
+    algebraic_value::Packed, i256, u256, AlgebraicType, AlgebraicValue, ArrayType, ArrayValue, MapType, ProductType,
+    ProductTypeElement, ProductValue, SumType, SumTypeVariant, SumValue,
+};
+
+/// For inspecting how much memory a value is using.
+///
+/// This trait specifically measures heap memory. If you want to measure stack memory too, add
+/// `mem::size_of_val()` to it. (This only really matters for the outermost type in a hierarchy.)
+pub trait MemoryUsage {
+    /// The **heap** memory usage of this type. The default implementation returns 0.
+    #[inline(always)]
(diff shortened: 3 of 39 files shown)


crates/core/src/database_instance_context_controller.rs
```diff
@@ -144,13 +146,14 @@ impl DatabaseInstanceContextController {
         }
     }
 
-    pub fn start_disk_monitor(&self) {
-        tokio::spawn(self.clone().disk_monitor());
+    fn start_usage_monitors(self: &Arc<Self>) {
+        tokio::spawn(Self::disk_monitor(Arc::downgrade(self)));
+        tokio::spawn(Self::memory_monitor(Arc::downgrade(self)));
```
joshua-spacetime, line 151: The downgrade is not strictly necessary right?

crates/core/src/database_instance_context_controller.rs
```diff
@@ -160,14 +163,43 @@ impl DatabaseInstanceContextController {
...
                         .record_disk_usage(&db.database, db.database_instance_id, disk_usage.sum(), dt);
                     *prev_disk_usage = disk_usage;
                 }
             }
+
+            prev_tick = tick;
+        }
+    }
+
+    const MEM_METERING_INTERVAL: Duration = Duration::from_secs(100);
```
joshua-spacetime, line 182: Why not use the same duration as disk metering?

crates/sats/src/algebraic_value.rs
```diff
@@ -110,6 +110,32 @@ pub enum AlgebraicValue {
...
+            | AlgebraicValue::I16(_)
+            | AlgebraicValue::U16(_)
+            | AlgebraicValue::I32(_)
+            | AlgebraicValue::U32(_)
+            | AlgebraicValue::I64(_)
+            | AlgebraicValue::U64(_)
+            | AlgebraicValue::I128(_)
+            | AlgebraicValue::U128(_)
+            | AlgebraicValue::F32(_)
+            | AlgebraicValue::F64(_) => 0,
```
joshua-spacetime, line 134: Is it true that AlgebraicValues are always transient?

crates/sats/src/array_value.rs
```diff
@@ -47,6 +49,32 @@ pub enum ArrayValue {
...
+            Self::I16(x) => x.get_heap_size(),
+            Self::U16(x) => x.get_heap_size(),
+            Self::I32(x) => x.get_heap_size(),
+            Self::U32(x) => x.get_heap_size(),
+            Self::I64(x) => x.get_heap_size(),
+            Self::U64(x) => x.get_heap_size(),
+            Self::I128(x) => x.get_heap_size(),
+            Self::U128(x) => x.get_heap_size(),
+            Self::F32(x) => x.capacity() * 4,
+            Self::F64(x) => x.capacity() * 8,
```
joshua-spacetime, line 70: Why are these different?

crates/sats/src/data_key.rs
```diff
@@ -16,6 +16,9 @@ pub enum DataKey {
     Hash(Hash),
 }
 
+#[cfg(feature = "get-size")]
+impl get_size::GetSize for DataKey {}
```
joshua-spacetime, line 20: Why can't we derive here?

crates/standalone/src/energy_monitor.rs
```diff
@@ -37,7 +37,12 @@ impl EnergyMonitor for StandaloneEnergyMonitor {
...
 
     fn record_disk_usage(&self, database: &Database, _instance_id: u64, disk_usage: u64, period: Duration) {
-        let amount = EnergyQuanta::from_disk_usage(disk_usage, period);
+        let amount = EnergyQuanta::from_data_storage(disk_usage, period);
+        self.withdraw_energy(database.identity, amount)
+    }
+
+    fn record_memory_usage(&self, database: &Database, _instance_id: u64, mem_usage: u64, period: Duration) {
+        let amount = EnergyQuanta::from_data_storage(mem_usage, period);
         self.withdraw_energy(database.identity, amount)
```
joshua-spacetime, line 46: Note, memory consumes more energy than disk. I believe Tyler suggested a 100X scale factor over disk.

Also can you add a new parameter to `withdraw_energy` to distinguish among the different dimensions?

cloutiertyler: I confirm that.

crates/core/src/database_instance_context_controller.rs
coolreader18: lmao, I thought Tyler said that the interval should be 100, not that the energy multiplier should be 100.

crates/core/src/database_instance_context_controller.rs
coolreader18: It makes it so the task doesn't continue on forever, even when no one else is holding onto the controller.

crates/standalone/src/energy_monitor.rs
coolreader18: Fixed the scale factor. I don't know if withdraw_energy needs a dimension param, cause it's just for standalone. Where would the dimensions be stored?

bfops:
@coolreader18 I just want to confirm for PR labeling purposes - this change is backwards-compatible right?

coolreader18:
@bfops yes, it is.

coolreader18:
Oh, from the user side at least - private will need an update (but I have it ready already)

coolreader18:
The hashbrown update is because 0.15 added an `allocation_size()` method, which is obviously very helpful. It also changes the default hasher, though, and so all the `HashCollectionExt` imports are because we can no longer use `HashMap::new()`.

joshua-spacetime (commented):
Thanks @coolreader18! Just a few typos, but otherwise looks good.

You've accounted for nearly all mem usage :)

I think it would have been sufficient to just track the pages, indexes, blob store, and pointer map. That probably could have reduced a lot of the boilerplate. But this is very nice nonetheless.

crates/table/src/memory_usage.rs
```diff
@@ -0,0 +1,252 @@
...
+                .map(|(k, v)| k.memory_usage() + v.memory_usage())
+                .sum::<usize>()
+    }
+}
+
+impl<K: MemoryUsage, V: MemoryUsage> MemoryUsage for std::collections::BTreeMap<K, V> {
+    fn memory_usage(&self) -> usize {
+        self.iter()
+            .map(|(k, v)| k.memory_usage() + v.memory_usage())
+            .sum::<usize>()
```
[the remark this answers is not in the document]
coolreader18: Yeah, I should put a comment here that this is best-effort.

crates/table/src/blob_store.rs
```diff
@@ -24,6 +26,8 @@ pub struct BlobHash {
     pub data: [u8; BlobHash::SIZE],
 }
 
+impl MemoryUsage for BlobHash {}
```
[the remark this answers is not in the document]
coolreader18: `memory_usage()` is only heap memory usage - the 32 is in the `mem::size_of::<BlobHash>()`, so it doesn't need to count it in from the function.

joshua-spacetime (approved):

PR #769, opened 2024-01-27
Integrate mem arch into `locking_tx_datastore`

Centril:
# Description of Changes

After several grueling months, this PR implements the shiny new SpacetimeDB datastore.

We retain the transactional semantics of the locking TX datastore, so we re-use the existing module `locking_tx_datastore` and struct `Locking`, updating them to use our new page-based tables rather than hash-maps of `ProductValue`.

For design and rationale, see the mem-arch proposal, which is not public.

No external APIs or ABIs are broken.

5, baby! We spent months considering the correct design here and then implementing that design to the best of our ability, but this is a huge change that touches large parts of the codebase.

Notably, everything in the `table` crate has already received thorough review from @kazimuth (modulo recent bugfixes and cleanups), but the changes in `core` are essentially fresh.

This does *not* mean that the changes are untested! @jdetter has tested this branch extensively with BitCraft.

# Review

I (pgoldman) am particularly interested in review from:

- [x] @kazimuth that our uses of the `table` crate line up with his expectations.
- [x] @Shubham8287 that our implementation of transactionality in `relational_db.rs` and `locking_tx_datastore` match the previous semantics.
- [x] @joshua-spacetime that we haven't messed up the metrics, and that we have included any recent changes to the metrics.
- [x] @jdetter that the performance regressions introduced by maintaining our existing APIs and ABIs are acceptable, and that he understands our plans to eliminate those regressions.
- [x] @cloutiertyler that our implementation here lines up with his vision.


jdetter (approved):
From a testing standpoint, I've heavily tested this PR for the last few days and it works - I'll give the code another review later today.

joshua-spacetime (approved):
The metrics look good, thank you both!

kazimuth (approved):
Table stuff LGTM!

Unsafe audit:
- [x] `Table::find_same_row`
- [x] `Table::row_hash_for`
- [x] `Table::delete_internal_skip_pointer_map`
- [x] `row_hash::hash_row_in_page`

Shubham8287 (approved):
Looks Good! Happy to see, mem-arch also broke up long locking_datastore file into small mods.

crates/core/src/db/datastore/locking_tx_datastore/state_view.rs
```diff
@@ -287,37 +284,33 @@ impl<'a> Iterator for Iter<'a> {
...
-                                    // and hasn't been removed in the current tx,
-                                    // so it exists and can be returned.
-                                    RowState::Absent => return Some(DataRef::new(row_id, row)),
-                                }
-                            }
-                            None => return Some(DataRef::new(row_id, row)),
+                        // Check the committed row's state in the current tx.
+                        // If it's been deleted, skip it.
+                        // If it's still present, yield it.
+                        if !self
```
cloutiertyler, line 293: This appears to be incorrect. We want to skip it if it's been inserted or deleted, and we only want to return it if it's not done either. See old code:

>                                   // Do nothing, via (3), we'll get it in the next stage (2).
>                                   RowState::Insert(_) |
>                                   // Skip it, it's been deleted.
 >                                   RowState::Delete => {}

cloutiertyler (changes requested):
Overall matches the vision. I left a few random comments and nits, which don't necessarily need to be addressed, however I also found what I believe to be a critical logic error in table iteration in the `Iter` struct in the `state_view.rs` file. Please address.

crates/core/src/db/datastore/locking_tx_datastore/tx.rs
```diff
@@ -1,20 +1,21 @@
...
 use spacetimedb_sats::{db::def::TableSchema, AlgebraicValue};
-use std::time::{Duration, Instant};
+use std::{
+    ops::RangeBounds,
+    time::{Duration, Instant},
+};
 
-#[allow(dead_code)]
 pub struct TxId {
-    pub(crate) committed_state_shared_lock: SharedReadGuard<CommittedState>,
```
cloutiertyler, line 15: Nit: Seems a little odd that these are pub.

crates/core/src/db/datastore/locking_tx_datastore/state_view.rs
cloutiertyler: I think we should also add a test case for this.

gefjon: The previous implementation did something silly when you deleted and re-inserted a row within a transaction: it had that row present in all three of the committed state, the delete table and the insert table. The new implementation is more principled; a row can be in only the following states:

- Not present anywhere.
- Unmodified; in committed state only.
- Newly inserted; in insert table only.
- Newly deleted; in committed state and delete table.

This means that the committed state and the insert table are disjoint sets.

crates/core/src/db/datastore/locking_tx_datastore/committed_state.rs
```diff
@@ -16,299 +15,254 @@ use crate::{
...
-                commit_table.insert(row_id, row.clone());
-                let pv = row;
-                let bytes = match row_id.0 {
-                    DataKey::Data(data) => Arc::new(data.to_vec()),
-                    DataKey::Hash(_) => memory.get(&row_id.0).unwrap().clone(),
-                };
-                TxRecord {
-                    op: TxOp::Insert(bytes),
+        // Insert the table row into st_tables, creating st_tables if it's missing
+        self.with_table_and_blob_store_or_create(ST_TABLES_ID, st_table_schema(), |st_tables, blob_store| {
```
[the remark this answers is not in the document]
gefjon: I happen to prefer it since it makes the lifetimes explicit, but since you don't agree I'm happy to change it.

crates/core/src/db/datastore/locking_tx_datastore/mut_tx.rs
```diff
@@ -790,214 +784,259 @@ impl MutTxId {
...
+    ) -> Result<IterByColRange<'a, R>> {
+        // We have to index_seek in both the committed state and the current tx state.
+        // First, we will check modifications in the current tx. It may be that the table
+        // has not been modified yet in the current tx, in which case we will only search
+        // the committed state. Finally, the table may not be indexed at all, in which case
+        // we fall back to iterating the entire table.
+        //
+        // We need to check the tx_state first. In particular, it may be that the index
+        // was only added in the current transaction.
+        // TODO(george): It's unclear that we truly support dynamically creating an index
```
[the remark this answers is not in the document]
gefjon: I don't feel that I know this. If you do, I think a follow-up PR to describe what's supported is warranted.

crates/core/src/db/datastore/locking_tx_datastore/state_view.rs
gefjon: Added a test in `relational_db.rs`, `test_insert_delete_insert_iter`, which exercises this path.

cloutiertyler (approved):
The new changes look great. Thanks for addressing all of that. It is time.

PR #774, opened 2024-01-28
table_name: use Arc<str>

Centril:
# Description of Changes

Switches most cases where a table / reducer name is stored to an `Arc<str>` rather than `String`.

None

1


cloutiertyler (approved):
This looks good to me in principle. We should just coordinate the merging.

mamcx:
Also, I think instead is the whole `schema` that should be under `Arc/Rc`

mamcx:
Because now `DbTable` & 'MemTable` both have ` pub head: Arc<Header>` this PR can be closed, right?

Centril:
This is being done in a different PR.

PR #775, opened 2024-01-29
core: Lock `unwritten_commit` across I/O

kim:
A database may continue to write to the log after a previous write
failed (due to an I/O error). At that point, offsets have already been
updated, which results in gaps in the offset sequence (which we can't
tolerate currently).

To fix this, the lock on `unwritten_commit` is acquired and held across
the write to the log segment. Only after that succeeded, offsets and
parent hash are updated.

### DIFF
--- a/crates/core/src/db/commit_log.rs
+++ b/crates/core/src/db/commit_log.rs
@@ -309,7 +309,7 @@ impl CommitLogMut {
         &self,
         ctx: &ExecutionContext,
         tx_data: &TxData,
-        datastore: &D,
+        _datastore: &D,
     ) -> Result<Option<usize>, DBError>
     where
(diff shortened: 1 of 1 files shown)


joshua-spacetime (approved):
Are you able to add a unit test?

cloutiertyler (approved):
This looks good to me. Thanks for this change. I think this will make issues much less likely.

I guess the issue would still be possible to happen if the write call doesn't throw an error, but the data is actually not persisted to disk for some reason. Don't know if that's actually possible with the unix write syscalls. Either way it should be much less likely with this change.

kim:
> Are you able to add a unit test?

I would not know how — the append call would need to be abstracted out.

Issue #758, opened 2024-01-25
Subscriptions are not consistent


joshua-spacetime:
Reducers return once they have sent their updates to the subscription actor. It is possible though that another reducer acquires the database lock before the subscription evaluation does. This is a potential a consistency violation since in the case of joins, incremental evaluation reads database state. It needs to be the case, under the current architecture, that subscriptions be evaluated logically as part of the originating reducer transaction.

joshua-spacetime:
Closed by #776

PR #776, opened 2024-01-29
fix broadcast_event_blocking

Shubham8287:
# Description of Changes
Solves #758 
current implementation of `broadcast_event_blocking`  was not exactly blocking to wait till broadcast finishes.
Fixed that, Added a Mutex as concurrent access through `broadcast_event_blocking` or `spawn` can lead to data race.

3, simple change but has potential to lead to deadlocks, perfomance regression.


crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -73,35 +77,52 @@ impl ModuleSubscriptionManager {
...
             EventStatus::OutOfEnergy => {} // ?
         }
     }
 
     pub fn broadcast_event_blocking(&self, client: Option<&ClientConnectionSender>, event: ModuleEvent) {
-        tokio::runtime::Handle::current().block_on(self.broadcast_event(client, event))
+        tokio::runtime::Handle::current().block_on(async move {
+            match event.status {
+                EventStatus::Committed(_) => {
+                    if let Err(e) = self.actor.lock_arc().broadcast_commit_event(event).await {
```
joshua-spacetime, line 98: Won't this lock be released once the message has been received on the other end? Don't we want to hold the lock until all of the subscriptions have been evaluated?

Shubham8287: thanks to point, I would live to hold this PR a bit, I think this can be done in much better way directly in subscription actor instead of here, so moved it to draft.

jdetter:
I will bot test this

crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -39,6 +39,8 @@ enum Command {
     },
     BroadcastCommitEvent {
         event: ModuleEvent,
+        // channel for signaling broadcast completion.
```
joshua-spacetime, line 42: nit: channel for signaling that all subscriptions have been evaluated.

crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -76,9 +78,13 @@ impl ModuleSubscriptionManager {
     pub async fn broadcast_event(&self, client: Option<&ClientConnectionSender>, mut event: ModuleEvent) {
         match event.status {
             EventStatus::Committed(_) => {
+                let (tx, rx) = oneshot::channel();
                 self.tx
-                    .send(Command::BroadcastCommitEvent { event })
+                    .send(Command::BroadcastCommitEvent { event, feedback_tx: tx })
                     .expect("subscription actor panicked");
+                if (rx.await).is_err() {
+                    log::error!("subscription actor not reported broadcast completion")
```
joshua-spacetime, line 86: nit: "failed to receive acknowledgement of successful evaluation from subscription actor"

joshua-spacetime (approved):
Just a couple comment revisions but otherwise LGTM.

PR #777, opened 2024-01-29
Fix bootstrapping ids for relating objects like indexes

mamcx:
# Description of Changes

The bootstrapping of system tables keeps the ids of related objects as `0` because is called `schema_for_table` that first consults the cache.

Now, after the bootstrapping is done, re-read the data from the database to be sure the correct `ids` are assigned.

This is an internal break because the `system tables` schemas get `0` for all the dependant objects `ids` and now get the correct one.

1


gefjon (changes requested):
Marking as request changes because I'd prefer to wait until after merging the new mem-arch before merging this.

gefjon (approved):

PR #779, opened 2024-01-29
Use rayon for broadcast_commit_event

coolreader18:
# Description of Changes

Switch to rayon for broadcast_commit_event - this makes the concurrency more straightforward to follow, and I think would make it easier to update in the future.

1


crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -202,29 +197,35 @@ impl ModuleSubscriptionActor {
...
+            |tx, subscription| -> Result<_, DBError> {
+                let database_update = event.status.database_update().unwrap();
+                let incr = subscription
+                    .queries
+                    .eval_incr(&self.relational_db, tx, database_update, auth)?;
+                Ok((subscription, incr))
+            },
+        );
+        let iter = iter.map(|r| -> Result<(), DBError> {
+            let (subscription, incr) = r?;
```
coolreader18, line 221: I'm not sure what the proper behavior is here, so I'm being conservative, but maybe someone else knows - if one subscriber's eval_incr fails, should we just log it and carry on? Before, it short-circuited the whole thing so that no one got a subscription update, but that doesn't seem super correct.

crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -202,29 +197,35 @@ impl ModuleSubscriptionActor {
...
-            mut event: ModuleEvent,
-            subscription: SubscriptionRw,
-            relational_db: Arc<RelationalDB>,
-        ) -> Result<(), DBError> {
-            let ctx = ExecutionContext::incremental_update(relational_db.address());
-            let mut tx = relational_db.begin_tx();
-            let database_update = event.status.database_update().unwrap();
-            let subscription = subscription.read_arc();
-            let futures = FuturesUnordered::new();
+    fn broadcast_commit_event(&mut self, event: ModuleEvent) -> Result<impl Future<Output = ()> + '_, DBError> {
```
joshua-spacetime, line 200: Nice to see this is no longer async!

joshua-spacetime (commented):
I like this change. My only question concerns performance - how much overhead does rayon add?

coolreader18:
@joshua-spacetime No overhead, actually! It should make things more efficient, since instead of spawning N threads for however many subscribers there are, it puts it on a threadpool, so it's still parallelized but there's less contention. (John and I just figured out that that attribute fixes the deadlock we've been having)

crates/core/src/host/mod.rs
```diff
@@ -45,13 +46,13 @@ impl ReducerArgs {
         Ok(match self {
             ReducerArgs::Json(json) => ArgsTuple {
                 tuple: from_json_seed(&json, SeedWrapper(ReducerDef::deserialize(schema)))?,
-                bsatn: None,
-                json: Some(json),
+                bsatn: OnceCell::new(),
+                json: OnceCell::with_value(json),
```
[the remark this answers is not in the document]
coolreader18: Yeah, cause the broadcast_commit tasks need to access it concurrently.

crates/core/src/subscription/module_subscription_actor.rs
coolreader18: Yeah, it still is sorta since it returns a future, but it's better encapsulated I feel.

jdetter (approved):
Tested using the bots (128CCU) and this appears to fix the deadlock issue, thank you Noa! 🎉

joshua-spacetime (approved):

PR #780, opened 2024-01-30
Add support to convert to chrono for Timestamp

### LINKED ISSUE #753: Add methods to Timestamp for deriving month and year
The Timestamp type in the bindings module should have methods `month()` and `year()`. Note this means adding a dependency on the `chrono` library in all likelihood.

mamcx:
# Description of Changes

Add `chrono` as a dependency for `bindings` so it is possible to query `month, year, ...`.

Closes  #753 

1

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4152,6 +4152,7 @@ dependencies = [
 name = "spacetimedb"
 version = "0.9.3"
 dependencies = [
+ "chrono",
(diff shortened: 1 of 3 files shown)


cloutiertyler:
https://github.com/clockworklabs/SpacetimeDB/issues/753#issuecomment-2229251123

PR #785, opened 2024-02-01
Fix assorted typos

bfops:
# Description of Changes
Just fixes typos in various strings, comments, etc.

Nope. Only edits to strings, comments, etc.

1


jdetter (approved):
LGTM 👍

PR #789, opened 2024-02-05
`trait ReadColumn`, to read a single column from a `RowRef`

gefjon:
# Description of Changes

This commit defines `trait ReadColumn`, which is implemented for types that can be stored in a table column and can be read out of said table column.

Its interesting method is `read_column(row: RowRef<'_>, idx: usize) -> Result<Self, TypeError>`, which attempts to read the `idx`th column of `row` as a value of type `Self`, returning a `TypeError` if the row in question does not have the appropriate types.

`ReadColumn` is implemented for Rust equivalents of all non-compound `AlgebraicType`s, i.e. integers, floats, `bool` and `String`. It is also implemented for all SATS value types, including those which represent values of compound types, i.e. `AlgebraicValue`, `ProductValue`, `SumValue`, `ArrayValue` and `MapValue`.

Note that this commit does not use the `ReadColumn` interface anywhere, except in tests. Follow-ups will use `ReadColumn` in various places, including:

- Interacting with system table rows in the datastore.
- Filtering rows in datastore iterators.
- Filtering rows in the VM.
- Inserting into indexes.

None.

3: non-trivial uses of `unsafe`; new uses of existing `bflatn_from` serialization.

### DIFF
--- a/crates/table/src/read_column.rs
+++ b/crates/table/src/read_column.rs
@@ -1,0 +1,507 @@
+//! Provides a trait [`ReadColumn`] for extracting a single column from a [`crate::table::RowRef`].
+//! This is desirable as frequently, e.g. when evaluating filtered queries,
+//! we are interested in only a single column (or a small set of columns),
+//! and would like to avoid the allocation required by a `ProductValue`.
+
+use crate::{
+    bflatn_from,
+    indexes::{PageOffset, Size},
+    layout::{AlgebraicTypeLayout, HasLayout, PrimitiveType, ProductTypeElementLayout, VarLenType},
+    table::RowRef,
+    util::slice_assume_init_ref,
+};
+use spacetimedb_sats::{
+    algebraic_value::ser::ValueSerializer, AlgebraicType, AlgebraicValue, ArrayValue, MapValue, ProductType,
+    ProductValue, SumValue,
+};
+use std::{cell::Cell, mem};
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum TypeError {
+    #[error(
+        "Attempt to read column {} of a product with only {} columns of type {:?}",
+        desired,
+        found.elements.len(),
+        found,
+    )]
+    IndexOutOfBounds { desired: usize, found: ProductType },
+    #[error("Attempt to read a column at type `{desired}`, but the column's type is {found:?}")]
+    WrongType {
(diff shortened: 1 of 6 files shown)


gefjon:
Also note that I ran the `read_column` tests under Miri, except:

- I changed the `ProptestConfig` to only 16 cases.
- I disabled `read_column_long_string`, since large blobs require `blake3`, which uses SIMD intrinsics and therefore doesn't work on Miri.

Other than that, the Miri tests passed (after about half an hour).

crates/table/src/read_column.rs
```diff
@@ -0,0 +1,507 @@
...
+                assert_wrong_type_error::<u32>(row_ref, idx, &col_ty.algebraic_type, AlgebraicType::U32)?;
+                assert_wrong_type_error::<i32>(row_ref, idx, &col_ty.algebraic_type, AlgebraicType::I32)?;
+                assert_wrong_type_error::<u64>(row_ref, idx, &col_ty.algebraic_type, AlgebraicType::U64)?;
+                assert_wrong_type_error::<i64>(row_ref, idx, &col_ty.algebraic_type, AlgebraicType::I64)?;
+                assert_wrong_type_error::<u128>(row_ref, idx, &col_ty.algebraic_type, AlgebraicType::U128)?;
+                assert_wrong_type_error::<i128>(row_ref, idx, &col_ty.algebraic_type, AlgebraicType::I128)?;
+                assert_wrong_type_error::<f32>(row_ref, idx, &col_ty.algebraic_type, AlgebraicType::F32)?;
+                assert_wrong_type_error::<f64>(row_ref, idx, &col_ty.algebraic_type, AlgebraicType::F64)?;
+                assert_wrong_type_error::<bool>(row_ref, idx, &col_ty.algebraic_type, AlgebraicType::Bool)?;
+                assert_wrong_type_error::<String>(row_ref, idx, &col_ty.algebraic_type, AlgebraicType::String)?;
```
kazimuth, line 392: that's an entertaining pattern

kazimuth (approved):
Looks good to me, clean tests and macros.

This is one of those things that seems particularly jittable.

crates/table/src/read_column.rs
gefjon: Can you tell I miss Common Lisp having types at runtime?

PR #791, opened 2024-02-05
Simplify broadcast_commit_event

coolreader18:
# Description of Changes

Simple change to remove an unnecessary mut reference and therefore simplify how the tx for broadcast_commit_event gets created.

### DIFF
--- a/crates/core/src/subscription/module_subscription_actor.rs
+++ b/crates/core/src/subscription/module_subscription_actor.rs
@@ -19,8 +19,8 @@ use crate::{
     db::relational_db::Tx,
     host::module_host::{EventStatus, ModuleEvent},
 };
-use futures::{stream::FuturesUnordered, Future, StreamExt};
-use rayon::iter::{IntoParallelRefMutIterator, ParallelIterator};
+use futures::Future;
+use rayon::iter::{IntoParallelRefIterator, ParallelIterator};
 use spacetimedb_lib::identity::AuthCtx;
 use spacetimedb_lib::Identity;
 use tokio::sync::{mpsc, oneshot};
@@ -126,7 +126,7 @@ impl ModuleSubscriptionActor {
             Command::AddSubscriber { sender, subscription } => self.add_subscription(sender, subscription).await?,
             Command::RemoveSubscriber { client_id } => self.remove_subscriber(client_id),
             Command::BroadcastCommitEvent { event, feedback_tx } => {
-                self.broadcast_commit_event(event).await;
+                tokio::task::block_in_place(|| self.broadcast_commit_event(event)).await;
                 let _ = feedback_tx.send(());
             }
         }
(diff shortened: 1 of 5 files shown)


crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -126,7 +126,7 @@ impl ModuleSubscriptionActor {
             Command::AddSubscriber { sender, subscription } => self.add_subscription(sender, subscription).await?,
             Command::RemoveSubscriber { client_id } => self.remove_subscriber(client_id),
             Command::BroadcastCommitEvent { event, feedback_tx } => {
-                self.broadcast_commit_event(event).await;
+                tokio::task::block_in_place(|| self.broadcast_commit_event(event)).await;
```
joshua-spacetime, line 129: Can wait to do this until we observe some measurements? It's not apparent that this is, or will be, a bottleneck.

joshua-spacetime (changes requested):

crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -206,57 +206,61 @@ impl ModuleSubscriptionActor {
...
+                };
+                let mut message = CachedMessage::new(message);
 
-        let iter = iter.map(|(subscription, incr)| {
-            let message = TransactionUpdateMessage {
-                event: &event,
-                database_update: incr,
-            };
-            let mut message = CachedMessage::new(message);
+                subscription.subscribers().iter().cloned().map(move |subscriber| {
```
Shubham8287, line 251: What do you think about doing it in single filter_map? So we don't have to clone. May be,  something like this?
```
            .filter_map(|subscription| {
                let incr = subscription
                    .queries
                    .eval_incr(&self.relational_db, tx, database_update, auth);
                match incr {
                    Ok(incr) if incr.tables.is_empty() => None,
                    Ok(incr) => {
                        let message = TransactionUpdateMessage {
                            event: &event,
                            database_update: incr,
                        };
                        let mut message = CachedMessage::new(message);
        
                        let tasks = subscription.subscribers().iter().map(move |subscriber| {
                            let message = message.serialize(subscriber.protocol);
                            tokio_handle.spawn(async move {
                                let _ = subscriber.send(message).await;
                            })
                        }).collect::<Vec<_>>();
                        Some(tasks)
                    },
                    Err(err) => {
                        // TODO: log an id for the subscription somehow as well
                        tracing::error!(err = &err as &dyn std::error::Error, "subscription eval_incr failed");
                        None
                    }
                }
            })
            .collect::<Vec<_>>();
            
```

Shubham8287 (commented):
Changes look good overall, I have added a comment for re-considering on a `.clone()`.

crates/core/src/subscription/module_subscription_actor.rs
coolreader18: The spawning is cheap, it's just an Arc inc ref. And the cloning is necessary because the future passed to spawn() needs to be `'static`

crates/core/src/subscription/module_subscription_actor.rs
coolreader18: The `block_in_place`? `broadcast_commit_event` evaluates an arbitrary set of queries, obviously the query executor has good performance but it still has the potential to block for an arbitrary amount of time.

crates/core/src/subscription/module_subscription_actor.rs
coolreader18: And I switched from `FuturesUnordered` to task spawning because otherwise the messages don't actually get sent until after *all* the queries are finished evaluating.

Shubham8287: makes sense, I agree it doesn't seems we can avoid clone but that doesn't looks like Arc inc to me.

Shubham8287:
LGTM :heavy_check_mark:

Shubham8287 (approved):

crates/core/src/subscription/module_subscription_actor.rs
joshua-spacetime: It really depends on the workload, right? My point is that we can't immediately tell that this will improve performance globally. It would be nice to have some measurements to back it up.

joshua-spacetime (approved):

crates/core/src/subscription/module_subscription_actor.rs
coolreader18: It was less about the performance aspect and more about the correctness of blocking inside of async; although, it's irrelevant since it only ever gets called from within the module host actor anyway.

Issue #788, opened 2024-02-02
Subscriptions are not consistent part II


joshua-spacetime:
As of #758, the ordering inconsistencies for incremental evaluation have been fixed. However the same inconsistency is still present for the initial subscribe call. In particular it is possible, subsequent to a transaction committing, for the corresponding incremental evaluation and an initial subscribe call to race such that the subscribe call is evaluated first. In this case it is possible for a client to receive two updates for the same state transition, putting the client in an invalid state.

From the perspective of the database, a transaction plus an incremental evaluation must be atomic to avoid such inconsistencies.

joshua-spacetime:
Closed by #792

PR #792, opened 2024-02-05
Add a RwLock around module subscriptions to prevent a race condition

coolreader18:
# Description of Changes

Fixes that issue. From a comment I added in module_subscription_actor:

> Take a lock on our subscriptions [before we commit the tx]. Otherwise, we could have a race condition where we commit the tx, someone adds a subscription and receives this tx as an update, and then receives the update again when we broadcast_event.

2

### DIFF
--- a/crates/core/src/host/wasm_common/module_host_actor.rs
+++ b/crates/core/src/host/wasm_common/module_host_actor.rs
@@ -2,6 +2,7 @@ use anyhow::{anyhow, Context};
 use bytes::Bytes;
 use std::sync::Arc;
 use std::time::Duration;
+use tokio::sync::RwLock;
 
 use spacetimedb_lib::buffer::DecodeError;
 use spacetimedb_lib::identity::AuthCtx;
@@ -24,7 +25,7 @@ use crate::host::{ArgsTuple, EntityDef, ReducerCallResult, ReducerId, ReducerOut
 use crate::identity::Identity;
 use crate::messages::control_db::Database;
 use crate::sql;
-use crate::subscription::module_subscription_actor::ModuleSubscriptionManager;
+use crate::subscription::module_subscription_actor::ModuleSubscriptions;
 use crate::util::{const_unwrap, ResultInspectExt};
 use crate::worker_metrics::WORKER_METRICS;
 use spacetimedb_sats::db::def::TableDef;
@@ -142,7 +143,7 @@ impl<T: WasmModule> WasmModuleHostActor<T> {
 
         let owner_identity = database_instance_context.identity;
         let relational_db = database_instance_context.relational_db.clone();
-        let subscription = ModuleSubscriptionManager::spawn(relational_db, owner_identity);
+        let subscriptions = Arc::new(RwLock::new(ModuleSubscriptions::new(relational_db, owner_identity)));
 
         let uninit_instance = module.instantiate_pre()?;
         let mut instance = uninit_instance.instantiate(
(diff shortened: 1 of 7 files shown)


crates/core/src/host/wasm_common/module_host_actor.rs
```diff
@@ -563,6 +564,11 @@ impl<T: WasmInstance> WasmModuleInstance<T> {
             .with_label_values(&address, reducer_name)
             .observe(timings.total_duration.as_secs_f64());
 
+        // Take a lock on our subscriptions now. Otherwise, we could have a race condition where we commit
+        // the tx, someone adds a subscription and receives this tx as an update, and then receives the
+        // update again when we broadcast_event.
+        let subscriptions = self.info.subscriptions.blocking_read();
```
joshua-spacetime, line 570: As of right now, we won't have concurrent readers, since reducers execute sequentially. So we probably want this to be a write lock, and use the read lock for the initial subscription, since there we can have concurrent readers.

joshua-spacetime (changes requested):
This change makes subscriptions consistent again, so I'm in favor of it, with a couple caveats.

1. I think we should reverse the order of the locks.
    That is, subscribe should take a read lock and the reducer a write lock.

3. I'd like this patch to consist of changes related to the RwLock only
    That is, separate from further refactoring to the subscription code. The reason being is that there are now 3 main locks that are used for synchronizing access to the database - the datastore lock, the reducer semaphore, and now this one. And their interactions can be quite subtle.

coolreader18:
@joshua-spacetime re: point 2, this is based off #791, which is why there's unrelated changes

coolreader18:
Also, point 1: ...how would that work? Adding a subscriber needs mutable access to the vec. Unless we write/find a concurrent mutable data structure, that write lock is physically necessary to modify the subscriber list.

crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -4,142 +4,46 @@ use super::{
...
-        }
-    }
-
-    pub fn broadcast_event_blocking(&self, client: Option<&ClientConnectionSender>, event: ModuleEvent) {
-        tokio::runtime::Handle::current().block_on(self.broadcast_event(client, event))
-    }
-}
-
-struct ModuleSubscriptionActor {
+pub struct ModuleSubscriptions {
```
Shubham8287, line 23: `Vec<Subscription>>` field of this should be wrap around RwLock, so that we can do concurrent `add_subscription` and `add_subscription` with broadcast.
Current approach of wrapping `ModuleSubscription` will not really allow that.

joshua-spacetime:
@coolreader18 

> how would that work? Adding a subscriber needs mutable access to the vec.

My point was that the rest of the time (aside from mutating the list) we don't need exclusive access. **HOWEVER**, I would actually like to rescind my comment and say that we should keep it the way you have it, because I do think it's a good idea to keep performance and correctness fixes separate. 

> re: point 2

It still seems like there are more changes that are not related to the new RwLock. I'd really like this patch to only contain those changes necessary to fix the race condition. Even the performance related changes, which I originally suggested we do here, should be made in a separate patch.

Shubham8287:
Resolved my comments, as we can merge it for fixing #788, I am writting a separate patch to make add_subscription concurrent.

coolreader18:
> It still seems like there are more changes that are not related to the new RwLock

I don't think there are - switching from an actor to shared state required touching other files, but we can't have a lock around shared state if the only access is via a message channel.

joshua-spacetime (approved):
This looks good. I think I was probably looking at an old diff. This is in line with what I was expecting.

PR #793, opened 2024-02-05
BTreeIndex: Specialize on primitive key types for great performance

gefjon:
# Description of Changes

Rewrite of Tyler's #771, because I thought this would be easier than rebasing.

This commit hoists branching on key types outside of comparisons and searches in `BTreeIndex`, so that in many cases we can use e.g. `u64::cmp` instead of the much slower `AlgebraicValue::cmp`.

This design is kind of ugly, and will likely change in the future, but for now it's good enough, and is a meaningful performance improvement.

No external interfaces change.

2 - ugly code with lots of repetitive matching may open us up to copy/paste errors.


kazimuth (approved):
👍

gefjon:
benchmarks please

PR #794, opened 2024-02-05
perf(747): Single query evaluation for multiple subscriptions

joshua-spacetime:
# Description of Changes

corresponding GitHub label.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4556,6 +4556,7 @@ dependencies = [
  "async-trait",
  "backtrace",
  "base64 0.21.4",
+ "blake3",
  "bytes",
  "bytestring",
  "chrono",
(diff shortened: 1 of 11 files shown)


crates/core/src/subscription/database_subscription_manager.rs
```diff
@@ -0,0 +1,241 @@
...
+/// and then serializes and broadcasts them to the client.
+/// It also performs row deduplication if necessary.
+#[derive(Default)]
+pub struct Aggregator {
+    // Table IDs to table names.
+    names: HashMap<TableId, String>,
+    // If a client is subscribed to multiple queries,
+    // and those queries return rows from the same table,
+    // we must dedup rows before relaying to the client.
+    dedup: HashMap<TableId, HashSet<TableOp>>,
```
cloutiertyler, line 38: Did we not have to dedup before?

PR #796, opened 2024-02-06
BTreeIndex: use `ReadColumn` instead of `ProductValue` fields

gefjon:
# Description of Changes

Prior to this commit, `BTreeIndex::insert` and `BTreeIndex::delete` took a `&ProductValue`, the row to be inserted or deleted, as an argument, and extracted the indexed column value(s) from it.

With this commit, we instead take a `RowRef` referring to the row to be inserted or deleted, and access the indexed column value(s) using `ReadColumn`. For specialized indexes, we extract them directly as native types, rather than as `AlgebraicValue`.

Making this change involved a nasty split borrow problem, as we now need to have a `RowRef` to the inserted/deleted row coexisting with a mutable borrow of the index to be modified. This required introducing a `TableInner` struct,
which is the referent of `RowRef`.

No external interfaces broken.

2 - minimal `unsafe`, semantics of `Table`, `RowRef` and `BTreeIndex` preserved.

### DIFF
--- a/crates/core/src/db/datastore/locking_tx_datastore/committed_state.rs
+++ b/crates/core/src/db/datastore/locking_tx_datastore/committed_state.rs
@@ -312,7 +312,7 @@ impl CommittedState {
             };
             let mut index = BTreeIndex::new(
                 index_row.index_id,
-                &table.row_layout(),
+                table.row_layout(),
                 &index_row.columns,
                 index_row.is_unique,
                 index_row.index_name,
--- a/crates/core/src/db/datastore/locking_tx_datastore/committed_state.rs
+++ b/crates/core/src/db/datastore/locking_tx_datastore/committed_state.rs
@@ -312,7 +312,7 @@ impl CommittedState {
             };
             let mut index = BTreeIndex::new(
                 index_row.index_id,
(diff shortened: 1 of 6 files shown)


gefjon:
Callgrind please

kazimuth (approved):
👍 

Unsafe audit:
- [x] RowRef::new

gefjon:
callgrind please

gefjon:
benchmarks please

gefjon:
Hard to tell what's noise and what's real changes from the wall-time benchmarks, but it seems like this is neither a major performance gain or loss. That's good enough for me to want to merge, as this PR will make it easier down the line to remove the `ProductValue` return from `Table::delete`, and serves as a good template for how we can use `ReadRow` in other parts of the codebase.

gefjon:
Running benchmarks locally (hopefully less noisy?) to confirm, then will merge.

gefjon:
Pretty much the same story locally. Some improvements, some regressions, some noise, nothing groundbreaking.

cloutiertyler:
It's surprising that multi row unique raw mem insertions have no impact given what I measured.

PR #798, opened 2024-02-06
Fix private not building with latest master

bfops:
# Description of Changes
Exposes `MutTxId` more than crate-local, which was causing SpacetimeDBPrivate to fail.

This is an unbreaking change :upside_down_face: 


jdetter (approved):
LGTM

cloutiertyler (approved):

PR #801, opened 2024-02-07
Adding game loading benchmarks

mamcx:
# Description of Changes

Adding a test load that mimics patterns common in games we are interested in.

Also documentation on how to use https://github.com/mstange/samply/, which allows to see perf data on Firefox with integrated flamegraph.

1

### DIFF
--- a/crates/bench/benches/special.rs
+++ b/crates/bench/benches/special.rs
@@ -1,6 +1,7 @@
 use criterion::{criterion_group, criterion_main, Criterion};
 use mimalloc::MiMalloc;
 use spacetimedb::db::{Config, Storage};
+use spacetimedb_bench::game::query_header;
 use spacetimedb_bench::{
     schemas::{create_sequential, BenchTable, Location, Person, RandomTable},
     spacetime_module::BENCHMARKS_MODULE,
@@ -16,6 +17,7 @@ fn criterion_benchmark(c: &mut Criterion) {
     serialize_benchmarks::<Location>(c);
 
     custom_module_benchmarks(c);
+    game_benchmarks(c);
 }
 
 fn custom_module_benchmarks(c: &mut Criterion) {
@@ -46,6 +48,10 @@ fn custom_module_benchmarks(c: &mut Criterion) {
             )
         });
     }
+}
+
+fn game_benchmarks(c: &mut Criterion) {
+    c.bench_function("game/header", |b| b.iter(|| query_header().unwrap()));
 }
 
 fn serialize_benchmarks<T: BenchTable + RandomTable>(c: &mut Criterion) {
--- a/crates/bench/src/game.rs
+++ b/crates/bench/src/game.rs
@@ -1,0 +1,264 @@
+/// Load test to see large variations on performance
+///
+/// NOTE: It should be running with `--release` or `--profile bench` when looking with perf tools...
+use criterion::black_box;
+use itertools::Itertools;
+use std::time::{Duration, Instant};
+use tempdir::TempDir;
+
+use spacetimedb::db::relational_db::{open_db, RelationalDB};
+use spacetimedb::error::DBError;
+use spacetimedb::execution_context::ExecutionContext;
+use spacetimedb::sql::execute::run;
+use spacetimedb::subscription::subscription::create_table;
+use spacetimedb_lib::error::ResultTest;
+use spacetimedb_lib::identity::AuthCtx;
+use spacetimedb_lib::operator::OpCmp;
+use spacetimedb_primitives::TableId;
+use spacetimedb_sats::relation::FieldName;
+use spacetimedb_sats::{product, AlgebraicType, AlgebraicValue, ProductType, ProductTypeElement, ProductValue};
+use spacetimedb_vm::dsl::mem_table;
+use spacetimedb_vm::eval::run_ast;
+use spacetimedb_vm::expr::{Code, ColumnOp, QueryExpr};
+use spacetimedb_vm::program::Program;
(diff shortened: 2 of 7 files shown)


crates/bench/src/game.rs
```diff
@@ -0,0 +1,198 @@
...
+pub struct GameData {
+    location_state: u64,
+    footprint_tile_state: u64,
+}
+
+impl GameData {
+    /// Will generate on `release`:
+    /// Chunks * 100
+    /// FootprintTileState * 1'000
+    /// LocationState * 1'000.000 rows
```
kazimuth, line 35: Could we make this smaller? Million row benchmarks add to the ci run time a lot

crates/bench/src/game.rs
```diff
@@ -0,0 +1,198 @@
...
+
+pub fn game_insert(base_rows: usize) -> ResultTest<Duration> {
+    let tables = make_tables()?;
+
+    let data = GameData::new(base_rows as u64);
+
+    let start = Instant::now();
+    black_box(fill_tables(&tables, data)?);
+
+    let result = black_box(run(&tables.db, "SELECT * FROM LocationState", AuthCtx::for_testing())?);
```
kazimuth, line 156: Technically this is now timing more than just insertions

crates/bench/src/game.rs
```diff
@@ -0,0 +1,198 @@
...
+
+    let start = Instant::now();
+
+    let result = query(&tables, data)?;
+
+    #[cfg(debug_assertions)]
+    assert_eq!(result, EXPECT_ROWS / 100);
+
+    #[cfg(not(debug_assertions))]
+    assert_eq!(result, EXPECT_ROWS / 10);
```
kazimuth, line 177: Slightly confused about the math here, could we define these as constants next to EXPECT_ROWS?

kazimuth (commented):
Generally looks good, it's nice to have some join benchmarks.

I'm also planning on adding some basic sql benches once https://github.com/clockworklabs/SpacetimeDB/pull/803 is merged, but those won't have joins so this is super helpful.

crates/bench/src/game.rs
mamcx: Maybe we should not add this to criterion anyway...

joshua-spacetime: I agree with Mario. This benchmark is mainly for performance analysis, so we don't need it to be run in CI. At least not right now.

crates/bench/src/game.rs
```diff
@@ -0,0 +1,198 @@
...
+use spacetimedb::sql::execute::run;
+use spacetimedb::subscription::subscription::create_table;
+use spacetimedb_lib::error::ResultTest;
+use spacetimedb_lib::identity::AuthCtx;
+use spacetimedb_primitives::TableId;
+use spacetimedb_sats::{product, AlgebraicType};
+
+fn make_test_db() -> Result<(RelationalDB, TempDir), DBError> {
+    let tmp_dir = TempDir::new("stdb_test")?;
+    let stdb = open_db(&tmp_dir, true, false)?.with_row_count(Arc::new(|_, _| i64::MAX));
```
joshua-spacetime, line 21: We'll want to use the default row count function so that we produce accurate plans.

crates/bench/benches/special.rs
```diff
@@ -48,6 +50,18 @@ fn custom_module_benchmarks(c: &mut Criterion) {
     }
 }
 
+fn game_sql_benchmarks(c: &mut Criterion) {
```
joshua-spacetime, line 53: I think we should just add the utility for right now. We can decide later if we want this to be a CI benchmark.

crates/bench/src/game.rs
```diff
@@ -0,0 +1,198 @@
...
+            location_state: base_rows * 10,
+        }
+    }
+}
+
+#[cfg(not(debug_assertions))]
+pub const BASE_ROWS: usize = 1_000;
+#[cfg(debug_assertions)]
+pub const BASE_ROWS: usize = 10;
+const EXPECT_ROWS: usize = BASE_ROWS * 1_000;
```
joshua-spacetime, line 48: Lets keep this simple for right now and just have a single table size. Let's do 1M rows each for both tables.

crates/bench/src/game.rs
```diff
@@ -0,0 +1,198 @@
...
+        (0.into(), "entity_id_index"),
+        (1.into(), "chunk_index"),
+        (2.into(), "x_index"),
+    ];
+    let location_state = create_table(&db, &mut tx, "LocationState", schema, indexes)?;
+
+    // FootprintTileState
+    let schema = &[
+        ("entity_id", AlgebraicType::U64),
+        ("type", AlgebraicType::I32),
```
joshua-spacetime, line 80: Can we use the actual enum type here?

crates/bench/src/game.rs
joshua-spacetime: Yeah as long as the inserts succeed we shouldn't need this validation.

crates/bench/src/game.rs
joshua-spacetime, line 35: @mamcx could you make each of these tables with 1M rows. Also could you make most chunks with 1200 rows, and add at least one chunk with 12,000 rows.

crates/bench/src/game.rs
mamcx: Importing the proc macros fail (by recursion).

joshua-spacetime:
Closing as this was merged as part of #839

PR #803, opened 2024-02-07
Rename types in benchmarks

kazimuth:
# Description of Changes

- Rename tables in benchmarks to describe their contents
- Make parameter names uniform between callgrind and criterion
- Disable stdb_module benchmarks on CI (too noisy)

This wants to be paired with https://github.com/clockworklabs/benchmarks-viewer/pull/2, which reworks some things about benchmark reports.

I'm considering disabling the in memory benchmarks entirely on CI, but haven't done that yet. Changing the benchmark report format may make them easier to read, so that that isn't necessary.

1

### DIFF
--- a/modules/benchmarks/src/lib.rs
+++ b/modules/benchmarks/src/lib.rs
@@ -16,13 +16,13 @@
 //! `{operation}_{index_strategy}_{table_name}`, in snake_case.
 //!
 //! The three index strategies are:
-//! - `Unique` / `unique`: a single unique key, declared first in the struct.
-//! - `NonUnique` / `non_unique`: no indexes.
-//! - `MultiIndex` / `multi_index`: one index for each row.
+//! - `unique`: a single unique key, declared first in the struct.
+//! - `no_index`: no indexes.
+//! - `btree_each_column`: one index for each column.
 //!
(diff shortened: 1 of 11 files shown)


modules/benchmarks/src/lib.rs
```diff
@@ -16,29 +16,29 @@
 //! `{operation}_{index_strategy}_{table_name}`, in snake_case.
 //!
 //! The three index strategies are:
-//! - `Unique` / `unique`: a single unique key, declared first in the struct.
-//! - `NonUnique` / `non_unique`: no indexes.
-//! - `MultiIndex` / `multi_index`: one index for each row.
+//! - `unique` / `unique`: a single unique key, declared first in the struct.
+//! - `no_index` / `no_index`: no indexes.
+//! - `multi_index` / `multi_index`: one index for each row.
```
gefjon, line 21:
```suggestion
//! - `unique`: a single unique key, declared first in the struct.
//! - `no_index`: no indexes.
//! - `btree_each_column`: one index for each column.
```

gefjon (approved):

kazimuth:
Merging this, I've successfully pulled the data out and it works correctly.

PR #805, opened 2024-02-08
perf(query): Moves Code::Block instead of cloning during eval

joshua-spacetime:
For a table scan of 1M rows, this clone was over 30% of the total execution time.

Screenshot 2024-02-07 at 6 18 20 PM


mamcx (approved):
LGTM

PR #806, opened 2024-02-08
This fixes replaying of the transaction log to no longer check constraints

cloutiertyler:
# Description of Changes

Copied from Discord:

> Okay the reason that this wasn't a problem before was that we were not checking index constraints on replay before. This means that we were relying on the fact that the constraints would have been checked at or before commit time and that they have been baked into the message log.

> We actually **do not** want to check them at replay time because not only is it a pain, but actually **subtly wrong** the way we have it implemented. It's wrong because the actual constraints of the database may change as different transactions are added to the state and you would actually have to change your indexes and constraints as you replayed the log. This we are not currently doing (we're building all the non-bootstrapped indexes at the end after replaying), and thus aren't implementing constraint checking correctly as it stands.

> However, the above is all rendered moot anyway because we don't need to check constraints while replaying if we just assume that they were all checked prior to the transaction committing in the first place.

> Note also that operation/mutation ordering **does not** matter for operations inside a transaction of the message log assuming we only ever insert **OR** delete a unique row in one transaction. If we ever insert **AND** delete then order does matter. The issue above does not imply that order matters. Ordering of operations would **only** matter if you wanted to view the state of the database as of a partially applied transaction. We never actually want to do this, because after a transaction has been committed, it is assumed that all operations happen instantaneously and atomically at the timestamp of the transaction. The only time that we actually want to view the state of a database while a transaction is partially applied is while the transaction is running **before** it commits. Thus, we only care about operation ordering while the transaction is running, but we do not care about it at all in the context of the commit log.

> Not caring about the order in the log, however, requires that we do not check index constraints during replay of transaction operatoins. We could check them in between transactions if we wanted to update the indexes and constraints as they changed during replay, but that is unnecessary. 

> It seems that prior to mem-arch at least we **DO** delete and insert the same row within a transaction. This is always either a noop `(insert -> delete) or (insert) -> (delete -> insert)` or an insert `(doesn't exist) -> (delete -> insert)` depending on the order in the transaction and if it had been inserted before.

> Since this could differ the order does matter unless we avoid writing delete && insert in the same transaction. I am going to assume that the ordering in the message log currently **does mean something** for the old implemenation and that the new implementation does **NOT** ever insert an insert && delete into the message log.

None

4


crates/core/src/db/datastore/locking_tx_datastore/datastore.rs
```diff
@@ -154,9 +194,14 @@ impl Locking {
                 Operation::Insert => {
                     let (table, blob_store) =
                         committed_state.get_table_and_blob_store_or_create_ref_schema(table_id, &schema);
-                    table.insert(blob_store, &row).unwrap_or_else(|e| {
-                        panic!("Failed to insert during transaction playback: {:?}", e);
-                    });
+                    table
+                        .insert_internal_allow_duplicate(blob_store, &row)
```
kim, line 198: Is this indeed enough? This method doesn't update the pointer map, which seems important.

crates/core/src/db/datastore/locking_tx_datastore/committed_state.rs
```diff
@@ -267,10 +268,12 @@ impl CommittedState {
...
         let blob_store = &mut self.blob_store;
-        table
-            .delete_equal_row(blob_store, rel)
-            .map_err(TableError::Insert)?
-            .ok_or_else(|| anyhow!("Delete for non-existent row when replaying transaction"))?;
+
+        // NOTE: We are eating the error here because during replay, it is assumed that
+        // the order of operations within a transaction mattered prior to mem-arch
+        // and that deleting and inserting the same row cannot occur in the commitlog
+        // after mem-arch.
```
kim, line 275:
```suggestion
        // the order of operations within a transaction does not matter
        // and that deleting and inserting the same row cannot occur in the commitlog.
```

It's not clear to me that or how this changed, namely the [TxState invariants](https://github.com/clockworklabs/SpacetimeDB/blob/master/crates/core/src/db/datastore/locking_tx_datastore/tx_state.rs#L36-L40). 

In any case, for a future reader it likely only matters that the order does not matter.

gefjon (changes requested):
I can fix these easily, so I'll open a PR based on this with an added commit.

gefjon (approved):

kim (approved):

PR #807, opened 2024-02-08
To make eval subscription concurrent

Shubham8287:
# Description of Changes
Moves RwLock from `SubscriptionManager` to its specific field `subscriptions` so that most of the heavy lifting part i.e `eval` of `add_susbcription` can be done concurrently.

3, It has potential to introduce concurrency bugs.

### DIFF
--- a/crates/core/src/subscription/module_subscription_actor.rs
+++ b/crates/core/src/subscription/module_subscription_actor.rs
@@ -20,10 +20,11 @@ use rayon::iter::{IntoParallelRefIterator, ParallelIterator};
 use spacetimedb_lib::identity::AuthCtx;
 use spacetimedb_lib::Identity;
 
+type Subscriptions = Arc<RwLock<Vec<Subscription>>>;
 #[derive(Debug)]
 pub struct ModuleSubscriptions {
     relational_db: Arc<RelationalDB>,
-    subscriptions: Arc<RwLock<Vec<Subscription>>>,
+    pub subscriptions: Subscriptions,
     owner_identity: Identity,
 }
 
@@ -38,8 +39,6 @@ impl ModuleSubscriptions {
 
     /// Add a subscriber to the module. NOTE: this function is blocking.
     pub fn add_subscriber(&self, sender: ClientConnectionSender, subscription: Subscribe) -> Result<(), DBError> {
-        self.remove_subscriber(sender.id);
-
         let tx = &mut *scopeguard::guard(self.relational_db.begin_tx(), |tx| {
             let ctx = ExecutionContext::subscribe(self.relational_db.address());
             self.relational_db.release_tx(&ctx, tx);
@@ -52,7 +51,12 @@ impl ModuleSubscriptions {
             queries.extend(qset);
         }
 
+        let database_update = tokio::task::block_in_place(|| queries.eval(&self.relational_db, tx, auth))?;
+        // It acquires the subscription lock after `eval`, allowing `add_subscription` to run concurrently.
+        // This also makes it possible for `broadcast_event` to get scheduled before the subsequent part here
+        // but that should not pose an issue.
         let mut subscriptions = self.subscriptions.write();
+        self._remove_subscriber(sender.id, &mut subscriptions);
         let subscription = match subscriptions.iter_mut().find(|s| s.queries == queries) {
             Some(sub) => {
                 sub.add_subscriber(sender);
@@ -73,7 +77,6 @@ impl ModuleSubscriptions {
             .initial_subscription_evals
             .with_label_values(&self.relational_db.address())
             .inc();
-        let database_update = tokio::task::block_in_place(|| subscription.queries.eval(&self.relational_db, tx, auth))?;
 
         let sender = subscription.subscribers().last().unwrap();
         // NOTE: It is important to send the state in this thread because if you spawn a new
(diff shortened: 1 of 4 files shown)


crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -51,38 +57,50 @@ impl ModuleSubscriptions {
...
-
-        let sender = subscription.subscribers().last().unwrap();
-
+        let sender = subscription.subscribers().last().unwrap().clone();
+        drop(subscriptions);
         // NOTE: It is important to send the state in this thread because if you spawn a new
         // thread it's possible for messages to get sent to the client out of order. If you do
         // spawn in another thread messages will need to be buffered until the state is sent out
         // on the wire
         let fut = sender.send_message(SubscriptionUpdateMessage { database_update });
```
joshua-spacetime, line 92: Correct me if I'm wrong, but technically a race condition is possible between 87 and 92.

Shubham8287: Actually yes, If same client subscribes frequently. Extended the lock to end of this function to fix.

crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -51,38 +57,49 @@ impl ModuleSubscriptions {
...
-        let subscription = match self.subscriptions.iter_mut().find(|s| s.queries == queries) {
+        let database_update = tokio::task::block_in_place(|| queries.eval(&self.relational_db, tx, auth))?;
+        Ok((queries, database_update))
+    }
+
+    /// Add a subscriber to the module. NOTE: this function is blocking.
+    pub fn add_subscriber(&self, sender: ClientConnectionSender, subscription: Subscribe) -> Result<(), DBError> {
+        let (queries, database_update) = self.db_updates(&sender, subscription)?;
+        // holds lock after `eval` so that it can be concurrent
+        let mut subscriptions = self.subscriptions.write();
```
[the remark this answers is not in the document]
Shubham8287: Ah! thanks alot for pointing this. Its basically the same problem as there was in reducer + eval_incr (releasing a tx lock before acquiring subscripton lock).
Sorry stupidly did it in last min refactoring.

crates/core/src/subscription/module_subscription_actor.rs
Shubham8287: Or no, If same client subscribes frequently i.e being serialized by this await - https://github.com/clockworklabs/SpacetimeDB/pull/807/files#diff-cde382f76a7190bac09b3ca01dc55e56abb79225fb34ec657d6e43ba7792da0eR172.

and sending updates together to different clients should not be a problem.

I feel we can do the `drop(subscriptions)` before blocking on send_message. (Being restrictive right now, and not dropping the lock early, until I am sure about this).

@coolreader18 , Would love to get your views on this.

joshua-spacetime (approved):
LGTM!

crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -51,38 +51,43 @@ impl ModuleSubscriptions {
             queries.extend(qset);
         }
 
-        let subscription = match self.subscriptions.iter_mut().find(|s| s.queries == queries) {
+        let database_update = tokio::task::block_in_place(|| queries.eval(&self.relational_db, tx, auth))?;
+        // holds lock after `eval` so that it can be concurrent
```
coolreader18, line 55: A longer comment here might be nice - I had to think about the correctness of this ordering for a few minutes, so laying it out explicitly would be helpful. We hold a read lock on the db via the tx, the module host needs a write tx to execute, etc, therefore this `queries.eval()` can't happen between a reducer being committed and it doing `broadcast_event()`. ~~I think this even holds for mvcc in the future, `begin_tx()` wouldn't lock but it would stick to the one commit it was started from; committing would put a commit on top of it but not change our tx once we've started it.~~ Oh. Wait no, this doesn't even work now, we need the secondary lock here. If `add_subscriber()` is called between commit and broadcast_event, the tx can contain the committed state, and then the broadcast_event would send a message with it and so would add_subscriber. I might push a commit to this branch with what I'm thinking, if that's ok?

coolreader18 (changes requested):

crates/core/src/subscription/module_subscription_actor.rs
Shubham8287: I am not sure If we can call it bug entirely, Client can receive one extra update of old subscription queuries after sending new subscription request because add_subscription has not processed yet, then receving a same data via add_subscriber must be expected.
but definitely go ahead if you can improve it.

Shubham8287: Or moving remove_subscription before eval will also solve it, it's just we have to acquire lock twice.

coolreader18: I thought that was a bug? Wasn't that the whole reason for #792?

coolreader18:
@Shubham8287 pushed; what do you think?

Shubham8287:
> @Shubham8287 pushed; what do you think?

What do you think about just doing `remove_subscription` in starting of `add_subscription` as we are doing in master, I think that will remove the need of this extra lock and still keep `add_subscription` to wait until broadcast finish.

crates/core/src/subscription/module_subscription_actor.rs
Shubham8287: > I thought that was a bug? Wasn't that the whole reason for #792?

I think main reason was other way around, i.e to receive duplicate state from broadcast after already having it from add_subscription is indeed a bug, Not really sure if vice-versa would also be a bug.

coolreader18: Oh, huh. I don't know if I realized that, but that makes sense. since `add_subscriber()` is gonna return a non-incremental update anyway, it doesn't matter what gets sent before. Alright.

coolreader18:
Okay, I think I see how this works. Feel free to reset my commit. (even without remove_subscription before evaluating the query, it should be fine). Though, a comment with all that would be very appreciated.

coolreader18 (approved):

PR #808, opened 2024-02-08
Fix replay: Phoebe's version

gefjon:
# Description of Changes

Modifications to #806 based on comments I made there.

- Inserts while replaying must still update the pointer map, maintain set semantics and
  insert into indexes, they just get to skip unique constraints.
- We should not swallow errors when replaying a delete.

# Testing

- Built SpacetimeDB v0.8.1 (pre-mem-arch), published BitCraft, uploaded world.
- Killed SpacetimeDB.
- Built SpacetimeDB from this branch, restarted BitCraft, signed in, walked around.
- Success!

None

4 - replay is, as evidenced by this PR's necessity, fiddly and complex.

### DIFF
--- a/crates/core/src/db/datastore/locking_tx_datastore/committed_state.rs
+++ b/crates/core/src/db/datastore/locking_tx_datastore/committed_state.rs
@@ -46,6 +46,9 @@ use std::{
 
 #[derive(Default)]
 pub(crate) struct CommittedState {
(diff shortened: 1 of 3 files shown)


gefjon:
Closing in favor of Tyler's PR, since we've determined that we need a middle ground between the two.

PR #810, opened 2024-02-08
Reduce allocations for prometheus with_label_values

coolreader18:
# Description of Changes

Based on #811 for RPITIT (return-position impl trait in trait)

1

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4676,6 +4676,7 @@ dependencies = [
 name = "spacetimedb-metrics"
 version = "0.8.1"
 dependencies = [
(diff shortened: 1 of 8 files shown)


coolreader18:
Nevermind I'll split it.

cloutiertyler:
benchmarks please

Shubham8287 (approved):
Reviewed for the following  - 
- Refactoring to adopt RPIT for AsRef<str>. (also my TIL)
- Perf optimisation by stopping allocation String for ints.

Looks Good!

PR #811, opened 2024-02-08
Bump Rust to 1.76.0

coolreader18:
# Description of Changes

Needed at least 1.75 for RPITIT (return-position impl trait in trait) for #810 and and since 1.76 just came out I figured may as well just skip to that. Might take a few hours before the 1.76 docker image is up though.

1


crates/cli/src/subcommands/generate/rust.rs
```diff
@@ -181,6 +180,7 @@ fn print_spacetimedb_imports(output: &mut Indenter) {
 
 fn print_file_header(output: &mut Indenter) {
     print_auto_generated_file_comment(output);
+    print_lines(output, &["#![allow(unused_imports)]"]);
```
coolreader18, line 183: This is because rustc now complains about the unused `pub use foo_bar_baz_reducer::*;` glob import.

crates/core/src/db/messages/transaction.rs
```diff
@@ -23,6 +23,7 @@ pub struct Transaction {
     pub writes: Vec<Write>,
 }
 
+#[allow(clippy::items_after_test_module)]
```
kim, line 26: :facepalm:

kim (approved):

crates/core/src/db/messages/transaction.rs
coolreader18: Opened rust-lang/rust-clippy#12247 bc of this

Issue #799, opened 2024-02-07
Rustup not found causes publish to fail


jdetter:
If you are trying to use `spacetime publish` or `spacetime build` and `rustup` isn't present on your system (e.g. you've installed cargo + rustc through your package manager) then it will fail with an error message:

![image](https://github.com/clockworklabs/SpacetimeDB/assets/4099508/3aadc0fa-7e29-40ab-98ab-1008335b0b66)

PR #814, opened 2024-02-09
Fix: Publish fails if rustup not present

bfops:
# Description of Changes
Allow users to run `spacetime publish` even if `rustup` is not present on their system. Typically we use it to install the `wasm32-unknown-unknown` target.

Addresses https://github.com/clockworklabs/SpacetimeDB/issues/799

These are CLI-only changes that don't affect our ABI.

1

### DIFF
--- a/crates/cli/src/tasks/rust.rs
+++ b/crates/cli/src/tasks/rust.rs
@@ -21,14 +21,11 @@ fn cargo_cmd(subcommand: &str, build_debug: bool, args: &[&str]) -> duct::Expres
 
 pub(crate) fn build_rust(project_path: &Path, skip_clippy: bool, build_debug: bool) -> anyhow::Result<PathBuf> {
     // Make sure that we have the wasm target installed (ok to run if its already installed)
-    match cmd!("rustup", "target", "add", "wasm32-unknown-unknown").run() {
-        Ok(_) => {}
-        Err(err) => {
-            println!(
-                "Warning: Failed to install wasm32-unknown-unknown target: {}. Is `rustup` installed?",
-                err
-            );
-        }
+    if let Err(err) = cmd!("rustup", "target", "add", "wasm32-unknown-unknown").run() {
+        println!(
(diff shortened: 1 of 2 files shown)


bfops:
Fixes https://github.com/clockworklabs/SpacetimeDB/issues/799

crates/cli/src/tasks/rust.rs
```diff
@@ -21,7 +21,12 @@ fn cargo_cmd(subcommand: &str, build_debug: bool, args: &[&str]) -> duct::Expres
 
 pub(crate) fn build_rust(project_path: &Path, skip_clippy: bool, build_debug: bool) -> anyhow::Result<PathBuf> {
     // Make sure that we have the wasm target installed (ok to run if its already installed)
-    cmd!("rustup", "target", "add", "wasm32-unknown-unknown").run()?;
+    match cmd!("rustup", "target", "add", "wasm32-unknown-unknown").run() {
+        Ok(_) => {}
+        Err(err) => {
+            println!("Warning: Failed to install wasm32-unknown-unknown target: {}. Is `rustup` installed?", err);
+        }
+    }
```
jdetter, line 32: You can simplify using something like this:

```rust
if let Err(err) = cmd!("rustup", "target", "add", "wasm32-unknown-unknown").run() {
    println!("Warning: Failed to install wasm32-unknown-unknown target: {}. Is `rustup` installed?", err);
}

```

jdetter (approved):
Small nit, but looks good to me 👍

bfops:
Oops I forgot to include this, but my testing was:

```bash
$ mv ~/.cargo/bin/rustup{,-backup}
$ cd $WORK/my_test_module
$ cargo run publish -c -S publish-test
Warning: Failed to install wasm32-unknown-unknown target: No such file or directory (os error 2). Is `rustup` installed?
    Finished release [optimized] target(s) in 0.05s
...
```

and then afterward to restore `rustup`:
```bash
mv ~/.cargo/bin/rustup{-backup,}
```

PR #826, opened 2024-02-12
Update upgrade-version tool

coolreader18:
# Description of Changes

Fixing after #802. Please let me know if there's anything you'd like me to change about this; I didn't wanna try extending the custom toml parsing going on so I just pulled in `toml_edit`; it does the right thing and the code is clearer. (also made `process_license_file` use regex replace_all, cause while testing the file renaming was failing for me. lemme know if I should revert that). I also, instead of doing the walkdir for `Cargo._toml`, just specify `crates/cli/src/subcommands/project/rust/Cargo._toml`, since that's gonna be the only ever one afaik.

### DIFF
--- a/tools/upgrade-version/src/main.rs
+++ b/tools/upgrade-version/src/main.rs
@@ -49,7 +49,7 @@ fn main() -> anyhow::Result<()> {
 
     let current_dir = env::current_dir().expect("No current directory!");
     let dir_name = current_dir.file_name().expect("No current directory!");
-    if dir_name != "SpacetimeDB" {
+    if dir_name != "SpacetimeDB" && dir_name != "public" {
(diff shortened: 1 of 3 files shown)


jdetter (approved):
Tested via: https://github.com/clockworklabs/SpacetimeDB/actions/runs/7879171217/job/21498928395?pr=825

tools/upgrade-version/src/main.rs
```diff
@@ -145,21 +50,29 @@ fn main() {
     let current_dir = env::current_dir().expect("No current directory!");
     let dir_name = current_dir.file_name().expect("No current directory!");
     if dir_name != "SpacetimeDB" {
```
jdetter, line 52: Before merging: please change this to `dir_name != "SpacetimeDB" && dir_name != "public"`

bfops (approved):
Works for me!

Issue #831, opened 2024-02-13
Subscriptions should not compute primary keys


joshua-spacetime:
The new datastore no longer stores a `DataKey` with each row. This means that the `id` field of a `RelValue` is always `None`, which in turn means that we will compute a `PrimaryKey` for each row returned in a subscription. This dominates the cost of subscription evaluation.

The following graphs were generated from a table scan of 1M fixed length rows. Here `PrimaryKey` represents ~50% of the total execution time. And the relative cost only grows as more rows are added to the result set.

![Image](https://github.com/clockworklabs/SpacetimeDB/assets/141075564/b6af7c8c-b64f-4218-ba92-2a44212cdc45)

![Image](https://github.com/clockworklabs/SpacetimeDB/assets/141075564/de239446-6929-4c6a-930f-74945a67043b)

Clients use the `PrimaryKey` to delete rows from their local cache. But in most cases hashing the row itself will be no more expensive than hashing the `PrimaryKey`.

PR #840, opened 2024-02-13
Remove `row_pk` from client API; hash rows not row_ids on client

gefjon:
# Description of Changes

This commit removes the `row_pk` field from `message TableRowOperation` in the client API. This is the beginning of addressing issue #831.

This has several implications for clients:
- Because we no longer have a stable content-addressed ID for each row, the client is responsible for hashing rows itself.
- In order to avoid requiring that generated types be `Hash + Eq`, we hash the BSATN-serialized bytes, not the deserialized domain type.
  - Requiring rows be `Hash + Eq` is undesirable because it would mean that generated types must use `sats::F32` or `sats::F64` rather than `f32` or `f64`, as the former are `Eq + Hash` while the latter are not. (Aside: it's stupid that Rust floats are not `Eq + Ord + Hash`, since IEEE-754 defines a total equality and total ordering on floats.)
- This means that, for large row types, the client cache will now be on the order of 2x as large. This seems like an acceptable trade-off.
  - Note that we have not introduced any new allocations in the Rust SDK, as the ProtoBuf messages already contain the `row_bytes: Vec<u8>` which we use as the keys in the client-cache `HashMap`.

This commit also makes some minor formatting changes to Rust codegen, since it was emitting code that rustc warned about.

This PR also removes the `row_pk` field from the host's definitions of the JSON WebSocket API. No client included in this repo uses that API, so the change is currently untested. Notably, this PR *does not* (yet?) remove places in the subscription engine where the `row_pk` is computed; it merely disables sending those pks to the client.

Still outstanding:

- [ ] Avoid computing row_pk in the subscription engine - will be a follow-up PR.
- [x] Update other SDKs.
  - [x] C# - https://github.com/clockworklabs/spacetimedb-csharp-sdk/pull/72
  - [x] TypeScript - https://github.com/clockworklabs/spacetimedb-typescript-sdk/pull/34
  - [ ] Python - tentatively assigned to @coolreader18 ; may be dropped.
- [ ] Update the website to use the altered JSON API.
  - Unclear if this is necessary; we suspect the website never touches the `row_pk`.
- [x] Update docs to reflect the change - https://github.com/clockworklabs/spacetime-docs/pull/29

Unsure whether I have the correct labels. This PR breaks:

- The client API, so SDKs will need to be rebuilt.
- Rust SDK codegen for types and tables which store floats, so client authors will need to sprinkle `.into()` calls throughout their float-ey code.

4 - breaks client API, so all SDKs will need to change; Cloud controldb may need fixing if it uses floats.


NateTheDev1:
Commenting for tracking

PR #847, opened 2024-02-14
Refactor some ReadColumn stuff + relational_db tests

Centril:
# Description of Changes

1. Add a method `read_col` to `RowRef` and use it.
2. Simplify `InvalidFieldError` creation.
3. Simplify and dedup various relational_db tests, including using `ReadColumn`.

Best reviewed by hiding whitespace changes.

# API and ABI breaking changes

None

# Expected complexity level and risk

1

PR #851, opened 2024-02-15
system_tables, mut_tx, and friends: bye bye `to_product_value`

Centril:
# Description of Changes

Get rid of a bunch of `to_product_value` calls.
Some associated cleanup and refactoring included.

Based on https://github.com/clockworklabs/SpacetimeDB/pull/847.

None

3, touches system table code, so make sure you check no semantics are changed.

### DIFF
--- a/crates/core/src/db/datastore/system_tables.rs
+++ b/crates/core/src/db/datastore/system_tables.rs
@@ -88,6 +88,12 @@ macro_rules! st_fields_enum {
             }
         }
 
+        impl From<$ty_name> for ColId {
+            fn from(value: $ty_name) -> Self {
+                value.col_id()
+            }
+        }
+
         impl From<$ty_name> for ColList {
             fn from(value: $ty_name) -> Self {
                 ColList::new(value.col_id())
@@ -313,10 +319,8 @@ impl TryFrom<RowRef<'_>> for StTableRow<String> {
     type Error = DBError;
     // TODO(cloutiertyler): Noa, can we just decorate `StTableRow` with Deserialize or something instead?
     fn try_from(row: RowRef<'_>) -> Result<StTableRow<String>, DBError> {
-        let table_id = row.read_col::<u32>(StTableFields::TableId.col_id())?.into();
-        let table_name = row.read_col(StTableFields::TableName.col_id())?;
         let table_type = row
-            .read_col::<String>(StTableFields::TableType.col_id())?
+            .read_col::<String>(StTableFields::TableType)?
             .deref()
             .try_into()
             .map_err(|x: &str| TableError::DecodeField {
@@ -327,7 +331,7 @@ impl TryFrom<RowRef<'_>> for StTableRow<String> {
             })?;
 
         let table_access = row
-            .read_col::<String>(StTableFields::TablesAccess.col_id())?
+            .read_col::<String>(StTableFields::TablesAccess)?
             .deref()
             .try_into()
             .map_err(|x: &str| TableError::DecodeField {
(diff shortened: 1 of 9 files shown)


coolreader18 (approved):
LGTM!

cloutiertyler:
benchmarks please

crates/core/src/db/datastore/system_tables.rs
```diff
@@ -307,14 +315,13 @@ pub struct StTableRow<Name: AsRef<str>> {
     pub(crate) table_access: StAccess,
 }
 
-impl<'a> TryFrom<&'a ProductValue> for StTableRow<&'a str> {
+impl TryFrom<RowRef<'_>> for StTableRow<String> {
     type Error = DBError;
     // TODO(cloutiertyler): Noa, can we just decorate `StTableRow` with Deserialize or something instead?
```
cloutiertyler, line 320: Would it be reasonable at some point to implement `SpacetimeType` for `StTableRow`?

Also I notice that we are moving to `String` from `&str`. I assume we always have to do that because now that we're going directly from bflatn there's otherwise no `&str` to return?

I would like to point out that once we implement the `iter_page` wasm abi call/syscall we will need a generalized way to create zero-copy Rust structs which point to page data. Something like:

```rust
#[repr(C)]
struct StTableRow {
    table_idx: u32,
    table_name: SatsZeroCopyStringPtr,
    ...
}
```

This would be as useful in this case as it would be for the code generated types in the module, so for example if you had this table:

```rust
#[spacetimedb(table)]
struct Person {
    age: u32,
    name: String
}
```

then we would also code generate a zero copy repr(C) "pointer" struct which could just be overlayed on top of the page. Something like: 

```rust
mod zero_copy {
    #[repr(C)]
    struct Person {
        age: u32,
        name: SatsZeroCopyStringPtr
    }
}
```

crates/core/src/db/datastore/system_tables.rs
```diff
@@ -372,31 +369,19 @@ pub struct StColumnRow<Name: AsRef<str>> {
...
-impl<'a> TryFrom<&'a ProductValue> for StColumnRow<&'a str> {
+impl TryFrom<RowRef<'_>> for StColumnRow<String> {
     type Error = DBError;
-    fn try_from(row: &'a ProductValue) -> Result<StColumnRow<&'a str>, DBError> {
-        let table_id: TableId = row.field_as_u32(StColumnFields::TableId.col_idx(), None)?.into();
-        let col_pos = row.field_as_u32(StColumnFields::ColPos.col_idx(), None)?.into();
-        let col_name = row.field_as_str(StColumnFields::ColName.col_idx(), None)?;
-        let bytes = row.field_as_bytes(StColumnFields::ColType.col_idx(), None)?;
+    fn try_from(row: RowRef<'_>) -> Result<StColumnRow<String>, DBError> {
+        let table_id = row.read_col(StColumnFields::TableId)?;
```
cloutiertyler, line 375: How does this know the type of the `TableId` field?

crates/core/src/db/datastore/traits.rs
```diff
@@ -45,16 +44,11 @@ pub trait Data: Into<ProductValue> {
 pub trait DataRow: Send + Sync {
     type RowId: Copy;
 
-    type DataRef<'a>;
```
cloutiertyler, line 48: Might this have a lot of conflicts with other PRs?

cloutiertyler: Perhaps this could be split out into a separate PR?

crates/table/src/read_column.rs
```diff
@@ -314,6 +314,31 @@ impl_read_column_via_av! {
     AlgebraicTypeLayout::Product(_) => into_product => ProductValue;
 }
 
+macro_rules! impl_read_column_via_from {
```
cloutiertyler, line 317: Devilry

cloutiertyler (commented):
Comments inline.

kazimuth (approved):
I have no particular complaints. It's neat to do this with the system tables, even the metadata reads are allocating less now.

PR #853, opened 2024-02-15
Fix publish issues 0.8.2

jdetter:
# Description of Changes

- Fixed any issues which prevented publishing release `v0.8.2-beta-hotfix7`

corresponding GitHub label.

### DIFF
--- a/crates/lib/Cargo.toml
+++ b/crates/lib/Cargo.toml
@@ -37,7 +37,7 @@ hex.workspace = true
 itertools.workspace = true
 once_cell.workspace = true
 prometheus.workspace = true
-serde = { workspace = true, optional = true }
+serde = { workspace = true, optional = true, features = ["derive"] }
 serde_with = {workspace = true, optional = true }
 thiserror.workspace = true
 tracing.workspace = true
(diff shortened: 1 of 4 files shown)


crates/lib/Cargo.toml
```diff
@@ -37,7 +37,7 @@ hex.workspace = true
 itertools.workspace = true
 once_cell.workspace = true
 prometheus.workspace = true
-serde = { workspace = true, optional = true }
```
jdetter, line 40: this was complaining about not being able to find serde Serialize/Deserialize

crates/table/Cargo.toml
```diff
@@ -29,8 +29,8 @@ proptest = ["dep:proptest", "dep:proptest-derive"]
 blake3_pure = ["blake3/pure"]
 
 [dependencies]
-spacetimedb-primitives = { path = "../primitives" }
-spacetimedb-sats = { path = "../sats" }
```
jdetter, line 33: I updated these to use the new workspace thing because cargo was complaining that these didn't specify version numbers

coolreader18 (approved):

PR #854, opened 2024-02-15
Fix deadlock

coolreader18:
# Description of Changes

Call order that would lead to deadlock:
```
module host: read-lock subscriptions
module host: commit tx, release lock on db
add_subscriber: take read lock on db
module host: call DatabaseUpdate::from_writes. tries to take write lock on db. cannot, add_subscriber has a read lock
add_subscriber: tries to take write lock on subscriptions. cannot, module host still holds it
deadlock
```

This fixes by making from_writes not take a write-lock on the db

### DIFF
--- a/crates/core/src/db/datastore/locking_tx_datastore/datastore.rs
+++ b/crates/core/src/db/datastore/locking_tx_datastore/datastore.rs
@@ -282,6 +282,10 @@ impl TxDatastore for Locking {
         tx.table_id_from_name(table_name, self.database_address)
     }
 
+    fn table_name_from_id_tx<'a>(&'a self, tx: &'a Self::Tx, table_id: TableId) -> Result<Option<Cow<'a, str>>> {
+        Ok(tx.table_exists(&table_id).map(Cow::Borrowed))
+    }
+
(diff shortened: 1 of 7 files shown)


crates/core/src/db/datastore/locking_tx_datastore/datastore.rs
```diff
@@ -282,6 +282,10 @@ impl TxDatastore for Locking {
         tx.table_id_from_name(table_name, self.database_address)
     }
 
+    fn table_name_from_id_tx<'a>(&'a self, tx: &'a Self::Tx, table_id: TableId) -> Result<Option<Cow<'a, str>>> {
+        Ok(tx.table_exists(&table_id).map(Cow::Borrowed))
```
coolreader18, line 286: Not sure if this is okay - the mut version does a db scan but this seems simpler but maybe it can be out of sync with the system table? idk, but this is fine for now.

jdetter (approved):
I've tested this and it does fix the deadlock, there does seem to be a correctness issue present in master but I think it probably was not introduced by this PR. On that basis I'm fine to approve 👍

PR #855, opened 2024-02-15
Fix bench errors and include in CI

coolreader18:
# Description of Changes

This doesn't affect much in practice, but rust-analyzer checks all crate targets by default and it's annoying to see errors in the status bar from something unrelated to what I'm working on.

1


kazimuth (approved):
LGTM.

PR #859, opened 2024-02-16
Add `delete_by_{nonunique_column}` method to `#[spacetimedb(table)]`

gefjon:
# Description of Changes

- Refactor `bindings::query::delete_by_field` to require `FilterableValue` and return `u32` count of deleted rows.
- Add `bindings::query::delete_by_unique_field` with the previous behavior of `delete_by_field`, i.e. requiring `UniqueValue` and returning `bool` with `true = deleted`. This is implemented in terms of `delete_by_field`.
- Macro-generated unique filter funcs call `delete_by_unique_field` rather than `delete_by_field`.
- The table macro now generates non-unique delete functions alongside the existing non-unique filter functions for non-unique fields with primitive types.
- Add uses of `::delete_by_{field}` for both unique and non-unique columns to the `rust-wasm-test` module, so we can see that they compile as expected.

Note that this PR does not include any host-side changes; the diff is localized to the `bindings` and `bindings-macro` crates.

Something to think about: it's possible we should not be emitting `filter_by_{field}` or `delete_by_{field}` methods for columns/fields which do not have an index, on the grounds that table scans for filtering are very slow and bad, and a performing such a filter is almost certainly a bug in the module. This PR does not change our existing behavior, which is to generate the methods for any field/column with a filterable type, even if unindexed.

N/A - `bindings::query` is for internal use only and is marked `#[doc(hidden)]`.

1.

### DIFF
--- a/crates/bindings-macro/src/lib.rs
+++ b/crates/bindings-macro/src/lib.rs
@@ -633,7 +633,7 @@ fn spacetimedb_tabletype_impl(item: syn::DeriveInput) -> syn::Result<TokenStream
 
         unique_delete_funcs.push(quote! {
             #vis fn #delete_func_ident(#column_ident: &#column_type) -> bool {
-                spacetimedb::query::delete_by_field::<Self, #column_type, #column_index>(#column_ident)
+                spacetimedb::query::delete_by_unique_field::<Self, #column_type, #column_index>(#column_ident)
             }
         });
     }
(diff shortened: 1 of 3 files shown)


kazimuth (approved):
This looks useful, I was looking for something like this a while back. 

It would also be useful to be able to delete by multiple non-unique fields, but then the question is how to codegen that in a sane way.

gefjon:
> It would also be useful to be able to delete by multiple non-unique fields, but then the question is how to codegen that in a sane way.

I think for this we'd want a `delete!` macro similar to the `query!` macro, rather than generating a method for every combination of fields.

PR #860, opened 2024-02-16
Replace DbProgram with a leaner version that compile directly to queries without environment

### LINKED ISSUE #815: Share DbProgram across queries
Currently we instantiate a `DbProgram` once per query execution. A `DbProgram` though does not hold any query specific state, so it should be possible to have a single global instantiation that's tied to the lifetime of `RelationalDB`. This should improve incremental join queries by ~15% by removing both construction and destruction from the hot path of query execution.

Image

Image

mamcx:
# Description of Changes

Experimental branch that compiles queries without running into a full interpreter (VM)

To be determined by more testing...

5 complexity 2 risk: AFAIK there are no paths that exercise the VM outside the internals that are removed here, but that is proving a negative!

```bash
     Running benches/subscription.rs

Benchmarking full-scan: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 53.5s, or reduce sample count to 10.
full-scan               time:   [523.87 ms 526.78 ms 530.67 ms]
                        change: [-4.1506% -2.1412% -0.6346%] (p = 0.01 < 0.05)
  2 (2.00%) high mild
  4 (4.00%) high severe

Benchmarking full-join: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 5.9s, enable flat sampling, or reduce sample count to 60.
full-join               time:   [1.1656 ms 1.1664 ms 1.1675 ms]
                        change: [-0.5557% -0.0309% +0.3747%] (p = 0.91 > 0.05)
                        No change in performance detected.
Found 11 outliers among 100 measurements (11.00%)
  2 (2.00%) high mild
  9 (9.00%) high severe

incr-select             time:   [8.9118 µs 8.9339 µs 8.9598 µs]
                        change: [-32.861% -32.358% -31.989%] (p = 0.00 < 0.05)
                        Performance has improved.

incr-join               time:   [55.282 µs 55.398 µs 55.531 µs]
                        change: [-16.661% -16.437% -16.236%] (p = 0.00 < 0.05)
                        Performance has improved.
```


crates/core/src/vm.rs
```diff
@@ -441,11 +429,11 @@ impl ProgramVm for DbProgram<'_, '_> {
...
 
     fn env(&self) -> &EnvDb {
-        &self.env
+        unreachable!()
     }
 
     fn env_mut(&mut self) -> &mut EnvDb {
-        &mut self.env
+        unreachable!()
     }
```
joshua-spacetime, line 437: Let's remove these from the trait entirely right now, since `DbProgram` no longer has an `EnvDb`.

crates/vm/src/eval.rs
```diff
@@ -44,6 +44,55 @@ fn build_query_opt(q: QueryExpr) -> QueryExprOpt {
     QueryExprOpt { source, query: q.query }
 }
 
+pub(crate) fn optimize_query(q: CrudExpr) -> ExprOpt {
```
joshua-spacetime, line 47: Can you make this patch, just about removing `EnvDb`? So essentially this patch should just have code removed, but not added - i.e. all red, no green.

crates/core/src/vm.rs
```diff
@@ -532,11 +511,7 @@ impl ProgramVm for DbProgram<'_, '_> {
...
 
     fn as_program_ref(&self) -> ProgramRef<'_> {
-        ProgramRef {
-            env: &self.env,
-            stats: &self.stats,
-            ctx: self.ctx(),
-        }
+        unreachable!()
```
joshua-spacetime, line 514: If it's unreachable, can we just remove it from the trait?

crates/vm/src/eval.rs
```diff
@@ -1,382 +1,60 @@
...
-    let mut code = HashMap::with_capacity(iter.len());
-
-    for (name, x) in iter {
-        code.insert(name, compile(p, x)?);
-    }
-    Ok(code)
-}
-
-fn compile_query(q: QueryExprOpt) -> QueryCode {
+pub(crate) fn compile_query(q: QueryExpr) -> QueryCode {
```
joshua-spacetime, line 11: So we're not constructing a `QueryExprOpt` anymore?

crates/vm/src/eval.rs
```diff
@@ -460,46 +133,63 @@ pub fn build_query(mut result: Box<IterRows>, query: Vec<Query>) -> Result<Box<I
...
 }
 
-/// Optimize & compile the [Expr] for late execution
+/// Optimize & compile the [CrudExpr] for late execution
 #[tracing::instrument(skip_all)]
-pub fn build_ast<P: ProgramVm>(p: &mut P, ast: Expr) -> Result<Code, ErrorVm> {
-    let ast = optimize(p, ast)?;
-    compile(p, ast)
+pub fn build_ast(ast: CrudExpr) -> Code {
+    compile_query_expr(ast.optimize(&|_, _| i64::MAX))
```
joshua-spacetime, line 139: We should only ever pass in `&|_, _| i64::MAX` when testing. Otherwise we want to use the table size metrics by default.

crates/vm/src/eval.rs
```diff
@@ -1,382 +1,60 @@
...
-        SourceExpr::MemTable(x) => {
-            SourceExprOpt::MemTable(TyExpr::new(x.clone(), Ty::Val(AlgebraicType::Product(x.head.ty()))))
-        }
-        SourceExpr::DbTable(x) => {
-            SourceExprOpt::DbTable(TyExpr::new(x.clone(), Ty::Val(AlgebraicType::Product(x.head.ty()))))
-        }
-    }
-}
-
-fn build_query_opt(q: QueryExpr) -> QueryExprOpt {
```
joshua-spacetime, line 41: Can we get rid of `QueryExprOpt` then? What about `ExprOpt`?

crates/vm/src/eval.rs
```diff
@@ -460,46 +133,63 @@ pub fn build_query(mut result: Box<IterRows>, query: Vec<Query>) -> Result<Box<I
...
+
+/// Optimize, compile & run the [Expr]
+#[tracing::instrument(skip_all)]
+pub fn run_ast<P: ProgramVm>(p: &mut P, ast: Expr) -> Code {
+    let code = match ast {
+        Expr::Block(x) => to_vec(x),
+        Expr::Crud(x) => build_ast(*x),
+        Expr::Value(x) => Code::Value(x),
+        Expr::Halt(err) => Code::Halt(err),
+        Expr::Ident(x) => Code::Halt(ErrorVm::Unsupported(format!("Ident {x}")).into()),
```
joshua-spacetime, line 190: Do we still need `Expr::Ident`?

crates/vm/src/expr.rs
```diff
@@ -1545,9 +1468,9 @@ impl From<QueryExpr> for Expr {
 
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub enum SourceExprOpt {
-    Value(TyExpr<AlgebraicValue>),
-    MemTable(TyExpr<MemTable>),
-    DbTable(TyExpr<DbTable>),
+    Value(AlgebraicValue),
+    MemTable(MemTable),
+    DbTable(DbTable),
```
joshua-spacetime, line 1473: Do we still need the `...Opt` types?

cloutiertyler:
@mamcx There's no way this is a 1 on the complexity scale, right?

gefjon:
> @mamcx There's no way this is a 1 on the complexity scale, right?

5 if I've ever seen one.

crates/vm/src/eval.rs
mamcx: Oh, correct, I miss that and the dependant code

crates/vm/src/eval.rs
mamcx: Yes, is not executed here but the upper layer building the `SELECT column_name <- ident` need it.

crates/vm/src/expr.rs
```diff
@@ -1545,56 +1468,9 @@ impl From<QueryExpr> for Expr {
 
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub enum SourceExprOpt {
```
joshua-spacetime, line 1470: @mamcx is this also dead code now?

crates/core/src/vm.rs
joshua-spacetime: @mamcx is `as_program_ref` safe to remove from the `ProgramVm` trait?

mamcx: Yes

joshua-spacetime:
Even better than expected
```
incr-select             time:   [5.2388 µs 5.2406 µs 5.2424 µs]
                        change: [-54.850% -54.809% -54.771%] (p = 0.00 < 0.05)
                        Performance has improved.

incr-join               time:   [39.819 µs 39.828 µs 39.837 µs]
                        change: [-31.888% -31.785% -31.705%] (p = 0.00 < 0.05)
                        Performance has improved.
```

joshua-spacetime (approved):

PR #862, opened 2024-02-16
Add GitHub workflow to check for PRs with merge-blocking labels

bfops:
# Description of Changes
Added a GitHub workflow to check for PRs with a `Do not merge` label, for ad-hoc PR blocking.

**Note:** A repo admin must [make this](https://github.com/clockworklabs/SpacetimeDB/settings/branches) a merge-blocking check!

No runtime changes.

1

### DIFF
--- a/.github/workflows/check-merge-labels.yml
+++ b/.github/workflows/check-merge-labels.yml
@@ -1,0 +1,29 @@
+name: Check merge labels
+
+on:
+  pull_request:
+    types: [labeled, unlabeled]
+  merge_group:
+permissions: read-all
+
+jobs:
+  check_merge_block_label:
+    if: ${{ contains(github.event.pull_request.labels.*.name, 'Do not merge') }}
+    runs-on: ubuntu-latest
+    steps:
+    - run: exit 1
+
+  check_future_release_label:
+    if: ${{ contains(github.event.pull_request.labels.*.name, 'release-0.9') }}
+    runs-on: ubuntu-latest
+    steps:
+    - run: exit 1
+
+  check_breaking_change_labels:
+    if: | ${{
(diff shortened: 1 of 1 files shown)


.github/workflows/check-merge-labels.yml
```diff
@@ -0,0 +1,33 @@
...
+          echo "This is targeted for a future release."
+          exit 1
+
+      - id: breaking_change
+        if: |
+          contains(github.event.pull_request.labels.*.name, 'abi-break') ||
+          contains(github.event.pull_request.labels.*.name, 'breaking change')
+        run: |
+          echo "This is a breaking change."
+          exit 1
```
bfops, line 33: I made these all `steps` instead of e.g. different `jobs` because it cluttered up the list of checks in the PR view, but I don't feel strongly about it if there's a better organization.

.github/workflows/check-merge-labels.yml
```diff
@@ -0,0 +1,33 @@
...
+
+on:
+  pull_request:
+    types: [labeled, unlabeled]
+  merge_group:
+permissions: read-all
+
+jobs:
+  label_checks:
+    runs-on: ubuntu-latest
```
bfops, line 11: should this always be `spacetimedb-runner`?

It seemed like that always forced the jobs onto a queue, even for these `run`s that don't require anything fancy.

jdetter: Yeah, generally the `spacetimedb-runner` is going to be much faster compared to the regular github ubuntu-latest runner, but for things like this its fine to just use the github runner.

jdetter (approved):
LGTM

PR #863, opened 2024-02-16
perf(832): Remove redundant row deduplication in subscriptions

### LINKED ISSUE #832: Remove unnecessary row deduplication in subscription evaluation
Clients may subscribe to multiple queries that return rows from the same table, however clients expect that rows do not show up more than once in the same update. Therefore we deduplicate subscription results before sending them to the client.

However the database ascribes to set semantics as it is. So query results are by definition sets, and hence for tables that are associated to one query only, we don't have to perform this extra deduplication. We can remove a substantial portion of subscription evaluation by only performing deduplication on query results that require it.

For example, this deduplication is ~25% of a table scan of 1M fixed length rows.

Image

Image

joshua-spacetime:
Closes #832.

The database already operates under set semantics, so unless multiple queries return rows from the same table, deduplication of the result set is not necessary.

## Performance numbers
```
> cargo bench --bench=subscription --profile=profiling -- full-scan --exact

full-scan               time:   [251.71 ms 251.99 ms 252.28 ms]
                        change: [-43.284% -43.172% -43.062%] (p = 0.00 < 0.05)
                        Performance has improved.
```

### Before
Screenshot 2024-02-16 at 2 30 53 PM

### After
Screenshot 2024-02-16 at 2 31 05 PM

### DIFF
--- a/crates/core/src/subscription/module_subscription_actor.rs
+++ b/crates/core/src/subscription/module_subscription_actor.rs
@@ -2,7 +2,7 @@ use std::sync::Arc;
 
 use super::{
     query::compile_read_only_query,
-    subscription::{QuerySet, Subscription},
+    subscription::{ExecutionSet, Subscription},
 };
 use crate::client::{
     messages::{CachedMessage, SubscriptionUpdateMessage, TransactionUpdateMessage},
@@ -15,6 +15,7 @@ use crate::host::module_host::{EventStatus, ModuleEvent};
 use crate::protobuf::client_api::Subscribe;
 use crate::worker_metrics::WORKER_METRICS;
 use futures::Future;
+use itertools::Itertools;
 use parking_lot::RwLock;
 use rayon::iter::{IntoParallelRefIterator, ParallelIterator};
 use spacetimedb_lib::identity::AuthCtx;
@@ -46,11 +47,14 @@ impl ModuleSubscriptions {
         });
 
         let auth = AuthCtx::new(self.owner_identity, sender.id.identity);
-        let mut queries = QuerySet::new();
-        for sql in subscription.query_strings {
-            let qset = compile_read_only_query(&self.relational_db, &tx, &auth, &sql)?;
-            queries.extend(qset);
-        }
+        let mut n_queries = 0;
+        let queries = subscription
+            .query_strings
+            .iter()
+            .map(|sql| compile_read_only_query(&self.relational_db, &tx, &auth, sql))
+            .flatten_ok()
+            .inspect(|_| n_queries += 1)
+            .collect::<Result<ExecutionSet, _>>()?;
 
         let database_update = queries.eval(&self.relational_db, &tx, auth)?;
         // It acquires the subscription lock after `eval`, allowing `add_subscription` to run concurrently.
@@ -65,12 +69,11 @@ impl ModuleSubscriptions {
                 sub
             }
             None => {
-                let n = queries.len();
                 subscriptions.push(Subscription::new(queries, sender));
                 WORKER_METRICS
                     .subscription_queries
                     .with_label_values(&self.relational_db.address())
-                    .add(n as i64);
+                    .add(n_queries as i64);
                 subscriptions.last_mut().unwrap()
             }
         };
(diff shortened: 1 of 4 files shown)


crates/core/src/subscription/subscription.rs
```diff
@@ -120,52 +119,6 @@ impl AsRef<QueryExpr> for SupportedQuery {
     }
 }
 
-/// A set of [supported][`SupportedQuery`] [`QueryExpr`]s.
-#[derive(Debug, Deref, DerefMut, PartialEq, From, IntoIterator)]
-pub struct QuerySet(BTreeSet<SupportedQuery>);
```
joshua-spacetime, line 125: Removed QuerySet in favor of ExecutionSet

crates/core/src/subscription/subscription.rs
```diff
@@ -233,200 +184,17 @@ fn evaluator_for_primary_updates(
...
-        Ok(Self(exprs))
-    }
-
-    /// Incremental evaluation of `rows` that matched the [Query] (aka subscriptions)
-    ///
-    /// This is equivalent to run a `trigger` on `INSERT/UPDATE/DELETE`, run the [Query] and see if the `row` is matched.
-    ///
-    /// NOTE: The returned `rows` in [DatabaseUpdate] are **deduplicated** so if 2 queries match the same `row`, only one copy is returned.
-    #[tracing::instrument(skip_all)]
-    pub fn eval_incr(
```
joshua-spacetime, line 273: I have split the logic for eval and eval_incr depending on whether we are executing a single query or a batch of queries that return rows from the same table.

crates/core/src/subscription/subscription.rs
```diff
@@ -744,6 +501,280 @@ fn with_delta_table(mut join: IndexJoin, index_side: bool, delta: DatabaseTableU
...
+            }
+        }
+        let mut exec_one = BTreeSet::new();
+        let mut exec_set = BTreeSet::new();
+        for (table_id, (name, mut queries)) in queries {
+            if queries.len() == 1 {
+                exec_one.insert(ExecutionUnit(table_id, name, queries.pop_first().unwrap()));
+            } else {
+                exec_set.insert(ExecutionUnit(table_id, name, queries));
+            }
```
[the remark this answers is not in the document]
joshua-spacetime: They're different types. One is a `SupportedQuery`, the other a `Vec<SupportedQuery>`.

crates/core/src/subscription/subscription.rs
```diff
@@ -744,6 +501,280 @@ fn with_delta_table(mut join: IndexJoin, index_side: bool, delta: DatabaseTableU
...
+        let mut ops = Vec::new();
+        let mut dup = HashSet::new();
+
+        for SupportedQuery { kind: _, expr } in &self.2 {
+            for table in run_query(&ctx, db, tx, expr, auth)? {
+                for row in table.data {
+                    let row_pk = pk_for_row(&row);
+
+                    // Skip duplicate rows
+                    if dup.contains(&row_pk) {
```
mamcx, line 587: Wait, is possible to get the same `row_pk` by different tables? because the comment say it need to dedup by table, not by row.

joshua-spacetime: It is possible, but that's fine, because the same row in two different tables is considered two different rows.

crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -45,12 +45,15 @@ impl ModuleSubscriptions {
...
         let auth = AuthCtx::new(self.owner_identity, sender.id.identity);
-        let mut queries = QuerySet::new();
+        let mut queries = Vec::new();
         for sql in subscription.query_strings {
             let qset = compile_read_only_query(&self.relational_db, tx, &auth, &sql)?;
             queries.extend(qset);
         }
 
+        let n = queries.len();
+        let queries = ExecutionSet::from_iter(queries);
```
coolreader18, line 55: This could probably just be all one iter chain now.

coolreader18:
@joshua-spacetime it'd probably be best to merge this first now.

coolreader18:
I could clean this up if that'd be easiest?

joshua-spacetime:
@coolreader18 that would be great! You should be able to simplify this greatly now that we don't have to do any deduplication at all.

joshua-spacetime (commented):
@coolreader18 LGTM!

There are two tests that are failing
1. test_subscribe_dedup
2. test_subscribe_private

For (1) just remove it since we're no longer deduping. In (2) we call two queries that overlap
```
SELECT * FROM inventory
SELECT * FROM inventory WHERE inventory_id = 1
```

Just remove one of those queries from the test. Feel free to merge after that.

PR #864, opened 2024-02-19
Adds an isolation_level argument to begin_mut_tx calls

cloutiertyler:
# Description of Changes

This adds an `isolation_level` argument to mutable transactions. This is currently unused and ignored by our datastore implementation as it already guarantees the highest isolation level, serializability. However, I am adding this in in preparation for introducing an implementation for a slightly weaker `Snapshot` isolation implementation that will allow for parallel execution of queries.

corresponding GitHub label.

2


cloutiertyler:
Did someone order some beefy docs?

PR #872, opened 2024-02-21
Metric for request RTT

Shubham8287:
# Description of Changes
Adds a new metric to measure end to end time of reducer, subscription and one_off_query call.

based on ticket https://github.com/orgs/clockworklabs/projects/22/views/21?pane=issue&itemId=53199356

### DIFF
--- a/crates/core/src/client/message_handlers.rs
+++ b/crates/core/src/client/message_handlers.rs
@@ -1,6 +1,5 @@
 use std::time::{Duration, Instant};
 
-use crate::db::db_metrics::DB_METRICS;
 use crate::energy::EnergyQuanta;
 use crate::execution_context::WorkloadType;
 use crate::host::module_host::{EventStatus, ModuleEvent, ModuleFunctionCall};
@@ -147,7 +146,7 @@ impl DecodedMessage<'_> {
         let res = match self {
             DecodedMessage::Call { reducer, args } => {
                 let res = client.call_reducer(reducer, args).await;
-                DB_METRICS
+                WORKER_METRICS
                     .request_round_trip
                     .with_label_values(&WorkloadType::Reducer, &address, reducer)
                     .observe(timer.elapsed().as_secs_f64());
@@ -155,7 +154,7 @@ impl DecodedMessage<'_> {
             }
             DecodedMessage::Subscribe(subscription) => {
                 let res = client.subscribe(subscription).await;
-                DB_METRICS
+                WORKER_METRICS
                     .request_round_trip
                     .with_label_values(&WorkloadType::Subscribe, &address, "")
                     .observe(timer.elapsed().as_secs_f64());
@@ -166,7 +165,7 @@ impl DecodedMessage<'_> {
                 message_id,
             } => {
                 let res = client.one_off_query(query, message_id).await;
-                DB_METRICS
+                WORKER_METRICS
                     .request_round_trip
                     .with_label_values(&WorkloadType::Sql, &address, "")
                     .observe(timer.elapsed().as_secs_f64());
--- a/crates/core/src/client/message_handlers.rs
+++ b/crates/core/src/client/message_handlers.rs
@@ -1,6 +1,8 @@
-use std::time::Duration;
+use std::time::{Duration, Instant};
 
+use crate::db::db_metrics::DB_METRICS;
 use crate::energy::EnergyQuanta;
+use crate::execution_context::WorkloadType;
 use crate::host::module_host::{EventStatus, ModuleEvent, ModuleFunctionCall};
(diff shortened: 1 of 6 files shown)


crates/core/src/db/db_metrics/mod.rs
```diff
@@ -129,6 +129,11 @@ metrics_group!(
         #[help = "For a given module, the size of its log file (in bytes)"]
         #[labels(db: Address)]
         pub module_log_file_size: IntGaugeVec,
+
+        #[name = spacetime_request_round_time]
+        #[help = "The total time it takes for request to complete"]
+        #[labels(txn_type: WorkloadType, database_address: Address, reducer_symbol: str)]
+        pub request_round_trip: HistogramVec,
```
Shubham8287, line 136: @joshua-spacetime , I am not very sure if this metric should be a part of db_metrics or worker_metrics.

joshua-spacetime: I don't think it matters, but maybe worker_metrics.

crates/core/src/db/db_metrics/mod.rs
```diff
@@ -129,6 +129,11 @@ metrics_group!(
         #[help = "For a given module, the size of its log file (in bytes)"]
         #[labels(db: Address)]
         pub module_log_file_size: IntGaugeVec,
+
+        #[name = spacetime_request_round_time]
```
joshua-spacetime, line 133: Maybe `spacetime_request_round_trip_time` instead?

crates/core/src/db/db_metrics/mod.rs
```diff
@@ -129,6 +129,11 @@ metrics_group!(
...
+        #[help = "The total time it takes for request to complete"]
+        #[labels(txn_type: WorkloadType, database_address: Address, reducer_symbol: str)]
```
joshua-spacetime, line 135: Is it important to be able to distinguish this metric by reducer? Or do we just care about the aggregate?

Shubham8287: I think if we have other metrics to know time spent on indivisual reducers then it can be avoided here.
@cloutiertyler, what's your thoughts on this?

crates/core/src/client/message_handlers.rs
```diff
@@ -136,19 +142,36 @@ enum DecodedMessage<'a> {
 }
 
 impl DecodedMessage<'_> {
-    async fn handle(self, client: &ClientConnection) -> Result<(), MessageExecutionError> {
+    async fn handle(self, client: &ClientConnection, timer: Instant) -> Result<(), MessageExecutionError> {
+        let address = client.module.info().address;
         let res = match self {
             DecodedMessage::Call { reducer, args } => {
                 let res = client.call_reducer(reducer, args).await;
```
cloutiertyler, line 148: Will this only return after all subscription messages have been sent out to clients? Or when exactly does this function return?

cloutiertyler: Could we add a comment explaining when this method returns?

crates/core/src/client/message_handlers.rs
```diff
@@ -136,19 +142,36 @@ enum DecodedMessage<'a> {
...
+                DB_METRICS
+                    .request_round_trip
+                    .with_label_values(&WorkloadType::Reducer, &address, reducer)
+                    .observe(timer.elapsed().as_secs_f64());
                 res.map(drop).map_err(|e| (Some(reducer), e.into()))
             }
             DecodedMessage::Subscribe(subscription) => {
-                client.subscribe(subscription).await.map_err(|e| (None, e.into()))
+                let res = client.subscribe(subscription).await;
```
cloutiertyler, line 156: Similarly when does this return? After the initial subscription has been processed or before that? And could we add a comment to that effect?

crates/core/src/db/db_metrics/mod.rs
cloutiertyler: I think we care about individual ones because some reducers will have much higher latency than others.

crates/core/src/client/message_handlers.rs
Shubham8287: this function `awaits` nestedly for `call_reducer_with_tx` https://github.com/clockworklabs/SpacetimeDB/blob/master/crates/core/src/host/wasm_common/module_host_actor.rs#L481, I think that's the function we know ends after finishing reducer and broadcasting for it.

crates/core/src/client/message_handlers.rs
Shubham8287: we do have await all the way down till initial subscription proccessed. I will add a comment

joshua-spacetime (approved):

PR #879, opened 2024-02-21
Revert "Use a copy-on-write structure for `RelValue`"

cloutiertyler:
Reverts clockworklabs/SpacetimeDB#869

I am reverting this PR as it does not have any demonstrable performance improvement and causes a lot of diff churn for Mario and Joshua to deal with while they're implementing their performance improvements. Specifically, this causes a lot of conflicts with https://github.com/clockworklabs/SpacetimeDB/pull/694 which we're waiting on.


cloutiertyler:
This PR did not end up cause a large number of conflicts and Mario is okay to deal with them in #694, so I will close this.

PR #891, opened 2024-02-26
Parallelize QuerySet::eval

coolreader18:
# Description of Changes

corresponding GitHub label.


crates/core/src/subscription/subscription.rs
```diff
@@ -342,47 +353,64 @@ impl QuerySet {
...
-                            op_type: 1, // Insert
-                            row_pk,
-                            row,
-                        });
-                    }
-                }
+                let tables = run_query(&ExecutionContext::subscribe(db.address()), db, tx, expr, auth)?;
+                v.push((t, tables));
                 #[cfg(feature = "metrics")]
                 record_query_duration_metrics(WorkloadType::Subscribe, &db.address(), start);
```
joshua-spacetime, line 366: Feel free to remove this. We're already collecting sufficient query metrics without it.

crates/core/src/subscription/subscription.rs
```diff
@@ -342,47 +353,64 @@ impl QuerySet {
...
     pub fn eval(&self, db: &RelationalDB, tx: &Tx, auth: AuthCtx) -> Result<DatabaseUpdate, DBError> {
-        let mut database_update: DatabaseUpdate = DatabaseUpdate { tables: vec![] };
-        let mut table_ops = HashMap::new();
-        let mut seen = HashSet::new();
-
-        for SupportedQuery { expr, .. } in self {
-            if let Some(t) = expr.source.get_db_table() {
+        // evaluate each of the queries in this QuerySet in parallel
+        let ops = self
+            .par_iter()
```
joshua-spacetime, line 358: For debugging reasons I would still like to have the option to run this single threaded. Could we add a compile time flag or something that would allow that?

coolreader18: Hmmm. That might be sorta tricky. Are you thinking in order to have the tracing spans be correct?

joshua-spacetime: It'll just be harder to identify issues with individual queries if they're all spread across different threads. If I can run all of them in a single thread, I can more easily see which parts are slow and and drill down into them.

crates/core/src/host/host_controller.rs
```diff
@@ -233,36 +169,38 @@ impl HostController {
...
+        let key = mhc.dbic.database_instance_id;
+
+        let mcc = ModuleCreationContext {
+            dbic: mhc.dbic,
+            scheduler: mhc.scheduler,
+            program_hash: hash_bytes(&mhc.program_bytes),
+            program_bytes: mhc.program_bytes,
+            energy_monitor: self.energy_monitor.clone(),
+        };
+        let module_host = spawn_rayon(move || Self::make_module_host(mhc.host_type, mcc)).await?;
```
joshua-spacetime, line 182: `make_module_host` doesn't acquire locks?

joshua-spacetime (commented):
This looks correct to me. I'm having @jdetter test it, and then I'll approve.

crates/core/src/host/host_controller.rs
coolreader18: Nope, not afaict

joshua-spacetime (approved):
LGTM

PR #894, opened 2024-02-27
Revert "Run initial subscriptions evals on rayon"

cloutiertyler:
Reverts clockworklabs/SpacetimeDB#888 due to a deadlocking issue with multiple clients connected.


jdetter:
symptoms:
- If you run a single client it will run just fine (subscription queries, sql command, etc.)
- If you run the bots against PR 888 then it will deadlock very quickly, no subscription data or anything can come back from SpacetimeDB after that point.
- Restarting the server resolves the issue temporarily until you connect multiple clients again

jdetter (approved):
LGTM

PR #896, opened 2024-02-27
made metric async and fix test

Shubham8287:
# Description of Changes

Made metrics reporting async,
- This should be the main file to review - crates/metrics/src/typed_prometheus.rs
- Made benchmarks to work with tokio rt
- annotated tests with `#[tokio::test]`

Unique index iter benchmark in master:
image

in current branch:
image

### DIFF
--- a/crates/metrics/src/typed_prometheus.rs
+++ b/crates/metrics/src/typed_prometheus.rs
@@ -1,3 +1,6 @@
+use std::fmt::Display;
+
+use arrayvec::ArrayString;
 use prometheus::core::{Metric, MetricVec, MetricVecBuilder};
 
 #[macro_export]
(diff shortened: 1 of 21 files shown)


crates/metrics/src/typed_prometheus.rs
```diff
@@ -13,8 +13,15 @@ macro_rules! metrics_group {
         impl $type_name {
             #[allow(clippy::new_without_default)]
             pub fn new() -> Self {
+                let (tx,mut rx) = tokio::sync::mpsc::unbounded_channel::<Box<dyn FnOnce() + Send>>();
```
joshua-spacetime, line 16: When exactly is this called? How long does this actor live for?

crates/metrics/src/typed_prometheus.rs
```diff
@@ -71,17 +78,34 @@ pub use make_collector;
...
+            pub fn with_label_values_async<F>(&self, $($labels: &$labelty),+, op: F)
+            where
+                F: FnOnce(
+                        <$vecty as $crate::typed_prometheus::ExtractMetricVecT>::M,
+                    ) + Send + 'static,
+            {
+                use $crate::typed_prometheus::AsPrometheusLabel as _;
+                let this = self.clone();
+                $(
+                    let $labels = $labels.to_owned();
```
joshua-spacetime, line 102: This seems very expensive especially since it happens before send.

crates/metrics/src/typed_prometheus.rs
```diff
@@ -71,17 +78,34 @@ pub use make_collector;
...
+                )+
+                let _ = self.1.send(Box::new(move || {
```
joshua-spacetime, line 104: I'm not sure about this allocation, I guess let's see what the benchmarks tell us.

joshua-spacetime (changes requested):
Based on the benchmark I ran, I think we'll have to address those allocations in `with_label_values_async`. It looks like they're just as expensive as `with_label_values`.

https://share.firefox.dev/4a97mJd

PR #901, opened 2024-02-28
Fixed an issue which caused metrics to only be recorded for on-disk databases

cloutiertyler:
# Description of Changes

Fixed an issue which caused metrics to only be recorded for on-disk databases.

Note to reviewer: I'm not sure that it makes sense to pass the ExecutionContext so deep, but this works.

None

2


jdetter (approved):
We tested this during a meeting using the bots, working from a functionality standpoint 👍

PR #903, opened 2024-02-28
acquire remove subscription lock in spawn_blocking

Shubham8287:
# Description of Changes

corresponding GitHub label.

### DIFF
--- a/crates/core/src/subscription/module_subscription_actor.rs
+++ b/crates/core/src/subscription/module_subscription_actor.rs
@@ -21,7 +21,7 @@ use spacetimedb_lib::identity::AuthCtx;
 use spacetimedb_lib::Identity;
 
 type Subscriptions = Arc<RwLock<Vec<Subscription>>>;
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 pub struct ModuleSubscriptions {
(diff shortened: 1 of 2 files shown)


crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -90,8 +90,11 @@ impl ModuleSubscriptions {
     }
 
     pub fn remove_subscriber(&self, client_id: ClientActorId) {
-        let mut subscriptions = self.subscriptions.write();
-        self._remove_subscriber(client_id, &mut subscriptions);
+        let this = self.clone();
+        tokio::task::spawn_blocking(move || {
+            let mut subscriptions = this.subscriptions.write();
+            this._remove_subscriber(client_id, &mut subscriptions);
+        });
```
coolreader18, line 97: I think it'd be better to do this spawn_blocking in remove_subscriber in module_host, not here. And make sure you `.await` on the spawned task, otherwise this becomes asynchronous which I don't think is what we want(?)

Shubham8287: Makes sense!

coolreader18 (approved):

PR #904, opened 2024-02-28
core: Fix host controller to not replace module if lifecyle hooks failed

kim:
Previously, `spawn_module_host` would unconditionally insert the new module into the controller state, and not remove it if the lifecycle hooks (`init_database` / `update_database`) returned an error.

This would mean that the module code was replaced with the new one, even if it should be rejected because the schema was not updated or the init / update reducer failed.

Fix this by starting the module, and later "committing" it to the controller state in two phases.

### DIFF
--- a/crates/core/src/host/host_controller.rs
+++ b/crates/core/src/host/host_controller.rs
@@ -179,6 +179,9 @@ impl HostController {
     /// reducer scheduler is started.
     ///
     /// Otherwise, if `F` returns an `Err` result, the module is discarded.
+    ///
+    /// In the `Err` case, `F` **MUST** roll back any modifications it has made
+    /// to the database passed in the [`ModuleHostContext`].
     async fn setup_module_host<F, Fut, T>(&self, mhc: ModuleHostContext, f: F) -> anyhow::Result<T>
     where
         F: FnOnce(ModuleHost) -> Fut,
@@ -250,6 +253,9 @@ impl HostController {
     ///
     /// This is a fairly expensive operation and should not be run on the async
     /// task threadpool.
+    ///
+    /// Note that this function **MUST NOT** make any modifications to the
+    /// database passed in as part of the [`ModuleCreationContext`].
     fn make_module_host(host_type: HostType, mcc: ModuleCreationContext) -> anyhow::Result<ModuleHost> {
         let module_host = match host_type {
             HostType::Wasm => {
--- a/crates/core/src/host/host_controller.rs
+++ b/crates/core/src/host/host_controller.rs
@@ -1,14 +1,18 @@
+use crate::db::update::UpdateDatabaseError;
 use crate::energy::{EnergyMonitor, EnergyQuanta, NullEnergyMonitor};
+use crate::execution_context::ExecutionContext;
 use crate::hash::hash_bytes;
 use crate::host;
 use crate::messages::control_db::HostType;
 use crate::module_host_context::{ModuleCreationContext, ModuleHostContext};
 use crate::util::spawn_rayon;
-use anyhow::Context;
+use anyhow::{ensure, Context};
+use futures::TryFutureExt;
 use parking_lot::Mutex;
 use serde::Serialize;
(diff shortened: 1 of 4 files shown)


crates/core/src/host/host_controller.rs
```diff
@@ -170,27 +168,88 @@ impl HostController {
...
             scheduler: mhc.scheduler,
-            program_hash: hash_bytes(&mhc.program_bytes),
+            program_hash,
             program_bytes: mhc.program_bytes,
             energy_monitor: self.energy_monitor.clone(),
         };
         let module_host = spawn_rayon(move || Self::make_module_host(mhc.host_type, mcc)).await?;
+        module_host.start();
+
+        let res = f(module_host.clone())
```
cloutiertyler, line 200: That's the same, I suppose with `f`. If `f` makes actual changes to the database even if it returns an error, then throwing away the module may still have modified the database. Perhaps we should just add a comment in the description that if and only if `f` returns an error it is guaranteed to have made no mutations to the database.

crates/core/src/host/host_controller.rs
```diff
@@ -170,27 +168,88 @@ impl HostController {
...
+            //
+            // It is unfortunate that [`Self::spawn_module_host`] is both public
+            // and takes the module's program bytes in its argument. Because it
+            // can be called concurrently to [`Self::init_module_host`] or
+            // [`Self::update_module_host`], we may end up with the wrong module
+            // version here.
+            //
+            // We should instead store the current program bytes verbatim in the
+            // database, such that `spawn_module_host` operates on the committed
+            // state only.
```
cloutiertyler, line 232: Amen

cloutiertyler (approved):
This looks good and makes sense to me. I have just requested some additional commentary.

crates/core/src/host/host_controller.rs
kim: We happen to control `f`, which is carefully crafted to run in a single transaction. But yes, adding more commentary can't hurt.

crates/core/src/host/host_controller.rs
```diff
@@ -170,27 +181,76 @@ impl HostController {
...
+    }
 
-        let old_module = self.modules.lock().insert(key, module_host.clone());
+    /// Complete the setup of the [`ModuleHost`] started using [`Self::start_module_host`].
+    ///
+    /// Exits the previous version of the module, if any, starts the new module's
+    /// scheduler, and registers the module with this host controller.
+    async fn commit_module_host(&self, spawned: SpawnedModuleHost) -> anyhow::Result<ModuleHost> {
+        {
+            // At this point, the supplied program must be the program stored
```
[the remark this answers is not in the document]
cloutiertyler: Yeah, in the future, I think `spawn_module_host` should read the program bytes from the database/commit log as you suggest. The way we could perhaps do it would be to send a message to the host_controller to stop all modules, then insert a new row in the datastore directly and then send a message to the host_controller to reload the modules.

Perhaps there's an even better way to do it, that's just off the top of my head.

crates/core/src/host/host_controller.rs
```diff
@@ -170,27 +168,88 @@ impl HostController {
...
         let mcc = ModuleCreationContext {
-            dbic: mhc.dbic,
+            dbic: mhc.dbic.clone(),
             scheduler: mhc.scheduler,
-            program_hash: hash_bytes(&mhc.program_bytes),
+            program_hash,
             program_bytes: mhc.program_bytes,
             energy_monitor: self.energy_monitor.clone(),
         };
         let module_host = spawn_rayon(move || Self::make_module_host(mhc.host_type, mcc)).await?;
```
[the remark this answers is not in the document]
kim: Interesting. 

We have to bind the database to the module in order to be able to call init / update reducers (which themselves run in a transactional context). So I suppose the contract of `host::wasmtime::make_actor` must just be to not touch the database that was passed in?

Maybe @coolreader18 has some thoughts on this?

PR #908, opened 2024-02-28
perf(831): Remove row_pk computation from query

### LINKED ISSUE #280: Remove PrimaryKey
The current `PrimaryKey` type is just a wrapper for `DataKey` which uniquely identifies each row in the database. It is not a column or row attribute and therefore does correspond to a traditional primary key in a relational database. We should remove to avoid confusion.

### LINKED ISSUE #831: Subscriptions should not compute primary keys

### LINKED ISSUE #874: Remove `RelationalDB::pk_for_row`
See e.g., https://github.com/clockworklabs/SpacetimeDB/pull/869/files#r1497828101 and friends.

joshua-spacetime:
Closes #831.
Closes #874.
Closes #280.

```
full-scan               time:   [105.90 ms 106.18 ms 106.51 ms]
                        change: [-53.505% -53.317% -53.150%] (p = 0.00 < 0.05)
                        Performance has improved.

full-join               time:   [312.20 µs 313.41 µs 314.63 µs]
                        change: [-33.453% -33.176% -32.873%] (p = 0.00 < 0.05)
                        Performance has improved.

incr-select             time:   [2.0535 µs 2.0584 µs 2.0649 µs]
                        change: [-12.335% -12.142% -11.936%] (p = 0.00 < 0.05)
                        Performance has improved.

incr-join               time:   [10.726 µs 10.747 µs 10.769 µs]
                        change: [+0.3173% +0.5243% +0.7765%] (p = 0.00 < 0.05)
```

Removes some `#[tracing::instrument(skip_all)]` as well.

1


crates/core/src/subscription/query.rs
```diff
@@ -404,46 +392,6 @@ mod tests {
         Ok(())
     }
 
-    #[test]
```
cloutiertyler, line 407: Maybe we want to check that the row values are equal?

crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -89,6 +89,7 @@ impl ModuleSubscriptions {
         // thread it's possible for messages to get sent to the client out of order. If you do
         // spawn in another thread messages will need to be buffered until the state is sent out
         // on the wire
+        // let _span = tracing::info_span!("send").entered();
```
cloutiertyler, line 92: Maybe pull this out?

cloutiertyler (commented):
This seems correct to me. Just two comments. Also wow, that scan speedup.

crates/core/src/subscription/module_subscription_actor.rs
joshua-spacetime: Ah yes, some left over debugging.

crates/core/src/subscription/query.rs
joshua-spacetime: Yes, but technically we already have tests for that.

PR #913, opened 2024-03-01
Use recv_many in ws_client_actor

coolreader18:
# Description of Changes

This batches messages onto the websocket instead of sending them one at a time. This might help latency with clients being paused for a long time, but I'm still unsure about the root cause of that issue so I wouldn't count much on it.

Supersedes #902

1

### DIFF
--- a/crates/client-api/src/routes/subscribe.rs
+++ b/crates/client-api/src/routes/subscribe.rs
@@ -5,7 +5,7 @@ use std::time::Duration;
 use axum::extract::{Path, Query, State};
 use axum::response::IntoResponse;
 use axum_extra::TypedHeader;
-use futures::{SinkExt, StreamExt};
+use futures::{FutureExt, SinkExt, StreamExt};
 use http::{HeaderValue, StatusCode};
 use serde::Deserialize;
 use spacetimedb::client::messages::{IdentityTokenMessage, ServerMessage};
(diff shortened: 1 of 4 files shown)


crates/client-api/src/routes/subscribe.rs
```diff
@@ -204,15 +207,15 @@ async fn ws_client_actor(client: ClientConnection, mut ws: WebSocketStream, mut
...
-            Some(message) = sendrx.recv() => {
+            Some(n) = sendrx.recv_many(&mut rx_buf, 32).map(NonZeroUsize::new) => {
                 if closed {
                     // TODO: this isn't great. when we receive a close request from the peer,
                     //       tungstenite doesn't let us send any new messages on the socket,
                     //       even though the websocket RFC allows it. should we fork tungstenite?
-                    log::info!("dropping message due to ws already being closed: {message:?}");
+                    log::info!("dropping messages due to ws already being closed: {:?}", &rx_buf[..n.get()]);
                 } else {
-                    // TODO: I think we can be smarter about feeding messages here?
```
jdetter, line 214: love to see a TODO being addressed 😅

jdetter (approved):
I've been testing on top of this for the last few bot tests. It seems to improve the gap issue, but not completely solve it. This change is just a generallly good change to make so let's keep it 👍

PR #915, opened 2024-03-01
CLI - Remove deprecated use of the `--project-path` flag

bfops:
# Description of Changes

Remove the deprecated use of `--project-path` to provide a wasm file.

Yes, this breaks the CLI API.

1


bfops:
(do not merge until we're ready to make this breaking change!)

cloutiertyler (approved):
Seems fine. Do we no longer allow you to specify the project for the wasm and just require you point directly to the wasm binary?

bfops:
> Seems fine. Do we no longer allow you to specify the project for the wasm and just require you point directly to the wasm binary?

I'm not 100% sure what you're asking so I hope this is clear: we never allowed users to specify both. Previously, `--project-path` could take _either_ a directory _or_ a wasm path. Then we added `--wasm-file` (now `--bin-path`) to make the behavior clearer, but we still quietly supported taking a wasm file via `--project-path`. After this PR, you use `--project-path` to provide a directory, and `--bin-path` to provide a wasm file.

PR #916, opened 2024-03-02
Batch metrics with ctx

Shubham8287:
# Description of Changes

On hotpath, prometheus `label_with_values` can is slow.
Attempt is to batch the metrics on Hotpath and report them while dropping `ctx`.
This PR has a merge from https://github.com/clockworklabs/SpacetimeDB/pull/1017, so diff is bit messy.

1


crates/core/src/db/datastore/locking_tx_datastore/committed_state.rs
```diff
@@ -569,32 +569,37 @@ impl<'a> CommittedIndexIter<'a> {
...
-        let reducer_name = self.ctx.reducer_name();
-        let table_id = &self.table_id.0;
         let table_name = self
             .committed_state
             .get_schema(&self.table_id)
             .map(|table| table.table_name.as_str())
             .unwrap_or_default();
 
-        DB_METRICS
+        let mut metrics = self.ctx.metrics.lock();
```
joshua-spacetime, line 578: Do we need a lock? Each transaction gets its own execution context.

crates/core/src/execution_context.rs
```diff
@@ -97,3 +118,52 @@ impl<'a> ExecutionContext<'a> {
...
 }
+
+impl Drop for ExecutionContext<'_> {
+    fn drop(&mut self) {
+        let workload = self.workload;
+        let database = self.database;
+        let reducer = self.reducer.unwrap_or_default().to_string();
+        let metric = self.metrics.clone();
+        log::info!("dropping execution context");
+        tokio::task::spawn_blocking(move || {
```
joshua-spacetime, line 129: Does this need to be spawn_blocking?

crates/core/src/execution_context.rs
```diff
@@ -1,17 +1,126 @@
...
+    pub fn new() -> Self {
+        Self(Vec::new())
+    }
+
+    pub fn add_metric(&mut self, table_id: TableId, table_name: &str) {
+        self.0.push(BufferMetric::new(table_id, table_name));
+    }
+
+    pub fn inc_by(&mut self, table_id: TableId, ty: MetricType, val: u64) {
+        if let Some(metric) = self.0.iter_mut().find(|x| x.table_id == table_id) {
```
joshua-spacetime, line 63: I tested this, and it doesn't add that much overhead to the benchmark (~1.5%). But that benchmark only reads from two tables. I am a little concerned about transactions where this vec could be much larger.

crates/core/src/execution_context.rs
```diff
@@ -1,17 +1,125 @@
+use std::cell::RefCell;
+
 use derive_more::Display;
 use spacetimedb_lib::Address;
+use spacetimedb_primitives::TableId;
+
+use crate::db::db_metrics::DB_METRICS;
+
+pub enum MetricType {
+    RdbNumIndexSeeks,
```
joshua-spacetime, line 10: Nit: We don't need to include the `RdbNum` prefix here.

crates/core/src/execution_context.rs
```diff
@@ -1,17 +1,125 @@
...
+    RdbNumKeysScanned,
+    RdbNumRowsFetched,
+}
+#[derive(Default, Clone)]
+struct BufferMetric {
+    pub table_id: TableId,
+    pub rdb_num_index_seeks: u64,
```
joshua-spacetime, line 17: Similarly here.

crates/core/src/execution_context.rs
```diff
@@ -1,17 +1,125 @@
...
+impl Metrics {
+    pub fn new() -> Self {
+        Self(Vec::new())
+    }
+
+    pub fn add_metric(&mut self, table_id: TableId, table_name: &str) {
+        self.0.push(BufferMetric::new(table_id, table_name));
+    }
+
+    pub fn inc_by(&mut self, table_id: TableId, ty: MetricType, val: u64) {
```
joshua-spacetime, line 62: Could `table_exists` and `add_metric` be merged into `inc_by`. In other words, can `inc_by` check if the table exists and if not, add the new metric to the list?

crates/core/src/execution_context.rs
```diff
@@ -1,17 +1,125 @@
...
+}
+
+impl BufferMetric {
+    pub fn new(table_id: TableId, table_name: &str) -> Self {
+        Self {
+            table_id,
+            rdb_num_index_seeks: 0,
+            rdb_num_keys_scanned: 0,
+            rdb_num_rows_fetched: 0,
+            cache_table_name: table_name.to_string(),
```
joshua-spacetime, line 46: Would it be relatively simple to take a reference here rather than clone?

crates/core/src/execution_context.rs
```diff
@@ -1,17 +1,125 @@
...
+#[derive(Default, Clone)]
 pub struct ExecutionContext<'a> {
     /// The database on which a transaction is being executed.
     database: Address,
     /// The reducer from which the current transaction originated.
     reducer: Option<&'a str>,
     /// The type of workload that is being executed.
     workload: WorkloadType,
+    /// The Metrics to be reported for this transaction.
+    pub metrics: RefCell<Metrics>,
```
joshua-spacetime, line 122: I'm guessing RefCell is to avoid changing all of the call sites from `ctx: &ExecutionContext` to `ctx: &mut ExecutionContext`?

crates/core/src/host/wasmtime/wasm_instance_env.rs
```diff
@@ -582,7 +582,7 @@ impl WasmInstanceEnv {
             let data = env
                 .instance_env
                 .iter_by_col_eq(&ctx, table_id.into(), col_id.into(), value)?;
-
+            drop(ctx);
```
joshua-spacetime, line 585: We probably don't want to be dropping this execution context for every iter_by_col_eq invocation since we could have many of those in a single transaction.

Instead of passing this to iter_by_col_eq, I think we should probably add an ExecutionContext to `TxSlot` and have iter_by_col_eq just mutate it on the TxSlot directly. That way we only drop it once and therefore only write the metrics once per transaction.

joshua-spacetime (changes requested):
Sorry for late review @Shubham8287, but I think this is definitely worth merging. This pretty much removes the current performance penalty associated with metrics collection from the hot path.

The only thing I think we'll want to do before merging is to make sure we're not dropping execution contexts, and therefore writing metrics, more than once per transaction, which is still possible through InstanceEnv.

crates/core/src/host/wasmtime/wasm_instance_env.rs
Shubham8287: Make sense, will do that.

crates/core/src/execution_context.rs
Shubham8287: yes

crates/core/src/execution_context.rs
Shubham8287: Lifetime issues can be pain there, had to make sure `ctx` always outlive `table_name` will anaylyze this bit more.

crates/core/src/execution_context.rs
Shubham8287: I am also not happy with the current way we are exposing methods here, have to refactor it, PR was mostly to test the idea.
Regarding `Could table_exists and add_metric be merged into inc_by.` 
Not very straightforward as right now, Idea is to only fetch `table_name` only once i.e when adding new metric. `inc_by` get calls more often, I would not like to make it dependent on `table_name` so will try to think something else

crates/core/src/host/wasmtime/wasm_instance_env.rs
Shubham8287: I tried this, but adding`ExecutionContext` to `TxSlot` is not very straightforward, it either requires to make TxSlot use a explicit lifetime or make `ExecutionContext` to not use lifetime.
I am making this change but either of this will be a significant change `ExecutionContext` thay can/should go with different PR.

joshua-spacetime: I think your suggestion to remove the lifetime parameter from ExecutionContext is the right approach.

Shubham8287: `drop`, here now if dropping of Gaurd not actual memory.

crates/core/src/execution_context.rs
```diff
@@ -1,17 +1,127 @@
...
+        });
+    }
+}
 
 /// Represents the context under which a database runtime method is executed.
 /// In particular it provides details about the currently executing txn to runtime operations.
 /// More generally it acts as a container for information that database operations may require to function correctly.
-#[derive(Default)]
-pub struct ExecutionContext<'a> {
+#[derive(Default, Clone)]
```
joshua-spacetime, line 115: Does it need to be Clone?

crates/core/src/execution_context.rs
```diff
@@ -1,17 +1,127 @@
...
+pub struct ExecutionContext {
     /// The database on which a transaction is being executed.
     database: Address,
     /// The reducer from which the current transaction originated.
-    reducer: Option<&'a str>,
+    reducer: Option<String>,
     /// The type of workload that is being executed.
     workload: WorkloadType,
+    /// The Metrics to be reported for this transaction.
+    pub metrics: Arc<RwLock<Metrics>>,
```
joshua-spacetime, line 124: Will there ever be concurrent access that this needs to be behind a mutex?

crates/core/src/host/wasm_common/module_host_actor.rs
```diff
@@ -531,7 +531,7 @@ impl<T: WasmInstance> WasmModuleInstance<T> {
         };
 
         let tx = tx.unwrap_or_else(|| stdb.begin_mut_tx(IsolationLevel::Serializable));
-        let tx_slot = self.instance.instance_env().tx.clone();
+        let mut tx_slot = self.instance.instance_env().tx.clone();
```
joshua-spacetime, line 534: Oh I see, we clone the tx slot, so we also need to clone the execution context that it holds.

crates/core/src/host/wasmtime/wasm_instance_env.rs
```diff
@@ -603,10 +606,10 @@ impl WasmInstanceEnv {
         Self::cvt_ret(caller, AbiCall::IterStart, out, |caller| {
             let env = caller.data_mut();
             // Retrieve the execution context for the current reducer.
-            let ctx = env.reducer_context();
+            let ctx = env.reducer_context()?;
             // Collect the iterator chunks.
             let chunks = env.instance_env.iter_chunks(&ctx, table_id.into())?;
-
+            drop(ctx);
```
joshua-spacetime, line 612: Do we still need the explicit drops?

joshua-spacetime:
Unfortunately it looks like it's still quite expensive
```
incr-join               time:   [2.2527 µs 2.2539 µs 2.2554 µs]
                        change: [+43.975% +44.175% +44.373%] (p = 0.00 < 0.05)
                        Performance has regressed.

incr-select             time:   [404.59 ns 405.26 ns 405.91 ns]
                        change: [+12.378% +12.551% +12.712%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

https://share.firefox.dev/4cJm2AO (incr-join)
https://share.firefox.dev/3TyifNO (incr-select)

Shubham8287:
> Unfortunately it looks like it's still quite expensive
> 
> ```
> incr-join               time:   [2.2527 µs 2.2539 µs 2.2554 µs]
>                         change: [+43.975% +44.175% +44.373%] (p = 0.00 < 0.05)
>                         Performance has regressed.
> 
> incr-select             time:   [404.59 ns 405.26 ns 405.91 ns]
>                         change: [+12.378% +12.551% +12.712%] (p = 0.00 < 0.05)
>                         Performance has regressed.
> ```
> 
> https://share.firefox.dev/4cJm2AO (incr-join) https://share.firefox.dev/3TyifNO (incr-select)

ah! It regressed because batching is not happening for `incr-join` rather they create new `ctx`  for each `ExecutionUnit`. Let me see, what can be done here.

Shubham8287:
benchmarks please

Shubham8287:
@joshua-spacetime , It looks regression you saw was because my branch was not synced with master, here are the results I measure after syncing 
```
full-scan               time:   [78.772 ms 79.304 ms 79.845 ms]
                        change: [-0.6241% +0.3867% +1.3107%] (p = 0.43 > 0.05)
                        No change in performance detected.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild

full-join               time:   [226.13 µs 226.76 µs 227.43 µs]
                        change: [-51.930% -51.767% -51.596%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild

incr-select             time:   [132.39 ns 132.87 ns 133.35 ns]
                        change: [-1.1616% -0.7623% -0.3836%] (p = 0.00 < 0.05)

incr-join               time:   [854.91 ns 857.69 ns 860.71 ns]
                        change: [-34.533% -34.222% -33.912%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  3 (3.00%) high mild
  1 (1.00%) high severe

query-indexes-multi     time:   [473.41 ns 474.45 ns 475.46 ns]
                        change: [-28.676% -28.413% -28.158%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
```

crates/core/src/host/wasm_common/module_host_actor.rs
Shubham8287: yeah that explains your above comment regarding `Arc<Mutex<>>`

crates/core/src/host/wasmtime/wasm_instance_env.rs
Shubham8287: Yes, ctx is shared reference from `env` (`TxSlot`) now and below line `env.iter.insert()` is mutable reference so can't exist together. Though, this a only a Mutex Gaurd drop not  object drop.

crates/core/src/execution_context.rs
Shubham8287: Yes, I can think of subscriptions where multiple threads may use same ctx.

joshua-spacetime:
@Shubham8287 when I ran the benchmarks I did update your branch with master. In order to get an accurate diff, you'll need to first run the benchmarks on master (with metrics disabled), and then run the benchmarks on your branch (with metrics enabled which is the default).

Shubham8287:
@joshua-spacetime , This is bench results with disabled metrics on master and then on my branch with metrics,
```
full-scan               time:   [77.284 ms 77.431 ms 77.578 ms]
                        change: [-1.3874% -0.9307% -0.5210%] (p = 0.00 < 0.05)
Found 3 outliers among 100 measurements (3.00%)
  2 (2.00%) high mild
  1 (1.00%) high severe

full-join               time:   [235.85 µs 236.73 µs 237.73 µs]
                        change: [+10.066% +10.541% +11.042%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild

incr-select             time:   [139.94 ns 140.55 ns 141.20 ns]
                        change: [-0.7931% -0.1905% +0.3154%] (p = 0.53 > 0.05)
                        No change in performance detected.
Found 15 outliers among 100 measurements (15.00%)
  9 (9.00%) low mild
  5 (5.00%) high mild
  1 (1.00%) high severe

incr-join               time:   [886.85 ns 888.99 ns 891.13 ns]
                        change: [+6.2229% +6.5017% +6.7732%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 4 outliers among 100 measurements (4.00%)
  3 (3.00%) high mild
  1 (1.00%) high severe

query-indexes-multi     time:   [472.24 ns 474.12 ns 475.82 ns]
                        change: [+10.995% +11.391% +11.811%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

Still not as bad as your bench, Lets meet sometime to find if there is any flaw in my way of doing bench?

crates/core/src/subscription/execution_unit.rs
```diff
@@ -206,8 +206,13 @@ impl ExecutionUnit {
 
     /// Evaluate this execution unit against the database using the json format.
     #[tracing::instrument(skip_all)]
-    pub fn eval_json(&self, db: &RelationalDB, tx: &Tx) -> Result<Option<TableUpdateJson>, DBError> {
-        let table_row_operations = Self::eval_query_expr(db, tx, &self.eval_plan, |row| {
+    pub fn eval_json(
+        &self,
+        ctx: &ExecutionContext,
+        db: &RelationalDB,
+        tx: &Tx,
```
joshua-spacetime, line 213: Because we want to write metrics at most once per transaction, it would probably be best to embed the ExecutionContext directly in the Tx. But that might require us to update a lot of function signatures.

crates/bench/benches/subscription.rs
```diff
@@ -103,8 +103,8 @@ fn eval(c: &mut Criterion) {
             let tx = db.begin_tx();
             let query = compile_read_only_query(&db, &tx, &auth, sql).unwrap();
             let query: ExecutionSet = query.into();
-
-            b.iter(|| drop(black_box(query.eval(Protocol::Binary, &db, &tx).unwrap())))
+            let ctx = &ExecutionContext::subscribe(db.address());
```
joshua-spacetime, line 106: Because `with_label_values` is now called when we drop the `ExecutionContext`, this should probably go inside the lambda.

joshua-spacetime (approved):
Excellent @Shubham8287! I profiled your changes and they add at most a ~2% overhead.

crates/core/src/subscription/execution_unit.rs
joshua-spacetime: Ah I saw you opened #1030. Excellent!

PR #917, opened 2024-03-03
perf(747): Single query execution for multiple subscriptions 

joshua-spacetime:
# Description of Changes

Previously we would evaluate each and every subscription on each and every row update.
If N subscriptions had a query Q in common, Q would be evaluated N different times.
With this change, queries are evaluated once, and the results transmitted to each client.
So in the example above, Q would be evaluated once, with the results duplicated to N different clients.

Some things to note:
1. It's currently single threaded
    In principle evaluating row updates should be very fast.
    On the order of microseconds.
    In that case parallelizing eval probably just adds overhead.
    However if we didn't have to aggregate the different table updates before sending to the client, it might be beneficial.
2. It's evaluated on a rayon thread.
    Again, if eval_incr is on the order of microseconds, is this just adding overhead?
3. When sending to clients, we clone `DatabaseTableUpdate`
    This means we clone product values.
    It would be better if we could serialize each DatabaseTableUpdate, so that we could just memcpy, but that would require changes to our protobuf schema.

None

4 Mainly due to subpar test coverage of eval_incr (although we have decent coverage, just not excellent)

### DIFF
--- a/crates/core/src/subscription/execution_unit.rs
+++ b/crates/core/src/subscription/execution_unit.rs
@@ -1,0 +1,161 @@
+use std::hash::Hash;
+
+use spacetimedb_lib::identity::AuthCtx;
+use spacetimedb_primitives::TableId;
+use spacetimedb_vm::expr::SourceSet;
+
+use crate::db::relational_db::{RelationalDB, Tx};
+use crate::error::DBError;
+use crate::execution_context::ExecutionContext;
+use crate::host::module_host::{DatabaseTableUpdate, TableOp};
+
+use super::query::{self, run_query, Supported};
+use super::subscription::{eval_primary_updates, IncrementalJoin, SupportedQuery};
+
+/// A hash for uniquely identifying query execution units,
+/// to avoid recompilation of queries that have an open subscription.
+///
+/// Currently we are using a cryptographic hash,
+/// which is most certainly overkill.
+/// However the benefits include uniqueness by definition,
+/// and a compact representation for equality comparisons.
+///
+/// It also decouples the hash from the physical plan.
+///
+/// Note that we could hash QueryExprs directly,
+/// using the standard library's hasher.
+/// However some execution units are comprised of several query plans,
+/// as is the case for incremental joins.
+/// And we want to associate a hash with the entire unit of execution,
+/// rather than an individual plan.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct QueryHash {
+    data: [u8; 32],
+}
+
+impl QueryHash {
+    pub const NONE: Self = Self { data: [0; 32] };
+
+    pub fn from_bytes(bytes: &[u8]) -> Self {
+        Self {
+            data: blake3::hash(bytes).into(),
+        }
+    }
+
+    pub fn from_string(str: &str) -> Self {
+        Self::from_bytes(str.as_bytes())
+    }
+}
+
+/// An atomic unit of execution within a subscription set.
+/// Currently just a single query plan,
+/// however in the future this could be multiple query plans,
+/// such as those of an incremental join.
+#[derive(Debug)]
+pub struct ExecutionUnit {
+    hash: QueryHash,
+    plan: SupportedQuery,
+}
+
+/// An ExecutionUnit is uniquely identified by its QueryHash.
+impl Eq for ExecutionUnit {}
+
+impl PartialEq for ExecutionUnit {
+    fn eq(&self, other: &Self) -> bool {
+        self.hash == other.hash
+    }
+}
+
+impl From<SupportedQuery> for ExecutionUnit {
+    fn from(plan: SupportedQuery) -> Self {
+        ExecutionUnit {
+            hash: QueryHash::NONE,
+            plan,
+        }
+    }
+}
+
+impl ExecutionUnit {
+    pub fn new(plan: SupportedQuery, hash: QueryHash) -> Self {
+        ExecutionUnit { hash, plan }
+    }
+
+    /// Is this a single table select or a semijoin?
+    pub fn kind(&self) -> Supported {
+        self.plan.kind
+    }
+
+    /// The unique query hash for this execution unit.
+    pub fn hash(&self) -> QueryHash {
+        self.hash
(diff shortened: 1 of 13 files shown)


crates/core/src/subscription/execution_unit.rs
```diff
@@ -0,0 +1,146 @@
...
+    /// this is the same as the return table.
+    /// In the case of a semijoin,
+    /// it is the auxiliary table against which we are joining.
+    pub fn filter_table(&self) -> TableId {
+        self.plan.filter_table()
+    }
+
+    /// Evaluate this execution unit against the database.
+    #[tracing::instrument(skip_all)]
+    pub fn eval(&self, db: &RelationalDB, tx: &Tx, auth: AuthCtx) -> Result<Option<DatabaseTableUpdate>, DBError> {
```
joshua-spacetime, line 98: Copied from the old eval code.

crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -47,44 +42,49 @@ impl ModuleSubscriptions {
...
+        for sql in subscription.query_strings {
+            let hash = QueryHash::from_string(&sql);
+            if let Some(unit) = guard.query(&hash) {
+                queries.push(unit);
+            } else {
+                queries.extend(
+                    compile_read_only_query(&self.relational_db, &tx, &auth, &sql)?
+                        .into_iter()
+                        .map(|plan| Arc::new(ExecutionUnit::new(plan, hash))),
+                )
```
gefjon, line 58: Why do we have `queries.push` in one branch, but `queries.extend` in the other? Surely we want to add exactly 1 element in either branch, right?

gefjon (commented):
This all seems reasonable. Minor requests for clarification/comments. Happy to approve once we get a PR description written.

crates/core/src/subscription/module_subscription_manager.rs
```diff
@@ -0,0 +1,389 @@
...
+use crate::execution_context::ExecutionContext;
+use crate::host::module_host::{DatabaseTableUpdate, DatabaseUpdate, ModuleEvent};
+use smallvec::SmallVec;
+use spacetimedb_lib::identity::AuthCtx;
+use spacetimedb_lib::Identity;
+use spacetimedb_primitives::TableId;
+use std::collections::{HashMap, HashSet};
+use std::sync::Arc;
+
+type Id = Identity;
```
mamcx, line 15: I think this alias makes things more opaque.

crates/core/src/subscription/module_subscription_manager.rs
```diff
@@ -0,0 +1,389 @@
...
+                        self.tables.remove(&query.filter_table());
+                    }
+                }
+            }
+            !ids.is_empty()
+        });
+    }
+
+    #[tracing::instrument(skip_all)]
+    pub async fn eval_updates(&self, db: &RelationalDB, auth: AuthCtx, event: Arc<ModuleEvent>) -> Result<(), DBError> {
```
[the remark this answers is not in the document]
mamcx: In concrete, what are the expectations/invariants when you evaluate the updates, so if somebody make changes here it knows...

crates/core/src/subscription/subscription.rs
```diff
@@ -101,6 +102,37 @@ impl SupportedQuery {
...
+
+    pub fn return_table(&self) -> TableId {
+        self.expr.source.get_db_table().unwrap().table_id
+    }
+
+    pub fn return_name(&self) -> String {
+        self.expr.source.table_name().to_owned()
+    }
+
+    pub fn filter_table(&self) -> TableId {
```
mamcx, line 114: What is the purpose of this filtering?

crates/core/src/subscription/execution_unit.rs
```diff
@@ -0,0 +1,146 @@
...
+use crate::db::relational_db::{RelationalDB, Tx};
+use crate::error::DBError;
+use crate::execution_context::ExecutionContext;
+use crate::host::module_host::{DatabaseTableUpdate, TableOp};
+
+use super::query::{self, run_query, Supported};
+use super::subscription::{eval_primary_updates, IncrementalJoin, SupportedQuery};
+
+/// A hash for uniquely identifying queries.
+/// Its purpose is to avoid recompilation of queries that have an open subscription.
```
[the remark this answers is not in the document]
joshua-spacetime: Mainly to avoid Eq comparisons on QueryExpr in the case of hash collisions.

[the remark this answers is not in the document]
joshua-spacetime: Also forward looking to when we are maintaining multiple QueryExprs for a join query.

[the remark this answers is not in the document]
joshua-spacetime: I will add a comment explaining :)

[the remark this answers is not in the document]
gefjon: Could you add a comment to that effect, which says that you believe blake3 hashing to be faster than `QueryExpr::eq`?

crates/core/src/subscription/module_subscription_actor.rs
joshua-spacetime: Correct, this assumes that compile_read_only_query returns a single element vec. I should either add an assert, or change the type signature.

crates/core/src/subscription/module_subscription_manager.rs
joshua-spacetime: I think you're right. I will remove this alias.

crates/core/src/subscription/module_subscription_actor.rs
gefjon: I would also accept a comment to that effect. Any of the three would be fine, just so long as a reader can tell what's going on.

crates/core/src/subscription/subscription.rs
joshua-spacetime: I'm maintaining a map from TableIds to the queries that read from them. This method is really only relevant for joins where there is a return table and then this filter table, which is the joining table. I will add a clarifying comment here.

PR #918, opened 2024-03-04
Re-land mult-col index selection for queries

Centril:
# Description of Changes

Lands https://github.com/clockworklabs/SpacetimeDB/pull/694 again by making `a != v` not use an index.
Adds a regression test that ensures `a != v` complies even if there is an index.

Also does some drive-by improvements in some other commits. This was beneficial for learning how `IndexArgument` is used.

None.

2

### DIFF
--- a/crates/core/src/sql/compiler.rs
+++ b/crates/core/src/sql/compiler.rs
@@ -283,13 +283,13 @@ mod tests {
     use std::ops::Bound;
 
     fn assert_index_scan(
-        op: Query,
+        op: &Query,
         cols: impl Into<ColList>,
         low_bound: Bound<AlgebraicValue>,
         up_bound: Bound<AlgebraicValue>,
     ) -> TableId {
         if let Query::IndexScan(IndexScan { table, columns, bounds }) = op {
-            assert_eq!(columns, cols.into(), "Columns don't match");
+            assert_eq!(columns, &cols.into(), "Columns don't match");
             assert_eq!(bounds.0, low_bound, "Lower bound don't match");
             assert_eq!(bounds.1, up_bound, "Upper bound don't match");
             table.table_id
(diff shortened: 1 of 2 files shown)


crates/vm/src/expr.rs
```diff
@@ -712,80 +738,313 @@ impl Query {
...
+            rhs,
+        } => ext_field_val(header, lhs, rhs).map(|(f, v)| (op, f, v)),
+        _ => None,
+    }
+}
+
+fn make_index_arg(cmp: OpCmp, columns: &ColList, value: AlgebraicValue) -> IndexColumnOp<'_> {
+    let arg = match cmp {
+        OpCmp::Eq => IndexArgument::Eq { columns, value },
+        OpCmp::NotEq => unreachable!("using a non-indexed scan for NotEq instead"),
```
mamcx, line 792: I think is more clear if says "NotEq have not IndexArgument, filtered-out by caller

jdetter:
Tested against the bots - works just as well as master. I mentioned an issue that I found but the issue is present in master and it is not present in a PR that Josh is intending on landing soon, so I'm good to call this working 👍

jdetter (approved):
Tested against the bots, working 👍

mamcx (approved):
LGTM

crates/core/src/sql/compiler.rs
```diff
@@ -332,6 +332,23 @@ mod tests {
...
+        // Create table [test] with cols [a, b] and index on [b].
+        db.create_table_for_test(
+            "test",
+            &[("a", AlgebraicType::U64), ("b", AlgebraicType::U64)],
+            &[(1.into(), "b")],
+        )?;
+
+        let tx = db.begin_tx();
+        // Should work with any qualified field.
+        assert!(compile_sql(&db, &tx, "select * from test where test.b <> 3").is_ok());
```
joshua-spacetime, line 348: Could you modify this test to be something like
```
select * from test where a = 1 and b <> 3
```
where there is also an index on `a`, so that we make sure we still select the index scan on `a`.

crates/vm/src/expr.rs
```diff
@@ -925,7 +923,15 @@ fn select_best_index<'a>(
 
     // Go through each operator and index,
     // consuming all field constraints that can be served by an index.
-    for (col_list, cmp) in [OpCmp::Eq, OpCmp::NotEq, OpCmp::Lt, OpCmp::LtEq, OpCmp::Gt, OpCmp::GtEq]
+    //
+    // NOTE: We do not consider `OpCmp::NotEq` at the moment.
+    // To provide such an index scan efficiently, we'd need to use two chained index scans
```
joshua-spacetime, line 928: I would remove this comment, since in general, you usually don't answer `!=` conditions using an index.

joshua-spacetime (approved):
I just requested a slight modification to a test case, but otherwise this looks good to go!

PR #919, opened 2024-03-04
(1/3) Commitlog: Base implementation "sans I/O"

kim:
First in a series of patches to implement the new commitlog format.

This patch implements the base format, leaving the transaction payload generic. Segment handling, writing and reading is implemented based on an in-memory backend, which greatly simplifies testing.

As a notable deviation from the previous implementation, segments are never implicitly trimmed. Instead, faulty commits are ignored if and only if the next commit in the log sequence is valid and has the right offset. On the write path, this entails closing the active segment when an (I/O) error occurs, but retaining the commit in memory such that it is written to the next segment.

Note that this patch does not define the final public API.

5

### DIFF
--- a/crates/commitlog/src/commit.rs
+++ b/crates/commitlog/src/commit.rs
@@ -1,0 +1,228 @@
+use std::{
+    io::{self, Read, Write},
+    ops::Range,
+};
+
+use crc32c::{Crc32cReader, Crc32cWriter};
+use spacetimedb_sats::buffer::{BufReader, DecodeError};
+
+use crate::{error::ChecksumMismatch, segment::CHECKSUM_ALGORITHM_CRC32C};
+
+pub struct Header {
+    min_tx_offset: u64,
+    n: u16,
+    len: u32,
+}
+
+impl Header {
+    pub const LEN: usize = /* offset */ 8 + /* n */ 2 + /* len */  4;
+
+    /// Read [`Self::LEN`] bytes from `reader` and interpret them as the
+    /// "header" of a [`Commit`].
+    ///
+    /// Returns `None` if:
+    ///
+    /// - The reader cannot provide exactly [`Self::LEN`] bytes
+    ///
+    ///   I.e. it is at EOF
+    ///
+    /// - Or, the read bytes are all zeroes
+    ///
+    ///   This is to allow preallocation of segments.
+    ///
+    pub fn decode<R: Read>(mut reader: R) -> io::Result<Option<Self>> {
+        let mut hdr = [0; Self::LEN];
+        if let Err(e) = reader.read_exact(&mut hdr) {
+            if e.kind() == io::ErrorKind::UnexpectedEof {
+                return Ok(None);
+            }
+
+            return Err(e);
+        }
+        match &mut hdr.as_slice() {
+            [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0] => Ok(None),
+            buf => {
+                let min_tx_offset = buf.get_u64().map_err(decode_error)?;
+                let n = buf.get_u16().map_err(decode_error)?;
+                let len = buf.get_u32().map_err(decode_error)?;
+
+                Ok(Some(Self { min_tx_offset, n, len }))
+            }
+        }
+    }
+}
+
+/// Entry type of a [`crate::Commitlog`].
+#[derive(Debug, Default, PartialEq)]
+pub struct Commit {
+    /// The offset of the first record in this commit.
+    ///
+    /// The offset starts from zero and is counted from the beginning of the
+    /// entire log.
+    pub min_tx_offset: u64,
+    /// The number of records in the commit.
+    pub n: u16,
+    /// A buffer of all records in the commit in serialized form.
+    ///
+    /// Readers must bring their own [`crate::Decoder`] to interpret this buffer.
+    /// `n` indicates how many records the buffer contains.
+    pub records: Vec<u8>,
+}
+
+impl Commit {
+    pub const FRAMING_LEN: usize = Header::LEN + /* crc32 */ 4;
+    pub const CHECKSUM_ALGORITHM: u8 = CHECKSUM_ALGORITHM_CRC32C;
+
+    /// The range of transaction offsets contained in this commit.
+    pub fn tx_range(&self) -> Range<u64> {
+        self.min_tx_offset..self.min_tx_offset + self.n as u64
+    }
+
+    /// Length in bytes of this commit when written to the log via [`Self::write`].
+    pub fn encoded_len(&self) -> usize {
+        Self::FRAMING_LEN + self.records.len()
+    }
+
+    /// Serialize and write `self` to `out`.
+    pub fn write<W: Write>(&self, out: W) -> io::Result<()> {
+        let mut out = Crc32cWriter::new(out);
+
+        let min_tx_offset = self.min_tx_offset.to_le_bytes();
+        let n = self.n.to_le_bytes();
+        let len = (self.records.len() as u32).to_le_bytes();
+
+        out.write_all(&min_tx_offset)?;
+        out.write_all(&n)?;
+        out.write_all(&len)?;
+        out.write_all(&self.records)?;
+
+        let crc = out.crc32c();
+        let mut out = out.into_inner();
+        out.write_all(&crc.to_le_bytes())?;
+
+        Ok(())
+    }
+
+    /// Attempt to read one [`Commit`] from the given [`Read`]er.
+    ///
+    /// Returns `None` if the reader is already at EOF.
+    ///
+    /// Verifies the checksum of the commit. If it doesn't match, an error of
+    /// kind [`io::ErrorKind::InvalidData`] with an inner error downcastable to
+    /// [`ChecksumMismatch`] is returned.
+    pub fn decode<R: Read>(reader: R) -> io::Result<Option<Self>> {
+        let mut reader = Crc32cReader::new(reader);
+
+        let Some(hdr) = Header::decode(&mut reader)? else {
+            return Ok(None);
+        };
+        let mut records = vec![0; hdr.len as usize];
+        reader.read_exact(&mut records)?;
+
+        let chk = reader.crc32c();
+        let crc = decode_u32(reader.into_inner())?;
+
+        if chk != crc {
+            return Err(invalid_data(ChecksumMismatch));
+        }
+
+        Ok(Some(Self {
+            min_tx_offset: hdr.min_tx_offset,
+            n: hdr.n,
+            records,
+        }))
+    }
+}
+
+/// Numbers needed to compute [`crate::segment::Header`].
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct Metadata {
+    pub tx_range: Range<u64>,
+    pub size_in_bytes: u64,
+}
+
+impl Metadata {
+    /// Extract the [`Metadata`] of a single [`Commit`] from the given reader.
+    ///
+    /// Note that this decodes the commit due to checksum verification.
+    /// Like [`Commit::decode`], returns `None` if the reader is at EOF already.
+    pub fn extract<R: io::Read>(reader: R) -> io::Result<Option<Self>> {
+        Commit::decode(reader).map(|maybe_commit| maybe_commit.map(Self::from))
+    }
+}
+
+impl From<Commit> for Metadata {
+    fn from(commit: Commit) -> Self {
+        Self {
+            tx_range: commit.tx_range(),
+            size_in_bytes: commit.encoded_len() as u64,
+        }
+    }
+}
+
+fn decode_u32<R: Read>(mut read: R) -> io::Result<u32> {
+    let mut buf = [0; 4];
+    read.read_exact(&mut buf)?;
+    Ok(u32::from_le_bytes(buf))
+}
+
+fn decode_error(e: DecodeError) -> io::Error {
+    invalid_data(e)
+}
+
+fn invalid_data<E>(e: E) -> io::Error
+where
+    E: Into<Box<dyn std::error::Error + Send + Sync>>,
+{
+    io::Error::new(io::ErrorKind::InvalidData, e)
+}
+
+#[cfg(test)]
+mod tests {
+    use rand::prelude::*;
+
+    use super::*;
+
+    #[test]
+    fn commit_roundtrip() {
+        let records = vec![0; 128];
+        let commit = Commit {
+            min_tx_offset: 0,
+            n: 3,
+            records,
+        };
+
+        let mut buf = Vec::with_capacity(commit.encoded_len());
+        commit.write(&mut buf).unwrap();
+        let commit2 = Commit::decode(&mut buf.as_slice()).unwrap();
+
+        assert_eq!(Some(commit), commit2);
+    }
+
+    #[test]
+    fn bitflip() {
+        let commit = Commit {
+            min_tx_offset: 42,
+            n: 10,
+            records: vec![1; 512],
+        };
+
+        let mut buf = Vec::with_capacity(commit.encoded_len());
+        commit.write(&mut buf).unwrap();
+
+        let mut rng = thread_rng();
+        let b = buf.choose_mut(&mut rng).unwrap();
+        *b ^= rng.gen::<u8>();
+
+        match Commit::decode(&mut buf.as_slice()) {
+            Err(e) => {
+                assert_eq!(e.kind(), io::ErrorKind::InvalidData);
+                e.into_inner()
+                    .unwrap()
+                    .downcast::<ChecksumMismatch>()
+                    .expect("IO inner should be checksum mismatch");
+            }
+            Ok(commit) => panic!("expected checksum mismatch, got valid commit: {commit:?}"),
+        }
+    }
+}
--- a/crates/commitlog/src/commitlog.rs
+++ b/crates/commitlog/src/commitlog.rs
@@ -1,0 +1,569 @@
+use std::{io, marker::PhantomData, mem, ops::Range, vec};
+
+use itertools::Itertools;
+use log::{debug, info, warn};
+
+use crate::{
+    error,
+    payload::Decoder,
+    repo::{self, Repo},
+    segment::{self, FileLike, Transaction, Writer},
+    Commit, Encode, Options,
+};
+
+#[derive(Debug)]
+pub struct Generic<R: Repo, T> {
+    pub(crate) repo: R,
+    pub(crate) head: Writer<R::Segment>,
+    tail: Vec<u64>,
+    opts: Options,
+    _record: PhantomData<T>,
+}
+
+impl<R: Repo, T> Generic<R, T> {
+    pub fn open(repo: R, opts: Options) -> io::Result<Self> {
+        let mut tail = repo.existing_offsets()?;
+        if !tail.is_empty() {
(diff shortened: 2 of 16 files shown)


kim:
If so desired, this patch could be split further into (perhaps) write path, read path, tests. I felt like the tests might be quite helpful to see, so both paths needed to be included.

gefjon (approved):
Probably don't merge this without Tyler's review, but I'm hitting the button to signify that I've read the code, I feel like I understand it, and I'm comfortable with it.

One small nit about a use of `Option<T>` where I feel `Result<(), T>` is more idiomatic.

This is really nice, clear code, and was very easy to review. Thanks!

crates/commitlog/src/repo/mod.rs
```diff
@@ -0,0 +1,153 @@
...
+/// If the traversal is successful, the segment header is checked against the
+/// `max_log_format_version`, and [`io::ErrorKind::InvalidData`] is returned if
+/// the segment's log format version is greater than the given value. Likewise
+/// if the checksum algorithm stored in the segment header cannot be handled
+/// by this crate.
+///
+/// If only a (non-empty) prefix of the segment could be read due to a failure
+/// to decode a [`Commit`], the segment [`Metadata`] read up to the faulty
+/// commit is returned in an `Err`. In this case, a new segment should be
+/// created for writing.
```
kazimuth, line 100: Should the segment also be truncated? Due to the language in the proposal [here](https://github.com/clockworklabs/SpacetimeDBPrivate/blob/master/proposals/0003-commitlog-format.md#log-integrity)

crates/commitlog/src/commit.rs
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    /// Attempt to read one [`Commit`] from the given [`Read`]er.
+    ///
+    /// Returns `None` if the reader is already at EOF.
+    ///
+    /// Verifies the checksum of the commit. If it doesn't match, an error of
+    /// kind [`io::ErrorKind::InvalidData`] with an inner error downcastable to
+    /// [`ChecksumMismatch`] is returned.
+    pub fn decode<R: Read>(reader: R) -> io::Result<Option<Self>> {
```
kazimuth, line 110: We could try to support reading into a user-supplied buffer / zero-copy reads with mmap in the future, but that should be left for later.

crates/commitlog/src/payload.rs
```diff
@@ -0,0 +1,55 @@
...
+
+pub struct ArrayDecoder<const N: usize>;
+
+impl<const N: usize> Decoder for ArrayDecoder<N> {
+    type Record = [u8; N];
+
+    fn decode_record<'a, R: BufReader<'a>>(&self, _version: u8, reader: &mut R) -> Result<Self::Record, DecodeError> {
+        reader.get_array()
+    }
+}
```
kazimuth, line 55: if only all serialization was this easy

kazimuth (approved):
This looks good. I have one question about how errors at the end of segments are handled, aside from that I'm happy.

crates/commitlog/src/repo/mod.rs
kim: Good catch, I did indeed deviate from this approach. Probably should submit an amendment to the spec.

I realized that, unless `write(2)` can lie and return an error even though it has written everything, we can fairly easily skip over partial data (not least thanks to checksumming).

Because a new segment is started when a write error occurs, we'll retain data potentially useful for forensics in case of a bad disk.

Explicit truncation is still needed for consensus-based replication, though.

crates/commitlog/src/commit.rs
kim: True. Gotta get some benchmarks in place...

crates/commitlog/src/repo/mod.rs
kazimuth: Made an issue for amending the spec, no rush on that though.

cloutiertyler: FWIW Kim now that I think about it

> unless write(2) can lie and return an error even though it has written everything

I think this may be possible. write(2) may return after data is written to a kernel buffer, but *before* it is actually written to the disk. Thus it would be possible to write(2) a commit, it return no error, only partially write that commit to disk, have a disk-write-error, and then have us write another commit to disk without us ever seeing an error.

cloutiertyler: It's possible I'm missing some other context like, we check that all of the commit offsets are sequential or something (and if not we truncate)

cloutiertyler: Sync'd with Kim. We are indeed checking for sequential commit numbers when reloading a database. Thus if we encounter this type of situation, it will require manual intervention to recover from (if that is even possible). This error must have occurred silently as illustrated in my scenario above, so there's nothing we could have really done to detect it when it happened anyway.

Could we expand the comment above to add the stipulation that this behavior of skipping corrupted messages is only valid if sequential ordering of commits is verified at startup, and provide my example as an illustration as why that is necessary.

kim: > but it can lie in this way:
> 
> > write(2) can return no error even though it _hasn't_ written everything

Wait... we cannot assume `write(2)` is behaving arbitrarily.
Namely, if we assume that write calls can be reordered arbitrarily, an append-only log is impossible.
We also must assume that `fsync(2)` won't return successfully if it hasn't flushed to disk.

What is indeed not correctly handled here is that an `fsync(2)` error should cause the active segment to be discarded. And because everything is happening asynchronously, failure to open a fresh segment must result in a panic (i.e. we must not keep writing to the old segment).

I'll make a new patch implementing this.

> Could we expand the comment above

I don't think a module which is not even exported from the crate is the right place to explain how different parts of the crate need to play together. Once we're overall confident with the implementation, I'd rather add some extensive crate-level docs, also summarizing the spec which is not accessible for everyone.

crates/commitlog/src/commit.rs
```diff
@@ -0,0 +1,233 @@
...
+/// Entry type of a [`crate::Commitlog`].
+#[derive(Debug, Default, PartialEq)]
+pub struct Commit {
+    /// The offset of the first record in this commit.
+    ///
+    /// The offset starts from zero and is counted from the beginning of the
+    /// entire log.
+    pub min_tx_offset: u64,
+    /// The number of records in the commit.
+    pub n: u16,
```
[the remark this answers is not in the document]
kim: It's `n` in the spec, and so it shall be `n` in the code.

crates/commitlog/src/commitlog.rs
```diff
@@ -0,0 +1,574 @@
...
+        self.tail.reserve(1);
+        self.tail.push(self.head.min_tx_offset);
+        for segment in self.tail.iter().rev() {
+            self.repo.remove_segment(*segment)?;
+        }
+
+        Self::open(self.repo.clone(), self.opts)
+    }
+
+    pub fn reset_to(mut self, offset: u64) -> io::Result<Self> {
```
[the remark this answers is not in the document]
kim: I'm deliberately not documenting private items which either

- Are trivial ("method foo makes a foo")
- Are documented as part of the API, and so would just repeat the text there
- Are merely "helpers" with no particular invariants to document

If you feel otherwise, for example because it aids your understanding, then please submit suggestions. Note also that I'll ask you to motivate why docstrings are needed in those instances.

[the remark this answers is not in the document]
kim: I'll resolve all "Docs :)" comments accordingly.

crates/commitlog/src/commit.rs
```diff
@@ -0,0 +1,233 @@
+use std::io::{self, Read, Write};
+
+use crc32c::{Crc32cReader, Crc32cWriter};
+use spacetimedb_sats::buffer::{BufReader, DecodeError};
+
+use crate::{error::ChecksumMismatch, segment::CHECKSUM_ALGORITHM_CRC32C};
+
+pub struct Header {
+    min_tx_offset: u64,
+    n: u16,
```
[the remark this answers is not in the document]
kim: As per spec.

crates/commitlog/src/payload.rs
```diff
@@ -0,0 +1,55 @@
+use std::sync::Arc;
+
+use spacetimedb_sats::buffer::{BufReader, BufWriter, DecodeError};
+
+/// A **datatype** which can be encoded.
+///
+/// The transaction payload of the commitlog (i.e. individual records in the log)
```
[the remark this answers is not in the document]
kim: What exactly is unclear? Familiarity with the format is obviously assumed: the log is composed of commits, which are composed of records, which we also refer to as "transactions" or "transaction payload". Each one of those must be [`Encode`]-able.

crates/commitlog/src/commitlog.rs
```diff
@@ -0,0 +1,574 @@
...
+
+    // Helper to obtain a list of the segment offsets which include transaction
+    // offset `offset`.
+    //
+    // The returned `Vec` is sorted in **ascending** order, such that the first
+    // element is the segment which contains `offset`.
+    //
+    // The offset of `self.head` is always included, regardless of how many
+    // entries it actually contains.
+    fn segment_offsets_from(&self, offset: u64) -> Vec<u64> {
```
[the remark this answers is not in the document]
kim: I don't think -- the `Commits` iterator must not borrow from self, because we don't want to hold a lock while iterating.

crates/commitlog/src/commitlog.rs
```diff
@@ -0,0 +1,574 @@
...
+                offs.push(self.head.min_tx_offset);
+            }
+
+            offs
+        }
+    }
+
+    pub fn commits_from(&self, offset: u64) -> Commits<R> {
+        let offsets = self.segment_offsets_from(offset);
+        let last_offset = offsets.first().cloned().unwrap_or(offset);
```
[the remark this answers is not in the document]
kim: See above

crates/commitlog/src/commit.rs
```diff
@@ -0,0 +1,233 @@
...
+
+#[cfg(test)]
+mod tests {
+    use rand::prelude::*;
+
+    use super::*;
+
+    #[test]
+    fn commit_roundtrip() {
+        let records = vec![0; 128];
```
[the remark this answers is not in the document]
kim: Why?

crates/commitlog/src/commit.rs
```diff
@@ -0,0 +1,233 @@
...
+        let min_tx_offset = self.min_tx_offset.to_le_bytes();
+        let n = self.n.to_le_bytes();
+        let len = (self.records.len() as u32).to_le_bytes();
+
+        out.write_all(&min_tx_offset)?;
+        out.write_all(&n)?;
+        out.write_all(&len)?;
+        out.write_all(&self.records)?;
+
+        let crc = out.crc32c();
```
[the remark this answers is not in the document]
kim: It is already computed. What does this comment add?

crates/commitlog/src/commit.rs
```diff
@@ -0,0 +1,233 @@
...
+        self.min_tx_offset + self.n as u64
+    }
+
+    /// Length in bytes of this commit when written to the log via [`Self::write`].
+    pub fn encoded_len(&self) -> usize {
+        Self::FRAMING_LEN + self.records.len()
+    }
+
+    /// Serialize and write `self` to `out`.
+    pub fn write<W: Write>(&self, out: W) -> io::Result<()> {
```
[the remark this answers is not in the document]
kim: Uhm... I'm really not a fan of this. Trait bounds, except maybe for trivial ones like `impl AsRef<str>`, tend to become longer over time, which either makes this syntax _less_ readable or creates larger diffs.

crates/commitlog/src/commitlog.rs
```diff
@@ -0,0 +1,574 @@
...
+    pub fn sync(&self) -> io::Result<()> {
+        self.head.fsync()
+    }
+
+    /// The last transaction offset written to disk, or `None` if nothing has
+    /// been written yet.
+    ///
+    /// Note that this does not imply durability: [`Self::sync`] may not have
+    /// been called at this offset.
+    pub fn max_committed_offset(&self) -> Option<u64> {
```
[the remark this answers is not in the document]
kim: `Some(0)` is valid and signifies that we know we have written the very first commit in the log.

crates/commitlog/src/commit.rs
```diff
@@ -0,0 +1,233 @@
...
+        Commit {
+            min_tx_offset,
+            n,
+            records,
+        }: Commit,
+    ) -> Self {
+        Self {
+            min_tx_offset,
+            max_tx_offset: min_tx_offset + n as u64,
+            size_in_bytes: Header::LEN as u64 + records.len() as u64 + /* crc32 */ 4,
```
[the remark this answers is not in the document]
kim: #985

crates/commitlog/src/commitlog.rs
```diff
@@ -0,0 +1,574 @@
...
+};
+
+#[derive(Debug)]
+pub struct Generic<R: Repo, T> {
+    pub(crate) repo: R,
+    pub(crate) head: Writer<R::Segment>,
+    tail: Vec<u64>,
+    opts: Options,
+    _record: PhantomData<T>,
+}
```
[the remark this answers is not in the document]
kim: #985

crates/commitlog/src/segment.rs
```diff
@@ -0,0 +1,554 @@
...
+pub const CHECKSUM_ALGORITHM_CRC32C: u8 = 0;
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct Header {
+    pub log_format_version: u8,
+    pub checksum_algorithm: u8,
+}
+
+impl Header {
+    pub const LEN: usize = MAGIC.len() + 4;
```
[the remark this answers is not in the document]
kim: #985

crates/commitlog/src/commitlog.rs
```diff
@@ -0,0 +1,574 @@
...
+        if let Err(e) = writer.commit() {
+            warn!("Commit failed: {e}");
+            self.start_new_segment()?;
+            Err(e)
+        } else {
+            Ok(sz)
+        }
+    }
+
+    pub fn sync(&self) -> io::Result<()> {
```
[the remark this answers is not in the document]
kim: #985

crates/commitlog/src/repo/mod.rs
cloutiertyler: This seems fine to me, as long as we don't forget to add that comment in at the end.

crates/commitlog/src/commit.rs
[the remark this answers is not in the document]
cloutiertyler: I'm going to resolve this without applying because it's just a nit.

crates/commitlog/src/commit.rs
```diff
@@ -0,0 +1,234 @@
...
+    fn from(
+        Commit {
+            min_tx_offset,
+            n,
+            records,
+        }: Commit,
+    ) -> Self {
+        Self {
+            min_tx_offset,
+            max_tx_offset: min_tx_offset + n as u64,
```
cloutiertyler, line 162: I could have sworn I wrote this somewhere before, but I can't find it. This seems wrong to me. If the `min_tx_offset` is 0 and there are 5 things in the list, then the `max_tx_offset` should be 4, not 5.

cloutiertyler: I'm a little unclear on the purpose of the metadata type generally, since you could just compute from the commit, but I'll reserve judgement for later.

crates/commitlog/src/commit.rs
```diff
@@ -0,0 +1,234 @@
...
+            size_in_bytes: Header::LEN as u64 + records.len() as u64 + /* crc32 */ 4,
```
cloutiertyler, line 163: Shouldn't this just be a function on the `Commit`?

crates/commitlog/src/commit.rs
```diff
@@ -0,0 +1,234 @@
...
+    /// Readers must bring their own [`crate::Decoder`] to interpret this buffer.
+    /// `n` indicates how many records the buffer contains.
+    pub records: Vec<u8>,
+}
+
+impl Commit {
+    pub const FRAMING_LEN: usize = Header::LEN + /* crc32 */ 4;
+    pub const CHECKSUM_ALGORITHM: u8 = CHECKSUM_ALGORITHM_CRC32C;
+
+    /// The largest transaction offset in this commit.
```
cloutiertyler, line 73: This comment is wrong AFAICT. This is the min_tx_offset of the next segment, not the max_tx_offset of this segment.

crates/commitlog/src/segment.rs
```diff
@@ -0,0 +1,554 @@
...
+        self.bytes_written <= Header::LEN as u64
+    }
+
+    /// Number of bytes written to this segment, including the header.
+    pub fn len(&self) -> u64 {
+        self.bytes_written
+    }
+}
+
+pub trait FileLike {
```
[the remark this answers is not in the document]
cloutiertyler: Not sure if you discussed with @Centril, but it would be nice to explain the motivation of this trait.

cloutiertyler (changes requested):
My biggest concern with this PR is that I think that `max_tx_offset` is improperly calculated. I thought I had already written this a few weeks ago, but I can't find the comment. I may have not submitted it accidentally.

I think this some naming and comments could be cleaned up a bit, but happy to defer that until later.

crates/commitlog/src/commit.rs
kim: The offsets are zero-based, which means that the max offset of one commit is equal to the min offset of the next commit. Because it has a potential to trip up the reader, this is spelled out explicitly in the spec.

Also note that the metadata types are merely internal helpers. You know, like semigroups.

crates/commitlog/src/commit.rs
kim: #985

crates/commitlog/src/commit.rs
kim: Both is true. But if its a method on commit, then it's the max offset of this commit.

crates/commitlog/src/commit.rs
cloutiertyler: I realize that they are zero based, that's why I'm making this comment. The maximum offset in this commit cannot be equal to the minimum offset in the next commit because offsets are unique.

The max offset cannot exist in this commit. If you mean something like next offset, I strongly suggest calling it that instead. This number as calculated does not represent the maximum transaction offset in this commit. There is no transaction with that offset in this commit.

cloutiertyler: Consider calling these upper and lower bound if what you mean to specify here is a half open range of tx offsets.

crates/commitlog/src/repo/mod.rs
kim: The commentary states what happens, namely that one cannot resume a segment when it contains data which cannot be decoded as a commit. A new segment is created then, from the offset of the last good commit.

This can go wrong due to caching effects: re-reading from disk might actually be reading from the page cache. There is no remedy for this, but the next best thing is to bypass the page cache.

I don't know what else should be added here.

crates/commitlog/src/commit.rs
kim: I changed this to be represented by a `Range<u64>` throughout the series.

cloutiertyler (approved):
This LGTM

PR #920, opened 2024-03-04
(2/3) Commitlog: Add I/O based on regular files

kim:
Provides a commitlog backing store based on files, and defines the exported `Commitlog` type which fixes the store to the file-based one.

**NOTE: This PR is stacked on top of #919**

1

### DIFF
--- a/crates/commitlog/src/lib.rs
+++ b/crates/commitlog/src/lib.rs
@@ -1,6 +1,17 @@
-#![allow(unused)]
+use std::{
+    io,
+    num::NonZeroU16,
+    path::PathBuf,
+    sync::{Arc, RwLock},
+    time::Duration,
+};
 
-use std::num::NonZeroU16;
+use log::debug;
+use tokio::{
+    sync::watch,
+    task::spawn_blocking,
+    time::{interval, MissedTickBehavior},
+};
 
 mod commit;
 mod commitlog;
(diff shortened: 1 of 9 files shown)


crates/commitlog/src/lib.rs
```diff
@@ -51,3 +64,250 @@ impl Default for Options {
...
+    ///
+    /// The returned iterator is not aware of segment rotation. That is, if a
+    /// new segment is created after this method returns, the iterator will not
+    /// traverse it.
+    ///
+    /// Commits appended to the log while it is being traversed are generally
+    /// visible to the iterator. Upon encountering [`io::ErrorKind::UnexpectedEof`],
+    /// however, a new iterator should be created using [`Self::commits_from`]
+    /// with the last transaction offset yielded.
+    pub fn commits(&self) -> impl Iterator<Item = Result<Commit, error::Traversal>> {
```
kim, line 147: slash-slash-todo: We're skipping bad commits in case the next commit is good. It's not obvious what to do when the iterator is exhausted and the last commit is bad -- there could be more coming at a later point in time. So for now, we'll yield the error.

This is fine in tooling, but requires special-casing when replaying the log to fold it into a datastore. Maybe a named iterator which can be turned into one which just ignores the error if it's the last item could be handy.

crates/commitlog/src/repo/fs.rs
```diff
@@ -0,0 +1,97 @@
...
+};
+
+use log::debug;
+
+use super::Repo;
+
+const SEGMENT_FILE_EXT: &str = ".stdb.log";
+
+pub fn segment_file_name(offset: u64) -> String {
+    format!("{offset:0>20}{SEGMENT_FILE_EXT}")
```
gefjon, line 14:
```suggestion
    // Left-pad the offset with zeros to 20 digits.
    format!("{offset:0>20}{SEGMENT_FILE_EXT}")
```

I can never remember the Rust format language, so anywhere we use a more-elaborate format specifier than `{}` or `{:?}`, I like a quick comment describing what it does.

crates/commitlog/src/repo/fs.rs
```diff
@@ -0,0 +1,97 @@
...
+//
+// Experiment:
+//
+// - O_DIRECT | O_DSYNC
+// - preallocation of disk space
+// - io_uring
+//
+
+#[derive(Clone, Debug)]
+pub struct Fs {
```
gefjon, line 29:
```suggestion
/// A commitlog repository [`Repo`] which stores commits in ordinary files on disk.
pub struct Fs {
```

crates/commitlog/src/repo/fs.rs
```diff
@@ -0,0 +1,97 @@
...
+    root: PathBuf,
```
gefjon, line 30:
```suggestion
    /// The base directory within which segment files will be stored.
    root: PathBuf,
```

crates/commitlog/src/repo/fs.rs
```diff
@@ -0,0 +1,97 @@
...
+}
+
+impl Fs {
+    pub fn new(root: impl Into<PathBuf>) -> Self {
```
gefjon, line 34:
```suggestion
    /// Create a commitlog repository which stores segments in the directory `root`.
    ///
    /// `root` must name an extant, accessible, writeable directory.
    pub fn new(root: impl Into<PathBuf>) -> Self {
```

Is this accurate?

crates/commitlog/src/repo/fs.rs
```diff
@@ -0,0 +1,97 @@
...
+        Self { root: root.into() }
+    }
+
+    pub fn segment_path(&self, offset: u64) -> PathBuf {
```
gefjon, line 38:
```suggestion
    /// Get the filename for a segment starting with `offset` within this repository.
    pub fn segment_path(&self, offset: u64) -> PathBuf {
```

gefjon (approved):
Looks great! This does exactly what I expected it to do based on the previous PR and the proposal. I suggested a few minor comments in `fs.rs`.

crates/commitlog/src/repo/fs.rs
kim: I had to look up "extant"

crates/commitlog/src/lib.rs
kim: I settled on yielding the error in iterators, but ignore it in folds (because the consumer of a fold doesn't have access to the traversal itself).

kazimuth (approved):
Nice and straightforward.

In the future it might be nice to try to run the existing tests against the file system Repo implementation as well. I don't think we need to do that right now though.

crates/commitlog/src/repo/fs.rs
```diff
@@ -0,0 +1,108 @@
...
+    fs::{self, File},
+    path::PathBuf,
+};
+
+use log::debug;
+
+use super::Repo;
+
+const SEGMENT_FILE_EXT: &str = ".stdb.log";
+
```
mamcx, line 12: `.log` extension could be confused easily as being text-based, how about `.log.stdb` or `.stdb.wbl` (write behind log)

kim: This file extension was requested by @Centril. If we don’t like it, an amendment to the spec shall be proposed.

crates/commitlog/src/lib.rs
```diff
@@ -53,3 +64,258 @@ impl Default for Options {
...
+        let inner = self.inner.read().unwrap();
+        debug!("sync commitlog");
+        inner.sync()?;
+
+        Ok(inner.max_committed_offset())
+    }
+
+    /// Write all outstanding transaction records to disk.
+    ///
+    /// Note that this does **not** force the OS to flush the data to the
```
cloutiertyler, line 107:
```suggestion
    /// Note that this does **not** force the OS to sync the data to the
```

crates/commitlog/src/lib.rs
```diff
@@ -53,3 +64,258 @@ impl Default for Options {
...
+    /// filesystem. Use [`Self::flush_and_sync`] or call [`Self::sync`] after
```
cloutiertyler, line 108:
```suggestion
    /// disk. Use [`Self::flush_and_sync`] or call [`Self::sync`] after
```

crates/commitlog/src/repo/fs.rs
cloutiertyler: .log is pretty standard. I think it's the same for kafka IIRC.

cloutiertyler (approved):
Seems fine given the previous PR.

PR #921, opened 2024-03-04
(3/3) Commitlog: Add canonical txdata payload

kim:
Defines the canonical commitlog payload.

Note that this may require some adjustments comes integration time. It is also worth considering to move the definitions to the core crate.

**NOTE: This PR is stacked on top of #920**

2


crates/commitlog/src/varint.rs
```diff
@@ -0,0 +1,47 @@
+//! Varint encoding and decoding functions.
```
gefjon, line 1:
```suggestion
//! Varint encoding and decoding functions.
//!
//! We use [Protobuf's Base-128 varint encoding](https://protobuf.dev/programming-guides/encoding/#varints).
//!
//! Unsigned integers are split into 7-bit chunks, with the least significant chunk first.
//! Each chunk is placed in the low 7 bits of a byte.
//! Non-terminal bytes have the high bit set.
//! The final byte in an integer has the high bit zeroed.
```

crates/commitlog/src/varchar.rs
```diff
@@ -0,0 +1,164 @@
...
+        } else {
+            let mut s = s;
+            while s.len() > N {
+                s.pop().unwrap();
+            }
+            Self { inner: s }
+        }
+    }
+
+    /// Discard the `Varchar` wrapper.`
```
gefjon, line 59:
```suggestion
    /// Discard the `Varchar` wrapper.
```

crates/commitlog/src/payload.rs
```diff
@@ -38,6 +52,20 @@ pub trait Decoder {
     // `decode_record` call.
 }
 
+/// A decoder which can decode one [`Ops::rowdata`] value, given the
+/// corresponding [`TableId`].
+///
+/// This is only relevant for the canonical payload [`Txdata`], and allows its
+/// [`Decoder`] [`TxdataDecoder`] to be defined in this crate (as opposed to in
```
gefjon, line 59:
```suggestion
/// [`Decoder`], [`TxdataDecoder`], to be defined in this crate (as opposed to in
```

crates/commitlog/src/payload.rs
```diff
@@ -53,3 +81,383 @@ impl<const N: usize> Decoder for ArrayDecoder<N> {
...
+    // module info.
+    //
+    // At that point, we can also remove the length prefix of
+    // `Inputs`.
+}
+
+impl Inputs {
+    pub fn encode(&self, buf: &mut impl BufWriter) {
+        let slen = self.reducer_name.len() as u8;
+        let len = /* slen */ 1 + slen as usize + self.reducer_args.len();
```
gefjon, line 171: What's the deal with the `/* slen */`? Is that related to the above comment about removing the length prefix?

crates/commitlog/src/payload.rs
```diff
@@ -53,3 +81,383 @@ impl<const N: usize> Decoder for ArrayDecoder<N> {
...
+    pub fn decode<'a, R: BufReader<'a>>(reader: &mut R) -> Result<Self, DecodeError> {
+        let len = reader.get_u32()?;
+        let slen = reader.get_u8()?;
+        let reducer_name = {
+            let bytes = reader.get_slice(slen as usize)?;
+            Varchar::from_str(std::str::from_utf8(bytes)?)
+                .expect("slice len cannot be > 255")
+                .into()
+        };
+        let reducer_args = reader.get_slice(len as usize - /* slen */ 1 - slen as usize)?.into();
```
gefjon, line 187: Same question re: `/* slen */`.

gefjon (commented):
One question about some commented-out code, and some suggested comments.

crates/commitlog/src/payload.rs
kim: It’s the one byte length of the string, mainly `reducer_name`. Which led me to introducing `Varchar` so that encoding remains infallible.

We could instead do one of the following:

- Use a varint
- Mandate that traversals must bring a way to resolve the type of the reducer args, so that the name can come after the args and we don’t need to store the length 
- Derive a stable numerical reducer id, which may however only be stable for the lifetime of the module revision

kim: We also likely want to settle on some refcounted, small string representation in core, so we can avoid copying the same strings over and over again.

crates/commitlog/src/payload.rs
```diff
@@ -53,3 +81,383 @@ impl<const N: usize> Decoder for ArrayDecoder<N> {
...
+        for TableId(table_id) in self.truncates.iter() {
+            buf.put_u32(*table_id);
+        }
+    }
+
+    /// Decode `Self` from the given buffer.
+    ///
+    /// A [`RowDecoder`] is supplied to allow to determine the row types of
+    /// individual [`Ops::rowdata`] rows.
+    pub fn decode<'a, D, R>(row_decoder: &D, reader: &mut R) -> Result<Self, DecodeError>
```
kim, line 262: Turns out this won’t work in practice because ops can depend on schema modifications within the same transaction. Which means that the row decoder would need to update its internal state after each decoder row.

cloutiertyler: I thought we decided that in the case of a schema mutation, all rows in the transactions would be converted to the target schema before saving:

https://github.com/clockworklabs/SpacetimeDBPrivate/pull/718/files#diff-84260e38cfce8e0fba4beb7475cba661892f4a1a66b669db5d49830cc3d43c11R388

kim: This has been re-implemented using a visitor pattern, which can be optimized further to not allocate the `Txdata` struct at all when replaying (aka folding) the history.

kazimuth (approved):
LGTM. I like the proptests.

crates/commitlog/src/varchar.rs
```diff
@@ -0,0 +1,166 @@
+use std::ops::Deref;
+
+/// An owned string with a max length of `N`, like the venerable VARCHAR.
```
cloutiertyler, line 3: I assume this is to implement the `reducer_name` and `outputs` fields as per the proposal? I should have asked this at proposal time, but why not just a string of unbounded length?

cloutiertyler (approved):
This seems fine to me. If I had one critique of this code, it's that it is IMO overly generic. It adds quite a bit of complexity to reading the code, with unclear benefit. 

e.g. Do we really need `Transaction` to be generic over the `Decoder::Record` type where `Record` is any type that impls `Encode` which is only impl'd for `ProductValue`? Why not keep it concrete with `ProductValue` for now and make it generic as needed?

kim:
> If I had one critique of this code, it's that it is IMO overly generic. It adds quite a bit of complexity to reading the code, with unclear benefit.

Unsure if this is what you're referring to, but I don't see how else the self-referential schema resolution could be implemented. Leaving the type of mutation rows generic doesn't make the code more or less complex, but it adds the benefit that optimizing away the `ProductValue` intermediate representation can be done incrementally:

It is not clear at this point how `ProductValue` can be avoided on the write path, but when replaying (folding) it is just overhead. The generic parameter allows to use `TxData<ProductValue>` for writing and `TxData<()>` for folding, so that we can e.g. replace decoding for the latter with a non-allocating version and perhaps later with a direct BSATN->BFLATN conversion, all without changing the `commitlog` crate.

crates/commitlog/src/varchar.rs
kim: The `Varchar` type exists because serialization must be infallible -- if that was not the case, a database could be rendered unusable as there is no way to recover from invalid input in a transaction. 

Logging inputs will create huge amounts of repeated data, so there needs to be _some_ limit.

Forcing the caller to prepare data within constraints helps to avoid runtime surprises, and even allows to reject modules early. The slight inconvenience also serves as a nag to come up with a fixed-size stable identifier for the reducer name down the line.

PR #922, opened 2024-03-04
Durability: Traits and implementation in terms of commitlog

kim:
Defines traits intended to abstract over the kind of persistence a
database utilizes. The only implementation is (host-)local durability in
terms of the new commitlog crate.

The trait definitions may not be considered stable yet, but are in their
tentative form needed for further integration of the new commitlog.

**NOTE: This PR is stacked on top of #921**

### DIFF
--- a/crates/durability/src/imp/local.rs
+++ b/crates/durability/src/imp/local.rs
@@ -1,0 +1,303 @@
+use std::{
+    io,
+    num::NonZeroU16,
+    panic,
+    path::PathBuf,
+    sync::{
+        atomic::{
+            AtomicI64, AtomicU64,
+            Ordering::{Acquire, Relaxed, Release},
+        },
+        Arc,
+    },
+    time::Duration,
+};
+
+use anyhow::Context as _;
+use log::{info, trace, warn};
+use spacetimedb_commitlog::{error, payload::Txdata, Commit, Commitlog, Decoder, Encode, Transaction};
+use tokio::{
+    sync::mpsc,
+    task::{spawn_blocking, JoinHandle},
+};
+use tracing::instrument;
+
+use crate::{Durability, History, TxOffset};
+
+/// [`Local`] configuration.
+#[derive(Clone, Copy, Debug)]
+pub struct Options {
+    /// Periodically flush and sync the log this often.
+    ///
+    /// Default: 500ms
+    pub sync_interval: Duration,
+    /// [`Commitlog`] configuration.
+    pub commitlog: spacetimedb_commitlog::Options,
+}
+
+impl Default for Options {
+    fn default() -> Self {
+        Self {
+            sync_interval: Duration::from_millis(500),
+            commitlog: Default::default(),
+        }
+    }
+}
+
+/// [`Durability`] implementation backed by a [`Commitlog`] on local storage.
+///
+/// The commitlog is constrained to store the canonical [`Txdata`] payload,
+/// where the generic parameter `T` is the type of the row data stored in
+/// the mutations section.
+///
+/// `T` is left generic in order to allow bypassing the `ProductValue`
+/// intermediate representation in the future.
+///
+/// Note, however, that instantiating `T` to a different type may require to
+/// change the log format version!
+pub struct Local<T> {
+    /// The [`Commitlog`] this [`Durability`] and [`History`] impl wraps.
+    clog: Arc<Commitlog<Txdata<T>>>,
+    /// The durable transaction offset, as reported by the background
+    /// [`FlushAndSyncTask`].
+    ///
+    /// A negative number indicates that we haven't flushed yet, or that the
+    /// number overflowed. In either case, appending new transactions shall panic.
+    ///
+    /// The offset will be used by the datastore to squash durable transactions
+    /// into the committed state, thereby making them visible to durable-only
+    /// readers.
+    ///
+    /// We don't want to hang on to those transactions longer than needed, so
+    /// acquire / release or stronger should be used to prevent stale reads.
+    durable_offset: Arc<AtomicI64>,
+    /// Backlog of transactions to be written to disk by the background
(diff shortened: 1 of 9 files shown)


crates/durability/src/imp/local.rs
```diff
@@ -0,0 +1,277 @@
...
+/// `T` is left generic in order to allow bypassing the `ProductValue`
+/// intermediate representation in the future.
+///
+/// Note, however, that instantiating `T` to a different type may require to
+/// change the log format version!
+pub struct Local<T> {
+    clog: Arc<Commitlog<Txdata<T>>>,
+    offset: Arc<AtomicI64>,
+    queue: mpsc::UnboundedSender<Txdata<T>>,
+    queue_depth: Arc<AtomicU64>,
```
gefjon, line 62: Could you add comments to the atomic members here describing what (if anything) they're used to synchronize, and what `Ordering` accomplishes that? It feels intuitive to me that `queue_depth` is just a shared counter for diagnostic/reporting purposes, and therefore can be `Relaxed`, but it's not obvious to me why the `offset` needs to be `SeqCst`.

gefjon (approved):

crates/durability/src/imp/local.rs
kim: The offset would be used by the datastore to squash transactions which became durable, and don’t need to be hidden from readers which requested durable-only data.

It seemed reasonable to keep this strongly ordered so as to not hang on to data longer than necessary.

But, I can be convinced of this being flawed reasoning. Wdyt?

kim: Updated to acquire / release and added commentary.

kazimuth (approved):
👍

crates/durability/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+    ///
+    /// This method must never block, and accept new transactions even if they
+    /// cannot be made durable immediately.
+    ///
+    /// A permanent failure of the durable storage may be signalled by panicking.
+    fn append_tx(&self, tx: Self::TxData);
+
+    /// The [`TxOffset`] considered durable.
+    ///
+    /// A `None` return value indicates that the durable offset is not known,
```
cloutiertyler, line 41: Do we want to return `Result<Option<TxOffset>, Error>` to distinguish between these two cases?

crates/durability/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+    /// either because nothing has been persisted yet, or because the status
+    /// cannot be retrieved.
+    fn durable_tx_offset(&self) -> Option<TxOffset>;
```
cloutiertyler, line 44: We could consider adding an observer pattern here if we feel we need to be notified immediately whenever the `durable_tx_offset` changes. Depends on the implementation. I pulled it out of Harald's thing because it was overly complicated and OmniPaxos works by polling anyway.

crates/durability/src/imp/local.rs
```diff
@@ -0,0 +1,300 @@
...
+    ///
+    /// The `root` directory must already exist.
+    ///
+    /// Background tasks are spawned onto the provided tokio runtime.
+    pub fn open(root: impl Into<PathBuf>, rt: tokio::runtime::Handle, opts: Options) -> io::Result<Self> {
+        info!("open local durability");
+
+        let clog = Arc::new(Commitlog::open(root, opts.commitlog)?);
+        let (queue, rx) = mpsc::unbounded_channel();
+        let queue_depth = Arc::new(AtomicU64::new(0));
```
cloutiertyler, line 100: Do you need to `Arc` it if it's already `Atomic`? I guess so, otherwise you need a `Mutex` for mutability.

crates/durability/src/imp/local.rs
```diff
@@ -0,0 +1,300 @@
...
+        let offset = Arc::new(AtomicI64::new(-1));
```
cloutiertyler, line 101: Shouldn't we be initializing this based on the clog's offset on disk after opening rather than `-1`? Am I misunderstanding?

cloutiertyler (commented):
This looks great to me. Just one important comment about initializing the durable offset.

crates/durability/src/imp/local.rs
kim: Yes we need the counter to be `Send`

crates/durability/src/lib.rs
kim: The thing is that there is no recoverable error -- at all, actually. Either things converge eventually, or the commitlog becomes unwriteable, in which case we panic on the next write. We could try to crash the datastore earlier if that's desired.

crates/durability/src/imp/local.rs
kim: Yeah can do.

I'll note, however, that it is basically non-sensical to call `durable_tx_offset` before anything was written by the owner of the `Durability`: we already know the offset from replaying the `History` (which, n.b., does not need to be the same source as the `Durability`).

Things would look a little different if `append_tx` would take the transaction offset, but then the question would be: what happens if one tries to `append_tx` with an offset that is _not_ `n + 1`?

Unfortunately, we have to keep track of the offset on both sides -- datastore and log.

crates/durability/src/imp/local.rs
```diff
@@ -0,0 +1,303 @@
...
+                    flush_error(source);
+                    retry = Some(txdata);
+                }
+            }
+
+            if flush_after {
+                clog.flush().map(drop).unwrap_or_else(flush_error);
+            }
+
+            debug!("flush-append succeeded");
```
gefjon, line 213: Kurtis asks that we drop this to `trace`.

kim: I would have preferred to update the log.conf -- the distinction between debug and trace is like between 10pt and 8pt.

But well...

PR #924, opened 2024-03-04
proto changes

Shubham8287:
# Description of Changes

- [x] Protobuf changes
- [x]  Server side changes
- [x] json changes

Not intented to, but could be possible. Reviewer to have attention on this.

### DIFF
--- a/crates/client-api-messages/protobuf/client_api.proto
+++ b/crates/client-api-messages/protobuf/client_api.proto
@@ -51,7 +51,7 @@ message IdentityToken {
 ///
 /// - `argBytes` is the arguments to the reducer, encoded as BSATN.
 ///
-/// - `messageId` is a identifier for client request.
+/// - `requestId` is a identifier for client request.
 /// SpacetimeDB models reducers as taking a single `AlgebraicValue` as an argument, which
 /// generally will be a `ProductValue` containing all of the args (except the
 /// `ReducerContext`, which is injected by the host, not provided in this API).
@@ -59,7 +59,7 @@ message FunctionCall {
     // TODO: Maybe this should be replaced with an int identifier for performance?
     string reducer = 1;
     bytes argBytes = 2;
-    bytes messageId = 3;
+    bytes requestId = 3;
 }
 
 /// Sent by client to database to register a set of queries, about which the client will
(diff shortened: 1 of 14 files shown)


jdetter (approved):
All of the choices here are reasonable:
- We're calling this thing `messageId` for now because the one-off queries already have this concept and there we call it `messageId`. We'll make a P2 ticket for 0.9 to rename these all to `requestId`
- We're putting the `messageId` inside of subscription update instead of transaction update in order to enable us to track round-trip times for subscription changes.

jdetter (changes requested):
Sorry I changed my mind 😅 

I think it makes sense to change messageId to requestId because renaming the field is not ABI breaking, and we have to go through and update the code everywhere anyway to support this.

crates/client-api-messages/protobuf/client_api.proto
```diff
@@ -51,13 +51,15 @@ message IdentityToken {
...
 ///
+/// - `requestId` is a identifier for client request.
 /// SpacetimeDB models reducers as taking a single `AlgebraicValue` as an argument, which
 /// generally will be a `ProductValue` containing all of the args (except the
 /// `ReducerContext`, which is injected by the host, not provided in this API).
 message FunctionCall {
     // TODO: Maybe this should be replaced with an int identifier for performance?
     string reducer = 1;
     bytes argBytes = 2;
+    bytes requestId = 3;
```
Shubham8287, line 62: @jdetter, pretty late but I was thinking, do we really need requestId as `bytes`? or something like `u64` would be sufficient?

jdetter: This definitely should just be a u64 or even just u32, but it would also break the ABI/API if we change it. I haven't looked at how its used, is there a reason other than breaking ABI/API why we can't just do u64 here?

Shubham8287: No, It's due to `oneOffQuery` uses bytes so to be consistent.
but now I feel to be right first, `OneOfQuery` has its own request and response message and doesn't collide with `TransactionUpdate` or `SubscriptionUpdate`. So We can leave that as it is until 0.9 and use `u32` for new messageId's I am adding. What do you say?

jdetter (approved):
Tested, works for me. Changes look reasonable 👍

PR #925, opened 2024-03-05
`iter_by_col_eq`: take AV by ref + impl RangeBounds<..> for &AV

Centril:
# Description of Changes

- Let `iter_by_col_eq` take `&AlgebraicValue` instead of taking ownership.
- Add `impl RangeBounds<AlgebraicValue> for &AlgebraicValue`.

Together, and with especially the latter, these will facilitate reusing queries.

None

1


kim (approved):
:+1:

PR #926, opened 2024-03-05
core: Integrate new commitlog + durability

kim:
# Description of Changes

This patch attempts to integrate the new commitlog with the minimum
changes.

Most of the diff comes from deletions of the legacy log and the need to
adjust tests due to the requirement for a tokio runtime when a durable
database is used in tests.

The "meat" of the patch are the `RelationalDB` constructors,
`RelationalDB::commit_tx`, and the replay logic in
`locking_tx_datastore`.

While `DataKey` is gone, there is still some redundant data being passed
around, which will be addressed in the follow-up patch.

Not compatible with logs written using the previous implementation.

5


crates/core/src/db/datastore/traits.rs
```diff
@@ -173,12 +172,8 @@ pub struct TxRecord {
     pub(crate) op: TxOp,
```
kim, line 172: Do we need `TxOp::Insert` still?

It does not seem like `DataKey` is used anywhere, so `TxData` could possibly replaced by `spacetimedb_commitlog::TxData`.

crates/core/src/db/relational_db.rs
```diff
@@ -724,48 +792,154 @@ impl RelationalDB {
...
+    /// A [`RelationalDB`] in a temporary directory.
+    ///
+    /// When dropped, any resources including the temporary directory will be
+    /// deleted.
+    ///
+    /// To keep the temporary directory, use [`Self::reopen`] or [`Self::into_parts`].
+    ///
+    /// [`TestDB`] is deref-coercible into [`RelationalDB`], which is dubious
+    /// but convenient.
+    pub struct TestDB {
```
kim, line 812: TODO: This should go into the `testing` crate (probably), so bugs around dropping the runtime are made in a single place.

crates/core/src/db/relational_db.rs
```diff
@@ -165,21 +159,9 @@ impl RelationalDB {
...
     pub fn message_log_size_on_disk(&self) -> u64 {
-        self.commit_log()
-            .map_or(0, |commit_log| commit_log.message_log_size_on_disk())
-    }
-
-    /// The number of bytes on disk occupied by the [ObjectDB].
-    pub fn object_db_size_on_disk(&self) -> std::result::Result<u64, DBError> {
-        self.commit_log()
-            .map_or(Ok(0), |commit_log| commit_log.object_db_size_on_disk())
+        unimplemented!()
```
kim, line 164: TODO

crates/core/src/host/wasm_common/module_host_actor.rs
```diff
@@ -608,16 +608,7 @@ impl<T: WasmInstance> WasmModuleInstance<T> {
...
-                    // TODO(cloutiertyler): This tracking doesn't really belong here if we want to write transactions to disk
-                    // in batches. This is because it's possible for a tiny reducer call to trigger a whole commit to be written to disk.
-                    // We should track the commit sizes instead internally to the CommitLog probably.
-                    if let Some(bytes_written) = bytes_written {
-                        WORKER_METRICS
-                            .reducer_write_size
-                            .with_label_values(&address, reducer_name)
-                            .observe(bytes_written as f64);
-                    }
+                if let Some(tx_data) = stdb.commit_tx(&ctx, tx).unwrap() {
```
kim, line 611: Note that it is no longer straightforward to report this metrics due to the durability layer being asynchronous and meant to buffer transaction data.

It is surely possible, however, to calculate the size of `tx_data` as it would be written to the log ... eventually. It's not clear if this is what we want, though.

/cc @joshua-spacetime @cloutiertyler

crates/core/src/db/datastore/locking_tx_datastore/datastore.rs
```diff
@@ -626,6 +561,135 @@ impl MutProgrammable for Locking {
...
+        reader: &mut R,
+    ) -> std::result::Result<Self::Row, Self::Error> {
+        let schema = self
+            .committed_state
+            .schema_for_table(&ExecutionContext::default(), table_id)?
+            .into_owned();
+        let row = ProductValue::decode(schema.get_row_type(), reader)?;
+
+        // TODO: remove
+        log::trace!("visit: insert {table_id} {row:?}");
```
gefjon, line 621: Did you intend to commit this line?

crates/core/src/db/datastore/locking_tx_datastore/datastore.rs
```diff
@@ -626,6 +561,135 @@ impl MutProgrammable for Locking {
...
+    ) -> std::result::Result<Self::Row, Self::Error> {
+        let schema = self
+            .committed_state
+            .schema_for_table(&ExecutionContext::default(), table_id)?;
+        // TODO: avoid clone
+        let table_name = schema.table_name.clone();
+        let row = ProductValue::decode(schema.get_row_type(), reader)?;
+
+        // TODO: remove
+        log::trace!("visit: delete {table_id} {row:?}");
```
gefjon, line 654: Ditto.

gefjon (approved):
Love a big red diff!

crates/core/src/host/wasm_common/module_host_actor.rs
joshua-spacetime: @kim If I'm not mistaken, I don't think this metric is even being used. So I think it's fine to delete.

crates/core/src/db/relational_db.rs
kim: In #955

kazimuth:
benchmarks please

kazimuth (approved):
-3000 is nice 😎 

Running benchmarks to double check for compile errors, the benches mess with relationaldb a lot. Otherwise this makes sense and looks good.

kim:
@kazimuth It is quite possible that this skews some benchmarks, because appending to the log is just appending to a channel. Maybe draining that channel and flushing its contents to disk needs to be included in the measurement?

crates/core/src/db/datastore/locking_tx_datastore/datastore.rs
```diff
@@ -626,6 +561,135 @@ impl MutProgrammable for Locking {
...
+    fn visit_insert<'a, R: BufReader<'a>>(
+        &mut self,
+        table_id: TableId,
+        reader: &mut R,
+    ) -> std::result::Result<Self::Row, Self::Error> {
+        let schema = self
+            .committed_state
+            .schema_for_table(&ExecutionContext::default(), table_id)?
+            .into_owned();
+        let row = ProductValue::decode(schema.get_row_type(), reader)?;
```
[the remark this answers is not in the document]
kim: I implemented the encoding of the mutations according to [this](https://github.com/clockworklabs/SpacetimeDBPrivate/pull/718). It’s possible that we could just convert all rows per table to bflatn at once, instead of each row individually. But it’s not clear to me how exactly that would work.

bfops:
I tested this:
> kim may disagree, but imo i'd say, deploy a world (any size), run bots for a while, disconnect all bots, kill server, restart, connect manually, ensure the world state looks reasonable, connect bots again, ensure the world looks reasonable

(from [here](https://discord.com/channels/931210784011321394/1150256030433955870/1219789729026015242))

The bots seem to be walking around happily after a restart!

crates/bench/src/spacetime_raw.rs
```diff
@@ -24,17 +23,16 @@ impl BenchDatabase for SpacetimeRaw {
     }
     type TableId = TableId;
 
-    fn build(in_memory: bool, fsync: bool) -> ResultBench<Self>
+    fn build(in_memory: bool, _fsync: bool) -> ResultBench<Self>
```
cloutiertyler, line 26: Should we deal with the unused _fsync here?

crates/core/benches/odb_flavor_bench.rs
cloutiertyler, line 1: Huzzah!

crates/bench/src/spacetime_raw.rs
kim: It does have some significance for impls of `BenchDatabase` for other databases.

crates/testing/src/modules.rs
```diff
@@ -192,16 +192,10 @@ impl CompiledModule {
...
-    fsync: FsyncPolicy::Never,
-};
+pub static DEFAULT_CONFIG: Config = Config { storage: Storage::Disk };
 
 /// For performance tests, do not persist to disk.
-pub static IN_MEMORY_CONFIG: Config = Config {
-    storage: Storage::Disk,
-    fsync: FsyncPolicy::Never,
-};
+pub static IN_MEMORY_CONFIG: Config = Config { storage: Storage::Disk };
```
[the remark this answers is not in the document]
kim: Git blame tells me it was introduced by @joshua-spacetime. Should I change it to `Storage::Memory`?

cloutiertyler (approved):
LGTM, goodbye DataKey!

crates/testing/src/modules.rs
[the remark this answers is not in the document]
RReverser: Oh hi, group chat. I just found this through git blame too and was about to ask.

PR #928, opened 2024-03-05
Single-table subscription queries: plan once, run repeatedly

gefjon:
# Description of Changes

Prior to this commit, every call to `ExecutionUnit::eval_incr` re-invoked the query planner to convert its `eval` plan into a plan suitable for reading from a `MemTable` of updates.

With this commit, specifically for single-table select queries, we invoke the query planner once during `ExecutionUnit::new`, and store the resuling `eval_incr_plan` for repeated use.

A follow-up will do the same for multi-table semijoins.

N/a

2

### DIFF
--- a/crates/core/src/subscription/execution_unit.rs
+++ b/crates/core/src/subscription/execution_unit.rs
@@ -1,0 +1,161 @@
+use std::hash::Hash;
+
+use spacetimedb_lib::identity::AuthCtx;
+use spacetimedb_primitives::TableId;
+use spacetimedb_vm::expr::SourceSet;
+
+use crate::db::relational_db::{RelationalDB, Tx};
+use crate::error::DBError;
+use crate::execution_context::ExecutionContext;
+use crate::host::module_host::{DatabaseTableUpdate, TableOp};
+
+use super::query::{self, run_query, Supported};
+use super::subscription::{eval_primary_updates, IncrementalJoin, SupportedQuery};
+
+/// A hash for uniquely identifying query execution units,
+/// to avoid recompilation of queries that have an open subscription.
+///
+/// Currently we are using a cryptographic hash,
+/// which is most certainly overkill.
+/// However the benefits include uniqueness by definition,
+/// and a compact representation for equality comparisons.
+///
+/// It also decouples the hash from the physical plan.
+///
+/// Note that we could hash QueryExprs directly,
+/// using the standard library's hasher.
+/// However some execution units are comprised of several query plans,
+/// as is the case for incremental joins.
+/// And we want to associate a hash with the entire unit of execution,
+/// rather than an individual plan.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct QueryHash {
+    data: [u8; 32],
+}
+
+impl QueryHash {
+    pub const NONE: Self = Self { data: [0; 32] };
+
+    pub fn from_bytes(bytes: &[u8]) -> Self {
+        Self {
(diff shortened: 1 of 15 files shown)


crates/core/src/subscription/execution_unit.rs
```diff
@@ -54,7 +51,17 @@ impl QueryHash {
 #[derive(Debug)]
 pub struct ExecutionUnit {
     hash: QueryHash,
-    plan: SupportedQuery,
+    eval_plan: SupportedQuery,
```
joshua-spacetime, line 54: Do we need to hold both the eval and eval_incr plans? The eval plan will only be executed once at subscribe time.

crates/core/src/subscription/execution_unit.rs
```diff
@@ -132,24 +183,38 @@ impl ExecutionUnit {
...
+            for table in tables.filter(|table| table.table_id == self.return_table()) {
+                // Build a `SourceSet` containing the updates from `table`.
+                let mem_table = query::to_mem_table_with_op_type(header.clone(), table_access, table);
+                let mut sources = SourceSet::default();
+                let _source_expr = sources.add_mem_table(mem_table);
+                debug_assert_eq!(_source_expr.source_id(), Some(_source_id));
+                // Evaluate the saved plan against the new `SourceSet`
+                // and capture the new row operations.
+                ops.extend(
+                    eval_primary_updates(db, auth, tx, eval_incr_plan, sources)?.map(|r| TableOp::new(r.0, r.1)),
```
joshua-spacetime, line 207: Ideally `eval_primary_updates` would just instantiate the iterator and run through the results. But right now it creates another intermediate representation `QueryCode`, and that's what is passed to `build_query`.

So to avoid cloning the QueryExpr, every time we call `eval_primary_updates`, we'll probably want to have it operate directly on `QueryCode`.

And I think ultimately this means that an ExecutionUnit should hold a QueryCode rather than a QueryExpr.

joshua-spacetime (commented):
I think we'll want to store the final representation (QueryCode) right before iterator in ExecutionUnit. But otherwise this looks good.

crates/core/src/subscription/execution_unit.rs
gefjon: I assumed (but have not checked) that we would re-use the `eval_plan` when a new subscriber joins with the same query. If that's not the case, then either:
- We should do that.
- We should re-structure `ExecutionUnit` so that it isn't responsible for `eval`.
Either way, I think this should be a follow-up.

crates/core/src/subscription/execution_unit.rs
gefjon: I changed `ExecutionUnit` so that, for `Select` queries, it stores `QueryCode` instead of `QueryExpr`. This does not remove the clones, as `build_query` takes an owned `QueryCode`. We could change that pretty easily, I think, but I'd like it to wait for a follow-up PR.

joshua-spacetime (approved):
After the first version of this patch that operated on QueryExpr
```
incr-select             time:   [860.02 ns 860.59 ns 861.18 ns]
                        change: [-54.939% -54.853% -54.763%] (p = 0.00 < 0.05)
                        Performance has improved.
```

After the second and final version of this patch that operates now on QueryCode
```
incr-select             time:   [650.26 ns 650.57 ns 650.93 ns]
                        change: [-24.252% -24.098% -23.938%] (p = 0.00 < 0.05)
                        Performance has improved.
```

Fantastic!

PR #930, opened 2024-03-05
Kick clients that are backing up their message channel

coolreader18:
# Description of Changes

This is a pretty forceful method of disconnecting them, but I think that's what the situation calls for. #913 should probably be merged before/at the same time as this, to lessen the chance of false positive kicking.

2-3? It's possible to get false positives from this, but I'm pretty confident that it won't be often, unless it turns out we need `CLIENT_CHANNEL_CAPACITY` to be higher than I set it.


crates/core/src/client/client_connection.rs
```diff
@@ -113,33 +146,50 @@ impl ClientConnection {
             .await?;
 
         // Buffer up to 64 client messages
```
jdetter, line 148: This comment seems to be out of date, could you update it?

coolreader18:
Split into two commits, to make it easier to review the logical changes in the second one.

PR #935, opened 2024-03-06
Avoid `Header::find_pos_by_name` in `eval_incr`.

gefjon:
# Description of Changes

Prior to this commit, `to_mem_table_with_op_type` would call `Header::find_pos_by_name` to locate the `__op_type` column, if it existed. This was slow, as `Header::find_pos_by_name` is a linear scan in increasing order, and the `__op_type` column is always either last or not present, so `to_mem_table_with_op_type` would traverse every column in the `Header`. This happened during every incremental evaluation for every query.

With this commit, we rely on the fact that `__op_type` is always the last column if present, and check only the last column of the header.

N/a.

2 - if I'm wrong about `__op_type` always being last, this will cause problems.

### DIFF
--- a/crates/core/src/subscription/query.rs
+++ b/crates/core/src/subscription/query.rs
@@ -10,6 +10,7 @@ use once_cell::sync::Lazy;
 use regex::Regex;
 use spacetimedb_lib::identity::AuthCtx;
 use spacetimedb_lib::Address;
+use spacetimedb_primitives::ColId;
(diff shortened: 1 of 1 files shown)


joshua-spacetime (approved):
```
incr-select             time:   [557.53 ns 561.97 ns 566.63 ns]
                        change: [-12.409% -12.113% -11.776%] (p = 0.00 < 0.05)
                        Performance has improved.
```

PR #936, opened 2024-03-06
chore: Remove debug instrumentation from hot path

joshua-spacetime:
Removes most instrumentation from RelationalDB,
but keeps all pre-existing instrumentation of the wasm api.

None

1


jdetter (approved):
Tested with bots - works for me and this seems to still have a decent amount of observability 👍

PR #937, opened 2024-03-06
Misc tweaks to `tools/perf.sh`

bfops:
# Description of Changes
Made a few minor updates to `tools/perf.sh`:
- explicitly check & log "found multiple candidate processes" separately from "no process found"
- only grabs processes that look like `spacetime start`
    - previously this would complain about e.g. `spacetime logs`
- show spacetime processes from **all** users on the machine

None

2


tools/perf.sh
```diff
@@ -85,12 +85,18 @@ if [[ -z "$DATAFILE" ]]; then
     DATAFILE=perf.data
 fi
 if [[ -z "$SPACETIME_PID" ]]; then
-    if [[ $(ps -a | grep spacetime | wc -l) != 1 ]] ; then
+    SPACETIMES="$(ps -a -e -f | grep '\<spacetime\>.*\<start\>' | grep -v '\<grep\>')"
```
bfops, line 88: `-f` allows us to get the args
`-e` lets us see all users

gefjon: This would make a great comment to add to the code.

gefjon (approved):
Can you go through and change all the error-path `echo`s to `>&2 echo`? Other than that, this looks good!

tools/perf.sh
cloutiertyler: Although, that's a bigger change

PR #938, opened 2024-03-06
Incremental joins: compile once, run repeatedly.

gefjon:
# Description of Changes

Well, more like, compile 3 times, run repeatedly, but 3 is approximately 1.

This commit re-writes `IncrementalJoin` to be a re-usable representation of a query plan for an incremental join,
where before it was a one-off worker for the same.

`IncrementalJoin` stores three copies of the query, compiled for `MemTable * DbTable`, `DbTable * MemTable` and `MemTable * MemTable`.

Related to this change, `eval_incr` no longer needs an `AuthCtx`, because we check permissions during query compilation, not execution, and all query planning is now done ahead of time during `add_subscriber`. As a result, many callsites, especially in tests, which used to pass an `AuthCtx` no longer do so.

N/a

corresponding GitHub label.

2 - I feel pretty good about this, having just done the select version.


crates/core/src/subscription/subscription.rs
```diff
@@ -262,178 +314,244 @@ impl JoinSide {
...
+    ) -> Result<Option<impl Iterator<Item = TableOp>>, DBError> {
+        let Some((index_side, probe_side)) = self.find_updates(updates) else {
+            return Ok(None);
+        };
+        let mut inserts = {
             // {A+ join B}
-            let a = eval_updates(db, *auth, tx, &join_a, self.join.return_index_rows, join_a_sources)?;
+            let index_inserts = index_side.inserts();
+            let index_inserts = to_mem_table_with_op_type(
+                self.index_table.head.clone(),
```
joshua-spacetime, line 454: Optional: We can further optimize this since we know this is the header of a MemTable, and so we can also compute the Header ahead of time.

joshua-spacetime (approved):
```
incr-join               time:   [7.5371 µs 7.5582 µs 7.5765 µs]
                        change: [-30.263% -29.969% -29.670%] (p = 0.00 < 0.05)
                        Performance has improved.
```

gefjon:
After the latest batch of commits, we're down to:
```
incr-join               time:   [4.6166 µs 4.6216 µs 4.6268 µs]
                        change: [-25.078% -24.958% -24.846%] (p = 0.00 < 0.05)
                        Performance has improved.
```
I don't actually know what the original time before this PR was, but this is a hefty improvement.

PR #939, opened 2024-03-07
fix: Optimize query plan in iter_filtered_chunks

Centril:
# Description of Changes

There were two bugs that prevented us from using multi-col indexes in `query!`.
None of these were in `find_sargable`, rather, they were in:
1. `InstanceEnv::iter_filtered_chunks`: it didn't call `q.optimize(...)` (so we may potentially see other improvements now).
2. `Header::column` would not find the `Column` due to `FieldName::Pos` being passed.

I've verified that the query ends up being optimized to a `Query::IndexScan` and that this actually reaches the index with the right name in `BTreeIndex::seek`.

None

2 - There are other users of `Header::column_pos_idx`.

### DIFF
--- a/crates/core/src/db/relational_db.rs
+++ b/crates/core/src/db/relational_db.rs
@@ -1126,7 +1126,6 @@ mod tests {
         stdb.commit_tx(&ExecutionContext::default(), tx)?;
         drop(stdb);
 
-        dbg!("reopen...");
         let stdb = open_db(&tmp_dir, false, true)?;
 
         let mut tx = stdb.begin_mut_tx(IsolationLevel::Serializable);
--- a/crates/core/src/host/instance_env.rs
+++ b/crates/core/src/host/instance_env.rs
@@ -367,8 +367,12 @@ impl InstanceEnv {
         .map_err(NodesError::DecodeFilter)?;
 
         // TODO(Centril): consider caching from `filter: &[u8] -> query: QueryExpr`.
-        let query =
-            spacetimedb_vm::dsl::query(schema.as_ref()).with_select(filter_to_column_op(&schema.table_name, filter));
+        let query = spacetimedb_vm::dsl::query(schema.as_ref())
+            .with_select(filter_to_column_op(&schema.table_name, filter))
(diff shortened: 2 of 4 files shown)


crates/sats/src/relation.rs
```diff
@@ -255,20 +255,34 @@ impl Header {
...
+                    .iter()
+                    .position(|f| &f.field == col)
+                    // ..and if we cannot find one,
+                    // check that the `field`th column has a matching table name.
+                    .or_else(|| {
+                        self.fields
+                            .get(*field)
+                            .filter(|c| c.field.table() == table)
+                            .map(|_| *field)
+                    })
```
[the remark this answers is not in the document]
mamcx: What could trigger the `or_else` scenario? It should not happened

modules/multi-col-index-test/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+pub fn test() -> anyhow::Result<()> {
+    log::info!("BEGIN");
+
+    // Try to find a row somewhere in the middle.
+    let offset = OffsetCoordinatesSmall {
+        x: 80,
+        z: 80,
+        dimension: 80,
+    };
+    let q = query!(|ls: LocationState| ls.x == offset.x && ls.z == offset.z && ls.dimension == offset.dimension);
```
joshua-spacetime, line 56: What are we testing?

joshua-spacetime: I'd rather we don't add it if it's not being exercised by CI.

joshua-spacetime (commented):
This looks good, I'd just prefer if we didn't include the module since it's not a proper test just yet.

modules/multi-col-index-test/src/lib.rs
cloutiertyler: Can we not just add a timer here to make sure it takes the appropriate amount of time based on the expected query plan? I've been asking for that for a while, and no one has explained why that cannot be done.

jdetter (approved):
I've posted the results in discord - this is a substantial improvement against the master baseline 👍

PR #940, opened 2024-03-07
Optimize the lookups of fields in the 'Header' and reduce the cost of allocations

### LINKED ISSUE #932: Make field lookups in Header faster
Image

This is a profile from a recent benchmark run for incremental select. The red is `Header::find_pos_by_name` which represents ~30% of eval_incr. This this function were just a lookup into a hash map, we could effectively remove this from the profile.

mamcx:
# Description of Changes

Eliminate the linear scans in lookups for `FieldName`:

```bash
# Before
find_by_name            time:   [13.951 µs 14.554 µs 15.303 µs]
  6 (6.00%) high severe  

#After
find_by_name            time:   [269.77 ns 270.10 ns 270.51 ns]
                        change: [-98.148% -98.082% -98.027%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild
```

Using a map of `FieldsName -> Postions`. 

NOTE: I don't use a regular `HashMap` or `Map<FieldName, Column>` because the order of the fields should be preserved as intended by the user (ie: `select c, a, b, i_want_this_order`).

Also, considering that the header is rarely updated, the `FieldNames` are now under `Arc`.

Closes #932.

1

### DIFF
--- a/crates/sats/src/relation.rs
+++ b/crates/sats/src/relation.rs
@@ -6,9 +6,13 @@ use crate::{algebraic_type, AlgebraicType, ProductType, ProductTypeElement, Type
 use derive_more::From;
 use spacetimedb_primitives::{ColId, ColList, ColListBuilder, Constraints, TableId};
 use std::collections::hash_map::DefaultHasher;
+use std::collections::HashMap;
 use std::fmt;
 use std::hash::{Hash, Hasher};
+use std::ops::Index;
+use std::slice::Iter;
 use std::sync::Arc;
+use std::vec::IntoIter;
 
 pub fn calculate_hash<T: Hash>(t: &T) -> u64 {
     let mut s = DefaultHasher::new();
@@ -95,9 +99,9 @@ impl FieldName {
         }
     }
 
-    pub fn into_field_name(self) -> Option<String> {
+    pub fn to_field_name(&self) -> Option<String> {
         match self {
-            FieldName::Name { field, .. } => Some(field),
+            FieldName::Name { field, .. } => Some(field.clone()),
             FieldName::Pos { .. } => None,
         }
     }
@@ -164,7 +168,7 @@ pub struct ColumnOnlyField<'a> {
 // TODO(perf): Remove `Clone` derivation.
 #[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
 pub struct Column {
-    pub field: FieldName,
+    pub field: Arc<FieldName>,
     pub algebraic_type: AlgebraicType,
     pub col_id: ColId,
 }
@@ -172,7 +176,7 @@ pub struct Column {
 impl Column {
     pub fn new(field: FieldName, algebraic_type: AlgebraicType, col_id: ColId) -> Self {
         Self {
-            field,
+            field: field.into(),
             algebraic_type,
             col_id,
         }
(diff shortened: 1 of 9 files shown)


crates/sats/src/relation.rs
```diff
@@ -147,15 +151,15 @@ pub struct ColumnOnlyField<'a> {
 // TODO(perf): Remove `Clone` derivation.
 #[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
 pub struct Column {
-    pub field: FieldName,
+    pub field: Arc<FieldName>,
```
joshua-spacetime, line 154: I actually think this doesn't need to be an Arc, for reasons I will explain below.

crates/sats/src/relation.rs
```diff
@@ -147,15 +151,15 @@ pub struct ColumnOnlyField<'a> {
...
     pub algebraic_type: AlgebraicType,
     pub col_id: ColId,
 }
 
 impl Column {
     pub fn new(field: FieldName, algebraic_type: AlgebraicType, col_id: ColId) -> Self {
         Self {
-            field,
+            field: field.into(),
```
joshua-spacetime, line 162: We're passing in an owned `FieldName`, so I don't think we need to clone here.

crates/sats/src/relation.rs
```diff
@@ -175,18 +179,139 @@ pub struct HeaderOnlyField<'a> {
...
 }
 
+/// Represents a collection of `fields` metadata made of [Column].
+///
+/// The `fields` need to be preserved by the order specified by the user.
+#[derive(Debug, Clone, Eq, PartialOrd, Ord, Hash)]
+pub struct Fields {
+    columns: Vec<Column>,
+    /// Keeps an index to quickly look up the position
+    idx: BTreeMap<Arc<FieldName>, usize>,
```
joshua-spacetime, line 189: Again, I actually don't think `Arc` is necessary here. Also I don't think we should use FieldName here but rather
```rust
idx: BTreeMap<(String, TableId), usize)>
```

I will explain below.

crates/sats/src/relation.rs
```diff
@@ -255,34 +384,77 @@ impl Header {
...
-                .position(|(pos, f)| &f.field == col || *field == pos),
+            FieldName::Name { .. } => self.fields.get_field(col),
+            FieldName::Pos { field, .. } => self.fields.column_by_pos(*field),
         }
-        .map(Into::into)
     }
 
-    pub fn column_pos_or_err<'a>(&'a self, col: &'a FieldName) -> Result<ColId, RelationError> {
-        self.column_pos(col)
+    /// Returns the [Column] with the specified `field_name`.
```
joshua-spacetime, line 410: Do we need all of these extra methods on Header? I was assuming we would just update the `find_pos_by_name` method.

crates/sats/src/relation.rs
```diff
@@ -255,34 +384,77 @@ impl Header {
...
             .ok_or_else(|| RelationError::FieldNotFound(self.clone_for_error(), col.clone()))
     }
 
-    /// Finds the position of a field with `name`.
-    pub fn find_pos_by_name(&self, name: &str) -> Option<ColId> {
-        self.column_pos(&FieldName::named(&self.table_name, name))
+    /// Finds the [ColId] of a field with `name`.
+    pub fn find_col_id_by_name(&self, name: &str) -> Option<ColId> {
+        self.col_id_by_field(&FieldName::named(&self.table_name, name))
+    }
```
joshua-spacetime, line 444: This still has an unneeded allocation. That's because `FieldName::named` takes a `&str` and creates an owned `String` from it. But we're just doing a lookup into a map. So we shouldn't need to allocate at all.

So we really need to update `col_id_by_field` to take a `&str` instead.

joshua-spacetime (changes requested):
So far you've replaced iteration over the fields with a btree lookup which is excellent. However the last piece that is missing is the removal of the string allocation when performing said lookup.

kazimuth:
Yeah, I was also wondering if it would be possible to switch entirely to integer identifiers for fields. (Also tables, but that would be later...)

crates/sats/src/relation.rs
mamcx: This is to mesh well with:

```rust
impl Fields {
    /// Creates a new instance of `Fields` with the given columns, and build an internal map of `FieldName` to their positions.
    pub fn new(columns: Vec<Column>) -> Self {
        let idx = columns
            .iter()
            .enumerate()
            .map(|(pos, x)| (x.field.clone(), pos))
            .collect();
        Self { columns, idx }
    }
```

Without the `arc` this will be another alloc.

crates/sats/src/relation.rs
mamcx: Is for streamlining the calls elsewhere...

bfops (changes requested):
# Bot test results
**Conclusion**: **Possible performance regression**

@mamcx It seems like this slows things down. I tested twice to check.
I have no idea why (maybe related to some of the allocations that @joshua-spacetime pointed out).

### Commits tested
before:
```
commit e4e80de664a882cefacbc9a03bf2d81bb3e23d71 (origin/bfops/perf-test-before)
Merge: a05aed4e 0ad01b77
Author: Zeke Foppa <196249+bfops@users.noreply.github.com>
Date:   Thu Mar 21 14:08:04 2024 -0700

    [bfops/perf-test-before]: Merge remote-tracking branch 'origin/bfops/enable-tracy' into bfops/perf-test-before

commit a05aed4e9a8b39e3be8c74e4a247a15d3d71c805
Author: Mazdak Farrokhzad <twingoow@gmail.com>
Date:   Thu Mar 14 19:16:04 2024 +0100

    Code motion: QueryCode => QueryExpr; CrudCode => CrudExpr (#975)
    
    * QueryCode.{table -> source}
    
    * nix QueryCode; identical to QueryExpr
    
    * nix CrudCode; use identical CrudExpr instead
```
after:
```
commit c3ab59e88852c4b6d327353731be65cf3a72902b (HEAD -> bfops/perf-test-after, origin/bfops/perf-test-after)
Merge: d7231c0b 0ad01b77
Author: Zeke Foppa <196249+bfops@users.noreply.github.com>
Date:   Thu Mar 21 14:08:08 2024 -0700

    [bfops/perf-test-after]: Merge remote-tracking branch 'origin/bfops/enable-tracy' into bfops/perf-test-after

commit d7231c0bb187ff9e04003fe3be3bcf4b1e229630 (mamcx/find_pos_by_name)
Author: Mario Alejandro Montoya Cortés <mamcx@elmalabarista.com>
Date:   Fri Mar 8 10:04:59 2024 -0500

    Lints
```
### `eval`
image

### `eval_updates`
image

### `enemy_ai_agent_loop`
image

### `npc_ai_agent_loop`
image

joshua-spacetime:
@mamcx since this is no longer critical for achieving our performance target, I suggest we close this and come back to it as part of refactoring the query engine.

PR #946, opened 2024-03-08
Truly remove perfcnt

mamcx:
# Description of Changes

Prior PR https://github.com/clockworklabs/SpacetimeDB/pull/941 only disabled `perfcnt` but because `criterion-perf-events` was in `Cargo.toml` it still compiles in `macOS` where is not supported.

1

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -111,7 +111,6 @@ colored = "2.0.0"
 console = { version = "0.15.6" }
 convert_case = "0.6.0"
 criterion = { version = "0.5.1", features = ["async", "async_tokio", "html_reports"] }
-criterion-perf-events = "0.4"
 crossbeam-channel = "0.5"
 cursive = { version = "0.20", default-features = false, features = ["crossterm-backend"] }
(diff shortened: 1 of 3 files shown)


kazimuth (approved):
Formatting change in cargo.toml but otherwise looks good, thanks for the catch

Cargo.toml
```diff
@@ -1,34 +1,34 @@
...
-  "modules/rust-wasm-test",
-  "modules/quickstart-chat",
-  "modules/sdk-test",
-  "modules/sdk-test-connect-disconnect",
-  "modules/spacetimedb-quickstart",
-  "crates/sdk/tests/test-client",
-  "crates/sdk/tests/test-counter",
-  "crates/sdk/tests/connect_disconnect_client",
-  "tools/upgrade-version",
+    "crates/bench",
```
cloutiertyler, line 3: Why the formatting change?

cloutiertyler: I'd rather not pollute the git history

mamcx: Done. Look like JetBrains add a formatter for toml.

PR #947, opened 2024-03-08
Also poll handle_queue while flushing

coolreader18:
# Description of Changes

Fixes the gap issue.

1

### DIFF
--- a/crates/client-api/src/routes/subscribe.rs
+++ b/crates/client-api/src/routes/subscribe.rs
@@ -196,7 +196,7 @@ async fn ws_client_actor_inner(
     // https://rust-lang.github.io/wg-async/vision/submitted_stories/status_quo/aws_engineer/solving_a_deadlock.html
     //
     // NOTE: never let this go unpolled while you're awaiting something; otherwise, it's possible
-    //       to deadlock or delay for a long time. see usage of `also_poll()` the branches of the
+    //       to deadlock or delay for a long time. see usage of `also_poll()` in the branches of the
     //       `select!` for examples of how to do this.
     //
     // TODO: do we want this to have a fixed capacity? or should it be unbounded
@@ -247,6 +247,8 @@ async fn ws_client_actor_inner(
                         // now we flush all the messages to the socket
                         ws.flush().await
                     };
+                    // Flush the websocket while continuing to poll the `handle_queue`,
+                    // to avoid deadlocks or delays due to enqueued futures holding resources.
                     let send_all = also_poll(send_all, handle_queue.make_progress());
                     let t1 = Instant::now();
                     if let Err(error) = send_all.await {
@@ -262,6 +264,8 @@ async fn ws_client_actor_inner(
 
             // If the module has exited, close the websocket.
             () = client.module.exited(), if !closed => {
+                // Send a close frame while continuing to poll the `handle_queue`,
+                // to avoid deadlocks or delays due to enqueued futures holding resources.
                 let close = also_poll(
                     ws.close(Some(CloseFrame { code: CloseCode::Away, reason: "module exited".into() })),
                     handle_queue.make_progress(),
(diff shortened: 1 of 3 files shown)


crates/client-api/src/routes/subscribe.rs
```diff
@@ -268,7 +277,7 @@ async fn ws_client_actor_inner(
             _ = liveness_check_interval.tick() => {
                 // If we received a pong at some point, send a fresh ping.
                 if mem::take(&mut got_pong) {
-                    if let Err(e) = ws.send(WsMessage::Ping(Vec::new())).await {
+                    if let Err(e) = also_poll(ws.send(WsMessage::Ping(Vec::new())), handle_queue.make_progress()).await {
```
gefjon, line 280:
```suggestion
                    // Send a ping message while continuing to poll the `handle_queue`,
                    // to avoid deadlocks or delays due to enqueued futures holding resources.
                    if let Err(e) = also_poll(ws.send(WsMessage::Ping(Vec::new())), handle_queue.make_progress()).await {
```

crates/client-api/src/routes/subscribe.rs
```diff
@@ -257,7 +262,11 @@ async fn ws_client_actor_inner(
 
             // If the module has exited, close the websocket.
             () = client.module.exited(), if !closed => {
-                if let Err(e) = ws.close(Some(CloseFrame { code: CloseCode::Away, reason: "module exited".into() })).await {
+                let close = also_poll(
```
gefjon, line 265:
```suggestion
                // Send a close frame while continuing to poll the `handle_queue`,
                // to avoid deadlocks or delays due to enqueued futures holding resources.
                let close = also_poll(
```

crates/client-api/src/routes/subscribe.rs
```diff
@@ -243,6 +247,7 @@ async fn ws_client_actor_inner(
                         // now we flush all the messages to the socket
                         ws.flush().await
                     };
+                    let send_all = also_poll(send_all, handle_queue.make_progress());
```
gefjon, line 250:
```suggestion
                    // Flush the websocket while continuing to poll the `handle_queue`,
                    // to avoid deadlocks or delays due to enqueued futures holding resources.
                    let send_all = also_poll(send_all, handle_queue.make_progress());
```

crates/client-api/src/routes/subscribe.rs
```diff
@@ -195,6 +195,10 @@ async fn ws_client_actor_inner(
     // awaited, which can lead to bugs like:
     // https://rust-lang.github.io/wg-async/vision/submitted_stories/status_quo/aws_engineer/solving_a_deadlock.html
     //
+    // NOTE: never let this go unpolled while you're awaiting something; otherwise, it's possible
+    //       to deadlock or delay for a long time. see usage of `also_poll()` the branches of the
+    //       `select!` for examples of how to do this.
+    //
```
gefjon, line 201:
```suggestion
    // NOTE: never let this go unpolled while you're awaiting something; otherwise, it's possible
    //       to deadlock or delay for a long time. see usage of `also_poll()` in the branches of the
    //       `select!` for examples of how to do this.
    //
```

gefjon (approved):
Suggested a few comments to make clear where the magic is happening.

gefjon:
This PR fixes what we've been calling the "gap issue," where a client with a poor connection could cause an entire module to freeze up, leaving characteristic "gaps" in a Tracy graph.

The problem was unpolled futures. The problem is always unpolled futures.

Each subscriber is associated with a `handle_queue` of incoming messages, e.g. reducer requests, and a `sendrx` of outgoing messages, e.g. subscription updates. The subscription worker (in `ws_client_actor_inner`) `select!`s among these and takes the appropriate action. For outgoing messages, the appropriate action is `send_all`, which feeds the messages into the WebSocket and then `flush`es it.

For clients with shoddy connections, the `flush` can take significant amounts of time, potentially several seconds.

Prior to this commit, during `send_all`, we were not polling the `handle_queue`. This meant that, during the long `flush`, any futures created from previously-received partially-processed messages were not advancing. This was a problem because these futures can (and do) contain contended resources. In our case, the contended resource was a semaphore token taken by `call_reducer`, but in general it could be e.g. a lock or a future's position in another queue.

As a result, if a client with a shoddy connection sent a reducer request, and then while that request was running, took a long time to receive and ACK an update on its WebSocket, the reducer request could block the server by holding a semaphore token in the `handle_queue` but not being polled.

With this commit, Noa has changed `ws_client_actor_inner` so that it continues to poll `handle_queue` while sending WebSocket messages. This means that, while the slow client will not receive its messages for a very long time, the reducer will still run and release its semaphore token, allowing other clients' reducer requests to proceed.

jdetter (approved):
Tested with bots - even playable over tailscale now! 👍

PR #950, opened 2024-03-11
core: More compact TxData

kim:
`TxData` is the representation of a transaction after it was committed,
and is passed around for evaluation of subscription queries and sending
the result to clients.

With the new commitlog, it can be represented more compactly, such that
copies for writing to the log can be avoided.

~~Note that this patch stops short of refactoring `DatabaseUpdate`, which
is another representation of the same information as sent to clients.
This means that `ProductValue`s need to be cloned from `TxData`, just as
before.~~ ProductValues are now `Arc<[ProductValue]>`, eliminating expensive clones.

**NOTE: This PR is stacked on top of #926**

### DIFF
--- a/crates/core/src/db/datastore/traits.rs
+++ b/crates/core/src/db/datastore/traits.rs
@@ -1,9 +1,11 @@
 use std::borrow::Cow;
+use std::collections::BTreeMap;
 use std::{ops::RangeBounds, sync::Arc};
 
 use super::Result;
 use crate::db::datastore::system_tables::ST_TABLES_ID;
 use crate::execution_context::ExecutionContext;
+use spacetimedb_data_structures::map::IntMap;
 use spacetimedb_primitives::*;
 use spacetimedb_sats::db::def::*;
 use spacetimedb_sats::hash::Hash;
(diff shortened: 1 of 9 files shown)


crates/core/src/db/datastore/traits.rs
```diff
@@ -158,29 +159,49 @@ pub enum IsolationLevel {
...
+/// A record of all the operations within a transaction.
+#[derive(Default)]
+pub struct TxData {
+    /// The inserted rows per table.
+    inserts: BTreeMap<TableId, Arc<[ProductValue]>>,
+    /// The deleted rows per table.
+    deletes: BTreeMap<TableId, Arc<[ProductValue]>>,
+    /// Map of all `TableId`s in both `inserts` and `deletes` to their
+    /// corresponding table name.
+    tables: BTreeMap<TableId, String>,
```
gefjon, line 171: Pure curiosity: Why `BTreeMap` instead of `HashMap`?

gefjon (approved):

crates/core/src/db/datastore/traits.rs
kim: It is significant when replaying that system tables (lower ids) come first, so creating and mutating a table in the same transaction works.

The way the visitor pattern reading works, this also means keeping them ordered on disk. This may be debatable…

kazimuth (approved):
👍

crates/core/src/host/module_host.rs
```diff
@@ -71,31 +69,25 @@ impl DatabaseUpdate {
...
-        for record in tx_data.records.iter() {
-            let pv = record.product_value.clone();
-            let table = map.entry(record.table_id).or_default();
-            match record.op {
-                TxOp::Delete => &mut table.0,
-                TxOp::Insert(_) => &mut table.1,
-            }
-            .push(pv);
+    pub fn from_writes(tx_data: &TxData) -> Self {
+        let mut map: HashMap<TableId, DatabaseTableUpdate> = HashMap::new();
```
kim, line 73: TODO: this changed since last rebase, we should be able to just clone the `Arc<[ProductValue]>` into the `DatabaseTableUpdate`.

kim: Changed `DatabaseTableUpdate` to contain `Arc[ProductValue]>` instead of `Vec<ProductValue>` as of [7065dad5bbca2c5271fab0705d7b8f0196eab1d5](https://github.com/clockworklabs/SpacetimeDB/compare/a5bd86d3f24a2db0a188514fd39c78492be49e57..7065dad5bbca2c5271fab0705d7b8f0196eab1d5)

cc @Centril as you have touched this code most recently

(Added missing #[inline] in [db0ff831c2540973f00c41d31ea6f4455d904614](https://github.com/clockworklabs/SpacetimeDB/compare/7065dad5bbca2c5271fab0705d7b8f0196eab1d5..db0ff831c2540973f00c41d31ea6f4455d904614))

crates/core/src/subscription/query.rs
```diff
@@ -1079,8 +1079,8 @@ mod tests {
             DatabaseTableUpdate {
                 table_id: lhs_id,
                 table_name: "lhs".to_string(),
-                deletes: vec![lhs_old],
-                inserts: vec![lhs_new],
+                deletes: vec![lhs_old].into(),
+                inserts: vec![lhs_new].into(),
```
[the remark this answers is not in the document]
kim: Ya, fair enough

crates/core/src/host/module_host.rs
```diff
@@ -71,31 +69,31 @@ impl DatabaseUpdate {
...
+        for (table_id, table_name, rows) in tx_data.inserts() {
+            map.entry(*table_id)
+                .and_modify(|update| update.inserts = rows.clone())
+                .or_insert(DatabaseTableUpdate {
+                    table_id: *table_id,
+                    table_name: table_name.to_string(),
+                    inserts: rows.clone(),
+                    deletes: Arc::new([]),
+                });
```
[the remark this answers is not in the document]
kim: Good catch, thanks

crates/core/src/db/datastore/traits.rs
kim: I changed `tables` to use `spacetimedb_data_structures::map::IntMap`, which is also used to build the `DatabaseUpdate`.

Or did you indeed mean `IndexMap` (i.e. insertion ordered)?

gefjon: Does `IntMap` provide the ordering guarantees you need?

kim: So `tables` doesn’t need ordering at all, and we “know” table ids are unique.

The ordering of inserts and deletes is significant, although if we _somehow_ restrict construction to the `merge` function, we’d also have them ordered already.

I just don’t want to docstring-type that, as I’ve had many issues in the past with implied ordering / set semantics.

cloutiertyler (approved):
LGTM

PR #951, opened 2024-03-11
perf(933): Clone bsatn instead of product values in incremental update

Centril:
# Description of Changes

"Cache" the BSATN serialization in incremental updates for subscribers using the same query.
Unfortunately, we have to deal with JSON as a possibility because of the website, so we have to thread through a bunch of `Either`s.

None

2


gefjon (approved):
Agree with Joshua that I want to understand the performance impact of this change before merging, but the code looks good.

kazimuth (approved):
I briefly looked at this before, this approach definitely makes the most sense.

joshua-spacetime (approved):

PR #952, opened 2024-03-11
refactor: Incremental evaluation tests for index joins

joshua-spacetime:
No functionality has been modified.
This patch just makes incremental eval test cases clearer and easier to extend.

1 - as long as tests pass


cloutiertyler (approved):
LGTM given that this is just moving code.

PR #955, opened 2024-03-12
core,commitlog: Re-instantiate commitlog disk usage reporting

kim:
Disk usage reporting was left unimplemented in previous patches of the series, as its semantics are slightly different from before.

Namely, inspecting the size of the commitlog now requires to `stat(2)` the segment files, and is thus fallible.

Also, a size reporting function is only defined for local durability (i.e. the commitlog). The behaviour when the database is in a follower state is left unspecified.

**NOTE: This patch is stacked on top of #950**

1.5


gefjon (approved):
Probably we want @joshua-spacetime to review this, but it looks reasonable to me.

joshua-spacetime (approved):
Everything looks good to me.

PR #957, opened 2024-03-12
fix: Incremental evaluation for index joins

### LINKED ISSUE #954: Correctness issue for incremental evaluation of index joins
Specifically for transactions that produce both inserts and deletes for each table.

joshua-spacetime:
Closes #954.

I previously avoided the evaluation of certain delta table joins as an optimization, which relied on the fact that a tx would not include inserts and deletes for both tables, which of course is not generally correct.

This patch includes the fully general solution.

3 - we have relatively good test coverage


crates/core/src/subscription/subscription.rs
```diff
@@ -350,134 +377,213 @@ impl IncrementalJoin {
...
 
+    /// Evaluate this [`IncrementalJoin`] over the row updates of a transaction t.
+    ///
+    /// In the comments that follow,
+    /// B(t) refers to the state of table B as of transaction t.
+    /// In particular, B(t) includes all of the changes from t.
+    /// B(s) refers to the state of table B as of transaction s,
+    /// where s is the transaction immediately preceeding t.
+    ///
+    /// Now we may ask,
```
joshua-spacetime, line 463: Here's an explanation of the math.

PR #962, opened 2024-03-13
Added 'override' keywords to functions and properties extending from …

hsreina:
What version of TypeScript are you targeting?

Since version 4.3 of TypeScript, override keyword has been introduced for properties or functions of classes extending from other classes.

Recent frameworks using Typescript might enforce the presence of the override keyword.

Of course implicit in projects can be disabled with the follow code. But I think it might be better to have it.
```
{
  "compilerOptions": {
    // ...
    "noImplicitOverride": true
  }
}
```

### DIFF
--- a/crates/cli/src/subcommands/generate/typescript.rs
+++ b/crates/cli/src/subcommands/generate/typescript.rs
@@ -687,8 +687,16 @@ fn autogen_typescript_product_table_common(
     {
         indent_scope!(output);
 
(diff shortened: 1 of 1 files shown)


hsreina:
I need to have a look how you run rust unit tests

cloutiertyler:
Hi @hsreina This is a long time coming, but this PR is actually no longer necessary. We've rewritten the TypeScript SDK to not use inheritance at all.

PR #969, opened 2024-03-13
Distinguish between inner and semijoins in `QueryExpr` AST.

gefjon:
# Description of Changes

This commit adds a flag `semi: bool` to `JoinExpr`, which signifies a semijoin, as opposed to an inner join.

A new optimization pass, `QueryExpr::try_semi_join`, is defined which can detect a certain common case of inner joins and rewrite them into semijoins.

The punchline here is that `core::vm::join_inner` used to accept a flag `semi: bool` which it could use to avoid some expensive `Header` mutations, but that flag was always passed as `false` because we had no way to distinguish semijoins. With this commit, the flag is actually used,
so evaluating non-indexed semijoins should avoid allocating a new `Header`.

N/a

3 - our query planner and evaluator depend strongly (more strongly than they should) on specific parses and representations in some places. I believe I found all such places and changed them, but am not confident.


crates/vm/src/eval.rs
```diff
@@ -329,6 +347,37 @@ pub mod tests {
...
+    fn test_semijoin() {
+        let p = &mut Program::new(AuthCtx::for_testing());
+        let table = MemTable::from_value(scalar(1));
+        let field = table.get_field_pos(0).unwrap().clone();
+
+        let mut sources = SourceSet::default();
+        let source_expr = sources.add_mem_table(table.clone());
+        let second_source_expr = sources.add_mem_table(table);
+
+        let q = query(source_expr).with_join_inner(second_source_expr, field.clone(), field, true);
```
joshua-spacetime, line 360: Again, I think I would prefer a higher level test that goes from sql to result set. You can use `RelationalDB::create_table_for_test` along with `sql::execute::run`.

crates/vm/src/expr.rs
```diff
@@ -1445,6 +1460,127 @@ impl QueryExpr {
         }
     }
 
+    /// Try to turn an inner join followed by a projection into a semijoin.
```
joshua-spacetime, line 1463: FWIW, this rewrite might not be exercised by the subscription path, since the original plans should all be index joins.

crates/vm/src/expr.rs
```diff
@@ -1445,6 +1460,127 @@ impl QueryExpr {
...
+    ///
+    /// Current limitations of this optimization:
+    /// - The `JoinInner` must be the first (0th) element of the `query`.
+    ///   Future work could search through the `query` to find any applicable `JoinInner`s,
+    ///   but the current implementation inspects only the first expr.
+    ///   This is likely sufficient because this optimization is primarily useful for enabling `try_index_join`,
+    ///   which is fundamentally limited to operate on the first expr.
+    /// - The `Project` must immediately follow the `JoinInner`, with no intervening exprs.
+    ///   Future work could search through intervening exprs to detect that the RHS table is unused.
+    /// - The LHS/source table must be a [`DbTable`], not a [`MemTable`].
```
joshua-spacetime, line 1492: Is this what we want? We would like to apply this optimization to delta table queries.

crates/vm/src/expr.rs
```diff
@@ -1611,12 +1741,14 @@ impl QueryExpr {
...
                 Query::JoinInner(join) => {
-                    q = q.with_join_inner(join.rhs.optimize(row_count), join.col_lhs, join.col_rhs)
+                    q = q.with_join_inner(join.rhs.optimize(row_count), join.col_lhs, join.col_rhs, join.semi);
                 }
                 _ => q.query.push(query),
             };
         }
 
+        // Make sure to `try_semi_join` before `try_index_join`, as the latter depends on the former.
+        let q = q.try_semi_join();
```
joshua-spacetime, line 1751: Ok I think I understand now. The compilation path is now
```
sql -> inner join + project -> semi join -> index join
```
And then for eval_incr, `to_inner_join` will now take an index join and produce an `inner_join{semi: true}`, whereas before it would produce an `inner_join + project`

Is this correct?

gefjon: Yes, that's correct.

joshua-spacetime (changes requested):
I just want to request one more test case? That we correctly transform an IndexJoin between delta tables to the corresponding semijoin.

joshua-spacetime:
But I should add that this change does improve the performance of incremental join exactly as expected.
```
incr-join               time:   [3.1750 µs 3.1777 µs 3.1807 µs]
                        change: [-27.737% -27.633% -27.519%] (p = 0.00 < 0.05)
                        Performance has improved.
```

crates/vm/src/expr.rs
gefjon: I expanded the comment to that effect.

crates/vm/src/expr.rs
```diff
@@ -2389,4 +2513,225 @@ mod tests {
...
+                    }),
+                ],
+                Some(TableId(1)),
+            );
+        let optimized = q.clone().optimize(&|_, _| 0);
+        assert_eq!(q, optimized);
+    }
+
+    #[test]
+    /// Tests that [`QueryExpr::optimize`] will not rewrite inner joins with an expression that uses the RHS fields before the projection.
```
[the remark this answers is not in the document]
gefjon: Will remove.

crates/vm/src/expr.rs
gefjon: Expanded comment here.

crates/vm/src/expr.rs
gefjon: Previously, this logic was part of `try_index_join`; it's now its own method. We still get here for SQL semijoins, since the SQL compiler emits a `JoinInner` followed by a `Project`. It's just that `try_index_join` used to recognize that whole pattern.

crates/core/src/vm.rs
```diff
@@ -702,6 +703,55 @@ pub(crate) mod tests {
...
+
+        let schema = TableDef::from_product("test", head).into_schema(table_id);
+
+        let data = MemTable::from_value(scalar(1u64));
+        let rhs = data.get_field_pos(0).unwrap().clone();
+
+        let mut sources = SourceSet::default();
+        let rhs_source_expr = sources.add_mem_table(data);
+
+        let q = query(&schema).with_join_inner(rhs_source_expr, FieldName::positional("inventory", 0), rhs, true);
```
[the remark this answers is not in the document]
joshua-spacetime: As you pointed out, we already have test coverage of this part, so these are redundant. Feel free to remove.

joshua-spacetime (approved):

PR #977, opened 2024-03-14
core: Make `init_module_host` consistent with `update_module_host`

kim:
`spawn_module_host` was changed in #904 to ensure that the supplied program is indeed the database's current program. There are, however, uses which would call `spawn_module_host`, followed by either `init_database` or `update_database` on the returned `ModuleHost`.

This would fail because the condition is not met.

While we cannot prevent misuse, this patch documents the various lifecycle methods, and adjusts the signature of `init_module_host` to be (somewhat) consistent with `update_module_host`. Namely, it will optionally return the result of the `init` reducer call.

---

It is probably preferable to re-arrange the error types and not use `anyhow`. I will leave that to a follow-up.

### DIFF
--- a/crates/core/src/host/host_controller.rs
+++ b/crates/core/src/host/host_controller.rs
@@ -6,7 +6,7 @@ use crate::host;
 use crate::messages::control_db::HostType;
 use crate::module_host_context::{ModuleCreationContext, ModuleHostContext};
 use crate::util::spawn_rayon;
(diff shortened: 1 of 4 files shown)


kazimuth (approved):
I think this makes sense. The nested error types are slightly confusing but the signature is structurally correct.

coolreader18 (approved):

PR #983, opened 2024-03-15
Remove `#[tracing::instrument...]` from `InstanceEnv::*_by_col_eq`

Centril:
# Description of Changes

See diff.


cloutiertyler (approved):
We are removing tracing that happens way too often and has significant overhead.

`iter_by_col_eq` takes like 1.5 us and tracing takes most of that time 1.2 us or so according to tracy.

PR #985, opened 2024-03-15
Commitlog: panic on fsync failure

kim:
Failure to `fsync(2)` can result in a strange state, in particular if the OS's page cache is used.
The most reasonable thing to do is to crash the application and force it re-read the filesystem state from disk.

Because we sync asynchronously, this means that we need to panic.

Please read the commit messages for more context.
You may also be interested in [this study](https://www.usenix.org/system/files/atc20-rebello.pdf).

### DIFF
--- a/crates/commitlog/src/commitlog.rs
+++ b/crates/commitlog/src/commitlog.rs
@@ -18,6 +18,12 @@ pub struct Generic<R: Repo, T> {
     tail: Vec<u64>,
     opts: Options,
     _record: PhantomData<T>,
+    /// Tracks panics/errors to control what happens on drop.
+    ///
+    /// Set to `true` before any I/O operation, and back to `false` after it
+    /// succeeded. This way, we won't try to perform I/O on drop when it is
+    /// unlikely to succeed, or even has a chance to panic.
+    panicked: bool,
 }
 
 impl<R: Repo, T> Generic<R, T> {
@@ -43,14 +49,16 @@ impl<R: Repo, T> Generic<R, T> {
             tail,
             opts,
             _record: PhantomData,
+            panicked: false,
         })
     }
 
-    /// Write the currently buffered data to disk and rotate segments as
+    /// Write the currently buffered data to storage and rotate segments as
     /// necessary.
     ///
-    /// Note that this does not imply that the data is durable, call
-    /// [`Self::sync`] to flush OS buffers.
+    /// Note that this does not imply that the data is durable, in particular
+    /// when a filesystem storage backend is used. Call [`Self::sync`] to flush
+    /// any OS buffers to stable storage.
     ///
     /// # Errors
     ///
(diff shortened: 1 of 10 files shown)


kazimuth (commented):
I think this makes sense so far. Panicking here seems like the best option generally speaking. Unfortunately, as you say, we would need to review what happens after the panic in the context of the larger/multi-tenant environment.

I think it would be worthwhile to merge this before switching to O_DIRECT. We can regard the page-cache-reboot issue as a separate bug -- panicking on fsync seems like a very good idea even if we don't have O_DIRECT, from reading the paper you linked.

I wonder how hard it would be to inject fs errors into a running instance of stdb...

kim:
I realize that the `Drop` impl for `commitlog::Generic` (not shown in the diff) can panic, unhelpfully. Instead, we probably want to flush the current commit to the current segment (regardless of max size).

kim:
I tested how standalone behaves by randomly panicking in the durability impl.

Given `panic = "unwind"`, it results in a permanently unusable database, as it seems impossible to bubble up the panic (or an `Err` result, for that matter). Setting the module trapped doesn't help, as that won't discard the database.

Honestly, I don't know if there is a solution without reworking the whole lifecycle management (which we need to do anyways). Unless @coolreader18 has an idea?

crates/commitlog/src/commitlog.rs
```diff
@@ -63,32 +88,48 @@ impl<R: Repo, T> Generic<R, T> {
...
             warn!("Commit failed: {e}");
+            // Nb.: Don't risk a panic by calling `self.sync()`.
+            // We already gave up on the last commit, and will retry it next time.
             self.start_new_segment()?;
             Err(e)
         } else {
             Ok(sz)
-        }
+        };
+        self.panicked = false;
```
coolreader18, line 114: This doesn't run even if its a normal error

crates/commitlog/src/commitlog.rs
```diff
@@ -137,62 +178,77 @@ impl<R: Repo, T> Generic<R, T> {
...
-            last_offset,
+            last_commit: CommitInfo::Initial { next_offset },
             last_error: None,
         }
     }
 
     pub fn reset(mut self) -> io::Result<Self> {
         info!("hard reset");
 
+        self.panicked = true;
```
coolreader18, line 189: true or false?

crates/commitlog/src/commitlog.rs
```diff
@@ -137,62 +178,77 @@ impl<R: Repo, T> Generic<R, T> {
...
         }
+        // Prevent finalizer from running by not updating self.panicked.
 
         Self::open(self.repo.clone(), self.opts)
     }
 
     pub fn reset_to(mut self, offset: u64) -> io::Result<Self> {
         info!("reset to {offset}");
 
+        self.panicked = true;
```
coolreader18, line 204: & this

coolreader18 (approved):

crates/commitlog/src/commitlog.rs
kim: Yes, because we might have deleted / truncated the segment the buffered commits are supposed to be appended to.

NOTE TO SELF: maybe we could flush before resetting.

PR #986, opened 2024-03-15
Don't do alignment computations during BFLATN ser/de

gefjon:
# Description of Changes

`AlgebraicTypeLayout` and friends already include full layout information, including properly-aligned offsets for `ProductTypeElementLayout`s. As such, there's no need to do any alignment computation during `serialize_value` or `write_value`.

Instead, while traversing a `ProductTypeLayout`, we can use each element's `offset` to update the `curr_offset`.

N/a

3 - minor change, but in close contact with some `unsafe` code.

# Testing

- [x] `table` tests pass, including proptests.
- [x] Wait for full CI to pass, incl. `core` tests and smoketests.


gefjon:
benchmarks please

joshua-spacetime (approved):
```
full-scan               time:   [59.269 ms 59.444 ms 59.639 ms]
                        change: [-28.386% -28.129% -27.845%] (p = 0.00 < 0.05)
                        Performance has improved.

full-join               time:   [210.61 µs 210.80 µs 211.01 µs]
                        change: [-13.686% -13.497% -13.312%] (p = 0.00 < 0.05)
                        Performance has improved.
```

PR #987, opened 2024-03-15
Remove useless usages of RowCount

mamcx:
# Description of Changes

`RowCount` is meant to be used to estimate the rows of a query, but we can't do that yet. Remove their uses in `Select` and others where the calculation is useless because in `collect_vec` is asked before is known the number of rows.

1

### DIFF
--- a/crates/vm/src/rel_ops.rs
+++ b/crates/vm/src/rel_ops.rs
@@ -12,7 +12,9 @@ pub(crate) trait ResultExt<T> {
 /// A trait for dealing with fallible iterators for the database.
 pub trait RelOps<'a> {
     fn head(&self) -> &Arc<Header>;
-    fn row_count(&self) -> RowCount;
+    fn row_count(&self) -> RowCount {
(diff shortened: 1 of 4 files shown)


crates/vm/src/rel_ops.rs
```diff
@@ -12,7 +12,9 @@ pub(crate) trait ResultExt<T> {
 /// A trait for dealing with fallible iterators for the database.
 pub trait RelOps<'a> {
     fn head(&self) -> &Arc<Header>;
-    fn row_count(&self) -> RowCount;
+    fn row_count(&self) -> RowCount {
```
joshua-spacetime, line 15: Should we remove it entirely or is it being used somewhere else?

mamcx: Is used in `RelOps::collect_vec` and other few places. Only beneficial for projections and queries without predicates.

joshua-spacetime (approved):

PR #993, opened 2024-03-19
drop subscription laster

Shubham8287:
# Description of Changes

corresponding GitHub label.

# Testing


crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -86,7 +86,6 @@ impl ModuleSubscriptions {
         subscriptions.remove_subscription(&sender.id.identity);
         subscriptions.add_subscription(sender.clone(), execution_set.into_iter());
         let num_queries = subscriptions.num_queries();
-        drop(subscriptions);
```
mamcx, line 89: Here or elsewhere an explanation of why this change?

joshua-spacetime:
Closing as this was transferred to #997.

PR #994, opened 2024-03-19
eval_incr: try to keep references to PVs till bsatn, not owned

Centril:
# Description of Changes

## Changes

Closes https://github.com/clockworklabs/SpacetimeDB/issues/998.

- Add `RelValue::ProjRef(&'a ProductValue)` to avoid cloning PVs. Various cow-like types are added so that we can instead go from a reference to a PV to bsatn in incremental evaluation.
- As a consequence to the above, the way sources are passed is changed to be generic, with the trait `SourceProvider`. An in-memory table is now any `Iterator<Item = RelValue<'a>>`.

```
Benchmarking incr-select: Collecting 100 samples in estimated 5.0000 
incr-select             time:   [238.90 ns 240.17 ns 241.72 ns]
                        change: [-45.577% -45.329% -45.079%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 13 outliers among 100 measurements (13.00%)
  3 (3.00%) high mild
  10 (10.00%) high severe

Benchmarking incr-join: Collecting 100 samples in estimated 5.0065 s 
incr-join               time:   [1.9816 µs 1.9931 µs 2.0088 µs]
                        change: [-25.185% -24.919% -24.651%] (p = 0.00 < 0.05)
                        Performance has improved.
```

Relative to before 1st commit, on i7-7700K, 64gb ram.

None

2

# Testing

No semantic changes, existing tests should cover this.

### DIFF
--- a/crates/bench/benches/subscription.rs
+++ b/crates/bench/benches/subscription.rs
@@ -130,6 +130,8 @@ fn eval(c: &mut Criterion) {
     );
     bench_eval(c, "full-join", &name);
 
+    let ctx_incr = &ExecutionContext::incremental_update(db.address());
+
     // To profile this benchmark for 30s
     // samply record -r 10000000 cargo bench --bench=subscription --profile=profiling -- incr-select --exact --profile-time=30
     c.bench_function("incr-select", |b| {
@@ -137,13 +139,14 @@ fn eval(c: &mut Criterion) {
         let select_lhs = "select * from footprint";
         let select_rhs = "select * from location";
         let auth = AuthCtx::for_testing();
-        let tx = db.begin_tx();
-        let query_lhs = compile_read_only_query(&db, &tx, &auth, select_lhs).unwrap();
-        let query_rhs = compile_read_only_query(&db, &tx, &auth, select_rhs).unwrap();
+        let tx = &db.begin_tx();
+        let query_lhs = compile_read_only_query(&db, tx, &auth, select_lhs).unwrap();
+        let query_rhs = compile_read_only_query(&db, tx, &auth, select_rhs).unwrap();
         let query = ExecutionSet::from_iter(query_lhs.into_iter().chain(query_rhs));
+        let tx = &tx.into();
 
         b.iter(|| {
-            let out = query.eval_incr(&db, &tx, &update).unwrap();
+            let out = query.eval_incr(ctx_incr, &db, tx, &update).unwrap();
             black_box(out);
         })
     });
@@ -159,12 +162,13 @@ fn eval(c: &mut Criterion) {
             where location.chunk_index = {chunk_index}"
         );
         let auth = AuthCtx::for_testing();
-        let tx = db.begin_tx();
-        let query = compile_read_only_query(&db, &tx, &auth, &join).unwrap();
+        let tx = &db.begin_tx();
+        let query = compile_read_only_query(&db, tx, &auth, &join).unwrap();
         let query: ExecutionSet = query.into();
+        let tx = &tx.into();
 
         b.iter(|| {
-            let out = query.eval_incr(&db, &tx, &update).unwrap();
+            let out = query.eval_incr(ctx_incr, &db, tx, &update).unwrap();
             black_box(out);
         })
     });
--- a/crates/core/src/vm.rs
+++ b/crates/core/src/vm.rs
@@ -224,16 +224,16 @@ fn join_inner<'a, I: 'a + Iterator<Item = RelValue<'a>>>(
     )
 }
 
-/// Resolve `query` to a table iterator, either an [`IterRows`] or a [`TableCursor`].
+/// Resolve `query` to a table iterator,
+/// either taken from an in-memory table, in the case of [`SourceExpr::InMemory`],
+/// or from a physical table, in the case of [`SourceExpr::DbTable`].
 ///
-/// If `query` refers to a `MemTable`, this will `Option::take` said `MemTable` out of `sources`,
-/// leaving `None`.
-/// This means that a query cannot refer to a `MemTable` multiple times,
-/// nor can a `SourceSet` which contains a `MemTable` be reused.
-/// This is because [`IterRows`] takes ownership of the `MemTable`.
+/// If `query` refers to an in memory table,
+/// `sources` will be used to fetch the table `I`.
+/// Examples of `I` could be derived from `MemTable` or `&'a [ProductValue]`
+/// whereas `sources` could be a closure that takes from a `SourceSet`.
 ///
-/// On the other hand, if the `query` is a `DbTable`, `sources` is unused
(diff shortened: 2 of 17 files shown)


crates/bench/benches/subscription.rs
```diff
@@ -45,7 +40,8 @@ fn insert_op(table_id: TableId, table_name: &str, row: ProductValue) -> Database
     DatabaseTableUpdate {
         table_id,
         table_name: table_name.to_string(),
-        ops: vec![TableOp::insert(row)],
+        inserts: vec![row],
```
mamcx, line 43: Here and elsewhere `TableOp` is replaced, is it correct that the order of operations doesn't matter? I will find it surprising that `delete + insert` == `insert + delete` because the potential of side-effects (aka triggers)

gefjon: We don't have triggers, and AFAIK have no plans to implement them in the near-to-medium-term future. If they're something we're interested in, we'll need to do some design work. In the meantime, I see no reason to make our performance worse in service of a feature that we don't have but might someday add.

Also, notably, we already shuffle the table ops relative to the order they're performed in a TX. `TxState` holds unordered sets of inserts and deletes, and has no way to reconstruct an ordered sequence of mutations.

mamcx: But we have reducers that are alike.

crates/core/src/vm.rs
```diff
@@ -41,12 +41,12 @@ impl<'a> From<&'a Tx> for TxMode<'a> {
...
-    ctx: &'a ExecutionContext,
+pub fn build_query<'a, I: 'a + Iterator<Item = RelValue<'a>>>(
+    ctx: &'a ExecutionContext<'a>,
     stdb: &'a RelationalDB,
-    tx: &'a TxMode,
+    tx: &'a TxMode<'a>,
     query: &'a QueryExpr,
-    sources: &mut SourceSet,
+    sources: &mut impl FnMut(SourceId) -> Option<I>,
 ) -> Result<Box<IterRows<'a>>, ErrorVm> {
```
gefjon, line 50: Needs comment describing the expected behavior of the `sources` function.

gefjon: I won't comment everywhere, but all the other functions that have been changed from `&mut SourceSet` to `&mut impl FnMut(SourceId) -> Option<I>` need the same comment, or at least a "see `build_query` for explanation of this argument."

crates/core/src/vm.rs
```diff
@@ -268,6 +267,18 @@ fn get_table<'a>(
     })
 }
 
+fn build_source_query<'a, I: 'a + Iterator<Item = RelValue<'a>>>(
```
gefjon, line 270: I find this name unhelpful; why does a "source query" need to be a `MemTable`? Surely compound queries are allowed to have `DbTable` sources.

crates/vm/src/expr.rs
```diff
@@ -273,57 +274,53 @@ impl From<Query> for Option<ColumnOp> {
...
 /// This allows query plans to be re-used, though each execution requires a new `SourceSet`.
 ///
-/// Internally, the `SourceSet` stores an `Option<MemTable>` for each planned [`SourceId`].
+/// Internally, the `SourceSet` stores an `Option<T>` for each planned [`SourceId`].
 /// During execution, the VM will [`Option::take`] the [`MemTable`] to consume them.
 /// This means that a query plan may not include multiple references to the same [`SourceId`].
-pub struct SourceSet(Vec<Option<MemTable>>);
+#[derive(Debug, PartialEq, Eq, Clone)]
+#[repr(transparent)]
+pub struct SourceSet<T, const N: usize>(ArrayVec<Option<T>, N>);
```
gefjon, line 290: Why the new `const N: usize` parameter? Do we have evidence that `ArrayVec` is meaningfully faster here than just `Vec`? Propagating the const parameter through everything that uses `SourceSet` adds a lot of noise to our code which I'd prefer to avoid without a good reason.

crates/core/src/vm.rs
gefjon: I'd also like this comment to strongly justify why these functions don't just take a `SourceSet` or `&mut SourceSet`, as I see this closure as a major increase in complexity and a barrier to at-a-glance understanding.

crates/vm/src/expr.rs
```diff
@@ -273,57 +274,53 @@ impl From<Query> for Option<ColumnOp> {
...
-/// Rather than embedding [`MemTable`]s in query plans, we store a `SourceExpr::MemTable`,
+/// Rather than embedding tables in query plans, we store a `SourceExpr::MemTable`,
 /// which contains the information necessary for optimization along with a [`SourceId`].
 /// Query execution then executes the plan, and when it encounters a `SourceExpr::MemTable`,
-/// retrieves the [`MemTable`] from the corresponding `SourceSet`.
+/// retrieves the `T` table from the corresponding `SourceSet`.
 /// This allows query plans to be re-used, though each execution requires a new `SourceSet`.
 ///
-/// Internally, the `SourceSet` stores an `Option<MemTable>` for each planned [`SourceId`].
+/// Internally, the `SourceSet` stores an `Option<T>` for each planned [`SourceId`].
```
gefjon, line 285: This comment needs to be revised to describe sensible choices for `T`, and what parts of our codebase use those values. In short, I want the comment to justify why this is generic over `T` rather than simply saying `MemTable`; the current comment is written in a way that makes me go, "What could a `SourceExpr::MemTable` ever refer to other than a `MemTable`?"

gefjon (changes requested):
In addition to the below comments, I need you to update your PR description to say what each of the 3 changes you're making (i.e. the 3 commits) do, not just how much they improve performance.

gefjon (approved):
I'd like review from @joshua-spacetime and/or @mamcx since I consider them the owners of the VM, but I'm happy.

mamcx:
LGTM code-wise, but I think this should be tested with bots to see if anything breaks.

joshua-spacetime (changes requested):
I thought we agreed to split this up, and I have to ask that we still do.

I expect the first commit to significantly improve the performance of `incr-select` and to a lesser extent `incr-join`. But it's unclear what the rest of the commits achieve, and there are too many changes for me to be able to make that determination.

I would really like a much smaller change set that addresses #998 specifically and exclusively.

joshua-spacetime (changes requested):
We have to be quite careful about what we merge at this point. I'd feel a lot better if there were still fewer changes. The changes to `RelValue` and `TableOpCow` make sense. But there are a lot of others, like changes to SourceSet, that look to be unnecessary.

joshua-spacetime (approved):

joshua-spacetime:
Closed by #1019

PR #995, opened 2024-03-20
test: Subscriptions should not drop read lock early

joshua-spacetime:
If a subscription drops its read lock on the database too early, that is before it sends its updates to the client, this test will fail.

# Description of Changes

corresponding GitHub label.

# Testing

### DIFF
--- a/crates/core/src/subscription/module_subscription_actor.rs
+++ b/crates/core/src/subscription/module_subscription_actor.rs
@@ -4,7 +4,7 @@ use super::query::compile_read_only_query;
 use super::subscription::ExecutionSet;
 use crate::client::messages::{SubscriptionUpdate, SubscriptionUpdateMessage, TransactionUpdateMessage};
 use crate::client::{ClientActorId, ClientConnectionSender};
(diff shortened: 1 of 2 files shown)


crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -166,3 +175,73 @@ impl ModuleSubscriptions {
...
+                    request_id: 0,
+                },
+                Instant::now(),
+                Some(Arc::new(move |tx: &_| {
+                    std::thread::sleep(Duration::from_secs(1));
+                    let ctx = ExecutionContext::default();
+                    // Assuming subscription evaluation holds a lock on the db,
+                    // any mutations to T will necessarily occur after,
+                    // and therefore we should only see a single row returned.
+                    assert_eq!(1, db.iter(&ctx, tx, table_id).unwrap().count());
```
joshua-spacetime, line 230: Note this test will fail without the fix in #997.

Shubham8287 (approved):
Really a good test, We should write tests  for validation synchronization correctness.

PR #997, opened 2024-03-20
fix(996): Do not release database lock early for subscriptions

### LINKED ISSUE #996: Subscriptions can send messages out of order
The following ordering is currently possible
```
Initial subscription evaluated 
    -> Tx starts
    -> Tx commits
    -> updates for Tx sent to client
    -> updates for initial subscription sent to client
```

That is, it is possible for a transaction to execute and send its updates to the client after an initial subscription has been evaluated but before the initial subscription has sent its results to the client. This will result in the client receiving an inconsistent view of the database.

joshua-spacetime:
Fixes #996.

# Description of Changes

corresponding GitHub label.

# Testing


crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -85,11 +85,9 @@ impl ModuleSubscriptions {
         // This also makes it possible for `broadcast_event` to get scheduled before the subsequent part here
         // but that should not pose an issue.
         let mut subscriptions = self.subscriptions.write();
-        drop(tx);
```
Shubham8287, line 88: I feel  we do need to `drop(tx)` early for any correction issue, we can do it being more conserative.

dropping subscription lock should be enough as that will also not let the reducers to commit (reducer need subscription read lock to commit), so their `eval_incr` will also not start.

if we not `drop(tx)` early, reducer execution will not happen parallel to `eval`, though right now we are letting execution to happen but not commit (which is also fine).

At the end, it depends whether we want to be more conservative or look for bit more perfomance.

joshua-spacetime: We discussed async: Holding the tx for the lifetime of the call does keep a reducer from executing until we push onto the send queue, but that shouldn't have a noticeable impact on throughput, plus it makes testing easier. So I think we're both in favor of this from a correctness standpoint.

Shubham8287 (approved):

PR #1001, opened 2024-03-21
test: Multi-column index selection through query macro

joshua-spacetime:
# Description of Changes

corresponding GitHub label.

# Testing


crates/testing/tests/standalone_integration_test.rs
```diff
@@ -139,3 +139,27 @@ fn test_call_query_macro() {
         },
     );
 }
+
+#[test]
+#[serial]
+fn test_index_scans() {
```
gefjon, line 145:
```suggestion
/// Assert that all of the timing spans in the `perf-test` module's `test_index_scans` reducer
/// are measured in micro- or nanoseconds.
///
/// This timing is a proxy for use of a multi-column index,
/// under the assumption that, because indexes provide an algorithmic speedup,
/// failing to use the index will result in execution being orders of magnitude worse,
// and taking milliseconds.
fn test_index_scans() {
```

gefjon (commented):
Could you add a `README.md` in the new module which says which test it's used for? I know we haven't been very good about including these so far, but I'd at least like to have them for any new modules we add.

crates/table/src/pointer_map.rs
```diff
@@ -234,8 +237,6 @@ impl PointerMap {
     ///
     /// Handles any hash conflicts for `hash`.
     pub fn insert(&mut self, hash: RowHash, ptr: RowPointer) -> bool {
-        debug_assert!(self.maintains_invariants());
```
joshua-spacetime, line 237: I'm not sure why these were left in, but they make populating a table quadratic time.

crates/testing/tests/standalone_integration_test.rs
```diff
@@ -139,3 +139,44 @@ fn test_call_query_macro() {
...
+            let json = r#"{"call": {"fn": "load_location_table", "args": []}}"#;
+            module.send(json.to_string()).await.unwrap();
+
+            let json = r#"{"call": {"fn": "test_index_scan_on_id", "args": []}}"#;
+            module.send(json.to_string()).await.unwrap();
+
+            let json = r#"{"call": {"fn": "test_index_scan_on_chunk", "args": []}}"#;
+            module.send(json.to_string()).await.unwrap();
+
+            let json = r#"{"call": {"fn": "test_index_scan_on_x_z_dimension", "args": []}}"#;
```
joshua-spacetime, line 162: This test will not pass without #1012

mamcx (approved):
LGTM

PR #1013, opened 2024-03-21
Preserve debuginfo (e.g. symbols) in wasm-opt

coolreader18:
# Description of Changes

This gives much better information for backtraces, and if the user does want to strip all symbols, they can change the `profile.release.strip` option in their `Cargo.toml`.

1


mamcx (approved):
LGTM

PR #1014, opened 2024-03-21
Prune deps from bindings dependency tree

coolreader18:
# Description of Changes

I noticed that somehow `protobuf` got into the transitive dependencies of `spacetimedb-bindings`. This brings the dependency count from 99 back to a respectable 56.

This makes modules compile quicker, which is good for our CI as well as first impressions of SpacetimeDB as a product. Speaking for myself, I know that I do often judge crates on dependency bloat, and I think 56 transitive dependencies is much more reasonable for a bindings crate than 99 (esp. when a lot of those had no need to be in there).

1

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -90,7 +90,7 @@ spacetimedb-table = { path = "crates/table", version = "0.8.2" }
 spacetimedb-vm = { path = "crates/vm", version = "0.8.2" }
 
 ahash = "0.8"
-anyhow = { version = "1.0.68", features = ["backtrace"] }
+anyhow = "1.0.68"
 anymap = "0.12"
 arrayvec = "0.7.2"
 async-trait = "0.1.68"
--- a/crates/lib/Cargo.toml
+++ b/crates/lib/Cargo.toml
@@ -15,17 +15,18 @@ name = "serde"
 required-features = ["serde"]
 
 [features]
-default = ["serde"]
+default = ["serde", "metrics"]
 serde = ["dep:serde", "spacetimedb-sats/serde", "dep:serde_with", "chrono/serde"]
 cli = ["clap"]
 # Allows using `Arbitrary` impls defined in this crate.
 proptest = ["dep:proptest", "dep:proptest-derive"]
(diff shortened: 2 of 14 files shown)


Cargo.toml
```diff
@@ -90,7 +90,7 @@ spacetimedb-table = { path = "crates/table", version = "0.8.2" }
 spacetimedb-vm = { path = "crates/vm", version = "0.8.2" }
 
 ahash = "0.8"
-anyhow = { version = "1.0.68", features = ["backtrace"] }
+anyhow = "1.0.68"
```
coolreader18, line 93: The `backtrace` crate was pulling in `libc` and `cc`

crates/lib/Cargo.toml
```diff
@@ -15,17 +15,18 @@ name = "serde"
 required-features = ["serde"]
 
 [features]
-default = ["serde"]
+default = ["serde", "metrics"]
 serde = ["dep:serde", "spacetimedb-sats/serde", "dep:serde_with", "chrono/serde"]
 cli = ["clap"]
 # Allows using `Arbitrary` impls defined in this crate.
 proptest = ["dep:proptest", "dep:proptest-derive"]
+metrics = ["dep:spacetimedb-metrics", "spacetimedb-sats/metrics"]
```
coolreader18, line 23: prometheus was pulling in `protobuf`

modules/rust-wasm-test/Cargo.toml
```diff
@@ -12,7 +12,6 @@ bench = false
 
 [dependencies]
 spacetimedb = { path = "../../crates/bindings" }
-spacetimedb-lib = { path = "../../crates/lib" } # TODO: this should not ever be necessary. `spacetimedb` should provide all module dependencies.
```
coolreader18, line 15: This was just cause I used crate to test, and I noticed that depending on lib directly like this enabled the default features and brought the dependency count way back up.

mamcx (approved):
LGTM

kazimuth (approved):
LGTM, is this ready for merge?

PR #1015, opened 2024-03-22
Updated the traits file - should be disallowed

jdetter:
# Description of Changes

corresponding GitHub label.

# Testing


bfops (approved):
"This looks great to me. We should merge."
- Someone who doesn't know what this is

cloutiertyler:
Closing since we verified this is disallowed unless a codeowner reviews.

PR #1016, opened 2024-03-22
Bench for eval add serialization steps

### LINKED ISSUE #999: Update incremental subscription benchmarks to include row serialization
Currently the incremental subscription benchmarks, `incr-join` and `incr-select`, only take into account query execution. They should also include row serialization. Specifically they should include both the cost of serializing product values into bsatn as well as bsatn to protobuf.

mamcx:
# Description of Changes

Closes #999

NOTE: The purpose of this PR was changed after talks with @joshua-spacetime.

OLD:

Another serialization benches for both protobuf & bsatn, using `select * from footprint` * 1_000_000 rows.

NOW:

Include the steps of serialization on the `eval` benches to have a full picture of the costs.

1

# Testing

### DIFF
--- a/crates/bench/benches/subscription.rs
+++ b/crates/bench/benches/subscription.rs
@@ -1,5 +1,8 @@
 use criterion::{black_box, criterion_group, criterion_main, Criterion};
+
+use spacetimedb::client::messages::{SubscriptionUpdate, SubscriptionUpdateMessage};
 use spacetimedb::client::Protocol;
+use spacetimedb::client::{ClientActorId, ClientConnectionSender};
 use spacetimedb::db::relational_db::RelationalDB;
 use spacetimedb::error::DBError;
 use spacetimedb::execution_context::ExecutionContext;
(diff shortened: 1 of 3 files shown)


crates/bench/benches/subscription.rs
```diff
@@ -148,6 +151,57 @@ fn eval(c: &mut Criterion) {
...
+        let out = query.eval_mem_table(&db, &tx).unwrap();
+
+        let out = DatabaseTableUpdate {
+            table_id: 0.into(),
+            table_name: "test".into(),
+            ops: out
+                .into_iter()
+                .flat_map(|x| x.data.into_iter().map(TableOp::insert))
+                .collect(),
+        };
```
joshua-spacetime, line 172: Right now `ExecutionSet::eval` returns a `ProtocolDatabaseUpdate`. Could you add a helper method that goes all the way to `DataMessage`?

What I'd like to benchmark is
```
ExecutionSet::eval
    -> ProtocolDatabaseUpdate (query execution)
    -> SubscriptionUpdateMessage (see ModuleSubscriptions::add_subscriber)
    -> DataMessage (see ClientConnectionSender::send_message)
```

That is, the full execution + the full serialization before we send over the wire.

crates/bench/benches/subscription.rs
```diff
@@ -148,6 +151,57 @@ fn eval(c: &mut Criterion) {
...
+            |out| {
+                let _ = TableUpdate::from(out);
+            },
+            criterion::BatchSize::PerIteration,
+        );
+    });
+
+    //To profile this benchmark for 30s
+    //samply record -r 10000000 cargo bench --bench=subscription --profile=profiling -- select-bsatn --exact --profile-time=30
+    c.bench_function("select-bsatn", |b| {
```
joshua-spacetime, line 185: I'd rather the benchmarks cover both bsatn and protobuf instead of individual benchmarks for each.

mamcx: Ok, I reworked the benches adding the `send_message` that executes the full pipeline and removing the different

PR #1017, opened 2024-03-22
Shub/remove lifetime from ctx

Shubham8287:
# Description of Changes
- remove lifeitme depency from `ExecutionContext` by making reducer_name from `&str` -> `String` with only one allocation per reducer call.
- Move Context inside TxSlot so it can be re-usable between ABIs & host for same call.

### DIFF
--- a/crates/core/src/host/instance_env.rs
+++ b/crates/core/src/host/instance_env.rs
@@ -33,6 +33,7 @@ pub struct InstanceEnv {
 #[derive(Clone, Default)]
 pub struct TxSlot {
     inner: Arc<Mutex<Option<MutTxId>>>,
+    ctx: Arc<Mutex<Option<ExecutionContext>>>,
 }
 
 #[derive(Default)]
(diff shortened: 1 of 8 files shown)


crates/core/src/host/wasmtime/wasm_instance_env.rs
```diff
@@ -153,15 +154,15 @@ impl WasmInstanceEnv {
...
+    fn reducer_context<'a>(&'a self) -> Result<impl DerefMut<Target = ExecutionContext> + 'a, WasmError> {
+        self.instance_env().get_ctx().map_err(|err| WasmError::Db(err.into()))
     }
 
     // TODO: make this part of cvt(), maybe?
     /// Gather the appropriate metadata and log a wasm_abi_call_duration_ns with the given AbiCall & duration
     #[cfg(feature = "metrics")]
     fn start_abi_call_timer(&self, call: AbiCall) -> prometheus::HistogramTimer {
-        let ctx = self.reducer_context();
+        let ctx = self.reducer_context().unwrap();
```
Shubham8287, line 165: Todo: handle this before mering.

crates/core/src/host/instance_env.rs
```diff
@@ -33,6 +34,7 @@ pub struct InstanceEnv {
 #[derive(Clone, Default)]
 pub struct TxSlot {
     inner: Arc<Mutex<Option<MutTxId>>>,
+    ctx: Arc<Mutex<Option<ExecutionContext>>>,
```
joshua-spacetime, line 37: Does it need to be wrapped in an Option?

joshua-spacetime (approved):

PR #1018, opened 2024-03-22
Implement 'sql' parsing for simple 'Enums'

mamcx:
# Description of Changes

Following the way enums are [used in postgres](https://www.postgresql.org/docs/current/datatype-enum.html), which are simple enums, we can now parse them and query for them:

It means if we have:

```rust
#[derive(spacetimedb::SpacetimeType, Clone, Copy, PartialEq, Debug)]
#[repr(i32)]
pub enum Role {
    Player,
    Admin,
}

#[spacetimedb(table)]
#[derive(Clone, Debug)]
pub struct IdentityRole {
    #[primarykey]
    pub identity: Identity,
    pub email: String,
    pub role: Role,
}
```

Now we can execute:

```sql
select * from IdentityRole where role = 'Player';
update IdentityRole set role =  'Admin' where email = 'test@clockworklabs';
```

Note: We can't use enums for `create table` or anywhere we could use the `type` of the enum on `sql`. It only can match the `value` specified by the enum from a module.

1
# Testing

- [x] *compile_enum_field*

### DIFF
--- a/crates/core/src/sql/ast.rs
+++ b/crates/core/src/sql/ast.rs
@@ -12,7 +12,7 @@ use spacetimedb_sats::{AlgebraicType, AlgebraicValue, ProductTypeElement};
 use spacetimedb_vm::errors::ErrorVm;
 use spacetimedb_vm::expr::{ColumnOp, DbType, Expr};
 use spacetimedb_vm::operator::{OpCmp, OpLogic, OpQuery};
-use spacetimedb_vm::ops::parse::parse;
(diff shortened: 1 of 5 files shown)


kazimuth (approved):
This is unambiguous, right? Because we have the type, this will never parse something that should be a string as an enum?

If so, LGTM.

mamcx:
Is unambiguous as long we have the `column = 'string'` (ie: should be ok for comparisions)

PR #1020, opened 2024-03-25
Bump Rust to 1.77 + fix warnings + use `Bound::map`

Centril:
# Description of Changes

- Bumps Rust to stable 1.77, https://blog.rust-lang.org/2024/03/21/Rust-1.77.0.html.
- Fixes some warnings that come as a result.
- Uses `Bound::map`, now available on stable.


gefjon (approved):

bfops:
(adding `breaks-library-compatibility` because I believe this requires changes by users of this library, correct me if I'm wrong?)

PR #1022, opened 2024-03-26
perf(1021): Compress subscription update messages

### LINKED ISSUE #1021: Compress protobuf messages using brotli
`SubscriptionUpdateMessage`s have the potential to be orders of magnitude larger than `TransactionUpdateMessage`s and could benefit from compression to reduce network latency. We should use compression level 1 (fastest).

Update(3-26-2024): We are going to compress all protobuf messages and measure the performance.

joshua-spacetime:
Closes #1021.

Subscription update messages can be quite large.
As such they can benefit a great deal from compression.

Unfortunately tungstenite does not support compression, so we add support for it here separately.

We chose Brotli for its universal support,
and its better compression ration compared to gzip.

# Description of Changes

corresponding GitHub label.

# Testing

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -50,6 +50,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "alloc-no-stdlib"
+version = "2.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cc7bb162ec39d46ab1ca8c77bf72e890535becd1751bb45f64c597edb4c8c6b3"
+
+[[package]]
+name = "alloc-stdlib"
+version = "0.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


gefjon:
I have some concerns about your implementation. It looks like you've set things up so that, rather than the binary protocol always receiving an uncompressed Protobuf `Message`, it now receives an untagged byte blob which may be either compressed or uncompressed. This seems problematic; I don't know how a client is supposed to know whether to immediately Protobuf-decode a message or to decompress it first and then decode it, short of unconditionally attempting to decompress, detecting an error, and Protobuf-decoding the raw message as a fallback.

I have an alternate implementation in #1023 which encloses the compressed `SubscriptionUpdate` message within the Protobuf `Message` so that the client can dispatch on the `type` to determine whether decompression is necessary. I also find this appealing because it's more backwards-compatible; your version will fail if a newer client connects to an older server, whereas mine won't.

PR #1031, opened 2024-03-27
Bump to reqwest 0.12 (uses hyper 1.0)

coolreader18:
# Description of Changes

Now we're fully onto hyper 1.0 in our deps tree!

1


mamcx (approved):
LGTM

PR #1033, opened 2024-03-29
Bot test driver script

kazimuth:
# Description of Changes

To allow running bot tests with a comment on a PR, like the other benchmarks.

Small PR but it took a lot of finagling. There might still be an extant tracy issue...

complexity 2, risk 0

### DIFF
--- a/crates/bench/bottest.sh
+++ b/crates/bench/bottest.sh
@@ -1,0 +1,123 @@
+#!/bin/bash
+
+# script to run a bot test. This is its own script for 2 reasons:
+# 1. you can run it off CI
+# 2. github actions can't run things in parallel anyway, so we need a shell script to run
+#    the spacetime server, tracy capture, and the bots in parallel.
+
+# move into project root, keep paths stable
+cd "$(dirname "$0")/../.."
+
+# color constants
+RED='\033[0;31m'
+NC='\033[0m'
+
+function assert_dir {
+    if [ ! -d "$1" ]; then
+        echo -e "${RED}$1 does not exist${NC}"
+        exit 1
+    fi
+}
+function assert_file {
+    if [ ! -f "$1" ]; then
+        echo -e "${RED}$1 does not exist${NC}"
+        exit 1
+    fi
+}
+function cleanup {
+    kill $SPACETIME_PROCESS
+    kill $TRACY_PROCESS
+    git checkout -- .
+}
+
+BOTS_DIR="${BOTS_DIR:-$HOME/bots}"
+TRACY_CAPTURE_BIN="${TRACY_CAPTURE_BIN:-$HOME/tracy/capture/build/unix/capture-release}"
+OUT_DIR="$PWD/crates/bench/bottest"
+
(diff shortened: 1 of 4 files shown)


kazimuth:
Man, I can't figure out why tracy isn't correctly producing a log. Sometimes it works, sometimes it doesn't. Really it doesn't seem set up for automated deployments like this. I think right now the issue is it's ignoring a SIGINT, which is supposed to shut it down and write out the log, but possibly there's also another issue related to spacetime server restarts. I'll try rearranging the script to let it shut down on its own...

crates/bench/bottest.sh
```diff
@@ -0,0 +1,128 @@
...
+
+echo ------- PREPARING WORLD -------
+
+target/release/spacetime start --listen-addr 0.0.0.0:3000 "$OUT_DIR/.spacetime" >"$OUT_DIR/spacetime_deploy_world.log" 2>&1 &
+SPACETIME_PROCESS=$!
+sleep 5
+
+target/release/spacetime publish -c bitcraft --wasm-file "$BOTS_DIR/bitcraft_spacetimedb_with_wasm_opt.wasm"
+
+python3 "$BOTS_DIR/deploy_world.py" -H http://127.0.0.1:3000 -m bitcraft -f "$BOTS_DIR/Spacetime128x128.snapshot" >"$OUT_DIR/deploy_world.log" 2>&1
```
bfops, line 99: I think we might be okay with running a smaller world, at least in the typical case? not sure

kazimuth: If you could send me a smaller world I can swap it out. But, it actually uploads in about 15 minutes so I don't think it's a huge problem.

bfops: yeah fair enough. I guess there's also a question about how to maintain the world snapshot? I believe it's essentially just a replay log of queries, so it'll change with some PRs.

kazimuth: John is providing snapshots and other support stuff for bot tests on gdrive periodically, but it wants to be in source control or something soon, maybe with git lfs.

kazimuth:
It looks like the issue is that spacetime is failing to reboot. Currently, the script:
1. runs spacetimedb without tracy to upload the world
2. uploads the world
3. shuts down spacetimedb
4. starts spacetimedb with tracy
5. runs tracy
6. runs a bot test
7. shuts down spacetimedb and tracy
8. zips up all the logs and tracy trace and uploads them

the problem right now seems to be step 4. Sometimes spacetimedb is failing to restart after being shutdown via a SIGINT (CTRL-C in bash.) The error is the following:

```
Error: database error: IO error: could not acquire lock on "/home/ubuntu/actions-runner-linux-x64-2.314.1/_work/SpacetimeDB/SpacetimeDB/crates/bench/bottest/.spacetime/control_node/control_db/db": Os { code: 11, kind: WouldBlock, message: "Resource temporarily unavailable" }
```

Not totally sure what to do next here. I could try and avoid restarting spacetimedb and just run tracy the whole time the world is uploading, but then the tracy logs are going to get huge.

cloutiertyler:
Closing in favor of #1255

PR #1036, opened 2024-04-01
Detect unsatisfiable range queries; warn and short-circuit.

gefjon:
# Description of Changes

This commit fixes a panic caused by unsatisfiable range bounds on an index query, e.g. `WHERE x < 5 AND x > 5`. These unsatisfiable bounds made Rust's `BTreeMap` angry (See https://doc.rust-lang.org/src/alloc/collections/btree/search.rs.html#106-124), and panicked. They also represent probable bugs, as it's silly to write a query which statically will return no rows.

With this commit, we detect statically unsatisfiable bounds in two cases:
- When compiling queries, we log a message at `WARN` containing the offending query.
- When evaluating queries, we silently construct an `EmptyRelOps` rather than a real query iterator.

This commit also adds a test that the offending queries can be compiled and executed without panicking, and select no rows.

N/a

corresponding GitHub label.

1

# Testing

- [x] Checked that the logs are emitted by enabling `env_logger` and passing `--nocapture` when running the tests.
- [x] Newly-added test `test_impossible_bounds_no_panic` used to panic but doesn't anymore.

### DIFF
--- a/crates/vm/src/expr.rs
+++ b/crates/vm/src/expr.rs
@@ -1356,8 +1356,18 @@ impl QueryExpr {
                 bounds: (Bound::Unbounded, Bound::Excluded(upper)),
                 ..
             }) if columns == lhs_col_id => {
+                // Queries like `WHERE x < 5 AND x > 5` never return any rows and are likely mistakes.
+                // Detect such queries and log a warning.
+                // Compute this condition early, then compute the resulting query and log it.
+                let is_never = !inclusive && value == upper;
+
                 let bounds = (Self::bound(value, inclusive), Bound::Excluded(upper));
                 self.query.push(Query::IndexScan(IndexScan { table, columns, bounds }));
+
+                if is_never {
+                    log::warn!("Query will select no rows due to equal excluded bounds: {self:?}")
+                }
+
                 self
             }
             // merge with a preceding select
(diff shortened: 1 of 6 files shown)


crates/vm/src/expr.rs
```diff
@@ -1436,14 +1446,24 @@ impl QueryExpr {
                 self.query.push(Query::IndexScan(IndexScan { table, columns, bounds }));
                 self
             }
-            // merge with a preceding lower bounded index scan (inclusive)
+            // merge with a preceding lower bounded index scan (enclusive)
             Query::IndexScan(IndexScan {
```
mamcx, line 1450: This word [exist](https://wikidiff.com/enclusive/inclusive)?

kazimuth (approved):
LGTM. 

Eventually we might want a system for lint passes but I think the places this happens right now make sense.

crates/vm/src/expr.rs
```diff
@@ -1356,8 +1356,18 @@ impl QueryExpr {
                 bounds: (Bound::Unbounded, Bound::Excluded(upper)),
                 ..
             }) if columns == lhs_col_id => {
+                // Queries like `WHERE x < 5 AND x > 5` never return any rows and are likely mistakes.
+                // Detect such queries and log a warning.
+                // Compute this condition early, then compute the resulting query and log it.
+                let is_never = !inclusive && value == upper;
```
joshua-spacetime, line 1362: We should not be creating an index scan plan in this case.

joshua-spacetime (commented):
If you'd like to merge this, that's fine, I'd just like comments around the log lines and the fix that say this is a temporary fix and that we should not be generating invalid index scan plans.

gefjon:
> If you'd like to merge this, that's fine, I'd just like comments around the log lines and the fix that say this is a temporary fix and that we should not be generating invalid index scan plans.

Agreed. I've added comments to that effect. I'll hold off on merging into master for now, though we may include this in the build used for the BitCraft alpha.

crates/vm/src/expr.rs
gefjon: 🤦. I replaced one letter but not the other when fixing a copy-paste error. Pushed a fix.

PR #1037, opened 2024-04-01
Reject large SQL queries

gefjon:
# Description of Changes

Return an error rather than attempting to compile queries longer than 50k bytes.

This avoids a stack overflow when compiling deeply-nested `AND` and `OR` expressions.

...Technically a breaking change, in that a case that used to sometimes-maybe-work is now always an error?

corresponding GitHub label.

1 - stupid hack, but trivial.

# Testing

- [x] Added a test which used to fail but now passes.

### DIFF
--- a/crates/core/src/sql/compiler.rs
+++ b/crates/core/src/sql/compiler.rs
@@ -12,8 +12,16 @@ use std::sync::Arc;
 
 use super::ast::TableSchemaView;
 
+/// DIRTY HACK ALERT: Maximum allowed length, in UTF-8 bytes, of SQL queries.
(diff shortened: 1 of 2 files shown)


gefjon:
> > This avoids a stack overflow when compiling deeply-nested `AND` and `OR` expressions.
> 
> Alternatively, we could use https://docs.rs/stacker/latest/stacker/ as done by rustc in similar scenarios.

I don't think compiling arbitrarily-large SQL expressions is a good idea. A real solution to this problem will be a more robust system for detecting and rejecting large queries, likely using the same infrastructure as we will be using to charge energy.

PR #1039, opened 2024-04-05
Remove dead code in subscription

Centril:
# Description of Changes

Removes some dead code.

Note: If we use `pub(crate)` or the least possible visibility, then this would have been caught by the compiler before, so I recommend avoiding needless `pub`s.


Shubham8287 (approved):
Looks good, nothing much to review.

PR #1040, opened 2024-04-05
use nohasher_hash and ahash instead of siphash13

Centril:
# Description of Changes

- Where possible, use `hashbrown::{HashMap, HashSet}` instead of `std::collections::{HashMap, HashSet}`. The main effect this has is to change the hash function from `SipHash13` to `ahash`. The latter does not have as strong DoS resistance, but it does provide some resistance.
- Where possible, use `nohash_hasher::NoHashHasher`. This hash function is the identity function and works on any integer type that can be transformed into `u64` without truncation. Notably, this includes types like `SequenceId` and `TableId`.
- All hash-map/set related collections and types necessary have been re-exported under `spacetimedb_data_structures::map`, which everyone can use throughout spacetimedb.

Subscription benchmarks (relative to branched-off-master, on i7-7700K, 64GB RAM):

```
full-join               time:   [315.13 µs 319.56 µs 325.98 µs]
                        change: [-4.5782% -3.6939% -2.6674%] (p = 0.00 < 0.05)
                        Performance has improved.

Benchmarking incr-select: Collecting 100 samples in estimated 
incr-select             time:   [179.91 ns 181.29 ns 183.30 ns]
                        change: [-7.6486% -7.0583% -6.2736%] (p = 0.00 < 0.05)
                        Performance has improved.

Benchmarking incr-join: Collecting 100 samples in estimated 5.
incr-join               time:   [1.0394 µs 1.0466 µs 1.0568 µs]
                        change: [-14.568% -13.685% -12.873%] (p = 0.00 < 0.05)
                        Performance has improved.
```

The main reason for `incr-join`'s more substantial perf improvement is likely the use of `HashSet`s within the incr-join algorithm itself.

There are other benchmarks we could run but I've limited myself to these as they were fast to run.

None

2

# Testing

No semantic changes apart from the hash functions, so the current test suite should probably suffice.


crates/core/src/auth/identity.rs
```diff
@@ -1,12 +1,12 @@
...
-use serde::{Deserialize, Serialize};
-use std::time::{Duration, UNIX_EPOCH};
-use std::{collections::HashSet, time::SystemTime};
-
 pub use jsonwebtoken::errors::Error as JwtError;
 pub use jsonwebtoken::errors::ErrorKind as JwtErrorKind;
+use jsonwebtoken::{decode, encode, Header, TokenData, Validation};
 pub use jsonwebtoken::{DecodingKey, EncodingKey};
+use serde::{Deserialize, Serialize};
+use std::collections::HashSet;
```
gefjon, line 7: Why did this `HashSet` not change to hashbrown?

crates/core/src/host/module_host.rs
```diff
@@ -72,7 +72,7 @@ impl DatabaseUpdate {
     }
 
     pub fn from_writes(stdb: &RelationalDB, tx_data: &TxData) -> Self {
-        let mut map: HashMap<TableId, (Vec<ProductValue>, Vec<ProductValue>)> = HashMap::new();
+        let mut map: IntMap<_, (Vec<_>, Vec<_>)> = IntMap::new();
```
gefjon, line 75: I'd ask that you retain the more precise type ascription here, as it makes the code easier to read in contexts where one doesn't have `rust-analyzer` providing inferred types.

crates/core/src/subscription/module_subscription_manager.rs
```diff
@@ -5,17 +5,19 @@ use crate::db::relational_db::RelationalDB;
...
 use smallvec::SmallVec;
 use spacetimedb_client_api_messages::client_api::{TableRowOperation, TableUpdate};
 use spacetimedb_lib::Identity;
 use spacetimedb_primitives::TableId;
-use std::collections::hash_map::Entry;
-use std::collections::{HashMap, HashSet};
 use std::ops::Deref;
 use std::sync::Arc;
 
+type IntMap<K, V> = HashMap<K, V, BuildNoHashHasher<K>>;
```
gefjon, line 19: Isn't this equivalent to `nohash_hasher::IntMap`?

gefjon (approved):
I'm glad you did this, but oh boy was it tedious to review. Anyway, a few questions, but this all looks reasonable and I'm happy to include it.

PR #1041, opened 2024-04-05
Consistently use `--identity` `-i` for identity args to CLI

gefjon:
# Description of Changes

Prior to this commit, some subcommands used `--as-identity` as the long, some of them used `-I` as the short, and there was no clear pattern.

With this commit, we just use `--identity` as the long everywhere, and `-i` as the short.

Breaking change to CLI arg parsing.

2 - I may have missed an arg, or some tool which depends on the previous arg format.

### DIFF
--- a/crates/cli/src/subcommands/call.rs
+++ b/crates/cli/src/subcommands/call.rs
@@ -58,7 +58,7 @@ pub async fn exec(mut config: Config, args: &ArgMatches) -> Result<(), Error> {
     let arguments = args.get_many::<String>("arguments");
     let server = args.get_one::<String>("server").map(|s| s.as_ref());
 
-    let as_identity = args.get_one::<String>("identity");
+    let identity = args.get_one::<String>("identity");
     let anon_identity = args.get_flag("anon_identity");
 
     let address = database_address(&config, database, server).await?;
@@ -69,7 +69,7 @@ pub async fn exec(mut config: Config, args: &ArgMatches) -> Result<(), Error> {
         address.clone(),
         reducer_name
     ));
-    let auth_header = get_auth_header_only(&mut config, anon_identity, as_identity, server).await?;
+    let auth_header = get_auth_header_only(&mut config, anon_identity, identity, server).await?;
     let builder = add_auth_header_opt(builder, &auth_header);
     let describe_reducer = util::describe_reducer(
         &mut config,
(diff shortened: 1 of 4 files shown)


crates/cli/src/subcommands/call.rs
```diff
@@ -58,7 +58,7 @@ pub async fn exec(mut config: Config, args: &ArgMatches) -> Result<(), Error> {
     let arguments = args.get_many::<String>("arguments");
     let server = args.get_one::<String>("server").map(|s| s.as_ref());
 
-    let as_identity = args.get_one::<String>("as_identity");
+    let as_identity = args.get_one::<String>("identity");
```
jdetter, line 61: Super small nit: I would probably rename this variable as well

crates/cli/src/subcommands/describe.rs
```diff
@@ -54,7 +54,7 @@ pub async fn exec(mut config: Config, args: &ArgMatches) -> Result<(), anyhow::E
     let entity_type = args.get_one::<String>("entity_type");
     let server = args.get_one::<String>("server").map(|s| s.as_ref());
 
-    let as_identity = args.get_one::<String>("as_identity");
+    let as_identity = args.get_one::<String>("identity");
```
jdetter, line 57: Super small nit: I would probably rename this variable as well

crates/cli/src/subcommands/publish.rs
```diff
@@ -53,21 +53,15 @@ pub fn cli() -> clap::Command {
...
-        //     Arg::new("as_identity")
-        //         .long("as-identity")
-        //         .short('i')
-        //         .required(false)
-        //         .conflicts_with("anon_identity"),
-        // )
+                .long_help("The identity that should own the database. If no identity is provided, your default identity will be used.")
+                .required(false)
+                .conflicts_with("anon_identity")
+        )
```
jdetter, line 64: Thank you for cleaning this up 👍

crates/cli/src/subcommands/sql.rs
```diff
@@ -65,7 +65,7 @@ pub fn cli() -> clap::Command {
 pub(crate) async fn parse_req(mut config: Config, args: &ArgMatches) -> Result<Connection, anyhow::Error> {
     let server = args.get_one::<String>("server").map(|s| s.as_ref());
     let database = args.get_one::<String>("database").unwrap();
-    let as_identity = args.get_one::<String>("as_identity");
+    let as_identity = args.get_one::<String>("identity");
```
jdetter, line 68: Super small nit: I would probably rename this variable as well

jdetter (approved):
This looks great, thanks phoebe 👍

PR #1044, opened 2024-04-05
`IndexSemijoin::next`: do a bit less cloning and work

Centril:
# Description of Changes

See diff.

Benchmark against based-on master on i7-7700K, 64GB RAM:

```
Benchmarking full-join: Collecting 100 samples in estimated 6.
full-join               time:   [306.04 µs 308.09 µs 310.89 µs]
                        change: [-8.2344% -7.6278% -7.0018%] (p = 0.00 < 0.05)
                        Performance has improved.
```

I tried this a few times just to make sure that the perf improvement wasn't a fluke.

None

1

### DIFF
--- a/crates/core/src/vm.rs
+++ b/crates/core/src/vm.rs
@@ -123,7 +123,10 @@ pub fn build_query<'a>(
                 let index_table = index_side.table_id().unwrap();
                 let index_header = index_side.head();
                 let probe_side = build_query(ctx, stdb, tx, probe_side, sources)?;
-                let probe_col = probe_side.head().column_pos(probe_field);
+                let probe_col = probe_side
+                    .head()
+                    .column_pos(probe_field)
+                    .expect("query compiler should have ensured the column exist");
                 Box::new(IndexSemiJoin {
                     ctx,
                     db: stdb,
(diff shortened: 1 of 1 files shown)


crates/core/src/vm.rs
```diff
@@ -291,8 +292,8 @@ pub struct IndexSemiJoin<'a, 'c, Rhs: RelOps<'a>> {
     /// An iterator for the probe side.
     /// The values returned will be used to probe the index.
     pub probe_side: Rhs,
-    /// The field whose value will be used to probe the index.
-    pub probe_field: &'c FieldName,
+    /// The column whose value will be used to probe the index.
+    pub probe_col: Option<ColId>,
```
joshua-spacetime, line 296: This shouldn't be optional.

kazimuth (approved):
Yeah, it doesn't really surprise me that looking those field names up repeatedly was slow.

Clean and simple, LGTM.

crates/core/src/vm.rs
```diff
@@ -301,7 +302,7 @@ pub struct IndexSemiJoin<'a, 'c, Rhs: RelOps<'a>> {
     pub index_table: TableId,
     /// The column id for which the index is defined.
     pub index_col: ColId,
-    /// Is this a left or right semijion?
+    /// Is this a left or right semi-join?
```
joshua-spacetime, line 305: Technically it's all one word.

PR #1047, opened 2024-04-08
Shrink AV by 8 bytes and AT by 16

Centril:
# Description of Changes

This primarily shrinks the size of types `AlgebraicValue` from 32 bytes to 24 and `AlgebraicType` from 32 to 16.
The size of some other types are also shrunk, notably, table, column, index, sequence, etc. definitions are shrunk by moving from `String` to `Box<str>`.  Moreover, `ProductValue` is shrunk to 16 bytes, a type that is used frequently in hot places.

Benchmarks below relative to based master, run on i7-7700K, 64GB.

```
full-scan               time:   [80.111 ms 80.529 ms 80.979 ms]
                        change: [-14.103% -13.427% -12.763%] (p = 0.00 < 0.05)
                        Performance has improved.

Benchmarking full-join: Collecting 100 samples in estimated 6.
full-join               time:   [301.14 µs 302.01 µs 303.25 µs]
                        change: [-9.6948% -9.1117% -8.5214%] (p = 0.00 < 0.05)
                        Performance has improved.

Benchmarking incr-select: Collecting 100 samples in estimated 
incr-select             time:   [183.21 ns 185.12 ns 188.14 ns]
                        change: [-5.7687% -4.8383% -3.6110%] (p = 0.00 < 0.05)
                        Performance has improved.
```

None

2

# Testing

No semantic changes, existing tests should suffice.

### DIFF
--- a/crates/sats/src/algebraic_value.rs
+++ b/crates/sats/src/algebraic_value.rs
@@ -64,7 +64,7 @@ pub enum AlgebraicValue {
     ///
     /// We box the `MapValue` to reduce size
     /// and because we assume that map values will be uncommon.
-    Map(MapValue),
+    Map(Box<MapValue>),
     /// A [`bool`] value of type [`AlgebraicType::Bool`].
     Bool(bool),
     /// An [`i8`] value of type [`AlgebraicType::I8`].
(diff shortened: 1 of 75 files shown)


crates/sats/src/algebraic_value.rs
```diff
@@ -86,11 +86,11 @@ pub enum AlgebraicValue {
     /// An [`i128`] value of type [`AlgebraicType::I128`].
     ///
     /// We box these up as they allow us to shrink `AlgebraicValue`.
-    I128(i128),
+    I128(Packed<i128>),
     /// A [`u128`] value of type [`AlgebraicType::U128`].
     ///
     /// We box these up as they allow us to shrink `AlgebraicValue`.
-    U128(u128),
+    U128(Packed<u128>),
```
kazimuth, line 93: This may inhibit SIMD optimizations for i/u128. Fortunately, that doesn't matter in the slightest -- packing the types in this PR is way more important.

kazimuth (approved):
Very nice. I see why this took longer than expected, this is the sort of thing that gets everywhere.

My only real concern right now would be that this could result in $O(n^2)$ behavior for loops that repeatedly append to `ArrayValue`s. I'm not aware of such loops off the top of my head, and I don't see any methods changed in this PR that expose such an interface for `ArrayValue`, so I don't think it's a problem.

cloutiertyler (approved):
This is fine in principle, but let's be deferential to higher prio things about merging this.

Issue #1048, opened 2024-04-09
Weaken `AlgebraicValue`'s `Hash` function to ignore enum discriminants


Centril:
The current implementation of `AlgebraicValue` is `#[derive(Hash)]`.
This causes the hash function to hash like so for e.g., `AlgebraicValue::U16(x)` to be:
```rust
hasher.write_u8(8); // The discriminant of `AlgebraicValue::U16` is 8.
hasher.write_u16(x); // The actual value `x: u16` in `::U16`.
```

Instead, based on my discussion with @gefjon, we agreed that, we should only hash each AV's payload and not the tag/discriminant of the AV variant itself. The benefits include:
1. This weakening should be more performant because its less work.
2. This is what the hash function for `RowRef` already does.
3. It's easier to define a hash function for BSATN-encoded rows this way.

The weakening should not introduce more collisions because AVs in e.g., subscriptions follow a schema, so when you have 100 `::U8`s, it does not reduce the rate of collisions when you hash the tag of `U8`.

PR #1057, opened 2024-04-09
Ignore discriminants in `Hash for AV + ArrayValue` + nix len-prefix for PV

Centril:
# Description of Changes

Fixes https://github.com/clockworklabs/SpacetimeDB/issues/1048.

- `Hash for AV` does not include `Discriminant<AV>` anymore.
- `Hash for PV ` is no longer length-prefixed.
- `Hash for ArrayValue` does not include `Discriminant<ArrayValue>` anymore.

In spacetimedb, we only compare hashes between homogeneously-typed values. Therefore, this PR does less work and does in fact not weaken the hash function. Rather, by avoiding to feed in known-equal data into the hasher, we can reduce the rate of conflicts, as we're now packing the same number of "interesting" bits alongside fewer "boring" bits into a hash.

Semantics in the APIs above are broken by this PR.

2

### DIFF
--- a/crates/sats/src/algebraic_value.rs
+++ b/crates/sats/src/algebraic_value.rs
@@ -22,7 +22,7 @@ pub type F64 = decorum::Total<f64>;
 /// These are only values and not expressions.
 /// That is, they are canonical and cannot be simplified further by some evaluation.
 /// So forms like `42 + 24` are not represented in an `AlgebraicValue`.
-#[derive(EnumAsInner, Debug, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, From)]
+#[derive(EnumAsInner, Debug, Clone, Eq, PartialEq, Ord, PartialOrd, From)]
 pub enum AlgebraicValue {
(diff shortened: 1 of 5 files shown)


gefjon (changes requested):
Does this make BFLATN hashing consistent with `ProductValue::hash`? I suspect not until we merge #1063 . If that's the case, I'd like to hold off on merging this until after that PR, and then add a (prop)test to this PR that the hash functions are indeed consistent.

I'd also like to see a comment on `SumValue` about what constraints need to be upheld by its `Hash` impl, and the fact that `derive(Hash)` does indeed uphold those constraints.

Otherwise, code looks good. I'll mention that, for our case of comparing hashes only between homogeneously-typed values, this does not actually weaken the hash function at all - in fact, avoiding feeding known-equal data into the hasher should improve the hashes, as we're now packing the same number of "interesting" bits alongside fewer "boring" bits into a hash.

joshua-spacetime (approved):

Issue #1006, opened 2024-03-21
BFLATN: store sum tags before payloads, same as BSATN


gefjon:
This would improve the fast-path BFLATN -> BSATN conversion, as it would allow larger `memcpy` blocks.

kazimuth:
I'm concerned that even with this change memcpy still won't be possible for most sum types, because BSATN does not take alignment into account.

e.g.
```rust
enum Thing {
     Stuff(u32)
}
```

bsatn layout:
| `0x0` | `0x1`  | `0x2`  | `0x3`  | `0x4`  |
| ----- | ------ | ------ | ------ | ------ |
| tag   | stuff0 | stuff1 | stuff2 | stuff3 |

old bflatn layout:
| `0x0`  | `0x1`  | `0x2`  | `0x3`  | `0x4` |
| ------ | ------ | ------ | ------ | ----- |
| stuff0 | stuff1 | stuff2 | stuff3 | tag   |

new bflatn layout (taking into account that `u32` has a required alignment of 4)
| `0x0` | `0x1` | `0x2` | `0x3` | `0x4`  | `0x5`  | `0x6`  | `0x7` |
| ----- | ----- | ----- | ----- | ------ | ------ | ------ | ------ |
| tag   | pad   | pad   | pad   | stuff0 | stuff1 | stuff2 | stuff3 |

gefjon:
You are correct, and we don't want to introduce padding into BSATN. The reason swapping the order would be beneficial is that, for enums where the fast-path does apply, we could merge copying the tag and/or the payload with preceeding and/or succeeding data.

kazimuth:
ahhh right okay cool.

PR #1063, opened 2024-04-09
Swap the location of tags in the BFLATN encoding

kazimuth:
# Description of Changes

Partially implements #1006. I haven't ensured we're actually doing the relevant memcpy optimizations yet.

Is BFLATN public ABI yet? It will be once we have snapshots I suppose. 
So, a breaking change on the BFLATN ABI.

3

# Testing

I want to run this under miri but haven't done so yet.

### DIFF
--- a/crates/table/src/bflatn_to_bsatn_fast_path.rs
+++ b/crates/table/src/bflatn_to_bsatn_fast_path.rs
@@ -240,9 +240,6 @@ impl LayoutBuilder {
 
         // Now that we've reached this point, we know that `first_variant_layout`
         // applies to the values of all the variants.
-        // Do a bit of hackery to re-order the tag, since BFLATN stores `(payload, tag)`,
-        // but BSATN stores `(tag, payload)`,
-        // then splice the `first_variant_layout` into `self`.
 
         let tag_bflatn_offset = self.next_bflatn_offset();
         let payload_bflatn_offset = tag_bflatn_offset + sum.payload_offset;
@@ -250,30 +247,23 @@ impl LayoutBuilder {
         let tag_bsatn_offset = self.next_bsatn_offset();
         let payload_bsatn_offset = tag_bsatn_offset + 1;
 
-        self.fields.push(MemcpyField {
-            bflatn_offset: tag_bflatn_offset,
-            bsatn_offset: tag_bsatn_offset,
-            length: 1,
-        });
+        // Serialize the tag, consolidating into the previous memcpy if possible.
+        self.visit_primitive(&PrimitiveType::U8);
 
-        for payload_field in &first_variant_layout.fields[..] {
+        if sum.payload_offset > 1 {
+            // Add an empty marker field to keep track of padding.
             self.fields.push(MemcpyField {
-                bflatn_offset: payload_bflatn_offset + payload_field.bflatn_offset,
-                bsatn_offset: payload_bsatn_offset + payload_field.bsatn_offset,
-                length: payload_field.length,
+                bflatn_offset: payload_bflatn_offset,
+                bsatn_offset: payload_bsatn_offset,
+                length: 0,
             });
+        } else if sum.payload_offset == 1 {
+            // Nothing to do.
         }
 
-        // Finally, start a new field which skips over the tag.
-        // This field will almost certainly end up empty,
-        // as there will generally be padding following the tag in `sum`,
-        // but that's okay, because `Self::build` strips empty fields.
-        let next_bsatn_offset = self.next_bsatn_offset();
-        self.fields.push(MemcpyField {
-            bflatn_offset: tag_bflatn_offset + 1,
-            bsatn_offset: next_bsatn_offset,
-            length: 0,
-        });
+        // Lay out the variants.
+        // Since all variants have the same layout, we just use the first one.
+        self.visit_value(&first_variant.ty)?;
 
         Some(())
     }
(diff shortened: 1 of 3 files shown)


crates/table/src/layout.rs
```diff
@@ -424,13 +425,17 @@ impl From<SumType> for SumTypeLayout {
...
 
-        let tag_offset = max_child_size as u16;
-        let size = align_to(max_child_size + 1, max_child_align) as u16;
-        let align = max_child_align as u16;
+        // guarantees tag fits inside align
+        let align = u16::max(max_child_align as u16, 1);
+
+        // [tag | pad to align | payload]
+        // payload may be empty
+        let size = align + align_to(max_child_size, max_child_align) as u16;
```
gefjon, line 433:
```suggestion
        // Ensure the payload field is sufficiently aligned for all its members.
        // `max_child_size` and `max_child_align` will already be consistent
        // if the most-aligned variant is also the largest,
        // but this is not necessarily the case.
        // E.g. if variant A is a product of 31 `u8`s, and variant B is a single `u64`,
        // `max_child_size` will be 31 and `max_child_align` will be 8.
        // Note that `payload_size` may be 0.
        let payload_size = align_to(max_child_size, max_child_align);

        // [tag | pad to align | payload]
        let size = align + payload_size as u16;
```

I had to do some thinking to convince myself this `align_to` was necessary, so I'm suggesting we add that reasoning as a comment.

crates/table/src/bflatn_to_bsatn_fast_path.rs
```diff
@@ -244,8 +244,8 @@ impl LayoutBuilder {
         // but BSATN stores `(tag, payload)`,
         // then splice the `first_variant_layout` into `self`.
 
-        let payload_bflatn_offset = self.next_bflatn_offset();
-        let tag_bflatn_offset = payload_bflatn_offset + sum.tag_offset;
+        let tag_bflatn_offset = self.next_bflatn_offset();
+        let payload_bflatn_offset = tag_bflatn_offset + sum.payload_offset;
```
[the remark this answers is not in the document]
gefjon: I disagree - it should be easy to change the `visit_sum` code to do a `visit_primitive(&PrimitiveType::U8)` for the tag, and then to advance to the payload and add the payload fields using logic similar to `visit_product`. I think this will be simpler than adding an additional pass, and I'd prefer not to add a post-processing step to handle a single case as it seems likely to go forgotten if this code ever changes.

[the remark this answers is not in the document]
gefjon: Certainly as a first step we should simply replace the `self.fields.push(...)` on line 253 with a call to `visit_primitive(&PrimitiveType::U8)`.

crates/table/src/layout.rs
kazimuth: Yeah good point, I had to stare at it for a while as well. 👍

crates/table/src/bflatn_to_bsatn_fast_path.rs
```diff
@@ -341,9 +341,9 @@ mod test {
                     AlgebraicType::Bool,
                 ])]),
                 2,
-                // Sums get wonky layouts
-                // because BFLATN and BSATN store the tag and the payload in opposite orders.
-                &[(1, 0, 1), (0, 1, 1)][..],
+                // in bflatn, sums have padding after the tag to the max alignment of any variant payload
+                // in this case, 0 bytes of padding, because all payloads are aligned to 1
+                &[(0, 0, 1), (1, 1, 1)][..],
             ),
```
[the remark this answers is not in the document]
kazimuth: Added all of these.

kazimuth:
Okay, the optimization is implemented.

> > Is BFLATN public ABI yet? It will be once we have snapshots I suppose. So, a breaking change on the BFLATN ABI.
> 
> BFLATN never leaks (yet) out of the db to consumers either in modules, queries, or subscriptions, so it's not public and thus this change isn't breaking.

Do we not consider stuff written to the filesystem as ABI? If this was merged after snapshots it would result in a change of our on-disk format, so that should be ABI right? (Whether or not it actually gets merged after snapshots is a different question.)

crates/table/src/bflatn_to_bsatn_fast_path.rs
```diff
@@ -240,40 +240,30 @@ impl LayoutBuilder {
...
-        // but that's okay, because `Self::build` strips empty fields.
-        let next_bsatn_offset = self.next_bsatn_offset();
-        self.fields.push(MemcpyField {
-            bflatn_offset: tag_bflatn_offset + 1,
-            bsatn_offset: next_bsatn_offset,
-            length: 0,
-        });
+        // Lay out the variants.
+        // Since all variants have the same layout, we just use the first one.
+        self.visit_value(&first_variant.ty)?;
```
kazimuth, line 266: This is where the optimization is implemented.

gefjon:
> Do we not consider stuff written to the filesystem as ABI? If this was merged after snapshots it would result in a change of our on-disk format, so that should be ABI right? (Whether or not it actually gets merged after snapshots is a different question.)

Well, given that you're implementing this now, and no one has started working on snapshots yet, I think we're good.

crates/table/src/layout.rs
```diff
@@ -424,13 +425,25 @@ impl From<SumType> for SumTypeLayout {
...
+        let align = u16::max(max_child_align as u16, 1);
+
+        // Ensure the payload field is sufficiently aligned for all its members.
+        // `max_child_size` and `max_child_align` will already be consistent
+        // if the most-aligned variant is also the largest,
+        // but this is not necessarily the case.
+        // E.g. if variant A is a product of 31 `u8`s, and variant B is a single `u64`,
+        // `max_child_size` will be 31 and `max_child_align` will be 8.
+        // Note that `payload_size` may be 0.
+        let payload_size = align_to(max_child_size, max_child_align);
```
kazimuth, line 438: Wait, on further reflection, maybe this isn't necessary? Because in that case, it's fine for the payload to be size 31, right?

kazimuth:
Okay I think this is ready

gefjon (approved):

Issue #1046, opened 2024-04-07
CLI:


xNul:

MichaelNey:
I would like this as well. It seems `spacetime list` definitely seems like a great spot for this. With a change to the command to include also fetching the DNS of the database via the /database/reverse_dns/:address endpoint, I was able to get an output like this: 
![image](https://github.com/clockworklabs/SpacetimeDB/assets/25756024/81fe4120-34bb-453f-b911-dea2876ee8c1)

I think this would be a nice QoL change that I would appreciate.

cloutiertyler:
This is a great idea! CLI needs some work generally I think

MichaelNey:
@xNul I made a change I've built locally that I like to use to get db names easily if you are looking to have that for now until the StDB team makes the addition.

Bechma:
Hello, I would like to bump this issue:

<img width="805" height="169" alt="Image" src="https://github.com/user-attachments/assets/862e6c1d-a343-4958-b785-6b6d2fb6b71d" />

It would be nice to have the database name, I'm completely blind until I `describe` each of the database, and based on the tables, I got the idea of which database is.

bfops:
Yeah that's a fair bump. We'll try to get this addressed soon!

PR #1072, opened 2024-04-10
Make spacetime list <identity> CLI command also show db_name associated with database

MichaelNey:
# Description of Changes

#1046 Mentions a need to see the database name when listing databases for an identity. I agree with this being useful so here is an implementation supporting that. There may be better ways that the StDB team could do this, but here is what I use locally when developing if I want to easily get a database name.

Updated the `spacetime list <identity>` to also display a `db_name` column in the response. This `db_name` column is populated by making a request to the `/database/reverse_dns/<db_address>` endpoint for each db_address returned for the identity.

 N/A ?

I believe the complexity and risk of this change to be a 1/5.

# Testing

I tested this by making multiple databases, then running the `spacetime list <identity>` command. This should show a response in the CLI as a table with two columns: `db_name` and `db_address`. You can verify the db_address' all match the db_names via spacetime profile dashboard, as well as by creating new databases, and renaming databases. After that, the command should reflect new database names correctly, as well as updated database names.

**Test 1**

- Create a new identity if you do not already have one
- Create a new database (or even multiple)
- Use the `spacetime list <identity>` command
- Verify the db_name is present, and that it matches the correct db_address
- Verify in the SpacetimeDB Profile page that they match as well if applicable

**Test 2**

- Follow test 1 if you do not already have an identity that owns at least one database
- Use the `spacetime list <identity>` command
- Verify the db_name is present, and that it matches the correct db_address
- Verify in the SpacetimeDB Profile page that they match as well if applicable
- Rename the database
- Use the `spacetime list <identity>` command
- Verify the db_address matches the new db_name you set for it

### DIFF
--- a/crates/cli/src/subcommands/list.rs
+++ b/crates/cli/src/subcommands/list.rs
@@ -1,9 +1,14 @@
-use crate::Config;
+use crate::{config::IdentityConfig, Config};
 use anyhow::Context;
 use clap::{Arg, ArgMatches, Command};
+use futures::future::join_all;
 use reqwest::StatusCode;
 use serde::Deserialize;
-use spacetimedb_lib::Address;
+
+use spacetimedb_lib::{
+    name::{DomainName, ReverseDNSResponse},
+    Address,
+};
 use tabled::{
     settings::{object::Columns, Alignment, Modify, Style},
     Table, Tabled,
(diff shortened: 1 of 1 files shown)


jdetter:
> Also I agree with BSL and am interested in completing the CLA to enable myself to make contributions that may be used.

Sorry for the late response on this, can you message us through discord to get the process started for this? We can email you the CLA

MichaelNey:
Spoke to TL and signed the CLA.

PR #1073, opened 2024-04-10
Drop commitlog logging to `trace` to avoid spamming host logs

gefjon:
# Description of Changes

Does what the title says.

N/a.

1

# Testing

Completely untested.


kurtismullins (approved):
LGTM -- Thank you @gefjon!

PR #1078, opened 2024-04-11
CLI - Set default server to `testnet`

dylanh724:
# Description of Changes

`testnet` is quite stable now and makes onboarding (both the UX and docs) a significantly more-streamlined experience

- from "local"
- Suggested by @cloutiertyler 

image

(This is technically a breaking change? I'm not quite sure what label to apply)

2

⚠️ There may be some tooling that assumes local if they never explicitly set it to testnet. 

# Testing

- [ ] *Try it in a new user env*


dylanh724:
Specific poke to @jdetter - heard you're working on CLI stuff now!👋

crates/cli/src/config.rs
```diff
@@ -107,7 +107,7 @@ pub struct RawConfig {
 
 const DEFAULT_HOST: &str = "127.0.0.1:3000";
 const DEFAULT_PROTOCOL: &str = "http";
-const DEFAULT_HOST_NICKNAME: &str = "local";
+const DEFAULT_HOST_NICKNAME: &str = "testnet";
```
jdetter, line 110: We would need to update all 3 of these:
```
const DEFAULT_HOST: &str = "testnet.spacetimedb.com";
const DEFAULT_PROTOCOL: &str = "https";
const DEFAULT_HOST_NICKNAME: &str = "testnet";
```

jdetter (changes requested):

jdetter:
We originally had the CLI pointing to testnet by default, I don't remember the exact reason why we changed it back to local. Changing it back to testnet has some implications but we could possibly get this into the next release. I'll talk to Tyler about it today.

crates/cli/src/config.rs
dylanh724: image
Thanks, got it

cloutiertyler:
> We originally had the CLI pointing to testnet by default, I don't remember the exact reason why we changed it back to local. Changing it back to testnet has some implications but we could possibly get this into the next release. I'll talk to Tyler about it today.

The reason was that testnet was not stable and we were afraid that people would be trying to do the tutorial while Testnet was down. Now that it has been more stable, I think this change makes sense.

jdetter (approved):
LGTM

jdetter (changes requested):
I have to double check that the smoketests are working properly after this change

jdetter:
We'll have to rework this PR after https://github.com/clockworklabs/SpacetimeDB/pull/1151 lands

jdetter (approved):
LGTM 👍

PR #1079, opened 2024-04-11
commitlog: Direct I/O

kim:
Introduce page-aligned buffered reads + writes. This is required for direct I/O (O_DIRECT or equivalent), which is now the default.

Direct I/O essentially means to bypass the OS's page cache and operate directly on the device. It does not mean that data is automatically more durable once written, `fsync` is still required. Optionally, the patch allows to enable O_DSYNC, which blocks a write until the equivalent of `fdatasync` has occurred.

Both options likely have a performance impact, which needs to be evaluated. There also some performance optimizations of the presented implemention possible, e.g. re-using buffer allocations or scatter/gather ("vectored") I/O.

**Stacked on top of #985**

4

# Testing

- [x] Run test suite under Linux
- [ ]  Check that the code compiles under occult operating systems (macOS / Windows)
- [ ] Check that the code works under occult operating systems.


crates/commitlog/src/dio/writer.rs
```diff
@@ -0,0 +1,214 @@
...
+/// # Flushing behaviour
+///
+/// [`io::Write::write`] calls will only flush the buffer when it is full.
+/// [`io::Write::flush`] calls, however, will flush the buffer up to the next
+/// [`BLOCK_SIZE`] boundary, padding the data with zeroes if necessary.
+///
+/// This is done so that partial writes to the underlying storage can be
+/// detected on the application layer, and retried if appropriate. It is also
+/// necessary to preserve `fsync` semantics: a [`PagedWriter`] replaces the OS
+/// page cache, where the latter is flushed to the device when `fsync` is called.
```
kim, line 29: Challenge this.

It seemed confusing having to remember to flush first, even though `BufWriter` behaves like this -- except that a flush there only means to move the buffered data to kernel space.

kim:
I might try to come up with an abstraction which uses the std `BufReader`/`BufWriter` when alignment is not needed, as that could be slightly more efficient.

crates/commitlog/src/dio/mod.rs
```diff
@@ -0,0 +1,111 @@
+mod page;
```
kim, line 1: Yes, I shall add some prose here explaining what direct I/O is and what's the deal with the alignment requirements.

crates/commitlog/src/segment.rs
```diff
@@ -411,15 +430,17 @@ mod tests {
...
-        let mut writer = repo::create_segment_writer(&repo, Options::default(), 0).unwrap();
-        writer.append([0; 32]).unwrap();
-        writer.append([0; 32]).unwrap();
-        writer.commit().unwrap();
-        writer.append([1; 32]).unwrap();
-        writer.commit().unwrap();
-        writer.append([2; 32]).unwrap();
-        writer.append([2; 32]).unwrap();
-        writer.commit().unwrap();
+        {
```
kim, line 433: Should move those blocks back to the left -- shouldn't depend on drop, actually.

crates/commitlog/src/tests/partial.rs
```diff
@@ -19,7 +21,7 @@ use crate::{
 fn traversal() {
     enable_logging();
 
-    let mut log = open_log::<[u8; 32]>(ShortMem::new(800));
+    let mut log = open_log::<[u8; 32]>(ShortMem::new(5120));
```
kim, line 24: Idk why I changed the parameters, should change back.

crates/commitlog/src/buf.rs
```diff
@@ -0,0 +1,114 @@
...
+    ptr,
+};
+
+/// A byte buffer of non-zero size and proper alignment.
+#[derive(Debug)]
+pub struct Aligned {
+    data: ptr::NonNull<u8>,
+    layout: alloc::Layout,
+    size: usize,
+}
```
gefjon, line 14: Why is this parametric over the layout at runtime, rather than using const generics? Do we ever need to dynamically compute the layout?

crates/commitlog/src/buf.rs
```diff
@@ -0,0 +1,114 @@
...
+        let data = unsafe { ptr::NonNull::new_unchecked(ptr) };
+
+        Self { data, layout, size }
+    }
+
+    /// Borrow the buffer as a slice.
+    pub fn as_bytes(&self) -> &[u8] {
+        // SAFETY: The allocation has non-zero size `self.size` and is properly
+        // aligned. `self.size` cannot be modified outside `Self`.
+        unsafe { std::slice::from_raw_parts(self.data.as_ptr(), self.size) }
```
gefjon, line 50: This method may create a reference to uninitialized memory, causing UB. It seems like `Aligned` should be a `NonNull<MaybeUninit<u8>>`, and this method should return `&[MaybeUninit<u8>]`.

crates/commitlog/src/buf.rs
kim: Hadn’t thought of const generics. I need to check for zeroes then, no?

crates/commitlog/src/buf.rs
kim: Is it? I mean, I really don’t know what I’m doing, but I thought the size parameter of the layout would mean it gets initialized.

I think the memory should be allocated up front instead of lazily. What’s the right way to do that?

gefjon: The memory is allocated up-front, but not initialized. This is the difference between `alloc` and `alloc_zeroed`. From an abstract machine perspective, the memory returned by `alloc` initially holds the special value `uninit`/`poison`, which is not a valid `u8`, but is a valid `MaybeUninit<u8>`. Constructing an `&u8` or `&mut u8` to `uninit` is UB, so you have to either allocate an initialized buffer with `alloc_zeroed`, or use `&MaybeUninit<u8>` instead.

kim: I see, thanks.

I took this from glommio btw, I guess I should maybe let them know.

crates/commitlog/src/dio/writer.rs
kazimuth: I think this makes sense for our use case. What exactly is the alternative?

kazimuth (commented):
This looks good as always. Would honestly be nice as a little crate.

PR #1086, opened 2024-04-12
core: Don't persist empty transactions

kim:
Fix a minor bug where completely empty transactions would still be
written to the commitlog. The bug is minor because, once we start
logging inputs, all transactions will be non-empty.

The check is done in relational DB rather than the durability crate,
because in principle empty transactions are permissible, and may be used
in the future (e.g. to confirm a certain offset).

1

# Testing

- [x] Test suite passes
- [x] Log output is absent as expected when calling a reducer which doesn't
      modify rows (where previously it was present).

### DIFF
--- a/crates/core/src/db/relational_db.rs
+++ b/crates/core/src/db/relational_db.rs
@@ -327,9 +327,12 @@ impl RelationalDB {
                     truncates: Box::new([]),
                 }),
             };
-            log::trace!("append {txdata:?}");
-            // TODO: Should measure queuing time + actual write
-            durability.append_tx(txdata);
+
+            if !txdata.is_empty() {
+                log::trace!("append {txdata:?}");
+                // TODO: Should measure queuing time + actual write
+                durability.append_tx(txdata);
+            }
         }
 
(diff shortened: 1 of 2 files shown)


kazimuth (approved):
👍

crates/core/src/db/relational_db.rs
```diff
@@ -327,9 +327,12 @@ impl RelationalDB {
                     truncates: Box::new([]),
                 }),
             };
-            log::trace!("append {txdata:?}");
-            // TODO: Should measure queuing time + actual write
-            durability.append_tx(txdata);
+
```
cloutiertyler, line 330:
```suggestion
// Fix a minor bug where completely empty transactions would still be
// written to the commitlog. The bug is minor because, once we start
// logging inputs, all transactions will be non-empty.

// The check is done in relational DB rather than the durability crate,
// because in principle empty transactions are permissible, and may be used
// in the future (e.g. to confirm a certain offset).
```

cloutiertyler (approved):
Approving with one suggestion to add the PR comment to the actual code.

PR #786, opened 2024-02-02
Implement 'sql' parsing for 'Identity, Address' in hex format

mamcx:
# Description of Changes

Allow to execute 

```sql
select * from test where 
identity = 0x0000000000000000000000000000000000000000000000000000000000000000 AND 
identity = x'93dda09db9a56d8fa6c024d843e805D8262191db3b4bA84c5efcd1ad451fed4e' AND 
identity = x'91DDA09DB9A56D8FA6C024D843E805D8262191DB3B4BA84C5EFCD1AD451FED4E' AND 
address = 0x00000000000000000000000000000000 AND 
address = x'00000000000000000000000000000000'
```

for `Identity` and `Address` new types.

NOTE: Only full hex are supported. This fails:

```rust
 Address::to_abbreviated_hex()
```
This requires a breaking change in how the `hex` are parsed:

```rust
pub fn from_hex(hex: &str) -> Result<Self, anyhow::Error> {
        <[u8; 16]>::from_hex(hex)
            .context("Addresses must be 32 hex characters (16 bytes) in length.")
            .map(|arr| Self::from_arr(&arr))
}
```

# Expected complexity level and risk

1

PR #1087, opened 2024-04-12
Fix output of binary, Identity, Address for `SQL` output and the 'Display' of them to show a full hex value

mamcx:
This is a companion for [Implement 'SQL' parsing for 'Identity, Address' in hex format](https://github.com/clockworklabs/SpacetimeDB/pull/786), so now we have a correct round-trip of binary values in `SQL`.

Now when printing an `SQL` output that has  `Binary`, `Identity`, or `Address` fields it correctly displays them as a `hex` value.

```bash
🪐quickstart>select i, i from OneIdentity
 i                                                                  | i
--------------------------------------------------------------------+--------------------------------------------------------------------
 0x93dda09db9a56d8fa6c024d843e805d8262191db3b4ba84c5efcd1ad451fed4e | 0x93dda09db9a56d8fa6c024d843e805d8262191db3b4ba84c5efcd1ad451fed4e
 0x93dda09db9a56d8fa6c024d843e805d8262191db3b4ba84c5efcd1ad451fed2e | 0x93dda09db9a56d8fa6c024d843e805d8262191db3b4ba84c5efcd1ad451fed2e
```

# Description of Changes

Because the print of `SQL' is changed for these types is also a breaking change anyway (if only for the small chance somebody depends on the old way of displaying them).

1

# Testing

- [x] *output_identity_address* confirmed it print correctly for `SQL`
- [x] *Manual inspection* of `SQL` output using the `cli`.


crates/lib/src/address.rs
```diff
@@ -38,7 +38,7 @@ impl Default for Address {
 
 impl fmt::Display for Address {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.pad(&self.to_hex())
+        write!(f, "0x{}", self.to_hex())
```
mamcx, line 41: This makes this a breaking change. Should I remove this?

crates/lib/src/identity.rs
```diff
@@ -99,7 +99,7 @@ impl Identity {
 
 impl fmt::Display for Identity {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.pad(&self.to_hex())
+        write!(f, "0x{}", self.to_hex())
```
mamcx, line 102: This makes this a breaking change. Should I remove this?

crates/lib/src/address.rs
mamcx: Removed

crates/lib/src/identity.rs
mamcx: Removed

smoketests/tests/new_user_flow.py
```diff
@@ -44,8 +44,9 @@ def test_new_user_flow(self):
         self.assertEqual(self.logs(5).count("Hello, Tyler!"), 1)
 
         out = self.spacetime("sql", self.address, "SELECT * FROM Person")
+        # The spaces after the name are important
         self.assertMultiLineEqual(out, """\
```
mamcx, line 48: Note this change of output...

smoketests/tests/sql.py
```diff
@@ -86,9 +86,9 @@ def test_sql_format(self):
 
         self.call("test")
         sql_out = self.spacetime("sql", self.address, "SELECT * FROM BuiltIn")
-
+        # The spaces after the last column are important
         self.assertMultiLineEqual(sql_out, """\
```
mamcx, line 90: Note this change of output...

crates/core/src/sql/compiler.rs
```diff
@@ -394,6 +396,65 @@ mod tests {
         Ok(())
     }
 
+    // Verify the output of `sql` match the inputs for `Identity`, 'Address' & bynary data.
```
joshua-spacetime, line 399:
```suggestion
    // Verify the output of `sql` matches the inputs for `Identity`, 'Address' & binary data.
```

crates/sats/src/satn.rs
```diff
@@ -501,80 +510,171 @@ impl<'a, 'b> ser::SerializeNamedProduct for NamedFormatter<'a, 'b> {
...
+
+    fn serialize_element<T: Satn + ser::Serialize + ?Sized>(
+        &mut self,
+        name: Option<&str>,
+        elem: &T,
+    ) -> Result<(), Self::Error> {
+        // For binary data, output in `hex` format and skip the tagging of each value
+        self.is_special = ProductType::is_special_tag(name.unwrap_or_default());
+        self.f.entry(|mut f| {
+            if !self.is_special {
```
joshua-spacetime, line 539: Is it possible to have the special logic only for the binary case, and use the default formatter otherwise?

mamcx: Not because the `identity` and `Address` are product types and they cause the normal formatter to print the internal tag like`__identity_bytes`, that is the reason of all the roundabouts...

joshua-spacetime (approved):

PR #1095, opened 2024-04-15
Nuke `#[spacetimedb(reducer, repeat = _)]`

Centril:
# Description of Changes

Get rid of the `repeat` modifier now that its unused in BitCraft and because it is a trap.

Breaks the module API/ABI.

3, someone (not BitCraft) might depend on this?


joshua-spacetime (approved):
I'm assuming we decided `repeat` can't be fixed?

PR #1096, opened 2024-04-16
Re-instantiate replay progress reporting

kim:
# Description of Changes

Output was wrongly logged at `debug`, and percentage reporting was missing a way to obtain the max offset to expect.

Note that this may need some reconsideration once we apply history _suffixes_. Do not rely on the output as is.

Nada

1

# Testing

- [ ] Manual testing: spin up an existing database and observe the log output

### DIFF
--- a/crates/core/src/db/relational_db.rs
+++ b/crates/core/src/db/relational_db.rs
@@ -158,14 +158,41 @@ impl RelationalDB {
     where
         T: durability::History<TxData = Txdata>,
     {
-        log::debug!("[{}] DATABASE: applying transaction history...", self.address);
-        // TODO: Output progress
+        log::info!("[{}] DATABASE: applying transaction history...", self.address);
+
+        // TODO: Revisit once we actually replay history suffixes, ie. starting
+        // from an offset larger than the history's min offset.
+        // TODO: We may want to require that a `tokio::runtime::Handle` is
+        // always supplied when constructing a `RelationalDB`. This would allow
+        // to spawn a timer task here which just prints the progress periodically
+        // in case the history is finite but very long.
+        let max_tx_offset = history.max_tx_offset();
+        let mut last_logged_percentage = 0;
+        let progress = |tx_offset: u64| {
+            if let Some(max_tx_offset) = max_tx_offset {
+                let percentage = f64::floor((tx_offset as f64 / max_tx_offset as f64) * 100.0) as i32;
+                if percentage > last_logged_percentage && percentage % 10 == 0 {
+                    log::info!(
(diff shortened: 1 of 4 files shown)


crates/core/src/db/relational_db.rs
```diff
@@ -151,14 +151,37 @@ impl RelationalDB {
...
+                        "[{}] Loaded {}% ({}/{})",
+                        self.address,
+                        percentage,
+                        tx_offset,
+                        max_tx_offset
+                    );
+                    last_logged_percentage = percentage;
+                }
+            // Print _something_ even if we don't know what's still ahead.
+            } else if tx_offset % 10_000 == 0 {
```
jdetter, line 174: Q: In what case would this happen? Would this flood the log?

crates/core/src/db/relational_db.rs
```diff
@@ -151,14 +151,37 @@ impl RelationalDB {
...
+        log::info!("[{}] DATABASE: applying transaction history...", self.address);
+
+        // TODO: Revisit once we actually replay history suffixes, ie. starting
+        // from an offset larger than the history's min offset.
+        let max_tx_offset = history.max_tx_offset();
+        let mut last_logged_percentage = 0;
+        let progress = |tx_offset: u64| {
+            if let Some(max_tx_offset) = max_tx_offset {
+                let percentage = f64::floor((tx_offset as f64 / max_tx_offset as f64) * 100.0) as i32;
+                if percentage > last_logged_percentage && percentage % 10 == 0 {
```
jdetter, line 163: Our general UX issue with this sort of logging would be that if it takes a really long time to get to the first 10% we really don't know how long it will take to complete the reload until we see that first 10% log. I would suggest changing this so that it outputs a log every 15 seconds instead of doing a percentage based log. If its not an easy change then don't worry about it this is fine for now.

jdetter (approved):
This is tested + working. I had some small feedback but nothing that would prevent merging. Thanks Kim!

crates/core/src/db/relational_db.rs
kim: This would happen for future implementations of the `History` trait which, for some reason, can’t determine how many txs are yet to come. For example, `apply` could be continuously fed from a replication stream.

In this case, we’d print a log every 10k txs, which seems reasonable.

crates/core/src/db/relational_db.rs
kim: Hm, ya, not sure if we'd want to spawn a thread just for this, or ask for the system time a couple of million times. I'll take a note that we might want to stick a `tokio::runtime::Handle` into `RelationalDB` for these kinds of things.

PR #1099, opened 2024-04-16
`call_reducer_with_tx`: move work before/after tx lock aquire/release

Centril:
# Description of Changes

Most of these were noticeable in the 380 CCU flame graph.

1. Create `ctx` & clone `tx_slot` before taking `tx` lock.
2. Move an allocation before taking `tx` lock.
3. Move `energy_monitor.record_reducer(..)` to after releasing `tx` lock.
4. Move tracing & metrics work to after releasing `tx` lock.
5. Move actual dropping of `Arc<ModuleEvent>` (last referent) to outside of `blocking_broadcast_event`, reducing the time the `subscriptions.read()` lock and `tx` lock is held.
6. Move work done by `record_metrics` after tx/subscription locks have been released.
7. Tweak `merge_apply_inserts` and `merge_apply_deletes` to update metrics at most once per table.

None

3 - We've had deadlocks and threading issues in this code path before, so review the logic with care.

# Testing

A playtest would be good to run before merging, primarily for safety reasons.

### DIFF
--- a/crates/core/src/host/wasm_common/module_host_actor.rs
+++ b/crates/core/src/host/wasm_common/module_host_actor.rs
@@ -549,7 +549,7 @@ impl<T: WasmInstance> WasmModuleInstance<T> {
         );
         let mut tx_slot = self.instance.instance_env().tx.clone();
         let tx = tx.unwrap_or_else(|| stdb.begin_mut_tx(IsolationLevel::Serializable));
-        let _guard = WORKER_METRICS
+        let metrics_reducer_query_dur_on_drop = WORKER_METRICS
             .reducer_plus_query_duration
             .with_label_values(&address, op.name)
             .with_timer(tx.timer);
@@ -648,6 +648,7 @@ impl<T: WasmInstance> WasmModuleInstance<T> {
             .subscriptions
             .blocking_broadcast_event(client.as_deref(), &subscriptions, &event);
         drop(subscriptions);
+        drop(metrics_reducer_query_dur_on_drop);
 
         ReducerCallResult {
             outcome,


crates/core/src/db/datastore/locking_tx_datastore/datastore.rs
```diff
@@ -417,45 +417,69 @@ impl MutTxDatastore for Locking {
...
-        })
-        .or_insert_with(|| cpu_time);
-
-    drop(guard);
-    DB_METRICS
-        .rdb_txn_cpu_time_sec_max
-        .with_label_values(workload, db, reducer)
-        .set(max_cpu_time);
+/// Holds sufficient information to record transaction metrics on drop.
+pub struct RecordTxMetricsOnDrop<'a> {
```
Shubham8287, line 421: This is very similar to what I am trying to do with iter metrics i.e to batch and delegate to later. I will rebase on your branch to make it consistent across code.

crates/core/src/db/datastore/traits.rs
```diff
@@ -243,18 +243,22 @@ pub trait Tx {
...
 
     fn begin_mut_tx(&self, isolation_level: IsolationLevel) -> Self::MutTx;
-    fn commit_mut_tx(&self, ctx: &ExecutionContext, tx: Self::MutTx) -> Result<Option<TxData>>;
-    fn rollback_mut_tx(&self, ctx: &ExecutionContext, tx: Self::MutTx);
+    fn commit_mut_tx<'ctx>(
+        &self,
+        ctx: &'ctx ExecutionContext,
+        tx: Self::MutTx,
+    ) -> Result<(TxData, RecordTxMetricsOnDrop<'ctx>)>;
+    fn rollback_mut_tx<'ctx>(&self, ctx: &'ctx ExecutionContext, tx: Self::MutTx) -> RecordTxMetricsOnDrop<'ctx>;
```
kim, line 258: Without additional context about locking it is quite odd to see this return type, the only thing one can do with is to drop it.

Some prose would certainly appreciated.

crates/core/src/host/wasm_common/module_host_actor.rs
```diff
@@ -589,47 +579,71 @@ impl<T: WasmInstance> WasmModuleInstance<T> {
...
+                reducer_name,
+                ?timings.total_duration,
+            );
+        }
+        reducer_span.exit();
+
+        WORKER_METRICS
+            .reducer_compute_time
+            .with_label_values(&address, reducer_name)
+            .observe(timings.total_duration.as_secs_f64());
```
kim, line 646: Any specific reason this is not in the `defer` above? Looks like only data from `ExecuteResult` is used.

crates/core/src/db/datastore/locking_tx_datastore/datastore.rs
```diff
@@ -478,37 +502,40 @@ impl MutTx for Locking {
...
         let timer = tx.timer;
         // TODO(cloutiertyler): We should probably track the tx.commit() time separately.
         let res = tx.commit(ctx);
 
-        // Record metrics for the transaction at the very end right before we drop
-        // the MutTx and release the lock.
-        record_metrics(ctx, timer, lock_wait_time, true);
-        Ok(Some(res))
+        // Keep what we need to record metrics, for the transaction, after we have released the lock.
+        let metrics = record_metrics_on_drop(ctx, timer, lock_wait_time, true);
```
cloutiertyler, line 526: I believe this is unnecessary because `tx.commit` takes ownership of the tx.

cloutiertyler (dismissed):
See comment.

coolreader18 (approved):

Issue #333, opened 2023-09-27
Getting a new identity every time I spacetime publish/call 


jdetter:
![image](https://github.com/clockworklabs/SpacetimeDB/assets/4099508/af58d0c0-066d-41e3-a415-b30e76b2d45c)

jdetter:
I randomly get into this state from time to time, next time it happens I'll dump my config here

kim:
I can reproduce this by not having a `default_identity` for a server. Which I think I don't have because `server add` doesn't set it.

kim:
Sometimes it seems the fresh identity is then set as the default, but sometimes not.

Maybe we should just prompt to select an identity if it wasn't set as a default nor an argument?

gefjon:
I suspect this is fixed by #1109 , #1196 and #1341 . Closing, but if it reappears we can reopen.

PR #1109, opened 2024-04-18
cli: Suggest to make new identity the default for server

kim:
When a server key rotation is suspected, suggest to make the new identity the default.

This usually happens during development using an ephemeral instance. If one follows the instructions, the `default_identity` is not set for the existing server, which makes the CLI generate a fresh identity every time.

See also: #333

1

# Testing

- Spin up an ephemeral stdb
- Try to do something using the CLI
- Observe that it complains about a rotated key 
- Run `spacetime identity new --no-email --server <the server>` (**without** `--default`)
- Try more things using the CLI
- Observe that a fresh identity is created every time
- Run `spacetime identity new --no-email --server <the server> --default` (as suggested)
- Observe that this identity is used subsequently

### DIFF
--- a/crates/cli/src/util.rs
+++ b/crates/cli/src/util.rs
@@ -361,7 +361,7 @@ Has the server rotated its keys?
 Remove the outdated identity with:
 \tspacetime identity remove {identity}
 Generate a new identity with:
-\tspacetime identity new --no-email --server {server}"
+\tspacetime identity new --no-email --server {server} --default"
         )
     })
 }


crates/cli/src/util.rs
```diff
@@ -361,7 +361,7 @@ Has the server rotated its keys?
 Remove the outdated identity with:
 \tspacetime identity remove {identity}
 Generate a new identity with:
-\tspacetime identity new --no-email --server {server}"
+\tspacetime identity new --no-email --server {server} --default"
```
jdetter, line 364: We should fix the UX issue in general, but regardless this is a good change to the warning to the user.

jdetter (approved):
LGTM

PR #1110, opened 2024-04-18
slow query log: only use `Instant::now` when needed + refactor

Centril:
# Description of Changes

- Refactor the slow query logger to reuse `WorkloadType` and other refactoring.
- Try to regain some of the performance lost in `incr-select` (see below).

After https://github.com/clockworklabs/SpacetimeDB/commit/7d5eb1532db173c032afa6b46b013148aec132c5 compared to https://github.com/clockworklabs/SpacetimeDB/commit/7d5eb1532db173c032afa6b46b013148aec132c5~1:
```
Benchmarking incr-select: Collecting 100 samples in estimated 5.0
incr-select             time:   [226.69 ns 228.30 ns 230.24 ns]
                        change: [+24.989% +26.097% +27.045%] (p = 0.00 < 0.05)
                        Performance has regressed.
```
After this PR compared to https://github.com/clockworklabs/SpacetimeDB/commit/7d5eb1532db173c032afa6b46b013148aec132c5~1:

```
Benchmarking incr-select: Collecting 100 samples in estimated 5.0
incr-select             time:   [195.23 ns 195.55 ns 196.04 ns]
                        change: [+7.5233% +8.3927% +9.1255%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

Run on i7-7770K, 64GB RAM.
As you can see, this recovers some of the `+26%` lost in performance such that `+8.4%` is lost instead.
The primary reason for the performance loss was the unconditional `Instant::now` which results in a system call.
This call is now conditional on whether there's also a `threshold`.

None

1

# Testing

No semantic changes.

### DIFF
--- a/crates/core/src/util/slow.rs
+++ b/crates/core/src/util/slow.rs
@@ -53,16 +53,16 @@ pub struct SlowQueryLogger<'a> {
     /// The threshold, if any, over which execution duration would result in logging.
     threshold: &'a Option<Duration>,
     /// The context the query is being run in.
-    ctx: WorkloadType,
+    workload: WorkloadType,
 }
 
 impl<'a> SlowQueryLogger<'a> {
-    pub fn new(sql: &'a str, threshold: &'a Option<Duration>, ctx: WorkloadType) -> Self {
+    pub fn new(sql: &'a str, threshold: &'a Option<Duration>, workload: WorkloadType) -> Self {
         Self {
             sql,
             start: threshold.map(|_| Instant::now()),
             threshold,
-            ctx,
+            workload,
         }
     }
 
@@ -86,8 +86,8 @@ impl<'a> SlowQueryLogger<'a> {
         if let Some((start, threshold)) = self.start.zip(self.threshold.as_ref()) {
             let elapsed = start.elapsed();
             if &elapsed > threshold {
-                let context = self.ctx.as_ref();
-                tracing::warn!(?context, ?threshold, ?elapsed, ?self.sql, "SLOW QUERY");
+                let workload = self.workload.as_ref();
+                tracing::warn!(?workload, ?threshold, ?elapsed, ?self.sql, "SLOW QUERY");
                 return Some(elapsed);
             }
         };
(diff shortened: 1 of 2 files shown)


mamcx (approved):
LGTM

joshua-spacetime:
Thanks for catching this @Centril. I'll take a look shortly. You're correct that we shouldn't be making any system calls per `eval_incr`.

crates/core/src/util/slow.rs
```diff
@@ -50,64 +43,51 @@ impl SlowQueryConfig {
...
     sql: &'a str,
     /// The start time of the query execution.
-    start: Instant,
-    /// Which [Threshold] to use.
-    threshold: Threshold,
+    start: Option<Instant>,
+    /// The threshold, if any, over which execution duration would result in logging.
+    threshold: &'a Option<Duration>,
+    /// The context the query is being run in.
+    ctx: WorkloadType,
```
joshua-spacetime, line 56: Can you name this field `workload`?

crates/core/src/util/slow.rs
```diff
@@ -50,64 +43,51 @@ impl SlowQueryConfig {
...
-            Threshold::Queries(dur) => ("Queries", dur),
-        };
-        if let Some(dur) = dur {
-            let elapsed = self.start.elapsed();
-            if elapsed > dur {
-                tracing::warn!(kind = kind, threshold = ?dur, elapsed = ?elapsed, sql = self.sql, "SLOW QUERY");
+        if let Some((start, threshold)) = self.start.zip(self.threshold.as_ref()) {
+            let elapsed = start.elapsed();
+            if &elapsed > threshold {
+                let context = self.ctx.as_ref();
```
joshua-spacetime, line 89: Similarly here as well.

joshua-spacetime (commented):
Thanks @Centril. I'm not sure where that 8% regression is coming from. It doesn't appear we've added any significant work.

I did notice though that we only ever call `SlowQueryLogger::query`, but we should be calling `SlowQueryLogger::subscription` in `eval` and `SlowQueryLogger::incremental_updates` in `eval_incr`. Could you make those changes here? It should only be two lines.

PR #1111, opened 2024-04-18
Always send TransactionUpdate to the sender

drogus:
# Description of Changes

This commit changes transaction update broadcast to always send an update to the sender, even if the sender is not subscribed to any data that was committed. In such a case, the transaction update is sent with an empty database update.

None

2

# Testing

* I tested using a TypeScript SDK test app
* I've added a unit test

- [x] I tested using a TypeScript SDK test app
- [x] I've added a unit test
- [ ] It would be best if someone can independently test that a transaction update is sent even without subscribe


crates/core/src/subscription/module_subscription_manager.rs
```diff
@@ -241,7 +246,21 @@ impl SubscriptionManager {
                 );
             drop(span);
 
-            let _span = tracing::info_span!("eval_send").entered();
+            if !eval.contains_key(&event.caller_identity) {
```
cloutiertyler, line 249: Do we want to send it to all clients with the same identity? I think probably we only want to send it to the (identity, address) pair, right?

drogus: Good question. This will *only* send to the original caller, cause I pass the caller connection client back from the reducer call, so we won't send to all of the clients with the same identity, but there is likely a bug in the `if` statement itself. If there are two clients connected with the same identity and if only the client that is *not* making the call is subscribed, this will return false and the caller will not get the update.

drogus: In order to fix the bug in this PR I have to fix the existing bug, so I'll submit the bugfix

drogus: So the bug was already fixed yesterday in #1121, I updated this PR to reflect the changes

cloutiertyler: Quite a saga! I'm glad it uncovered other issues though 🙏

crates/core/src/subscription/module_subscription_manager.rs
```diff
@@ -245,7 +250,23 @@ impl SubscriptionManager {
                 );
             drop(span);
 
-            let _span = tracing::info_span!("eval_send").entered();
```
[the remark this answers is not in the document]
drogus: oops, a mistake, thanks, good catch!

drogus:
@Centril thanks for a review, I fixed the stuff you pointed out

PR #1112, opened 2024-04-18
Define `Hash + Eq` for BSATN

Centril:
# Description of Changes

- Defines `hash_bsatn` to hash an AV encoded as BSATN.
- Defines `eq_bsatn` to compare AV encoded BSATN against an AV.

None.

2, we need to make sure the algorithms are consistent with AVs.

# Testing

- proptests are provided


kazimuth (approved):
LGTM, everything looks clean.

I'm curious, what do the `reborrow`s on the deserializer do? This may have been discussed before.

PR #1114, opened 2024-04-18
Disallow calling random lifecycle reducers

coolreader18:

1

### DIFF
--- a/crates/core/src/host/wasm_common/module_host_actor.rs
+++ b/crates/core/src/host/wasm_common/module_host_actor.rs
@@ -5,7 +5,7 @@ use std::time::Duration;
 
 use spacetimedb_lib::buffer::DecodeError;
 use spacetimedb_lib::identity::AuthCtx;
-use spacetimedb_lib::{bsatn, Address, ModuleDef, TableDesc};
+use spacetimedb_lib::{bsatn, Address, ModuleDef, ModuleValidationError, TableDesc};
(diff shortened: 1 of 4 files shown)


crates/core/src/host/wasm_common/module_host_actor.rs
```diff
@@ -173,7 +176,10 @@ impl<T: WasmModule> WasmModuleHostActor<T> {
             tables
                 .into_iter()
                 .map(|x| (x.schema.table_name.clone(), EntityDef::Table(x))),
-            reducers.iter().map(|x| (x.name.clone(), EntityDef::Reducer(x.clone()))),
+            reducers
+                .iter()
+                .filter(|r| !(r.name.starts_with("__") && r.name.ends_with("__")))
```
kim, line 181: Is this not the same map that is used to resolve a reducer in `ModuleHost::call_reducer_inner`?
This would mean that `ModuleHost::call_identity_connected_disconnected` never actually calls a corresponding reducer, even if the module defines them.

jdetter:
Suggestion: it would be really easy to add a smoketest for this

crates/core/src/host/wasm_common/module_host_actor.rs
kim: Oh I see, there's another binding

kim (approved):

PR #1123, opened 2024-04-19
Default commitlog output to info instead of debug

jdetter:
# Description of Changes

Right now the default log level for the commitlog is debug which is pretty noisy when loading a database from disk:

image

After defaulting it to info its much more reasonable:

image

corresponding GitHub label.

No API/ABI break

0 - this is just changing log levels

# Testing

- [x] Try loading a database before this change, the log is pretty spammy, try with this change and its much better. See screenshots.


kim (approved):
👍

PR #1125, opened 2024-04-19
Add a test for #1121

drogus:
# Description of Changes

While working on #1111 I realised that we have a bug with subscriptions not being unique when multiple clients with the same identity are connected. I fixed the bug and only then realised it was already fixed yesterday in #1121. When working on my changes I created a test for the issue, so I guess it doesn't hurt to at least submit it.

1


crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -190,7 +193,6 @@ mod tests {
     use tokio::sync::mpsc;
 
     #[test]
-    /// Asserts that a subscription holds a tx handle for the entire length of its evaluation.
```
joshua-spacetime, line 193: Did you mean to remove this comment?

crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -250,4 +252,93 @@ mod tests {
...
+
+        let id = ClientActorId::for_test();
+        let sender = Arc::new(ClientConnectionSender::dummy(id, Protocol::Binary));
+        let module_subscriptions = ModuleSubscriptions::new(db.clone(), id.identity);
+
+        let client_id0 = ClientActorId::for_test();
+        let (client0, mut rx0) = ClientConnectionSender::dummy_with_channel(client_id0, Protocol::Binary);
+
+        let client_id1 = ClientActorId::for_test();
+        let (client1, mut rx1) = ClientConnectionSender::dummy_with_channel(client_id1, Protocol::Binary);
```
joshua-spacetime, line 274: Does this test what you want? Won't the Identities be different here?

drogus: you're right, no idea what I was thinking 😅 I'll push a fix

crates/core/src/subscription/module_subscription_actor.rs
drogus: mistake, fixed now

drogus:
@joshua-spacetime I update the PR, so that the test is actually testing what I wanted to test

joshua-spacetime (approved):

PR #1130, opened 2024-04-22
Use -s to specify server for spacetime identity list -s <server>

jdetter:
# Description of Changes

Previous command was:
```
spacetime identity list <server>
```

New command is:
```
spacetime identity list -s <server>
```

The new command is more consistent with all of the other `spacetime` CLI commands which take a server argument.

corresponding GitHub label.

Technically yes, just a CLI change.

0

# Testing

- [ ] Do `spacetime identity list -s <server>` or leave `-s <server>` unset to list identities for the default server.


bfops (approved):
:heavy_check_mark: The hero we need

PR #1132, opened 2024-04-22
Disconnect dangling clients

### LINKED ISSUE #1178: On server start after a crash, call on_disconnect for all clients for whom on_connect was called in a module but on_disconnect was not called
We could *maybe* get away with not having this for BitCraft Alpha, but it means that we'll have issues if the server crashes during the playtest.

kim:
When replaying the commitlog, keep track of unpaired connect/disconnect calls and call disconnect when instantiating the module.

2

# Testing

- [x] Tested with `quickstart-chat` like so:

  - edit `quickstart-chat` to log connect / disconnect calls
  - publish `quickstart-chat` module
  - start client
  - shutdown server
  - start server
  - start client (to force module to be loaded)
  - stop client
  - examine module logs to convince thyself that connect is always paired with disconnect

### DIFF
--- a/crates/core/src/db/relational_db.rs
+++ b/crates/core/src/db/relational_db.rs
@@ -20,6 +20,7 @@ use fs2::FileExt;
 use parking_lot::RwLock;
 use spacetimedb_commitlog as commitlog;
 use spacetimedb_durability::{self as durability, Durability};
+use spacetimedb_lib::Identity;
 use spacetimedb_primitives::*;
 use spacetimedb_sats::db::auth::{StAccess, StTableType};
 use spacetimedb_sats::db::def::{ColumnDef, IndexDef, SequenceDef, TableDef, TableSchema};
@@ -27,6 +28,7 @@ use spacetimedb_sats::{AlgebraicType, AlgebraicValue, ProductType, ProductValue}
 use spacetimedb_table::indexes::RowPointer;
 use spacetimedb_vm::errors::ErrorVm;
 use std::borrow::Cow;
+use std::collections::HashSet;
 use std::fs::{create_dir_all, File};
 use std::io;
 use std::ops::RangeBounds;
@@ -49,6 +51,10 @@ type DiskSizeFn = Arc<dyn Fn() -> io::Result<u64> + Send + Sync>;
 
 pub type Txdata = commitlog::payload::Txdata<ProductValue>;
 
+/// Clients for which a connect reducer call was found in the [`History`], but
+/// no corresponding disconnect.
+pub type ConnectedClients = HashSet<(Identity, Address)>;
+
 #[derive(Clone)]
 pub struct RelationalDB {
     // TODO(cloutiertyler): This should not be public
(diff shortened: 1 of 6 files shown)


crates/core/src/db/relational_db.rs
```diff
@@ -61,9 +67,12 @@ pub struct RelationalDB {
...
     disk_size_fn: Option<DiskSizeFn>,
 
-    // Release file lock last when dropping.
-    _lock: Arc<File>,
     config: Arc<RwLock<DatabaseConfig>>,
+
+    // DO NOT ADD FIELDS AFTER THIS.
+    // By default, fields are dropped in declaration order.
+    // We want to release the file lock last.
+    _lock: Arc<File>,
```
gefjon, line 75: This is pretty nasty, tbh. I know it's not your doing, and I don't consider this a blocker, but you might consider writing a manual `impl Drop` rather than relying on the automatic one in order to be more explicit about this constraint.

crates/core/src/db/relational_db.rs
```diff
@@ -84,7 +93,11 @@ impl RelationalDB {
     ///
     /// The [`tokio::runtime::Handle`] is used to spawn background tasks which
     /// take care of flushing and syncing the log.
-    pub fn local(root: impl AsRef<Path>, rt: tokio::runtime::Handle, address: Address) -> Result<Self, DBError> {
+    pub fn local(
+        root: impl AsRef<Path>,
+        rt: tokio::runtime::Handle,
+        address: Address,
+    ) -> Result<(Self, ConnectedClients), DBError> {
```
gefjon, line 100: Can you add to the docs describing the return values? Github won't let me drop a suggestion on an unchanged line, but I'd add a paragraph reading:

```rs
/// Returns the set of clients who were connected as of the most recent transaction
/// as a [`ConnectedClients`]
/// so that the newly-restored module can call disconnect for each of them.
```

crates/core/src/db/relational_db.rs
```diff
@@ -154,7 +168,7 @@ impl RelationalDB {
     /// of the database in case of an incomplete replay.
     ///
     /// This restriction may be lifted in the future to allow for "live" followers.
-    pub fn apply<T>(self, history: T) -> Result<Self, DBError>
+    pub fn apply<T>(self, history: T) -> Result<(Self, ConnectedClients), DBError>
```
gefjon, line 171: Similarly, I'd add to the docs:

```rs
/// Watches for `__identity_connected__` / `__identity_disconnected__` calls in the log
/// in order to construct a [`ConnectedClients`],
/// the set of clients who were connected as of the most recent TX,
/// so that the newly-constructed module can call disconnect for each of them.
```

gefjon (commented):
Looks good. Minor suggestions for doc comments.

I hate to do this to you after you implemented this functionality, but I think we should seriously consider an `st_connected_clients` system table rather than this strategy because of the interaction with snapshotting.

bfops:
Fixes https://github.com/clockworklabs/SpacetimeDB/issues/1178

crates/core/src/db/relational_db.rs
kim: Actually, dropping some field _first_ in an `impl Drop` I know how to do, but not how to drop it _last_. Would it not require to wrap all the fields in `ManuallyDrop`?

gefjon: Oh, right, I forgot that even when you manually `impl Drop` you still get the auto-generated drop method as an epilogue. Disregard.

gefjon:
> > we should seriously consider an st_connected_clients system table
> 
> I'm not opposed to this, but it will be a breaking change. If that's ok, or we're willing to delay this change until it is ok, we can go the other route np.

I think certainly we're fine with breaking changes at this stage. I don't know to what extent we're willing to delay this PR; that's a question for @cloutiertyler .

cloutiertyler (approved):
If tested, then this LGTM

Issue #1116, opened 2024-04-18
`Column`: do not store `col_id` -- its already inferrable by `Header.fields`


Centril:
Remove `Column.col_id` as it can be inferred from the index in `Header.fields`.

cc https://github.com/clockworklabs/SpacetimeDB/issues/1113
cc @joshua-spacetime

PR #1136, opened 2024-04-22
Remove `Column.col_id` field

Centril:
# Description of Changes

It is always the same as the index in `Header.fields`.
This get `Column` down to 64 bytes (still large!) which fits in a cache line.

Fixes https://github.com/clockworklabs/SpacetimeDB/issues/1116.

None

1


mamcx (approved):
LGTM

PR #1139, opened 2024-04-23
Prevent importing an identity with a name that already exists in the config

jdetter:
# Description of Changes

Reported in the discord by `halu`. If you have an identity in your config with some name, like `admin`, you can still import another identity with this same name. This PR changes the behavior so that an error is printed telling the user to select another name.

corresponding GitHub label.

No

0

# Testing

- [x] Create a new identity with the name `admin`. Then create another identity with no name. Export this new identity. Delete the identity. Now import this identity with the name `admin`, you will get an error.


jdetter:
Closing in favor of https://github.com/clockworklabs/SpacetimeDB/pull/1140

PR #1140, opened 2024-04-23
Make identity naming more consistent + fix bugs

jdetter:
# Description of Changes

- There are several places in the CLI where you can set the name of an identity:
  - `spacetime identity new --name <name> --no-email`
  - `spacetime identity set-name <identity> <name>`
  - `spacetime identity import <identity> <token> --name <name>`
 
In these places we had slightly different logic for validating that an identity name looks okay. I've made the logic in these places more consistent. This will prevent people from accidentally adding identities with the same name or with an invalid name (identity names cannot look like identities).

corresponding GitHub label.

Not breaking

0

# Testing

- [x] Create a new identity with the name admin. Then create another identity with no name. Export this new identity. Delete the identity. Now import this identity with the name admin, you will get an error.

### DIFF
--- a/crates/cli/src/config.rs
+++ b/crates/cli/src/config.rs
@@ -978,22 +978,27 @@ Import an existing identity with:
         })
     }
 
-    pub fn name_exists(&self, name: &str) -> bool {
-        self.get_identity_config_by_name(name).is_some()
-    }
+    pub fn can_set_name(&self, new_nickname: Option<&str>) -> Result<(), anyhow::Error> {
+        if let Some(new_nickname) = new_nickname {
+            if self.name_exists(new_nickname) {
+                return Err(anyhow::anyhow!("An identity with that name already exists."));
+            }
 
-    pub fn identity_exists(&self, identity: &Identity) -> bool {
-        self.get_identity_config_by_identity(identity).is_some()
-    }
+            if is_hex_identity(new_nickname) {
+                return Err(anyhow::anyhow!("An identity name cannot be an identity."));
+            }
+        }
 
-    pub fn can_set_name(&self, new_nickname: &str) -> Result<(), anyhow::Error> {
-        if self.name_exists(new_nickname) {
-            return Err(anyhow::anyhow!("An identity with that name already exists."));
-        }
-        if is_hex_identity(new_nickname) {
-            return Err(anyhow::anyhow!("An identity name cannot be an identity."));
-        }
         Ok(())
+    }
+
+    pub fn name_exists(&self, nickname: &str) -> bool {
+        for name in self.identity_configs().iter().map(|c| &c.nickname) {
+            if name.as_ref() == Some(&nickname.to_string()) {
+                return true;
+            }
+        }
+        false
     }
 
     pub fn get_identity_config_by_name(&self, name: &str) -> Option<&IdentityConfig> {
--- a/crates/cli/src/config.rs
+++ b/crates/cli/src/config.rs
@@ -978,17 +978,13 @@ Import an existing identity with:
         })
     }
 
-    pub fn can_set_name(&self, new_nickname: Option<&str>) -> Result<(), anyhow::Error> {
-        if let Some(new_nickname) = new_nickname {
-            if self.name_exists(new_nickname) {
-                return Err(anyhow::anyhow!("An identity with that name already exists."));
-            }
-
-            if is_hex_identity(new_nickname) {
-                return Err(anyhow::anyhow!("An identity name cannot be an identity."));
-            }
+    pub fn can_set_name(&self, new_nickname: &str) -> Result<(), anyhow::Error> {
+        if self.name_exists(new_nickname) {
+            return Err(anyhow::anyhow!("An identity with that name already exists."));
         }
-
+        if is_hex_identity(new_nickname) {
+            return Err(anyhow::anyhow!("An identity name cannot be an identity."));
+        }
         Ok(())
     }
 
--- a/crates/cli/src/config.rs
+++ b/crates/cli/src/config.rs
@@ -988,13 +988,12 @@ Import an existing identity with:
         Ok(())
     }
 
-    pub fn name_exists(&self, nickname: &str) -> bool {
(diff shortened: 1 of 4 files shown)


crates/cli/src/subcommands/identity.rs
```diff
@@ -585,6 +577,7 @@ async fn exec_token(config: Config, args: &ArgMatches) -> Result<(), anyhow::Err
 async fn exec_set_name(mut config: Config, args: &ArgMatches) -> Result<(), anyhow::Error> {
     let new_name = args.get_one::<String>("name").unwrap();
     let identity_or_name = args.get_one::<String>("identity").unwrap();
+    config.can_set_name(Some(identity_or_name.as_str()))?;
```
bfops, line 580: This is a regression. It essentially renders unnamed identities unusable.
```
   $ spacetime identity new -s local --no-email | tee output
    IDENTITY     631d3896d10a5fa9db73a95a5a2e20981ab71bc32cb908859e449b5eeb886e9a
    NAME         
    EMAIL        
 
   $ cat output | grep 'IDENTITY' | awk '{print $2;}' | tee identity
   631d3896d10a5fa9db73a95a5a2e20981ab71bc32cb908859e449b5eeb886e9a
 
   $ spacetime identity set-name $(cat identity) new-identity-name
-  Updated identity:
-   IDENTITY  631d3896d10a5fa9db73a95a5a2e20981ab71bc32cb908859e449b5eeb886e9a
-   NAME      new-identity-name
+  Error: An identity name cannot be an identity.
+  [1]
```

bfops (changes requested):

crates/cli/src/subcommands/identity.rs
jdetter: Thank you for catching this Zeke that would have been a disaster 😦

crates/cli/src/subcommands/identity.rs
```diff
@@ -454,6 +445,11 @@ async fn exec_import(mut config: Config, args: &ArgMatches) -> Result<(), anyhow
 
     //optional
     let nickname = args.get_one::<String>("name").cloned();
+    config.can_set_name(nickname.as_deref())?;
+
+    if config.identity_configs().iter().any(|ic| ic.identity == identity) {
+        return Err(anyhow::anyhow!("Identity already exists in config"));
+    };
```
jdetter, line 452: I've also added a check here to prevent adding the same identity multiple times

crates/cli/src/util.rs
```diff
@@ -149,6 +149,10 @@ pub async fn select_identity_config(
     server: Option<&str>,
 ) -> Result<IdentityConfig, anyhow::Error> {
     if let Some(identity_or_name) = identity_or_name {
+        if identity_or_name.is_empty() {
+            return Err(anyhow::anyhow!("Identity value cannot be empty."));
+        }
```
bfops, line 154: why did we need this check? shouldn't `get_identity_config("")` fail anyway?

crates/cli/src/config.rs
```diff
@@ -889,6 +889,20 @@ Import an existing identity with:
...
+                return Err(anyhow::anyhow!("An identity with that name already exists."));
+            }
+
+            if is_hex_identity(new_nickname) {
+                return Err(anyhow::anyhow!("An identity name cannot be an identity."));
+            }
+        }
+
+        Ok(())
+    }
```
bfops, line 904: how would you feel about making this like:
```rs
pub fn can_set_name(&self, new_nickname: &str) -> Result<(), anyhow::Error> {
```
(it seems a bit weird to call `can_set_name(None)`?)

crates/cli/src/subcommands/publish.rs
```diff
@@ -111,6 +112,8 @@ pub async fn exec(mut config: Config, args: &ArgMatches) -> Result<(), anyhow::E
     let build_debug = args.get_flag("debug");
     let wasm_file = args.get_one::<PathBuf>("wasm_file");
     let database_host = config.get_host_url(server)?;
+    // Validate the identity first to provide an early error message
+    util::select_identity_config(&mut config, identity, server).await?;
```
bfops, line 116: could we just move the later code to here, rather than effectively duplicating this check? like could we put this up here:
```
    let (auth_header, identity) = get_auth_header(&mut config, anon_identity, identity, server)
        .await?
        .unzip();
```

crates/cli/src/util.rs
```diff
@@ -152,7 +152,7 @@ pub async fn select_identity_config(
         config
             .get_identity_config(identity_or_name)
             .cloned()
-            .ok_or_else(|| anyhow::anyhow!("No such identity credentials for identity: {}", identity_or_name))
+            .ok_or_else(|| anyhow::anyhow!("No identity credentials for identity \"{}\"", identity_or_name))
```
bfops, line 155: (the extra `""` make it clearer when it's just the empty string)

crates/cli/src/subcommands/publish.rs
```diff
@@ -112,6 +112,14 @@ pub async fn exec(mut config: Config, args: &ArgMatches) -> Result<(), anyhow::E
     let wasm_file = args.get_one::<PathBuf>("wasm_file");
     let database_host = config.get_host_url(server)?;
 
+    // If the user didn't specify an identity and we didn't specify an anonymous identity, then
+    // we want to use the default identity
+    // TODO(jdetter): We should maybe have some sort of user prompt here for them to be able to
+    //  easily create a new identity with an email
+    let (auth_header, identity) = get_auth_header(&mut config, anon_identity, identity, server)
+        .await?
+        .unzip();
```
bfops, line 121: moving this earlier means that we check `identity` validity before we go through a (potentially very long) build.

crates/cli/src/subcommands/identity.rs
bfops: (factored out `config.identity_exists` so it's somewhat self-describing)

bfops (approved):
LGTM once the base PR merges

PR #1141, opened 2024-04-23
host: Unregister panicked modules from the host controller

kim:
Add a Fn object to `ModuleHost` which is invoked whenever a reducer call on this host panics. When created from the `HostController`, the function simply unregisters the module from the controller.

This should release both the module and database resources, such that the database restarts on subsequent access. This is the intended behavior to land #985.

Note that this patch does **not** address the disjoint lifecycles of modules and databases, thus a panic when executing an SQL command is likely to leave the database in a permanently inaccessible state.

2

# Testing

- [x] Tested manually by randomly panicking in the `durability` background tasks, while accessing the database using the CLI. Note that, due to lazy database instantiation, the first command after a panic returns an error but the next one doesn't.
- [ ] Write me a test harness which would allow to run above in an automated fashion ;)

### DIFF
--- a/crates/core/src/host/host_controller.rs
+++ b/crates/core/src/host/host_controller.rs
@@ -19,7 +19,7 @@ use std::sync::Arc;
 use std::time::{Duration, Instant};
 
 pub struct HostController {
-    modules: Mutex<IntMap<u64, ModuleHost>>,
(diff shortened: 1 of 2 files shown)


kim:
Does not actually work as written.

PR #1143, opened 2024-04-23
Simplify C# source generator csproj

RReverser:
# Description of Changes

Apparently custom scripts are no longer necessary, so removing to make maintenance simpler (noticed this while working on yet another source generator).

corresponding GitHub label.

1

# Testing

- [x] Built the NuGet package, referenced in a new project, checked that it still builds as expected.


RReverser:
Testing instructions (this one is a bit tricky to test because it only affects built NuGet package and can't be tested with folder references):

 - Go into `SpacetimeDB\crates\bindings-csharp\Codegen` in your terminal.
 - Run `dotnet pack` to create a NuGet package, this will be built in `bin\Release` subfolder.
 - Go into `bindings-csharp\Runtime` and run the same `dotnet pack`.
 - Create a new C# module somewhere on your disk with `spacetime init --lang csharp <folder-name>`.
 - Open the generated `StdbModule.csproj`.
 - Add something like this (with your local paths) to `<PropertyGroup>`:
    ```xml
    <RestoreSources>D:\SpacetimeDB\crates\bindings-csharp\Codegen\bin\Release;D:\SpacetimeDB\crates\bindings-csharp\Runtime\bin\Release</RestoreSources>
    ```
 - Build the C# module with `dotnet build`.

It should build without errors - if it does, the NuGet packages are still packed correctly after this PR.

kazimuth:
I think it would make sense to add that as a smoketest, I can do that on top of this branch if needed.

RReverser:
> I think it would make sense to add that as a smoketest, I can do that on top of this branch if needed.

Sounds great if you could. Existing tests already cover any code changes, but yeah, we don't have anything in place for Nuget config changes specifically.

kazimuth:
@RReverser done. I also made the smoketests run on windows, at least that one -- although you have to run them through Powershell, not Git Bash.

kazimuth (approved):
I'm good with this pending somebody else looking at the test I wrote.

smoketests/tests/csharp_module.py
```diff
@@ -0,0 +1,50 @@
...
+                codegen_bin = codegen / "bin" / "Release"
+                runtime_bin = runtime / "bin" / "Release"
+
+                csproj = Path(tmpdir) / "StdbModule.csproj"
+                with open(csproj, "r") as f:
+                    contents = f.read()
+
+                contents = contents.replace(
+                    "</PropertyGroup>",
+                    f"""<RestoreAdditionalProjectSources>{codegen_bin.absolute()};{runtime_bin.absolute()}</RestoreAdditionalProjectSources>
```
RReverser, line 38: As mentioned in comment, I'd opt for `RestoreSources` rather than `RestoreAdditionalProjectSources` here to ensure that we're getting and testing the local package instead of silently getting one from nuget.org.

smoketests/tests/csharp_module.py
```diff
@@ -0,0 +1,50 @@
...
+</PropertyGroup>""",
+                )
+                with open(csproj, "w") as f:
+                    f.write(contents)
+
+                run_cmd("dotnet", "build", cwd=tmpdir, capture_stderr=True)
+
+        except subprocess.CalledProcessError as e:
+            print(e)
```
RReverser, line 47: Why is `print` necessary in addition to `raise` - wouldn't Python print errors by default?

RReverser:
@kazimuth Added couple of nits. Also - does this new test run automatically on CI or is it for manual run only for now?

smoketests/tests/csharp_module.py
kazimuth: Unfortunately we have additional dependencies that do not resolve when we do that.

smoketests/tests/csharp_module.py
kazimuth: Ah, that was debugging, I'll fix it.

smoketests/tests/csharp_module.py
kazimuth: Oh, I know, I'll patch the version!

RReverser: You should be able to add nuget URL explicitly as the last element.

kazimuth:
It runs automatically if CI has dotnet sdk 8.0 installed -- I am not sure if it does or not.

The issue you raised is patched :)

smoketests/tests/csharp_module.py
kazimuth: Wait, that is a far better solution than what I just did. I will do that instead.

kazimuth:
Okay done

smoketests/tests/csharp_module.py
kazimuth: The python unittest library prints raised errors in a different place from test stdout, so it's a little harder to read if you do that. It would work though, I could changes it.

RReverser: Meh, it's probably fine.

RReverser:
Wanted to approve the PR, but of course I can't because I created it 😅 Between the two of us, I think it should be good now, so added to the merge queue.

RReverser:
Thanks again for the test!

PR #1146, opened 2024-04-23
Fix codegen tests

RReverser:
# Description of Changes

While working on the new C# codegen, I accidentally noticed that those tests were passing even when they clearly should've been failing due to changed output.

After running with `--nocapture`, I found out it's because the tests are silently skipped and reported as successful when `rust_wasm_test.wasm` isn't built.

This further led to finding that `rust_wasm_test.wasm` is never built - the relevant module results in `rust_wasm_test_module.wasm` instead - so these tests have been incorrectly passing for ages.

This PR changes them to actually build the module as part of testing and updates the snapshots to latest master.

corresponding GitHub label.

1

# Testing


jdetter (approved):
Tested by pushing a codegen change to this PR, which caused the workflow to fail

image

PR #1149, opened 2024-04-24
NFC: format C# module SDK code

RReverser:
# Description of Changes

While looking through the large diffs while splitting out small PRs out of my refactor branch, I noticed that quite a lot of noise is from me working on a formatted code and using primary constructors while the one in master is not.

As such, I'm splitting out just those automated / non-functional changes into a separate PR to make subsequent functional diffs easier to read.

corresponding GitHub label.

1

# Testing

- [x] Checked that it still builds.
- [ ] *Write a test you want a reviewer to do here, so they can check it off when they're satisfied.* All the changes here are automated via CSharpier + IDE, so checking that it builds is enough.

### DIFF
--- a/crates/bindings-csharp/Codegen/Type.cs
+++ b/crates/bindings-csharp/Codegen/Type.cs
@@ -64,8 +64,7 @@ public class Type : IIncrementalGenerator
     )
     {
         context
-            .SyntaxProvider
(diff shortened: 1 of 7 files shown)


kazimuth (approved):
👍

PR #1151, opened 2024-04-24
CLI - Small refactor to default server configs

bfops:
# Description of Changes
We previously had our default server configs partially defined in a few places, and it led to surprising results. This refactor centralizes the default server configs.

No change to behavior.

1

# Testing
- [x] Existing CI coverage


crates/cli/src/config.rs
```diff
@@ -105,10 +105,6 @@ pub struct RawConfig {
     server_configs: Vec<ServerConfig>,
```
bfops, line 105: [Review with recommended settings](https://github.com/clockworklabs/SpacetimeDB/pull/1151/files?diff=split&w=1)

crates/cli/src/config.rs
```diff
@@ -105,10 +105,6 @@ pub struct RawConfig {
...
 }
 
-const DEFAULT_HOST: &str = "127.0.0.1:3000";
-const DEFAULT_PROTOCOL: &str = "http";
-const DEFAULT_HOST_NICKNAME: &str = "local";
-
```
bfops, line 111: for some reason, we previously had the `local` server config hardcoded here, and the `testnet` one hardcoded elsewhere.

This led to surprising results from innocuous diffs, e.g. [this change](https://github.com/clockworklabs/SpacetimeDB/pull/1078/files/124ca39c1012acb31d6cb39b783c5c1e996b5c74) would leave us with two `testnet`s and no `local`!

joshua-spacetime (approved):

PR #1153, opened 2024-04-25
NFC: inline format args in `generate` code

RReverser:
CLI `generate` code is a particularly heavy user of format-based macros, and so it benefits most from inlining format args where possible.

This is done by adding `#![warn(clippy::uninlined_format_args)]` + running `cargo clippy --fix` followed by `cargo fmt`, so shouldn't require manual review.

This is a follow-up to #1142 and, like that PR, is mainly done to make generate's code a bit cleaner and diffs simpler.

# Description of Changes

corresponding GitHub label.

# Testing


kazimuth (approved):
👍

PR #1155, opened 2024-04-25
Fail CI on smoketest failures

coolreader18:
# Description of Changes

corresponding GitHub label.

1

# Testing

### DIFF
--- a/smoketests/__init__.py
+++ b/smoketests/__init__.py
@@ -143,7 +143,7 @@ class Smoketest(unittest.TestCase):
 
     def fingerprint(self):
         # Fetch the server's fingerprint; required for `identity list`.
-        self.spacetime("server", "fingerprint", "localhost", "-f")
+        self.spacetime("server", "fingerprint", "-s", "localhost", "-f")
     
     def new_identity(self, *, email, default=False):
         output = self.spacetime("identity", "new", "--no-email" if email is None else f"--email={email}")
--- a/smoketests/tests/servers.py
+++ b/smoketests/tests/servers.py
@@ -15,11 +15,11 @@ class Servers(Smoketest):
         self.assertRegex(servers, re.compile(r"^\s*testnet\.spacetimedb\.com\s+https\s+testnet\s*$", re.M))
         self.assertRegex(servers, re.compile(r"^\s*\*\*\*\s+127\.0\.0\.1:3000\s+http\s+localhost\s*$", re.M))
 
-        out = self.spacetime("server", "fingerprint", "127.0.0.1:3000", "-f")
+        out = self.spacetime("server", "fingerprint", "-s", "http://127.0.0.1:3000", "-f")
         self.assertIn("No saved fingerprint for server 127.0.0.1:3000.", out)
 
-        out = self.spacetime("server", "fingerprint", "127.0.0.1:3000")
+        out = self.spacetime("server", "fingerprint", "-s", "http://127.0.0.1:3000")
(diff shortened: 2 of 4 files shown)


smoketests/__init__.py
```diff
@@ -143,7 +143,7 @@ def reset_config(cls):
 
     def fingerprint(self):
         # Fetch the server's fingerprint; required for `identity list`.
-        self.spacetime("server", "fingerprint", "localhost", "-f")
+        self.spacetime("server", "fingerprint", "-s", "localhost", "-f")
```
jdetter, line 146: the local server is called `local` not `localhost` I believe

smoketests/tests/servers.py
```diff
@@ -15,11 +15,11 @@ def test_servers(self):
         self.assertRegex(servers, re.compile(r"^\s*testnet\.spacetimedb\.com\s+https\s+testnet\s*$", re.M))
         self.assertRegex(servers, re.compile(r"^\s*\*\*\*\s+127\.0\.0\.1:3000\s+http\s+localhost\s*$", re.M))
 
-        out = self.spacetime("server", "fingerprint", "127.0.0.1:3000", "-f")
+        out = self.spacetime("server", "fingerprint", "-s", "http://127.0.0.1:3000", "-f")
```
jdetter, line 18: instead of `http://127.0.0.1:3000` I would recommend just using `local`

smoketests/tests/servers.py
```diff
@@ -15,11 +15,11 @@ def test_servers(self):
...
         self.assertIn("No saved fingerprint for server 127.0.0.1:3000.", out)
 
-        out = self.spacetime("server", "fingerprint", "127.0.0.1:3000")
+        out = self.spacetime("server", "fingerprint", "-s", "http://127.0.0.1:3000")
```
jdetter, line 21: instead of `http://127.0.0.1:3000` I would recommend just using `local`

smoketests/__init__.py
jdetter: also we've changed this recently so it makes sense this is failing

smoketests/tests/servers.py
bfops, line 21:
```suggestion
        out = self.spacetime("server", "fingerprint", "-s", "local")
```

smoketests/tests/servers.py
bfops:
```suggestion
        out = self.spacetime("server", "fingerprint", "-s", "local", "-f")
```

smoketests/tests/servers.py
bfops:
```suggestion
        out = self.spacetime("server", "fingerprint", "-s", "local")
```

smoketests/__init__.py
bfops: it looks like we explicitly configure the server in `smoketests/config.toml`

bfops (approved):
LGTM, indifferent on remaining nits

smoketests/tests/servers.py
coolreader18: mm... this is testing the `fingerprint` subcommand specifically, that it accepts using urls instead of names. I'm gonna leave it as is.

smoketests/__init__.py
coolreader18: No, that's not the issue - it's passing now.

jdetter (approved):
LGTM

PR #1162, opened 2024-04-26
Fix soundness hole in `Table::delete` + don't make & immedately drop PVs in the method

### LINKED ISSUE #1083: `MutTxId::insert_row_internal`: don't create PV and throw it away
It's this stuff:
```rust
                        tx_table
                            .delete(tx_blob_store, ptr)
                            .expect("Failed to delete a row we just inserted");
```

Centril:
# Description of Changes

- There was a soundness hole in `Table::delete` where we didn't confirm that the row was present.
- In many cases of `Table::delete`, we make a PV just to immediately drop it. Fixes #1083.

Future work: consider changing `Table::find_same_row` to work on `&ProductValue` instead, so we don't have to insert rows temporarily anywhere.

None

2

### DIFF
--- a/crates/table/src/table.rs
+++ b/crates/table/src/table.rs
@@ -415,6 +415,13 @@ impl Table {
     }
 
     /// Deletes the row identified by `ptr` from the table.
+    ///
+    /// The function `before` is run on the to-be-deleted row,
+    /// if it is present, before deleting.
+    /// This enables callers to extract the deleted row.
+    /// E.g. applying deletes when squashing/merging a transaction into the committed state
+    /// passes `|row| row.to_product_value()` as `before`
+    /// so that the resulting `ProductValue`s can be passed to the subscription evaluator.
     pub fn delete<'a, R>(
         &'a mut self,
         blob_store: &'a mut dyn BlobStore,


crates/table/src/table.rs
```diff
@@ -426,7 +410,30 @@ impl Table {
             debug_assert!(deleted);
         }
 
-        self.delete_internal(blob_store, ptr)
+        // SAFETY: We've checked above that `self.is_row_present(ptr)`.
+        unsafe { self.delete_internal(blob_store, ptr) }
+    }
+
+    /// Deletes the row identified by `ptr` from the table.
```
gefjon, line 417:
```suggestion
    /// Deletes the row identified by `ptr` from the table.
    ///
    /// The function `before` is run on the to-be-deleted row,
    /// if it is present, before deleting.
    /// This enables callers to extract the deleted row.
    /// E.g. applying deletes when squashing/merging a transaction into the committed state
    /// passes `|row| row.to_product_value()` as `before`
    /// so that the resulting `ProductValue`s can be passed to the subscription evaluator.
```

gefjon (approved):

Issue #1118, opened 2024-04-18
Don't use `ColumnOp` in query plans; store a different version where `FieldName` becomes `AbsColPos`


Centril:
This is a stepping stone towards implementing https://github.com/clockworklabs/SpacetimeDB/issues/1113.

cc @joshua-spacetime

Issue #1117, opened 2024-04-18
`Column`: replace `FieldName` with `AbsColPos = (TableId, ColId)`


Centril:
This is a more compact representation!

cc https://github.com/clockworklabs/SpacetimeDB/issues/1113
cc @joshua-spacetime

PR #1165, opened 2024-04-26
Redefine FieldName as `(TableId, ColId)`

Centril:
# Description of Changes

Do not store table names and field names in `FieldName`.
Rather, store the table id and the column id which is substantially cheaper (8 bytes).
Fixes a bug in `fn column_pos` along the way and does some drive-by cleanup.

The second commit adds both necessary changes to tests and refactoring around them, and similar changes which I had to do to understand and be confident in the changes to the tests. The diff got larger than I wanted, but once merged, I think it will help us long term.

Fixes https://github.com/clockworklabs/SpacetimeDB/issues/1118.
Fixes https://github.com/clockworklabs/SpacetimeDB/issues/1117.

None

2


gefjon (approved):
Tests look cleaner, column accesses look more efficient.

PR #1166, opened 2024-04-26
`IndexJoin`/`JoinInner`: store `ColId`

Centril:
# Description of Changes

1. Store `ColId` in `IndexJoin` and `JoinInner` as opposed to `FieldName`. This avoids querying `Header` in `build_query`.
2. Store `Header`s for inner joins and projections. `QueryExpr::head` now provides the end-result header for the entire query, which is the same header as you'd get from `build_query`. This means that we can eventually stop making a `Header` in `build_query` and thereby move more runtime to query compilation and spend less in query building / evaluation.
3. Dedup logic in the two `build_query` functions.

None

2


gefjon:
Oh, I see. I missed the fact that the row in question is not within the range.

gefjon:
If we had `BTreeIndexIter::<K, V>::skip_to_next_greater_than(&mut self, min_bound: &K)` (the leapfrog join operator) we could use multi-column indexes to answer range queries...

PR #1172, opened 2024-04-26
fix(1170): Use scope guard to decrement reducer queue length

### LINKED ISSUE #1170: Fix reducer queue length metric
It's still not correct

joshua-spacetime:
Fixes #1170.

Also updates the bucket values for the queue length histogram. Also removes the max queue length metric, since the histogram should suffice.

# Description of Changes

corresponding GitHub label.

# Testing

### DIFF
--- a/crates/core/src/util/lending_pool.rs
+++ b/crates/core/src/util/lending_pool.rs
@@ -12,7 +12,7 @@ use parking_lot::Mutex;
 use spacetimedb_lib::Address;
 use tokio::sync::{OwnedSemaphorePermit, Semaphore};
 
-use crate::worker_metrics::{MAX_QUEUE_LEN, WORKER_METRICS};
+use crate::worker_metrics::WORKER_METRICS;
 
 use super::notify_once::{NotifiedOnce, NotifyOnce};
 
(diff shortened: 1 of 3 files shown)


crates/core/src/util/lending_pool.rs
```diff
@@ -58,31 +87,9 @@ impl<T> LendingPool<T> {
...
-                }
-            })
-            .or_insert_with(|| new_queue_len);
-
-        drop(guard);
-        queue_len_max.set(max_queue_len);
-
         async move {
+            let _guard = QueueMetric::inc(db);
             let permit = acq.await.map_err(|_| PoolClosed)?;
```
joshua-spacetime, line 92: I'm fairly confident we should not have any more issues with this metric not resetting, or this metric going negative, with this scope guard in place.

crates/core/src/util/lending_pool.rs
```diff
@@ -49,6 +49,35 @@ struct PoolVec<T> {
...
+}
+
+impl Drop for QueueMetric {
+    fn drop(&mut self) {
+        WORKER_METRICS.instance_queue_length.with_label_values(&self.db).dec();
+        let queue_length = WORKER_METRICS.instance_queue_length.with_label_values(&self.db).get();
+        WORKER_METRICS
+            .instance_queue_length_histogram
+            .with_label_values(&self.db)
+            .observe(queue_length as f64);
```
coolreader18, line 65: why do we need a separate histogram? shouldn't grafana accumulate the data for us over time anyway?

coolreader18 (approved):
LGTM!

crates/core/src/util/lending_pool.rs
coolreader18: if you didn't need to se the histogram you could just use the `inc_scope` utility function I made a while ago, which does this same sorta thing.

joshua-spacetime: > why do we need a separate histogram? shouldn't grafana accumulate the data for us over time anyway?

The `instance_queue_length` metric is a gauge which is only sampled once per sampling interval. So all it gives us is the queue length at the end of the sampling interval over time. But it doesn't tell us anything about the queue length at the beginning or middle of the sampling interval.

`instance_queue_length_histogram` on the other hand will build a histogram of all the different values the queue length takes throughout the sampling interval, and this will give us percentiles.

It is true that `instance_queue_length` really doesn't give us any new information over the histogram. We just need some variable to track the current length of the queue. I just chose a prometheus metric for that purpose mainly for ease of use.

joshua-spacetime: > you could just use the inc_scope utility function I made a while ago

I would have done that if it weren't for the `get`

PR #1174, opened 2024-04-26
fix(1173): Record wait time for all reducers

### LINKED ISSUE #1173: Fix reducer wait time metric
At the moment we're only recording this metric for scheduled reducers. We should be recording it for all reducers. We're also recording it before we acquire the module instance semaphore.

joshua-spacetime:
Fixes #1173.

Previously we were only recording this metric for scheduled reducers. We were also recording it before we acquired access to the module instance. Now we record it for all reducers after we acquire access to the module instance.

This patch also removes max wait time since the histogram should suffice.

# Description of Changes

corresponding GitHub label.

# Testing

### DIFF
--- a/crates/core/src/host/module_host.rs
+++ b/crates/core/src/host/module_host.rs
@@ -14,6 +14,7 @@ use crate::protobuf::client_api::{TableRowOperation, TableUpdate};
 use crate::subscription::module_subscription_actor::ModuleSubscriptions;
 use crate::util::lending_pool::{Closed, LendingPool, LentResource, PoolClosed};
 use crate::util::notify_once::NotifyOnce;
+use crate::worker_metrics::WORKER_METRICS;
 use bytes::Bytes;
 use derive_more::{From, Into};
 use futures::{Future, FutureExt};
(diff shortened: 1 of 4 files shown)


crates/core/src/host/module_host.rs
```diff
@@ -636,12 +637,19 @@ impl ModuleHost {
...
         R: Send + 'static,
     {
-        let mut inst = self.inner.get_instance(self.info.address).await?;
+        let mut inst = {
+            // Record the time spent waiting in the queue
+            let _guard = WORKER_METRICS
+                .reducer_wait_time
+                .with_label_values(&self.info.address, reducer)
+                .start_timer();
+            self.inner.get_instance(self.info.address).await?
```
joshua-spacetime, line 651: @coolreader18 do you agree that this is the correct place to be recording this metric? And that this works for all reducers?

coolreader18: for all reducers and for other such calls. but yeah, this seems right.

coolreader18 (approved):

PR #1177, opened 2024-04-29
Switch to a better API for tagged enums for C#

RReverser:
# Description of Changes

With C# records - which are available since C# 9, so covers Unity requirements as well - we can use subclassing and pattern matching to get sum types that look a lot more like Rust tagged enums than my previous approach.

For example, the following Rust code:

```rust
let foo = match some_tagged_value {
  MyEnum::A(x) => ...,
  MyEnum::B(y) => ...
};
```

can now be expressed via almost equivalent C# code

```cs
var foo = someTaggedValue switch {
  MyEnum.A(var x) => ...,
  MyEnum.B(var y) => ...,
  // main difference is that, unlike in Rust, they're not exhaustive so you need to handle edge cases yourself
  _ => throw new ArgumentException(...)
};
```

This is a breaking change but IMO worth it for the better API going forward.

corresponding GitHub label.

2

# Testing

- [x] Checked that C# SDK tests are still working.

### DIFF
--- a/crates/bindings-csharp/Codegen/Module.cs
+++ b/crates/bindings-csharp/Codegen/Module.cs
@@ -128,7 +128,7 @@ public class Module : IIncrementalGenerator
                                         )
                                     "))} }}
                                 ),
-                                {t.Name}.GetSatsTypeInfo().AlgebraicType.TypeRef
+                                (SpacetimeDB.SATS.AlgebraicType.Ref){t.Name}.GetSatsTypeInfo().AlgebraicType
                             );
 
                             public static IEnumerable<{t.Name}> Query(System.Linq.Expressions.Expression<Func<{t.Name}, bool>> filter) =>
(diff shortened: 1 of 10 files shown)


kazimuth:
Oh this is extremely nice but it is going to be a huge patch for BitCraft right? How many enums are they using?

These records will also allocate, `record struct` isn't supported until C# 10 :/ I think the niceness of the API is worth the potential perf hit, we would have to profile to see if there was a serious difference.

Also note we can't use Unity serialization with records according to this: https://docs.unity3d.com/Manual/CSharpCompiler.html
I'm not sure if this will be a problem.

RReverser:
> Also note we can't use Unity serialization with records according to this: [docs.unity3d.com/Manual/CSharpCompiler.html](https://docs.unity3d.com/Manual/CSharpCompiler.html)
> I'm not sure if this will be a problem.

As for this, I'm not sure either. Given that Unity / client C# code in general doesn't support tagged enums altogether for now, this is at least not a breaking change, but we can return to this question in the future after we integrate it into the client SDK.

kazimuth:
Oh perfect, no complaints from me then.

kazimuth (approved):
👍

jdetter (approved):
I'm fine with this - we don't really have support for this on the client right now anyway so nothing to break there. We'll just have to include this in the 0.10 release notes because this would break modules who were already using tagged enums

PR #1186, opened 2024-04-30
core: Collapse DBIC into HostController

kim:
Make it so `HostController` manages both the module host (wasm machinery) and the database (`RelationalDB` / `DatabaseInstanceContext`) of spacetime databases deployed to a server.

The `DatabaseInstanceContextController` (DBIC) is removed in the process.

This allows to make database accesses panic-safe, in that uncaught panics will cause all resouces to be released and the database to be restarted on subsequent access. This is a prerequisite for #985.

It also allows to move towards storage of the module binary directly in the database / commitlog. This patch, however, makes some contortions in order to **not** introduce a breaking change just yet.

2.5

# Testing

- [x] All API things should still work without changes
- [x] Insert a `panic!` in the `durability` crate's background tasks, which is triggered with some probability.
       Deploy a test module and access it using the CLI (call reducer, get logs, sql). 
       Observe that the panic is triggered (in the server logs).
       Run another command and observe that it returns an error.
       Run it again and observe that the database restarts and the command succeeds.
       On a scale from 1-10, rate how confusing the error messages are.
       On a scale from 1-10, rate how confusing the fact is that the database does not eagerly restart.


kim:
Also, I am unsure if we tracked energy in standalone before, or if I introduced it accidentally.

crates/standalone/src/lib.rs
```diff
@@ -289,51 +275,53 @@ impl spacetimedb_client_api::ControlStateWriteAccess for StandaloneEnv {
...
+                let program_bytes_address = self.object_db.insert_object(spec.program_bytes)?;
+                let database = Database {
+                    num_replicas: spec.num_replicas,
+                    program_bytes_address,
+                    publisher_address,
+                    ..existing_db.clone()
+                };
+                let database_id = database.id;
+                let database_addr = database.address;
+                self.control_db.update_database(database.clone())?;
```
kim, line 316: The logic did not change in this patch, but I'll point out that it is flawed:

If we fail to run the update, the new database will keep running.
There is no obvious solution for that within the current event-driven design, because it is impossible to guarantee that we can restore the previous state.

It will need to be revised in a later patch.

kim:
@kazimuth @RReverser It is somewhat mysterious to me how the repeated `StandaloneEnv::init` with same database directory could possibly work. Is there something I'm missing, or was it just incidental that it worked?

kim:
> In both cases, the database was inaccessible after the panic.

That sounds wrong, it should come up again (lazily). Will investigate…

Thanks for testing!

crates/standalone/src/control_db.rs
```diff
@@ -457,15 +457,6 @@ impl ControlDb {
         Ok(id)
     }
 
-    pub fn update_database_instance(&self, database_instance: DatabaseInstance) -> Result<()> {
```
kim, line 460: Turns out the `DatabaseInstance` never actually changes, so this is not needed.

gefjon:
> > In both cases, the database was inaccessible after the panic.
> 
> That sounds wrong, it should come up again (lazily). Will investigate…

Disregard. It does in fact seem to come up again, I was just mistaking my `quickstart-chat` being unable to connect due to stale credentials.

crates/core/src/host/host_controller.rs
```diff
@@ -101,249 +194,649 @@ impl From<&EventStatus> for ReducerOutcome {
...
-            host.exit().await;
-        }
-        Ok(())
+    /// An error is returned if the host's program does not match the hash given
+    /// in [`Database`].
+    ///
+    /// If the computation `F` panics, the host is removed from this controller,
+    /// releasing its resources.
+    #[tracing::instrument(skip_all)]
+    pub async fn using_database<F, T>(&self, database: Database, instance_id: u64, f: F) -> anyhow::Result<T>
```
kim, line 291: Less than pleasant, arguably.
Can't think of a better way, though, because relational db is not `UnwindSafe`.

coolreader18: You could just use `AssertUnwindSafe` - we don't much care about unwinding anyway

kim: Maybe, but that would still mean to catch unwind in every single method of relational db. That seems silly.

crates/core/src/host/host_controller.rs
```diff
@@ -1,26 +1,105 @@
...
+///
+/// This is a temporary measure until modules are stored in their database itself.
+#[derive(Clone)]
+pub enum ProgramStorage {
+    /// External storage, addressable by program hash.
+    External(Arc<ExternalStorage>),
+    /// Storage inside the [`RelationalDB`] itself, but with a retrieval
+    /// function chosen by the user.
+    SameDb(Arc<SameDbStorage>),
+}
```
cloutiertyler, line 62: Ah this is cool

crates/core/src/host/host_controller.rs
```diff
@@ -1,26 +1,105 @@
...
+
+/// A host controller manages the lifecycle of spacetime databases and their
+/// associated modules.
+#[derive(Clone)]
 pub struct HostController {
-    modules: Mutex<IntMap<u64, ModuleHost>>,
-    pub energy_monitor: Arc<dyn EnergyMonitor>,
+    /// Map of all modules managed by this controller,
+    /// keyed by database instance id.
+    modules: Modules,
```
cloutiertyler, line 88: I would rename this to `hosts: Hosts` for consistency

crates/core/src/host/host_controller.rs
cloutiertyler: Just out of pure curiosity, what makes `RelationalDB` not `UnwindSafe`? Keep in mind I know very little about Rust stack unwinding

cloutiertyler (approved):
This is quite a positive improvement. I also like your naming scheme. I thought a bit and I could not come up with anything better than `Host`. I think it works.

I left one nit about renaming `modules: Modules` to `hosts: Hosts`. Could you make that change before merging?

crates/core/src/host/host_controller.rs
kim: `parking_lot` locks used in the datastore keep their data in a literal `UnsafeCell`. Catching a panic means code can continue to run (e.g. destructors), and the compiler cannot prove anything about the memory safety of an `UnsafeCell` (because that is what it's for :)). At least that's how I understand it.

I believe what Noa is saying is that our goal is to _drop_ the whole `Host` when there is a panic, so unless we have `Drop` impls which do silly things, this should be safe.

Thinking about it some more, it might be possible and sufficient to only catch panicking of `durability.append_tx` in `RelationalDB::commit_tx`. This is the place where we _deliberately_ panic, and once the method panicked it will continue to panic (to signal that the durability layer is in an irrecoverably broken state). Any other panic (e.g. bad unwrap) will make the current request panic, but the next one is likely to succeed.

This still requires to assert unwind safety, so I'll try it in a follow up for easier review.

crates/core/src/host/host_controller.rs
```diff
@@ -1,26 +1,105 @@
...
+//   spacetimedb, because a disconnect will make a node miss history events.
+//
+
+/// A shared mutable cell containing a module host and associated database.
+type HostCell = Arc<AsyncRwLock<Option<Host>>>;
+
+/// The registry of all running modules.
+type Modules = Arc<Mutex<IntMap<u64, HostCell>>>;
+
+type ExternalStorage = dyn Fn(&Hash) -> anyhow::Result<Option<AnyBytes>> + Send + Sync + 'static;
```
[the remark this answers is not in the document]
kim: To what?

crates/client-api/src/routes/database.rs
```diff
@@ -89,17 +88,10 @@ pub async fn call<S: ControlStateDelegate + NodeDelegate>(
...
-                .load_module_host_context(database, instance_id)
-                .await
-                .map_err(log_and_500)?;
-            host.spawn_module_host(dbic).await.map_err(log_and_500)?
-        }
-    };
+    let module = host
+        .get_or_launch_module_host(database, instance_id)
+        .await
+        .map_err(log_and_500)?;
```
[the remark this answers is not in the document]
kim: At least I have reduced the line count of the 4 occurrences by ~50% each :P

I think it is courtesy to #1147 to not make those lines conflict in a non-obvious way.

The whole `routes` module tree could certainly benefit from a grand cleanup, tho.

crates/core/src/host/host_controller.rs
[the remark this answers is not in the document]
kim: Maybe it's just me, but I find it common that the idiom `Result<Option<T>>` denotes "may error or not found". Obscuring that for the sake of less angle brackets doesn't help recognizing the pattern imho, so I'll ask for a more precise name. Not sure if I could be convinced to name everything sentence-like, like `MaybeErrorOrNotFound`.

coolreader18:
just fyi, I've got a version of #1147 ready for when this merges.

Issue #104, opened 2023-07-27
Unsoundness in `type_of_sum` returning a product, not sum type


Centril:
See https://github.com/clockworklabs/SpacetimeDB/blob/c3526ee61db444b7af850a0ad870352fd045425f/crates/sats/src/algebraic_value.rs#L373-L386.

marvin-hansen:
Does SATN actually support recursive types?
 I don't remember having seen it in the docs.

cloutiertyler:
> Does SATN actually support recursive types? I don't remember having seen it in the docs.

It does, yes.

PR #1189, opened 2024-04-30
Fix unsoundness in `AlgebraicValue::type_of` by making it partial

Centril:
# Description of Changes

First we reduce the number of `.type_of()` calls to one.
Then we make the method sound by making it partial without an answer in e.g., the case of sums.
Also cleans up the bsatn roundtrip test.

Fixes https://github.com/clockworklabs/SpacetimeDB/issues/104.

None

1

### DIFF
--- a/crates/sats/src/algebraic_value.rs
+++ b/crates/sats/src/algebraic_value.rs
@@ -230,6 +230,17 @@ impl AlgebraicValue {
     /// This function is partial
     /// as type inference is not possible for `AlgebraicValue` in the case of sums.
     /// Thus the method only answers for the decidable subset.
+    ///
+    /// # A note on sums
+    ///
+    /// The type of a sum value must be a sum type and *not* a product type.
+    /// Suppose `x.tag` is for the variant `VarName(VarType)`.
+    /// Then `VarType` is *not* the same type as `{ VarName(VarType) | r }`
+    /// where `r` represents a polymorphic variants component.
+    ///
+    /// To assign this a correct type we either have to store the type with the value
+    /// r alternatively, we must have polymorphic variants (see row polymorphism)
+    /// *and* derive the correct variant name.
     pub fn type_of(&self) -> Option<AlgebraicType> {
         match self {
             Self::Sum(_) => None,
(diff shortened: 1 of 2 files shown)


crates/sats/src/algebraic_value.rs
```diff
@@ -210,61 +210,46 @@ impl AlgebraicValue {
...
+        Some(AlgebraicType::product([k, v]))
     }
 
     /// Infer the [`AlgebraicType`] of an [`AlgebraicValue`].
-    pub fn type_of(&self) -> AlgebraicType {
-        // TODO: What are the types of empty arrays/maps/sums?
+    ///
+    /// This function is partial
+    /// as type inference is not possible for `AlgebraicValue` in the case of sums.
+    /// Thus the method only answers for the decidable subset.
```
mamcx, line 232: Why is not possible?

kazimuth: Because if you see `VariantZ(data)`, you can't, in general, know what other variants are possible.

crates/sats/src/array_value.rs
```diff
@@ -85,8 +85,11 @@ impl ArrayValue {
     /// This leads to e.g., an empty array of products having the type "never".
     /// This is the most conservative choice
     /// and has the consequence that no values can be added to such an array.
-    fn first_type_of<T>(arr: &[T], then: impl FnOnce(&T) -> AlgebraicType) -> AlgebraicType {
-        arr.first().map(then).unwrap_or_else(AlgebraicType::never)
+    fn first_type_of<T>(arr: &[T], then: impl FnOnce(&T) -> Option<AlgebraicType>) -> Option<AlgebraicType> {
+        match arr.first() {
+            None => Some(AlgebraicType::never()),
```
kazimuth, line 90: Should we just make this return None now?

crates/sats/src/algebraic_value.rs
mamcx: Oh right, we don't tag a unique id for each `sum`.

mamcx: I think this explanation should be in the comment

PR #1191, opened 2024-05-01
C#: rename DbEventArgs to ReducerContext

RReverser:
# Description of Changes

This is no longer an event type, so remove subclassing and rename to match the Rust API.

corresponding GitHub label.

1

# Testing

- [x] Checked that C# SDK tests still build.

### DIFF
--- a/crates/bindings-csharp/Codegen/Module.cs
+++ b/crates/bindings-csharp/Codegen/Module.cs
@@ -222,7 +222,8 @@ public class Module : IIncrementalGenerator
                             (
                                 p.Name,
                                 p.Type,
-                                IsDbEvent: p.Type.ToString() == "SpacetimeDB.Runtime.ReducerContext"
+                                IsContextArg: p.Type.ToString()
+                                    == "SpacetimeDB.Runtime.ReducerContext"
                             )
                         )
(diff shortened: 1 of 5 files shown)


crates/bindings-csharp/Codegen/Module.cs
```diff
@@ -247,7 +247,7 @@ class {r.Name}: IReducer {{
                                     );
                                 }}
 
-                                void IReducer.Invoke(BinaryReader reader, SpacetimeDB.Runtime.DbEventArgs dbEvent) {{
+                                void IReducer.Invoke(BinaryReader reader, SpacetimeDB.Runtime.ReducerContext dbEvent) {{
                                     {r.FullName}({string.Join(", ", r.Args.Select(a => a.IsDbEvent ? "dbEvent" : $"{a.Name}.Read(reader)"))});
```
cloutiertyler, line 251: This still references `"dbEvent"`, is that fine?

RReverser: Good catch. I stupidly replaced all instances of the class but not variable names.

kazimuth (approved):
More uniformity between module languages is great. 👍

PR #1192, opened 2024-05-02
Some clarifications to dropping the updates in eval incr

Centril:
# Description of Changes

Based on the latest flame-graph, we are spending notable time doing `drop_in_place` of the iterators of `IncrementalJoin::eval`.
At first, I thought this was because we were cloning and dropping 4x the `SmallVec`, but as these are inline and not heap allocated, that seems unlikely. So this PR tries to clarify in code that this isn't the case by using `ArrayVec` and borrows. Moreover, all the dropping work of the hash map is moved outside the rayon parallel iteration. Also, this makes the benchmarks slightly more accurate as they now do not clone-allocate + drop in place.

None

1

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4483,6 +4483,7 @@ name = "spacetimedb-core"
 version = "0.8.2"
 dependencies = [
  "anyhow",
+ "arrayvec",
  "async-trait",
  "backtrace",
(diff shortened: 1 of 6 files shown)


mamcx (approved):
LGTM

PR #1194, opened 2024-05-02
perf(`JoinInner`): use AVs for keys instead of PVs

Centril:
# Description of Changes

1. Reuse `join_inner` logic from spacetimedb_vm.
2. Dedup query tests slightly.
3. Precompute `ColIds` in `join_inner` and don't recompute it for every element.
4. Store `AlgebraicValue` keys in `JoinInner` as opposed to `ProductValue`. This removes an allocation per map entry.
5. Make `JoinInner` callbacks infallible.
6. In `JoinInner::next`, avoid clones unless necessary.

3-6 collectively, but primarily 4, results in the following perf improvements compared to HEAD~1 on i7-7700K, 64GB RAM:

```
incr-join               time:   [788.24 ns 790.71 ns 794.34 ns]
                        change: [-15.369% -14.407% -13.168%] (p = 0.00 < 0.05)
                        Performance has improved.
```

None

1

### DIFF
--- a/crates/vm/src/rel_ops.rs
+++ b/crates/vm/src/rel_ops.rs
@@ -272,7 +272,6 @@ impl<'a, Lhs, Rhs, KeyLhs, KeyRhs, Pred, Proj> RelOps<'a> for JoinInner<'a, Lhs,
 where
     Lhs: RelOps<'a>,
     Rhs: RelOps<'a>,
-    // TODO(Centril): consider using keys that aren't `ProductValue`s.
     KeyLhs: FnMut(&RelValue<'a>) -> AlgebraicValue,
     KeyRhs: FnMut(&RelValue<'a>) -> AlgebraicValue,
     Pred: FnMut(&RelValue<'a>, &RelValue<'a>) -> bool,


crates/vm/src/rel_ops.rs
```diff
@@ -273,9 +273,9 @@ where
     Lhs: RelOps<'a>,
     Rhs: RelOps<'a>,
     // TODO(Centril): consider using keys that aren't `ProductValue`s.
```
kazimuth, line 275:
```suggestion
```

crates/vm/src/rel_ops.rs
```diff
@@ -273,9 +273,9 @@ where
...
-    KeyLhs: FnMut(&RelValue<'a>) -> Result<ProductValue, ErrorVm>,
-    KeyRhs: FnMut(&RelValue<'a>) -> Result<ProductValue, ErrorVm>,
-    Pred: FnMut(&RelValue<'a>, &RelValue<'a>) -> Result<bool, ErrorVm>,
+    KeyLhs: FnMut(&RelValue<'a>) -> AlgebraicValue,
```
kazimuth, line 276: This can't return &AlgebraicValue because it might be freshly allocated, right?

kazimuth (approved):
LGTM 👍

PR #1195, opened 2024-05-02
feat(1168): Track subscription query duration for each reducer

### LINKED ISSUE #1168: Combine reducer and incremental update duration metrics
Right now reducer durations and incremental update durations are tracked using two different metrics. We should also have a metric that combines them since a reducer call and an incremental update are logically an atomic unit of execution.

This will give users additional observability into their reducers as they'll be able to see which reducers have more expensive incremental updates.

Note this will be a natural fallout of #1157 since both will be executed as part of the same transaction, but we'll want to make sure to maintain the separated metrics as well.

joshua-spacetime:
Closes #1168.

# Description of Changes

corresponding GitHub label.

# Testing

### DIFF
--- a/crates/core/src/host/wasm_common/module_host_actor.rs
+++ b/crates/core/src/host/wasm_common/module_host_actor.rs
@@ -27,6 +27,7 @@ use crate::module_host_context::ModuleCreationContext;
 use crate::sql;
 use crate::subscription::module_subscription_actor::ModuleSubscriptions;
 use crate::util::const_unwrap;
+use crate::util::prometheus_handle::HistogramExt;
 use crate::worker_metrics::WORKER_METRICS;
 use spacetimedb_sats::db::def::TableDef;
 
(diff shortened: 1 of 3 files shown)


crates/core/src/host/wasm_common/module_host_actor.rs
```diff
@@ -534,7 +535,26 @@ impl<T: WasmInstance> WasmModuleInstance<T> {
...
+            |tx @ MutTxId { timer, .. }| {
+                (
+                    tx,
+                    WORKER_METRICS
+                        .reducer_plus_query_duration
+                        .with_label_values(&address, op.name)
+                        .with_timer(timer),
+                )
+            },
+        );
```
coolreader18, line 557: why not just
```suggestion
        let tx = tx.unwrap_or_else(|| stdb.begin_mut_tx(IsolationLevel::Serializable));
        let _guard = WORKER_METRICS
            .reducer_plus_query_duration
            .with_label_values(&address, op.name)
            .with_timer(tx.timer);
```
?

coolreader18 (approved):

crates/core/src/host/wasm_common/module_host_actor.rs
joshua-spacetime: I don't know :)

PR #1196, opened 2024-05-02
Create a lockfile when opening config files

gefjon:
# Description of Changes

In the past, we've had issues where multiple concurrent CLI processes would race to read and write the CLI config file,
leading to data loss.

We considered using `flock`/`LockFileEx` and blocking until the file became available, but unfortunately it's not possible to atomically create and lock a nonexistent file, which we need to do in the case where the configuration doesn't yet exist.

Instead, we opt for a classic lockfile-based scheme: Before opening a config file `foo.conf`, attempt to exclusively create `foo.lock`, and panic if the exclusive creation fails. Once it becomes clear that we will not write the config any more, i.e. in `Config::drop`, delete the lockfile, allowing another process to operate.

This means that attempting to run multiple concurrent Spacetime CLI processes with the same config file is now a hard error.

Arguably this is a CLI interface break, in the sense that running multiple concurrent CLI processes with the same config file is now a hard error, where before it sometimes worked.

3 - CLI stuff is finicky and requires manual testing.

# Testing

- [x] Using the new CLI:
  - Create an identity.
  - List identities for a particular server, a non-default server, and all servers.
  - Publish a database.
  - Fingerprint a server.
  - Add a server configuration.
  - Delete an identity.
  - Access a module's logs.
  - Run a SQL query against a module.
- [x] Create a file `~/.spacetime/config.lock`, then attempt the above, and observe the error.
- [x] Delete `~/.spacetime/config.lock`, try again, and see that it works again.
- [x] Invoke the CLI with invalid arguments to trigger a Clap parse error (e.g. `spacetime lsit`), then run again with a valid parse (e.g. `spacetime identity list`), and verify that there is no stale lockfile preventing the corrected command.
- [x] Delete the `~/.spacetime` directory, then run a CLI command and verify that the lockfile can be created despite the directory not existing.
- [ ] Do whatever cursed concurrent invocations were exhibiting the bug before, and observe that they now show an error rather than silently corrupting or losing data.

### DIFF
--- a/crates/cli/src/config.rs
+++ b/crates/cli/src/config.rs
@@ -109,6 +109,48 @@ pub struct RawConfig {
 pub struct Config {
     proj: RawConfig,
     home: RawConfig,
+
+    /// Lockfile for the project config, if a file exists for the project config.
+    ///
+    /// Created before reading the project config, and closed upon dropping the config.
+    /// Allows us to mutate the config via [`Config::save`] without worrying about other processes.
+    proj_lock: Option<PathBuf>,
+
+    /// Lockfile for the system config.
+    ///
+    /// Created before reading or creating the home config, and closed upon dropping the config.
+    /// Allows us to mutate the config via [`Config::save`] without worrying about other processes.
+    home_lock: PathBuf,
+}
+
+impl Drop for Config {
+    fn drop(&mut self) {
+        // Delete lockfiles so other processes can access the config.
+        let proj_result = self.proj_lock.as_ref().map(std::fs::remove_file);
+        let home_result = std::fs::remove_file(&self.home_lock);
+
+        // Release both locks before checking for errors to avoid stale lockfiles sitting around.
+        match (proj_result, home_result) {
+            (Some(Err(proj_err)), Err(home_err)) => panic!(
+                "Encountered errors while releasing both the project configuration lock and system configuration lock.
+
+System: {home_err:#?}
+
+Project: {proj_err:#?}"
+            ),
+            (Some(Err(proj_err)), _) => panic!(
+                "Encountered error while releasing the project configuration lock.
+
+{proj_err:#?}"
+            ),
+            (_, Err(home_err)) => panic!(
+                "Encountered error while releasing the system configuration lock.
+
+{home_err:#?}"
+            ),
+            _ => (),
+        }
+    }
 }
 
 const HOME_CONFIG_DIR: &str = ".spacetime";
@@ -790,13 +832,52 @@ impl Config {
             .find(|path| path.exists())
     }
 
-    fn load_raw(config_dir: PathBuf, is_project: bool) -> RawConfig {
+    fn lockfile_path(config_path: impl AsRef<Path>) -> PathBuf {
+        let mut lockfile = config_path.as_ref().to_path_buf();
+        lockfile.set_extension("lock");
+        lockfile
+    }
+
+    fn create_lockfile_or_panic(config_path: impl AsRef<Path> + std::fmt::Debug) -> PathBuf {
+        let lockfile = Self::lockfile_path(&config_path);
+        std::fs::File::create_new(&lockfile).unwrap_or_else(|e| {
+            panic!(
+                "Unable to acquire lock on config file: {config_path:?}.
+
+Attempted to create a lockfile {lockfile:?}, but got: {e:#?}"
+            )
+        });
+        lockfile
+    }
+
+    /// Load a [`RawConfig`] from a file, creating a lockfile.
+    ///
+    /// The second value is the path of the created lockfile.
+    ///
+    /// Panics if unable to create the lockfile, e.g. because another process has already locked the config.
+    ///
+    /// If `!is_project` and the env var `SPACETIME_CONFIG_FILE` is set,
+    /// use that rather than `config_dir`.
+    ///
+    /// If `is_project` and the path does not exist, return an empty config.
+    /// Do not create a lockfile.
+    ///
+    /// If `!is_project` and the path does not exist, return a config with sensible defaults.
+    /// Do create a lockfile.
+    fn load_raw(config_dir: PathBuf, is_project: bool) -> (RawConfig, Option<PathBuf>) {
         // If a config file overload has been specified, use that instead
         if !is_project {
             if let Some(config_path) = std::env::var_os("SPACETIME_CONFIG_FILE") {
-                return Self::load_from_file(config_path.as_ref())
-                    .inspect_err(|e| eprintln!("SPACETIME_CONFIG_FILE does not point to a valid config file: {e:#}"))
-                    .unwrap_or_default();
+                let lockfile = Self::create_lockfile_or_panic(&config_path);
+
+                return (
+                    Self::load_from_file(config_path.as_ref())
+                        .inspect_err(|e| {
+                            eprintln!("SPACETIME_CONFIG_FILE does not point to a valid config file: {e:#}")
+                        })
+                        .unwrap_or_default(),
+                    Some(lockfile),
+                );
             }
         }
         if !config_dir.exists() {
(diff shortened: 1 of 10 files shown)


gefjon:
Ugh, this is no good because `spacetime start` reads the config and acquires the lock, even though it doesn't ever inspect the config.

gefjon:
Ugh, if Clap parsing fails, it doesn't panic; instead it directly calls `exit` (presumably in order to suppress the panic message). This means that unwinding doesn't happen and destructors are never called, so lockfiles get left in place. Gross.

crates/cli/src/config.rs
```diff
@@ -109,6 +109,51 @@ pub struct RawConfig {
...
+    proj_lock: Option<PathBuf>,
+
+    /// Lockfile for the system config.
+    ///
+    /// Created before reading or creating the home config, and closed upon dropping the config.
+    /// Allows us to mutate the config via [`Config::save`] without worrying about other processes.
+    ///
+    /// None only in tests, which are allowed to concurrently mutate configurations as much as they want,
+    /// since they use a hardcoded configuration rather than loading from a file.
+    home_lock: Option<PathBuf>,
```
bfops, line 126: Instead of shuffling around `PathBuf`s, how would you feel about making an actual type `LockFile` or something? It would have its own `drop`, and instead of `create_lockfile_or_panic`, we'd have a function that either panicks or returns a `LockFile`.

Then, it wouldn't just take a `PathBuf` to construct a `Config` - it requires passing a `LockFile`, which represents an actual acquired `LockFile`. (It still doesn't stop you from passing one `LockFile` in place of another.. but it helps prevent passing along a path without remembering to call the lock function)

bfops: I guess one could also imagine a `LockedConfig` class or some such, rather than bundling the lockfiles into `Config`. It would just tightly couple access to the the `Config` instance, with holding the lockfile for the correct  corresponding path, and when one is dropped the other is dropped. Then we wouldn't be passing along multiple paths/lockfiles with any risk for confusing them with one another (they'd be tightly coupled with their corresponding `Config`s, and share the same lifetime constraints with those configs rather than with one another)

gefjon: It turns out we don't actually need to lock the `proj` config, since we never write to it. So now we only have a single `Lockfile` on the `home` config.

crates/cli/src/config.rs
```diff
@@ -790,86 +852,96 @@ impl Config {
...
         // TODO(cloutiertyler): For now we're checking for a spacetime.toml file
         // in the current directory. Eventually this should really be that we
         // search parent directories above the current directory to find
         // spacetime.toml files like a .gitignore file
         let cur_dir = std::env::current_dir().expect("No current working directory!");
-        let cur_config = Self::load_raw(cur_dir, true);
+        if let Some(config_path) = Self::find_config_path(&cur_dir) {
+            Self::load_from_file(&config_path)
+                .inspect_err(|e| eprintln!("config file {config_path:?} is invalid: {e:#}"))
+                .unwrap_or_default()
```
kazimuth, line 879: I know it was like this before, but it might be worth raising the error here. I would prefer a CLI tool crashes rather than continuing with unknown defaults when there's a config error.

kazimuth (approved):
One comment, otherwise LGTM 👍

crates/cli/src/config.rs
gefjon: I agree with you, but I'd like @jdetter to weigh in before I change any semantics more than is required for the ticket.

crates/cli/src/config.rs
```diff
@@ -790,86 +852,96 @@ impl Config {
...
+        Config {
+            home: self.home.clone(),
+            proj: self.proj.clone(),
+            home_file: None,
+        }
+    }
 
-            Self::find_config_path(&config_dir).unwrap_or_else(|| config_dir.join(CONFIG_FILENAME))
+    pub fn save(&self) {
+        let Some((home_path, _)) = &self.home_file else {
```
bfops, line 932: is this definitely correct? any chance that this drops the lock immediately?

bfops: sounds like this might insta-drop? https://stackoverflow.com/questions/76311007/what-happens-when-assigning-to-the-underscore-pattern

(either way, nit: explicit drop might make it clearer for everyone :stuck_out_tongue: )

crates/cli/src/subcommands/generate/mod.rs
```diff
@@ -98,7 +100,10 @@ pub fn cli() -> clap::Command {
         .after_help("Run `spacetime help publish` for more detailed information.")
 }
 
-pub fn exec(args: &clap::ArgMatches) -> anyhow::Result<()> {
+pub fn exec(config: Config, args: &clap::ArgMatches) -> anyhow::Result<()> {
+    // Release the lockfile on the config, since we don't need it.
+    drop(config);
```
bfops, line 105: nit on this and others: it sounds like matching with `_` might implicitly do this?

If not, might be nice to add a `config.release_lock()` or something convenience function, so this docstring can be in one place and the code can self-document a bit more.

crates/cli/src/config.rs
```diff
@@ -105,10 +105,72 @@ pub struct RawConfig {
...
+    }
+}
+
+impl Drop for Lockfile {
+    fn drop(&mut self) {
+        std::fs::remove_file(&self.path)
+            .with_context(|| format!("Unable to remove lockfile {:?}", self.path))
+            .unwrap();
+    }
+}
```
bfops, line 155: nit: FWIW this is veeeeery close to a `Config`-independent struct.

do you have any opinions on factoring it out a la https://github.com/clockworklabs/SpacetimeDB/pull/1202/files

crates/cli/src/config.rs
```diff
@@ -790,86 +852,96 @@ impl Config {
...
+    }
 
-            Self::find_config_path(&config_dir).unwrap_or_else(|| config_dir.join(CONFIG_FILENAME))
+    pub fn save(&self) {
+        let Some((home_path, _)) = &self.home_file else {
+            return;
         };
 
+        let parent = home_path.parent().unwrap();
+        if parent != Path::new("") {
```
bfops, line 937: why did this become a `parent != Path::new("")` instead of a `parent.exists()`?

bfops (commented):
Looks pretty good, left some nits and one I-think-bug.

What do you think about splitting out the internal-only refactors into a base PR, a la https://github.com/clockworklabs/SpacetimeDB/pull/1203/files

crates/cli/src/config.rs
bfops: (I'm assuming we want to hold the lock until after the write.. right..?)

gefjon: No, this does not drop the lock. If we were taking ownership of `self.home_file` it would, but because we're just borrowing the lock stays intact within `self.home_file`.

crates/cli/src/subcommands/generate/mod.rs
gefjon: Moving into a `_` pattern does drop, but I prefer being more explicit because I think that fact is not obvious. It would be a valid choice for Rust to make it so that moving into `_` defers the drop until the end of scope, similar to moving into an unused binding like `_foo`.

I like the `release_lock` idea.

crates/cli/src/config.rs
gefjon: No strong feelings one way or the other. If you prefer that, I'm into it.

crates/cli/src/config.rs
gefjon: - `make_dir_all` on an existing directory is a no-op, so we don't need to check for that.
- `make_dir_all` on the empty path is invalid, so we do need to avoid doing that to handle the case where the config path is a relative path with no directory component.

crates/cli/src/config.rs
bfops: ahh I see - we're never dropping the lock in this function. idk why I was sure this must drop the lock :smiling_face_with_tear:

crates/cli/src/subcommands/generate/mod.rs
bfops: yeah it does seem easy to accidentally change semantics by "trivially" changing code (e.g. literally the variable names)

crates/cli/src/config.rs
gefjon: Encapsulated in a helper function with comments.

crates/cli/src/subcommands/generate/mod.rs
gefjon: Added a `release_lock` method.

jdetter (approved):
Tested + working. Originally I hit a snag on windows but I am now not able to reproduce it at all and it might be a somewhat rare issue. Even if the issue occurs it didn't currupt my config file which is what I care about.

Thanks Phoebe 👍

PR #1198, opened 2024-05-03
Jeremie/notify sql

lcodes:
# Description of Changes

- Added a new parameter on execute_sql to bring ModuleSubscriptions into scope.
- Augmented Code::Pass with an optional Update structure to thread inserts/deletes back into execute_sql where they can be broadcast as an event to subscribers.
- Updated tests to reflect these changes.

2? Still a few things I'm unsure of:
- Changed the return from evaluating delete from a count to an Update structure holding the count, but (afaik) the count wasn't used.
- The new Update structure is almost identical to the one from core it then converts into, would it be preferable to move that one instead?
- Moved ast transform and collect_result outside the transaction scope in execute_sql. Assuming these changes are safe.

# Testing

Still need to add tests to validate that changes broadcast correctly.

### DIFF
--- a/crates/client-api/src/routes/database.rs
+++ b/crates/client-api/src/routes/database.rs
@@ -523,20 +523,25 @@ where
     let instance_id = database_instance.id;
 
     let host = worker_ctx.host_controller();
-    let module_host = host.get_or_launch_module_host(database.clone(), instance_id).await.map_err(log_and_500)?;
-    let json = host.using_database(
-        database,
-        instance_id,
-        move |db| -> axum::response::Result<_, (StatusCode, String)> {
-            tracing::info!(sql = body);
-            let results = sql::execute::run(db, &body, auth, Some(&module_host.info().subscriptions)).map_err(|e| {
-                log::warn!("{}", e);
-                if let Some(auth_err) = e.get_auth_error() {
-                    (StatusCode::UNAUTHORIZED, auth_err.to_string())
-                } else {
-                    (StatusCode::BAD_REQUEST, e.to_string())
-                }
-            })?;
+    let module_host = host
+        .get_or_launch_module_host(database.clone(), instance_id)
+        .await
+        .map_err(log_and_500)?;
+    let json = host
+        .using_database(
+            database,
+            instance_id,
+            move |db| -> axum::response::Result<_, (StatusCode, String)> {
+                tracing::info!(sql = body);
+                let results =
+                    sql::execute::run(db, &body, auth, Some(&module_host.info().subscriptions)).map_err(|e| {
+                        log::warn!("{}", e);
+                        if let Some(auth_err) = e.get_auth_error() {
+                            (StatusCode::UNAUTHORIZED, auth_err.to_string())
+                        } else {
+                            (StatusCode::BAD_REQUEST, e.to_string())
+                        }
+                    })?;
 
                 let json = db.with_read_only(&ctx_sql(db), |tx| {
                     results
--- a/crates/client-api/src/routes/database.rs
+++ b/crates/client-api/src/routes/database.rs
@@ -523,22 +523,13 @@ where
     let instance_id = database_instance.id;
 
     let host = worker_ctx.host_controller();
-    let module_host = match host.get_module_host(instance_id) {
-        Ok(module_host) => module_host,
-        Err(_) => {
-            let dbic = worker_ctx
-                .load_module_host_context(database, instance_id)
-                .await
-                .map_err(log_and_500)?;
-            host.spawn_module_host(dbic).await.map_err(log_and_500)?
-        }
-    };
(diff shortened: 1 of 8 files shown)


crates/core/src/sql/execute.rs
```diff
@@ -56,39 +74,61 @@ pub(crate) fn collect_result(result: &mut Vec<MemTable>, r: CodeResult) -> Resul
...
+    };
     slow_logger.log();
 
-    Ok(result)
+    let mut tables = Vec::with_capacity(total);
+    let mut updates = Vec::with_capacity(total);
+    collect_result(&mut tables, &mut updates, code?.into())?;
+
+    if !reads && !updates.is_empty() {
+        let event = ModuleEvent {
```
lcodes, line 107: Left most of the values to defaults here, assuming these relate to reflecting the state of a reducer/client which aren't present in this case.

Only handling the committed case, as collect_result short-circuits the function on error.

crates/core/src/vm.rs
```diff
@@ -447,10 +447,17 @@ impl<'db, 'tx> DbProgram<'db, 'tx> {
                 // TODO: How do we deal with mutating values?
                 Table::MemTable(_) => Err(ErrorVm::Other(anyhow::anyhow!("How deal with mutating values?"))),
                 Table::DbTable(x) => {
+                    let inserts = rows.clone();
```
lcodes, line 450: Would prefer not to clone here, but db.insert wants a row, not a &row. Is it worth extending the refactor in there too?

crates/core/src/vm.rs
```diff
@@ -518,15 +533,15 @@ impl<'db, 'tx> DbProgram<'db, 'tx> {
                         }
                     }
                 }
-                Ok(Code::Pass)
+                Ok(Code::Pass(None)) // TODO going to assume dropping a table doesn't need to broadcast its entire contents as a notify
```
lcodes, line 536: Assumption here is deleting a table a reducer/subscriber depends on is either undefined behavior or breaking things up.

crates/core/src/vm.rs
```diff
@@ -518,15 +533,15 @@ impl<'db, 'tx> DbProgram<'db, 'tx> {
...
             }
             TxMode::Tx(_) => unreachable!("mutable operation is invalid with read tx"),
         }
     }
 
     fn _set_config(&mut self, name: String, value: AlgebraicValue) -> Result<Code, ErrorVm> {
         self.db.set_config(&name, value)?;
-        Ok(Code::Pass)
+        Ok(Code::Pass(None)) // TODO can subscriptions watch config values?
```
lcodes, line 544: Unsure how to broadcast these, as there's no "sql update" happening here. Is there a path to notify config changes?

crates/core/src/vm.rs
gefjon: Correct; modifying the schema which clients depend on is unsupported.

crates/core/src/vm.rs
gefjon: Clients should only care about inserts and deletes in non-system tables; anything else doesn't need to be broadcast.

crates/core/src/sql/execute.rs
```diff
@@ -21,17 +28,27 @@ pub struct StmtResult {
...
-        CodeResult::Pass => {}
+        CodeResult::Pass(x) => match x {
+            None => {},
+            Some(update) => {
+                updates.push(DatabaseTableUpdate {
+                    table_name: update.table_name,
+                    table_id: update.table_id,
+                    inserts: update.inserts.into(),
+                    deletes: update.deletes.into()
+                });
```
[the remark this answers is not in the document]
lcodes: That's the idea, except the types are slightly different. Where would you place DatabaseTableUpdate so its visible to both crates, and it can still impl its traits?

Or is it preferred to add spacetimedb-core as a vm dependency?

crates/vm/src/expr.rs
```diff
@@ -2057,14 +2057,23 @@ impl AuthAccess for CrudExpr {
     }
 }
 
+#[derive(Debug, PartialEq)]
+pub struct Update {
+    pub affected_rows: u32,
```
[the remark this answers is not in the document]
lcodes: You're right. Was it also unused from the delete op before?

crates/client-api/src/routes/database.rs
```diff
@@ -522,45 +522,47 @@ where
...
-                        .collect::<Vec<_>>()
-                });
-
-                Ok(json)
-            },
-        )
+    let host = worker_ctx.host_controller();
+    let module_host = host.get_or_launch_module_host(database.clone(), instance_id).await.map_err(|e| {
+        log::warn!("{}", e);
+        (StatusCode::INTERNAL_SERVER_ERROR, "Internal Server Error")
```
lcodes, line 528: Okay this is a bit aggressive, but I think this is the wanted behavior

PR #1202, opened 2024-05-03
Move Lockfile to its own file

bfops:
# Description of Changes

corresponding GitHub label.

# Testing

### DIFF
--- a/crates/cli/src/lockfile.rs
+++ b/crates/cli/src/lockfile.rs
@@ -1,0 +1,51 @@
+use anyhow::Context;
+use std::path::{Path, PathBuf};
+
+#[derive(Debug)]
+/// A file used as an exclusive lock on access to another file.
+///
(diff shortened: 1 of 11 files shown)


crates/cli/src/lockfile.rs
```diff
@@ -0,0 +1,58 @@
...
+impl Lockfile {
+    /// Acquire an exclusive lock on the file `lock_path`.
+    pub fn acquire(lock_path: impl AsRef<Path>) -> anyhow::Result<Self> {
+        // Ensure the directory exists before attempting to create the lockfile.
+        let parent = lock_path.as_ref().parent().unwrap();
+        if parent != Path::new("") {
+            std::fs::create_dir_all(parent).with_context(|| {
+                format!(
+                    "Unable to create directory {parent:?} to build lockfile for {:?}",
+                    lock_path.as_ref(),
```
bfops, line 27: note this does degrade our error messages slightly in some places, but I think they are still pretty explanatory?

PR #1207, opened 2024-05-06
Split `ColumnOp` into one with row indices and one with `FieldName` & other enabled changes

Centril:
# Description of Changes

## The main goal of this PR

This PR has some main goals, while other changes are less important and are done as additional work (see commit descriptions). I recommend reviewing this PR commit by commit after having reviewed https://github.com/clockworklabs/SpacetimeDB/pull/1166 upon which this PR is based.

In this PR, we want to:

1. Let `build_query` and query execution work on a `ColumnOp`  that stores `ColId` rather than `FieldName` (23e00d23cd22bc6d495c26e036c84503a9ef9db3). The previous version of `ColumnOp` is now called `FieldOp` and uses `FieldName` as before. This type is used by SQL compilation and query planning and is eventually compiled down to a `ColumnOp`. This means that `Header` is no longer used by `ColumnOp` nor `build_query`.
2. Making `build_query` itself (20340c88606d6e0af4be5e962938b2f453d33da7) and query execution infallible (3415702df3bba5fe2485c2ddb2a218fddbc614c5 and 986d80e70e5ee500bdcca7d10b8d54f3d5ec9119). To make this possible, we must ensure that we only get `bool`s where anything else does not make sense (977b5bb9db7636c38e6e1121222aea24415dfd9b).
3. Split `IndexSemiJoin` into left and right versions (e0c12cef61ed2eb51f361741ce5e60c57a9d51d2 and 3968f4fec39a8d1a91e14571674e3b39d5bcb51d). This enables us to make both versions less branchy and to significantly shrink the size of the right case (312 bytes vs 64).

## Benchmarks

Benchmarks relative to based master on i7-7700K, 64GB RAM:

```
Benchmarking full-join: Collecting 100 samples in estimated 5.
full-join               time:   [270.49 µs 271.55 µs 273.56 µs]
                        change: [-11.546% -9.6360% -8.2237%] (p = 0.00 < 0.05)
                        Performance has improved.

Benchmarking incr-select: Collecting 100 samples in estimated 
incr-select             time:   [151.49 ns 152.82 ns 154.70 ns]
                        change: [-22.532% -22.144% -21.611%] (p = 0.00 < 0.05)
                        Performance has improved.

Benchmarking incr-join: Collecting 100 samples in estimated 5.
incr-join               time:   [684.71 ns 687.90 ns 692.34 ns]
                        change: [-16.301% -15.580% -14.816%] (p = 0.00 < 0.05)
                        Performance has improved.

[...]

PR #1208, opened 2024-05-07
From.{join => joins} + remove Option layer + Remove RelOps::try_fold

Centril:
# Description of Changes

Some drive-by refactorings extracted from https://github.com/clockworklabs/SpacetimeDB/pull/1207 in the interest of smaller PRs.

- The field `From.join` is renamed to `joins` and its `Option` layer is removed which is redundant as `Vec` can represent absence as well.
- `RelOps::try_fold` is removed as it is dead code.

Best reviewed with whitespace changes ignored.

None

1


mamcx (approved):
LGTM

PR #1209, opened 2024-05-07
subscription benches: reorder footprint fields

Centril:
# Description of Changes

Reorder `footprint` table in subscription benchmarks to be more optimal wrt. layout such that there is less padding and so to-bsatn serialization uses fewer memcpy fields. This does not improve the performance of spacetimedb but represents better what spacetimedb is capable of when a table is optimized.

Extracted from https://github.com/clockworklabs/SpacetimeDB/pull/1207 in the interest of smaller PRs.

None

1


mamcx (approved):
LGTM

PR #1210, opened 2024-05-07
Remove some impls that were dead code

Centril:
# Description of Changes

- Remove `impl From<FieldName> for Box<ColumnOp>`
- Remove `impl From<AlgebraicValue> for Box<ColumnOp>`

Both are dead code.


mamcx (approved):
LGTM

PR #1211, opened 2024-05-08
CrudExpr::{Delete, Insert, Update}: only DbTables are possible, so cleanup InMem path

Centril:
# Description of Changes

- Commit 1 removes a bunch of methods that were unused.
- Commit 2 realizes that all of this `Table` business is in fact redundant and can be removed as `SourceExpr::InMem` is impossible and thus we can assume we have a `DbTable`.
- Commit 3 moves `CrudExpr::Update`'s `fn eval_query` logic to its own method, like all the other cases, for readability and consistency.
- Commit 4 uses `TxMode::unwrap_mut` that was introduced in commit 2 in 2 more cases.

None

1


mamcx (approved):
LGTM

PR #1212, opened 2024-05-09
`spacetime build` uses `--project-path` param

jdetter:
# Description of Changes

- `spacetime publish` and `spacetime generate` both use `--project-path <path>`, so `spacetime build` should use the same.

corresponding GitHub label.

Breaks the CLI params

1

# Testing

- [x] `spacetime build --project-path <path>` works as expected.


bfops (approved):
Looks good - let's see if it breaks something!

jdetter:
Our CI is happy and this is another step towards a more cohesive CLI 👍

PR #1214, opened 2024-05-13
Remove a few dead impls

Centril:
# Description of Changes

See diff. These are unused.


kim (approved):
If it ain't used it ain't used

PR #1218, opened 2024-05-13
Fix select * from *

### LINKED ISSUE #1199: `SELECT * FROM *` no longer allowed in subscriptions
Errors with `Unsupported query on subscription: "Multiple statements in subscription query"`.

Should also add a test to the SDK tests which exercises this.

coolreader18:
# Description of Changes

We should really remove this.

Fixes #1199


gefjon (approved):

jdetter:
@bfops please do a functional test on this with the C# SDK and make sure that it works properly

bfops (approved):
Worked for me in a one-off test

I also tested the unit test by adding it alone in https://github.com/clockworklabs/SpacetimeDB/pull/1223, and it fails appropriately.

PR #1225, opened 2024-05-14
Add CODEOWNERS for `LICENSE.txt`

bfops:
# Description of Changes

Adds a `CODEOWNERS` entry for all `LICENSE.txt` files.

This also makes sure that @cloutiertyler signs off on any **version bumps**, which update the root `LICENSE.txt`.

This means that @cloutiertyler is the signoff on every release.

No.

0

# Testing

- [x] See https://github.com/clockworklabs/SpacetimeDB/pull/1228


jdetter (approved):
Seems like a good protection 👍

cloutiertyler (approved):
LGTM

PR #1238, opened 2024-05-15
Add test for `select * from B join A on B.y = A.x`

Centril:
# Description of Changes

Adds a test for a join with different column names in the `ON` condition.


mamcx (approved):
LGTM

PR #1242, opened 2024-05-16
row_count field to table

Shubham8287:
# Description of Changes
Meta data field in `Table` struct for row count.

N/A

2

### DIFF
--- a/crates/core/src/db/relational_db.rs
+++ b/crates/core/src/db/relational_db.rs
@@ -1377,6 +1377,23 @@ mod tests {
     }
 
     #[test]
+    fn test_row_count() -> ResultTest<()> {
+        let stdb = TestDB::durable()?;
+
+        let mut tx = stdb.begin_mut_tx(IsolationLevel::Serializable);
+        let schema = my_table(AlgebraicType::I64);
+        let table_id = stdb.create_table(&mut tx, schema)?;
+        stdb.insert(&mut tx, table_id, product![1i64])?;
+        stdb.insert(&mut tx, table_id, product![2i64])?;
+        stdb.commit_tx(&ExecutionContext::default(), tx)?;
+
+        let stdb = stdb.reopen()?;
+        let tx = stdb.begin_tx();
+        assert_eq!(tx.get_row_count(table_id).unwrap(), 2);
+        Ok(())
+    }
+
+    #[test]
     fn test_unique() -> ResultTest<()> {
         let stdb = TestDB::durable()?;
 
--- a/crates/table/src/table.rs
+++ b/crates/table/src/table.rs
@@ -52,6 +52,8 @@ pub struct Table {
     /// depending on whether this is a tx scratchpad table
     /// or a committed table.
     squashed_offset: SquashedOffset,
+    /// Store number of rows present in table.
+    pub row_count: u64,
 }
 
 impl Table {
@@ -111,7 +113,7 @@ impl TableInner {
     }
 }
 
-static_assert_size!(Table, 240);
+static_assert_size!(Table, 248);
 
 /// Various error that can happen on table insertion.
 #[derive(Error, Debug)]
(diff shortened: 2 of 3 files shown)


crates/table/src/table.rs
```diff
@@ -197,6 +201,7 @@ impl Table {
 
         // Insert the row into the page manager.
         let (hash, ptr) = self.insert_internal(blob_store, row)?;
+        self.row_count += 1;
```
joshua-spacetime, line 204: Should this happen inside `insert_internal`?

crates/table/src/table.rs
```diff
@@ -439,7 +444,7 @@ impl Table {
 
         // SAFETY: We've checked above that `self.is_row_present(ptr)`.
         unsafe { self.delete_unchecked(blob_store, ptr) }
-
+        self.row_count -= 1;
```
joshua-spacetime, line 447: Similarly should the decrement happen inside `delete_internal`? I think replay might not go through this code path. It might call `delete_internal` directly.

crates/table/src/table.rs
```diff
@@ -33,6 +33,8 @@ use spacetimedb_sats::{
 use std::sync::Arc;
 use thiserror::Error;
 
+pub type RowCount = u64;
```
joshua-spacetime, line 36: nit: Probably don't need the type alias

crates/core/src/db/datastore/locking_tx_datastore/tx_state.rs
```diff
@@ -150,4 +150,16 @@ impl TxState {
         };
         Some((tbl, blob_store, delete_tables.entry(table_id).or_default()))
     }
+
+    /// `get_row_count` returns`i64` because its possible to have net negative row count in TxState
```
joshua-spacetime, line 154: FWIW, I don't think this is necessary yet since row counts are only used in read transactions.

crates/core/src/db/datastore/locking_tx_datastore/mut_tx.rs
```diff
@@ -692,6 +692,23 @@ impl MutTxId {
     pub fn rollback(self) {
         // TODO: Check that no sequences exceed their allocation after the rollback.
     }
+
+    pub fn get_row_count(&self, table_id: TableId) -> RowCount {
```
joshua-spacetime, line 696: Same here. I'm fine if you want to remove this for now and then add it back when we start using cardinality estimation for DML.

crates/table/src/table.rs
Shubham8287: you are right

gefjon:
Please add test(s) which exercise the following paths and assert the row count is correct:
- Inserts.
- Inserts followed by deletes.
- Insert of an already-present row, causing a set-semantic violation.
- `core::db::datastore::locking_tx_datastore::mut_tx::MutTx::delete_by_row`.
- Transaction replay.

Shubham8287:
> * `core::db::datastore::locking_tx_datastore::mut_tx::MutTx::delete_by_row`.

I didn't get what it means, considering `row_count` is only implemented for read tx. I may add test if you explain.

Added test for rest of the paths.

crates/core/src/db/relational_db.rs
```diff
@@ -1376,6 +1376,23 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn test_row_count() -> ResultTest<()> {
```
joshua-spacetime, line 1380: @Shubham8287 I think @gefjon is suggesting to add a duplicate insert and a delete in either this test or a separate one.

joshua-spacetime (approved):

crates/core/src/db/relational_db.rs
Shubham8287: I added them here already with other similar tests - https://github.com/clockworklabs/SpacetimeDB/pull/1242/files#diff-2e3e28b72d24c870151676e95eadbbcf99aa90580a9023cf85d3b6244be0ed23R1160

PR #1252, opened 2024-05-17
Atomically downgrade lock when committing tx to prevent deadlock

coolreader18:
# Description of Changes

Also, move the logic with committing and taking the subscription lock into module_subscription_actor. Best reviewed ignoring whitespace.

before/after:
diagram

### DIFF
--- a/crates/core/src/host/wasm_common/module_host_actor.rs
+++ b/crates/core/src/host/wasm_common/module_host_actor.rs
@@ -25,7 +25,7 @@ use crate::identity::Identity;
 use crate::messages::control_db::Database;
 use crate::module_host_context::ModuleCreationContext;
 use crate::sql;
-use crate::subscription::module_subscription_actor::ModuleSubscriptions;
+use crate::subscription::module_subscription_actor::{ModuleSubscriptions, WriteSkew};
 use crate::util::const_unwrap;
 use crate::util::prometheus_handle::HistogramExt;
 use crate::worker_metrics::WORKER_METRICS;
@@ -579,14 +579,8 @@ impl<T: WasmInstance> WasmModuleInstance<T> {
         }
         reducer_span.exit();
 
-        // Take a lock on our subscriptions now. Otherwise, we could have a race condition where we commit
-        // the tx, someone adds a subscription and receives this tx as an initial update, and then receives the
-        // update again when we broadcast_event.
-        let subscriptions = self.info.subscriptions.subscriptions.read();
         let status = match call_result {
             Err(err) => {
-                stdb.rollback_mut_tx(&ctx, tx);
-
                 T::log_traceback("reducer", reducer_name, &err);
 
                 WORKER_METRICS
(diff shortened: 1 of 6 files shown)


Shubham8287:
Ack it. Here is another visualisation
```
add_subcription:  begin_tx()                 subscription.write()
reducer1:      subscription.read()              begin_tx()
reducer2:                           begin_mut_tx()
```
So, the reducer 2 lock will not be acquired but It will block the queue (further reads) due to fairess.

crates/core/src/host/wasm_common/module_host_actor.rs
```diff
@@ -604,23 +598,13 @@ impl<T: WasmInstance> WasmModuleInstance<T> {
...
                 EventStatus::Failed(errmsg.into())
             }
-            Ok(Ok(())) => {
-                if let Some(tx_data) = stdb.commit_tx(&ctx, tx).unwrap() {
-                    EventStatus::Committed(DatabaseUpdate::from_writes(&tx_data))
-                } else {
-                    todo!("Write skew, you need to implement retries my man, T-dawg.");
-                }
-            }
+            Ok(Ok(())) => EventStatus::Committed(DatabaseUpdate::default()),
```
Shubham8287, line 605: nit: As commit is not happening here anymore, may be we can choose better name for`EventStatus::Committed` -> `EventStatus::YetToCommitted` or `EventStatus::ReducerSuccess`

crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -142,25 +144,35 @@ impl ModuleSubscriptions {
...
-        event: Arc<ModuleEvent>,
-    ) {
-        match event.status {
-            EventStatus::Committed(_) => {
-                tokio::task::block_in_place(|| self.broadcast_commit_event(subscriptions, event, client))
+        mut event: ModuleEvent,
+        ctx: &ExecutionContext,
+        tx: MutTx,
+    ) -> Result<Result<Arc<ModuleEvent>, WriteSkew>, DBError> {
+        let subscriptions = self.subscriptions.read();
```
Shubham8287, line 155:
```suggestion
       // Take a read lock on `subscriptions` before committing tx
       // else it can result in subscriber receiving duplicate updates.
        let subscriptions = self.subscriptions.read();
```
Will prefer to not still not drop this commment.

crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -170,34 +182,13 @@ impl ModuleSubscriptions {
...
-        subscriptions: &SubscriptionManager,
-        event: Arc<ModuleEvent>,
-        sender_client: Option<&ClientConnectionSender>,
-    ) {
-        subscriptions.eval_updates(&self.relational_db, event, sender_client)
+        Ok(Ok(event))
     }
 }
 
+pub struct WriteSkew;
```
Shubham8287, line 190: nit: Maybe better name is `WriteConflict` because by defination `WriteSkew` anomoly is something transaction doesn't really detect while commiting unless transaction is tracking its reads.

Shubham8287 (approved):

PR #1253, opened 2024-05-18
More trace logs for TX locking

bfops:
# Description of Changes

Some small changes to make the trace logs more consistent for tx locks.

Nope

1

# Testing

This is a holdover from a debugging session. It was tested-by-fire.


crates/core/src/db/relational_db.rs
```diff
@@ -440,9 +444,9 @@ impl RelationalDB {
     where
         F: FnOnce(&mut Tx) -> T,
     {
-        let mut tx = self.inner.begin_tx();
+        let mut tx = self.begin_tx();
         let res = f(&mut tx);
-        self.inner.release_tx(ctx, tx);
+        self.release_tx(ctx, tx);
```
bfops, line 449: these are so we get the log messages consistently

crates/core/src/subscription/module_subscription_manager.rs
```diff
@@ -125,7 +125,7 @@ impl SubscriptionManager {
         let tables = &event.status.database_update().unwrap().tables;
         let slow = db.read_config().slow_query;
         let tx = scopeguard::guard(db.begin_tx(), |tx| {
-            tx.release(&ExecutionContext::incremental_update(db.address(), slow));
+            db.release_tx(&ExecutionContext::incremental_update(db.address(), slow), tx);
```
bfops, line 128: so we get the log messages consistently

crates/core/src/db/relational_db.rs
```diff
@@ -300,7 +304,7 @@ impl RelationalDB {
 
     #[tracing::instrument(skip_all)]
     pub fn release_tx(&self, ctx: &ExecutionContext, tx: Tx) {
-        log::trace!("ROLLBACK TX");
+        log::trace!("RELEASE TX");
```
bfops, line 307: I think this was a typo?

crates/core/src/db/relational_db.rs
```diff
@@ -283,13 +283,17 @@ impl RelationalDB {
...
+        log::trace!("ACQUIRED MUT TX");
+        r
     }
 
     #[tracing::instrument(skip_all)]
     pub fn begin_tx(&self) -> Tx {
         log::trace!("BEGIN TX");
-        self.inner.begin_tx()
+        let r = self.inner.begin_tx();
+        log::trace!("ACQUIRED TX");
```
bfops, line 295: when we were debugging lock issues, we appreciated not only having a "trying to acquire lock" message but also a "I am now holding the lock" message.

mamcx (approved):
LGTM

PR #1263, opened 2024-05-20
Misc datastore cleanups & less cloning

Centril:
# Description of Changes

1. Privatize `st_*_schema` functions, only used in system_tables() now.
2. Only call `system_tables()` once in `bootstrap_system_tables`. Eagerly `Arc` those and then clone as little as possible.
3. Take `TableId` by value more
4. Rename `table_exists` -> `table_name`
5. Dedup `table_name` by using `get_schema()`.
6. Other minor misc deduping.

None

2, touches some system table code but in a fairly small way.
Still, be sure to double check `ST_TABLES_IDX` and the `bootstrap_system_tables` changes.

### DIFF
--- a/crates/core/src/db/datastore/locking_tx_datastore/committed_state.rs
+++ b/crates/core/src/db/datastore/locking_tx_datastore/committed_state.rs
@@ -9,11 +9,11 @@ use crate::{
     db::{
         datastore::{
             system_tables::{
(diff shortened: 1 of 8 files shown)


Shubham8287 (approved):
I extensively checked for system_table and related schema changes, Looks fine to me.

PR #1274, opened 2024-05-21
`execution_set.check_auth(...)` on initial subscription

Centril:
# Description of Changes

We simply didn't check auth for initial subscriptions, and this PR fixes that.

Technically this breaks the API in the sense that we check auth now, but that is a bug fix...

1

# Testing

A test is added that ensures that we check auth.

### DIFF
--- a/crates/core/src/db/relational_db.rs
+++ b/crates/core/src/db/relational_db.rs
@@ -540,11 +540,12 @@ impl RelationalDB {
             .collect()
     }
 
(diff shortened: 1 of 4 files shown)


joshua-spacetime:
> Also simplifies fn sources(), using internal iteration rather than external, avoiding some allocation when checking auth.

Bug fixes and refactors really should be separate.

mamcx (approved):
LGTM

Issue #1256, opened 2024-05-20
Implement consistent filtering rules proposal


cloutiertyler:
Progress:

- [x] Rust client-side: https://github.com/clockworklabs/SpacetimeDB/pull/1275
- [x] C# client-side: https://github.com/clockworklabs/SpacetimeDB/pull/1277 + https://github.com/clockworklabs/spacetimedb-csharp-sdk/pull/93
- [x] TypeScript client-side: https://github.com/clockworklabs/SpacetimeDB/pull/1276
- [x] Rust server-side: https://github.com/clockworklabs/SpacetimeDB/pull/1280
- [x] C# server-side: https://github.com/clockworklabs/SpacetimeDB/pull/1282

RReverser:
Progress:

- Rust client-side: https://github.com/clockworklabs/SpacetimeDB/pull/1275
- C# client-side: https://github.com/clockworklabs/SpacetimeDB/pull/1277 + https://github.com/clockworklabs/spacetimedb-csharp-sdk/pull/93
- TypeScript client-side: https://github.com/clockworklabs/SpacetimeDB/pull/1276
- Rust server-side: TODO
- C# server-side: TODO

PR #1275, opened 2024-05-22
Consistent filtering in Rust client + minor fixes

RReverser:
# Description of Changes

- Extend codegen tests to Rust.
- Fix found issue with non-deterministic order of imports in generated code caused by usage of `HashSet`.
- Replace `cursive-chat`'s `module_bindings` folder with a symlink to `quickstart-chat`'s one (as they're supposed to be the same, and this makes diffs smaller).
- Implement Rust client part of #1256.
- Regenerated tests.

Consistent filtering rules proposal limits types on which filtering can be done, as well as changes `filter_by` functions to always return an opaque iterator instead of iterator or option as it did before.

1

# Testing

- [x] Ran codegen tests.
- [x] Regenerated all `module_bindings` in the repo manually.
- [x] Reran SDK tests that depend on `module_bindings`.


kazimuth (approved):
Looks good, more consistent. I love reviewing these codegen changes by just looking at changes to the generated code lol.

RReverser:
> I love reviewing these codegen changes by just looking at changes to the generated code lol.

Heh doesn't help that in this case we didn't have baseline codegen snapshot before those changes - that's why I kept the commit introducing that baseline separate.

PR #1277, opened 2024-05-22
Update C# codegen to consistent filtering rules

RReverser:
# Description of Changes

- Limit types as per proposal.
- Add `Query` client-side SDK helper for API parity with server-side modules (on client-side it's a simple wrapper around Iter + Where).
- Change return type of `FilterBy` to always be iterable, with new `FindBy` function for unique fields.
- Simplify the way primary keys are handled - must go with https://github.com/clockworklabs/spacetimedb-csharp-sdk/pull/93 for the client SDK counterpart.

Limits equatable types and changes return type of `FilterBy` functions as per proposal.

2

# Testing

- [x] Built & ran the client-side chat example in C# SDK.
- [x] Ran codegen tests and updated the snapshot.


kazimuth (approved):
👍

crates/cli/src/subcommands/generate/csharp.rs
```diff
@@ -832,25 +827,30 @@ fn autogen_csharp_access_funcs_for_struct(
...
     });
+    writeln!(output);
+
+    // Simple alias for Iter().Where(...) for API parity with C# server-side modules.
+    writeln!(
+        output,
+        "public static IEnumerable<{struct_name_pascal_case}> Query(Func<{struct_name_pascal_case}, bool> filter)"
+    );
+    indented_block(output, |output| {
+        writeln!(output, "return Iter().Where(filter);",);
```
jdetter, line 845: You need to `include System.Linq;` at the top of the file otherwise the automatically generated files no longer compile.

jdetter (changes requested):
Small issue

crates/cli/src/subcommands/generate/csharp.rs
RReverser: Hm good point, but I'm surprised CI on https://github.com/clockworklabs/spacetimedb-csharp-sdk/pull/93 didn't catch it, as it does build the example...

RReverser: A-ha, it's because the sample project uses `<ImplicitUsings>enable</ImplicitUsings>` in its `csproj` so C# adds `System.Linq` among other usings automatically. I'm going to change that to make those issues easier to catch.

jdetter (approved):
Import issue fixed, thanks 👍

PR #1278, opened 2024-05-22
Tables only become public explicitly via `#[spacetimedb(table(public))]`

### LINKED ISSUE #1268: Rust - All tables must be private by default and should not use an underscore
In Rust:

```rust
#[spacetimedb(table, public)]
#[spacetimedb(table, private)] // equivalent to below
#[spacetimedb(table)]
```

Centril:
# Description of Changes

Implements and fixes #1268.

- `#[spacetimedb(table)]` means `#[spacetimedb(table(private))]`
- If a public table is desired, you can use `#[spacetimedb(table(public))]`
- This is implemented by translating to `#[sats(public)]` which then results in `<Table as TableType>::TABLE_ACCESS` becoming public rather than private.

This is a major breaking change as previously public tables now become private, so users will need to add annotations to recover old semantics.

3, the main risk is the breaking change.

# Testing

Static assertions are added to `rust-wasm-test´.


RReverser:
Not related to implementation, rather the issue, but the syntax seems somewhat unidiomatic to me.

Given that it's a meta parameter to the proc-macro for tables, shouldn't it be `#[spacetimedb(table(public))]` instead?

That would also simplify parsing somewhat.

Alternatively, any reason we can't use struct visibility as an indicator of table visibility?

So `pub` structs become public tables, a d private structs become private tables - that seems nice as it would behave like visibility in any other "export" context in Rust.

RReverser:
> I would however want to keep the `table, public` syntax for now, and then revisit all of the macro syntax holistically, e.g., using (not set in stone...):

Yeah this would be good. If they become truly independent markers, then comma-separate syntax as "shorthand" makes more sense.

What irks me in current syntax is that - IIUC the implementation correctly - those markers appear independent rather than hierarchical, but if you try to change order to `#[spacetimedb(public, table)]`, that won't work, which can be rather surprising to users.

Same for not being able to write `#[spacetimedb(table)] #[spacetimedb(public)]` which a user would probably also expect to work given the precedent from other attribute proc-macros in the ecosystem.

> I think I might partner with @coolreader18 to write up a proposal at some point before 1.0.

Feel free to tag me as well, would be happy to at least review. 

> I think this risks conflating encapsulation within the codebase vs. information flow control (IFC).

Isn't that sort of the idea of SpacetimeDB modules, to make those 1:1? If you can define tables themselves using Rust syntax, why not their privacy?

Especially given that we build modules as Rust libraries, it doesn't seem like a long stretch for same exporting rules to apply - in libraries if something is public at the top level, then it's exported to external users as well, so in STDB modules we could also say if something is public at the top level, it's similarly exported.

Admittedly, I can see how separate attribute would reduce risk of accidentally exposing tables even further, it just seems so much more elegant to apply Rust-esque exporting/visibility rules 😅

gefjon:
We can't conflate in-language `pub` with table visibility to clients; that breaks the way BitCraft structures their code. Developers must be able to have a table which is defined in one Rust `mod` but accessed from another in the same WASM module (i.e. is `pub`), but is not visible to clients (i.e. is not `public`).

mamcx:
> Isn't that sort of the idea of SpacetimeDB modules, to make those 1:1? If you can define tables themselves using Rust syntax, why not their privacy?
 
In a lot of cases, it is good to think "But what about SQL?"

Tables live in the database, not the module. It could (eventually) be executed `create table private ...` and now we see Rust is not the authoritative source of truth, is the system tables.

mamcx:
> What irks me in current syntax is that - IIUC the implementation correctly - those markers appear independent rather than hierarchical, but if you try to change order to `#[spacetimedb(public, table)]`, that won't work, which can be rather surprising to users.

Now seeing this I'm inclined to say is better to implement now `#[spacetimedb(table(public))]`.

coolreader18:
There's a lot of things in the current syntax that are like that Mario - this would be in line with how it is currently. I agree w Mazdak that we should revisit this holistically.

gefjon:
I mean, I'm pretty sure I just made up the `spacetimedb(table, public)` syntax on the spot during our meeting and everyone went with it; it's not like it was a well-considered design. I agree that `spacetimedb(table(public))` is better, and we might as well do it right the first time.

coolreader18 (approved):

coolreader18:
@Centril you can run `cargo insta test -p spacetimedb --accept` to update the snap tests

crates/bindings-macro/src/lib.rs
```diff
@@ -75,7 +78,7 @@ mod sym {
 /// The macro takes this `input`, which defines what the attribute does,
 /// and it is structured roughly like so:
 /// ```ignore
-/// input = table | init | connect | disconnect | migrate
+/// input = table [, private | public] | init | connect | disconnect | migrate
```
RReverser, line 81: Doc needs update.

crates/bindings-macro/src/module.rs
```diff
@@ -115,9 +123,18 @@ pub(crate) fn extract_sats_type<'a>(
         krate,
         original_attrs: attrs,
         data,
+        public,
     })
 }
 
+/// Ensures that the `public` modifier
```
RReverser, line 130: Doc comment seems incomplete?

PR #1279, opened 2024-05-22
Phoebe/snapshot/naive capture

gefjon:
# Description of Changes

Implements snapshotting, as per the snapshotting proposal (not currently available to the public).

corresponding GitHub label.

# Testing


gefjon:
I have now successfully restored from a snapshot, and grabbed a preliminary timing for it.

This is once again with the modified `quickstart-chat`, and the snapshot in question is 5.2 MiB on-disk:

```sh
du -hs ~/.spacetime/worker_node/database_instances/93dda09db9a56d8fa6c024d843e805d8/2/database/snapshots/00000000000000028672.snapshot_dir
5.2M	/home/phoebe/.spacetime/worker_node/database_instances/93dda09db9a56d8fa6c024d843e805d8/2/database/snapshots/00000000000000028672.snapshot_dir
```

There are three interesting time spans to restore from a snapshot:
- The time to read the snapshot from disk into memory, incl. parsing `Page`s from their serialized form and verifying a bunch of hashes.
- The time to install the already-in-memory snapshot into a `Locking` datastore.
- The total time from zero to complete, including both of the above plus replaying the suffix of the log.

In the tested case, these were 4.3 ms, 93 us and 10.7 ms, respectively.

Note that:
- This is a very small database, so obviously the times will be small.
- The value I have chosen for `SNAPSHOT_FREQUENCY`, `0x1000`, is much smaller than the one we expect to use, 1 million, so the suffix of the log will be on average ~100x as many transactions.
- Each individual transaction in this database is tiny, so we will see a faster rate of log replay in tx/s than we would expect from a more interesting module.
- I have a pretty fast SSD.

As a result, there's not a whole to glean from this, except that these numbers are encouraging and we aren't obviously doing something horribly wrong that would cause us to rethink our design.

gefjon:
Separate PRs to split this into once it's ready for review:

- [ ] Fix bug in `commitlog` to make it respect requested `offset`.
- [ ] Move `Lockfile` to new crate `fs-utils`.
- [ ] Add `DirTrie` to `fs-utils`.
- [ ] Implement ser/de for `Page` and friends.
- [ ] Add `snapshot` crate without integration.
- [ ] Integrate `snapshot` crate into `core`.

gefjon:
Closing in favor of #1344 .

PR #1283, opened 2024-05-22
Implement new rand api

coolreader18:
# Description of Changes

Implements the improved rng interface as proposed by @gefjon.

Only an API break if people were using the `getrandom` feature.

2

I'm probably gonna add a smoketest for this.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -139,7 +139,7 @@ fs-err = "2.9.0"
 fs2 = "0.4.3"
 futures = "0.3"
 futures-channel = "0.3"
-getrandom = { version = "0.2.7", features = ["custom"] }
+getrandom = "0.2.7"
 glob = "0.3.1"
 hashbrown = { version = "0.14", features = ["rayon", "serde"] }
 headers = "0.4"
--- a/crates/bindings/src/rng.rs
+++ b/crates/bindings/src/rng.rs
@@ -1,0 +1,195 @@
+use std::cell::{OnceCell, UnsafeCell};
+use std::fmt;
+use std::marker::PhantomData;
+use std::sync::atomic::{AtomicU64, Ordering::Relaxed};
+
+use rand::distributions::{Distribution, Standard};
+use rand::rngs::StdRng;
+use rand::{RngCore, SeedableRng};
+
+use crate::Timestamp;
+
+scoped_tls::scoped_thread_local! {
+    static RNG: OnceCell<RngContext>
+}
+
+// this ensures that you can't stash an StdbRng somewhere and use it
+// at a later point, breaking atomic reproducibility of transactions.
+static RNG_GENERATION: AtomicU64 = AtomicU64::new(0);
+
+struct RngContext {
+    rng: StdCellRng,
+    generation: u64,
+}
+
+impl RngContext {
+    fn seed() -> Self {
+        Self {
+            rng: StdCellRng::new(StdRng::seed_from_u64(Timestamp::now().micros_since_epoch)),
+            generation: RNG_GENERATION.fetch_add(1, Relaxed),
+        }
+    }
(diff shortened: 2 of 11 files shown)


crates/bindings/src/rng.rs
```diff
@@ -0,0 +1,195 @@
...
+/// A reference to the random number generator for this reducer call.
+///
+/// An instance can be obtained via [`spacetimedb::rng()`][rng()]. Import
+/// [`rand::Rng`] to get access to useful methods
+///
+/// This handle
+/// can only be used in the context of this reducer call; it cannot be
+/// stashed in a global and used in a later reducer call, as that would break
+/// the atomicity of transactions.
+///
```
gefjon, line 85: Again, I dispute this point, since the handle doesn't contain any information.

Users shouldn't be stashing anything in globals, but `StdbRng` is not worse than any other type in that regard, and I don't think we should treat it specially.

crates/bindings/src/rng.rs
```diff
@@ -0,0 +1,195 @@
...
+/// `StdbRng` uses the same PRNG as `rand`'s [`StdRng`], however, because it
+/// is seeded from a timestamp, it is not cryptographically secure.
+///
+/// The type of reproducibility you're looking for be finer grained than
+/// "if it happens at the exact same time, you get the same result" -- if
+/// so, just seed an [`StdRng`] directly, or use another rng like those
+/// listed [here](https://rust-random.github.io/book/guide-rngs.html).
+/// Just note that you should never depend on state from outside of
+/// the current reducer call (by e.g. storing an rng in a global variable),
+/// as that can and will break things in the database.
```
gefjon, line 95:
```suggestion
/// Note that you must not store any state,
/// including an Rng,
/// in a global variable or any other in-WASM side channel.
/// All state persisted across reducer calls must be stored in the database.
```

Cargo.toml
```diff
@@ -139,7 +139,7 @@ fs-err = "2.9.0"
 fs2 = "0.4.3"
 futures = "0.3"
 futures-channel = "0.3"
-getrandom = { version = "0.2.7", features = ["custom"] }
+getrandom = "0.2.7"
```
gefjon, line 142: Why do we still depend on `getrandom`?

crates/bindings/src/rng.rs
```diff
@@ -0,0 +1,195 @@
...
+
+use crate::Timestamp;
+
+scoped_tls::scoped_thread_local! {
+    static RNG: OnceCell<RngContext>
+}
+
+// this ensures that you can't stash an StdbRng somewhere and use it
+// at a later point, breaking atomic reproducibility of transactions.
+static RNG_GENERATION: AtomicU64 = AtomicU64::new(0);
```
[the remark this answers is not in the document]
coolreader18: hm. I was imagining a ReseedingRng wrapping StdbRng, but I guess you're right that the issue there would be storing the rng in a global, and not the use of StdbRng. I guess I was imagining that if you were depending on the rng producing a certain reproducible stream of entropy, that wouldn't hold across reducers, but yeah you shouldn't be doing that anyway.

Cargo.toml
coolreader18: See comment in `bindings/Cargo.toml` - to detect it if someone tries to use it. Also, we can't depend on the `std` feature of rand without pulling in getrandom, and getrandom will yell at you if you try to compile for wasm without enabling either `js` or `custom`. And if we did forgo `std`, every crate in a module's dependency tree would have to as well - they'd have to do `rand = { version  = "..", default-features = false }` which feels like some unnecessary friction, especially with the unhelpful advice from the `getrandom` crate telling you to enable the `js` feature.

gefjon (approved):

PR #1284, opened 2024-05-22
Centralize system tables with `system_table!` macro

Centril:
# Description of Changes

1. Refactor system tables under a single `system_table!` macro which stores the table id, index, and name in consts together with the columns which remain variants. The schema is also now part of e.g., `StTable` as a function rather than the logic being spread out. Column positions in the `schema()` are also verified at compile time against the manually provided column positions. We've had bugs due to inconsistent ordering in the past, so its good to check things.
2. `IdNotFound` is split into more better and more relevant error variants.
3. Add `table_name_or_err` to refactor some error handling.

None

3, there's some risk in that I mixed up some table for some other or a col for some other, but I used find and replace so the risk should be small. Nonetheless, please look at things with care.

### DIFF
--- a/crates/core/src/db/datastore/locking_tx_datastore/state_view.rs
+++ b/crates/core/src/db/datastore/locking_tx_datastore/state_view.rs
@@ -171,7 +171,7 @@ pub(crate) trait StateView {
 pub struct Iter<'a> {
     ctx: &'a ExecutionContext,
     table_id: TableId,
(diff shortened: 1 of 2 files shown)


cloutiertyler:
I am closing this PR as it serves only to cause churn in the codebase. The best case scenario is that this PR does nothing and the worst case is that we introduce a subtle bug. This is not remotely high priority at the moment.

PR #1286, opened 2024-05-23
`merge_apply_inserts/deletes`: do metrics work once per table, not per row

Centril:
# Description of Changes

Rather than doing metrics work once per row in `merge_apply_inserts/deletes`, just take the per-table lengths and update metrics once per table.

None

1


crates/core/src/db/datastore/locking_tx_datastore/committed_state.rs
```diff
@@ -399,21 +399,25 @@ impl CommittedState {
                     let pv = table
                         .delete(blob_store, row_ptr, |row| row.to_product_value())
                         .expect("Delete for non-existent row!");
```
Shubham8287, line 401: If this panics, then metrics count in prometheus will permanently become incosistent with table but then we will also have other bigger troubles?

Shubham8287 (approved):

PR #1288, opened 2024-05-23
Shub/st connected clients

Shubham8287:
# Description of Changes

corresponding GitHub label.

# Testing


crates/core/src/db/datastore/system_tables.rs
```diff
@@ -78,7 +83,8 @@ pub(crate) const ST_COLUMNS_IDX: usize = 1;
 pub(crate) const ST_INDEXES_IDX: usize = 2;
 pub(crate) const ST_CONSTRAINTS_IDX: usize = 3;
 pub(crate) const ST_MODULE_IDX: usize = 4;
-pub(crate) const ST_SEQUENCES_IDX: usize = 5;
+pub(crate) const ST_SEQUENCES_IDX: usize = 6;
```
Shubham8287, line 86: this is breaking

crates/core/src/host/module_host.rs
```diff
@@ -668,45 +671,98 @@ impl ModuleHost {
...
+                        InvalidReducerArguments {
+                            err: err.into(),
+                            reducer: reducer_name.into(),
+                        }
+                        .into()
+                    }),
+                e => Err(e),
+            });
+
+        // Deleting client from `st_clients`does not depend upon result of disconnect reducer hence done in a separate tx.
```
kim, line 723: This will create two `__identity_disconnected__` entries in the commitlog in the case disconnect is defined and succeeds.

That's fine for the purpose of the auto-disconnect mechanism, but we shouldn't be afraid of over-documenting it. E.g. offline analysis like @lcodes is doing might find it surprising.

Maybe just add commentary to the enclosing method explaining what is going on.

kim: The behavior is correct, though, without support for nested transactions.

kim (approved):
Looks good!

Modulo docs + system schema tests.

crates/core/src/host/module_host.rs
```diff
@@ -668,45 +671,98 @@ impl ModuleHost {
...
-
-                Ok(())
-            }
-            e => Err(e),
-        })
+        let db = &self.inner.dbic().relational_db;
+        let ctx = &ExecutionContext::reducer(
+            db.address(),
+            ReducerContext {
+                name: reducer_name.to_owned(),
```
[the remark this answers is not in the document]
kim: Wait… if the context is not consumed, that’d be a bug. There must be a transaction in the log in any case.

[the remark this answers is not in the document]
Shubham8287: @kim, if reducers exist, they create `ctx` in `module_host_actor` and not using this one.

crates/core/src/host/module_host.rs
Shubham8287: Agreed, makes sense.

PR #1289, opened 2024-05-23
CI workflows support `workflow_dispatch` event

bfops:
# Description of Changes

Add `workflow_dispatch` event support to `ci.yml`.

This allows us to trigger workflow runs manually (see [benchmarks](https://github.com/clockworklabs/SpacetimeDB/actions/workflows/benchmarks.yml)).

No

2

# Testing

- [x] We're at least still running the _normal_ events properly https://github.com/clockworklabs/SpacetimeDB/actions/runs/9213356867/job/25347061394
- [ ] Can't test the manual `workflow_dispatch` step until it's merged

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ jobs:
     steps:
       - name: Find Git ref
         run: |
-          PR_NUMBER="${{ github.event.inputs.pr_number || github.event.issue.number || null }}"
+          PR_NUMBER="${{ github.event.inputs.pr_number || null }}"
           if test -n "${PR_NUMBER}"; then
             GIT_REF="$( gh pr view $PR_NUMBER --json headRefName --jq .headRefName )"
           else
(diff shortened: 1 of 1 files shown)


kazimuth (approved):
Makes sense to me, 👍

kurtismullins (approved):
Nice! Just one comment: The Pull Request CI is usually a combination of a given branch merged with master to make sure that it will behave as expected when merged into master. This appears to fetch just the branch. It might not make as much sense to use "Pull Request Number" as it would to maybe use `commit` or `branch_name` for example.

That is just a nitpick -- otherwise it looks good to me!

bfops:
> Nice! Just one comment: The Pull Request CI is usually a combination of a given branch merged with master to make sure that it will behave as expected when merged into master. This appears to fetch just the branch. It might not make as much sense to use "Pull Request Number" as it would to maybe use `commit` or `branch_name` for example.
> 
> That is just a nitpick -- otherwise it looks good to me!

That's a good point. I can also see an argument for keeping inputs the same between workflows.

I'm going to merge as-is, but I promise not to take it personally if someone else cares enough to changes it!

bfops:
I have successfully triggered a CI workflow from the `Actions` page!

PR #1291, opened 2024-05-23
Let ProgramStorage::external be async

coolreader18:
# Description of Changes

Also: remove sled object_db. Merging that in from #1245.

1


crates/core/src/host/disk_storage.rs
```diff
@@ -0,0 +1,88 @@
...
+
+        match fs::File::options().write(true).create_new(true).open(&tmp).await {
+            Ok(mut f) => {
+                f.write_all(value).await?;
+                f.sync_data().await?;
+                drop(f);
+
+                fs::rename(tmp, path).await?;
+            }
+            Err(e) if e.kind() == io::ErrorKind::AlreadyExists => {
```
kim, line 68: Is this even possible?

crates/core/src/host/host_controller.rs
```diff
@@ -46,37 +48,57 @@ type HostCell = Arc<AsyncRwLock<Option<Host>>>;
 /// The registry of all running hosts.
 type Hosts = Arc<Mutex<IntMap<u64, HostCell>>>;
 
-type ExternalStorage = dyn Fn(&Hash) -> anyhow::Result<Option<AnyBytes>> + Send + Sync + 'static;
-type SameDbStorage = dyn Fn(&RelationalDB, &Hash) -> anyhow::Result<Option<AnyBytes>> + Send + Sync + 'static;
+#[async_trait]
```
kim, line 51: Do we need `async_trait` still?

crates/core/src/host/host_controller.rs
```diff
@@ -46,37 +48,57 @@ type HostCell = Arc<AsyncRwLock<Option<Host>>>;
...
+        F: Fn(&RelationalDB, Hash) -> anyhow::Result<Option<AnyBytes>> + Send + Sync + 'static,
     {
         Self::SameDb(Arc::new(f))
     }
+
+    async fn load_program(&self, db: &RelationalDB, program_hash: Hash) -> anyhow::Result<AnyBytes> {
+        debug!("lookup program {}", program_hash);
+        match self {
+            ProgramStorage::External(external) => external.lookup(program_hash).await,
+            ProgramStorage::SameDb(f) => f(db, program_hash),
```
kim, line 98: Wondering if this should be spawn blocking, just in case the same thread has a tx open…

crates/core/src/host/host_controller.rs
```diff
@@ -89,7 +111,7 @@ pub struct HostController {
     /// The directory to create database instances in.
     ///
     /// For example:
-    ///
+    ///>
```
kim, line 114: What does this do?

crates/core/src/object_db.rs
```diff
@@ -1,34 +0,0 @@
-use crate::{
```
kim, line 1: 🥳

kim (approved):

crates/core/src/host/host_controller.rs
coolreader18: Yes - it lets the future borrow from the ExternalStorage, rather than having to be `'static`

crates/core/src/host/host_controller.rs
coolreader18: oops, nothing

crates/core/src/host/host_controller.rs
coolreader18: Hmm, probably - but we can cross that when we get to actually using it.

PR #1292, opened 2024-05-23
Release/candidate/v0.9.2

bfops:
# Description of Changes

corresponding GitHub label.

# Testing

### DIFF
--- a/crates/bindings-csharp/Runtime/Runtime.csproj
+++ b/crates/bindings-csharp/Runtime/Runtime.csproj
@@ -5,7 +5,7 @@
     <LangVersion>latest</LangVersion>
     <Nullable>enable</Nullable>
     <AssemblyName>SpacetimeDB.Runtime</AssemblyName>
-    <AssemblyVersion>0.9.0</AssemblyVersion>
+    <AssemblyVersion>0.9.2</AssemblyVersion>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <EmitCompilerGeneratedFiles>true</EmitCompilerGeneratedFiles>
   </PropertyGroup>
--- a/crates/bindings-csharp/Runtime/Runtime.csproj
+++ b/crates/bindings-csharp/Runtime/Runtime.csproj
@@ -5,7 +5,7 @@
     <LangVersion>latest</LangVersion>
     <Nullable>enable</Nullable>
     <AssemblyName>SpacetimeDB.Runtime</AssemblyName>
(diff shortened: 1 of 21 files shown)


crates/bindings-csharp/Runtime/Runtime.csproj
```diff
@@ -5,7 +5,7 @@
     <LangVersion>latest</LangVersion>
     <Nullable>enable</Nullable>
     <AssemblyName>SpacetimeDB.Runtime</AssemblyName>
-    <AssemblyVersion>0.8.2</AssemblyVersion>
+    <AssemblyVersion>0.9.0</AssemblyVersion>
```
kurtismullins, line 8: @bfops I wonder if these should be 0.9.2 now?

@RReverser had used `0.9.0` for the `0.9.1` release but *I believe* that was only because they appeared skipped for the actual `0.9.0` release.

tools/publish-crates.sh
```diff
@@ -37,7 +37,7 @@ for crate in "${CRATES[@]}" ; do
 	if [ $DRY_RUN == 1 ] ; then
 		cargo publish --dry-run
 	else
-		cargo publish
+		cargo publish --allow-dirty
```
kurtismullins, line 40: Oops! I should not have left this line change in. Although we will probably technically need it for every single release unless someone removes this line permanently or we find another way to work around the error you get when attempting to push it to crates.io.

```rust
#![doc = include_str!("../../../../docs/incremental-joins.md")]
```
**Source:** https://github.com/clockworklabs/SpacetimeDB/blob/master/crates/core/src/subscription/subscription.rs#L24

**Error Message:**
```
error: couldn't read src/subscription/../../../../docs/incremental-joins.md: No such file or directory (os error 2)
  --> src/subscription/subscription.rs:24:10
   |
24 | #![doc = include_str!("../../../../docs/incremental-joins.md")]
   |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: this error originates in the macro `include_str` (in Nightly builds, run with -Z macro-backtrace for more info)
   ```

Thoughts?

crates/bindings-csharp/Runtime/Runtime.csproj
RReverser: As discussed in chat, we only bump one up rather than keeping those in sync with STDB itself. So it's enough to bump to 0.9.1 (or just not bump at all, as we haven't had any changes since, up to you). Definitely not 0.9.2.

PR #1302, opened 2024-05-28
CLI - Fix `Lockfile` error message format strings

bfops:
# Description of Changes

Previous error text:
```
thread 'main' panicked at crates/cli/src/config.rs:911:58:
called `Result::unwrap()` on an `Err` value: Unable to acquire lock on config file {config_path:?}: failed to create lockfile {path:?}

Caused by:
    File exists (os error 17)
```

New error message:
```
thread 'main' panicked at crates/cli/src/config.rs:911:58:
called `Result::unwrap()` on an `Err` value: Unable to acquire lock on config file "/home/lead/.spacetime/config.toml": failed to create lockfile "/home/lead/.spacetime/config.lock"

Caused by:
    File exists (os error 17)
```

Nope

1

# Testing

- [x] Start `spacetime logs -f`, then try to run `spacetime call`, observe error

### DIFF
--- a/crates/cli/src/config.rs
+++ b/crates/cli/src/config.rs
@@ -149,7 +149,7 @@ impl Lockfile {
         path.set_extension("lock");
         // Open with `create_new`, which fails if the file already exists.
         std::fs::File::create_new(&path).with_context(|| {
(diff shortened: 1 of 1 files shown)


gefjon (approved):
:facepalm:

PR #1304, opened 2024-05-28
Refactor `check_auth` + `(walk_)sources` using internal iteration in the latter

Centril:
# Description of Changes

- Simplify `fn sources(..)`, now `fn walk_sources(...)`, using internal iteration rather than external, avoiding some allocation when checking auth.
- Simplify some `AuthAccess` impls using the above.

None

1


Shubham8287 (approved):
Looks good!

mamcx (approved):
LGTM

PR #1309, opened 2024-05-28
Fix FilterBy regression in C#

RReverser:
# Description of Changes

Fixes regression accidentally introduced in #1277: if FindBy returns null, FilterBy will return an iterable with a single null item instead of an empty iterable.

corresponding GitHub label.

# Testing

### DIFF
--- a/crates/cli/tests/snapshots/codegen__codegen_csharp.snap
+++ b/crates/cli/tests/snapshots/codegen__codegen_csharp.snap
@@ -374,7 +374,10 @@ namespace SpacetimeDB
 
 		public static IEnumerable<PkMultiIdentity> FilterById(uint value)
 		{
-			return new[] { FindById(value) };
+			if (FindById(value) is {} found)
+			{
+				yield return found;
+			}
 		}
 
 		public static PkMultiIdentity FindByOther(uint value)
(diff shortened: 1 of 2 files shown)


crates/cli/tests/snapshots/codegen__codegen_csharp.snap
```diff
@@ -374,7 +374,10 @@ namespace SpacetimeDB
 
 		public static IEnumerable<PkMultiIdentity> FilterById(uint value)
 		{
-			return new[] { FindById(value) };
+			if (FindById(value) is {} found)
```
kazimuth, line 377: What a funny idiom

kazimuth (approved):
Ah, good catch.

It's too bad we can't use nullity annotations to mark that `FindBy` returns a nullable object. Unity...

crates/cli/tests/snapshots/codegen__codegen_csharp.snap
RReverser: I know, right 😅

RReverser:
> It's too bad we can't use nullity annotations to mark that `FindBy` returns a nullable object. Unity...

Oh, we can - that's continued in https://github.com/clockworklabs/SpacetimeDB/pull/1312/files#diff-7e5af7fb5d724ff748e59bc91da14a16e2132302343bb4ac823c1829ed25649eL886-R433.

kazimuth:
Blah, right, I keep mixing up modules and client, sorry. Nice!

RReverser:
Oh that one is for client. Unity doesn't mind nullability.

PR #1311, opened 2024-05-29
Bump ABI version for C# AOT

RReverser:
# Description of Changes

Looks like these constants got missed during upgrade, making AOT version ABI-incompatible.

Unfortunately, there's no way to use a central constant here, and we don't have the Windows + Emscripten setup for testing the AOT variant yet, so we probably should add it to some "steps to do during ABI version bump" doc.

corresponding GitHub label.

# Testing

### DIFF
--- a/crates/bindings-csharp/Runtime/RawBindings.cs
+++ b/crates/bindings-csharp/Runtime/RawBindings.cs
@@ -11,7 +11,7 @@ public static partial class RawBindings
     // `LibraryImport` directly.
     const string StdbNamespace =
 #if EXPERIMENTAL_WASM_AOT
-        "spacetime_7.0"
+        "spacetime_8.0"
 #else
         "bindings"
 #endif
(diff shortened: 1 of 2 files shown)


kazimuth (approved):
It would be nice to get an automated test harness up to catch these things, but that is going to be a huge pain.

I think github actions technically can use windows runners... you probably have to write PowerShell though. I've heard PowerShell is good, but never really learned it...

crates/bindings-csharp/Runtime/RawBindings.cs
```diff
@@ -63,28 +62,36 @@ public class StdbException : Exception
 
     public class NoSuchTableException : StdbException
     {
-        internal NoSuchTableException() : base(ERRNO_NO_SUCH_TABLE) { }
+        internal NoSuchTableException()
```
kazimuth, line 65: How come this formatting change ended up in this PR? Did the autoformatter in use change or something?

RReverser: I think @coolreader18 just didn't have autoformatted configured in their editor, but I do, so it formatted code on save.

RReverser:
> It would be nice to get an automated test harness up to catch these things, but that is going to be a huge pain.

TBH we could just do a search for `spacetime_7.0` (substitute current version) in the codebase. I'm intentionally keeping those strings raw (not interpolated) to make search-replace easier, so in theory we could do a dumb search in CI as well.

Issue #1258, opened 2024-05-20
Fix bug with multi-col index query "WHERE a > 2 AND a < 6 AND b < 3 AND b < 7"


Centril:
It's currently translated to an index scan on the range `(2, 3) .. (6, 7)` but that is not sound.
We can only accept a constraint `a = X` together with a range scan on `b`.

PR #1316, opened 2024-05-30
 Restrict multi-col index scans to `=` (`OpCmp::Eq`) on all columns 

Centril:
# Description of Changes

Restrict multi-col index scans to `=` (`OpCmp::Eq`) on all columns to fix the current unsound index optimization bug.
Later, we can lift the restriction with a more elaborate and correct implementation.

Fixes https://github.com/clockworklabs/SpacetimeDB/issues/1258.

None

1, localized changes.

# Testing

The tests have been hardened to catch this, both at query compilation time and query execution run-time.

### DIFF
--- a/crates/vm/src/expr.rs
+++ b/crates/vm/src/expr.rs
@@ -917,10 +917,8 @@ type FieldsIndexed = HashSet<(FieldName, OpCmp)>;
 ///   i.e., both `WHERE a = 1 AND b = 2`
 ///   and `WHERE b = 2 AND a = 1` are valid.
 ///
-/// - Queries against multi-col indices must use the same operator in their constraints.
+/// - Queries against multi-col indices must use `=`, for now, in their constraints.
 ///   Otherwise, the index cannot be used.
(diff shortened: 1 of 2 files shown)


crates/vm/src/expr.rs
```diff
@@ -982,57 +975,61 @@ fn select_best_index<'a>(
...
+                .min()
+                .unwrap_or_default();
+
+            // For all of these sets of constraints,
+            // construct the value to compare against.
+            while min_all_cols_num_eq > 0 {
+                let mut elems = Vec::with_capacity(col_list.len() as usize);
+                for col in col_list.iter() {
+                    // Cannot panic as `min_all_cols_num_eq > 0`.
+                    let field = pop_multimap(&mut fields_map, (col, cmp)).unwrap();
```
mamcx, line 1023: Why not use `let Some(...)` here to skip the two passes for the search of `col, cmp`?

mamcx (approved):
LGTM

PR #1318, opened 2024-05-30
NFC: use record struct for auto-equality in C#

RReverser:
The new `record struct` construct is the same as `struct`, but automatically derives equality and hashing by walking through the fields - exactly what we need for those 2 types.

# Description of Changes

corresponding GitHub label.

# Testing


crates/bindings-csharp/Runtime/RawBindings.cs
```diff
@@ -63,28 +62,36 @@ public class StdbException : Exception
 
     public class NoSuchTableException : StdbException
     {
-        internal NoSuchTableException() : base(ERRNO_NO_SUCH_TABLE) { }
+        internal NoSuchTableException()
```
kazimuth, line 65: These formatting changes are also in https://github.com/clockworklabs/SpacetimeDB/pull/1309

kazimuth (approved):
Lovely red minuses.

PR #1330, opened 2024-06-03
Fix commitlog `fold_transactions_from` ignoring requested offset

gefjon:
# Description of Changes

Prior to this commit, `fold_transactions_from` on a durability backed by a commitlog would discard the requested offset and unconditionally yield all txes in the relevant segments.

This commit changes that behavior so that `fold_transactions_from` skips commitlog commits (which contain many txes) less than the reqested offset, and skips txes using `consume_record`.

N/a

3 - commitlog code is complicated. Is my test correct and sufficient? Did I correctly use `decode_record` and `consume_record`? Have I accidentally broken replay? Do I have an off-by-one error in my comparison operators?

# Testing

- [x] Wrote a new unit test for the affected functionality.

### DIFF
--- a/crates/commitlog/src/commitlog.rs
+++ b/crates/commitlog/src/commitlog.rs
@@ -399,9 +399,9 @@ where
         for n in 0..commit.n {
             let tx_offset = commit.min_tx_offset + n as u64;
             if tx_offset < from {
-                de.consume_record(version, tx_offset, records)?;
+                de.skip_record(version, tx_offset, records)?;
             } else {
-                de.decode_record(version, tx_offset, records)?;
+                de.consume_record(version, tx_offset, records)?;
             }
         }
     }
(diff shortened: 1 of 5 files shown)


crates/commitlog/src/commitlog.rs
```diff
@@ -390,10 +390,19 @@ where
...
+        }
+
         let records = &mut commit.records.as_slice();
         for n in 0..commit.n {
             let tx_offset = commit.min_tx_offset + n as u64;
-            de.consume_record(version, tx_offset, records)?;
+            if tx_offset < from {
+                de.consume_record(version, tx_offset, records)?;
+            } else {
+                de.decode_record(version, tx_offset, records)?;
```
kim, line 404: Both consume and decode drive the visitor, but decode may allocate the txdata (and then drop it immediately). I think you just want to `continue` if the offset is smaller than `from`.

gefjon: Don't you need some way to consume the bytes from the `records` stream?

gefjon: Oh, this is going to be more of a pain than I realized. Sigh.

kim (approved):

PR #1332, opened 2024-06-03
Make both TX offset counters agree

gefjon:
# Description of Changes

Prior to this commit, `CommittedState::next_tx_offset` got out of sync with the commitlog/durability's notion of the tx offset, because the former counted all committed TXes, while the latter excluded certain empty TXes (namely, empty TXes which were not `__identity_connected__` or `__identity_disconnected__` reducers).

With this commit, the skipping logic is moved earlier into `CommittedState`, so that it can maintain a counter consistent with that used by the commitlog.

N/a

2 - I may have inadvertently changed the logic of whether to skip a TX while moving it.

# Testing

Nothing comes to mind as useful.

### DIFF
--- a/crates/core/src/db/datastore/locking_tx_datastore/committed_state.rs
+++ b/crates/core/src/db/datastore/locking_tx_datastore/committed_state.rs
@@ -360,6 +360,33 @@ impl CommittedState {
         table.get_row_ref(&self.blob_store, row_ptr).unwrap()
     }
 
+    /// True if the transaction `(tx_data, ctx)` will be written to the commitlog,
+    /// and therefore consumes a value from `self.next_tx_offset`.
+    ///
+    /// A TX is written to the logs if any of the following holds:
+    /// - The TX inserted at least one row.
+    /// - The TX deleted at least one row.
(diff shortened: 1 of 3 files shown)


crates/core/src/db/datastore/locking_tx_datastore/committed_state.rs
```diff
@@ -373,6 +400,13 @@ impl CommittedState {
 
         self.merge_apply_inserts(&mut tx_data, tx_state.insert_tables, tx_state.blob_store, ctx);
 
+        // If the TX will be logged, record its projected tx offset,
+        // then increment the counter.
+        if self.tx_consumes_offset(&tx_data, ctx) {
+            tx_data.set_tx_offset(self.next_tx_offset);
+            self.next_tx_offset += 1;
```
kim, line 407: Did you forget to remove the earlier increment?

gefjon: D'oh! Pushed.

cloutiertyler (approved):
This is fine by me

PR #1338, opened 2024-06-04
Make `update_database` insensitive to table access changes

Centril:
# Description of Changes

The change here does two things:

1. It no longer considers table access changes to a table to be a schema incompatibility.
2. A change in the table access results in a change in `tx_state.insert_tables` as well as the `st_tables` system table.

None

3, touches system table logic, review with care wrt. transactionality?

### DIFF
--- a/crates/core/src/db/update.rs
+++ b/crates/core/src/db/update.rs
@@ -4,11 +4,12 @@ use crate::database_logger::SystemLogger;
 use crate::error::DBError;
 use crate::execution_context::ExecutionContext;
 use anyhow::Context;
-use core::fmt;
+use core::{fmt, mem};
 use itertools::Itertools;
 use similar::{Algorithm, TextDiff};
 use spacetimedb_data_structures::map::HashMap;
 use spacetimedb_primitives::ConstraintKind;
+use spacetimedb_sats::db::auth::StAccess;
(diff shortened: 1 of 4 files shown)


crates/core/src/db/update.rs
```diff
@@ -197,7 +208,32 @@ pub fn schema_updates(
             );
             let proposed_schema = TableSchema::from_def(known_schema.table_id, proposed_schema_def);
 
-            if proposed_schema != known_schema {
+            // HACK(Centril): Compute whether the new schema is in-compatible with the old,
```
kim, line 211: Is in-compatible dual to out-compatible? :stuck_out_tongue:

crates/core/src/db/update.rs
```diff
@@ -197,7 +208,32 @@ pub fn schema_updates(
...
+            // while ignoring `.table_access`.
+            let (schema_is_incompatible, proposed_schema, known_schema) = {
+                // Change the access of both known and proposed to `Public`.
+                // We could have also picked `Private`.
+                // It does not matter. We just want access to be irrelevant wrt. schema compatibility.
```
kim, line 216: "wrt" or "w.r.t." (I consulted the Oxford reference!)

kim (approved):
l.g.t.m.
Nits are purely for entertainment purposes.

Issue #1339, opened 2024-06-05
(latest master) Issue with config lock file


jdetter:
I keep hitting this on macos when trying to publish BitCraft, I have a feeling the BitCraft team will run into this as well. The problem resolves if I delete the lock file but then it comes back.

```
(python-env) boppy@macbook game % spacetime publish -c bitcraft -s local -S
thread 'main' panicked at crates/cli/src/config.rs:911:58:
called `Result::unwrap()` on an `Err` value: Unable to acquire lock on config file "/Users/boppy/.spacetime/config.toml": failed to create lockfile "/Users/boppy/.spacetime/config.lock"

Caused by:
    File exists (os error 17)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

jdetter:
FWIW I also was running into this on windows a while ago so its probably present there as well

bfops:
I was able to reproduce this by killing a `spacetime publish` (with `Ctrl-C`).

Notably, `spacetime publish` holds the lock for the entire duration, which includes compiling and crossing a network.

PR #1341, opened 2024-06-05
Fix bug with `Lockfile` sticking around

bfops:
# Description of Changes
Fixes https://github.com/clockworklabs/SpacetimeDB/issues/1339.

I updated `Config::save` to use an atomic write, so we no longer have the race condition that the `Lockfile` was originally created to address.

Correspondingly, I've removed `Config`'s usage of `Lockfile` entirely.

Nope

2

# Testing

- [ ] Test that `spacetime identity add` still works as expected
- [x] CI passes

### DIFF
--- a/crates/cli/src/config.rs
+++ b/crates/cli/src/config.rs
@@ -860,8 +860,8 @@ impl Config {
         // that approach (see https://github.com/clockworklabs/SpacetimeDB/issues/1339, and the
         // TODO in `lockfile.rs`).
         //
-        // We should address this issue, but we currently don't expect it to arise very frequently.
-        // (https://github.com/clockworklabs/SpacetimeDB/pull/1341#issuecomment-2150857432)
+        // We should address this issue, but we currently don't expect it to arise very frequently
+        // (see https://github.com/clockworklabs/SpacetimeDB/pull/1341#issuecomment-2150857432).
         if let Err(e) = atomic_write(&home_path, config) {
             eprintln!("could not save config file: {e}")
         }
--- a/crates/cli/src/config.rs
+++ b/crates/cli/src/config.rs
@@ -106,6 +106,7 @@ pub struct RawConfig {
     server_configs: Vec<ServerConfig>,
 }
 
+#[derive(Debug, Clone)]
 pub struct Config {
     proj: RawConfig,
     home: RawConfig,
--- a/crates/cli/src/config.rs
+++ b/crates/cli/src/config.rs
@@ -3,13 +3,12 @@ use anyhow::Context;
 use jsonwebtoken::DecodingKey;
 use serde::{Deserialize, Serialize};
 use spacetimedb::auth::identity::decode_token;
-use spacetimedb_fs_utils::create_parent_dir;
+use spacetimedb_fs_utils::{atomic_write, create_parent_dir};
 use spacetimedb_lib::Identity;
 use std::{
     fs,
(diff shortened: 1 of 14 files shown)


crates/cli/src/config.rs
```diff
@@ -594,21 +581,16 @@ Fetch the server's fingerprint with:
...
-        // e.g. by creating a new identity if none exists.
-    }
+fn atomic_write(file_path: &PathBuf, data: String) -> anyhow::Result<()> {
+    let temp_file = NamedTempFile::new()?;
+    // Close the file, but keep the path to it around.
+    let temp_path = temp_file.into_temp_path();
+    std::fs::write(&temp_path, data)?;
+    std::fs::rename(&temp_path, file_path)?;
+    Ok(())
+}
```
bfops, line 591: move this to the `fs_utils` crate

crates/cli/src/config.rs
```diff
@@ -861,50 +833,26 @@ impl Config {
...
         Self {
             home: RawConfig::new_with_localhost(),
             proj: RawConfig::default(),
-
-            home_file: None,
-        }
-    }
-
-    #[doc(hidden)]
-    pub fn clone_for_test(&self) -> Self {
```
bfops, line 884: no longer needed since `Config` is back to being "plain ol' data"

gefjon:
> If we were to address this remaining config-clobbering issue, I would actually push to just have save check whether the config file has been changed since the initial read, and fail in that edge case.

This is insufficient because the CLI may have already performed observable side-effects by the time you abort.

> But, given that this case seems pretty unlikely, I'm not compelled to fix it now.

🤷

gefjon (approved):
As I've said, I don't agree with this direction, but I'm willing to defer to you on the design here. Code is clear and does the thing it's supposed to.

bfops:
> > If we were to address this remaining config-clobbering issue, I would actually push to just have save check whether the config file has been changed since the initial read, and fail in that edge case.
> 
> This is insufficient because the CLI may have already performed observable side-effects by the time you abort.

I agree with this as a general point, but I didn't see any obvious commands where this seemed like an unrecoverable loss. Are there specific side-effects you're thinking about / know of?

bfops:
> As I've said, I don't agree with this direction, but I'm willing to defer to you on the design here. Code is clear and does the thing it's supposed to.

I appreciate it! But I'm pretty new here - let's get someone else to weigh in before I just override your objections.

crates/cli/src/config.rs
```diff
@@ -852,6 +852,18 @@ impl Config {
...
         let config = toml::to_string_pretty(&self.home).unwrap();
 
+        // TODO: We currently have a race condition if multiple processes are modifying the config.
+        // If process X and process Y read the config, each make independent changes, and then save
+        // the config, the first writer will have its changes clobbered by the second writer.
+        //
+        // We used to use `Lockfile` to prevent this from happening, but we had other issues with
+        // that approach (see https://github.com/clockworklabs/SpacetimeDB/issues/1339).
+        // We should eventually reintroduce `Lockfile`, with further fixes (see the TODO in
+        // `lockfile.rs`).
```
cloutiertyler, line 862: I don't know that we should ever reintroduce `Lockfile`. I think we need a more general solution to the problem. The issue with `Lockfile` is that it may never be released.

cloutiertyler (approved):
LGTM

PR #1342, opened 2024-06-05
feat(1329): System table for system variables

### LINKED ISSUE #1329: Keep SQL variables in a system table
Right now we have environment variables that are runtime configurable through the SQL shell. These variables control database state but are not written to the commitlog as of yet. We should store them in a system table in order to track their state properly.

joshua-spacetime:
Closes #1329.

Used to persist global parameters for slow query logging and cardinality limits.

2

# Testing

### DIFF
--- a/crates/core/src/db/datastore/system_tables.rs
+++ b/crates/core/src/db/datastore/system_tables.rs
@@ -1,17 +1,25 @@
+use crate::db::relational_db::RelationalDB;
 use crate::error::{DBError, TableError};
+use crate::execution_context::ExecutionContext;
 use core::fmt;
-use spacetimedb_lib::{Address, Identity};
+use derive_more::From;
+use spacetimedb_lib::{Address, Identity, SumType};
 use spacetimedb_primitives::*;
 use spacetimedb_sats::db::auth::{StAccess, StTableType};
 use spacetimedb_sats::db::def::*;
 use spacetimedb_sats::hash::Hash;
 use spacetimedb_sats::product_value::InvalidFieldError;
 use spacetimedb_sats::{
-    impl_deserialize, impl_serialize, product, AlgebraicType, AlgebraicValue, ArrayValue, ProductValue,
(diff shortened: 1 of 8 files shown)


crates/core/src/db/datastore/system_tables.rs
```diff
@@ -716,10 +751,419 @@ impl From<&StClientsRow> for ProductValue {
...
+            | StVarName::SlowQueryThreshold
+            | StVarName::SlowSubThreshold
+            | StVarName::SlowIncThreshold => StVarType::U64,
+        }
+    }
+}
+
+/// The type of a system variable in `st_var`
+#[repr(u8)]
+pub enum StVarType {
```
mamcx, line 901: Why do we need another `type` for this? I think just matching on `AlgebraicType` if wanna restrict to `primitive` types is enough

joshua-spacetime: Can a column be typed as `AlgebraicValue`? I was under the assumption it needed a concrete type. In this case, `AlgebraicType::Sum(...)`.

gefjon: We could conceivably `impl SpacetimeType for AlgebraicValue` by having an `AlgebraicType` which describes `AlgebraicValue`.

crates/core/src/vm.rs
```diff
@@ -402,13 +405,14 @@ pub struct DbProgram<'db, 'tx> {
...
+    ctx: &ExecutionContext,
+    db: &RelationalDB,
+    tx: &TxId,
+    row_est: impl Fn(&QuerySet, &TxId) -> u64,
     auth: &AuthCtx,
-    config: &DatabaseConfig,
 ) -> Result<(), DBError> {
     if auth.caller != auth.owner {
-        if let Some(limit) = config.row_limit {
+        if let Some(limit) = StVarTable::row_limit(ctx, db, tx)? {
```
joshua-spacetime, line 415: We use the `row_limit` in `st_var` when estimating cardinality.

mamcx (approved):
LGTM

PR #1344, opened 2024-06-06
Integrate snapshotting into core

gefjon:
# Description of Changes

This PR integrates snapshotting into `RelationalDB`.

In this initial version, we take a snapshot every 1 million txes written to the commitlog.

Part of this involves making the `committed_state` module `pub(crate)` where previously it was private.
In order to preserve the previous visibility of its members, anything that was previously `pub` is now `pub(super)`.

Notable TODOs, some or all of which may be left as future work:

- [x] Mark snapshots newer than the `durable_tx_offset` as invalid when restoring.
- [ ] Write a test that exercises snapshot capture and restore. This is challenging because of the large `SNAPSHOT_INTERVAL`; unless/until we make that configurable, such a test would have to commit 1 million txes.
- [ ] Make `SNAPSHOT_INTERVAL` configurable.

Nothing broken, but we expose a new API and on-disk format which we will have to maintain in the future.

4 - bootstrapping and replaying a database is complex and error-prone, and this PR adds significantly more surface area to that codepath which can break.

# Testing

- [x] Manual testing in #1279 .
- [ ] Edit the definition of `SNAPSHOT_INTERVAL` to a small number, possibly 1000, then publish your favorite module, run it to capture several snapshots, then restart, observe the host logs about the snapshot being restored, and verify the state is correct.
  - I did this, but you should too.
- [ ] BitCraft bot test long enough to generate snapshots.


gefjon:
I just ran a test with my `counter-module`, which is:
```rust
use spacetimedb::{schedule, spacetimedb};

#[spacetimedb(table)]
pub struct Counter {
    #[primarykey]
    id: u64,
    count: u64,
}

#[spacetimedb(init)]
pub fn init() {
    Counter::insert(Counter { id: 0, count: 0 }).unwrap();
}

#[spacetimedb(reducer)]
pub fn count(repeat: bool) {
    let mut current = Counter::filter_by_id(&0).unwrap();
    current.count += 1;
    log::info!("Counted to {}", current.count);
    Counter::update_by_id(&0, current);
    if repeat {
        schedule!("0ms", count(true));
    }
}
```

The test was:
- Build StDB with `SNAPSHOT_FREQUENCY = 1000`.
- Run:
  ```bash
  for i in {0..3000}; do spacetime call counter count false; done
  ```
- Interrupt the above when I got bored, which turned out to be at 1989.
- Run:
  ```bash
  spacetime sql counter 'SELECT * FROM Counter'
  ```
  and see 1989.
- Restart StDB and run the above query again, see the same value.

This worked and the host logged appropriately about restoring from a snapshot (which took ~500 us to capture and < 1 ms to read and restore, since it contained only 8 pages).

Curiously, significantly more TXes happened than I was expecting. I made 1989 calls to the reducer. Including connect/disconnect calls, this amounts to ~6000 TXes. But the tx offset of my snapshot was 7000, and the `durable_tx_offset` at restart time was 7956. The astute reader will notice that 7956 / 1989 = 4, meaning each reducer call via the CLI resulted in 4 mutable TXes, not the 3 I was expecting.

I'm not worried about this, but I am confused and intrigued.

kim:
> each reducer call via the CLI resulted in 4 mutable TXes, not the 3 I was expecting.

This could be the extra disconnect tx introduced in #1288

gefjon:
> > each reducer call via the CLI resulted in 4 mutable TXes, not the 3 I was expecting.
> 
> This could be the extra disconnect tx introduced in #1288

Oh yeah, that'd do it. Thanks Kim!

crates/core/src/db/relational_db.rs
```diff
@@ -924,6 +1097,9 @@ pub mod tests_utils {
         /// Note that flushing the log is an asynchronous process. [`Self::reopen`]
         /// ensures all data has been flushed to disk before re-opening the
         /// database.
+        ///
+        /// Requires access to a Tokio runtime in order to spawn a snapshot worker.
```
kim, line 1101: Do we want or need to start two runtimes? `durable` could be passed in a handle, or the handle can be obtained from the `TestDB`.

crates/core/src/db/relational_db.rs
```diff
@@ -60,6 +65,7 @@ pub struct RelationalDB {
     // TODO(cloutiertyler): This should not be public
     pub(crate) inner: Locking,
     durability: Option<Arc<dyn Durability<TxData = Txdata>>>,
+    snapshots: Option<Arc<SnapshotWorker>>,
```
cloutiertyler, line 68:
```suggestion
    snapshotter: Option<Arc<SnapshotWorker>>,
```
perhaps?

I wonder if this should be owned by the `RelationalDB` or if it would just be better to have this be something held by the `Host` or `DatabaseInstanceContext` and to put `maybe_do_snapshot` in there as well. Although I haven't through through exactly who should call `maybe_do_snapshot`

gefjon: I think it makes the most sense to store in the `RelationalDB` because:
- It acts a lot like the `Durability`, which is also owned by the `RelationalDB`.
- Opportunities to call `maybe_do_snapshot` become increasingly tenuous as you get more abstracted than this.
- I would prefer not to leak the `CommittedState`, or even the `Locking`, any further than this.

gefjon: I've renamed all the `snapshots` vars and members to either `snapshot_worker` or `snapshot_repo` as appropriate.

crates/core/src/db/relational_db.rs
```diff
@@ -419,6 +571,20 @@ impl RelationalDB {
         }
     }
 
+    /// Decide based on the `committed_state.next_tx_offset`
+    /// whether to request that the [`SnapshotWorker`] in `self` capture a snapshot of the database.
+    ///
+    /// Actual snapshotting happens asynchronously in a Tokio worker.
+    fn maybe_do_snapshot(&self, tx_data: &TxData) {
+        if let Some(snapshot_worker) = &self.snapshot_worker {
+            if let Some(tx_offset) = tx_data.tx_offset() {
```
kim, line 580: I know it's been discussed, but maybe it can be explained in the commentary: why are we snapshotting independent of the durable tx offset?

crates/core/src/db/relational_db.rs
```diff
@@ -121,7 +198,21 @@ impl RelationalDB {
...
+        let snapshot_repo = SnapshotRepository::open(snapshot_dir, address, instance_id).map(Arc::new)?;
+
+        let start = std::time::Instant::now();
+        let res = Self::open(
+            root,
+            address,
+            Some((durability.clone(), disk_size_fn)),
+            Some(snapshot_repo),
+        )?
+        .apply(durability);
```
kim, line 212: I think we will need to determine the `ConnectedClients` entirely from `st_clients`, as it appears right after this point.

Consider: the database crashed right after a snapshot, so the history suffix is empty -- but there might be connected clients in `st_clients`. Before we had snapshotting, an empty history meant an empty database.

gefjon:
Rebased and performed manual testing. Appears to still work.

crates/core/src/host/host_controller.rs
```diff
@@ -624,7 +624,9 @@ async fn make_module_host(
     mcc: ModuleCreationContext,
     unregister: impl Fn() + Send + Sync + 'static,
 ) -> anyhow::Result<ModuleHost> {
+    let rt = tokio::runtime::Handle::current();
     spawn_rayon(move || {
+        let _rt = rt.enter();
```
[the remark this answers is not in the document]
gefjon: When I first did this, failing to include the enter guard here caused the `testing` crate to fail its tests for reasons I don't understand. This appears to no longer be the case, again for reasons I don't understand, so I'll remove this.

crates/core/src/db/relational_db.rs
gefjon: I believe this is the case now that I've "rebased" onto latest master, including `st_clients`. Please verify that I'm doing this correctly.

crates/core/src/db/relational_db.rs
gefjon: Added comment.

crates/table/src/table.rs
```diff
@@ -1055,6 +1070,17 @@ impl Table {
     pub(crate) fn static_bsatn_layout(&self) -> Option<&StaticBsatnLayout> {
         self.inner.static_bsatn_layout.as_ref()
     }
+
+    /// Rebuild the [`PointerMap`] by iterating over all the rows in `self` and inserting them.
+    ///
+    /// Called when restoring from a snapshot after installing the pages.
+    fn rebuild_pointer_map(&mut self, blob_store: &dyn BlobStore) {
+        let ptrs = self
```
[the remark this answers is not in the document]
gefjon: Oh, I'm definitely not computing `self.row_count` when restarting. Will amend (and also add this comment).

crates/core/src/db/datastore/system_tables.rs
```diff
@@ -334,6 +334,19 @@ fn st_module_schema() -> TableSchema {
...
+pub(crate) fn system_table_schema(table_id: TableId) -> Option<TableSchema> {
+    match table_id {
+        ST_TABLES_ID => Some(st_table_schema()),
+        ST_COLUMNS_ID => Some(st_columns_schema()),
+        ST_SEQUENCES_ID => Some(st_sequences_schema()),
+        ST_INDEXES_ID => Some(st_indexes_schema()),
+        ST_CONSTRAINTS_ID => Some(st_constraints_schema()),
+        ST_MODULE_ID => Some(st_module_schema()),
+        ST_CLIENTS_ID => Some(st_clients_schema()),
+        _ => None,
```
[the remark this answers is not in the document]
gefjon: Done during "rebase."

crates/core/src/db/datastore/locking_tx_datastore/committed_state.rs
```diff
@@ -233,20 +233,30 @@ impl CommittedState {
...
         let ctx = ExecutionContext::internal(database_address);
-        for table_id in ref_schemas.map(|s| s.table_id) {
-            let schema = self.schema_for_table_raw(&ctx, table_id)?;
-            self.tables
-                .get_mut(&table_id)
-                .unwrap()
-                .with_mut_schema(|ts| *ts = schema);
+        for schema in system_tables() {
+            self.tables.get_mut(&schema.table_id).unwrap().schema =
+                Arc::new(self.schema_for_table_raw(&ctx, schema.table_id)?);
```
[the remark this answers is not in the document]
gefjon: This is the way it was written in `bootstrap_system_tables`; I copied the body of this function from there.

crates/core/src/db/relational_db.rs
```diff
@@ -75,6 +81,76 @@ pub struct RelationalDB {
...
+                .unwrap();
+        }
+    }
+
+    fn take_snapshot(committed_state: &RwLock<CommittedState>, snapshot_repo: &SnapshotRepository) {
+        let mut committed_state = committed_state.write();
+        let Some(tx_offset) = committed_state.next_tx_offset.checked_sub(1) else {
+            log::info!("SnapshotWorker::take_snapshot: refusing to take snapshot at tx_offset -1");
+            return;
+        };
```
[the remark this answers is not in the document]
gefjon: No, there'd be a weird corner case where we could try to snapshot TX -1 as if it was TX 0. I don't know if we'd ever hit that case - it's possible that this never gets called until at least one TX has been committed - but I don't feel comfortable asserting that is true and will always be true, so I'd rather just include the more verbose `checked_sub`.

crates/core/src/db/datastore/locking_tx_datastore/committed_state.rs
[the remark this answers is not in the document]
cloutiertyler: Unrelatedly, I don't really understand what this is doing or what the point of it is really. I don't recall in what ways the system tables change from original hardcoded schemas on startup. So I don't really understand why this needs to change

crates/core/src/db/datastore/locking_tx_datastore/datastore.rs
```diff
@@ -129,6 +130,58 @@ impl Locking {
...
+        } = snapshot;
+
+        let datastore = Self::new(database_address);
+        let mut commit_state = datastore.committed_state.write_arc();
+        commit_state.blob_store = blob_store;
+
+        let ctx = ExecutionContext::internal(datastore.database_address);
+
+        for (table_id, pages) in tables {
+            let schema = match system_table_schema(table_id) {
```
cloutiertyler, line 161: Won't this have the problem that the changes that are applied during bootstrapping (referenced above) won't be applied to the system tables?

crates/core/src/db/datastore/locking_tx_datastore/datastore.rs
```diff
@@ -129,6 +130,58 @@ impl Locking {
...
+        let ReconstructedSnapshot {
+            database_address,
+            tx_offset,
+            blob_store,
+            tables,
+            ..
+        } = snapshot;
+
+        let datastore = Self::new(database_address);
+        let mut commit_state = datastore.committed_state.write_arc();
```
cloutiertyler, line 155: Please rename this to `committed_state` for consistency.

crates/core/src/db/datastore/locking_tx_datastore/datastore.rs
gefjon: The later call to `reset_system_table_schemas` does this.

crates/core/src/db/datastore/locking_tx_datastore/datastore.rs
gefjon: It is also named `commit_state` in `bootstrap`. I will rename it here, but I think adjusting `bootstrap` should happen in a follow-up.

crates/core/src/db/relational_db.rs
```diff
@@ -64,6 +69,7 @@ pub struct RelationalDB {
 
     inner: Locking,
     durability: Option<Arc<dyn Durability<TxData = Txdata>>>,
+    snapshot_worker: Option<Arc<SnapshotWorker>>,
```
cloutiertyler, line 72: I wonder if it would be best to move this outside of the `RelationalDB` (`DatabaseEngine` future name), to something slightly higher level (e.g. `DatabaseInstanceContext`). That way we maintain that `RelationalDB` remains runtime independent. I don't feel it strongly enough to block the review, but it makes me uncomfortable to have the `DatabaseEngine`/`RelationalDB` dependent on tokio.

crates/core/src/db/datastore/locking_tx_datastore/datastore.rs
```diff
@@ -129,6 +130,58 @@ impl Locking {
...
+        commit_state.blob_store = blob_store;
+
+        let ctx = ExecutionContext::internal(datastore.database_address);
+
+        for (table_id, pages) in tables {
+            let schema = match system_table_schema(table_id) {
+                Some(schema) => Arc::new(schema),
+                None => commit_state.schema_for_table(&ctx, table_id)?,
```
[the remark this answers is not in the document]
gefjon: There cannot be a schema cached in the committed state because we just constructed the committed state new and empty. I will add a comment to this effect.

crates/core/src/db/relational_db.rs
cloutiertyler: To be clear, I think the ability to snapshot itself would remain inside of `RelationalDB`, I'm specifically just referring to the scheduling/triggering system.

crates/core/src/db/relational_db.rs
```diff
@@ -76,6 +82,80 @@ pub struct RelationalDB {
...
+        while let Some(()) = trigger.next().await {
+            let committed_state = committed_state.clone();
+            let repo = repo.clone();
+            tokio::task::spawn_blocking(move || Self::take_snapshot(&committed_state, &repo))
+                .await
+                .unwrap();
+        }
+    }
+
+    fn take_snapshot(committed_state: &RwLock<CommittedState>, snapshot_repo: &SnapshotRepository) {
```
cloutiertyler, line 118: In reference to the scheduling comment above, I would imagine that this function would be added to the `RelationalDB` type.

crates/core/src/db/relational_db.rs
```diff
@@ -292,6 +396,48 @@ impl RelationalDB {
...
+                    if snapshot.database_address != address {
+                        // TODO: return a proper typed error
+                        return Err(anyhow::anyhow!(
+                            "Snapshot has incorrect database_address: expected {address} but found {}",
+                            snapshot.database_address,
+                        )
+                        .into());
+                    }
+                    let start = std::time::Instant::now();
+                    let res = Locking::restore_from_snapshot(snapshot);
```
cloutiertyler, line 427: This will eventually need to be a datastore trait function.

crates/core/src/db/datastore/locking_tx_datastore/datastore.rs
cloutiertyler: Ah I missed that.

cloutiertyler (approved):
This now looks good to me. 🙏

crates/core/src/db/relational_db.rs
kim: Great, thanks!

crates/core/src/db/datastore/locking_tx_datastore/committed_state.rs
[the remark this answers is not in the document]
gefjon: There is no reason not to use `.with_mut_schema` as far as I am aware, but I did not go that route because I did not change this code.

What's happening is that the hardcoded schemas have 0 for all the autoinc IDs of columns and indices and constrants, AFAICT because we don't want to manually assign and maintain IDs for all the objects in the metaschemas. So when you initially install the `struct Schema` in the `struct Table` from the hardcoded stuff in `system_tables.rs`, you have zeros in places that the actual `st_index` &c rows have real values. This method recomputes the `struct Schema`s to match what's in the system tables.

crates/core/src/db/datastore/locking_tx_datastore/datastore.rs
gefjon: Added a comment to draw attention to it.

crates/core/src/db/relational_db.rs
kim: `apply_history` returns the unmatched `__connect__`s during replay -- but what I meant was, what if the snapshot's tx offset is the last offset in the commitlog (hence the history is empty)?

Arguably an edge case, so fine by me to merge an fix in a later patch (I can submit one if you want).

kim (approved):

crates/core/src/db/relational_db.rs
gefjon: Oh, I assumed it would read the connected clients out of the table. I thought we had done away with groveling the logs for connected clients. I think a follow-up would be good, and if you're willing to do so, that would be great!

kim: https://github.com/clockworklabs/SpacetimeDB/pull/1366

PR #1350, opened 2024-06-06
feat: System table based slow query logging

joshua-spacetime:
Before this patch,
we had a system table for storing slow query logging parameters, as well as a separate in memory structure.

Both paths were updated simultaneously via a SQL SET directive, however both paths were completely independent,
and slow query logging was driven only from the in memory config.

With this change,
the in memory config has been removed,
and slow query logging is only based on what is stored in st_var.

As a result,
you can now make a config change to query logging, either through the SQL shell or by updating the st_var table directly. And any config changes made at runtime will persist after restart.

2

* Removes the previous config for slow query logging, and updates type signatures accordingly
* Replaces the previous config with lookups into `st_var` via the `StVarTable` utility

# Testing


crates/core/src/db/datastore/system_tables.rs
```diff
@@ -820,6 +822,12 @@ impl StVarTable {
         db.insert(tx, ST_VAR_ID, ProductValue::from(StVarRow { name, value }))?;
         Ok(())
     }
+
+    /// Parse the literal representation of a system variable
+    fn parse_var(name: StVarName, literal: &str) -> Result<StVarValue, DBError> {
+        StVarValue::try_from_primitive(parse::parse(literal, &name.type_of())?)
```
mamcx, line 828: In `vm` we have `ErrorType` for this kind of error

crates/core/src/db/datastore/system_tables.rs
```diff
@@ -1095,7 +1103,7 @@ mod tests {
         let db = TestDB::durable().expect("failed to create db");
         let ctx = ExecutionContext::default();
         let _ = db.with_auto_commit(&ctx, |tx| {
-            StVarTable::write_var(&ctx, &db, tx, StVarName::RowLimit, StVarValue::U64(5))
+            StVarTable::write_var(&ctx, &db, tx, StVarName::RowLimit, "5")
```
mamcx, line 1106: It should not be an `AlgebraicValue` that matches the `typeof` of `variables`?

crates/core/src/sql/ast.rs
```diff
@@ -283,7 +281,7 @@ pub enum SqlAst {
     },
     SetVar {
         name: String,
-        value: AlgebraicValue,
+        literal: String,
     },
```
mamcx, line 285: Same here. If is a reason to  switch to stringy values....

crates/core/src/db/datastore/system_tables.rs
joshua-spacetime: This is mainly for UX. I figured it's nicer to write `SET row_limit TO 5` rather than `SET row_limit TO (u64 = 5)`

crates/core/src/sql/ast.rs
joshua-spacetime: Yes and no. For the AST, a string makes sense because we haven't started type checking or variable binding. But once we have a CrudExpr, then yes I agree, types should already be resolved and we shouldn't need to pass around unparsed strings. I've saved that part for future work when we update type checking in general.

mamcx (approved):
LGTM. Must create ticket for adding typed values.

PR #1366, opened 2024-06-11
core: Determine dangling clients from `st_clients`

kim:
# Description of Changes

When opening a `RelationalDB`, determine the set of dangling clients
(`ConnectedClients`) by scanning the `st_clients` table instead of
tracking unmatched `__identity_connected__` calls during history replay.

We left the replay tracking in place in #1288, treating the commitlog as
the sole source of truth. With #1344 (snapshotting), this is no longer
correct: the snapshot may contain rows in `st_clients`, but leave no
history suffix for replay.

2

It may be the case that the somewhat peculiar transactional logic of
connect/disconnect can go wrong in a way which requires consulting the commitlog
inputs. If so, we'd need to read from `st_clients` only in the case of an empty
history after restoring from a snapshot.

---

Depends-on: #1344

### DIFF
--- a/crates/core/src/db/datastore/system_tables.rs
+++ b/crates/core/src/db/datastore/system_tables.rs
@@ -758,16 +758,22 @@ pub fn read_bytes_from_col(row: RowRef<'_>, col: impl StFields) -> Result<Box<[u
 }
 
 /// Read an [`Address`] directly from the column `col` in `row`.
+///
+/// The [`Address`] is assumed to be stored as a flat byte array.
 pub fn read_addr_from_col(row: RowRef<'_>, col: impl StFields) -> Result<Address, DBError> {
     read_bytes_from_col(row, col).map(Address::from_slice)
 }
 
 /// Read an [`Identity`] directly from the column `col` in `row`.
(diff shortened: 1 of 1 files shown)


crates/core/src/db/datastore/system_tables.rs
```diff
@@ -800,6 +800,27 @@ impl From<&StClientsRow> for ProductValue {
     }
 }
 
+impl TryFrom<RowRef<'_>> for StClientsRow {
+    type Error = DBError;
+
+    fn try_from(row: RowRef<'_>) -> Result<Self, Self::Error> {
+        fn read_col<T: TryFrom<AlgebraicValue>>(row: RowRef<'_>, col: StClientsFields) -> Result<T, DBError> {
```
kim, line 807: Is there a less pedestrian way to do this? As in, can I not somehow `Deserialize` the whole `RowRef`?

/cc @Centril

gefjon (approved):

Shubham8287 (approved):
Looks good, I like how this simplify reading st table columns.

PR #1367, opened 2024-06-11
Smoketest auto-disconnect after restart

kim:
(Re-)enables the `zz-docker` tests with some tweaking to increase the chances that a restart actually works.
Adds a smoketest which tests that clients are auto-disconnected after unclean shutdown.

Depends-on: #1343


cloutiertyler (approved):
LGTM!

PR #1390, opened 2024-06-11
Make some commitlog helpers public

lcodes:
# Description of Changes

To scan the commitlog, access to CommittedState is required to maintain the schemas needed to decode the transaction records. To use CommittedState, access to the StateView trait is also needed.

To query the data for existing rows given a new row, an immutable getter is added to Table as well.

No breaks, no risks, just marking a handful of existing functions to be public outside their crates.

### DIFF
--- a/crates/commitlog/src/error.rs
+++ b/crates/commitlog/src/error.rs
@@ -62,7 +62,7 @@ pub struct Append<T> {
 pub struct ChecksumMismatch;
 
 #[derive(Debug, Error)]
-pub enum SegmentMetadata {
+pub(crate) enum SegmentMetadata {
     #[error("invalid commit encountered")]
     InvalidCommit {
         sofar: segment::Metadata,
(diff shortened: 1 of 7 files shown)


kim:
Re: commitlog I'm not against exporting those things (although maybe we could be a bit more selective?). Would need to make sure though that all such "stabilized" internals are properly documented.

lcodes:
> Re: commitlog I'm not against exporting those things (although maybe we could be a bit more selective?). Would need to make sure though that all such "stabilized" internals are properly documented.

Could we mark these new public APIs as "alpha"? (where alpha means the API isn't finalized/stabilized and can change in the future)

Or, we could document these APIs as a means to directly read through the commitlog without having to instantiate a full stdb instance. I think this is the preferred route, as it leads into things like future Kafka integrations.

kim:
Documentation would do, I guess. From a cursory look at how you're using things, I'd say that the `Repo` trait is unstable -- clearly, there are impls which don't support creation of segments.

(Nb. I can only speak for the commitlog here, not the other exports)

Centril (approved):

kim (approved):

PR #1400, opened 2024-06-11
[NFC] [C#] Minor simplifications to type checks in Roslyn codegen

RReverser:
# Description of Changes

Minor simplifications to type checks in Roslyn codegen.

corresponding GitHub label.

1

# Testing

- [x] `dotnet test`

### DIFF
--- a/crates/bindings-csharp/BSATN.Codegen/Utils.cs
+++ b/crates/bindings-csharp/BSATN.Codegen/Utils.cs
@@ -49,16 +49,16 @@ public static class Utils
     public static string GetTypeInfo(ITypeSymbol type)
     {
         // We need to distinguish handle nullable reference types specially:
-        // compiler expands something like `int?` to `System.Nullable<int>` but with the nullable annotation set to `Annotated`
+        // compiler expands something like `int?` to `System.Nullable<int>` with the nullable annotation set to `Annotated`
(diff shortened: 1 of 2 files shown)


crates/bindings-csharp/BSATN.Codegen/Utils.cs
```diff
@@ -49,16 +49,16 @@ IncrementalGeneratorInitializationContext context
...
     {
         // We need to distinguish handle nullable reference types specially:
-        // compiler expands something like `int?` to `System.Nullable<int>` but with the nullable annotation set to `Annotated`
+        // compiler expands something like `int?` to `System.Nullable<int>` with the nullable annotation set to `Annotated`
         // while something like `string?` is expanded to `string` with the nullable annotation set to `Annotated`...
         // Beautiful design requires beautiful hacks.
         if (
             type.NullableAnnotation == NullableAnnotation.Annotated
-            && type.OriginalDefinition.ToString() != "System.Nullable<T>"
+            && type.OriginalDefinition.SpecialType != SpecialType.System_Nullable_T
```
kazimuth, line 57: Hey, at least you got rid of the string comparison 👍

kazimuth (approved):
Looks good to me.

PR #1403, opened 2024-06-12
smoketests: Add test for module hotswapping

kim:
The test establishes a subscription, updates the module (`clear=False`) and asserts that the subscription was not terminated and received an update made using the new module.

---
Depends-on: ~~#1343~~ #1367

### DIFF
--- a/smoketests/tests/modules.py
+++ b/smoketests/tests/modules.py
@@ -149,7 +149,7 @@ fn init() {
 
 #[spacetimedb(reducer)]
 pub fn my_repeating_reducer(prev: Timestamp) {
-    println!("Invoked: ts={:?}, delta={:?}", Timestamp::now(), prev.elapsed());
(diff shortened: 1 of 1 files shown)


smoketests/tests/modules.py
```diff
@@ -161,3 +161,79 @@ def test_upload_module_2(self):
...
+        name = random_string()
+        self.publish_module(name, clear=False)
+        sub = self.subscribe("SELECT * FROM *", n=2)
+
+        # Trigger event on the subscription
+        self.call("add_person", "Horst")
+
+        # Update the module
+        self.write_module_code(self.MODULE_CODE_B)
+        self.publish_module(name, clear=False)
```
cloutiertyler, line 227: Are these synchronous calls?

kim: Anything interacting with the database just shells out to the CLI. So, yes.

cloutiertyler (approved):
LGTM. Merging this.

PR #1404, opened 2024-06-12
core: Simplify custom bootstrap

kim:
Lifecycle management is simplified enough to render the `post_boot`
closure obsolete. Instead the `ModuleHost` is returned, which allows
calling reducers on it.

Even though the method remains useful mainly for control db
bootstrapping, it seems general enough to be part of `HostController`'s
API.

1


cloutiertyler (approved):
Makes sense to me. Appreciate the simplification

PR #1407, opened 2024-06-12
New schema implementation WIP

kazimuth:
# Description of Changes

Starting to implement the new schema proposal.

Will eventually change field canonical field ordering on all client schemas, but that may be a separate PR.

4

# Testing


kazimuth:
Closed in favor of #1448

PR #1413, opened 2024-06-12
Fix auth regression

coolreader18:


crates/client-api/src/auth.rs
```diff
@@ -65,6 +65,12 @@ impl SpacetimeCreds {
             Err(e) => Err(e),
         };
         if let Ok(Query(creds)) = Query::<Self>::try_from_uri(&parts.uri) {
+            // TODO: do we want to have the `?token=` query param just be the jwt, instead of this?
```
cloutiertyler, line 68:
```suggestion
            // TODO STABILITY: do we want to have the `?token=` query param just be the jwt, instead of this?
```

cloutiertyler (approved):
This LGTM. For context to future me, the format of the `token` query param was introduced in https://github.com/clockworklabs/SpacetimeDB/pull/51 which was only meant for short lived queries. We should evaluate whether this query param is still necessary, and what format it should be if it is.

NOTE: We should also consider only accepting it for the websocket route, as it was originally only meant as a short lived token for this route only.

PR #1419, opened 2024-06-13
`TableDef`: clarify `generated_*` methods

Centril:
# Description of Changes

To improve my understanding of what `TableDef` and `TableSchema` actually do wrt. indices & constraints, I slightly refactored the `TableDef::generated_*` methods and added commentary. This should make it easier to understand the implementation of the index hack I am working on.

One semantic aspect that changed here is that I removed the binary searches -- they don't seem to be correct, as I did not determine that we actually keep a sorted order by name in a `TableDef`.

This patch is likely to become obsolete once the new schema impl is written, but I found it helpful in the interim.

None

1

### DIFF
--- a/crates/bindings/src/rt.rs
+++ b/crates/bindings/src/rt.rs
@@ -395,20 +395,14 @@ pub fn register_reftype<T: SpacetimeType>() {
 pub fn register_table<T: TableType>() {
     register_describer(|module| {
         let data = *T::make_type(module).as_ref().unwrap();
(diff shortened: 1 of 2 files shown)


kazimuth (approved):
Makes sense.

This is convincing me that `IndexId`, etc should be kept separate from `Ref<IndexDef>` because there aren't corresponding indices like this in the schema.

Unless I should construct those when building a schema... hmmm....

PR #1421, opened 2024-06-13
Generate tagged enums in C# client code

RReverser:
# Description of Changes

Now that we use a shared BSATN runtime with C# modules, we can leverage our special TaggedEnum construct to support tagged enums in the client as well.

As a drive-by fix, also slightly simplifies TypeScript tagged enum output and marked C# TaggedEnum base record as abstract (since it shouldn't be constructed).

corresponding GitHub label.

1

# Testing

- [x] Ran codegen tests and updated snapshots.


gefjon (approved):

gefjon:
Thanks! For posterity, I asked Ingvar to do this because it's useful for #1077 , as it will allow us to use the CLI to codegen the websocket bindings.

Issue #581, opened 2023-11-21
Track max transaction cpu time


joshua-spacetime:
`histogram_quantile(1, <metric>)` only approximates the maximum value. We should track certain max values in their own metric.

PR #1423, opened 2024-06-13
[STABILITY] Migrate db log paths(waiting folder proposal)

mamcx:
# Description of Changes

Closes [#581](https://github.com/clockworklabs/SpacetimeDBPrivate/issues/581)

Stores module logs in `$STDB_PATH/database_instances`

corresponding GitHub label.

1

# Testing

- [x] *Manually inspect the logs created in the correct path.*

### DIFF
--- a/crates/client-api/src/routes/database.rs
+++ b/crates/client-api/src/routes/database.rs
@@ -435,7 +435,7 @@ where
         ))?;
     let instance_id = database_instance.id;
 
-    let filepath = DatabaseLogger::filepath(&address, instance_id);
+    let filepath = DatabaseLogger::filepath(instance_id);
     let lines = DatabaseLogger::read_latest(&filepath, num_lines).await;
 
     let body = if follow {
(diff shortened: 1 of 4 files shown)


kim:
The full context got lost it seems: the distinction `control_node` / `worker_node` still exists in standalone.

We can either:

- store things in `worker_node` even if it does not make a lot of sense, or
- determine the instances root by a parametrized function relative to STDB_ROOT

I’m more in favor of the latter, so as to not have hardcoded `stdb_path(“some/path”)` all over the place.

mamcx:
Ok, so I need to differentiate by edition?

kim:
I would propose to simply retire the `worker_node` directory. We would then have:

```
STDB_PATH
| -- conf
| -- config.toml
| -- logs
| -- database_instances
|     `-- <address>
|           `-- <instance_id>
|                 | -- database
|                 | -- module_logs
|                 `-- scheduler 
`-- wasmtime_cache
```

Standalone can keep using `control_node` then.

mamcx:
Closed, will be made by #1657

PR #1434, opened 2024-06-13
HACK: Tweak `schema_updates` to allow adding/removing non-unique indices

Centril:
# Description of Changes

You can now add and remove non-generated indices.
That is, the indices that you actually define with `#[spacetimedb(index(...))]`.

None

3, some questions re. transactionality...

# Testing

- Created a project with a table without an index, then added the index, republished, then removed the index and republished (https://github.com/clockworklabs/SpacetimeDB/pull/1434#issuecomment-2167964422).

# Review notes

- Hide whitespace changes on github.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4366,6 +4366,7 @@ dependencies = [
  "derive_more",
  "dirs",
  "email_address",
+ "enum-as-inner",
  "enum-map",
  "env_logger",
  "flate2",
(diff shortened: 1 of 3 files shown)


kim (approved):
Seems to do the right things.

Maybe the smoke tests could be extended to cover what is accepted as an update and what is not (I don't think we can test that an index is present or absent).

Note that adding or dropping an index is not transactional. However, the modifications of the system tables are, so re-publishing a failed update should converge to the correct state.

PR #1440, opened 2024-06-17
Restructure NuGet packaging

RReverser:
Since the 0.10 release and before this change we had 5 NuGet packages that need to be pushed on each release:

- **SpacetimeDB.BSATN.Codegen**: Roslyn codegen for handling `[SpacetimeDB.Type]` attributes and generating [de]serialization code for marked types.
- **SpacetimeDB.BSATN.Runtime**: Runtime API for [de]serialization.
- **SpacetimeDB.Runtime**: Runtime APIs for interacting with SpacetimeDB host.
- **SpacetimeDB.Codegen**: Roslyn codegen for handling `[SpacetimeDB.Table]` and `[SpacetimeDB.Reducer]` attributes and generating module definitions, reducer schedulers and table APIs on marked types and method. It also automatically included **SpacetimeDB.BSATN.Codegen** to generate ser/de APIs as well.
- **SpacetimeDB.ClientSDK**: Client C# SDK.

C# users had to explicitly include:

- Both **SpacetimeDB.ClientSDK** and **SpacetimeDB.BSATN.Codegen** as dependencies if they're developing a client.
- Both **SpacetimeDB.Runtime** and **SpacetimeDB.Codegen** as dependencies if they're developing a module.

---

This PR restructures csproj configurations very differently, making both publishing and end-user consumption simpler. Now both `*Codegen` packages are internal, non-packable, dependencies and their DLLs are automatically included in `*Runtime` counterparts.

As maintainers, we only need to `nuget push` 3 packages now instead of 5:

- **SpacetimeDB.BSATN.Runtime** - it will automatically include **SpacetimeDB.BSATN.Codegen** as part of the same package.
- **SpacetimeDB.Runtime** - it will automatically include **SpacetimeDB.Codegen** in the same package.
- **SpacetimeDB.ClientSDK**: Client C# SDK (it already has **SpacetimeDB.BSATN.Runtime** as a transitive dependency).

As a user, you'll need only 1 dependency now instead of 2:

- **SpacetimeDB.ClientSDK** if you're developing a client.
- **SpacetimeDB.Runtime** if you're developing a module.

# Description of Changes

corresponding GitHub label.

1

# Testing

- [x] Manually ran the updated csharp_module smoketest to verify that packaging still works as expected.

### DIFF
--- a/crates/bindings-csharp/BSATN.Codegen/BSATN.Codegen.csproj
+++ b/crates/bindings-csharp/BSATN.Codegen/BSATN.Codegen.csproj
@@ -1,21 +1,23 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
-    <EnforceExtendedAnalyzerRules>true</EnforceExtendedAnalyzerRules>
-    <IsRoslynComponent>true</IsRoslynComponent>
-    <LangVersion>latest</LangVersion>
-    <Nullable>enable</Nullable>
     <AssemblyName>SpacetimeDB.BSATN.Codegen</AssemblyName>
     <Version>0.10.0</Version>
-    <IncludeBuildOutput>false</IncludeBuildOutput>
-    <!-- We're going to include this analyzer in the Runtime package instead of publishing separately. -->
-    <IsPackable>false</IsPackable>
+    <Title>SpacetimeDB BSATN Codegen</Title>
+    <Description>The SpacetimeDB BSATN Codegen implements the Roslyn incremental generators for BSATN serialization/deserialization in C#.</Description>
   </PropertyGroup>
 
   <PropertyGroup>
-    <Title>SpacetimeDB BSATN Codegen</Title>
-    <Description>The SpacetimeDB BSATN Codegen implements the Roslyn incremental generators for BSATN serialization/deserialization in C#.</Description>
+    <!-- We're going to include this analyzer in the Runtime package instead of publishing separately. -->
+    <IsPackable>false</IsPackable>
+    <IsRoslynComponent>true</IsRoslynComponent>
+    <EnforceExtendedAnalyzerRules>true</EnforceExtendedAnalyzerRules>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <LangVersion>latest</LangVersion>
+    <Nullable>enable</Nullable>
   </PropertyGroup>
 
   <ItemGroup>
--- a/crates/bindings-csharp/BSATN.Codegen/BSATN.Codegen.csproj
+++ b/crates/bindings-csharp/BSATN.Codegen/BSATN.Codegen.csproj
@@ -7,26 +7,15 @@
     <LangVersion>latest</LangVersion>
     <Nullable>enable</Nullable>
     <AssemblyName>SpacetimeDB.BSATN.Codegen</AssemblyName>
-    <AssemblyVersion>0.10.0</AssemblyVersion>
+    <Version>0.10.0</Version>
     <IncludeBuildOutput>false</IncludeBuildOutput>
     <!-- We're going to include this analyzer in the Runtime package instead of publishing separately. -->
     <IsPackable>false</IsPackable>
   </PropertyGroup>
 
   <PropertyGroup>
-    <Version>$(AssemblyVersion)</Version>
     <Title>SpacetimeDB BSATN Codegen</Title>
-    <Authors>RReverser</Authors>
-    <Company>Clockwork Labs</Company>
-    <Product>SpacetimeDB</Product>
     <Description>The SpacetimeDB BSATN Codegen implements the Roslyn incremental generators for BSATN serialization/deserialization in C#.</Description>
-    <Copyright>2023</Copyright>
-    <PackageProjectUrl>https://spacetimedb.com/</PackageProjectUrl>
-    <PackageIcon>logo.png</PackageIcon>
-    <PackageReadmeFile>README.md</PackageReadmeFile>
-    <RepositoryType>git</RepositoryType>
-    <RepositoryUrl>https://github.com/clockworklabs/SpacetimeDB</RepositoryUrl>
-    <PackageLicenseFile>LICENSE</PackageLicenseFile>
   </PropertyGroup>
 
   <ItemGroup>
--- a/crates/bindings-csharp/BSATN.Codegen/BSATN.Codegen.csproj
+++ b/crates/bindings-csharp/BSATN.Codegen/BSATN.Codegen.csproj
@@ -9,6 +9,8 @@
     <AssemblyName>SpacetimeDB.BSATN.Codegen</AssemblyName>
     <AssemblyVersion>0.10.0</AssemblyVersion>
     <IncludeBuildOutput>false</IncludeBuildOutput>
+    <!-- We're going to include this analyzer in the Runtime package instead of publishing separately. -->
+    <IsPackable>false</IsPackable>
   </PropertyGroup>
 
   <PropertyGroup>
@@ -29,13 +31,6 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.3.1" PrivateAssets="all" />
-  </ItemGroup>
-
-  <ItemGroup>
-    <None Include="../logo.png" Pack="true" PackagePath="" />
-    <None Include="../Codegen/README.md" Pack="true" PackagePath="" />
-    <None Include="../LICENSE" Pack="true" PackagePath="" />
-    <None Include="$(OutputPath)/$(AssemblyName).dll" Pack="true" PackagePath="analyzers/dotnet/cs" Visible="false" />
   </ItemGroup>
 
   <ItemGroup>
--- a/crates/bindings-csharp/BSATN.Runtime/BSATN.Runtime.csproj
+++ b/crates/bindings-csharp/BSATN.Runtime/BSATN.Runtime.csproj
@@ -15,7 +15,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <ProjectReference Include="../BSATN.Codegen/BSATN.Codegen.csproj" OutputItemType="Analyzer" PrivateAssets="all" ReferenceOutputAssembly="false" />
+    <!-- We want to build BSATN.Codegen both to include it in our NuGet package but also we want it to transform [SpacetimeDB.Type] usages in BSATN.Runtime code itself. -->
+    <ProjectReference Include="../BSATN.Codegen/BSATN.Codegen.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" />
   </ItemGroup>
 
   <ItemGroup>
--- a/crates/bindings-csharp/BSATN.Runtime/BSATN.Runtime.csproj
+++ b/crates/bindings-csharp/BSATN.Runtime/BSATN.Runtime.csproj
@@ -26,13 +26,15 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <ProjectReference Include="../BSATN.Codegen/BSATN.Codegen.csproj" OutputItemType="Analyzer" PrivateAssets="all" />
+    <ProjectReference Include="../BSATN.Codegen/BSATN.Codegen.csproj" PrivateAssets="all" OutputItemType="analyzer" ReferenceOutputAssembly="false" />
   </ItemGroup>
 
   <ItemGroup>
     <None Include="../logo.png" Pack="true" PackagePath="" />
     <None Include="../Runtime/README.md" Pack="true" PackagePath="" />
     <None Include="../LICENSE" Pack="true" PackagePath="" />
+    <!-- We want all users who depends on BSATN.Runtime to automatically get the Roslyn codegen component as well. -->
+    <None Include="../BSATN.Codegen/bin/$(Configuration)/netstandard2.0/SpacetimeDB.BSATN.Codegen.dll" Pack="true" PackagePath="analyzers/dotnet/cs" />
   </ItemGroup>
 
   <ItemGroup>
--- a/crates/bindings-csharp/BSATN.Runtime/BSATN.Runtime.csproj
+++ b/crates/bindings-csharp/BSATN.Runtime/BSATN.Runtime.csproj
@@ -26,7 +26,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <ProjectReference Include="../BSATN.Codegen/BSATN.Codegen.csproj" PrivateAssets="all" OutputItemType="analyzer" ReferenceOutputAssembly="false" />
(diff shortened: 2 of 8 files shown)


bfops:
Does this require a minor version bump? At first glance it seems like if we uploaded these as e.g. `0.10.1`, then it would break users who are relying on `0.10.*`? Because some of the packages would be updated and some wouldn't?

RReverser:
> Does this require a minor version bump? At first glance it seems like if we uploaded these as e.g. `0.10.1`, then it would break users who are relying on `0.10.*`? Because some of the packages would be updated and some wouldn't?

Correct, we need to leave this until 0.11 because existing users must remove the `SpacetimeDB.Codegen` from their deps after this change. I should've added a label.

RReverser:
I created a `release-0.11` label and assigned it here.

smoketests/tests/csharp_module.py
```diff
@@ -41,7 +41,7 @@ def test_build_csharp_module(self):
                 with open(csproj, "w") as f:
                     f.write(contents)
 
-                run_cmd("dotnet", "build", cwd=tmpdir, capture_stderr=True)
+                run_cmd("dotnet", "publish", cwd=tmpdir, capture_stderr=True)
```
bfops, line 44: is this related to the PR contents, or separate optimization?

crates/bindings-csharp/BSATN.Runtime/BSATN.Runtime.csproj
```diff
@@ -26,13 +26,15 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <ProjectReference Include="../BSATN.Codegen/BSATN.Codegen.csproj" OutputItemType="Analyzer" PrivateAssets="all" />
+    <ProjectReference Include="../BSATN.Codegen/BSATN.Codegen.csproj" PrivateAssets="all" OutputItemType="analyzer" ReferenceOutputAssembly="false" />
```
bfops, line 29: Sanity checking - I don't know enough to know whether this change from `Analyzer` to `analyzer` matters?

crates/bindings-csharp/BSATN.Runtime/BSATN.Runtime.csproj
```diff
@@ -26,13 +26,15 @@
...
   </ItemGroup>
 
   <ItemGroup>
     <None Include="../logo.png" Pack="true" PackagePath="" />
     <None Include="../Runtime/README.md" Pack="true" PackagePath="" />
     <None Include="../LICENSE" Pack="true" PackagePath="" />
+    <!-- We want all users who depends on BSATN.Runtime to automatically get the Roslyn codegen component as well. -->
+    <None Include="../BSATN.Codegen/bin/$(Configuration)/netstandard2.0/SpacetimeDB.BSATN.Codegen.dll" Pack="true" PackagePath="analyzers/dotnet/cs" />
```
bfops, line 37: Dumb question - this says `Pack="true"` but the project above says `<IsPackable>false</IsPackable>` - I assume those don't actually conflict despite kinda sounding like they would?

crates/bindings-csharp/Codegen/Codegen.csproj
```diff
@@ -23,38 +25,24 @@
...
   <ItemGroup>
-    <ProjectReference Include="../BSATN.Codegen/BSATN.Codegen.csproj" PrivateAssets="all" />
+    <!-- This analyzer uses some shared helpers from the BSATN.Codegen one, so include it as a regular dependency -->
+    <!-- (*not* with OutputItemType=analyzer because we don't want to use that source generator on source of Codegen itself) -->
+    <ProjectReference Include="../BSATN.Codegen/BSATN.Codegen.csproj" />
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.5.0" PrivateAssets="all" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.3.1" PrivateAssets="all" />
```
bfops, line 39: why did this get downgraded?

crates/bindings-csharp/Codegen/Codegen.csproj
```diff
@@ -23,38 +25,24 @@
...
-    <RepositoryUrl>https://github.com/clockworklabs/SpacetimeDBLibCSharp</RepositoryUrl>
+    <RepositoryUrl>https://github.com/clockworklabs/SpacetimeDB</RepositoryUrl>
     <PackageLicenseFile>LICENSE</PackageLicenseFile>
   </PropertyGroup>
 
   <ItemGroup>
-    <ProjectReference Include="../BSATN.Codegen/BSATN.Codegen.csproj" PrivateAssets="all" />
+    <!-- This analyzer uses some shared helpers from the BSATN.Codegen one, so include it as a regular dependency -->
+    <!-- (*not* with OutputItemType=analyzer because we don't want to use that source generator on source of Codegen itself) -->
+    <ProjectReference Include="../BSATN.Codegen/BSATN.Codegen.csproj" />
```
bfops, line 35: sanity checking my understanding - we removed `PrivateAssets="all"` because we (now) _do_ want the contents of `BSATN.Codegen.csproj` to be available to people who include `Codegen.csproj`, is that right?

crates/bindings-csharp/BSATN.Codegen/BSATN.Codegen.csproj
```diff
@@ -31,13 +33,6 @@
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.3.1" PrivateAssets="all" />
   </ItemGroup>
 
-  <ItemGroup>
-    <None Include="../logo.png" Pack="true" PackagePath="" />
-    <None Include="../Codegen/README.md" Pack="true" PackagePath="" />
-    <None Include="../LICENSE" Pack="true" PackagePath="" />
-    <None Include="$(OutputPath)/$(AssemblyName).dll" Pack="true" PackagePath="analyzers/dotnet/cs" Visible="false" />
-  </ItemGroup>
```
bfops, line 39: Why did we get to remove these?

crates/bindings-csharp/Codegen/Codegen.csproj
```diff
@@ -23,38 +25,24 @@
...
+  <!-- We need a different hack to include BSATN.Codegen as a transitive analyzer dependency. https://github.com/dotnet/roslyn/issues/47275#issuecomment-685482999 -->
   <ItemGroup>
-    <None Include="../logo.png" Pack="true" PackagePath="" />
-    <None Include="README.md" Pack="true" PackagePath="" />
-    <None Include="../LICENSE" Pack="true" PackagePath="" />
-    <!-- source generators need to be put in a specific folder -->
-    <!-- see https://github.com/dotnet/roslyn/blob/main/docs/features/source-generators.cookbook.md#package-a-generator-as-a-nuget-package -->
-    <None Include="$(OutputPath)/$(AssemblyName).dll" Pack="true" PackagePath="analyzers/dotnet/cs" Visible="false" />
-    <!-- We want BSATN.Codegen to be an internal dependency shipped in the same package as Codegen itself. -->
-    <None Include="$(OutputPath)/SpacetimeDB.BSATN.Codegen.dll" Pack="true" PackagePath="analyzers/dotnet/cs" Visible="false" />
```
bfops, line 46: Why did we get to remove these?

crates/bindings-csharp/BSATN.Runtime/BSATN.Runtime.csproj
bfops: another one - does this require any `IncludeRuntimeDependency="false"` like `Codegen.csproj` does below?

crates/bindings-csharp/Runtime/Runtime.csproj
```diff
@@ -41,6 +47,8 @@
     <None Include="build/" Pack="true" PackagePath="build" />
     <None Include="bindings.c" Pack="true" PackagePath="" />
     <None Include="driver.h" Pack="true" PackagePath="" />
+    <!-- We want all users who depends on Runtime to automatically get the Roslyn codegen component as well. -->
+    <None Include="../Codegen/bin/$(Configuration)/netstandard2.0/SpacetimeDB.Codegen.dll" Pack="true" PackagePath="analyzers/dotnet/cs" />
```
bfops, line 51: Again asking for my understanding, feel free to just say "trust me" if it's too much explaining - what's the difference between depending on the `Codegen.csproj` "directly" vs forcing it to build above and then depending on the `dll`?

smoketests/tests/csharp_module.py
RReverser: It's related to testing this PR, see commit description. https://github.com/clockworklabs/SpacetimeDB/pull/1440/commits/79bb911c056324fcecc87a30c352cbc5b0264159

crates/bindings-csharp/BSATN.Runtime/BSATN.Runtime.csproj
RReverser: I don't think so... but it's not really intentional, must have gotten it while changing forth and back and testing.

```suggestion
    <ProjectReference Include="../BSATN.Codegen/BSATN.Codegen.csproj" OutputItemType="Analyzer" PrivateAssets="all" ReferenceOutputAssembly="false" />
```

crates/bindings-csharp/BSATN.Runtime/BSATN.Runtime.csproj
RReverser: > Dumb question - this says `Pack="true"` but the project above says `<IsPackable>false</IsPackable>` - I assume those don't actually conflict despite kinda sounding like they would?

Yeah the idea here is that BSATN.Codegen should not be packable (aka publishable) as a separate NuGet package after this change, _but_ we explicitly pack the DLL it produces in the BSATN.Runtime package instead.

crates/bindings-csharp/Codegen/Codegen.csproj
RReverser: For consistency with version used in BSATN.Codegen. It's not a strictly required downgrade, but that version already has the APIs we need and is available in Unity, so why not.

crates/bindings-csharp/Runtime/Runtime.csproj
```diff
@@ -21,13 +21,19 @@
...
     <PackageLicenseFile>LICENSE</PackageLicenseFile>
   </PropertyGroup>
 
   <ItemGroup>
+    <!-- Note: BSATN.Codegen is included in the BSATN.Runtime NuGet package. -->
     <ProjectReference Include="../BSATN.Runtime/BSATN.Runtime.csproj" />
-    <ProjectReference Include="../BSATN.Codegen/BSATN.Codegen.csproj" OutputItemType="Analyzer" PrivateAssets="all" />
+    <!-- ...But we also need BSATN.Codegen to transform `[SpacetimeDB.Type]` usages in source of Runtime itself. -->
+    <!-- For that, it has to be included as an explicit private dependency. -->
+    <ProjectReference Include="../BSATN.Codegen/BSATN.Codegen.csproj" PrivateAssets="all" OutputItemType="analyzer" ReferenceOutputAssembly="false" />
```
bfops, line 33:
```suggestion
    <ProjectReference Include="../BSATN.Codegen/BSATN.Codegen.csproj" OutputItemType="Analyzer" PrivateAssets="all" ReferenceOutputAssembly="false" />
```

crates/bindings-csharp/Codegen/Codegen.csproj
bfops: Cool, ty. Leaving this visible in case other/future reviewers have the same question.

crates/bindings-csharp/BSATN.Runtime/BSATN.Runtime.csproj
bfops: Perfect, thanks!

smoketests/tests/csharp_module.py
bfops: oh neat! I had no idea.

crates/bindings-csharp/Codegen/Codegen.csproj
RReverser: It's just another one of "forth and back" changest that don't matter in the end.

Here it just doesn't matter one way or another, because `PrivateAssets="all"` only matters for NuGet packaging and neither Codegen nor BSATN.Codegen are packable after this PR, so having or removing that attribute has zero difference.

I'd probably keep it removed just to avoid someone asking why we need `PrivateAssets="all"` on a seemingly no-op package in the future.

crates/bindings-csharp/BSATN.Codegen/BSATN.Codegen.csproj
RReverser: Because BSATN.Codegen is no longer packable, and all of these are packing directives for extra files.

crates/bindings-csharp/Codegen/Codegen.csproj
RReverser: Same as above, Codegen is no longer packable so these are no longer necessary.

crates/bindings-csharp/Runtime/Runtime.csproj
RReverser: Depending "directly" results in a Runtime nuget package with a dependency on a Codegen nuget package, so we'd still need to pack and publish both on each release.

However, including the DLL of the Codegen package directly in the Runtime one simply "merges" them together so we only need to publish a single package instead.

crates/bindings-csharp/BSATN.Codegen/BSATN.Codegen.csproj
bfops: Ah yeah I see now they're still included in `Runtime.csproj`

bfops (approved):
To the best of my ability to understand these packages, this seems good to me. If it ends up breaking something, maybe we can add more tests.

crates/bindings-csharp/Directory.Build.props
```diff
@@ -0,0 +1,19 @@
...
+    <RepositoryType>git</RepositoryType>
+    <RepositoryUrl>https://github.com/clockworklabs/SpacetimeDB</RepositoryUrl>
+    <PackageLicenseFile>LICENSE</PackageLicenseFile>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <None Include="$(MSBuildThisFileDirectory)/logo.png" Pack="true" PackagePath="" />
+    <None Include="$(MSBuildThisFileDirectory)/LICENSE" Pack="true" PackagePath="" />
+  </ItemGroup>
+</Project>
```
bfops, line 19: TIL this file! :+1:

crates/bindings-csharp/Codegen/Codegen.csproj
RReverser: I removed `PrivateAssets="all"` on other codegen references as well for consistency & to reduce noise.

kurtismullins (approved):
This is awesome, thank you @RReverser! 

I may need a hand to figure out which packages no longer need to be published when we do our releases (although it seems self-evident here) but overall, LGTM!

RReverser:
> I may need a hand to figure out which packages no longer need to be published when we do our releases (although it seems self-evident here)

I documented that in the 2nd part of the PR description, but let me know if you need anything else.

bfops:
These codegen changes broke SpacetimeDB `master`, because the updated NuGet packages aren't published. We probably should have done these things separately.

PR #1441, opened 2024-06-17
Don't run bot tests on every master commit whoops

kazimuth:
# Description of Changes

We may want to do this once we have frozen ABI and we can be guaranteed it won't break on ABI changes, but we don't have that yet.

0

### DIFF
--- a/.github/workflows/bottest.yml
+++ b/.github/workflows/bottest.yml
@@ -1,7 +1,6 @@
 on:
   push:
     branches:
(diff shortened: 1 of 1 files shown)


bfops (approved):
Seems right to me, assuming we're fine with having this run _only_ when people request it. It's different from how we do the benchmarks etc but I agree there are different trade-offs with this one being much more likely to get broken.

PR #1442, opened 2024-06-17
Dedupe Spacetime ABI namespace in MSBuild

RReverser:
# Description of Changes

Some MSBuild magic to dedupe and centralise the ABI string in C#. After this change you should only update it in 1 place instead of 3, hopefully preventing issues like #1311 from reappearing.

corresponding GitHub label.

2 - simple changes, but since it's MSBuild, there's always a risk I broke something.

# Testing


RReverser:
Heh clearly I at least broke builds on Linux, even though it works fine on Windows. Marking as draft for now.

PR #1447, opened 2024-06-19
Bump version to 0.11.0

bfops:
# Description of Changes

https://github.com/clockworklabs/SpacetimeDB/pull/1440 has merged, requiring an upgrade to 0.11.0.

No

1

# Testing

None

### DIFF
--- a/crates/cli/src/subcommands/project/csharp/StdbModule._csproj
+++ b/crates/cli/src/subcommands/project/csharp/StdbModule._csproj
@@ -8,7 +8,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="SpacetimeDB.Runtime" Version="0.10.*" />
+    <PackageReference Include="SpacetimeDB.Runtime" Version="0.11.*" />
   </ItemGroup>
 
 </Project>
--- a/crates/cli/src/subcommands/project/rust/Cargo._toml
+++ b/crates/cli/src/subcommands/project/rust/Cargo._toml
@@ -9,5 +9,5 @@ edition = "2021"
 crate-type = ["cdylib"]
 
 [dependencies]
-spacetimedb = "0.10.1"
+spacetimedb = "0.11.0"
 log = "0.4"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -767,7 +767,7 @@ dependencies = [
 
 [[package]]
 name = "connect_disconnect_client"
-version = "0.10.1"
+version = "0.11.0"
 dependencies = [
  "anyhow",
  "spacetimedb-sdk",
@@ -4150,7 +4150,7 @@ dependencies = [
 
 [[package]]
 name = "spacetimedb"
-version = "0.10.1"
+version = "0.11.0"
 dependencies = [
  "derive_more",
  "getrandom",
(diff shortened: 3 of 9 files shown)


crates/cli/src/subcommands/project/csharp/StdbModule._csproj
```diff
@@ -8,7 +8,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="SpacetimeDB.Runtime" Version="0.10.*" />
+    <PackageReference Include="SpacetimeDB.Runtime" Version="0.11.*" />
```
bfops, line 11: @RReverser I guess we should only bump this when we're about to publish to NuGet, right? If we don't bump this now, will devs using `spacetime init --lang csharp` have issues?

crates/cli/src/subcommands/project/rust/Cargo._toml
```diff
@@ -9,5 +9,5 @@ edition = "2021"
 crate-type = ["cdylib"]
 
 [dependencies]
-spacetimedb = "0.10.1"
+spacetimedb = "0.11.0"
```
bfops, line 12: Ah this fetches from crates.io, right? I guess this is one we shouldn't bump until we publish?

Or we should encourage devs to have local overrides for upcoming versions? (Since presumably you can't use updated codegen while pointing to old libraries?)

crates/cli/src/subcommands/project/csharp/StdbModule._csproj
RReverser: They shouldn't. As long as releases are ABI compatible - and I believe 0.11 is - staying on the older version is safe.

crates/cli/src/subcommands/project/rust/Cargo._toml
RReverser: Yup, I would say let's not bump those until we publish.

RReverser:
I'd like to squeeze in one more change / refactoring I'm currently working on (close-ish to finishing) in this release if it doesn't have to go out urgently.

RReverser:
Alternatively, we can keep this for 0.12 to lump with all other breaking changes I guess?

crates/cli/src/subcommands/project/csharp/StdbModule._csproj
bfops: I think I'm asking the other way around - if we bump this now, then anybody internally using a recently-built `spacetime` binary (off `master`) will generate a project that references nonexistent NuGet packages, right?

RReverser: Yes, sorry, that's what I meant - staying on the older version is _safer_ than switching to the new one too early.

bfops: Yeah sorry I think I misread.. or something..

bfops:
> I'd like to squeeze in one more change / refactoring I'm currently working on (close-ish to finishing) in this release if it doesn't have to go out urgently.

Sure, I don't think the 0.11 release is imminent. I'm bumping the version number because we have already merged breaking changes, so we are certainly not a 0.10.* version anymore.

RReverser:
> I'd like to squeeze in one more change / refactoring I'm currently working on (close-ish to finishing) in this release if it doesn't have to go out urgently.

FWIW https://github.com/clockworklabs/SpacetimeDB/pull/1455 is the one I had in mind - it's a fairly large breaking change, so thought might be good to lump them together and use 0.11 for basically public testing.

PR #1449, opened 2024-06-19
Timer Table Implementation

Shubham8287:
# Description of Changes
Host side implementation of scheduled tables as per proposal - https://github.com/clockworklabs/SpacetimeDBPrivate/blob/master/proposals/0005-timers.md

yes

3

# Testing
Tested via running through example apps.


crates/sats/src/db/def.rs
```diff
@@ -485,6 +485,7 @@ pub struct TableSchema {
     pub sequences: Vec<SequenceSchema>,
     pub table_type: StTableType,
     pub table_access: StAccess,
+    pub scheduled: Option<Box<str>>,
```
kazimuth, line 488: In the PR #1448 I'm rewriting Def. My plan was to add:

```rust
struct ScheduleDef {
    reducer_name: Box<str>,
    at_column: ColId
}
```

And then having an `Option<ScheduleDef>` on `TableDef`.
If you don't mind restructuring this that way it would make it easier for me down the line.

crates/core/src/host/scheduler.rs
```diff
@@ -249,52 +264,173 @@ impl SchedulerActor {
...
+                                identity,
+                                None,
+                                None,
+                                None,
+                                None,
+                                &reducer.reducer,
+                                ReducerArgs::Bsatn(reducer.bsatn_args.into()),
+                            )
+                            .await;
+                        // Delete the entry from table if it is not repeated schedule in new tx
```
kim, line 293: You may be aware of this: the check for `Err(NoSuchModule(_))` in the old code is quite important, in case the module host is exited in between the upgrade of the weak reference succeeding and the actual call. We want to re-schedule once the new module host is instantiated.

crates/core/src/host/scheduler.rs
```diff
@@ -249,52 +264,173 @@ impl SchedulerActor {
...
+    let mut tx = db.begin_mut_tx(IsolationLevel::Serializable);
+    let ctx = ExecutionContext::internal(db.address());
+
+    let schedule_id = AlgebraicValue::from(id.schedule_id);
+    let table_id = id.table_id;
+
+    let schedule_id_pos = db
+        .schema_for_table_mut(&tx, table_id)?
+        .get_column_id_by_name(SCHEDULE_ID_FIELD)
+        .ok_or(anyhow!("SCHEDULE_ID_FIELD not found"))?;
```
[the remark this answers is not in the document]
kim: Can we have a workspace-level clippy conf? Clippy tends to think strict is better, and I think this lint should just be turned off.

crates/sats/src/db/def.rs
Shubham8287: How do you translate this in `TableSchema`?

crates/core/src/db/datastore/locking_tx_datastore/state_view.rs
```diff
@@ -140,6 +140,15 @@ pub(crate) trait StateView {
...
             .collect::<Result<Vec<_>>>()?;
 
+        let scheduled = self
+            .iter_by_col_eq(ctx, ST_SCHEDULED_ID, StScheduledFields::TableId, value_eq)?
+            .next()
+            .map(|row| -> Result<_> {
+                let row = StScheduledRow::try_from(row)?;
+                Ok(row.reducer_name)
+            })
+            .transpose()?;
```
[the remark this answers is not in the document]
Shubham8287: ummm, `scheduled` is supposed to be an `Option` enum, provided suggestion does not looks to evaluates to that.

crates/core/src/host/wasm_common.rs
```diff
@@ -359,7 +359,6 @@ macro_rules! abi_funcs {
             "spacetime_8.0"::buffer_alloc,
             "spacetime_8.0"::buffer_consume,
             "spacetime_8.0"::buffer_len,
```
Shubham8287, line 361: What semantic do we use here? should this be updated to 9.0 now?

crates/core/src/host/scheduler.rs
```diff
@@ -249,52 +224,248 @@ impl SchedulerActor {
...
+                                // the ScheduledReducer in the database for when the module restarts
+                                // and also not delete it if it's a repeated schedule
+                                if matches!(res, Err(ReducerCallError::NoSuchModule(_))) && !is_repeated {
+                                    //`call_reducer` above moves the tx, so its safe to create new here
+                                    let _ = db
+                                        .with_auto_commit(&ctx, |tx| {
+                                            db.delete(tx, id.table_id, [schedule_row_ptr]);
+                                            Ok::<(), anyhow::Error>(())
+                                        })
+                                        .map_err(|e| log::error!("deleting scheduled reducer failed: {e:#}"));
```
kim, line 274: Strangely, the types are lying -- this transaction cannot actually fail.

But if it could, we might run the scheduler again in the future, is that correct?

I think it's important to note both in the proposal and user-facing documentation that scheduled reducers have "at least once" semantics.

crates/core/src/host/scheduler.rs
```diff
@@ -249,52 +224,248 @@ impl SchedulerActor {
...
+    let schedule_id = row.field_as_u64(schedule_id_col_pos, SCHEDULED_ID_FIELD.into())?;
+
+    let schedule_at_ab = row.get_field(schedule_at_col_pos, SCHEDULED_ID_FIELD.into())?;
+    let schedule_at = ScheduleAt::try_from(schedule_at_ab.clone()).map_err(|e| {
+        anyhow!(
+            "Failed to convert field '{}' to ScheduleAt: {:?}",
+            SCHEDULED_AT_FIELD,
+            e
+        )
+    })?;
```
kim, line 420:
```suggestion
    let schedule_at = ScheduleAt::try_from(schedule_at_ab.clone()).with_context(|| {
        format!(
            "Failed to convert field '{}' to ScheduleAt",
            SCHEDULED_AT_FIELD,
        )
    })?;
```

kim (commented):
Looks good overall.
I'll want to do the manual testing before approving.

crates/core/src/host/scheduler.rs
Shubham8287: But does that also means that there is a bug in our logic?

kim: I think it means that, like in many instances in the codebase, we formulate types not to express what is the case, but what might be the case in the future.

In this case: transactions may become fallible under MVCC, if and when that is supported, but they are, in fact, not.

If we resist debating the code structure, the possibility of this particular transaction being fallible means that we may run the scheduled reducer more than once.

I think that’s fine, but we need to be diligent in documenting the behavior.

crates/bindings-macro/src/lib.rs
```diff
@@ -436,13 +435,80 @@ struct Column<'a> {
...
+    let reducer_name = reducer.to_string();
+    let mut modified_item = syn::parse2::<DeriveInput>(item)?;
+    let type_check = reducer_type_check(&modified_item, &reducer)?;
+    add_scheduled_fields(&mut modified_item)?;
+    // codegen call all the `__preinit`s to add it to module Typespace
+    // codegen does not try to replace ref types with their product types.
+    // hence known types get registered as ref types.
+    let ref_register = quote! {
+        const _: () = {
+            #[export_name = "__preinit__20_register_describer_ScheduleAt"]
```
kim, line 463: This means that there can only be a single scheduled reducer per stdb module.

What would speak against adding "known types" to the typespace by default? 
`__describe_module__()` stands out as a place to do that.

cc @Centril @gefjon

crates/core/src/host/scheduler.rs
```diff
@@ -249,52 +226,247 @@ impl SchedulerActor {
...
+        let ctx = ExecutionContext::internal(db.address());
+        let caller_identity = module_host.info().identity;
+        let module_info = module_host.info.clone();
+
+        let call_reducer_params = move |tx: &MutTxId| -> Result<Option<CallReducerParams>, anyhow::Error> {
+            let schedule_row = match get_schedule_row_mut(&ctx, tx, &db, id) {
+                Ok(schedule_row) => schedule_row,
+                // if the row is not found, it means the schedule is cancelled by the user
+                Err(_) => {
+                    log::info!(
```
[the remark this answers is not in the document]
Shubham8287: Did you put it inside `init`?

crates/bindings-macro/src/lib.rs
Shubham8287: I think the problem is because `ScheduleAt` ref regestering is happening for all scheduled tables.
On anoher note, code generation is working fine without doing this as all. May be, `make_type` of `ScheduleAt` getting called early enough through some way. Still checking.

Shubham8287: It is because `ScheduleAt` is always a field of a struct, and parents `preinit` register describer defined for parent calls it `make_type`.

crates/core/src/host/scheduler.rs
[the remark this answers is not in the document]
kim: Yes

crates/bindings-macro/src/lib.rs
kim: Ok, so we don't actually need this preinit at all?

Shubham8287: Yes, I removed it testing generating and publishing module. Even by raw printing module typespace with and without this `preinit` they are exactly same.

crates/sats/src/db/def.rs
kazimuth: I guess just assert that `at_column` is named `at` and convert the reducer name.

crates/core/src/host/wasm_common.rs
cloutiertyler: This is a breaking change and should probably be set to 9.0

crates/core/src/db/datastore/locking_tx_datastore/state_view.rs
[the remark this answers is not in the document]
cloutiertyler: Given that @Centril did not provide more context I would mark this as resolved.

crates/core/src/host/wasm_common.rs
Shubham8287: yeah, I have already done that

Shubham8287:
new commits after last update are 
- Few times rebasing with master.
- merging #1459, which was required for smoketests to pass.
- Simplify Timestamp usage which was conflicting with protobufectomy.
All of them seems trivial or not to be concerned with PR main logic, hence didn't seek new approval.

PR #1459, opened 2024-06-26
[DO NOT MERGE] Timer tables implementation for C#

RReverser:
**[DO NOT MERGE]** - this PR is currently created against https://github.com/clockworklabs/SpacetimeDB/pull/1455 instead of `master` as its base.

This is done in order to make diff readable and exclude the diff of the larger PR here, but it means that attempting to merge this PR will result in it merging into the wrong branch.

Once https://github.com/clockworklabs/SpacetimeDB/pull/1455 is merged, this PR should be simply retargeted against master and only then merged.

# Description of Changes

C# implementation of the new scheduling API via timer tables + removal of the old one.

Removes the old scheduling API.

corresponding GitHub label.

2

# Testing

- [x] `dotnet test`
- [ ] *Write a test you want a reviewer to do here, so they can check it off when they're satisfied.* Would be good to add C# SDK tests for this as well.


Shubham8287:
This PR has been merged to master via https://github.com/clockworklabs/SpacetimeDB/pull/1449. To pass the CI, it was needed to merge Csharp and rust bindings together.

I manually tested this PR via publishing module, and scheduling a reducer before merging.

PR #1462, opened 2024-06-26
Ensure MacOS builds from CI are executable

kurtismullins:
# Description of Changes

The `spacetime` binary in the MacOS build's tarball is not flagged as executable.

This makes it executable.

n/a

It only affects future release automation. I assume there is next to no risk.

# Testing

Untested as this specific CI is only ran on `master`, `release` branches, and tags.


bfops (approved):
hard to argue with that!

PR #1466, opened 2024-06-28
Protobufectomy: c# codegen

lcodes:
# Description of Changes

C# side of the rust protobufectomy. Updates the codegen to reflect matching changes to the C# client SDK.

corresponding GitHub label.

# Testing


crates/cli/src/subcommands/generate/csharp.rs
```diff
@@ -647,7 +652,7 @@ pub fn autogen_csharp_globals(items: &[GenItem], namespace: &str) -> Vec<(String
         );
         writeln!(
             output,
-            "public ReducerEvent(ClientApi.Event dbEvent, IReducerArgs? args) : base(dbEvent) => Args = args;"
+            "public ReducerEvent(TransactionUpdate update, IReducerArgs? args) : base(update) => Args = args;"
```
jdetter, line 655: Either we need to add `using SpacetimeDB.ClientApi;` to the top of this file or you need to fully qualify this path because it causes a compilation issue:

```
Assets/_Project/autogen/_Globals/SpacetimeDBClient.cs(515,23): error CS0246: The type or namespace name 'TransactionUpdate' could not be found (are you missing a using directive or an assembly reference?)
```

jdetter (changes requested):
Small nit that needs to be addressed - I'm still testing right now

crates/cli/src/subcommands/generate/csharp.rs
```diff
@@ -312,8 +312,8 @@ fn autogen_csharp_product_table_common(
         ],
     );
 
-    writeln!(output, "[SpacetimeDB.Type]");
     writeln!(output, "[DataContract]");
+    writeln!(output, "[SpacetimeDB.Type]");
```
jdetter, line 316: Please revert this transposition

crates/cli/src/subcommands/generate/csharp.rs
```diff
@@ -343,11 +345,14 @@ fn autogen_csharp_product_table_common(
                 field_name.to_case(Case::Pascal),
                 default_init(ctx, &field.algebraic_type)
             );
+            if i + 1 != num_fields {
+                writeln!(output);
```
jdetter, line 349: This is causing an additional space in between each field, do we need this change? This is causing the diff for the API upgrade to be pretty insane: https://github.com/clockworklabs/BitCraftClient/pull/3586/files

jdetter (changes requested):
Easy to fix whitespace nits

crates/cli/src/subcommands/generate/csharp.rs
lcodes: Oh right, I added this to make the emitted fields more readable, as the lines can get long. It's not a needed change tho, purely style. Should we keep or remove it?

jdetter: please remove it, if you feel strongly about the style change we can do it in a separate PR IMO - I just want the API upgrade to be as simple as possible.

crates/cli/src/subcommands/generate/csharp.rs
jdetter: I believe this still doesn't compile. When I run the autogen for the BitCraft module I still end up with this compilation error:

image

```
Assets/_Project/autogen/_Globals/SpacetimeDBClient.cs(524,23): error CS0246: The type or namespace name 'TransactionUpdate' could not be found (are you missing a using directive or an assembly reference?)
```

jdetter (changes requested):
Let me know when you want me to take another look

lcodes:
Moving contents of this into noa's branch

PR #1474, opened 2024-07-02
`ColList`: preserve list order on `list.push(..)`

Centril:
# Description of Changes

Preserve the list order of `ColList` for e.g., `col_list![2, 0]` by heapifying in those cases.
This might not be the optimal performance solution, but this fixes the bug while preserving that `size_of::<ColList> == 8` and that it is still inline in the common case.

Rather unlikely that anyone relied on this bug.

2, some unsafe and someone might have relied on this bug.

# Testing

- Tests are amended to ensure that the list order is preserved.


kazimuth (approved):
Ha! Brilliant.  LGTM 👍 

(I was going to suggest sorting inputs in the FromIterator impl, but then I realized that would defeat the purpose of this PR... it would make sense if you were implementing From<HashSet<ColId>>, but there's no iterator equivalent of that.)

PR #1475, opened 2024-07-02
Fix C# module smoketests

bfops:
# Description of Changes

https://github.com/clockworklabs/SpacetimeDB/pull/1440 restructured our NuGet packages, and correspondingly updated the `.csproj` files we generate. However, since we haven't actually published the restructured NuGet packages, `.csproj`s expecting the _new_ NuGet package structure will be broken, including the smoketests.

It seems like this behavior was _sometimes_ causing smoketests to fail. It's unclear why it's only sometimes, but our best guess is "something something caching", so this PR forcibly clears the NuGet package cache.

Additionally, we bump the versions of the local NuGet packages to 0.11, so that they cannot be found on the NuGet servers. This means that the smoketests can _only_ find them in local directories, so they do not get confused about where to get these packages.

No

# Testing

- [x] CI passes
- [x] `spacetime init --lang csharp` followed by `spacetime build` works successfully
- [x] smoketests pass when merged with https://github.com/clockworklabs/SpacetimeDB/pull/1459 in https://github.com/clockworklabs/SpacetimeDB/pull/1476

### DIFF
--- a/smoketests/tests/csharp_module.py
+++ b/smoketests/tests/csharp_module.py
@@ -22,25 +22,6 @@ class CreateProject(unittest.TestCase):
 
             with tempfile.TemporaryDirectory() as tmpdir:
                 spacetime("init", "--lang=csharp", tmpdir)
-
-                packed_projects = ["BSATN.Runtime", "Runtime"]
-                restore_sources = [str(bindings / project / "bin" / "Release") for project in packed_projects]
-                # note that nuget URL comes last, which ensures local sources should override it.
-                restore_sources.append("https://api.nuget.org/v3/index.json")
-
-                csproj = Path(tmpdir) / "StdbModule.csproj"
-                with open(csproj, "r") as f:
-                    contents = f.read()
-
-                contents = contents.replace(
-                    "</PropertyGroup>",
-                    # note that nuget URL comes last, which ensures local sources should override it.
-                    f"""<RestoreSources>{str.join(";", restore_sources)}</RestoreSources>
-</PropertyGroup>""",
-                )
-                with open(csproj, "w") as f:
-                    f.write(contents)
-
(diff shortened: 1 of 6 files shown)


Shubham8287 (approved):
Gets the job done!

Shubham8287:
Oh wait, The PR in description fails with this error - 
```
/tmp/tmp6wbnwk4t/Lib.cs(8,29): error CS0103: The name 'ColumnAttrs' does not exist in the current context [/tmp/tmp6wbnwk4t/StdbModule.csproj]
```
Is that means, error is specific to those PRs?

smoketests/tests/csharp_module.py
```diff
@@ -17,6 +17,7 @@ def test_build_csharp_module(self):
 
         try:
 
+            run_cmd("dotnet", "nuget", "locals", "all", "--clear", cwd=bindings, capture_stderr=True)
```
bfops, line 20: clear package cache. this isn't required to make the CI work again, but there did seem to be some kind of caching happening because the smoketests weren't passing or failing consistently. This hopefully makes them more consistent.

bfops: (we think this also resolves a hypothetical issue where multiple open PRs make non-backwards-compatible changes to the NuGet packages)

Shubham8287 (approved):
This feels like a good mitigation to fix the master. Thanks for working on this.
But I think there is still a need to have a general solution so that we don't run into the problem in future (not to be done by you or under the scope of this PR).

bfops:
> This feels like a good mitigation to fix the master. Thanks for working on this. But I think there is still a need to have a general solution so that we don't run into the problem in future (not to be done by you or under the scope of this PR).

The general solution is to bump the package versions when there are breaking changes

PR #1477, opened 2024-07-03
Add support for `I256` and `U256`

### LINKED ISSUE #1097: Add `AlgebraicType/Value::U256`
Many cryptographic hashes result in 32 byte (256 bit) outputs, and storing this as a `Box<U256>` is about as cheap as we can make it. This new variant/type would be helpful for e.g., `Identity`.

Centril:
# Description of Changes

Fixes #1097.

Implements support for `256`-width integers in SATS and friends.
The integer types are backed by the crate `ethnum` which provides support as-if they were primitive Rust types.

Breaks the BSATN-encoding of `AlgebraicType`.

2 -- some subtle considerations:
- sequences (supported, but can at most become `u128::MAX` for e.g., `u256`)

# Testing

Unit tests and SDK tests have been amended.

### DIFF
--- a/crates/bindings-csharp/BSATN.Codegen/Utils.cs
+++ b/crates/bindings-csharp/BSATN.Codegen/Utils.cs
@@ -131,8 +131,10 @@ public static class Utils
             var result = type.OriginalDefinition.ToString() switch
             {
                 // {U/I}{128/256} are not treated by C# as regular primitives, so we need to match them by type name.
-                "System.I128" or "SpacetimeDB.I128" => "SpacetimeDB.BSATN.I128",
-                "System.U128" or "SpacetimeDB.U128" => "SpacetimeDB.BSATN.U128",
+                "System.Int128" => "SpacetimeDB.BSATN.I128",
+                "System.UInt128" => "SpacetimeDB.BSATN.U128",
+                "SpacetimeDB.I128" => "SpacetimeDB.BSATN.I128Stdb",
+                "SpacetimeDB.U128" => "SpacetimeDB.BSATN.U128Stdb",
                 "SpacetimeDB.I256" => "SpacetimeDB.BSATN.I256",
                 "SpacetimeDB.U256" => "SpacetimeDB.BSATN.U256",
                 "System.Collections.Generic.List<T>" => $"SpacetimeDB.BSATN.List",
@@ -149,6 +151,7 @@ public static class Utils
                 result =
                     $"{result}<{string.Join(", ", type.TypeArguments.Select(SymbolToName).Concat(type.TypeArguments.Select(GetTypeInfo)))}>";
             }
+
             return result;
         }
     }
--- a/crates/bindings-csharp/BSATN.Runtime/BSATN/Runtime.cs
+++ b/crates/bindings-csharp/BSATN.Runtime/BSATN/Runtime.cs
@@ -175,7 +175,7 @@ public readonly struct U64 : IReadWrite<ulong>
         new AlgebraicType.U64(default);
 }
 
-public readonly struct U128 : IReadWrite<SpacetimeDB.U128>
+public readonly struct U128Stdb : IReadWrite<SpacetimeDB.U128>
 {
     public SpacetimeDB.U128 Read(BinaryReader reader) =>
         new(reader.ReadUInt64(), reader.ReadUInt64());
@@ -190,17 +190,34 @@ public readonly struct U128 : IReadWrite<SpacetimeDB.U128>
         new AlgebraicType.U128(default);
 }
 
+#if NET7_0_OR_GREATER
+public readonly struct U128 : IReadWrite<System.UInt128>
+{
+    public System.UInt128 Read(BinaryReader reader) =>
+        new(reader.ReadUInt64(), reader.ReadUInt64());
+
+    public void Write(BinaryWriter writer, System.UInt128 value)
+    {
+        writer.Write((ulong)(value >> 64));
+        writer.Write((ulong)value);
+    }
+
+    public AlgebraicType GetAlgebraicType(ITypeRegistrar registrar) =>
+        new AlgebraicType.U128(default);
+}
+#endif
+
 public readonly struct U256 : IReadWrite<SpacetimeDB.U256>
 {
     public SpacetimeDB.U256 Read(BinaryReader reader)
     {
-        var bsatn = new U128();
+        var bsatn = new U128Stdb();
         return new(bsatn.Read(reader), bsatn.Read(reader));
     }
 
     public void Write(BinaryWriter writer, SpacetimeDB.U256 value)
     {
-        var bsatn = new U128();
+        var bsatn = new U128Stdb();
         bsatn.Write(writer, value.Upper);
         bsatn.Write(writer, value.Lower);
     }
@@ -249,7 +266,7 @@ public readonly struct I64 : IReadWrite<long>
         new AlgebraicType.I64(default);
 }
 
-public readonly struct I128 : IReadWrite<SpacetimeDB.I128>
+public readonly struct I128Stdb : IReadWrite<SpacetimeDB.I128>
 {
     public SpacetimeDB.I128 Read(BinaryReader reader) =>
         new(reader.ReadUInt64(), reader.ReadUInt64());
@@ -264,17 +281,33 @@ public readonly struct I128 : IReadWrite<SpacetimeDB.I128>
         new AlgebraicType.I128(default);
 }
 
+#if NET7_0_OR_GREATER
+public readonly struct I128 : IReadWrite<System.Int128>
+{
+    public System.Int128 Read(BinaryReader reader) => new(reader.ReadUInt64(), reader.ReadUInt64());
+
+    public void Write(BinaryWriter writer, System.Int128 value)
+    {
+        writer.Write((long)(value >> 64));
+        writer.Write((long)value);
+    }
+
+    public AlgebraicType GetAlgebraicType(ITypeRegistrar registrar) =>
+        new AlgebraicType.I128(default);
+}
+#endif
+
 public readonly struct I256 : IReadWrite<SpacetimeDB.I256>
 {
     public SpacetimeDB.I256 Read(BinaryReader reader)
     {
-        var bsatn = new U128();
+        var bsatn = new U128Stdb();
         return new(bsatn.Read(reader), bsatn.Read(reader));
     }
 
     public void Write(BinaryWriter writer, SpacetimeDB.I256 value)
     {
-        var bsatn = new U128();
+        var bsatn = new U128Stdb();
         bsatn.Write(writer, value.Upper);
         bsatn.Write(writer, value.Lower);
     }
--- a/crates/bindings-csharp/Runtime/Internal/Module.cs
+++ b/crates/bindings-csharp/Runtime/Internal/Module.cs
@@ -2,7 +2,6 @@ namespace SpacetimeDB.Internal;
 
 using SpacetimeDB;
 using SpacetimeDB.BSATN;
-using I128 = SpacetimeDB.I128;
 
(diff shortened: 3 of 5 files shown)


crates/bindings-csharp/BSATN.Runtime/BSATN/Int128.cs
```diff
@@ -0,0 +1,128 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+#if !NET7_0_OR_GREATER
+using System.Diagnostics.CodeAnalysis;
+using System.Runtime.InteropServices;
+
+namespace System
```
RReverser, line 8: This would be nice, but might be problematic if another package that user depends on declares its own `System.Int128` polyfill. It's ok to declare `internal` System types for things like marker attributes, but public types can lead to all sorts of conflicts.

I think we should keep this type under our own `SpacetimeDB` namespace.

crates/bindings-csharp/BSATN.Runtime/BSATN.Runtime.csproj
```diff
@@ -8,7 +8,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <!-- Note: the source code of this package is used in Unity too, which is limited to C# 9 and .NET Standard 2.1. -->
+    <!-- Note: the source code of this package is used in Unity too, which is limited to C# 11 and .NET Standard 2.1. -->
```
RReverser, line 11: Good catch, this comment is outdated as of recently. It should be something like this though:

```suggestion
    <!-- Note: the binary produced by this package is used in Unity too, which is limited to .NET Standard 2.1. -->
```

crates/bindings-csharp/BSATN.Codegen/Utils.cs
```diff
@@ -133,6 +133,8 @@ static string GetTypeInfoForNamedType(INamedTypeSymbol type)
                 // (U)Int128 are not treated by C# as regular primitives, so we need to match them by type name.
                 "System.Int128" => "SpacetimeDB.BSATN.I128",
                 "System.UInt128" => "SpacetimeDB.BSATN.U128",
+                "SpacetimeDB.Int256" => "SpacetimeDB.BSATN.I256",
```
RReverser, line 136: For custom types it's better to inline the `struct BSATN` implementation right inside the type instead of augmenting the codegen.

The only reason this codegen has a large switch is because C# doesn't have Rust-like traits that would allow us to declare BSATN implementations on built-in types - but where types are not built-in, we don't need to do this.

crates/bindings-csharp/BSATN.Runtime/BSATN/Int256.cs
```diff
@@ -0,0 +1,134 @@
...
+
+using System;
+using System.Diagnostics.CodeAnalysis;
+using System.Runtime.InteropServices;
+
+namespace SpacetimeDB
+{
+    /// <summary>Represents a 256-bit signed integer.</summary>
+    [StructLayout(LayoutKind.Sequential)]
+    public readonly struct Int256 : IEquatable<Int256>
```
RReverser, line 12: Might want to add `IComparable<Int256>` here and to the other types.

crates/bindings-csharp/BSATN.Runtime/BSATN/Int256.cs
```diff
@@ -0,0 +1,134 @@
...
+        }
+
+        /// <inheritdoc cref="IEquatable{T}.Equals(T)" />
+        public bool Equals(Int256 x) => _upper == x._upper && _lower == x._lower;
+
+        /// <inheritdoc cref="object.GetHashCode()" />
+        public override int GetHashCode() => HashCode.Combine(_lower, _upper);
+
+        /// <inheritdoc cref="object.ToString()" />
+        public override string ToString() => $"Int256({_upper},{_lower})";
```
RReverser, line 132: Do we want to provide proper ToString() maybe? Should be easy to do via BigInteger - not the most efficient method perhaps, but should be fine for stringification.

crates/bindings-csharp/BSATN.Runtime/BSATN/Int256.cs
```diff
@@ -0,0 +1,134 @@
...
+            _lower = lower;
+            _upper = upper;
+        }
+
+        internal UInt128 Lower => _lower;
+
+        internal UInt128 Upper => _upper;
+
+        /// <inheritdoc cref="IComparable.CompareTo(object)" />
+        public int CompareTo(object? value)
```
RReverser, line 38: I think these all implementations could be significantly simplified if you declared this type as a wrapper around `(UInt128 Upper, UInt128 Lower)` tuple field instead. In C# tuples automatically implement `CompareTo`, equality and `GetHashCode`, so you'd be able to simply forward all those methods to the field instead of reimplementing them from scratch.

RReverser: Signed types might need a bit more care though, not sure.

crates/bindings-csharp/BSATN.Codegen/Utils.cs
RReverser: In fact, you don't even need to provide one manually - just use lower-endian field order and then you can use `[SpacetimeDB.Type]` for automatic read/write generation on those types.

RReverser: I'd just really like to get rid of this hardcoded `switch` in the future - e.g. one idea I had was just defining a `SystemShims` namespace that would be used for any types from the `System` namespace - and adding non-`System` types kinda interferes with that.

I can remove it when I get to it, but I'd prefer not to have non-System types in here to begin with.

crates/bindings-csharp/BSATN.Codegen/Utils.cs
```diff
@@ -131,8 +131,10 @@ static string GetTypeInfoForNamedType(INamedTypeSymbol type)
             var result = type.OriginalDefinition.ToString() switch
             {
                 // (U)Int128 are not treated by C# as regular primitives, so we need to match them by type name.
-                "System.Int128" => "SpacetimeDB.BSATN.I128",
-                "System.UInt128" => "SpacetimeDB.BSATN.U128",
+                "SpacetimeDB.Int128" => "SpacetimeDB.BSATN.I128",
+                "SpacetimeDB.UInt128" => "SpacetimeDB.BSATN.U128",
```
RReverser, line 135: Both need to be checked here, or it will break existing `System.Int128` support in C# modules.
```suggestion
                "System.Int128" or "SpacetimeDB.Int128" => "SpacetimeDB.BSATN.I128",
                "System.UInt128" or "SpacetimeDB.UInt128" => "SpacetimeDB.BSATN.U128",
```

crates/bindings-csharp/BSATN.Runtime/BSATN/I128.cs
```diff
@@ -0,0 +1,135 @@
...
+using System.Runtime.InteropServices;
+
+/// <summary>Represents a 128-bit signed integer.</summary>
+[StructLayout(LayoutKind.Sequential)]
+public readonly struct I128 : IEquatable<I128>, IComparable, IComparable<I128>
+{
+    internal const int Size = 16;
+
+#if BIGENDIAN
+        private readonly ulong _upper;
```
RReverser, line 16: Nit: broken formatting, probably after changing the namespace.

Unfortunately https://github.com/clockworklabs/SpacetimeDB/pull/1567 is still waiting for review, but can you meanwhile install CSharpier and run `dotnet csharpier .` in `crates/bindings-csharp`? That way we have less code to fix once CI check lands.

crates/bindings-csharp/Runtime/Internal/Module.cs
```diff
@@ -41,20 +43,20 @@ public partial struct ConstraintDef(string name, ColumnAttrs kind, uint[] column
     public partial struct SequenceDef(
         string sequenceName,
         uint colPos,
-        Int128? increment = null,
-        Int128? start = null,
-        Int128? min_value = null,
-        Int128? max_value = null,
-        Int128? allocated = null
+        I128? increment = null,
```
RReverser, line 46: This shouldn't be necessary as module runtime only targets latest .NET. Let's keep native Int128 here instead rather than the shim.

crates/bindings-csharp/Runtime/Runtime.csproj
```diff
@@ -11,6 +11,7 @@
     <TargetFramework>net8.0</TargetFramework>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <RootNamespace>SpacetimeDB</RootNamespace>
+    <Nullable>enable</Nullable>
```
RReverser, line 14: This is already enabled for all projects in `Directory.Build.props`, not necessary here.

crates/bindings-csharp/BSATN.Runtime/BSATN/Int256.cs
```diff
@@ -1,134 +1,127 @@
...
-    /// <summary>Represents a 256-bit signed integer.</summary>
-    [StructLayout(LayoutKind.Sequential)]
-    public readonly struct Int256 : IEquatable<Int256>
-    {
-        internal const int Size = 32;
+    internal const int Size = 32;
 
 #if BIGENDIAN
         private readonly UInt128 _upper;
         private readonly UInt128 _lower;
```
lcodes: Autoformat missed the inside of preprocessor conditionals?

crates/bindings-csharp/BSATN.Runtime/BSATN/Runtime.cs
```diff
@@ -248,32 +241,25 @@ public void Write(BinaryWriter writer, UInt256 value)
     public AlgebraicType GetAlgebraicType(ITypeRegistrar registrar) => new AlgebraicType.I64(default);
 }
 
-public readonly struct I128 : IReadWrite<System.Int128>
+public readonly struct I128 : IReadWrite<SpacetimeDB.Int128>
```
lcodes: What happens now if devs declare a field as System.Int128?

crates/bindings-csharp/BSATN.Codegen/Utils.cs
```diff
@@ -130,11 +130,11 @@ static string GetTypeInfoForNamedType(INamedTypeSymbol type)
             }
             var result = type.OriginalDefinition.ToString() switch
             {
-                // (U)Int128 are not treated by C# as regular primitives, so we need to match them by type name.
-                "System.Int128" or "SpacetimeDB.Int128" => "SpacetimeDB.BSATN.I128",
-                "System.UInt128" or "SpacetimeDB.UInt128" => "SpacetimeDB.BSATN.U128",
-                "SpacetimeDB.Int256" => "SpacetimeDB.BSATN.I256",
-                "SpacetimeDB.UInt256" => "SpacetimeDB.BSATN.U256",
+                // {U/I}{128/256} are not treated by C# as regular primitives, so we need to match them by type name.
+                "System.I128" or "SpacetimeDB.I128" => "SpacetimeDB.BSATN.I128",
```
lcodes: System.I128?
System.U128?

crates/bindings-csharp/BSATN.Runtime/BSATN/Runtime.cs
```diff
@@ -175,7 +175,7 @@ public void Write(BinaryWriter writer, Inner? value)
         new AlgebraicType.U64(default);
 }
 
-public readonly struct U128 : IReadWrite<SpacetimeDB.U128>
+public readonly struct U128Stdb : IReadWrite<SpacetimeDB.U128>
```
lcodes, line 178: Why not call this one U128 and the other UInt128?
Or make both explicit by naming the other U128Sys?

lcodes (approved):

PR #1479, opened 2024-07-03
Move `db` module from `spacetimedb_sats` to `spacetimedb_lib`

kazimuth:
# Description of Changes

As well as the `relation` module.

Split this out of the schema rewrite, it is largely a find-replace change (except for one method that had to be renamed due to orphan rule).

This is an API-breaking change, but curiously not an ABI-breaking change.

1

# Testing

Tests and smoketests pass locally, except for the C# issues floating around. (e: this has now been fixed.)


crates/lib/src/db/def.rs
```diff
@@ -300,14 +292,13 @@ impl From<FieldDef<'_>> for ProductTypeElement {
...
-#[sats(crate = crate)]
 pub struct ColumnDef {
     pub col_name: Box<str>,
     pub col_type: AlgebraicType,
 }
 
-impl From<ProductType> for Vec<ColumnDef> {
-    fn from(value: ProductType) -> Self {
+impl ColumnDef {
+    pub fn from_product_type(value: ProductType) -> Vec<ColumnDef> {
```
kazimuth, line 301: This change was necessary due to the orphan rule (you can't impl From<T> for Vec<StructDefinedInThisCrate>)

mamcx (approved):
LGTM

cloutiertyler (approved):
This LGTM

PR #1482, opened 2024-07-08
CLI: Use `--server` and `--identity` args consistently instead of anonymous args

bfops:
# Description of Changes

Updates the CLI's `energy`, `identity`, `list`, and `server` subcommands to consistently use `--server`/`-s` for a server arg and `--identity`/`-i` for an identity arg.

All the subcommands altered by this PR used to use anonymous args for these.

Closes https://github.com/clockworklabs/SpacetimeDB/issues/1486.

This **breaks** the CLI API for all altered subcommands!

**This may necessitate updates to scripts that use the CLI**... but it's hard to figure out which ones until they break :smiling_face_with_tear: 

1

# Testing

- [x] CI passes

### DIFF
--- a/crates/cli/src/subcommands/identity.rs
+++ b/crates/cli/src/subcommands/identity.rs
@@ -1,4 +1,5 @@
 use crate::{
+    common_args,
     config::{Config, IdentityConfig},
     util::{init_default, y_or_n, IdentityTokenJson, InitDefaultResultType},
 };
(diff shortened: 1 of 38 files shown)


joshua-spacetime (approved):

jdetter (approved):
I haven't tested this myself but the code looks good. Another win for consistency, thanks Zeke 👍

crates/cli/src/subcommands/identity.rs
```diff
@@ -152,7 +152,7 @@ fn get_subcommands() -> Vec<Command> {
             ,
         Command::new("remove")
             .about("Removes a saved identity from your spacetime config")
-            .arg(Arg::new("identity")
+            .arg(common_args::identity()
```
bfops, line 155: revert this one

bfops: (nevermind - we thought this was the only param to `spacetime identity remove`, but it turns out this is not true)

PR #1487, opened 2024-07-09
Track typescript import statements in a btree to deduplicate them

jsdt:
# Description of Changes

When generating TS code, this tracks import statements in a set instead of a vector to avoid duplicates. With the previous code, you could have duplicate import statements if there are two enum cases with field types that require an import.

None.

1. This is very simple and low risk.

# Testing

This fixed an issue I had with this type:
```
#[derive(Debug, Clone, Copy, SpacetimeType, Hash, Eq, PartialEq)]
enum Role {
    Spectator,
    Player(Team),
    Spymaster(Team),
}
```

Previously the import for `Team` would get generated twice.

- [x] Worked for previously failing example.

### DIFF
--- a/crates/cli/src/subcommands/generate/typescript.rs
+++ b/crates/cli/src/subcommands/generate/typescript.rs
@@ -1,5 +1,6 @@
 use super::util::fmt_fn;
 
+use std::collections::BTreeSet;
 use std::fmt;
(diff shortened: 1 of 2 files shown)


mamcx (approved):
LGTM

PR #1488, opened 2024-07-09
Filter extra deps on keccak on macOS that fails snap testing

mamcx:
# Description of Changes

This is a workaround to make it possible to run `cargo test --all`.

Only on `macOS` when running the `cargo insta` on the `deptree_snapshot` test it gets an extra deps of

```bash
sha3 v0.10.8
└── keccak v0.1.4
            └── cpufeatures v0.2.9 # Extra on macOS
                        └── libc v0.2.153
```
 
... that causes all the other test fails by `PoisonError`.

1

# Testing

- [x] *Confirm with joshua this happens on macOS*
- [x] *Check `cargo test --all` now can run.*

### DIFF
--- a/crates/bindings/tests/deps.rs
+++ b/crates/bindings/tests/deps.rs
@@ -2,12 +2,41 @@
 //! to make sure we don't unknowningly add a bunch of dependencies here,
 //! slowing down compilation for every spacetime module.
 
+#[cfg(target_os = "macos")]
+/// On macOS, the `keccak` crate depends on `cpufeatures/libc`, which don't on Linux.
+fn filter_keccak(input: String) -> String {
+    let mut lines = input.lines();
+    let mut found_keccak = false;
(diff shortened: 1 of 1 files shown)


crates/bindings/tests/deps.rs
```diff
@@ -2,12 +2,41 @@
...
+        } else if found_cpufeatures && line.contains("libc") {
+        } else {
+            result.push(line.to_string());
+        }
+    }
+
+    result.join("\n")
+}
+
+#[cfg(not(target_os = "macos"))]
```
joshua-spacetime, line 28: Instead of filtering on macos, can we have two different snapshots? I.e. add a new snapshot for macos. And then use a compile-time flag inside of `deptree_snapshot` to use the correct one depending on the platform?

joshua-spacetime (changes requested):

PR #1489, opened 2024-07-09
Add support for Gzip compression on server

lcodes:
# Description of Changes

Removed the brotli dependency and replaced it with the flate2 library already being included.

Changes the communications between the STDB server and its SDK clients, but opaque to users.

1. Straightforward change. Needs matching updates for the client repos.

# Testing

Running the sdk test suite


joshua-spacetime:
Note, I updated the PR title to reflect the decision to support Gzip in addition to Brotli compression.

Cargo.toml
```diff
@@ -138,7 +138,7 @@ email_address = "0.2.4"
 enum-as-inner = "0.6"
 enum-map = "2.6.3"
 env_logger = "0.10"
-flate2 = "1.0.24"
+flate2 = "1.0.24" # TODO features = ["zlib-ng"], default-features = false ?
```
lcodes, line 141: Should we consider the C zlib library instead of the (slower) rust one? This is from the flate2 docs

joshua-spacetime: Note that we tested the speed of the rust implementation and it's very similar to the current brotli implementation.

joshua-spacetime (commented):
Can you add a test?

bfops:
Adding the `api-break` label because it seems like this requires SDK changes to provide the choice of compression? Not sure though, I didn't look to closely. Feel free to remove.

bfops:
Nevermind, I read a bit more and it looks like we gracefully default to brotli.

joshua-spacetime:
@lcodes this looks like it's ready although the sdk tests appear to be failing. Also if it's not too difficult, it would be really nice to add a test exercising gzip specifically.

lcodes:
Had issues with getting sdk tests to pass on this branch, moved the impl back in parts to a new branch until tests pass again; moving the PR there https://github.com/clockworklabs/SpacetimeDB/pull/1642

Issue #1371, opened 2024-06-11
[STABILITY] Return empty success from /identity GET rather than 404 when no identities match the queried email address


cloutiertyler:
Currently, /identity?email=no-one-owns-this@email.address returns a 404 when the email address has no associated identities. It could, and probably should, just return an empty list instead, so that the website could detect the difference between 404: wrong url and no-such-identity.

PR #1491, opened 2024-07-10
Return empty success from /identity GET

lcodes:
# Description of Changes

https://github.com/clockworklabs/SpacetimeDB/issues/1371

Swap out the previous 404 for a default response.

Changes the behavior in case of an empty email match, going from a client error to an empty success for its response.

0

# Testing

```
λ curl localhost:3000/identity?email=me@example.com
{"identities":[]}

λ curl -I localhost:3000/identity
HTTP/1.1 400 Bad Request
```

### DIFF
--- a/crates/client-api/src/routes/identity.rs
+++ b/crates/client-api/src/routes/identity.rs
@@ -44,7 +44,7 @@ pub async fn create_identity<S: ControlStateDelegate + NodeDelegate>(
     Ok(axum::Json(identity_response))
 }
 
-#[derive(Debug, Clone, Serialize)]
+#[derive(Debug, Clone, Serialize, Default)]
 pub struct GetIdentityResponse {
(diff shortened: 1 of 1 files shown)


crates/client-api/src/routes/identity.rs
```diff
@@ -77,7 +77,7 @@ pub async fn get_identity<S: ControlStateDelegate>(
             (!identities.is_empty()).then_some(GetIdentityResponse { identities })
         }
     };
-    let identity_response = lookup.ok_or(StatusCode::NOT_FOUND)?;
+    let identity_response = lookup.unwrap_or_default();
```
bfops, line 80: nit: it seems like we should return 400 / Bad Request if no `email` is provided?

bfops (approved):
Nice, this simplified version is even clearer. LGTM!

Issue #1372, opened 2024-06-11
[STABILITY] /database/request_recovery_code and /database/confirm_recovery_code should be POST, not GET.


cloutiertyler:
Also, to be honest, they should be `/identity/request_recovery_code` and `/identity/confirm_recovery_code`, not in `/database`.

PR #1492, opened 2024-07-10
Moving recovery code APIs under /identity and using POST

lcodes:
# Description of Changes

https://github.com/clockworklabs/SpacetimeDB/issues/1372

Added `control_routes` to identity, based off database's.
Moved the recovery code APIs there.
Switched them to expect POST.

Changes both the request method and the router endpoints of the recovery code API.

0

# Testing

```
curl -X POST "localhost:3000/identity/request_recovery_code?email=me@example.com&identity=1234567890123456789012345678901234567890123456789012345678901234"
curl -X POST "localhost:3000/identity/confirm_recovery_code?email=me@example.com&identity=1234567890123456789012345678901234567890123456789012345678901234&code=1"
```

Assuming the previous handler code is correct, this validates the routes and request method switches.

### DIFF
--- a/crates/cli/src/subcommands/identity.rs
+++ b/crates/cli/src/subcommands/identity.rs
@@ -628,8 +628,8 @@ async fn exec_recover(mut config: Config, args: &ArgMatches) -> Result<(), anyho
     }
 
     let client = reqwest::Client::new();
-    let builder = client.get(Url::parse_with_params(
-        format!("{}/database/request_recovery_code", config.get_host_url(server)?,).as_str(),
+    let builder = client.post(Url::parse_with_params(
+        format!("{}/identity/request_recovery_code", config.get_host_url(server)?,).as_str(),
         query_params,
     )?);
     let res = builder.send().await?;
--- a/crates/cli/src/subcommands/identity.rs
+++ b/crates/cli/src/subcommands/identity.rs
@@ -653,8 +653,8 @@ async fn exec_recover(mut config: Config, args: &ArgMatches) -> Result<(), anyho
         };
 
         let client = reqwest::Client::new();
-        let builder = client.get(Url::parse_with_params(
-            format!("{}/database/confirm_recovery_code", config.get_host_url(server)?,).as_str(),
+        let builder = client.post(Url::parse_with_params(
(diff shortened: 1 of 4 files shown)


crates/client-api/src/routes/identity.rs
```diff
@@ -233,3 +236,117 @@ where
...
+
+pub fn control_routes<S>(_: S) -> axum::Router<S>
+where
+    S: NodeDelegate + ControlStateDelegate + Clone + 'static,
+{
+    use axum::routing::post;
+    axum::Router::new()
+        .route("/request_recovery_code", post(request_recovery_code::<S>))
+        .route("/confirm_recovery_code", post(confirm_recovery_code::<S>))
+}
```
bfops, line 352: why is this a separate `Router` rather than just being part of the existing one?

crates/cli/src/subcommands/identity.rs
```diff
@@ -628,8 +628,8 @@ async fn exec_recover(mut config: Config, args: &ArgMatches) -> Result<(), anyho
     }
 
     let client = reqwest::Client::new();
-    let builder = client.get(Url::parse_with_params(
-        format!("{}/database/request_recovery_code", config.get_host_url(server)?,).as_str(),
+    let builder = client.post(Url::parse_with_params(
+        format!("{}/identity/request_recovery_code", config.get_host_url(server)?,).as_str(),
```
bfops, line 632: shouldn't we also be updating `confirm_recovery_code`?

lcodes: Good catch

crates/client-api/src/routes/identity.rs
bfops: Yeah I think this makes sense. I did a bit of investigating and it seems like even in `database.rs` the distinction isn't important anymore.

bfops:
@lcodes It looks like the smoketests are failing, but I don't _think_ it has anything to do with this PR - can you update with latest `master` and see if that fixes it?

Unfortunately, I don't think there's a way to test this recovery flow locally, so we'll have to merge and test it once deployed somewhere with SendGrid enabled.

bfops (approved):
Seems reasonable to me! Let's try to remember to test this once it lands somewhere with SendGrid enabled.

PR #1498, opened 2024-07-11
Move schemas to `schema` crate, rename `Def` to `RawDefV8`

kazimuth:
# Description of Changes

This is in preparation for further schema rewrites, extracted from #1448, based on #1479. There are only name changes (well, and one bindings change due to the name changes).

I've chosen aggressive, ugly names for the existing `Def` types. This is because soon, these types will be considered unvalidated user input, which should not be used anywhere except at the module boundary, where they are immediately validated and converted to trusted types. I want these types to stick out in code to encourage migration away from them.

This is technically an API break, but the APIs involved are only used by macros, so 🤷 

0


mamcx (commented):
LGTM. 


NOTE: The removal of names for sequence and such reordering break the system tables, so I think this should be marked.

kazimuth:
Yeah, which is why I might not remove them after all, together with SQL compatibility. But this PR doesn't actually implement that, it just renames things to make future PRs easier to read.

kazimuth:
WAIT A MINUTE. We're on ABI version 8. Should this all be renamed `RawModuleDefV8`? Or possibly `V9`, since we are about to introduce an ABI break to make this versioned?

cloutiertyler:
Two suggestions:

1. Use v8 instead of v0
1. remove the V0 from the type and make it `v8::RawTableDef` instead using a rust module

mamcx:
Good question. I think if we need constant warnings then having the version in the structs is better, else if wanna clean refactorings in the module.

kazimuth:
Yeah, we want warnings because these are now considered unvalidated user input types, and we want people to move to the new `ModuleDef` in my upcoming PR.

cloutiertyler (approved):
This looks good to me. I think we're going to have to coordinate merging with with @gefjon's P0

kazimuth:
Rebased over that, this should be good to go

PR #1500, opened 2024-07-12
C# smoketests use `nuget.config`

bfops:
# Description of Changes

Our C# smoketests used to manually modify the `.csproj` file generated by `spacetime init --lang csharp`.

This PR modifies the smoketests to not modify the `.csproj`, and to instead add a separate `nuget.config` file. This is easier to maintain / less prone to breakage. (It was also partly just practice of using `nuget.config`, because it's easier to port it over to other scenarios)

This seems to have resolved the issue where pointing to `0.10.0` might still pull from NuGet instead of the local repo. I'm not 100% sure that it's actually due to this change - I may have tested that issue incorrectly before (I may have pointed at `0.10.*` and updated the local package version to `0.10.0`, and it might have decided to pull `0.10.1` from NuGet because it was more recent; not completely sure)

Nope.

1

# Testing

- [x] Smoketests pass
- [x] Smoketests pass [even when we downgrade versions to 0.10.0](https://github.com/clockworklabs/SpacetimeDB/pull/1502)

### DIFF
--- a/crates/bindings-csharp/BSATN.Codegen/BSATN.Codegen.csproj
+++ b/crates/bindings-csharp/BSATN.Codegen/BSATN.Codegen.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <AssemblyName>SpacetimeDB.BSATN.Codegen</AssemblyName>
-    <Version>0.10.0</Version>
+    <Version>0.11.0</Version>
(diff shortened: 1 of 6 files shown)


jdetter (approved):
I would ask @RReverser to take another look at this when he gets back, but for now this is fine with me 👍

PR #1501, opened 2024-07-12
Add some helper scripts

kazimuth:
# Description of Changes

I keep forgetting the right invocations to run all the tests & clippy correctly, so I added some scripts that do this.

0

# Testing

I ran the scripts :)


tools/run-all-tests.sh
```diff
@@ -0,0 +1,52 @@
...
+set -euo pipefail
+
+script_dir="$(cd -P -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd -P)/"
+stdb_root="$(realpath "$script_dir/../")"
+
+# Find cargo.
+cargo=$(which cargo)
+if [ ! -x "$cargo" ] ; then
+    echo "Could not find 'cargo' executable, please install it."
+    exit 1
```
bfops, line 12: will we ever get here? won't `which` return nonzero, causing the script to exit?

tools/run-all-tests.sh
```diff
@@ -0,0 +1,52 @@
...
+
+# Find dotnet.
+dotnet=$(which dotnet)
+if [ -x "$dotnet" ] ; then
+    dotnet_version=$($dotnet --version 2>&1 | awk -F '.' '{print $1}')
+    if [ "$dotnet_version" -lt 8 ] ; then
+        dotnet=""
+        echo "Could not find 'dotnet' version 8 executable ($dotnet is version $($dotnet --version 2>&1)), skipping dotnet tests."
+    fi
+fi
```
bfops, line 23: ditto above and/or shouldn't there be an `else`?

tools/run-all-tests.sh
```diff
@@ -0,0 +1,52 @@
...
+
+cd "$stdb_root"
+tools/clippy.sh
+cargo test --all
+if [ -x "$python" ] ; then
+    $python -m smoketests
+fi
+if [ -x "$dotnet" ] ; then
+    $dotnet test crates/bindings-csharp
+fi
```
bfops, line 51: won't the tests themselves generate the errors about `dotnet` not being found, being the wrong version, etc?

if `dotnet` is in the path, then this would work.. couldn't we just.. run `dotnet test crates/bindings-csharp` without really bothering to do any looking for it in the first place?

tools/clippy.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/env bash
+
+script_dir="$(cd -P -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd -P)/"
```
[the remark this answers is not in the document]
kazimuth: The above incantation also works when the script is being sourced, but that doesn't matter in this case so I swapped it

tools/run-all-tests.sh
kazimuth: ...you are correct

tools/clippy.sh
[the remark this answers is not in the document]
bfops: (fwiw I care less about the `BASH_SOURCE` and more about `readlink` vs `cd && pwd -P`)

tools/run-all-tests.sh
```diff
@@ -0,0 +1,29 @@
...
+set -euo pipefail
+
+script_dir="$(readlink -f "$(dirname "$0")")"
+stdb_root="$(realpath "$script_dir/../")"
+
+set -euox pipefail
+
+cd "$stdb_root"
+
+tools/clippy.sh
```
bfops, line 12: this requires `tools/clippy.sh` to be executable. I pushed a commit to make both of these executable (so we can run them both directly rather than e.g. `bash tools/clippy.sh`)

bfops (approved):
Works for me!

This will be nicer than me always doing `Ctrl-R clippy`.

tools/run-all-tests.sh
kazimuth: Thank you. For some reason my git executable kept ignoring when I did that, no idea why.

bfops: Yeah I was wondering if you were on Windows in some form. I think it comes down to the fact that Windows doesn't have the same filesystem bits for controlling permissions. There's some other way to manually force the changes into git..

bfops: Looks like this is how to write it directly into the git index:
```bash
git update-index --chmod=+x <file>
```
https://stackoverflow.com/questions/6476513/git-file-permissions-on-windows

PR #1503, opened 2024-07-12
CI - SpacetimeDB PRs run the C# SDK tests

bfops:
# Description of Changes

Added a CI step that checks out `spacetimedb-csharp-sdk` and runs its test suite.

In the current PR incarnation, if the local major/minor package versions are bumped, the C# SDK tests will continue to use the old ones. This is because the C# SDK repo will still have its version numbers pointed to e.g. `0.10.*` rather than e.g. `0.11.*`. Is this a bug or a feature?

I kind of think that the current behavior is correct - if we bump the major/minor version numbers, we're explicitly aware that we're releasing a breaking change. This effectively tests the experience of a current _user_ of the C# SDK (on `master`). In that case, we can make this a required check.

On the other hand, if we decide we want the C# SDK to _always_ use the local versions, then this should **not** be made a required check. We may sometimes release breaking changes, which will need to happen in this repository before they happen in that one.

No

2

# Testing

- [x] CI passes
  - [x] CI logs show that `dotnet test` is, in fact, running
  - [x] If we manually downgrade the `bindings-csharp` versions to 0.10.1 (to match the latest version on NuGet), the tests use the local versions (and fail, because the C# SDK repo is not yet updated for the latest packages)
    - https://github.com/clockworklabs/SpacetimeDB/actions/runs/9944746191/job/27471462469?pr=1510

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -137,7 +137,7 @@ jobs:
             </configuration>
           EOF
           # clear package caches, so we get fresh ones even if version numbers haven't changed
-          dotnet nuget local all --clear
+          dotnet nuget locals all --clear
           dotnet test
 
   lints:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -120,17 +120,13 @@ jobs:
       - name: C# SDK tests
         run: |
           ( cd crates/bindings-csharp/BSATN.Runtime && dotnet pack )
-          ( cd crates/bindings-csharp/BSATN.Codegen && dotnet pack )
           cd spacetimedb-csharp-sdk
-          # only for testing - remove this before merging!
-          sed -i'' 's/Version="0.10.\*"/Version="0.12.*"/' SpacetimeDB.ClientSDK.csproj tests/tests.csproj
           cat >nuget.config <<EOF
             <?xml version="1.0" encoding="utf-8"?>
             <configuration>
               <packageSources>
                 <!-- Local NuGet repositories -->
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -77,9 +77,67 @@ jobs:
...
+      - uses: dsherret/rust-toolchain-file@v1
+
+      - uses: actions/setup-dotnet@v3
+        with:
+          dotnet-version: "8.x"
+
+      - name: Create /stdb dir
+        run: |
+          sudo mkdir /stdb
+          sudo chmod 777 /stdb
```
jdetter, line 113: 🌶️

.github/workflows/ci.yml
```diff
@@ -77,9 +77,67 @@ jobs:
...
+        with:
+          repository: clockworklabs/spacetimedb-csharp-sdk
+          path: spacetimedb-csharp-sdk
+
+      - name: C# SDK tests
+        run: |
+          ( cd crates/bindings-csharp/BSATN.Runtime && dotnet pack )
+          cd spacetimedb-csharp-sdk
+
+          cat >nuget.config <<EOF
```
jdetter, line 125: can we throw a comment on this explaining what we're doing here?

jdetter (approved):
Small nit but this looks great, thanks zeke 🙏

.github/workflows/ci.yml
bfops: added one - how does this look?

.github/workflows/ci.yml
bfops: good ol' copy-:spaghetti:

bfops:
(did a final round of testing by manually downgrading the local package versions to match the NuGet versions, see the `Testing` section)

PR #1518, opened 2024-07-15
ABI BREAK: Add versioning for RawModuleDef

kazimuth:
# Description of Changes

Modules now return a serialized enum `RawModuleDef` from `__describe_module__` for future versioning extensibility. This is an ABI break, but it will make it easier to avoid ABI breaks in the future.

1

# Testing

I still need to run the full test suite which is why I'm opening this PR.

### DIFF
--- a/crates/bindings-csharp/Runtime/Internal/Module.cs
+++ b/crates/bindings-csharp/Runtime/Internal/Module.cs
@@ -122,7 +122,7 @@ public static partial class Module
         : SpacetimeDB.TaggedEnum<(TypeAlias TypeAlias, Unit _Reserved)>;
 
     [SpacetimeDB.Type]
-    public partial struct ModuleDef()
+    public partial struct RawModuleDefV8()
     {
         List<AlgebraicType> Types = [];
         List<TableDesc> Tables = [];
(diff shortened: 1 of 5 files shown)


kazimuth:
Notes on ABI versioning:

This is an ABI break, so it should technically be called ABI version 9. But then we will soon have

```rust
enum RawModuleDef {
   V9(v8::RawModuleDefV8), // old ABI
   V10(v10::RawModuleDefV10) // new ABI based on the schema proposal
}
```

which is very confusing.

Note that V10 is also sort of *not* an ABI break, because it's adding an enum variant.

We could also say something like:

```rust
enum RawModuleDef {
   V8BackCompat(v8::RawModuleDefV8), // old ABI
   V9(v9::RawModuleDefV9) // new ABI based on the schema proposal
}
```
which I am leaning towards.

cloutiertyler:
I would either do:

```rust
enum RawModuleDef {
   V9(v9::RawModuleDefV9), // old ABI
   V10(v10::RawModuleDefV10) // new ABI based on the schema proposal
}
```
or 
```rust
enum RawModuleDef {
   V8BackCompat(v8::RawModuleDefV8), // old ABI
   V9(v9::RawModuleDefV9) // new ABI based on the schema proposal
}
```

kazimuth:
I'll stick with `V8BackCompat` then cuz it's less churn

cloutiertyler (approved):
This LGTM now.

PR #1519, opened 2024-07-15
Added v8 mod and removed V0 from names

cloutiertyler:


kazimuth:
This also needs to be rebased, I'll just do it in my branch

kazimuth:
Implemented in #1498

PR #1524, opened 2024-07-17
Ignore some sequence allocation mismatches when checking for compatible updates

gefjon:
# Description of Changes

When checking for schema compatibility before an update, we erroneously considered sequence allocations to be an incompatible change. This means that any module which contained at least one sequence/`autoinc` column which had allocated more than 4096 values would be impossible to ever update.

With this change, we permit updates where the proposed sequence allocation is less than the known sequence allocation, and use the larger known sequence allocation as the value. This makes it possible to update modules which contain `autoinc` columns that have allocated more than 4096 values.

N/a.

1

# Testing

- [x] Published new compatible module on mirror of BitCraft alpha 2; John connected and walked around and stuff and said it looked reasonable.

### DIFF
--- a/crates/core/src/db/update.rs
+++ b/crates/core/src/db/update.rs
@@ -282,6 +282,16 @@ fn schema_compat_and_updates_hack(
         updates.changed_access.insert(table_name(), prop_access);
     }
 
+    // If a proposed sequence has a lower allocation than an existing sequence,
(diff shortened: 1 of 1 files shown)


jdetter (approved):
Tested against BitCraft, working. We'll get this deployed ASAP. Thanks Phoebe!

kazimuth (approved):
👍

PR #1528, opened 2024-07-18
commitlog: Fix single-commit bitflip test

kim:
It was sometimes generating a zero bitmask.
Also turn into proptest, so we get the randomized values as inputs.

1


bfops (approved):
Seems reasonable to me, as far as I understand!

PR #1530, opened 2024-07-19
table: Make `with_mut_schema` clone-on-write

kim:
Avoid incrementing the refcount of the schema before mutating, so it
gets modified in place if no other reference is live.

1


bfops:
I'm removing the `no runtime change` label since this is labeled as a bugfix :monocle_face: Am I crazy?

PR #1535, opened 2024-07-23
core: Store system tables in commitlog

### LINKED ISSUE #1433: Bootstrap the system tables from the commitlog

kim:
Changes the bootstrap sequence, such that the initial system table
schemas and contents are stored in the commitlog.

Opening a database proceeds as follows:

- If a snapshot exists, restore it and replay the history suffix
- Otherwise:
  - Create the "seed tables" `st_table` and `st_columns` in-memory
  - Replay the history
    - Whenever a table in the reserved system table range is found,
      reset the system table schema and layout information in the
      page manager, so as to repair incomplete schema information.
  - If the history was empty:
    - Create the rest of the system tables (i.e. sans seed tables) in a
      transaction.
    - Commit and persist that transaction.

## Alternatives Considered

The already brittle bootrapping process is now even more brittle. In particular
the various incantations of repair / rebuild routines is hard to follow and easy
to break.

It is also unfortunate that the commitlog will be non-empty in case the
initialization from the user-supplied module fails.

As an alternative, a snapshot could be taken right after the database was fully
initialized, incl. the module. Instead of hashes, this snapshot could contain
their preimages and be inlined into the commitlog.

Restoring a database from such a snapshot requires only the statically known
schemas of the seed tables and a single repair pass.

It will not be possible to open a database from a commitlog created by an older
version, because bootstrapping the system tables will only occur if the history
is empty.

5 - database bootstrap is highly fragile

### DIFF
--- a/crates/core/src/db/datastore/locking_tx_datastore/committed_state.rs
+++ b/crates/core/src/db/datastore/locking_tx_datastore/committed_state.rs
@@ -8,12 +8,9 @@ use crate::{
     db::{
         datastore::{
             system_tables::{
-                system_tables, StColumnRow, StConstraintRow, StIndexRow, StSequenceRow, StTableFields, StTableRow,
-                SystemTable, ST_CLIENTS_ID, ST_CLIENT_IDX, ST_COLUMNS_ID, ST_COLUMNS_IDX, ST_COLUMNS_NAME,
-                ST_CONSTRAINTS_ID, ST_CONSTRAINTS_IDX, ST_CONSTRAINTS_NAME, ST_INDEXES_ID, ST_INDEXES_IDX,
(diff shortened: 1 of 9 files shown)


crates/core/src/db/datastore/locking_tx_datastore/committed_state.rs
```diff
@@ -135,9 +130,9 @@ impl CommittedState {
             // If the row is already there, no-op.
             ignore_duplicate_insert_error(st_tables.insert(blob_store, &row))?;
         }
-
+        //
```
gefjon, line 133:
```suggestion
```

gefjon (commented):
I do not love this. Having reviewed it, I am now much more convinced by your inline snapshots proposal and am inclined to push that forward rather than this.

Issue #1527, opened 2024-07-17
CI - Fix flaky smoketests that subscribe to queries


bfops:
See e.g. https://github.com/clockworklabs/SpacetimeDB/actions/runs/9977698168/job/27575374257?pr=1479

I believe this is due to these smoketests using the CLI to subscribe to queries. They asynchronously read errors/exceptions out from the CLI output, but the CLI call is wrapped in a **synchronous** exception handler, so this handler may or may not be triggered properly depending on the timing. For example, in `smoketests/tests/permissions.py`:

```python
        with self.assertRaises(Exception):
            self.subscribe("SELECT * FROM Secret", n=0)
```

PR #1536, opened 2024-07-23
Fix flaky subscription smoketests

bfops:
# Description of Changes

Fixes https://github.com/clockworklabs/SpacetimeDB/issues/1527. From there:

> See e.g. https://github.com/clockworklabs/SpacetimeDB/actions/runs/9977698168/job/27575374257?pr=1479
>
> I believe this is due to these smoketests using the CLI to subscribe to queries. They asynchronously read errors/exceptions out from the CLI output, but the CLI call is wrapped in a **synchronous** exception handler, so this handler may or may not be triggered properly depending on the timing. For example, in `smoketests/tests/permissions.py`:
> 
> ```python
>         with self.assertRaises(Exception):
>             self.subscribe("SELECT * FROM Secret", n=0)
> ```

Those blocks of code would create a subscription (in a background thread), but not necessarily wait for it to complete. I believe there was a small race condition where we could exit the `assertRaises` block before getting the exception (see the logs in the linked ticket).

I have updated these calls to wait for the subscription to finish, which re-raises any exceptions that occur in the background thread. Note this is due to `subscribe` using a custom thread class - this re-raising does not happen by default in Python.

No

1

# Testing

The CI currently passes, so it's not obviously-broken.

I was not able to reproduce the failure locally (at least, for the amount of time I was willing to wait). So, my only testing is that the tests still _pass_. It's probably not _more_ flaky.

If anybody else wants to try, this was the loop I used:

```bash
true;
while [[ $? == 0 ]]; do
  python3 -m smoketests;
done;
discord-ping 'test failed';
```


smoketests/tests/add_remove_index.py
```diff
@@ -64,7 +64,7 @@ def test_add_then_remove_index(self):
         # There are no indices, resulting in an unsupported unindexed join.
         self.publish_module(name, clear = False)
         with self.assertRaises(Exception):
-            self.subscribe(self.JOIN_QUERY, n = 0)
+            self.subscribe(self.JOIN_QUERY, n = 0)()
```
bfops, line 67: adding this `()` causes us to wait for the subscription results (which, in this case, will throw an exception).

This is the same for the lines below.

joshua-spacetime:
> I did not actually address the race condition in the subscribe function.

@bfops what's the race condition and why does this change make it so that we're no longer relying on it?

smoketests/__init__.py
```diff
@@ -237,17 +237,16 @@ def stderr_task():
         threading.Thread(target=stderr_task).start()
 
         print("initial update:", proc.stdout.readline())
-        if proc.poll() is not None:
-            if proc.returncode:
-                raise subprocess.CalledProcessError(proc.returncode, fake_args)
-            return lambda timeout=None: []
```
bfops, line 243: This was the source of the race condition - if the process failed quickly enough, we would raise this exception synchronously in the call to `subscribe`. But if the process took a bit longer, then the block below would be the one to raise the exception, which would only be raised to the caller when the caller actually joined the background thread and waited for the subscription to complete.

By removing this block, we now _only_ get exceptions when we wait for the subscription result. We will never get the exception synchronously. This eliminates the race condition - we must always wait for the subscription to complete in order to see the exception.

I have updated the tests accordingly.

smoketests/__init__.py
```diff
@@ -237,17 +237,16 @@ def stderr_task():
...
 
         def run():
             updates = list(map(json.loads, proc.stdout))
             code = proc.wait()
             if code:
                 raise subprocess.CalledProcessError(code, fake_args)
             return updates
+        # Note that we're returning `.join`, not `.join()`; this returns something that the caller can call in order to
+        # join the thread and wait for the results.
+        # If the caller does not invoke this returned value, the thread will just run in the background and not be awaited.
```
joshua-spacetime, line 249: Excellent!

smoketests/__init__.py
```diff
@@ -299,6 +298,9 @@ def enterClassContext(cls, cm):
             return result
 
 
+# This is a custom thread class that will propagate an exception to the caller of `.join()`.
+# This is required because, by default, threads do not propagate exceptions to their callers,
+# even callers who have called `join`.
```
joshua-spacetime, line 303: Excellent!

joshua-spacetime (approved):

PR #1542, opened 2024-07-24
Add lib::db::{column_ordering, raw_def::v9}

kazimuth:
# Description of Changes

This is NOT an ABI or API break because we have enums now!!

I decided to split these out from the validation & migration code PR to make it easier to review.
I haven't added any code to work with these, just the definitions.

The new "test" features are going to be exercised in my next PR.

a mild 1/5: this adds some new types that aren't used anywhere, which is perhaps slightly confusing. 

# Testing

CI should suffice.

### DIFF
--- a/crates/lib/src/db/raw_def/v9.rs
+++ b/crates/lib/src/db/raw_def/v9.rs
@@ -1,5 +1,3 @@
-//! This module contains the "raw definitions" returned by modules to the database.
-
 use crate::db::auth::{StAccess, StTableType};
 use itertools::Itertools;
 use spacetimedb_primitives::*;
@@ -180,6 +178,7 @@ pub struct RawTableDefV9 {
     /// Must be a valid [crate::db::identifier::Identifier].
     pub name: RawIdentifier,
     /// A reference to a product type containing the columns of this table.
+    /// This is the single source of truth for the table's columns.
     ///
     /// Like all types in the module, this must have the [default element ordering](crate::db::default_element_ordering), UNLESS a custom ordering is declared via `ModuleDef.misc_exports` for this type.
     pub product_type_ref: AlgebraicTypeRef,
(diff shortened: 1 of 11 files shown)


crates/lib/src/db/raw_def/v9.rs
```diff
@@ -0,0 +1,467 @@
...
+    /// The types used in the module.
+    ///
+    /// `AlgebraicTypeRef`s in the table, reducer, and type alias declarations refer to this typespace.
+    ///
+    /// Any `Product` or `Sum` types used transitively by the module MUST be declared in this typespace.
+    /// Each should have EXACTLY ONE corresponding `TypeAlias` declaration in the `misc_exports` field,
+    /// with a module-unique name.
+    /// This ensures that client code generation will work correctly.
+    ///
+    /// It is permitted but not required to refer to `Builtin` types via this typespace.
```
cloutiertyler, line 49:
```suggestion
    /// It is permitted but not required to refer to `Builtin` or "primitive" types via this typespace.
```
We're eventually removing the `Builtin` distinction so it'll be a little confusing

cloutiertyler (approved):
This looks good to me. Just one small nit.

crates/lib/src/db/column_ordering.rs
```diff
@@ -0,0 +1,129 @@
+use std::cmp::Ordering;
+
+use spacetimedb_sats::{AlgebraicType, BuiltinType, Typespace};
+
+/// Canonical ordering for fields of a Table.
+///
+/// Fields are ordered by their `SpacetimeDB` alignment, then alphabetically by name.
+///
+/// Does not currently apply to all AlgebraicTypes.
```
gefjon, line 9:
```suggestion
/// Does not currently apply to all AlgebraicTypes, only to table types.
/// This means that, if a table contains a `ProductValue` as a column,
/// that column's fields will not be sorted.
```

crates/lib/src/db/column_ordering.rs
```diff
@@ -0,0 +1,129 @@
...
+            .unwrap_or(1),
+        AlgebraicType::Sum(ref sum) => sum
+            .variants
+            .iter()
+            .map(|variant| alignment(&variant.algebraic_type, ctx))
+            .max()
+            .unwrap_or(1),
+        AlgebraicType::Ref(ref ref_) => alignment(&ctx[*ref_], ctx),
+    }
+}
```
gefjon, line 57: I do not feel good about duplicating alignment computation in addition to the version in the table crate (in `layout.rs`). It may be unavoidable, but if so please add comments *to both implementations* to the effect that:
1. It is very important that they stay consistent.
2. Why the duplication is necessary.

crates/lib/src/db/column_ordering.rs
```diff
@@ -0,0 +1,129 @@
...
+
+/// Check that a slice is sorted by an ordering.
+///
+/// TODO: remove this when `[T]::is_sorted_by` is stabilized.
```
gefjon, line 61:
```suggestion
/// TODO: remove this when [`[T]::is_sorted_by`](https://doc.rust-lang.org/std/primitive.slice.html#method.is_sorted_by) is stabilized.
```

crates/lib/src/db/column_ordering.rs
kazimuth: Removed!

crates/lib/src/db/raw_def/v9.rs
```diff
@@ -0,0 +1,521 @@
...
+            name,
+            at_column,
+            reducer_name,
+        });
+        self
+    }
+
+    /// Build the table and add it to the module.
+    pub fn finish(self) {
+        // self is now dropped.
```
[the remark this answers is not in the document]
kazimuth: It was how the existing table building works, the methods operated on `mut self` and returned `Self`. I'm not particularly attached to it

crates/lib/src/db/raw_def/v9.rs
```diff
@@ -0,0 +1,521 @@
...
+
+    /// Get the column ID of the column with the specified name, if any.
+    ///
+    /// Returns `None` if this `TableDef` has been constructed with an invalid `ProductTypeRef`,
+    /// or if no column exists with that name.
+    pub fn get_col_id(&self, column: impl AsRef<str>) -> Option<ColId> {
+        let column = column.as_ref();
+        self.columns()?
+            .iter()
+            .position(|x| x.name.as_ref().map(|s| &s[..]) == Some(column))
```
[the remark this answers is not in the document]
kazimuth: Turns out I should have been using `.name()` anyway :)

Centril (approved):

PR #1543, opened 2024-07-25
Add ErrorStream combinator

kazimuth:
# Description of Changes

This was pulled out of the schema refactor because it's standalone (and imo pretty cute.)

There are some crates on crates.io that do something similar, but they all look either heavy-duty or unmaintained.

0

### DIFF
--- a/crates/data-structures/src/error_stream.rs
+++ b/crates/data-structures/src/error_stream.rs
@@ -1,0 +1,254 @@
+//! Methods and traits for dealing with multiple errors simultaneously.
+
+/// A non-empty stream of errors.
+///
+/// Logically, this type is unordered, and it is not guaranteed that the errors will be returned in the order they were added. Attach identifying information to your errors if you want to sort them.
+///
+/// This struct is intended to be used with:
+/// - The [CombineErrors] trait, which allows you to combine a tuples of results.
+/// - The [CollectAllErrors] trait, which allows you to collect errors from an iterator of results.
+#[derive(thiserror::Error, Debug, Clone, Default, PartialEq, Eq)]
+pub struct ErrorStream<E>(smallvec::SmallVec<[E; 1]>);
+
+impl<E> ErrorStream<E> {
+    /// Returns an iterator over the errors in the stream.
+    pub fn iter(&self) -> impl Iterator<Item = &E> {
+        self.0.iter()
+    }
+
+    /// Returns a mutable iterator over the errors in the stream.
+    pub fn iter_mut(&mut self) -> impl Iterator<Item = &mut E> {
+        self.0.iter_mut()
+    }
+
+    /// Returns an iterator over the errors in the stream, consuming the stream.
+    pub fn drain(&mut self) -> impl Iterator<Item = E> + '_ {
+        self.0.drain(..)
+    }
+
+    /// Push an error onto the stream.
+    pub fn push(&mut self, error: E) {
+        self.0.push(error);
+    }
+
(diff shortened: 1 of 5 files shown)


cloutiertyler:
Seems fine if we fix the snapshot errors. What is the motivation to add this though?

crates/data-structures/src/error_stream.rs
```diff
@@ -0,0 +1,248 @@
...
+    ///     let (age, name) = (age(), name()).combine_errors()?;
+    ///     Ok(format!("{} is {} years old", name, age))
+    /// }
+    /// ```
+    fn combine_errors(self) -> Result<Self::Ok, ErrorStream<Self::Error>>;
+}
+
+macro_rules! tuple_combine_errors {
+    ($($T:ident),*) => {
+        impl<$($T,)* E> CombineErrors for ($(Result<$T, ErrorStream<E>>,)*) {
```
gefjon, line 103: Why is this impl'd for `Result<T, ErrorStream<E>>` rather than for `Result<T, E>`? Is it possible to do some type magic to permit either `E` or `ErrorStream<E>` as the err variant of the result in this impl? Something like,

```rust
trait ExtendErrorStream<E> {
  fn add_error(&mut self, e: E);
}
impl<E> ExtendErrorStream<E> for ErrorStream<E> {
  fn add_error(&mut self, e: E) { self.0.push(e); }
}
impl<E> ExtendErrorStream<ErrorStream<E>> for ErrorStream<E> {
  fn add_error(&mut self, e: ErrorStream<E>) { self.0.extend(e.0); }
}

// macro needs both `T` ok types and `En` err types, and expands:
impl<$($T,)* $($En,)* E> CombineErrors for ($(Result<$T, $En>,)*)
where
  $(ErrorStream<E>: ExtendErrorStream<$En>,)*
{
  type Ok = ($($T,)*);
  type Error = E;
  fn combine_errors(self) -> Result<Self::Ok, ErrorStream<Self::Error>> {
    // pretty much the same, but with `add_error` instead of `unpack`.
  }
}
```

crates/data-structures/src/error_stream.rs
```diff
@@ -0,0 +1,248 @@
...
+impl<E> AddAssign<E> for ErrorStream<E> {
+    fn add_assign(&mut self, error: E) {
+        self.0.push(error);
+    }
+}
+impl<E> AddAssign<ErrorStream<E>> for ErrorStream<E> {
+    fn add_assign(&mut self, other: ErrorStream<E>) {
+        self.0.extend(other.0);
+    }
+}
```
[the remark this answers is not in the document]
kazimuth: Yeah I thought about doing that, this was an idle thought. I'll swap it out.

crates/data-structures/src/error_stream.rs
kazimuth: I tried to impl for both `Result<T, E>` and `Result<T, ErrorStream<E>>`, but that fails because it overlaps. Lemme try that and see if it works.

kazimuth: Ah, you still run into issues. `(Result<A, ErrorStream<E>>, Result<B, ErrorStream<E>>)` ends up ambiguous. Should the output type be `Result<(A, B). ErrorStream<E>>` or `Result<(A, B), ErrorStream<ErrorStream<E>>>`?

gefjon: Darn! Oh well.

gefjon (approved):

PR #1544, opened 2024-07-25
core: Start system table ids at 1

kim:
Start all autoinc id columns within the system schema with 1 instead of
0.

Zero is the sentinel value to increment the autoinc sequence when
inserting, so using 0 as an actual value can lead to confusion and
subtle bugs.

1


kim:
Updated to start _all_ autoinc columns of system tables at 1.

bfops:
Wait, is this _not_ an ABI break?

kim:
@bfops Sorry, no, it's not. I initially thought it was, but we're not storing any id in the commitlog, and user-defined tables start at 4097 (not system tables + 1).

PR #1545, opened 2024-07-25
C#: ignore fields in partial declaration not marked with BSATN attribute

RReverser:
# Description of Changes

Some of the recent changes broke BitCraft which relies on being able to add extra fields to structures marked with `[SpacetimeDB.Type]`.

It wants those fields to be available in runtime, but ignored for BSATN and table purposes, which is a usecase I didn't consider in the past so it caused a few issues here and there. This one in particular is a regression from #1401.

corresponding GitHub label.

1

# Testing

- [x] Added a failing codegen test + verified that it's no longer failing after a fix.
- [ ] Run `dotnet test` as well, or try to load BitCraft with updated codegen DLL.

### DIFF
--- a/crates/bindings-csharp/BSATN.Codegen/Utils.cs
+++ b/crates/bindings-csharp/BSATN.Codegen/Utils.cs
@@ -151,9 +151,38 @@ public static class Utils
         }
     }
 
-    public static IEnumerable<IFieldSymbol> GetFields(INamedTypeSymbol type)
+    public static IEnumerable<IFieldSymbol> GetFields(
+        TypeDeclarationSyntax typeSyntax,
+        INamedTypeSymbol type
+    )
     {
-        return type.GetMembers().OfType<IFieldSymbol>().Where(f => !f.IsStatic);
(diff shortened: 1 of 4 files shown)


crates/bindings-csharp/BSATN.Codegen/Utils.cs
```diff
@@ -151,9 +151,38 @@ static string GetTypeInfoForNamedType(INamedTypeSymbol type)
...
+        // {
+        //     public int ExtraField;
+        // }
+        // ```
+        //
+        // In this scenario, only fields declared inside the declaration with the `[SpacetimeDB.Type]` attribute
+        // should be considered as BSATN fields, and others are expected to be ignored.
+        //
+        // To achieve this, we need to walk over the annotated type syntax node, collect the field names,
+        // and look up the resolved field symbols only for those fields.
```
bfops, line 179: :100:

bfops (approved):
seems smart to me. Excellent comment made review easier :rocket:

PR #1546, opened 2024-07-25
C#: remove `using SpacetimeDB.ClientApi;`

RReverser:
# Description of Changes

This using was added in the protobufectomy PR to all autogenerated files, even though it's unused in all of them but one (SpacetimeDBClient.cs).

As a result, it makes the autogen diff very noisy on large projects like BitCraft for no benefit.

This PR fixes it by only adding using in that single file where it's necessary.

corresponding GitHub label.

# Testing

- [x] `dotnet test` for snapshot testing


bfops (approved):
This worked correctly for me in BitCraftClient!

PR #1548, opened 2024-07-25
Use latest C# in BSATN.Runtime

RReverser:
# Description of Changes

Now that we include it by DLL rather than by source in the Unity SDK, we no longer need to limit ourselves to C# 9 - the only limitation is now the .NET Standard.

This allows to slightly clean up the code by leveraging newer syntactic features.

(best to review ignoring whitespace)

corresponding GitHub label.

# Testing


kazimuth (approved):
I am happy with this. From my reading there are no semantic changes, just formatting, right? Looks nice.

RReverser:
Yeah all changes are syntax upgrades suggested by IDE once I updated the lang version. Should be a non-functional change.

PR #1550, opened 2024-07-26
core: Singular system table names

kim:
For consistency reasons, name all system tables in singular form.

1


cloutiertyler (approved):
LGTM, thank you! A great weight has been lifted from my mind haha

PR #1558, opened 2024-07-31
core: Use `Program` instead of `(Hash, Box<[u8]>)`

kim:
Refactor to use `Program` (introduced in #1532) throughout.

Also avoid cloning the program bytes to instantiate a wasm module.
This makes the `AnyBytes` type unused, so remove it.

No

1 -- mostly mechanical

# Testing

n/a

### DIFF
--- a/crates/core/src/db/datastore/traits.rs
+++ b/crates/core/src/db/datastore/traits.rs
@@ -335,10 +335,7 @@ impl Program {
 
     /// Create a [`Program`] with no bytes.
     pub fn empty() -> Self {
-        Self {
-            hash: hash_bytes([].into()),
-            bytes: [].into(),
-        }
+        Self::from_bytes([])
     }
 }
 
--- a/crates/core/src/db/datastore/traits.rs
+++ b/crates/core/src/db/datastore/traits.rs
@@ -9,7 +9,7 @@ use crate::db::datastore::system_tables::ST_TABLE_ID;
 use crate::execution_context::ExecutionContext;
 use spacetimedb_data_structures::map::IntMap;
 use spacetimedb_lib::db::raw_def::*;
-use spacetimedb_lib::{Address, Identity};
+use spacetimedb_lib::{hash_bytes, Address, Identity};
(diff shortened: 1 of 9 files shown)


Centril (approved):
Love the red numbers!

crates/core/src/db/datastore/traits.rs
```diff
@@ -315,13 +315,34 @@ pub struct Metadata {
 }
 
 /// Program associated with a database.
+#[derive(Clone)]
```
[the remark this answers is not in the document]
kim: Ah yeah, thanks. I first tried to some `Arc`-ing, but then realized we only need to get the bytes across `spawn_rayon`.

crates/core/src/db/datastore/traits.rs
```diff
@@ -322,6 +322,26 @@ pub struct Program {
...
+    /// This computes the hash over `bytes`, so prefer constructing [`Program`]
+    /// directly if the hash is already known.
+    pub fn from_bytes(bytes: impl Into<Box<[u8]>>) -> Self {
+        let bytes = bytes.into();
+        let hash = hash_bytes(&bytes);
+        Self { hash, bytes }
+    }
+
+    /// Create a [`Program`] with no bytes.
+    pub fn empty() -> Self {
```
cloutiertyler, line 337: Technically the address of a zero byte program is: `c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470`

crates/core/src/db/datastore/traits.rs
```diff
@@ -322,6 +322,26 @@ pub struct Program {
...
+        Self {
+            hash: Hash::ZERO,
```
cloutiertyler, line 339: Not sure if this is used anywhere that would matter, but couldn't hurt to do this:

```suggestion
            hash: hash_bytes([].into()),
```

cloutiertyler (approved):
LGTM now

PR #1561, opened 2024-08-01
Get rid of python SDK

Centril:
# Description of Changes & API and ABI breaking changes

Removes python SDK, which we already regard as unsupported and unmaintained.

1


kim (approved):

kim:
We should probably mark https://github.com/clockworklabs/spacetimedb-python-sdk as archived.

PR #1562, opened 2024-08-01
Minor NFC improvements to C# Roslyn codegen

RReverser:
# Description of Changes

 - Switch to C#11 raw string literals. Raw string literals automatically ignore indentation and can be correctly auto-formatted (although there are still some issues, eg https://github.com/belav/csharpier/issues/1308), which makes them more suitable for multiline templates. This commit shows up as the most complex diff due to changes like `{` -> `{{` and formatting, but all the transformations were automatically done by IDE action.
 - Use StringBuilder for building up the codegen output where possible. Aside from being more efficient for appending literal strings, it has specialized overload for interpolated strings as well, avoiding bunch of intermediate buffers (https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/6.0/stringbuilder-append-evaluation-order).
 - Extract more of the common compilation options to directory-level C# config.

corresponding GitHub label.

# Testing


crates/bindings-csharp/Runtime/Internal/Module.cs
```diff
@@ -258,14 +258,15 @@ Buffer args
...
-            );
+            reducers[(int)id]
+                .Invoke(
+                    reader,
+                    new(
+                        Identity.From(caller_identity.Consume()),
+                        Address.From(caller_address.Consume()),
+                        timestamp.ToStd()
+                    )
+                );
```
kazimuth, line 269: This seems to be an autoformatter change.
We might want to add a test to CI to validate that all C# checked in to the repo is formatted with some formatter? (in another PR).

RReverser: Yeah I want to add that too. I'm using CSharpier but this particular change happened because I upgraded their version locally. I'll pin its version & add CI in a separate PR.

kazimuth (approved):
Looks good. Sadly the GitHub syntax highlighting doesn't seem to work quite as well with the new syntax, but oh well.

crates/bindings-csharp/Runtime/Internal/Module.cs
RReverser: @kazimuth Here goes https://github.com/clockworklabs/SpacetimeDB/pull/1567

PR #1567, opened 2024-08-06
NFC: enforce style for C# locally and on CI

RReverser:
# Description of Changes

As suggested in the other PR, adds C# style checks to the CI so that minor issues and reformatting are fixed early.

corresponding GitHub label.

1

# Testing

### DIFF
--- a/crates/bindings-csharp/Codegen/Module.cs
+++ b/crates/bindings-csharp/Codegen/Module.cs
@@ -176,7 +176,7 @@ public class Module : IIncrementalGenerator
                 transform: (context, ct) =>
                 {
                     var tableSyntax = (TypeDeclarationSyntax)context.TargetNode;
-                    var table = context.SemanticModel.GetDeclaredSymbol(tableSyntax)!;
+                    var table = context.SemanticModel.GetDeclaredSymbol(tableSyntax, ct)!;
 
                     var fields = GetFields(tableSyntax, table)
                         .Select(f =>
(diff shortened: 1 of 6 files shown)


crates/bindings-csharp/Runtime/Runtime.csproj
```diff
@@ -26,7 +26,7 @@
 
   <ItemGroup>
     <None Include="README.md" Pack="true" PackagePath="" />
-    <None Include="build/" Pack="true" PackagePath="build" />
+    <None Include="build/*" Pack="true" PackagePath="build" />
```
kazimuth, line 29: Is this a recursive glob?

crates/bindings-csharp/Codegen/Module.cs
```diff
@@ -176,7 +176,7 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
                 transform: (context, ct) =>
                 {
                     var tableSyntax = (TypeDeclarationSyntax)context.TargetNode;
-                    var table = context.SemanticModel.GetDeclaredSymbol(tableSyntax)!;
+                    var table = context.SemanticModel.GetDeclaredSymbol(tableSyntax, ct)!;
```
bfops, line 179: this looks like an actual runtime change, yes? Why is that?

bfops: (I removed the `no-runtime-change` label)

RReverser: Well as it's purely codegen, technically it's still compile-time change, and one that shouldn't matter at that... but I don't have a strong opinion on that.

crates/bindings-csharp/Runtime/Runtime.csproj
RReverser: No, but we don't have any nesting there, just 2 files.

bfops (approved):

PR #1572, opened 2024-08-07
ABI v9 validation code

kazimuth:
# Description of Changes

WIP validation code for moduledefv9, plus the definition for `ModuleDef`. The new module-def should also be backwards-compatible with `V8BackCompat` and the existing `TableSchema`, `Header`, etc types. I have not implemented those bridges yet.

A few API changes in components I have recently added, otherwise this is just adding code.

1, there is a good amount of code but I am generally happy with the redesigned interfaces. It's fun playing with my new multi-error combinator.

# Testing

Test porting is WIP, I have more proptest helpers now and I am going to use those to fuzz this once I've got the rest of the stuff ported.


crates/schema/src/def/validate/v9_old.rs
```diff
@@ -0,0 +1,921 @@
+use itertools::Itertools;
```
kazimuth, line 1: This file will be removed before this is undrafted, just here while I'm porting it.

crates/schema/src/migrate.rs
```diff
@@ -0,0 +1,568 @@
+/*
```
kazimuth, line 1: This file is also being ported and I think I will leave it to a followup PR.

crates/sats/src/algebraic_type.rs
```diff
@@ -300,6 +327,38 @@ impl AlgebraicType {
             _ => None,
         }
     }
+
+    /// Non-recursively validates that the type is in "nominal normal form".
```
kazimuth, line 331: This function is in fact recursive, so this should be changed to:
`Validates that the type is in "nominal normal form". Does not follow Refs.`

crates/schema/src/def/validate/v9.rs
```diff
@@ -0,0 +1,437 @@
...
+        .into_iter()
+        .map(|ty| {
+            validator
+                .validate_type_def(ty)
+                .map(|type_def| (type_def.name.clone(), type_def))
+        })
+        .collect_all_errors::<HashMap<_, _>>()
+        .and_then(|types| {
+            // We need to validate the typespace after we have all the type definitions.
+            // This is because we need to check that every type in the typespace has a corresponding
```
kazimuth, line 55: Every `Product`, `Sum`, and `Ref` type in the typespace, actually

crates/schema/src/def/validate/v9.rs
```diff
@@ -0,0 +1,437 @@
...
+
+        let params_valid: Result<()> = params
+            .iter()
+            .enumerate()
+            .map(|(position, param)| {
+                self.validate_algebraic_type(
+                    TypeLocation::ReducerArg {
+                        reducer_name: (&*name).into(),
+                        position,
+                        arg_name: param.name().map(Into::into),
```
kazimuth, line 264: Add comment: this does not allocate, since `TypeLocation` stores a `Cow`.

crates/schema/src/error.rs
```diff
@@ -0,0 +1,181 @@
...
+
+use crate::identifier::Identifier;
+
+pub type ValidationErrors = ErrorStream<ValidationError>;
+
+#[derive(thiserror::Error, Debug, PartialOrd, Ord, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum ValidationError {
+    #[error("name `{name}` is used for multiple entities")]
+    DuplicateName { name: Identifier }, // use an Identifier because this is checked for validated names.
```
kazimuth, line 19: add: most other variants of this enum store un-validated strings, since we want to support giving an error message even if names aren't valid.

crates/schema/src/def/validate/v9.rs
```diff
@@ -0,0 +1,437 @@
...
+            schedule,
+            table_type,
+            table_access,
+        })
+    }
+
+    /// Validate a column.
+    ///
+    /// Note that this accepts a `ProductTypeElement` rather than a `ColumnDef`,
+    /// because all information about tables is stored in the `Typespace` in ABI version 9.
```
kazimuth, line 172: About columns, rather

kazimuth:
@Centril @cloutiertyler This is ready for review.

kazimuth:
I didn't touch C#, so not sure where those SDK failures are coming from.

Centril (commented):
Partial review; more to come.

crates/lib/src/db/raw_def/v9.rs
```diff
@@ -34,19 +41,21 @@ pub type RawIdentifier = Box<str>;
...
     ///
-    /// Any `Product` or `Sum` types used transitively by the module MUST be declared in this typespace.
-    ///
-    /// Every `Product`, `Sum`, and `Ref` type in this typespace MUST have a corresponding `RawTypeDefV9` declaration in the `types` field, with a module-unique name.
+    /// The typespace must satisfy `Typespace::is_nominal_normal_form`. That is, all types referenced by this typespace must either:
+    /// 1. satisfy `AlgebraicType::is_nominal_normal_form`,
+    /// 2. or be a `Sum`/`ProductType`s whose fields satisfy the same.
     ///
-    /// All product and sum types in this typespace MUST have the [default element ordering](crate::db::default_element_ordering) UNLESS they declare a custom ordering via their `RawTypeDefV9`.
+    /// Types satisfying condition 2 are called *non-nominal types*.
```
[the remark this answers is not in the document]
kazimuth: It's definitely a bad name because that's not what it's supposed to imply... Open to suggestions.

What I'm calling a "nominal" type is something that can be used by client codegen to generate a reference to a type. If you have a `Ref(...)`, of course, then you generate `"{type_names[ref_]}"` in the client. Similarly, if you have `Array(Ref(...))`, you generate `"Array<{type_names[ref_]}>"`, etc.
There are also some special types, which look structural, but have special reference-generating code, like `AlgebraicType::option(AlgebraicType::ref(...))` or `AlgebraicType::identity`.

What I'm calling a "non-nominal" type is something that can be used by client codegen to generate a type *definition*. That is, something in the module's `typespace` with an attached name. Non-nominal is really not the right word for this.

crates/lib/src/db/raw_def/v9.rs
```diff
@@ -270,15 +193,16 @@ pub struct RawUniqueConstraintDefV9 {
...
+/// - `scheduled_id` of type `u64`.
+/// - `scheduled_at` of type `ScheduleAt`.
 #[derive(Debug, Clone, ser::Serialize, de::Deserialize)]
 #[cfg_attr(feature = "test", derive(PartialEq, Eq, PartialOrd, Ord))]
 pub struct RawScheduleDefV9 {
     /// The name of the schedule. Must be unique within the containing `RawDatabaseDef`.
     pub name: RawIdentifier,
 
-    /// The column that stores the desired invocation time.
-    pub at_column: ColId,
```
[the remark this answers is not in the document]
kazimuth: Currently all downstream code assumes the columns are named `scheduled_at` and `scheduled_id`, so I decided to just look for those column names instead. We're discussing what the API should look like in #1577

crates/sats/src/product_type.rs
```diff
@@ -88,6 +90,16 @@ impl ProductType {
...
     }
+
+    /// Iterate over the `ColId`s for this `ProductType`.
+    pub fn col_ids(&self) -> impl Iterator<Item = ColId> + '_ {
+        (0..self.elements.len()).map(ColId::from)
+    }
+
+    /// Get a column of this `ProductType` by `ColId`.
+    pub fn get_column(&self, col_id: impl Into<ColId>) -> Option<&ProductTypeElement> {
+        self.elements.get(col_id.into().0 as usize)
```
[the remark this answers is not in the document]
kazimuth: ah, i knew there had to be an API for this but missed this. 👍

kazimuth:
@Centril, please review #1584, #1584, #1586 before coming back to this one.

crates/schema/src/error.rs
```diff
@@ -0,0 +1,189 @@
...
+        ref_: AlgebraicTypeRef,
+    },
+    #[error("Type {type_name} has invalid ref: {ref_}")]
+    InvalidTypeRef {
+        type_name: RawIdentifier,
+        ref_: AlgebraicTypeRef,
+    },
+    #[error("A scheduled table must have columns `scheduled_id: u64` and `scheduled_at: ScheduledAt`, but table `{table}` has columns {columns:?}")]
+    ScheduledIncorrectColumns { table: RawIdentifier, columns: ProductType },
+    #[error("{location} is not in nominal normal form: {ty:?}")]
```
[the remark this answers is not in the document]
kazimuth: Yeah it's not great.
This error should never happen to users using the official SDKs though, it would only happen if the SDK is generating malformed ModuleDefs. Nothing the user can do should be able to cause this, it's really an error for SDK developers.

crates/schema/src/def/validate/v9.rs
```diff
@@ -0,0 +1,1124 @@
...
+    ) -> Result<ScheduleDef> {
+        let RawScheduleDefV9 { name, reducer_name } = schedule;
+
+        // Find the appropriate columns.
+        let at_column = table_in_progress
+            .product_type
+            .elements
+            .iter()
+            .enumerate()
+            .find(|(_, element)| element.name() == Some("scheduled_at"));
```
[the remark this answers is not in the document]
kazimuth: Moved to #1590

crates/lib/src/db/raw_def/v9.rs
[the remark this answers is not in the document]
kazimuth: Fixed in #1590

kazimuth:
@Centril @cloutiertyler I consider this ready for merge.

crates/lib/src/db/raw_def/v9.rs
```diff
@@ -174,22 +95,38 @@ pub struct RawTableDefV9 {
...
     /// Must be a valid [crate::db::identifier::Identifier].
     pub name: RawIdentifier,
+
     /// A reference to a `ProductType` containing the columns of this table.
     /// This is the single source of truth for the table's columns.
     /// All elements of the `ProductType` must have names.
     ///
-    /// Like all types in the module, this must have the [default element ordering](crate::db::default_element_ordering), UNLESS a custom ordering is declared via `ModuleDef.misc_exports` for this type.
+    /// Like all types in the module, this must have the [default element ordering](crate::db::default_element_ordering),
+    /// UNLESS a custom ordering is declared via `ModuleDef.misc_exports` for this type.
```
kazimuth, line 104:
```suggestion
    /// UNLESS a custom ordering is declared via a `RawTypeDefv9` for this type.
```

crates/lib/src/db/raw_def/v9.rs
```diff
@@ -315,8 +294,179 @@ pub struct RawReducerDefV9 {
...
+
+impl RawModuleDefV9Builder {
+    /// Create a new, empty `RawModuleDefBuilder`.
+    pub fn new() -> Self {
+        Default::default()
+    }
+
+    /// Add a type to the in-progress module.
+    ///
+    /// The returned type must satisfy `AlgebraicType::is_nominal_normal_form`
```
kazimuth, line 318:
```suggestion
    /// The returned type must satisfy `AlgebraicType::is_valid_for_client_type_definition` or  `AlgebraicType::is_valid_for_client_type_use` .
```

crates/lib/src/db/raw_def/v9.rs
```diff
@@ -315,8 +294,179 @@ pub struct RawReducerDefV9 {
...
+            btree_map::Entry::Vacant(v) => {
+                // Bind a fresh alias to the unit type.
+                let slot_ref = self.module.typespace.add(AlgebraicType::unit());
+                // Relate `typeid -> fresh alias`.
+                v.insert(slot_ref);
+
+                // Alias provided? Relate `name -> slot_ref`.
+                if let Some(name) = name {
+                    // Right now, we just split the name on patterns likely to split up module paths.
+                    // TODO(1.0): build namespacing directly into the bindings macros so that we don't need to do this.
```
kazimuth, line 442: Note: this will not be an ABI break because we can rewrite the `TypespaceBuilder` trait when we are redoing the Rust binding macros to return v9. `TypespaceBuilder` lives *inside* the module, so it is not part of the ABI. It is part of the API surface touched by the macros, so this will be an *API* break, but it's one of those weird actually-private-despite-being-public APIs you get with macros.

crates/schema/src/def.rs
```diff
@@ -0,0 +1,650 @@
...
+            .stored_in_table_def(&self.table_name)
+            .expect("validated column");
+        let product_type = module_def.typespace()[table_def.product_type_ref]
+            .as_product()
+            .expect("validated column");
+        &product_type.elements[self.col_id.idx()].algebraic_type
+    }
+}
+
+/// Requires that the projection of the table onto these columns is an bijection.
```
kazimuth, line 413:
```suggestion
/// Requires that the projection of the table onto these columns is an bijection.
///
/// That is, there must be a one-to-one relationship between a row and the `columns` of that row.
```

crates/schema/src/def.rs
```diff
@@ -0,0 +1,650 @@
...
+        let scope = scope
+            .into_iter()
+            .map(|chunk| Identifier::new(chunk).map_err(ErrorStream::from))
+            .collect_all_errors();
+        let name = Identifier::new(name.into()).map_err(ErrorStream::from);
+        let (scope, name) = (scope, name).combine_errors()?;
+        Ok(ScopedTypeName { scope, name })
+    }
+
+    pub fn from_name(name: Identifier) -> Self {
```
kazimuth, line 498:
```suggestion
    /// Create a new `ScopedTypeName` with an empty scope.
    pub fn from_name(name: Identifier) -> Self {
```

crates/schema/src/def.rs
```diff
@@ -0,0 +1,640 @@
...
+    tables: IdentifierMap<TableDef>,
+
+    /// The reducers of the module definition.
+    reducers: IdentifierMap<ReducerDef>,
+
+    /// The type definitions of the module definition.
+    types: HashMap<ScopedTypeName, TypeDef>,
+
+    /// The typespace of the module definition.
+    typespace: Typespace,
```
kazimuth, line 85: When we add `AlgebraicTypeUse` and `AlgebraicTypeDefinition`, we may want to change this to a datastructure that stores those for client codegen. On the other hand, I think the database engine still wants to think about algebraic types structurally, so we may be in the uncomfortable position of needing both.

crates/schema/src/def.rs
```diff
@@ -0,0 +1,640 @@
...
+            }
+        }
+    }
+}
+
+impl TryFrom<RawModuleDef> for ModuleDef {
+    type Error = ValidationErrors;
+
+    fn try_from(raw: RawModuleDef) -> Result<Self, Self::Error> {
+        Self::validate(raw)
```
RReverser, line 200: Nit but I'd just inline the `validate` implementation here; seems pointless to have custom `ModuleDef::validate` when a standard API will suffice.

crates/schema/src/def.rs
```diff
@@ -0,0 +1,640 @@
...
+    pub sequences: IdentifierMap<SequenceDef>,
+
+    /// The schedule for the table, if present.
+    pub schedule: Option<ScheduleDef>,
+
+    /// Whether this is a system- or user-created table.
+    pub table_type: StTableType,
+
+    /// Whether this table is public or private.
+    pub table_access: StAccess,
```
RReverser, line 288: Could we change those to be enums in the new version?

Those 2 being special strings has always been awkward to work with in the SDKs. The only argument for them being strings I heard is extensibility, but 1) we can add more variants to the enums just as easily and 2) we have ModuleDef versioning now anyways.

crates/schema/src/def.rs
```diff
@@ -0,0 +1,640 @@
...
+use spacetimedb_lib::db::raw_def::v9::{RawIdentifier, RawScopedTypeNameV9};
+use spacetimedb_lib::{ProductType, RawModuleDef};
+use spacetimedb_primitives::{ColId, ColList};
+use spacetimedb_sats::AlgebraicType;
+use spacetimedb_sats::{AlgebraicTypeRef, Typespace};
+
+pub mod validate;
+
+/// A map from `Identifier`s to values of type `T`.
+pub type IdentifierMap<T> = HashMap<Identifier, T>;
```
RReverser, line 36: As far as I can tell, all the types you store here implement your nice `ModuleDefLookup` trait and already allow to get a map key from `T`.

One thing I've done in the past for similar scenarios is use `HashSet` instead + a newtype that implements `Eq + Hash + Borrow<Key>` by delegating to the key comparisons.

That way you should be able to avoid a lot of allocations in all these `(foobar.name.clone(), foobar)` KV construction lines, as well as save some memory by simply reusing `foobar` itself both for the key and the value parts of each entry.

Idk how critical it is, so feel free to leave this as a future optimisation, just thought I'd throw it out there as I noticed a lot of those clones in the PR.

kazimuth: Oh that's a great idea. That should get rid of almost all of these clones and get rid of some possible invalid states.

I'm tempted to leave it as a future optimization for now though just in favor of getting this merged.

crates/schema/src/def.rs
kazimuth: Uh, these are enums. Possibly they are serialized as strings?

RReverser: Yeah they are enums in Rust but their SATS implementations manually serialise them as strings.

Since it's a ModuleDef-breaking change, I figured this PR would be a good place to introduce new versions of these enums with the standard enum serialisation instead.

kazimuth: Ahh, I missed that entirely. I'll make the change.

crates/schema/src/def.rs
cloutiertyler: I think in the future we will want to think about how we can implement nominal types within a structural type system on a whollistic level.

crates/schema/src/def.rs
cloutiertyler: Agreed with this by the way

cloutiertyler (approved):
I don't see any significant issues with this change, provided you address the things you've already responded to. Excited to merge!

crates/schema/src/def.rs
kazimuth: Yeah, I think that would be very worthwhile.

PR #1574, opened 2024-08-07
Updating the change date in anticipation of the anniversary

cloutiertyler:
# Description of Changes

I'm updating the change date for the license. The change date is 5 years out, even though the code becomes open source after 4 years so that we don't have to constantly update the change date.

# Testing


bfops (approved):
Okay, seems sane to me!

PR #1576, opened 2024-08-08
Verify Unity compatibility in the shared C# codegen

RReverser:
# Description of Changes

So far, the codegen tests verified that `[SpacetimeDB.Type]` and `[SpacetimeDB.Table]` markers result in expected generated methods, and that the generated code is valid and can be compiled successfully.

They did this by combining snapshot testing with a C# module example that has those attributes applied to a variety of types.

This works well enough, but now that C# codegen is split into two parts - BSATN.Codegen and Codegen - and the former is also used for C# clients like Unity, we want to make sure they work independently.

We also want to validate that the shared part that handles only the `[SpacetimeDB.Type]` attributes generates code compatible with Unity limitations - that is, it must be limited to .NET Standard 2.1 APIs and C# 9 syntax.

In this PR, I'm doing just that - creating two separate project fixtures. `client` one has limited .NET and C# versions, and only runs the BSATN.Codegen generator for the `[Type]` markers, whereas the `server` one does everything the old test did and tests modules with types, tables and reducers.

In the process I also simplified the way test projects are built by using a programmatic MSBuild API instead of manually building and collecting necessary DLLs into a compilation object. This brings test fixtures much closer to real-world projects, simplifies test maintainability, while still allowing to do all the low-level checks like those for codegen caching.

corresponding GitHub label.

# Testing

- [x] `dotnet test`


bfops:
FYI @RReverser this is giving cryptic failures on the C# SDK tests, idk if that's expected.

RReverser:
> Edit: Nvm that's not related to this PR.

@bfops So is this one good to go then?

kazimuth (approved):
This makes sense, good to split the tests apart. The microsoft plumbing seems so much more pleasant than the proc-macro plumbing in Rust. I have no particular comments on the implementation, unfortunately.

PR #1582, opened 2024-08-13
Fix [iu](128|256) BSATN implementations in C#

RReverser:
# Description of Changes

128- and, as of recently, 256-bit integers were implemented incorrectly in C#. They would write upper half of the number first, followed by the lower part, making the resulting encoding mixed-endian.

Unfortunately, this flew under the radar because SDK tests only verify round-trip: reducer arguments are passed in via BSATN encoding, and the table events are also retrieved in BSATN encoding.

As a result, if any encoding implementation is incorrect in an idempotent way - that is, you can only observe the bug from inside a module via stringification or arithmetic operations, but roundtrip restores the correct value - the SDK test doesn't see anything wrong with it, as what happened here.

I added one more test that, instead of round-tripping, sends values to the module and expects it to stringify them and insert strings into a table. This allows us to see the deserialized values as the module does, without going back via the original BSATN.

To implement this, I added proper stringification of large integers to C#, going via BigInteger conversion, which can be somewhat costly, but unlikely to matter in ToString().

I also extended sample integers in the primitive struct example to fill all bytes, as previously one half of int128/int256 would be unused and stay zero, which can also hide potential issues.

I want to extract common parts later and perhaps extend tests to do real property testing in C# instead of simply hardcoding values, but for now I'm satisfied this particular issue is fixed so wanted to submit as a separate PR.

corresponding GitHub label.

# Testing


gefjon (approved):
Have looked at only the tests, but those look good.

PR #1583, opened 2024-08-13
[CI] Do not use branch names when publishing container images

kurtismullins:
# Description of Changes

Removes the creation of Docker Container Images with branch names as tags.

**AMD/Intel Image Build**
Screenshot 2024-08-13 at 12 57 01 PM

**ARM Image Build**
Screenshot 2024-08-13 at 1 00 10 PM

**Both images merged into a multi-platform image**
Screenshot 2024-08-13 at 1 01 28 PM

**The created images in Dockerhub**
Screenshot 2024-08-13 at 1 02 35 PM

Side Note: It is much nicer to cross-reference related/identical images in Github's Artifact Repository than it is in Dockerhub. Perhaps we should consider only using Dockerhub for `spacetimedb:latest` and tagged releases?

n/a

1

n/a

# Testing

- See the included screenshots. I created tags which matched the blob pattern `v*` and verified that the tag names are correct by looking at the CI output and cross-referencing DockerHub.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -25,7 +25,6 @@ jobs:
           images: |
             clockworklabs/spacetimedb
           tags: |
-            type=ref,event=branch,prefix=branch-
             type=sha,prefix=commit-
             type=ref,event=tag
             type=sha,suffix=-amd64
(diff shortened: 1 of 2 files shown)


cloutiertyler (approved):
LGTM

cloutiertyler (approved):
LGTM

cloutiertyler (approved):
LGTM

PR #1584, opened 2024-08-13
Preliminary Identifier validation

kazimuth:
# Description of Changes

Extracted from #1572. This implements validation for identifiers used in a module. Sorry for the delay @Centril, I got sidetracked yesterday.
It's only a preliminary implementation -- we can add more validation later once the identifier proposal is merged.

N/A

0

# Testing

As seen in PR.


crates/schema/src/identifier.rs
```diff
@@ -125,6 +130,10 @@ mod tests {
     proptest! {
         #[test]
         fn test_standard_ascii_identifiers(s in "[a-zA-Z_][a-zA-Z0-9_]*") {
+            // Ha! Proptest will reliably find these.
+            if Identifier::is_reserved(&s) {
+                return Ok(());
+            }
```
gefjon, line 136: Proptest has a way of skipping cases: https://docs.rs/proptest/latest/proptest/test_runner/enum.TestCaseError.html .

gefjon: https://docs.rs/proptest/latest/proptest/macro.prop_assume.html

PR #1585, opened 2024-08-13
Add `try_from_iter` to `ColList` & fix bug in `ColListBuilder`

kazimuth:
# Description of Changes

I originally had some `ColId`-related methods added to `ProductType` as well, but then I decided that `sats` really shouldn't care about database notions like Column IDs. So this only adds a single `ColList` method. I also found a bug! `ColListBuilder` was using `is_empty` to check if it should error, but `is_empty` always returns `false`.

N/A

0

This does add `itertools` as a dependency of `primitives`, unfortunately.

# Testing

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4464,6 +4464,7 @@ version = "0.12.0"
 dependencies = [
  "bitflags 2.4.1",
  "either",
+ "itertools 0.12.0",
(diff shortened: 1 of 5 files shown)


kazimuth:
Perf, this isn't needed then.

PR #1588, opened 2024-08-14
Allow empty `ColList`

Centril:
# Description of Changes

Fix the bug found in https://github.com/clockworklabs/SpacetimeDB/pull/1585 by allowing `ColList`s to be empty, as we actually try to use the type as a possibly-empty list in some places (which is why the tests failed in #1585).

None

2, some unsafe related code affected, but there's less unsafe code now.

# Testing

- [x] Tests were extended to test the empty aspect and the new functionality.
- [x] All the ColList proptests were run in miri.

### DIFF
--- a/crates/primitives/src/col_list.rs
+++ b/crates/primitives/src/col_list.rs
@@ -19,79 +19,19 @@ use either::Either;
 
 /// Constructs a `ColList` like so `col_list![0, 2]`.
 ///
-/// A head element is required.
 /// Mostly provided for testing.
 #[macro_export]
 macro_rules! col_list {
-    ($head:expr $(, $elem:expr)* $(,)?) => {{
-        let mut list = $crate::ColList::new($head.into());
-        $(list.push($elem.into());)*
-        list
+    ($($elem:expr),* $(,)?) => {{
+        [$($elem),*].into_iter().collect::<$crate::ColList>()
     }};
 }
 
-/// An error signalling that a `ColList` was empty.
-#[derive(Debug)]
-pub struct EmptyColListError;
-
-/// A builder for a [`ColList`] making sure a non-empty one is built.
-pub struct ColListBuilder {
-    /// The in-progress list.
-    list: ColList,
-}
-
-impl Default for ColListBuilder {
-    fn default() -> Self {
-        Self::new()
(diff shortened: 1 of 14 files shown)


Cargo.toml
```diff
@@ -81,7 +81,10 @@ debug = true
 version = "0.12.0"
 edition = "2021"
 # update rust-toolchain.toml too!
-rust-version = "1.78.0"
+rust-version = "1.80.1"
```
kazimuth, line 84: I wonder if we could get a badge for this on the top line of the main README

crates/client-api/src/routes/tracelog.rs
```diff
@@ -1,107 +0,0 @@
-use axum::body::Bytes;
```
kazimuth, line 1: Slightly confused how this file ended up here, maybe needs rebase?

crates/primitives/src/col_list.rs
```diff
@@ -19,79 +19,18 @@ use either::Either;
...
-        }
-        builder
-    }
-}
-
-/// This represents a non-empty list of [`ColId`]
+/// This represents a list of [`ColId`]s
 /// but packed into a `u64` in a way that takes advantage of the fact that
 /// in almost all cases, we won't store a `ColId` larger than 62.
 /// In the rare case that we store larger ids, we fall back to a thin vec approach.
```
kazimuth, line 33:
```suggestion
/// In the rare case that we store larger ids, we fall back to a thin vec approach.
/// We also fall back to a thin vec if the ids stored are not in sorted order, from low to high.
```

kazimuth (approved):
Looks good. Nice to be able to use `collect`.

I was sketching an implementation of `sort` as well, since I need it for finding indexes for unique constraints, but it's a little too much added code to do with suggestions.
I'll open a followup PR once this is merged.

gefjon (dismissed):
> Bump Rust to 1.80.1 as this turned out to be required due to compiler bugs on 1.78.

I would like a more detailed description than this, and would also like to know why bumping to 1.79 is insufficient. I would also strongly prefer the version bump happen in a separate PR, rather than sneaking it in alongside another change.

PR #1590, opened 2024-08-14
Make `ScheduleAt` special + `Typespace::is_valid_for_client_code_generation`

kazimuth:
# Description of Changes

This PR does 2 things:
- It makes the `ScheduleAt` type special, and reworks the `is_special` infrastructure a little. Now we have `SumType::is_special` and `AlgebraicType::is_special` in addition to `ProductType::is_special`. I also reordered `ScheduleAt` so that its variants are in alphabetical order, in preparation for the upcoming "default element ordering" changes.

- It renames some things in sats to get rid of the name "nominal normal form", which by all accounts
  was confusing to everybody. The new APIs are named:
  - `AlgebraicType::is_valid_for_client_type_definition`
  - `AlgebraicType::is_valid_for_client_type_use`
  - `Typespace::is_valid_for_client_code_generation`

This is an API and ABI break.

2. I have a sneaking suspicion this may break client code generation in odd ways.
The new `AlgebraicType::is_special` may also need to be integrated in various places.

# Testing
Running test suite, I'd also like to check how badly this breaks Bitcraft & the SDKs...


crates/sats/src/sum_type.rs
```diff
@@ -61,20 +70,50 @@ impl SumType {
...
         self.variants.is_empty()
     }
 
+    /// Return whether this sum type is the special `ScheduleAt` type,
+    /// `Interval(u64) | Time(u64)`.
+    /// Does not follow `Ref`s.
+    pub fn is_schedule_at(&self) -> bool {
+        match &*self.variants {
+            [first, second] => {
+                first.has_name("Interval")
```
Shubham8287, line 103: Nit: Seems you declared but forget to use Tags here?

crates/cli/src/subcommands/generate/csharp.rs
```diff
@@ -42,17 +42,31 @@ fn scalar_or_string_name(b: &AlgebraicType) -> Option<&str> {
 fn ty_fmt<'a>(ctx: &'a GenCtx, ty: &'a AlgebraicType, namespace: &'a str) -> impl fmt::Display + 'a {
     fmt_fn(move |f| match ty {
         AlgebraicType::Sum(sum_type) => {
-            // This better be an option type
-            if let Some(inner_ty) = sum_type.as_option() {
-                write!(f, "{}?", ty_fmt(ctx, inner_ty, namespace))
+            if sum_type.is_special() {
```
RReverser, line 45: Looking at those `if is_... { if let Some(...) ...` I get an impression that we would be better served by `as_special()` + an enum instead of a simple `is_special() -> bool`.

Then you could use it with a much cleaner code like

```
match ty.as_special() {
  Some(SpecialType::Identity) => ...,
  Some(SpecialType::Option(ty)) => ...,
  ...
  None => ...handle non-special types...
}
```

crates/cli/tests/snapshots/codegen__codegen_rust.snap
```diff
@@ -874,7 +855,11 @@ impl TableType for TestD {
 
 
 impl TestD {
-	}
+	#[allow(unused)]
+	pub fn filter_by_test_c(test_c: Option::<Namespace.testC>) -> TableIter<Self> {
```
RReverser, line 859: FWIW we decided not to allow filtering by optionals for now back in the consistent filtering proposal (https://github.com/clockworklabs/SpacetimeDBPrivate/pull/783).

So this method shouldn't have appeared.

crates/cli/tests/snapshots/codegen__codegen_rust.snap
```diff
@@ -693,6 +690,10 @@ impl RepeatingTestArg {
 	#[allow(unused)]
 	pub fn find_by_scheduled_id(scheduled_id: u64) -> Option<Self> {
 		Self::find(|row| row.scheduled_id == scheduled_id)
+}
+	#[allow(unused)]
+	pub fn filter_by_scheduled_at(scheduled_at: ScheduleAt) -> TableIter<Self> {
```
RReverser, line 695: This got added in Rust output, but not other langs, creating inconsistency between language APIs.

Let's exclude it for now - if we want to allow more types in addition to what we decided in the consistent filtering proposal, we should do this (ugh can't avoid repeating that word) consistently.

crates/cli/src/subcommands/generate/typescript.rs
```diff
@@ -40,16 +40,32 @@ fn scalar_or_string_to_ts(ty: &AlgebraicType) -> Option<(&str, &str)> {
...
-                unimplemented!()
+                panic!("Cannot format non-special sum type {sum_type:?}")
             }
         }
-        ty if ty.is_identity() => write!(f, "Identity"),
-        ty if ty.is_address() => write!(f, "Address"),
         // All other types should fail.
-        AlgebraicType::Product(_) => unimplemented!(),
+        AlgebraicType::Product(product_type) => {
+            if product_type.is_special() {
```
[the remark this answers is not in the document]
kazimuth: It is not redundant in the presence of code changes. If you add a new `special` type, and miss updating code generation, the old code won't catch it.

I'll revert in favor of doing the enum refactoring later, since that is really the correct way to express this.

crates/cli/src/subcommands/generate/csharp.rs
kazimuth: At that point we could bite the bullet and add `TypespaceForGenerate`, which should express exactly the typespaces that satisfy `is_valid_for_client_code_generation`.

crates/cli/tests/snapshots/codegen__codegen_rust.snap
kazimuth: Yeah, I was just updating everywhere `is_special` was used but it looks like the logic is structured a lil differently across generators. Will revert.

kazimuth:
I've fixed this partway but it looks like the typescript SDK needs to be updated with a `ScheduleAt` enum for this to merge. I am not sure of the right way to set that up, it looks like typescript has the client API auto-generated?

kazimuth:
https://github.com/clockworklabs/spacetimedb-typescript-sdk/pull/88 contains the typescript side of the new changes here. I still need to do an end-to-end test of this with that.

RReverser (approved):
Nits about using `enum SpecialType` or something aside, the snapshots look good now.

Centril (approved):

RReverser:
> * It makes the `ScheduleAt` type special

Did we ever add it to the C# SDK? When trying to regenerate BitCraft files, I'm getting compilation errors due to `SpacetimeDB.ScheduleAt` being undefined.

PR #1606, opened 2024-08-16
Add backwards-compatible validation for RawModuleDefV8.

kazimuth:
# Description of Changes

Plus unit and integration tests.
I'm still working through errors in the integration tests -- I think I'm going to need to something slightly more clever in a few places -- but generally this is ready for review.

N/A

This is tricky because we're adding a new code path that's supposed to replicate the behavior of another complex code path, but I think once we add the other end (ModuleDef -> TableSchema) comparing the results will be enough to validate that stuff works.

# Testing

I test on both synthetic v8 module definitions and actual module definitions from the repo.


crates/schema/src/def/validate/v8.rs
```diff
@@ -1,9 +1,716 @@
...
+    #[test]
+    fn not_nominal() {
+        let inner_not_nominal_type = AlgebraicType::product([("b", AlgebraicType::U32)]);
+        let invalid_type = AlgebraicType::product([("a", inner_not_nominal_type.clone())]);
+        let mut builder = RawModuleDefV8Builder::default();
+        builder.add_type_for_tests("Invalid", invalid_type.clone());
+        builder.add_reducer_for_tests("silly", ProductType::from([("a", invalid_type.clone())]));
+        let result: Result<ModuleDef> = builder.finish().try_into();
+
+        expect_error_matching!(result, ValidationError::NotNominalNormalForm { location, ty } => {
```
kazimuth, line 648: Note that these names will be changed once #1590 is merged

kazimuth:
@Centril @cloutiertyler I consider this ready for merge.

crates/schema/src/def/validate/v8.rs
```diff
@@ -1,9 +1,761 @@
...
+    let RawConstraintDefV8 {
+        constraint_name,
+        constraints,
+        columns,
+    } = constraint;
+
+    if constraints.has_primary_key() {
+        if let Some(col) = columns.as_singleton() {
+            if primary_key.is_none() {
+                *primary_key = Some(col);
```
cloutiertyler, line 260: What's the point of this. Seems unused? Or am I misunderstanding? Is this mutating the argument?

kazimuth: Yeah, I'll rewrite this with comments.

cloutiertyler (approved):
In principle this all seems fine to me.

crates/schema/src/def/validate/v9.rs
```diff
@@ -755,11 +769,73 @@ enum TypeDefOrUse {
     Use,
 }
 
+fn check_table_names_consistent_with_type_names(
```
cloutiertyler, line 772: Why do we have this requirement? Can I not have this:

```rust
#[table(name="logged_in_user")]
#[table(name="logged_out_user")]
struct User {
    ...
}
```

kazimuth: oh you're completely right, i was just pondering if I actually needed this. I will discard.

Issue #1460, opened 2024-06-26
Tracking issue for WASM ABI changes


Centril:
This issue is meant to track the implementation work necessary to adjust the WASM ABI to what is proposed in the proposal.

DETAILS TODO.

PR #1609, opened 2024-08-19
[WASM ABI 1.0] impl `__call_reducer__` using `bytes_source_read`

Centril:
# Description of Changes

This adds `_bytes_source_read` and implements `__call_reducer__` with it.
Meanwhile, `_buffer_len` and `_buffer_consume` are removed.

### DIFF
--- a/crates/bindings-csharp/Runtime/Internal/FFI.cs
+++ b/crates/bindings-csharp/Runtime/Internal/FFI.cs
@@ -7,6 +7,15 @@ using System.Runtime.InteropServices.Marshalling;
 // forwarding in the codegen for `__describe_module__` and `__call_reducer__` exports which both
 // use this type.
 [StructLayout(LayoutKind.Sequential)]
+public readonly record struct BytesSource(uint Handle)
+{
+    public static readonly BytesSource INVALID = new(0);
+}
+
+// This type is outside of the hidden `FFI` class because for now we need to do some public
+// forwarding in the codegen for `__describe_module__` and `__call_reducer__` exports which both
+// use this type.
+[StructLayout(LayoutKind.Sequential)]
 [NativeMarshalling(typeof(Marshaller))]
 public readonly record struct Buffer(uint Handle)
 {
@@ -49,6 +58,7 @@ internal static partial class FFI
             LOOKUP_NOT_FOUND = 2,
             UNIQUE_ALREADY_EXISTS = 3,
             BUFFER_TOO_SMALL = 4,
+            NO_SUCH_BYTES = 8,
         }
 
         // This custom marshaller takes care of checking the status code
@@ -75,6 +85,7 @@ internal static partial class FFI
                     Errno.LOOKUP_NOT_FOUND => new LookupNotFoundException(),
                     Errno.UNIQUE_ALREADY_EXISTS => new UniqueAlreadyExistsException(),
                     Errno.BUFFER_TOO_SMALL => new BufferTooSmallException(),
+                    Errno.NO_SUCH_BYTES => new NoSuchBytesException(),
                     _ => new UnknownException(status),
                 };
             }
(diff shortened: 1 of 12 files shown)


crates/bindings-csharp/Runtime/Internal/FFI.cs
```diff
@@ -3,6 +3,30 @@ namespace SpacetimeDB.Internal;
...
+// This type is outside of the hidden `FFI` class because for now we need to do some public
+// forwarding in the codegen for `__describe_module__` and `__call_reducer__` exports which both
+// use this type.
+[StructLayout(LayoutKind.Sequential)]
+[NativeMarshalling(typeof(Marshaller))]
+public readonly record struct BytesSource(uint Handle)
+{
+    public static readonly BytesSource INVALID = new(0);
+
+    // TODO(centril): actually, we don't return this by value, so can we remove this?
```
RReverser, line 15: Yeah if we don't pass it by value, it should work without custom marshaller.

crates/bindings-csharp/Runtime/Internal/Module.cs
```diff
@@ -218,12 +218,54 @@ public static void RegisterTable<T>()
...
-        FFI._buffer_consume(buffer, result, len);
-        return result;
+        if (source == BytesSource.INVALID) {
+            return new byte[0];
+        }
+        else
+        {
+            byte[] buffer = new byte[0x20_000];
+            uint written = 0;
+            while (true)
```
RReverser, line 230: Haven't used it myself, but it looks like [ArrayBufferWriter](https://learn.microsoft.com/en-us/dotnet/api/system.buffers.arraybufferwriter-1?view=net-8.0) could help reduce some boilerplate here.

Its `GetSpan(extraSize)` performs automatic resizing and returns a buffer to write into.

crates/bindings-csharp/Runtime/Internal/Module.cs
```diff
@@ -218,12 +218,54 @@ public static void RegisterTable<T>()
...
 
-    private static byte[] Consume(this Buffer buffer)
+    private static byte[] Consume(this BytesSource source)
     {
-        var len = FFI._buffer_len(buffer);
-        var result = new byte[len];
-        FFI._buffer_consume(buffer, result, len);
-        return result;
+        if (source == BytesSource.INVALID) {
+            return new byte[0];
```
RReverser, line 224:
```suggestion
            return [];
```

crates/bindings-csharp/Runtime/Internal/FFI.cs
```diff
@@ -183,13 +209,10 @@ uint message_len
...
-    public static partial uint _buffer_len(Buffer buf_handle);
-
-    [LibraryImport(StdbNamespace)]
-    public static partial void _buffer_consume(
-        Buffer buf_handle,
-        [MarshalUsing(CountElementName = nameof(dst_len))] [Out] byte[] dst,
-        uint dst_len
+    public static unsafe partial short _bytes_source_read(
+        BytesSource source,
+        byte* buffer_ptr,
```
RReverser, line 214: I think you can use `Span<byte>` here and then you don't need to mark this function as `unsafe`, and the caller doesn't need to do the `unsafe`+`fixed` dance either - `LibraryImport` attribute takes care of that during marshalling.

crates/bindings-csharp/Runtime/Internal/Module.cs
RReverser: Leaving this for later.

crates/bindings/src/rt.rs
```diff
@@ -451,8 +457,59 @@ extern "C" fn __call_reducer__(
...
-    let args = args.read();
-    reducers[id](ctx, &args)
+    with_read_args(args, |args| reducers[id](ctx, args))
+}
+
+/// Run `logic` with `args` read from the host into a `&[u8]`.
+fn with_read_args<R>(args: BytesSource, logic: impl FnOnce(&[u8]) -> R) -> R {
+    if args == BytesSource::INVALID {
+        return logic(&[]);
+    }
```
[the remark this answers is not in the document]
cloutiertyler: If we're tweaking the proposal, please make sure changes make it back into the proposal.

coolreader18 (approved):

PR #1613, opened 2024-08-19
fix: Always remember to release a read only tx

joshua-spacetime:
# Description of Changes

Previously we were not releasing downgraded transactions for incremental subscription updates.

None

1


mamcx (approved):
LGTM

PR #1614, opened 2024-08-19
fix: Correctly label transactions as committed

joshua-spacetime:
# Description of Changes

Committed transactions were being labeled as rollbacks.

None

1


mamcx (approved):
LGTM

PR #1615, opened 2024-08-20
 [WASM ABI 1.0] impl `__call_reducer__` & `__describe_module__` using `bytes_sink_write`

Centril:
# Description of Changes

This adds `_bytes_sink_write` and implements `__call_reducer__`, `__describe_module__`, and `__setup__` with it.
Meanwhile, `_buffer_len`, `_buffer_consume`, and `_buffer_alloc` are fully removed this time around (some missing stuff from previous PR).

cc #1460

### DIFF
--- a/crates/bindings-csharp/Runtime/Internal/Module.cs
+++ b/crates/bindings-csharp/Runtime/Internal/Module.cs
@@ -252,15 +252,29 @@ public static partial class Module
                 // but a module should be prepared for it.
                 case 0:
                     break;
+                case (short)(ushort)FFI.Errno.NO_SUCH_BYTES:
+                    throw new NoSuchBytesException();
                 default:
-                    throw new UnreachableException();
+                    throw new UnknownException((FFI.Errno)(ushort)ret);
             }
+        }
+    }
+
+    private static void Write(this BytesSink sink, byte[] bytes)
+    {
+        var start = 0U;
+        while (start != bytes.Length)
+        {
+            var written = (uint)bytes.Length;
+            var buffer = bytes.AsSpan((int)start);
+            FFI._bytes_sink_write(sink, buffer, ref written);
+            start += written;
         }
     }
 
 #pragma warning disable IDE1006 // Naming Styles - methods below are meant for FFI.
 
-    public static Buffer __describe_module__()
+    public static void __describe_module__(BytesSink description)
     {
         // replace `module` with a temporary internal module that will register RawModuleDefV8, AlgebraicType and other internal types
         // during the RawModuleDefV8.GetSatsTypeInfo() instead of exposing them via user's module.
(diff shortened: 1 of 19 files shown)


crates/bindings-csharp/Runtime/Internal/Module.cs
```diff
@@ -252,15 +252,33 @@ private static byte[] Consume(this BytesSource source)
                 // but a module should be prepared for it.
                 case 0:
                     break;
+                case (short)(ushort)FFI.Errno.NO_SUCH_BYTES:
```
RReverser, line 255: Same questions as in the other PR about just changing Errno to `short` and whether `(short)FFI.Errno.` would suffice here.

crates/bindings-csharp/Runtime/Internal/Module.cs
```diff
@@ -252,15 +252,33 @@ private static byte[] Consume(this BytesSource source)
...
+    private static void Write(this BytesSink sink, byte[] bytes)
+    {
+        var start = 0U;
+        while (true)
+        {
+            var written = (uint)bytes.Length;
+            var buffer = bytes.AsSpan((int)start);
+            FFI._bytes_sink_write(sink, buffer, ref written);
+            start += written;
+            if (start == bytes.Length)
```
RReverser, line 272: Nit: just change the condition to `while (start != bytes.Length)` instead of short-circuiting?

crates/bindings-csharp/Runtime/Internal/Module.cs
```diff
@@ -269,17 +287,15 @@ public static Buffer __describe_module__()
...
         }
         catch (Exception e)
         {
             Runtime.Log($"Error while describing the module: {e}", Runtime.LogLevel.Error);
-            return Buffer.INVALID;
         }
     }
 
-    public static Buffer __call_reducer__(
+    public static short __call_reducer__(
```
RReverser, line 298: Maybe change the return value to the `Errno` so that we could return typechecked enum directly?

crates/bindings-csharp/Runtime/bindings.c
```diff
@@ -104,20 +105,26 @@ PREINIT(10, startup) {
     MonoObject* res;                                                          \
     mono_wasm_invoke_method_ref(ffi_method_##name, NULL, (void*[]){args},     \
                                 NULL, &res);                                  \
-    return *(ret*)mono_object_unbox(res);                                     \
+    res_code                                                                  \
   }
 
-EXPORT(Buffer, __describe_module__, ());
+#define EXPORT(ret, name, params, args...)                                             \
+  EXPORT_WITH_MONO_RES(ret, return *(ret*)mono_object_unbox(res);, name, params, args) \
```
RReverser, line 112: God I can't wait for when we'll be able to ditch all this Mono stuff. .NET 9, I'm counting on you...

coolreader18 (approved):

PR #1616, opened 2024-08-20
Redefine Address as `U128` + `Identity` and Hash as `U256`

Centril:
# Description of Changes

Redefines `Address` as `u128`.

Breaks the stored type of `Address` in system tables and in user tables.

### DIFF
--- a/crates/bindings-csharp/BSATN.Runtime/Builtins.cs
+++ b/crates/bindings-csharp/BSATN.Runtime/Builtins.cs
@@ -6,6 +6,15 @@ using System.Runtime.InteropServices;
 using SpacetimeDB.BSATN;
 using SpacetimeDB.Internal;
 
+internal static class Util
+{
+    // Same as `Convert.ToHexString`, but that method is not available in .NET Standard
+    // which we need to target for Unity support.
+    public static string ToHex<T>(T val)
+        where T : struct =>
+        BitConverter.ToString(MemoryMarshal.AsBytes([val]).ToArray()).Replace("-", "");
+}
+
 public readonly partial struct Unit
 {
     // Custom BSATN that returns an inline empty product type that can be recognised by SpacetimeDB.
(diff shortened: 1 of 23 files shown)


crates/bindings-csharp/BSATN.Runtime/Builtins.cs
```diff
@@ -61,67 +70,69 @@ string wrapperPropertyName
...
-
-    public static Address Random()
-    {
-        var random = new Random();
-        var addr = new Address();
-        random.NextBytes(addr.bytes);
-        return addr;
+    public static Address? From(byte[] bytes) {
+        Debug.Assert(bytes.Length == 16);
+        return new Address(MemoryMarshal.Read<U128>(bytes));
```
RReverser, line 91: This still needs to check that given bytes do not represent a zero address, like before.

You should be able to do this via simply

```cs
var addr = new Address(...);
return addr == default ? null : addr;
```

crates/bindings-csharp/BSATN.Runtime/Builtins.cs
```diff
@@ -61,67 +70,69 @@ string wrapperPropertyName
...
             return null;
         }
-        return new(bytes);
+        else
+        {
+            return new Address(new U128(upper, lower));
+        }
     }
-
-    public static Address Random()
```
RReverser, line 82: This is a breaking change, the method needs to be restored.

crates/bindings-csharp/BSATN.Runtime/Builtins.cs
```diff
@@ -61,67 +70,69 @@ string wrapperPropertyName
         );
 }
 
-public record Address : BytesWrapper
+public readonly record struct Address
 {
-    protected override int SIZE => 16;
+    private readonly U128 value;
 
     public Address() { }
```
RReverser, line 77: These default constructors here and in `Identity` can be removed now; they were necessary for `record` but any `struct` has them automatically.

crates/bindings-csharp/BSATN.Runtime/Builtins.cs
```diff
@@ -61,67 +70,66 @@ string wrapperPropertyName
...
-        : base(bytes) { }
-
-    public static Address? From(byte[] bytes)
-    {
-        if (bytes.All(b => b == 0))
-        {
-            return null;
-        }
-        return new(bytes);
+    internal Address(U128 v) => value = v;
```
RReverser, line 77: Hm, at this point are these constructors still used anywhere?

RReverser (approved):
C# LGTM modulo one final nit. Haven't looked at Rust much.

crates/lib/tests/serde.rs
```diff
@@ -58,25 +58,27 @@ fn test_json_mappings() {
...
+
     let data = r#"
 {
     "foo": 42,
     "bar": "404040FFFF0A48656C6C6F",
     "baz": ["heyyyyyy", "hooo"],
     "quux": { "Hash": "54a3e6d2b0959deaacf102292b1cbd6fcbb8cf237f73306e27ed82c3153878aa" },
     "and_peggy": { "some": 3.141592653589793238426 },
-    "identity": ["0000000000000000000000000000000000000000000000000000000000000000"]
+    "identity": ["0x0"]
```
RReverser, line 69: Ah the Rust implementation probably needs to revert to original hex stringification as well. We do use identity/address strings in a few places (e.g. configs).

crates/lib/src/identity.rs
```diff
@@ -63,36 +71,32 @@ impl Identity {
 
     #[doc(hidden)]
     pub fn __dummy() -> Self {
```
RReverser, line 73: I know this wasn't added in this PR but I looked at usages, and I'm unclear why we even need those functions instead of just using `::ZERO`.

kazimuth:
I got the Rust part of the PR mostly rebased, still need to do the C# part. If there are open design discussions remaining we should probably have them.

bfops (approved):
CLI changes LGTM

kazimuth:
This is now fully ready for merge together with https://github.com/clockworklabs/SpacetimeDBPrivate/pull/1060 @Centril

kazimuth:
> PR looks good, but did you give any thought to the formatting changes of identity? I personally think they are an improvement, but ymmv.

Ah, see Ingvar's comment above.

> Ah the Rust implementation probably needs to revert to original hex stringification as well. We do use identity/address strings in a few places (e.g. configs), so this would be a breaking change.

This is a breaking change anyway but it's a wipe-the-database breaking change rather than a rewrite-your-configs breaking change at the moment.

PR #1617, opened 2024-08-20
Fix `cargo test --all` on Windows

kazimuth:
# Description of Changes

It was previously broken due to linker errors, now it works.

N/A

0

# Testing

I ran `cargo test --all` on windows.

### DIFF
--- a/crates/bindings-sys/src/lib.rs
+++ b/crates/bindings-sys/src/lib.rs
@@ -18,6 +18,8 @@ pub mod raw {
     // on. Any non-breaking additions to the abi surface should be put in a new `extern {}` block
     // with a module identifier with a minor version 1 above the previous highest minor version.
     // For breaking changes, all functions should be moved into one new `spacetime_X.0` block.
+    //
+    // If you update this, make sure to update the `windows_test_shims` module below.
     #[link(wasm_import_module = "spacetime_10.0")]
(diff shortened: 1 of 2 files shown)


crates/bindings-sys/src/lib.rs
```diff
@@ -745,3 +747,155 @@ impl Drop for RowIter {
         unsafe { raw::_iter_drop(self.raw) }
     }
 }
+
+/// Rust tries to build tests for our various modules on the host system, not for wasm.
+/// On Linux, this succeeds, despite there being unresolved symbols, because the linker is okay with
+/// unresolved symbols.
+/// On Windows, this fails, because the linker errors on unresolved symbols.
+/// By adding these shims, we can build and run tests on windows, as long as they don't use module ABI calls.
+#[cfg(target_family = "windows")]
```
kazimuth, line 756: This DOES mean that if you, for some reason, wanted to run modules in native code, it would be forbidden on windows.
I thought about alternatively doing `#[cfg(not(target_family = "wasm"))]` so that it would be forbidden everywhere non-wasm instead. Not sure what the right choice is.

cloutiertyler:
I don't think we've run into this in CI, @jdetter do we run all the test in the windows tests? I've closed this, but if this is a known issue, please reopen.

PR #1619, opened 2024-08-20
Add snapshot tests for normalized ModuleDef

### LINKED ISSUE #1589: Test coverage - assert that equivalent Rust and C# modules produce the same DescribeModule output
We want to test that C# and Rust modules, when written to expose the same set of tables, types and reducers, `spacetime generate` the same bindings. Actual codegen goes through the same path, and is checked by the SDK tests, so what's missing is a test that they return the same results from `describe_module` (or whatever it's called).

Conveniently, we already have interchangeable Rust and C# modules - the SDK test modules are known to be equivalent because we run the same client against both of them without rebuilding its bindings in between.

Write a test which, for each pair of Rust/C# SDK-test modules we happen to have sitting around, calls `describe_module` on each, deserializes/processes the result as necessary, and then asserts that they're the same.

RReverser:
# Description of Changes

This provides snapshot testing for ModuleDef returned from Wasm modules. It will automatically compare ModuleDefs of test modules against both the previous snapshots, and across different languages.

Note that comparison doesn't (can't) use raw ModuleDef. There are some minor non-functional differences between Rust and C# that need to be cleaned up before comparing - most notably, the order in which types are registered during table traverse, plus some smaller things like Rust registering column constraints even where they're no-op. I believe those changes don't matter and are fine to remove before comparing.

The history is split into two commits - one that adds working testing with all the cleanups, and in a separate one I added more compact debug representations for AlgebraicType and child types. That commit could be reverted, but it makes the snapshots much easier to read (see the diff) and I believe will be useful for regular debugging outside this test as well.

Fixes #1589.

corresponding GitHub label.

# Testing

- [x] Ran SDK tests (`cargo test -p spacetimedb-sdk`).

### DIFF
--- a/crates/testing/src/sdk.rs
+++ b/crates/testing/src/sdk.rs
@@ -8,6 +8,11 @@ use std::thread::JoinHandle;
 
 use crate::invoke_cli;
 use crate::modules::{CompilationMode, CompiledModule};
+use spacetimedb_lib::db::raw_def::RawTableDefV8;
+use spacetimedb_lib::sats::primitives::Constraints;
+use spacetimedb_lib::{AlgebraicType, MiscModuleExport, ProductType, RawModuleDefV8, ReducerDef};
+use std::collections::BTreeMap;
+use std::path::Path;
 use tempfile::TempDir;
 
 pub fn ensure_standalone_process() {
@@ -122,7 +127,6 @@ macro_rules! memoized {
             .get_or_insert_with(HashMap::new)
             .entry($key)
             .or_insert_with_key(|$key| -> $value_ty { $body })
-            .clone()
     }};
 }
 
@@ -139,6 +143,7 @@ fn compile_module(module: &str) -> String {
         let module = CompiledModule::compile(module, CompilationMode::Debug);
         module.path().to_str().unwrap().to_owned()
     })
+    .clone()
 }
 
 // Note: this function does not memoize because we want each test to publish the same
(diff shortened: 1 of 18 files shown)


kazimuth:
This is great, but possibly should be done against the new `ModuleDef` type from #1572 and #1606? Those are going to merge this week, they have been a little delayed because I kept getting blocked on other PRs merging. I would really love to implement pretty-printing for that `ModuleDef`.

crates/testing/src/sdk.rs
```diff
@@ -203,7 +208,93 @@ fn generate_bindings(language: &str, wasm_file: &str, client_project: &str, gene
...
+        } = descriptions;
+
+        for reducer in &mut reducers {
+            for arg in &mut reducer.args {
+                typespace.inline_typerefs_in_type(&mut arg.algebraic_type).unwrap();
+            }
+        }
+
+        for table in &mut tables {
+            // Rust generates same number of constraints as number of columns,
```
RReverser, line 241: @kazimuth Is this something that will be gone in the new schema by any chance?

kazimuth: It is indeed! We still generate indexes for unique constraints, but all the other generated stuff is gone.

RReverser:
@kazimuth I rebased this, but on the 2nd thought we shouldn't compare V9 moduledef just yet.

The purpose of the original issue / of adding those tests was to verify that the module structure used for generation is the same across different implementations (Rust and C#) of the same module.

Since `spacetime generate` and `extract_descriptions` which we verify here still uses the V8 module definition, I think we should merge this as-is first and upgrade the test only after generation for V9 is implemented.

Meanwhile, we'll at least already have testing in place.

kazimuth (approved):
I like the new debug formatting for types, that's very convenient.

I'm happy merging this for now and updating it once the stuff it depends on has been rewritten. I agree that it kind of doesn't make sense to test all the new stuff while it hasn't been fully plumbed through everything.

PR #1622, opened 2024-08-21
WASM ABI: implement `row_iter_bsatn_advance` & `row_iter_bsatn_close`

Centril:
# Description of Changes

Implements `row_iter_bsatn_advance` & `row_iter_bsatn_close`.

### DIFF
--- a/crates/bindings-csharp/Runtime/Internal/FFI.cs
+++ b/crates/bindings-csharp/Runtime/Internal/FFI.cs
@@ -18,6 +18,20 @@ public readonly record struct BytesSource(uint Handle)
 [StructLayout(LayoutKind.Sequential)]
 public readonly record struct BytesSink(uint Handle) { }
 
+public enum Errno : short
+{
+    EXHAUSTED = -1,
+    OK = 0,
+    HOST_CALL_FAILURE = 1,
+    NO_SUCH_TABLE = 4,
+    LOOKUP_NOT_FOUND = 2,
+    NO_SUCH_ITER = 6,
+    NO_SUCH_BYTES = 8,
+    NO_SPACE = 9,
+    BUFFER_TOO_SMALL = 11,
+    UNIQUE_ALREADY_EXISTS = 12,
+}
+
 #pragma warning disable IDE1006 // Naming Styles - Not applicable to FFI stuff.
 internal static partial class FFI
 {
@@ -31,18 +45,6 @@ internal static partial class FFI
         "bindings"
 #endif
     ;
-
-    public enum Errno : ushort
-    {
-        OK = 0,
-        HOST_CALL_FAILURE = 1,
-        NO_SUCH_TABLE = 4,
-        LOOKUP_NOT_FOUND = 2,
-        NO_SUCH_BYTES = 8,
-        NO_SPACE = 9,
-        BUFFER_TOO_SMALL = 11,
-        UNIQUE_ALREADY_EXISTS = 12,
-    }
 
     [NativeMarshalling(typeof(Marshaller))]
     public struct CheckedStatus
@@ -108,7 +110,7 @@ internal static partial class FFI
     [StructLayout(LayoutKind.Sequential)]
     public readonly record struct RowIter(uint Handle)
     {
-        public static readonly RowIter INVALID = new(uint.MaxValue);
+        public static readonly RowIter INVALID = new(0);
     }
 
     [LibraryImport(StdbNamespace)]
@@ -159,14 +161,14 @@ internal static partial class FFI
     );
 
     [LibraryImport(StdbNamespace)]
-    public static partial CheckedStatus _iter_advance(
+    public static partial Errno _row_iter_bsatn_advance(
         RowIter iter_handle,
         [MarshalUsing(CountElementName = nameof(buffer_len))] [Out] byte[] buffer,
         ref uint buffer_len
     );
 
     [LibraryImport(StdbNamespace)]
-    public static partial void _iter_drop(RowIter iter_handle);
+    public static partial CheckedStatus _row_iter_bsatn_close(RowIter iter_handle);
 
     [LibraryImport(StdbNamespace)]
     public static partial void _volatile_nonatomic_schedule_immediate(
@@ -189,7 +191,7 @@ internal static partial class FFI
     );
 
     [LibraryImport(StdbNamespace)]
-    public static partial short _bytes_source_read(
+    public static partial Errno _bytes_source_read(
         BytesSource source,
         Span<byte> buffer,
         ref uint buffer_len
--- a/crates/bindings-csharp/Runtime/Internal/ITable.cs
+++ b/crates/bindings-csharp/Runtime/Internal/ITable.cs
@@ -22,31 +22,53 @@ public interface ITable<T> : IStructuralReadWrite
 
             public bool MoveNext()
             {
+                if (handle == FFI.RowIter.INVALID)
+                {
+                    return false;
+                }
+
                 uint buffer_len;
                 while (true)
                 {
                     buffer_len = (uint)buffer.Length;
-                    try
+                    var ret = FFI._row_iter_bsatn_advance(handle, buffer, ref buffer_len);
+                    if (ret == Errno.EXHAUSTED)
                     {
-                        FFI._iter_advance(handle, buffer, ref buffer_len);
+                        handle = FFI.RowIter.INVALID;
                     }
-                    catch (BufferTooSmallException)
+                    // On success, the only way `buffer_len == 0` is for the iterator to be exhausted.
+                    // This happens when the host iterator was empty from the start.
+                    System.Diagnostics.Debug.Assert(!(ret == Errno.OK && buffer_len == 0));
+                    switch (ret)
                     {
-                        buffer = new byte[buffer_len];
-                        continue;
+                        // Iterator advanced and may also be `EXHAUSTED`.
+                        // When `OK`, we'll need to advance the iterator in the next call to `MoveNext`.
+                        // In both cases, copy over the row data to `Current` from the scratch `buffer`.
+                        case Errno.EXHAUSTED
+                        or Errno.OK:
+                            Current = new byte[buffer_len];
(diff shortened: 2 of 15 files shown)


crates/core/src/host/wasmtime/wasm_instance_env.rs
```diff
@@ -582,69 +582,121 @@ impl WasmInstanceEnv {
...
+            let buffer_len = u32::read_from(mem, buffer_len_ptr)?;
+            // Get a mutable view to the `buffer`.
+            let mut buffer = mem.deref_slice_mut(buffer_ptr, buffer_len)?;
 
             let mut written = 0;
-            // Advance the iterator.
+            // Fill the buffer as much as possible.
             while let Some(chunk) = iter.as_slice().first() {
-                let Some(buf_chunk) = wasm_buffer.get_mut(..chunk.len()) else {
+                // TODO(Centril): refactor using `split_at_mut_checked`.
```
RReverser, line 639: Alternatively, could leverage the `std::io::Write` implementation on slices. `.write()` on a slice results in the same zero-cost memcpy code and automatically takes care of "advancing" the slice to the remainder part: https://rust.godbolt.org/z/1x4W6KeYx

crates/bindings-csharp/Runtime/Internal/ITable.cs
```diff
@@ -22,31 +22,45 @@ public class Enumerator(FFI.RowIter handle) : IDisposable
...
+                        // Iterator exhausted, we're done.
+                        case -1:
+                            handle = FFI.RowIter.INVALID;
+                            return buffer_len != 0;
+                        case 0:
+                            return buffer_len != 0;
+                        case (short)(ushort)FFI.Errno.NO_SUCH_ITER:
+                            throw new NoSuchIterException();
+                        case (short)(ushort)FFI.Errno.BUFFER_TOO_SMALL:
+                            buffer = new byte[buffer_len];
```
RReverser, line 51: Shouldn't this copy the original buffer? Not sure if any data is written to the buffer at that point or not.

crates/bindings-csharp/Runtime/Internal/FFI.cs
```diff
@@ -19,7 +19,7 @@ public readonly record struct BytesSource(uint Handle)
 public readonly record struct BytesSink(uint Handle) { }
 
 #pragma warning disable IDE1006 // Naming Styles - Not applicable to FFI stuff.
-internal static partial class FFI
+public static partial class FFI
```
RReverser, line 22: Is this just for Errno? Can we extract it like BytesSource above if so? I'd prefer not to expose APIs we can easily keep hidden.

crates/bindings-csharp/Runtime/Internal/FFI.cs
```diff
@@ -159,14 +161,14 @@ out RowIter out_
     );
 
     [LibraryImport(StdbNamespace)]
-    public static partial CheckedStatus _iter_advance(
+    public static partial short _row_iter_bsatn_advance(
```
RReverser, line 164: This can be just `Errno` now, right?

crates/bindings-csharp/Runtime/Internal/ITable.cs
```diff
@@ -22,31 +22,51 @@ public class Enumerator(FFI.RowIter handle) : IDisposable
...
+                {
+                    return false;
+                }
+
                 uint buffer_len;
                 while (true)
                 {
                     buffer_len = (uint)buffer.Length;
-                    try
+                    var ret = (FFI.Errno)
```
RReverser, line 34: Then you don't need the cast here.

crates/bindings-csharp/Runtime/Internal/ITable.cs
```diff
@@ -22,31 +22,51 @@ public class Enumerator(FFI.RowIter handle) : IDisposable
...
+                        handle = FFI.RowIter.INVALID;
                     }
-                    catch (BufferTooSmallException)
+                    switch (ret)
                     {
-                        buffer = new byte[buffer_len];
-                        continue;
+                        // Iterator exhausted, we're done.
+                        case FFI.Errno.EXHAUSTED:
+                        // Iterator advanced but not exhausted,
```
RReverser, line 44: I'm confused by this comment, is something missing? It's under `EXHAUSTED:` but says "iterator advanced but not exhausted". Also there is no `break;` for this `case` 🤔

crates/bindings-csharp/Runtime/Internal/ITable.cs
```diff
@@ -22,31 +22,51 @@ public class Enumerator(FFI.RowIter handle) : IDisposable
...
+                            Current = new byte[buffer_len];
+                            Array.Copy(buffer, 0, Current, 0, buffer_len);
+                            return buffer_len != 0;
+                        // Couldn't find the iterator, error!
+                        case FFI.Errno.NO_SUCH_ITER:
+                            throw new NoSuchIterException();
+                        // The scratch `buffer` is too small to fit a row / chunk.
+                        // Grow `buffer` and try again.
+                        case FFI.Errno.BUFFER_TOO_SMALL:
+                            buffer = new byte[buffer_len];
```
RReverser, line 57: Unlike the comment says, it doesn't grow the `buffer` here?

RReverser: Ahh I see, `buffer_len` has a new length at this point. Might be worth adding to the comment, it's easy to miss.

crates/bindings-csharp/Runtime/Internal/ITable.cs
```diff
@@ -22,31 +22,51 @@ public class Enumerator(FFI.RowIter handle) : IDisposable
...
-                        continue;
+                        // Iterator exhausted, we're done.
+                        case FFI.Errno.EXHAUSTED:
+                        // Iterator advanced but not exhausted,
+                        // we'll need to advance the iterator in the next call to `MoveNext`.
+                        case FFI.Errno.OK:
+                            // Success! Copy over the row data to `Current` from the scratch `buffer`.
+                            Current = new byte[buffer_len];
+                            Array.Copy(buffer, 0, Current, 0, buffer_len);
+                            return buffer_len != 0;
```
RReverser, line 50: FWIW `MoveNext()` should return `false` when the iterator is exhausted. Empty returned buffer used to indicate that before, but IIUC not anymore - now the special status should indicate that instead?

crates/bindings-csharp/Runtime/Internal/Module.cs
```diff
@@ -233,29 +233,29 @@ private static byte[] Consume(this BytesSource source)
             // Write into the spare capacity of the buffer.
             var spare = buffer.AsSpan((int)written);
             var buf_len = (uint)spare.Length;
-            var ret = FFI._bytes_source_read(source, spare, ref buf_len);
+            var ret = (FFI.Errno)FFI._bytes_source_read(source, spare, ref buf_len);
```
RReverser, line 236: I'd change `bytes_source_read` to return `Errno` directly instead, to avoid casting.

crates/bindings-csharp/Runtime/Internal/ITable.cs
RReverser: Right, but what I meant is what if `ret == 0 && buffer_len == 0`? Right now the iterator will stop on that too - is that something we want intentionally or do we want to stop _only_ on `ret == -1`?

RReverser: I was thinking of an edge case where host - for whatever reason - returned one of the interim buffers as genuinely empty, so `buffer_len == 0` but the iterator is not over yet.

If you think it should be unreachable, let's add `Debug.Assert` to indicate that and to check for that at least in debug builds.

But even with the assertion it still sounds like this particular line can be simplified to `return true;` as the "iterator finished" condition only depends on `handle == FFI.RowIter.INVALID` (after having received "exhausted" return code) and not depend on `buffer_len` at all.

crates/core/src/host/wasmtime/wasm_instance_env.rs
RReverser: Yeah either way its a future TODO, just wanted to comment with an alternative :)

crates/bindings-csharp/Runtime/Internal/ITable.cs
RReverser: Ah yeah you can, in this case it was just unclear whether the fallthrough is intentional or not.

Nitpicking as the comment is enough but I'd probably just rewrite it to `case Errno.EXHAUSTED or Errno.OK:` like you said, which would make the intent clear instantly.

crates/bindings-csharp/Runtime/Internal/ITable.cs
RReverser: > Counter-example: Consider an empty table. You do the equivalent of `MyTable::iter()`. This makes the host side iterator contain zero chunks. A call to `_row_iter_bsatn_advance` will return `-1` the first time and you don't know that `handle == FFI.RowIter.INVALID` yet. Yet, the call to `.MoveNext()` should return `false`, as otherwise, the `foreach` will think there was a chunk/rows, but there wasn't (technically, `while (stream.Position < stream.Length)` is tolerant to this).

You're right, that's actually a great counter-example.

Will `Debug.Assert` suggested above fail on this? Either way, could be nice to have some test for this edge case. I'm guessing you'd like to do that outside this PR, but if so, can you at least create an issue so that we don't forget to add such SDK test later?

RReverser (approved):

crates/bindings-sys/src/lib.rs
```diff
@@ -355,6 +380,11 @@ pub mod raw {
     #[repr(transparent)]
     pub struct RowIter(u32);
 
+    impl RowIter {
+        /// An invalid handle, used e.g., when the iterator has been exhausted.
+        pub const INVALID: Self = Self(0);
+    }
```
coolreader18, line 386: Would it make more sense to have RowIter be a `NonZeroU32`, and use `Option<RowIter>` to represent `INVALID`?

Issue #1563, opened 2024-08-01
Update SQL AST in accordance with the SQL spec


joshua-spacetime:
The `SqlAst` should represent the SQL grammar as defined in the spec. As such there are a few changes we should make.

1. No name resolution or type checking

    We should have a purely syntactic representation. Type checking and name resolution should be done after we've derived this representation.

2. No column expressions

    Right now we have `Column::UnnamedExpr(Expr)`, but we decided in the spec to only support column names, qualified and unqualified.

3. Errors

    We want good internal as well as user facing error messages. [error-stack](https://docs.rs/error-stack/latest/error_stack) looks like a way to achieve the former, [ariadne](https://docs.rs/ariadne/latest/ariadne) the latter. Now is a good time to experiment with both.

We should probably define a new AST type first. We can replace `SqlAst` with the new type after we've implemented type checking and name resolution for the new type.

PR #1623, opened 2024-08-21
Update SQL AST in accordance with the SQL spec

mamcx:
# Description of Changes

The first step towards updating the parsing of `SQL` to match the spec for `SpacetimeDB`, as asked by ticket #1563.

This is the syntactic parser, without being integrated yet into the codebase, to be done later.

1

# Testing

- [x] *Generate test for both valid and invalid cases, and pretty printer of the `ast`*

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4574,6 +4574,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "spacetimedb-sql-parser"
+version = "0.12.0"
+dependencies = [
+ "derive_more",
+ "sqlparser",
+ "thiserror",
+]
+
+[[package]]
 name = "spacetimedb-standalone"
 version = "0.12.0"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ members = [
   "crates/sdk",
   "crates/snapshot",
   "crates/sqltest",
+  "crates/sql-parser",
   "crates/standalone",
   "crates/table",
   "crates/testing",
@@ -100,6 +101,7 @@ spacetimedb-primitives = { path = "crates/primitives", version = "0.12.0" }
 spacetimedb-sats = { path = "crates/sats", version = "0.12.0" }
 spacetimedb-schema = { path = "crates/schema", version = "0.12.0" }
 spacetimedb-standalone = { path = "crates/standalone", version = "0.12.0" }
+spacetimedb-sql-parser = { path = "crates/sql-parser", version = "0.12.0" }
 spacetimedb-table = { path = "crates/table", version = "0.12.0" }
 spacetimedb-vm = { path = "crates/vm", version = "0.12.0" }
 spacetimedb-fs-utils = { path = "crates/fs-utils", version = "0.12.0" }
@@ -247,14 +249,14 @@ wasmtime = { version = "15", default-features = false, features = ["cranelift", 
 # and reconnecting, from the tracy client to the database.
 # TODO(George): Need to be able to remove "broadcast" in some build configurations.
 tracing-tracy = { version = "0.10.4", features = [
-  "enable",
-  "system-tracing",
-  "context-switch-tracing",
-  "sampling",
-  "code-transfer",
-  "broadcast",
-  "ondemand",
-] }
+   "enable",
+   "system-tracing",
+   "context-switch-tracing",
+   "sampling",
+   "code-transfer",
+   "broadcast",
+   "ondemand",
+ ] }
 
 # Vendor the openssl we rely on, rather than depend on a
 # potentially very old system version.
--- a/crates/sql-parser/Cargo.toml
+++ b/crates/sql-parser/Cargo.toml
@@ -1,0 +1,12 @@
+[package]
+name = "spacetimedb-sql-parser"
+version.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+license-file = "LICENSE"
+description = "The SpacetimeDB SQL AST and Parser"
+
+[dependencies]
+derive_more.workspace = true
+sqlparser.workspace = true
+thiserror.workspace = true
--- a/crates/sql-parser/src/ast/mod.rs
+++ b/crates/sql-parser/src/ast/mod.rs
@@ -1,0 +1,99 @@
+use sqlparser::ast::Ident;
+
+pub mod sql;
+pub mod sub;
+
+/// The FROM clause is either a [RelExpr] or a JOIN
+pub enum SqlFrom<Ast> {
+    Expr(RelExpr<Ast>, Option<SqlIdent>),
+    Join(RelExpr<Ast>, SqlIdent, Vec<SqlJoin<Ast>>),
+}
+
+/// A RelExpr is an expression that produces a relation
+pub enum RelExpr<Ast> {
+    Var(SqlIdent),
+    Ast(Box<Ast>),
+}
+
+/// An inner join in a FROM clause
+pub struct SqlJoin<Ast> {
+    pub expr: RelExpr<Ast>,
+    pub alias: SqlIdent,
+    pub on: Option<SqlExpr>,
+}
+
+/// A projection expression in a SELECT clause
+pub struct ProjectElem(pub SqlExpr, pub Option<SqlIdent>);
+
+/// A SQL SELECT clause
+pub enum Project {
+    /// SELECT *
+    /// SELECT a.*
+    Star(Option<SqlIdent>),
+    /// SELECT a, b
+    Exprs(Vec<ProjectElem>),
+}
+
+/// A scalar SQL expression
+pub enum SqlExpr {
+    /// A constant expression
+    Lit(SqlLiteral),
+    /// Unqualified column ref
+    Var(SqlIdent),
+    /// Qualified column ref
+    Field(SqlIdent, SqlIdent),
+    /// A binary infix expression
+    Bin(Box<SqlExpr>, Box<SqlExpr>, BinOp),
+}
+
+/// A SQL identifier or named reference
+#[derive(Clone)]
+pub struct SqlIdent {
+    pub name: String,
+    pub case_sensitive: bool,
+}
+
+impl From<Ident> for SqlIdent {
+    fn from(value: Ident) -> Self {
+        match value {
+            Ident {
+                value: name,
+                quote_style: None,
+            } => SqlIdent {
+                name,
+                case_sensitive: false,
+            },
+            Ident {
+                value: name,
+                quote_style: Some(_),
+            } => SqlIdent {
+                name,
+                case_sensitive: true,
+            },
+        }
+    }
+}
+
+/// A SQL constant expression
+pub enum SqlLiteral {
+    /// A boolean constant
+    Bool(bool),
+    /// A hex value like 0xFF or x'FF'
+    Hex(String),
+    /// An integer or float value
+    Num(String),
+    /// A string value
+    Str(String),
+}
+
+/// Binary infix operators
+pub enum BinOp {
+    Eq,
+    Ne,
+    Lt,
+    Gt,
+    Lte,
+    Gte,
+    And,
+    Or,
+}
--- a/crates/sql-parser/src/ast/sql.rs
+++ b/crates/sql-parser/src/ast/sql.rs
@@ -1,0 +1,73 @@
+use super::{Project, SqlExpr, SqlFrom, SqlIdent, SqlLiteral};
+
+/// The AST for the SQL DML and query language
+pub enum SqlAst {
+    /// SELECT ...
+    Query(QueryAst),
+    /// INSERT INTO ...
+    Insert(SqlInsert),
+    /// UPDATE ...
+    Update(SqlUpdate),
+    /// DELETE FROM ...
+    Delete(SqlDelete),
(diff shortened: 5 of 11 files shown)


crates/sql/src/ast/subscription.rs
```diff
@@ -0,0 +1,323 @@
...
+//! predicate
+//!     =  expr
+//!     | predicate AND predicate
+//!     ;
+//!
+//! predicate_on
+//!     =  columnExpr '=' columnExpr
+//!     | predicate_on AND predicate_on
+//!     ;
+//!
```
mamcx, line 62: This is an addition to the spec, because the `predicate` allows expressions that will not optimize.

crates/sql/src/errors.rs
```diff
@@ -0,0 +1,106 @@
...
+#[derive(Debug, Display)]
+pub enum Suggestion {
+    #[display(fmt = "Change it to an INNER JOIN")]
+    ChangeJoinToInner,
+    #[display(fmt = "Change to a valid comparison operator: `=, <, >, <=, >=`")]
+    ChangeOpToCmp,
+    #[display(fmt = "Change to several queries")]
+    ChangeOrToManyQueries,
+}
+
```
mamcx, line 17: It could be good idea to brainstorm which suggestion could enhance the experience of errors.

crates/sql/src/errors.rs
```diff
@@ -0,0 +1,106 @@
...
+#[derive(Error, Debug)]
+pub enum SqlUnsupported {
```
mamcx, line 19: A single place to see what we need to add later, maybe.

crates/sql/src/ast/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+///    = ident
+///    ;
+///
+/// columnExpr
+///     = column
+///     | qualfied
+///     ;
+/// ```
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct SqlColumn {
```
joshua-spacetime, line 65: Is this a project element? We should probably move it into the `ad_hoc` module.

crates/sql/src/ast/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+                    write!(f, "OR")
+                }
+            },
+        }
+    }
+}
+
+/// Raw `SQL` value.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum SqlValue {
```
joshua-spacetime, line 186: I think `Literal` would be a more descriptive name than `Value`.

crates/sql/src/ast/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+    /// Hexadecimal value
+    /// ```sql
+    /// 0x0000FCAB
+    /// x'93dda0'
+    /// ```
+    Hex(String),
+    /// Unparsed number (int, float, etc.)
+    Num(String),
+    /// Unparsed long number (bigint, etc.)
+    NumLong(String),
```
joshua-spacetime, line 198: We should probably just have a single variant. The exact int/float type will be determined during type checking.

crates/sql/src/ast/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+    /// String value
+    String(String),
+    /// Identifier (table, column, etc.)
+    ///
+    /// ```sql
+    ///  "column_name"
+    /// ```
+    Identifier(Identifier),
```
joshua-spacetime, line 206: Identifiers should be separate from literals.

crates/sql/src/ast/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+    Null,
```
joshua-spacetime, line 207: We don't support nulls yet.

crates/sql/src/ast/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+        }
+    }
+}
+
+/// Field expression, with optional `alias`.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, From)]
+pub enum SqlFieldExpr {
+    Name(SqlColumn),
+    Value { value: SqlValue, alias: Option<Identifier> },
+}
```
joshua-spacetime, line 229: I think we want a field expression to just represent field access like `r.a`. In other words, `a` on its own should not be considered a field expression. And it shouldn't have an alias associated with it.

crates/sql/src/ast/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+            SqlFieldExpr::Name(column) => write!(f, "{}", column),
+            SqlFieldExpr::Value { value, alias } => match alias {
+                Some(alias) => write!(f, "{} AS {}", value, alias),
+                None => write!(f, "{}", value),
+            },
+        }
+    }
+}
+
+impl From<Ident> for SqlFieldExpr {
```
joshua-spacetime, line 253: Just to reiterate, we should remove this impl since a singular Ident shouldn't be considered a field expression.

crates/sql/src/ast/subscription.rs
```diff
@@ -0,0 +1,258 @@
...
+/// predicate_on
+///    = columnExpr '=' columnExpr
+///   | predicate_on AND predicate_on
+///  ;
+/// ```
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum SqlOn {
+    And { lhs: Box<SqlOn>, rhs: Box<SqlOn> },
+    Eq { lhs: SqlColumn, rhs: SqlColumn },
+}
```
joshua-spacetime, line 159: We could be permissive, allow any expression here, and then catch invalid ones during semantic analysis. I think that's probably what we should do.

crates/sql/src/ast/subscription.rs
```diff
@@ -0,0 +1,258 @@
...
+/// ```ebnf
+/// expr
+///     = literal
+///     | ident
+///     | column
+///     | expr op expr
+///     ;
+/// ```
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum SqlExpr {
```
joshua-spacetime, line 205: This should have at least 4 variants: for literals, idents, field expressions, and binary expressions.

crates/sql/src/ast/ad_hoc.rs
```diff
@@ -0,0 +1,449 @@
...
+//!     | FALSE
+//!     ;
+//! ```
+use crate::ast::*;
+
+/// Arguments for `SQL` projection.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum SqlArg {
+    Column(SqlColumn),
+    QualifiedWildcard { table: Identifier },
```
joshua-spacetime, line 181: Note, this isn't a valid argument to an aggregate function.

crates/sql/src/ast/ad_hoc.rs
```diff
@@ -0,0 +1,449 @@
...
+///
+/// ```ebnf
+/// aggrExpr
+///    = COUNT '(' STAR ')' [AS] ident
+///    | COUNT '(' DISTINCT columnExpr ')' [AS] ident
+///    | SUM   '(' columnExpr ')' [AS] ident
+/// ;
+/// ```
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum SqlSummarize {
```
joshua-spacetime, line 195:
```suggestion
pub enum SqlAgg {
```

crates/sql/src/ast/ad_hoc.rs
```diff
@@ -0,0 +1,449 @@
...
+    Count {
+        arg: SqlArg,
+        distinct: bool,
+        alias: Option<Identifier>,
```
joshua-spacetime, line 199: An Alias is required for count.

crates/sql/src/ast/ad_hoc.rs
```diff
@@ -0,0 +1,449 @@
...
+    },
+    Sum {
+        column: SqlColumn,
+        alias: Option<Identifier>,
```
joshua-spacetime, line 203: And for sum.

crates/sql/src/ast/ad_hoc.rs
```diff
@@ -0,0 +1,449 @@
...
+///
+/// ```ebnf
+/// predicate
+///     = expr
+///    | predicate AND predicate
+///    | predicate OR predicate
+///    ;
+/// ```
+#[derive(Debug, PartialEq, Eq)]
+pub enum SqlOn {
```
joshua-spacetime, line 263: As mentioned previously, let's not have a separate expression type for `ON`.

crates/sql/src/ast/ad_hoc.rs
```diff
@@ -0,0 +1,449 @@
...
+/// A valid expression for `SQL` queries.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum SqlExpr {
+    Field(SqlFieldExpr),
+    BinOp {
+        op: OpQuery,
+        lhs: Box<SqlExpr>,
+        rhs: Box<SqlExpr>,
+    },
+    Summarize(SqlSummarize),
```
joshua-spacetime, line 318: Agg should not be included in the expression type.

crates/sql/src/ast/ad_hoc.rs
```diff
@@ -0,0 +1,449 @@
...
+/// The `INSERT INTO` clause for `SQL` queries.
+///
+/// ```ebnf
+/// INSERT INTO table [ '(' column { ',' column } ')' ] VALUES '(' literal { ',' literal } ')'
+/// ```
+#[derive(Debug, PartialEq, Eq)]
+pub struct SqlInsert {
+    pub table: SqlTable,
+    pub columns: Box<[SqlColumn]>,
+    pub values: Box<[Box<[SqlValue]>]>,
```
joshua-spacetime, line 378: Why a nested list?

crates/sql/src/parser/mod.rs
```diff
@@ -0,0 +1,270 @@
...
+    }
+}
+
+#[macro_export]
+macro_rules! unsupported {
+    ($name:literal,$a:expr)=>{{
+        if $a.unsupported() {
+            return Err(SqlError::Unsupported(SqlUnsupported::FeatureNamed {
+                name: stringify!($name),
+                it: stringify!($a),
```
joshua-spacetime, line 77: Does this return a readable error message?

crates/sql/src/parser/mod.rs
```diff
@@ -0,0 +1,270 @@
...
+
+/// Unified AST evaluation for the SQL parser
+pub trait SqlParser {
+    type Ast;
+    type Expr;
+    type Selection;
+    type Projection;
+    type InnerJoin;
+    type JoinOn;
+    type From;
```
joshua-spacetime, line 171: I think a trait is probably overkill. Modules should be sufficient.

crates/sql/src/ast/mod.rs
mamcx: Done

crates/sql/src/ast/subscription.rs
mamcx: The `sql parser` doesn't separate `ident` from `column` because is not possible to know without type checking, so neither I can add an `ident` here. I assume this is always a `SqlColumn`. If we say we can neither, then we need to change how `Identifier`  works, and remove `SqlColumn`

crates/sql/src/ast/ad_hoc.rs
joshua-spacetime: Sorry, I meant it shouldn't be optional.

crates/sql/src/ast/mod.rs
```diff
@@ -0,0 +1,309 @@
+pub mod ad_hoc;
+pub mod subscription;
+
+use std::fmt;
+
+use derive_more::{Display, From};
+use spacetimedb_lib::operator::{OpCmp, OpLogic, OpQuery};
+use sqlparser::ast::{Ident, ObjectName};
+
+// TODO: Validate if is a valid identifier...
```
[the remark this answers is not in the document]
joshua-spacetime: I think you can remove this TODO. At this point we know it's valid syntactically. Whether it points to a valid object will be determined during name resolution.

crates/sql/src/ast/ad_hoc.rs
mamcx: Done

crates/sql/src/ast/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+    fn from(value: ObjectName) -> Self {
+        Identifier::new(value.to_string())
+    }
+}
+
+impl From<&str> for Identifier {
+    fn from(value: &str) -> Self {
+        Identifier::new(value.to_string())
+    }
+}
```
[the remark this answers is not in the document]
mamcx: I gate this `impl` with `#[cfg(test)]` now.

crates/sql/src/ast/subscription.rs
mamcx: Changed `SqlColumn` to `SqlIdentifier` so is more explicit.

joshua-spacetime (approved):

PR #1624, opened 2024-08-21
Remove hardcoded types from C# codegen

RReverser:
# Description of Changes

We leverage the fact that all primitives in C# have a corresponding fully-qualified name, and declare a single namespace with all their BSATN implementations.

This makes type resolution in the codegen simpler - it doesn't need to hold an ever-growing switch of special types - and allows contributors to add new System implementations (e.g. for collections) without having to deal with Roslyn at all. 

corresponding GitHub label.

# Testing


crates/bindings-csharp/BSATN.Codegen/Utils.cs
```diff
@@ -126,25 +113,19 @@ static string GetTypeInfoForNamedType(INamedTypeSymbol type)
                 }
                 return $"SpacetimeDB.BSATN.Enum<{SymbolToName(type)}>";
             }
-            var result = type.OriginalDefinition.ToString() switch
+            var fullName = type.ToDisplayString(SymbolFormatForBSATN);
+            string result;
+            if (fullName.StartsWith("System."))
```
lcodes, line 118: Can't you compare against type.Namespace here?

crates/bindings-csharp/BSATN.Codegen/Utils.cs
```diff
@@ -126,25 +113,19 @@ static string GetTypeInfoForNamedType(INamedTypeSymbol type)
...
-                // If we're here, then this is nullable *value* type like `int?`.
-                "System.Nullable<T>" => $"SpacetimeDB.BSATN.ValueOption",
-                var name when name.StartsWith("System.") => throw new InvalidOperationException(
-                    $"Unsupported system type {name}"
-                ),
-                _ => $"{SymbolToName(type)}.BSATN",
-            };
+                // C# doesn't have traits, so we put all the implementations for built-in types
+                // in a special namespace with 1:1 name mapping.
+                result = $"SpacetimeDB.BSATN.SystemImpls.{fullName[7..]}";
```
lcodes, line 122: There's a whole lot of System.* types, do we really want to allow System.Reflection.Metadata.MetadataReader to be caught by this for example?

crates/bindings-csharp/BSATN.Runtime/BSATN/Runtime.cs
```diff
@@ -1,476 +1,414 @@
...
+    // without having to touch the codegen at all, and without codegen having to hardcode any special types.
+    namespace SystemImpls
+    {
+        // This implementation is nearly identical to RefOption. The only difference is the constraint on T.
+        // Yes, this is dumb, but apparently you can't have *really* generic `T?` because,
+        // despite identical bodies, compiler will desugar it to very different
+        // types based on whether the constraint makes it a reference type or a value type.
+        public readonly struct Nullable<Inner, InnerRW> : IReadWrite<System.Nullable<Inner>>
+            where Inner : struct
+            where InnerRW : IReadWrite<Inner>, new()
```
lcodes, line 176: the RW can be constrained to a struct instead of new(), would that make the above comment fixed?

crates/bindings-csharp/BSATN.Codegen/Utils.cs
RReverser: `.ContainingNamespace` would have the full namespace, eg `System.Collections.Generic`. You'd then need to loop through parents to get to the actual top. That's what I did initially, but since we need this stringified name anyway, it's much simpler to do a string-based check as well.

crates/bindings-csharp/BSATN.Codegen/Utils.cs
RReverser: IMO yes, one way or another we should (and will) get a compilation error for unsupported types.

PR #1630, opened 2024-08-23
Allow converting new ModuleDef to old TableSchema

kazimuth:
# Description of Changes

This was not painful at all! The code is a mess but the tests pass, the paths commute! Gonna rewrite to be cleaner, but the logic is there.

It should be possible to go in reverse from TableSchema -> ModuleDef via RawModuleDefv8, I have not implemented that yet. Eventually we'll update TableSchema to look like v9 and then we can throw out that path. 

None

1, this is tricky but I think I've got it right.

# Testing

We test that this produces identical results for the old and the new path for all of the modules in `modules` (up to a small amount of sorting).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4512,9 +4512,11 @@ dependencies = [
 name = "spacetimedb-schema"
 version = "0.12.0"
 dependencies = [
+ "hashbrown 0.14.1",
  "itertools 0.12.0",
  "lazy_static",
  "proptest",
+ "serde_json",
  "smallvec",
  "spacetimedb-cli",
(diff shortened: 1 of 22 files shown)


kazimuth:
This is ready for review.

I've ended up just using `#[allow(deprecated)]` in a number of places using the old path, because most of those places are going to need to be retouched as we rewrite stuff to use the new APIs (#1648, #1649, #1650, #1651, #1652, #1653, #1654)

cloutiertyler (approved):
This LGTM. I'm sorry for all the mapping, but at least it makes you verify familiar with the old and new stuff.

PR #1632, opened 2024-08-26
smoketests: Test unauthorized publish without -c

kim:
When updating a database as an unauthorized user, also test that this is rejected without -c


bfops (approved):
Looks right to me!

bfops:
Although hmm.. I would expect the `Internal tests` check to be failing on this PR, since it adds a smoketest for something that's currently broken in private..

PR #1634, opened 2024-08-26
WASM ABI: `get_table_id` -> `table_id_from_name`

Centril:
# Description of Changes

Rename as the title says.

### DIFF
--- a/crates/bindings-csharp/Runtime/Internal/FFI.cs
+++ b/crates/bindings-csharp/Runtime/Internal/FFI.cs
@@ -23,8 +23,8 @@ public enum Errno : short
     EXHAUSTED = -1,
     OK = 0,
     HOST_CALL_FAILURE = 1,
+    NOT_IN_TRANSACTION = 2,
     NO_SUCH_TABLE = 4,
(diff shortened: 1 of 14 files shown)


RReverser (approved):

crates/bindings-csharp/Runtime/Internal/FFI.cs
```diff
@@ -23,13 +23,16 @@ public enum Errno : short
...
     NO_SUCH_TABLE = 4,
-    LOOKUP_NOT_FOUND = 2,
     NO_SUCH_ITER = 6,
     NO_SUCH_BYTES = 8,
     NO_SPACE = 9,
     BUFFER_TOO_SMALL = 11,
     UNIQUE_ALREADY_EXISTS = 12,
+
+    // TODO(1.0): remove this.
+    LOOKUP_NOT_FOUND = 99,
```
RReverser, line 35: It seems entirely unused? (well, the `ColumnValueNotFound` and `RangeNotFound` variants)

If so, let's just remove it here instead of changing opcode and moving around and leaving a comment - if anything, simply removing it seems like a smaller change.

PR #1639, opened 2024-08-27
WASM ABI: `insert` -> `datastore_insert_bsatn` & impl new semantics

Centril:
# Description of Changes

Rename `insert` to `datastore_insert_bsatn` and impl the new semantics wrt. auto-inc columns.
Also, fix the semantics when multiple sequences exist on a table.

cc #1460 

Fixes https://github.com/clockworklabs/SpacetimeDB/issues/1640

### DIFF
--- a/crates/bindings-csharp/Codegen.Tests/fixtures/server/snapshots/Module#PublicTable.verified.cs
+++ b/crates/bindings-csharp/Codegen.Tests/fixtures/server/snapshots/Module#PublicTable.verified.cs
@@ -188,7 +188,8 @@ partial struct PublicTable : SpacetimeDB.Internal.ITable<PublicTable>
             ));
     }
 
-    public void ReadGenFields(System.IO.BinaryReader reader) {
+    public void ReadGenFields(System.IO.BinaryReader reader)
+    {
         if (Id == default)
         {
             Id = BSATN.Id.Read(reader);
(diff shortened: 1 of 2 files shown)


crates/bindings-csharp/Runtime/Internal/FFI.cs
```diff
@@ -145,7 +147,11 @@ out RowIter out_
     );
 
     [LibraryImport(StdbNamespace)]
-    public static partial CheckedStatus _insert(TableId table_id, byte[] row, uint row_len);
+    public static partial CheckedStatus _datastore_insert_bsatn(
+        TableId table_id,
+        [MarshalUsing(CountElementName = nameof(row_len))] [Out] byte[] row,
```
RReverser, line 152: This array is used both as an input and output.

```suggestion
        [MarshalUsing(CountElementName = nameof(row_len))] [In, Out] byte[] row,
```

RReverser (approved):

RReverser:
Reviewed both Rust and C# on this one; aside from one minor issue above, the rest looks good.

crates/bindings-csharp/Runtime/Internal/FFI.cs
RReverser: Alternatively, let's maybe start using `Span` / `ReadOnlySpan` (same as Rust's `&mut [T]` / `&[T]`) when adding the new APIs.

They don't need all those extra attributes, and generally will be more flexible than arrays in terms of what they can accept as input.

We talked about switching to spans before, but it was never important enough to work on it as a refactor, but since you're adding new methods, we might as well use the better types :)

cloutiertyler (approved):
I approve the changes to the trait.rs file.

PR #1642, opened 2024-08-27
Added None and Gzip compression modes to the client protocol

lcodes:
Added None and Gzip options to the client protocol, and a new Compression type threaded back to the connect() function to configure it.


crates/core/src/client/messages.rs
```diff
@@ -28,46 +30,67 @@ pub trait ToProtocol {
...
+
+    let mut encoder = CompressorReader::new(reader, BUFFER_SIZE, COMPRESSION_LEVEL, LG_WIN);
+
+    let mut out = Vec::new();
+    encoder
+        .read_to_end(&mut out)
+        .expect("Failed to Brotli compress `SubscriptionUpdateMessage`");
+    out
+}
+
```
mamcx, line 74: The combination of both looks like will benefit from a shared trait or similar so we encode the invariants described in the comments and the `const` used there.


Like:

```rust
struct Compresor {
    algo: Compression,
}

impl Compresor {
    const BUFFER_SIZE: usize = 32 * 1024;
    // Separated values by algorithm
    const BROTOLI_LG_WIN: u32 = 22;
    
    fn compress(&self, data: &[u8]) -> Vec<u8> {
        match self.algo {
            Compression::None => data.to_vec(),
            Compression::Gzip => {
               ...
            }
            Compression::Brotli => {
             ...
            }
        }
    }
}
```

mamcx: It also could be moved all the compression into another crate or module

PR #1643, opened 2024-08-27
Move cursive-chat's example module-bindings away from symlinks

lcodes:
Cursive-chat fails to compile on windows because git doesn't recreate the symlinks correctly.

This fixes it by copying the module bindings from quickstart-chat instead of symlinking them.


RReverser:
> Cursive-chat fails to compile on windows because git doesn't recreate the symlinks correctly.

Huh weird, did you enable developer mode on your Windows for symlink support?

I was the one who moved this demo away from copying to symlinks in the past 😅 (and I'm on Windows too)

lcodes:
Ah hrm, I was under the impression symlinks on windows need extra setup and need git to run as administrator.

But you're right, it seems `git config --global core.symlinks true` alone works. Good enough for me.

lcodes:
Ah but we might want to document this in a developer setup somewhere

PR #1644, opened 2024-08-27
Fix issue with identity creation

jdetter:
# Description of Changes

This fixes the following issue:
```
REDACTED@geralt MINGW64 ~/REDACTED/REDACTED
$ bash ./sp_deploy.sh REDACTED-bot-test
Error: Identity 93dda09db9a56d8fa6c024d843e805d8262191db3b4ba84c5efcd1ad451fed4e is not valid for server REDACTED-bot-test
List valid identities for server REDACTED-bot-test with:
        spacetime identity list -s REDACTED-bot-test

REDACTED@geralt MINGW64 ~/REDACTED/REDACTED
$ spacetime server ^C

REDACTED@geralt MINGW64 ~/REDACTED/REDACTED
$ spacetime identity new --no-email -s REDACTED-bot-test
Error: Identity 5ebd1221afb10aaf61d6c8a527413d82a1ad9ee9ac9ea420e8dc80dc9350394e is not valid for server REDACTED-bot-test
List valid identities for server REDACTED-bot-test with:
        spacetime identity list -s REDACTED-bot-test

REDACTED@geralt MINGW64 ~/REDACTED/REDACTED
$ spacetime identity list -s REDACTED-bot-test
Identities for REDACTED-bot-test:
 DEFAULT  IDENTITY  NAME
```

@gefjon There is probably a deeper issue here that needs investigating but this fixes the surface level problem which is all I care about from the UX side.

corresponding GitHub label.

Not breaking

1

# Testing

- [x] `spacetime identity new -s <server>` works regardless of what existing identities you have.

### DIFF
--- a/crates/cli/src/subcommands/identity.rs
+++ b/crates/cli/src/subcommands/identity.rs
@@ -398,10 +398,6 @@ async fn exec_new(mut config: Config, args: &ArgMatches) -> Result<(), anyhow::E
         query_params,
     )?);
 
-    if let Ok(identity_token) = config.get_default_identity_config(server) {
-        builder = builder.basic_auth("token", Some(identity_token.token.clone()));
-    }
-
     let identity_token: IdentityTokenJson = builder.send().await?.error_for_status()?.json().await?;
     let identity = identity_token.identity;
 
--- a/crates/cli/src/subcommands/identity.rs
+++ b/crates/cli/src/subcommands/identity.rs
@@ -393,7 +393,7 @@ async fn exec_new(mut config: Config, args: &ArgMatches) -> Result<(), anyhow::E
         query_params.push(("email", email.as_str()))
     }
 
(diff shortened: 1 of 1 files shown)


crates/cli/src/subcommands/identity.rs
```diff
@@ -398,10 +398,6 @@ async fn exec_new(mut config: Config, args: &ArgMatches) -> Result<(), anyhow::E
         query_params,
     )?);
 
-    if let Ok(identity_token) = config.get_default_identity_config(server) {
```
jdetter, line 401: For some reason the identity grabbed here is not the correct identity for the server I'm connecting to and I'm getting an invalid identity response. We just shouldn't be supplying auth here at all IMO.

bfops: any chance there's a `spacetime.toml` in the project directory?

jdetter:
this doesn't work and we don't know why

PR #1646, opened 2024-08-28
[WASM ABI 1.0] Fix `bytes_source_read`

kim:
When decoding reducer args, `bytes_source_read` would incorrectly set
instead of advance the read cursor, causing decoding of large args to
fail.

1

# Testing

This would cause non-public tests to fail, which use large reducer args.
We may want to create a large-args test case here as well.

### DIFF
--- a/crates/core/src/host/wasmtime/wasm_instance_env.rs
+++ b/crates/core/src/host/wasmtime/wasm_instance_env.rs
@@ -848,7 +848,7 @@ impl WasmInstanceEnv {
                 env.call_reducer_args = None;
                 Ok(-1i32)
             } else {
-                *cursor = can_read_len;
+                *cursor += can_read_len;
                 Ok(0)
             }
         })


crates/core/src/host/wasmtime/wasm_instance_env.rs
```diff
@@ -848,7 +848,7 @@ impl WasmInstanceEnv {
                 env.call_reducer_args = None;
                 Ok(-1i32)
             } else {
-                *cursor = can_read_len;
+                *cursor += can_read_len;
```
bfops, line 851: @Centril why did this pass all the smoketests? Can we add test coverage for this?

coolreader18 (approved):

crates/core/src/host/wasmtime/wasm_instance_env.rs
kim: afaiu, it only fails when the reducer args are past a certain size. I didn't look close enough to know what size exactly, perhaps then we can make a test which fails in this repo.

PR #1656, opened 2024-08-29
Fix flaky smoketest due to race condition in subscribe

coolreader18:
This should fix it - now it'll only error if the initial update was never printed, i.e. the subscription never started. I'm about 90% sure that this should fix any race condition issues, but if there's any flakiness we can just go back to the fix from before.

### DIFF
--- a/smoketests/__init__.py
+++ b/smoketests/__init__.py
@@ -237,11 +237,14 @@ class Smoketest(unittest.TestCase):
             sys.stderr.writelines(proc.stderr)
         threading.Thread(target=stderr_task).start()
 
-        init_update = proc.stdout.readline()
-        code = proc.poll()
-        if code:
-            raise subprocess.CalledProcessError(code, fake_args)
-        print("initial update:", init_update)
+        init_update = proc.stdout.readline().strip()
+        if init_update:
+            print("initial update:", init_update)
+        else:
+            code = proc.poll()
+            if code:
+                raise subprocess.CalledProcessError(code, fake_args)
+            print("no inital update, but no error code either")
 
         def run():
             updates = list(map(json.loads, proc.stdout))
--- a/smoketests/__init__.py
+++ b/smoketests/__init__.py
@@ -241,10 +241,13 @@ class Smoketest(unittest.TestCase):
         if init_update:
             print("initial update:", init_update)
         else:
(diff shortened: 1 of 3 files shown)


smoketests/tests/add_remove_index.py
```diff
@@ -72,7 +72,7 @@ def test_add_then_remove_index(self):
         # There are no indices, resulting in an unsupported unindexed join.
         self.publish_module(name, clear = False)
         with self.assertRaises(Exception):
-            self.subscribe(self.JOIN_QUERY, n = 0)()
+            self.subscribe(self.JOIN_QUERY, n = 0)
```
bfops, line 75: if we've fixed the race condition, this change from #1536 is no longer necessary

smoketests/__init__.py
```diff
@@ -237,11 +237,14 @@ def stderr_task():
...
-            raise subprocess.CalledProcessError(code, fake_args)
-        print("initial update:", init_update)
+        init_update = proc.stdout.readline().strip()
+        if init_update:
+            print("initial update:", init_update)
+        else:
+            code = proc.poll()
+            if code:
+                raise subprocess.CalledProcessError(code, fake_args)
+            print("no inital update, but no error code either")
```
bfops, line 247: to double-check - is it intended that we trigger this _both_ if the `code` is `None` _and_` if the `code` is `0`?

coolreader18: Correct. Maybe it should check for 0, but it doesn't make much of a difference anyway.

bfops (approved):
Thanks for looking into this! Seems worth trying. It's definitely nicer UX if it's free of race conditions.

smoketests/__init__.py
coolreader18: Oh - it should wait. I see the potential race condition here.

bfops: Yeah I think that's right. I suspect it's basically when the process hasn't quite exited yet, but I was unable to reproduce locally so I couldn't really narrow done.

coolreader18:
Ran it `test_add_then_remove_index` 20 times with no failure, so it seems fine now, hopefully.

bfops (approved):
LGTM let's give it a shot

PR #1657, opened 2024-08-30
Directories structures

mamcx:
# Description of Changes

The first step is cleaning the way the directories of the database are used.

This PR implements a series of helpers that describe the desired layout of the files and directories of the database, binaries and installations.

This PR only define an API, no yet change the way how we manage the files. But when it get integrated will be a major break.

1

# Testing

- [x] *Print the layout of the directories as part of testing*


mamcx:
Closed because implement by @coolreader18

PR #1658, opened 2024-08-30
Automatic migration planning

kazimuth:
# Description of Changes

Implements planning, does not yet integrate.

N/A

0

# Testing

See added tests.

### DIFF
--- a/crates/schema/src/auto_migrate.rs
+++ b/crates/schema/src/auto_migrate.rs
@@ -1,0 +1,716 @@
+use crate::{def::*, error::PrettyAlgebraicType, identifier::Identifier};
+use spacetimedb_data_structures::error_stream::{CollectAllErrors, CombineErrors, ErrorStream};
+use spacetimedb_lib::db::raw_def::v9::TableType;
+use spacetimedb_sats::WithTypespace;
+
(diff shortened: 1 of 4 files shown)


crates/schema/src/auto_migrate.rs
```diff
@@ -0,0 +1,718 @@
...
+pub struct AutoMigratePlan<'def> {
+    /// The old database definition.
+    pub old: &'def ModuleDef,
+    /// The new database definition.
+    pub new: &'def ModuleDef,
+    /// The checks to perform before the automatic migration.
+    pub prechecks: Vec<AutoMigratePrecheck<'def>>,
+    /// The migration steps to perform.
+    /// Order should not matter, as the steps are independent.
+    pub steps: Vec<AutoMigrateStep<'def>>,
```
cloutiertyler, line 19: Is it guaranteed to not matter, or does it just happen to be that way because of the types of automigration we do now?

cloutiertyler (approved):
This looks really good. Left a comment.

kazimuth:
This now also fixes some stuff in ModuleDef introduced by miscommunication while collaborating with Noa

PR #1660, opened 2024-09-03
Rust module API rework

coolreader18:
WIP - waiting on abi rework implementation.


crates/bindings/src/lib.rs
```diff
@@ -65,23 +66,58 @@ pub struct ReducerContext {
...
+    type DbView;
+
+    /// Get a view into the tables.
+    ///
+    /// This method is provided for times when a programmer wants to be generic over the `DbContext` type.
+    /// Concrete-typed code is expected to read the `.db` field off the particular `DbContext` implementor.
+    /// Currently, being this generic is only meaningful in clients,
+    /// as modules have only a single implementor of `DbContext`.
+    fn db(&self) -> &Self::DbView;
+}
```
gefjon, line 100: `DbContext` in modules also has methods `rng`, `identity` and `address`. `identity` and `address` are subject to change and need not be implemented yet, but `rng` is important and should be included.

crates/bindings/src/lib.rs
```diff
@@ -235,204 +233,77 @@ pub fn delete_by_rel(table_id: TableId, relation: &[impl Serialize]) -> Result<u
...
-
-    /// A trait of result types which know how to convert a `Result<T: TableType>` into itself.
-    pub trait InsertResult {
-        type T: TableType;
-        fn from_res(res: Result<Self::T>) -> Self;
-    }
-}
+#[doc(hidden)]
+pub trait Captures<'a> {}
+impl<'a, T> Captures<'a> for T {}
```
gefjon, line 285: Now I am even more confused.

crates/bindings/src/lib.rs
```diff
@@ -235,204 +233,77 @@ pub fn delete_by_rel(table_id: TableId, relation: &[impl Serialize]) -> Result<u
...
 
 /// A UNIQUE constraint violation on table type `T` was attempted.
-pub struct UniqueConstraintViolation<T: TableType>(PhantomData<T>);
-impl<T: TableType> fmt::Debug for UniqueConstraintViolation<T> {
+#[derive(Debug)]
+#[non_exhaustive]
+pub struct UniqueConstraintViolation;
```
gefjon, line 290: This could potentially contain `table: String, column: String` for example. That can be future work, but please leave a TODO comment in here to get better error messages.

crates/bindings/src/lib.rs
```diff
@@ -454,133 +325,91 @@ pub mod query {
...
+                Some(row)
+            })
+        }
 
-        // Insert the new row.
-        Table::insert(new);
+        fn delete(&self, col_val: Self::ColType) -> bool {
+            delete_by_col_eq(Tbl::table_id(), COL_IDX.into(), &col_val)
+                // TODO: Returning `Err` here was supposed to signify an error,
+                //       but it can also return `Err(_)` when there is nothing to delete.
```
gefjon, line 383: @Centril will this still be true with the new ABI?

crates/bindings/src/lib.rs
```diff
@@ -454,133 +325,91 @@ pub mod query {
...
+                .unwrap_or(0)
+                > 0
+        }
 
-        // TODO: For now this is always successful.
-        //       In the future, this could return what `delete_by_field` returns?
-        true
+        fn update(&self, new_row: Self::Row) -> Self::Row {
+            todo!()
```
gefjon, line 389: Impl this in terms of `self.delete` and an insert. There's not going to be a host call which implements this method.

crates/bindings/src/lib.rs
```diff
@@ -455,133 +326,91 @@ pub mod query {
...
-    where
-        Table: TableType,
-    {
-        type Item = Table;
+    impl<Tbl: Table, IndexType, const INDEX_INDEX: usize> BTreeIndex for BTreeIndexImpl<Tbl, IndexType, INDEX_INDEX> {
+        type Row = Tbl::Row;
+        type IndexType = IndexType;
+
+        fn filter<B: table::BTreeIndexBounds<Self::IndexType, K>, K>(&self, b: B) -> impl Iterator<Item = Self::Row> {
+            todo!();
```
gefjon, line 408:
```suggestion
            todo!("Implement this method once the new ABI exposes physical index ops to WASM");
```

crates/bindings/src/lib.rs
```diff
@@ -455,133 +326,91 @@ pub mod query {
...
+            [].into_iter()
+        }
 
-        fn next(&mut self) -> Option<Self::Item> {
-            let mut cursor = &self.cursor;
-            (cursor.remaining() != 0).then(|| bsatn::from_reader(&mut cursor).unwrap())
+        fn delete<B: table::BTreeIndexBounds<Self::IndexType, K>, K>(&self, b: B) -> u64 {
+            todo!()
```
gefjon, line 413:
```suggestion
            todo!("Implement this method once the new ABI exposes physical index ops to WASM");
```

crates/bindings/src/rt.rs
```diff
@@ -217,7 +217,7 @@ macro_rules! impl_reducer {
             Ret: IntoReducerResult
         {
             #[allow(non_snake_case)]
-            fn invoke(&self, _ctx: ReducerContext, args: ($($T,)*)) -> Result<(), Box<str>> {
+            fn invoke(&self, _ctx: &ReducerContext, args: ($($T,)*)) -> Result<(), Box<str>> {
```
gefjon, line 220: I believe this branch is no longer valid because the `ctx` arg is now mandatory.

crates/bindings/src/table.rs
```diff
@@ -0,0 +1,414 @@
...
+    }
+
+    /// The error type for this table for unique constraint violations. Will either be
+    /// [`UniqueConstraintViolation`] if the table has any unique constraints, or [`Infallible`]
+    /// otherwise.
+    type UniqueConstraintViolation: MaybeError<UniqueConstraintViolation>;
+    /// The error type for this table for auto-increment overflows. Will either be
+    /// [`AutoIncOverflow`] if the table has any auto-incrementing columns, or [`Infallible`]
+    /// otherwise.
+    type AutoIncOverflow: MaybeError<AutoIncOverflow>;
```
gefjon, line 70: Again, not how the `Table` trait is specified.

crates/bindings/src/table.rs
```diff
@@ -0,0 +1,414 @@
...
+    () => {};
+    (@impl ($($V:ident),*), ($T:ident $(, $U:ident)+)) => {
+        impl<$($V,)* $T, $($U,)+ Term: BTreeIndexBoundsTerminator<$T>> BTreeIndexBounds<($($U,)+ $T, $($V,)*)> for ($($U,)+ Term,) {}
+        impl_btree_index_bounds!(@impl ($($V,)* $T), ($($U),*));
+    };
+    (@impl ($($V:ident),*), ($T:ident)) => {
+        impl<$($V,)* $T, Term: BTreeIndexBoundsTerminator<$T>> BTreeIndexBounds<($T, $($V,)*)> for (Term,) {}
+        impl<$($V,)* $T, Term: BTreeIndexBoundsTerminator<$T>> BTreeIndexBounds<($T, $($V,)*), SingleBound> for Term {}
+    };
+}
```
gefjon, line 280: Neat macro-ology!

crates/bindings/src/table.rs
```diff
@@ -0,0 +1,414 @@
...
+impl<T> BTreeIndexBoundsTerminator<T> for &T {}
+impl<T> BTreeIndexBoundsTerminator<T> for ops::Range<T> {}
+impl<T> BTreeIndexBoundsTerminator<T> for ops::RangeFrom<&T> {}
+impl<T> BTreeIndexBoundsTerminator<T> for ops::RangeFrom<T> {}
+impl<T> BTreeIndexBoundsTerminator<T> for ops::RangeInclusive<&T> {}
+impl<T> BTreeIndexBoundsTerminator<T> for ops::RangeInclusive<T> {}
+impl<T> BTreeIndexBoundsTerminator<T> for ops::RangeTo<&T> {}
+impl<T> BTreeIndexBoundsTerminator<T> for ops::RangeTo<T> {}
+impl<T> BTreeIndexBoundsTerminator<T> for ops::RangeToInclusive<&T> {}
+impl<T> BTreeIndexBoundsTerminator<T> for ops::RangeToInclusive<T> {}
```
gefjon, line 298: We need to discuss (on the proposal) whether we actually want to accept all combinations of `T` and `&T`.

gefjon (changes requested):
- You still have the macros looking for known concrete types for columns marked `autoinc`.
- I would really appreciate some comments and docs on the macros. In particular, listing some examples of expected inputs and outputs, and some explanation of why the different parts of the outputs are there and what parts of the runtime system they plug into.

crates/bindings-macro/src/lib.rs
```diff
@@ -626,119 +626,73 @@ fn spacetimedb_tabletype_impl(item: syn::DeriveInput) -> syn::Result<TokenStream
...
-                spacetimedb::query::update_by_field::<Self, #column_type, #column_index>(#column_ident, value)
-            }
-        });
-
-        unique_delete_funcs.push(quote! {
-            #vis fn #delete_func_ident(#column_ident: &#column_type) -> bool {
-                spacetimedb::query::delete_by_unique_field::<Self, #column_type, #column_index>(#column_ident)
+        let index_ty = quote!((#(#col_tys,)*));
+        index_accessors.push(quote! {
+            fn #ident(&self) -> impl spacetimedb::BTreeIndex<Row = #row_type, IndexType = #index_ty>
```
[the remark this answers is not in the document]
coolreader18: I had hoped/thought so, but in practice, no, you have to import the trait to call the methods.

crates/bindings-macro/src/lib.rs
```diff
@@ -626,119 +626,73 @@ fn spacetimedb_tabletype_impl(item: syn::DeriveInput) -> syn::Result<TokenStream
...
+        let ident = name.unwrap_or_else(|| {
+            Ident::new(
+                &field_names
+                    .iter()
+                    .map(ToString::to_string)
+                    .collect::<Vec<_>>()
+                    .join("_"),
+                quote!(#(#field_names)*).span(),
+            )
+        });
```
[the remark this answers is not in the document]
coolreader18: Both. I'll change it to require it be specified.

crates/bindings-macro/src/lib.rs
```diff
@@ -137,7 +134,7 @@ enum MacroInput {
     Disconnect,
     Index {
         ty: IndexType,
-        name: Option<String>,
+        name: Option<Ident>,
```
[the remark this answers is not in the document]
coolreader18: Both. I was trying to change the macros as little as possible, since there's already #1314 which I had hoped would be able to land first.

crates/bindings-macro/src/lib.rs
```diff
@@ -626,119 +626,73 @@ fn spacetimedb_tabletype_impl(item: syn::DeriveInput) -> syn::Result<TokenStream
...
-            }
-        });
-
-        unique_delete_funcs.push(quote! {
-            #vis fn #delete_func_ident(#column_ident: &#column_type) -> bool {
-                spacetimedb::query::delete_by_unique_field::<Self, #column_type, #column_index>(#column_ident)
+        let index_ty = quote!((#(#col_tys,)*));
+        index_accessors.push(quote! {
+            fn #ident(&self) -> impl spacetimedb::BTreeIndex<Row = #row_type, IndexType = #index_ty>
+                + spacetimedb::Captures<'_> + spacetimedb::Captures<'a> {
```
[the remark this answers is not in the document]
coolreader18: Annoying hacky way to get around a compilation error until rust-lang/rfcs#3498 drops

crates/bindings/src/rt.rs
coolreader18: Ah, I missed that

crates/bindings/src/table.rs
coolreader18: I couldn't tell how it was supposed to be from the proposal - "For tables with constraints, [TryInsertError] is an enum generated by the `#[spacetimedb::table]` macro. [...] TODO: specify a generated name for this type?". In the interest of generating as few identifiers as possible, I made it so that it would be an enum defined in the bindings crate, but with variants defined based on the given `Table`. Are you amenable to using this implementation? IMO it's a better UX & better docs than the generated enum version mentioned in the proposal.

crates/bindings/src/lib.rs
gefjon: If we add the trait later and move inherent methods to trait methods, that's a breaking change.

crates/bindings/src/table.rs
gefjon: I am amenable to this. Leave a comment on the proposal and I'll update it to reflect your version.

crates/bindings-macro/src/lib.rs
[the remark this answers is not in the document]
gefjon: Do you think using an `impl Trait` type here is tangibly better in some way than a concrete type? If so, I'm open to revising the proposal, but my intuition is that we should just return a `BTreeIndex<#row_type, #index_ty>`.

crates/bindings-macro/src/lib.rs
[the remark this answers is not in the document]
gefjon: Can you elaborate? What's the compiler error, and how does adding this do-nothing trait fix it?

crates/bindings-macro/src/lib.rs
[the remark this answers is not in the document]
coolreader18: Hm, I guess that's good - I assumed the proposal meant an autogenerated concrete type, but if it can be a lib-defined concrete type, then that sounds good.

crates/bindings-macro/src/lib.rs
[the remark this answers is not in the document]
coolreader18: It's explained more in depth in that [rfc](https://rust-lang.github.io/rfcs/3498-lifetime-capture-rules-2024.html), but basically - return types often "capture" lifetime & type parameters, which for a concrete type means like, mentioning `'a` in the type (`fn foo() -> X<'a>`). impl traits represent opaque types, which are implicitly defined and thus need to have rules of how and when to capture lifetimes. If the opaque type doesn't capture the lifetime, but the underlying type it represents does, that's a compile error. In edition2024, those rules will be changed to capture all in-scope lifetimes by default, with the opt-in ability to specify a specific set of lifetimes to capture instead (`fn foo() -> impl Trait + use<'a>`). For now, we have to polyfill that behavior by adding a dummy trait that just mentions the lifetime, but does nothing else - telling the compiler that the opaque type captures the lifetime.

crates/bindings-macro/src/lib.rs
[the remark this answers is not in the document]
coolreader18: The main thing is that we have to expose the implementation detail of the `const COL_IDX: u16` arg, but I think that's fine.

crates/bindings-macro/src/lib.rs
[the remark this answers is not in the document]
gefjon: I thought that `impl Trait + 'a` already did this.

crates/bindings/src/lib.rs
coolreader18: Well, it wouldn't necessarily be moving them - we could just keep the inherent methods and add the trait impl. From a user perspective, having to always import this weird `DbContext` trait just to use an rng seems unergonomic. I think using trait syntax to define this API is really helpful, but that doesn't mean we have to have a singleton trait with only one impl in the actual crate.

crates/bindings-macro/src/lib.rs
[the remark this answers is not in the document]
coolreader18: oh, that's why I assumed :upside_down_face:

> ```rust
> // Name may vary. Type may be generic over a lifetime bound.
> struct MyTableMyUniqueColumnUniqueIndex {
>     // private fields
> }
> 
> impl TableHandle {
>     // May return a shared reference `&MyTableMyUniqueColumnUniqueIndex`.
>     fn my_unique_column(&self) -> MyTableMyUniqueColumnUniqueIndex;
> }
> ```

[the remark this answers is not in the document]
gefjon: Ah, this is more than I intended to specify. Returning a concrete library type is fine. I'll make a note to update the proposal.

crates/bindings/src/lib.rs
gefjon: Does the same reasoning apply to clients, where you have to import the `DbContext` trait in order to do things like subscribe to queries and inspect the connection's status?

crates/bindings-macro/src/lib.rs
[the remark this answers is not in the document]
coolreader18: Welp, they're gone anyway, since I switched over to concrete types.

coolreader18:
Currently just blocked on #1704.

coolreader18:
I guess now it's just blocked on an implementation in ModuleHost

coolreader18:
Differences from proposal:
- TryInsertError is a generic enum, not an associated type (this has since been amended in the proposal)
  - `Table::UniqueConstraintViolation` and `AutoIncOverflow` associated types that define each variant of the TryInsertError enum.
  - `ctx.db.my_table()` returns a `&'_ my_table_Handle`, not a `my_table_Handle<'_>` (this is so that `TryInsertError<Self>` is trivially `'static`, and can be returned as a `Box<dyn Error>`)

coolreader18:
The merge is just for testing - I'll rebase once #1697 actually merges.

coolreader18:
@Centril do you know why this might be failing? In `BTreeIndex::filter()`:

```rust
        let index_id = Idx::index_id();
        let args = b.get_args();
        let (prefix, prefix_elems, rstart, rend) = args.args_for_syscall();
        let iter = sys::datastore_btree_scan_bsatn(index_id, prefix, prefix_elems, rstart, rend)
            .unwrap_or_else(|e| panic!("unexpected error from datastore_btree_scan_bsatn: {e}"));
        TableIter::new(iter)
```
It panics not at the lookup of the index id by name (searches the system tables), but in datastore_btree_scan_bsatn with "No such index" (searches `index_id_map`). Is the index not getting properly committed to the database state? As far as I can tell it is, but I don't know why else this would be failing.

coolreader18:
gruhhh.... the tests are failing because the rust module now declares snake_case table names but the csharp module doesn't

coolreader18:
theoretically this is all tests passing, besides that

coolreader18:
maybe

coolreader18:
yeah, rust tests are all passing for me locally

coolreader18:
Reviewer's discretion whether this is an acceptable solution, I guess.

gefjon:
Can you alter the C# module to define snake_case table names by changing `class MyTable` to `class my_table`?

gefjon (approved):

gefjon:
Don't merge until we have a companion PR for private.

PR #1661, opened 2024-09-03
Add schema::type_for_generate, update validation to use it

kazimuth:
# Description of Changes

This should significantly ease writing client codegen. @RReverser, the added types aren't serializable yet, I'll leave that to you if you'd like to do it in a further PR.

Note: I've added a requirement that all product and sum types in a ModuleDef have names for all elements, because I believe this is currently a requirement for client codegen. This can be removed if we ever figure out how to do client codegen without it.

None

1, the logic added here is somewhat intricate. It's tested against a good number of test cases, all of the modules in the repo, and a proptest, which gives me some confidence in it.

# Testing
See the added tests in the source code.


coolreader18 (approved):

cloutiertyler (approved):
These changes make sense to me in light of our discussion yesterday.

crates/schema/src/type_for_generate.rs
```diff
@@ -0,0 +1,516 @@
...
+                    } else {
+                        // We haven't processed it yet. It's either a ref to a valid use, or invalid.
+                        let def = self
+                            .typespace
+                            .get(ref_)
+                            .ok_or(TypeRefError::InvalidTypeRef(ref_))
+                            .and_then(|def| {
+                                if def == &AlgebraicType::Ref(ref_) {
+                                    // Self-reference.
+                                    Err(TypeRefError::RecursiveTypeRef(ref_))
```
RReverser, line 249: So this now prohibits recursive types altogether? Can we please keep them?

I currently can autogenerate C# AlgebraicType structure - which is recursive - with all its children types directly from Rust, and I'd like to make use of it after landing ModuleDef autogen in #1670.

If we error out on recursive types, then AlgebraicType as well as anything that depends on it - which includes ModuleDef - will error out and won't codegen anymore.

I think instead of an error we should have `AlgebraicTypeUse::Ref { ref_, is_recursive }` variant so that langs that need to explicitly box the reference (only Rust among langs we currently support), have boolean tag to know that they need to do so, whereas others that already keep all the objects on the heap, can deal with it as a normal reference.

crates/schema/src/type_for_generate.rs
```diff
@@ -0,0 +1,516 @@
...
+    )]
+    NonSpecialTypeNotAUse { ty: PrettyAlgebraicType },
+
+    #[error("internal codegen error: invalid AlgebraicTypeRef")]
+    TypeRefError(#[from] TypeRefError),
+
+    #[error("internal codegen error: type ref {ref_} was not pre-declared as a definition")]
+    NonDeclaredTypeDef { ref_: AlgebraicTypeRef },
+
+    #[error("internal codegen error: all type elements require names: {ty}")]
```
RReverser, line 31: I'd reword this, "type elements" sounds like talking about elements of the type (ProductTypeElement, SumTypeVariant), not types themselves, and can be confusing.

crates/schema/src/type_for_generate.rs
```diff
@@ -0,0 +1,516 @@
...
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub enum AlgebraicTypeUse {
+    /// A type where the definition is given by the typing context (`Typespace`).
+    /// In other words, this is defined by a pointer to another `AlgebraicType`.
+    /// An AlgebraicTypeUse must point to an `AlgebraicTypeDef`.
+    Ref(AlgebraicTypeRef),
+
+    /// The type of array values where elements are of a base type `elem_ty`.
+    /// Values [`AlgebraicValue::Array(array)`](crate::AlgebraicValue::Array) will have this type.
+    /// Stores an `Arc` for compression, the `Arc` is not semantically meaningful.
```
RReverser, line 115: I'm confused by this somewhat, why `Arc` and not plain `Box` if it's just for compression? Do we store reference-counted copies of the same type in several places?

crates/schema/src/type_for_generate.rs
```diff
@@ -0,0 +1,516 @@
...
+
+/// A typespace for generating client code.
+///
+/// The key difference is that this typespace stores only `AlgebraicTypeDef`s, not `AlgebraicType`s.
+/// We use the same `AlgebraicTypeRef`s from the original typespace.
+/// The difference is that `AlgebraicTypeRef`s ONLY point to `AlgebraicTypeDef`s.
+/// Chains of `AlgebraicTypeRef`s in the original `Typespace` are contracted to point to their ending `AlgebraicTypeDef`.
+#[derive(Debug, Clone)]
+pub struct TypespaceForGenerate {
+    defs: HashMap<AlgebraicTypeRef, AlgebraicTypeDef>,
```
RReverser, line 54: Could we also provide fully resolved type names here?

It would be nice not to have to worry about `#[sats(name)]`, `--namespace` and so on in each generator, and instead just have something akin to your `ScopedTypeName` prepared and ready to be used for codegen.

crates/schema/src/type_for_generate.rs
```diff
@@ -0,0 +1,516 @@
...
+    }
+}
+
+/// An algebraic type definition.
+#[derive(Debug, Clone)]
+pub enum AlgebraicTypeDef {
+    /// A product type declaration.
+    Product(ProductTypeDef),
+    /// A sum type declaration.
+    Sum(SumTypeDef),
```
RReverser, line 88: Can we have a separate variant for plain enum? They already need to be handled as a different type in languages without tagged enum support, and I imagine we might need that even more in the future.

coolreader18:
I added some utilities to ModuleDef that make it easier/possible to do codegen - would it be ok to add those here? Maybe I'll make a PR into this branch.

crates/schema/src/type_for_generate.rs
kazimuth: Ah, I thought we didn't support recursive types in modules though. Does the code generation for that pass through this, like are you using the module codegen to generate those? I'm happy to add that tag if so.

crates/schema/src/type_for_generate.rs
kazimuth: Ah, in fact that was the original error here but the error message persisted through a refactor.

crates/schema/src/type_for_generate.rs
kazimuth: If multiple types contain identical sub-`AlgebraicTypeUse`s, they will use the same `Arc`, yeah. It's acyclic unless you pass through a Ref, just hash-consing. I can replace with Box if we don't wanna pay the Arc cost.

crates/schema/src/type_for_generate.rs
kazimuth: The relevant ScopedTypeNames are stored in the containing ModuleDef, I can add a helper accessor, or just clone them in here.

crates/schema/src/type_for_generate.rs
kazimuth: 👍

crates/schema/src/type_for_generate.rs
```diff
@@ -0,0 +1,516 @@
...
+    /// The `U128` type. Values [`AlgebraicValue::U128(v)`](crate::AlgebraicValue::U128) will have this type.
+    U128,
+    /// The `I256` type. Values [`AlgebraicValue::I256(v)`](crate::AlgebraicValue::I256) will have this type.
+    I256,
+    /// The `U256` type. Values [`AlgebraicValue::U256(v)`](crate::AlgebraicValue::U256) will have this type.
+    U256,
+    /// The `F32` type. Values [`AlgebraicValue::F32(v)`](crate::AlgebraicValue::F32) will have this type.
+    F32,
+    /// The `F64` type. Values [`AlgebraicValue::F64(v)`](crate::AlgebraicValue::F64) will have this type.
+    F64,
```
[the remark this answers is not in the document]
kazimuth: Cyclic dependency. I could move it to the `schema` crate?

crates/schema/src/type_for_generate.rs
RReverser: > Does the code generation for that pass through this, like are you using the module codegen to generate those?

Yeah, since we don't have any other codegen, we use regular `spacetime generate` for ClientApi and for this new ModuleDef and then delete files we're not interested in heh.

Someday we might clean it up, but I imagine the basic codegen pipeline is still going to be shared between modules and "just types".

crates/schema/src/type_for_generate.rs
RReverser: Nah, makes sense. Although I guess `Rc` would work here too, since we never touch those structures concurrently?

kazimuth: I was worried about some obscure thing demanding `Sync` later and got cold feet

crates/schema/src/type_for_generate.rs
kazimuth: Cool, I'll fix that then.
Ugh now I need to actually handle cycles though. ;-;

crates/schema/src/type_for_generate.rs
RReverser: Fair enough, doesn't matter much anyway.

crates/schema/src/type_for_generate.rs
RReverser: Luckily, you use `Arc` so it should be as easy as downgrading it all to `sync::Weak`... maybe? hopefully? 😅

kazimuth: The cycles all pass through `AlgebraicTypeRef` so no need for that even.

crates/schema/src/type_for_generate.rs
kazimuth: Wait, the enum variant is confusingly named, the error message was right. Renamed.

crates/schema/src/type_for_generate.rs
kazimuth: In particular, you should be looking things up in here based on the `TypeDef`s in `ModuleDef`, so you'll have it available.

crates/lib/src/db/raw_def/v9.rs
```diff
@@ -296,6 +296,37 @@ pub struct RawScheduleDefV9 {
     pub reducer_name: RawIdentifier,
 }
 
+/// A constraint definition attached to a table.
```
kazimuth, line 299: This is also important

crates/sats/src/algebraic_type.rs
```diff
@@ -182,6 +182,22 @@ impl AlgebraicType {
...
+
+    /// Returns whether this type is a never type.
+    pub fn is_never(&self) -> bool {
+        matches!(self, Self::Sum(p) if p.is_empty())
+    }
+
+    /// If this type is the standard option type, returns the type of the `some` variant.
+    /// Otherwise, returns `None`.
+    pub fn as_option(&self) -> Option<&AlgebraicType> {
+        self.as_sum().and_then(SumType::as_option)
```
RReverser, line 198: Purely personal preference / nitpick, but I'd do

```suggestion
        self.as_sum()?.as_option()
```

crates/lib/src/db/raw_def/v9.rs
```diff
@@ -296,6 +296,37 @@ pub struct RawScheduleDefV9 {
...
+#[derive(Debug, Clone, SpacetimeType)]
+#[sats(crate = crate)]
+#[cfg_attr(feature = "test", derive(PartialEq, Eq, PartialOrd, Ord))]
+pub struct RawConstraintDefV9 {
+    /// The name of the constraint. Must be unique within the containing `RawDatabaseDef`.
```
RReverser, line 304: I see this type (`RawDatabaseDef`) mentioned in a lot of comments in `def.rs` and `v9.rs`, but it doesn't seem defined anywhere?

kazimuth: Yep, it became `RawModuleDef`. Good catch.

PR #1662, opened 2024-09-03
WIP: updating Schema and system tables

kazimuth:
# Description of Changes

Sketch of updating the system tables + updating ModuleDef to be a little easier to add constraints to, as suggested by Mazdak a while back.

I will probably split off the ModuleDef update into its own PR, since it is fairly contained and can be merged in the next day or two.
That is blocked on merging https://github.com/clockworklabs/SpacetimeDB/pull/1658 and https://github.com/clockworklabs/SpacetimeDB/pull/1661.

This code is full of errors, the draft is just here to allow review of the new Schema definitions.

This breaks the ABI and API, it rearranges system tables.

4, this is going to reverberate across the entire codebase.

# Testing

### DIFF
--- a/crates/core/src/db/datastore/system_tables.rs
+++ b/crates/core/src/db/datastore/system_tables.rs
@@ -173,7 +173,7 @@ macro_rules! st_fields_enum {
 // WARNING: For a stable schema, don't change the field names and discriminants.
 st_fields_enum!(enum StTableFields {
     "table_id", TableId = 0,
-    "table_name", TableName = 1,
+    "name", TableName = 1,
     "table_type", TableType = 2,
     "table_access", TablesAccess = 3,
 });
@@ -181,18 +181,16 @@ st_fields_enum!(enum StTableFields {
 st_fields_enum!(enum StColumnFields {
     "table_id", TableId = 0,
     "col_pos", ColPos = 1,
-    "col_name", ColName = 2,
-    "col_type", ColType = 3,
+    "name", ColName = 2,
+    "type", ColType = 3,
 });
 // WARNING: For a stable schema, don't change the field names and discriminants.
-st_fields_enum!(enum StIndexFields {
-    "index_id", IndexId = 0,
-    "table_id", TableId = 1,
-    "index_name", IndexName = 2,
-    "columns", Columns = 3,
-    "is_unique", IsUnique = 4,
-    "index_type", IndexType = 5,
-});
+st_fields_enum!(
+    enum StIndexFields {
+        // TODO(jgilles): how do we add a system table that stores only an enum?
+        // What if we need to add new variants to the enum?
+    }
+);
 // WARNING: For a stable schema, don't change the field names and discriminants.
 st_fields_enum!(
     /// The fields that define the internal table [crate::db::relational_db::ST_SEQUENCES_NAME].
@@ -208,13 +206,12 @@ st_fields_enum!(
     "allocated", Allocated = 8,
 });
 // WARNING: For a stable schema, don't change the field names and discriminants.
-st_fields_enum!(enum StConstraintFields {
-    "constraint_id", ConstraintId = 0,
-    "constraint_name", ConstraintName = 1,
-    "constraints", Constraints = 2,
-    "table_id", TableId = 3,
-    "columns", Columns = 4,
-});
+st_fields_enum!(
+    enum StConstraintFields {
+        // TODO(jgilles): how do we add a system table that stores only an enum?
+        // What if we need to add new variants to the enum?
+    }
+);
 // WARNING: For a stable schema, don't change the field names and discriminants.
 st_fields_enum!(enum StModuleFields {
     "database_address", DatabaseAddress = 0,
@@ -233,10 +230,10 @@ st_fields_enum!(enum StVarFields {
     "name", Name = 0,
     "value", Value = 1,
 });
-
 st_fields_enum!(enum StScheduledFields {
     "table_id", TableId = 0,
     "reducer_name", ReducerName = 1,
+    "name", Name = 2,
 });
 
 /// System Table [ST_TABLE_NAME]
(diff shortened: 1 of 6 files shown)


kazimuth:
Key questions: 
- Can the system tables store enums (with data) at all?
- Can we add variants to an enum stored in the system tables without it being a breaking change?
- Can we add a new system table without it being a breaking change?

gefjon:
First let me say I'm addressing this from the perspective of: we maintain a known schema for each system table, supplied by the host, and the data in the snapshot and/or log must conform to that schema. That may not always be the case, but it is the case now, and frankly I don't think we have time to change it before 1.0.

> Can the system tables store enums (with data) at all?

Sure. Why wouldn't they be able to?

> Can we add variants to an enum stored in the system tables without it being a breaking change?

First of all, obviously you can't take a newer module, commitlog or snapshot and run it on an older host if you do that. The old host will see what it believes to be an invalid discriminant and likely panic.

Second of all, you can only take an older module, commitlog or snapshot and run it on a newer host if the new variant doesn't change the BFLATN layout of the sum type. This means that the new variant's payload must have size and alignment less than or equal to the largest and most-aligned previously-existing variant(s).

> Can we add a new system table without it being a breaking change?

Again only in the new-host-old-data sense, and not the old-host-new-data sense, but sure, I don't see why not. I would expect a not-present table to behave the same as an empty table in either a snapshot or a commitlog. 'Course, we should test that, because there's a good chance I made an incorrect assumption somewhere in the snapshot-restoration code, but I don't think there's a fundamental problem or even a significant barrier to this working.

crates/core/src/db/datastore/system_tables.rs
```diff
@@ -208,13 +212,16 @@ st_fields_enum!(
...
+st_fields_enum!(
+    enum StConstraintFields {
+        "constraint_id", ConstraintId = 0,
+        "name", ConstraintName = 1,
+        // TODO(jgilles): can a system table store an enum?
+        // What if we need to add new variants to the enum? Will that break
+        // the schema? Does this need to store a byte array instead?
+        "data", ConstraintData = 3,
+    }
+);
```
cloutiertyler, line 224: I would expect this to mirror how constraints are represented in `ModuleDef`. If we're going to have an enum here, that's what we should do in `ModuleDef` as well.

Notably postgres appears to put them all in a single table, but I would like you to confirm that.

cloutiertyler: Ah JK you seem to have done that below.

crates/core/src/db/datastore/system_tables.rs
```diff
@@ -173,26 +173,30 @@ macro_rules! st_fields_enum {
 // WARNING: For a stable schema, don't change the field names and discriminants.
 st_fields_enum!(enum StTableFields {
     "table_id", TableId = 0,
-    "table_name", TableName = 1,
+    "name", TableName = 1,
     "table_type", TableType = 2,
     "table_access", TablesAccess = 3,
 });
```
cloutiertyler, line 179: If you're going to rename `table_name` to name, might as well do `table_type` and `table_access`.

Incidentally, I don't know what `table_type` even is. 

Or better yet, I would just leave these names as is to reducer code churn and errors and then we can do a final pass at the end. So leave it as:

```rust
st_fields_enum!(enum StTableFields {
    "table_id", TableId = 0,
    "table_name", TableName = 1,
    "table_name", TableName = 1,
    "table_type", TableType = 2,
    "table_access", TablesAccess = 3,
});
```

crates/core/src/db/datastore/system_tables.rs
```diff
@@ -173,26 +173,30 @@ macro_rules! st_fields_enum {
...
+st_fields_enum!(
+    enum StIndexFields {
+        "index_id", IndexId = 0,
+        "table_id", TableId = 1,
+        "index_name", IndexName = 2,
+
+        // TODO(jgilles): can a system table store an enum?
+        // What if we need to add new variants to the enum? Will that break
+        // the schema? Does this need to store a byte array instead?
+        "index_algorithm", IndexAlgorithm = 3,
```
cloutiertyler, line 197: This unfortunately doesn't tell me what the type of this field is.

Could you put a schema for all the tables in a comment like this:

```md
| table_id | table_name  | table_type | foobar          |
|----------|-------------|------------|-----------------|
| u32      | String      | u32        | IndexAlgorithm  |

enum IndexAlgorithm {
    Foobar(String)
}
```

Or something similar? It's just hard to comment on the changes without the type info.

kazimuth:
One other questions from Tyler:
> Should we store primary_key in the schema? I realize we don't currently use it and that it's not meaningful except to determine what should be an update vs an insert/delete, but how do you compare two ModuleDefs which differ by PrimaryKey in ponderautomigrate? I guess it just ignores them?

`primary_key` can always be extracted by loading the `ModuleDef`. I guess we need to decide if you need to be able to exactly reconstruct the `ModuleDef` from the system tables. Choices:
1. Keep information that only matters to client codegen in `ModuleDef`, not in the system tables.
    We'll need to keep `ModuleDef`s around next to the system tables for performing `ponder_automigrate`.
2. Add client codegen information (Type names, reducer names, misc exports, the typespace) to the system tables.
    Then we can exactly reconstruct the input `ModuleDef` as needed.

I'm not sure which is the best option here. 2 leaks some "nominal" information into the system tables, on the other hand, 1 stores information redundantly. 🤷

crates/core/src/db/datastore/system_tables.rs
```diff
@@ -173,26 +173,30 @@ macro_rules! st_fields_enum {
...
-    "index_type", IndexType = 5,
-});
+st_fields_enum!(
+    enum StIndexFields {
+        "index_id", IndexId = 0,
+        "table_id", TableId = 1,
+        "index_name", IndexName = 2,
+
+        // TODO(jgilles): can a system table store an enum?
+        // What if we need to add new variants to the enum? Will that break
```
mamcx, line 195: Yes, system tables can store any SATN type.

crates/core/src/db/datastore/system_tables.rs
mamcx: I agree to keep the prefix for the columns, that helps with debugging.

kazimuth: Yeah and joins.

kazimuth:
Closed in favor of #1697

PR #1665, opened 2024-09-04
CLI - Remove project-specific config files

bfops:
# Description of Changes

Remove support for project-specific spacetime config files. The global Spacetime config files are unchanged.

We plan to add these back once we have a more principled approach to configs.

Fixes https://github.com/clockworklabs/SpacetimeDB/issues/1666.

Breaks anybody using project-specific config files.

2

# Testing

Just existing smoketests


jdetter (approved):
I did not test this myself but we weren't using the project config anywhere so the likelyhood that this breaks something is very low.

Issue #1571, opened 2024-08-07
CLI - Help text is wrong for `spacetime sql`


bfops:
The help text suggests providing the anonymous args in the wrong order:

```
$ spacetime sql
error: the following required arguments were not provided:
  <--interactive|query>
  <database>

Usage: spacetime sql <--interactive|query> <database>

For more information, try '--help'.
```

But the database name actually needs to come before the query, not after:
```
$ spacetime sql -s local 'select * from Person' test-module
Error: HTTP status client error (400 Bad Request) for url (http://127.0.0.1:3000/database/dns/select%20*%20from%20Person)

$ spacetime sql -s local test-module 'select * from Person'
 name 
------
```

PR #1667, opened 2024-09-04
CLI - Fix helptext for `spacetime sql`

bfops:
# Description of Changes

Fixes https://github.com/clockworklabs/SpacetimeDB/issues/1571

No

1

# Testing

```
$ cargo run -- sql --help
   Compiling spacetimedb-cli v0.12.0 (/home/lead/work/clockwork-localhd/SpacetimeDBPrivate/public/crates/cli)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 14.07s
     Running `target/debug/spacetime sql --help`
Runs a SQL query on the database.

Usage: spacetime sql [OPTIONS] <database> <query>

Arguments:
  <database>
          The domain or address of the database you would like to query

  <query>
          The SQL query to execute

Options:
      --interactive
          Instead of using a query, run an interactive command prompt for `SQL` expressions

  -i, --identity <identity>
          The identity to use for querying the database. If no identity is provided, the default one will be used.

  -a, --anon-identity
          If this flag is present, no identity will be provided when querying the database

  -s, --server <server>
          The nickname, host name or URL of the server hosting the database

  -h, --help
          Print help (see a summary with '-h')
```


crates/cli/src/subcommands/sql.rs
```diff
@@ -32,13 +32,7 @@ pub fn cli() -> clap::Command {
                  .long("interactive")
                  .action(ArgAction::SetTrue)
                  .conflicts_with("query")
-                 .help("Runs an interactive command prompt for `SQL` expressions"),)
-        .group(
-            ArgGroup::new("mode")
-                .args(["interactive","query"])
-                .multiple(false)
-                .required(true)
-        )
```
bfops, line 41: this was redundant since both the args use `.conflicts_with`. However, putting them in a group caused `clap` to put the **anonymous** query arg before `database` in the help text, even though it was actually the second anonymous arg.

mamcx (approved):
LGTM

PR #1670, opened 2024-09-04
Migrate C# ModuleDef to V9

### LINKED ISSUE #1589: Test coverage - assert that equivalent Rust and C# modules produce the same DescribeModule output

### LINKED ISSUE #1948: Rewrite C# server modules off V8 to V9

### LINKED ISSUE #2061: Migrate C# scheduling to new API

RReverser:
# Description of Changes

This migrates module definition to V9 for C# modules as we're planning to drop V8, and it can't represent some of the new features.

In the process, also improved Debug pretty-printing for the Rust typespace and nested types (to use in future diffs in tests).

 - Fixes #2061.
 - Fixes #1589.
 - Fixes #1948.

corresponding GitHub label.

# Testing


RReverser:
Updated this PR to latest definitions. Thanks to #1661 + @coolreader18's #1675, I can verify that I'm now producing correct ModuleDef as it can generate Rust sources from a C# module. Further testing is still blocked by lack of V9 support in the actual runtime - I'm hitting https://github.com/clockworklabs/SpacetimeDB/blob/71c772248b527e790bf704e51206bdc6a550f330/crates/core/src/host/wasm_common/module_host_actor.rs#L159-L163 so leaving this PR as a draft.

crates/bindings-csharp/BSATN.Codegen/Type.cs
```diff
@@ -234,10 +234,13 @@ public void Write(System.IO.BinaryWriter writer, {{FullName}} value) {
                     {{write}}
                 }
 
-                public SpacetimeDB.BSATN.AlgebraicType GetAlgebraicType(SpacetimeDB.BSATN.ITypeRegistrar registrar) =>
+                public SpacetimeDB.BSATN.AlgebraicType.Ref GetAlgebraicType(SpacetimeDB.BSATN.ITypeRegistrar registrar) =>
```
kazimuth, line 237: I wish you could do this in Rust, types for specific enum elements.

crates/bindings-csharp/Codegen/Module.cs
```diff
@@ -431,11 +454,38 @@ public IEnumerable<View> GenerateViews()
...
             // Important: the position must be stored here, before filtering.
-            .Select((col, pos) => new Constraint(col, pos, col.GetAttrs(viewName)))
-            .Where(c => c.attr != ColumnAttrs.UnSet);
+            .Select((col, pos) => new Constraint(col, pos, col.GetAttrs(view)))
+            .Where(c => c.attr.HasFlag(filterByAttr));
+
+    public IEnumerable<ViewIndex> GetIndexes(TableView view) =>
+        Indexes.Where(i => i.Table == null || i.Table == view.Name);
+
+    // Reimplementation of V8 -> V9 constraint conversion in Rust.
```
kazimuth, line 469: I've ripped most of this logic out in Rust modules FYI. They no longer have a notion of the "constraints" bitfield at all. The v9 format doesn't need it. In particular you don't have to generate indexes for unique constraints, the v9 validation code will do that while parsing; unless you want to control the names of the generated indexes. (Which I was TRYING to completely standardize, but alas, overruled.)

crates/bindings-csharp/BSATN.Codegen/Type.cs
```diff
@@ -234,10 +234,13 @@ public void Write(System.IO.BinaryWriter writer, {{FullName}} value) {
...
                     registrar.RegisterType<{{FullName}}>(_ => new SpacetimeDB.BSATN.AlgebraicType.{{Kind}}(new SpacetimeDB.BSATN.AggregateElement[] {
                         {{MemberDeclaration.GenerateDefs(Members)}}
                     }));
+
+                SpacetimeDB.BSATN.AlgebraicType SpacetimeDB.BSATN.IReadWrite<{{FullName}}>.GetAlgebraicType(SpacetimeDB.BSATN.ITypeRegistrar registrar) =>
```
kazimuth, line 242: I'd never seen this syntax before, wouldn't mind a comment mentioning it's specifically overriding the superclass method due to type stuff.

modules/spacetimedb-quickstart/src/lib.rs
```diff
@@ -1,6 +1,6 @@
 use spacetimedb::{log, ReducerContext, Table};
 
-#[spacetimedb::table(name = person, public)]
+#[spacetimedb::table(name = person, public, index(name = age, btree(columns = [age])))]
```
kazimuth, line 3: You could also do an index annotation on `age`. Except it might not have an index generated clientside. Hmm. I wonder if it would cause Rust module compilation errors to have both a column and unique index on a field.

crates/schema/tests/ensure_same_schema.rs
```diff
@@ -0,0 +1,50 @@
...
+        diff.is_empty(),
+        "Rust and C# modules are not identical. Here are the steps to migrate from C# to Rust: {diff:#?}"
+    );
+}
+
+macro_rules! declare_tests {
+    ($($name:ident => $path:literal,)*) => {
+        $(
+            #[test]
+            #[serial]
```
kazimuth, line 37: I never knew about this annotation, that's useful.

crates/sats/src/lib.rs
```diff
@@ -234,3 +234,8 @@ where
 macro_rules! __make_register_reftype {
     ($ty:ty, $name:literal) => {};
 }
+
+// A helper for prettier Debug implementation, without extra indirection around Some("name").
```
kazimuth, line 238: Might as well make this a doc comment.

crates/bindings-csharp/Runtime/Internal/ITable.cs
```diff
@@ -2,121 +2,120 @@ namespace SpacetimeDB.Internal;
...
-    static abstract IEnumerable<TableDesc> MakeTableDesc(ITypeRegistrar registrar);
-
-    internal abstract class RawTableIterBase
+    public sealed class Enumerator(FFI.RowIter handle) : IDisposable
     {
-        public sealed class Enumerator(FFI.RowIter handle) : IDisposable
-        {
-            byte[] buffer = new byte[0x20_000];
-            public byte[] Current { get; private set; } = [];
+        byte[] buffer = new byte[0x20_000];
```
kazimuth, line 10: How often is one of these created? I've heard wasm memory allocators are often pretty slow, but I don't know wrt C#. It would be nice to poke at a C# module with a memory profiler, but I'm not sure if any such thing exists for wasm...

kazimuth (commented):
All of this looks good. The codegen changes look sensible and the better debug-formatting Rust-side is a good idea. My big question is whether it would make sense to get rid of the ColumnAttrs bitfield entirely.

crates/bindings-csharp/Codegen/Module.cs
RReverser: Ah nice, happy to remove it here.

crates/bindings-csharp/BSATN.Codegen/Type.cs
RReverser: It's an explicit interface implementation in C#, we use it quite a bit actually.

crates/schema/tests/ensure_same_schema.rs
RReverser: Yeah it's a separate proc-macro crate.

crates/bindings-csharp/Runtime/Internal/ITable.cs
RReverser: Yeah few people were concerned about it. I tried couple of approaches to replace it with something more sensible, but it didn't make any noticeable improvement to benchmarks - I assume having memory reused within Wasm instance helps.

Might "optimise" it anyway in the future, but definitely out of scope of this PR.

modules/spacetimedb-quickstart/src/lib.rs
RReverser: I think I did this because we don't yet have syntax for that in C# and I wasn't sure if an inline annotation in Rust would be completely equivalent to outer one in C#.

Might be better to keep them the same until we implement inline annotations in C# too.

RReverser:
> My big question is whether it would make sense to get rid of the ColumnAttrs bitfield entirely.

What would you replace it with?

crates/bindings-csharp/Codegen/Module.cs
RReverser: But wait, V9 still has `Constraints` and `Sequences` collections, so this particular function is still needed.

It's only conversion to indexes that I can remove.

RReverser: Looks like that breaks due to accessor name comparisons in "ponder auto diff" function, which now fail because auto-generated accessor names are not the same.

I guess that's what you meant by

> unless you want to control the names of the generated indexes. (Which I was TRYING to completely standardize, but alas, overruled.)

but I think autogenerated accessor names are the ones that are wrong in this case?

RReverser: I'm going to just revert this change for now, but I think we might need to fix autogeneration code regardless.

kazimuth:
> > My big question is whether it would make sense to get rid of the ColumnAttrs bitfield entirely.
> 
> What would you replace it with?

On the Rust side, we have a data structure listing indexes, a data structure listing constraints, a data structure listing sequences, and that's it. It seems to work fine, but the logic in C# may be structured differently so that that isn't as easy to do. No need to make the change if it's just makework.

kazimuth (approved):
I have no complaints. I think we should merge this ASAP.

crates/bindings-csharp/Codegen/Module.cs
kazimuth: That's fine. It can be a followup PR to pull this out and double check the names are standardized between C# and rust.

PR #1671, opened 2024-09-04
Commitlog offset index

Shubham8287:
# Description of Changes
Implements Commitlog offset proposal -https://github.com/clockworklabs/SpacetimeDBPrivate/pull/867

Reader integration is left to do in another PR.

N/A

# Testing
- unit tests, and manual run for integration testing

### DIFF
--- a/crates/commitlog/src/index/indexfile.rs
+++ b/crates/commitlog/src/index/indexfile.rs
@@ -1,0 +1,346 @@
+use std::{
+    fs::{self, File},
+    io,
+    marker::PhantomData,
+    mem,
+    path::PathBuf,
+};
+
+use log::debug;
+use memmap2::MmapMut;
+
+use super::{IndexError, IndexRead, IndexWrite};
+const OFFSET_INDEX_FILE_EXT: &str = ".stdb.ofs";
+const KEY_SIZE: usize = mem::size_of::<u64>();
+const ENTRY_SIZE: usize = KEY_SIZE + mem::size_of::<u64>();
+
+/// Returns the offset index file name based on the root path and offset
+pub fn offset_index_file_name(root: &PathBuf, offset: u64) -> PathBuf {
+    root.join(format!("{offset:0>20}{OFFSET_INDEX_FILE_EXT}"))
+}
+
+/// A mutable representation of an index file using memory-mapped I/O.
+///
+/// `IndexFileMut` provides efficient read and write access to an index file, which stores
+/// key-value pairs
+/// Succesive key written should be sorted in ascending order, 0 is invalid-key value
+#[derive(Debug)]
+pub struct IndexFileMut<Key: Into<u64> + From<u64>> {
+    // A mutable memory-mapped buffer that represents the file contents.
+    inner: MmapMut,
+    /// The number of entries currently stored in the index file.
+    num_entries: usize,
+
+    _marker: PhantomData<Key>,
+}
+
+impl<Key: Into<u64> + From<u64>> IndexFileMut<Key> {
+    fn num_entries(&self) -> Result<usize, IndexError> {
+        for index in 0.. {
+            match self.index_lookup(index) {
+                Ok((entry, _)) => {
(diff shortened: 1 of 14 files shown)


crates/commitlog/src/lib.rs
```diff
@@ -45,6 +51,16 @@ pub struct Options {
     ///
     /// Default: 65,535
     pub max_records_in_commit: NonZeroU16,
+    /// Whenever at least this many bytes have been written to the currently
+    /// active segment, an entry is added to its offset index.
+    ///
+    /// Default: 4096
+    pub offset_index_interval_bytes: NonZeroU64,
+    /// If `true`, require that the segment must be synced to disk before an
+    /// index entry is added.
```
kim, line 60:
```suggestion
    /// index entry is added.
    ///
    /// Setting this to `true` implies that the commitlog is synced to disk
    /// every `offset_index_interval_bytes`, regardless of other settings.
    /// 
    /// Setting it to `false` (the default) will update the index whenever 
    /// the commitlog is synced _and_ `offset_index_interval_bytes` have
    /// been written. This may mean that less entries are written to the index
    /// than strictly every `offset_index_interval_bytes`.
```

crates/commitlog/src/repo/mod.rs
```diff
@@ -17,6 +18,23 @@ pub use fs::Fs;
 #[cfg(test)]
 pub use mem::Memory;
 
+#[derive(Debug, Default, Clone, Copy)]
+pub struct TxOffset(u64);
```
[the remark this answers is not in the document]
Shubham8287: Made it alias, newtype is not useful right now but may be when we use different Indexes like timestamps based.
Though, It will require bunch of mechanical change, can be done separately or when needed.

crates/commitlog/src/lib.rs
Shubham8287: ``` Setting this to `true` implies that the commitlog is synced to disk
   every `offset_index_interval_bytes`, regardless of other settings.```

umm why offset index setting will make commitlog to fsync?

What I understood from the proposa isl, if this is `true`, commitlog offset index will only be written when commitlog syncs to disk, and it does not care about `offset_index_interval_bytes` setting.
So, this would be the case where less entry may written to the disk than every `offset_index_interval_bytes`.

If set to `false`,  index entries will be written every `offset_index_interval_bytes` without irrespective about commitlog fsync.

This is implemented as per my understanding only in the PR.

crates/commitlog/src/index/indexfile.rs
```diff
@@ -0,0 +1,334 @@
...
+    offset: u64,
+    cap: u64,
+) -> Result<IndexFileMut<Key>, IndexError> {
+    File::options()
+        .write(true)
+        .read(true)
+        .create_new(true)
+        .open(offset_index_file_name(path, offset))
+        .and_then(|file| {
+            file.set_len(cap * ENTRY_SIZE as u64)?;
```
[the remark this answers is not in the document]
Shubham8287: Added `set_len` in `open_index_file` as well.

crates/commitlog/src/lib.rs
kim:
```suggestion    
    /// index entry is added.
    ///
    /// Setting this to `false` (the default) will update the index every
    /// `offset_index_interval_bytes`, even if the commitlog wasn't synced.
    /// This means that the index could contain non-existent entries in the
    /// event of a crash.
    ///
    /// Setting it to `true` will update the index when the commitlog is synced,
    /// and `offset_index_interval_bytes` have been written.
    /// This means that the index could contain fewer index entries than
    //// strictly every `offset_index_interval_bytes`.

```

More like this?

kim (approved):

crates/commitlog/src/lib.rs
Shubham8287: Yes, this looks accurate

Issue #1672, opened 2024-09-04
CLI - Print out the path of the config file being updated


bfops:
# Background
This improves the clarity of where config updates are happening.

# Definition of "Done"
When the CLI mutates the config, it prints out the path of the file being updated.

PR #1673, opened 2024-09-04
CLI - Print out config path when saving

bfops:
# Description of Changes

Fixes https://github.com/clockworklabs/SpacetimeDB/issues/1672.

Nope!

1

# Testing
```
$ cargo run -- call foo bar
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.21s
     Running `target/debug/spacetime call foo bar`
Saving config to /home/lead/.spacetime/config.toml..
Error: The dns resolution of `foo` failed.
```


mamcx (approved):
LGTM

PR #1675, opened 2024-09-05
Switch Rust codegen to use new ModuleDef

coolreader18:
# Description of Changes

Based on #1661. The way I did this also allows for incremental porting of the typescript & csharp codegens - obviously it should happen soon, but this is nice cause it allows avoiding conflicts (unless Ingvar's already started working on the csharp one).

2

# Testing

Snapshot tests all pass.


crates/cli/src/subcommands/generate/mod.rs
```diff
@@ -240,114 +245,132 @@ pub struct GenCtx {
...
-    let items: Vec<GenItem> = items.collect();
-    let mut files: Vec<(String, String)> = items
-        .iter()
-        .filter_map(|item| item.generate(&ctx, lang, namespace))
-        .collect();
-    files.extend(generate_globals(&ctx, lang, namespace, &items));
-
-    Ok(files)
+pub fn generate(module: RawModuleDef, lang: Language, namespace: &str) -> anyhow::Result<Vec<(String, String)>> {
+    let module = match module {
```
kazimuth, line 249: You can just use `.try_into()`

kazimuth, line 249: Also this whole function looks good, this was exactly the structure I thought would make sense here while we port

crates/cli/src/subcommands/generate/rust.rs
```diff
@@ -16,60 +18,88 @@ type Indenter = CodeIndenter<String>;
...
-            }
-        }
-        AlgebraicType::Product(ProductType { elements }) => {
-            print_comma_sep_braced(out, elements, |out: &mut W, elem: &ProductTypeElement| {
-                if let Some(name) = &elem.name {
-                    write!(out, "{name}: ")?;
-                }
-                write_type(ctx, out, &elem.algebraic_type)
-            })?;
+        AlgebraicTypeUse::Unit => write!(out, "()")?,
```
kazimuth, line 69: Happy with this patch :)

crates/schema/tests/validate_integration.rs
```diff
@@ -18,8 +18,11 @@ const TEST_TABLE_ID: TableId = TableId(1337);
 #[allow(clippy::disallowed_macros)] // LET ME PRINTLN >:(
 fn validate_module(module_name: &str) {
     let module = CompiledModule::compile(module_name, CompilationMode::Debug);
-    let raw_module_def: RawModuleDefV8 =
+    let raw_module_def: RawModuleDef =
         extract_descriptions(module.path()).expect("failed to extract module descriptions");
+    let RawModuleDef::V8BackCompat(raw_module_def) = raw_module_def else {
+        panic!("no more v8")
+    };
```
kazimuth, line 25: So both C# and rust are patched to produce v9 now?

crates/schema/src/type_for_generate.rs
```diff
@@ -332,16 +332,24 @@ pub enum AlgebraicTypeUse {
 impl AlgebraicTypeUse {
     /// Extract all `AlgebraicTypeRef`s that are used in this type and add them to `buf`.`
     fn extract_refs(&self, buf: &mut HashSet<AlgebraicTypeRef>) {
+        self.for_each_ref(|r| {
+            buf.insert(r);
+        })
+    }
+
+    pub fn for_each_ref(&self, mut f: impl FnMut(AlgebraicTypeRef)) {
```
kazimuth, line 340: Comment pls

crates/cli/src/subcommands/generate/rust.rs
```diff
@@ -769,38 +711,39 @@ fn print_dispatch_imports(out: &mut Indenter) {
...
 
-fn iter_reducer_items(items: &[GenItem]) -> impl Iterator<Item = &ReducerDef> {
-    items.iter().filter_map(|item| match item {
-        GenItem::Reducer(reducer) => Some(reducer),
-        _ => None,
-    })
+fn iter_module_names(module: &ModuleDef) -> impl Iterator<Item = String> + '_ {
+    dbg!(module.types().map(|ty| (&ty.name, ty.ty)).collect::<Vec<_>>());
+    itertools::chain!(
+        // iter_tables(module).map(|table| table.name.deref().to_case(Case::Snake)),
```
kazimuth, line 717: comment

kazimuth (approved):
Minor nits otherwise I'm happy with this. I'm less familiar with how the codegen works in the nitty gritty but everything here makes sense to me.

crates/schema/tests/validate_integration.rs
coolreader18: No, not yet - the `panic!()` will happen when v8 is no longer being produced.

PR #1679, opened 2024-09-05
C# module benchmarks

RReverser:
# Description of Changes

These are ported from Rust benchmarks 1:1 (similarly to how `sdk-test-cs` is ported from `sdk-test`) by using Copilot + elbow grease for manual fixes.

When you set an environment variable `STDB_BENCH_CS=1`, this allows to run all the same module benchmarks as Rust currently does, except against an analogous C# implementation.

corresponding GitHub label.

# Testing

- [x] *Write a test you've completed here.* Tried running C# benchmarks.
- [ ] *Write a test you want a reviewer to do here, so they can check it off when they're satisfied.* Try running C# benchmarks.


RReverser:
Example run in Rust

```
special/stdb_module/large_arguments/64KiB
                        time:   [75.082 µs 75.849 µs 76.733 µs]
                        change: [-9.7556% -7.9926% -6.4687%] (p = 0.00 < 0.05)
                        Performance has improved.
```

vs C#

```
special/stdb_module/large_arguments/64KiB
                        time:   [1.2303 ms 1.2388 ms 1.2474 ms]
                        change: [+1500.6% +1523.1% +1543.4%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

RReverser:
For the sake of posterity, here's EXPERIMENTAL_WASM_AOT vs regular C#:

```
special/stdb_module/large_arguments/64KiB
                        time:   [147.81 µs 149.06 µs 150.45 µs]
                        change: [-86.822% -86.542% -86.228%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 14 outliers among 100 measurements (14.00%)
```

only 2x slower than Rust (which is in line with my early microbenchmarks).

cloutiertyler:
This is great to have!

modules/benchmarks-cs/ia_loop.cs
```diff
@@ -0,0 +1,372 @@
...
+        public float z = z;
+        public float vx = x + 10.0f;
+        public float vy = y + 20.0f;
+        public float vz = z + 30.0f;
+    }
+
+    public static ulong MomentMilliseconds()
+    {
+        // Idk why Rust uses complicated math, but it seems like it should always return 1.
+        return 1;
```
mamcx, line 33: It was purely to copy the way was done and use the time methods, but is likely that this could be just a `const`

modules/benchmarks-cs/ia_loop.cs
```diff
@@ -0,0 +1,372 @@
...
+            AgentLoop(newAgent, agent_targetable, surrounding_agents, current_time_ms);
+
+            count++;
+        }
+
+        Runtime.Log($"ENEMY IA LOOP PLAYERS: {players}, processed: {count}");
+    }
+
+    [SpacetimeDB.Reducer]
+    public static void init_game_ia_loop(uint initial_load)
```
mamcx, line 353:
```suggestion
   // We check only for a single pass in the game loop.
    public static void init_game_ia_loop(uint initial_load)
```

I think is good idea to preserve this comment

mamcx (approved):
LGTM

modules/benchmarks-cs/ia_loop.cs
RReverser: There are quite a few comments that I didn't copy from Rust version. I just added README to C# modules saying that code is inherited from Rust version, and that Rust version should act as a reference that people should look at.

PR #1680, opened 2024-09-05
Auto-generate C# ModuleDef bindings from Rust

RReverser:
# Description of Changes

This PR marks entire `ModuleDef` type tree (with all its dependencies) with `derive(SpacetimeType)`, so that it can be used in any SATS context.

Then, it's used to generate all the C# raw module types instead of us maintaining our own copies, so this PR is a net negative in terms of lines of code we need to maintain, despite the diff stats saying otherwise due to the `Autogen` folder.

This is the first part of migration to V9 (#1670) and makes the actual migration much easier, since we don't need to update types by hand or worry about mismatches between C# and Rust definitions. We should be able to reuse the same principle and derives to autogenerate type bindings to more languages in the future as well.

In the process I had to make a few semi-related fixes to the `generate/csharp.rs`, I'm happy to extract them out into their own PR if that's easier to review but it will be a dependency of this one.

corresponding GitHub label.

1

# Testing

- [x] Ran the usual C# tests.


coolreader18 (approved):
LGTM from the rust side

RReverser:
Tested with BitCraft; it will need some adjustments (they have some of the constructors that we now autogenerate here, so those manual copies will need to be removed), but otherwise seems to work.

PR #1681, opened 2024-09-06
Stopwatch and reducer rng

SteveBoytsun:
# Description of Changes

- Renamed `ConsoleTimer` to `LogStopwatch`
- Implemented `LogStopwatch` support in C# module SDK
- Implemented `ReducerContext.Rng` and removed old `Runtime.Random`

corresponding GitHub label.

2. There are a lot of files changed, but they're all fairly trivial (most are just renames)

# Testing

- [x] C# testing was done with the following reducer: 
```cs
[SpacetimeDB.Reducer]
public static void test(ReducerContext ctx) 
{
    LogStopwatch sw = new LogStopwatch("test12312");
    for (int i = 0; i < 10; i++)
    {
        SpacetimeDB.Runtime.Log(ctx.Rng.Next(10).ToString());
    }
    sw.End();
    sw.End();
}
```
- [x] Rust testing was done with the following reducer
```rs
#[spacetimedb(reducer)]
pub fn test() {
	let sw = spacetimedb::log_stopwatch::LogStopwatch::new("test");
	sw.end();
}
```


RReverser:
I feel like you and @Centril  are following different ABIs here... Here's his commit from yesterday. https://github.com/clockworklabs/SpacetimeDB/commit/df5ef6c81d9e9ad231f847a9cff8b9231675802f

ABI should stay intact in high-level API PRs.

crates/bindings-csharp/Runtime/LogStopwatch.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace SpacetimeDB
+{
+    public class LogStopwatch
```
RReverser, line 9: Can't look up the spec right now, but I believe we agreed that timers should also be IDisposable (ending the timer at the end of the using scope).

SteveBoytsun: I was back-and-forth on this, but I guess there's no harm in adding it

gefjon (changes requested):
Don't change the ABI or the Rust bindings in this PR, only do the C# API changes. The C# interface and implementation looks good.

SteveBoytsun:
> Don't change the ABI or the Rust bindings in this PR, only do the C# API changes. The C# interface and implementation looks good.

Reverted (and re-tested)

gefjon (approved):

crates/bindings-csharp/Runtime/LogStopwatch.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace SpacetimeDB
```
RReverser, line 7: Please remove unused usings (I believe none of these should be necessary? VSCode warns me about them, but maybe VS doesn't?) and use file-scoped namespace for consistency with other code.

crates/bindings-csharp/Runtime/LogStopwatch.cs
```diff
@@ -0,0 +1,30 @@
...
+        {
+            StopwatchId = StartStopwatchInternal(name);
+        }
+
+        private static uint StartStopwatchInternal(string name)
+        {
+            var name_bytes = Encoding.UTF8.GetBytes(name);
+            var id = Internal.FFI._log_stopwatch_start(name_bytes, (uint)name_bytes.Length);
+            return id;
+        }
```
RReverser: This can be just inlined into the constructor, no need for extra indirection and should be easier to read.

```suggestion
        public LogStopwatch(string name)
        {
            var name_bytes = Encoding.UTF8.GetBytes(name);
            StopwatchId = Internal.FFI._log_stopwatch_start(name_bytes, (uint)name_bytes.Length);
        }
```

crates/bindings-csharp/Runtime/LogStopwatch.cs
```diff
@@ -0,0 +1,35 @@
...
+        private static uint StartStopwatchInternal(string name)
+        {
+            var name_bytes = Encoding.UTF8.GetBytes(name);
+            var id = Internal.FFI._console_timer_start(name_bytes, (uint)name_bytes.Length);
+            return id;
+        }
+
+        void IDisposable.Dispose()
+        {
+            End();
```
RReverser, line 27: I believe we'll have a problem if someone does `using var ...` but decides to `.End()` manually early somewhere. In Rust it's not an issue because `end` consumes the instance by value, so `Drop` is guaranteed to run only once, but in C# that's not the case so we'll have a risk of double-free, leading to an error due to the timer already having been deleted.

crates/bindings-csharp/Runtime/Runtime.cs
```diff
@@ -11,15 +11,21 @@ public class ReducerContext
     public readonly DateTimeOffset Time;
     public readonly Address? Address;
 
+    /// <summary>
+    /// A reducer-specific instance of `System.Random` that is seeded by current reducer's timestamp. This object is unchanged throught the entire reducer call
+    /// </summary>
+    public Random Rng { get; private set; }
```
RReverser, line 17: Since we don't need to set it dynamically anymore, this can be just a readonly field like all the others in this context.

```suggestion
    public readonly Random Rng;
```

crates/bindings-csharp/Runtime/Internal/FFI.cs
```diff
@@ -221,4 +223,10 @@ public static partial CheckedStatus _bytes_sink_write(
         ReadOnlySpan<byte> buffer,
         ref uint buffer_len
     );
+
+    [LibraryImport(StdbNamespace)]
+    public static partial uint _console_timer_start([In] byte[] name, uint name_len);
```
RReverser, line 228: Please add strongly-typed id wrapper struct like we do for other IDs. It serves both as a documentation and makes it harder to pass the wrong integer into the wrong API, but in marshalling it behaves exactly the same way.

crates/bindings-csharp/Runtime/LogStopwatch.cs
```diff
@@ -0,0 +1,35 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace SpacetimeDB
+{
+    public sealed class LogStopwatch : IDisposable
+    {
+        private uint StopwatchId;
```
RReverser, line 11:
```suggestion
        private readonly uint StopwatchId;
```

Doesn't have to be mutable as it's supposed to be assigned only once upon construction.

crates/bindings-csharp/Runtime/LogStopwatch.cs
RReverser: Ah looks like it's because I didn't add `.editorconfig` to the solution, and VS can't find it otherwise.

These legacy solution workflows are annoying sometimes 😅 I'll add it in a PR.

SteveBoytsun: Done

crates/bindings-csharp/Runtime/LogStopwatch.cs
SteveBoytsun: Oh, guess it's leftover from when we still had `LogTimer.Start()`. Fixed

crates/bindings-csharp/Runtime/Runtime.cs
SteveBoytsun: I don't necessarily agree with with approach, but I changed it for the sake of consistency

crates/bindings-csharp/Runtime/Internal/FFI.cs
SteveBoytsun: Done, although it didn't like PInvoke method having a custom managed return type, so I had to add a custom Marshaller. Doesn't look like we had `TableId` or any other managed types as method returns previously (aside from `CheckedStatus`, which also has a custom Marshaller), so it may be a limitation of C# / WASM/ whatever

crates/bindings-csharp/Runtime/LogStopwatch.cs
SteveBoytsun: Done

crates/bindings-csharp/Runtime/Internal/FFI.cs
RReverser: Ugh yeah we used to have to do that for `Buffer`, we just don't return it by value anymore. In theory this should be among things the new FFI in .NET 9 fixes, fingers crossed.

crates/bindings-csharp/Runtime/LogStopwatch.cs
RReverser: Yeah that's pretty reasonable.

RReverser (approved):

crates/bindings-csharp/Runtime/LogStopwatch.cs
```diff
@@ -0,0 +1,30 @@
...
+using SpacetimeDB.Internal;
+
+namespace SpacetimeDB;
+
+public sealed class LogStopwatch : IDisposable
+{
+    private readonly FFI.ConsoleTimerId StopwatchId;
+    private bool WasStopped;
+
+    public LogStopwatch(string name)
```
RReverser, line 11: Looks like we need to update the proposal, it still has `private` constructor + a separate `Start()` method.

PR #1687, opened 2024-09-10
NFC: enforce more C# code style + enable it in Visual Studio

RReverser:
# Description of Changes

This is a follow-up to #1567:

 - Add `.editorconfig` to Visual Studio solution, otherwise devs on VS don't see same code style suggestions as VSCode and/or MSBuild. Note that Visual Studio automatically regenerates `.editorconfig` with all the properties explicitly set, hence the larger file than before.
 - Configure and auto-fix few more minor code style deviations.

corresponding GitHub label.

# Testing

### DIFF
--- a/crates/bindings-csharp/SpacetimeSharpSATS.sln
+++ b/crates/bindings-csharp/SpacetimeSharpSATS.sln
@@ -3,23 +3,25 @@ Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio Version 17
 VisualStudioVersion = 17.0.31903.59
 MinimumVisualStudioVersion = 10.0.40219.1
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Codegen", "Codegen\Codegen.csproj", "{1E37FF71-567A-4AC8-947E-117098F01142}"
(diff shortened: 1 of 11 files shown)


crates/bindings-csharp/SpacetimeSharpSATS.sln
```diff
@@ -3,24 +3,26 @@ Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio Version 17
 VisualStudioVersion = 17.0.31903.59
 MinimumVisualStudioVersion = 10.0.40219.1
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Codegen", "Codegen\Codegen.csproj", "{1E37FF71-567A-4AC8-947E-117098F01142}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Codegen", "Codegen\Codegen.csproj", "{1E37FF71-567A-4AC8-947E-117098F01142}"
```
kazimuth, line 6: Do we need to worry about these strings for packaging reasons or are they purely internal?

kazimuth (approved):
I Love Lint.

We probably want to mirror this in the C# SDK as well.

crates/bindings-csharp/SpacetimeSharpSATS.sln
RReverser: Purely internal. I don't even know why VS decided to regenerate them - I tried reverting, but it regenerated them on the next time to the same strings again, so I'm going to keep it.

PR #1688, opened 2024-09-10
Add C# diagnostics support

### LINKED ISSUE #1541: Improve Roslyn diagnostics
Currently having any issue in C# code that results in an exception inside Roslyn codegen, ends up printing a useless wall of errors (999+ in case of BitCraftClient), making it very hard to pin down the root cause, wasting precious development time.

Instead, I want to emit proper diagnostics and show them in IDE and MSBuild without all the supplemental noise.

RReverser:
# Description of Changes

Previously, whenever a Roslyn codegen (handler of `[SpacetimeDB.Type]` and other attributes) failed, MSBuild would simply skip it altogether and proceed to try and compile the rest of the codebase, just without autogenerated methods.

This would result in a terrible development experience: the CLI logs would contain tons of useless errors due to missing methods, losing the original message among the noise, and IDEs wouldn't show the error at the source location at all. On large projects like BitCraft any BSATN error (e.g. due to missing type / out-of-date SDK) would simply result in "999+ errors" view.

This PR adds proper Roslyn-based diagnostics for IDEs and CLI builds alike, with error descriptors pointing to the correct location in the source. It also implements graceful recovery instead of throwing an exception - we now keep trying to generate what we can in face of errors, minimising the noise from missing methods and making IntelliSense experience smoother.

Fixes #1541. 

corresponding GitHub label.

2

# Testing

- [x] `dotnet test` with new tests


crates/bindings-csharp/BSATN.Codegen/Diag.cs
```diff
@@ -0,0 +1,162 @@
...
+    private int idCounter = 0;
+
+    public string NextId() => $"{Tag}{++idCounter:D4}";
+}
+
+// Roslyn `Diagnostics` requires declaration using old-school string format ("foo {0} bar {1}")
+// which is not very type-safe when you want to pass arguments from arbitrary places in the codebase.
+//
+// This helper class allows us to define diagnostics in a type-safe way by using LINQ expressions
+// to extract the format and arguments from a normal-looking lambda.
```
kazimuth, line 20: Could you add examples of the two styles of usage?

crates/bindings-csharp/BSATN.Codegen/Type.cs
```diff
@@ -89,16 +100,16 @@ public BaseTypeDeclaration(GeneratorAttributeSyntaxContext context)
             Kind = TypeKind.Sum;
 
             if (
-                type.BaseType.TypeArguments[0]
+                type.BaseType.TypeArguments.FirstOrDefault()
```
kazimuth, line 103: I wonder if there's a lint for this, not using `[0]`.

crates/bindings-csharp/BSATN.Codegen/Utils.cs
```diff
@@ -1,30 +1,33 @@
...
 
 public static class Utils
 {
-    // Even `ImmutableArray<T>` is not deeply equatable, which makes it a common
-    // pain point for source generators as they must use only cacheable types.
+    // None of the standard collections - even `ImmutableArray<T>` - are deeply equatable,
+    // which makes it a common pain point for source generators as they must use only cacheable types.
     // As a result, everyone builds their own `EquatableArray<T>` type.
-    public readonly record struct EquatableArray<T>(ImmutableArray<T> Array) : IEnumerable<T>
+    // In our case, we use a `List<T>`-backed type because we want to extend it dynamically in a few places.
```
kazimuth, line 13: Is there any way to prevent extending a list that's already stored as a cache key?

crates/bindings-csharp/BSATN.Codegen/Diag.cs
```diff
@@ -0,0 +1,162 @@
...
+
+public interface IDiagHolder
+{
+    Diagnostics Diag { get; }
+}
+
+public class Diagnostics : EquatableList<Diagnostic>, IDiagHolder
+{
+    Diagnostics IDiagHolder.Diag => this;
+}
```
kazimuth, line 162: Slightly confused about this setup, what exactly is going on here?

Would appreciate a comment on IDiagHolder saying "should be implemented by ..., make sure you call Utils.ReportDiagnostics after you ..."

crates/bindings-csharp/Codegen.Tests/TestInit.cs
```diff
@@ -1,14 +1,69 @@
...
+        }
+    }
+
     [ModuleInitializer]
     public static void Initialize()
     {
-        VerifierSettings.OmitContentFromException();
+        if (Environment.GetEnvironmentVariable("CI") is null)
+        {
+            // Don't show the full diff in the console when running tests locally as Verify will open the diff tool automatically.
```
kazimuth, line 64: Will this happen in vscode? 🥺 

I guess I can just use visual studio for editing C# stuff.

crates/bindings-csharp/Codegen.Tests/Tests.cs
```diff
@@ -130,6 +138,38 @@ await Task.WhenAll(
...
+
+    [Fact]
+    public static async Task TestDiagnostics()
+    {
+        var sampleCompilation = await CompileFixture("diag");
+
+        // Run generators similarly to regular tests.
+        //
+        // However, Roslyn `Diagnostic` objects (error representation) are not cacheable.
+        // We can come up with custom intermediate representation, but for now we just skip caching checks.
```
kazimuth, line 152: 👍

kazimuth (commented):
This looks good to me, but I'll leave approval for other reviewers as requested.

crates/bindings-csharp/Codegen.Tests/TestInit.cs
RReverser: > Will this happen in vscode? 🥺

Yeah, it just opens whatever default tool you have set up as a diff editor. I think it picks one from Git settings maybe? For me it opens VSCode at least.

crates/bindings-csharp/BSATN.Codegen/Diag.cs
RReverser: Oh this is just a convenience for `ReportDiagnostics` to handle anything that has `Diagnostics Diag;` inside, as well as raw `Diagnostics` object itself (the last one is used in enum checks).

I'll add comments.

crates/bindings-csharp/BSATN.Codegen/Utils.cs
RReverser: We could split up types into separate "builder" and end list. Or use `ImmutableList` which has a more costly implementation, but, well, it's immutable. I'm not sure if it's worth the effort, but if you think so, I could add something.

crates/bindings-csharp/BSATN.Codegen/Type.cs
RReverser: Can't find anything at the moment.

crates/bindings-csharp/BSATN.Codegen/Diag.cs
RReverser: Do you mean the regular `Diagnostic` constructor vs this helper? Or which 2 styles?

RReverser:
@kazimuth To address the feedback, reverted back to regular `EquatableArray` like before this PR and now I instead have a `DiagReporter` class that is literally just a builder passed to constructors and not stored in data structures.

The data structures themselves no longer store the diagnostics and don't need the new interface, which makes the design simpler than in the initial implementation of this PR, and makes the data structures cacheable even in face of errors, so it's a win-win.

crates/bindings-csharp/BSATN.Codegen/Type.cs
SteveBoytsun: FYI `[0]` is not the same as `FirstOrDefault()`, and, unless you know optimization context of target platform, `[0]` can be considerably faster

RReverser: Yeah but this is not about speed, but about actually wanting the semantics of `FirstOrDefault()` where it returns nullable instead of throwing if there is no element. We could, of course, instead do manual length check + `[0]` access, but it shouldn't matter in practice, and this API is a bit cleaner.

crates/bindings-csharp/Codegen/Module.cs
```diff
@@ -155,11 +152,18 @@ public override Scope.Extensions ToExtensions()
...
+        {
+            return extensions;
+        }
+
         var autoIncFields = Members
             .Where(f => f.Attrs.HasFlag(ColumnAttrs.AutoInc))
             .Select(f => f.Name);
 
-        var iTable = $"SpacetimeDB.Internal.ITable<{ShortName}>";
+        var iTable = $"SpacetimeDB.Internal.ITable<{FullName}>";
```
lcodes, line 166: Is the switch from Short to Full name related to diagnostics? The module SDK PR also does these, but this should merge fine I think.

crates/bindings-csharp/Codegen.Tests/fixtures/server/server.csproj
```diff
@@ -2,11 +2,7 @@
 
   <PropertyGroup>
     <TargetFramework>net8.0</TargetFramework>
-  </PropertyGroup>
-
-  <PropertyGroup>
-    <IsPackable>false</IsPackable>
-    <DefaultItemExcludes>$(DefaultItemExcludes);snapshots/**/*</DefaultItemExcludes>
+    <ImplicitUsings>enable</ImplicitUsings>
```
lcodes, line 5: This seems like a bigger change than diagnostics. Does it become the new convention for all C# modules/client libraries too?

crates/bindings-csharp/Codegen.Tests/fixtures/diag/Lib.cs
```diff
@@ -0,0 +1,328 @@
...
+    EnumVariant248,
+    EnumVariant249,
+    EnumVariant250,
+    EnumVariant251,
+    EnumVariant252,
+    EnumVariant253,
+    EnumVariant254,
+    EnumVariant255,
+    EnumVariant256,
+    EnumVariant257,
```
lcodes, line 280: Sometimes I wish there were codegen inside codegen :)

crates/bindings-csharp/Codegen.Tests/fixtures/diag/Lib.cs
```diff
@@ -0,0 +1,328 @@
+﻿using SpacetimeDB;
+
+public enum LocalEnum { }
+
+[SpacetimeDB.Type]
+public partial struct TestUnsupportedType
+{
+    public DateTime UnsupportedSpecialTime;
```
lcodes, line 8: Shouldn't this say UnsupportedSpecialType?

crates/bindings-csharp/Codegen.Tests/fixtures/client/client.csproj
```diff
@@ -3,14 +3,8 @@
   <PropertyGroup>
     <!-- We want to ensure that BSATN.Codegen on its own produces code compatible with Unity. -->
     <!-- This means limiting project to .NET Standard 2.1 and C# 9. -->
-    <TargetFrameworks>netstandard2.1</TargetFrameworks>
+    <TargetFramework>netstandard2.1</TargetFramework>
```
lcodes, line 6: How are we testing against multiple dotnet language and runtime versions, out of curiosity?

crates/bindings-csharp/BSATN.Codegen/Type.cs
```diff
@@ -253,36 +263,47 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
...
-                        );
-                    }
-
-                    // Check that enums are compatible with SATS but otherwise skip from extra processing.
-                    return false;
-                },
-                transform: (context, ct) => new TypeDeclaration(context)
+        context
+            .SyntaxProvider.ForAttributeWithMetadataName(
+                "SpacetimeDB.TypeAttribute",
```
lcodes, line 300: Forgot to typeof(TypeAttribute).FullName this one?

crates/bindings-csharp/BSATN.Codegen/Diag.cs
```diff
@@ -0,0 +1,232 @@
...
+    public void Report<TContext>(ErrorDescriptor<TContext> descriptor, TContext ctx)
+    {
+        builder.Add(descriptor.ToDiag(ctx));
+    }
+
+    private DiagReporter() { }
+
+    private static readonly ErrorDescriptor<(SyntaxNode node, Exception e)> InternalError =
+        new(
+            new("STDBINT", "SpacetimeDB.Internal"),
```
lcodes, line 180: STDBINT? What about "STDB._" ?

lcodes: And can these be lowercase? I don't think diagnostics should scream at users :)

crates/bindings-csharp/BSATN.Codegen/Type.cs
```diff
@@ -8,11 +8,23 @@ namespace SpacetimeDB.Codegen;
...
-    public MemberDeclaration(string name, ITypeSymbol type)
-        : this(name, SymbolToName(type), GetTypeInfo(type)) { }
+    public MemberDeclaration(ISymbol member, ITypeSymbol type, DiagReporter diag)
+        : this(member.Name, SymbolToName(type), "")
+    {
+        try
+        {
+            TypeInfo = GetTypeInfo(type);
+        }
+        catch (Exception e)
```
lcodes, line 18: Is that wide a catch type needed? Won't that turn an internal error into an unsupported type?

crates/bindings-csharp/Codegen/Module.cs
RReverser: Yeah, otherwise it was emitting more residual errors (in the last test that checks for MSBuild noise). I don't remember which ones anymore, but I remember it was necessary to reduce the total count.

crates/bindings-csharp/Codegen.Tests/fixtures/server/server.csproj
RReverser: No, it did the same before, it was just inherited from top-level `Directory.Build.props`. But now that I added a separate `Directory.Build.props` in `fixtures`, I had to manually copy-paste properties that were no longer inherited (`.props` does not inherit from parent directories).

crates/bindings-csharp/Codegen.Tests/fixtures/diag/Lib.cs
RReverser: Well this is a `DateTime` property, so I figured "Time" naming would be slightly clearer. But I can see how this can look like a typo :)

crates/bindings-csharp/Codegen.Tests/fixtures/client/client.csproj
RReverser: We don't really, we have only one specific target we support on module SDK side (only .NET 8) and on client SDK side (only .NET Standard 2.1 for Unity compat).

crates/bindings-csharp/BSATN.Codegen/Type.cs
RReverser: No, BSATN.Codegen doesn't include BSATN.Runtime/Attrs.cs, at least yet. We don't really have useful attribute info, so I didn't bother sharing it between those packages too, for now that only happens in module's Runtime+Codegen.

crates/bindings-csharp/BSATN.Codegen/Type.cs
```diff
@@ -253,36 +263,47 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
         context
             .SyntaxProvider.ForAttributeWithMetadataName(
                 "SpacetimeDB.TypeAttribute",
```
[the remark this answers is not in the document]
RReverser: Same answer as above.

crates/bindings-csharp/BSATN.Codegen/Diag.cs
RReverser: These are diagnostic IDs, they're uppercase letters + 4 digits by convention, and most tooling likely expects them to remain so. The 2nd one is the "human-readable" category.

crates/bindings-csharp/BSATN.Codegen/Type.cs
RReverser: Yeah but an internal error during type resolution is due to an unsupported type one way or another, and the detailed error message will contain the original exception text anyway.

So even if there's an internal error, this will at least limit the context to one specific field and the text will explain why it couldn't be handled.

crates/bindings-csharp/Codegen.Tests/fixtures/diag/Lib.cs
lcodes: Yeah I think it's best since it matches the SpecialType enum we're seeing in codegen

lcodes (approved):

crates/bindings-csharp/Codegen/Module.cs
RReverser: Looks like this is not necessary anymore, probably because we have table views now which already use full name. Going to revert and see if there's anything else I can revert as well.

PR #1689, opened 2024-09-10
Fix table codegen wrt table names

coolreader18:
# Description of Changes

Fixes the issue encountered by @kim, introduced in https://github.com/clockworklabs/SpacetimeDB/pull/1314.


kim (approved):
Add a regression test to `rust-wasm-test` maybe?

```rust
#[spacetimedb::table(name = _LegacyPrivate)]
pub struct LegacyPrivate {
    sensitive: String
}
```

coolreader18:
The test is already in there - it wasn't underscore stripping from the table name, it was just using the type name instead of the table name. The snapshot changes are a result, so they'll monitor this regression.

kim:
I’m not so much concerned about the underscore, but any potential future change which would restrict what is a valid table name. Ideally, I’d like to see a randomized test which for now accepts any sequence of characters.

PR #1692, opened 2024-09-10
Rework Rust SDK for new API

gefjon:
# Description of Changes

The Rust SDK rework for the new stable API!

Still TODO, to be follow-ups:
- Subscription modification / mutable subscription sets. This is waiting on actually implementing the internals.
- Identity and auth rework. This is waiting on that proposal being accepted and implemented.
- Non-unique BTree indices.
- Supporting PKs of arbitrary SATS types. Currently we impose `Hash + Eq` because it's easier.

Reworks the Rust SDK to match the new stable API. This includes a significant rewrite of the SDK internals, since the previous internals were ill-suited to implement the new API.

Oh baby it sure is!

3 - A pretty major rewrite of the Rust SDK, but we have reasonable test coverage.

# Testing

- [x] SDK tests rewritten and pass.
  - [ ] Except `resubscribe`, which is currently disabled and would be broken.
- [x] Ran quickstart-chat and played around a bit.


kim (approved):

coolreader18 (approved):
I don't think I have the familiarity with the sdk (nor the time) to review & approve the entire thing, but from a `Table` API and codegen perspective, LGTM.

gefjon:
> I don't think I have the familiarity with the sdk (nor the time) to review & approve the entire thing, but from a `Table` API and codegen perspective, LGTM.

Luckily we have pretty good test coverage to demonstrate that the thing works, so my main concern is that it matches the proposal and the other impls.

PR #1694, opened 2024-09-10
CLI - `dns` subcommand becomes `rename`

bfops:
# Description of Changes

The `dns` subcommand is half-finished. We are simplifying it for now: it is now just `spacetime rename`.

This removes several `dns` subcommands, and renames `dns set-name` to just `rename`.

2

# Testing

Existing CI has been updated (see https://github.com/clockworklabs/SpacetimeDB/pull/1694/files#diff-c05a1f93f7bbe434d3be3d080b9aff1750af73d7e1fc2870a8a9c9c9550e89e9).


cloutiertyler:
I'm recommending we keep this as a P1 because if we don't then we need the docs to explain what `dns` is and all that and we don't want to bother with that because it's going to be confusing since it's not literally `dns` at this point.

gefjon:
Tyler says: we should have a `spacetime database rename` command (or `spacetime rename`? Like what we do for servers.) which renames a database. We should call these "database names," and not use the word "DNS."

crates/cli/src/util.rs
```diff
@@ -46,7 +46,7 @@ pub async fn spacetime_dns(
 /// identity will be looked up in the config and it will be used instead. Returns Ok() if the
 /// domain is successfully registered, returns Err otherwise.
 pub async fn spacetime_register_tld(
-    config: &mut Config,
+    config: &Config,
```
bfops, line 49: unrelated cleanup. I think this function used to manage identities in some way, but no longer needs a `&mut Config` since our identity management is more global now.

crates/cli/src/subcommands/dns.rs
```diff
@@ -1,97 +1,37 @@
...
-async fn exec_register_tld(mut config: Config, args: &ArgMatches) -> Result<(), anyhow::Error> {
-    let tld = args.get_one::<String>("tld").unwrap().clone();
+    let domain = args.get_one::<String>("new-name").unwrap();
+    let database_identity = args.get_one::<String>("database-identity").unwrap();
     let server = args.get_one::<String>("server").map(|s| s.as_ref());
+    let identity = decode_identity(&config)?;
+    let auth_header = get_auth_header(&config, false)?;
 
-    match spacetime_register_tld(&mut config, &tld, server).await? {
+    match spacetime_register_tld(&config, domain, server).await? {
```
bfops, line 34: note: this new `exec` function is a combination of the previous `exec_register_tld` and `exec_set_name` functions, now all-in-one.

cloutiertyler:
Could you add a test for this? Or point to the specific CI test that tests this command?

bfops:
> Could you add a test for this? Or point to the specific CI test that tests this command?

https://github.com/clockworklabs/SpacetimeDB/pull/1694/files#diff-c05a1f93f7bbe434d3be3d080b9aff1750af73d7e1fc2870a8a9c9c9550e89e9 has been updated

smoketests/tests/domains.py
```diff
@@ -3,47 +3,18 @@
...
+        # This should throw an exception before there's a db with this name
+        with self.assertRaises(Exception):
+            self.spacetime("logs", rand_name)
+
+        self.spacetime("rename", rand_name, self.database_identity)
 
-        names = self.spacetime("dns", "reverse-lookup", self.database_identity).splitlines()
-        self.assertIn(rand_name, names)
+        # Now we're essentially just testing that it *doesn't* throw an exception
+        self.spacetime("logs", rand_name)
```
cloutiertyler, line 20: Not to be a stickler, but would it be possible to add the following tests:

- Test that you can't rename to a database that already exists
- Test that you can't publish a database with an invalid name (e.g. that includes special characters)

I would also keep the tests that test the slash functionality. We should only remove those tests when we actually change the server behavior. It's currently testing that behavior not this CLI comamnd.

cloutiertyler (changes requested):
See single comment.

smoketests/tests/domains.py
bfops: I've added the renaming test.

What special characters should be disallowed? I tried `!@#$%^&*()` and was able to publish successfully.

In a perfect world, I would argue that the "disallowed characters" test should be more internal, i.e. testing that the endpoints will return an error when using one of those symbols. Ideally I would say the CLI tests are relatively "thin" and test that the CLI does indeed correctly trigger the endpoints we expect it to.

bfops: I've added back separate tests for the behavior of `/`

cloutiertyler: Well these are not solely CLI tests but just generally smoke tests, although I do agree that should be a unit test/http test as well.

cloutiertyler (approved):
These changes now look good to me.

PR #1695, opened 2024-09-10
CLI - Remove admin `energy` subcommands

bfops:
# Description of Changes

The `set-balance` subcommand is half-finished. We are removing it for now. We may add it back in future reworks of the energy system.

Removes the `spacetime energy set-balance` subcommand.

1

# Testing

Existing CI only.


mamcx (approved):
LGTM

Issue #1370, opened 2024-06-11
Remove module `update` functionality in favor of future migration/publish argument implementation


cloutiertyler:
# Callback when publishing without -c

## Problem

The BitCraft team wants some custom logic to execute when they publish a module without using `-c`. One thing Alexander wants to be able to do is change the duration of a scheduled reducer (because the durations are part of the static data, which they can update from the client).

## Proposed Solution #1

We could add a callback function that is analogous to the `init(..)` function, except its executed when publishing without `-c`. Basically in this new `update(..)` function he could lookup any scheduled reducers, cancel them all, and then reschedule them with the new updated value.

Perhaps we decorate like this:

```rust
#[spacetimedb(init)]
pub fn init() {
  // Do stuff in here that should only be done when starting from scratch
}

#[spacetimedb(update)]
pub fn update() {
  // Do stuff in here that should be done when starting from scratch *or* when a module is updated
}
```

This solution is in the same vein as a migration handler. As specified here this function would not have any ability to migrate any data, but it would give you a handle to be able to do operations on your module

Some questions to answer before we move forward on this:

- As part of the module lifecycle, maybe `init` is called when you publish with `-c` and `module_updated(..)` is also called when you publish with `-c`, *but* when you publish without `-c` only `module_updated(..)` is called.

## Proposed Solution #2

(pasted from comments section)

For now we could just tell the BitCraft team to refactor their code. Any code that manages repeating reducers should be in a new reducer called `manage_reducers(..)` . This reducer would be called from `init`, but it would also be manually invoked after publishing without `-c`.

This ticket is about number 1 here:*

1. You haven’t changed the schema and you’re intending to update the module and you want to run something when that happens
2. You have changed the schema, but you haven’t written a migration function. If we continue with this you will be in a borked state so we should throw an error
3. You want to do a migration, you’ve done a schema change and you’ve written a migration function.

cloutiertyler:
I believe this is possibly already implemented, but we should use this ticket to make sure we're happy with the current design.

kim:
Yes we have that: `init` is called on the first publish, and `update` is called on every subsequent update (aka publish-without-dash-c). Both are transactional, i.e. if the reducers return an error, nothing is committed to the database and either no database is running (on `init`), or the old database continues running (on `update`).

cloutiertyler:
[...]

PR #1697, opened 2024-09-11
Update TableSchema & system tables to resemble ABI V9

kazimuth:
# Description of Changes

This gets rid of most of the usages of `RawModuleDefV8` in `core`. It replaces it with `ModuleDef`, which means that the new validation path is used consistently throughout the codebase. It also refactors the system tables & `TableSchema` types to match the new `ModuleDef`, the `update` functionality to use the new auto-migration functionality based on `ModuleDef`, and does a little bit of other cleanup.

The major user-visible changes from this PR are:
- Some modules that used to be allowed may be rejected by the new, intensive validation rules.
- The `describe` endpoint now returns a json-serialized `RawModuleDefV9` rather than `RawModuleDefV8`.

The major logical changes are:
- `ConstraintSchema` and `IndexSchema` (and their corresponding system tables) no longer store redundant information. Constraints are completely separate from indices; to determine if an index is unique, look up a constraint with its columns. (This is not *quite* true in `mut_tx.rs` and the `table` crate, see my discussion with tyler below, but this is now true throughout more of the codebase.)
- Also note that `IndexSchema` stores a `ColList`, to keep track of field ordering, whereas `ConstraintSchema` stores a `ColSet`, which discards ordering, but can be cheaply constructed from a `ColList`.
- The `Constraints` type is now constructed on-demand from `TableSchema`, rather than being stored directly in `TableSchema`. This means we have a single source of truth for the data in `Constraints` and are less likely to muck it up. `Constraints` is still used in the `vm` and `sql` modules.

ABI break in the system tables, API break at the describe endpoint.

4, this is a system table update and touches things all over `core`.

# Testing

Updated all tests and added some. See also https://github.com/clockworklabs/SpacetimeDB/pull/1737


crates/core/src/db/datastore/system_tables.rs
```diff
@@ -237,229 +243,271 @@ st_fields_enum!(enum StVarFields {
...
+/// See the comment on [`SYSTEM_MODULE_DEF`].
+fn system_module_def() -> ModuleDef {
+    let mut builder = RawModuleDefV9Builder::new();
+
+    // System Table [ST_TABLE_NAME]
+    //
+    // | table_id | table_name  | table_type | table_access |
+    // |----------|-------------|----------- |------------- |
+    // | 4        | "customers" | "user"     | "public"     |
+    builder
```
kazimuth, line 260: Note: I am going to change these to use `add_type`, since `St(...)Row` now implements `SpacetimeType`. I'll add asserts in this method that the fields are in the correct position and have the right labels.

crates/core/src/db/datastore/system_tables.rs
```diff
@@ -1242,11 +1329,22 @@ impl TryFrom<RowRef<'_>> for StScheduledRow<Box<str>> {
...
+*/
+
+thread_local! {
+    static READ_BUF: RefCell<Vec<u8>> = RefCell::new(Vec::new());
+}
+
+/// Read a value from a system table via BSatn.
+fn read_via_bsatn<T: DeserializeOwned>(row: RowRef<'_>) -> Result<T, DBError> {
+    READ_BUF.with_borrow_mut(|buf| Ok(row.read_via_bsatn::<T>(buf)?))
+}
```
kazimuth, line 1347: This is now how all system table rows are going to be read. I haven't finished updating this file yet.

crates/primitives/src/ids.rs
```diff
@@ -53,3 +53,4 @@ system_id!(SequenceId, u32);
 system_id!(IndexId, u32);
```
kazimuth, line 53: These now need to implement Serialize, Deserialize, and SpacetimeType. They are going to pretend to be their wrapped types.

crates/core/src/sql/compiler.rs
```diff
@@ -135,18 +135,15 @@ fn compile_columns(table: &TableSchema, cols: &[ColId]) -> DbTable {
     let mut columns = Vec::with_capacity(cols.len());
     let cols = cols
         .iter()
+        // TODO: should we error here instead?
+        // When would the user be passing in columns that aren't present?
         .filter_map(|col| table.get_column(col.idx()))
```
mamcx, line 140: This will be gone when the new `sql parser` lands, and it will do the proper checkings.

crates/lib/src/relation.rs
```diff
@@ -92,16 +93,26 @@ pub struct Header {
...
 
 impl Header {
+    /// Create a new header.
+    /// Note that equal ColLists with different Constraints will have their constraints unioned.
     pub fn new(
         table_id: TableId,
         table_name: Box<str>,
         fields: Vec<Column>,
-        constraints: Vec<(ColList, Constraints)>,
+        uncompressed_constraints: impl IntoIterator<Item = (ColList, Constraints)>,
```
mamcx, line 106:
```suggestion
       proposed_constraints: impl IntoIterator<Item = (ColList, Constraints)>,
```

I think `compression` conveys a different thing.

crates/schema/src/schema.rs
```diff
@@ -46,23 +49,41 @@ trait Schema: Sized {
...
     columns: Vec<ColumnSchema>,
+
+    /// The primary key of the table, if present. Must refer to a valid column.
+    ///
+    /// Currently, there must be a unique constraint and an index corresponding to the primary key.
+    /// Eventually, we may remove the requirement for an index.
+    ///
+    /// The database engine does not actually care about this, but client code generation does.
+    pub primary_key: Option<ColId>,
+
```
mamcx, line 68: I think this should instead be a method that returns the `pk` from the `constraints`. There is a good reason to cache this?

kazimuth: There no longer are any `Constraints` stored in the TableSchema. This is the single source of truth for the primary key.

crates/core/src/db/datastore/system_tables.rs
```diff
@@ -235,231 +248,264 @@ st_fields_enum!(enum StVarFields {
...
+    //
+    // | table_id | table_name  | table_type | table_access |
+    // |----------|-------------|----------- |------------- |
+    // | 4        | "customers" | "user"     | "public"     |
+    let st_table_type = builder.add_type::<StTableRow>();
+    builder
+        .build_table(ST_TABLE_NAME, *st_table_type.as_ref().expect("should be ref"))
+        .with_type(TableType::System)
+        .with_primary_key(StTableFields::TableId)
+        .with_column_sequence(StTableFields::TableId, None)
```
mamcx, line 287: A variant `with_primary_key_auto` should exist that also calls the other steps, and is less verbose and error-prone...

crates/core/src/db/datastore/system_tables.rs
```diff
@@ -235,231 +248,264 @@ st_fields_enum!(enum StVarFields {
...
+        .with_index(
+            v9::RawIndexAlgorithm::BTree {
+                columns: col_list![StIndexFields::IndexId],
+            },
+            "accessor_name_doesnt_matter",
+            None,
+        )
+        .with_column_sequence(StIndexFields::IndexId, None)
+        .with_unique_constraint(StIndexFields::IndexId, None);
+
```
mamcx, line 324: They used to be comments like `// TODO: Unique constraint on index name?` that this pr made as it was resolved as `no`. Is this the case? If not, then we need to decide about what to do with names.

crates/core/src/db/datastore/system_tables.rs
```diff
@@ -235,231 +248,264 @@ st_fields_enum!(enum StVarFields {
...
+    let st_index_type = builder.add_type::<StIndexRow>();
+    builder
+        .build_table(ST_INDEX_NAME, *st_index_type.as_ref().expect("should be ref"))
+        .with_type(TableType::System)
+        .with_primary_key(StIndexFields::IndexId)
+        .with_index(
+            v9::RawIndexAlgorithm::BTree {
+                columns: col_list![StIndexFields::IndexId],
+            },
+            "accessor_name_doesnt_matter",
```
mamcx, line 319:
```suggestion
            "accessor_name_doesnt_matter",
            DEFAULT_ACCESOR_NAME
```

crates/core/src/db/update.rs
```diff
@@ -1,627 +1,204 @@
...
-    pub reason: TaintReason,
+/// Manually migrate a database.
+fn manual_migrate_database(
+    _stdb: &RelationalDB,
+    _tx: &mut MutTxId,
+    _plan: ManualMigratePlan,
+    _system_logger: &SystemLogger,
+    _existing_tables: Vec<Arc<TableSchema>>,
+) -> anyhow::Result<()> {
+    unimplemented!("Manual database migrations are not yet implemented")
```
kazimuth, line 50: TODO: need to check for a reducer marked `LifeCycle::Update` and call it here? Or should it be called after auto-migration?

crates/core/src/db/datastore/system_tables.rs
kazimuth: Ah, I'll just put them back in. We should decide that before 1.0 though.

crates/lib/src/relation.rs
kazimuth: I reworked the structure here a little and changed this to `uncombined_constraints`. See the doc comment on `relation.rs:combine_constraints`.

crates/core/src/db/datastore/traits.rs
```diff
@@ -419,7 +420,10 @@ pub trait MutTxDatastore: TxDatastore + MutTx {
     }
 
     // Indexes
-    fn create_index_mut_tx(&self, tx: &mut Self::MutTx, table_id: TableId, index: RawIndexDefV8) -> Result<IndexId>;
+
+    /// Create a new index with the given schema.
+    /// The index's ID field should be set to 0 (the autoinc sentinel value), but the table ID should not.
+    fn create_index_mut_tx(&self, tx: &mut Self::MutTx, index_schema: IndexSchema, is_unique: bool) -> Result<IndexId>;
```
cloutiertyler, line 426: Why is `is_unique` necessary here? Also should it be included in the `index_schema` or no?

kazimuth: See the comment added in `mut_tx.rs`. (It's a little hard to document these APIs because there's multiple layers with nearly-identical structures, not always clear where to put the comments.)

It is still necessary because the `table` crate still takes an `is_unique` parameter. We can eventually move the enforcement of unique constraints outside `table` and then it won't be necessary anymore.

It would be better to just look up the corresponding unique constraint in the system tables, but this might be being called in a transaction where the relevant unique constraint will be created later (in particular during `create_table`), so we need to hedge our bets and give the caller control.

kazimuth:
There's now a fix for the internal tests in private, so this is all ready.

crates/core/src/db/update.rs
kazimuth: We're dropping it (https://github.com/clockworklabs/SpacetimeDB/issues/1370), so this is fine.

crates/core/src/db/update.rs
```diff
@@ -1,627 +1,220 @@
 use super::datastore::locking_tx_datastore::MutTxId;
 use super::relational_db::RelationalDB;
 use crate::database_logger::SystemLogger;
-use crate::error::{DBError, TableError};
-use anyhow::Context;
-use core::{fmt, mem};
-use enum_as_inner::EnumAsInner;
-use itertools::Itertools;
-use similar::{Algorithm, TextDiff};
+use crate::execution_context::ExecutionContext;
```
kazimuth, line 4: This file was completely rewritten so should probably not be reviewed in diff view.

crates/core/src/vm.rs
```diff
@@ -541,39 +540,6 @@ impl<'db, 'tx> DbProgram<'db, 'tx> {
         }
     }
 
-    fn _create_table(&mut self, table: RawTableDefV8) -> Result<Code, ErrorVm> {
```
kazimuth, line 544: This code was dead and now out-of-date so I buried it.

crates/core/src/db/datastore/locking_tx_datastore/datastore.rs
```diff
@@ -1231,23 +1198,59 @@ mod tests {
...
+            ],
+            vec![SequenceSchema {
+                sequence_id: SequenceId::SENTINEL,
+                table_id: TableId::SENTINEL,
+                col_pos: 0.into(),
+                sequence_name: "id_sequence".into(),
+                start: 1,
+                increment: 1,
+                min_value: 1,
+                max_value: 170141183460469231731687303715884105727,
```
cloutiertyler, line 1246: NIT: Replace with `i128::MAX`

crates/core/src/db/datastore/system_tables.rs
```diff
@@ -235,231 +248,223 @@ st_fields_enum!(enum StVarFields {
...
+    //
+    // * `database_address` is the [`Address`] of the database.
+    // * `owner_identity` is the [`Identity`] of the owner of the database.
+    // * `program_kind` is the [`ModuleKind`] (currently always [`WASM_MODULE`]).
+    // * `program_hash` is the [`Hash`] of the raw bytes of the (compiled) module.
+    // * `program_bytes` are the raw bytes of the (compiled) module.
+    //
+    // | database_address | owner_identity |  program_kind | program_bytes | program_hash        |
+    // |------------------|----------------|---------------|---------------|---------------------|
+    // | <bytes>          | <bytes>        |  0            | <bytes>       | <bytes>             |
```
cloutiertyler, line 352: NIT: This diagram is missing `module_version`.

cloutiertyler (approved):
I scanned the whole PR. I left a few small nits. AFAICT, this looks really good. Very nice job here @kazimuth.

I am approving this PR from a `traits.rs` perspective and no major objection to the rest, but would probably be good to get another approval as well.

kim (approved):
Looks good.

Error propagation needs another pass.

crates/core/src/db/relational_db.rs
```diff
@@ -60,6 +61,9 @@ pub type DiskSizeFn = Arc<dyn Fn() -> io::Result<u64> + Send + Sync>;
 
 pub type Txdata = commitlog::payload::Txdata<ProductValue>;
 
+/// The only version modules can have at the moment.
+pub const ONLY_MODULE_VERSION: &str = "0.0.1";
```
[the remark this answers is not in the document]
kazimuth: This is a versioning field for the client module, I believe Lightfox requested this. Adding it now for backwards-compatibility.

crates/core/src/db/datastore/locking_tx_datastore/datastore.rs
kazimuth: I was just copying from the other cases.

crates/core/src/host/host_controller.rs
```diff
@@ -393,11 +387,10 @@ impl HostController {
             let program = load_program(&self.program_storage, program_hash).await?;
             let update_result = host
                 .update_module(host_type, program, self.unregister_fn(instance_id))
-                .await?;
+                .await;
```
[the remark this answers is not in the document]
kazimuth: Ah, I refactored `UpdateDatabaseResult` slightly but didn't need to do this here, will rewrite.

crates/core/src/db/relational_db.rs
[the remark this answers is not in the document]
kim: Can we have a docstring telling us that? :cat_with_heart_eyes:

crates/core/src/db/datastore/system_tables.rs
kazimuth: Fixed, and I moved these to doc comments on the `St...Row` types since they make sense there.

crates/core/src/host/wasm_common/module_host_actor.rs
```diff
@@ -352,30 +315,40 @@ impl<T: WasmInstance> ModuleInstance for WasmModuleInstance<T> {
...
-        let (tx, res) = stdb.with_auto_rollback(&ctx, tx, |tx| {
-            crate::db::update::update_database(stdb, tx, proposed_tables, self.system_logger())
-        })?;
+        let res = crate::db::update::update_database(stdb, &mut tx, plan, self.system_logger());
+
         match res {
             Err(e) => {
-                self.system_logger().warn("Database update failed");
+                log::warn!("Database update failed: {} @ {}", e, stdb.address());
+                self.system_logger().warn(&format!("Database update failed: {e}"));
```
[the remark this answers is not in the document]
kazimuth: Would be a good idea to just log everything that goes through the system loggers yeah. I'll make that a separate issue tho

PR #1699, opened 2024-09-11
WASM ABI: add `datastore_btree_scan_bsatn` & `index_id_from_name`

Centril:
# Description of Changes

Adds the ABIs:
- `index_id_from_name`
- `datastore_btree_scan_bsatn`

cc #1460 

## Reviewer notes:

- The `IndexIdMap` stuff is based on my discussion with @gefjon re. having in `TxState` & `CommittedState`.
- The `AlgebraicValue::{Min, Max}` variant additions are based on discussions with @joshua-spacetime  wrt. the need to have a min/max. These variants are  ephemeral and don't exist in the codec.
- Only consulting the `TxState` wrt. an index exists semantically or not is based on discussions with @joshua-spacetime .
- In the interest of time and implementing this quickly, this PR doesn't care a lot about making abstractions nice. So it wholesale ignores `traits.rs` and just pierces through all layers directly to `MutTxId`. To facilitate understanding however, I've focused on doc comments and comments for the added code.

None, this is new ABI.

3 - some complicated logic re. transactionality and how they interact with indices.

# TODO (to unblock folks, I suggest that these be follow ups)

- testing
- C# module bindings changes


crates/bindings-sys/src/lib.rs
```diff
@@ -714,6 +812,68 @@ pub fn datastore_table_scan_bsatn(table_id: TableId) -> Result<RowIter, Errno> {
...
+/// Note that the `prefix` in this case can be empty (`prefix_elems = 0`),
+/// in which case this becomes a ranged index scan on a single-col index
+/// or even a full table scan if `rstart` and `rend` are both unbounded.
+///
+/// The relevant table for the index is found implicitly via the `index_id`,
+/// which is unique for the module.
+///
+/// On success, the iterator handle is written to the `out` pointer.
+/// This handle can be advanced by [`row_iter_bsatn_advance`].
+///
```
gefjon, line 838:
```suggestion
///
/// # Non-obvious queries
///
/// For an index on columns `[a, b, c]`:
///
/// - `a = x, b = y` is encoded as a prefix `[x, y]`
///   and a range `Range::Unbounded`,
///   or as a  prefix `[x]` and a range `rstart = rend = Range::Inclusive(y)`.
/// - `a = x, b = y, c = z` is encoded as a prefix `[x, y]`
///   and a  range `rstart = rend = Range::Inclusive(z)`.
/// - A sorted full scan is encoded as an empty prefix
///   and a range `Range::Unbounded`.
///
```

gefjon (approved):

bfops:
Removing and re-adding to merge queue since we un-required the problematic check

bfops:
...nevermind, apparently we _haven't_ un-required the problematic check.

Issue #1504, opened 2024-07-12
Remove dependency on openssl


Centril:
It's being pulled in by reqwest, which comes as a default feature we're not using (see https://github.com/seanmonstar/reqwest/blob/master/Cargo.toml)

We're also using it in standalone to generate keys, which can be delegated to shell calls.

Removing this dependency removes a long chunk of compilation time on windows.

jdetter:
> Removing this dependency removes a long chunk of compilation time on windows.

Really it's hard to understate the amount of time this is taking in windows. I believe the vast majority of the time spent compiling SpacetimeDB on windows is just openssl.

mamcx:
For generating the keys we can use `ring`

PR #1700, opened 2024-09-11
Remove openssl

mamcx:
# Description of Changes

Using the `openssl` crate significantly increases the build times on Windows. Switching to `rustls` improves it:

```bash
# Windows 11
# 1.79.0-x86_64-pc-windows-msvc 
# AMD Ryzen 9 7940HS w/ Radeon 780M Graphics        4.00 GHz
#  32.0 GB (25.8 GB usable)

   Compiling spacetimedb-core v0.12.0 (C:\Proyectos\space\SpacetimeDB\crates\core)
   Compiling spacetimedb-client-api v0.12.0 (C:\Proyectos\space\SpacetimeDB\crates\client-api)
   Compiling spacetimedb-standalone v0.12.0 (C:\Proyectos\space\SpacetimeDB\crates\standalone)   

OpenSSL:

    Finished `dev` profile [unoptimized + debuginfo] target(s) in 4m 01s
    Finished `release` profile [optimized] target(s) in 4m 39s    

RusTls:

    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1m 21s
    Finished `release` profile [optimized] target(s) in 2m 02s    
```

3

It looks correct to me the generation and reading of keys.  However, even if trivial, changing libraries that do crypto requires careful eyes.

# Testing
- [x] * Check using https://8gwifi.org/PemParserFunctions.jsp that we use the same algorithms and settings as from `openssl`.*
- [x] *Publish `quickstart` and re-read the keys*
- [x] *Manual inspection of the keys and config files*


gefjon:
I strongly request that we not do potentially dangerous changes which are not intended to be API breaks within the 0.12 release window. This can, and therefore should, wait.

mamcx:
Changed to draft so we don't merge yet

mamcx:
Btw, this adds [`cmake`](https://aws.github.io/aws-lc-rs/requirements/windows0 as a requirement on Windows.

RReverser:
> Btw, this adds [`cmake`](https://aws.github.io/aws-lc-rs/requirements/windows0 as a requirement on Windows.

Sigh. It's somewhat unfortunate to replace one external dependency for Windows devs (Perl) with another (CMake). Why does rustls need it?

dare3path:
closes https://github.com/clockworklabs/SpacetimeDB/issues/1504

bfops:
`openssl` is a bunch of my build time on Ubuntu as well, so I was curious how this PR affects total build time (after `rm -rf target/*`).

master:
```
real	6m20.941s
user	7m28.604s
sys		1m14.316s
```
This branch:
```
real	4m48.625s
user	6m54.506s
sys		1m0.312s
```
So a meaningful improvement for me!

I don't really love that this swaps one outside-of-cargo dependency for another on Windows, but I can't speak to which one is more imposing for Windows users.

.github/workflows/ci.yml
```diff
@@ -146,6 +146,20 @@ jobs:
             cargo build -p v8
           fi
 
+      # This step shouldn't be needed, but somehow we end up with caches that are missing librusty_v8.a.
```
bfops, line 149: I added this since we now build in release mode here, and we started hitting this issue again.

I still have no idea why this happens since we no longer `cache-on-failure`, but.. /shrug.

PR #1707, opened 2024-09-12
c# client generate

lcodes:
Rust side generating module bindings for https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/pull/131


RReverser:
> @[lcodes](https://github.com/lcodes) lcodes requested a review from [RReverser](https://github.com/RReverser) [18 minutes ago](https://github.com/clockworklabs/SpacetimeDB/pull/1707#event-14276822043)

FWIW I'm going to review this one in more detail after the runtime counterpart.

crates/bindings-csharp/Codegen.Tests/fixtures/diag/snapshots/ExtraCompilationErrors.verified.txt
```diff
@@ -552,12 +552,60 @@ public readonly struct TestTableTaggedEnum : SpacetimeDB.Internal.ITableView<Tes
...
+      ]
     }
-    public IEnumerable<global::TestTableTaggedEnum> Iter() => SpacetimeDB.Internal.ITableView<TestTableTaggedEnum, global::TestTableTaggedEnum>.Iter();
+  },
+  {/*
+    public ulong Count => SpacetimeDB.Internal.ITableView<TestTableTaggedEnum, global::TestTableTaggedEnum>.DoCount();
+    public IEnumerable<global::TestTableTaggedEnum> Iter() => SpacetimeDB.Internal.ITableView<TestTableTaggedEnum, global::TestTableTaggedEnum>.DoIter();
                                                                                                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^
-    public global::TestTableTaggedEnum Insert(global::TestTableTaggedEnum row) => SpacetimeDB.Internal.ITableView<TestTableTaggedEnum, global::TestTableTaggedEnum>.Insert(row);
+    public IEnumerable<global::TestTableTaggedEnum> Query(System.Linq.Expressions.Expression<Func<global::TestTableTaggedEnum, bool>> predicate) => SpacetimeDB.Internal.ITableView<TestTableTaggedEnum, global::TestTableTaggedEnum>.DoQuery(predicate);
```
RReverser, line 606: FWIW we removed the server-side `Query` API, so it's not necessary on client-side either anymore. Feel free to either make it private, or remove altogether, whichever is easier.

crates/bindings-csharp/Runtime/Attrs.cs
```diff
@@ -51,6 +51,17 @@ public sealed class TableAttribute : Attribute
...
     }
 
+    public static class Index {
+        [AttributeUsage(AttributeTargets.Struct | AttributeTargets.Class, AllowMultiple = true)]
+        public sealed class BTreeAttribute : Attribute {
+            public string? Table { get; init; }
+
+            public string? Name { get; init; }
+
+            public string[] Columns { get; init; } = [];
```
RReverser, line 61:
```suggestion
            public required string[] Columns { get; init; };
```

IIUC, index without columns is never useful, so let's use the neat `required` feature to make sure that user always specifies columns.

crates/bindings-csharp/Runtime/Internal/Bounds.cs
```diff
@@ -0,0 +1,305 @@
...
+    where URW : struct, IReadWrite<U>
+    where VRW : struct, IReadWrite<V>
+    where WRW : struct, IReadWrite<W>
+    where XRW : struct, IReadWrite<X>
+    where YRW : struct, IReadWrite<Y>
+    where ZRW : struct, IReadWrite<Z>
+    where ARW : struct, IReadWrite<A>
+    where BRW : struct, IReadWrite<B>
+    where CRW : struct, IReadWrite<C> {
+    public ushort PrefixElems => 9;
```
RReverser, line 282: Oh my god. This is getting out of hand 😅

I suppose we can fix it later though.

crates/bindings-csharp/Runtime/Internal/FFI.cs
```diff
@@ -165,6 +177,26 @@ ref uint buffer_len
...
+    public static partial CheckedStatus datastore_insert_bsatn(
+        TableId table_id,
+        Span<byte> row,
+        ref uint row_len
+    );
+
+    [LibraryImport(StdbNamespace)]
+    public static unsafe partial CheckedStatus datastore_delete_by_btree_scan_bsatn(
+        IndexId index_id,
+        byte* prefix,
```
RReverser, line 190: Use `byte[]` or `Span` or `ReadonlySpan` for safety. Then you won't need those manual `fixed (...)` wrappers on the consumption side.

crates/bindings-csharp/Runtime/Internal/FFI.cs
```diff
@@ -174,13 +206,28 @@ out uint out_
     );
 
     [LibraryImport(StdbNamespace)]
-    public static partial void volatile_nonatomic_schedule_immediate(
-        [In] byte[] name,
-        uint name_len,
-        [In] byte[] args,
-        uint args_len
+    public static partial Errno bytes_source_read(
```
RReverser, line 209: Why were those moved around? It makes diff harder to read.

crates/bindings-csharp/Runtime/Internal/IIndex.cs
```diff
@@ -0,0 +1,106 @@
...
+        w.Write(0); bsatn.Write(w, key); // inclusive rstart
+        idx = (uint)s.Length;
+        w.Write(0); bsatn.Write(w, key); // inclusive rend
+        end = (uint)s.Length;
+        return s.ToArray();
+    }
+
+    private class RawTableIter(T key, RW bsatn) : ITable<Row>.RawTableIterBase {
+        protected override unsafe void IterStart(out FFI.RowIter handle) {
+            fixed (byte* bytes = Params(key, bsatn, out var idx, out var end)) {
```
RReverser, line 36: See above, we should never have to write those unsafe wrappers ourselves. Marshalling takes care of it as long as we specify correct types.

crates/bindings-csharp/Runtime/Internal/IIndex.cs
```diff
@@ -0,0 +1,106 @@
...
+
+public interface IUniqueIndex<View, Row, T, RW> : IIndex<View>
+    where View : IUniqueIndex<View, Row, T, RW>
+    where Row : ITable<Row>, new()
+    where RW : BSATN.IReadWrite<T>
+{
+    private static byte[] Params(T key, RW bsatn, out uint idx, out uint end) {
+        using var s = new MemoryStream();
+        using var w = new BinaryWriter(s);
+        w.Write(0); bsatn.Write(w, key); // inclusive rstart
```
RReverser, line 27: These are kinda dangerous, I don't like relying on inference to ensure that `Write(0)` will result in the correct integer type.

Can we use a `const` with explicit type? That would also help readability to clarify what this zero represents.

crates/bindings-csharp/Runtime/Internal/FFI.cs
lcodes: There's no way to offset the buffer otherwise; this mirrors the rust side doing one allocation for the prefix/rstart/rend and then indexing into it.

crates/bindings-csharp/Runtime/Internal/FFI.cs
lcodes: Ah my bad, I made them in the same order as the wasm proposal to make it easier to see if they were all present

crates/bindings-csharp/Runtime/Internal/FFI.cs
RReverser: > no way to offset the buffer otherwise

You should be able to do that on Span with simple range indexing.

crates/bindings-csharp/Codegen/Module.cs
```diff
@@ -716,6 +716,7 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
                     $$"""
                     // <auto-generated />
                     #nullable enable
+                    #pragma warning disable CS8981
```
RReverser, line 719: Do we really need to do this in _all_ generated code? I'd say let's disable it in our own test projects, user projects usually shouldn't need it.

crates/bindings-csharp/Runtime/Internal/Bounds.cs
lcodes: yeah hehe

crates/bindings-csharp/Codegen/Module.cs
lcodes: Ah there could be points made for both directions;

ie reducers being callable from the cmdline there's a point to name them in snake_case, and it fits the naming convention of tables too.

But you're also right it opens the door for non-idiomatic C# names to make it through. I'm fine with either solutions.

RReverser:
@lcodes `AddPlayerArgsStruct` and such are different from naming in Phoebe's and your own proposal (https://github.com/clockworklabs/SpacetimeDBPrivate/pull/949).

I always wanted to remove that `ArgsStruct` suffix too, but was punting off as it would be a breaking change, but now we have a perfect excuse for breaking changes - it's in the spec!

RReverser:
In a scenario like this:

```cs
internal static partial class Outer
{
    [SpacetimeDB.Table]
    public partial struct InnerTable
    {
        public string Name;
    }
}
```

we will currently generate an ITableView implementation with `public` visibility, which will result in

```
   Assert.Empty() Failure: Collection was not empty
Collection: [SpacetimeDB.Codegen\SpacetimeDB.Codegen.Module\FFI.cs(142,50): error CS0050: Inconsistent accessibility: return type 'IEnumerable<Outer.InnerTable>' is less accessible than method 'InnerTable.Iter()', SpacetimeDB.Codegen\SpacetimeDB.Codegen.Module\FFI.cs(143,37): error CS0050: Inconsistent accessibility: return type 'Outer.InnerTable' is less accessible than method 'InnerTable.Insert(Outer.InnerTable)', SpacetimeDB.Codegen\SpacetimeDB.Codegen.Module\FFI.cs(143,37): error CS0051: Inconsistent accessibility: parameter type 'Outer.InnerTable' is less accessible than method 'InnerTable.Insert(Outer.InnerTable)', SpacetimeDB.Codegen\SpacetimeDB.Codegen.Module\FFI.cs(144,17): error CS0051: Inconsistent accessibility: parameter type 'Outer.InnerTable' is less accessible than method 'InnerTable.Delete(Outer.InnerTable)']
  Stack Trace:
     at SpacetimeDB.Codegen.Tests.GeneratorSnapshotTests.TypeAndModuleGeneratorsOnServer() in D:\SpacetimeDB\crates\bindings-csharp\Codegen.Tests\Tests.cs:line 152
--- End of stack trace from previous location ---
```

We should instead generate it as an `internal`. I think in the visibility loop you need to take the minimum of any visibility you've seen while traversing up.

RReverser:
Also mentioned this before, but right now `Module.verified.txt` results in couple of `An internal error occurred during codegen: ...` due to usage of exceptions for columns and visibility checks.

Could you please convert those to 2 new diagnostics with location context attached?

crates/bindings-csharp/Runtime/Internal/ITable.cs
```diff
@@ -108,29 +100,27 @@ public IEnumerable<T> Parse()
...
-        protected override void IterStart(out FFI.RowIter handle)
-        {
-            _ = (tableId, colId, value);
-            // Needs to be implemented via datastore_table_scan_bsatn
-            throw new NotImplementedException();
-        }
-    }
-
     // Note: this must be Lazy to ensure that we don't try to get the tableId during startup, before the module is initialized.
     private static readonly Lazy<FFI.TableId> tableId_ =
```
RReverser, line 125: These no longer need to be Lazy. As the comment says, it was necessary only because we used static tables. Converting them to regular properties should make code a bit simpler.

crates/bindings-csharp/Runtime/Internal/FFI.cs
```diff
@@ -229,4 +271,20 @@ internal static class ConsoleTimerIdMarshaller
 
     [LibraryImport(StdbNamespace)]
     public static partial CheckedStatus console_timer_end(ConsoleTimerId stopwatch_id);
+
+    [LibraryImport(StdbNamespace)]
+    public static partial CheckedStatus iter_start_filtered(
```
RReverser, line 276: Looks like a rebase issue. This ABI was removed and shouldn't be used anymore.

crates/bindings-csharp/Runtime/Internal/FFI.cs
```diff
@@ -131,10 +146,25 @@ out RowIter out_
...
+        ColId prefix_elems,
+        ReadOnlySpan<byte> rstart,
+        uint rstart_len,
+        ReadOnlySpan<byte> rend,
+        uint rend_len,
+        out RowIter out_
+    );
+
+    [LibraryImport(StdbNamespace)]
+    public static partial CheckedStatus iter_by_col_eq(
```
RReverser, line 162: This ABI was removed as well and shouldn't be used anymore.

crates/bindings-csharp/Runtime/bindings.c
```diff
@@ -36,44 +37,63 @@ OPAQUE_TYPEDEF(ConsoleTimerId, uint32_t);
...
+        const uint8_t* rstart, uint32_t rstart_len, const uint8_t* rend, uint32_t rend_len, RowIter* iter),
+       (index_id, prefix, prefix_len, prefix_elems, rstart, rstart_len, rend, rend_len, iter));
+IMPORT(Status, iter_by_col_eq,
+       (TableId table_id, ColId col_id, const uint8_t* value,
+        uint32_t value_len, RowIter* iter),
+       (table_id, col_id, value, value_len, iter));
+IMPORT(Status, iter_start_filtered,
+       (TableId table_id, const uint8_t* filter, uint32_t filter_len,
+        RowIter* iter),
+       (table_id, filter, filter_len, iter));
```
RReverser, line 63: Also seems to accidentally re-add removed ABIs.

In general, couldn't you just revert this file (almost) entirely? Looks like all the necessary functions were present already, just shuffled around.

crates/bindings-csharp/Runtime/Internal/FFI.cs
RReverser: Same as with `bindings.c`, maybe worth reverting the reordering here to make it obvious what's actually added and what is an accidental regression.

modules/sdk-test-multi-cs/sdk-test-multi-cs.csproj
```diff
@@ -0,0 +1,22 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
RReverser, line 1: Looks like a rebase artifact, this file should be removed. We don't have any sources in this folder since the multi-column test was moved to Codegen.Tests.

crates/bindings-csharp/Codegen.Tests/fixtures/diag/snapshots/Module#FFI.verified.cs
```diff
@@ -54,63 +55,37 @@ public readonly struct TestAutoIncNotInteger
...
-                    .Update(@this);
+                >.DoInsert(row);
+
+            public bool Delete(global::TestAutoIncNotInteger row) =>
+                SpacetimeDB.Internal.ITableView<
+                    TestAutoIncNotInteger,
+                    global::TestAutoIncNotInteger
+                >.DoDelete(row);
+
+            public UniqueIndex<
```
RReverser, line 82: Nit for the future: we should try and generate table views in separate files.

FFI.cs should be the last resort, just for module definitions and such, where items truly need to be combined together.

Otherwise it makes caching that much worse when we need to recompute and regenerate entire FFI.cs whenever e.g. just one index in one table has changed.

crates/bindings-csharp/Codegen.Tests/fixtures/server/snapshots/Module#MultiTableRow.verified.cs
```diff
@@ -92,18 +94,7 @@ SpacetimeDB.BSATN.ITypeRegistrar registrar
                     Indexes: [],
                     Constraints:
                     [
-                        new(
```
RReverser, line 95: @lcodes This looks like a notable regression. Shouldn't `Foo` still be autoinc in MultiTable2?

crates/bindings-csharp/Codegen.Tests/fixtures/diag/snapshots/Module#TestAutoIncNotInteger.verified.cs
```diff
@@ -59,16 +59,9 @@ SpacetimeDB.BSATN.ITypeRegistrar registrar
                     Constraints:
                     [
                         new(
-                            nameof(SpacetimeDB.Local.TestAutoIncNotInteger),
```
RReverser, line 62: Same here, AutoInc just disappeared for some reason.

crates/bindings-csharp/Codegen/Module.cs
```diff
@@ -716,7 +717,7 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
                     $$"""
                     // <auto-generated />
                     #nullable enable
-                    #pragma warning disable CS8981
+                    #pragma warning disable
```
RReverser: Why do we need this? I'd rather know if generated code is problematic.

jdetter (approved):
I have not done a code review on this but I've tested it against circle game and it works.

PR #1708, opened 2024-09-13
feat: sql type checking and logical plan

joshua-spacetime:
# Description of Changes

Defines the logical query plan and implements lowering from ast to logical plan which includes variable binding and type checking. DML is still left to do.

None

3

### DIFF
--- a/crates/lib/src/lib.rs
+++ b/crates/lib/src/lib.rs
@@ -379,6 +379,12 @@ pub fn from_hex_pad<R: hex::FromHex<Error = hex::FromHexError>, T: AsRef<[u8]>>(
     hex: T,
 ) -> Result<R, hex::FromHexError> {
     let hex = hex.as_ref();
-    let hex = if hex.starts_with(b"0x") { &hex[2..] } else { hex };
+    let hex = if hex.starts_with(b"0x") {
+        &hex[2..]
+    } else if hex.starts_with(b"X'") {
+        &hex[2..hex.len()]
+    } else {
+        hex
+    };
     hex::FromHex::from_hex(hex)
 }
--- a/crates/planner/src/logical/bind.rs
+++ b/crates/planner/src/logical/bind.rs
@@ -1,0 +1,414 @@
+use std::sync::Arc;
+
+use spacetimedb_lib::{from_hex_pad, Address, AlgebraicValue, Identity};
+use spacetimedb_sats::{AlgebraicType, ProductType, ProductTypeElement};
+use spacetimedb_schema::schema::TableSchema;
+use spacetimedb_sql_parser::{
+    ast::{
+        self,
+        sub::{SqlAst, SqlSelect},
+        BinOp, ProjectElem, SqlExpr, SqlFrom, SqlLiteral,
+    },
+    parser::sub::parse_subscription,
+};
+
+use super::errors::{ConstraintViolation, ResolutionError, TypingError, Unsupported};
+use super::expr::{Expr, Ref, RelExpr, Type, Vars};
+
+pub type TypingResult<T> = core::result::Result<T, TypingError>;
+
+pub trait SchemaView {
+    fn schema(&self, name: &str, case_sensitive: bool) -> Option<Arc<TableSchema>>;
+}
+
+/// Parse and type check a subscription query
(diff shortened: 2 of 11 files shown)


crates/lib/src/lib.rs
```diff
@@ -379,6 +379,12 @@ pub fn from_hex_pad<R: hex::FromHex<Error = hex::FromHexError>, T: AsRef<[u8]>>(
     hex: T,
 ) -> Result<R, hex::FromHexError> {
     let hex = hex.as_ref();
-    let hex = if hex.starts_with(b"0x") { &hex[2..] } else { hex };
+    let hex = if hex.starts_with(b"0x") {
+        &hex[2..]
+    } else if hex.starts_with(b"X'") {
+        &hex[2..hex.len()]
```
mamcx, line 385: The test `compile_eq_identity_address` also check for `x`

crates/planner/src/logical/bind.rs
```diff
@@ -0,0 +1,414 @@
...
+/// Type check and lower a [SqlAst] into a [RelExpr].
+/// This includes name resolution and variable binding.
+pub fn type_ast(expr: SqlAst, tx: &impl SchemaView) -> TypingResult<RelExpr> {
+    match expr {
+        SqlAst::Union(a, b) => {
+            let a = type_ast(*a, tx)?;
+            let b = type_ast(*b, tx)?;
+            assert_eq_types(a.ty(), b.ty())?;
+            Ok(RelExpr::Union(Box::new(a), Box::new(b)))
+        }
```
mamcx, line 38: Remember to add a comment or check here about checking the schemas are compatible with the union/minus projection

crates/planner/src/logical/bind.rs
```diff
@@ -0,0 +1,414 @@
...
+fn expect_op_type(op: BinOp, expr: Expr) -> TypingResult<Expr> {
+    match (op, expr.ty()) {
+        // Logic operators take booleans
+        (BinOp::And | BinOp::Or, Type::Alg(AlgebraicType::Bool)) => Ok(expr),
+        // Comparison operators take integers or floats
+        (BinOp::Lt | BinOp::Gt | BinOp::Lte | BinOp::Gte, Type::Alg(t)) if t.is_integer() || t.is_float() => Ok(expr),
+        // Equality supports numerics, strings, and bytes
+        (BinOp::Eq | BinOp::Ne, Type::Alg(t))
+            if t.is_bool()
+                || t.is_integer()
```
mamcx, line 285: I think we should add a trait similar to `eq, ord` so we instead do `if t.is_comparable`

crates/planner/src/logical/bind.rs
```diff
@@ -0,0 +1,414 @@
...
+            "select * from s where s.bytes = 0xABCD",
+            "select * from s where s.bytes = X'ABCD'",
+            "select * from s as r where r.bytes = 0xABCD",
+            "select * from (select t.* from t join s)",
+            "select * from (select t.* from t join s on t.u32 = s.u32 where t.f32 = 0.1)",
+            "select * from (select t.* from t join (select s.u32 from s) s on t.u32 = s.u32)",
+        ] {
+            let result = parse_and_type_sub(sql, &tx).inspect_err(|_| {
+                // println!("sql: {}\n\n\terr: {}\n", sql, err);
+            });
```
mamcx, line 371: I think this way of testing could blindly let pass an unrelated error and mask something is broken.
 
If the idea is to make checking for the error easier without heavy matching, how about adding a simple enum `Error{TypeCheck, FieldNotFound}` that we return for the error?

crates/planner/src/logical/expr.rs
```diff
@@ -0,0 +1,317 @@
...
+    /// Find an in scope table variable and field.
+    /// Return its type, variable and field positions if found.
+    /// Return an error otherwise.
+    pub fn expect_field(
+        &self,
+        table: &str,
+        field: &str,
+        expected: Option<&Type>,
+    ) -> TypingResult<(usize, usize, &AlgebraicType)> {
+        self.find(table)
```
mamcx, line 207: I think is better to make a struct that describes what are the `usize, usize` here...

mamcx (approved):
I approve after agree with @joshua-spacetime that some changes are for the next step

PR #1711, opened 2024-09-16
Filter out deps that only is added to keccack on arm that broke local testing

mamcx:
# Description of Changes

This is to let the developers that run test on  `arm` (like `Apple M1`) pass the `deptree_snapshot` check.

1
# Testing

This must be checked on `arm`. In other targets is as before.


mamcx:
An alternative is to only run this test on `linux`.

cloutiertyler (approved):
This LGTM and helps me run my tests on mac

PR #1712, opened 2024-09-16
Fix heisenbug by using petgraph

kazimuth:
# Description of Changes

I added a nondeterministic test failure, whoops! That algorithm I was using was misremembered and completely wrong.

None

0

# Testing

Tests pass always now.

### DIFF
--- a/crates/schema/src/type_for_generate.rs
+++ b/crates/schema/src/type_for_generate.rs
@@ -188,7 +188,7 @@ impl AlgebraicTypeDef {
                 }
                 AlgebraicTypeDef::PlainEnum(_) => {}
             }
-            buf.iter().cloned().collect()
+            buf.drain().collect()
         })
     }
 
--- a/crates/schema/src/type_for_generate.rs
+++ b/crates/schema/src/type_for_generate.rs
@@ -1,6 +1,10 @@
 //! `AlgebraicType` extensions for generating client code.
 
 use enum_as_inner::EnumAsInner;
+use petgraph::{
+    algo::tarjan_scc,
+    visit::{GraphBase, IntoNeighbors, IntoNodeIdentifiers, NodeIndexable},
+};
 use smallvec::SmallVec;
 use spacetimedb_data_structures::{
     error_stream::{CollectAllErrors, CombineErrors, ErrorStream},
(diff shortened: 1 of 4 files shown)


RReverser (approved):

crates/schema/src/type_for_generate.rs
```diff
@@ -161,21 +170,26 @@ impl AlgebraicTypeDef {
...
+                AlgebraicTypeDef::Sum(SumTypeDef { variants, .. }) => {
+                    for (_, use_) in variants.iter() {
+                        use_.extract_refs(buf);
+                    }
                 }
+                AlgebraicTypeDef::PlainEnum(_) => {}
             }
-            AlgebraicTypeDef::PlainEnum(_) => {}
-        }
+            buf.iter().cloned().collect()
```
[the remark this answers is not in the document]
kazimuth: Oh true

crates/schema/src/type_for_generate.rs
```diff
@@ -151,6 +155,11 @@ pub enum AlgebraicTypeDef {
     PlainEnum(PlainEnumTypeDef),
 }
 
+thread_local! {
+    /// Used to efficiently extract refs from a def.
+    static EXTRACT_REFS_BUF: RefCell<HashSet<AlgebraicTypeRef>> = RefCell::new(HashSet::new());
```
RReverser, line 160: Very nitpicky, but I tend to prefer to move such static / thread-local vars into the function itself, to make it more obvious it's not used anywhere else.

crates/schema/src/type_for_generate.rs
```diff
@@ -161,21 +170,26 @@ impl AlgebraicTypeDef {
...
-    fn extract_refs(&self, buf: &mut HashSet<AlgebraicTypeRef>) {
-        match self {
-            AlgebraicTypeDef::Product(ProductTypeDef { elements, .. }) => {
-                for (_, ty) in elements.iter() {
-                    ty.extract_refs(buf);
+    /// Extract all `AlgebraicTypeRef`s that are used in this type into a buffer.
+    /// The buffer may be in arbitrary order, but will not contain duplicates.
+    fn extract_refs(&self) -> SmallVec<[AlgebraicTypeRef; 16]> {
+        EXTRACT_REFS_BUF.with_borrow_mut(|buf| {
+            buf.clear();
```
RReverser, line 177: This shouldn't be necessary anymore (not immediately obvious due to the other comment, var being outside the function, but seems to be the case?).

PR #1714, opened 2024-09-17
Logging API

SteveBoytsun:
# Description of Changes

Changed logging based on [this proposal](https://github.com/clockworklabs/SpacetimeDBPrivate/pull/981)

corresponding GitHub label.

1

# Testing

- [x] You can test it by calling the following reducer
 ```cs
[Reducer]
public static void testt(ReducerContext ctx)
{
    Log.Debug("1");
    Log.Trace("2");
    Log.Info("3");
    Log.Warn("4");
    Log.Error("5");
    Log.Exception("6");
}
```
Note: STDB doesn't differentiate between error and exception, so the last two logs are both errors


RReverser:
We have quite a few modules and templates using `Runtime.Log`, including in runtime tests. You're going to want to update all of those too (just search the entire codebase).

SteveBoytsun:
> We have quite a few modules and templates using `Runtime.Log`, including in runtime tests. You're going to want to update all of those too (just search the entire codebase).

Good catch, fixed (not as many as I thought there would be)

RReverser:
> Good catch, fixed (not as many as I thought there would be)

CI still complains about a missed template for `spacetime generate` used in smoketests. You probably used "find references" instead of textual search.

SteveBoytsun:
> CI still complains about a missed template for `spacetime generate` used in smoketests. You probably used "find references" instead of textual search.

I did do textual search. It actually seems to be the opposite problem - smoketests are using old version of SDK for some reason. I'm not sure how to fix it, let's discuss in today's meeting
>  /tmp/tmp4jexmxj5/Lib.cs(19,9): error CS0103: The name 'Log' does not exist in the current context [/tmp/tmp4jexmxj5/StdbModule.csproj]

RReverser:
Ahh ok, sorry. Yeah it checks out C# SDK here https://github.com/clockworklabs/SpacetimeDB/blob/3fcb5056b205f69c84e9e6888264fad28032a182/.github/workflows/ci.yml#L117-L119 - it points to master by default, but for your usecase it would need to be changed to point to your branch.

Cross-repo testing is annoying.

RReverser (approved):
Looks good, thanks. The CI issue is not something we can fix at the moment, but as long as you tested them locally, it should be fine.

cloutiertyler:
I force merged this because of the API breaking changes at Steve's and Ingvar's request

joshua-spacetime:
@RReverser is there a PR that fixes the failing smoketest?

PR #1717, opened 2024-09-18
[C#] NFC: split out IStructuralWrite interface

RReverser:
Structural read implementation can be only directly implemented on types with default constructors, but writing should not have such limitations.

In this PR I'm splitting out writing into its own interface, so that it can be implemented even on types like tagged enums, which simplifies their own `BSATN.Write` implementation and allows to use them with helpers like `ToBytes`.

I'm also making the generated tagged enum record `abstract` - which it, arguably, should've been from the beginning - but it should be a non-functional change because its constructor has always been `private` anyway.

# Description of Changes

corresponding GitHub label.

# Testing


cloutiertyler:
I'm closing this as stale, but I've passed it to @lisandroct to investigate if it's worth doing.

PR #1718, opened 2024-09-18
[C#] [NFC] Reuse attribute classes from runtime

RReverser:
Currently, we have to copy-paste and hardcode property names and types from the runtime in the codegen, which is pretty error-prone, and is getting harder as our API is going to rapidly expand.

In this PR I'm implementing a basic helper that allows to reuse original attribute classes from the runtime in the codegen by taking the syntax data and reconstructing the same attribute instance via reflection.

This ensures that we have all the same typechecked properties on both sides instead of having to extract them manually by name + explicit casts.

# Description of Changes

corresponding GitHub label.

# Testing

### DIFF
--- a/crates/bindings-csharp/Codegen/Codegen.csproj
+++ b/crates/bindings-csharp/Codegen/Codegen.csproj
@@ -29,6 +29,16 @@
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.3.1" PrivateAssets="all" />
   </ItemGroup>
 
+  <ItemGroup>
+    <!-- dev-dependency to add internal C# classes and attributes not included in .NET Standard -->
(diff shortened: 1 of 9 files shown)


lcodes (approved):

crates/bindings-csharp/Codegen/Codegen.csproj
```diff
@@ -29,6 +29,16 @@
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.3.1" PrivateAssets="all" />
   </ItemGroup>
 
+  <ItemGroup>
+    <!-- dev-dependency to add internal C# classes and attributes not included in .NET Standard -->
+    <PackageReference Include="PolySharp" Version="1.14.1" PrivateAssets="all" />
```
lcodes, line 34: Is this needed inside codegen? It should normally run on a recent enough version of dotnet no?

PR #1719, opened 2024-09-18
Jeremie/cs module

lcodes:
Updates runtime and codegen to support multiple tables per row type


RReverser:
Just to duplicate a note from DMs - I noticed a few things in the PR that would be caught by the codegen tests, and looks like the snapshots haven't been updated in the PR. Can you please run `...\bindings-csharp> dotnet test` and fixup the issues it finds and commit the updated snapshots? Otherwise I'm afraid I might be catching things in review that those tests would already catch as well.

Also #1718 might help with some attribute parsing here.

crates/bindings-csharp/Runtime/Runtime.cs
```diff
@@ -77,6 +59,14 @@ public AlgebraicType GetAlgebraicType(ITypeRegistrar registrar) =>
     }
 }
 
+public abstract class BaseReducerContext<DbView> : DbContext<DbView>, IReducerContext
+    where DbView : class, new()
```
RReverser, line 63: Unless we have an interface to uphold, I'd rather not spell out extra constraints. In this case it doesn't seem to add much value in terms of checks.

```suggestion
    where DbView : new()
```

crates/bindings-csharp/Runtime/Runtime.cs
```diff
@@ -77,6 +59,14 @@ public AlgebraicType GetAlgebraicType(ITypeRegistrar registrar) =>
...
+{
+    public Identity Sender => Runtime.SenderIdentity!;
```
RReverser, line 65: Why are these set on the singleton now? What happens if someone wants to store a `ReducerContext`? It would now lose its own identity and point to the one from the currently running reducer instead.

I think those should be own fields, like before.

crates/bindings-csharp/Runtime/Attrs.cs
```diff
@@ -40,8 +22,65 @@ public enum ColumnAttrs : byte
...
+    public const string Init = "__init__";
+    public const string Update = "__update__";
+    public const string Connect = "__identity_connected__";
+    public const string Disconnect = "__identity_disconnected__";
+}
+
+[AttributeUsage(AttributeTargets.Method, Inherited = false)]
+public sealed class ReducerAttribute : Attribute
+{
+    public string? Name;
```
RReverser, line 83: Reducer kind being a special name was more of an implementation detail, and that will be changing in the new proposal (WIP in https://github.com/clockworklabs/SpacetimeDB/pull/1670) where ReducerKind is stored separately.

AFAIK the SDK proposal also doesn't allow custom `#[spacetimedb::reducer(name = "...")]`, so we shouldn't be doing that in C# either - can you please revert this to a regular attribute parameter, so it would be easier to migrate to non-name-based ReducerKind in the future.

crates/bindings-csharp/Runtime/Runtime.cs
lcodes: I made it a class because it does carry state; the connection handle in this case. A struct could be copied/defaulted and introduce a silent null connection. This forces it to be a reference type, and given there's only one per application the distinction between value/ref is no longer relevant, I think.

crates/bindings-csharp/Codegen/Module.cs
```diff
@@ -129,14 +162,21 @@ public TableDeclaration(GeneratorAttributeSyntaxContext context)
             throw new InvalidOperationException("Tagged enums cannot be tables.");
         }
 
-        var attrArgs = context.Attributes.Single().NamedArguments;
+        Visibility = context.TargetSymbol.DeclaredAccessibility switch {
```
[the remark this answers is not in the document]
lcodes: It's in the proposal that the generated types match the visibility of their source ones.

So if a table is internal, its handle view should also be internal, and so on.

crates/bindings-csharp/Runtime/Runtime.cs
lcodes: Well in this case Local is stateless, but the same base class has to support DbConnection on the client side.

crates/bindings-csharp/Runtime/Runtime.cs
lcodes: Reducer context is created once for the duration of the module, and @gefjon says each reducer invocation should be seen as its own isolated instance with no memory able to outlive that invocation.

Making them fields would need a new ReducerContext per reducer invocation, which means its DbView instance has to either also be created per invocation, or stored somewhere.

lcodes: I'm fine with either approaches; I guess it depends on what semantics we want for memory across reducer invocations, and if we want to leverage isolation in the impl like this ReducerContext does here.

RReverser: > Making them fields would need a new ReducerContext per reducer invocation

Yeah, that's what I'm suggesting and what Rust is doing as well. Even if users shouldn't rely on memory sharing, we still should avoid singletons to prevent accidentally leaking data between different invocations that might belong to different identities.

crates/bindings-csharp/Codegen/Module.cs
```diff
@@ -19,7 +19,7 @@ enum ColumnAttrs : byte
 
 record ColumnDeclaration : MemberDeclaration
 {
-    public readonly ColumnAttrs Attrs;
+    public readonly EquatableArray<(string?, ColumnAttrs)> Attrs;
```
RReverser, line 22: Can you use named tuple fields, or better yet, extract this to a `record struct`? All the `.Item1` / `.Item2` tend to be not very clear when reading.

crates/bindings-csharp/Codegen/Module.cs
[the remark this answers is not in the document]
RReverser: I see, makes sense. I'd suggest to just store the `Accessibility` in the struct - it's just a plain enum, so it's fine to do so - and use `SyntaxFacts.GetText()` during output stringification.

crates/bindings-csharp/Codegen/Module.cs
```diff
@@ -129,17 +160,31 @@ public TableDeclaration(GeneratorAttributeSyntaxContext context)
...
+        };
 
-        IsPublic = attrArgs.Any(pair => pair is { Key: "Public", Value.Value: true });
+        Views = new(context.Attributes
+            .Where(a => a.AttributeClass?.ToDisplayString() == "SpacetimeDB.TableAttribute")
+            .Select(a => new TableView(this, a))
+            .ToImmutableArray());
+
+        var schedules = Views.Where(t => t.Scheduled != null).Select(t => t.Scheduled);
+        var numSchedules = schedules.Count();
```
RReverser, line 179: This will cause the iterable to be fully walked over twice.

Looks like the logic could be simplified if you just kept the nullable `t.Scheduled` before doing `.Distinct()` - it would catch both `null` as well as some non-null values as individual items, so you'd only need to do one check instead of several.

crates/bindings-csharp/Runtime/Attrs.cs
```diff
@@ -1,28 +1,10 @@
 ﻿namespace SpacetimeDB;
 
-public static class ReducerKind
+[Type]
+public enum IndexType : byte
```
RReverser, line 4: It doesn't look like this is used anywhere - a leftover?

modules/sdk-test-multi-cs/sdk-test-multi-cs.csproj
```diff
@@ -1,6 +1,8 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
+    <TargetName>StdbModule</TargetName>
+    <EmitCompilerGeneratedFiles>true</EmitCompilerGeneratedFiles>
```
RReverser, line 5: Let's remove these? They're useful for local debugging, but not necessary for e.g. CI. For checking the output, we have Codegen.Tests anyway these days.

crates/bindings-csharp/Codegen.Tests/fixtures/server/Lib.cs
```diff
@@ -39,12 +39,13 @@ public partial record CustomTaggedEnum
     : SpacetimeDB.TaggedEnum<(int IntVariant, string StringVariant)>;
 
 [SpacetimeDB.Table]
-public partial class PrivateTable { }
+public partial struct PrivateTable { }
```
RReverser, line 42: Is this change necessary? It had a secondary purpose of checking that `class` tables work as well. Admittedly, that's not very clear, so if you want to instead add a class-based table separately, that would be fine too.

crates/bindings-csharp/Codegen/Module.cs
```diff
@@ -310,38 +363,36 @@ public ReducerDeclaration(GeneratorAttributeSyntaxContext context)
         {
             throw new Exception($"Reducer {method} must return void");
         }
+        if (method.Parameters.FirstOrDefault()?.Type is not INamedTypeSymbol namedType || namedType.Name != "ReducerContext") {
```
RReverser, line 366: Please check the full type name via `.ToString()` or `.FullName`, to make sure it's _our_ `ReducerContext`, just in case.

crates/bindings-csharp/Runtime/Runtime.cs
RReverser: It appears we could also merge `BaseReducerContext` and `DbContext` into one thing, since they always should be the same?

modules/sdk-test-multi-cs/sdk-test-multi-cs.csproj
lcodes: Ah yeah the second one is a mistake, but I'd like to keep the former; for the simple reason it allows all test modules to be loaded in the sln found under bindings-csharp; otherwise they conflict on the same project name.

For wasm StdbModule is hardcoded, so TargetName fixes that.

crates/bindings-csharp/Codegen/Module.cs
lcodes: Hrm the reason I did this cast to INamedTypeSymbol is because `ReducerContext` doesn't exist. Trying to make the ISymbol resolve the type its supposed to point to causes a generator exception.

Ours is `BaseReducerContext`, so there shouldn't be any conflicts possible.

lcodes: Or rather, this very codegen pass is what generates `ReducerContext` now

crates/bindings-csharp/Codegen.Tests/fixtures/server/Lib.cs
lcodes: Ah hmm, I'll have to relax other constraints then.

RReverser: Ah heh then test catches an issue as expected.

crates/bindings-csharp/Codegen/Module.cs
RReverser: Oh right, because we autogenerate it... That's unfortunate, but makes sense.

modules/sdk-test-multi-cs/sdk-test-multi-cs.csproj
RReverser: That's fine by me.

crates/bindings-csharp/Codegen/Module.cs
```diff
@@ -129,17 +160,31 @@ public TableDeclaration(GeneratorAttributeSyntaxContext context)
...
+            or Accessibility.Internal => "internal",
+            Accessibility.Public => "public",
+            _ => throw new Exception(
+                "Table row type visibility must be public or internal."
+            ),
+        };
 
-        IsPublic = attrArgs.Any(pair => pair is { Key: "Public", Value.Value: true });
+        Views = new(context.Attributes
+            .Where(a => a.AttributeClass?.ToDisplayString() == "SpacetimeDB.TableAttribute")
```
RReverser, line 174: You don't need this `.Where` - `context.Attributes` is already contextual attribute and is guaranteed to match what was passed in `ForAttributeWithMetadataName`.

crates/bindings-csharp/Codegen/Module.cs
```diff
@@ -209,15 +301,15 @@ public override Scope.Extensions ToExtensions()
...
+                            .Select((col, pos) => (col, pos, col.GetAttrs(v.Name)))
+                            .Where(tuple => tuple.Item3 != ColumnAttrs.UnSet)
                             .Select(pair =>
                                 $$"""
                                 new (
                                     nameof({{ShortName}}),
                                     {{pair.pos}},
                                     nameof({{pair.col.Name}}),
-                                    SpacetimeDB.ColumnAttrs.{{pair.col.Attrs}}
+                                    (SpacetimeDB.ColumnAttrs){{(int)pair.Item3}}
```
RReverser, line 312: Let's keep the typechecked enum instead, like it did before by stringifying the enum's member name.

crates/bindings-csharp/Codegen/Module.cs
```diff
@@ -227,68 +319,29 @@ public override Scope.Extensions ToExtensions()
...
-                    """
-                );
-            }
-        }
-
         return extensions;
     }
 }
 
 record ReducerParamDeclaration : MemberDeclaration
```
RReverser, line 340: Since it doesn't have any own data, I'd probably just remove this class and store `MemberDeclaration`. You'd only need to change `ConvertMember`.

crates/bindings-csharp/Runtime/Runtime.cs
RReverser: (I accidentally marked this resolved, but unresolved back because the earlier singleton comment is still valid)

crates/bindings-csharp/Codegen.Tests/fixtures/server/Lib.cs
```diff
@@ -13,7 +13,7 @@ public partial struct CustomStruct
 }
 
 [SpacetimeDB.Type]
-public partial struct CustomClass
+public partial class CustomClass
```
RReverser, line 16: Huh when did those become structs? I somehow missed it. That was a bug, especially noticeable since those tests were named CustomClass to specifically differentiate them from CustomStructs with the same contents.

crates/bindings-csharp/Codegen/Module.cs
```diff
@@ -31,18 +33,29 @@ bool isEquatable
     )
         : base(name, type, typeInfo)
     {
-        Attrs = attrs;
+        (string?, ColumnAttrs)[] x = [(default(string), attrs)];
```
RReverser, line 36: What's happening here?

crates/bindings-csharp/Codegen/Module.cs
```diff
@@ -31,18 +33,29 @@ bool isEquatable
...
-            .Select(a => (ColumnAttrs)a.ConstructorArguments[0].Value!)
-            .SingleOrDefault();
+            .Select(a => (a.NamedArguments.FirstOrDefault(a => a.Key == "Table").Value.Value as string,
+                a.AttributeClass?.ToString() switch
+                {
+                    "SpacetimeDB.AutoIncAttribute" => ColumnAttrs.AutoInc,
+                    "SpacetimeDB.PrimaryKeyAttribute" => ColumnAttrs.PrimaryKey,
+                    "SpacetimeDB.UniqueAttribute" => ColumnAttrs.Unique,
+                    "SpacetimeDB.IndexedAttribute" => ColumnAttrs.Indexed,
+                    _ => ColumnAttrs.UnSet,
```
RReverser, line 56: Looks like this will even store separate array entries for 3rd-party attributes that don't have anything to do with SpacetimeDB. Those entries will be pretty much useless `(null, UnSet)`, so I'd suggest skipping them over to improve cacheability.

crates/bindings-csharp/Codegen/Module.cs
lcodes: [].ToImmutableArray() doesn't compile without this type hint.

modules/sdk-test-connect-disconnect-cs/Lib.cs
```diff
@@ -1,28 +1,30 @@
...
 
+static partial class Module
+{
     [SpacetimeDB.Reducer(ReducerKind.Connect)]
-    public static void OnConnect(ReducerContext e)
+    public static void OnConnect(ReducerContext ctx)
     {
-        new Connected { identity = e.Sender }.Insert();
+        var row = new Connected { identity = ctx.Sender };
+        ctx.Db.Connected.Insert(ref row);
```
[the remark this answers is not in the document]
lcodes: Ah that's an interesting find!

crates/bindings-csharp/Codegen.Tests/fixtures/server/Lib.cs
lcodes: Ah yeah I was renaming the other one and saw Class in this name but struct used :)

crates/bindings-csharp/Codegen/Module.cs
lcodes: Hrm, isn't this out of scope for this PR? Happy to do it in a future one

crates/bindings-csharp/Codegen/Module.cs
lcodes: Ah yeah now there's separate attributes, like [AutoInc] and [PrimaryKey], so either its a cast to int here or emitting the full bitwise or expression.

lcodes: Figured (cast)15 is better than (ColumnAttrs.PrimaryKey | ColumnAttrs.Unique) from a codegen complexity vs effect perspective.

crates/bindings-csharp/Codegen/Module.cs
RReverser: I meant more generally. Why do we create this `x` with empty table and provided attrs in the first place?

RReverser: Ah this is for fake scheduling fields.

Hm can you just extract `(string? table, ColumnAttrs mask)` into `record struct` after all? It will be basically the same, but a bit more readable in both debug view (which doesn't preserve tuple field names) and because you'll be able to name the type here more clearly.

Also take a look at https://learn.microsoft.com/en-us/dotnet/api/system.collections.immutable.immutablearray.create?view=net-8.0#system-collections-immutable-immutablearray-create-1(-0) which will simplify this one step further.

crates/bindings-csharp/Codegen/Module.cs
```diff
@@ -87,32 +102,57 @@ or SpecialType.System_Int64
...
-        if (Attrs.HasFlag(ColumnAttrs.Unique) && !IsEquatable)
+        if (attrs.HasFlag(ColumnAttrs.Unique) && !IsEquatable)
         {
             throw new Exception(
                 $"{type} {Name} is not valid for Identity, PrimaryKey or PrimaryKeyAuto as it's not an equatable primitive."
             );
         }
     }
 
+    public ColumnAttrs GetAttrs(string tableName) => Attrs
```
RReverser, line 113: It looks like we only ever retrieve those arguments as a mask. Maybe better to store it as pre-aggregated mask grouped by table in the first place?

That would both make the stored array smaller, and those checks simpler.

RReverser: That is, you would have something like

```cs
field
.GetAttributes()
.GroupBy(attr => /* table name */, attr => /* mask */)
.Select(group => new TableAttrs(group.Key, attr.Aggregate(...)))
.ToImmutableArray()
```

and store that.

crates/bindings-csharp/Codegen.Tests/fixtures/server/Lib.cs
RReverser: Looks like I added like that from the very beginning. Yay for copy-paste 🤦‍♂️

crates/bindings-csharp/BSATN.Runtime/Db.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace SpacetimeDB;
+
+public abstract class DbContext<DbView>
```
RReverser, line 3: Let's turn this into `abstract record`. Looks like in all the descendants it's just a data holder for things passed to the constructor, and `record`s have nicer syntax for this use-case.

crates/bindings-csharp/Codegen/Module.cs
lcodes: Hmm, the grouping needs to be more complex tho because its not just by tableName, its tableName or null.

ie all attributes for the column regardless of what table handle is instantiated, masked into attributes for the column specific to the table handle being instantiated

At that point it becomes a tradeoff between memory and compute; there's a good case that recomputing data is either fast or just as fast as keeping it cached, when transforms are this simple and called this infrequently.

crates/bindings-csharp/Codegen/Module.cs
lcodes: Ah nice, didn't know about IA.Create

crates/bindings-csharp/Codegen/Module.cs
RReverser: > its not just by tableName, its tableName or null.

`null` is still fine as a key though, and it will allow to iterate over fewer items each time you call `GetAttrs` - and that happens quite a lot as it's used in couple of loops.

`GetAttrs` will stay either way, it's just that it will need to do less work for each individual column, so you both store less stuff in memory and spend less CPU time - not really a tradeoff.

crates/bindings-csharp/Codegen.Tests/fixtures/server/snapshots/Module#PublicTable.verified.cs
```diff
@@ -188,165 +188,290 @@ SpacetimeDB.BSATN.ITypeRegistrar registrar
...
+                            nameof(ByteField),
+                            global::PublicTable.BSATN.ByteField.GetAlgebraicType(registrar)
+                        ),
+                        new(
+                            nameof(UshortField),
+                            global::PublicTable.BSATN.UshortField.GetAlgebraicType(registrar)
+                        ),
+                        new(
+                            nameof(UintField),
+                            global::PublicTable.BSATN.UintField.GetAlgebraicType(registrar)
```
RReverser, line 211: Are all these changes necessary? I think in this case `BSATN` is still unambiguous even in face of multiple table views?

crates/bindings-csharp/Codegen/Module.cs
RReverser: Well you did remove `IsContextArg` in this PR, which makes this type 1:1 same as `MemberDeclaration`, so might as well remove the entire class.

crates/bindings-csharp/Codegen/Module.cs
```diff
@@ -294,36 +370,35 @@ public ReducerDeclaration(GeneratorAttributeSyntaxContext context)
...
         );
         Scope = new Scope(methodSyntax.Parent as MemberDeclarationSyntax);
     }
 
-    private IEnumerable<ReducerParamDeclaration> NonContextArgs => Args.Where(a => !a.IsContextArg);
-
     public KeyValuePair<string, string> GenerateClass()
     {
+        var args = string.Join(", ", Args.Select(a => $"{a.Name}.Read(reader)"));
+        var argsSep = args == "" ? "" : ", ";
```
RReverser, line 390: Or just `.Prepend("ctx")` above before joining?

crates/bindings-csharp/Codegen/Module.cs
```diff
@@ -394,21 +467,24 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
...
                 // Sort tables and reducers by name to match Rust behaviour.
                 // Not really important outside of testing, but for testing
                 // it matters because we commit module-bindings
                 // so they need to match 1:1 between different langs.
-                var tableNames = tuple.Left.Sort();
+                var tableViews = tuple.Left.Sort((a, b) => a.Key.viewName.CompareTo(b.Key.viewName));
                 var addReducers = tuple.Right.Sort((a, b) => a.Key.CompareTo(b.Key));
                 // Don't generate the FFI boilerplate if there are no tables or reducers.
-                if (tableNames.IsEmpty && addReducers.IsEmpty)
-                {
```
RReverser, line 409: Why did these get removed? I thought we even have mandatory braces in editorconfig so it should tell you about style mismatch 🤔

crates/bindings-csharp/Codegen/Module.cs
lcodes: Good point

crates/bindings-csharp/BSATN.Runtime/Db.cs
```diff
@@ -0,0 +1,11 @@
...
+
+public abstract record DbContext<DbView>
+    where DbView : class, new()
+{
+    public readonly DbView Db;
+
+    public DbContext() => Db = new();
+
+    public DbContext(DbView db) => Db = db;
+}
```
RReverser, line 11: I meant you can then use the shorthand syntax, not just change class to struct.

So just:

```suggestion
public abstract record DbContext<DbView>(DbView Db)
    where DbView : class, new()
{
    // Do we really need this constructor btw?
    public DbContext() : this(new()) { }
}
```

crates/bindings-csharp/Codegen.Tests/fixtures/server/Lib.cs
RReverser: (still needs reverted so we test class-based tables)

crates/bindings-csharp/Codegen/Module.cs
lcodes: Huh, didn't even realize I removed these, I'll add them back

crates/bindings-csharp/BSATN.Runtime/Db.cs
lcodes: Ah I keep forgetting that syntax exists hah, thanks!
The empty ctor is needed for derived classes that don't specify one.

RReverser: > The empty ctor is needed for derived classes that don't specify one.

It seems there is only one such class (`BaseReducerContext`) - let's move the creation of `DbView` to it instead? Then it will be just `record BaseReducerContext<DbView> : DbContext<DbView>(new())`.

crates/bindings-csharp/Codegen.Tests/fixtures/server/snapshots/Module#PublicTable.verified.cs
lcodes: Good catch, yeah this one doesn't need it.

modules/sdk-test-connect-disconnect-cs/Lib.cs
```diff
@@ -1,28 +1,28 @@
...
 
-static partial class Module
-{
-    [SpacetimeDB.Table(Public = true)]
-    public partial struct Connected
-    {
-        public Identity identity;
-    }
+[SpacetimeDB.Table(Public = true)]
+public partial struct Connected {
```
RReverser, line 6: What happened to all these? 😢

lcodes: Dotnet csharpier doesn't seem to work in the module projects

lcodes: used `dotnet format` instead

crates/bindings-csharp/Runtime/Runtime.cs
```diff
@@ -61,18 +61,21 @@ public AlgebraicType GetAlgebraicType(ITypeRegistrar registrar) =>
...
-    public static Identity? SenderIdentity { get; internal set; }
+    public readonly Identity Sender;
+    public readonly Address? Address;
+    public readonly Random Random;
+    public readonly DateTimeOffset Time;
 
-    public static Address? SenderAddress { get; internal set; }
-
-    public static DateTimeOffset Time { get; internal set; }
+    protected BaseReducerContext(
```
RReverser, line 69: Looks like this can be converted into shorthand record syntax now?

crates/bindings-csharp/Runtime/Internal/Module.cs
```diff
@@ -53,6 +53,13 @@ public static class Module
     private static readonly RawModuleDefV8 moduleDef = new();
     private static readonly List<IReducer> reducers = [];
 
+    private static Func<Identity, Address?, Random, DateTimeOffset, IReducerContext>? newContext =
+        null;
+
+    public static void Initialize(
```
RReverser, line 59: Technically we have some other initialize methods here - RegisterTable and RegisterReducer are part of initialization - so this name is somewhat non-descriptive. I'd suggest renaming it to explicit `SetReducerContextCtor` or something.

crates/cli/src/subcommands/generate/csharp.rs
```diff
@@ -706,11 +700,18 @@ pub fn autogen_csharp_globals(ctx: &GenCtx, items: &[GenItem], namespace: &str)
 
     writeln!(
         output,
-        "public class SpacetimeDBClient : SpacetimeDBClientBase<ReducerEvent>"
+        "public class DbConnection : DbConnectionBase<DbConnection, EventContext>"
     );
     indented_block(&mut output, |output| {
-        writeln!(output, "protected SpacetimeDBClient()");
+        writeln!(output, "public readonly RemoteTables RemoteTables = new();");
+        writeln!(output, "public readonly RemoteReducers RemoteReducers = new();");
```
RReverser, line 707: Shouldn't this entire file be excluded from this PR? IIUC you're making client-side changes in #1707 and that one seems a bit further along.

crates/bindings-csharp/Runtime/Runtime.cs
RReverser: At that point I'm actually not even sure that we need `BaseReducerContext`, as `ReducerContext` just forwards all arguments to it.

Could we make `ReducerContext` inherit from `DbContext` on its own?

We have `IReducerContext` interface for erased types anyway.

crates/bindings-csharp/Codegen/Module.cs
```diff
@@ -117,21 +175,122 @@ public TableDeclaration(GeneratorAttributeSyntaxContext context)
             throw new InvalidOperationException("Tagged enums cannot be tables.");
         }
 
-        var attr = context.Attributes.Single().ParseAs<TableAttribute>();
+        Visibility = context.TargetSymbol.DeclaredAccessibility;
+        switch (Visibility)
+        {
+            case Accessibility.ProtectedAndInternal:
+            case Accessibility.NotApplicable:
```
RReverser, line 182: This will include private types if a table class is nested inside another class, because for nested types the default visibility is private.

Is that intentional?

crates/bindings-csharp/Codegen/Module.cs
RReverser: This code still looks suspicious (maybe more so than last time?).

Now you have `numSchedules = schedules.Count();`, but because `schedules` is just a `Views.Select(...)`, you're essentially doing `numSchedules = Views.Length`.

And then inside the if condition you're checking `numSchedules != Views.Length`, which can never be true...

crates/bindings-csharp/Codegen/Module.cs
```diff
@@ -117,21 +175,122 @@ public TableDeclaration(GeneratorAttributeSyntaxContext context)
...
+                    public bool UpdateBy{f.Name}({f.Type} {f.Name}, {globalName} @this) =>
+                        {colEqWhere}.Update(@this);
+                    """;
+            }
         }
     }
 
-    protected override ColumnDeclaration ConvertMember(IFieldSymbol field) => new(field);
+    public IEnumerable<
+        KeyValuePair<(string viewName, string tableName), (string view, string getter)>
```
RReverser, line 253: Nit - it doesn't look like this is even used as actual key-values, but rather each field has its own meaning... so can it be just

```suggestion
        (string viewName, string tableName, string view, string getter)
```

instead to reduce nesting?

RReverser: Also I'd still suggest extracting it into `record struct TableView`.

Unlike type fields, tuple names are compile-time-only syntax sugar, they don't even show up in reflection, and `(string Item1, string Item2)` is rather unhelpful to look at in a debugger or in any serialization context.

crates/bindings-csharp/Codegen/Module.cs
lcodes: Do you mean
`private class Outer { public class Inner {} }`
or
`public class Outer { private class Inner {} }`
?

Only the former should allow [Table] attributes, and being part of a private Outer should make codegen emit an error. Yeah I'll have to add something here.

crates/cli/src/subcommands/generate/csharp.rs
lcodes: Ah right it slipped in from the client sdk pr

crates/cli/src/subcommands/project/csharp/Lib._cs
```diff
@@ -1,28 +1,28 @@
 using SpacetimeDB;
 
-static partial class Module
+public static partial class Module
 {
     [SpacetimeDB.Table]
     public partial struct Person
     {
-        [SpacetimeDB.Column(ColumnAttrs.Unique | ColumnAttrs.AutoInc)]
+	    [SpacetimeDB.AutoInc]
```
RReverser, line 8: Nit: indentation is off?

crates/bindings-csharp/Codegen/Module.cs
```diff
@@ -294,36 +393,42 @@ public ReducerDeclaration(GeneratorAttributeSyntaxContext context)
...
 
-                public void Invoke(BinaryReader reader, SpacetimeDB.ReducerContext ctx) {
-                    {{FullName}}(
-                        {{string.Join(
-                            ", ",
-                            Args.Select(a => a.IsContextArg ? "ctx" : $"{a.Name}.Read(reader)")
-                        )}}
-                    );
+                public void Invoke(BinaryReader reader, SpacetimeDB.Internal.IReducerContext ctx) {
+                    {{FullName}}((SpacetimeDB.ReducerContext){{args}});
```
RReverser, line 431: Nit: I'd move `(SpacetimeDB.ReducerContext)` into the string in the `Prepend` above, it took me a moment to understand that it's not casting all the args, but that the first one happens to be our context, since they're declared and used in different places.

crates/bindings-csharp/Runtime/Internal/ITable.cs
```diff
@@ -7,12 +7,18 @@ public interface ITable<T> : IStructuralReadWrite
...
+
     static abstract Filter CreateFilter();
+}
 
-    // These are static helpers that codegen can use.
+public interface ITableView<View, T>
+    where View : ITableView<View, T>
+    where T : ITable<T>, new()
+{
+    static abstract T ReadGenFields(BinaryReader reader, T row);
```
RReverser, line 19: We can leave this as a future cleanup, but it seems like it would be easier to make this a regular method on `ITable<T>` instead of `static abstract` on `ITableView`? Since it operates on `row` instance anyway.

lcodes: That was my first approach, but with each table view able to have a different set of unique keys, they become per-view functions.

crates/cli/src/subcommands/project/csharp/Lib._cs
lcodes: ah a tab got inserted there

RReverser (approved):
🎉 LGTM

crates/bindings-csharp/Codegen/Module.cs
RReverser: > is better than (ColumnAttrs.PrimaryKey | ColumnAttrs.Unique) from a codegen complexity vs effect perspective

We actually have names for their combinations too, so this shouldn't be a problem. At least before you could also specify `[Column(ColumnAttrs.PrimaryKey | ColumnAttrs.Unique)]` and it would get stringified as `PrimaryKeyAuto` automatically.

I'm probably going to change this.

PR #1720, opened 2024-09-18
Fixes showing server errors on `cli sql` calls

mamcx:
# Description of Changes

The `cli` incorrectly tries to parse as `json` an `http error` masquing the problem. Now it checks that the `StatusCode` is within `200-299.` before returning.

1

# Testing

- [x] *Check the `sql` calls that could generar `error 500`.*
- [x] *Use my branch of adding a new system table that is broken to get errors.


bfops (approved):
Seems reasonable enough!

PR #1722, opened 2024-09-19
cli: Remove dbg!

kim:
Our clippy.toml should disallow this, but doesn't.
Anyway, running codegen with a debug build is not so uncommon, and this
dbg! output is not too informative (unless debugging codegen).

0


mamcx (approved):
LGTM

PR #1725, opened 2024-09-19
perf: Use type ids for type checking

joshua-spacetime:
A `RelExpr` now holds type ids instead of types.
This reduces the size of `RelExpr`.
It avoids copying and cloning types during type checking. Comparisons are also cheaper for primitive types.

### DIFF
--- a/crates/planner/src/logical/ty.rs
+++ b/crates/planner/src/logical/ty.rs
@@ -1,0 +1,240 @@
+use std::fmt::{Display, Formatter};
+
+use spacetimedb_lib::AlgebraicType;
+use spacetimedb_sats::algebraic_type::fmt::fmt_algebraic_type;
+use spacetimedb_sql_parser::ast::BinOp;
+use thiserror::Error;
+
+use crate::static_assert_size;
+
+/// When type checking a [super::expr::RelExpr],
+/// types are stored in a typing context [TyCtx].
+/// It will then hold references, in the form of [TyId]s,
+/// to the types defined in the [TyCtx].
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct TyId(u32);
+
+impl TyId {
+    /// The number of primitive types whose [TyId]s are statically defined.
+    pub const N: usize = 16;
+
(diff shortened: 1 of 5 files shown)


crates/planner/src/logical/ty.rs
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    /// Is this type compatible with this binary operator?
+    pub fn is_compatible_with(&self, op: BinOp) -> bool {
+        match (op, self) {
+            // Logic operators take booleans
+            (BinOp::And | BinOp::Or, Type::Alg(AlgebraicType::Bool)) => true,
+            // Comparison operators take integers or floats
+            (BinOp::Lt | BinOp::Gt | BinOp::Lte | BinOp::Gte, Type::Alg(t)) => t.is_integer() || t.is_float(),
+            // Equality supports numerics, strings, and bytes
```
mamcx, line 202: Why can't the others?

crates/planner/src/logical/ty.rs
```diff
@@ -0,0 +1,243 @@
...
+
+    /// Resolve an id to its [Type].
+    /// Panics if id is out of bounds.
+    pub fn resolve(&self, id: TyId) -> &Type {
+        &self.types[id.0 as usize]
+    }
+
+    /// Add a type to the context and return a [TyId] for it.
+    /// The [TyId] is not guaranteed to be unique to the type.
+    /// However for primitive types it will be.
```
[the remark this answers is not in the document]
joshua-spacetime: No problem, it just requires more memory and it becomes a little more expensive to generate type ids. But the benefit is that we only need type ids to test for equality.

crates/planner/src/logical/ty.rs
joshua-spacetime: Mainly to avoid inconsistencies with other type systems. In the case of Rust, sums and products correspond to enum and struct types. Those types could have custom Eq and Ord implementations and as of right now, we don't have a way of representing that in SpacetimeDB.

mamcx (approved):
LGTM

PR #1726, opened 2024-09-20
smoketests: Allow to skip tests requiring dotnet

kim:
Adds a CLI option `--skip-dotnet`, which skips all dotnet tests.
If not given, and no dotnet was found, exit nonzero.

This is useful in environments where dotnet is not installed, but is
deliberately opt-in so that those tests don't rot.

1

# Testing

smoketests in CI should still run all the dotnet tests

### DIFF
--- a/smoketests/__init__.py
+++ b/smoketests/__init__.py
@@ -31,6 +31,9 @@ TEMPLATE_CARGO_TOML = (re.compile(r"^spacetimedb\s*=.*$", re.M) \
 
 # this is set to true when the --docker flag is passed to the cli
 HAVE_DOCKER = False
+# this is set to true when the --skip-dotnet flag is not passed to the cli,
+# and a dotnet installation is detected
+HAVE_DOTNET = False
 
 # we need to late-bind the output stream to allow unittests to capture stdout/stderr.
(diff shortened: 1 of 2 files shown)


bfops:
This actually makes it look like those tests should have been skipped instead of failing.. I wonder what was wrong with the logic.

bfops:
Ah yeah I see that it used to re-raise, but now we just return false in either case. That seems better.

smoketests/__init__.py
```diff
@@ -138,21 +141,6 @@ def run_cmd(*args, capture_stderr=True, check=True, full_output=False, cmd_name=
 def spacetime(*args, **kwargs):
     return run_cmd(SPACETIME_BIN, *args, cmd_name="spacetime", **kwargs)
 
-
-def _check_for_dotnet() -> bool:
```
jdetter, line 142: I see we moved this check below where we're processing arguments, that's probably a good place for it 👍

jdetter (approved):
If I run on a system where dotnet isn't installed I get the following output with an exit code of `1`:

```
no suitable dotnet installation found
```

If I then run with the `--skip-dotnet` flag it skips the dotnet tests as expected.

PR #1735, opened 2024-09-24
Tighten package sources in C# smoketests

RReverser:
# Description of Changes

Before this, we couple of times ran into issues where a version in SpacetimeDB template was outdated, or NuGet sources were in incorrect order, and so on, which led to either tests failing with misleading error (https://github.com/clockworklabs/SpacetimeDB/commit/2b09c8d1f8868c22f8d3b6af30a492ef1ac9569c being most recent example), or, worse, quietly passing even when code or versions are actually broken.

Apparently, NuGet has a package source mapping feature which allows to ensure that specific packages are only pulled from specific custom sources, which is exactly what we want here. It allows us to ensure that SpacetimeDB.* pacakges for smoketests are pulled from our local build folders, and not from nuget.org.

corresponding GitHub label.

# Testing

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -137,9 +137,14 @@ jobs:
             <packageSources>
               <!-- Local NuGet repositories -->
               <add key="Local SpacetimeDB.BSATN.Runtime" value="../crates/bindings-csharp/BSATN.Runtime/bin/Release" />
-              <!-- Official NuGet.org server -->
-              <add key="NuGet.org" value="https://api.nuget.org/v3/index.json" />
             </packageSources>
+            <packageSourceMapping>
+              <!-- Ensure that SpacetimeDB.BSATN.Runtime is used from the local folder. -->
+              <!-- Otherwise we risk an outdated version being quietly pulled from NuGet for testing. -->
+              <packageSource key="Local SpacetimeDB.BSATN.Runtime">
+                <package pattern="SpacetimeDB.BSATN.Runtime" />
+              </packageSource>
+            </packageSourceMapping>
           </configuration>
           EOF
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -144,6 +144,10 @@ jobs:
               <packageSource key="Local SpacetimeDB.BSATN.Runtime">
                 <package pattern="SpacetimeDB.BSATN.Runtime" />
               </packageSource>
+              <!-- Fallback to NuGet for other packages. -->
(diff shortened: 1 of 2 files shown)


mamcx (approved):
LGTM

.github/workflows/ci.yml
```diff
@@ -137,9 +137,14 @@ jobs:
...
-              <!-- Official NuGet.org server -->
-              <add key="NuGet.org" value="https://api.nuget.org/v3/index.json" />
             </packageSources>
+            <packageSourceMapping>
+              <!-- Ensure that SpacetimeDB.BSATN.Runtime is used from the local folder. -->
+              <!-- Otherwise we risk an outdated version being quietly pulled from NuGet for testing. -->
+              <packageSource key="Local SpacetimeDB.BSATN.Runtime">
+                <package pattern="SpacetimeDB.BSATN.Runtime" />
+              </packageSource>
+            </packageSourceMapping>
```
bfops, line 147: FWIW, the previous nuget-fallback behavior was entirely intentional (see the PR description that introduced it: https://github.com/clockworklabs/SpacetimeDB/pull/1503).

tl;dr if we release breaking changes in this repo, the SDK tests would continue to use a package they were compatible with, rather than breaking for every future SpacetimeDB PR until someone fixed the C# SDK.

bfops: This way, the C# SDK tests _only_ break if you fail to bump the version numbers when you make a breaking change.

I don't feel strongly about keeping it that way (and I believe it's not a required check, so this doesn't "deadlock" us), but that was the original intent.

RReverser: > This way, the C# SDK tests only break if you fail to bump the version numbers when you make a breaking change.

Problem is, we also have the opposite sometimes - we bumped SpacetimeDB version, made some breaking changes, but C# SDK tests are still passing "successfully" because they are still pointing to old BSATN.* packages.

This is too easy to miss at the moment, so I do think the noisy error would be better in this context.

RReverser: Although I can see the arguments for the original behaviour too...

bfops: Sounds good - we should probably add a requirement that people breaking a downstream repo (in this case, the C# SDK) have a fix PR ready to go before merging their breaking change in this repo.

RReverser: I'd say let's try this stricter mode to catch breaking behaviour sooner; if it does prove too annoying, we can always revert those changes.

PR #1740, opened 2024-09-24
CLI - Unify copies of the confirmation/`--force` logic

bfops:
# Description of Changes

We had several different copies and half-factorings of the logic for confirmation prompts / `--force` param. This PR unifies them.

No

2

# Testing

Only existing smoketests for `init`, `publish`, and `identity`. I could be persuaded to do more thorough testing if anybody's concerned.


crates/cli/src/subcommands/publish.rs
```diff
@@ -166,25 +166,21 @@ pub async fn exec(mut config: Config, args: &ArgMatches) -> Result<(), anyhow::E
...
-            println!("Skipping confirmation due to --force.");
-        } else {
-            // Note: `name_or_address` should be set, because it is `required` in the CLI arg config.
-            println!(
-                "This will DESTROY the current {} module, and ALL corresponding data.",
+        // Note: `name_or_address` should be set, because it is `required` in the CLI arg config.
+        println!(
+            "This will DESTROY the current {} module, and ALL corresponding data.",
+            name_or_address.unwrap()
+        );
```
bfops, line 173: Note: This `println!` was moved out of the conditional

crates/cli/src/subcommands/upgrade.rs
```diff
@@ -152,10 +153,8 @@ pub async fn exec(_config: Config, args: &ArgMatches) -> Result<(), anyhow::Erro
 
     if release_version == version::CLI_VERSION {
         println!("You're already running the latest version: {}", version::CLI_VERSION);
-        if !force {
+        if !confirm_prompt(force, "Do you want to reinstall? ")? {
             return Ok(());
-        } else {
-            println!("Force flag is set, continuing with upgrade.");
```
bfops, line 158: Here we added a **new** confirmation prompt, to bring the behavior in line with all the other commands that have a `--force`.

crates/cli/src/util.rs
```diff
@@ -344,13 +344,17 @@ pub fn host_or_url_to_host_and_protocol(host_or_url: &str) -> (&str, Option<&str
...
 /// Return `false` unless the input is `y`.
-pub fn y_or_n(prompt: &str) -> anyhow::Result<bool> {
+pub fn y_or_n(force: bool, prompt: &str) -> anyhow::Result<bool> {
+    if force {
+        println!("Force flag is present, skipping confirmation");
+        return Ok(true);
+    }
     let mut input = String::new();
-    print!("{} (y/n)", prompt);
+    print!("{} [y/N]", prompt);
```
jdetter, line 353: The choice here of capital N was interesting here to me, then I asked chatgpt about it and this explanation makes sense

image

This styling seems exactly correct, well done Zeke 👍

jdetter (approved):
I have not tested this myself but the code looks excellent and I have no objections here 👍

crates/cli/src/util.rs
bfops: I copied it from one of the commands you made, so well done past-John :smile:

PR #1741, opened 2024-09-24
CLI stabilization

bfops:
# Description of Changes

This is a collection of a bunch of small argument changes for the CLI, in preparation to stabilize this API.

Yes, this breaks a lot of the CLI API.

2

# Testing
Existing tests have been updated


crates/cli/src/subcommands/build.rs
```diff
@@ -19,7 +19,6 @@ pub fn cli() -> clap::Command {
                 .long("skip_clippy")
                 .short('S')
                 .action(SetTrue)
-                .env("SPACETIME_SKIP_CLIPPY")
```
bfops, line 22: I did this because we've generally backed off of supporting env vars for now - should I keep it in? (The smoketests use it)

crates/cli/src/subcommands/identity.rs
```diff
@@ -309,12 +305,6 @@ async fn exec_remove(mut config: Config, args: &ArgMatches) -> Result<(), anyhow
         return Err(anyhow::anyhow!("Must provide an identity or name to remove"));
     }
 
-    if force && !(all || all_server.is_some()) {
-        return Err(anyhow::anyhow!(
-            "The --force flag can only be used with --all or --all-server"
-        ));
-    }
```
bfops, line 316: I removed this check since the `--yes` param is a bit more generic than `--force` was. It's kind of fine to provide it when it won't make a difference, since it's just a general "skip any confirmations if they come up" flag now.

crates/cli/src/subcommands/describe.rs
```diff
@@ -13,6 +13,8 @@ pub fn cli() -> clap::Command {
         )
         .arg(
             Arg::new("entity_type")
+                .long("type")
+                .short('t')
```
bfops, line 17: I think we might undo this change. This is required to use `entity_name`, and it's ignored unless we use `entity_name`.

crates/cli/src/subcommands/build.rs
jdetter: I did not know this even existed and I think now that we're solving the clippy problem this is irrelevent. If we remove this logic here and somewhere else in our infrastructure we're setting this environment variable then that should have no impact because we're just ignoring it here. I would recommend we remove it as you've done here.

crates/cli/src/subcommands/logs.rs
```diff
@@ -28,6 +28,8 @@ pub fn cli() -> clap::Command {
         )
         .arg(
             Arg::new("num_lines")
+                .long("num-lines")
+                .short('n')
```
jdetter, line 32: 👍

crates/cli/src/subcommands/generate/mod.rs
```diff
@@ -89,11 +89,12 @@ pub fn cli() -> clap::Command {
...
                 .help("The language to generate"),
         )
-        .arg(Arg::new("delete_files").long("delete-files").action(SetTrue).help(
-            "Delete outdated generated files whose definitions have been \
-             removed from the module. Prompts before deleting unless --force is \
-             supplied.",
-        ))
+        .arg(
+            Arg::new("delete_files")
+                .long("clear-target-directory")
```
jdetter, line 94: The rest of the changes here are non-controvertial other than this one - I might recommend pulling this out so that we can get the rest of these changes in through this PR. We can discuss more what we want to do about `--delete-files`.

crates/cli/src/subcommands/generate/mod.rs
```diff
@@ -89,11 +89,12 @@ pub fn cli() -> clap::Command {
                 .value_parser(clap::value_parser!(Language))
                 .help("The language to generate"),
         )
-        .arg(Arg::new("delete_files").long("delete-files").action(SetTrue).help(
-            "Delete outdated generated files whose definitions have been \
-             removed from the module. Prompts before deleting unless --force is \
```
jdetter, line 94: I'll comment here as well that this help text is also super misleading. Files which existed from the previous generation will be overridden for the new generation (for example, when you update the schema of a table). This will completely overwrite that file including any modifications to it as long as we detect that our special file header is uneditied.

crates/cli/src/subcommands/build.rs
```diff
@@ -16,10 +16,8 @@ pub fn cli() -> clap::Command {
         )
         .arg(
             Arg::new("skip_clippy")
-                .long("skip_clippy")
-                .short('S')
+                .long("skip-println-checks")
```
jdetter, line 19: I like this change 👍

crates/cli/src/subcommands/publish.rs
```diff
@@ -128,13 +112,9 @@ pub async fn exec(mut config: Config, args: &ArgMatches) -> Result<(), anyhow::E
         ));
     }
 
-    if trace_log {
-        query_params.push(("trace_log", "true"));
-    }
-
     let path_to_wasm = if !path_to_project.is_dir() && path_to_project.extension().map_or(false, |ext| ext == "wasm") {
```
jdetter, line 115: @bfops When is the proper time to remove this? Post 0.12?

jdetter (commented):
Just comments for now but will approve after we figure out what's going on with `--delete-files` or we split those changes out into a separate PR.

jdetter (approved):
Approving because it sounds like we're backing out the `--delete-files` changes for now

crates/cli/src/subcommands/publish.rs
bfops: honestly probably 0.12. I'll see if I can revive the PR that removes this. I think multiple subcommands do this iirc.

Issue #846, opened 2024-02-14
CLI - `spacetime publish` should have flags for configuring `wasm-opt`


gefjon:
Optimizing modules with `wasm-opt`, at least under the default `wasm-opt` config, strips their debug symbols, making profiler output (e.g. `perf`) incomprehensible. We need some way to disable this other than "uninstall `wasm-opt`."

### Bare minimum:

Flag to `spacetime publish` like `--no-wasm-opt` which still builds the module in release mode, but does not run `wasm-opt`. (Currently the only way to publish without running `wasm-opt` is via `--debug`, which is obviously undesirable for profiling.)

### Ideal:

- Accept arbitrary flags to pass to `wasm-opt`, either in an environment variable or a flag.
  - `WASM_OPT_FLAGS='-O2 -g -ffm -uim -all' spacetime publish -c`
  - `spacetime publish -c --wasm-opt-flags '-O2 -g -ffm -uim -all'`
- If a set of flags is not supplied, use a sensible set of defaults.
  - Currently we pass `-O2 -all`, though it's not clear to me how much thought went into that choice.
  - Someone should check if we can depend on `--low-memory-unused` and `--zero-filled-memory`.
  - Determining a set of default flags can wait for a follow-up ticket.
- Regardless of whether a set of flags is supplied or not, add input and output arguments.
- If a set of flags is supplied, `wasm-opt` not found becomes a hard error, rather than a warning.

RReverser:
> * Currently we pass `-O2 -all`, though it's not clear to me how much thought went into that choice.

`-O2` has been chosen after careful comparisons of different modes on some example STDB modules. It had the best balance between performance and compilation times (`-O3` is rarely useful in `wasm-opt`).

`-all` was added because C# doesn't emit proper `target_features` section, without which `wasm-opt` errors out on post-MVP Wasm features like bulk memory operations.

RReverser:
FWIW I think `-g` should be fine to always keep around - if input Wasm was already stripped, it won't add anything, and if it has debug info, then it will be simply preserved. But might be better to only add it in debug builds (which STDB CLI already distinguishes).

gefjon:
> FWIW I think `-g` should be fine to always keep around - if input Wasm was already stripped, it won't add anything, and if it has debug info, then it will be simply preserved. But might be better to only add it in debug builds (which STDB CLI already distinguishes).

Well, we'd like a way to have release builds with symbols; that's the impetus for this ticket.

[...]

PR #1743, opened 2024-09-25
CLI - `build` command has a param for wasm-opt flags

bfops:
# Description of Changes

Add a `spacetime build --wasm-opt=false` option, so users can skip wasm-opt even if they have it installed.

Note that this is mutually exclusive with `--debug`. I could be persuaded to merge `--wasm-opt` and `--debug` into a `--profile=debug|release|release-no-wasm-opt` flag if that's preferred.

It also doesn't error if you provide `--wasm-opt=true` and then it can't find wasm-opt. Should that be an error instead of a "warning"? (And the default would become "wasm-opt-if-available" or something).

Addresses https://github.com/clockworklabs/SpacetimeDB/issues/846.

No. This is a new param.

2

# Testing

### DIFF
--- a/crates/cli/src/subcommands/build.rs
+++ b/crates/cli/src/subcommands/build.rs
@@ -57,3 +57,11 @@ pub async fn exec(_config: Config, args: &ArgMatches) -> Result<PathBuf, anyhow:
 
     Ok(bin_path)
 }
+
+pub async fn exec_with_argstring(config: Config, project_path: &PathBuf, args: &str) -> Result<PathBuf, anyhow::Error> {
+    // Note: "build" must be the start of the string, because `build::cli()` is the entire build subcommand.
+    // If we don't include this, the args will be misinterpreted (e.g. as commands).
+    let build_options = format!("build {} --project-path {}", args, project_path.display());
+    let build_args = cli().get_matches_from(build_options.split_whitespace());
+    exec(config.clone(), &build_args).await
+}
--- a/crates/cli/src/subcommands/build.rs
+++ b/crates/cli/src/subcommands/build.rs
@@ -1,3 +1,4 @@
+use crate::tasks::ReleaseLevel;
 use crate::Config;
 use clap::ArgAction::SetTrue;
 use clap::{Arg, ArgMatches};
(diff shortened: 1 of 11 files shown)


crates/testing/tests/standalone_integration_test.rs
```diff
@@ -43,39 +43,34 @@ async fn read_logs(module: &ModuleHandle) -> Vec<String> {
 fn test_calling_a_reducer_in_module(module_name: &'static str) {
     init();
 
-    CompiledModule::compile(module_name, CompilationMode::Debug).with_module_async(
```
bfops, line 46: clippy really wanted to reformat this. It helps to review with whitespace hidden, but it's still not great.

crates/cli/src/subcommands/build.rs
```diff
@@ -30,12 +31,21 @@ pub fn cli() -> clap::Command {
...
                 .help("Builds the module using debug instead of release (intended to speed up local iteration, not recommended for CI)"),
         )
+        .arg(
+            Arg::new("wasm_opt")
+                .long("wasm-opt")
+                .value_parser(clap::builder::BoolishValueParser::new())
+                .default_value("true")
+                .conflicts_with("debug")
+                .help("Whether to optimize the compiled module with wasm-opt (default: true)"),
+        )
```
bfops, line 41: I could also be persuaded to merge `--wasm-opt` and `--debug` into a `--profile=debug|release|release-no-wasm-opt` flag if that's preferred.

RReverser:
> Add a `spacetime build --wasm-opt=false` option, so users can skip wasm-opt even if they have it installed.

Hm what's the usecase? If someone wants quick iteration, I'd say they should use debug builds (which already don't run wasm-opt), and if someone wants to publish a fast module, they should always run wasm-opt as it yield noticeable boost.

bfops:
> > Add a `spacetime build --wasm-opt=false` option, so users can skip wasm-opt even if they have it installed.
> 
> Hm what's the usecase? If someone wants quick iteration, I'd say they should use debug builds (which already don't run wasm-opt), and if someone wants to publish a fast module, they should always run wasm-opt as it yield noticeable boost.

See https://github.com/clockworklabs/SpacetimeDB/issues/846

RReverser:
> at least under the default wasm-opt config, strips their debug symbols, making profiler output (e.g. perf) incomprehensible.

IIRC the original motivation is no longer relevant as we now correctly pass a flag to preserve debug info in wasm-opt (wasm-opt has always supported preserving debug symbols, we just didn't pass that flag originally).

bfops:
> > at least under the default wasm-opt config, strips their debug symbols, making profiler output (e.g. perf) incomprehensible.
> 
> IIRC the original motivation is no longer relevant as we now correctly pass a flag to preserve debug info in wasm-opt (wasm-opt has always supported preserving debug symbols, we just didn't pass that flag originally).

I have no context - I'd love for this to be true! All I know is I was assigned this ticket last week :shrug: I'll ping @cloutiertyler and/or @gefjon.

RReverser:
Just for the context, the fix I'm referring to was here, about a month after the original issue. https://github.com/clockworklabs/SpacetimeDB/pull/1013

gefjon:
So long as we're preserving debug symbols, I consider the original issue solved and see no need to remove wasm-opt. I also don't see any downside to providing the flag, but certainly it's not a priority. Long-term I'd like to have a mechanism for passing arbitrary wasm-opt flags through `spacetime build`/`spacetime publish`, but again, not a priority.

RReverser:
> Long-term I'd like to have a mechanism for passing arbitrary wasm-opt flags

Yeah I think long-term we should have something like `spacetime.toml` where we can specify all sorts of configs, including this one. Command-line args can only hold so much.

PR #1745, opened 2024-09-25
CLI - Change default server back to `local`

bfops:
# Description of Changes

We have resolved to change the default server back to `local` for the time being.

Mostly no. This changes the behavior of e.g. any automation that creates a fresh CLI install, but that doesn't seem very widespread.

1

# Testing
Literally none

### DIFF
--- a/crates/cli/src/config.rs
+++ b/crates/cli/src/config.rs
@@ -151,7 +151,7 @@ impl RawConfig {
             ecdsa_public_key: None,
         };
         RawConfig {
-            default_server: testnet.nickname.clone(),
+            default_server: local.nickname.clone(),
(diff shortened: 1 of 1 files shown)


crates/cli/src/config.rs
```diff
@@ -151,7 +151,7 @@ impl RawConfig {
             ecdsa_public_key: None,
         };
         RawConfig {
-            default_server: testnet.nickname.clone(),
+            default_server: local.nickname.clone(),
```
jdetter, line 154: Hopefully this will be the last time right? 😅

jdetter (approved):
I did bother testing this myself because the change is extremely simple and looks correct. Thanks Zeke 👍

Issue #1602, opened 2024-08-16
RLS: Update describe_module schema


joshua-spacetime:
`__describe_module__` needs to return RLS queries. One way would be to include them as part of the table def.

mamcx:
in this ticket will be validated that the module should fail to compile if the queries fail to compile, as was first requested on for #1600

PR #1746, opened 2024-09-26
Adding the system table for row level security

### LINKED ISSUE #1600: RLS: Store module defined RLS queries in system table
This ticket includes (1) creating a system table for RLS queries, and (2) updating the drivers of `__describe_module__` (database init and update) to insert RLS queries into this system table.

Note, the module should fail to compile if the queries fail to compile.

mamcx:
# Description of Changes
 NOTE: The design changes after several meetings with the team to improve the ergonomics of module compilation.

Add a new system table `st_row_level_security` to store the `SQL` queries for row-level security.

This is a first step to add support for authorization. This PR doesn't validate that the queries are valid, that will done when #1602 is implemented.

Closes #1600.

2: It adds a new `system table` and because we haven't yet migrated support for them, it needs a recreate of the db.

# Testing

- [x] *Add extra test for RLS*
- [x] *Because we don't have any RLS in the initial bootstrap I also add a test for reload after creating one*
- [x] *Check using the `SQL cli` the new table*


crates/schema/src/def/validate/v9.rs
```diff
@@ -582,6 +601,12 @@ impl TableValidator<'_, '_> {
         }
     }
 
+    fn validate_row_level_security_def(&mut self, policy: RawRowLevelSecurityDefV9) -> Result<RowLevelSecurityDef> {
+        let RawRowLevelSecurityDefV9 { name, sql } = policy;
```
kazimuth, line 605: It would be nice to also parse the sql here as well and validate it. Maybe add a TODO?

crates/core/src/db/datastore/system_tables.rs
```diff
@@ -715,6 +746,43 @@ impl From<StConstraintRow> for ConstraintSchema {
...
+/// | row_level_security_id | row_level_security_name | table_id | sql          |
+/// |-----------------------|-------------------------|----------|--------------|
+/// | 1                     | "rls_test"              | 1        | "SELECT ..." |
+#[derive(Debug, Clone, PartialEq, Eq, SpacetimeType)]
+#[sats(crate = spacetimedb_lib)]
+pub struct StRowLevelSecurityRow {
+    pub(crate) row_level_security_id: RowLevelSecurityId,
+    pub(crate) row_level_security_name: Box<str>,
+    pub(crate) table_id: TableId,
+    pub(crate) sql: Box<str>,
```
kazimuth, line 760: I was wondering if we might want to add another field here storing a cache for a compiled expression of some sort, but on further reflection I think that's liable to cause problems, better to have the system tables be less redundant.

kazimuth (approved):
Looks great. We've ended up with a lot of boxes to check when adding stuff to the schema, good job navigating all that. Maybe we should document the process somewhere.

You may also want to add row-level-security to `crates/schema/src/auto_migrate.rs`. It seems to me that arbitrary changes to row-level security can be performed automatically.

crates/schema/src/def/validate/v9.rs
mamcx: This is for a follow-up pr. We are still waiting to integrate the new parser...

mamcx:
> It seems to me that arbitrary changes to row-level security can be performed automatically.

If we need to check the validity of the `sql` then probably not.

crates/primitives/src/lib.rs
```diff
@@ -7,7 +7,7 @@ mod ids;
 
 pub use attr::{AttributeKind, ColumnAttribute, ConstraintKind, Constraints};
 pub use col_list::{ColList, ColSet};
-pub use ids::{ColId, ConstraintId, IndexId, ScheduleId, SequenceId, TableId};
+pub use ids::{ColId, ConstraintId, IndexId, RowLevelSecurityId, ScheduleId, SequenceId, TableId};
```
joshua-spacetime, line 10: This was imported but not used.

crates/core/src/db/datastore/traits.rs
```diff
@@ -438,6 +438,24 @@ pub trait MutTxDatastore: TxDatastore + MutTx {
     fn drop_constraint_mut_tx(&self, tx: &mut Self::MutTx, constraint_id: ConstraintId) -> super::Result<()>;
     fn constraint_id_from_name(&self, tx: &Self::MutTx, constraint_name: &str) -> super::Result<Option<ConstraintId>>;
 
+    // Row-level security
+    fn create_row_level_security_mut_tx(
```
joshua-spacetime, line 442: Does this need to be part of the datastore trait?

crates/lib/src/db/raw_def/v9.rs
```diff
@@ -330,6 +333,17 @@ pub struct RawUniqueConstraintDataV9 {
     pub columns: ColList,
 }
 
+/// Data for the `RLS` policy on a table.
+#[derive(Debug, Clone, SpacetimeType)]
+#[sats(crate = crate)]
+#[cfg_attr(feature = "test", derive(PartialEq, Eq, PartialOrd, Ord))]
+pub struct RawRowLevelSecurityDefV9 {
+    /// The name of the policy. Must be unique within the containing `ModuleDef`.
+    pub name: RawIdentifier,
```
joshua-spacetime, line 342: Can we remove the name? I don't think it's needed in any rls workflow.

crates/schema/src/def.rs
```diff
@@ -416,6 +416,9 @@ pub struct TableDef {
     /// The sequences for the table, indexed by name.
     pub sequences: IdentifierMap<SequenceDef>,
 
+    /// The row-level security policies for the table, indexed by name.
+    pub row_level_security: IdentifierMap<RowLevelSecurityDef>,
```
joshua-spacetime, line 420: If we remove the name, can this just be a `Vec`?

crates/schema/src/def.rs
```diff
@@ -937,6 +960,18 @@ impl ModuleDefLookup for ConstraintDef {
     }
 }
 
+impl ModuleDefLookup for RowLevelSecurityDef {
+    type Key<'a> = &'a Identifier;
```
joshua-spacetime, line 964: What is `ModuleDefLookup` used for? Does `RowLevelSecurityDef` need to implement it? If so, could the `Key` just be the sql string itself?

crates/schema/src/def/validate/v9.rs
joshua-spacetime: Where is this validation utility called? Is this called by the rust bindings for example? If so, we can't do any validation here, because currently semantic analysis can only be run against the server.

crates/schema/src/schema.rs
```diff
@@ -1008,3 +1041,44 @@ impl Schema for ConstraintSchema {
         Ok(())
     }
 }
+
+/// A struct representing the schema of a row-level security policy.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct RowLevelSecuritySchema {
+    pub row_level_security_id: RowLevelSecurityId,
+    pub row_level_security_name: Box<str>,
```
joshua-spacetime, line 1049: Again I think we can remove the name field. Do we even need the primary key column `row_level_security_id`?

joshua-spacetime (changes requested):
My main questions are:

1. Can we get rid of the `name` column? I believe we can.
2. Do we need a primary key column? I believe we don't.

crates/schema/src/def/validate/v9.rs
kazimuth: @mamcx ah yeah I meant later, for now just a TODO is fine.

kazimuth: @joshua-spacetime this is called while parsing the `RawModuleDef` that comes out of a module, both in the server while instantiating a module and in, say, the CLI, while performing tasks like generating client bindings.

crates/schema/src/schema.rs
joshua-spacetime: There can be multiple RLS queries per table. But given a table, will it ever be necessary to ask for a specific RLS query? If that's not really a plausible use case, then an separate id field seems redundant, as does the name.

kazimuth: Chatting with Tyler and Mazdak, we had decided we wanted all entities in the database to have names so that they could be referred to and queried from the system tables. I had initially tried to get rid of all the names, since they were redundant, but was overruled. The intent is now that all entities in the system tables have names so that they can be referred to by the user, even if autogenerated.

crates/core/src/db/datastore/traits.rs
mamcx: Yes, is required because the accessors elsewhere

crates/primitives/src/lib.rs
mamcx: Is used in several parts.

crates/core/src/db/datastore/locking_tx_datastore/datastore.rs
```diff
@@ -428,6 +432,34 @@ impl MutTxDatastore for Locking {
...
+    ) -> Result<RowLevelSecurityId> {
+        let ctx = &ExecutionContext::internal(self.database_address);
+        tx.create_row_level_security(ctx, table_id, row_level_security_schema)
+    }
+
+    fn drop_row_level_security_mut_tx(
+        &self,
+        tx: &mut Self::MutTx,
+        row_level_security_policy_id: RowLevelSecurityId,
+    ) -> crate::db::datastore::Result<()> {
```
joshua-spacetime, line 449: I don't think we'll ever drop an RLS filter individually.

crates/core/src/db/datastore/locking_tx_datastore/datastore.rs
```diff
@@ -428,6 +432,34 @@ impl MutTxDatastore for Locking {
...
+        let ctx = &ExecutionContext::internal(self.database_address);
+        tx.drop_row_level_security(ctx, row_level_security_policy_id)
+    }
+
+    fn row_level_security_for_table_id(
+        &self,
+        tx: &Self::MutTx,
+        table_id: TableId,
+    ) -> crate::db::datastore::Result<Vec<RowLevelSecuritySchema>> {
```
joshua-spacetime, line 458: This should probably return an iterator.

crates/core/src/db/datastore/traits.rs
joshua-spacetime: The datastore traits are a little bloated IMO.

We should really think of the database as being implemented using the datastore. In other words, the datastore should be suitably generic so as to allow multiple database implementations. RLS is not a feature of a datastore but rather a database. So I would argue these methods are better suited to RelationalDB than anything else.

mamcx: I that deserve another pr.

joshua-spacetime (approved):

mamcx:
> My main questions are:
> 
> 1. Can we get rid of the `name` column? I believe we can.
> 2. Do we need a primary key column? I believe we don't.

Yes, is already done

PR #1750, opened 2024-09-27
client-api: Fix publish error reporting

kim:
Since #1679, a call to the publish endpoint that doesn't update but create a database would drop errors and falsely return a success response.

While the `ControlStateWriteAccess::publish` method can't currently express the different error branches, we make it so creation errors are returned as `Some(UpdateDatabaseResult::ErrorExecutingMigration)`.


kazimuth (approved):
Thanks, good catch.

PR #1751, opened 2024-09-27
CLI - `--anon-identity` -> `--anonymous`

bfops:
# Description of Changes

Updates `--anon-identity` to be `--anonymous`. We also remove the shorthand `-a` because we're planning to make identities much more robust, so this should be much less frequent.

Yes, it changes a CLI arg.

2

# Testing

Existing smoketests have been updated

### DIFF
--- a/crates/cli/src/common_args.rs
+++ b/crates/cli/src/common_args.rs
@@ -1,4 +1,5 @@
 use clap::Arg;
+use clap::ArgAction::SetTrue;
 
 pub fn server() -> Arg {
     Arg::new("server")
@@ -12,4 +13,11 @@ pub fn identity() -> Arg {
         .long("identity")
         .short('i')
         .help("The identity to use")
+}
+
+pub fn anonymous() -> Arg {
+    Arg::new("anon_identity")
+        .long("anonymous")
+        .action(SetTrue)
+        .help("Perform this action as an arbitrary newly-created user instead of as myself")
 }
--- a/crates/cli/src/common_args.rs
+++ b/crates/cli/src/common_args.rs
@@ -19,5 +19,5 @@ pub fn anonymous() -> Arg {
     Arg::new("anon_identity")
         .long("anonymous")
         .action(SetTrue)
-        .help("Perform this action as an arbitrary newly-created user instead of as myself")
+        .help("Perform this action with an anonymous identity")
 }
--- a/crates/cli/src/subcommands/call.rs
+++ b/crates/cli/src/subcommands/call.rs
@@ -33,7 +33,7 @@ pub fn cli() -> clap::Command {
                 .conflicts_with("anon_identity")
                 .help("The identity to use for the call"),
         )
-        .arg(common_args::anonymous().conflicts_with("identity"))
+        .arg(common_args::anonymous())
         .after_help("Run `spacetime help call` for more detailed information.\n")
 }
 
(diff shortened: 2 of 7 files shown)


crates/cli/src/subcommands/subscribe.rs
```diff
@@ -71,14 +71,7 @@ pub fn cli() -> clap::Command {
...
         )
-        .arg(
-            Arg::new("anon_identity")
-                .long("anon-identity")
-                .short('a')
-                .conflicts_with("identity")
-                .action(ArgAction::SetTrue)
-                .help("If this flag is present, no identity will be provided when querying the database"),
-        )
+        .arg(common_args::anonymous())
```
bfops, line 74: removed the `.conflicts_with("identity")` because the identity param is already marked as conflicting with this one

crates/cli/src/subcommands/sql.rs
```diff
@@ -40,12 +40,7 @@ pub fn cli() -> clap::Command {
                 .long_help("The identity to use for querying the database. If no identity is provided, the default one will be used."),
         )
         .arg(
-            Arg::new("anon_identity")
-                .long("anon-identity")
-                .short('a')
-                .conflicts_with("identity")
-                .action(ArgAction::SetTrue)
-                .help("If this flag is present, no identity will be provided when querying the database")
+            common_args::anonymous()
```
bfops, line 43: removed the `.conflicts_with("identity")` because the identity param is already marked as conflicting with this one

crates/cli/src/subcommands/describe.rs
```diff
@@ -30,12 +30,7 @@ pub fn cli() -> clap::Command {
                 .long_help("The identity to use to describe the entity. If no identity is provided, the default one will be used."),
         )
         .arg(
-            Arg::new("anon_identity")
-                .long("anon-identity")
-                .short('a')
-                .conflicts_with("identity")
-                .action(SetTrue)
-                .help("If this flag is present, no identity will be provided when describing the database"),
+            common_args::anonymous()
```
bfops, line 33: removed the `.conflicts_with("identity")` because the identity param is already marked as conflicting with this one

crates/cli/src/subcommands/call.rs
```diff
@@ -33,14 +33,7 @@ pub fn cli() -> clap::Command {
...
         )
-        .arg(
-            Arg::new("anon_identity")
-                .long("anon-identity")
-                .short('a')
-                .conflicts_with("identity")
-                .action(ArgAction::SetTrue)
-                .help("If this flag is present, the call will be executed with no identity provided"),
-        )
+        .arg(common_args::anonymous())
```
bfops, line 36: removed the `.conflicts_with("identity")` because the identity param is already marked as conflicting with this one

crates/cli/src/common_args.rs
```diff
@@ -13,3 +14,10 @@ pub fn identity() -> Arg {
         .short('i')
         .help("The identity to use")
 }
+
+pub fn anonymous() -> Arg {
+    Arg::new("anon_identity")
```
bfops, line 19: I kept the old internal name to minimize churn, but I'm happy to rename it too.

crates/cli/src/common_args.rs
```diff
@@ -13,3 +14,10 @@ pub fn identity() -> Arg {
...
+        .long("anonymous")
+        .action(SetTrue)
+        .help("Perform this action as an arbitrary newly-created user instead of as myself")
```
jdetter, line 22: I would update this to be a bit more exact, something like:
> Perform this action with an anonymous identity

specifically, using "user" here I think could be confusing

jdetter (approved):
Approved but please update the help text of the `-a` flag before merging. Otherwise this looks great, thanks Zeke 👍

PR #1752, opened 2024-09-27
CLI - `--force` -> `--yes`/`-y`

bfops:
# Description of Changes

Replace `--force` with a more generic `--yes`/`-y` that just generally skips all confirmation prompts.

Yes, renames a CLI arg.

2

# Testing

Existing tests have been updated


crates/cli/src/subcommands/identity.rs
```diff
@@ -309,12 +305,6 @@ async fn exec_remove(mut config: Config, args: &ArgMatches) -> Result<(), anyhow
         return Err(anyhow::anyhow!("Must provide an identity or name to remove"));
     }
 
-    if force && !(all || all_server.is_some()) {
-        return Err(anyhow::anyhow!(
-            "The --force flag can only be used with --all or --all-server"
-        ));
-    }
```
bfops, line 316: I removed this check since the `--yes` param is a bit more generic than `--force` was. It's kind of fine to provide it when it won't make a difference, since it's just a general "skip any confirmations if they come up" flag now.

crates/cli/src/subcommands/generate/mod.rs
```diff
@@ -105,13 +106,7 @@ pub fn cli() -> clap::Command {
              removed from the module. Prompts before deleting unless --force is \
              supplied.",
         ))
-        .arg(
-            Arg::new("force")
-                .long("force")
-                .action(SetTrue)
-                .requires("delete_files")
```
jdetter, line 112: This change makes `--yes` do nothing without the `--delete-files` flag. I would recommend that if the user passes `--yes` here they probably also meant to pass `--delete-files` here and we should print an error for that. What do you think @bfops?

jdetter (approved):
Optional nit, otherwise this looks good and has been tested by me. Thanks @bfops !

crates/cli/src/subcommands/generate/mod.rs
bfops: I'm fine with that. I feel the same way about it as I do [here](https://github.com/clockworklabs/SpacetimeDB/pull/1752/files#r1779176636), which is that `--yes` is now just kind of a general "please skip any confirmations" parameter, so it feels more reasonable for users to slap it on even when a command won't have any confirmations. It's not _nonsensical_, it just doesn't do anything.

I don't feel very strongly about it, though. I can see it either way.

jdetter: My somewhat controversial solution to this (probably as a separate PR) is to remove `--delete-files` as a flag and just make it the default behavior. My concern here is that many times I've ran the generate command without `--delete-files` and I've ended up in an inconsistent state because some of the old filenames didn't conflict with the new file names and so those files were left behind by the previous run of the generate command. Originally we added this flag because we were worried about just deleting files off of people's computers but we've rectified this concern in 2 ways:
1. Files now have a header on them and we only delete files which have that special header. Users can even prevent the file from being deleted if they modify the header.
2. There is a prompt which shows you the files which are about to be deleted.

Given those 2 things are now implemented, I would say the behavior of `--delete-files` should now be the default behavior and actually running without `--delete-files` when you're overwriting generated files is wrong.

IMO the steps forward are:
1. I have no issue with this merging as is
2. We create a ticket for 1.0 to remove the `--delete-files` flag and make that the default behavior. By default you will get a (y/N) prompt when running generate to delete files (if there are files to be deleted) and this prompt will be skipped if you pass `-y/--yes`.

@bfops What do you think about this?

bfops: We discussed this and settled on keeping things the way they are, and instead potentially add options to the (upcoming) project config files that include a default output directory, and a delete-files-by-default option.

PR #1753, opened 2024-09-27
Minor attribute cleanups for C# module

RReverser:
# Description of Changes

A follow-up to #1719:

 - Hide `ColumnAttrs` and base `ColumnAttribute` classes into `SpacetimeDB.Internal` since they shouldn't be accessed by user.
 - Remove legacy `PrimaryKeyIdentity` alias as `ColumnAttrs` is no longer part of the public API.
 - Extend `ParseAs` helper to support abstract classes by accepting an explicit type.
 - Reuse `ParseAs` in couple of places instead of manual parsing.
 - Combine attributes by table early during parsing to reduce number of stored elements that need to be cached and iterated over.
 - Stringify ColumnAttrs as variant name again.

corresponding GitHub label.

# Testing

### DIFF
--- a/crates/bindings-csharp/Runtime/Attrs.cs
+++ b/crates/bindings-csharp/Runtime/Attrs.cs
@@ -1,73 +1,87 @@
-﻿namespace SpacetimeDB;
-
-[Flags]
-public enum ColumnAttrs : byte
+﻿namespace SpacetimeDB
 {
-    UnSet = 0b0000,
-    Indexed = 0b0001,
-    AutoInc = 0b0010,
-    Unique = Indexed | 0b0100,
-    Identity = Unique | AutoInc,
-    PrimaryKey = Unique | 0b1000,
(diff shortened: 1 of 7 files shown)


crates/bindings-csharp/Runtime/Attrs.cs
```diff
@@ -1,73 +1,87 @@
-﻿namespace SpacetimeDB;
-
-[Flags]
-public enum ColumnAttrs : byte
+﻿namespace SpacetimeDB
 {
```
lcodes, line 2: Don't think the toplevel ns needs {}; namespace Internal {} is fine to use with namespace SpacetimeDB; afaik

RReverser: It's necessary here because you can't mix file-scoped and regular namespace syntax, C# will complain. Idk why they made it so, but 🤷‍♂️

lcodes (approved):

PR #1756, opened 2024-09-28
feat: Integrate type checking into current query code

joshua-spacetime:
# Description of Changes

corresponding GitHub label.

# Testing

### DIFF
--- a/crates/core/src/estimation.rs
+++ b/crates/core/src/estimation.rs
@@ -73,7 +73,7 @@ mod tests {
         execution_context::ExecutionContext,
         sql::compiler::compile_sql,
     };
-    use spacetimedb_lib::AlgebraicType;
+    use spacetimedb_lib::{identity::AuthCtx, AlgebraicType};
     use spacetimedb_sats::product;
     use spacetimedb_vm::expr::CrudExpr;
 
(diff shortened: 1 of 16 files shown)


crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -83,7 +85,11 @@ impl ModuleSubscriptions {
...
             } else {
-                let mut compiled = compile_read_only_query(&self.relational_db, &tx, sql)?;
+                // NOTE: The following ensures compliance with the 1.0 sql api.
+                // Come 1.0, it will have replaced the current compilation stack.
+                parse_and_type_sub(sql, &SchemaViewer::new(&self.relational_db, &*tx, &auth))?;
+
+                let mut compiled = compile_read_only_query(&self.relational_db, &auth, &tx, sql)?;
                 if compiled.len() > 1 {
                     return Result::Err(
                         SubscriptionError::Unsupported(String::from("Multiple statements in subscription query"))
```
mamcx, line 95: Is not this error caught by the new checker  before?

crates/core/src/estimation.rs
```diff
@@ -82,6 +82,7 @@ mod tests {
     }
 
     fn num_rows_for(db: &RelationalDB, sql: &str) -> u64 {
+        let compile_sql = |db, tx, sql_text| compile_sql(db, &AuthCtx::for_testing(), tx, sql_text);
         let tx = db.begin_tx();
```
mamcx, line 86: This looks unnecesary

crates/core/src/sql/ast.rs
```diff
@@ -469,6 +473,33 @@ fn compile_where(table: &From, filter: Option<SqlExpr>) -> Result<Option<Selecti
...
+}
+
+impl<T: TableSchemaView> SchemaView for SchemaViewer<'_, T> {
+    fn schema(&self, name: &str, _: bool) -> Option<Arc<TableSchema>> {
+        let name = name.to_owned().into_boxed_str();
+        let schema = self
+            .tx
+            .find_table(self.db, Table { name })
+            .map(Some)
+            .unwrap_or_else(|_| None)?;
```
mamcx, line 489: Here we should add a comment of why this is correct

mamcx (approved):
LGTM

PR #1757, opened 2024-09-28
[C#] Fix table view names and duplicate table registration

RReverser:
# Description of Changes

A follow-up to #1719.

In Codegen.Tests this test will be actually validated locally & on CI, snapshotted for reference, and checked that it compiles properly, whereas in that folder it was more or less dormant as the SDK test runner is intended only for tests that match pairwise between Rust/C# and doesn't know about the new folder.

This reveals at least one minor issue, but worth checking that there're no others.

corresponding GitHub label.

# Testing


crates/bindings-csharp/Codegen.Tests/fixtures/server/snapshots/Module#FFI.verified.cs
```diff
@@ -471,8 +635,11 @@ public static void Main()
         SpacetimeDB.Internal.Module.RegisterReducer<Init>();
         SpacetimeDB.Internal.Module.RegisterReducer<InsertData>();
         SpacetimeDB.Internal.Module.RegisterReducer<InsertData2>();
+        SpacetimeDB.Internal.Module.RegisterReducer<InsertMultiData>();
         SpacetimeDB.Internal.Module.RegisterReducer<ScheduleImmediate>();
         SpacetimeDB.Internal.Module.RegisterReducer<SendScheduledMessage>();
+        SpacetimeDB.Internal.Module.RegisterTable<global::MultiTable>();
+        SpacetimeDB.Internal.Module.RegisterTable<global::MultiTable>();
```
RReverser, line 642: @lcodes This needs to be fixed up.

crates/bindings-csharp/Codegen.Tests/fixtures/server/snapshots/Module#MultiTable.verified.cs
```diff
@@ -0,0 +1,130 @@
...
+        [
+            new(
+                new(
+                    TableName: nameof(MultiTable),
+                    Columns:
+                    [
+                        new(nameof(Name), BSATN.Name.GetAlgebraicType(registrar)),
+                        new(nameof(Foo), BSATN.Foo.GetAlgebraicType(registrar)),
+                        new(nameof(Bar), BSATN.Bar.GetAlgebraicType(registrar))
+                    ],
```
RReverser, line 60: Nit for future work: we should be able to create and reuse this list instead of generating codegen for duplicates, as all table views have the same list of column names and types.

crates/bindings-csharp/Codegen.Tests/fixtures/server/Lib.cs
```diff
@@ -143,3 +143,25 @@ public static void Init(ReducerContext ctx)
         );
     }
 }
+
+[Table(Name = "MultiTable1", Public = true)]
+[Table(Name = "MultiTable2")]
+public partial struct MultiTable
```
lcodes, line 149: Nitpick; could we name this one RowType or something?

ie making it clear the distinction between the struct Foo being the row type, and the [Table(Name = "Foo")] being the table defined from that row type.

lcodes (approved):

crates/bindings-csharp/Codegen.Tests/fixtures/server/Lib.cs
RReverser: Done

PR #1774, opened 2024-09-30
CLI - `spacetime generate` always deletes stale files

bfops:
# Description of Changes

`spacetime generate --delete-files` deletes all files in the target directory that start with the SpacetimeDB autogenerated file header. This PR makes turns that behavior permanently on, and removes the flag.

We think this is unlikely to negatively affect users, since it only deletes files that start with the header that is automatically  generated by SpacetimeDB. If users do experience negative effects, we can add a flag to disable this behavior.

Yes. We no longer support `--delete-files`.

2

# Testing

```
$ cargo run -- generate --lang csharp --out-dir $WORK/my_new_project/ --project-path $WORK/my_new_project/
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.23s
     Running `target/debug/spacetime generate --lang csharp --out-dir /home/lead/work/clockwork/my_new_project/ --project-path /home/lead/work/clockwork/my_new_project/`
Saving config to /home/lead/.spacetime/config.toml.
    Checking spacetime-module v0.1.0 (/mnt/nutera/work/my_new_project)
    Finished `release` profile [optimized] target(s) in 0.15s
    Finished `release` profile [optimized] target(s) in 0.07s
For best performance install wasm-opt from https://github.com/WebAssembly/binaryen/releases.
Continuing with unoptimised module.
Build finished successfully.
compilation took 2.092558726s

$ cp $WORK/my_new_project/AddReducer.{cs,cs.stale}

$ cargo run -- generate --lang csharp --out-dir $WORK/my_new_project/ --project-path $WORK/my_new_project/
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.23s
     Running `target/debug/spacetime generate --lang csharp --out-dir /home/lead/work/clockwork/my_new_project/ --project-path /home/lead/work/clockwork/my_new_project/`
Saving config to /home/lead/.spacetime/config.toml.
    Checking spacetime-module v0.1.0 (/mnt/nutera/work/my_new_project)
    Finished `release` profile [optimized] target(s) in 0.14s
    Finished `release` profile [optimized] target(s) in 0.07s
For best performance install wasm-opt from https://github.com/WebAssembly/binaryen/releases.
Continuing with unoptimised module.
Build finished successfully.
compilation took 2.110778341s
The following files were not generated by this command and will be deleted:
  /home/lead/work/clockwork/my_new_project/AddReducer.cs.stale
Are you sure you want to delete these files? [y/N]y
Files deleted successfully.
```


crates/cli/src/subcommands/generate/mod.rs
```diff
@@ -157,41 +152,40 @@ pub async fn exec(config: Config, args: &clap::ArgMatches) -> anyhow::Result<()>
...
-                    if !contents.starts_with("// THIS FILE IS AUTOMATICALLY GENERATED BY SPACETIMEDB.") {
-                        continue;
-                    }
-                }
-
-                if paths
-                    .iter()
-                    .any(|x| x.file_name().unwrap() == path.file_name().unwrap())
-                {
+    // TODO: We should probably just delete all generated files before we generate any, rather than selectively deleting some afterward.
```
jdetter, line 155: @bfops Yeah hold on I think there are some bugs here. We only prompt users *after* we've already done the generation? That seems wrong to me. Also we should only be replacing files which contain our special header, not just overwriting whatever file is already there. Can we fix those issues here or do you want to do a separate PR?

bfops: Let's do those things in a separate PR since they're not API-breaking.
Also, it looks like we have the capacity to _generate_ files in subdirectories, but it's not clear to me that we can _delete_ them from subdirectories..

jdetter (approved):
I'm approving this - we're fixing the other existing issues in a separate PR

crates/cli/src/subcommands/generate/mod.rs
bfops: Created https://github.com/clockworklabs/SpacetimeDB/issues/1786

PR #1779, opened 2024-10-01
offset index read integration

Shubham8287:
# Description of Changes
1. Create `IndexFile`: a wrapper around `IndexFileMut` for exposing only read apis.
2. Update `Repo` trait to get `IndexFile`
3. Refactoring: Method naming changes to show clear distinction between read and write version of `IndexFile`
4. Modify `Commits` iterator to lookup through Index for initial commit. 

N/A

# Testing
image
Did some manual testing

### DIFF
--- a/crates/commitlog/src/commitlog.rs
+++ b/crates/commitlog/src/commitlog.rs
@@ -468,7 +468,13 @@ impl CommitInfo {
         }
     }
 
-    // If initial offset falls within a commit, adjust it to the commit boundary
+    // If initial offset falls within a commit, adjust it to the commit boundary.
+    //
+    // Returns `true` if the initial offset is past `commit`.
+    // Returns `false` if `self` isn't `Self::Initial`,
+    // or the initial offset has been adjusted to the starting offset of `commit`.
+    //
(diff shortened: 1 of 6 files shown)


crates/commitlog/src/commitlog.rs
```diff
@@ -496,8 +504,11 @@ impl<R: Repo> Commits<R> {
         // interesting.
         let prev_error = self.last_error.take();
 
+        // Skip entries before the initial commit.
+        if self.last_commit.offset_before_initial(&commit.min_tx_offset) {
```
Shubham8287, line 508: Ah! I see problem with skipping commits, you will be able to jump to the tx falling in middle of some commit.

Shubham8287: but that will not work with current implementation as well. Will report `out-of-order`.
So I am confused whether the change to make `CommitlogInfo::Initial` any `commit.min_tx_offset` rather `segment.min_tx_offset` is Correct? @kim 
Looks fine to me.

Shubham8287: Went little dirty to detect and adjust offset to commit boundary inside the iterator itself.

PR #1784, opened 2024-10-01
CLI - Fix identity removal helptext

bfops:
# Description of Changes

If an identity has become invalid, we advise users to call `spacetime identity remove ${IDENTITY}`.. but it should be `spacetime identity remove -i ${IDENTITY}` since https://github.com/clockworklabs/SpacetimeDB/pull/1482/ landed.

Nope, just helptext.

1

# Testing
Literally none

### DIFF
--- a/crates/cli/src/util.rs
+++ b/crates/cli/src/util.rs
@@ -363,7 +363,7 @@ pub fn unauth_error_context<T>(res: anyhow::Result<T>, identity: &str, server: &
             "Identity {identity} is not valid for server {server}.
 Has the server rotated its keys?
 Remove the outdated identity with:
-\tspacetime identity remove {identity}
+\tspacetime identity remove -i {identity}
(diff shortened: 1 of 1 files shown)


crates/cli/src/util.rs
```diff
@@ -363,7 +363,7 @@ pub fn unauth_error_context<T>(res: anyhow::Result<T>, identity: &str, server: &
             "Identity {identity} is not valid for server {server}.
 Has the server rotated its keys?
 Remove the outdated identity with:
-\tspacetime identity remove {identity}
+\tspacetime identity remove -i {identity}
 Generate a new identity with:
 \tspacetime identity new --no-email --server {server} --default"
```
bfops, line 368: should we really be telling users to use `--no-email`?

jdetter (approved):

PR #1761, opened 2024-09-29
Generic and optimized WebSocket API

Centril:
# Description of Changes

Implements https://github.com/clockworklabs/SpacetimeDBPrivate/pull/993 but with some changes.

Fixes https://github.com/clockworklabs/SpacetimeDB/issues/1080.
Fixes https://github.com/clockworklabs/SpacetimeDB/issues/1450.
Fixes https://github.com/clockworklabs/SpacetimeDB/issues/1201.
Closes https://github.com/clockworklabs/SpacetimeDB/pull/1200.

# API and ABI breaking changes

Yes We Can.

PR #1785, opened 2024-10-01
Fix `RowSizeHint::index_to_range`

Centril:
# Description of Changes

Make sure `index_to_range` doesn't advance beyond `rows_data`.


bfops:
This fixes an issue originally introduced in https://github.com/clockworklabs/SpacetimeDB/pull/1761.

bfops (approved):
I don't have all the context for this change, but this looks like a generally correct end condition check fix for an iterator, and I've confirmed it fixes at least one bug that was causing crashes.

PR #1791, opened 2024-10-02
commitlog: Yield `StoredCommit` in iterators

kim:
This provides access to the checksum, which may be of interest.
Leaving the `History` trait as-is, as there is no immediate need for the
checksum.

Also fix rustdoc warnings.


Shubham8287 (approved):
change looks reasonable.

Issue #1762, opened 2024-09-30
CLI - Config-modifying commands print the "saving config" text twice


bfops:
# Background
Commands like `spacetime server set-default` print the "Saving config" text twice.

This is because all commands save the config once right after loading it. (So this issue likely applies to all commands that actually modify the config).

# Definition of Done
TODO

PR #1799, opened 2024-10-02
CLI - Remove the always-save-config behavior

bfops:
# Description of Changes

We used to unconditionally call `config.save()` at the start of every CLI command, in order to ensure that a default version of the config is written to disk if it wasn't already there.

This PR updates the behavior to only save to disk _if_ the file didn't exist.

This addresses https://github.com/clockworklabs/SpacetimeDB/issues/1762.

None.

2

# Testing

```
$ cargo run -- server list
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.25s
     Running `target/debug/spacetime server list`
 DEFAULT  HOSTNAME                                              PROTOCOL  NICKNAME                    
     ***  127.0.0.1:3000                                        http      local                       
          testnet.spacetimedb.com                               https     testnet                     
          bitcraft-team.spacetimedb.org                         https     bitcraft-internal           
          bitcraft-internal-bitcraft-internal-spacetimedb:3000  http      bitcraft-internal-tailscale 
          testnet.staging.spacetimedb.com                       https     staging                     

$ mv ~/.spacetime/config.toml{,.bak} 

$ cargo run -- server list
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.25s
     Running `target/debug/spacetime server list`
Saving config to /home/lead/.spacetime/config.toml.
 DEFAULT  HOSTNAME                 PROTOCOL  NICKNAME 
     ***  127.0.0.1:3000           http      local    
          testnet.spacetimedb.com  https     testnet  
```

### DIFF
--- a/crates/cli/src/config.rs
+++ b/crates/cli/src/config.rs
@@ -792,14 +792,19 @@ impl Config {
 
     pub fn load() -> anyhow::Result<Self> {
         let home_path = Self::system_config_path();
-        let home = if home_path.exists() {
-            Self::load_from_file(&home_path)
-                .inspect_err(|e| eprintln!("config file {home_path:?} is invalid: {e:#?}"))?
+        let config = if home_path.exists() {
+            Self {
+                home: Self::load_from_file(&home_path)
+                    .inspect_err(|e| eprintln!("config file {home_path:?} is invalid: {e:#?}"))?,
+            }
         } else {
-            RawConfig::new_with_localhost()
+            let config = Self {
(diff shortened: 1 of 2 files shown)


jdetter (approved):
Tested by me, the commands I was running before are no longer printing the log and I've ran some commands which do write to the config and it does print the log. Also FWIW I think this implementation is more correct than what we had before. Thanks Zeke! 🙏

Issue #1463, opened 2024-06-26
Switch to/Provide option for gzip


cloutiertyler:
Puru: 

> Provide an option in WS API to send the response encoded in gzip via `?gzip` or `?encoding=gzip`
>
> OR
>
> Totally switch to gzip, which should be a non-breaking change(AFAIK).
>
> From discord:
>
> "Just outlining the general outline:
> If we wait for brotli in web, and it comes after 1 year, it'll take over 1 more to > spread across all the browsers(* cough * Safari firefox * cough *) and have > enuf usage as well(Like safari being tied to mac updates so it'll be the slowest to get). After that, it would take time for Node to get it as well, after which we gotta wait for us bumping our TS SDK's minimum node version to the one which supports DecompressionStream(brotli) So we're looking at 2-3 years from now at a minimum. Which means 100KB for our SDK for the said period of time
> 
> If we switch to gzip or provide it as an option, we can have a non-breaking > change in the JS SDK(which internally starts using DecompressionStream('gzip') widely supported). We can also do this as a breaking change while bumping up a minimum node version, pre-1.0. And the bundle size can go down very easily to 5-10KBs, which is really good to have. Sizes don't matter in the long run, but it does if people start complaining and this causes others to try to use HTTP API instead, which is a failure of the SDK.
>
> Something to think about pre-1.0"

cloutiertyler:
Why wouldn't a change to gzip be a breaking change?

cloutiertyler:
Puru:

> If we wait for brotli in web, and it comes after 1 year, it'll take over 1 more to spread across all the browsers(* cough * Safari firefox * cough *) and have enuf usage as well(Like safari being tied to mac updates so it'll be the slowest to get). After that, it would take time for Node to get it as well, after which we gotta wait for us bumping our TS SDK's minimum node version to the one which supports DecompressionStream(brotli) So we're looking at 2-3 years from now at a minimum. Which means 100KB for our SDK for the said period of time

> If we switch to gzip or provide it as an option, we can have a non-breaking change in the JS SDK(which internally starts using DecompressionStream('gzip') widely supported). We can also do this as a breaking change while bumping up a minimum node version, pre-1.0. And the bundle size can go down very easily to 5-10KBs, which is really good to have. Sizes don't matter in the long run, but it does if people start complaining and this causes others to try to use HTTP API instead, which is a failure of the SDK.

> Something to think about pre-1.0

PuruVJ:
[...]

PR #1802, opened 2024-10-03
Add gzip + none compression algos and let SDK pick compression

Centril:
# Description of Changes

Adds gzip and none compression choices, and expose it to the rust sdk via `DbConnectionBuilder::with_compression`.

- [x] Implement in Host
- [x] Implement in Rust SDK
- [x] Implement in TS SDK, https://github.com/clockworklabs/spacetimedb-typescript-sdk/pull/109
- [x] Implement in C# SDK, https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/pull/155

Fixes https://github.com/clockworklabs/SpacetimeDB/issues/1463.
Closes https://github.com/clockworklabs/SpacetimeDB/pull/1642.

None. If a compression is not chosen brotli is used.

### DIFF
--- a/crates/client-api-messages/Cargo.toml
+++ b/crates/client-api-messages/Cargo.toml
@@ -15,6 +15,7 @@ bytestring.workspace = true
 brotli.workspace = true
 chrono = { workspace = true, features = ["serde"] }
 enum-as-inner.workspace = true
+flate2.workspace = true
 serde = { workspace = true, features = ["derive"] }
 serde_json.workspace = true
 smallvec.workspace = true
--- a/crates/client-api-messages/src/websocket.rs
+++ b/crates/client-api-messages/src/websocket.rs
@@ -33,7 +33,7 @@ use spacetimedb_sats::{
     SpacetimeType,
 };
 use std::{
-    io::{self, Read as _},
+    io::{self, Read as _, Write as _},
     sync::Arc,
 };
 
@@ -74,7 +74,7 @@ pub trait WebsocketFormat: Sized {
 
     /// Convert a `QueryUpdate` into `Self::QueryUpdate`.
     /// This allows some formats to e.g., compress the update.
-    fn into_query_update(qu: QueryUpdate<Self>) -> Self::QueryUpdate;
+    fn into_query_update(qu: QueryUpdate<Self>, compression: Compression) -> Self::QueryUpdate;
 }
 
 /// Messages sent from the client to the server.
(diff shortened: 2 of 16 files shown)


crates/client-api-messages/src/websocket.rs
```diff
@@ -165,12 +165,15 @@ pub struct OneOffQuery {
...
 }
 
-/// The tag recognized by ghe host and SDKs to mean no compression of a [`ServerMessage`].
+/// The tag recognized by the host and SDKs to mean no compression of a [`ServerMessage`].
 pub const SERVER_MSG_COMPRESSION_TAG_NONE: u8 = 0;
 
 /// The tag recognized by the host and SDKs to mean brotli compression  of a [`ServerMessage`].
 pub const SERVER_MSG_COMPRESSION_TAG_BROTLI: u8 = 1;
 
+/// The tag recognized by the host and SDKs to mean brotli compression  of a [`ServerMessage`].
```
gefjon, line 174:
```suggestion
/// The tag recognized by the host and SDKs to mean GZip compression  of a [`ServerMessage`].
```

crates/sdk/src/db_connection.rs
```diff
@@ -877,6 +881,16 @@ but you must call one of them, or else the connection will never progress.
...
 
+    /// Sets the compression used when a certain threshold in the message size has been reached.
+    ///
+    /// The current threshold used by the host is 1KiB for the entire server message
+    /// and for individual query updates.
+    /// Note however that this threshold is not guaranteed and may change without notice.
+    pub fn with_compression(mut self, compression: Compression) -> Self {
+        self.compression = compression;
+        self
+    }
```
gefjon, line 892: I don't love exposing this here, honestly. I'm not sure we're prepared to commit to the set of compression algorithms we expose, the default, or to what extent the host promises to comply with the client's requested format. I also don't honestly see a strong need for it - as I understand it we wanted to offer GZip because JS engines have it built-in, so we remove a dependency from the TypeScript client SDK and likely get more optimized decompression there. Neither of these things apply to the Rust client SDK, so I don't see the value in exposing this. Notably I think we actively do not want people choosing `Compression::None`.

All this to say, I would feel better leaving this method unstable for now.

```suggestion
    #[doc(hidden)]
    /// Sets the compression used when a certain threshold in the message size has been reached.
    ///
    /// The current threshold used by the host is 1KiB for the entire server message
    /// and for individual query updates.
    /// Note however that this threshold is not guaranteed and may change without notice.
    ///
    /// This method is unstable and is subject to change incompatibly without a major version bump.
    pub fn with_compression(mut self, compression: Compression) -> Self {
        self.compression = compression;
        self
    }
```

crates/client-api-messages/Cargo.toml
```diff
@@ -15,6 +15,7 @@ bytestring.workspace = true
 brotli.workspace = true
 chrono = { workspace = true, features = ["serde"] }
 enum-as-inner.workspace = true
+flate2.workspace = true
```
gefjon, line 18: Could be a follow-up, but I think we should consider feature-gating this (and brotli, I suppose). There's at least some value to reducing the Rust client SDK's dependency tree, especially since (as discussed in another comment) I see very little value for an external user of the SDK in switching away from the default compression format.

gefjon (approved):
Looks good! One doc-comment typo, one request to mark something unstable, one note about low-priority future work. Apply my suggestions, or argue that `with_compression` should be stable, and then go ahead and merge.

PR #1804, opened 2024-10-04
Make smoketests not log expected teardown errors

kazimuth:
# Description of Changes

There were teardown errors in the logs despite the tests not failing (because our testing framework doesn't instrument stdout/stderr during teardown, I think). Getting rid of them makes it easier to diagnose actual test failures

# Testing

Ran the smoketests


bfops (approved):
This seems reasonable to me!

PR #1812, opened 2024-10-09
Websocket API: Light transaction updates & `NoSuccessNotify`

Centril:
# Description of Changes

Does two things:

1. Adds `TransactionUpdateLight` that consists only of the `request_id` and the `DatabaseUpdate<F>`. This is emitted when the query param `&light=true` is set (unless the update is sent to the caller in which case a full update is sent).
2. Adds `flags: enum CallReducerFlags { FullUpdate, NoSuccessNotify }` to `ClientMessage`. When `NoSuccessNotify` is set, the host will not send back an empty update to the caller on success without the caller having subscribed to any affected queries. This is exposed in the SDK via `ctx.set_reducer_flags.my_reducer(CallReducerFlags::NoSuccessNotify)`. When the default `FullUpdate` is used, full updates are sent, whether the caller was subscribed to the relevant queries or not.

TODO:
- [x] Host implementation of the above.
- [x] Rust SDK implement of the above.
- [x] C# SDK implementation of the above, https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/pull/170
- [x] TS SDK implementation of the above, https://github.com/clockworklabs/spacetimedb-typescript-sdk/pull/110

This breaks the websocket API/protocol.

### DIFF
--- a/crates/cli/src/subcommands/generate/typescript.rs
+++ b/crates/cli/src/subcommands/generate/typescript.rs
@@ -390,7 +390,7 @@ dbViewConstructor: (imp: DBConnectionImpl) => {{
 }},
 reducersConstructor: (imp: DBConnectionImpl, setReducerFlags: SetReducerFlags) => {{
   return new RemoteReducers(imp, setReducerFlags);
-}}
+}},
 setReducerFlagsConstructor: () => {{
   return new SetReducerFlags();
 }}"
--- a/crates/cli/tests/snapshots/codegen__codegen_typescript.snap
+++ b/crates/cli/tests/snapshots/codegen__codegen_typescript.snap
@@ -893,7 +893,7 @@ const REMOTE_MODULE = {
   },
   reducersConstructor: (imp: DBConnectionImpl, setReducerFlags: SetReducerFlags) => {
     return new RemoteReducers(imp, setReducerFlags);
-  }
+  },
   setReducerFlagsConstructor: () => {
     return new SetReducerFlags();
   }


bfops:
@Centril I'm happy to review the CLI changes - Do you have recommendations on how to test this? Should I just regenerate bindings for an example client for each language?

crates/core/src/client/client_connection.rs
```diff
@@ -244,12 +261,20 @@ impl ClientConnection {
         args: ReducerArgs,
         request_id: RequestId,
         timer: Instant,
+        flags: CallReducerFlags,
     ) -> Result<ReducerCallResult, ReducerCallError> {
+        let caller = match flags {
+            CallReducerFlags::FullUpdate => Some(self.sender()),
+            // Setting `sender = None` causes `eval_updates` to skip sending to the caller
+            // as it has no access to the caller other than by id/addr.
+            CallReducerFlags::NoSuccessNotify => None,
```
gefjon, line 270: Does this also affect the `sender: Identity` and client address passed to the reducer in WASM? I feel like those should still be passed.

gefjon (commented):
Ideally I would like to see tests added to the Rust SDK test suite which exercise this new functionality. If time is a concern, these could be added in a follow-up. I am happy to assist in writing the tests, but I don't think I have a strong enough grasp of the intended semantics to actually write them myself.

I would very much like clarification on the `ReducerContext.sender` passed to reducers whose callers requested no-notify.

I've also left many suggestions which are all: don't document the WS protocol on user-facing parts of the SDK. The WS protocol should be abstracted away to users of the SDK.

crates/core/src/client/client_connection.rs
gefjon: Excellent, thanks!

gefjon (approved):

crates/cli/src/subcommands/subscribe.rs
```diff
@@ -205,6 +205,8 @@ where
     S: TryStream<Ok = WsMessage> + Unpin,
     S::Error: std::error::Error + Send + Sync + 'static,
 {
+    const RECV_TX_UPDATE: &str = "protocol error: received transaction update before initial subscription update";
```
bfops, line 208: nit: this name doesn't seem to match the purpose of this constant

bfops (approved):
(codeowner review) the `crates/cli/src/subcommands/subscribe.rs` changes look benign to me :+1:

crates/client-api/src/routes/subscribe.rs
```diff
@@ -43,7 +43,12 @@ pub struct SubscribeParams {
 #[derive(Deserialize)]
 pub struct SubscribeQueryParams {
     pub client_address: Option<AddressForUrl>,
-    pub compression: Option<Compression>,
+    #[serde(default)]
+    pub compression: Compression,
+    /// Whether we want "light" responses, tailored to network bandwidth constrained clients.
+    /// This knob works by setting other, more specifc, knobs to the value.
+    #[serde(default)]
+    pub light: bool,
```
cloutiertyler, line 51: I thought the plan was to encode this as part of the protocol? Something like `v1.bin.spacetimedb.light` or am I misremembering? I could definitely be misremembering.

cloutiertyler (approved):
In principle, this looks good and correct to me. I just want to be cautious about merging, particularly if we have not yet tested this. Maybe we hold off on doing a basic CircleGame test tomorrow before merging.

crates/client-api/src/routes/subscribe.rs
cloutiertyler: Ack. It's definitely a bikeshed. Just wondering if you knew what we had originally said.

PR #1813, opened 2024-10-09
Correct `Schema::check_compatible` to resolve types in a ModuleDef

kazimuth:
# Description of Changes

When performing an automatic migration, we check the old moduledef is compatible with the data stored in the system tables.
But right now, the types stored in the system tables are resolved, but the types in the moduledef are not. So the check breaks on nested types. This PR fixes that and adds a regression test for it.

I'd like to do a followup PR adding more intensive automigration testing.

0

# Testing

Added a failing test, fixed the test.

### DIFF
--- a/crates/core/src/db/update.rs
+++ b/crates/core/src/db/update.rs
@@ -30,16 +30,16 @@ pub fn update_database(
     let existing_tables = stdb.get_all_tables_mut(tx)?;
 
     // TODO: consider using `ErrorStream` here.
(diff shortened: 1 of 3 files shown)


bfops:
This fixes the issue for me! I don't really understand the context around this code but seems like an improvement.

cloutiertyler (approved):
I do understand the context, and this LGTM. Thank you for the regression test as well.

PR #1815, opened 2024-10-09
Rust modules: Revise `BTreeIndexBounds`

gefjon:
# Description of Changes

- Make BTree filter operators reject owned values of non-`Copy` types, so that diagnostics prefer to suggest `&x` instead of `x.clone()`.
- Restrict BTree filter args to a small set of types defined by our previous "Consistent filtering rules" proposal, per @RReverser 's request.
- Also make it so that `String` columns can be queried by `&str` arguments and ranges thereof.
- Add some comments to the macros that implement `BTreeIndexBounds`, because it's a doozy.

For reasons that escape me, this breaks inference on integer literals: they get defaulted too early to `i32`, even when the indexed column is of a different type like `u32` or `i64`. We could potentially work around this by making `i32` a `FilterableValue` for any integer column type, but that opens the door for some nasty truncation-related bugs. Failing that, we'll have to decide whether allowing `&str` for `String`, and owned values for `Copy` types, is more valuable than correctly inferring the types of integer literals (which would in that case be `&0` instead of `0i64`).

`BTreeIndexBounds` is now less permissive than it was in the 0.12 release. Also the integer literal thing.

3 - Complex macros and traits, but relatively contained. The only undesirable interactions I can forsee are other cases like the integer literal thing where type inference gets less reliable because we have more complex traits involved.

# Testing

- [x] Wrote a new test case in `rust-wasm-test` which demonstrates many combinations of valid args.
- [x] Added some comments in the same place with some invalid combinations, and tested uncommenting them and seeing that they do not compile.


crates/bindings/src/table.rs
```diff
@@ -434,12 +435,84 @@ impl<Tbl: Table, IndexType, Idx: Index> BTreeIndex<Tbl, IndexType, Idx> {
...
+///   and arrays thereof,
+///   should not implement it, as C# and TypeScript use reference equality for those types.
+/// - It should only be implemented for owned values if those values are `Copy`.
+///   Otherwise it should only be implemented for references.
+///   This is so that rustc and IDEs will recommend rewriting `x` to `&x` rather than `x.clone()`.
+/// - `Arg: FilterableValue<Column = Col>`
+///   for any pair of types `(Arg, Col)` which meet the above criteria
+///   is desirable if `Arg` and `Col` have the same BSATN layout.
+///   E.g. `&str: FilterableValue<Column = String>` is desirable.
+pub trait FilterableValue: Serialize {
```
coolreader18, line 458: Wait, why are we going back to a closed set of types that can be filtered on?

coolreader18 (approved):
I guess if that's the direction we wanna go, these changes look good to me.

PR #1833, opened 2024-10-10
Remove email and recovery related identity endpoints

jsdt:
# Description of Changes

This removes code for handling emails and identity recovery, since this should move elsewhere.

corresponding GitHub label.

2. This is not very complex. We are just removing code. The risk here is in missing something that depends on these endpoints.

# Testing

This is failing the internal tests, but the internal `cargo test`s pass with [this PR](https://github.com/clockworklabs/SpacetimeDBPrivate/pull/1041).


cloutiertyler:
This generally looks good to me. It's correct, although it will need a corresponding internal PR.

crates/client-api/src/auth.rs
```diff
@@ -96,7 +97,15 @@ pub struct SpacetimeAuth {
 impl SpacetimeAuth {
     /// Allocate a new identity, and mint a new token for it.
     pub async fn alloc(ctx: &(impl NodeDelegate + ControlStateDelegate + ?Sized)) -> axum::response::Result<Self> {
-        let identity = ctx.create_identity().await.map_err(log_and_500)?;
+        // TODO: I'm just sticking in a random string until we change how identities are generated.
```
cloutiertyler, line 100: We'll just have to remember to go back and swap this.

cloutiertyler (approved):
This LGTM

crates/client-api/src/auth.rs
jsdt: This will be fixed in the next PR.

PR #1836, opened 2024-10-10
Make `Timestamp` a special SATS type

gefjon:
# Description of Changes

- Add timestamp as a new "special" `AlgebraicType`.
- Move `Timestamp` to the SATS crate, where previously it was in both `bindings` and `client-api-messages`.
  - Other special types are mostly defined in `lib`, which seems wrong to me.
- Update Rust client SDK and codegen to use `Timestamp` rather than `u64`.
- Make `Timestamp` store `micros_since_unix_epoch: i64`, where previously it had `micros_since_unix_epoch: u64`.
   - Signed value lets us represent dates before the Unix epoch, which Rust `SystemTime` and C# `DateTime(Offset)` can do.
   - Retaining 64-bit microseconds keeps us binary-compatible with past versions.
- Add `TimeDuration`, analogous to `Duration` and `TimeSpan`, as another new "special" `AlgebraicType`.
- `TimeDuration` is `micros: i64`, to match `Timestamp`.
  - I'm not sure this is the right choice; C# has signed `TimeSpan` but Rust has unsigned `Duration`.
  - Different operations become fallible or infallible depending on which you choose.
  - Honestly I doubt it really matters.
- `ScheduleAt` stores `Timestamp` or `TimeDuration` in its two variants.
  - Use in `ScheduleAt::Interval` means arguably we want unsigned `TimeDuration`, as a negative interval seems meaningless.
  - Currently using the magnitude / absolute value, so that `ScheduleAt::Interval(-10 minutes)` is the same as `ScheduleAt::Interval(10 minutes)`.
  - Again, probably doesn't really matter,  as most users will construct `ScheduleAt`s via `Duration::into`, I hope.
- Update the WebSocket protocol to use these new SATS types rather than raw `u64`s.

Companion PRs:
- [TypeScript SDK](https://github.com/clockworklabs/spacetimedb-typescript-sdk/pull/114).
- [Private](https://github.com/clockworklabs/SpacetimeDBPrivate/pull/1068).
- [C# client SDK](https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/pull/233).

Still TODO:
- Update docs.

- `Timestamp` type (e.g. in `ReducerContext`) is now defined in the SATS crate, re-exported from lib and bindings, and has slightly different methods.
  - Most notably, `Timestamp::now` is deprecated and stubbed in the WASM target.
- The SATS type and serialization of `Timestamp` is changed from `U64` to `Product { __timestamp_micros_since_unix_epoch__: I64 }`. This is layout-equivalent in BSATN and I think BFLATN, but changes the JSON format. It will also make old `ModuleDef`s / system tables detect a migration.
- WebSocket message schema changed to use new types.

2 - Changing the SATS definitions of types has been shown recently to be potentially scary. On the other hand, this redefinition is layout-compatible, so should be much less breaking than when we redefined `Identity`.

# Testing

- [x] Added unit tests that `Timestamp` is well-behaved and compatible with `SystemTime`.
- [x] Added SDK tests for the same.
- [ ] Whatever needs to happen to test with C# and TypeScript.


gefjon:
Having discussed this more with the team, there's an additional wrinkle: while C#'s `DateTime` is semantically correct as a translation for `AlgebraicType::timestamp()`, use of that type is fraught because of its locale-dependent printing. I have been encouraged to prefer `DateTimeOffset`, which maintains locale-independent printing by encoding a time zone, but that's no good because we don't store the time zone in `AlgebraicType::timestamp()` and so would have to discard it when serializing. We would still be representing the same point in time, but it would be, say, 12:00 UTC rather than 7:00 EST. Implicitly losing information in this way seems undesirable.

My disposition is to avoid these tough questions by retaining `spacetimedb::Timestamp` as a user-facing type (adding it to the client SDKs) and providing methods on it to convert to and from `SystemTime` or `DateTimeOffset`.

I'm also interested in providing a special SATS type analogous to Rust `Duration` or C# `TimeSpan`, for storage in `ScheduleAt::Interval`. I'm not sure what to name this type - I would prefer not to use either `Duration` or `TimeSpan` to avoid confusion when writing Rust and C# respectively. I'm considering `TimeDuration`, which is redundant but clear and unambiguous.

kazimuth:
Are we sure we don't want locale-dependent printing? That seems like a reasonable default to me but I could be convinced otherwise.

crates/sats/src/product_type.rs
```diff
@@ -11,6 +11,10 @@ use crate::{AlgebraicType, AlgebraicValue, ProductTypeElement, SpacetimeType, Va
 pub const IDENTITY_TAG: &str = "__identity__";
 /// The tag used inside the special `Address` product type.
 pub const ADDRESS_TAG: &str = "__address__";
+/// The tag used inside the special `Timestamp` product type.
+pub const TIMESTAMP_TAG: &str = "__timestamp_nanos_since_unix_epoch";
```
kazimuth, line 15: Nit: should this be `__timestamp_nanos_since_unix_epoch__`?

gefjon:
TODO: Switch back to `micros: u64`, rather than `nanos: i64`. The latter is likely more correct, but the former is BSATN- and BFLATN-compatible with our old code, which is more valuable than the incremental improvement in precision or the ability to represent negative durations or pre-Unix timestamps.

gefjon:
@kazimuth to help me figure out why C# is busted, and to update the C# client SDK by re-generating the `client-api-messages` bindings.

kazimuth (commented):
A couple of comments from catching up to work on the C# stuff

crates/lib/src/scheduler.rs
```diff
@@ -18,38 +19,64 @@ use spacetimedb_sats::{
...
+                let now = SystemTime::now();
+                let time = SystemTime::from(*time);
+                time.duration_since(now).unwrap_or(Duration::ZERO)
             }
-            ScheduleAt::Interval(dur) => Duration::from_micros(*dur),
+            // TODO(correctness): Determine useful behavior on negative intervals,
+            // as that's the case where `to_duration` fails.
+            // Currently, we use the magnitude / absolute value,
+            // which seems at least less stupid than clamping to zero.
+            ScheduleAt::Interval(dur) => dur.to_duration_abs(),
```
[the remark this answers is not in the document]
gefjon: I am not sure I fully understand the distinction here, or where such a comment would go.

kazimuth:
https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/pull/233

modules/benchmarks-cs/circles.cs
```diff
@@ -31,7 +31,7 @@ public partial struct Circle(uint entity_id, uint player_id, float x, float y, f
         public uint player_id = player_id;
         public Vector2 direction = new(x, y);
         public float magnitude = magnitude;
-        public ulong last_split_time = (ulong)(DateTimeOffset.UtcNow.Ticks / 10);
+        public Timestamp last_split_time = new Timestamp(DateTimeOffset.UtcNow.Ticks / 10);
```
kazimuth, line 34:
```suggestion
        public Timestamp last_split_time = (Timestamp)DateTimeOffset.UtcNow;
```

crates/core/src/messages/instance_db_trace_log.rs
```diff
@@ -64,7 +64,7 @@ pub struct CreateIndex {
 }
 #[derive(Clone, Serialize, Deserialize)]
 pub struct InstanceEvent {
-    pub event_start_epoch_micros: Timestamp,
+    pub event_start_epoch_nanos: Timestamp,
```
kazimuth, line 67:
```suggestion
    pub event_start: Timestamp,
```

kazimuth (approved):
Approving RE: `Timestamp`, `core`, and `schema`.

kazimuth:
TODO: test `elapsed` does not panic on 0ns duration

gefjon:
> TODO: test `elapsed` does not panic on 0ns duration

Turns out we don't even define `elapsed`, so this is a non-issue. We can add it in a follow-up.

crates/cli/src/tasks/csharp.rs
```diff
@@ -74,19 +74,33 @@ pub(crate) fn build_csharp(project_path: &Path, build_debug: bool) -> anyhow::Re
...
+    }
+    let possible_output_paths = [
+        project_path.join(format!("bin/{config_name}/net8.0/wasi-wasm/{subdir}/StdbModule.wasm")),
+        project_path.join(format!("bin~/{config_name}/net8.0/wasi-wasm/{subdir}/StdbModule.wasm")),
+    ];
+    if possible_output_paths.iter().all(|p| p.exists()) {
+        anyhow::bail!(concat!(
+            "For some reason, your project has both a `bin` and a `bin~` folder.\n",
+            "I don't know which to use, so please delete both and rerun this command so that we can see which is up-to-date."
+        ));
```
bfops, line 98: should this be a first step? Like if both `bin` and `bin~` exist, then maybe _one_ of them is outdated, so the above "you're using deprecated bindings" messages might be erroneous.

bfops: that said, I'm surprised that we newly need this behavior. I know we hardcode the C#/Unity SDK project itself to generate `bin~` so that the output isn't picked up by Unity (which ignores things with a `~` suffix) but I didn't think that leaked into users' projects..

crates/cli/src/tasks/csharp.rs
```diff
@@ -74,19 +74,33 @@ pub(crate) fn build_csharp(project_path: &Path, build_debug: bool) -> anyhow::Re
     } else {
         "AppBundle"
     };
-    let mut output_path = project_path.join(format!("bin/{config_name}/net8.0/wasi-wasm/{subdir}/StdbModule.wasm"));
-    if !output_path.exists() {
-        // check for the old .NET 7 path for projects that haven't migrated yet
-        output_path = project_path.join(format!("bin/{config_name}/net7.0/StdbModule.wasm"));
+
```
bfops, line 77:
```suggestion
    // TODO: This code looks for build outputs in both `bin` and `bin~` as output directories. @bfops feels like we shouldn't have to look for `bin~`, since the `~` suffix is just intended to cause Unity to ignore directories, and that shouldn't be relevant here. We do think we've seen `bin~` appear though, and it's not harmful to do the extra checks, so we're merging for now due to imminent code freeze. At some point, it would be good to figure out if we do actually see `bin~` in module directories, and where that's coming from (which could suggest a bug).
```

bfops (approved):
My codeowned files look good given the time-sensitivity.
`crates/cli/src/tasks/csharp.rs`

PR #1845, opened 2024-10-11
CLI - Stabilize `spacetime server` subcommands

bfops:
# Description of Changes

- Made the `server` argument anonymous
- Named the `url` arg on `spacetime server add`
- Replaced `--host` and `--protocol` params with `--url`

Yes, this breaks the `spacetime server` subcommand APIs.

1

# Testing

- [x] Updated existing smoketests


smoketests/tests/servers.py
```diff
@@ -7,19 +7,31 @@ class Servers(Smoketest):
...
 
-        out = self.spacetime("server", "fingerprint", "-s", "localhost")
+        out = self.spacetime("server", "fingerprint", "localhost")
         self.assertIn("Fingerprint is unchanged for server localhost", out)
+
+    def edit_server(self):
+        """Verify that we can edit server configurations"""
+
+        out = self.spacetime("server", "add", "--url", "https://foo.com", "foo", "--no-fingerprint")
+        out = self.spacetime("server", "edit", "--url", "https://testnet.spacetimedb.com", "--new-name", "testnet", "--no-fingerprint")
```
bfops, line 31: so it turns out that `edit` can implicitly use the default server - do we like this behavior?

bfops: ah `ping` and `fingerprint` also implicitly use the default server. I kinda feel like all of these commands should take an explicit server?

crates/cli/src/subcommands/server.rs
```diff
@@ -75,15 +80,14 @@ fn get_subcommands() -> Vec<Command> {
...
-            .arg(common_args::server().help("The nickname, host name or URL of the server to ping")),
+            .arg(Arg::new("server").help("The nickname, host name or URL of the server to ping")),
         Command::new("edit")
             .about("Update a saved server's nickname, host name or protocol")
-            .arg(common_args::server().help("The nickname, host name or URL of the server"))
+            .arg(Arg::new("server").help("The nickname, host name or URL of the server"))
             .arg(
                 Arg::new("nickname")
                     .help("A new nickname to assign the server configuration")
-                    .short('n')
```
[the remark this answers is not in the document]
cloutiertyler: Nah, I think ditching it is fine. You don't run the command all that often and I think the explicitness helps demonstrate what's going on.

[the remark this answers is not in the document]
cloutiertyler: Can always add it, but can't take it away.

smoketests/tests/servers.py
cloutiertyler: I feel like explicit is probably the way to go here. I don't think the implicit assumption adds a lot to the user experience and we can always add it back later if we want. (can't remove it though).

cloutiertyler (approved):

bfops:
(I just wanted to check that `Internal tests` had passed first)

PR #1848, opened 2024-10-14
Add btree index filters to the C# sdk with codegen

lcodes:
# Description of Changes

Adds support for multi-column btree index filtering in the C# client
- codegen emitting Filter() overloads in the index handles
- codegen emitting BTree storage and comparison operators
- runtime changes to make Address and Identity comparable

No breaks.

Most complexity is in the codegen, the runtime behavior is straightforward; a SortedSet<> per btree with a comparator matching its key columns, and Filter() methods querying that sorted set with min/max values derived from the table row type.

# Testing

### DIFF
--- a/crates/cli/src/subcommands/generate/csharp.rs
+++ b/crates/cli/src/subcommands/generate/csharp.rs
@@ -401,6 +401,7 @@ fn autogen_csharp_access_funcs_for_struct(
     product_type: &ProductType,
     table_name: &str,
     schema: &TableSchema,
+    btrees: &Vec<BTreeIndex>
 ) {
     let csharp_table_name = table_name.to_case(Case::Pascal);
     let constraints = schema.backcompat_column_constraints();
@@ -417,6 +418,7 @@ fn autogen_csharp_access_funcs_for_struct(
             None => continue,
             Some(x) => x,
         };
+        writeln!(output);
         writeln!(
             output,
             "public readonly ref struct {csharp_field_name_pascal}UniqueIndex"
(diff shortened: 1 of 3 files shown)


bfops:
What's the testing for this change? It looks like the test suite is failing - do existing tests need updating, and/or does it have new tests for the new behavior?

RReverser:
This needs rebase now that Identity/Address representations have changed + snapshot regen.

crates/cli/src/subcommands/generate/csharp.rs
```diff
@@ -430,69 +431,152 @@ fn autogen_csharp_access_funcs_for_struct(
...
+                let field = &product_type.elements[col.idx()];
+                let field_type = &field.algebraic_type;
+                match csharp_field_type(field_type) {
+                    None => todo!(),
+                    Some(x) => {
+                        let field_name = field.name.as_ref().expect("autogen'd tuples should have field names");
+                        let csharp_field_name_pascal = field_name.replace("r#", "").to_case(Case::Pascal);
+                        let range = match field_type {
+                            AlgebraicType::Product(pt) =>
+                                if pt.elements[0].name == Some("__address_bytes".into()) { ("Address.MinValue", "Address.MaxValue") }
```
RReverser, line 449: We should reuse the `is_address` and other helpers so that we don't hardcode those names in more places.

crates/cli/src/subcommands/generate/csharp.rs
```diff
@@ -430,69 +431,152 @@ fn autogen_csharp_access_funcs_for_struct(
...
+                            AlgebraicType::I32 => ("int.MinValue", "int.MaxValue"),
+                            AlgebraicType::U32 => ("uint.MinValue", "uint.MaxValue"),
+                            AlgebraicType::I64 => ("long.MinValue", "long.MaxValue"),
+                            AlgebraicType::U64 => ("ulong.MinValue", "ulong.MaxValue"),
+                            AlgebraicType::I128 => ("I128.MinValue", "I128.MaxValue"),
+                            AlgebraicType::U128 => ("U128.MinValue", "U128.MaxValue"),
+                            AlgebraicType::I256 => ("I256.MinValue", "I256.MaxValue"),
+                            AlgebraicType::U256 => ("U256.MinValue", "U256.MaxValue"),
+                            AlgebraicType::F32 => ("float.MinValue", "float.MaxValue"),
+                            AlgebraicType::F64 => ("double.MinValue", "double.MaxValue"),
```
RReverser, line 467: You should be able to use `is_integer() || is_float()` + `scalar_or_string_name` instead of repeating those.

crates/cli/src/subcommands/generate/csharp.rs
```diff
@@ -430,69 +431,152 @@ fn autogen_csharp_access_funcs_for_struct(
...
+                            writeln!(output, "var {0} = a.{0}.CompareTo(b.{0});", field.csharp_field_name_pascal);
+                            writeln!(output, "if ({0} != 0) return {0};", field.csharp_field_name_pascal);
+                        },
+                        AlgebraicType::String => {
+                            writeln!(output, "var {0} = String.Compare(a.{0}, b.{0}, StringComparison.InvariantCulture);", field.csharp_field_name_pascal);
+                            writeln!(output, "if ({0} != 0) return {0};", field.csharp_field_name_pascal);
+                        },
+                        _ => {
+                            writeln!(output, "if (a.{0} < b.{0}) return -1;", field.csharp_field_name_pascal);
+                            writeln!(output, "if (a.{0} > b.{0}) return 1;", field.csharp_field_name_pascal);
```
RReverser, line 494: All this shouldn't be necessary, you can just return the result of `String.Compare` as-is.

crates/cli/src/subcommands/generate/csharp.rs
```diff
@@ -430,69 +431,152 @@ fn autogen_csharp_access_funcs_for_struct(
...
+        writeln!(output, "sealed class {0}Comparer : IComparer<{struct_name_pascal_case}>", index.csharp_index_name);
+        indented_block(output, |output| {
+            writeln!(output, "public int Compare({struct_name_pascal_case}? a, {struct_name_pascal_case}? b)");
+            indented_block(output, |output| {
+                writeln!(output, "if (a == null || b == null) return -1;");
+                for field in &fields {
+                    match field.field_type {
+                        AlgebraicType::Product(_) => {
+                            writeln!(output, "var {0} = a.{0}.CompareTo(b.{0});", field.csharp_field_name_pascal);
+                            writeln!(output, "if ({0} != 0) return {0};", field.csharp_field_name_pascal);
```
RReverser, line 486: This shouldn't be necessary, you can just return result of `CompareTo`.

crates/cli/src/subcommands/generate/csharp.rs
```diff
@@ -430,69 +431,152 @@ fn autogen_csharp_access_funcs_for_struct(
...
+                    None => todo!(),
+                    Some(x) => {
+                        let field_name = field.name.as_ref().expect("autogen'd tuples should have field names");
+                        let csharp_field_name_pascal = field_name.replace("r#", "").to_case(Case::Pascal);
+                        let range = match field_type {
+                            AlgebraicType::Product(pt) =>
+                                if pt.elements[0].name == Some("__address_bytes".into()) { ("Address.MinValue", "Address.MaxValue") }
+                                else if pt.elements[0].name == Some("__identity_bytes".into()) { ("Identity.MinValue", "Identity.MaxValue") }
+                                else { todo!() },
+                            AlgebraicType::String => ("\"\"", "\"\\uFFFF\\uFFFF\""),
```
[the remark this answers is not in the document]
RReverser: This will probably require some logic revamp to account for types without a maximum.

PR #1849, opened 2024-10-14
RLS: Adding a new filter! macro

### LINKED ISSUE #1601: RLS: Implement the filter! proc macro
As defined in the RLS proposal. All that is required for this ticket is that `filter!` add its SQL query to the output of `__describe_module__`.

### LINKED ISSUE #1602: RLS: Update describe_module schema

mamcx:
# Description of Changes

Add a new `filter!` `proc-macro` for defining row-level security policies (`rls`) using `SQL`. 

It is a continuation of #1746.

Closes #1602.

The new macro:

* Only checks the syntax of the `SQL` query at compile time using the subscription dialect. It is impossible to fully validate it until we publish the module and all the table definitions are already committed to the database.
* Can be placed anywhere in the module

To avoid tracking a graph of dependencies, we remove and then add against all the `rls` and finally compile them to validate it.

1

# Testing

- [x] *Unit testing that assert the presence of the migration steps.*
- [x] *Update smoke test `auto_migration.py`.*
- [x] *Manual inspection using the `cli`*

### DIFF
--- a/crates/bindings-macro/Cargo.toml
+++ b/crates/bindings-macro/Cargo.toml
@@ -13,6 +13,7 @@ bench = false
 
 [dependencies]
 spacetimedb-primitives.workspace = true
+spacetimedb-sql-parser.workspace = true
 
 bitflags.workspace = true
 humantime.workspace = true
--- a/crates/bindings-macro/src/lib.rs
+++ b/crates/bindings-macro/src/lib.rs
@@ -17,7 +17,6 @@ use proc_macro::TokenStream as StdTokenStream;
 use proc_macro2::{Span, TokenStream};
 use quote::{format_ident, quote, quote_spanned, ToTokens};
 use std::borrow::Cow;
-use std::collections::HashMap;
 use std::hash::{DefaultHasher, Hash, Hasher};
 use std::time::Duration;
 use syn::ext::IdentExt;
--- a/crates/bindings-macro/src/lib.rs
+++ b/crates/bindings-macro/src/lib.rs
@@ -17,6 +17,8 @@ use proc_macro::TokenStream as StdTokenStream;
 use proc_macro2::{Span, TokenStream};
 use quote::{format_ident, quote, quote_spanned, ToTokens};
 use std::borrow::Cow;
+use std::collections::HashMap;
+use std::hash::{DefaultHasher, Hash, Hasher};
 use std::time::Duration;
 use syn::ext::IdentExt;
 use syn::meta::ParseNestedMeta;
@@ -1236,4 +1238,75 @@ pub fn schema_type(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
     })()
     .unwrap_or_else(syn::Error::into_compile_error)
     .into()
+}
+
+fn parse_sql(input: ParseStream) -> syn::Result<String> {
+    use spacetimedb_sql_parser::parser::sub;
+
+    let lookahead = input.lookahead1();
+    let sql = if lookahead.peek(syn::LitStr) {
+        let s = input.parse::<syn::LitStr>()?;
+        // Checks the query is syntactically valid
+        let _ = sub::parse_subscription(&s.value()).map_err(|e| syn::Error::new(s.span(), format_args!("{e}")))?;
+
+        s.value()
+    } else {
+        return Err(lookahead.error());
+    };
+
+    Ok(sql)
+}
+
+/// Generates code for registering a row-level security `SQL` function.
+///
+/// A row-level security function takes a `SQL` query expression that is used to filter rows.
+///
+/// The query follows the same syntax as a subscription query.
+///
+/// **Example:**
+///
+/// ```rust,ignore
+/// /// Players can only see what's in their chunk
+/// spacetimedb::filter!("
+///     SELECT * FROM LocationState WHERE chunk_index IN (
+///         SELECT chunk_index FROM LocationState WHERE entity_id IN (
+///             SELECT entity_id FROM UserState WHERE identity = @sender
+///         )
+///     )
+/// ");
+/// ```
+///
+/// **NOTE:** The `SQL` query expression is pre-parsed at compile time, but only check is a valid
+/// subscription query *syntactically*, not that the query is valid when executed.
+///
+/// For example, it could refer to a non-existent table.
+#[proc_macro]
+pub fn filter(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
+    let rls_sql = syn::parse_macro_input!(input with parse_sql);
+
+    let mut hasher = DefaultHasher::new();
+    rls_sql.hash(&mut hasher);
+    let rls_name = format_ident!("rls_{}", hasher.finish());
+
+    let register_rls_symbol = format!("__preinit__20_register_{rls_name}");
+
+    let generated_describe_function = quote! {
+        #[export_name = #register_rls_symbol]
+        extern "C" fn __register_rls() {
+            spacetimedb::rt::register_row_level_security::<#rls_name>()
+        }
+    };
+
+    let emission = quote! {
+        const _: () = {
+            #generated_describe_function
+        };
+        #[allow(non_camel_case_types)]
+        struct #rls_name { _never: ::core::convert::Infallible }
+        impl spacetimedb::rt::RowLevelSecurityInfo for #rls_name {
+            const SQL: &'static str = #rls_sql;
+        }
+    };
+
+    emission.into()
 }
--- a/crates/core/src/db/update.rs
+++ b/crates/core/src/db/update.rs
@@ -2,12 +2,14 @@ use super::datastore::locking_tx_datastore::MutTxId;
 use super::relational_db::RelationalDB;
 use crate::database_logger::SystemLogger;
 use crate::execution_context::ExecutionContext;
+use crate::sql::parser::RowLevelExpr;
 use spacetimedb_data_structures::map::HashMap;
 use spacetimedb_lib::db::auth::StTableType;
+use spacetimedb_lib::db::raw_def::v9::RawRowLevelSecurityDefV9;
 use spacetimedb_lib::AlgebraicValue;
 use spacetimedb_primitives::ColSet;
 use spacetimedb_schema::auto_migrate::{AutoMigratePlan, ManualMigratePlan, MigratePlan};
-use spacetimedb_schema::def::TableDef;
+use spacetimedb_schema::def::{ModuleDefLookup, TableDef};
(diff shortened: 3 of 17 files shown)


crates/core/src/db/update.rs
```diff
@@ -221,6 +229,20 @@ fn auto_migrate_database(
             spacetimedb_schema::auto_migrate::AutoMigrateStep::RemoveSchedule(_) => {
                 anyhow::bail!("Removing schedules is not yet implemented");
             }
+            spacetimedb_schema::auto_migrate::AutoMigrateStep::AddRowLevelSecurity(sql_rls) => {
+                system_logger.info(&format!("Adding row-level security `{sql_rls}`"));
+                log::info!("Adding row-level security `{sql_rls}`");
+                let tables = all_tables.values().cloned().collect::<Vec<_>>();
+                let rls = RawRowLevelSecurityDefV9::lookup(plan.new, sql_rls).unwrap();
```
mamcx, line 236: Yes, this is sad. I ponder the possibility of adding an alternative `lookup`

crates/bindings-macro/Cargo.toml
```diff
@@ -13,6 +13,7 @@ bench = false
 
 [dependencies]
 spacetimedb-primitives.workspace = true
+spacetimedb-sql-parser.workspace = true
```
joshua-spacetime, line 16: We have to parse on the server regardless, so I would say let's not add it to the bindings.

crates/bindings-macro/src/lib.rs
```diff
@@ -1241,3 +1242,74 @@ pub fn schema_type(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+/// For example, it could refer to a non-existent table.
+#[proc_macro]
+pub fn filter(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
+    let rls_sql = syn::parse_macro_input!(input with parse_sql);
+
+    let mut hasher = DefaultHasher::new();
+    rls_sql.hash(&mut hasher);
+    let rls_name = format_ident!("rls_{}", hasher.finish());
+
+    let register_rls_symbol = format!("__preinit__20_register_{rls_name}");
```
joshua-spacetime, line 1294: Is it possible to scope this to a block instead of globally? So that we don't have to generate a unique name for each registration?

crates/bindings-macro/src/lib.rs
```diff
@@ -1241,3 +1242,74 @@ pub fn schema_type(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+        #[export_name = #register_rls_symbol]
+        extern "C" fn __register_rls() {
+            spacetimedb::rt::register_row_level_security::<#rls_name>()
+        }
+    };
+
+    let emission = quote! {
+        const _: () = {
+            #generated_describe_function
+        };
```
joshua-spacetime, line 1306: Can we place all definitions and impls associated with this filter in this block?

crates/bindings-macro/src/lib.rs
```diff
@@ -1241,3 +1242,74 @@ pub fn schema_type(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+        #[allow(non_camel_case_types)]
+        struct #rls_name { _never: ::core::convert::Infallible }
```
joshua-spacetime, line 1308: What does `_never: ::core::convert::Infallible` do?

crates/core/src/db/update.rs
joshua-spacetime: What about resolving the rls filters **after** all the other auto-migrate steps?

crates/bindings-macro/src/lib.rs
gefjon: WASM imports and exports always have to be unique - this is going to be in the resulting WASM module's exported symbol table.

crates/bindings-macro/src/lib.rs
gefjon: It makes it so this struct can never be constructed, effectively limiting it to only type-level computations. It's not strictly necessary - the struct is private anyways, and even if someone *could* construct an instance it wouldn't hurt anything - but it's a nice way to communicate the intent that this be a compile-time type name with no runtime representation.

crates/core/src/db/update.rs
mamcx: Done

crates/schema/src/auto_migrate.rs
```diff
@@ -218,6 +224,7 @@ fn auto_migrate_tables(plan: &mut AutoMigratePlan<'_>) -> Result<()> {
                     plan.steps.push(AutoMigrateStep::AddTable(new.key()));
                     Ok(())
                 }
+                // TODO: When we remove tables, we should also remove their dependencies, including row-level security.
```
joshua-spacetime, line 227: Is this TODO still applicable?

crates/schema/src/auto_migrate.rs
```diff
@@ -170,8 +174,10 @@ pub fn ponder_auto_migrate<'def>(old: &'def ModuleDef, new: &'def ModuleDef) ->
     let indexes_ok = auto_migrate_indexes(&mut plan, &new_tables);
     let sequences_ok = auto_migrate_sequences(&mut plan, &new_tables);
     let constraints_ok = auto_migrate_constraints(&mut plan, &new_tables);
+    // Is important that this is the last step, because it needs the list of tables
```
joshua-spacetime, line 177:
```suggestion
    // IMPORTANT: RLS auto-migrate steps must come last,
    // since they assume that any schema changes, like adding or dropping tables,
    // have already been reflected in the database state.
```

crates/expr/src/errors.rs
```diff
@@ -134,6 +134,10 @@ impl UnexpectedType {
 #[error("Duplicate name `{0}`")]
 pub struct DuplicateName(pub String);
 
+#[derive(Debug, Error)]
+#[error("No `TableId` found in `sql` expression")]
+pub struct NoTableId;
```
joshua-spacetime, line 139:
```suggestion
#[derive(Debug, Error)]
#[error("`filter!` does not support column projections; Must return table rows")]
pub struct FilterReturnType;
```

Is this better?

crates/core/src/host/wasm_common/module_host_actor.rs
```diff
@@ -273,9 +274,22 @@ impl<T: WasmInstance> ModuleInstance for WasmModuleInstance<T> {
                     let table_name = &def.name;
                     self.system_logger().info(&format!("Creating table `{table_name}`"));
                     let schema = TableSchema::from_module_def(&self.info.module_def, def, (), TableId::SENTINEL);
-                    stdb.create_table(tx, schema)
+                    stdb.create_table(tx, schema.clone())
```
joshua-spacetime, line 277: Nit: This clone is stale now. You can remove it.

crates/core/src/sql/parser.rs
```diff
@@ -0,0 +1,38 @@
...
+use spacetimedb_expr::expr::RelExpr;
+use spacetimedb_expr::ty::TyCtx;
+use spacetimedb_lib::db::raw_def::v9::RawRowLevelSecurityDefV9;
+use spacetimedb_lib::identity::AuthCtx;
+use spacetimedb_schema::schema::RowLevelSecuritySchema;
+
+pub struct RowLevelExpr {
+    pub sql: RelExpr,
+    pub def: RowLevelSecuritySchema,
+}
```
joshua-spacetime, line 15: Do we need a new type? We don't use the `RelExpr` at all, other than getting the table id from it.

crates/core/src/sql/parser.rs
```diff
@@ -0,0 +1,38 @@
...
+    pub fn build_row_level_expr(
+        stdb: &RelationalDB,
+        tx: &mut MutTxId,
+        rls: &RawRowLevelSecurityDefV9,
+    ) -> Result<Self, TypingError> {
+        let mut ctx = TyCtx::default();
+        let sql = parse_and_type_sub(
+            &mut ctx,
+            &rls.sql,
+            &SchemaViewer::new(stdb, tx, &AuthCtx::for_testing()),
```
joshua-spacetime, line 27: We should use a real `AuthCtx` here.

crates/core/src/db/update.rs
```diff
@@ -221,6 +224,19 @@ fn auto_migrate_database(
             spacetimedb_schema::auto_migrate::AutoMigrateStep::RemoveSchedule(_) => {
                 anyhow::bail!("Removing schedules is not yet implemented");
             }
+            spacetimedb_schema::auto_migrate::AutoMigrateStep::AddRowLevelSecurity(sql_rls) => {
+                system_logger.info(&format!("Adding row-level security `{sql_rls}`"));
+                log::info!("Adding row-level security `{sql_rls}`");
+                let rls = RawRowLevelSecurityDefV9::lookup(plan.new, sql_rls).unwrap();
```
joshua-spacetime, line 230:
```suggestion
                let rls = plan.new.lookup_expect(sql_rls);
```

crates/core/src/db/update.rs
```diff
@@ -221,6 +224,19 @@ fn auto_migrate_database(
...
+                let rls = RowLevelExpr::build_row_level_expr(stdb, tx, rls)?;
+
+                stdb.create_row_level_security(tx, rls.def)?;
+            }
+            spacetimedb_schema::auto_migrate::AutoMigrateStep::RemoveRowLevelSecurity(sql_rls) => {
+                system_logger.info(&format!("Removing-row level security `{sql_rls}`"));
+                log::info!("Removing row-level security `{sql_rls}`");
+                stdb.drop_row_level_security(tx, sql_rls.clone())?;
```
joshua-spacetime, line 238:
```suggestion
                stdb.drop_row_level_security(tx, sql_rls)?;
```

crates/core/src/host/wasm_common/module_host_actor.rs
joshua-spacetime, line 277:
```suggestion
                    stdb.create_table(tx, schema)
```

crates/schema/src/auto_migrate.rs
mamcx: Yes, is for the future, until we have triggers or `CASCADE` support

crates/core/src/db/update.rs
mamcx: Not allowed by borrow checker

crates/core/src/sql/parser.rs
mamcx: I expect we need this when later we execute the `rls` policy

joshua-spacetime (approved):

PR #1851, opened 2024-10-15
commitlog: Introduce epoch

kim:
Adds a field `epoch` typed u64 to the commit header. It signifies the
term number of the leader when writing to a commitlog in a distributed
setting.

It will help detecting and recovering from divergent logs, a situation
which can arise during a leader failover.


Shubham8287 (approved):

cloutiertyler (approved):
This is fine by me, but could you please update the commitlog proposal as a way to document this change?

PR #1854, opened 2024-10-15
Use new token format when generating new id tokens

jsdt:
# Description of Changes

Send tokens with `iss` and `sub` claims, and an identity computed from those. This is compatible with the way we are currently parsing tokens, so it doesn't change much code.

corresponding GitHub label.

2. This is a low risk change that just depends on clients allowing tokens with more fields. When the server parses incoming tokens, we don't require these new fields yet.

# Testing

No testing beyond CI checks here and on the [corresponding private PR](https://github.com/clockworklabs/SpacetimeDBPrivate/pull/1059).


cloutiertyler (approved):
This LGTM

crates/client-api/src/auth.rs
```diff
@@ -58,6 +59,9 @@ impl SpacetimeCreds {
     pub fn decode_token(&self, public_key: &DecodingKey) -> Result<SpacetimeIdentityClaims, JwtError> {
         decode_token(public_key, self.token()).map(|x| x.claims)
     }
+    fn from_signed_token(token: String) -> Self {
+        Self { token }
+    }
     /// Mint a new credentials JWT for an identity.
     pub fn encode_token(private_key: &EncodingKey, identity: Identity) -> Result<Self, JwtError> {
```
cloutiertyler, line 66: Are we unable to remove this function yet?

jsdt: Not yet, because we are using it in some internal functions. We will need to remove it when we change how we are accepting tokens, but it doesn't need to be removed for this PR.

PR #1862, opened 2024-10-15
Limit update() to only work on primary keys

coolreader18:
# Description of Changes

@cloutiertyler and I were talking about this. Makes the semantic mapping for people easier, a row update on the client is an update() on the server. Doesn't make sense for non-pk to have `.update()` but it doesn't trigger an update.

1


gefjon:
Please also update the API proposal

gefjon:
Should we change the API? If updates are only by PK, and a table has at most one PK, then they don't really belong on unique indices, do they? Should it be `ctx.db.my_table().update(new_row)` rather than `ctx.db.my_table.().my_pk_field().update(new_row)`?

coolreader18:
> Or do we want to leave it on the PK index to make it clear how the old row is found during the update?

I think that's beneficial - I just added a section in the doc comment for `update()` about what's considered an update and why it only exists on the primary key. We could probably consider adding `.update()` on the table in the future, if it's enough of a ergonomics thing.

cloutiertyler:
I'm going to try to get this to merge, but I don't want to make the `.update` on the table instead of the column change because I think that's too drastic of an API change at this point in time. It can always be added later too.

coolreader18:
That should be working for the rust module bindings.

cloutiertyler:
Closing in favor of #4279, which re-implements this on top of current master and extends the change to C# and TypeScript module bindings as well.

Issue #1847, opened 2024-10-11
Implement `spacetime login`


cloutiertyler:
Steps:

1.  The CLI starts a websocket connection to `https://spacetimedb.com/login-cli`
2. It passes in a random nonce that it created to the websocket
3. Zeke figures out how to open the default browser and set the url to be `https://spacetimedb.com/login-cli?nonce={the nonce}`
4. The user logs in on the browser
5. The API backend sends the session token to the CLI over the websocket connection
6. The CLI provides the session token on every subsequent request to the website API
7. The CLI hits the `/v1/identity` endpoint to receive a SpacetimeDB token
8. The CLI provides this token to all SpacetimeDB requests (unless they are flagged `--anonymous`.

bfops:
ChatGPT:
```rust
fn open_browser(url: &str) {
    let open_command = match env::consts::OS {
        "windows" => Command::new("cmd").args(&["/C", "start", url]),
        "macos" => Command::new("open").arg(url),
        "linux" => Command::new("xdg-open").arg(url),
        _ => {
            eprintln!("Unsupported OS");
            return;
        }
    };

    if let Err(e) = open_command.status() {
        eprintln!("Failed to open browser: {}", e);
    }
}
```

PuruVJ:
I saw that we also pull in this crate somewhere in spacetimedb https://crates.io/crates/webbrowser, so caninstead just reuse it

PR #1863, opened 2024-10-16
CLI - Implement new `login` flow

bfops:
# Description of Changes

Add the `spacetime login` command, which gets a login token from the web server.

This currently just implements the command and adds the returned session token into the config file. It does **not** use the token anywhere!

Getting the token requires https://github.com/clockworklabs/spacetime-web/pull/464.
Using the token requires https://github.com/clockworklabs/SpacetimeDB/pull/1854.

Closes https://github.com/clockworklabs/SpacetimeDB/issues/1847.

This breaks the current flow for using any CLI commands that interact with a server. They will now require logging in first.

# Testing


bfops:
Combining this into https://github.com/clockworklabs/SpacetimeDB/pull/1878

PR #1864, opened 2024-10-16
Durability Provider

Shubham8287:
# Description of Changes
- Added `fn leader` to `NodeDelegate`, to abstract `HostController`.
- Added `dyn DurabilityProvider` in `HostController`.
- Update `message::Database` to include `ReplicationConfig`


crates/core/src/host/host_controller.rs
```diff
@@ -40,6 +41,9 @@ type HostCell = Arc<AsyncRwLock<Option<Host>>>;
 /// The registry of all running hosts.
 type Hosts = Arc<Mutex<IntMap<u64, HostCell>>>;
 
+/// A future that resolves to a [`durability::Durability`] instance.
+pub type DynDurabilityFut = BoxFuture<'static, anyhow::Result<(Arc<dyn Durability<TxData = Txdata>>, DiskSizeFn)>>;
```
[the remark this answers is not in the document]
Shubham8287: It is dyn because it needs to be wrapped around Futture, I needs to be future because we do `get_or_launch_module_host` everywhere which require this as param.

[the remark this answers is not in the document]
kim: I played around with this a bit, and I think it could work more elegantly if we have a trait like `ExternalStorage`:

```rust
pub type DurabilityAndMetrics = (Arc<dyn Durability<TxData = Txdata>>, DiskSizeFn);

// dumb name...
pub trait ExternalDurability {
    // return `None` if this replica should be in-memory only
    async fn get_durability(&self, replica_id: u64) -> anyhow::Result<Option<DurabilityAndMetrics>>;
}
```

We then store `Arc<dyn ExternalDurability>` in `HostController` instead of passing it in every method call.

crates/core/src/host/host_controller.rs
```diff
@@ -205,8 +209,15 @@ impl HostController {
     ///
     /// See also: [`Self::get_module_host`]
     #[tracing::instrument(skip_all)]
-    pub async fn get_or_launch_module_host(&self, database: Database, replica_id: u64) -> anyhow::Result<ModuleHost> {
-        let mut rx = self.watch_maybe_launch_module_host(database, replica_id).await?;
+    pub async fn get_or_launch_module_host(
+        &self,
+        database: Database,
+        replica_id: u64,
+        durability: DynDurabilityFut,
```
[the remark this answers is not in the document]
Shubham8287: Oh, I was under the impression that `Storage::Memory` will always be used for testing only. Makes sense.

crates/core/src/host/host_controller.rs
```diff
@@ -685,15 +714,17 @@ impl Host {
                 None,
             )?,
             db::Storage::Disk => {
-                let (durability, disk_size_fn) = relational_db::local_durability(&db_path).await?;
+                let durability = durability.await?;
+                // History trait does not let use `DynDurabilityFut` directly
```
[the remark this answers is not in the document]
kim: Is this comment still relevant?

I think the behavior as written is correct -- the history is always the local history, while the durability can be something else.

crates/standalone/src/lib.rs
```diff
@@ -107,6 +130,28 @@ impl spacetimedb_client_api::NodeDelegate for StandaloneEnv {
...
+    async fn leader(&self, database_id: u64) -> anyhow::Result<Option<Host>> {
+        let leader = match self.control_db.get_leader_replica_by_database(database_id) {
+            Some(leader) => leader,
+            None => return Ok(None),
+        };
+
+        let database = self
+            .control_db
+            .get_database_by_id(database_id)?
+            .ok_or_else(|| anyhow::anyhow!("Database {} has no associated database", database_id))?;
```
kim, line 143:
```suggestion
            .with_context(|| format!("Database {} not found", database_id))?;
```

crates/client-api/src/lib.rs
```diff
@@ -20,13 +27,110 @@ pub mod util;
...
     fn client_actor_index(&self) -> &ClientActorIndex;
 
     type JwtAuthProviderT: auth::JwtAuthProvider;
     fn jwt_auth_provider(&self) -> &Self::JwtAuthProviderT;
+    /// Return the leader [`Host`] of `database_id`.
+    ///
+    /// Returns `None` if the current leader is not hosted by this node.
+    /// The [`Host`] is spawned implicitly if not already running.
+    async fn leader(&self, database_id: u64) -> anyhow::Result<Option<Host>>;
+    fn module_logs_dir(&self, replica_id: u64) -> ModuleLogsDir;
```
kim, line 42: How do you feel about adding the following to `Host` instead:

```rust
pub async fn module_logs(&self, num_lines: Option<u32>) -> String {
    let logs_dir = self.host_controller.data_dir.replica(self.replica_id).module_logs();
    DatabaseLogger::read_latest(logs_dir, num_lines).await
}
```

kim: (Not sure why we're reading all of those into a single big string instead of a stream of lines, but that'd be for another time)

crates/core/src/messages/control_db.rs
```diff
@@ -36,6 +36,15 @@ pub struct Database {
     ///
     /// Updating the database's module will **not** change this value.
     pub initial_program: Hash,
+
+    /// Replication configuration for the database.
+    pub replication: ReplicationConfig,
+}
+
+#[derive(Clone, Default, PartialEq, Serialize, Deserialize)]
+pub struct ReplicationConfig {
```
kim, line 45: I might be missing something, but I think we can omit this here -- we only need the type from the control _module_.

kim (approved):
Looks good!

crates/client-api/src/lib.rs
Shubham8287: > How do you feel about adding the following to Host instead:

I think we want to fetch these module_logs without actually launching host if it is not already up, actually depends on `follow` - https://github.com/clockworklabs/SpacetimeDB/blob/master/crates/client-api/src/routes/database.rs#L457

PR #1867, opened 2024-10-16
Fix compression threshold bug

jdetter:
# Description of Changes

I discovered this issue when testing deliveryz. The packets that are supposed to be compressed aren't compressed and the ones that aren't supposed to be compressed are compressed. We had a simple logic inversion here which I've corrected.

corresponding GitHub label.

Not breaking

1

# Testing

- [x] I ran deliveryz against this PR and the compression is now correct


jdetter:
This has been incorporated into `v0.12.0-beta-hotfix3`

PR #1869, opened 2024-10-17
C# codegen making state relative to a DbConnection

lcodes:
# Description of Changes

Make the static fields indexing primary keys instance fields.

# Testing

Codegen compiles. Same code as before but going through a `this` pointer instead of a `static` .

Context from Boppy: The Unity testsuite exercises this feature in the testsuite: https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/pull/168

### DIFF
--- a/crates/cli/src/subcommands/generate/csharp.rs
+++ b/crates/cli/src/subcommands/generate/csharp.rs
@@ -412,6 +412,14 @@ fn autogen_csharp_access_funcs_for_struct(
             "public readonly ref struct {csharp_field_name_pascal}UniqueIndex"
         );
         indented_block(output, |output| {
+            writeln!(output, "readonly {csharp_table_name}Handle Handle;");
+            write!(
+                output,
+                "internal {csharp_field_name_pascal}UniqueIndex({csharp_table_name}Handle handle) => "
+            );
+            writeln!(output, "Handle = handle;");
+            writeln!(output);
+
             writeln!(
                 output,
                 "public {struct_name_pascal_case}? Find({csharp_field_type} value)"
@@ -419,7 +427,7 @@ fn autogen_csharp_access_funcs_for_struct(
             indented_block(output, |output| {
                 writeln!(
                     output,
-                    "{csharp_field_name_pascal}_Index.TryGetValue(value, out var r);"
+                    "Handle.{csharp_field_name_pascal}_Index.TryGetValue(value, out var r);"
                 );
                 writeln!(output, "return r;");
             });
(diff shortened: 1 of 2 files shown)


jdetter (approved):
Tested via https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/pull/176

Code looks good to me, thanks @lcodes!

crates/cli/tests/snapshots/codegen__codegen_csharp.snap
```diff
@@ -672,27 +672,33 @@ namespace SpacetimeDB
 
 			public readonly ref struct IdUniqueIndex
 			{
+				readonly PkMultiIdentityHandle Handle;
```
RReverser, line 675: The logic seems unnecessarily complicated.

Table handle stores the Dictionary<> caches.
Table handles also store instances of index structs like this IdUniqueIndex.
Then, index structs store a reference back to the parent table handles.
And then, index structs only access said relevant cache via those table handles.

Instead, let's move the Dictionary<> cache directly into the index struct. This way we'll encapsulate the cache data inside the relevant index struct, and won't need this kind of 3-way access just to retrieve values from the cache from inside the index.

(originally posted at https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/pull/168#discussion_r1810863587 )

RReverser (changes requested):

jdetter (changes requested):
Ingvar left some requested changes here: https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/pull/168#discussion_r1810863587

I'm temporarily removing my approval so that he has an opportunity to review here

crates/cli/src/subcommands/generate/csharp.rs
```diff
@@ -407,28 +407,28 @@ fn autogen_csharp_access_funcs_for_struct(
...
         indented_block(output, |output| {
+            write!(
+                output,
+                "internal readonly Dictionary<{csharp_field_type}, {csharp_table_name}> __Cache = new(16);"
+            );
+            writeln!(output);
+
             writeln!(
                 output,
                 "public {struct_name_pascal_case}? Find({csharp_field_type} value)"
```
RReverser, line 420: If we're using a class, might as well extract those into a shared base class, but should be good to go as-is for now too.

crates/cli/src/subcommands/generate/csharp.rs
```diff
@@ -407,28 +407,28 @@ fn autogen_csharp_access_funcs_for_struct(
...
         };
-        writeln!(
-            output,
-            "public readonly ref struct {csharp_field_name_pascal}UniqueIndex"
-        );
+        writeln!(output, "public class {csharp_field_name_pascal}UniqueIndex");
         indented_block(output, |output| {
+            write!(
+                output,
+                "internal readonly Dictionary<{csharp_field_type}, {csharp_table_name}> __Cache = new(16);"
```
RReverser, line 414: Why do we need `__` - can it clash with anything provided by the user? I don't think so because we don't add any user-generated names in this class, do we?

RReverser (approved):
Minor nits, but fine to go as-is too.

PR #1877, opened 2024-10-17
CLI - Fix `spacetime server edit` failing on renames

bfops:
# Description of Changes

`spacetime server edit` had a bug during renames, where it would try to access the server config by the old name, causing an error.

This must have been broken for a long time. The code has been in this state for a while, looks like about a year from the git history (a1e9984840ff60d77023f04927e50929cc7e0102).

No. Bugfix.

1

# Testing

I added a smoketest.

### DIFF
--- a/crates/cli/src/subcommands/server.rs
+++ b/crates/cli/src/subcommands/server.rs
@@ -379,6 +379,7 @@ pub async fn exec_edit(mut config: Config, args: &ArgMatches) -> Result<(), anyh
     }
 
     let (old_nick, old_host, old_proto) = config.edit_server(server, new_nick, new_host, new_proto)?;
+    let server = new_nick.unwrap_or(server);
 
     if let (Some(new_nick), Some(old_nick)) = (new_nick, old_nick) {
         println!("Changing nickname from {} to {}", old_nick, new_nick);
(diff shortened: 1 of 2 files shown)


smoketests/tests/servers.py
```diff
@@ -23,3 +23,15 @@ def test_servers(self):
 
         out = self.spacetime("server", "fingerprint", "-s", "localhost")
         self.assertIn("Fingerprint is unchanged for server localhost", out)
+
+    def test_edit_server(self):
```
bfops, line 27: This test was originally part of https://github.com/clockworklabs/SpacetimeDB/pull/1845/files#diff-a953574fc278eedbeb3eb9090a80a7fcc97d5ea94a1e5612b25d54d877c43a4a, but I moved it here to test the current behavior.

crates/cli/src/subcommands/server.rs
```diff
@@ -379,6 +379,7 @@ pub async fn exec_edit(mut config: Config, args: &ArgMatches) -> Result<(), anyh
     }
 
     let (old_nick, old_host, old_proto) = config.edit_server(server, new_nick, new_host, new_proto)?;
+    let server = new_nick.unwrap_or(server);
```
bfops, line 382: this causes the value of `server` to be correct for any code below (i.e. any code after `config.edit_server`).

cloutiertyler (approved):
Good catch

PR #1878, opened 2024-10-18
CLI - Login flow

bfops:
# Description of Changes

Add the `spacetime login` command, which gets a login token from the web server.

I have **removed** `IdentityConfig` in this PR, because we don't want to be storing identities anymore. I've also removed almost all of the `--identity` CLI parameters, because we want users to be going through `spacetime login` to switch identities, rather than "just in time" via CLI params.

Closes https://github.com/clockworklabs/SpacetimeDB/issues/1847.

This breaks existing stored identities in the config file.

3

# Testing

- Get a token using `spacetime login --host [web-host]` (a local one works fine)
- Run a local server and get a token using `spacetime login --direct local`
- With either token, publish a module, call a function, and print logs.

# Future work
- Consider how `spacetimedb_token` should look to be backwards-compatible with the future changes (e.g. we may want to allow users to `login` to switch between multiple accounts - what will we cache and where?)
- Move `spacetimedb_token` out of config so we're no longer storing sensitive tokens in a human-edited file


crates/testing/src/sdk.rs
```diff
@@ -145,7 +145,15 @@ fn compile_module(module: &str) -> String {
 // module as a separate clean database instance for isolation purposes.
 fn publish_module(wasm_file: &str) -> String {
     let name = random_module_name();
-    invoke_cli(&["publish", "--server", "local", "--bin-path", wasm_file, &name]);
+    invoke_cli(&[
+        "publish",
+        "--anonymous",
```
bfops, line 150: added `--anonymous` so we don't need to log in

crates/cli/src/subcommands/login.rs
```diff
@@ -0,0 +1,169 @@
...
+                .long("token")
+                .conflicts_with("host")
+                .conflicts_with("new")
+                .help("Bypass the login flow and use a login token directly"),
+        )
+        .arg(
+            Arg::new("refresh-cache")
+                .long("refresh-cache")
+                .action(ArgAction::SetTrue)
+                .help("Clear the cached tokens and re-fetch them"),
```
cloutiertyler, line 26: Tokens? Does this apply to the website session token or the spacetimedb token?

cloutiertyler (approved):
This looks great. Love the red. I left two small comments.

crates/cli/src/subcommands/server.rs
```diff
@@ -59,13 +59,6 @@ fn get_subcommands() -> Vec<Command> {
...
             )
-            .arg(
-                Arg::new("delete-identities")
-                    .help("Also delete all identities which apply to the server")
-                    .long("delete-identities")
-                    .short('I')
-                    .action(ArgAction::SetTrue),
-            )
             .arg(common_args::yes()),
         Command::new("fingerprint")
```
cloutiertyler, line 63: Are we keeping fingerprints?

crates/cli/src/subcommands/dns.rs
```diff
@@ -31,41 +31,48 @@ fn get_subcommands() -> Vec<Command> {
...
             )
             .after_help("Run `spacetime dns lookup --help` for more detailed information"),
         Command::new("reverse-lookup")
             .about("Returns the domains for the provided database identity")
-            .arg(Arg::new("database-identity").required(true).help("The database identity you would like to find all of the known domains for"))
+            .arg(
+                Arg::new("database-identity")
+                    .required(true)
+                    .help("The database identity you would like to find all of the known domains for"),
+            )
```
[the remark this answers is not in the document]
bfops: One or two commands still have --identity, so I would probably keep this distinct for now.

crates/cli/src/subcommands/login.rs
bfops: It clears both. It was useful for me while testing since we kept updating server-side stuff, but I'm not particularly attached to any specific behavior (or necessarily keeping this at all).

crates/cli/src/subcommands/server.rs
bfops: I removed --identity params and IdentityConfig in this PR because they no longer functioned. I know we plan to remove fingerprints, but maybe we should do this in a followup PR since this one is already pretty big?

cloutiertyler: Sure thing.

bfops:
(Going to fix `Internal tests` before merging)

crates/cli/src/subcommands/login.rs
```diff
@@ -167,3 +186,19 @@ async fn spacetimedb_login(remote: &str, web_session_id: &String) -> Result<Stri
...
 }
+
+#[derive(Debug, Clone, Deserialize)]
+struct LocalLoginResponse {
+    pub token: String,
+}
+
+async fn spacetimedb_direct_login(host: &str) -> Result<String, anyhow::Error> {
+    // Users like to provide URLs with trailing slashes, which can cause issues due to double-slashes in the routes below.
+    let host = host.trim_end_matches('/');
```
cloutiertyler, line 197: I believe there is a way with `PathBuf` to just have this do this for you. https://doc.rust-lang.org/std/path/struct.Path.html#method.join I believe this merges trailing slashes.

crates/cli/src/subcommands/login.rs
```diff
@@ -7,16 +7,22 @@ use webbrowser;
...
-            Arg::new("host")
-                .long("host")
+            Arg::new("auth-host")
+                .long("auth-host")
                 .default_value("https://spacetimedb.com")
                 .help("Fetch login token from a different host"),
         )
+        .arg(
+            Arg::new("server")
+                .long("direct")
```
cloutiertyler, line 17: Is "server" a typo? Why is "direct" the long of "server". Also I think perhaps something like "cluster-issued-login" or "server-issued-login" is a more apt name.

bfops: In the CLI helptext, the arg name is used as the name of the value, whereas the `long` is the name of the param itself:

```
Options:
      --auth-host <auth-host>      Fetch login token from a different host [default: https://spacetimedb.com]
      --direct <server>            Log in to a SpacetimeDB server directly, without going through a global auth server
      --token <spacetimedb-token>  Bypass the login flow and use a login token directly
      --refresh-cache              Clear the cached tokens and re-fetch them
  -h, --help                       Print help
```

I'm happy with a more descriptive name!

crates/cli/src/subcommands/login.rs
bfops: I don't believe pathbuf is suitable for URLs unless we keep the protocol separate until the very end (`http://foo.com` is parsed differently as a path than as a url)

crates/cli/src/subcommands/login.rs
bfops: (renamed to `--server-issued-login`)

cloutiertyler (approved):
This LGTM. I might want to bikeshed `server-issued-login` before final 1.0 release, but it is more descriptive now already.

PR #1879, opened 2024-10-18
Directory structure impl

coolreader18:
# Description of Changes

Implements the new directory structure proposal, using strongly-typed newtypes to represent paths. All places in the database/server/cli that previously would have been passing around `PathBuf`s and `.join()`ing them now instead pass around these types, calling methods on them that internally call `.join()` and return another strongly-typed child node in the directory hierarchy.

Changes from the proposal implemented here:
- the replicas directory in the data-dir is flattened from having database files be at `replicas/$database_identity/$replica_id/database/...`, to just `replicas/$replica_id/...`
- all environment variables are removed, besides a few (`SPACETIMEDB_DISABLE_DISK_LOGGING`, `SPACETIMEDB_TRACY`, `SPACETIMEDB_FLAMEGRAPH{,_PATH}`) that seem to be just for debugging purposes. I can turn them into command-line args if that's desired.
- I split the `~/.spacetime` directory across the xdg base directories on unix-likes (so `~/.config/spacetime`, `~/.local/share/spacetime`, etc). On windows it's moved to `%LocalAppData%/SpacetimeDB`
- Split the `config.toml` into `cli.toml` and data-dir `config.toml`, and implemented `data-dir/metadata.toml`.
- `data-dir/wasmtime_cache` -> `data-dir/cache/wasmtime`
- module logs are dated instead of just `0.log` (though a proper logrolling mechanism hasn't been implemented - that likely needs further work), database logs are in `data-dir/logs` instead of `/var/logs/spacetime`.
- `data-dir/spacetime.pid` lockfile, so multiple processes can't use the same data-dir at the same time.
- All the root paths (`cli-{bin-file,bin-dir,config-dir}`, `data-dir`) are now bundled in the `SpacetimePaths` type. server processes like standalone only interact with the data-dir.
- the CLI has 2 new flags for the base command itself, not on subcommands - `spacetime --root-dir` and `spacetime --config-path`, to set a custom root directory for the spacetime installation and to specify a custom `cli.toml`, respectively. this was not in the proposal itself but something like this had to be implemented for our tests to still work.

Changes that are not directly related to that, but facilitate it:
- the `spacetimedb::auth::JwtKeys` type, which deduplicates jwt-key-file code across our server implementations. This is something I have less good motivation for, and I could try and revert if desired.
- `Host::try_init` now takes `&HostController`, instead of like 5 different args that were just fields on `HostController` passed individually. I was adding a 6th arg and clippy was getting mad at me for having too many parameters.
- the wasmtime `Engine` and `Linker` are now stored in a `WasmtimeRuntime` struct. Since we have to give `Engine` the `{data-dir}/cache/wasmtime` path, it made more sense for that to just get eagerly initialized by `HostController`, who has access to the data-dir path, instead of threading `data_dir` all the way through to `wasmtime::make_actor`.
- `relational_db::test_utils::TempReplicaDir`, which merges a `TempDir` and `spacetimedb_paths::server::ReplicaDir`.
- Added `spacetimedb::startup::TracingOptions` got added - this is so we wouldn't read environment variables in `core` anymore.


crates/paths/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "spacetimedb-paths"
+version.workspace = true
+edition.workspace = true
+rust-version.workspace = true
```
bfops, line 5: minor thing - can you add a description and license file for this? Otherwise it will fail to publish. (I think our convention for other crates is to just use an empty license file called `LICENSE`).

crates/bench/src/spacetime_module.rs
```diff
@@ -77,9 +78,8 @@ impl BenchDatabase for SpacetimeModule {
         let module = runtime.block_on(async {
             // We keep a saved database at "crates/bench/.spacetime".
             // This is mainly used for caching wasmtime native artifacts.
-            BENCHMARKS_MODULE
-                .load_module(config, Some(Path::new(env!("CARGO_MANIFEST_DIR"))))
-                .await
+            let path = RootDir(Path::new(env!("CARGO_MANIFEST_DIR")).join(".spacetime"));
+            BENCHMARKS_MODULE.load_module(config, Some(&path)).await
```
mamcx, line 82: Ideally, no code outside the handler of 'paths' should construct the layout.

ie: `Paths::from_cargo_manifest`

crates/cli/src/subcommands/server.rs
```diff
@@ -425,26 +432,10 @@ pub async fn exec_edit(mut config: Config, args: &ArgMatches) -> Result<(), anyh
...
-
-    let control_node_dir = stdb_path("control_node");
-    let worker_node_dir = stdb_path("worker_node");
-    if control_node_dir.exists() || worker_node_dir.exists() {
-        if control_node_dir.exists() {
-            println!("Control node database path: {}", control_node_dir.to_str().unwrap());
-        } else {
-            println!("Control node database path: <not found>");
-        }
+    let data_dir = args.get_one::<PathBuf>("data_dir").unwrap();
```
mamcx, line 435: I think it is better if anything that is related to the `Layout` is described by constants instead of hardcode strings.

ie: `Paths::DATA_DIR`

crates/core/src/db/relational_db.rs
```diff
@@ -2434,8 +2436,8 @@ mod tests {
             row_ty,
         }));
         {
-            let clog =
-                Commitlog::<()>::open(dir.path().join("clog"), Default::default()).expect("failed to open commitlog");
+            let clog = Commitlog::<()>::open(ReplicaPath(dir.path().into()).commit_log(), Default::default())
```
mamcx, line 2439: This should also be done in the `Paths`.

crates/core/src/startup.rs
```diff
@@ -55,29 +80,31 @@ fn configure_tracing() {
...
-    let write_to = if disable_disk_logging {
-        BoxMakeWriter::new(std::io::stdout)
+    let write_to = if let Some(logs_dir) = opts.disk_logging {
+        let roller = rolling::Builder::new()
+            .filename_prefix(format!("spacetimedb-{}", opts.edition))
+            .filename_suffix("log")
+            .build(logs_dir)
+            .unwrap();
+        // TODO: syslog?
+        BoxMakeWriter::new(std::io::stdout.and(roller))
```
mamcx, line 90: This also should be build by the `Paths`

crates/paths/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+license-file = "LICENSE"
+description = "The spacetimedb directory tructure, represented as a type hierarchy"
```
bfops, line 7:
```suggestion
description = "The spacetimedb directory structure, represented as a type hierarchy"
```

crates/paths/src/lib.rs
```diff
@@ -0,0 +1,87 @@
...
+    /// Get the default directories for the currrent platform.
+    ///
+    /// Returns an error if the platform director(y/ies) cannot be found.
+    pub fn platform_defaults() -> anyhow::Result<Self> {
+        #[cfg(windows)]
+        {
+            let data_dir = dirs::data_local_dir().ok_or_else(|| anyhow::anyhow!("Could not find LocalAppData"))?;
+            let root_dir = RootDir(data_dir.joined("SpacetimeDB"));
+            Ok(Self::from_root_dir(&root_dir))
+        }
```
bfops, line 56: the `dirs` crate says that it works with the XDG directories under Linux - is there a reason we have to manually switch to XDG directories when not on windows?

bfops: also does this work properly on MacOS?

crates/bench/src/spacetime_module.rs
coolreader18: I feel like this case doesn't matter much - to me, this is more or less equivalent to a user passing `--root-dir=$CARGO_MANIFEST_DIR/.spacetime` on the command line. I can add a comment addressing this.

crates/cli/src/subcommands/server.rs
coolreader18: This is just the clap argument though, it's defined in the same file and doesn't matter anywhere else. It could be renamed to `"foo"` and it wouldn't be observable to the user at all.

crates/paths/src/lib.rs
coolreader18: I'm not able to test on macOS, but it should. The reason we use this (~~and there's a comment saying this right above here~~ oops - there used to be) is because `dirs` doesn't use the xdg dirs on macOS, which the proposal says we should.

crates/paths/src/lib.rs
```diff
@@ -0,0 +1,321 @@
...
+    #[test]
+    fn custom() {
+        let root = Path::new("/custom/path");
+        let paths = SpacetimePaths::from_root_dir(&RootDir(root.to_owned()));
+        assert_eq!(paths.cli_config_dir.0, root.join("config"));
+        assert_eq!(
+            paths.cli_bin_file.0,
+            root.join("spacetime.exe").with_extension(EXE_EXTENSION)
+        );
+        assert_eq!(paths.cli_bin_dir.0, root.join("bin"));
```
mamcx, line 318: Why assume `custom` is for Windows? 

BTW I have [this solution](https://github.com/clockworklabs/SpacetimeDB/blob/79d1654add17a3a99e80dfd5cc42a90c11ba3507/crates/platforms/src/directories.rs#L129) to make it work everywhere.

mamcx:
Question: Is for a later pr the changes to the scripts:

* `run_standalone_temp.sh`
* `run_cluster_temp.sh`

?

crates/paths/src/lib.rs
coolreader18: Because that's what the proposal is - it doesn't say it outright, but it's directly equivalent to a RootDir based at `%LocalAppData%/SpacetimeDB`. So, why duplicate the implementation if it's no different?

coolreader18:
I think so - someone mentioned that they aren't really used anymore, anyway.

bfops:
@coolreader18 Two things:

1. Can you update the PR description to have all the usual details?
2. Is it possible/reasonable to split this into one PR that does the internal refactor and switches everything to use strongly-typed paths, and a separate PR that changes the user-facing behavior and moves the files into the new paths? I'm having trouble figuring out what's happening where, because some files have huge rewrites that are mostly-independent of the actual files moving. It would be easier to review if there were a clear split between functional/non-functional changes.

gefjon (approved):
Many thanks for the in-depth PR description. That makes review much easier.

crates/paths/src/server.rs
```diff
@@ -0,0 +1,176 @@
...
+            }
+            Err(e) => return Err(e.into()),
+        }
+        let mut pidfile = PidFile { file, path };
+        pidfile.file.set_len(0)?;
+        write!(pidfile.file, "{}", std::process::id())?;
+        Ok(pidfile)
+    }
+
+    pub fn replica(&self, replica_id: u64) -> ReplicaDir {
```
jsdt, line 52: What's the motivation for having multiple replica directories?

jsdt (commented):
This seems like an improvement to me, though it would be a good time to add some more comments. I also wrote a few unit tests that you could add: https://github.com/clockworklabs/SpacetimeDB/compare/noa/directory-structure...jsdt/lock-test

crates/paths/src/server.rs
```diff
@@ -0,0 +1,176 @@
...
+            Err(e) if e.kind() == io::ErrorKind::WouldBlock => {
+                let mut s = String::new();
+                let pid = file.read_to_string(&mut s).ok().and_then(|_| s.trim().parse().ok());
+                return Err(PidFileError::Exists { pid });
+            }
+            Err(e) => return Err(e.into()),
+        }
+        let mut pidfile = PidFile { file, path };
+        pidfile.file.set_len(0)?;
+        write!(pidfile.file, "{}", std::process::id())?;
```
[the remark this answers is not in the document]
coolreader18: `File::flush()` is actually a no-op, but it's probably good practice anyway.

crates/paths/src/server.rs
coolreader18: A replica is just what spacetime calls a database instance now, I guess - so each replica directory contains the database for a separate module.

crates/paths/src/standalone.rs
```diff
@@ -0,0 +1,21 @@
+use crate::server::ServerDataDir;
+use crate::utils::path_type;
+
+mod sealed {
+    pub trait Sealed: AsRef<std::path::Path> {}
+}
```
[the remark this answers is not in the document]
coolreader18: Fair enough

crates/standalone/src/subcommands/start.rs
```diff
@@ -50,170 +56,146 @@ impl ProgramMode {
...
-                \n\tSPACETIMEDB_TRACY: Set to 1 to enable Tracy profiling.\
-                \n\nWarning: If you set a value on the command line, it will override the value set in the environment variable.")
+        // .long_about("Starts a standalone SpacetimeDB instance. This command recognizes the following environment variables: \
+        //         \n\tSPACETIMEDB_LOG_CONFIG: The path to the log configuration file. \
+        //         \n\tSPACETIMEDB_LOGS_PATH: The path to the directory that should contain logs for SpacetimeDB. \
+        //         \n\tSTDB_PATH: The path to the directory that should contain the database files for SpacetimeDB. \
+        //         \n\tSPACETIMEDB_JWT_PUB_KEY: The path to the public jwt key for verifying identities. \
+        //         \n\tSPACETIMEDB_JWT_PRIV_KEY: The path to the private jwt key for issuing identities. \
+        //         \n\tSPACETIMEDB_TRACY: Set to 1 to enable Tracy profiling.\
+        //         \n\nWarning: If you set a value on the command line, it will override the value set in the environment variable.")
```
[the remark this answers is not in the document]
coolreader18: Those are the old env vars that we don't read anymore - I figured it might be nice to leave for future reference, but I guess that's what commit history is for.

crates/paths/src/utils.rs
```diff
@@ -0,0 +1,122 @@
...
+/// }
+/// ```
+#[macro_export]
+macro_rules! path_type {
+    ($(#[doc = $doc:literal])* $(#[non_exhaustive($non_exhaustive:tt)])? $name:ident) => {
+        $(#[doc = $doc])*
+        #[derive(Clone, Debug, $crate::__serde::Serialize, $crate::__serde::Deserialize)]
+        #[serde(transparent)]
+        #[cfg_attr(all($($non_exhaustive)?), non_exhaustive)]
+        pub struct $name(pub std::path::PathBuf);
```
bfops, line 37: nit: I don't love the `.0` pattern for unwrapping these paths. Imo a `.untyped()` or `.to_pathbuf()` or something would be clearer.

PR #1882, opened 2024-10-21
Handle JWT tokens with oidc providers

jsdt:
# Description of Changes

This handles JWT tokens from other providers, as long as they host an openid config and public keys.

Tokens with the issuer "localhost" will go through the older flow.

There are some things that should be improved in future PRs:
 1. This uses a global JWK cache. The cache should probably be tied to the `NodeDelegate`.
 2. This could use more logging and metrics around key fetching and failures.
 3. This is using forks of two jwt/jwk libraries, which we should replace longer term.

4. This doesn't change much aside from including more tokens that can be accepted.

# Testing

This has unit tests that spin up a server hosting public keys to test fetching and validation.


gefjon (approved):
Needs rebase, fmt, clippy, insta and to pass tests, but I trust that you'll get all those things done. Code & logic looks completely reasonable.

crates/client-api/src/auth.rs
```diff
@@ -120,6 +158,51 @@ impl SpacetimeAuth {
...
+            audience: vec!["spacetimedb".to_string()],
+        };
+        let id = claims.id();
+        let token = claims.encode_and_sign(&kp.private_key)?;
+
+        let decoded = identity::decode_token(&kp.public_key, &token)?;
+        assert_eq!(decoded.claims.identity, id);
+        Ok(())
+    }
+}
```
cloutiertyler, line 204: Thank you for the test.

crates/core/src/auth/identity.rs
```diff
@@ -8,6 +8,8 @@ use std::collections::HashSet;
 use std::time::Duration;
 use std::time::SystemTime;
 
+use super::token_validation::TokenValidationError;
+
 #[serde_with::serde_as]
 #[derive(Debug, Serialize, Deserialize)]
 pub struct SpacetimeIdentityClaims {
```
cloutiertyler, line 15: Is this still used? Is this being removed in another PR?

crates/core/src/auth/identity.rs
```diff
@@ -20,6 +22,105 @@ pub struct SpacetimeIdentityClaims {
     pub exp: Option<SystemTime>,
 }
 
+// The new token format that we are sending out.
+#[serde_with::serde_as]
+#[derive(Debug, Serialize, Deserialize)]
+pub struct SpacetimeIdentityClaims2 {
+    #[serde(rename = "hex_identity")]
+    pub identity: Identity,
```
cloutiertyler, line 30: Do we still want to keep the `hex_identity` at all? It's derived from the subject and issuer, right? I suppose it doesn't hurt, but if we do keep it around it should probably be verified to be correctly computed by the subject and issuer.

I would advocate for just removing it.

cloutiertyler: If we do keep it, I think it should be serialized as `id` or `identity` rather than `hex_identity`

Cargo.toml
```diff
@@ -169,7 +169,7 @@ indicatif = "0.16"
 insta = { version = "1.21.0", features = ["toml"] }
 is-terminal = "0.4"
 itertools = "0.12"
-jsonwebtoken = { version = "8.1.0" }
+jsonwebtoken = { git = "https://github.com/jsdt/jsonwebtoken.git", rev = "7f0cef63c74f58dfa912e88da844932ba4c71562"}
```
cloutiertyler, line 172: NIT: We should move this fork under Clockwork Labs in another PR.

jsdt: Either that or switch to a better jwt library.

crates/core/src/auth/identity.rs
jsdt: I'm not sure, but I can add that to the list of clean up needed to follow up on this PR.

crates/core/src/auth/identity.rs
jsdt: This does verify that the identity matches. I'm not sure if we need it to be in the token, but we should send the identity to clients to avoid needing to compute it on the client side.

crates/client-api/src/auth.rs
```diff
@@ -94,19 +100,51 @@ pub struct SpacetimeAuth {
...
-            rng.fill(&mut random_bytes);
-
-            let preimg = [b"clockworklabs:", &random_bytes[..]].concat();
-            Identity::from_hashing_bytes(preimg)
+        // Generate claims with a random subject.
+        let claims = TokenClaims {
+            issuer: ctx.local_issuer(),
+            subject: Uuid::new_v4().to_string(),
+            // Placeholder audience.
+            audience: vec!["spacetimedb".to_string()],
```
cloutiertyler, line 139: NIT: Not sure how I feel about this, but I don't think it harms us

crates/core/src/auth/token_validation.rs
```diff
@@ -0,0 +1,613 @@
...
+            let validated_claims = validator.validate_token(&token).await?;
+            assert_eq!(validated_claims.issuer, issuer);
+            assert_eq!(validated_claims.subject, subject);
+            assert_eq!(validated_claims.identity, Identity::from_claims(&issuer, subject));
+        }
+
+        let invalid_token = jsonwebtoken::encode(&header, &orig_claims, &invalid_kp.private_key)?;
+        assert!(validator.validate_token(&invalid_token).await.is_err());
+        //assert!(OidcTokenValidator.validate_token(&invalid_token).await.is_err());
+        // tokio::spawn(server);
```
cloutiertyler, line 599: NIT: remove comments

crates/core/src/auth/token_validation.rs
```diff
@@ -0,0 +1,613 @@
...
+
+    async fn fetch(&self, key: FastStr) -> Result<Arc<JwksValidator>, Self::Error> {
+        // TODO: Make this stored in the struct so we don't need to keep creating it.
+        // let raw_issuer = get_raw_issuer(token)?;
+        let raw_issuer = key.to_string();
+        log::info!("Fetching key for issuer {}", raw_issuer.clone());
+        // TODO: Consider checking for trailing slashes or requiring a scheme.
+        let oidc_url = format!("{}/.well-known/openid-configuration", raw_issuer);
+        // TODO: log errors here.
+        let keys = Jwks::from_oidc_url(oidc_url).await?;
```
cloutiertyler, line 203: Oh, convenient that there's a library for it.

cloutiertyler (approved):
This generally looks good to me. I would not block this from merging. However, I left several nit comments. Perhaps we can address them in another PR.

Cargo.toml
cloutiertyler: That too is an option.

crates/core/src/auth/identity.rs
cloutiertyler: If we want to send it to clients, might as well put it in the token then.

crates/core/src/auth/token_validation.rs
jsdt: It won't be convenient if we switch `jwt` libraries, but it's simple to reimplement it.

PR #1883, opened 2024-10-21
 websocket: send/receive reducer & table ids instead of names 

Centril:
# Description of Changes

Receive and send reducer ids and table ids as opposed to names in the SDK.

TODO:
- [x] Host implementation of the above.
- [x] Rust SDK implement of the above.
- [x] C# SDK implementation of the above, https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/pull/178
- [ ] TS SDK implementation of the above

Fixes https://github.com/clockworklabs/SpacetimeDBPrivate/issues/1091.

Yes.

### DIFF
--- a/crates/core/src/db/relational_db.rs
+++ b/crates/core/src/db/relational_db.rs
@@ -2424,9 +2424,9 @@ mod tests {
                 arg_bsatn,
             } = ReducerContext::try_from(&input).unwrap();
             if i == 0 {
-                assert_eq!(reducer_name, "__identity_connected__");
+                assert_eq!(&*reducer_name, "__identity_connected__");
             } else {
-                assert_eq!(reducer_name, "abstract_concrete_proxy_factory_impl");
+                assert_eq!(&*reducer_name, "abstract_concrete_proxy_factory_impl");
             }
             assert!(
                 arg_bsatn.is_empty(),
--- a/crates/core/src/execution_context.rs
+++ b/crates/core/src/execution_context.rs
@@ -26,7 +26,7 @@ pub struct ExecutionContext {
 #[derive(Clone)]
 pub struct ReducerContext {
     /// The name of the reducer.
-    pub name: String,
+    pub name: Arc<str>,
     /// The [`Identity`] of the caller.
     pub caller_identity: Identity,
     /// The [`Address`] of the caller.
@@ -81,7 +81,7 @@ impl TryFrom<&txdata::Inputs> for ReducerContext {
         let timestamp = bsatn::from_reader(args)?;
 
         Ok(Self {
-            name: inputs.reducer_name.to_string(),
+            name: inputs.reducer_name.as_str().into(),
             caller_identity,
             caller_address,
             timestamp,
@@ -200,7 +200,7 @@ impl ExecutionContext {
     /// If this is a reducer context, returns the name of the reducer.
     #[inline]
     pub fn reducer_name(&self) -> &str {
-        self.reducer.as_ref().map(|ctx| ctx.name.as_str()).unwrap_or_default()
+        self.reducer.as_ref().map(|ctx| &*ctx.name).unwrap_or_default()
     }
 
     /// If this is a reducer context, returns the full reducer metadata.
--- a/crates/core/src/host/module_host.rs
+++ b/crates/core/src/host/module_host.rs
@@ -592,7 +592,8 @@ impl ModuleHost {
         let db = self.db();
         let workload = || {
             Workload::Reducer(ReducerContext {
-                name: reducer_name.to_owned(),
+                // TODO(perf, centril): consider allowing `&'static str | Arc<str>`, perhaps flexstr.
+                name: reducer_name.into(),
                 caller_identity,
                 caller_address,
                 timestamp: Timestamp::now(),
@@ -794,7 +795,7 @@ impl ModuleHost {
                     tx.ctx = ExecutionContext::with_workload(
                         tx.ctx.database_identity(),
                         Workload::Reducer(ReducerContext {
-                            name: reducer.into(),
+                            name: reducer,
                             caller_identity: params.caller_identity,
                             caller_address: params.caller_address,
                             timestamp: Timestamp::now(),
@@ -932,9 +933,7 @@ impl ModuleHost {
         let db = self.db();
         let auth = self.auth_ctx_for(identity);
         let (table_names, table_ids): (Vec<_>, Vec<_>) = db
-            .with_read_only(&ExecutionContext::internal(db.database_identity()), |tx| {
-                db.get_all_tables(tx)
-            })
+            .with_read_only(Workload::Internal, |tx| db.get_all_tables(tx))
             .expect("ids_to_name: database in a broken state?")
             .iter()
             .filter(|schema| is_table_visible(schema, &auth))
--- a/crates/core/src/host/wasm_common/module_host_actor.rs
+++ b/crates/core/src/host/wasm_common/module_host_actor.rs
@@ -429,7 +429,7 @@ impl<T: WasmInstance> WasmModuleInstance<T> {
 
         let op = ReducerOp {
             id: reducer_id,
-            name: reducer_name,
+            name: reducer_name_arc.clone(),
             caller_identity: &caller_identity,
             caller_address: &caller_address,
             timestamp,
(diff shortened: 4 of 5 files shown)


crates/cli/src/subcommands/generate/mod.rs
```diff
@@ -312,9 +312,9 @@ trait Lang {
     fn type_filename(&self, type_name: &ScopedTypeName) -> String;
     fn reducer_filename(&self, reducer_name: &Identifier) -> String;
 
-    fn generate_table(&self, module: &ModuleDef, namespace: &str, tbl: &TableDef) -> String;
+    fn generate_table(&self, idx: u32, module: &ModuleDef, namespace: &str, tbl: &TableDef) -> String;
     fn generate_type(&self, module: &ModuleDef, namespace: &str, typ: &TypeDef) -> String;
-    fn generate_reducer(&self, module: &ModuleDef, namespace: &str, reducer: &ReducerDef) -> String;
+    fn generate_reducer(&self, idx: u32, module: &ModuleDef, namespace: &str, reducer: &ReducerDef) -> String;
```
gefjon, line 317: Why this change to add indices to the client codegen? I believe, but would like it confirmed, that these are not table IDs which could be sent over a WebSocket; rather, they appear to be an unrelated optimization to the SDK internals.

If that is the case, I would strongly prefer this be broken out into a separate PR, and frankly I would see very little reason to approve such a PR without measurements showing that `HashMap` lookups of table and reducer name strings are a significant performance overhead in the SDKs. I would also question why we'+ prefer to introduce a new ID to this interface, rather than a change purely internal to the codegen, possibly involving either generating these indices within the codegen, or possibly involving a perfect hash function.

If that is not the case, and this change is necessary or useful towards the actual goal of this PR, namely transmitting integer IDs rather than name strings over the WebSocket API, then I would like to see documentation on this trait's methods describing what these indices mean, where they come from and how they're used. I would also like the PR description amended to describe this change.

crates/cli/src/subcommands/generate/mod.rs
```diff
@@ -366,7 +372,8 @@ impl GenItem {
                 _ => todo!(),
             },
             GenItem::Reducer(reducer) => {
-                let code = csharp::autogen_csharp_reducer(ctx, reducer, namespace);
+                let code = csharp::autogen_csharp_reducer(ctx, *reducer_idx, reducer, namespace);
+                *reducer_idx += 1;
```
gefjon, line 376: My kingdom for `usize::post_increment(&mut self, inc: Self) -> Self`...

crates/core/src/db/datastore/locking_tx_datastore/committed_state.rs
```diff
@@ -510,9 +510,8 @@ impl CommittedState {
                     deletes.push(pv);
                 }
 
-                let table_name = &*table.get_schema().table_name;
-
                 if !deletes.is_empty() {
+                    let table_name = &table.get_schema().table_name;
```
gefjon, line 514: Is this related to the PR, or just a drive-by change? Is it actually an optimization, i.e. is `get_schema` expensive?

crates/core/src/db/datastore/locking_tx_datastore/committed_state.rs
```diff
@@ -551,9 +550,8 @@ impl CommittedState {
                 inserts.push(pv);
             }
 
-            let table_name = &*commit_table.get_schema().table_name;
-
             if !inserts.is_empty() {
+                let table_name = &commit_table.get_schema().table_name;
```
gefjon, line 554: Same questions.

crates/core/src/db/datastore/locking_tx_datastore/datastore.rs
```diff
@@ -899,7 +899,6 @@ impl<F: FnMut(u64)> spacetimedb_commitlog::payload::txdata::Visitor for ReplayVi
         reader: &mut R,
     ) -> std::result::Result<Self::Row, Self::Error> {
         let schema = self.committed_state.schema_for_table(table_id)?;
-        // TODO: avoid clone
```
gefjon, line 902: Why is this comment removed?

crates/cli/src/subcommands/generate/typescript.rs
```diff
@@ -82,7 +82,7 @@ Requested namespace: {namespace}",
         output.into_inner()
     }
 
-    fn generate_table(&self, module: &ModuleDef, namespace: &str, table: &TableDef) -> String {
+    fn generate_table(&self, _idx: u32, module: &ModuleDef, namespace: &str, table: &TableDef) -> String {
```
gefjon, line 85: Why does TypeScript not need or use these indices?

crates/core/src/error.rs
```diff
@@ -114,11 +114,11 @@ pub enum PlanError {
...
     TableNotFoundQualified { expect: String },
     #[error("Unknown field: `{field}` not found in the table(s): `{tables:?}`")]
-    UnknownField { field: String, tables: Vec<Box<str>> },
+    UnknownField { field: String, tables: Vec<Arc<str>> },
     #[error("Unknown field name: `{field}` not found in the table(s): `{tables:?}`")]
-    UnknownFieldName { field: FieldName, tables: Vec<Box<str>> },
+    UnknownFieldName { field: FieldName, tables: Vec<Arc<str>> },
     #[error("Field(s): `{fields:?}` not found in the table(s): `{tables:?}`")]
-    UnknownFields { fields: Vec<String>, tables: Vec<Box<str>> },
+    UnknownFields { fields: Vec<String>, tables: Vec<Arc<str>> },
```
gefjon, line 121: Are these (and similar changes from `Box` to `Arc` in this PR) related to the goal of the PR? Could they be split into a separate PR?

crates/sdk/src/callbacks.rs
```diff
@@ -51,34 +51,34 @@ impl CallbackId {
 /// Manages row callbacks for a `DbContext`/`DbConnection`.
 pub struct DbCallbacks<M: SpacetimeModule> {
     /// Maps table name to a set of callbacks.
-    table_callbacks: HashMap<&'static str, TableCallbacks<M>>,
+    table_callbacks: IntMap<u32, TableCallbacks<M>>,
```
gefjon, line 54: Comment is no longer correct. It's also not clear what the actual key is, now - is it the table ID received during handshake, or the table index computed at codegen time?

gefjon (changes requested):
- Please update the PR description to match the template.
- I would like to see a Rust SDK test added which verifies that it's possible to construct a connection and then immediately invoke a reducer, without first waiting for the on-connect callback. I would expect this to queue the reducer call until the handshake message is received. The code appears to do this correctly, but I want a test to make sure we don't regress it in the future.

bfops (approved):
LGTM on the CLI changes that I'm a codeowner for - I did not review `crates/cli/src/subcommands/generate/mod.rs` because that diff is more about "code generation" than about "CLI".

crates/cli/src/subcommands/generate/mod.rs
gefjon: I do not dispute that this is obviously better for runtime performance, but there are non-technical costs to making changes like this. For example:

- I have a stack of outstanding PRs, ending in an actual P1 ticket with actual performance and usability implications for the SDK, which will significantly conflict with this change.
- Reviewing this PR was made artificially and unnecessarily more difficult because it included this change, and I had to sort through which parts of the diff related to the stated goal of the PR versus which parts related to this change.
- Introducing a third identifier for database objects, in addition to the string names and the runtime IDs, adds significant potential for confusion. It's already not great that tables have both names and IDs, with different semantics, but at least the names are of an obviously distinct type from the IDs, so there's less risk of confusing the two. I believe there to be a very real cost in code complexity and maintenance to adding another integer identifier for database objects which is not interchangeable with IDs.

cloutiertyler: There is the additional cost that making a larger change may introduce more bugs. As the person who has to deal with all the bugs and their downstream ramifications, I can tell you that the cost is very very high. We just dealt with it most recently with the change to `u256` for `Identity`. The nature of these things is that it always feels like it will probably be fine, but it rarely is and then we pay the price downstream.

I agree with @gefjon, please separate the PR. I don't think this one can go in, but we'll get it in eventually. Either that or we can just have this be a different version of the API and maintain both and eventually deprecate the one Lightfox is using currently.

bfops:
@Centril is this API breaking _and_ ABI breaking? Can you add the appropriate labels?

PR #1885, opened 2024-10-22
core: Make `Replay::next_tx_offset` pub

kim:
Tooling should be able to restore from a snapshot, so that traversing
transactions doesn't require to traverse the entire commitlog from the
start.

To do so, it needs access to the tx offset post restoration.


kim:
superseded by #1899

PR #1894, opened 2024-10-23
Don't autogen schedule fields

### LINKED ISSUE #1818: `#[spacetimedb::table(scheduled())]` should probably not autogen the scheduled_id/at fields for you - should just require they already be present
UX story:

> Someone new comes to the codebase and wants to insert a row into the table. the compiler complains "missing fields scheduled_id and schedule_at" and so they go to the struct definition but they're not there. what the heck are they?? what do they do?? how is the compiler complaining about these fields that don't exist??

coolreader18:
# Description of Changes

Resolves #1818

module API–breaking, but a very very easy fix (copy-paste from the error message into your code and it's fixed). was deemed an okay breakage for rc1->rc2, if that's what ends up happening.

1


crates/bindings-macro/src/table.rs
```diff
@@ -543,17 +547,46 @@ pub(crate) fn table_impl(mut args: TableArgs, mut item: MutItem<syn::DeriveInput
...
-            let scheduled_at_id = (fields.len() - 1) as u16;
-            quote!(spacetimedb::table::ScheduleDesc {
+            // better error message when both are missing
+            if scheduled_at_column.is_none() && primary_key_column.is_none() {
+                return Err(syn::Error::new(
+                    original_struct_ident.span(),
+                    "scheduled table missing required columns; add these to your struct:\n\
+                             #[primary_key]\n\
+                             #[auto_inc]\n\
+                             scheduled_id: u64,\n\
```
gefjon, line 566: Is `scheduled_id` required to have both of these attributes? In particular, I could imagine `auto_inc` not being required, though usually correct. (To be clear, I am not advocating for changing this message, just asking.)

gefjon (approved):
:excellent: Thanks for updating the test!

crates/bindings-macro/src/table.rs
coolreader18: auto_inc isn't required, no

PR #1897, opened 2024-10-24
Eliminate dependency on `im`; just mutate the client cache

gefjon:
# Description of Changes

This turns out to be effectively a prerequisite for client-side indices. Storing indexes in a type-erased way
while making them `Clone` for storage in `im::HashMap` is likely possible, but it's at least very challenging and complex. Given that we no longer actually need the concurrent persistence offered by `im::HashMap`, it seems easier to just revert to using `std::collections::HashMap`.

Breaks SDK codegen. No user-facing breakage.

2 - minor risk of deadlocks because the SDK's concurrency story is pretty complex, but well-tested so likely safe.

# Testing

- [x] SDK test suite


mamcx (commented):
LGTM, but I think needs more eyes.

kazimuth (commented):
This all makes sense to me, but I'm not that familiar with the rust sdk so I'm not confident approving.

coolreader18 (approved):

gefjon:
Included in #1909 ; merging that and closing this.

PR #1900, opened 2024-10-24
Rust SDK: group all codegen'd names exported from the SDK under `pub mod __codegen`

gefjon:
# Description of Changes

Based on #1897 . While writing that PR, I moved a definition from `db_connection.rs` to `client_cache.rs`. This needlessly broke codegen.

Rather than having the codegen rely on a whole bunch of different `#[doc(hidden)] pub mod`s in the SDK library by path, group every name the codegen refers to under `spacetimedb_sdk::__codegen`. This has several benefits:
- It makes very clear that users should not be referring to those interfaces.
- It allows us to move definitions around in the SDK library without changing the codegen, so long as we update the re-exports from the `__codegen` module.

In a similar vein, the unstable interface Cloud use(s|d) to set its client address is moved to `spacetimedb_sdk::unstable`. This may end up getting removed anyways; I'm not sure what the plan is going forward for how cloud nodes will identify themselves.

Breaks codegen. Minor user-facing breakage: `u256, i256` are now re-exported from the SDK root rather than via the `spacetimedb_sdk::sats::{u256, i256}`. The `sats` crate is no longer fully re-exported, so users who want it will need a separate dependency.

1

# Testing

- [x] SDK test suite.


crates/sdk/src/client_cache.rs
```diff
@@ -112,41 +116,186 @@ impl<M: SpacetimeModule> ClientCache<M> {
...
 }
 
-/// A shared view into a particular state of the `ClientCache`.
-pub(crate) type ClientCacheView<M> = Arc<ClientCache<M>>;
+/// Internal implementation of a generated `TableHandle` struct,
+/// which mediates access to a table in the client cache.
+///
+/// `TableHandle`s don't actually hold a direct reference to the table they access,
+/// as that would require both gnarly lifetimes and also a `MutexGuard` on the client cache.
+/// Instead, they hold an `Arc<Mutex>` on the whole [`ClientCache`],
```
kazimuth, line 130: obligatory:
- what if we used `RwLock`
- nah we can do that later.

kazimuth (commented):
Everything here makes sense to me, but again, I'm not the most familiar with the Rust SDK so I'd like it if this got more eyes on it.

crates/sdk/src/client_cache.rs
gefjon: That doesn't fix the "gnarly lifetimes" issue. And it actually seems likely to introduce *more* deadlocks, not less, as in that version we would have long-standing read guards on the client cache, whereas in this version we have only short-lived exclusive guards.

coolreader18 (approved):

gefjon:
Included in #1909 ; merging that and closing this.

PR #1902, opened 2024-10-24
commitlog: Improve skipping behavior of traversals

kim:
The `*_from` style traversals have historically yielded commits or transactions before the given from-offset, leaving it to downstream consumers to handle skipping.

While folding handles it internally, this behavior is not great for transaction iterators, due to the statefulness of decoding -- it is usually necessary to call `Decoder::skip_record` until the desired offset is found.

We would also yield all commits from the start of the nearest segment boundary, which can be quite confusing when using the commit iterators directly.

This patch fixes the situation by:

* Setting the desired offset as the inital next offset in the `Commits` iterator, instead of the nearest segment boundary.
* Looping instead of recursing in the `Commits` iterator while skipping commits, so we can skip until the initial offset without blowing the stack.
* Passing the desired offset to `Commit::into_transactions`, such that `Decoder::skip_record` can be called if the offset doesn't lie on the commit boundary.

2.5 -- makes head spin

**NOTE**: On top of #1901, because tooling work depends on the whole series

### DIFF
--- a/crates/commitlog/src/commit.rs
+++ b/crates/commitlog/src/commit.rs
@@ -123,19 +123,49 @@ impl Commit {
     ///
     /// The supplied [`Decoder`] is responsible for extracting individual
     /// transactions from the `records` buffer.
+    ///
+    /// `version` is the log format version of the current segment, and gets
+    /// passed to [`Decoder::decode_record`].
+    ///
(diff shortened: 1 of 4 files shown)


Shubham8287 (approved):
I like these improvments. :+1:

PR #1907, opened 2024-10-24
Fix the endpoint for short-lived tokens

jsdt:
# Description of Changes

NB: This was branched off https://github.com/clockworklabs/SpacetimeDB/pull/1892, so you can ignore the earlier commits.

The main change here is to fix the `identity/websocket_token` token route that clients can use to get a short-lived token. That endpoint will now validate the token normally, and return a token with the same clients, signed with the local key. This means the short lived token is not necessarily signed with `issuer`s key. This is ok, because when we validate tokens, we first check if the token was signed with our key before we even look at the issuer.

This also has a few innocuous changes to work with this private PR https://github.com/clockworklabs/SpacetimeDBPrivate/pull/1081.

1.

# Testing

The tests in `token_validation` verify how/when we are supposed to look at the issuer when validating keys.


coolreader18 (approved):
Seems reasonable to me.

PR #1908, opened 2024-10-25
feat: Add non-unique index join iterator

joshua-spacetime:
* Adds a non-unique (constraint) index join iterator
* Abstracts over index key projections in order to reuse code between unique and non-unique index joins


mamcx (approved):
LGTM

PR #1909, opened 2024-10-25
Rust SDK: Actual client-side indices for unique constraints

gefjon:
# Description of Changes

Based on #1900 , which in turn is based on #1897 .

Updates unique constraint `.find` methods on clients to be backed by actual indices, rather than O(n) full scans. Specifically, unique indices on clients are `HashMap`s, whereas on the server they are `BTreeMap`s. This is because:

- It was easier given that I haven't implemented client-side non-unique indices yet.
- We artifically restrict filter methods to a small set of known types, all of which are both `Ord` and `Hash`, so there's not much difference.
- `HashMap` is marginally more efficient, I guess?

Associated changes:

- Codegen now includes a mechanism for registering each table in the `ClientCache` during `DbConnectionBuilder::build`, where previously they were lazily added the first time a row was inserted into them. This is necessary so that the codegen can install unique indices up front.
- The codegen actually discriminates on filterable types now, and does not emit indices or access methods for unique columns of non-filterable types.

Rust SDK codegen changes and will require re-running `spacetime generate`. Unique filter methods on columns of non-filterable types which were erroneously generated prior to now are no longer generated, so clients which depend on them will break.

2 - codegen's always fiddly, but slapping a few extra `HashMap`s around is pretty simple.

# Testing

- [x] Rust SDK tests still work, and rely on the `.find` methods behaving correctly.
- [ ] Measure or assert performance somehow? Frankly I don't know how much we care.


coolreader18 (approved):

gefjon:
Merging this and closing its two parents, rather than going through the rigamarole of merging each individually and rebasing.

PR #1911, opened 2024-10-25
CLI - Add `login show` and `logout`

bfops:
# Description of Changes

- Added the subcommand `login show` (to show the current login)
- Replaced `--clear-cache` with a separate `logout` subcommand (which imo is easier to understand)
- Made a small change to replace a `.conflicts_with` chain with an `ArgGroup`

Yes; this removes `--clear-cache`.

2

# Testing

- [x] `login clear` removed the logins from my config file
- [x] `login show` when not logged in gives an error message
- [x] `login show` when logged in shows my identity
- [x] `login show --token` when logged in shows my identity + token


cloutiertyler (approved):

crates/cli/src/subcommands/logout.rs
```diff
@@ -0,0 +1,12 @@
+use crate::Config;
+use clap::{ArgMatches, Command};
+
+pub fn cli() -> Command {
+    Command::new("logout")
+}
+
+pub async fn exec(mut config: Config, _args: &ArgMatches) -> Result<(), anyhow::Error> {
```
PuruVJ, line 8: You also have to do `POST /auth/cli/logout` with same Authorizaton headers as sent on `/api/spacetimedb-token`, to delete this session from auth service also

PuruVJ (changes requested):
Just one comment about logout

crates/cli/src/subcommands/logout.rs
PuruVJ: It also should take `auth-type` flag for that I assume

bfops: @PuruVJ what's the `auth-type` flag? What would I pass?

PuruVJ: Oh sorry I meant auth-host 😅

bfops: ah okay :+1: in that case done

PR #1912, opened 2024-10-28
Companion to spacetime-web#470

PuruVJ:
This modifies login.rs with all the namign changes in https://github.com/clockworklabs/spacetime-web/pull/470


crates/cli/src/subcommands/login.rs
```diff
@@ -120,43 +120,76 @@ async fn spacetimedb_token_cached(
...
+                        session_token: session_token.clone(),
+                    }));
+                } else {
+                    // If approved is true but session_token is missing, return an error
+                    return Err("Session token is missing in response.".to_string());
+                }
+            } else {
+                // Approved is false, no session token expected
+                return Ok(None);
+            }
```
bfops, line 188: nit: I don't think we should be using both `if-else` cascades _and_ early return - can you do one or the other?

crates/cli/src/subcommands/login.rs
```diff
@@ -120,43 +120,76 @@ async fn spacetimedb_token_cached(
...
-struct WebLoginTokenResponse {
+struct WebLoginTokenData {
     token: String,
 }
 
+#[derive(Deserialize)]
+struct WebLoginTokenResponse {
+    success: bool,
+    data: WebLoginTokenData,
+}
```
bfops, line 144: nit: is it really necessary/useful to have this wrapper type if it just has a single member?

crates/cli/src/subcommands/login.rs
PuruVJ: Makes sense. I'm turning in for the day and won't be able to change it for next 12 hours or so. If you have bandwidth, feel free to give it a go

crates/cli/src/subcommands/login.rs
PuruVJ: I have standardized all the api routes with this structure, so consumer can check with a simple .success check whether it worked or not. Previously they were all being done either by returning json success result, or a text error, which would cause deserializing errors anywhere. This solid structure makes sure it can be avoided

bfops: Is this a general standard for all of our API routes, or just the auth-related ones? I'm not seeing this same pattern in other areas of our repo.

PuruVJ: This is for anything on the website

bfops: I notice that `/login/request-token` endpoint doesn't return the `error` member - is that on purpose?

PuruVJ: There isn’t anything on it that could give an error. Maybe I can add something but for now it’s not that worth it

bfops (approved):
This worked for me when I tested with https://github.com/clockworklabs/spacetime-web/pull/470 (of course we shouldn't merge until that's merged)

crates/cli/src/subcommands/login.rs
PuruVJ: Ohhh. I mean I dont have any opinion, I just keep these things an object so later if we add more properties we don't break stuff. Prolly doesn't apply here, its just part of my personal philosophy

Issue #1893, opened 2024-10-23
`RawModuleDef` should include fewer names, allowing the host to generate names consistently


kazimuth:
With phoebe's addition of `accessor_name` for indexes, the `name` field in various RawModuleDefs become redundant. Currently, the module bindings are allowed to specify names for `RawIndexDef` / `RawConstraintDef` / `RawScheduleDef`. These names are (almost?) completely auto-generated. The user-specified name for indexes is already its own field, `accessor_name`, distinct from these.

The problem right now is that we have to re-implement name generation in both the C# and Rust module bindings. It's a pain to make these consistent. It's much easier to just not bother, and then have the host generate all names uniformly during `RawModuleDef` validation. Names are almost entirely an internal implementation detail anyway.

bfops:
Closed because it is marked done

Issue #1891, opened 2024-10-23
Special reducers don't preserve user-defined names


RReverser:
As per recent API revamp proposal, special reducers attached to `init`/`client_connected`/`client_disconnected` events should preserve user-provided names and use the V9 moduledef out-of-band signal to indicate the lifecycle event it's attached to.

I'm working on the C# counterpart of this (V8 -> V9 migration), but I noticed that Rust implementation of the bindings macro, which already supports V9 moduledef, doesn't preserve the reducer names either.

cc @gefjon

bfops:
Adding this to 1.0 release since it was created recently.

bfops:
This is _technically_ ABI-breaking because this could hypothetically create naming conflicts in the module and/or generated code, but only in cases where users are being silly. (And users can address by fixing it in their module code).

cloutiertyler:
This is actually a P1 because it causes the host to not call lifecycle reducers on the C# modules unless they share the name of the reducer. This is because the module host is incorrectly implemented and is still calling lifecycles based on their names and not on their V9 flags.

The only reason the Rust is working now is because the Rust macros in the module library rewrite the names of lifecycle reducers to the names that the module host (incorrectly) expects.

kazimuth:
https://github.com/clockworklabs/SpacetimeDB/blob/master/crates/schema/src/def.rs#L903 is where the data is stored in the ModuleDef. There's a ModuleDef stored in the modulehost that you can query to get this information. You could also cache the names of the lifecycle reducers on ModuleDef construction; ModuleDef is an ephemeral in-memory data structure so changes to it are not breaking.

PR #1918, opened 2024-10-28
Rip useless names out of RawDef

kazimuth:
# Description of Changes

Addresses https://github.com/clockworklabs/SpacetimeDB/issues/1893
Does NOT address https://github.com/clockworklabs/SpacetimeDB/issues/1891 since the special names turned out to be used throughout the codebase and I think addressing that will require a separate big PR.

While we're reworking naming, we may want to revisit the naming scheme here. It has a problem.
Consider the following declaration:

```rust
#[spacetimedb(table)]
#[derive(Clone, Debug)]
#[spacetimedb(index(name = "idx1", btree(columns = [a, b]))]
#[spacetimedb(index(name = "idx2", btree(columns = [a_b]))]
pub struct EdgeCase {
    pub a: u32,
    pub b: u32,
    pub a_b: u64,
}
```

These indexes will BOTH receive the generated name `"idx_EdgeCase_btree_a_b"`. Weird errors result.

(`idx1` and `idx2` are now referred to as `accessor_name`s and are stored separately from the name of the index itself.)

It would be better if we used out-of-band characters in the index name, like, say, `"EdgeCase.index.btree([a,b])"`. However, if we do this, we will need to relax the `Identifier` validation rules for indexes, since normally those will reject special characters. Also, it means that these identifiers can no longer be used directly as identifiers in SQL. @mamcx, I would welcome your opinion on this. I think it's fine, since system table queries can just wrap names in quotes.

We could also use a more elaborate mangling scheme, but this may penalize readability of the system tables.

ABI breaking.

Reduces complexity overall so let's say 2->1.

# Testing

Fixed all relevant tests.


RReverser (approved):
Thanks, LGTM. My only comments would be what's already covered by your TODOs.

mamcx:
I don't understand why it will generate `"idx_EdgeCase_btree_a_b"`. I expect it to be `"idx1_EdgeCase_btree_a_b"/"idx2_EdgeCase_btree_a_b"`

coolreader18 (approved):

kazimuth:
See also https://github.com/clockworklabs/SpacetimeDBPrivate/pull/915

crates/bindings-csharp/Codegen.Tests/fixtures/server/snapshots/Module#FFI.verified.cs
```diff
@@ -231,7 +229,7 @@ internal BTreeViewsUniqueIndex(BTreeViews handle)
             internal BTreeViewsUniqueIndex Id => new(this);
 
             internal sealed class LocationIndex()
-                : SpacetimeDB.Internal.IndexBase<global::BTreeViews>("bt_BTreeViews_Location")
+                : SpacetimeDB.Internal.IndexBase<global::BTreeViews>("index(,[X,Y])")
```
kazimuth, line 232: ah, this is wrong, wait.

kazimuth:
I think the remaining test failure here could be related to https://github.com/clockworklabs/SpacetimeDB/pull/1987 so I'm going to work on that first.

crates/core/src/db/datastore/locking_tx_datastore/datastore.rs
```diff
@@ -1449,17 +1449,17 @@ mod tests {
...
+            ConstraintRow { constraint_id: 6, table_id: ST_CONSTRAINT_ID.into(), unique_columns: col(0), constraint_name: "constraint.unique(st_constraint,[constraint_id])", },
+            ConstraintRow { constraint_id: 7, table_id: ST_CLIENT_ID.into(), unique_columns: col_list![0, 1], constraint_name: "constraint.unique(st_client,[identity,address])", },
+            ConstraintRow { constraint_id: 8, table_id: ST_VAR_ID.into(), unique_columns: col(0), constraint_name: "constraint.unique(st_var,[name])", },
+            ConstraintRow { constraint_id: 9, table_id: ST_SCHEDULED_ID.into(), unique_columns: col(0), constraint_name: "constraint.unique(st_scheduled,[schedule_id])", },
+            ConstraintRow { constraint_id: 10, table_id: ST_SCHEDULED_ID.into(), unique_columns: col(1), constraint_name: "constraint.unique(st_scheduled,[table_id])", },
+            ConstraintRow { constraint_id: 11, table_id: ST_ROW_LEVEL_SECURITY_ID.into(), unique_columns: col(1), constraint_name: "constraint.unique(st_row_level_security,[sql])", },
```
kazimuth, line 1462: Here are examples of the new auto-generated names in action.

kazimuth, line 1462: Here are examples of the new auto-generated default names in action.

kazimuth, line 1462: Here are examples of the new auto-generated default names in action.

cloutiertyler: Please use the Postgres standard naming scheme: https://stackoverflow.com/a/4108266/1445441

cloutiertyler (changes requested):
Please see comment.

crates/core/src/db/datastore/locking_tx_datastore/datastore.rs
kazimuth: That is sensitive to name collisions but it's easy to adopt.

cloutiertyler (approved):
Test suite is failing, but this now LGTM from my perspective on names

PR #1922, opened 2024-10-30
Fix benches for ia_loop, circles

mamcx:
# Description of Changes

Corrections to how are benched `ia_loop ` & `circles`.

Basically, `circles` have `auto_inc` in their tables definitions, but `ia_loop` does not. 

The consequence is that `circles` take a lot of time because the `init_game_circles` get called by each iteration of the benches, resulting in a big table with up to `1_000_000` rows when the setup was meant to be at most `10_000`. 

Then for the `ia_loop` the lack of `auto_inc` causes a `unique constraint violation` because the table is forbidden to grow. 

Now:

* In both, the database is initialized just once with the maximum amount of rows
* Is removed `group.measurement_time(Duration::from_secs(60 * 2));` that causes the bench to run longer than must
* Removed the filtering in the `ia_loop` game loop based on `timestamps`, it causes the code to run differently between the integration test and benches.
*  Increase the size of `ia_loop` from `[10, 100]` to `[500, 5_000]` so criterion has something to measure:
```bash
# BEFORE
special/db_game/ia_loop/load=100
                        time:   [0.0000 ps 0.0000 ps 0.0000 ps]
# AFTER
special/db_game/ia_loop/load=500
                        time:   [386.62 ps 461.53 ps 640.54 ps]
```

Nothing, but this change invalidates the numbers of previous benchmarks.

1

# Testing

- [x] *Runing a lot of benches with different parameters*

### DIFF
--- a/crates/bench/benches/special.rs
+++ b/crates/bench/benches/special.rs
@@ -10,7 +10,6 @@ use spacetimedb_lib::{bsatn::ToBsatn as _, sats, ProductValue};
 use spacetimedb_schema::schema::TableSchema;
 use spacetimedb_testing::modules::start_runtime;
 use std::sync::OnceLock;
-use std::time::Instant;
 use std::{hint::black_box, sync::Arc};
 
 #[global_allocator]
@@ -79,18 +78,10 @@ fn custom_db_benchmarks(c: &mut Criterion) {
                 });
             });
 
-            b.iter_custom(|iters| {
-                let start = Instant::now();
-
-                for _ in 0..iters {
-                    black_box(
-                        runtime.block_on(async {
-                            module.call_reducer_binary("run_game_circles", args.clone()).await.ok()
-                        }),
-                    );
-                }
-
-                start.elapsed()
+            b.iter(|| {
+                black_box(
+                    runtime.block_on(async { module.call_reducer_binary("run_game_circles", args.clone()).await.ok() }),
+                );
             })
         });
     }
@@ -109,14 +100,10 @@ fn custom_db_benchmarks(c: &mut Criterion) {
                 })
             });
 
-            b.iter_custom(|_iters| {
-                let start = Instant::now();
-
+            b.iter(|| {
                 black_box(
                     runtime.block_on(async { module.call_reducer_binary("run_game_ia_loop", args.clone()).await.ok() }),
                 );
-
-                start.elapsed()
             })
         });
     }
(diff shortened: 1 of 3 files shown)


mamcx:
benchmarks please

crates/bench/benches/special.rs
```diff
@@ -62,20 +63,23 @@ fn custom_db_benchmarks(c: &mut Criterion) {
     let module = runtime.block_on(async { BENCHMARKS_MODULE.load_module(config, None).await });
     let mut group = c.benchmark_group("special");
     // This bench take long, so adjust for it
-    group.measurement_time(Duration::from_secs(60 * 2));
     group.sample_size(10);
```
RReverser, line 66: Is this necessary or can be removed as well in favour of defaults? Also the comment above should be removed I think.

crates/bench/benches/special.rs
```diff
@@ -62,20 +63,23 @@ fn custom_db_benchmarks(c: &mut Criterion) {
...
                 runtime.block_on(async {
                     module
-                        .call_reducer_binary("init_game_circles", args.clone())
+                        .call_reducer_binary("init_game_circles", sats::product![100])
                         .await
                         .unwrap()
                 });
+            });
 
+            b.iter_custom(|iters| {
```
RReverser, line 82: Do we still need `iter_custom` after the initializer change? It seems that we can now instead use regular `b.iter`.

crates/bench/benches/special.rs
mamcx: Without limiting the sample the test goes from ~2min to ~10min in my machine, without any major improvement in the result.

crates/bench/benches/special.rs
mamcx: Ok, removed it, and also I made both benches to be the same logic (circles have another inner iteration).

crates/bench/benches/special.rs
RReverser: Hm for me it doesn't, Criterion still defaults to same 10 samples...

RReverser: But fair enough.

RReverser: Ah maybe it's because you changed params in the circle benchmark to larger ones but I didn't locally.

RReverser (approved):

PR #1923, opened 2024-10-30
Cached query execution plan

joshua-spacetime:
Defines an execution plan that is cached for subscription evaluation. This will be the lowering target for the physical plan.

### DIFF
--- a/crates/execution/src/iter.rs
+++ b/crates/execution/src/iter.rs
@@ -1,6 +1,7 @@
-use std::borrow::Cow;
+use std::ops::{Bound, RangeBounds};
 
 use spacetimedb_lib::{AlgebraicValue, ProductValue};
+use spacetimedb_primitives::{IndexId, TableId};
 use spacetimedb_table::{
     blob_store::BlobStore,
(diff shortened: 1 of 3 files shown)


crates/execution/src/iter.rs
```diff
@@ -165,96 +322,63 @@ impl<'a> Iterator for Iter<'a> {
...
+                iter.next().map(|(tuple, ptr)| {
+                    match (tuple, ptr) {
+                        // A leaf join
+                        //   x
+                        //  / \
+                        // a   b
+                        (Tuple::Row(u), ptr) => {
+                            // Returns a 2-tuple
+                            Tuple::Join(vec![u, Row::Ptr(ptr)])
+                        }
```
mamcx, line 361: If we always return 2 then how about `[2; Row]`?

mamcx (approved):
LGTM

crates/execution/src/iter.rs
joshua-spacetime: This is for the leaves of the join

Issue #1819, opened 2024-10-09
CLI - Replace clippy checks


bfops:
# Background
Clippy is slow. Since we really only use it for one things, we'd like to replace it with a "manual" search for the string `println!` and just warn about that.

# Definition of Done
- [ ] Replace use of `clippy` with a text-based search for `println!` in the project
- [ ] Remove `spacetimedb::println!`

bfops:
all linked PRs are done!

PR #1928, opened 2024-11-01
CLI - Replace clippy with a manual check

bfops:
# Description of Changes

We were apparently only using clippy to check for nonfunctional print statements (https://github.com/clockworklabs/SpacetimeDB/issues/1819).

This PR replaces our use of clippy (when building modules) with a manual check for disallowed print statements.

For me, this reduced the time of `cargo clean && time spacetime build` from `2m15s` (with clippy) to `1m30s` (without clippy).

Not breaking.

2

# Future work
The other items in https://github.com/clockworklabs/SpacetimeDB/issues/1819.

# Testing

- [x] Manually added a `println!` to new module and observed that `spacetime build` produces an error:
```
$ spacetimedb-lcli build

Detected nonfunctional print statements:

./src/lib.rs:30:     println!("Hello world");

Error: Found 1 disallowed print statement(s).
These will not be printed from SpacetimeDB modules.
If you need to print something, use the `log` crate
and the `log::info!` macro instead.
```

- [x] BitCraft fails by default
```
BitCraft/packages/game$ spacetimedb-cli build

Detected nonfunctional print statements:

./src/game/world_gen/world_generator.rs:53: //         //println!("{} {}", x, y);

Error: Found 1 disallowed print statement(s).
These will not be printed from SpacetimeDB modules.
If you need to print something, use the `log` crate
and the `log::info!` macro instead.
```

- [x] BitCraft passes once the `println!` is removed:
```
BitCraft/packages/game$ sed -i'' 's/println!/log::info!/g' src/game/world_gen/world_generator.rs
BitCraft/packages/game$ spacetimedb-cli build
    Updating crates.io index
  Downloaded spacetimedb v1.0.0-rc3
  Downloaded 1 crate (26.7 KB) in 0.25s
   Compiling thiserror v1.0.58
   Compiling log v0.4.21
   Compiling shlex v1.3.0
   Compiling itertools v0.12.1
   Compiling cc v1.2.9
...
```

- [x] BitCraft fails if I forcibly pass a different `lint-dir` that includes the `println!`s in `build.rs`:
```
BitCraft/packages/game$ spacetimedb-cli build --lint-dir .

Detected nonfunctional print statements:

././src/game/world_gen/world_generator.rs:53: //         //println!("{} {}", x, y);
././build.rs:45:     println!("cargo:rerun-if-changed={dir}");
././build.rs:98:     println!("cargo:rerun-if-changed=src/messages/components.rs");
././build.rs:492:         // output_commit.push_str(format!("            println!(\"{}: {{:?}}\", knowledge.entries);\n", n).as_str());

Error: Found 4 disallowed print statement(s).
These will not be printed from SpacetimeDB modules.
If you need to print something, use the `log` crate
and the `log::info!` macro instead.
```

- [x] `--lint-dir ''` skips linting:
```
$ spacetimedb-cli build --lint-dir ''
Warning: Skipping checks for nonfunctional print statements.
If you have used builtin macros for printing, such as println!,
your logs will not show up.
   Compiling bitcraft-spacetimedb v0.1.0 (/mnt/nutera/work/BitCraft/packages/game)
```


crates/cli/src/tasks/rust.rs
```diff
@@ -28,22 +29,35 @@ pub(crate) fn build_rust(project_path: &Path, skip_clippy: bool, build_debug: bo
...
-            &["--", "--no-deps", "-Aclippy::all", "-Dclippy::disallowed-macros"],
-        )
-        .dir(project_path)
-        .env("CLIPPY_DISABLE_DOCS_LINKS", "1")
-        .env("CLIPPY_CONF_DIR", clippy_conf_dir.path())
-        .unchecked()
-        .run()?;
-        anyhow::ensure!(out.status.success(), "clippy found a lint error");
+        let mut err_count: u32 = 0;
+        for file in walkdir::WalkDir::new(project_path).into_iter().filter_map(Result::ok) {
```
bfops, line 34: should we be walking the whole project directory? As-is, this erroneously catches some `println`s in BitCraft's `build.rs`. Should we be hardcoding `src/` or something...?

bfops: There does not appear to be a way to use e.g. `cargo` to list all the source files, as far as I can find.

crates/cli/src/tasks/rust.rs
```diff
@@ -28,22 +29,36 @@ pub(crate) fn build_rust(project_path: &Path, skip_clippy: bool, build_debug: bo
...
-            &["--", "--no-deps", "-Aclippy::all", "-Dclippy::disallowed-macros"],
-        )
-        .dir(project_path)
-        .env("CLIPPY_DISABLE_DOCS_LINKS", "1")
-        .env("CLIPPY_CONF_DIR", clippy_conf_dir.path())
-        .unchecked()
-        .run()?;
-        anyhow::ensure!(out.status.success(), "clippy found a lint error");
+        let mut err_count: u32 = 0;
+        for file in walkdir::WalkDir::new(project_path).into_iter() {
```
bfops, line 34: should we be walking the whole project directory? As-is, this erroneously catches some `println`s in BitCraft's `build.rs`. Should we be hardcoding `src/` or something...?

bfops: There does not appear to be a way to use e.g. `cargo` to list all the source files, as far as I can find. It looks like the best we could do is parse the output of the cargo build in order to figure out which files were involved.

bfops: Alternatively, we could add an extra argument that defaults to `src/`, e.g. `--lint-path` or something.

cloutiertyler: I think the last option is the best.

bfops: Done. Update the `Testing` section.

cloutiertyler (approved):
Thank you for the detailed testing spec. It's a little unfortunate that we warn even in commented code. I didn't think about that, but also that's not a very common case (although in principle it could come up). I'm good to approve this. We can consider an even better solution post 1.0

PR #1931, opened 2024-11-01
Use actual types for api responses, not ad-hoc `json!()` objects

coolreader18:
# Description of Changes

No behavior change, but it feels important to be able to see the actual structure of the response, especially as we're stabilizing.

1

# Testing

Was not able to get spacetime-web working to test this, but it's the same schema as it was before, and it matches with the typedef in typescript in spacetime-web.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4779,6 +4779,7 @@ dependencies = [
  "scopeguard",
  "serde",
  "serde_json",
+ "serde_with",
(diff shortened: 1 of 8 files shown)


gefjon (approved):
This seems like a significant improvement in code quality. Please coordinate with @PuruVJ to make sure the website still works, and double-check that the "Internal Tests" passes after you rebase-onto or merge-from master.

PR #1932, opened 2024-11-01
Use previous reducer context when downgrading a tx

joshua-spacetime:
So that we can correlate subscription updates with reducers


mamcx (approved):
LGTM

PR #1938, opened 2024-11-05
refactor: encode type info implicitly in the physical plan

joshua-spacetime:
Should simplify optimization

### DIFF
--- a/crates/physical-plan/src/plan.rs
+++ b/crates/physical-plan/src/plan.rs
@@ -1,61 +1,42 @@
-use spacetimedb_expr::ty::TyId;
+use std::{ops::Bound, sync::Arc};
+
 use spacetimedb_expr::StatementSource;
 use spacetimedb_lib::AlgebraicValue;
 use spacetimedb_primitives::{ColId, IndexId};
 use spacetimedb_schema::schema::TableSchema;
 use spacetimedb_sql_parser::ast::BinOp;
-use std::{ops::Bound, sync::Arc};
 
-/// A physical plan is a concrete query evaluation strategy.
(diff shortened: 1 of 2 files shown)


crates/physical-plan/src/plan.rs
```diff
@@ -1,32 +1,57 @@
...
 pub enum PhysicalPlan {
     /// Scan a table row by row, returning row ids
     TableScan(Arc<TableSchema>),
-    /// Fetch and return row ids from an index
-    IndexScan(IndexScan),
-    /// Join an input relation with a base table using an index
-    IndexJoin(IndexJoin),
+    /// Fetch row ids from an index
+    IxScan(IxScan),
+    /// Join a relation to a table using an index
```
mamcx, line 18: I rather use `IndexScan` like before, think is more descriptive.

crates/physical-plan/src/plan.rs
```diff
@@ -44,13 +69,19 @@ pub struct IndexScan {
 
 /// BTrees support equality and range scans
 pub enum IndexOp {
-    Eq(AlgebraicValue, TyId),
-    Range(Bound<AlgebraicValue>, Bound<AlgebraicValue>, TyId),
+    Eq(AlgebraicValue),
+    Range(Bound<AlgebraicValue>, Bound<AlgebraicValue>),
```
mamcx, line 73: I think this deserve a type, so is more explicit how we convert from `Ops` to this, ie:: 

```rust
struct BoundsOp {
   from:Bound<AlgebraicValue>, 
   to: Bound<AlgebraicValue>
}
```

mamcx (approved):
LGTM

PR #1940, opened 2024-11-05
 websocket: send/receive reducer & table ids instead of names (take 2)

Centril:
# Description of Changes

Receive and send reducer ids and table ids as opposed to names in the SDK.

Alternative to https://github.com/clockworklabs/SpacetimeDB/pull/1883 which is more efficient, but also uses indices internally rather than strings, making the diff larger.

Yes.


crates/cli/src/subcommands/generate/rust.rs
```diff
@@ -851,42 +837,45 @@ impl __sdk::spacetime_module::InModule for Reducer {{
...
-                                        reducer.name.deref(),
-                                    );
-                                }
+                "fn parse_call_info(
+                    reducer_id_to_name: &impl Fn(__ws::ReducerId) -> anyhow::Result<&'static str>,
+                    raw: __ws::ReducerCallInfo<__ws::BsatnFormat>,
+                ) -> anyhow::Result<Self> {",
+                |out| {
+                    out.delimited_block(
+                        "use __sdk::spacetime_module::parse_reducer_args;
```
gefjon, line 847: I generally prefer to avoid `use`-importing definitions within generated code, as I'm wary of name conflicts. In this specific case I don't see any possibility of a conflict, but just as a rule I prefer not thinking about whether it's even possible.

Don't change this, I'm just musing. In the future I may do some revisions to this file and add a file-level doc comment with a style guide.

crates/sdk/src/db_connection.rs
```diff
@@ -123,12 +131,36 @@ impl<M: SpacetimeModule> DbContextImpl<M> {
...
+                    .collect();
+                let check_eq = |got: &[Arc<_>], need: &[&_]| got.iter().map(|n| &**n).eq(need.iter().copied());
+                anyhow::ensure!(
+                    check_eq(&ids_to_names.reducer_names, M::Reducer::reducer_names()),
+                    "actual and expected reducer names do not match"
+                );
+                anyhow::ensure!(
+                    check_eq(&ids_to_names.table_names, M::DbUpdate::table_names()),
+                    "actual and expected reducer names do not match"
+                );
```
gefjon, line 153: Why require this? It seems to me that the SDK should be able to function even if it's not aware of the full set of reducers or of tables. Recall that unrecognized reducers are (or at least should be; it's possible this is mis-implemented) parsed as `Event::UnknownTransaction`, so the SDK can survive a `TransactionUpdate` message for an unknown reducer. We can reasonably assume that the client will never subscribe to a table which it doesn't have bindings for, so there's no risk of `TableUpdate`s for unknown tables.

We have already seen users in the wild manually removing bindings for reducers and tables they don't intend to access from the client, which seems like a silly thing to do, but I see no reason not to support. More importantly, it should be possible to use outdated codegen to connect to a newer module so long as the module's schema has not changed incompatibly, i.e. it should be possible to add new tables or reducers to a module without distributing a new client. Our customers depend on both of these abilities.

gefjon (commented):
Thanks for splitting this out. It's a much easier review this way. Notably, I was now able to find an actual semantic concern within this smaller diff. It should be easy to address.

gefjon (approved):

bfops (approved):
CLI changes in `generate/mod.rs` and `subscribe.rs` LGTM

cloutiertyler (dismissed):
Requires a corresponding internal change to address the Internal Test failure.

bfops:
@Centril is this both API and ABI breaking? Can you add the appropriate labels?

PR #1952, opened 2024-11-06
Decode identity correctly

jsdt:
# Description of Changes

The `token show` command was decoding tokens in a way that required the identity field to be set. When we parse tokens from clients, we can compute the identity from the other claims. This PR fixes the `token show` command to do the same.

1

# Testing

@bfops tested this manually.

### DIFF
--- a/crates/cli/src/util.rs
+++ b/crates/cli/src/util.rs
@@ -5,7 +5,7 @@ use base64::{
 };
 use reqwest::RequestBuilder;
 use serde::Deserialize;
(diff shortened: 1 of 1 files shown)


bfops (approved):
Works for me:
```
$ cargo run -- login show
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.36s
     Running `target/debug/spacetime login show`
You are logged in as c2002013212deb0bd4ce851bd6fc7fa4117b52b2af1d2341f3d39c16374a94b5
```

PR #1954, opened 2024-11-06
CLI - Fix stale helptext instructing users to use `spacetime identity`

bfops:
# Description of Changes

Update stale help text that referred to `spacetime identity` (which has been removed).

No breaking changes

1

# Testing

Existing tests only


cloutiertyler (approved):
Much better, thank you!

PR #1955, opened 2024-11-06
CLI - Tell users if they're already logged in

bfops:
# Description of Changes

Currently, `spacetime login` just silently returns if the user is already logged in.

This PR adds a message informing the user that they are already logged in.

None

1

# Testing

- [x] When I run `spacetime login`, I see the message.
```
$ cargo run -- login
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.27s
     Running `target/debug/spacetime login`
You are already logged in.
If you want to log out, use spacetime logout.
```


cloutiertyler (approved):
Seems reasonable.

Issue #666, opened 2023-12-13
Track energy for database compute


joshua-spacetime:
We need to track the energy cost of datastore operations like inserts, deletes, index scans, etc. This can be done with varying degrees of specificity, from tracking the bytes that each operation touches to using the time of each operation as a proxy. We should pick something and implement it, as we can always refine the implementation later.

cloutiertyler:
In principle, couldn't we base this on whatever energy cost the cost based optimizer will assign to these operations? e.g. rows inserted, deleted, index lookups, etc?

joshua-spacetime:
> In principle, couldn't we base this on whatever energy cost the cost based optimizer will assign to these operations? e.g. rows inserted, deleted, index lookups, etc?

The optimizer will estimate the energy cost of a query by estimating (1) the physical bytes to be read, (2) memory allocations, (3) memory access patters, etc. The datastore operations can track all of these, but it can also track the total time spent in the datastore as a reasonable proxy.

joshua-spacetime:
Fixed by https://github.com/clockworklabs/SpacetimeDB/pull/2140

PR #1957, opened 2024-11-07
Track time spent in datastore methods; use this to charge energy

gefjon:
# Description of Changes

`TxId` and `MutTxId` now read from the `Instant` clock around all of their public methods, and accumulate the total `Duration` spent in those methods throughout the life of the transaction. Then, when committing or aborting the transaction, the `RelationalDB` charges energy via a new `EnergyMonitor` method, `record_datastore_time`.

This necessitates having `RelationalDB` hold an `Arc<dyn EnergyMonitor>`, as previously it had no mechanism by which to access the energy monitor.

Note that this approach is intended as a stopgap, rather than a permanent solution. Notable flaws include:

- Energy consumed by datastore operations is not included in the WS broadcasts sent to clients; these contain only WASM execution energy. This is because we construct the `ModuleEvent` before downgrading the TX and running incremental queries, and the total datastore energy cost is only available after those operations.

- Wall time is a very noisy measurement. It is not necessarily incorrect to pass this noise on to our customers, since many of the unpredictable costs are caused by events like page allocations which are caused by the user code, and in general we expect any variance not caused by user code to disappear as the total time grows large. Still, it's not wonderful that we will end up charging different amounts of energy to run the same operations against the same data sets multiple times.

- This patch does not track time spent in iterator or `RowRef` operations, only methods on the (`Mut`-) `TxId` (which are generally accessed through the `RelationalDB`).

- We also need to take care that this does not significantly regress performance, since it involves a potentially large number of added clock reads and relaxed atomic increments. My hope is that the atomics are uncontended and so it's not a significant cost, but hey, this is why we have benchmarks, right?
  - Update: Callgrind benchmarks look fine.

Will require private PR, which I will do after we determine that this is an acceptable approach and that the overhead is low. No user-facing interfaces change.

2 - performance risks, and may under- or over-charge energy, but seems unlikely to break functionality.

# Testing

- [ ] None - curious to hear reviewers' thoughts on how best to test this!

### DIFF
--- a/crates/client-api-messages/src/energy.rs
+++ b/crates/client-api-messages/src/energy.rs
@@ -13,6 +13,21 @@ pub struct EnergyQuanta {
     pub quanta: u128,
 }
 
+#[derive(Copy, Clone)]
+#[repr(transparent)]
+#[must_use]
+/// Time taken while executing datastore operations, for which energy should be charged.
+///
+/// A transparent newtype around [`Duration`], to enable a `must_use`
+/// annotation so that we don't forget to charge energy.
+pub struct DatastoreComputeDuration(pub Duration);
+
+impl DatastoreComputeDuration {
+    pub fn from_micros(micros: u64) -> Self {
+        Self(Duration::from_micros(micros))
+    }
+}
+
 impl EnergyQuanta {
     pub const ZERO: Self = EnergyQuanta { quanta: 0 };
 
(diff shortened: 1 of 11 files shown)


gefjon:
Closes #666 . Nice.

gefjon:
benchmarks please

kazimuth:
Uhhh keep in mind callgrind might not simulate delays due to atomics. But if they're uncontended it probably doesn't matter.

gefjon:
> Uhhh keep in mind callgrind might not simulate delays due to atomics. But if they're uncontended it probably doesn't matter.

Ah, I hadn't considered that. Are the wall-time benchmarks working again? Do I need to merge something into this branch?

gefjon:
@Centril has offered to run some benchmarks. This is low-priority among his work.

joshua-spacetime (changes requested):
Sorry for the delay in providing feedback here.

We're most likely going to have to rethink this approach since the majority of energy usage is going to come from reads, i.e. iterators. There are really only two operations that determine the cost of reads (in the datastore):

1. Index seeks
2. Row ID fetches, or `RowPointer` dereferences.

Of course the cost of both of these operations is highly data dependent, but just counting them should give us an initial estimate that's good enough modulo some constant factor. Ultimately we'll want to take the table schema into account to more accurately estimate the cost of a fetch.

cloutiertyler:
Closing this in favor of @jsdt's sku system.

cloutiertyler:
Closing this in favor of @jsdt's sku system.

PR #1964, opened 2024-11-08
C# endianness fixes

kazimuth:
# Description of Changes

Similar fixes to the previous endianness fixes (https://github.com/clockworklabs/SpacetimeDB/pull/1926) for C#.

Reverts a previous API break which (I believe) printed Identities and Addresses backwards in C#. Otherwise, only documents existing behavior & adds methods.

1

# Testing

I am not sure how to add unit tests to the C# bindings, @RReverser please advise. E: NVM, figured it out.


crates/bindings-csharp/BSATN.Runtime.Tests/BSATN.Runtime.Tests.csproj
```diff
@@ -0,0 +1,25 @@
...
+    <IsTestProject>true</IsTestProject>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <TargetFrameworks>net8.0</TargetFrameworks>
+    <RootNamespace>SpacetimeDB</RootNamespace>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="CSharpier.Core" Version="0.28.2" />
```
RReverser, line 14: I don't think you need this and the Logging one - they are necessary for codegen tests, which use it to format snapshots, but shouldn't be necessary for runtime ones.

crates/bindings-csharp/BSATN.Runtime.Tests/Tests.cs
```diff
@@ -0,0 +1,162 @@
...
+        var ident4 = bsatn.Read(reader);
+        Assert.Equal(ident4, ident);
+
+        // Note: From = FromLittleEndian
+        var ident5 = Identity.From(littleEndianBytes);
+        Assert.Equal(ident5, ident);
+    }
+
+    [Fact]
+    public static void IdentityLengthCheck()
```
RReverser, line 148: Nit: looks like property testing via [CsCheck](https://github.com/AnthonyLloyd/CsCheck) could make this test both a bit cleaner and perhaps more reliable.

RReverser: Same for other tests here too actually; but we can do that in the future.

crates/bindings-csharp/BSATN.Runtime/Builtins.cs
```diff
@@ -8,11 +8,136 @@ namespace SpacetimeDB;
...
+    /// <typeparam name="T"></typeparam>
+    /// <param name="source"></param>
+    /// <param name="reverse"></param>
+    /// <returns></returns>
+    static T Read<T>(ReadOnlySpan<byte> source, bool reverse)
+        where T : struct
+    {
+        if (reverse)
+        {
+            Span<byte> reversed = stackalloc byte[source.Length];
```
RReverser, line 64: I'd suggest to instead: 1) read T, 2) create a byte view over it using [MemoryMarshal.CreateSpan](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.memorymarshal.createspan?view=net-8.0) + [MemoryMarshal.AsBytes](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.memorymarshal.asbytes?view=net-8.0), 3) reverse bytes in-place.

That way you can avoid intermediate array and copying.

crates/bindings-csharp/BSATN.Runtime/Builtins.cs
```diff
@@ -35,10 +160,50 @@ public readonly record struct Address
...
+    }
+
+    /// <summary>
+    /// Create an Address from a hex string.
+    /// </summary>
+    /// <param name="hex"></param>
+    /// <returns></returns>
+    public static Address? FromHexString(string hex)
+    {
+        var addr = new Address(Util.ReadFromBigEndianHexString<U128>(hex));
```
RReverser, line 206: Since this doesn't go via FromBigEndian, it should have its own length check. It probably needs its own anyway, since hex string parsing doesn't check for odd length.

crates/bindings-csharp/BSATN.Runtime.Tests/Tests.cs
kazimuth: Ooh, very nice

crates/bindings-csharp/BSATN.Runtime/Builtins.cs
kazimuth: Good call.

crates/bindings-csharp/BSATN.Runtime/Builtins.cs
```diff
@@ -8,11 +8,139 @@ namespace SpacetimeDB;
...
+    /// (The endianness of the host system can be checked via System.BitConverter.IsLittleEndian.)
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    /// <param name="source"></param>
+    /// <param name="reverse"></param>
+    /// <returns></returns>
+    static T Read<T>(ReadOnlySpan<byte> source, bool reverse)
+        where T : struct
+    {
+        Debug.Assert(
```
kazimuth, line 62: Now everything gets a length check :)

crates/bindings-csharp/BSATN.Runtime/Builtins.cs
```diff
@@ -8,11 +8,139 @@ namespace SpacetimeDB;
...
+            source.Length == Marshal.SizeOf<T>(),
+            $"Error while reading ${typeof(T).FullName}: expected source span to be {Marshal.SizeOf<T>()} bytes long, but was {source.Length} bytes."
+        );
```
kazimuth, line 65: Now everything gets a length check :)

kazimuth (commented):
This should be ready now @RReverser

RReverser:
> It probably needs its own anyway, since hex string parsing doesn't check for odd length.

I think this part got missed - passing an odd-length hex string like `012` would just ignore the `2` instead of throwing an error.

The rest looks good, thanks!

RReverser (approved):
Approving with one final comment.

crates/bindings-csharp/BSATN.Runtime.Tests/Tests.cs
```diff
@@ -0,0 +1,179 @@
...
+
+using CsCheck;
+using Xunit;
+
+public static class BSATNRuntimeTests
+{
+    [Fact]
+    public static void AddressRoundtrips()
+    {
+        var str = "00112233445566778899AABBCCDDEEFF";
```
RReverser, line 11: Do we need a test that checks that non-hex characters don't pass?

crates/bindings-csharp/BSATN.Runtime.Tests/Tests.cs
```diff
@@ -0,0 +1,179 @@
...
+        var bsatn = new Address.BSATN();
+        if (addr is { } addrNotNull)
+        {
+            bsatn.Write(writer, addrNotNull);
+        }
+        else
+        {
+            Assert.Fail("Impossible");
+        }
+        writer.Flush();
```
RReverser, line 54: Very minor nitpick, but I'd instead wrap this into a block and use `using ` for memorystream&writer inside, so that RAII takes care of flushing for us.

crates/bindings-csharp/BSATN.Runtime.Tests/Tests.cs
```diff
@@ -0,0 +1,179 @@
...
+    [Fact]
+    public static void AddressRoundtrips()
+    {
+        var str = "00112233445566778899AABBCCDDEEFF";
+        var addr = Address.FromHexString(str);
+
+        Assert.NotNull(addr);
+        Assert.Equal(addr.ToString(), str);
+
+        byte[] bytes =
```
RReverser, line 17: If you want, you can use [`Convert.FromHexString`](https://learn.microsoft.com/en-us/dotnet/api/system.convert.fromhexstring?view=net-8.0) here.

We can't use it in BSATN.Runtime because it's not part of .NET Standard which we need for Unity, but no harm in using it in tests.

crates/bindings-csharp/BSATN.Runtime/Builtins.cs
```diff
@@ -8,11 +8,152 @@ namespace SpacetimeDB;
...
+    public static byte[] StringToByteArray(string hex)
+    {
+        Debug.Assert(
+            hex.Length % 2 == 0,
+            $"Expected input string (\"{hex}\") to be of even length"
+        );
+
+        foreach (var c in hex)
+        {
+            Debug.Assert(
```
RReverser, line 132: I don't think this is necessary, `Convert.ToByte` should do it by itself. When I said "do we need a test", I really only meant a unit test :)

kazimuth:
I'm nervous about these internal test failures, I don't know what's causing them.

gefjon:
> I'm nervous about these internal test failures, I don't know what's causing them.

These are a known issue that should hopefully be fixed by tomorrow.

crates/bindings-csharp/BSATN.Runtime.Tests/Tests.cs
RReverser: I see you used it in the identity roundtrips tests below, but not in this function. Seems unintentional?

crates/bindings-csharp/BSATN.Runtime.Tests/Tests.cs
```diff
@@ -0,0 +1,158 @@
...
+
+        Array.Reverse(bytes);
+        var ident3 = Identity.From(bytes);
+        Assert.Equal(ident3, ident);
+
+        var memoryStream = new MemoryStream();
+        var writer = new BinaryWriter(memoryStream);
+        var bsatn = new Identity.BSATN();
+        bsatn.Write(writer, ident);
+        writer.Flush();
```
RReverser, line 115: Same nit re: `using (...)` block, like you do in address now.

PR #1965, opened 2024-11-08
Run and test benchmarks against C# as well

RReverser:
# Description of Changes

Turns out, there are some APIs that are not covered by any of our tests, except benchmarks.

Benchmarks are executed as both actual benchmarks and as integration tests for Rust, but not for C#. This PR fixes that to ensure that both Rust and C# are tested and measured on the CI from now on.

It also makes some adjustments to integration tests themselves, in particular fixes an issue where in integration tests reducers would be called but the response would never be read, so if a reducer panics, the tests would still succeed. Additionally, on failures the SpacetimeDB logs will be read and added to the error message so that you get a useful stacktrace from the module itself.

I also had to adjust some benchmark numbers (in particular, for `ia_loop`) since they were already very slow on CI for Rust, but become even more so for C#. The new loads are much lower, but should still be very usable for benchmark comparisons.

The only remaining issue is that the benchmark names have changed, and the benchmark comparison tool tries to compare mismatched pairs - I'm not sure what the solution to this is. cc @kazimuth 

corresponding GitHub label.

# Testing

### DIFF
--- a/crates/bench/src/spacetime_module.rs
+++ b/crates/bench/src/spacetime_module.rs
@@ -1,4 +1,4 @@
-use std::path::Path;
+use std::{marker::PhantomData, path::Path};
 
 use spacetimedb::db::{Config, Storage};
 use spacetimedb_lib::{
@@ -7,7 +7,7 @@ use spacetimedb_lib::{
 };
 use spacetimedb_paths::RootDir;
 use spacetimedb_primitives::ColId;
-use spacetimedb_testing::modules::{start_runtime, CompilationMode, CompiledModule, LoggerRecord, ModuleHandle};
+use spacetimedb_testing::modules::{start_runtime, LoggerRecord, ModuleHandle, ModuleLanguage};
 use tokio::runtime::Runtime;
 
 use crate::{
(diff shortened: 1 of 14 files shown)


RReverser:
benchmarks please

RReverser:
benchmarks please

crates/testing/src/modules.rs
```diff
@@ -223,3 +226,43 @@ pub struct LoggerRecord {
     pub line_number: Option<u32>,
     pub message: String,
 }
+
+const COMPILATION_MODE: CompilationMode = if cfg!(debug_assertions) {
```
kazimuth, line 230: Do we know if the `debug_assertions` feature is currently enabled in the transitive dependencies of the `bench` crate?

crates/core/src/host/module_host.rs
```diff
@@ -238,14 +238,12 @@ impl ModuleInfo {
...
-/// Invariant: the reducer names are in alphabetical order.
+/// Invariant: the reducer names are in the same order as they were declared in the `ModuleDef`.
 pub struct ReducersMap(IndexSet<Box<str>>);
 
 impl<'a> FromIterator<&'a str> for ReducersMap {
     fn from_iter<T: IntoIterator<Item = &'a str>>(iter: T) -> Self {
-        let mut sorted = Vec::from_iter(iter);
-        sorted.sort();
-        ReducersMap(sorted.into_iter().map_into().collect())
+        Self(iter.into_iter().map_into().collect())
```
kazimuth, line 246: If this PR is delayed we should do a separate PR with this change and merge it right away cause it's a straight up correctness fix.

crates/testing/src/modules.rs
RReverser: No, but it doesn't have to be - the point here is to use that mode only in integration test mode for easier debugging (it helped me with a stacktrace).

RReverser:
@kazimuth FWIW this is marked as draft because it's not quite ready to look at. I only created a PR because without it our CI doesn't run on branches :)

kazimuth:
Well, looks good to me so far :)

RReverser:
benchmarks please

crates/core/src/host/module_host.rs
RReverser: Done in #1980.

RReverser:
benchmarks please

RReverser:
The failure here is the one that will be fixed by #1987.

RReverser:
Copied relevant change myself to unblock the CI.

crates/bench/src/spacetime_module.rs
```diff
@@ -42,26 +25,27 @@ lazy_static::lazy_static! {
 ///
 /// See the doc comment there for information on the formatting expected for
 /// table and reducer names.
-pub struct SpacetimeModule {
+pub struct SpacetimeModule<L> {
```
kazimuth, line 28: neat, I like this.

gefjon (approved):

gefjon:
Internal tests failures appear to be spurious/unrelated, but let's wait to merge until they're fixed just in case.

PR #1966, opened 2024-11-08
Remove some Identity conversions that could panic

jsdt:
# Description of Changes

This removes `Identity::{from_slice,from_be_slice}`, which previously had the potential to panic. Since callers can just use `try_into`, I don't think we need this in the API.

This removes two functions from the `Identity` lib. If we find that these are widely used, we could add back a safe version. Making this safe will always be a breaking change though, since we need to change the return type to a Result.

1

# Testing

Not much added testing, since this is a pretty mechanical change.


kazimuth (commented):
This is a good idea. I was just retouching these and didn't like that `unwrap`. However, it is an API break. My concern was that `from_slice` might already be being used by clients so I wasn't sure about changing it. `from_be_slice` is definitely fine to remove since it was just added.

We could also change them to return `Option<Self>` instead.

crates/lib/src/identity.rs
```diff
@@ -269,6 +244,17 @@ mod tests {
         );
     }
 
+    // Make sure the checksum is valid.
```
gefjon, line 247:
```suggestion
    /// Make sure the checksum is valid.
```

gefjon (approved):
Thanks for cleaning this up. One minor terminology request: in the future, please don't use "unsafe" to describe Rust code unless it is `unsafe`, i.e. liable to invoke UB. Your PR title made me somewhat worried about this, when the actual changes are trivial and easy to approve.

gefjon:
Spurious test failure should be fixed by #1979 . Please rebase onto master or merge master into this branch, at your preference, to confirm.

PR #1967, opened 2024-11-08
Implement DataStore trait for Tx/MutTx

mamcx:
# Description of Changes

First step into the integration of the new query planner.

This implement the new `DataStore` trait for both `Tx`/`MutTx` and introduce a new `Query<Tx | MutTc>` struct that implement the internal query methods. This will eliminate the need of having the many `iter / iter_mut` versions around.

**NOTE**: Is draft because wanna coordinate how will be the direction with @joshua-spacetime, because this is part of a large set of steps for the new query engine. That is why not documentation is written.

1

# Testing

- [x] *New test for the trait*

### DIFF
--- a/crates/core/src/db/query.rs
+++ b/crates/core/src/db/query.rs
@@ -1,0 +1,297 @@
+use crate::db::datastore::locking_tx_datastore::committed_state::CommittedState;
+use crate::db::datastore::locking_tx_datastore::tx::TxId;
+use crate::db::datastore::locking_tx_datastore::MutTxId;
+use crate::db::datastore::traits::Tx;
+use spacetimedb_data_structures::map::IntMap;
+use spacetimedb_execution::iter::{Datastore, DeltaScanIter};
+use spacetimedb_primitives::{ColList, IndexId, TableId};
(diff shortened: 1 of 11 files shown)


crates/core/src/db/query.rs
```diff
@@ -0,0 +1,297 @@
...
+    fn get_committed_state(&self) -> &CommittedState;
+    fn get_tables(&self) -> &IntMap<TableId, Table>;
+}
+
+impl Datastore for TxId {
+    fn delta_scan_iter(&self, _table_id: TableId) -> DeltaScanIter {
+        DeltaScanIter::empty_iter()
+    }
+
+    fn table_scan_iter(&self, table_id: TableId) -> TableScanIter {
```
mamcx, line 24: All this methods requiere `unwrap` and assume the acquiring of the `iter` will not fail. I think we should make them return a `Result` so we can report `Index/Table not found`

Centril (changes requested):

mamcx:
Closed because @joshua-spacetime implements it in his prs.

PR #1969, opened 2024-11-08
CLI - Remove unused `default_identity` stuff

bfops:
# Description of Changes

I removed the `default_identity` member and the associated functions, since they're not currently used anywhere.

None

1

# Testing

Existing tests only


cloutiertyler (approved):
LGTM

PR #1987, opened 2024-11-12
Preserve reducer order in schema conversion

RReverser:
# Description of Changes

This was found while working on #1965. The tests will come in that other PR, but it was suggested that we should land the correctness fix itself sooner.

corresponding GitHub label.

# Testing

### DIFF
--- a/crates/schema/src/def.rs
+++ b/crates/schema/src/def.rs
@@ -24,6 +24,7 @@ use crate::schema::{Schema, TableSchema};
 use crate::type_for_generate::{AlgebraicTypeUse, ProductTypeDef, TypespaceForGenerate};
 use deserialize::ReducerArgsDeserializeSeed;
 use hashbrown::Equivalent;
+use indexmap::IndexMap;
 use itertools::Itertools;
 use spacetimedb_data_structures::error_stream::{CollectAllErrors, CombineErrors, ErrorStream};
 use spacetimedb_data_structures::map::HashMap;
@@ -84,7 +85,9 @@ pub struct ModuleDef {
     tables: IdentifierMap<TableDef>,
 
     /// The reducers of the module definition.
-    reducers: IdentifierMap<ReducerDef>,
+    /// Note: this is using IndexMap because reducer order is important
+    /// and must be preserved for future calls to `__call_reducer__`.
+    reducers: IndexMap<Identifier, ReducerDef>,
 
     /// The type definitions of the module definition.
     types: HashMap<ScopedTypeName, TypeDef>,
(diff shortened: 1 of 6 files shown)


crates/schema/src/def.rs
```diff
@@ -84,7 +85,9 @@ pub struct ModuleDef {
     tables: IdentifierMap<TableDef>,
 
     /// The reducers of the module definition.
-    reducers: IdentifierMap<ReducerDef>,
+    /// Note: this is using IndexMap because reducer order is important
```
kazimuth, line 88: Can we get a corresponding comment on `RawModuleDefv9` as well

kazimuth (commented):
This looks good, I think will need a corresponding `Private` PR.

This *is* inconsistent with how IDs work for other entity types.
Two alternatives would be:
- Provide a `_reducer_id_from_name` API similar to indexes/tables
- Require the user to sort their reducers when computing IDs

This is definitely the easiest patch though.

RReverser:
> * Require the user to sort their reducers when computing IDs

Do you mean with validation check? That would also be fine by me as an alternative.

crates/schema/src/def.rs
RReverser: Where do you mean? I didn't change its type.

kazimuth: For most of the vectors in that struct, order does not matter, the contained data is sorted. However, this is no longer true for the `reducers` vector, whose order is now semantically important.

kazimuth:
Yeah, I think that's less confusing overall. I can do another branch that adds that check in `schema`.

RReverser:
If you think that's better, sure, go for it.

Personally I don't care too much. I do think that preserving whatever order happens to be, is easier for module binding libraries, particularly because a validation error would be visible a bit late in the pipeline (user would get an error only when an example module with mismatched order is compiled & published), but given that we only have 2 such libraries and not creating new one every month or so, it shouldn't be a big deal.

crates/schema/src/def.rs
kazimuth: Actually, this sort of doesn't matter, since the module itself never uses the Reducer ID.

kazimuth:
Ah, no, you're right. I was confused -- it really doesn't matter inside the module, the module never uses reducer IDs. It's just an interface somewhere else. So, I'm happy with this.

kazimuth (approved):
LGTM with tests passing

RReverser:
That's... some new test failure.

> This looks good, I think will need a corresponding `Private` PR.

Also, does this still stand / do I need to change private too?

kazimuth:
It looks like both test failures here were caused by unrelated heisenbugs on master, we don't need a Private fix. This one might also go away on rerun.

PR #1988, opened 2024-11-12
Clean up auth related code with traits, and reuse existing key generation code.

jsdt:
# Description of Changes

This is an attempt to clean up auth-related code. To summarize the main changes:

1. We don't use a global OIDC key cache any more.
2. The 4 JWT related functions in the `NodeDelegate` trait were removed, and the functionality was put in the `JwtAuthProvider` trait. `NodeDelegate` now just has one function to return an auth provider. This should make it easier to change this in the future. The new trait also exposes functions for signing and verifying keys, rather than giving the concrete keys.
3. `SpacetimeIdentityClaims2` is now the only `SpacetimeIdentityClaims`.
4. Instead of using `jsonwebkey` in the token validation tests, we use the same key generate code that is used for standalone key generation. There were some small changes to the key generate code to make it easier to reuse.

There is a corresponding [private PR](https://github.com/clockworklabs/SpacetimeDBPrivate/pull/1158).

None.

1. This is close to functionally equivalent, and should make future auth changes less complex.

# Testing

Most of the unit tests for this code are in `token_validation.rs`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -76,9 +76,9 @@ dependencies = [
 
 [[package]]
 name = "allocator-api2"
-version = "0.2.18"
+version = "0.2.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5c6cb57a04249c6480766f7f7cef5467412af1490f8d1e243141daddada3264f"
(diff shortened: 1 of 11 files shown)


gefjon (approved):

cloutiertyler (approved):
Approving the code owner parts I'm responsible for.

PR #1989, opened 2024-11-12
Add /health endpoint

drogus:
# Description of Changes

This PR adds a health endpoint with some potentailly usefeul information like:

* package name (so we can distinguish between standalone and cloud)
* package version
* if the current node is schedulable
* a list of nodes in the cluster

No breakage, the endpoint didn't exist in standalone before.

1

# Testing

- [x] *Local test*
- [x] *I have started the server, and checked if the `/health` endpoint responds with 200 and intended response JSON (version, schedulable status etc)*


kurtismullins (commented):
It _looks_ good to me but I don't know enough about our internal system (e.g `get_nodes()`) to be a good reviewer of this. 

I think that as long as it actually represents the database being healthy, its better than what we have now!

cloutiertyler (approved):
I am satisfied with @drogus's level of testing.

PR #1991, opened 2024-11-12
Use `NonZeroU64` for `num_distinct_values` to avoid div-by-zero

gefjon:
# Description of Changes

Hotfix for a div-by-zero panic encountered by the BitCraft Alpha.

What was happening was, `num_distinct_values` on an empty table returned 0, which `index_row_est` then used as a divisor, leading to a panic. It seems reasonably clear to me that the correct estimate of distinct values in an empty index is 0, not bottom, so adjusting `num_distinct_values` to return `Option<NonZeroU64>`, and then making `index_row_est` return 0 in that case, seems an obvious fix.

N/a.

1 - Hard to be more broken.

# Testing

- [ ] I am unsure how to test this. Perhaps we can deploy to staging, then do whatever the BitCraft team did that was broken, and verify that it no longer panics?

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -6,6 +6,7 @@ on:
       - master
       - staging
       - dev
+      - phoebe/hotfix/0.11.1/divide-by-zero
     tags:
       - 'v*'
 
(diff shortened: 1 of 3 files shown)


gefjon:
Closing in favor of #1985 , which merged long ago.

PR #1992, opened 2024-11-12
Avoid div-by-zero in row count estimations for empty tables

gefjon:
# Description of Changes

Forward-port of #1991 , an 0.11.1 hotfix for the BitCraft alpha, into master.

What was happening was, `num_distinct_values` on an empty table returned 0, which `index_row_est` then used as a divisor, leading to a panic. It seems reasonably clear to me that
the correct estimate of distinct values in an empty index is 0, not bottom, so adjusting `num_distinct_values` to return `Option<NonZeroU64>`, and then making `index_row_est` return 0 in that case, seems an obvious fix.

None

1

# Testing

- [ ] Check that it works for the BitCraft alpha.
- [ ] Maybe write a smoketest?


kim:
#1985 and #1986 addressed the same, you may want to pick #1986 as a regression test

gefjon:
> #1985 and #1986 addressed the same, you may want to pick #1986 as a regression test

Ah, I didn't realize we already had patches against master for this. Will close.

PR #1994, opened 2024-11-13
Fix inference for UNION

mamcx:
# Description of Changes

When trying to infer the type for `UNION` like queries we get:

```bash
 Unexpected type: (expected) (u32: U32f32: F32str: String) != (u32: U32f32: F32str: String) (inferred)
```

Because the check is only 'nominal' (aka: test both `table_id` are the same), this patch also makes a structural check.

1
# Testing


cloutiertyler:
bots please

(doing this for unrelated reasons to this PR)

mamcx:
Closed because need more changes on the new query engine that has obsoleted this code.

PR #1996, opened 2024-11-14
CLI should not depend on rustup

### LINKED ISSUE #1971: CLI should not depend on rustup
Phoebe: A small but meaningful subset of users (myself included) use Rust without `rustup`. Currently, this is painful, because the CLI unconditionally invokes `rustup` to ensure the `wasm32-unknown-unknown` target is installed. I ended up having to put a stub shell script into my path as `rustup` which does nothing and returns 0.

The CLI should verify that `cargo` has the `wasm32-unknown-unknown` target installed, and if it is present, not invoke `rustup`.

Bonus points if the CLI checks for `rustup` on the system when `wasm32-unknown-unknown` is not present, and if neither the target nor `rustup` are present, prints an error message telling the user to install the target.

mamcx:
# Description of Changes

Rustup could not be installed (for example, for people who use `nix`). 

Now, the `cli` defensively check for it. It also try to find the target `wasm32-unknown-unknown` in case is not installed.

Closes #1971.

corresponding GitHub label.

1

# Testing

- [x] *Write a test*
- [x] *Run manually in windows*

### DIFF
--- a/crates/cli/src/detect.rs
+++ b/crates/cli/src/detect.rs
@@ -6,54 +6,56 @@ use std::path::Path;
 pub(crate) fn find_executable(exe_name: impl AsRef<Path>) -> Option<std::path::PathBuf> {
     std::env::var_os("PATH").and_then(|paths| {
         std::env::split_paths(&paths)
-            .filter_map(|dir| {
-                let full_path = dir.join(&exe_name);
-                full_path.is_file().then_some(full_path)
-            })
-            .next()
+            .map(|dir| dir.join(&exe_name))
+            .find(|x| x.is_file())
     })
 }
 
 /// Check if `rustup` is installed (aka: Is in the `PATH`).
-pub(crate) fn has_rust_up() -> anyhow::Result<bool> {
-    Ok(match std::env::consts::OS {
+pub(crate) fn has_rust_up() -> bool {
+    match std::env::consts::OS {
         "linux" | "freebsd" | "netbsd" | "openbsd" | "solaris" | "macos" => find_executable("rustup").is_some(),
         "windows" => find_executable("rustup.exe").is_some(),
         unsupported_os => {
-            return Err(anyhow::anyhow!(
-                "This OS may be unsupported for `rustup`: {unsupported_os}"
-            ));
+            eprintln!("This OS may be unsupported for `rustup`: {unsupported_os}");
+            false
         }
-    })
+    }
 }
 
 /// Check if `rustfmt` is installed (aka: Is in the `PATH`).
-pub(crate) fn has_rust_fmt() -> anyhow::Result<bool> {
-    Ok(match std::env::consts::OS {
+pub(crate) fn has_rust_fmt() -> bool {
+    match std::env::consts::OS {
         "linux" | "freebsd" | "netbsd" | "openbsd" | "solaris" | "macos" => find_executable("rustfmt").is_some(),
         "windows" => find_executable("rustfmt.exe").is_some(),
         unsupported_os => {
-            return Err(anyhow::anyhow!(
-                "This OS may be unsupported for `rustfmt`: {unsupported_os}"
-            ));
+            eprintln!("This OS may be unsupported for `rustfmt`: {unsupported_os}");
+            false
         }
-    })
+    }
 }
 
 /// Check if the [Target] is installed.
 ///
 /// **NOTE:** If `rustup` is not installed, we check inside the `rustc sysroot` directory.
-pub(crate) fn has_wasm32_target() -> anyhow::Result<bool> {
-    if has_rust_up()? {
-        let output = cmd!("rustup", "target", "list", "--installed").read()?;
-        Ok(output.contains("wasm32-unknown-unknown"))
-    } else {
-        // When `rustup` is not installed, we need to manually check the [Target] inside the sysroot directory
-        let root = cmd!("rustc", "--print", "sysroot").read()?;
-        Path::new(&format!("{}/lib/rustlib/{}", root.trim(), "wasm32-unknown-unknown"))
-            .try_exists()
-            .map_err(|err: io::Error| anyhow::anyhow!(err))
-    }
+pub(crate) fn has_wasm32_target() -> bool {
+    let result = || {
+        if has_rust_up() {
+            let output = cmd!("rustup", "target", "list", "--installed").read()?;
+            Ok(output.contains("wasm32-unknown-unknown"))
+        } else {
+            // When `rustup` is not installed, we need to manually check the [Target] inside the sysroot directory
+            let root = cmd!("rustc", "--print", "sysroot").read()?;
+            Path::new(&format!("{}/lib/rustlib/{}", root.trim(), "wasm32-unknown-unknown"))
+                .try_exists()
+                .map_err(|err: io::Error| anyhow::anyhow!(err))
+        }
+    };
+
+    result().unwrap_or_else(|err| {
+        eprintln!("Error checking for wasm32 target: {err}");
+        false
+    })
 }
 
 #[cfg(test)]
@@ -62,7 +64,7 @@ mod tests {
 
     #[test]
     fn test_has_target() -> anyhow::Result<()> {
-        assert!(has_wasm32_target()?);
+        assert!(has_wasm32_target());
         Ok(())
     }
 }
--- a/crates/cli/src/detect.rs
+++ b/crates/cli/src/detect.rs
@@ -1,0 +1,68 @@
+use duct::cmd;
+use std::io;
+use std::path::Path;
+
+/// Find an executable in the `PATH`.
+pub(crate) fn find_executable(exe_name: impl AsRef<Path>) -> Option<std::path::PathBuf> {
+    std::env::var_os("PATH").and_then(|paths| {
+        std::env::split_paths(&paths)
+            .filter_map(|dir| {
+                let full_path = dir.join(&exe_name);
+                full_path.is_file().then_some(full_path)
+            })
+            .next()
+    })
+}
+
+/// Check if `rustup` is installed (aka: Is in the `PATH`).
+pub(crate) fn has_rust_up() -> anyhow::Result<bool> {
+    Ok(match std::env::consts::OS {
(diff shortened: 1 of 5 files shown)


crates/cli/src/detect.rs
```diff
@@ -0,0 +1,68 @@
...
+            .map_err(|err: io::Error| anyhow::anyhow!(err))
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_has_target() -> anyhow::Result<()> {
```
kazimuth, line 64: It would be cool if we could, like, configure a system here with both configurations and check, but  that's not really feasible in a Rust test. I think this test is ok as-is, assuming we have wasm32 on CI.

crates/cli/src/detect.rs
```diff
@@ -0,0 +1,68 @@
+use duct::cmd;
+use std::io;
+use std::path::Path;
+
+/// Find an executable in the `PATH`.
+pub(crate) fn find_executable(exe_name: impl AsRef<Path>) -> Option<std::path::PathBuf> {
```
kazimuth, line 6: There's also https://github.com/harryfei/which-rs/, but it seems to mostly do stuff we don't need, so I think this is fine for now.

kazimuth (approved):
👍

crates/cli/src/detect.rs
```diff
@@ -0,0 +1,68 @@
...
+    std::env::var_os("PATH").and_then(|paths| {
+        std::env::split_paths(&paths)
+            .filter_map(|dir| {
+                let full_path = dir.join(&exe_name);
+                full_path.is_file().then_some(full_path)
+            })
+            .next()
```
RReverser, line 13: Nit: would be slightly cleaner IMO

```suggestion
            .map(|dir| dir.join(&exe_name))
            .find(Path::is_file)
```

crates/cli/src/detect.rs
```diff
@@ -0,0 +1,68 @@
...
+    })
+}
+
+/// Check if `rustfmt` is installed (aka: Is in the `PATH`).
+pub(crate) fn has_rust_fmt() -> anyhow::Result<bool> {
+    Ok(match std::env::consts::OS {
+        "linux" | "freebsd" | "netbsd" | "openbsd" | "solaris" | "macos" => find_executable("rustfmt").is_some(),
+        "windows" => find_executable("rustfmt.exe").is_some(),
+        unsupported_os => {
+            return Err(anyhow::anyhow!(
```
RReverser, line 36: I think we'd want to just return `false` here - if OS is unsupported by Rustup, that doesn't mean it's not supported by Rust altogether, so no need to return an error.

crates/cli/src/detect.rs
```diff
@@ -0,0 +1,68 @@
...
+    })
+}
+
+/// Check if `rustup` is installed (aka: Is in the `PATH`).
+pub(crate) fn has_rust_up() -> anyhow::Result<bool> {
+    Ok(match std::env::consts::OS {
+        "linux" | "freebsd" | "netbsd" | "openbsd" | "solaris" | "macos" => find_executable("rustup").is_some(),
+        "windows" => find_executable("rustup.exe").is_some(),
+        unsupported_os => {
+            return Err(anyhow::anyhow!(
```
RReverser, line 23: I think we'd want to just return `false` here - if OS is unsupported by Rustup, that doesn't mean it's not supported by Rust altogether, so no need to return an error.

mamcx: But what is the recovery with `false` in this case? The callers are asking `dustup` because is the one with the functionality to proceed (or not). Is only for detect the presence of a target that exist a know workaround.

RReverser: This function is called from `has_wasm32_target` which has an alternative branch that goes via regular `rustc`. If Rustup is not available for this platform, then that 2nd branch should still be taken instead of entire `has_wasm32_target` returning an error.

RReverser: Basically, it's already up to the caller to decide what to do in `false` case. I'm just saying "Rustup is not present on this platform" and "Rustup _cannot_ be present on this platform" should have the same behaviour from caller's PoV.

crates/cli/src/detect.rs
```diff
@@ -6,54 +6,56 @@ use std::path::Path;
...
-            .map_err(|err: io::Error| anyhow::anyhow!(err))
-    }
+pub(crate) fn has_wasm32_target() -> bool {
+    let result = || {
+        if has_rust_up() {
+            let output = cmd!("rustup", "target", "list", "--installed").read()?;
+            Ok(output.contains("wasm32-unknown-unknown"))
+        } else {
+            // When `rustup` is not installed, we need to manually check the [Target] inside the sysroot directory
+            let root = cmd!("rustc", "--print", "sysroot").read()?;
```
RReverser, line 48: You can use eg `rustc --print target-libdir --target wasm32-unknown-unknown` instead.

crates/cli/src/detect.rs
```diff
@@ -6,54 +6,56 @@ use std::path::Path;
...
-    } else {
-        // When `rustup` is not installed, we need to manually check the [Target] inside the sysroot directory
-        let root = cmd!("rustc", "--print", "sysroot").read()?;
-        Path::new(&format!("{}/lib/rustlib/{}", root.trim(), "wasm32-unknown-unknown"))
-            .try_exists()
-            .map_err(|err: io::Error| anyhow::anyhow!(err))
-    }
+pub(crate) fn has_wasm32_target() -> bool {
+    let result = || {
+        if has_rust_up() {
```
RReverser, line 43: Actually, why do we need a separate branch for Rustup in this function? Rustc check should cover both Rustup and non-Rustup cases, right?

mamcx: I worry that `rustc` flags were not stable, so [I checked](https://www.reddit.com/r/rust/comments/1gvrnp3/can_rustc_flags_be_considered_stable_interfaces/) and you are right, so remove the branch.

crates/cli/src/subcommands/generate/mod.rs
```diff
@@ -471,7 +472,15 @@ impl WasmCtx {
...
     match lang {
         Language::Rust => {
-            cmd!("rustup", "component", "add", "rustfmt").run()?;
+            if !has_rust_fmt() {
+                if has_rust_up() {
+                    if let Err(err) = cmd!("rustup", "component", "add", "rustfmt").run() {
+                        println!("Warning: Failed to install rustfmt: {err}");
+                    }
+                } else {
+                    return Err(anyhow::anyhow!("rustfmt is not installed. Please install it."));
```
bfops, line 481: it's a bit strange to me that if we _fail_ to install `rustfmt` above, we just print a warning, but if we don't have `rustup` at all then we return an error. Should we print a warning in this case too, or should we return an `Err` above?

crates/cli/src/tasks/rust.rs
```diff
@@ -20,12 +21,15 @@ fn cargo_cmd(subcommand: &str, build_debug: bool, args: &[&str]) -> duct::Expres
...
-            err
-        );
+    if !has_wasm32_target() {
+        if has_rust_up() {
+            // Make sure that we have the wasm target installed (ok to run if its already installed)
+            if let Err(err) = cmd!("rustup", "target", "add", "wasm32-unknown-unknown").run() {
+                println!("Warning: Failed to install wasm32-unknown-unknown target: {}", err);
+            }
+        } else {
+            anyhow::bail!("wasm32-unknown-unknown target is not installed. Please install it.");
```
bfops, line 31: some comment here as above - strange that we `bail` if rustup is missing, but only warn if rustup fails to install the target.

crates/cli/src/subcommands/generate/mod.rs
```diff
@@ -471,7 +472,15 @@ impl WasmCtx {
 fn format_files(generated_files: Vec<PathBuf>, lang: Language) -> anyhow::Result<()> {
     match lang {
         Language::Rust => {
-            cmd!("rustup", "component", "add", "rustfmt").run()?;
+            if !has_rust_fmt() {
+                if has_rust_up() {
+                    if let Err(err) = cmd!("rustup", "component", "add", "rustfmt").run() {
+                        return Err(anyhow::anyhow!("Failed to install rustfmt with rustup: {err}"));
```
RReverser, line 478: A better way to write this would be `.run().context("Failed to install rustfmt with Rustup")?`.

crates/cli/src/subcommands/generate/mod.rs
```diff
@@ -471,7 +473,15 @@ impl WasmCtx {
...
     match lang {
         Language::Rust => {
-            cmd!("rustup", "component", "add", "rustfmt").run()?;
+            if !has_rust_fmt() {
+                if has_rust_up() {
+                    cmd!("rustup", "component", "add", "rustfmt")
+                        .run()
+                        .context("Failed to install rustfmt with Rustup")?;
+                } else {
+                    return Err(anyhow::anyhow!("rustfmt is not installed. Please install it."));
```
bfops, line 482: nit: `anyhow::bail` like we do below?

crates/cli/src/tasks/rust.rs
```diff
@@ -20,12 +21,15 @@ fn cargo_cmd(subcommand: &str, build_debug: bool, args: &[&str]) -> duct::Expres
...
 pub(crate) fn build_rust(project_path: &Path, skip_clippy: bool, build_debug: bool) -> anyhow::Result<PathBuf> {
-    // Make sure that we have the wasm target installed (ok to run if its already installed)
-    if let Err(err) = cmd!("rustup", "target", "add", "wasm32-unknown-unknown").run() {
-        println!(
-            "Warning: Failed to install wasm32-unknown-unknown target: {}. Is `rustup` installed?",
-            err
-        );
+    if !has_wasm32_target() {
+        if has_rust_up() {
+            // Make sure that we have the wasm target installed (ok to run if its already installed)
```
bfops, line 26: at least part of this comment is no longer relevant (since if we run this code, we have already checked that the wasm32 target is _not_ installed)

bfops (approved):
LGTM

PR #2002, opened 2024-11-19
Preserve the comments of the "cli.toml" on changes

### LINKED ISSUE #1798: CLI - Preserve comments when mutating the config file
# Background
We currently parse, update, and write out the entire config file whenever the CLI is mutating the file. This has the side effect of removing any comments.

We would like to preserve a user's manual edits, including comments, whenever possible.

# Definition of "Done"
- We use `toml_edit` or some similar crate in order to apply edits to the spacetime config file.
- When the CLI mutates the config, user comments are preserved.

mamcx:
# Description of Changes

We would like to preserve a user's manual edits, including comments, so now 'tolm_edit' is used for mutating the config file.

Closes #1798.

Marked breaking because of the change of crate to parse and save. Is the low-level used by 'toml'.

2:

Marked breaking because of the change of how to parse and save the file (that mutates instead of generating a fresh immutable view of the config) even if according to test there is none.

# Testing

- [x] *Write unit testing that simulates different edits.*
- [x] *Run 'standalone' and execute several 'cli' commands to see manually if the config was updated without issues.*


crates/cli/src/config.rs
```diff
@@ -367,6 +390,41 @@ Fetch the server's fingerprint with:
...
+                        let nickname = table.get("nickname").and_then(Item::as_str).map(String::from);
+                        let host = table.get("host").and_then(Item::as_str).map(String::from).unwrap();
+                        let protocol = table.get("protocol").and_then(Item::as_str).map(String::from).unwrap();
+                        let ecdsa_public_key = table.get("ecdsa_public_key").and_then(Item::as_str).map(String::from);
+                        ServerConfig {
+                            nickname,
+                            host,
+                            protocol,
+                            ecdsa_public_key,
+                        }
```
bfops, line 414: nit: Should we have a `ServerConfig::from_table`?

crates/cli/src/config.rs
```diff
@@ -511,15 +570,69 @@ impl Config {
...
             home: RawConfig::new_with_localhost(),
             home_path,
+            doc: Default::default(),
         }
     }
 
+    /// Returns a preserving copy of [Config].
+    fn doc(&self) -> DocumentMut {
+        let mut doc = self.doc.clone();
+
```
bfops, line 580: nit: can we do something like `let RawConfig { ... } = home` so people don't forget to update this function if they add more fields?

crates/cli/src/config.rs
bfops:
```rust
    pub fn from_table:(table: &Table) -> Self {
        let nickname = table.get("nickname").and_then(Item::as_str).map(String::from);
        let host = table.get("host").and_then(Item::as_str).map(String::from).unwrap();
        let protocol = table.get("protocol").and_then(Item::as_str).map(String::from).unwrap();
        let ecdsa_public_key = table.get("ecdsa_public_key").and_then(Item::as_str).map(String::from);
        ServerConfig {
            nickname,
            host,
            protocol,
            ecdsa_public_key,
        }
    }
```

mamcx: Done

crates/cli/src/config.rs
mamcx: Done

crates/core/src/config.rs
```diff
@@ -93,8 +122,39 @@ pub struct LogConfig {
...
-    pub fn read(path: &ConfigToml) -> anyhow::Result<Option<Self>> {
-        parse_config(path.as_ref())
+/// Patch the value of a key in a TOML document,
+/// preserving the formatting and comments of the original value.
+fn patch_value(item: Option<&Item>, value: Option<&str>) -> Option<Item> {
+    match (value, item) {
+        (Some(value), Some(Item::Value(Value::String(v)))) => {
+            let mut new = Value::String(Formatted::new(value.to_string()));
+            let decor = new.decor_mut();
+            *decor = v.decor().clone();
```
bfops, line 132: why do we only do this `decor` dance if the original value is a string? It looks like any `Table` can have `decor`, is that right?

crates/core/src/config.rs
```diff
@@ -93,8 +122,39 @@ pub struct LogConfig {
...
+            Some(new.into())
+        }
+        (Some(val), _) => Some(val.into()),
+        (None, _) => None,
```
bfops, line 136: nit: the caller is just going to match on the returned value anyway. The only time it's `None` is if `value` was `None`, so it seems easier to have the caller match on `value` before calling this, instead of the return value after calling this. Then it would be `fn patch_value(item: Option<&item>, value: &str) -> Item`, which seems a little simpler.

crates/core/src/config.rs
```diff
@@ -93,8 +122,39 @@ pub struct LogConfig {
     pub directives: Vec<String>,
 }
 
-impl ConfigFile {
-    pub fn read(path: &ConfigToml) -> anyhow::Result<Option<Self>> {
-        parse_config(path.as_ref())
+/// Patch the value of a key in a TOML document,
+/// preserving the formatting and comments of the original value.
+fn patch_value(item: Option<&Item>, value: Option<&str>) -> Option<Item> {
```
bfops, line 127: nit: maybe we could call this something like `update_value_preserve_decor` or something? I didn't really understand the point of `patch_value` without reading through the cases.

crates/core/src/config.rs
```diff
@@ -1,18 +1,37 @@
...
 use spacetimedb_lib::Address;
 use spacetimedb_paths::cli::{ConfigDir, PrivKeyPath, PubKeyPath};
 use spacetimedb_paths::server::{ConfigToml, MetadataTomlPath};
 
-#[derive(serde::Serialize, serde::Deserialize)]
-pub struct MetadataFile {
-    pub version: semver::Version,
-    pub edition: String,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub client_address: Option<Address>,
```
bfops, line 13: nit: this diff is a little harder to review because several things moved without changing, like `MetadataFile`, `current_version`, and `impl ConfigFile`.

It's fine, I was able to review it, but in the future maybe those changes could be done separately?

crates/core/src/config.rs
```diff
@@ -93,8 +122,39 @@ pub struct LogConfig {
...
+        doc[key] = value;
+    } else {
+        doc.remove(key);
+    }
+}
+
+/// Set the value of a key in a `TOML` table, removing the key if the value is `None`.
+///
+/// **NOTE**: This function will preserve the formatting and comments of the original value.
+pub fn set_table_opt_value(table: &mut Table, key: &str, value: Option<&str>) {
```
bfops, line 154: nit: why are these functions in this file? It looks like nothing else in this file has anything to do with tomls, so it's a bit surprising that they're added here.

crates/cli/src/config.rs
```diff
@@ -511,15 +586,76 @@ impl Config {
...
+        set_value(WEB_SESSION_TOKEN_KEY, web_session_token.as_deref());
+        set_value(SPACETIMEDB_TOKEN_KEY, spacetimedb_token.as_deref());
+
+        // Short-circuit if there are no servers.
+        if server_configs.is_empty() {
+            doc.remove(SERVER_CONFIGS_KEY);
+            return doc;
+        }
+        // ... or if there are no server_configs to edit.
+        let server_configs = if let Some(cfg) = doc.get_mut(SERVER_CONFIGS_KEY).and_then(Item::as_array_of_tables_mut) {
```
bfops, line 618: nit: I got confused a few times because here we shadow `server_configs`; the first time we use `server_configs` above, it's the _new_ server configs. Here, we shadow it to be the _old_ server configs. Then, below, we read the new server configs out of `self.home` again. Maybe we could just use the names `new_server_configs` and `old_server_configs` and not do any shadowing?

bfops (approved):
left lots of nits, none of them critical

crates/cli/src/config.rs
```diff
@@ -1,11 +1,22 @@
...
 use anyhow::Context;
 use jsonwebtoken::DecodingKey;
-use serde::{Deserialize, Serialize};
+use spacetimedb::config::{set_opt_value, set_table_opt_value};
 use spacetimedb_fs_utils::atomic_write;
 use spacetimedb_paths::cli::CliTomlPath;
-
-#[derive(Serialize, Deserialize, Clone, Debug)]
+use std::collections::HashMap;
+use toml_edit::{ArrayOfTables, DocumentMut, Item, Table};
```
gefjon, line 8: I would prefer to see these, especially `Table`, used by full path or with a qualified import. `Item` and `Table` in the code below are at best unhelpful names, and at worst actively confusing, given that we use the word "Table" to mean something very different in most of our code. I would do `use toml_edit as toml;` and then write `toml::Table`, `toml::DocumentMut`, &c. below.

crates/cli/src/config.rs
```diff
@@ -14,6 +25,21 @@ pub struct ServerConfig {
 }
 
 impl ServerConfig {
+    /// Generate a new [Table] representing this [ServerConfig].
```
gefjon, line 28:
```suggestion
    /// Generate a new [`Table`] representing this [`ServerConfig`].
```

crates/cli/src/config.rs
```diff
@@ -14,6 +25,21 @@ pub struct ServerConfig {
...
+    pub fn as_table(&self) -> Table {
+        let mut table = Table::new();
+        Self::update_table(&mut table, self);
+        table
+    }
+
+    /// Update an existing [Table] with the values of a [ServerConfig].
```
gefjon, line 35:
```suggestion
    /// Update an existing [`Table`] with the values of a [`ServerConfig`].
```

This needs a paragraph describing its specific behavior. What data from the existing `Table` are preserved? What are overwritten?

crates/cli/src/config.rs
```diff
@@ -33,10 +59,29 @@ impl ServerConfig {
...
+            .unwrap();
+        let ecdsa_public_key = table.get(ECDSA_PUBLIC_KEY).and_then(Item::as_str).map(String::from);
+        ServerConfig {
+            nickname,
+            host,
+            protocol,
+            ecdsa_public_key,
+        }
+    }
+}
```
gefjon, line 79: Should `ServerConfig` really be `From<&Table>`? Can we convert *any* `Table` into a meaningful `ServerConfig`? It seems to me that this should more likely be a `TryFrom` impl, though I am not sure. E.g. what happens if I do:
```rust
let table = toml::Table::new();
let server_config = ServerConfig::from(&table);
```

crates/cli/src/config.rs
```diff
@@ -33,10 +59,29 @@ impl ServerConfig {
...
+
+// Any change here must be coordinated with Config::doc
```
gefjon, line 81: Please expand this, and make it into a doc comment. Do you mean:

- Changes at the source level by SpacetimeDB developers to this type definition?
- Changes at runtime to the data contained in a `RawConfig` instance?

Who is responsible for the coordination? In our actual uses of this type, how do those changes propagate?

crates/cli/src/config.rs
```diff
@@ -48,6 +93,9 @@ pub struct RawConfig {
 pub struct Config {
     home: RawConfig,
     home_path: CliTomlPath,
+    // The TOML document that was parsed to create `home`.
+    // We need to keep it to preserve comments and formatting when saving the config.
```
gefjon, line 97:
```suggestion
    /// The TOML document that was parsed to create `home`.
    ///
    /// We need to keep it to preserve comments and formatting when saving the config.
```

crates/cli/src/config.rs
```diff
@@ -367,6 +415,32 @@ Fetch the server's fingerprint with:
...
+            .unwrap_or_default();
+
+        RawConfig {
+            default_server,
+            server_configs,
+            web_session_token,
+            spacetimedb_token,
+        }
+    }
+}
```
gefjon, line 443: Similar questions here as on the `From<&Table> for ServerConfig`.

crates/core/src/config.rs
```diff
@@ -1,18 +1,37 @@
...
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub client_address: Option<Address>,
+pub fn current_version() -> semver::Version {
+    env!("CARGO_PKG_VERSION").parse().unwrap()
+}
+
+/// Parse a TOML file at the given path, returning `None` if the file does not exist.
+///
+/// **NOTE**: Comments and formatting in the file could be preserved.
+pub fn parse_preserving_config<T: for<'a> From<&'a DocumentMut>>(
```
gefjon, line 18: This is a silly name. Of course parsing the config preserves the config. Do you want to call this `parse_config_preserving_formatting` or something?

gefjon: Also, why is this given such a general type? Can we not just have a function that parses a `RawConfig`? Or, failing that, why is this not just a function that reads a `DocumentMut` from a file, and then the caller gets to parse it into whatever `T` they want on their own?

crates/cli/src/config.rs
```diff
@@ -511,15 +586,76 @@ impl Config {
         Self {
             home: RawConfig::new_with_localhost(),
             home_path,
+            doc: Default::default(),
         }
     }
 
+    /// Returns a preserving copy of [Config].
```
gefjon, line 593: Preserving what? Also, use backticks when linking to code names in doc comments:

```suggestion
    /// Returns a preserving copy of [`Config`].
```

crates/core/src/config.rs
gefjon: This file does contain some code that *reads* TOML files. See e.g. `MetadataFile::read` above. That said, the only place we ever *write* TOML files AFAICT is from the CLI, so I think these functions for writing TOML should be moved into the CLI code.

crates/core/src/config.rs
gefjon: Also, if this is only going to be called by the CLI, then it should be in the CLI's config.rs, not here.

crates/cli/src/config.rs
```diff
@@ -511,15 +586,76 @@ impl Config {
...
+        };
+
+        let mut new_configs = self
+            .home
+            .server_configs
+            .iter()
+            .map(|cfg| (cfg.nick_or_host(), cfg))
+            .collect::<HashMap<_, _>>();
+
+        // Update the existing servers.
```
gefjon, line 632:
```suggestion
        // Update the existing servers, and remove deleted servers.
        // We'll add new servers later.
        // We do this somewhat elaborate dance rather than just overwriting the config
        // in order to preserve the order and formatting of pre-existing server configs in the file.
```

crates/cli/src/config.rs
```diff
@@ -511,15 +586,76 @@ impl Config {
...
+                .or_else(|| old_config.get(HOST_KEY))
+                .and_then(|v| v.as_str())
+                .unwrap();
+            if let Some(new_config) = new_configs.remove(nick_or_host) {
+                ServerConfig::update_table(old_config, new_config);
+                new_vec.push(old_config.clone());
+            }
+        }
+
+        // Add the new servers.
```
gefjon, line 646:
```suggestion
        // Add the new servers. This appends them to the end of the config file,
        // after the (preserved) existing configs.
```

gefjon (changes requested):
Additional (manual) testing I would like to see done:

- Run multiple CLI processes concurrently / in parallel, such that they race to mutate a shared config file. Demonstrate that this does not corrupt or clear the config file, and describe the specific behavior.
- Manually create an ill-formed config file, run a CLI process that would mutate it, and demonstrate that the ill-formed file is not (further) corrupted or cleared. Describe the specific behavior.

crates/core/src/config.rs
bfops: Good point

crates/core/src/config.rs
mamcx: > This is a silly name. Of course, parsing the config preserves the config. Do you want to call this `parse_config_preserving_formatting` or something?

This is to make a difference with `parse_config` that uses the regular `serde` deserialization that doesn't preserve the formatting.

crates/core/src/config.rs
mamcx: I consider that keeping all the stuff that touches `toml` in one place makes life easier.

crates/core/src/config.rs
mamcx: > Also, why is this given such a general type? Can we not just have a function that parses a `RawConfig`? Or, failing that, why is this not just a function that reads a `DocumentMut` from a file, and then the caller gets to parse it into whatever `T` they want on their own?

I do this in case other files need preserving, and because `RawConfig` is not accessible here.

Then, because `parse_config` is here too, move this out will hide later the fact we have 2 ways to read it.

mamcx: We decided to move this because the need for a generic version was considered unnecessary for now.

crates/core/src/config.rs
mamcx: It will need that logic duplicated.

crates/core/src/config.rs
bfops: I think I'm asking why this _only_ applies when the item is a string, since it seems like all `Table`s have `decor`? (https://docs.rs/toml_edit/latest/toml_edit/struct.Table.html#method.decor_mut)

Does `decor` not apply to anything besides strings?

crates/core/src/config.rs
mamcx: Done

crates/core/src/config.rs
mamcx: Done

mamcx:
> Additional (manual) testing I would like to see done:
> 
> * Run multiple CLI processes concurrently / in parallel, such that they race to mutate a shared config file. Demonstrate that this does not corrupt or clear the config file, and describe the specific behavior.
> * Manually create an ill-formed config file, run a CLI process that would mutate it, and demonstrate that the ill-formed file is not (further) corrupted or cleared. Describe the specific behavior.

I added an automated `test` instead.

crates/core/src/config.rs
mamcx: I added a test that confirms the comments are preserved for tables as-is. I agree it looks weird it works

crates/cli/src/config.rs
mamcx: Done

crates/cli/src/config.rs
mamcx: Done

crates/cli/src/config.rs
mamcx: Done

bfops (approved):
Thank you for addressing those comments!

gefjon (approved):
Looks great!

PR #2007, opened 2024-11-21
Fix declaration order of ScheduleAt in C#

kazimuth:
And add some tests to sdk-test-cs. This was causing `ScheduleAt` to not get recognized by the schema validation code.

These tests don't fully exercise scheduling/indexes yet, but at least they check if they compile & are accepted by the host.

0

# Testing

Added tests.


RReverser:
Hm this doesn't look right. It seems it's Rust that is wrong here as C# implemented the enum according to the spec: https://github.com/clockworklabs/SpacetimeDBPrivate/blob/592a34c0cb82d587843d2b2309b824e94c035d26/proposals/0005-timers.md?plain=1#L82

Or, I guess, we can do this PR, but then we need to change the spec as well.

crates/bindings-csharp/BSATN.Runtime/Builtins.cs
```diff
@@ -349,8 +349,8 @@ public AlgebraicType GetAlgebraicType(ITypeRegistrar registrar) =>
             // to avoid leaking the internal *Repr wrappers in generated SATS.
             new AlgebraicType.Sum(
                 [
-                    new("Time", new AlgebraicType.U64(default)),
                     new("Interval", new AlgebraicType.U64(default)),
+                    new("Time", new AlgebraicType.U64(default)),
```
RReverser, line 353: FWIW this is the only necessary reordering, the rest don't matter and can be safely reverted.

kazimuth:
Ah whoops. I think it's probably easier to change the spec at this point as it's less of a breaking change.

crates/bindings-csharp/BSATN.Runtime/Builtins.cs
kazimuth: I figured, but I thought we might as well update all of them for consistency.

kazimuth:
https://github.com/clockworklabs/SpacetimeDBPrivate/pull/1180

RReverser (approved):

PR #2011, opened 2024-11-26
Split binaries

### LINKED ISSUE #1838: Directory structure - Split `spacetime` into multiple binaries
# Background
We would like to split `spacetime` into `spacetimedb-cli` and `spacetimedb-standalone`.

# Definition of "Done"
TODO

coolreader18:
# Description of Changes

Resolves #1838

Adds a `spacetimedb-update` multicall binary that proxies to `spacetimedb-cli`, and makes `spacetime run` call out to the `spacetimedb-standalone` binary instead of embedding the whole server in the cli. Functionality for the `update` binary itself forthcoming in another PR.

3 - adds the proxy binary and changes how `run` works under-the-hood, but the deeper changes will come in the follow-up.

# Testing

- [x] Smoketests use new proxy binary and work.
- [x] Manually set up a test environment and `spacetimedb-update` proxied to the cli binary.

### DIFF
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -4,20 +4,19 @@ mod config;
 pub(crate) mod detect;
 mod edit_distance;
 mod errors;
+mod start;
 mod subcommands;
 mod tasks;
 pub mod util;
+
+use std::process::ExitCode;
 
 use clap::{ArgMatches, Command};
 
 pub use config::Config;
 use spacetimedb_paths::SpacetimePaths;
-use spacetimedb_standalone::subcommands::start::ProgramMode;
 pub use subcommands::*;
 pub use tasks::build;
-
-#[cfg(feature = "standalone")]
-use spacetimedb_standalone::subcommands::start;
 
 pub fn get_subcommands() -> Vec<Command> {
     vec![
@@ -39,8 +38,7 @@ pub fn get_subcommands() -> Vec<Command> {
         server::cli(),
         upgrade::cli(),
         subscribe::cli(),
-        #[cfg(feature = "standalone")]
-        start::cli(ProgramMode::CLI),
+        start::cli(),
     ]
 }
 
@@ -49,7 +47,7 @@ pub async fn exec_subcommand(
     paths: &SpacetimePaths,
     cmd: &str,
     args: &ArgMatches,
-) -> Result<(), anyhow::Error> {
+) -> anyhow::Result<ExitCode> {
     match cmd {
         "version" => version::exec(config, args).await,
         "call" => call::exec(config, args).await,
@@ -66,11 +64,11 @@ pub async fn exec_subcommand(
         "build" => build::exec(config, args).await.map(drop),
         "server" => server::exec(config, paths, args).await,
         "subscribe" => subscribe::exec(config, args).await,
-        #[cfg(feature = "standalone")]
-        "start" => start::exec(Some(paths), args).await,
+        "start" => return start::exec(paths, args).await,
         "login" => login::exec(config, args).await,
         "logout" => logout::exec(config, args).await,
         "upgrade" => upgrade::exec(config, args).await,
         unknown => Err(anyhow::anyhow!("Invalid subcommand: {}", unknown)),
     }
+    .map(|()| ExitCode::SUCCESS)
 }
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -1,3 +1,5 @@
+use std::process::ExitCode;
+
 use clap::{Arg, Command};
 use mimalloc::MiMalloc;
 use spacetimedb_cli::*;
@@ -8,7 +10,7 @@ use spacetimedb_paths::{RootDir, SpacetimePaths};
 static GLOBAL: MiMalloc = MiMalloc;
 
 #[tokio::main]
-async fn main() -> Result<(), anyhow::Error> {
+async fn main() -> anyhow::Result<ExitCode> {
     // Compute matches before loading the config, because `Config` has an observable `drop` method
     // (which deletes a lockfile),
     // and Clap calls `exit` on parse failure rather than panicing, so destructors never run.
@@ -25,9 +27,7 @@ async fn main() -> Result<(), anyhow::Error> {
         .unwrap_or_else(|| paths.cli_config_dir.cli_toml());
     let config = Config::load(cli_toml)?;
 
-    exec_subcommand(config, &paths, cmd, subcommand_args).await?;
-
-    Ok(())
+    exec_subcommand(config, &paths, cmd, subcommand_args).await
 }
 
 fn get_command() -> Command {
--- a/crates/cli/src/start.rs
+++ b/crates/cli/src/start.rs
@@ -1,0 +1,68 @@
+use std::ffi::OsString;
+use std::process::{Command, ExitCode};
+
+use anyhow::Context;
+use clap::{Arg, ArgMatches};
+use spacetimedb_paths::SpacetimePaths;
+
+pub fn cli() -> clap::Command {
+    clap::Command::new("start")
+        .about("Start a local SpacetimeDB instance")
+        .disable_help_flag(true)
+        .arg(
+            Arg::new("edition")
+                .long("edition")
+                .help("The edition of SpacetimeDB to start up")
+                .value_parser(clap::value_parser!(Edition))
+                .default_value("standalone"),
+        )
+        .arg(
+            Arg::new("args")
+                .help("The args to pass to `spacetimedb-{edition} start`")
+                .value_parser(clap::value_parser!(OsString))
+                .allow_hyphen_values(true)
+                .num_args(0..),
+        )
+}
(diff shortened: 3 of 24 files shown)


crates/update/src/main.rs
```diff
@@ -0,0 +1,95 @@
...
+        None => SpacetimePaths::platform_defaults()?,
+    };
+    match args.cmd {
+        Subcommand::Cli { args: mut cli_args } => {
+            if let Some(root_dir) = &args.root_dir {
+                cli_args.insert(0, OsString::from_iter(["--root-dir=".as_ref(), root_dir.as_ref()]));
+            }
+            proxy::run_cli(&paths, None, cli_args)
+        }
+        _ => unimplemented!(),
```
cloutiertyler, line 93: Is the idea that we'll implement these in another PR? Note that `spacetimedb upgrade` is already a thing. We probably want to remove that with that PR.

gefjon:
Needs PR description according to template.

gefjon:
PR description should include what the correct way for a reviewer to invoke the CLI is now. In the past I've been doing `cargo run --manifest-path ~/clockworklabs/SpacetimeDB/Cargo.toml -- $@`. This appears to not hit the new functionality. Changing the manifest-path to `SpaceitmeDB/crates/update/Cargo.toml` appears to not exercise the new functionality either, as that binary is looking to be invoked as `spacetime`. `cargo install --path SpacetimeDB/crates/update` *also* does not allow me to test the new functionality, as it installs as `spacetimedb-update`.

I also notice that you've updated the smoketests to invoke the inner `spacetimedb-cli` rather than the `spacetime` wrapper, which concerns me, as it seems the new functionality is untested.

gefjon (approved):
Looks great, thanks!

coolreader18:
@bfops @Centril could you review as codeowners?

crates/cli/src/subcommands/subscribe.rs
```diff
@@ -144,7 +143,10 @@ pub async fn exec(config: Config, args: &ArgMatches) -> Result<(), anyhow::Error
 
     // Create the websocket request.
     let mut req = http::Uri::from_parts(uri)?.into_client_request()?;
-    req.headers_mut().insert(header::SEC_WEBSOCKET_PROTOCOL, TEXT_PROTOCOL);
+    req.headers_mut().insert(
+        header::SEC_WEBSOCKET_PROTOCOL,
+        http::HeaderValue::from_static(ws::TEXT_PROTOCOL),
+    );
```
bfops, line 149: why did this need to change?

crates/cli/src/main.rs
```diff
@@ -8,7 +10,7 @@ use spacetimedb_paths::{RootDir, SpacetimePaths};
 static GLOBAL: MiMalloc = MiMalloc;
 
 #[tokio::main]
-async fn main() -> Result<(), anyhow::Error> {
+async fn main() -> anyhow::Result<ExitCode> {
```
bfops, line 13: (ditto above)

crates/cli/src/start.rs
```diff
@@ -0,0 +1,68 @@
...
+        Err(err).context(format!("exec failed for {}", bin_path.display()))
+    }
+    #[cfg(windows)]
+    {
+        use std::os::windows::process::ExitCodeExt;
+        let status = cmd
+            .status()
+            .with_context(|| format!("failed to run {}", cli_path.display()))?;
+        Ok(ExitCode::from_raw(status.code().unwrap_or(1) as u32))
+    }
```
bfops, line 67: do we need a variant for mac?

crates/cli/src/subcommands/subscribe.rs
bfops: ah I see - this didn't just get moved, it went from having `http::HeaderValue` already baked in, to being a raw string.

(I'm curious why this change had to happen in this PR, but this seems fine as far as my code-owned files are concerned.)

coolreader18: Because `client_api_messages` doesn't depend on `http`, so with removing cli's dependency on `client_api`, it now has to be wrapped in cli.

crates/cli/src/lib.rs
```diff
@@ -49,7 +47,7 @@ pub async fn exec_subcommand(
     paths: &SpacetimePaths,
     cmd: &str,
     args: &ArgMatches,
-) -> Result<(), anyhow::Error> {
+) -> anyhow::Result<ExitCode> {
```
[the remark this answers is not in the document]
coolreader18: Oh, I think I was probably playing around with rewriting these and when I went to change it back I typed it this way, since in the rest of the codebase that's what it always is.

crates/cli/src/start.rs
coolreader18: mac is unix

crates/cli/src/subcommands/subscribe.rs
bfops: okay, then why did we have to remove the CLI's dependency on `client_api` in this PR?

bfops:
I would ideally like to confirm that this has been tested on windows and mac before merging, but that feels out of scope for a codeowner review and @gefjon's already approved.

crates/cli/src/lib.rs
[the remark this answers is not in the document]
bfops: Ah got it! FWIW I think it's good PR-hygiene to review the diff before opening a PR for review, but again this seems completely fine :+1:

crates/cli/src/start.rs
bfops: oh blerp of course thank you!

crates/cli/src/start.rs
```diff
@@ -0,0 +1,68 @@
...
+        use std::os::unix::process::CommandExt;
+        let err = cmd.exec();
+        Err(err).context(format!("exec failed for {}", bin_path.display()))
+    }
+    #[cfg(windows)]
+    {
+        use std::os::windows::process::ExitCodeExt;
+        let status = cmd
+            .status()
+            .with_context(|| format!("failed to run {}", cli_path.display()))?;
```
bfops, line 65: is `cli_path` defined?

crates/cli/src/start.rs
```diff
@@ -0,0 +1,68 @@
...
+        .args(args);
+    #[cfg(unix)]
+    {
+        use std::os::unix::process::CommandExt;
+        let err = cmd.exec();
+        Err(err).context(format!("exec failed for {}", bin_path.display()))
+    }
+    #[cfg(windows)]
+    {
+        use std::os::windows::process::ExitCodeExt;
```
bfops, line 62: probably a dumb question - is this import needed?

bfops: nevermind I see the `from_raw` below now, I'm dumb

crates/cli/src/start.rs
```diff
@@ -0,0 +1,68 @@
...
+        let status = cmd
+            .status()
+            .with_context(|| format!("failed to run {}", cli_path.display()))?;
+        Ok(ExitCode::from_raw(status.code().unwrap_or(1) as u32))
```
bfops, line 66: why do we return an `Err` in unix but an `Ok(ExitCode)` in Windows? Is there a downside to just returning an `Err` in both cases?

coolreader18: `exec()` never returns except in an error case; it replaces the current running process with the new one. Windows doesn't have an equivalent, so we just run it as a subprocess and return the exit code.

crates/cli/src/start.rs
coolreader18: Ah, nice catch. We should probably run ci on windows

crates/cli/src/subcommands/subscribe.rs
coolreader18: It's part of splitting the binaries; spacetimedb-cli shouldn't need to run the `client_api` http server because that's what the spacetimedb-standalone binary does. The follow up also fully removes cli's dependency on `spacetimedb_core`, which cuts out nearly 100 unnecessary dependencies.

crates/cli/src/start.rs
bfops: Er sorry what I should have asked is *is there a downside to returning an `Err` for all non-success cases? On Windows, we have the option of returning failing exit codes instead of `Err`s. The caller of this function will only get an `ExitCode` if we're on windows, so it seems like we end up having potential for meaningfully different functionality. If we returned `Result<()>` then unix and windows would "look" the same from the caller's point of view.

crates/cli/src/subcommands/subscribe.rs
bfops: oh I'm definitely not saying the CLI should be running the server.. but keeping the compile-time dependency doesn't seem like something that had to be changed in this PR.

Either way, not a big deal, it just added some diff and I wasn't sure if I was missing an underlying reason.

bfops (approved):
The files I'm a codeowner for look great to me. Thanks for bearing with me @coolreader18.

PR #2025, opened 2024-11-28
Add `Layout.fixed: bool` for a fast path in `eq_row_in_page` & `row_type_visitor`

Centril:
# Description of Changes

Adds to `Layout` whether it is the layout of a "fixed type" and not the layout of a var-len type's fixed component.
This is then exploited to 2x-10x speed up `eq_row_in_page` and speed up `row_type_visitor` for the fixed len case at the cost of some minor regression for small VLOs.

This in turn will speed up insertions and subscriptions.

```
Benchmarking eq_in_page/U32: Collecting 100 samples in estimated 5.0000 s (1.5B iter
eq_in_page/U32          time:   [3.4175 ns 3.4689 ns 3.5373 ns]
                        change: [-55.828% -54.348% -52.236%] (p = 0.00 < 0.05)
                        Performance has improved.
Benchmarking eq_in_page/Option<U32>/None: Collecting 100 samples in estimated 5.0000
eq_in_page/Option<U32>/None
                        time:   [3.3766 ns 3.3865 ns 3.4052 ns]
                        change: [-65.726% -65.473% -65.207%] (p = 0.00 < 0.05)
                        Performance has improved.
Benchmarking eq_in_page/Option<U32>/Some: Collecting 100 samples in estimated 5.0000
eq_in_page/Option<U32>/Some
                        time:   [3.3739 ns 3.3803 ns 3.3942 ns]
                        change: [-74.766% -74.605% -74.449%] (p = 0.00 < 0.05)
                        Performance has improved.
Benchmarking eq_in_page/U32x2: Collecting 100 samples in estimated 5.0000 s (1.5B it
eq_in_page/U32x2        time:   [3.3937 ns 3.4211 ns 3.4583 ns]
                        change: [-72.515% -71.694% -71.042%] (p = 0.00 < 0.05)
                        Performance has improved.
Benchmarking eq_in_page/U32x4: Collecting 100 samples in estimated 5.0000 s (1.5B it
eq_in_page/U32x4        time:   [3.3896 ns 3.4206 ns 3.4578 ns]
                        change: [-84.075% -83.915% -83.753%] (p = 0.00 < 0.05)
                        Performance has improved.
Benchmarking eq_in_page/U32x8: Collecting 100 samples in estimated 5.0000 s (1.4B it
eq_in_page/U32x8        time:   [3.6620 ns 3.7123 ns 3.7736 ns]
                        change: [-90.794% -90.690% -90.574%] (p = 0.00 < 0.05)
                        Performance has improved.
Benchmarking eq_in_page/String/0: Collecting 100 samples in estimated 5.0000 s (673M
eq_in_page/String/0     time:   [7.5180 ns 7.5381 ns 7.5708 ns]
                        change: [+5.5214% +6.0313% +6.4757%] (p = 0.00 < 0.05)
                        Performance has regressed.
Benchmarking eq_in_page/String/16: Collecting 100 samples in estimated 5.0000 s (439
eq_in_page/String/16    time:   [10.089 ns 10.194 ns 10.345 ns]
                        change: [+9.1368% +10.153% +11.231%] (p = 0.00 < 0.05)
                        Performance has regressed.
Benchmarking eq_in_page/String/128: Collecting 100 samples in estimated 5.0000 s (31
eq_in_page/String/128   time:   [15.956 ns 16.020 ns 16.148 ns]
                        change: [+3.3239% +4.2398% +5.1960%] (p = 0.00 < 0.05)
                        Performance has regressed.
Benchmarking eq_in_page/String/512: Collecting 100 samples in estimated 5.0001 s (16
eq_in_page/String/512   time:   [29.182 ns 29.352 ns 29.569 ns]
                        change: [-1.7566% -0.9731% -0.3028%] (p = 0.00 < 0.05)
Benchmarking eq_in_page/U32x2x2: Collecting 100 samples in estimated 5.0000 s (1.5B 
eq_in_page/U32x2x2      time:   [3.3729 ns 3.3938 ns 3.4353 ns]
                        change: [-88.097% -88.009% -87.929%] (p = 0.00 < 0.05)
                        Performance has improved.
Benchmarking eq_in_page/U32x4x2: Collecting 100 samples in estimated 5.0000 s (1.4B 
eq_in_page/U32x4x2      time:   [3.6449 ns 3.6607 ns 3.6853 ns]
                        change: [-92.001% -91.920% -91.870%] (p = 0.00 < 0.05)
                        Performance has improved.
Benchmarking eq_in_page/U32x2x4: Collecting 100 samples in estimated 5.0000 s (1.4B 
eq_in_page/U32x2x4      time:   [3.6469 ns 3.6907 ns 3.7494 ns]
                        change: [-92.797% -92.713% -92.629%] (p = 0.00 < 0.05)
                        Performance has improved.
Benchmarking eq_in_page/Option<U32x2>/None: Collecting 100 samples in estimated 5.00
eq_in_page/Option<U32x2>/None
                        time:   [3.3798 ns 3.3966 ns 3.4241 ns]
                        change: [-65.092% -63.961% -61.660%] (p = 0.00 < 0.05)
                        Performance has improved.
Benchmarking eq_in_page/Option<U32x2>/Some: Collecting 100 samples in estimated 5.00
eq_in_page/Option<U32x2>/Some
                        time:   [3.3756 ns 3.4034 ns 3.4457 ns]
                        change: [-81.915% -81.656% -81.353%] (p = 0.00 < 0.05)
                        Performance has improved.
```

None

2, fairly self-contained.

### DIFF
--- a/crates/table/src/eq.rs
+++ b/crates/table/src/eq.rs
@@ -4,6 +4,7 @@
 
 use super::{
     bflatn_from::read_tag,
+    bflatn_to_bsatn_fast_path::StaticBsatnLayout,
     indexes::{Bytes, PageOffset},
     layout::{align_to, AlgebraicTypeLayout, HasLayout, ProductTypeLayout, RowTypeLayout},
     page::Page,
@@ -24,13 +25,14 @@ use super::{
 /// 1. `fixed_offset_a/b` are valid offsets for rows typed at `ty` in `page_a/b`.
 /// 2. for any `vlr_a/b: VarLenRef` in the fixed parts of row `a` and `b`,
 ///   `vlr_a/b.first_offset` must either be `NULL` or point to a valid granule in `page_a/b`.
+/// 3. the `static_bsatn_layout` must be derived from `ty`.
 pub unsafe fn eq_row_in_page(
     page_a: &Page,
     page_b: &Page,
     fixed_offset_a: PageOffset,
     fixed_offset_b: PageOffset,
     ty: &RowTypeLayout,
-    only_fixed_parts: bool,
+    static_bsatn_layout: Option<&StaticBsatnLayout>,
 ) -> bool {
     // Contexts for rows `a` and `b`.
     let a = BytesPage::new(page_a, fixed_offset_a, ty);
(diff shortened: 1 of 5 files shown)


crates/table/src/eq.rs
```diff
@@ -30,14 +30,23 @@ pub unsafe fn eq_row_in_page(
...
+    only_fixed_parts: bool,
 ) -> bool {
+    // Contexts for rows `a` and `b`.
+    let a = BytesPage::new(page_a, fixed_offset_a, ty);
+    let b = BytesPage::new(page_b, fixed_offset_b, ty);
+
+    // If there are only fixed parts in the layout,
+    // there are no pointers to anywhere,
+    // So it is sound to simply check for byte-wise equality
+    // and we need not do the tree traversal at all.
```
gefjon, line 42: I would like to see additional justification here that we don't have to worry about padding bytes within the rows causing false negatives. I believe the code is correct, since we zero-initialize pages and never write to padding bytes, but I think that's non-obvious and important to record. E.g. one could imagine a version of `Page` where padding bytes held arbitrary non-poison values, and therefore this comparison could return false for two equal rows.

crates/table/src/layout.rs
```diff
@@ -329,12 +332,36 @@ impl MemoryUsage for PrimitiveType {}
...
+            Self::I128 | Self::U128 => &Layout {
+                size: 16,
+                align: 16,
+                fixed: true,
+            },
+            Self::I256 | Self::U256 => &Layout {
+                size: 32,
+                align: 32,
+                fixed: true,
+            },
```
gefjon, line 364: Consider pulling this into a constructor `Layout::new_primitive(size_and_align: u16) -> Self { Self { size: size_and_align, align: size_and_align, fixed: true } }`?

crates/table/src/eq.rs
gefjon: Edit: I actually don't think this is correct in the presence of sums with variants that have distinct layouts.

Imagine a row `Product { x: Sum { full: u64, padding: Product { a: u8, b: u32 } } }`.

1. Insert two rows:
  - A_0: `{ x: full(-1i64 as u64) }`.
  - B_0: `{ x: full(u64::ALTERNATING_BIT_PATTERN) }`.
2. Delete both rows.
3. Insert two new rows, which will occupy the same slots as the previous two rows:
  - A_1: `{ x: padding({ a: 0, b: 0 }) }`.
  - B_1: `{ x: padding({ a: 0, b: 0 }) }`.
4. Compare the two new rows with `eq_row_in_page`. Expect `true`, as they are equal, but get `false`, since `A_0[1..4] != B_0[1..4]`.

gefjon (changes requested):
Possible correctness bug. Please convince me otherwise.

crates/table/src/eq.rs
gefjon: I'm a little surprised too. I don't actually care, though; I think these new numbers seem like a reasonable and productive trade-off.

gefjon (approved):

gefjon:
Please rebase or merge on top of 80dff96723a9c963af36920a7501cdbb2fdd9d56 (#1909 ); that should fix the internal tests.

PR #2026, opened 2024-11-28
Decide that BSATN `byte` that ain't 0 or 1 is not a valid `AV::Bool`

Centril:
# Description of Changes

Yes, this no longer accepts e.g., `2` as `true`.

# Testing

A proptest is added to ensure this new (correct) behavior does not regress.

### DIFF
--- a/crates/sats/src/bsatn.rs
+++ b/crates/sats/src/bsatn.rs
@@ -152,7 +152,7 @@ impl ToBsatn for ProductValue {
 
 #[cfg(test)]
 mod tests {
-    use super::to_vec;
+    use super::{to_vec, DecodeError};
     use crate::proptest::generate_typed_value;
     use crate::{meta_type::MetaType, AlgebraicType, AlgebraicValue};
     use proptest::prelude::*;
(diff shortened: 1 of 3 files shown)


crates/sats/src/buffer.rs
```diff
@@ -20,6 +20,8 @@ pub enum DecodeError {
     InvalidTag { tag: u8, sum_name: Option<String> },
     /// Expected data to be UTF-8 but it wasn't.
     InvalidUtf8,
+    /// Expected the byte to be 0 or 1 to be a valid bool.
+    InvalidBool(u8),
```
gefjon, line 24: I wonder if there's some better way we can encode the set of errors "value is not valid at expected type." I don't see much of a difference between `InvalidTag`, `InvalidUtf8` and `InvalidBool`.

This is not a request for any changes, just some musing.

crates/sats/src/bsatn.rs
```diff
@@ -179,5 +179,11 @@ mod tests {
             let val_decoded = AlgebraicValue::decode(&ty, &mut &bytes[..]).unwrap();
             prop_assert_eq!(val, val_decoded);
         }
+
+        #[test]
+        fn bsatn_non_zero_one_u8_aint_bool(val in 2u8..) {
+            let bytes = [val];
+            prop_assert!(AlgebraicValue::decode(&AlgebraicType::Bool, &mut &bytes[..]).is_err());
```
gefjon, line 186: Do you want to assert the specific error? Fine if not, just wondering.

gefjon (approved):

cloutiertyler (commented):
Just also saying that I approve of this. Good catch.

PR #2031, opened 2024-12-02
query rewriter

joshua-spacetime:
First batch of query rewrite rules optimizing index joins


crates/physical-plan/src/plan.rs
```diff
@@ -69,11 +161,15 @@ pub enum IndexOp {
...
-    pub input: Box<PhysicalPlan>,
+    pub lhs: Box<PhysicalPlan>,
     /// The rhs indexed table
-    pub table: Arc<TableSchema>,
-    /// The rhs index
-    pub index: IndexId,
+    pub rhs: Arc<TableSchema>,
+    /// The rhs field name
+    pub rhs_label: Symbol,
+    /// The index id
```
mamcx, line 169: I think it will be nice if all the index info is in their own `struct`:

```rust
struct Idx {
index_id: IndexId,
index_cols: ColList,
unique: bool,
}
```

... it will simplify later when we need to know for example if the `idx` comes from a `btree` or a `hash`.

mamcx (commented):
Looks good, just fix the test...

mamcx (approved):
LGTM. The comments will be in another pr as discussed...

PR #2033, opened 2024-12-03
Migrate C# client-side reducer enum to the new syntax

RReverser:
# Description of Changes

As per https://github.com/clockworklabs/SpacetimeDBPrivate/pull/1198.

This makes reducers use the same codegen structure as tables and other tuple types, but nests them inside a `Reducer` class as subclasses for pattern-matching, like we allow on tagged enums.

The reason we don't use a tagged enum here anymore is because we don't need BSATN sum type support for the Reducer enum, and it imposes unnecessary restrictions like max 256 elements.

While testing the changes, simplified the C# CI a bit by reusing the `tools~/write-nuget-config.sh` script from the C# SDK repo instead of duplicating its logic.

API breaking change for the client.

2 - a breaking change, but shouldn't affect too much client code.

# Testing

Waiting for CI to test it against client SDK, it's a bit quicker than manual.


.github/workflows/ci.yml
```diff
@@ -120,49 +120,26 @@ jobs:
         run: |
           sudo mkdir /stdb
           sudo chmod 777 /stdb
+
```
gefjon, line 123: Why did this file change? Is that related to this PR?

crates/bindings-csharp/Runtime/Internal/Autogen/RawRowLevelSecurityDefV9.cs
```diff
@@ -18,9 +18,7 @@ public partial class RawRowLevelSecurityDefV9
 		[DataMember(Name = "sql")]
 		public string Sql;
 
-		public RawRowLevelSecurityDefV9(
-			string Sql
-		)
+		public RawRowLevelSecurityDefV9(string Sql)
```
gefjon, line 21: Do these files get auto-formatted? If so, why did they change?

gefjon (approved):
I have no way to test this, but it appears reasonable. I'd like clarity on the CI changes before you merge, but otherwise this looks good to go.

.github/workflows/ci.yml
RReverser: I was trying to test this and wanted to regenerate client SDK as part of the CI. That backfired - cross-repo testing is still pretty damn hard - so the test will have to be as part of client SDK PR, so I reverted the `spacetime generate` changes but left this one as it's purely for clarity. I'm happy to split it out as it's not related anymore.

crates/bindings-csharp/Runtime/Internal/Autogen/RawRowLevelSecurityDefV9.cs
RReverser: No, I changed the C# codegen logic while unifying reducer, tables and type generation. They're not autoformatted.

.github/workflows/ci.yml
gefjon: I'm fine to keep it in as long as you add a note about it to the PR description.

RReverser: Done

bfops (approved):
Approving CLI changes as a codeowner. I didn't really review; the changes are in C#-specific code that probably should be in an un-owned file anyway.

PR #2036, opened 2024-12-03
Rust SDK: no more reducer args structs

gefjon:
# Description of Changes

Well, at least, the reducer arg structs are no longer user-facing. They're still generated and used internally, for ser/de.

This mostly involves a lot of uninteresting shuffling things around in the codegen. One notable piece is that we must now be more precise about what names are re-exported by the generated `mod.rs`. The (now-internal) arg structs are `pub(super)` where `super` is that `mod.rs`, but the previous system of emitting `pub use foo_reducer::*;` would expose it anyways, causing the exact same name conflict that we're trying to avoid here. As such, this commit includes a rework of `print_module_reexports` which avoids glob reexports for types and reducers. Tables still use glob re-exports,
since reexporting callback ids would otherwise be complicated.

Minor incompatible change to the Rust SDK: code which inspects `ReducerEvent`s sees one less layer of nesting. Some patterns may need to be rewritten, but this should be easy.

1: this code is well-tested, and the majority of this diff is autogenerated.

# Testing

- [x] Test suite!
- [ ] Private companion & its tests.


RReverser:
> They're still generated and used internally, for ser/de.

Is this because we don't support enum variants with fields in our macros?

crates/cli/src/subcommands/generate/rust.rs
```diff
@@ -859,27 +935,13 @@ impl __sdk::InModule for Reducer {{
...
-                "fn reducer_args(&self) -> &dyn std::any::Any {",
-                |out| {
-                    out.delimited_block(
-                        "match self {",
-                        |out| {
-                            for reducer in iter_reducers(module) {
-                                writeln!(out, "Reducer::{}(args) => args,", reducer_variant_name(&reducer.name));
+                                write!(out, "Reducer::{}", reducer_variant_name(&reducer.name));
+                                if !reducer.params_for_generate.elements.is_empty() {
+                                    // Because Rust prohibits empty struct variants,
```
RReverser, line 940: I don't think it does? Can you show an example?

crates/cli/src/subcommands/generate/rust.rs
```diff
@@ -828,13 +903,14 @@ fn print_reducer_enum_defn(module: &ModuleDef, out: &mut Indenter) {
...
-                    "{}({}::{}),",
-                    reducer_variant_name(&reducer.name),
-                    reducer_module_name(&reducer.name),
-                    reducer_args_type_name(&reducer.name),
-                );
+                write!(out, "{} ", reducer_variant_name(&reducer.name));
+                if !reducer.params_for_generate.elements.is_empty() {
+                    // If the reducer has any arguments, generate a "struct variant."
+                    // If it doesn't, generate a "unit variant" instead.
+                    // For some reason, Rust prohibits empty struct variants.
```
RReverser, line 910: Unless I'm misunderstanding something, it doesn't. Enums like

```rust
enum Foo {
    A {}
}
```

are fine.

gefjon: Strange - I tried this and got an error. I wonder if I misunderstood and it was actually from the SATS derive macros, before I removed them.

gefjon: Ah, yep, it was the SATS macros. Well, now that I've done the work to generate unit variants, I figure we might as well keep it - it's better UX. I'll just change the comments.

RReverser: Yeah I agree about better UX either way.

crates/cli/src/subcommands/generate/rust.rs
```diff
@@ -937,7 +941,7 @@ impl __sdk::InModule for Reducer {{
                             for reducer in iter_reducers(module) {
                                 write!(out, "Reducer::{}", reducer_variant_name(&reducer.name));
                                 if !reducer.params_for_generate.elements.is_empty() {
-                                    // Because Rust prohibits empty struct variants,
+                                    // Because we're emitting unit variants when the payload is empty,
```
RReverser, line 944: FWIW here you can always use `{{ .. }}` to simplify the codegen logic a bit (remove the condition) if you want. Rust allows matching unit variants as if they were empty struct variants, so both `Foo::Bar => ...` and `Foo::Bar {} => ...` work fine.

gefjon:
> > They're still generated and used internally, for ser/de.
> 
> Is this because we don't support enum variants with fields in our macros?

There's a few things going on here:

- We can't `derive(Serialize, Deserialize)` on `enum Reducer`, for a whole host of reasons.
- Even if we could, or if we implemented those traits manually, the WS format doesn't contain a BSATN-encoded `Reducer`; it has a dependent pair of `(reducer_name: String, args: (reducer_type reducer_name))`, where the client has access to a partial function `reducer_type : String -> Type`.
- This means that what we actually need is to implement `reducer_type`, and then a way to deserialize each different reducer args on its own. Hypothetically, one could work up construction with `DeserializeSeed` which does this straight into `Reducer` without the intermediate per-reducer struct, but that seems like a lot more work for no clear gain.

I will add comments to this effect.

PR #2037, opened 2024-12-04
move unique constraint checking until after optimistic insertion

Centril:
# Description of Changes

Move unique constraint checking until after the row has been inserted and roll back if there was a violation.
We need to do this to enable insertion via BSATN as we no longer will have a PV.
After insertion, we can instead project `RowRef`s to the index types.
This should also facilitate moving indices out of tables.

# Testing

Proptests are amended to reflect new order.

### DIFF
--- a/crates/core/src/db/datastore/locking_tx_datastore/mut_tx.rs
+++ b/crates/core/src/db/datastore/locking_tx_datastore/mut_tx.rs
@@ -1261,7 +1261,8 @@ impl MutTxId {
                     }
 
                     // Pacify the borrow checker.
-                    // SAFETY: `ptr` came from `tx_table.insert` just now without any interleaving calls.
+                    // SAFETY: `tx_row_ptr` came from `tx_table.insert` just now
+                    // without any interleaving `&mut` calls that could invalidate the pointer.
                     let tx_row_ref = unsafe { tx_table.get_row_ref_unchecked(tx_blob_store, tx_row_ptr) };
 
                     // (2) The `tx_row_ref` did not violate a unique constraint *within* the `tx_table`,
--- a/crates/table/src/btree_index.rs
+++ b/crates/table/src/btree_index.rs
@@ -202,7 +202,9 @@ impl TypedIndex {
     /// or may insert a nonsense value into the index.
     /// Note, however, that it will not invoke undefined behavior.
     ///
-    /// Assumes that this is a unique constraint and returns `Ok(Some(existing_row))` if it's violated.
+    /// Assumes that this is a unique constraint
+    /// and returns `Ok(Some(existing_row))` if it's violated.
+    /// The index is not inserted to in that case.
     fn insert_unique(&mut self, cols: &ColList, row_ref: RowRef<'_>) -> Result<Option<RowPointer>, InvalidFieldError> {
         fn insert_at_type<T: Ord + ReadColumn>(
             this: &mut Index<T>,
--- a/crates/table/src/btree_index/multimap.rs
+++ b/crates/table/src/btree_index/multimap.rs
@@ -39,7 +39,8 @@ impl<K: Ord, V: Ord> MultiMap<K, V> {
 
     /// Inserts the relation `key -> val` to this multimap.
     ///
-    /// Returns back the value if the `key` was already present in the map.
+    /// If `key` was already present in the map, does not add an association with `val`.
+    /// Returns the existing associated value instead.
     pub fn insert_unique(&mut self, key: K, val: V) -> Option<&V> {
         // TODO(perf, centril): don't use a multimap at all for unique indices.
         let vals = self.map.entry(key).or_default();


gefjon:
Can you bring in latest master to fix the "Internal Tests" job?

crates/core/src/db/datastore/locking_tx_datastore/mut_tx.rs
```diff
@@ -1196,35 +1196,27 @@ impl MutTxId {
...
-                // but may still be a set-semantic conflict with a row
-                // in the committed state.
-
-                let ptr = row_ref.pointer();
+            Ok((tx_row_hash, tx_row_ref)) => {
+                let tx_row_ptr = tx_row_ref.pointer();
                 if let Some(commit_table) = commit_table {
-                    // Safety:
+                    // The `tx_row_ref` was not previously present in insert tables,
+                    // but may still be a set-semantic conflict with a row
```
gefjon, line 1210: Comment should be expanded to say that it may also be a constraint violation. I'm on mobile RN and can't find the suggestion button, but would add "or may violate a constraint" after this sentence.

gefjon (commented):
AFK and not gonna do a full review RN; will follow up in the morning. Just don't want to forget this.

crates/core/src/db/datastore/locking_tx_datastore/mut_tx.rs
```diff
@@ -1250,26 +1244,42 @@ impl MutTxId {
...
                         let blob_store = &self.committed_state_write_lock.blob_store;
                         return Ok(RowRefInsertion::Existed(
                             // SAFETY: `find_same_row` told us that `ptr` refers to a valid row in `commit_table`.
-                            unsafe { commit_table.get_row_ref_unchecked(blob_store, committed_ptr) },
+                            unsafe { commit_table.get_row_ref_unchecked(blob_store, commit_ptr) },
                         ));
                     }
+
+                    // Pacify the borrow checker.
+                    // SAFETY: `ptr` came from `tx_table.insert` just now without any interleaving calls.
```
gefjon, line 1264: This comment is no longer accurate. `tx_row_ptr` (which is renamed) came from the `tx_row_ref`. There have been intervening calls - `Table::find_same_row` notably - and this comment should justify (or at least assert) that the pointer is not invalidated by those calls.

crates/table/src/btree_index.rs
```diff
@@ -186,13 +186,57 @@ impl TypedIndex {
...
+    /// Add the row referred to by `row_ref` to the index `self`,
+    /// which must be keyed at `cols`.
+    ///
+    /// If `cols` is inconsistent with `self`,
+    /// or the `row_ref` has a row type other than that used for `self`,
+    /// this will behave oddly; it may return an error,
+    /// or may insert a nonsense value into the index.
+    /// Note, however, that it will not invoke undefined behavior.
+    ///
+    /// Assumes that this is a unique constraint and returns `Ok(Some(existing_row))` if it's violated.
```
gefjon, line 205: Add a line saying whether the index is mutated in that case.

I don't love this return type: it seems like a unique constraint violation is error-ey. I would consider `Result<Result<(), RowPointer>, InvalidFieldError>` or defining `enum InsertUniqueError { InvalidField(InvalidFieldError), AlreadyPresent(RowPointer) }` and returning `Result<(), InsertUniqueError>`. But this is a small nit and it's fine with me if we merge with the type as-is. I do care about the comment, though.

crates/table/src/btree_index.rs
```diff
@@ -445,12 +494,13 @@ impl BTreeIndex {
...
 
     /// Extends [`BTreeIndex`] with `rows`.
-    /// Returns whether every element in `rows` was inserted.
     pub fn build_from_rows<'table>(
         &mut self,
         cols: &ColList,
         rows: impl IntoIterator<Item = RowRef<'table>>,
     ) -> Result<(), InvalidFieldError> {
+        // TODO(centril, correctness): consider `self.is_unique`.
+        //   Should not be able to add an index which would cause unique constraint violations.
```
gefjon, line 503: Do you just want to call `self.check_and_insert` here instead of `self.insert`?

crates/table/src/btree_index/multimap.rs
```diff
@@ -37,6 +37,20 @@ impl<K: Ord, V: Ord> MultiMap<K, V> {
         self.map.entry(key).or_default().push(val);
     }
 
+    /// Inserts the relation `key -> val` to this multimap.
+    ///
+    /// Returns back the value if the `key` was already present in the map.
```
gefjon, line 42:
```suggestion
    /// If `key` was already present in the map, does not add an association with `val`.
    /// Returns the existing associated value instead.
```

gefjon (approved):
Minor nits, mostly comment-ey. Code looks good, and is correct AFAICT.

Issue #2020, opened 2024-11-27
perf: pool chunks used by row iterators in `InstanceEnv`


Centril:
Used by e.g., table_scan and btree_scan

PR #2038, opened 2024-12-05
Allocation pool the chunked iters in `InstanceEnv`

Centril:
# Description of Changes

This adds a `ChunkPool` which `ChunkedWriter` uses.
Also removes some dead code in `InstanceEnv`.

~Based on flame graphs and timings, I'm not sure this changes much :/
We might want to try a different approach that doesn't do any chunking at all but rather writes to a singe buffer that we pool.
We can then keep an offset when iterating instead. Not sure how this will work with row boundaries though.~

I measured things again for `.iter()`, and in this case, it seems like the time spent in `datastore_table_scan_bsatn` decreased by a few %. Not a huge amount, but still an improvement. The index scan bit not improving makes sense -- there's a unique index there and only a single element is fetched.

Fixes https://github.com/clockworklabs/SpacetimeDB/issues/2020.

### DIFF
--- a/crates/core/src/host/instance_env.rs
+++ b/crates/core/src/host/instance_env.rs
@@ -3,6 +3,7 @@ use crate::database_logger::{BacktraceProvider, LogLevel, Record};
 use crate::db::datastore::locking_tx_datastore::MutTxId;
 use crate::error::{IndexError, NodesError};
 use crate::replica_context::ReplicaContext;
+use core::mem;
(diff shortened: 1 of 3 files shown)


coolreader18 (approved):
LGTM. Do you know if we have any tests that actually exercise the `chunk.len() > ROW_ITER_CHUNK_SIZE` codepath? That feels like it'd be good to have if not especially as this gets optimized more.

PR #2044, opened 2024-12-09
Avoid a multimap when the index is unique

Centril:
# Description of Changes

Fixes https://github.com/clockworklabs/SpacetimeDB/issues/2035.

Also fixes a bug in `create_index` where it doesn't error when a unique index is added where the existing rows would violate the constraint.

Also reduces the work done in `Table::clone_structure`.


coolreader18 (approved):

cloutiertyler:
What's the benchmark diff from this out of curiosity. I saw you updated the benches.

PR #2046, opened 2024-12-10
Enforce .NET 8 SDK for new C# modules

RReverser:
# Description of Changes

This fixes an issue where users with .NET 9 SDK installed can't build C# modules even if .NET 8 SDK is installed as well.

Unfortunately, this covers only new projects created with `spacetime init` as that's the only place where we control the project directory. For everyone else, we print instructions with a link to the official configuration docs.

corresponding GitHub label.

# Testing

Have both .NET 8 and .NET 9 installed globally (e.g. latest Visual Studio includes .NET 9 SDK by default, and you can install .NET 8 SDK manually from the official website).

Try to init a new Spacetime C# module and publish it. Before this PR, it will fail, with this PR it will work.

I have run the above testing manually.

### DIFF
--- a/crates/cli/src/subcommands/init.rs
+++ b/crates/cli/src/subcommands/init.rs
@@ -178,8 +178,7 @@ pub async fn exec_init_csharp(args: &ArgMatches) -> anyhow::Result<()> {
         (include_str!("project/csharp/StdbModule._csproj"), "StdbModule.csproj"),
         (include_str!("project/csharp/Lib._cs"), "Lib.cs"),
         (include_str!("project/csharp/_gitignore"), ".gitignore"),
-        // Reuse global.json from modules instead of a separate template to ensure that we stay in sync.
-        (include_str!("../../../../modules/global.json"), "global.json"),
+        (include_str!("project/csharp/global._json"), "global.json"),
     ];
 
     // Check all dependencies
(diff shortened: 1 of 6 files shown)


crates/cli/src/subcommands/init.rs
```diff
@@ -178,6 +178,8 @@ pub async fn exec_init_csharp(args: &ArgMatches) -> anyhow::Result<()> {
         (include_str!("project/csharp/StdbModule._csproj"), "StdbModule.csproj"),
         (include_str!("project/csharp/Lib._cs"), "Lib.cs"),
         (include_str!("project/csharp/_gitignore"), ".gitignore"),
+        // Reuse global.json from modules instead of a separate template to ensure that we stay in sync.
+        (include_str!("../../../../modules/global.json"), "global.json"),
```
cloutiertyler, line 182: This is going to cause an issue with publishing the module: https://github.com/rust-lang/cargo/issues/13309

Please revise.

cloutiertyler (changes requested):
Have you tested this? Could you please update the testing section of the PR template?

crates/cli/src/subcommands/init.rs
cloutiertyler: We discussed that, but we can't do that because the dry-run fails because we aren't actually pushing things and they depend on each other.

RReverser: Ok not a huge fan of copying, but given that this is going to change _very_ rarely (once a year at worst), I just copied the JSON file into the template directory like all other files.

RReverser:
> Have you tested this? Could you please update the testing section of the PR template?

Yeah, without this PR publishing C# modules stopped working altogether after Visual Studio update. Updated the testing section.

cloutiertyler:
@RReverser could you just confirm that you ran the tests in the testing section?

RReverser:
> @RReverser could you just confirm that you ran the tests in the testing section?

Yes. Without this fix I can't publish modules on my machine since I installed .NET 9 (at least without manual workarounds), but with this PR I can - at least for the new modules.

cloutiertyler (approved):
Given that you have run the tests specified in the testing section, I will approve this PR.

PR #2050, opened 2024-12-10
Update Rust module documentation

kazimuth:
# Description of Changes

While I'm rewriting the reference, I figured I'd do this as well.

0

# Testing

Added doctests


kazimuth:
There are still a number of doctest failures that need to be fixed, and some warnings in upstream crates of `bindings` as well, but I'm opening this up for textual review now.

crates/bindings/README.md
```diff
@@ -0,0 +1,517 @@
+# SpacetimeDB Rust Module Library
```
kazimuth, line 1: I recommend reviewing starting here

crates/bindings/README.md
```diff
@@ -0,0 +1,517 @@
...
+}
+```
+
+Every reducer runs inside a [database transaction](https://en.wikipedia.org/wiki/Database_transaction). <!-- TODO: specific transaction level guarantees. --> This means that reducers will not observe the effects of other reducers modifying the database while they run. Also, if a reducer fails, all of its changes to the database will automatically be rolled back. Reducers can fail by [panicking](::std::panic!) or by returning an `Err`.
+
+#### The `ReducerContext` Type
+
+Reducers have access to a special [`ReducerContext`] argument. This argument allows reading and writing the database attached to a module. It also provides some additional functionality, like generating random numbers and scheduling future operations.
+
+The most important field of [`ReducerContext`] is [`.db`](ReducerContext#structfield.db). This field provides a view of the module's database. The [`#[table]`](macro@crate::table) macro generates traits that add accessor methods to this field.
```
gefjon, line 438: "The most important field" seems like a silly thing to say, IMO. Does `ctx.sender` not matter? Should `ctx.identity()` just go screw itself?

```suggestion
[`ReducerContext`] provides access to the database tables via [the `.db` field](ReducerContext#structfield.db). The [`#[table]`](macro@crate::table) macro generates traits that add accessor methods to this field.
```

crates/bindings/README.md
```diff
@@ -0,0 +1,517 @@
...
+
+<!-- TODO: this seems to work sometimes, but not always... Sometimes the links to downstream trait implementations aren't generated for some reason. Maybe it only works in the same cargo workspace?
+
+To see all of the available methods on `ctx.db`, run `cargo doc` in your module's directory, and navigate to the `spacetimedb::Local` struct in the generated documentation. This will be at the path:
+- `[your_project_directory]/target/doc/spacetimedb/struct.Local.html` (non-Windows)
+- `[your_project_directory]\target\doc\spacetimedb\struct.Local.html` (Windows)
```
gefjon, line 444: I don't think you actually need to distinguish Windows vs Unix paths in 2025 Anno Domini. Just use `/` as a path separator.

crates/bindings/README.md
```diff
@@ -0,0 +1,517 @@
...
+
+#### Lifecycle Reducers
+
+A small group of reducers are called at set points in the module lifecycle. These are used to initialize
+the database and respond to client connections. See [Lifecycle Reducers](macro@crate::reducer#lifecycle-reducers).
+
+#### Scheduled Reducers
+
+Reducers can be scheduled to run repeatedly. This can be used to implement timers, game loops, and
+maintenance tasks. See [Scheduled Reducers](macro@crate::reducer#scheduled-reducers).
```
gefjon, line 465: Is scheduled-ness a feature of the reducer, or of the table? If the latter, we may want to document this as "scheduled tables" or something. I'm unsure on what's best.

crates/bindings/README.md
```diff
@@ -0,0 +1,517 @@
...
+- ✅ **Adding or removing `#[autoinc]` annotations.**
+- ✅ **Changing tables from private to public**.
+- ✅ **Adding reducers**.
+- ✅ **Removing `#[unique]`  annotations.**
+
+The following changes are allowed, but may break clients:
+
+- ⚠️ **Changing or removing reducers**. Clients that attempt to call the old version of a changed reducer will receive runtime errors.
+- ⚠️ **Changing tables from public to private**. Clients that are subscribed to a newly-private table will receive runtime errors.
+- ⚠️ **Removing `#[primary_key]` annotations**. Non-updated clients will still use the old `#[primary_key]` as a unique key in their local cache, which can result in non-deterministic behavior when updates are received.
```
gefjon, line 489: I don't think this breaks clients. It's certainly not non-deterministic. It would be a breaking change to downgrade `primary_key` to not-unique-at-all, but I think we only allow `primary_key` -> `unique`, which doesn't actually change client behavior at all.

crates/bindings/README.md
```diff
@@ -0,0 +1,517 @@
...
+- ⚠️ **Removing indexes**. This is only breaking in some situtations.
```
gefjon, line 490: Do we support removing unique indexes (which come with unique constraints)?

crates/bindings/README.md
```diff
@@ -0,0 +1,517 @@
...
+  The specific problem is subscription queries <!-- TODO: clientside link --> involving semijoins, such as:
+    ```text
+    SELECT *
+    FROM Employee
+    WHERE DeptName IN (
+    SELECT DeptName
+    FROM Dept
+    )
```
gefjon, line 498: I don't know if we actually support this construction of semijoins. (CC @joshua-spacetime ). We've been using the equivalent:
```sql
SELECT Employee.* FROM Employee JOIN Dept ON Employee.DeptName = Dept.DeptName
```
We also only support semijoins that "look like" PK/FK joins, i.e. where both columns are unique (* I don't actually know if we enforce this, or just assume it. @joshua-spacetime may know more), and where there's an additional `WHERE` filter on the "filter table" (`Dept` in this case) (* This should be changing in the future, but I believe this is the current status).

crates/bindings/README.md
```diff
@@ -0,0 +1,517 @@
...
+- ✅ **Removing `#[unique]`  annotations.**
+
+The following changes are allowed, but may break clients:
+
+- ⚠️ **Changing or removing reducers**. Clients that attempt to call the old version of a changed reducer will receive runtime errors.
+- ⚠️ **Changing tables from public to private**. Clients that are subscribed to a newly-private table will receive runtime errors.
+- ⚠️ **Removing `#[primary_key]` annotations**. Non-updated clients will still use the old `#[primary_key]` as a unique key in their local cache, which can result in non-deterministic behavior when updates are received.
+- ⚠️ **Removing indexes**. This is only breaking in some situtations.
+  The specific problem is subscription queries <!-- TODO: clientside link --> involving semijoins, such as:
+    ```text
```
gefjon, line 492:
```suggestion
    ```sql
```

crates/bindings/src/lib.rs
```diff
@@ -45,13 +42,610 @@ pub use timestamp::Timestamp;
...
+/// /*
+/// TODO: whoops, use filter, and popularity_and_username.
+///     // For every named `index`, the table has an extra method
+///     // for getting a corresponding `spacetimedb::BTreeIndex`.
+///     let by_id_and_username: spacetimedb::BTreeIndex<_, (u32, String), _> =
+///         users.id_and_username();
+///     let mut billy: User = by_id_and_username.find((&57, &"Billy".to_string()));
+///     billy.popularity += 5;
+///     by_id_and_username.update(billy);
+/// */
```
gefjon, line 132: ?

crates/bindings/src/lib.rs
```diff
@@ -45,13 +42,610 @@ pub use timestamp::Timestamp;
...
+/// Instead, a type implementing [`Table<Row = Self>`] is generated. This can be looked up in a [`ReducerContext`]
+/// using `ctx.db.{table_name}()`. This type represents a handle to a database table, and can be used to
+/// iterate and modify the table's elements. It is a view of the entire table -- the entire set of rows at the time of the reducer call.
+///
+/// # Example
+///
+/// ```ignore
+/// use spacetimedb::{table, ReducerContext};
+///
+/// #[table(name = users, public,
```
gefjon, line 98:
```suggestion
/// #[table(name = user, public,
```

Table names should be singular.

gefjon: The rest of this doc comment will need to be updated too, but I'm too lazy to add suggestions to all of the uses.

crates/bindings/src/lib.rs
```diff
@@ -45,13 +42,610 @@ pub use timestamp::Timestamp;
...
+///
+/// The `#[table(...)]` attribute accepts any number of the following arguments, separated by commas.
+///
+/// Multiple `table` annotations can be present on the same type. This will generate
+/// multiple tables of the same row type, but with different names.
+///
+/// ### `name`
+///
+/// Specify the name of the table in the database, if you want it to be different from
+/// the name of the struct. The name can be any valid Rust identifier.
```
gefjon, line 167:
```suggestion
/// Specify the name of the table in the database.
```

It's not optional.

crates/bindings/src/lib.rs
```diff
@@ -45,13 +42,610 @@ pub use timestamp::Timestamp;
...
+///
+/// The table name is used to get a handle to the table from a [`ReducerContext`].
+/// For a table *table*, use `ctx.db.{table}()` to do this.
+/// For example:
+/// ```ignore
+///  #[table(name = users)]
+///  pub struct User {
+///      #[autoinc]
```
gefjon, line 175:
```suggestion
///      #[auto_inc]
```

crates/bindings/src/lib.rs
```diff
@@ -45,13 +42,610 @@ pub use timestamp::Timestamp;
...
+/// system; `pub struct` or `pub(crate) struct` do not affect the table visibility, only
+/// the visibility of the items in your own source code.
+///
+/// ### `index(...)`
+///
+/// You can specify an index on one or more of the table's columns with the syntax:
+/// `index(name = my_index, btree(columns = [a, b, c]))`
+///
+/// You can also just put `#[index(btree)]` on the field itself if you only need
+/// a single-column attribute; see column attributes below.
```
gefjon, line 209:
```suggestion
/// a single-column index; see column attributes below.
```

crates/bindings/src/lib.rs
```diff
@@ -45,13 +42,610 @@ pub use timestamp::Timestamp;
...
+///
+/// Multiple indexes are permitted.
```
gefjon, line 211:
```suggestion
/// A table may declare any number of indexes.
```

crates/bindings/src/lib.rs
```diff
@@ -45,13 +42,610 @@ pub use timestamp::Timestamp;
...
+/// Creates an auto-increment constraint.
+///
+/// When a row is inserted with the annotated field set to `0` (zero),
+/// the sequence is incremented, and this value is used instead.
+///
+/// Can only be used on numeric types.
+///
+/// May be combined with indexes or unique constraints.
+///
+/// Note that using `#[autoinc]` on a field does not also imply `#[primary_key]` or `#[unique]`.
```
gefjon, line 250:
```suggestion
/// Note that using `#[auto_inc]` on a field does not also imply `#[primary_key]` or `#[unique]`.
```

crates/bindings/src/lib.rs
```diff
@@ -45,13 +42,610 @@ pub use timestamp::Timestamp;
...
+/// If those semantics are desired, those attributes should also be used.
+///
+/// <!-- TODO: What happens if a reducer tries to insert a row that has an already-existing unique
+///            auto-inc column? Like, if the user inserts a row ahead of the auto-inc, then
+///            the auto-inc catches up? -->
```
gefjon, line 255: Normal unique constraint violation.

crates/bindings/src/lib.rs
```diff
@@ -45,13 +42,610 @@ pub use timestamp::Timestamp;
...
+/// When there is a unique column constraint on the table, insertion can fail if a uniqueness constraint is violated.
+/// If we insert two rows which have the same value of a unique column, the second will fail.
+/// This will be via a panic with [`Table::insert`] or via a `Result::Err` with [`Table::try_insert`].
+///
+/// For example:
+/// ```rust
+/// type CountryCode = String;
+/// type NationalBird = String;
+///
+/// #[table(name = countries)]
```
gefjon, line 285: Again, table names should be singular.

crates/bindings/src/lib.rs
```diff
@@ -45,13 +42,610 @@ pub use timestamp::Timestamp;
...
+///
+/// ### `#[index(btree)]`
+///
+/// Creates a single-column index with the specified algorithm.
+///
+/// It is an error to specify this attribute together with `#[unique]`.
+/// Unique constraints implicitly create a unique index, which is accessed using the [`UniqueColumn`] struct instead of the
+/// [`BTreeIndex`] struct.
+///
+/// The created index has the same name as the column. <!-- TODO: this may change if we do the unify-index-names PR. -->
```
gefjon, line 332: > this may change if we do the unify-index-names PR.

It what? News to me. Can you link me the PR?

crates/bindings/src/lib.rs
```diff
@@ -45,13 +42,610 @@ pub use timestamp::Timestamp;
...
+/// ```
+/// to get a [`BTreeIndex`] from a [`ReducerContext`].
+///
+/// For example:
+///
+/// ```ignore
+/// ctx.db.cities().latitude()
+/// ```
+///    
+/// # Generated code
```
gefjon, line 346: I kind of hate this whole section, honestly. Not in the sense that it's poorly written, but in that I don't want to document specifics of the codegen in this way. I'd much rather document how users will interact with the generated code, rather than describing the code itself. The whole point of generating types with horrible ugly names like `country__TableHandle` is that users aren't supposed to name those types. And the whole point of using macros to generate code is that users don't need to know the implementation details of that generated code.

crates/bindings/src/lib.rs
```diff
@@ -45,13 +42,610 @@ pub use timestamp::Timestamp;
...
+/// The first argument of a reducer is always a [`&ReducerContext`]. This context object
+/// allows accessing the database and viewing information about the caller, among other things.
+///
+/// After this, a reducer can take any number of arguments.
+/// These arguments must implement the [`SpacetimeType`], [`Serialize`], and [`Deserialize`] traits.
+/// All of these traits can be derived at once by marking a type with `#[derive(SpacetimeType)]`.
+///
+/// Reducers may return either `()` or `Result<(), E>` where `E: Debug`.
+///
+/// ```rust,ignore
```
gefjon, line 413: Why is this `ignore`? In general, why are so many of the code blocks in your docs marked `ignore`? I fear that ignored code blocks are doomed to bitrot and become incorrect, which is why Rustdoc supports doctests.

crates/bindings/src/lib.rs
```diff
@@ -45,13 +42,610 @@ pub use timestamp::Timestamp;
...
+///     x: f32,
+///     y: f32,
+///     area: f32,
+/// ) -> Result<(), AddPlaceError> {
+///     // add a "place" to the database.
+/// }
+/// ```
+///
+/// Reducers may fail by returning a [`Result::Err`](std::result::Result) or by [panicking](std::panic!).
+/// Such a failure will be printed to the module logs and abort the active database transaction.
```
gefjon, line 452: I don't think we actually print `Err` returns to the module logs, only panic payloads. We may change that in the future, but I think we should be more vague here about how these messages actually get used as diagnostics. We should also mentioned that they get returned to the caller as a failure response.

Also, Tyler feels quite strongly that `panic` = bug, `Err` = expected failure, so you may want to mention that here.

crates/bindings/src/lib.rs
```diff
@@ -45,13 +42,610 @@ pub use timestamp::Timestamp;
...
+/// Any changes to the database will be rolled back.
```
gefjon, line 453:
```suggestion
/// Any changes to the database by failed reducers will be rolled back.
```

crates/bindings/src/lib.rs
```diff
@@ -45,13 +42,610 @@ pub use timestamp::Timestamp;
...
+/// and any time the database is cleared.
+///
+/// If an error occurs when initializing, the module will not be published.
+///
+/// ### `#[spacetimedb::reducer(client_connected)]`
+///
+/// This reducer is run when a client connects to the SpacetimeDB module.
+/// Their identity can be found in the sender value of the `ReducerContext`.
+///
+/// If an error occurs in the reducer, the client will be disconnected.
```
gefjon, line 492: I don't think we've implemented this behavior. CC @jsdt .

crates/bindings/src/lib.rs
```diff
@@ -45,13 +42,610 @@ pub use timestamp::Timestamp;
...
+///
+/// ### `#[spacetimedb::reducer(update)]`
+///
+/// This reducer is run when the module is updated,
+/// i.e., when publishing a module for a database that has already been initialized.
+///
+/// If an error occurs when updating, the module will not be published,
+/// and the previous version of the module attached to the database will continue executing.
+///
+/// # Scheduled reducers
```
gefjon, line 510: This section should link to the corresponding section of the table macro docs at some point.

crates/bindings/src/lib.rs
```diff
@@ -45,13 +42,610 @@ pub use timestamp::Timestamp;
...
+///
+/// In addition to life cycle annotations, reducers can be made **scheduled**.
+/// This allows calling the reducers at a particular time, or in a loop.
+/// This can be used for game loops.
+///
```
gefjon, line 515: Should mention that scheduled reducers can still be called by clients, and suggest gating on `ctx.identity() == ctx.sender` to prevent that.

crates/bindings/src/lib.rs
```diff
@@ -45,13 +42,610 @@ pub use timestamp::Timestamp;
...
+///     /// An identifier for the scheduled reducer call.
+///     #[primary_key]
+///     #[autoinc]
+///     scheduled_id: u64,
+///
+///     /// Information about when the reducer should be called.
+///     #[scheduled_at]
+///     scheduled_at: ScheduleAt,
+///
+///     // After the mandatory fields, any number of fields can be added.
```
gefjon, line 546:
```suggestion
///     // In addition to the mandatory fields, any number of fields can be added.
```

Order doesn't matter. Field names don't matter either, actually, only type and attributes, but we don't need to mention that.

crates/bindings/src/lib.rs
```diff
@@ -45,13 +42,610 @@ pub use timestamp::Timestamp;
...
+///    #[autoinc]
+///    id: u64,
+///    name: String,
+/// }
+/// ```
+///
+/// will generate accessors that allow looking up the `people` table in a `ReducerContext`.
+///
+pub use table::__MapRowTypeToTable;
+*/
```
gefjon, line 625: Dear god, please no...

crates/bindings/src/lib.rs
```diff
@@ -118,6 +745,19 @@ impl DbContext for ReducerContext {
     }
 }
 
+/// Allows accessing the local database attached to the module.
+///
+/// This slightly strange type appears to have no methods, but that is misleading.
+/// The `#[table]` macro uses the trait system to add table accessors to this type.
+/// These are generated methods that allow you to access specific tables.
+///
+/// <!-- TODO(THIS PR): THIS IS A LIE. In my testing I thought it worked, but it seems less reliable than I thought.
```
gefjon, line 754: :sob:

crates/bindings/src/table.rs
```diff
@@ -12,49 +12,45 @@ pub use spacetimedb_primitives::{ColId, IndexId};
...
 pub trait Table: TableInternal {
     /// The type of rows stored in this table.
     type Row: SpacetimeType + Serialize + DeserializeOwned + Sized + 'static;
 
-    /// Returns the number of rows of this table in the TX state,
-    /// i.e. num(committed_state) + num(insert_table) - num(delete_table).
-    ///
-    /// This API is new to modules (though it previously existed in the Rust SDK)
-    /// and will require a new host function in the ABI.
+    /// Returns the number of rows of this table as of this transaction.
```
gefjon, line 25:
```suggestion
    /// Returns the number of rows of this table.
    ///
    /// This takes into account modifications by the current transaction,
    /// even though those modifications have not yet been committed or broadcast to clients.
```

crates/bindings/src/table.rs
```diff
@@ -12,49 +12,45 @@ pub use spacetimedb_primitives::{ColId, IndexId};
...
 
-    /// Inserts `row` into the TX state,
-    /// i.e. removes it from the delete table or adds it to the insert table as appropriate.
+    /// Inserts `row` into the table.
     ///
     /// The return value is the inserted row, with any auto-incrementing columns replaced with computed values.
     /// The `insert` method always returns the inserted row,
     /// even when the table contains no auto-incrementing columns.
     ///
+    /// The returned row is a COPY of the row in the database.
```
gefjon, line 48:
```suggestion
    /// The returned row is a copy of the row in the database.
```

I don't feel like we need capital letters for this.

crates/bindings/src/table.rs
```diff
@@ -12,49 +12,45 @@ pub use spacetimedb_primitives::{ColId, IndexId};
...
-    /// Note that, in languages where error handling is based on exceptions,
-    /// no distinction is provided between `Table::insert` and `Table::try_insert`.
-    /// A single method `insert` is defined which throws an exception on failure,
-    /// and callers may either install handlers around it or allow the exception to bubble up.
-    ///
-    /// Note on MVCC: because callers have no way to determine if the row was previously present,
-    /// two concurrent transactions which delete the same row
-    /// may be ordered arbitrarily with respect to one another
-    /// while maintaining sequential consistency, assuming no other conflicts.
+    /// <!-- TODO: semantics of insert duplicate row with no unique constraints -->
```
gefjon, line 53: It's a no-op, not an error. Unless there's a unique constraint, in which case it is an error.

crates/bindings/src/table.rs
gefjon: May be better to just splice this into the previous paragraph: "The return value is a copy of the inserted row, [...]."

crates/bindings/src/table.rs
```diff
@@ -98,14 +89,8 @@ pub trait Table: TableInternal {
...
-    /// we should define `Self::try_delete` and `Self::TryDeleteError`,
-    /// analogous to [`Self::try_insert`] and [`Self::TryInsertError`].
-    ///
-    /// Note on MVCC: the return value means that logically a `delete` performs a query
-    /// to see if the row is present.
-    /// As such, two concurrent transactions which delete the same row
-    /// cannot be placed in a sequentially-consistent ordering,
-    /// and one of them must be retried.
+    /// we should define `Self::ForeignKeyViolation`,
+    /// analogous to [`Self::UniqueConstraintViolation`].
```
gefjon, line 93: Please remove this section; I wrote it as internal docs and did not intend for it to become public.

crates/bindings/src/table.rs
```diff
@@ -263,6 +248,41 @@ pub trait Column {
...
+///
+/// For a table *table* with a column *column*, use `ctx.db.{table}().{column}()`
+/// to get a `UniqueColumn` from a [`ReducerContext`](crate::ReducerContext).
+///
+/// Example:
+///
+/// ```no_run
+/// use spacetimedb::{table, UniqueColumn, ReducerContext, DbContext};
+///
+/// #[table(name = users)]
```
gefjon, line 262: singular table name

crates/bindings/src/table.rs
```diff
@@ -374,6 +394,51 @@ pub trait Index {
...
+/// ctx.db.{table}().{index}()
+/// ```
+/// for a table *table* and an index *index*.
+///
+/// Example:
+///
+/// ```no_run
+/// use spacetimedb::{table, BTreeIndex, ReducerContext, DbContext};
+///
+/// #[table(name = users,
```
gefjon, line 410: singular table name

crates/bindings/src/table.rs
```diff
@@ -374,6 +394,51 @@ pub trait Index {
...
+///     let by_dogs_and_name: BTreeIndex<_, (u64, String), _> = ctx.db.users().dogs_and_name();
+/// }
+/// ```
+///
+/// For single-column indices, use the name of the column:
+///
+/// ```no_run
+/// use spacetimedb::{table, BTreeIndex, ReducerContext, DbContext};
+///
+/// #[table(name = users)]
```
gefjon, line 429: singular table name

crates/bindings/src/table.rs
```diff
@@ -384,10 +449,75 @@ impl<Tbl: Table, IndexType, Idx: Index> BTreeIndex<Tbl, IndexType, Idx> {
...
+    /// This method accepts a variable numbers of arguments using the [`BTreeIndexBounds`] trait.
+    /// This depends on the type of the B-Tree index. `b` may be:
     /// - A value for the first indexed column.
     /// - A range of values for the first indexed column.
     /// - A tuple of values for any prefix of the indexed columns, optionally terminated by a range for the next.
+    ///
+    /// For example:
+    ///
+    /// ```no_run
+    /// #[table(name = users,
```
gefjon, line 461: I'm gonna stop leaving these comments. You should just grep for `name =` and fix them all.

crates/sats/src/de.rs
```diff
@@ -551,6 +551,9 @@ pub use spacetimedb_bindings_macro::Deserialize;
 ///
 /// The lifetime `'de` allows us to deserialize lifetime-generic types in a zero-copy fashion.
 ///
+/// Do not manually implement this trait unless you know what you are doing.
+/// Incorrect implementations are safe, but can result in data loss.
```
gefjon, line 555: I would prefer a more precise warning that implementations of this trait must be consistent with all of `Serialize for T`, `SpacetimeType for T` and `Serialize, Deserialize for AlgebraicValue`. That should be plenty to scare off anyone who doesn't know what they're doing.

crates/sats/src/ser.rs
```diff
@@ -203,13 +203,17 @@ pub use spacetimedb_bindings_macro::Serialize;
 
 use crate::AlgebraicType;
 
-/// A **data structure** that can be serialized into any data format supported by SATS.
+/// A **data structure** that can be serialized into any data format supported by
```
gefjon, line 206: Why is "data structure" emphasized here, and can it not be?

crates/sats/src/de.rs
```diff
@@ -542,7 +542,7 @@ pub trait DeserializeSeed<'de> {
 use crate::de::impls::BorrowedSliceVisitor;
 pub use spacetimedb_bindings_macro::Deserialize;
 
-/// A **datastructure** that can be deserialized from any data format supported by SATS.
+/// A **data structure** that can be deserialized from any data format supported by the SpacetimeDB Algebraic Type System.
```
gefjon, line 545: Emphasis is weird and ugly here.

crates/sats/src/ser.rs
```diff
@@ -203,13 +203,17 @@ pub use spacetimedb_bindings_macro::Serialize;
...
+/// the SpacetimeDB Algebraic Type System.
 ///
 /// In most cases, implementations of `Serialize` may be `#[derive(Serialize)]`d.
 ///
 /// The `Serialize` trait in SATS performs the same function as [`serde::Serialize`] in [`serde`].
 /// See the documentation of [`serde::Serialize`] for more information of the data model.
 ///
+/// Do not manually implement this trait unless you know what you are doing.
+/// Incorrect implementations are safe, but can result in data loss.
+///
```
gefjon, line 216: Same note about this warning.

crates/sats/src/typespace.rs
```diff
@@ -232,14 +232,65 @@ pub trait GroundSpacetimeType {
...
 
-/// A trait for Rust types that can be represented as an [`AlgebraicType`]
-/// provided a typing context `typespace`.
+/// This trait makes types self-describing, allowing them to automatically register their structure
+/// with SpacetimeDB. This is used to tell SpacetimeDB about the structure of a module's tables and
+/// reducers.
+///
+/// Deriving this trait also derives [`Serialize`](crate::ser::Serialize), [`Deserialize`](crate::de::Deserialize),
+/// and [`Debug`](std::fmt::Debug). (There are currently no trait bounds on `SpacetimeType` documenting this fact.)
+/// These traits are used to convert Rust data structures to other formats, suitable for storing on disk or passing over the network. `Debug` is simply for debugging convenience.
```
gefjon, line 241:
```suggestion
/// `Serialize` and `Deserialize` are used to convert Rust data structures to other formats, suitable for storing on disk or passing over the network. `Debug` is simply for debugging convenience.
```

crates/sats/src/typespace.rs
```diff
@@ -232,14 +232,65 @@ pub trait GroundSpacetimeType {
...
+///
+/// Any Rust type implementing `SpacetimeType` can be used in table and reducer declarations. A derive macro is provided, and can be used on both structs and enums:
```
gefjon, line 243: Can you be more specific? I believe we don't allow tuple structs. I'm not sure whether or not we allow unit structs. I think on enums we only allow unit variants and single-element tuple variants, not struct variants or n-tuple variants. I could be wrong though; you should check.

crates/sats/src/typespace.rs
```diff
@@ -232,14 +232,65 @@ pub trait GroundSpacetimeType {
...
+/// And common data structures:
+///
+/// - `String` and `&str`, utf-8 string data
+/// - `()`, the unit type
+/// - `Option<T> where T: SpacetimeType`
+/// - `Vec<T> where T: SpacetimeType`
+///
+/// (Storing collections in rows of a database table is a form of [denormalization](https://en.wikipedia.org/wiki/Denormalization).)
+///
+/// Do not manually implement this trait unless you are VERY sure you know what you're doing. Incorrect implementations can result in data loss!
```
gefjon, line 281: Same thing about the warning here.

gefjon (commented):
This is great stuff, thanks!

crates/bindings/src/lib.rs
jsdt: I didn't implement it, but from looking at the code, it looks like that is true. That's the first thing we do when trying to spawn a client connection [here](https://github.com/clockworklabs/SpacetimeDB/blob/a54ea3a391c2828c6fa49bb905312e0154be6b40/crates/core/src/client/client_connection.rs#L184), and we give up if it fails.

crates/bindings/README.md
joshua-spacetime: Phoebe is correct, we don't support `IN`, so
```sql
SELECT Employee.* FROM Employee JOIN Dept ON Employee.DeptName = Dept.DeptName
```
is the correct way to express it atm.

As of 1.0, a `WHERE` clause is no longer required.

Regarding `PK/FK` joins, previously the server did not enforce join columns having unique constraints, but the client could produce incorrect results in such cases. However, with the new 1.0 subscription api, the client does/will have handling for non-unique join columns.

crates/bindings/README.md
```diff
@@ -0,0 +1,517 @@
...
+
+From the perspective of clients, this process is mostly seamless. Open connections are maintained and subscriptions continue functioning. [Automatic migrations](#automatic-migrations) forbid most table changes except for adding new tables, so client code does not need to be recompiled.
+However:
+- Clients may witness a brief interruption in the execution of scheduled reducers (for example, game loops.)
+- New versions of a module may remove or change reducers that were previously present. Client code calling those reducers will receive runtime errors.
+
+
+## Tables
+
+Tables are declared using the [`#[table(name = table_name)]` macro](macro@crate::table).
```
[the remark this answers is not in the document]
kazimuth: It is a rustdoc link: https://rust-lang.github.io/rfcs/1946-intra-rustdoc-links.html#path-ambiguities

crates/bindings/README.md
```diff
@@ -0,0 +1,517 @@
...
+│  +─────────────────┐  │ call_reducer() │  ┌─────────────────┐  │
+│  │   Client Code   │─────────────────────>│   Module Code   │  │
+│  └─────────────────┘  │                │  └─────────────────┘  │
+│                       │                │                       │
+└───────────────────────┘                └───────────────────────┘
+```
+
+Rust modules are written with the the Rust Module Library (this crate). They are built using [cargo](https://doc.rust-lang.org/cargo/) and deployed using the [`spacetime` CLI tool](https://spacetimedb.com/install). Rust modules can import any Rust [crate](https://crates.io/) that supports being compiled to WebAssembly.
+
+(Note: Rust can also be used to write **clients** of SpacetimeDB databases, but this requires using a completely different library, the SpacetimeDB Rust Client SDK. See the documentation on [clients] for more information.)
```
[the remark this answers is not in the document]
kazimuth: There's an anchor at the bottom of the file.

crates/bindings/README.md
```diff
@@ -0,0 +1,517 @@
...
+- New versions of a module may remove or change reducers that were previously present. Client code calling those reducers will receive runtime errors.
+
+
+## Tables
+
+Tables are declared using the [`#[table(name = table_name)]` macro](macro@crate::table).
+
+This macro is applied to a Rust struct with named fields. All of the fields of the table must implement [`SpacetimeType`].
+
+The resulting type is used to store rows of the table. It is normal struct type. Row values are not special -- operations on row types do not, by themselves, modify the table. Instead, a [`ReducerContext`](#reducercontext) is needed to get a handle to the table.
```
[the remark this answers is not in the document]
kazimuth: I think it's important to clarify exactly what magic is happening.

crates/bindings/README.md
kazimuth: ok but

```
james@DESKTOP-1O4DK4H MINGW64 /f/clockwork/SpacetimeDB/modules (jgilles/rust-docs-update)
$ rg 'ctx.db' --stats | grep matches
159 matches
8 files contained matches

james@DESKTOP-1O4DK4H MINGW64 /f/clockwork/SpacetimeDB/modules (jgilles/rust-docs-update)
$ rg 'ctx.sender' --stats | grep matches
14 matches
4 files contained matches

james@DESKTOP-1O4DK4H MINGW64 /f/clockwork/SpacetimeDB/modules (jgilles/rust-docs-update)
$ rg 'ctx.identity' --stats | grep matches
2 matches
1 files contained matches
```

:3

crates/bindings/README.md
kazimuth: old habits

crates/bindings/README.md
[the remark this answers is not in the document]
kazimuth: Additionally, I'd propose https://github.com/clockworklabs/SpacetimeDB/issues/2261

crates/bindings/README.md
```diff
@@ -0,0 +1,517 @@
...
+Notice that updating a row is only possible if a row has a unique column -- there is no `update` method in the base [`Table`] trait. SpacetimeDB has no notion of rows having an "identity" aside from their unique / primary keys.
+
+The `#[primary_key]` annotation is similar to the `#[unique]` annotation, except that it leads to additional methods being made available in the [client]-side SDKs. <!-- TODO: better link. -->
+
+<!-- TODO: does this actually work? Are there still limitations on what we can mark unique?
+
+It is not currently possible to mark a group of fields as collectively unique. You can, however, derive `#[SpacetimeType]` on a sub-struct and mark that unique. For example:
+
+...
+-->
```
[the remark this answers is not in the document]
gefjon: Oh, it seems like we actually *don't* impose this limit on unique columns in Rust, only non-unique BTree indexes. That's... not ideal.

crates/bindings/README.md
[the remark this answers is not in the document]
gefjon: D'oh! Not sure how I missed that.

crates/bindings/README.md
kazimuth: Removing `#[unique]` is allowed, so I don't see any logic that would catch what you're saying. 

...And the `primary_key` field is ignored by `auto_migrate_table`, whoops! https://github.com/clockworklabs/SpacetimeDB/issues/2262

crates/bindings/README.md
kazimuth: Yes.

crates/bindings/src/lib.rs
kazimuth: On first insert, or when the auto inc catches up?

crates/bindings/src/lib.rs
kazimuth: My main goal was just to communicate that there is a separate handle type from the row type itself, with a completely different API. Naming the types makes it clear what is going on, even if it is ugly.

Also note that if users look at the rustdoc for the generated code, they will see the TableHandle types.

crates/bindings/src/lib.rs
kazimuth: Fixed

crates/bindings/src/lib.rs
kazimuth: Not convinced that's a meaningful distinction in modules but whatever

crates/sats/src/ser.rs
kazimuth: tyler

crates/sats/src/typespace.rs
kazimuth: Should have said "Any Rust type implementing `SpacetimeType` can be used as a table column or reducer argument."

crates/bindings/README.md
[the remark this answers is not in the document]
gefjon: Fixed in #2267 . I will update the docs here.

crates/bindings/README.md
gefjon: I'm deciding I don't really care. This is fine.

crates/bindings/src/lib.rs
gefjon: When the auto-inc catches up.

gefjon: Added a note about this in place of the TODO.

crates/bindings/src/lib.rs
gefjon: I'm adding a note encouraging users not to name the table handle explicitly.

crates/bindings/src/table.rs
gefjon: Adding this note.

gefjon (approved):

PR #2057, opened 2024-12-13
Use a spinner to show that wasm is compiling in spacetime generate

coolreader18:
# Description of Changes

Little QOL improvement.

1

# Testing

I ran `spacetime generate` to ensure that the spinner shows up and tells the user what it's doing.

### DIFF
--- a/crates/cli/src/subcommands/generate/mod.rs
+++ b/crates/cli/src/subcommands/generate/mod.rs
@@ -132,7 +132,12 @@ pub async fn exec(config: Config, args: &clap::ArgMatches) -> anyhow::Result<()>
         } else {
             build::exec_with_argstring(config.clone(), project_path, build_options).await?
         };
-        extract_descriptions(&wasm_path)?
+        let spinner = indicatif::ProgressBar::new_spinner();
+        spinner.enable_steady_tick(60);
+        spinner.set_message("Compiling wasm...");
+        let module = compile_wasm(&wasm_path)?;
+        spinner.set_message("Extracting schema from wasm...");
+        extract_descriptions_from_module(module)?
     };
 
     fs::create_dir_all(out_dir)?;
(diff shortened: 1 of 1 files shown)


cloutiertyler (dismissed):
Could you please add to the description what you did, if anything, to test this?

crates/cli/src/subcommands/generate/mod.rs
```diff
@@ -132,7 +132,12 @@ pub async fn exec(config: Config, args: &clap::ArgMatches) -> anyhow::Result<()>
         } else {
             build::exec_with_argstring(config.clone(), project_path, build_options).await?
         };
-        extract_descriptions(&wasm_path)?
+        let spinner = indicatif::ProgressBar::new_spinner();
+        spinner.enable_steady_tick(60);
+        spinner.set_message("Compiling wasm...");
+        let module = compile_wasm(&wasm_path)?;
+        spinner.set_message("Extracting schema from wasm...");
+        extract_descriptions_from_module(module)?
```
bfops, line 140: my main nit is that it seems like there are several places where we take a long time to compile something, and I'm confused why we're adding a spinner for this specific one. Like, it's an improvement, but it's also a bit confusing to not be doing this everywhere if we're doing it (and potentially factoring out some common helpers etc.)

cloutiertyler: Which other places did you have in mind?

bfops: I mean, actual `build` is the main one that comes to mind that I feel like people have mentioned before.

crates/cli/src/subcommands/generate/mod.rs
```diff
@@ -372,16 +377,22 @@ impl GenItem {
 }
 
 pub fn extract_descriptions(wasm_file: &Path) -> anyhow::Result<RawModuleDef> {
-    let engine = wasmtime::Engine::default();
-    let t = std::time::Instant::now();
-    let module = wasmtime::Module::from_file(&engine, wasm_file)?;
-    println!("compilation took {:?}", t.elapsed());
```
bfops, line 378: did we remove this summary message entirely, or does the new spinner do the same?

bfops: I guess either way users could just run things through `time` if they're trying to do a before-and-after comparison of something.

crates/cli/src/subcommands/generate/mod.rs
coolreader18: Build is quirky because cargo already shows a progress bar, and trying to show both would leave them conflicting.

crates/cli/src/subcommands/generate/mod.rs
coolreader18: Yeah, iirc that println was more a debugging thing that got left in too long.

bfops (approved):
Looks fine to me!

PR #2063, opened 2024-12-16
fix: filter pushdown over table scan

joshua-spacetime:
# Description of Changes

1. Optimizes `PhysicalProject::None`
2. Fixes filter pushdown rewrite rule so as not to match a `table_scan + filter`

1


mamcx (approved):
LGTM

PR #2065, opened 2024-12-16
Pretty print auto-migration plans

kazimuth:
# Description of Changes

Implements https://github.com/orgs/clockworklabs/projects/22/views/16?filterQuery=migration&pane=issue&itemId=91111218.

While implementing this, I discovered auto-migration plans applied their steps in a nondeterministic order (I was iterating a hashmap and had forgotten to sort.)

This interacted in a particularly bad way with row-level-security policies:
- Row level security policies are removed and re-added in every automigration, to allow the core crate to re-initialize the relevant queries. (I disagree with this -- it should be an implementation detail of the core crate, not mentioned in auto-migrations -- but I'm not changing it in this PR; see my comments about it in the code.)
- Since ordering was random, the Add step for the policy could be run before the Remove step.
- Adding an already existing policy was implemented in an idempotent way by the core crate.
- So, any auto-migration could nondeterministically drop row-level-security policies!

This is now fixed by simply sorting the steps of an auto-migration in a prescribed way.

This is a minor breaking change to the client API since it adds data to the result of an update call.

0

# Testing

Added a smoketest.
TODO: I want to add to the integration tests of the `schema` crate as well.

### DIFF
--- a/crates/cli/src/subcommands/publish.rs
+++ b/crates/cli/src/subcommands/publish.rs
@@ -179,6 +179,7 @@ pub async fn exec(config: Config, args: &ArgMatches) -> Result<(), anyhow::Error
             domain,
             database_identity,
             op,
+            update_summary,
         } => {
             let op = match op {
                 PublishOp::Created => "Created new",
@@ -188,6 +189,9 @@ pub async fn exec(config: Config, args: &ArgMatches) -> Result<(), anyhow::Error
                 println!("{} database with name: {}, identity: {}", op, domain, database_identity);
             } else {
                 println!("{} database with identity: {}", op, database_identity);
+            }
+            if let Some(update_summary) = update_summary {
+                println!("{}", update_summary);
             }
         }
         PublishResult::TldNotRegistered { domain } => {
--- a/crates/schema/src/auto_migrate.rs
+++ b/crates/schema/src/auto_migrate.rs
@@ -62,12 +62,14 @@ pub enum AutoMigrateStep<'def> {
     // It is important FOR CORRECTNESS that `Remove` variants are declared before `Add` variants in this enum!
     //
     // The ordering is used to sort the steps of an auto-migration.
-    // If adds go before removes, the following can occur:
+    // If adds go before removes, and the user tries to remove an index and then re-add it with new configuration,
+    // the following can occur:
     //
-    // 1. `AddIndex("my_special_boy)`
-    // 2. `RemoveIndex("my_special_boy)`
+    // 1. `AddIndex("indexname")`
+    // 2. `RemoveIndex("indexname")`
     //
-    // You see the problem.
+    // This results in the index being added -- which, at time of writing, does nothing -- and then removed,
+    // resulting in the intended index not being created.
     //
     // For now, we just ensure that we declare all `Remove` variants before `Add` variants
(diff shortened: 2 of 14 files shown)


crates/cli/src/subcommands/publish.rs
```diff
@@ -189,6 +190,9 @@ pub async fn exec(config: Config, args: &ArgMatches) -> Result<(), anyhow::Error
             } else {
                 println!("{} database with identity: {}", op, database_identity);
             }
+            if let Some(update_summary) = update_summary {
+                println!("{}", update_summary);
```
kazimuth, line 194: TODO: I guess I should check if the terminal supports ANSI color codes here, since the summary may include those.

bfops: Is this TODO meant for this PR, or should it be a comment/ticket?

bfops: I'm not sure that this is the best UX. It seems like these update summaries can be pretty large, and I don't think that most users are interested in reviewing them. Could we put this behind a CLI flag `--show-migration-summary` or something like that?

gefjon:
Please separate the bugfix from the new feature. These should be separate PRs.

kazimuth:
Sure, https://github.com/clockworklabs/SpacetimeDB/pull/2104

crates/cli/src/subcommands/publish.rs
kazimuth: Just a reminder to myself.

We could put these behind a flag, but then 99.9% of users will never see them. Our tools are already pretty chatty and I don't think this adds an overwhelming amount of information. It's good to tell the user when they're modifying their database.

It's particularly important to print an explanation when automatic migrations fail -- but, come to think of it, this PR hasn't implemented that, so I suppose I should do it as well, either here or in another PR.

bfops: Is that reminder meant to be implemented in this PR? (i.e. should I hold off on reviewing?)

kazimuth: For this PR yeah, I'll mark as draft until it's ready again.

bfops:
Converting to draft per https://github.com/clockworklabs/SpacetimeDB/pull/2065/files#r1910890326

crates/schema/src/auto_migrate.rs
```diff
@@ -52,45 +43,64 @@ pub struct AutoMigratePlan<'def> {
...
+    /// Remove an index.
+    RemoveIndex(<IndexDef as ModuleDefLookup>::Key<'def>),
+    /// Remove a constraint.
+    RemoveConstraint(<ConstraintDef as ModuleDefLookup>::Key<'def>),
+    /// Remove a sequence.
+    RemoveSequence(<SequenceDef as ModuleDefLookup>::Key<'def>),
+    /// Remove a schedule annotation from a table.
+    RemoveSchedule(<ScheduleDef as ModuleDefLookup>::Key<'def>),
+    /// Remove a row-level security query.
+    RemoveRowLevelSecurity(<RawRowLevelSecurityDefV9 as ModuleDefLookup>::Key<'def>),
```
kazimuth, line 88: Note: this change got pulled out of this PR.

PR #2067, opened 2024-12-17
Reduce noisy C# compilation errors

RReverser:
# Description of Changes

When source generator processes code that misuses SpacetimeDB APIs, has wrong signatures, uses types incompatible with BSATN etc, it reports diagnostics but still emits new source code. In some cases, this improves autocomplete and diagnostics, since types and methods are still there even while code is not 100% valid yet, but in other cases the output code is so broken that .NET attempting to compile it will only produce even more noisy errors, which drown out the custom actionable diagnostic we emitted earlier in the pipeline.

This PR fixes couple of such cases by either fixing up or skipping particularly problematic entries during the codegen phase.

You can see the difference in the `crates/bindings-csharp/Codegen.Tests/fixtures/diag/snapshots/ExtraCompilationErrors.verified.txt` snapshot, which captures all the .NET 
compilation errors that didn't come from our own source generator - there is a lot less noise after this PR.

corresponding GitHub label.

1

# Testing

`dotnet test` snapshot tests


crates/bindings-csharp/Codegen/Diag.cs
```diff
@@ -34,20 +34,24 @@ internal static class ErrorDescriptor
...
+        AttributeData attr
+    )> EmptyIndexColumns =
         new(
             group,
-            "Index attribute must specify index Columns.",
-            table => $"Table {table.Identifier} has an Index.BTree attribute, but no columns.",
-            table => table.BaseList!
+            "Index attribute must specify columns.",
+            ctx =>
+                $"{(ctx.index.AccessorName != "" ? ctx.index.AccessorName : "Index")} has an Index.BTree attribute, but no columns.",
```
kazimuth, line 45: In database dependency theory, it actually does make sense to key things based on the empty set of columns. It's equivalent to having an index on a column of type `()`. However, it's pretty silly to make such an index, so I'm fine with this error.

We might want to consider whether to allow empty unique constraints. This is semantically meaningful: it means the table can have at most one row. This might be useful if a client wants to have "global state" tables. I opened https://github.com/clockworklabs/SpacetimeDB/issues/2080 to discuss this.

kazimuth (approved):
This looks good, I like `Unsupported<T>`. My only concern is that this change might allow erroring modules to compile in unusual situations. However, this PR does a good job of having comments everywhere errors are suppressed, documenting that some other error has already been emitted. So I'm happy with this.

RReverser:
> My only concern is that this change might allow erroring modules to compile in unusual situations. However, this PR does a good job of having comments everywhere errors are suppressed, documenting that some other error has already been emitted.

Hm yeah that's a good point. FWIW the other mitigating factor is that in all those scenarios, worst that can happen is a runtime error instead of a compile one. But yeah, something to look out for if it slips through the tests.

Issue #2017, opened 2024-11-27
perf: Use `StaticLayout` to insert BSATN -> BFLATN


Centril:
Could be faster than going through the tree of `ProductType`.
One issue here though is that we need to validate the shape of the BSATN at some point against the schema.
The tree-walking aspect does that implicitly, but if we use the static layout, that might not work.

PR #2069, opened 2024-12-18
datastore: insert via BSATN instead of via PV

Centril:
# Description of Changes

This PR changes the structure of the datastore to insert via BSATN rather than a PV. But this isn't entirely true, we just use the fast-path for BSATN -> BFLATN but in the general case we still do BSATN -> PV -> BFLATN. Later, I'll fully remove the temporary step. One of the other required changes here is that we write and generate sequence values directly in BFLATN.

Fixes https://github.com/clockworklabs/SpacetimeDB/issues/2017.

Perf numbers on master

```
2025-01-03T09:30:37.776998Z  INFO: : Timing span "update_positions_by_id": 1.969988054s
2025-01-03T09:30:43.829835Z  INFO: : Timing span "update_positions_by_collect": 1.518314839s
```

with this PR:

```
2025-01-03T09:35:54.415502Z  INFO: : Timing span "update_positions_by_id": 1.64947372s
2025-01-03T09:36:00.880068Z  INFO: : Timing span "update_positions_by_collect": 1.26073871s
```

That is, this removes ~0.4516s and 0.3887s from the benchmarks respectively.

Flamegraph of `InstanceEnv::insert`:
Screenshot 2025-01-03 at 10-53-47 Flame Graph

Full flamegraph: https://flamegraph.com/share/14b571da-c9ba-11ef-9832-26c3e5347170

The next step after this PR is to avoid the pointer map in case there is a unique index.
After that, the next step is to add an `update` ABI.

None

4 -- limited scope, but lots of unsafe code and complicated logic.

# Testing

The internal `spacetimedb_table` tests still use the old `table.insert`, so the new path. To compensate for this, without duplicating tests, a new proptest `insert_bsatn_same_as_pv` is added asserting that the result and side-effects of inserting via PV and BSATN are the same. Moreover, both insert paths try to share as much code as possible to improve test coverage. The higher level tests, starting with `MutTxId` now use the new path. Over time, we can replace the remaining old paths with the new and then move all tests to the new as well.

# Reviewer notes

I would recommend reviewing in this order:
1. Start with `spacetimedb_table`
2. Move on to `locking_tx_datastore`
3. Move on to `InstanceEnv`
4. Review all the other boring changes to files like `estimation.rs` that were unfortunately necessary.

# Review notes for Tyler

In `traits.rs` the following changes:
- `insert_mut_tx` changes from:
   ```rust
   fn insert_mut_tx<'a>(..., row: ProductValue) -> Result<(AlgebraicValue, RowRef<'a>)>;
   ```
   to:
   ```rust
   fn insert_mut_tx<'a>(..., row: &[u8]) -> Result<(ColList, RowRef<'a>)>;
   ```

### DIFF
--- a/crates/core/src/db/datastore/locking_tx_datastore/mut_tx.rs
+++ b/crates/core/src/db/datastore/locking_tx_datastore/mut_tx.rs
@@ -1178,6 +1178,11 @@ impl MutTxId {
 
     /// Insert a row, encoded in BSATN, into a table.
     ///
+    /// Zero placeholders, i.e., sequence triggers,
+    /// in auto-inc columns in the new row will be replaced with generated values
+    /// if and only if `GENERATE` is true.
+    /// This method is called with `GENERATE` false when updating the `st_sequence` system table.
+    ///
     /// Requires:
     /// - `TableId` must refer to a valid table for the database at `database_address`.
     /// - `row` must be a valid row for the table at `table_id`.
--- a/crates/core/src/db/datastore/traits.rs
+++ b/crates/core/src/db/datastore/traits.rs
@@ -484,7 +484,7 @@ pub trait MutTxDatastore: TxDatastore + MutTx {
     ) -> u32;
     /// Inserts `row`, encoded in BSATN, into the table identified by `table_id`.
     ///
-    /// Returns the list of columns where values were replaced with generated ones
+    /// Returns the list of columns with sequence-trigger values that were replaced with generated ones
     /// and a reference to the row as a [`RowRef`].
     ///
     /// Generated columns are columns with an auto-inc sequence
--- a/crates/table/src/table.rs
+++ b/crates/table/src/table.rs
@@ -97,6 +97,9 @@ pub(crate) struct TableInner {
     row_layout: RowTypeLayout,
     /// A [`StaticLayout`] for fast BFLATN <-> BSATN conversion,
     /// if the [`RowTypeLayout`] has a static BSATN length and layout.
+    ///
+    /// A [`StaticBsatnValidator`] is also included.
+    /// It's used to validate BSATN-encoded rows before converting to BFLATN.
     static_layout: Option<(StaticLayout, StaticBsatnValidator)>,
     /// The visitor program for `row_layout`.
     ///
@@ -321,6 +324,9 @@ impl Table {
     ///
     /// This is also useful when we need to insert a row temporarily to get back a `RowPointer`.
     /// In this case, A call to this method should be followed by a call to [`delete_internal_skip_pointer_map`].
+    ///
+    /// When `row` is not valid BSATN at the table's row type,
+    /// an error is returned and there will be nothing for the caller to revert.
     pub fn insert_physically_bsatn<'a>(
         &'a mut self,
         blob_store: &'a mut dyn BlobStore,
@@ -399,7 +405,7 @@ impl Table {
                 // as `row_ty` was derived from the same schema as `seq` is part of.
                 let elem_ty = unsafe { &row_ty.elements.get_unchecked(seq.col_pos.idx()) };
                 // SAFETY:
-                // - `elem_ty` appears as a column in th row type.
+                // - `elem_ty` appears as a column in the row type.
                 // - `AlgebraicValue` is compatible with all types.
                 let val = unsafe { AlgebraicValue::unchecked_read_column(row_ref, elem_ty) };
                 val.is_numeric_zero()
@@ -415,9 +421,10 @@ impl Table {
     /// # Safety
     ///
     /// - `self.is_row_present(row)` must hold.
-    /// - `col_id` must be a valid column, with a primiive type, of the row type.
+    /// - `col_id` must be a valid column, with a primitive integer type, of the row type.
     pub unsafe fn write_gen_val_to_col(&mut self, col_id: ColId, ptr: RowPointer, seq_val: i128) {
         let row_ty = self.inner.row_layout.product();
+        // SAFETY: Caller promised that `col_id` was a valid column.
         let elem_ty = unsafe { row_ty.elements.get_unchecked(col_id.idx()) };
         let AlgebraicTypeLayout::Primitive(col_typ) = elem_ty.ty else {
             // SAFETY: Columns with sequences must be primitive types.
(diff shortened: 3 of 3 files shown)


crates/table/src/table.rs
```diff
@@ -234,20 +250,242 @@ impl Table {
...
+    /// Physically insert a `row`, encoded in BSATN, into this table,
+    /// storing its large var-len members in the `blob_store`.
+    ///
+    /// On success, returns the hash of the newly-inserted row,
+    /// and a `RowRef` referring to the row.
+    ///
+    /// This does not check for set semantic or unique constraints.
+    ///
+    /// This is also useful when we need to insert a row temporarily to get back a `RowPointer`.
+    /// In this case, A call to this method should be followed by a call to [`delete_internal_skip_pointer_map`].
```
gefjon, line 323: Can you expand this comment with a short paragraph about what will happen if this is called with a `row` that is not valid BSATN at the table's row type? Doesn't have to be super in-depth, but just to set our expectations between "Precisely detects all type errors" and "corrupts the table and burns down your house."

crates/table/src/table.rs
```diff
@@ -234,20 +250,242 @@ impl Table {
...
+
+        sequences
+            .iter()
+            // Find all the sequences that are triggered by this row.
+            .filter(|seq| {
+                // SAFETY: `seq.col_pos` is in-bounds of `row_ty.elements`
+                // as `row_ty` was derived from the same schema as `seq` is part of.
+                let elem_ty = unsafe { &row_ty.elements.get_unchecked(seq.col_pos.idx()) };
+                // SAFETY:
+                // - `elem_ty` appears as a column in th row type.
```
gefjon, line 402:
```suggestion
                // - `elem_ty` appears as a column in the row type.
```

crates/table/src/table.rs
```diff
@@ -234,20 +250,242 @@ impl Table {
...
+    }
+
+    /// Writes `seq_val` to the column at `col_id` in the row identified by `ptr`.
+    ///
+    /// Truncates the `seq_val` to fit the type of the column.
+    ///
+    /// # Safety
+    ///
+    /// - `self.is_row_present(row)` must hold.
+    /// - `col_id` must be a valid column, with a primiive type, of the row type.
```
gefjon, line 418:
```suggestion
    /// - `col_id` must be a valid column, with a primitive integer type, of the row type.
```

crates/table/src/table.rs
```diff
@@ -234,20 +250,242 @@ impl Table {
...
+            PrimitiveType::U32 => write(fixed_buf, elem_ty.offset, (seq_val as u32).to_le_bytes()),
+            PrimitiveType::I64 => write(fixed_buf, elem_ty.offset, (seq_val as i64).to_le_bytes()),
+            PrimitiveType::U64 => write(fixed_buf, elem_ty.offset, (seq_val as u64).to_le_bytes()),
+            PrimitiveType::I128 => write(fixed_buf, elem_ty.offset, seq_val.to_le_bytes()),
+            PrimitiveType::U128 => write(fixed_buf, elem_ty.offset, (seq_val as u128).to_le_bytes()),
+            PrimitiveType::I256 => write(fixed_buf, elem_ty.offset, (i256::from(seq_val)).to_le_bytes()),
+            PrimitiveType::U256 => write(fixed_buf, elem_ty.offset, (u256::from(seq_val as u128)).to_le_bytes()),
+            PrimitiveType::Bool | PrimitiveType::F32 | PrimitiveType::F64 => {
+                panic!("`{:?}` is not a sequence integer type", &elem_ty.ty)
+            }
```
gefjon, line 450: It feels mildly odd that we panic here, but `unreachable_unchecked` above. I don't feel strongly, but I might replace this with `unreachable_unchecked` just for consistency's sake.

crates/table/src/table.rs
gefjon: Spelling on "primitive." Being an integer is not a safety requirement as written, but I think it could be (see other comment). Whether or not we actually invoke UB if the column is non-integer, I think it's simplest in terms of docs to just say that we don't define behavior in that case, i.e. to make it a safety requirement as I'm doing here.

crates/table/src/table.rs
```diff
@@ -234,20 +250,242 @@ impl Table {
...
+        }
+    }
+
+    /// Performs all the checks necessary after having fully decided on a rows contents.
+    ///
```
gefjon, line 455:
```suggestion
    ///
    /// This includes inserting the row into any applicable indices and/or the pointer map.
    ///
```

crates/core/src/db/datastore/locking_tx_datastore/mut_tx.rs
```diff
@@ -1171,65 +1161,111 @@ impl<'a> Iterator for IndexScanFilterDeleted<'a> {
...
+            self.insert::<true>(table_id, buf)
+        })
+    }
+
+    /// Insert a row, encoded in BSATN, into a table.
     ///
     /// Requires:
     /// - `TableId` must refer to a valid table for the database at `database_address`.
     /// - `row` must be a valid row for the table at `table_id`.
     ///
```
gefjon, line 1184:
```suggestion
    ///
    /// Zero placeholders in auto-inc columns in the new row will be replaced with generated values
    /// if and only if `GENERATE` is true.
    /// This method is called with `GENERATE` false when updating the `st_sequence` system table.
    ///
```

crates/core/src/db/datastore/locking_tx_datastore/mod.rs
```diff
@@ -1,4 +1,4 @@
-#![forbid(unsafe_op_in_unsafe_fn)]
+#![deny(unsafe_op_in_unsafe_fn)]
```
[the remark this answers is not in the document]
gefjon: TIL that `deny` and `forbid` are different lint levels. I have no clue which is worse.

gefjon (approved):
I have only reviewed the `table` and `locking_tx_datastore` changes with any particular care; everything else I glanced at, but didn't worry too much about, on the grounds that it's all safe code, it's barely changing and it's tested. The `table` and `datastore` parts I looked at quite carefully. I agree with your safety reasoning, and your benchmark results speak for themselves. I've left a few minor comment-related requests, like I always do. Great work on this! And thanks again for writing such a detailed PR description; it made review much smoother.

crates/core/src/db/datastore/traits.rs
```diff
@@ -482,23 +482,19 @@ pub trait MutTxDatastore: TxDatastore + MutTx {
         table_id: TableId,
         relation: impl IntoIterator<Item = ProductValue>,
     ) -> u32;
-    /// Inserts `row` into the table identified by `table_id`.
+    /// Inserts `row`, encoded in BSATN, into the table identified by `table_id`.
     ///
-    /// Returns the generated column values (the [`AlgebraicValue`])
+    /// Returns the list of columns where values were replaced with generated ones
```
cloutiertyler, line 487:
```suggestion
    /// Returns the list of columns where values that were replaced with generated ones
```

cloutiertyler (approved):
Approved pending the suggested change. I have only reviewed the files I'm the owner for.

crates/core/src/db/datastore/traits.rs
gefjon: This doesn't scan. "[...] where values that were [...]" is incorrect.

PR #2070, opened 2024-12-18
fix: qualified projection columns

joshua-spacetime:
# Description of Changes

The parser qualifies all names. Any names that remain unqualified is a parsing error.

n/a

1

# Testing

Added applicable unit test to the parser


mamcx (approved):
LGTM

PR #2071, opened 2024-12-18
Fix install for callgrind runner docker image

kazimuth:
# Description of Changes

It seems the cargo solver has chosen a crate that doesn't compile with the rustc version in our docker image. This image is only used for benchmarks, so rather than trying to change our iai-callgrind-runner fork, I'm just updating the rustc version in the image.

None

0

# Testing

Going to run the updated actions yml on this branch to see if it works


kazimuth:
Hmm, the runner is now working. That error is actually an error in the benchmarks comparison application. I'll go poke at that.

RReverser:
Maybe now it's down to that regex update that you mentioned will be necessary after C# benchmarks are merged?

cloutiertyler (approved):

cloutiertyler:
Closing as this has been done in another commit.

PR #2073, opened 2024-12-19
commitlog: Fix offset index truncation

kim:
Fix the `ftruncate` implementation for the offset index writer to use the correct argument (transaction not byte offset). Add a test for it.

Also change the commitlog `reset_to` method to truncate to `offset + 1`, so as to retain the `offset` in the index if it is there. This matches the semantics of `reset_to`, which keeps the given offset.

1

# Testing

This patch adds a test.


Shubham8287 (approved):
Thanks for test

PR #2074, opened 2024-12-19
query engine integration

joshua-spacetime:
# Description of Changes

This patch integrates the new query engine into the query code path.

> What is implemented?

1. Instantiating tuple-at-a-time iterators from a physical plan
2. The addition of a new `query` crate which is the top level crate for invoking the query engine.
3. Invoking the new engine for initial subscriptions, one off queries, and incremental update
4. Benchmarks

n/a

4

The query engine is responsible for implementing the subscription logic, which is **the** only way to read data out of SpacetimeDB. As such, its correctness and performance is paramount.

# Testing

> Does the new engine preserve correctness? Does it also preserve performance?

We have already added tests to ensure we preserve semantics, with more being worked on currently. This patch updates the applicable benchmarks to ensure we maintain performance characteristics.

```
full-scan               time:   [10.988 ms 11.039 ms 11.098 ms]
                        change: [-24.628% -24.131% -23.622%] (p = 0.00 < 0.05)
                        Performance has improved.

full-join               time:   [92.660 µs 92.817 µs 92.976 µs]
                        change: [-34.859% -34.730% -34.595%] (p = 0.00 < 0.05)
                        Performance has improved.

query-indexes-multi     time:   [361.15 ns 361.59 ns 362.10 ns]
                        change: [-30.401% -30.266% -30.125%] (p = 0.00 < 0.05)
                        Performance has improved.
```

### DIFF
--- a/crates/core/src/host/module_host.rs
+++ b/crates/core/src/host/module_host.rs
@@ -7,14 +7,17 @@ use crate::db::datastore::system_tables::{StClientFields, StClientRow, ST_CLIENT
 use crate::db::datastore::traits::{IsolationLevel, Program, TxData};
 use crate::energy::EnergyQuanta;
 use crate::error::DBError;
+use crate::estimation::estimate_rows_scanned;
 use crate::execution_context::{ExecutionContext, ReducerContext, Workload};
 use crate::hash::Hash;
 use crate::identity::Identity;
 use crate::messages::control_db::Database;
 use crate::replica_context::ReplicaContext;
-use crate::sql;
+use crate::sql::ast::SchemaViewer;
 use crate::subscription::module_subscription_actor::ModuleSubscriptions;
+use crate::subscription::tx::DeltaTx;
 use crate::util::lending_pool::{Closed, LendingPool, LentResource, PoolClosed};
+use crate::vm::check_row_limit;
 use crate::worker_metrics::WORKER_METRICS;
 use anyhow::Context;
 use bytes::Bytes;
@@ -24,17 +27,18 @@ use indexmap::IndexSet;
 use itertools::Itertools;
 use smallvec::SmallVec;
 use spacetimedb_client_api_messages::timestamp::Timestamp;
-use spacetimedb_client_api_messages::websocket::{Compression, QueryUpdate, WebsocketFormat};
+use spacetimedb_client_api_messages::websocket::{Compression, OneOffTable, QueryUpdate, WebsocketFormat};
 use spacetimedb_data_structures::error_stream::ErrorStream;
 use spacetimedb_data_structures::map::{HashCollectionExt as _, IntMap};
 use spacetimedb_lib::identity::{AuthCtx, RequestId};
 use spacetimedb_lib::Address;
 use spacetimedb_primitives::{col_list, TableId};
+use spacetimedb_query::SubscribePlan;
 use spacetimedb_sats::{algebraic_value, ProductValue};
 use spacetimedb_schema::auto_migrate::AutoMigrateError;
 use spacetimedb_schema::def::deserialize::ReducerArgsDeserializeSeed;
 use spacetimedb_schema::def::{ModuleDef, ReducerDef};
-use spacetimedb_vm::relation::{MemTable, RelValue};
+use spacetimedb_vm::relation::RelValue;
 use std::fmt;
 use std::sync::{Arc, Weak};
 use std::time::{Duration, Instant};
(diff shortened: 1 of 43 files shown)


crates/table/src/btree_index.rs
```diff
@@ -33,10 +33,10 @@ use spacetimedb_sats::{
...
 mod uniquemap;
 
-type Index<K> = multimap::MultiMap<K, RowPointer>;
-type IndexIter<'a, K> = multimap::MultiMapRangeIter<'a, K, RowPointer>;
-type UniqueIndex<K> = uniquemap::UniqueMap<K, RowPointer>;
-type UniqueIndexIter<'a, K> = uniquemap::UniqueMapRangeIter<'a, K, RowPointer>;
+pub type Index<K> = multimap::MultiMap<K, RowPointer>;
+pub type IndexIter<'a, K> = multimap::MultiMapRangeIter<'a, K, RowPointer>;
+pub type UniqueIndex<K> = uniquemap::UniqueMap<K, RowPointer>;
+pub type UniqueIndexIter<'a, K> = uniquemap::UniqueMapRangeIter<'a, K, RowPointer>;
```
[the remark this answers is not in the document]
joshua-spacetime: Yeah not needed. At least for right now.

crates/core/src/host/module_host.rs
```diff
@@ -836,16 +837,24 @@ impl ModuleHost {
     }
```
mamcx, line 838: This is now dead code? Why still here?

crates/core/src/sql/ast.rs
```diff
@@ -474,32 +475,34 @@ fn compile_where(table: &From, filter: Option<SqlExpr>) -> Result<Option<Selecti
...
-            // It will be surfaced as a name resolution error instead.
-            .unwrap_or_else(|_| None)?;
-        if schema.table_access == StAccess::Private && self.auth.caller != self.auth.owner {
-            return None;
-        }
-        Some(schema)
+impl<T: StateView> SchemaView for SchemaViewer<'_, T> {
+    fn table_id(&self, name: &str) -> Option<TableId> {
+        let AuthCtx { owner, caller } = self.auth;
+        self.tx
```
mamcx, line 485:
```suggestion
// Get the schema from the in-memory state instead of fetching from the database for speed
```

crates/core/src/host/module_host.rs
```diff
@@ -836,16 +837,24 @@ impl ModuleHost {
...
 
     #[tracing::instrument(skip_all)]
-    pub fn one_off_query(&self, caller_identity: Identity, query: String) -> Result<Vec<MemTable>, anyhow::Error> {
+    pub fn one_off_query<F: WebsocketFormat>(
+        &self,
```
mamcx, line 841: This is now dead code? Why still here?

crates/physical-plan/src/plan.rs
```diff
@@ -618,7 +743,7 @@ pub struct IxJoin {
     /// The expression for computing probe values.
     /// Values are projected from the lhs,
     /// and used to probe the index on the rhs.
-    pub lhs_probe_expr: ProjectField,
+    pub lhs_field: TupleField,
```
mamcx, line 746: Can the `rhs_label` turn into `TupleField` instead/

crates/physical-plan/src/plan.rs
```diff
@@ -575,25 +711,14 @@ pub enum Sarg {
     Range(ColId, Bound<AlgebraicValue>, Bound<AlgebraicValue>),
```
mamcx, line 711: In 

```rust
pub enum Sarg {
     Range(ColId, Bound<AlgebraicValue>, Bound<AlgebraicValue>),
```
```suggestion
    Range(BinOp,ColId, Bound<AlgebraicValue>, Bound<AlgebraicValue>),
``` 

(for the plan printer)

crates/execution/src/iter.rs
```diff
@@ -1,747 +1,1020 @@
...
-/// An iterator for a left deep join tree
-pub enum LeftDeepJoin<Iter> {
-    /// A standard join
-    Eq(Iter),
-    /// A semijoin that returns the lhs
-    SemiLhs(Iter),
-    /// A semijion that returns the rhs
-    SemiRhs(Iter),
+/// A semijoin iterator.
+/// Returns [RowRef]s for right semijoins.
```
mamcx, line 315:
```suggestion
/// Returns [RowRef]s for the right side.
```

crates/core/src/host/module_host.rs
joshua-spacetime: I don't believe so. Why do you think it's dead?

crates/physical-plan/src/plan.rs
joshua-spacetime: It can, but `TupleField` can handle multiple in scope relvars. But for index joins, the rhs is always a single table, which is why I chose to use `ColId`.

crates/physical-plan/src/plan.rs
joshua-spacetime: Can we defer this decision until we optimize range scans? I'm not sure which representation is best until we add rewrites that use it.

mamcx: I need the `BinOp` for printing the `plan`

mamcx (approved):
LGTM

PR #2085, opened 2025-01-03
Implement the `identity` host call and `ctx.identity()` in Rust

gefjon:
# Description of Changes

This PR adds `ReducerContext::identity(&self) -> Identity`, a method which reads the module identity. It's useful e.g. so modules can assert that scheduled reducers are being called by the scheduler, rather than by a client.
It's also a part of the API spec that fell by the wayside during 0.12.

This PR also adds a smoketest that said method does what it's supposed to, and adds a call to it in the `rust-wasm-test` module that I used during manual testing.

Still TODO: implement the same method in C# modules.

Adds a new API and a new ABI function, but doesn't break anything.

1: it's easy!

# Testing

- [x] Wrote a new smoketest of the new functionality.
- [x] Manually called the `assert_caller_identity_is_module_identity` reducer of `rust-wasm-test` and watched it panic with the expected log output for a client caller.


coolreader18 (approved):
LGTM! Was gonna suggest caching it in a `static _: OnceLock` but host calls are quick and I don't think this is gonna be in a hot path anyway.

gefjon:
> LGTM! Was gonna suggest caching it in a `static _: OnceLock` but host calls are quick and I don't think this is gonna be in a hot path anyway.

Yeah, I considered that, but I figure we don't care and I wanted to keep the PR simple.

PR #2087, opened 2025-01-06
CLI - Fix `spacetime energy balance` not working

bfops:
# Description of Changes

`spacetime energy status` was renamed to `spacetime energy balance`, but only partially. This completely borked using the command with _either_ name.

This completes a previous API break.

1

# Testing

- [x] `cargo run -- energy balance` runs successfully now
```
{"balance":"0"}
```
- [x] Added a smoketest


cloutiertyler (approved):
Love the regression test.

PR #2088, opened 2025-01-06
C#: fix class names for unique column indices

RReverser:
# Description of Changes

Couple of issues reported by @SteveBoytsun turned out to be a class name conflict for unique indexes - the class name was based on the name of the table instead of the name of the column, leading to duplicate classes with different contents, which, in turn, caused all sorts of weird compilation errors.

The fix is pretty trivial, and added a regression test in the PR (fails before the PR, works after).

corresponding GitHub label.

# Testing

- [x] `dotnet test` + used @SteveBoytsun's repo for manual testing


gefjon (approved):
Thanks for the tests!

PR #2091, opened 2025-01-07
Implement `ctx.Identity` for C#

### LINKED ISSUE #2089: Implement `ctx.Identity` getter in C# modules
Analogous to the `ctx.identity()` method in Rust, as implemented in [SpacetimeDB#2085](https://github.com/clockworklabs/SpacetimeDB/pull/2085).

TBH I forget whether we specified this as a method or an attribute in C#, but whatever. Either way, the getter makes a call to the `identity` host function (added in the above PR), treats its result as a little-endian `Identity`, and returns that `Identity`. This should take all of 15 minutes to implement, hopefully.

RReverser:
# Description of Changes

Fixes #2089.

Since C# doesn't have fixed-size arrays available to safe code, I decided it's simpler to accept a pointer to `Identity` itself. It's stored as little-endian anyway, so the only change that was necessary is to mark the struct as FFI-compatible. The actual FFI declaration for `identity` looks a bit different from others for .NET reasons - see comment on the declaration for more details, but the code itself is still pretty simply.

Marking the struct as FFI-compatible also allowed to remove a bunch of helpers, casts and intermediate allocations, since we can now read/write `Identity` and `Address` directly instead of reaching into their internal representation.

For testing, I moved the Rust-specific integration test from smoketests to our "standalone integration test" which runs for both Rust and C#, and implemented C# counterpart to it.

corresponding GitHub label.

1

# Testing

- [x] Ran relevant standalone integration tests and fixed the issues.


gefjon (approved):
Looks great! Thanks for calling out that you moved the test in your PR description; that saved me a lot of confusion.

Issue #2019, opened 2024-11-27
perf: Do not use pointer map for tables with a unique constraint


Centril:

gefjon:
I would perhaps reframe this as making it so that every table has a single primary key index, which by default is on the tuple of all its columns, but can be narrowed to a single row or a smaller tuple. But that may be too large of a rearchitecting to make in our timeframe.

In the simpler ("hacky") version, `Table::delete`, aka `delete_by_row_value`, which devolves to `Table::find_same_row`, will need to contain additional logic to locate a unique index and use it to scan for the sought row. I am not concerned about the performance of this, but it may require significant fiddling just to make it work.

PR #2092, opened 2025-01-07
perf(2019): skip `PointerMap` given a unique index

Centril:
# Description of Changes

This PR removes the pointer map entirely if there is a unique index and uses that instead to maintain set semantics.
Unique indices are also used to scan for rows to delete in that case when doing "delete by value".

Fixes https://github.com/clockworklabs/SpacetimeDB/issues/2019.

Builds upon https://github.com/clockworklabs/SpacetimeDB/pull/2069.
With that PR, the numbers on master are:
```
2025-01-03T09:35:54.415502Z  INFO: : Timing span "update_positions_by_id": 1.64947372s
2025-01-03T09:36:00.880068Z  INFO: : Timing span "update_positions_by_collect": 1.26073871s
```

Perf numbers on this PR.

```
for i in {1..5}; do spacetime call basics update_positions_by_id -s local; done

2025-01-07T23:56:50.151582Z  INFO: : Timing span "update_positions_by_id": 1.516822083s
2025-01-07T23:56:52.097840Z  INFO: : Timing span "update_positions_by_id": 1.451819161s
2025-01-07T23:56:54.108770Z  INFO: : Timing span "update_positions_by_id": 1.547491017s
2025-01-07T23:56:56.071318Z  INFO: : Timing span "update_positions_by_id": 1.50619533s
2025-01-07T23:56:58.155637Z  INFO: : Timing span "update_positions_by_id": 1.569858904s

for i in {1..5}; do spacetime call basics update_positions_by_collect -s local; done

2025-01-07T23:52:44.146575Z  INFO: : Timing span "update_positions_by_collect": 1.10285933s
2025-01-07T23:52:45.743800Z  INFO: : Timing span "update_positions_by_collect": 1.071438136s
2025-01-07T23:52:47.273903Z  INFO: : Timing span "update_positions_by_collect": 1.060757941s
2025-01-07T23:52:48.852988Z  INFO: : Timing span "update_positions_by_collect": 1.089778179s
2025-01-07T23:52:50.448466Z  INFO: : Timing span "update_positions_by_collect": 1.058645907s
```

Flamegraph of `InstanceEnv::insert`:

Screenshot 2025-01-08 at 01-03-40 Flame Graph

The flamegraph indicates that as a %, more time is being spent in `insert_physically_bsatn`, and since that hasn't changed, this entails that less time is being spent elsewhere. `CommittedState::merge` and the delete path is also faster now, but is not shown in this zoomed in part of the flamegraph. The general flamegraph is: https://flamegraph.com/share/cd26706c-cd55-11ef-9832-26c3e5347170

The next step is to add the `update` ABI.

None

3, Fairly simple set of changes, but in a pretty core place so there's some risk of "did we miss to change in this other location".
To mitigate this risk, the code has been grepped and refactored to unify code paths so logic can be changed in fewer place.

# Testing

The old pointer map set semantic logic is still tested in e.g., `insert_duplicate_set_semantic` and the higher up tests (in the datastore) have tests for unique indices.

### DIFF
--- a/crates/table/src/table.rs
+++ b/crates/table/src/table.rs
@@ -65,7 +65,13 @@ pub struct Table {
     /// Page manager and row layout grouped together, for `RowRef` purposes.
     inner: TableInner,
     /// Maps `RowHash -> [RowPointer]` where a [`RowPointer`] points into `pages`.
-    /// If `None`, there is a unique index that implies the set semantic constraint.
+    /// A [`PointerMap`] is effectively a specialized unique index on all the columns.
+    ///
+    /// In tables without any other unique constraints,
+    /// the pointer map is used to enforce set semantics,
+    /// i.e. to prevent duplicate rows.
+    /// If `self.indexes` contains at least one unique index,
+    /// duplicate rows are impossible regardless, so this will be `None`.
     pointer_map: Option<PointerMap>,
     /// The indices associated with a set of columns of the table.
     /// The order is used here to keep the smallest indices first.
@@ -214,7 +220,7 @@ impl Table {
         let static_layout = StaticLayout::for_row_type(&row_layout).map(|sl| (sl, static_bsatn_validator(&row_layout)));
         let visitor_prog = row_type_visitor(&row_layout);
         // By default, we start off with an empty pointer map,
-        // which is removed when the first unique index is.
+        // which is removed when the first unique index is added.
         let pm = Some(PointerMap::default());
         Self::new_with_indexes_capacity(schema, row_layout, static_layout, visitor_prog, squashed_offset, pm)
     }
@@ -243,7 +249,11 @@ impl Table {
     /// Insert a `row` into this table, storing its large var-len members in the `blob_store`.
     ///
     /// On success, returns the hash, if any, of the newly-inserted row,
-    /// and a `RowRef` referring to the row.
+    /// and a `RowRef` referring to the row.s
+    /// The hash is only computed if this table has a [`PointerMap`],
+    /// i.e., does not have any unique indexes.
+    /// If the table has unique indexes,
+    /// the returned `Option<RowHash>` will be `None`.
     ///
     /// When a row equal to `row` already exists in `self`,
     /// returns `InsertError::Duplicate(existing_row_pointer)`,
@@ -552,7 +562,7 @@ impl Table {
             .indexes
             .iter()
             .find(|(_, idx)| idx.is_unique())
-            .expect("there is at least one unique index");
+            .expect("there should be at least one unique index");
         // Project the needle row to the columns of the index, and then seek.
         // As this is a unique index, there are 0-1 rows for this key.
         let needle_row = unsafe { needle_table.get_row_ref_unchecked(needle_bs, needle_ptr) };
@@ -560,10 +570,17 @@ impl Table {
         idx.seek(&key).next()
     }
 
-    /// Insert row identified by `ptr` into the pointer map.
+    /// Insert the row identified by `ptr` into the table's [`PointerMap`],
+    /// if the table has one.
+    ///
     /// This checks for set semantic violations.
-    /// Deletes the row and returns an error if there were any violations.
-    /// Returns the row hash computed, if any.
+    /// If a set semantic conflict (i.e. duplicate row) is detected by the pointer map,
+    /// the row will be deleted and an error returned.
+    /// If the pointer map confirms that the row was unique, returns the `RowHash` of that row.
+    ///
+    /// If this table has no `PointerMap`, returns `Ok(None)`.
+    /// In that case, the row's uniqueness will be verified by [`Self::insert_into_indices`],
+    /// as this table has at least one unique index.
     ///
     /// SAFETY: `self.is_row_present(row)` must hold.
     /// Post-condition: If this method returns `Ok(_)`, the row still exists.
@@ -600,25 +617,31 @@ impl Table {
         // add it to the `pointer_map`.
         self.pointer_map
             .as_mut()
-            .expect("pointer map was confirmed to exist previously")
+            .expect("pointer map should exist, as it did previously")
             .insert(hash, ptr);
 
         Ok(Some(hash))
     }
 
-    /// Returns the list of pointers for this `row_hash`.
+    /// Returns the list of pointers to rows which hash to `row_hash`.
+    ///
+    /// If `self` does not have a [`PointerMap`], always returns the empty slice.
     fn pointers_for(&self, row_hash: RowHash) -> &[RowPointer] {
         self.pointer_map.as_ref().map_or(&[], |pm| pm.pointers_for(row_hash))
     }
 
-    /// Finds the [`RowPointer`] to the row in `target_table` equal, if any,
-    /// by [`eq_row_in_page`], to the row at `needle_ptr` within `needle_table`,
-    /// using the pointer map.
+    /// Using the [`PointerMap`],
+    /// searches `target_table` for a row equal to `needle_table[needle_ptr]`.
+    ///
+    /// Rows are compared for equality by [`eq_row_in_page`].
     ///
     /// Lazily computes the row hash if needed and returns it, or uses the one provided, if any.
     ///
     /// Used for detecting set-semantic duplicates when inserting
-    /// but does nothing if there is a unique constraint that implies a set-semantic constraint.
+    /// into tables without any unique constraints.
+    ///
+    /// Does nothing and always returns `None` if `target_table` does not have a `PointerMap`,
+    /// in which case the caller should instead use [`Self::find_same_row_via_unique_index`].
     ///
     /// Note that we don't need the blob store to compute equality,
     /// as content-addressing means it's sufficient to compare the hashes of large blobs.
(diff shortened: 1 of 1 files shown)


crates/table/src/table.rs
```diff
@@ -66,9 +65,11 @@ pub struct Table {
     /// Page manager and row layout grouped together, for `RowRef` purposes.
     inner: TableInner,
     /// Maps `RowHash -> [RowPointer]` where a [`RowPointer`] points into `pages`.
-    pointer_map: PointerMap,
+    /// If `None`, there is a unique index that implies the set semantic constraint.
```
gefjon, line 68:
```suggestion
    /// The [`PointerMap`], which is effectively a specialized unique index on all the columns.
    ///
    /// In tables without any other unique constraints, the pointer map is used to enforce set semantics,
    /// i.e. to prevent duplicate rows.
    /// If `self.indexes` contains at least one unique index, duplicate rows are impossible regardless,
    /// so this will be `None`.
```

crates/table/src/table.rs
```diff
@@ -144,7 +145,7 @@ impl TableInner {
     }
 }
 
-static_assert_size!(Table, 272);
+static_assert_size!(Table, 264);
```
gefjon, line 148: How did this get smaller?

crates/table/src/table.rs
```diff
@@ -212,7 +213,10 @@ impl Table {
         let row_layout: RowTypeLayout = schema.get_row_type().clone().into();
         let static_layout = StaticLayout::for_row_type(&row_layout).map(|sl| (sl, static_bsatn_validator(&row_layout)));
         let visitor_prog = row_type_visitor(&row_layout);
-        Self::new_with_indexes_capacity(schema, row_layout, static_layout, visitor_prog, squashed_offset, 0)
+        // By default, we start off with an empty pointer map,
+        // which is removed when the first unique index is.
```
gefjon, line 217:
```suggestion
        // which is removed when the first unique index is added.
```

crates/table/src/table.rs
```diff
@@ -238,7 +242,7 @@ impl Table {
 
     /// Insert a `row` into this table, storing its large var-len members in the `blob_store`.
     ///
-    /// On success, returns the hash of the newly-inserted row,
+    /// On success, returns the hash, if any, of the newly-inserted row,
     /// and a `RowRef` referring to the row.
```
gefjon, line 246:
```suggestion
    /// On success, returns the hash, if any, of the newly-inserted row,
    /// and a `RowRef` referring to the row.
    /// The hash is only computed if this table has a [`PointerMap`],
    /// i.e. does not have any unique indexes. If the table has unique indexes,
    /// the returned `Option<RowHash>` will be `None`.
```

crates/table/src/table.rs
```diff
@@ -530,26 +534,53 @@ impl Table {
...
+        needle_table: &Table,
+        needle_bs: &dyn BlobStore,
+        needle_ptr: RowPointer,
+    ) -> Option<RowPointer> {
+        // Find the smallest unique index.
+        let (cols, idx) = target_table
+            .indexes
+            .iter()
+            .find(|(_, idx)| idx.is_unique())
+            .expect("there is at least one unique index");
```
gefjon, line 555:
```suggestion
            .expect("There should be at least one unique index");
```

Formatting per https://doc.rust-lang.org/std/result/enum.Result.html#method.expect .

crates/table/src/table.rs
```diff
@@ -530,26 +534,53 @@ impl Table {
...
+        let needle_row = unsafe { needle_table.get_row_ref_unchecked(needle_bs, needle_ptr) };
+        let key = needle_row.project(cols).expect("needle row should be valid");
+        idx.seek(&key).next()
+    }
+
     /// Insert row identified by `ptr` into the pointer map.
     /// This checks for set semantic violations.
     /// Deletes the row and returns an error if there were any violations.
-    /// Returns the row hash computed.
+    /// Returns the row hash computed, if any.
```
gefjon, line 566: Please update this doc comment to describe that this method should be called even if `self.pointer_map.is_none()`. I suggest:

```rust
/// If this table has a [`PointerMap`], insert the row identified by `ptr` into that map.
///
/// This checks for set semantic violations.
/// If a set semantic conflict (i.e. duplicate row) is detected by the pointer map,
/// the row will be deleted and an error returned.
/// If the pointer map confirms that the row was unique, returns the `RowHash` of that row.
///
/// If this table has no [`PointerMap`], returns `Ok(None)`.
/// In that case, the row's uniqueness will be verified by [`Self::insert_into_indices`],
/// as this table has at least one unique index.

File: crates/table/src/table.rs [hunk 7]
```diff
@@ -567,39 +598,57 @@ impl Table {
         // If the optimistic insertion was correct,
         // i.e. this is not a set-semantic duplicate,
         // add it to the `pointer_map`.
-        self.pointer_map.insert(hash, ptr);
+        self.pointer_map
+            .as_mut()
+            .expect("pointer map was confirmed to exist previously")
```
gefjon, line 603:
```suggestion
            .expect("Pointer map should exist, as it did previously")
```

crates/table/src/table.rs
```diff
@@ -567,39 +598,57 @@ impl Table {
...
+            .insert(hash, ptr);
 
-        Ok(hash)
+        Ok(Some(hash))
     }
 
-    /// Finds the [`RowPointer`] to the row in `committed_table`
-    /// equal, by [`eq_row_in_page`], to the row at `tx_ptr` within `tx_table`, if any.
+    /// Returns the list of pointers for this `row_hash`.
```
gefjon, line 609:
```suggestion
    /// Returns the list of pointers to rows whose hash is equal to `row_hash`.
    ///
    /// If `self` does not have a [`PointerMap]`, always returns the empty slice.
```

crates/table/src/table.rs
```diff
@@ -567,39 +598,57 @@ impl Table {
...
+    fn pointers_for(&self, row_hash: RowHash) -> &[RowPointer] {
+        self.pointer_map.as_ref().map_or(&[], |pm| pm.pointers_for(row_hash))
+    }
+
+    /// Finds the [`RowPointer`] to the row in `target_table` equal, if any,
+    /// by [`eq_row_in_page`], to the row at `needle_ptr` within `needle_table`,
+    /// using the pointer map.
```
gefjon, line 616:
```suggestion
    /// Using the [`PointerMap`], searches `target_table` for a row equal to `needle_table[needle_ptr]`.
    ///
    /// Rows are compared for equality by [`eq_row_in_page`].
```

crates/table/src/table.rs
```diff
@@ -567,39 +598,57 @@ impl Table {
...
+    ///
+    /// Lazily computes the row hash if needed and returns it, or uses the one provided, if any.
     ///
-    /// Used for detecting set-semantic duplicates when inserting.
+    /// Used for detecting set-semantic duplicates when inserting
+    /// but does nothing if there is a unique constraint that implies a set-semantic constraint.
```
gefjon, line 621:
```suggestion
    /// Used for detecting set-semantic duplicates when inserting
    /// into tables without any unique constraints.
    ///
    /// Does nothing and always returns `None`  if `target_table` does not have a [`PointerMap`],
    /// in which case the caller should instead use [`Self::find_same_row_via_unique_index`].
```

crates/table/src/table.rs
```diff
@@ -614,11 +663,45 @@ impl Table {
...
                     )
                 }
-            })
+            });
+
+        (row_hash, row_ptr)
+    }
+
+    /// Finds the [`RowPointer`] to the row in `target_table` equal, if any,
+    /// to the row at `needle_ptr` within `needle_table`.
```
gefjon, line 676:
```suggestion
    /// Searches `target_table` for a row equal to `needle_table[needle_ptr]`,
    /// and returns the [`RowPointer`] to that row in `target_table`, if it exists.
    ///
    /// Searches using the [`PointerMap`] or a unique index, as appropriate for the table.
```

crates/table/src/table.rs
```diff
@@ -768,13 +853,15 @@ impl Table {
         // as the row is already present, set-semantically.
         let (temp_row, _) = self.insert_physically_pv(blob_store, row)?;
         let temp_ptr = temp_row.pointer();
-        let hash = temp_row.row_hash();
 
         // Find the row equal to the passed-in `row`.
+        // This uses one of two approaches.
+        // Either there is a pointer map, so we use that,
+        // or, here is at least one unique index, so we use the smallest one.
```
gefjon, line 860: What does it mean for one index to be "smaller" than another?

gefjon (approved):
>  To mitigate this risk, the code has been grepped and refactored to unify code paths so logic can be changed in fewer place.

Refactorings like this only make review easier if they can be separated from the risky changes. When the changes are combined like this, my work is approximately tripled, as I now have to:

- Determine which changes are related to the refactor versus which are related to the risky change.
- Determine whether the refactor would be correct in isolation.
- Determine whether the risky change is correct given the refactor.

In the future, if you want to do a refactoring with the intention of increasing confidence in a risky change, make the refactoring in a first commit, and the risky change in a second. These don't need to be separate PRs (though they can be), since reviewers can use GitHub's **Changes from all commits** review drop-down to inspect them separately.

I'm going to review this PR anyways since we're on a tight deadline, but in the future when I see things like this I'm going to request changes and refuse to review until things are separated.

PR #1929, opened 2024-11-01
Remove IO macros

bfops:
# Description of Changes

Remove the spacetimedb version of common macros like `println!`, per https://github.com/clockworklabs/SpacetimeDB/issues/1819.

# API and ABI breaking changes

Yes. This removes several spacetimedb macros. They should be replaced with `spacetimedb::log::info!`, `spacetimedb::log::error!`, or `spacetimedb::log::debug!`.

# Expected complexity level and risk

2

# Testing

Existing automated tests have been updated.

PR #2099, opened 2025-01-09
Use `println`/`eprintln` in the quickstart-chat client, not `log`

gefjon:
# Description of Changes

Likely by mistake, #1929 rewrote the `quickstart-chat` client to use `log::info!` and `log::error!` for output.
In addition to needlessly adding complexity to a quickstart project, this was broken because that program didn't install a logger, so it produced no output at all.

This commit reverts those changes, returning to `println`/`eprintln` output in the quickstart-chat client.
We could use `log::` macros and install a logger,
but then we'd have to explain installing a logger in the quickstart guide.

This commit also makes me a codeowner of the `quickstart-chat` project, both the module and the client.
I think extra scrutiny on these files is important because they're always supposed to exactly match the code presented in the tutorials, which prior to this commit they clearly did not.

N/a.

1

# Testing

- [x] Ran `quickstart-chat` locally, typed a few messages, set my name, saw output as expected.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -7,3 +7,6 @@ LICENSE.txt @cloutiertyler
 /crates/cli/src/ @bfops @cloutiertyler
 /crates/cli/src/subcommands/generate/ # No owners
 /crates/cli/src/subcommands/generate/mod.rs @bfops @cloutiertyler # These codeowners should be the same as the "root" CLI codeowners
+
+/crates/sdk/examples/quickstart-chat/ @gefjon
+/modules/quickstart-chat/ @gefjon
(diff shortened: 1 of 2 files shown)


cloutiertyler:
> In addition to needlessly adding complexity to a quickstart project

What is the needless complexity that you're referring to? This change was necessary in the module code to implement the planned removal of running clippy by default, and enforcing that people don't use `println!` via grep. It should not have been made in the client however.

gefjon:
> > In addition to needlessly adding complexity to a quickstart project
> 
> What is the needless complexity that you're referring to? This change was necessary in the module code to implement the planned removal of running clippy by default, and enforcing that people don't use `println!` via grep. It should not have been made in the client however.

Correct. It is needless added complexity *in clients*, since it adds an additional dependency on `log` and setting up a logger. I thought it obvious that my statements here were about the change to the quickstart-chat client, not about any other part of our codebase.

PR #2102, opened 2025-01-09
add signature of `datastore_update_bsatn` host call

Centril:
# Description of Changes

Adds the ABI `datastore_update_bsatn`, defining the signature in `bindings-sys` and in the host.
However, the ABI is not used by the `bindings` crate (todo) and the `InstanceEnv::update` method added also is `todo!()`.

This PR is opened so that we can discuss and decide on the ABI signature and semantics.

Some design choices made here:
1. Sequences *are* generated for the row provided, matching the semantics of `insert`. This does have a minor adverse perf impact. If it turns that this is a problem, we can nix that. There are also avenues for optimizing sequence generation by e.g., storing a `(ColList, Box<[SequenceId]>)` directly in a `Table`  to avoid going to the `TableSchema`.
2. The search for the row-to-up-date is done after generating and writing back sequence values.
3. Unique constraints that aren't the one in the index are checked. This is not really optional, as it would otherwise allow unique constraints to be violated.
4. When a row wasn't found, an error is returned.

None.

3, its not used yet, but we are deciding on a signature and ABI call semantics in this PR.

# Testing

Nothing to test yet. This has not been integrated yet.

### DIFF
--- a/crates/bindings-sys/src/lib.rs
+++ b/crates/bindings-sys/src/lib.rs
@@ -346,7 +346,7 @@ pub mod raw {
         ///
         /// The row to update is found by projecting `row`
         /// to the type of the *unique* index identified by `index_id`.
-        /// If no row is found, `row` is inserted.
+        /// If no row is found, the error `NO_SUCH_ROW` is returned.
         ///
         /// To handle auto-incrementing columns,
         /// when the call is successful,
@@ -372,6 +372,7 @@ pub mod raw {
         /// - `NO_SUCH_TABLE`, when `table_id` is not a known ID of a table.
         /// - `NO_SUCH_INDEX`, when `index_id` is not a known ID of an index.
         /// - `INDEX_NOT_UNIQUE`, when the index was not unique.
+        /// - `NO_SUCH_ROW`, when the row was not found in the unique index.
         /// - `BSATN_DECODE_ERROR`, when `row` cannot be decoded to a `ProductValue`
         ///    typed at the `ProductType` the table's schema specifies
         ///    or when it cannot be projected to the index identified by `index_id`.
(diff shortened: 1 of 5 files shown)


crates/bindings-sys/src/lib.rs
```diff
@@ -332,11 +332,58 @@ pub mod raw {
...
+        /// Updates a row in the table identified by `table_id` to `row`
+        /// where the row is read from the byte string `row = row_ptr[..row_len]` in WASM memory
+        /// where `row_len = row_len_ptr[..size_of::<usize>()]` stores the capacity of `row`.
+        ///
+        /// The byte string `row` must be a BSATN-encoded `ProductValue`
+        /// typed at the table's `ProductType` row-schema.
+        ///
+        /// The row to update is found by projecting `row`
+        /// to the type of the *unique* index identified by `index_id`.
+        /// If no row is found, `row` is inserted.
```
gefjon, line 349: > If no row is found, `row` is inserted.

I do not agree with this choice. My intuition is that an `update` call where no such row previously existed is more likely to be a bug than intended behavior. As such, I think an `update` that finds no such row should roll back the insertion and return an error. I don't see this being a performance concern, as we expect that branch to be rarely taken.

gefjon (commented):
> Sequences are generated for the row provided, matching the semantics of insert.

I agree with this choice. I think we do not want to expose users to a way to accidentally skip auto-inc processing.

> The search for the row-to-up-date is done after generating and writing back sequence values.

:shrug: I don't think it really matters. AFAICT this is only observably different when updating by a `#[unique] #[auto_inc]` column and supplying 0 as that column's value in the new row, which is a nonsense thing to do.

> Unique constraints that aren't the one in the index are checked. This is not really optional, as it would otherwise allow unique constraints to be violated.

I agree with this assessment.

PR #2103, opened 2025-01-09
Fix defining multiple tables on the same `struct` in Rust modules

### LINKED ISSUE #2097: BUG: Cannot define multiple tables with the same type in Rust
Image

gefjon:
# Description of Changes

Fixes #2097 .

This PR makes it possible to apply the `#[spacetimedb::table]` attribute multiple times to the same `struct` item, resulting in multiple tables with the same row type.

Tip for reviewers: This PR contains 3 distinct commits with meaningful changes (plus some more with tests and stuff), the first two of which are refactorings and the third of which is the actual fix. I recommend reviewing these separately, using the "Changes from **all commits**" dropdown in the top-left corner of the GitHub review UI. Commit messages, in order, reproduced for your reviewing pleasure:

## Nix `MapRowTypeToTable`; make table handles nameable

In a noble but misguided effort to make table handles not nameable, previous versions of our `bindings-macro` `#[table]` attribute hid the table struct away in a `const _: () = {}` block, and used a hidden trait `MapRowTypeToTable` to name it where necessary in the macro output.

This imposed the requirement that each row type corresponded to exactly one table. This is not the only part of the macros that imposes that requirement, so further commits will follow.

## Remove use of `cvt_attr` in `table` attribute macro

Making multiple tables on the same type work is going to require some amount of splicing the attributes list of the `item` passed to `table`. The `cvt_attr` helper prevents this, allowing only prepending and appending token streams. This commit inlines that use of `cvt_attr`, so that future commits can make the necessary changes.

It defines a new helper, `ok_or_compile_error`, which does the useful part of `cvt_ret`, namely allowing `?` short circuiting with errors returning `e.into_compile_error()`.

## Support multiple tables with the same row type

This commit reworks the `spacetimedb::table` macro such that it never emits more than one `impl` of each of the relevant traits (`Serialize`, `Deserialize`, `SpacetimeType`).
This is accomplished by:
- Moving the `impl`s of those traits out of `table_impl` and into `derive(__TableHelper)`.
- `pushnew`-ing a `#[derive(__TableHelper)]` into the end of the `attrs: Vec<Attribute>` in the `DeriveInput`. That way, later `#[spacetimedb::table]` macros are processed *before* the `__TableHelper` is expanded, so they will see that the `#[derive(__TableHelper)]` in that `attrs` list, and know not to add another.

N/a

3: Our macros are complex, under-documented and brittle.

# Testing

- [x] Manually built and published a module containing the `player` and `logged_out_player` example from #2097 .
- [x] Added the same example to the `rust-wasm-test` module, which gets built during CI.


crates/bindings-macro/src/lib.rs
```diff
@@ -159,6 +160,28 @@ pub fn reducer(args: StdTokenStream, item: StdTokenStream) -> StdTokenStream {
...
+
+    impl Parse for Wrapper {
+        fn parse(input: ParseStream) -> syn::Result<Self> {
+            input
+                .call(Attribute::parse_outer)
+                .map(|attrs| Wrapper(attrs.into_iter().next().unwrap()))
+        }
+    }
+
+    syn::parse2::<Wrapper>(source).unwrap().0
```
coolreader18, line 182: I do believe this can just be expressed as `syn::Parser::parse2(Attribute::parse_outer, source).unwrap().into_iter().next().unwrap()`

coolreader18: or `Attribute::parse_outer.parse2(source)....` if you print `syn::parse::Parser` into scope

gefjon: Neat! Applied, thanks.

cloutiertyler (approved):
I have confirmed that this works with https://github.com/clockworklabs/SpacetimeDBCircleGame. I updated that game to have two tables use the same type and I was able to play the game and move around.

I also confirmed that I was able to publish a module with a scheduled table and a regular table:

```rust
#[spacetimedb::table(name = move_all_players_timer, scheduled(move_all_players))]
#[spacetimedb::table(name = move_all_player_timer_2, public)]
pub struct MoveAllPlayersTimer {
    #[primary_key]
    #[auto_inc]
    scheduled_id: u64,
    scheduled_at: spacetimedb::ScheduleAt,
}
```

I also reviewed the code itself and did not find issues.

The update to the test module and snapshot tests also look good to me.

crates/bindings/src/table.rs
```diff
@@ -149,11 +149,6 @@ pub struct ScheduleDesc<'a> {
     pub scheduled_at_column: u16,
 }
 
-#[doc(hidden)]
-pub trait __MapRowTypeToTable {
```
kazimuth, line 153: The downside of this removal is that there is now no way to link from row types to their table types in locally generated documentation.

We could consider adding a new trait:

```rust
/// Trait that documents which tables a row type is used for.
pub trait RowTypeForTable<T: Table> {}
```

implemented once for each table declared on a row type. This would serve no purpose aside from documentation.

kazimuth (commented):
Everything here makes sense.

crates/bindings/src/table.rs
gefjon: If we wanted to link from row types to tables in docs (which I would not take as a given that we do), I think we would do it by having the `table` attribute macro add a `doc = "Used as the row type for the table [this table handle type]"` attribute to the `struct` item. No trait is necessary.

PR #2105, opened 2025-01-10
Rework RLS macro for amended syntax

gefjon:
# Description of Changes

Out-of-band discussion resulted in a different syntax for registering RLS rules in module code. This PR implements that new syntax in Rust modules. C# modules, which did not have an implementation prior, are unaffected.

The new macro's implementation is significantly simpler, with all the interesting work done by "runtime" code rather than in the macro. One effect of this is that SQL syntax errors are no longer detected at compile time. Instead, they will only be checked at `spacetime publish` time, when the filter query is processed by the SpacetimeDB host. I consider this reasonable, as even in the previous design, most "interesting" errors (e.g. name resolution, type checking) could only be detected at publish time anyways. The new macro could perform SQL syntax checking in the same way as the old one, but I consider the benefits of this to be outweighed by the added complexity in our macro code it would require.

This commit also applies `#[doc(hidden)]` to everything RLS-related, with TODO comments to make them visible once RLS is fully implemented.

No, in the sense that RLS is not actually implemented, and so no one could possibly have been depending on it.

1: new macro is much simpler; internals are unchanged.

# Testing

- [x] Updated smoketests and SDK test module to use the new macro syntax. The smoketest makes assertions about the resulting system tables.
- [ ] It's essentially impossible to do any more testing, as RLS is not implemented, so this macro doesn't actually do anything currently.


mamcx (approved):
LGTM

crates/bindings-macro/src/lib.rs
```diff
@@ -360,17 +359,25 @@ pub fn schema_type(input: StdTokenStream) -> StdTokenStream {
     })
 }
 
-/// Generates code for registering a row-level security `SQL` function.
+/// Generates code for registering a row-level security visibility.
```
joshua-spacetime, line 362:
```suggestion
/// Generates code for registering a row-level security rule.
```

crates/bindings-macro/src/lib.rs
```diff
@@ -360,17 +359,25 @@ pub fn schema_type(input: StdTokenStream) -> StdTokenStream {
...
 ///
-/// A row-level security function takes a `SQL` query expression that is used to filter rows.
+/// This attribute must be applied to a `const` binding of type [`Filter`].
+/// The filter will be installed as if an additional SQL `WHERE` predicate
+/// in all client queries against the table to which the filter applies.
```
joshua-spacetime, line 366:
```suggestion
/// It will be interpreted as a filter on the table to which it applies, for all client queries.
```

joshua-spacetime (approved):

PR #2118, opened 2025-01-13
Rust SDK: `with_credentials` -> `with_token`

gefjon:
# Description of Changes

See discussion on https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/pull/212

This PR replaces `DbConnectionBuilder::with_credentials` with `::with_token`, which does not require an `Identity` as argument. It also amends our machinery for saving and loading credentials to and from disk to suit the new API.

Yep!

1.

# Testing

- [x] I believe automated tests to be sufficient.

### DIFF
--- a/crates/sdk/examples/quickstart-chat/main.rs
+++ b/crates/sdk/examples/quickstart-chat/main.rs
@@ -43,8 +43,8 @@ fn creds_store() -> credentials::File {
 }
 
 /// Our `on_connect` callback: save our credentials to a file.
-fn on_connected(_ctx: &DbConnection, identity: Identity, token: &str) {
-    if let Err(e) = creds_store().save(identity, token) {
+fn on_connected(_ctx: &DbConnection, _identity: Identity, token: &str) {
+    if let Err(e) = creds_store().save(token) {
         eprintln!("Failed to save credentials: {:?}", e);
     }
 }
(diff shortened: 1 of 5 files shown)


RReverser (approved):
I'm not a fan of `impl ToString` since this accepts almost any type in Rust ecosystem (any type that implements `Display`), eg user can accidentally pass `Some(true)` as a token and it will be silently stringified, but it's up to you.

No concerns otherwise.

gefjon:
> I'm not a fan of `impl ToString` since this accepts almost any type in Rust ecosystem (any type that implements `Display`), eg user can accidentally pass `Some(true)` as a token and it will be silently stringified, but it's up to you.
> 
> No concerns otherwise.

We use `impl ToString` elsewhere. I'm not aware of any other trait in `std` that means "string-like thing that can be converted to `String`."

RReverser:
> We use `impl ToString` elsewhere. I'm not aware of any other trait in `std` that means "string-like thing that can be converted to `String`."

Literally `impl Into<String>` is much better IMO. It covers all the usual suspects - `String`, `&str`, `Cow<str>` etc.

gefjon:
> Literally impl Into<String> is much better IMO. It covers all the usual suspects - String, &str, Cow<str> etc.

If you want to open a follow-up PR that changes all uses of `impl ToString` in `DbConnectionBuilder`, I will happily accept it.

PR #2125, opened 2025-01-14
Add new crates to `publish-crates.sh` script

bfops:
# Description of Changes

We had several crates that were missing from the `publish-crates.sh` script, meaning that manual intervention was required for it to run properly.

No code changes

1

# Testing

I believe this is where I added them in order to successfully publish to crates.io, but it's hard to re-test after the fact :shrug: Unfortunately the `--dry-run` option doesn't "just work" the way we'd want.

We'd have to set up our own temporary package repository in order to test this comprehensively.

The worst case is that we have to make further corrections next time we publish, which is the same state we're in without this PR.

### DIFF
--- a/tools/publish-crates.sh
+++ b/tools/publish-crates.sh
@@ -23,7 +23,7 @@ if [ $DRY_RUN != 1 ] ; then
 fi
 
 BASEDIR=$(pwd)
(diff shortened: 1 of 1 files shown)


cloutiertyler (approved):
It's very low risk, so although the testing is suspect, I think it's fine to merge.

bfops:
Yeah, unfortunately the `--dry-run` option doesn't "just work" the way we'd want.

We'd have to set up our own temporary package repository in order to test this comprehensively.

The worst case is that we have to make further corrections next time we publish, which is the same state we're in without this PR.

PR #2126, opened 2025-01-14
Finish spacetimedb-update functionality

coolreader18:
# Description of Changes

based off #2011

Adds implementations to all the subcommands of `spacetime version`.

Expected github release artifact structure:

* artifact filename: `spacetime-{rustc-target-triple}.{ext}`
  * `ext = if windows { "zip" } else { "tar.gz" }`
* artifact contents:
  * `spacetimedb-cli[.exe]`, `spacetimedb-standalone[.exe]`, `spacetimedb-update[.exe]`

~~cc @clockworklabs/devops - feedback on this^^ would be nice. I kept the edge cases since that's how it already was and I figured that might make it easier in the release workflow, but honestly I'd prefer if it was just `spacetime-{rustc-target-triple}.{ext}` if possible. Not sure where the release flow is or if it's even automated.~~ Implemented in #2193.

3 - a complex bit of code that's implementing a theoretically perpetual contract between `spacetimedb-update`, github release artifacts, and the directory structure.

# Testing

- [x] Manually tested that `spacetime version install` works
- [ ] I'm working on an automated regrtest to make sure the functionality works.
- [ ] reviewer: if you wanna, play around with it and try and catch out any edge cases?


bfops:
(It looks like this currently isn't passing CI, so I'm holding off on review)

crates/auth/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+edition.workspace = true
+rust-version.workspace = true
+
+[dependencies]
+spacetimedb-lib.workspace = true
+
+anyhow.workspace = true
+serde.workspace = true
+serde_with.workspace = true
+jsonwebtoken.workspace = true
```
bfops, line 13: at first glance this seems unrelated to the `update` rework.. was there a circular dependency or something?

crates/cli/src/config.rs
```diff
@@ -830,6 +829,53 @@ Update the server's fingerprint with:
...
+/// **NOTE**: This function will preserve the formatting and comments of the original value.
+pub fn set_table_opt_value(table: &mut toml_edit::Table, key: &str, value: Option<&str>) {
+    let old_value = table.get(key);
+    if let Some(new) = value {
+        table[key] = copy_value_with_decor(old_value, new);
+    } else {
+        table.remove(key);
+    }
+}
+
```
bfops, line 878: similar to above - why did these functions need to get moved?

crates/auth/Cargo.toml
coolreader18: CLI depended on this component of core, so I split it out to a separate crate. This was in the [`Remove spacetimedb-core as a dep of cli`](https://github.com/clockworklabs/SpacetimeDB/pull/2126/commits/35e54883eb11abdf19b529faec45b13b2f1c1eb5) commit.

crates/cli/src/config.rs
coolreader18: They were in core and depended upon by cli.

crates/update/src/cli/list.rs
```diff
@@ -0,0 +1,29 @@
...
+        for ver in versions {
+            print!("{ver}");
+            if Some(&ver) == current.as_ref() {
+                print!(" (current)");
+            }
+            println!();
+        }
+        Ok(())
+    }
+}
```
bfops, line 29: I get this error:
```
$ cargo run -- version list
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.38s
     Running `target/debug/spacetimedb-cli version list`
Error: No such file or directory (os error 2)
```

Is it just because I'm running from `target` instead of a proper install?

crates/cli/src/lib.rs
```diff
@@ -4,23 +4,22 @@ mod config;
...
 
 pub use config::Config;
-use spacetimedb_paths::SpacetimePaths;
+use spacetimedb_paths::{RootDir, SpacetimePaths};
 pub use subcommands::*;
 pub use tasks::build;
 
 pub fn get_subcommands() -> Vec<Command> {
     vec![
-        version::cli(),
```
bfops, line 23: is this meant to be replaced with `version list`?

coolreader18: Yes, and it's added just further down.

crates/update/src/cli/list.rs
coolreader18: Probably, yes. I can improve the error message.

crates/paths/src/cli.rs
```diff
@@ -15,19 +17,90 @@ impl ConfigDir {
...
+            Err(e) if e.kind() == std::io::ErrorKind::NotFound => Ok(None),
+            Err(e) => Err(e.into()),
+        }
+    }
+
+    pub fn installed_versions(&self) -> impl Iterator<Item = anyhow::Result<semver::Version>> {
+        enum Either<A, B> {
+            A(A),
+            B(B),
+        }
```
bfops, line 80: why do we need an entirely new type just to return a list of versions here?

coolreader18: Idk, I guess I can just collect it to vec

crates/cli/src/lib.rs
bfops: Ah it looks like `--version` does this now?

crates/cli/src/subcommands/generate/mod.rs
```diff
@@ -129,7 +128,7 @@ pub async fn exec(config: Config, args: &clap::ArgMatches) -> anyhow::Result<()>
             build::exec_with_argstring(config.clone(), project_path, build_options).await?
         };
         let spinner = indicatif::ProgressBar::new_spinner();
-        spinner.enable_steady_tick(60);
+        spinner.enable_steady_tick(std::time::Duration::from_millis(60));
```
bfops, line 131: After some digging, it looks like this must have been due to a package version upgrade. Did we need to bump this package version in this PR for some reason?

crates/auth/Cargo.toml
bfops: Was there some reason the CLI couldn't (continue to) depend on core? PR diffs get complicated quickly by having lots of changes/cleanups/refactorings that aren't related to the core change.. If you think reviewers will have an easier time reviewing commit-by-commit, please put that in the PR description because it's not necessarily the way that PRs are reviewed by default.

crates/cli/src/subcommands/generate/mod.rs
```diff
@@ -463,9 +469,9 @@ impl WasmCtx {
         let (mem, env) = Self::mem_env(&mut caller);
 
         // Read `buffer_len`, i.e., the capacity of `buffer` pointed to by `buffer_ptr`.
-        let buffer_len = u32::read_from(mem, buffer_len_ptr)?;
+        let buffer_len = u32::from_le_bytes(deref_slice(mem, buffer_len_ptr, 4)?.try_into().unwrap());
         // Write `buffer` to `sink`.
-        let buffer = mem.deref_slice(buffer_ptr, buffer_len)?;
+        let buffer = deref_slice(mem, buffer_ptr, buffer_len)?;
```
bfops, line 474: hoo boy am I not equipped to review these changes - why is this related to the spacetimedb-update changes?

crates/core/src/auth/token_validation.rs
```diff
@@ -157,7 +157,7 @@ impl TokenValidator for DecodingKey {
 
         let data = decode::<IncomingClaims>(token, self, &validation)?;
         let claims = data.claims;
-        claims.try_into()
+        claims.try_into().map_err(TokenValidationError::Other)
```
bfops, line 160: same question here - why is this change related to the `update` binary?

crates/data-structures/Cargo.toml
```diff
@@ -6,7 +6,7 @@ license-file = "LICENSE"
 description = "Assorted data structures used in spacetimedb"
 
 [features]
-serde = ["dep:serde"]
+serde = ["dep:serde", "hashbrown/serde"]
```
bfops, line 9: ?

crates/standalone/src/lib.rs
```diff
@@ -438,13 +439,12 @@ fn withdraw_energy(control_db: &ControlDb, identity: &Identity, amount: EnergyQu
 pub async fn exec_subcommand(cmd: &str, args: &ArgMatches) -> Result<(), anyhow::Error> {
     match cmd {
         "start" => start::exec(args).await,
-        "version" => version::exec(args).await,
```
bfops, line 441: reviewer note: this is now provided by `--version` in `crates/standalone/src/main.rs`

crates/cli/src/subcommands/start.rs
```diff
@@ -52,13 +58,16 @@ pub async fn exec(paths: &SpacetimePaths, args: &ArgMatches) -> anyhow::Result<E
...
+        let err = cmd.exec();
+        Err::<std::process::ExitStatus, _>(err)
+    };
+    #[cfg(not(unix))]
+    let result = cmd.status();
 
-    let status = cmd
-        .status()
-        .with_context(|| format!("exec failed for {}", bin_path.display()))?;
+    let status = result.with_context(|| format!("exec failed for {}", bin_path.display()))?;
```
bfops, line 71: I believe these changes are to address https://github.com/clockworklabs/SpacetimeDB/issues/2133. 

Given that that issue has a different purpose and surface area for testing than the update functionality, I would like to suggest addressing it in a separate PR.

bfops:
Note to self: I still need to review:
- `crates/cli/src/subcommands/version.rs`
- `crates/paths`
- `crates/update` itself

crates/update/src/cli/upgrade.rs
```diff
@@ -0,0 +1,27 @@
...
+                if new_update_binary.exists() {
+                    tokio::fs::copy(new_update_binary, &paths.cli_bin_file).await?;
+                    eprintln!("Self-updated `spacetime version` command")
+                }
+            }
+
+            Ok(())
+        })?
+    }
+}
```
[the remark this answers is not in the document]
bfops: ah it's because the current artifact is called `spacetime.linux-amd64.tar.gz` rather than `spacetime-linux-amd64.tar.gz`

crates/update/src/cli/install.rs
```diff
@@ -0,0 +1,207 @@
...
+
+pub(super) async fn available_releases(client: &reqwest::Client) -> anyhow::Result<Vec<semver::Version>> {
+    let url = "https://api.github.com/repos/clockworklabs/SpacetimeDB/releases";
+    let releases: Vec<Release> = client.get(url).send().await?.json().await?;
+
+    releases.into_iter().map(|release| release.version()).collect()
+}
+
+#[derive(Deserialize)]
+struct ReleaseAsset {
```
bfops, line 155: can we rename these to indicate that they're API types (not types that we get to choose).

e.g.
```suggestion
// The type returned by GitHub's API to describe a release asset
struct GHApiReleaseAsset {
```

bfops: nevermind, I realize these were transplanted from the previous code :+1: can always clean it up later

crates/update/src/cli/install.rs
```diff
@@ -0,0 +1,207 @@
...
+    yes: ForceYes,
+}
+
+impl Install {
+    pub(super) fn exec(self, paths: &SpacetimePaths) -> anyhow::Result<()> {
+        super::tokio_block_on(async {
+            anyhow::ensure!(
+                self.edition == ["standalone"],
+                "can only install spacetimedb-standalone at the moment"
+            );
```
bfops, line 40: wait why can we only install standalone right now? The current upgrade flow definitely allows users to install previous versions.

This is already a big PR, so I think it would be fine to leave the scope as-is but we should at least create a followup ticket.

crates/paths/src/cli.rs
```diff
@@ -15,19 +17,90 @@ impl ConfigDir {
...
+            // remove the link if it already exists
+            std::fs::remove_file(&link_path).ok();
+            std::os::unix::fs::symlink(version.to_string(), link_path)?;
+        }
+        #[cfg(windows)]
+        {
+            junction::delete(&link_path).ok();
+            let version_path = self.version_dir(version);
+            junction::create(&version_path, &link_path)
+                .or_else(|err| std::os::windows::fs::symlink_dir(version.to_string(), &link_path).or(Err(err)))?;
```
bfops, line 58: I don't have any experience with windows junctions - what's the reasoning behind this "try `version_path` junction first, then fallback to `symlink_dir(version.to_string(), ...)`" logic?

crates/paths/src/cli.rs
```diff
@@ -15,19 +17,90 @@ impl ConfigDir {
     pub fn cli_toml(&self) -> CliTomlPath {
         CliTomlPath(self.0.join("cli.toml"))
     }
+    pub fn default_version_toml(&self) -> DefaultVersionToml {
```
bfops, line 20: is this used?

crates/update/src/cli.rs
```diff
@@ -0,0 +1,108 @@
...
+#[derive(clap::Subcommand)]
+enum VersionSubcommand {
+    List(list::List),
+    Use(r#use::Use),
+    Upgrade(upgrade::Upgrade),
+    Install(install::Install),
+    Uninstall(uninstall::Uninstall),
+}
+
+fn reqwest_client() -> anyhow::Result<reqwest::Client> {
```
bfops, line 82: is this used?

bfops: ah it's used in the submodules got it :+1:

crates/cli/src/subcommands/generate/mod.rs
coolreader18: There's a function in it I was looking to use that wasn't in the version we had specified, I believe

coolreader18:
I removed the rm-core-as-dep-of-cli commit from this branch, I'll do it as a follow-up PR.

crates/paths/src/cli.rs
coolreader18: oops, no

crates/paths/src/cli.rs
coolreader18: If the filesystem isn't NTFS, junctions won't work. Very uncommon situation though. I'll add a comment.

crates/update/src/cli/install.rs
coolreader18: Standalone as opposed to cloud. I don't believe we are currently able to install cloud with `spacetime update`.

bfops: ah gotcha that makes sense :+1:

crates/update/src/cli/uninstall.rs
```diff
@@ -0,0 +1,22 @@
...
+    yes: ForceYes,
+}
+
+impl Uninstall {
+    pub(super) fn exec(self, paths: &SpacetimePaths) -> anyhow::Result<()> {
+        let Self { version, yes } = self;
+        if yes.confirm(format!("Uninstall v{version}?"))? {
+            let dir = paths.cli_bin_dir.version_dir(&version);
+            std::fs::remove_dir_all(dir)?;
+        }
```
bfops, line 19: if we uninstall the current version, will that leave something hanging in an awkward way? Should we remove the `current` link in that case?

crates/cli/src/subcommands/version.rs
```diff
@@ -1,33 +1,82 @@
...
+    #[cfg(unix)]
+    let result = {
+        use std::os::unix::process::CommandExt;
+        cmd.arg0(applet);
+        let err = cmd.exec();
+        Err::<std::process::ExitStatus, _>(err)
+    };
+    #[cfg(windows)]
+    let result = {
+        cmd.env("SPACETIMEDB_UPDATE_MULTICALL_APPLET", applet);
```
bfops, line 57: is this needed because the `args[0]` behavior is different on windows than on linux e.g. if going via a shortcut/symlink?

What's the advantage of doing this differently on windows vs linux than just always doing this env-based approach?

bfops:
(I merged in `master` to bump CI. The `Internal tests` were failing due to them running before https://github.com/clockworklabs/SpacetimeDBPrivate/pull/1279 merged).

crates/paths/src/cli.rs
```diff
@@ -21,14 +23,65 @@ impl ConfigDir {
...
+            Err(e) if e.kind() == std::io::ErrorKind::NotFound => Ok(None),
+            Err(e) => Err(e.into()),
+        }
+    }
+
+    pub fn installed_versions(&self) -> anyhow::Result<Vec<semver::Version>> {
+        self.read_dir()?
+            .filter_map(|r| match r {
+                Ok(entry) => {
+                    let parsed: semver::Version = entry.file_name().to_str()?.parse().ok()?;
```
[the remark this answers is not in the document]
coolreader18: Now made it so not everything has to be a `semver::Version`, so it works now.

crates/cli/src/subcommands/start.rs
coolreader18: Split out into #2200.

crates/cli/src/subcommands/version.rs
coolreader18: Ultimately this is a hack to get behavior like argv0 on windows. In the future, there'll hopefully be a function akin to argv0 on windows, but for now this is the best approximation. https://github.com/rust-lang/rust/issues/66510

crates/update/src/main.rs
```diff
@@ -19,7 +19,13 @@ fn main() -> anyhow::Result<ExitCode> {
     if cmd == "spacetimedb-update" {
         spacetimedb_update_main()
     } else if cmd == "spacetime" {
-        proxy::run_cli(None, Some(argv0.as_os_str()), args.collect())
+        let args = args.collect::<Vec<_>>();
+        if args.first().is_some_and(|s| s == "version") {
+            // if the first arg is unambiguously `version`, go straight to `spacetime version`
+            spacetimedb_update_main()
```
bfops, line 25: Q: is this case necessary for some reason? Won't the CLI end up delegating back to `spacetimedb-update` anyway?

crates/cli/src/subcommands/version.rs
bfops: That makes sense. I think my question remains though: if the "hack" is good enough for windows, why not just use it on both platforms and have less surface area for edge cases?

crates/update/src/main.rs
coolreader18: Yes, but there's no perfect way for the cli to find the update binary again. If you're running via cargo run or something, it'll find `cli_bin_path` in `~/.local`, but then `spacetime version` will delegate to a different binary. It all sucks, but I feel this avoids the most bad edge cases.

crates/cli/src/subcommands/version.rs
coolreader18: I probably don't have an answer for that besides "that gives bad vibes". would you like me to change it?

crates/update/src/cli.rs
```diff
@@ -0,0 +1,138 @@
...
+
+#[derive(clap::Subcommand)]
+enum Subcommand {
+    Version(Version),
+    #[command(hide = true)]
+    Cli {
+        #[clap(allow_hyphen_values = true)]
+        args: Vec<OsString>,
+    },
+    SelfInstall {
```
bfops, line 73: I think this subcommand is new since my last review? What motivated this?

crates/update/build.rs
```diff
@@ -0,0 +1,5 @@
+#![allow(clippy::disallowed_macros)]
+fn main() {
+    let target = std::env::var("TARGET").unwrap();
+    println!("cargo::rustc-env=BUILD_TARGET={target}");
+}
```
bfops, line 5: I'm not very familiar with massaging the build process - what's this for?

crates/update/src/cli/link.rs
```diff
@@ -0,0 +1,38 @@
+use std::path::PathBuf;
+
+use anyhow::Context;
+use spacetimedb_paths::cli::BinDir;
+use spacetimedb_paths::SpacetimePaths;
+
+/// Set a local installation of SpacetimeDB as a custom version.
+#[derive(clap::Args)]
+pub(super) struct Link {
```
bfops, line 9: I think this is new since my last review - why did we add this in here? It does seem convenient but it's whole extra bit of surface area that I think we could add in a backwards-compatible way later. Was it part of some spec somewhere or did Tyler request it or something?

crates/update/build.rs
coolreader18: For some reason you can't access the compilation target without it setting it in a build script like this. This is for knowing what release artifact to download.

crates/update/src/cli/link.rs
coolreader18: This wasn't spec'd, and I should've left it as a separate commit, sorry, but this felt important for dog-food QA testing - letting us install `spacetime-update` globally but still use our local version for testing and stuff.

coolreader18: Also, it's left hidden for a reason - this isn't meant to be a stable cli surface.

crates/update/src/cli.rs
coolreader18: This is for the install script, that we can let spacetimedb-update handle the directory structure stuff instead of the bash or powershell needing to. This is how rustup does it.

crates/update/src/cli/link.rs
coolreader18: Now they're as a separate commit.

bfops (approved):
After the conversation elsewhere today, I'm approving this as-is. If we need to do further testing, changes, or code cleanups, we will have to do them in followup PRs. For now let's merge this and start getting it tested more widely.

PR #2127, opened 2025-01-15
Use Into<String> instead of ToString

RReverser:
# Description of Changes

As discussed on https://github.com/clockworklabs/SpacetimeDB/pull/2118, ToString is overly generic because it includes any type that implements Display.

Instead, we only want to accept string-like types (whether owned or not), which is better covered by Into<String>.

corresponding GitHub label.

2

# Testing

- [x] Ran CI tests, internal tests failed, fixed them.


crates/sats/src/hex.rs
```diff
@@ -36,6 +36,18 @@ impl<const N: usize> ops::Deref for HexString<N> {
...
+    fn from(hex: &HexString<N>) -> Self {
+        hex.as_str().into()
+    }
+}
+
+impl<const N: usize> From<HexString<N>> for String {
+    fn from(hex: HexString<N>) -> Self {
+        hex.as_str().into()
+    }
+}
```
gefjon, line 49: Why are these impls necessary? Where are they used?

gefjon:
How confident are you that our existing demos (I believe this is just `quickstart-chat`) do not depend on the previous behavior? What about external customers' code? What is the list of string-like types that implement `Into<String>` which you believe we want to allow? Are there any types which are `ToString` but not `Into<String>` which you think our customers may try to pass to these methods?

I can answer all of these questions, but it is not my job as reviewer to do so, it is your job as PR author. The answers should be recorded in your PR description for posterity.

crates/sats/src/hex.rs
RReverser: Internal tests pass hex strings to these APIs and fail without those impls.

RReverser:
> How confident are you that our existing demos (I believe this is just `quickstart-chat`) do not depend on the previous behavior?

I'm not, at all. I just offered a comment on the previous PR and you said you'd accept a followup PR changing those APIs. If there is no interest in changing them after all, feel free to close it.

cloutiertyler:
@gefjon I would like to merge this one because it narrows the API in what I feel is an appropriate way. It also brings the API more in line with what we have in the TypeScript SDK:

```ts
  withModuleName(nameOrAddress: string): DBConnectionBuilder<DBConnection> {
    this.#nameOrAddress = nameOrAddress;
    return this;
  }

  withToken(token?: string): DBConnectionBuilder<DBConnection> {
    this.#token = token;
    return this;
  }
```

and C# SDK:

https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/blob/bb22fcc762343a1a36e380c4ddf7b97a2938292c/src/SpacetimeDBClient.cs#L60


The Rust `quickstart-chat` does not rely on the old behavior: 

https://github.com/clockworklabs/SpacetimeDB/blob/249910b07907e817003738ee717f97d6bc9f2ce7/crates/sdk/examples/quickstart-chat/main.rs#L48

cloutiertyler (approved):
These changes are important and AFAICT good to merge IMO.

PR #2129, opened 2025-01-15
Run smoketests on windows in ci

coolreader18:
# Description of Changes

~~I don't intend to spend a ton of time on this, but maybe it'll just work~~ yay :)

Only some slight modifications to the smoketest code needed to get this working, and then wrestling with dotnet-sdk in the ci environment. For now this is running on GitHub's windows-latest runner, but we could switch to a self-hosted runner at some point if we wanted to.

1

# Testing

That's what we're doing here :)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,28 +14,8 @@ on:
 name: CI
 
 jobs:
-  find_ref:
-    name: Find Git ref
-    runs-on: spacetimedb-runner
-    outputs:
-      git_ref: ${{ steps.find_ref.outputs.git_ref }}
-    steps:
-      - name: Find Git ref
-        id: find_ref
-        env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: |
-          PR_NUMBER="${{ github.event.inputs.pr_number || null }}"
-          if test -n "${PR_NUMBER}"; then
-            GIT_REF="$( gh pr view --repo clockworklabs/SpacetimeDB $PR_NUMBER --json headRefName --jq .headRefName )"
-          else
-            GIT_REF="${{ github.ref }}"
-          fi
-          echo "git_ref=${GIT_REF}" >>"$GITHUB_OUTPUT"
-
   docker_smoketests:
     name: Smoketests
-    needs: find_ref
     strategy:
       matrix:
         include:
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -28,21 +31,42 @@ jobs:
...
         uses: actions/checkout@v4
         with:
-          ref: ${{ env.GIT_REF }}
+          ref: ${{ needs.find_ref.outputs.git_ref }}
       - uses: dsherret/rust-toolchain-file@v1
-      - uses: actions/setup-dotnet@v3
+      - uses: actions/setup-dotnet@v4
         with:
           global-json-file: modules/global.json
       - name: Start containers
```
bfops, line 55: So it's easier to understand when looking at the CI:
```suggestion
      - name: Start containers (Linux)
```

.github/workflows/ci.yml
```diff
@@ -28,21 +31,42 @@ jobs:
...
+        if: runner.os == 'Linux'
         run: docker compose up -d
+      - name: Start database
```
bfops, line 58: So that it's easier to understand when looking at the CI output:
```suggestion
      - name: Start database (Windows)
```

.github/workflows/ci.yml
```diff
@@ -14,11 +14,14 @@ on:
 name: CI
 
 jobs:
-  docker_smoketests:
-    name: Smoketests
+  find_ref:
+    name: Find Git ref
```
bfops, line 18: was there a reason we separated this out? I like having stuff factored, but it adds an extra line in the list of checks in the PR review, which makes me a bit sad:
image

.github/workflows/ci.yml
```diff
@@ -28,21 +31,42 @@ jobs:
...
+        if: runner.os == 'Linux'
         run: docker compose up -d
+      - name: Start database
+        if: runner.os == 'Windows'
+        run: |
+          cargo build -p spacetimedb-cli -p spacetimedb-standalone -p spacetimedb-update
+          Start-Process target/debug/spacetimedb-cli.exe start
+          cd modules
+          dotnet workload config --update-mode workload-set
+          dotnet workload update
```
bfops, line 65: dumb question - do you know why this dotnet-juggling is only needed on windows?

crates/update/src/proxy.rs
```diff
@@ -39,11 +39,10 @@ pub(crate) fn run_cli(paths: &SpacetimePaths, argv0: Option<&OsStr>, args: Vec<O
     }
     #[cfg(windows)]
     {
-        use std::os::windows::process::ExitCodeExt;
         let status = cmd
             .status()
             .with_context(|| format!("failed to run {}", cli_path.display()))?;
-        Ok(ExitCode::from_raw(status.code().unwrap_or(1) as u32))
+        Ok(ExitCode::from(status.code().unwrap_or(1) as u8))
```
bfops, line 45: was this causing an issue in some edge case? at first glance they seem 99% similar.

smoketests/__main__.py
```diff
@@ -80,10 +83,10 @@ def main():
         except FileNotFoundError:
             pass
         try:
-            os.symlink("spacetimedb-update", SPACETIME_BIN)
+            os.symlink("spacetimedb-update" + exe_suffix, SPACETIME_BIN)
         except OSError:
             import shutil
-            shutil.copyfile(SPACETIME_BIN.with_name("spacetimedb-update"), SPACETIME_BIN)
+            shutil.copyfile(SPACETIME_BIN.with_name("spacetimedb-update" + exe_suffix), SPACETIME_BIN)
```
bfops, line 89: could we maybe group together the `"spacetimedb-update" + exe_suffix` into a `spacetimedb_update_bin` variable or something so we don't have to do this in multiple places?

smoketests/tests/csharp_module.py
```diff
@@ -19,7 +19,7 @@ def test_build_csharp_module(self):
         try:
 
             run_cmd("dotnet", "nuget", "locals", "all", "--clear", cwd=bindings, capture_stderr=True)
-            run_cmd("dotnet", "workload", "install", "wasi-experimental", "--skip-manifest-update")
+            run_cmd("dotnet", "workload", "install", "wasi-experimental", "--skip-manifest-update", cwd=STDB_DIR / "modules")
```
bfops, line 22: was this only needed on windows or something? o_O

bfops (commented):
Looks pretty great to me :100: I commented on a few small things.

.github/workflows/ci.yml
coolreader18: I wasn't sure how to port it to Powershell, so I did this. I might be able to make it run in sh even on windows though

.github/workflows/ci.yml
coolreader18: No - I think GitHub's runner environment is just messed up.

crates/update/src/proxy.rs
coolreader18: That API wasn't actually stable, but the code wasn't being checked before.

smoketests/tests/csharp_module.py
coolreader18: something to do with dotnet sdk version selection; the windows-latest runner has v9 installed but our runner doesn't, I think.

.github/workflows/ci.yml
bfops: ahh it runs on a different runner. that makes sense. eh, probably fine for now then.

maybe there's a less dumb way to get the git ref.. I'm pretty sure I added this.

.github/workflows/ci.yml
bfops: Can you just add a comment saying as much? just a "we're not sure why this is needed on windows, but it made the thing work" type of thing so that future maintainers aren't terrified of changing it.

crates/update/src/proxy.rs
bfops: oh nice!

.github/workflows/ci.yml
bfops: honestly.. I'm not even 100% sure why we need to get the ref at all.. at first glance it looks like we're recreating the default behavior.

anyway probably fine to leave it for now.

bfops: Oh it's so we can trigger it manually from the `Actions` page

PR #2130, opened 2025-01-15
Query execution updates for cpu metrics

joshua-spacetime:
# Description of Changes

This patch is a prerequisite for computing query cpu metrics.

> Why is a new executor needed?

Tuple at a time iterators are not well suited to gathering per operator metrics.

None

2

Adding new query executors is relatively straightforward since they work off of the same physical plan. We can rely on our current test coverage to catch any issues.

# Testing

No new tests have been added. We already have correctness tests for the query and subscription code that are executor agnostic.


mamcx (approved):
LGTM

PR #2140, opened 2025-01-17
Track query and datastore cpu usage metrics

joshua-spacetime:
# Description of Changes

Tracks cpu usage metrics in both query and datastore. It does **not** track wasmtime fuel usage

None

3

The changes are relatively straightforward, just a lot of plumbing.

# Testing

An existing query test was updated to test the query metrics. New tests were added for the datastore.


joshua-spacetime:
@Centril the impact for these is negligible.

```
full-join        time:   [91.830 µs 92.193 µs 92.510 µs]
                 change: [-1.6552% -1.2958% -0.9618%] (p = 0.00 < 0.05)
```

However once we introduce "bytes scanned" for example, I fully expect the difference will no longer be negligible.

jsdt (commented):
These metrics look good, though I'm curious about how we will be able to add byte-based read stats.

crates/execution/src/pipelined.rs
```diff
@@ -29,34 +29,40 @@ impl From<ProjectPlan> for PipelinedProject {
 }
 
 impl PipelinedProject {
-    pub fn execute<'a, Tx: Datastore + DeltaStore>(
+    pub fn execute<'a, Tx: Datastore + DeltaStore, Metrics: ExecutionMetrics>(
         &self,
         tx: &'a Tx,
+        metrics: &mut Metrics,
```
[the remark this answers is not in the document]
joshua-spacetime: > Would it be easier if we attached the metrics to Tx?

For reducers, yes. For subscriptions, it depends. A single subscription update is evaluated across multiple threads using a single read-only transaction, which means the handle to that transaction cannot be mutable. We'd have to use atomics in that case.

crates/execution/src/pipelined.rs
```diff
@@ -29,34 +29,40 @@ impl From<ProjectPlan> for PipelinedProject {
 }
 
 impl PipelinedProject {
-    pub fn execute<'a, Tx: Datastore + DeltaStore>(
+    pub fn execute<'a, Tx: Datastore + DeltaStore, Metrics: ExecutionMetrics>(
```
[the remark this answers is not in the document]
joshua-spacetime: No, there's not. The semantics are the same.

crates/sats/src/size_of.rs
```diff
@@ -0,0 +1,121 @@
+use ethnum::{i256, u256};
+
+use crate::{algebraic_value::Packed, AlgebraicValue, ArrayValue, ProductValue, SumValue, F32, F64};
+
+pub trait SizeOf {
```
joshua-spacetime, line 5: This trait and impls borrowed from #2098

crates/client-api-messages/src/websocket.rs
```diff
@@ -67,7 +78,14 @@ pub trait WebsocketFormat: Sized {
     type Single: SpacetimeType + for<'de> Deserialize<'de> + Serialize + Debug + Clone;
 
     /// The type used for the encoding of a list of items.
-    type List: SpacetimeType + for<'de> Deserialize<'de> + Serialize + RowListLen + Debug + Clone + Default;
+    type List: SpacetimeType
+        + for<'de> Deserialize<'de>
+        + Serialize
+        + RowListLen
+        + ByteListLen
```
joshua-spacetime, line 85: @Centril I added a new type constraint to `WebsocketFormat` to be able to compute a subscription update's size in bytes. Can you review this change?

gefjon (approved):
Approving as code owner for websocket.rs . Have not checked the rest.

crates/sats/src/size_of.rs
gefjon: Whoever merges first will have the canonical version, and the other PR will be updated to use it.

crates/core/src/host/instance_env.rs
```diff
@@ -162,6 +174,10 @@ impl InstanceEnv {
 
         self.maybe_schedule_row(stdb, tx, table_id, row_ptr)?;
 
+        // Note, we update the metric for bytes written after the insert.
+        // This is to capture auto-inc columns.
+        tx.metrics.bytes_written += buffer.len();
```
joshua-spacetime, line 179: @jsdt here's where we're tracking bytes written on insert

jsdt: Oops, not sure how I missed that.

PR #2150, opened 2025-01-21
Relax dotnet version requirements

kim:
Verion 8.0.400 is not necessarily available in Linux package repositories, causing the `csharp_module.py` smoketest to fail (including on CI).

Until and unless a specific issue arises, allow any 8.0 minor version.

1

# Testing

Fixes a smoketest.

### DIFF
--- a/crates/bindings-csharp/global.json
+++ b/crates/bindings-csharp/global.json
@@ -1,0 +1,1 @@
+../../modules/global.json
--- a/crates/cli/src/subcommands/project/csharp/global._json
+++ b/crates/cli/src/subcommands/project/csharp/global._json
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "8.0",
+    "version": "8.0.0",
     "rollForward": "latestMinor"
   }
 }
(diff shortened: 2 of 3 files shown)


RReverser:
Looks like this now ends up selecting `9.0` on the CI, which is not compatible with our code (hence the failure).

Maybe try `"version": "8.0.0"` - I'm not sure it accepts the shorthand version.

kim:
> Maybe try "version": "8.0.0"

Well, this does not work either

kim:
Why is it behaving differently on Windows? Is it, like, looking in a different path or something?

RReverser:
Maybe because you introduced symlink, which might or might not work on Windows? At least it didn't behave differently on your original commit.

I tried using symlinks in original PR I think, but copy was in the end easier, especially since we shouldn't change this more often than once a year or so.

RReverser:
Could also be that setup-dotnet action doesn't respect rollForward, unlike .NET itself.

I'd suggest to:
1) Try to revert the symlink thingy.
2) Change version to `8.0.100` which was in fact the first version in `8.0` release (yeah, their versioning scheme is weird...).

kim:
Thanks @RReverser, I'll close this and instead endure the pain of installing dotnet in our various other CI environments.

PR #2151, opened 2025-01-21
[C#] Reuse buffers in table iterations

RReverser:
# Description of Changes

This is one of the branches I had for a while, but never submitted as a PR as it wasn't a priority. Still might be nice to have eventually.

Benchmark comparisons:

| Group                                                                                  | Speedup | Master Timing         | Reuse-C#-Buffers-Iter Timing |
|----------------------------------------------------------------------------------------|---------|-----------------------|------------------------------|
| special/db_game/csharp/circles/load=10                                                 | 1.50    | 1528.5±31.20ms        | 1015.9±13.92ms              |
| special/db_game/csharp/circles/load=100                                                | 1.48    | 1522.5±21.20ms        | 1027.5±15.85ms              |
| special/db_game/csharp/ia_loop/load=10                                                 | 1.83    | 242.8±5.78ms          | 132.3±3.28ms                |
| special/db_game/csharp/ia_loop/load=100                                                | 2.16    | 624.8±15.15ms         | 289.7±3.82ms                |
| stdb_module/csharp/mem/filter/string/index/load=2048/count=256                         | 1.09    | 865.0±29.83µs         | 790.6±29.09µs               |
| stdb_module/csharp/mem/filter/u64/index/load=2048/count=256                            | 1.16    | 525.3±9.86µs          | 452.1±6.28µs                |
| stdb_module/csharp/mem/insert_bulk/u32_u64_u64/btree_each_column/load=2048/count=256   | 1.07    | 2.0±0.12ms            | 1900.5±104.00µs             |
| stdb_module/csharp/mem/iterate/u32_u64_str/unique_0/count=256                          | 1.06    | 790.5±46.63µs         | 746.3±13.33µs               |
| stdb_module/csharp/mem/iterate/u32_u64_u64/unique_0/count=256                          | 1.10    | 475.0±26.96µs         | 432.1±11.19µs               |

That is, the difference is fairly marginal in microbenchmarks, but really shines in integration ones which are arguably more representative of real-world scenarios.

In those, the speedup ranges from 1.5x to 2.1x.

corresponding GitHub label.

1

# Testing

- [x] Ran SDK tests for C#.
- [x] Ran `dotnet test`.

### DIFF
--- a/crates/bindings-csharp/Runtime/Internal/Module.cs
+++ b/crates/bindings-csharp/Runtime/Internal/Module.cs
@@ -124,11 +124,8 @@ public static class Module
                 // The host will likely not trigger this branch (current host doesn't),
                 // but a module should be prepared for it.
                 case Errno.OK:
+                    ret.Check();
                     break;
-                case Errno.NO_SUCH_BYTES:
-                    throw new NoSuchBytesException();
-                default:
-                    throw new UnknownException(ret);
             }
         }
     }
--- a/crates/bindings-csharp/Runtime/Internal/Module.cs
+++ b/crates/bindings-csharp/Runtime/Internal/Module.cs
@@ -95,11 +95,11 @@ public static class Module
         moduleDef.RegisterTable(View.MakeTableDesc(typeRegistrar));
     }
 
-    private static byte[] Consume(this BytesSource source, ref byte[] buffer)
+    private static MemoryStream Consume(this BytesSource source, ref byte[] buffer)
     {
         if (source == BytesSource.INVALID)
         {
-            return [];
+            return new();
         }
         var written = 0U;
         while (true)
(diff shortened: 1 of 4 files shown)


crates/bindings-csharp/Runtime/Internal/Module.cs
```diff
@@ -1,311 +1,55 @@
...
-                    SourceName: table.SourceName,
-                    ProductTypeRef: table.ProductTypeRef,
-                    PrimaryKey: table.PrimaryKey,
-                    Indexes: table.Indexes,
-                    Constraints: table.Constraints,
-                    Sequences: table.Sequences,
-                    TableType: table.TableType,
-                    TableAccess: table.TableAccess,
-                    DefaultValues: defaults is null
-                        ? []
```
JasonAtClockwork, line 133: This looks like it has the potential for an infinite loop, should we keep a `default` just in case? I might be missing another out with just the changed code.

JasonAtClockwork (approved):
~~One small thing that might not need a change otherwise looks good.~~

Not sure how but my review went under the wrong PR, the liens listed aren't even correct to the code I reviewed in the comment.

Issue #2134, opened 2025-01-17
Update violates set semantics


kim:
Using the `update` method on a `#[primary_key]`'ed table inserts a new row even if the row is unchanged.
This prevents the database to restart from the commitlog:

```
internal error: failed to get or launch module host: Error inserting row ProductValue { elements: [U64(1), U64(1)] } during transaction 6 playback: TableError: Duplicate insertion of row RowPointer(r: 0, pi: 0, po: 0, so: 1) violates set semantics: Duplicate insertion of row RowPointer(r: 0, pi: 0, po: 0, so: 1) violates set semantics    
```

To reproduce, create a module like so:

```rust
use spacetimedb::{ReducerContext, Table};

#[spacetimedb::table(name = boeuf)]
pub struct Boeuf {
    #[primary_key]
    pub id: u64,
    pub n: u64,
}

#[spacetimedb::reducer]
pub fn add(ctx: &ReducerContext, id: u64, n: u64) {
    match ctx.db.boeuf().id().find(id) {
        Some(row) => ctx.db.boeuf().id().update(Boeuf { n, ..row }),
        None => ctx.db.boeuf().insert(Boeuf { id, n }),
    };
}
```

* Publish the module, then call the `add` reducer twice with the same arguments.
* Restart the server
* Access the database, e.g. by running an SQL statement or calling the reducer again
* Observe above mentioned error

Shubham8287:
+1, this is failing replication cluster restart smoketest

kim:
I'm marking this P0.

cloutiertyler:
@kim could you post the result of your bisect here even if you don't believe it was correct?

kim:
f136670420e13148b305c5deb4a416833dcaa7fb

kim:
The issue is not fully fixed, potentially due to special handling of connect/disconnect transactions.

imho, replaying should only warn, but never error on set semantic violations.

kim:
Steps to reproduce:

* run a standalone server
* publish the quickstart-chat example
* run the quickstart-chat example and exit it
  This will only create connect/disconnect transactions.
* stop the server
* remove the snapshot of the quickstart-chat replica
* start the server

Observe:

```
[...]

PR #2156, opened 2025-01-21
Fix unique index + `MutTxId::insert` not un-deleteting

Centril:
# Description of Changes

Fixes https://github.com/clockworklabs/SpacetimeDB/issues/2134.
The resolution was a fix to a silly bug where we did not un-delete in case of a unique index.
Also fixes an issue in `find_same_row_via_unique_index` where it didn't check row equality for the found pointer.

None.

1

# Testing

A regression test is added.

Also ran the repro in the issue:
```
w1$ spacetime start
w2$ spacetime publish basics -s local -c -y
w2$ spacetime call basics add 1 1 -s local
w2$ spacetime call basics add 1 1 -s local
w1$ ^C
w1$ spacetime start
w2$ spacetime sql basics 'select * from boeuf'
 id | n 
----+---
 1  | 1 
```

### DIFF
--- a/smoketests/tests/auto_migration.py
+++ b/smoketests/tests/auto_migration.py
@@ -48,7 +48,7 @@ const PERSON_VISIBLE: spacetimedb::Filter = spacetimedb::Filter::Sql("SELECT * F
 pub struct Book {
     isbn: String,
 }
- 
+
 #[spacetimedb::reducer]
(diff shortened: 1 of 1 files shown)


gefjon (approved):
Looks good to me. Thanks for the new test. Once you get CI happy, this is good to merge as far as I'm concerned.

gefjon (approved):

PR #2157, opened 2025-01-21
Fix autocomplete issues in bindings crate

coolreader18:
# Description of Changes

Fixes 2 specific issues:
- invalid syntax (namely, at least, an incomplete if block: `if foo`) in a reducer body would cause autocomplete to break.
  - This is fixed by upgrading the `proc-macro2` dependency.
- No completions for methods on unique columns.
  - This is fixed by fiddling around with the bounds on the `impl` block for `UniqueColumn`. Best I can tell is that this was hitting some poorly-handled edge case in chalk, related to evaluating bounds that are too dependent on each other (?)

1

# Testing

- [x] Both cases now provide autocomplete with rust-analyzer.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3539,9 +3539,9 @@ dependencies = [
 
 [[package]]
 name = "proc-macro2"
-version = "1.0.89"
+version = "1.0.93"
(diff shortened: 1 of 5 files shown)


gefjon (approved):
:pray: thank you so much for figuring this out.

PR #2158, opened 2025-01-22
CLI - Just-in-time login flow

bfops:
# Description of Changes

When a user goes to perform an action that requires logging in, we prompt them whether they want to log in.

If they decline to log in, we auth them directly with the server that they're trying to access.

**Note: We are still using the endpoint-based auth flow, and not doing any kind of scan for a key on the local machine!** I have never gotten an answer for what code I'm supposed to use to encrypt a token "manually" once I have the key in hand. IIRC when we last discussed, we said it was fine to keep using this endpoint, but if not then I would still need to know where that code lives / how to use it.

The following commands now have the potential to prompt the user to log in:
- `call`
- `delete`
- `describe`
- `rename`
- `energy`
- `list`
- `logs`
- `publish`
- `sql`
- `subscribe`

All of these commands have been updated to have a `--yes` parameter (if they didn't already). Somewhat unintuitively, this will actually answer "no" to the "log in with spacetimedb.com?" prompt so that it can obey its deeper mission of proceeding non-interactively.

The `login show` command has **not** been updated to this "just in time" login behavior. It seems unintuitive to prompt the user to log in when they're trying to `login show`. Instead, it tells them they're not logged in.

API breaking in the sense that it adds a confirmation prompt to many CLI actions.

2

# Testing

- [x] Logging out and then publishing prompts to log in
```
$ spacetimedb-cli logout
Saving config to /home/lead/.config/spacetime/cli.toml.

$ spacetimedb-cli publish -s local 
You are not logged in. Would you like to log in with spacetimedb.com? [y/N]
```

- [x] Staying logged in and then publishing does not prompt
```
$ spacetimedb-cli publish -s local 
    Checking spacetime-module v0.1.0 (/mnt/nutera/work/my_new_project)
    Finished `release` profile [optimized] target(s) in 0.14s
    Finished `release` profile [optimized] target(s) in 0.07s
For best performance install wasm-opt from https://github.com/WebAssembly/binaryen/releases.
Continuing with unoptimised module.
Build finished successfully.
Uploading to local => http://127.0.0.1:3000
Publishing module...
Created new database with identity: c2004ae02e7a187daeb619cc2e21739ac62ee18b9dec7576720585c1ea8a09ee
```

- [x] Logging out and then declining to log in causes a direct server log in, and successfully publishes
```
$ spacetimedb-cli logout
Saving config to /home/lead/.config/spacetime/cli.toml.

$ spacetimedb-cli publish -s local 
You are not logged in. Would you like to log in with spacetimedb.com? [y/N]N
Saving config to /home/lead/.config/spacetime/cli.toml.
    Finished `release` profile [optimized] target(s) in 0.07s
For best performance install wasm-opt from https://github.com/WebAssembly/binaryen/releases.
Continuing with unoptimised module.
Build finished successfully.
Uploading to local => http://127.0.0.1:3000
Publishing module...
Created new database with identity: c200fe686e34bd022c3d64bac8227b979ff1a714e57bc540fe106e69c96d7775
```

- [x] ...and that login does not work for other servers
```
$ spacetimedb-cli publish -s testnet
    Finished `release` profile [optimized] target(s) in 0.04s
For best performance install wasm-opt from https://github.com/WebAssembly/binaryen/releases.
Continuing with unoptimised module.
Build finished successfully.
You are about to publish to a non-local server: testnet.spacetimedb.com
Are you sure you want to proceed? [y/N]y
Uploading to testnet => https://testnet.spacetimedb.com
Publishing module...
Error: Identity c200115452f755e4bc722b7b32cb440b338e50899049b89b6d22e856cbfdb070 is not valid for server testnet.
Please log back in with `spacetime logout` and then `spacetime login`.

Caused by:
    TokenError(Error(InvalidSignature))
```

- [x] Logging out and then agreeing to log in opens a browser window, then successfully publishes
```
$ spacetimedb-cli logout
Saving config to /home/lead/.config/spacetime/cli.toml.

$ spacetimedb-cli publish -s local
You are not logged in. Would you like to log in with spacetimedb.com? [y/N]y
Opening https://spacetimedb.com/login/cli?token=eyJhbGciOiJIUzI1NiJ9.eyJleHAiOjE3Mzc2NTc5OTMwODQsInN1YiI6IjAxSkpBNzc3V0s0WlYwMFcwV0gwSEVXSDlGIn0.aqFlrhqGW0kbfqYr7ZW2vluCPu4zR6z5-UJeYRRlfiQ in your browser.
Waiting to hear response from the server...
Login successful!
Saving config to /home/lead/.config/spacetime/cli.toml.
Saving config to /home/lead/.config/spacetime/cli.toml.
    Finished `release` profile [optimized] target(s) in 0.05s
For best performance install wasm-opt from https://github.com/WebAssembly/binaryen/releases.
Continuing with unoptimised module.
Build finished successfully.
Uploading to local => http://127.0.0.1:3000
Publishing module...
Created new database with identity: c2004444682cf6d87407940e88592e0c6c51c8ca0d73f74a00578a567a52673e
```

- [x] ...and that login **does** work for other servers
```
$ spacetimedb-cli publish -s testnet
    Finished `release` profile [optimized] target(s) in 0.07s
For best performance install wasm-opt from https://github.com/WebAssembly/binaryen/releases.
Continuing with unoptimised module.
Build finished successfully.
You are about to publish to a non-local server: testnet.spacetimedb.com
Are you sure you want to proceed? [y/N]y
Uploading to testnet => https://testnet.spacetimedb.com
Publishing module...
Created new database with identity: c2002493cea901e0ac0888a4a68df2fcc99991700456b0961d5a594d38433cd3
```

- [x] `login show` works properly when logged in
```
$ spacetimedb-cli login show
You are logged in as c20058f5a5bbcbfd65110416110492e91d5ab64054640deb456bd5fa8a2ab9d5
```

- [x] `login show` gives an error when not logged in
```
$ spacetimedb-cli login show
You are not logged in. Run `spacetime login` to log in.
```

- [x] Existing automated smoketests still pass (so any CLI coverage in the smoketests still works as expected)


crates/cli/src/common_args.rs
```diff
@@ -27,5 +27,5 @@ pub fn yes() -> Arg {
         .long("yes")
         .short('y')
         .action(SetTrue)
-        .help("Assume \"yes\" as answer to all prompts and run non-interactively")
+        .help("Run non-interactively wherever possible. This will answer \"yes\" to almost all prompts.")
```
bfops, line 30: I tweaked this helptext since this flag really prioritizes "proceed as non-interactively as possible"; in the case of logging in on-demand, that means answering _no_ to "login with spacetimedb.com".

crates/cli/src/subcommands/login.rs
```diff
@@ -100,18 +105,28 @@ async fn spacetimedb_token_cached(config: &mut Config, host: &Url, direct_login:
...
-        let token = if direct_login {
-            spacetimedb_direct_login(host).await?
-        } else {
-            let session_token = web_login_cached(config, host).await?;
-            spacetimedb_login(host, &session_token).await?
-        };
-        config.set_spacetimedb_token(token.clone());
-        config.save();
-        Ok(token)
+        spacetimedb_login_force(config, host, direct_login).await
```
bfops, line 108: I factored this out into a separate function so that it could be reused from other files.

crates/cli/src/util.rs
```diff
@@ -264,8 +272,7 @@ Please log back in with `spacetime logout` and then `spacetime login`."
     })
 }
 
-pub fn decode_identity(config: &Config) -> anyhow::Result<String> {
-    let token = config.spacetimedb_token_or_error()?;
+pub fn decode_identity(token: &String) -> anyhow::Result<String> {
```
bfops, line 275: this function now takes a `token` directly, because otherwise it was going to need a bunch of extra params to pass to `get_login_token_or_log_in` and it made a bunch of stuff more messy.

crates/cli/src/subcommands/publish.rs
```diff
@@ -198,7 +199,8 @@ pub async fn exec(config: Config, args: &ArgMatches) -> Result<(), anyhow::Error
             ));
         }
         PublishResult::PermissionDenied { domain } => {
-            let identity = decode_identity(&config)?;
+            let token = get_login_token_or_log_in(&mut config, server, !force).await?;
```
bfops, line 202: this has a weird side effect: if we're not logged in, and we use `--anon-identity`, then a permission denied error will prompt the user to log in :sweat_smile:

bfops: this is an exacerbation of an issue we've had for a while (namely a permission denied error _used_ to just say "error: you're not logged in. please log in"). I feel like I should add an `if anon_identity` case here too?

crates/cli/src/subcommands/login.rs
```diff
@@ -79,13 +81,17 @@ async fn exec_subcommand(config: Config, cmd: &str, args: &ArgMatches) -> Result
 async fn exec_show(config: Config, args: &ArgMatches) -> Result<(), anyhow::Error> {
     let include_token = args.get_flag("token");
 
-    let identity = decode_identity(&config)?;
+    let token = if let Some(token) = config.spacetimedb_token() {
+        token
+    } else {
+        println!("You are not logged in. Run `spacetime login` to log in.");
+        return Ok(());
+    };
```
bfops, line 89: the old version of `decode_identity` was fetching the token from the config and then producing a similar error if there wasn't one, but now we've simplified `decode_identity` to take a token as a param instead of the config. This is the only call site where we had to do this manual compensation.

crates/cli/src/subcommands/login.rs
```diff
@@ -79,13 +81,17 @@ async fn exec_subcommand(config: Config, cmd: &str, args: &ArgMatches) -> Result
...
+
+    let identity = decode_identity(&token)?;
     println!("You are logged in as {}", identity);
 
     if include_token {
-        // We can `unwrap` because `decode_identity` fetches this too.
-        // TODO: maybe decode_identity should take token as a param.
-        let token = config.spacetimedb_token().unwrap();
```
bfops, line 88: a nice side effect of our `decode_identity` refactor is that we have the token in hand already at this point!

crates/cli/src/subcommands/publish.rs
```diff
@@ -198,7 +199,12 @@ pub async fn exec(config: Config, args: &ArgMatches) -> Result<(), anyhow::Error
             ));
         }
         PublishResult::PermissionDenied { domain } => {
-            let identity = decode_identity(&config)?;
+            if anon_identity {
+                anyhow::bail!("You need to be logged in to publish to {}", domain.tld());
+            }
```
bfops, line 204: We used to just rely on the error from `decode_identity` that said something like "you're not logged in, please log in". That somewhat obfuscated that we had a `PermissionDenied` error, so now we emit a more descriptive error.

(This was necessary because our new helper function `get_login_token_or_log_in` actually _logs the user in_, which is a very weird UX to have in an error case, especially when the user passed `--anon-identity`.)

crates/cli/src/common_args.rs
cloutiertyler: Perhaps we should call that example out in the help text?

bfops: I'm worried that it will be _more_ confusing if we make users think about it too hard, since I think "run non-interactively" is what they want deep down.

I guess it's probably fair to add it.

bfops: Updated

crates/cli/src/subcommands/publish.rs
```diff
@@ -198,7 +199,18 @@ pub async fn exec(config: Config, args: &ArgMatches) -> Result<(), anyhow::Error
...
+            let token = if anon_identity {
+                None
+            } else {
+                config.spacetimedb_token()
+            };
+            let token = if let Some(token) = token {
+                token
+            } else {
+                anyhow::bail!("You need to be logged in to publish to {}", domain.tld());
+            };
```
bfops, line 211: We used to just rely on the error from `decode_identity` that said something like "you're not logged in, please log in". It was somewhat-coincidentally a reasonable error message, in that the error message was _really_ saying "I can't decode your identity because you're not logged in", but what the user would infer was "I need to be logged in to publish here".

This also had a bug in the case where we passed `--anon-identity`, where it would use our _actual_ logged-in identity (if we had one) in the error message, even though that wasn't the identity we used for the request.

Now, 1) we explicitly check the `anon_identity` case, and 2) `decode_identity` no longer generates that coincidentally-correct-ish error message, so we produce an error message that somewhat benefits from the extra context of being a `PermissionDenied` case.

cloutiertyler:
>  Logging out and then declining to log in causes a direct server log in, and successfully publishes

We should probably print something explaining what happens. e.g.

```
Logging in with a server issued identity: 0xc200deadbeef
WARNING: This identity is only valid on the server that issued it.
```

bfops:
> > Logging out and then declining to log in causes a direct server log in, and successfully publishes
> 
> We should probably print something explaining what happens. e.g.
> 
> ```
> Logging in with a server issued identity: 0xc200deadbeef
> WARNING: This identity is only valid on the server that issued it.
> ```

Done. I just did a simple message though. Feel free to iterate on it more (inline in the diff).

crates/cli/src/common_args.rs
cloutiertyler: Yeah that's a fair point, but probably good to call out.

crates/cli/src/subcommands/describe.rs
```diff
@@ -23,14 +23,16 @@ pub fn cli() -> clap::Command {
...
         .after_help("Run `spacetime help describe` for more detailed information.\n")
 }
 
-pub async fn exec(config: Config, args: &ArgMatches) -> Result<(), anyhow::Error> {
+pub async fn exec(mut config: Config, args: &ArgMatches) -> Result<(), anyhow::Error> {
     let database = args.get_one::<String>("database").unwrap();
     let entity_name = args.get_one::<String>("entity_name");
     let entity_type = args.get_one::<String>("entity_type");
     let server = args.get_one::<String>("server").map(|s| s.as_ref());
+    let force = args.get_flag("force");
```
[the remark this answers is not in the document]
bfops: In the definition of `common_args::yes` - https://github.com/clockworklabs/SpacetimeDB/blob/master/crates/cli/src/common_args.rs#L25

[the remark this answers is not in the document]
bfops: (at some point we should probably rename the internal flag name to `yes` as well, but I didn't do it in the original PR that added `common_args::yes()`, to minimize the diff / chance for oversights)

crates/cli/src/subcommands/publish.rs
cloutiertyler: Let's just update the error message to say "Hey you need to be logged in as the owner of foobar to publish to foobar."

cloutiertyler (approved):
This looks really good to me. Thanks for addressing the feedback and also for the detailed testing section and comments. It made review very straightforward.

bfops:
I'm going to merge this given how much trouble I'm having triggering the `PermissionDenied` or `Unauthorized` flows for testing. We can iterate on these cases later if something's not working properly.

PR #2161, opened 2025-01-22
Don't create indexes during bootstrapping; wait until after replay

gefjon:
# Description of Changes

This PR fixes a bug where replaying databases without a snapshot of TX 0, i.e. using the entire commitlog,
would fail when replaying deletes on `st_client`.

This is because `st_client` was in the unique (*) position of:
- Being a system table,
- with a unique index,
- which is mutated after bootstrapping.

(*) Other system tables meet all three criteria in edge cases with schema-changing publishes, but only `st_client` meets them in everyday testing.

In this case, when replaying the entire commitlog, we create the indexes on the system tables during bootstrapping before replay. (I theorize, but have not checked,
that when restoring from a snapshot we do not create these indexes so eagerly, instead waiting for after replaying the commitlog suffix to create them.)

However, we cannot update or use indexes during bootstrapping, at least currently, because doing so causes unique constraint violations. Prior to this commit, we had a bunch of complicated code which avoided updating indexes during replay.

PR #2092, aka commit f667f65b159f0ffbd00a3bdb24afee79c0084070, introduced a case where we would use an index during replay: when replaying a delete for a table which had an index, and therefore had no pointer map.
Prior to that commit, we would never use an index during replay, even if it existed.

This all came together to mean that `st_client` had an empty index, and no pointer map, and so calls to `replay_delete_by_rel` on that table would always fail. I theorize that no one noticed this sooner
because it did not manifest when restoring from a snapshot, and there is always a snapshot of TX 0 unless you manually delete it, as Kim did during his testing.

The fix is simple: we do not construct indexes on the system tables during bootstrapping before replaying the commitlog, instead waiting until after replay like we do for user-defined indexes. This allows us to rip out all the aforementioned fiddly code for dodging indexes during replay,
since now there will not be any indexes and therefore we will never update them.

This PR also includes some new testing infra in `relational_db.rs`, and exports some test helpers from the `commitlog` crate under its `test` feature.

N/a

4: I don't feel like I understand our bootstrapping process terribly well, it is quite fiddly, and I cannot explain why we were previously constructing system table indexes before replay. It may have been important in some way I do not realize.

# Testing

- [x] Replayed a commitlog containing a conn/disconn pair after deleting snapshot 0, and saw that it successfully reconstructed the (or at least an) in-memory state.
  - I did not do any more work to verify that the reconstructed state was correct.
- [x] Did the same thing with a commitlog with snapshot 0 still intact, to make sure I didn't somehow break that.
- [x] Wrote a cool new regression test, `replay_delete_from_st_client`, which fails on master but passes on this branch.

### DIFF
--- a/crates/commitlog/src/tests/helpers.rs
+++ b/crates/commitlog/src/tests/helpers.rs
@@ -38,6 +38,17 @@ where
     total_txs
 }
 
+pub fn fill_log_with<R, T>(log: &mut commitlog::Generic<R, T>, txes: impl IntoIterator<Item = T>)
+where
+    R: Repo,
+    T: Debug + Encode,
+{
+    for tx in txes {
+        log.append(tx).unwrap();
+    }
+    log.commit().unwrap();
+}
+
 pub fn enable_logging() {
     let _ = env_logger::builder()
         .filter_level(log::LevelFilter::Trace)
--- a/crates/commitlog/src/tests/helpers.rs
+++ b/crates/commitlog/src/tests/helpers.rs
@@ -38,6 +38,9 @@ where
     total_txs
 }
 
+/// Put the `txes` into a single commit within `log`.
+///
+/// Panics if `txes` does not fit within a single commit within `log`.
 pub fn fill_log_with<R, T>(log: &mut commitlog::Generic<R, T>, txes: impl IntoIterator<Item = T>)
 where
     R: Repo,
--- a/crates/commitlog/src/tests/helpers.rs
+++ b/crates/commitlog/src/tests/helpers.rs
@@ -38,9 +38,9 @@ where
     total_txs
 }
 
-/// Put the `txes` into a single commit within `log`.
+/// Put the `txes` into `log`.
 ///
-/// Panics if `txes` does not fit within a single commit within `log`.
+/// Each TX from `txes` will be placed in its own commit within `log`.
 pub fn fill_log_with<R, T>(log: &mut commitlog::Generic<R, T>, txes: impl IntoIterator<Item = T>)
 where
     R: Repo,
@@ -48,8 +48,8 @@ where
 {
     for tx in txes {
         log.append(tx).unwrap();
+        log.commit().unwrap();
     }
-    log.commit().unwrap();
 }
 
 pub fn enable_logging() {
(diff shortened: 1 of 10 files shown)


cloutiertyler:
> However, we cannot update or use indexes during bootstrapping, at least currently, because doing so causes unique constraint violations.
> ...
> This all came together to mean that st_client had an empty index, and no pointer map, and so calls to replay_delete_by_rel on that table would always fail.

I don't understand why that follows from this, could you explain? In particular how is the double delete thing we talked about relevant?

gefjon:
Mazdak is correct;  there is only one delete, for a row which is present, but is impossible to find because its table has no pointer map.

gefjon:
More specifically, as of  #2092 , `delete_equal_row` (called by `replay_delete_by_rel`) has two branches: one for tables with a unique constraint, and the other for tables with a pointer map. (Prior to that PR, it always used the pointer map,  as all tables had pointer maps.) The bug is that `delete_equal_row` during replay on `st_client` would see a unique index and use it, but that unique index was unpopulated, so the `delete_equal_row` call would fail. This appears to the datastore like deleting of a not-present row, even though the row is present in the table. Doing a full scan followed by a delete-by-pointer would succeed, but we don't do that because replay is slow enough already even without doing a bunch of full scans.

kim:
> why we were previously constructing system table indexes before replay

As far as I can tell, this has been there since inception of the locking datastore and could have been removed with the introduction of `rebuild_state_after_replay` in d8576bfcc94c0b79f2756703e4e1ad35b658b448

kim (approved):

kim:
As a follow-up, we could also consider to remove the code which lets empty connect/disconnect transactions be committed to the commitlog, because there are no actually empty transactions?

crates/commitlog/src/tests/helpers.rs
```diff
@@ -38,6 +38,17 @@ where
     total_txs
 }
 
+pub fn fill_log_with<R, T>(log: &mut commitlog::Generic<R, T>, txes: impl IntoIterator<Item = T>)
```
kim, line 41: I think this should at least have commentary which explains that `txes` must fit into a single commit as per the options of the `commitlog::Generic` being passed in.
And perhaps that it will create a single commit.

PR #2163, opened 2025-01-22
Fix ScheduleAt in C#

kazimuth:
# Description of Changes

Fix the implementation of C# timestamps. While addressing issues on blackholio I noticed these were incorrect. The specific issue was a missed subtraction of DateTimeOffset.UnixEpoch when converting from DateTimeOffset to milliseconds.

A fix is needed if the C# server version of Blackholio is to work.

https://github.com/clockworklabs/SpacetimeDB/pull/1836 is a larger change that will rework Timestamp in C#, that will want to be rebased onto this.

Added APIs in the C# SDK.

0

# Testing

Added a unit test.

### DIFF
--- a/crates/bindings-csharp/BSATN.Runtime/Builtins.cs
+++ b/crates/bindings-csharp/BSATN.Runtime/Builtins.cs
@@ -257,6 +257,31 @@ public abstract partial record ScheduleAt
 
     public static implicit operator ScheduleAt(DateTimeOffset time) => new Time(time);
 
+    /// <summary>
+    /// There are 10 C# Timestamp "Ticks" per microsecond.
+    /// </summary>
+    public static readonly ulong TicksPerMicrosecond = 10;
+
+    public static ulong ToMicroseconds(TimeSpan interval)
+    {
+        return (ulong)interval.Ticks / TicksPerMicrosecond;
+    }
+
+    public static TimeSpan TimeSpanFromMicroseconds(ulong intervalMicros)
+    {
+        return TimeSpan.FromTicks((long)(TicksPerMicrosecond * intervalMicros));
+    }
+
+    public static ulong ToMicrosecondsSinceUnixEpoch(DateTimeOffset time)
+    {
+        return ToMicroseconds(time - DateTimeOffset.UnixEpoch);
+    }
+
+    public static DateTimeOffset DateTimeOffsetFromMicrosSinceUnixEpoch(ulong microsSinceUnixEpoch)
+    {
+        return DateTimeOffset.UnixEpoch + TimeSpanFromMicroseconds(microsSinceUnixEpoch);
+    }
+
     public readonly partial struct BSATN : IReadWrite<ScheduleAt>
     {
         [SpacetimeDB.Type]
@@ -289,6 +314,10 @@ public abstract partial record ScheduleAt
         public AlgebraicType GetAlgebraicType(ITypeRegistrar registrar) =>
             // Constructing a custom one instead of ScheduleAtRepr.GetAlgebraicType()
             // to avoid leaking the internal *Repr wrappers in generated SATS.
+            // We are leveraging the fact that single-element structs are byte-compatible with their elements
+            // when parsing BSATN.
+            // TODO: this might break when working with other formats like JSON, but this is all going to be rewritten
+            // anyway with Phoebe's Timestamp PR.
             new AlgebraicType.Sum(
                 [
                     new("Interval", new AlgebraicType.U64(default)),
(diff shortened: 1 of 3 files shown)


crates/bindings-csharp/BSATN.Runtime/Repr.cs
```diff
@@ -1,27 +0,0 @@
...
-using System.Runtime.InteropServices;
-
-// We store time information in microseconds in internal usages.
-//
-// These utils allow to encode it as such in FFI and BSATN contexts
-// and convert to standard C# types.
-
-[StructLayout(LayoutKind.Sequential)] // we should be able to use it in FFI
-[SpacetimeDB.Type] // we should be able to encode it to BSATN too
-public partial struct DateTimeOffsetRepr(DateTimeOffset time)
```
RReverser, line 12: It's a matter of taste, of course, but I kind of liked the intermediate representation as a separate type you can convert from, especially since `struct`s are cheap / zero-copy wrappers.

crates/bindings-csharp/BSATN.Runtime/Builtins.cs
```diff
@@ -257,19 +256,49 @@ public sealed record Time(DateTimeOffset Time_) : ScheduleAt;
 
     public static implicit operator ScheduleAt(DateTimeOffset time) => new Time(time);
 
+    /// <summary>
+    /// There are 10 C# Timestamp "Ticks" per microsecond.
+    /// </summary>
+    public static readonly ulong TicksPerMicrosecond = 10;
```
RReverser, line 262: Shame that we can't use [TimeSpan.TicksPerMicrosecond](https://learn.microsoft.com/en-us/dotnet/api/system.timespan.tickspermicrosecond?view=net-9.0#system-timespan-tickspermicrosecond) in .NET Standard...

crates/bindings-csharp/BSATN.Runtime/Builtins.cs
```diff
@@ -257,19 +256,49 @@ public sealed record Time(DateTimeOffset Time_) : ScheduleAt;
...
+    }
+
+    public static ulong ToMicrosecondsSinceUnixEpoch(DateTimeOffset time)
+    {
+        return ToMicroseconds(time - DateTimeOffset.UnixEpoch);
+    }
+
+    public static DateTimeOffset DateTimeOffsetFromMicrosSinceUnixEpoch(ulong microsSinceUnixEpoch)
+    {
+        return DateTimeOffset.UnixEpoch.AddTicks((long)(TicksPerMicrosecond * microsSinceUnixEpoch));
```
RReverser, line 282: Nit: might be slightly cleaner to reuse the `TimeSpan` conversion.

```suggestion
        return DateTimeOffset.UnixEpoch + TimeSpanFromMicroseconds(microsSinceUnixEpoch);
```

crates/bindings-csharp/BSATN.Runtime/Repr.cs
kazimuth: I had found a server-side bsatn parsing issue in blackholio and was wondering if this was related. Turned out not to be, the actual fix is here: https://github.com/clockworklabs/SpacetimeDB/pull/2164/files#diff-8b70d4eeef215ef2f1e1a16e36b20ae9a943e1a3aeb823c260fc056842b570a8R158 . I can revert this. I need to implement a custom Timestamp type per phoebe's proposal anyway which this can be used for.

RReverser: It's fine either way. As you said, it's going to change soon anyway.

kazimuth: Ah, and you'd already fixed that, I missed it. Welp!

crates/bindings-csharp/BSATN.Runtime/Builtins.cs
kazimuth: I'm inclined to avoid polyfills for now since they're trickier to test; this is going to get rewritten in https://github.com/clockworklabs/SpacetimeDB/pull/1836 anyway. But yeah, would be nice to just have these available. Unity...

crates/bindings-csharp/BSATN.Runtime/Repr.cs
```diff
@@ -11,17 +11,17 @@ namespace SpacetimeDB.Internal;
 [SpacetimeDB.Type] // we should be able to encode it to BSATN too
 public partial struct DateTimeOffsetRepr(DateTimeOffset time)
 {
-    public ulong MicrosecondsSinceEpoch = (ulong)time.Ticks / 10;
```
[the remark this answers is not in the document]
RReverser: Yeah, saw it, that's definitely a good catch. Always wishing we had more tests.

> presumably the birth of Christ.

Good guess. The docs:

> The value of the [Ticks](https://learn.microsoft.com/en-us/dotnet/api/system.datetimeoffset.ticks?view=net-9.0) property represents the number of 100-nanosecond intervals that have elapsed since 12:00:00 midnight on January 1, 0001

RReverser (approved):

gefjon (approved):

PR #2164, opened 2025-01-22
WIP: Start SDK tests for delete_all_by_eq_bsatn

kazimuth:
# Description of Changes

This API used to be unimplemented and the SDK tests did not exercise it. Now it is implemented but while playing with blackholio I noticed the C# implementation was wrong.

For now I am going to fix blackholio by avoiding use of this API for now, but we should also correct the implementation and test it.

None 

0

# Testing

Working on adding tests. If someone is more familiar with the SDK tests I would appreciate help amending them.


kazimuth:
This is almost entirely redundant with https://github.com/clockworklabs/SpacetimeDB/pull/2146 (merged just before and I didn't see) except that I wanna add a few more tests to that.

cloutiertyler (approved):
Just tests mainly, so I'd like to merge this.

PR #2166, opened 2025-01-23
Fix sending of subscription errors.

jsdt:
# Description of Changes

This is pulling out the server side changes that were originally in https://github.com/clockworklabs/SpacetimeDB/pull/2111.

Previously the code was not sending subscription errors for invalid queries (because we were treating those errors as generic internal errors). This fixes that. It also adds the query id to those subscription error messages, so that the clients aren't forced to track the request ids.

1

# Testing

This change is covered with some end-to-end tests in https://github.com/clockworklabs/SpacetimeDB/pull/2111.

Adding unit tests in the future would be good.


gefjon (approved):
I approved these in the original PR; I'll approve em here too.

PR #2169, opened 2025-01-24
Rust SDK: remove `anyhow`, use `thiserror` w/ structured error types

gefjon:
# Description of Changes

Our use of `anyhow` in the Rust SDK was originally due to my being lazy. I figured I'd kick the can on writing actual error types as long as possible,  and it turns out now's the time that's no longer possible, because while working on another ticket I needed to `clone` an error, and `anyhow::Error: !Clone`.

This commit removes the Rust SDK's use of `anyhow`, replacing it with a `thiserror`-annotated `enum Error`.

This resulted in essentially simple mechanical changes to  a lot of the SDK code, and also to the codegen, resulting in a truly enormous diff.

## Advice for reviewers

Start in `crates/sdk/src/error.rs` and see if the error enum makes sense. Then look through the rest of the SDK `src` to verify that I'm constructing them sensibly. Then check `crates/cli/src/subcommands/generate/rust.rs`. Do not at any point look at the generated code inside `crates/sdk/tests`, as it will be uninteresting and verbose.

Yep! The errors returned from the SDK functions are now of a different type.

1: large volume of mechanical changes. It was essentially impossible to depend on the actual nature of the emitted errors before, as `anyhow` isn't really useful for that, so we're effectively adding strictly more information in a way that should not break any existing client code.

# Testing

- [x] I believe our existing automated tests are sufficient.


gefjon:
Based  on the lints failure, it seems we need to get #2111 merged, then get this updated to latest master.

crates/sdk/src/error.rs
```diff
@@ -0,0 +1,96 @@
+use std::sync::Arc;
+
+use thiserror::Error;
+
+#[derive(Error, Debug, Clone)]
+#[non_exhaustive]
+pub enum Error {
+    #[error("Disconnected normally following a call to `DbContext::disconnect`")]
+    Disconnected,
```
jsdt, line 9: It looks like there are 3 errors that effectively mean we are disconnected (`Disconnected`, `AlreadyDisconnected`, and `Ws`). Can we combine these three? Maybe `Disconnected` with an optional underlying error?

crates/sdk/src/error.rs
```diff
@@ -0,0 +1,96 @@
...
+        Self {
+            cause: Some(Arc::new(cause)),
+            ..self
+        }
+    }
+
+    #[doc(hidden)]
+    /// Called by codegen. Not part of this library's stable API.
+    pub fn failed_parse(ty: &'static str, container: &'static str) -> Self {
+        Self::new(format!("Failed to parse {ty} from {container}"))
```
jsdt, line 56: Does it make sense to add a suggestion in the error message to regen the client API? Or is this usually due to actual bugs.

jsdt (approved):
I'd really prefer to combine all the flavors of disconnection errors, but I'm going to approve.

crates/sdk/src/error.rs
gefjon: I agree that `Disconnected` and `AlreadyDisconnected` are redundant, and am happy to combine them. I think I should restructure `Ws` so that we can distinguish between failed-to-connect versus erroneous-disconnect-after-the-fact, since I believe it can occur in both of those cases. I am unsure whether erroneous-disconnect should be the same error as disconnected-on-purpose; my first intuition was no, but if you think that's wrong, I'm happy to combine the two categories, s.t. `Disconnected { source: None }` means normal disconnect and  `Disconnected { source: Some(err) }` means abnormal disconnect.

crates/sdk/src/error.rs
gefjon: That is an excellent idea, and I will do so.

gefjon:
Test failure is unrelated and is fixed by #2199 .

PR #2173, opened 2025-01-25
Adds improvements & new crates to `publish-crates.sh`

cloutiertyler:
# Description of Changes

This PR modifies `publish-crates.sh`:

1. to take a new argument `--allow-dirty` which is forwarded to `crates publish`
2. to not fail if a crate has already been published before
3. to add `tyler@clockworklabs.io` and `zeke@clockworklabs.io` to be an owner on all newly published crates

NOTE! I recommend we investigate https://github.com/crate-ci/cargo-release in the future rather than to improve this script ad infinitum.

None

1, it can only break this script, but nothing else.

# Testing

I have not tested the new improvements to the script since it's quite hard to test without bumping the versions on things. The risk however is very low and I figure we can fix issues as they come.

### DIFF
--- a/crates/cli/src/subcommands/project/rust/Cargo._toml
+++ b/crates/cli/src/subcommands/project/rust/Cargo._toml
@@ -9,5 +9,5 @@ edition = "2021"
 crate-type = ["cdylib"]
 
 [dependencies]
-spacetimedb = "1.0.0-rc3.1"
+spacetimedb = "1.0.0-rc3"
 log = "0.4"
--- a/crates/cli/src/subcommands/project/rust/Cargo._toml
+++ b/crates/cli/src/subcommands/project/rust/Cargo._toml
@@ -9,5 +9,5 @@ edition = "2021"
 crate-type = ["cdylib"]
 
 [dependencies]
-spacetimedb = "1.0.0-rc3"
+spacetimedb = "1.0.0-rc3.1"
 log = "0.4"
--- a/crates/schema/Cargo.toml
+++ b/crates/schema/Cargo.toml
@@ -33,8 +33,8 @@ enum-map.workspace = true
 [dev-dependencies]
 spacetimedb-lib = { workspace = true, features = ["test"] }
 # these are circular dependencies, but only in tests, so it's fine
-spacetimedb-testing = { path = "../testing" }
-spacetimedb-cli.workspace = true
+# spacetimedb-testing = { path = "../testing" }
+# spacetimedb-cli.workspace = true
 
 proptest.workspace = true
 serial_test.workspace = true
--- a/crates/schema/Cargo.toml
+++ b/crates/schema/Cargo.toml
@@ -33,8 +33,8 @@ enum-map.workspace = true
 [dev-dependencies]
 spacetimedb-lib = { workspace = true, features = ["test"] }
 # these are circular dependencies, but only in tests, so it's fine
-# spacetimedb-testing = { path = "../testing" }
-# spacetimedb-cli.workspace = true
+spacetimedb-testing = { path = "../testing" }
+spacetimedb-cli.workspace = true
 
 proptest.workspace = true
 serial_test.workspace = true
--- a/tools/publish-crates.sh
+++ b/tools/publish-crates.sh
@@ -4,41 +4,61 @@ set -euo pipefail
 cd "$(dirname "$0")/.."
 
 DRY_RUN=0
+ALLOW_DIRTY=0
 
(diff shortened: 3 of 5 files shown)


crates/cli/src/subcommands/project/rust/Cargo._toml
```diff
@@ -9,5 +9,5 @@ edition = "2021"
 crate-type = ["cdylib"]
 
 [dependencies]
-spacetimedb = "1.0.0-rc3"
+spacetimedb = "1.0.0-rc3.1"
```
bfops, line 12: Given that these changes have been released, should we be bumping to a later version? (Or, if API/ABI-breaking changes have merged on master since `rc3` was released, should we be bumping to `rc4` here?)

tools/publish-crates.sh
```diff
@@ -4,41 +4,78 @@ set -euo pipefail
...
-	fi
+for crate in "${CRATES[@]}"; do
+    cd "${BASEDIR}/crates/${crate}"
+
+    PUBLISH_CMD="cargo publish"
+    [[ $DRY_RUN -eq 1 ]] && PUBLISH_CMD+=" --dry-run"
+    [[ $ALLOW_DIRTY -eq 1 ]] && PUBLISH_CMD+=" --allow-dirty"
+
+    # Check if crate exists on crates.io
+    if ! cargo search "$crate" --limit 1 | grep -q "^$crate ="; then
```
bfops, line 53: I'm not sure that this `--limit 1` is helping us if we're already `grep`ing. Does `cargo search` guarantee that an exact string match would appear first?

Here's my canonical "cargo search is weird" example:
```
$ cargo search asdfasdf
lisho = "0.1.5"                 # A simple personal link shortener with no external dependencies in under 200 lines of Rust.
okapi-response-mac = "0.1.1"    # A macro for creating response code based on responses structs rather than typing and implementi…
umami_metrics = "0.1.0"         # A rust library to send metrics data to the Umami metrics platform
```

```suggestion
    if ! cargo search "$crate" | grep -q "^$crate ="; then
```

tools/publish-crates.sh
```diff
@@ -4,41 +4,78 @@ set -euo pipefail
...
+        IS_NEW_CRATE=1
+        echo "INFO: Detected $crate as a new crate on crates.io!"
+    else
+        IS_NEW_CRATE=0
+    fi
+
+    echo "Publishing crate: $crate with command: $PUBLISH_CMD"
+    if ! OUTPUT=$($PUBLISH_CMD 2>&1); then
+        if echo "$OUTPUT" | grep -q "crate version .* is already uploaded"; then
+            echo "WARNING: Crate $crate version is already published. Skipping..."
```
bfops, line 63: as I've mentioned before, I think we need to make this configurable with a command-line option. This could be a genuine failure if e.g. we failed to bump package versions entirely.

Can we add e.g. `--skip-already-published`?

bfops (changes requested):
Thank you for doing this! This looks better than what I got from copilot in my spike of making these changes.

That said, I think it needs another quick pass.

Ideally these would also be separate changes since they're logically independent:
- Bump crate versions
- Add new crates to publish script
- Add new functionality to publish script

That said, they're all small and self-contained so I think it's okay to leave these in the same PR this time (as long as you're fine with all of them waiting to merge until all of them are approved).

crates/schema/Cargo.toml
```diff
@@ -33,8 +33,8 @@ enum-map.workspace = true
 [dev-dependencies]
 spacetimedb-lib = { workspace = true, features = ["test"] }
 # these are circular dependencies, but only in tests, so it's fine
-spacetimedb-testing = { path = "../testing" }
-spacetimedb-cli.workspace = true
+# spacetimedb-testing = { path = "../testing" }
+# spacetimedb-cli.workspace = true
```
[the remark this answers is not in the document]
cloutiertyler: You are correct.

tools/publish-crates.sh
bfops: (I can make this change if you're busy, it's easy now that you've refactored our argument parsing)

bfops:
Created https://github.com/clockworklabs/SpacetimeDB/issues/2176 to track us investigating the `cargo-release` tool.

bfops:
I moved the version bumps into https://github.com/clockworklabs/SpacetimeDB/pull/2177.

bfops (approved):
This version LGTM

cloutiertyler:
Thank you for making the changes, this LGTM.

PR #2177, opened 2025-01-27
Bump version to `1.0.0-rc4`

bfops:
# Description of Changes

We have released `v1.0.0-rc3`. This PR bumps us to the next RC version. (Hotfixes will happen on cherrypicked branches, and will therefore not include this PR).

No, just a version bump.

1

# Testing

None. This just bumps the version.


cloutiertyler (approved):
This LGTM as well, given that there's no need to actually bump to rc3.1

PR #2181, opened 2025-01-27
Remove janky schema from http API

coolreader18:
# Description of Changes

Get rid of the weird stuff in `describe` and just return a RawModuleDef.

2 - changes api a decent amount but only to return an already well-established format.

# Testing

- [x] Smoketests cover this wrt to CLI changes.

### DIFF
--- a/crates/cli/src/subcommands/call.rs
+++ b/crates/cli/src/subcommands/call.rs
@@ -1,18 +1,19 @@
+use crate::api::ClientApi;
 use crate::common_args;
 use crate::config::Config;
 use crate::edit_distance::{edit_distance, find_best_match_for_name};
-use crate::util::{self, UNSTABLE_WARNING};
-use crate::util::{add_auth_header_opt, database_identity, get_auth_header};
+use crate::util::UNSTABLE_WARNING;
 use anyhow::{bail, Context, Error};
 use clap::{Arg, ArgMatches};
-use itertools::Either;
-use serde_json::Value;
+use convert_case::{Case, Casing};
+use itertools::Itertools;
 use spacetimedb::Identity;
-use spacetimedb_lib::de::serde::deserialize_from;
-use spacetimedb_lib::sats::{AlgebraicType, AlgebraicTypeRef, Typespace};
+use spacetimedb_lib::sats::{self, AlgebraicType, Typespace};
 use spacetimedb_lib::ProductTypeElement;
+use spacetimedb_schema::def::{ModuleDef, ReducerDef};
 use std::fmt::Write;
-use std::iter;
+
+use super::sql::parse_req;
 
 pub fn cli() -> clap::Command {
     clap::Command::new("call")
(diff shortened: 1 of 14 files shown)


coolreader18:
I semi-rewrote the `invalid_arguments` stuff because it was interacting with the api through `serde_json::Value`s instead of ModuleDef, and it was much easier to just make it use ModuleDef than to keep the weird stuff.

bfops:
Question about testing - Did this PR change the expected output format of `spacetime describe` at all? The smoketests just seem to test that the CLI exits successfully, but I don't see, like, snapshot tests or anything. I don't specifically think that we need them, just trying to understand the expected impact here.

coolreader18:
I don't think there's really any use case for `spacetime describe` atm besides some sort of scripting. Ideally I think its default behavior would be to show a nice human-readable description of the module or the selected item. Maybe its current behavior should be put behind a `--json` flag, or it should just be removed entirely until it has a better spec?

bfops:
> I don't think there's really any use case for `spacetime describe` atm besides some sort of scripting. Ideally I think its default behavior would be to show a nice human-readable description of the module or the selected item. Maybe its current behavior should be put behind a `--json` flag, or it should just be removed entirely until it has a better spec?

Yeah I don't think we particularly need to preserve the old output, but I do think we want to make sure that `describe` and `call` still have reasonable-looking outputs.

I do think we've told some folks with larger DBs to inspect them via `describe`, so I don't think we should remove it entirely. We're going to mark it as "unstable" for 1.0 though, so we can totally keep changing it / breaking it.

crates/cli/src/subcommands/call.rs
```diff
@@ -37,50 +38,36 @@ pub fn cli() -> clap::Command {
         .after_help("Run `spacetime help call` for more detailed information.\n")
 }
 
-pub async fn exec(mut config: Config, args: &ArgMatches) -> Result<(), Error> {
+pub async fn exec(config: Config, args: &ArgMatches) -> Result<(), Error> {
     eprintln!("{}\n", UNSTABLE_WARNING);
-    let database = args.get_one::<String>("database").unwrap();
```
cloutiertyler, line 42: What happened to the `database` arg? You're not using it but it's still an argument.

crates/cli/src/subcommands/call.rs
```diff
@@ -37,50 +38,36 @@ pub fn cli() -> clap::Command {
...
-        anon_identity,
-        !force,
-    )
-    .await?;
+
+    let conn = parse_req(config, args).await?;
+    let api = ClientApi::new(conn);
+
+    let database_identity = api.con.database_identity;
+    let database = &api.con.database;
```
cloutiertyler, line 50: You get it from the `api.con` but that does not take into account the args. What's going on here?

crates/cli/src/subcommands/call.rs
```diff
@@ -1,18 +1,19 @@
...
 use spacetimedb::Identity;
-use spacetimedb_lib::de::serde::deserialize_from;
-use spacetimedb_lib::sats::{AlgebraicType, AlgebraicTypeRef, Typespace};
+use spacetimedb_lib::sats::{self, AlgebraicType, Typespace};
 use spacetimedb_lib::ProductTypeElement;
+use spacetimedb_schema::def::{ModuleDef, ReducerDef};
 use std::fmt::Write;
-use std::iter;
+
+use super::sql::parse_req;
```
cloutiertyler, line 16: Why is `parse_req` in the `sql` module? Also what does `parse_req` mean. I realize that this is a pre-existing function.

crates/cli/src/subcommands/call.rs
cloutiertyler: Looks like it does take into account the args via the `parse_req` function. That is just confusingly in the `sql` `mod`.

crates/cli/src/subcommands/call.rs
coolreader18: We pass `args` to `parse_req()` which extracts it. We do this in a couple other places in the cli, and I figured if I'm reworking the api that I'd move it to the centralized ClientApi struct in the process, so it's easier to see where we're calling from, rather than the http API interaction being spread out all across the crate.

crates/cli/src/subcommands/call.rs
coolreader18: Yeah, I dunno why it's there. I have a follow-up branch that moves most of the cli to use ClientApi and renames parse_req that I was gonna try and squeeze into http-glowup, but I decided against it. I'll probably open it post-1.0.

cloutiertyler: Roger. Thank you!

cloutiertyler (approved):
This all looks quite good to me. It's definitely an improvement. I ran it on `quickstart-chat`

```json
{
  "typespace": {
    "types": [
      {
        "Product": {
          "elements": [
            {
              "name": {
                "some": "sender"
              },
              "algebraic_type": {
                "Product": {
                  "elements": [
                    {
                      "name": {
                        "some": "__identity__"
                      },
                      "algebraic_type": {
                        "U256": []
                      }
                    }
                  ]
                }
              }
            },
            {
              "name": {
                "some": "sent"
              },
              "algebraic_type": {
                "Product": {
                  "elements": [
                    {
                      "name": {
                        "some": "__timestamp_micros_since_unix_epoch__"
                      },
                      "algebraic_type": {
                        "I64": []
                      }
                    }
                  ]
                }
              }
            },
            {
              "name": {
                "some": "text"
              },
              "algebraic_type": {
                "String": []
              }
            }
          ]
        }
      },
      {
        "Product": {
          "elements": [
            {
              "name": {
                "some": "identity"
              },
              "algebraic_type": {
                "Product": {
                  "elements": [
                    {
                      "name": {
                        "some": "__identity__"
                      },
                      "algebraic_type": {
                        "U256": []
                      }
                    }
                  ]
                }
              }
            },
            {
              "name": {
                "some": "name"
              },
              "algebraic_type": {
                "Sum": {
                  "variants": [
                    {
                      "name": {
                        "some": "some"
                      },
                      "algebraic_type": {
                        "String": []
                      }
                    },
                    {
                      "name": {
                        "some": "none"
                      },
                      "algebraic_type": {
                        "Product": {
                          "elements": []
                        }
                      }
                    }
                  ]
                }
              }
            },
            {
              "name": {
                "some": "online"
              },
              "algebraic_type": {
                "Bool": []
              }
            }
          ]
        }
      }
    ]
  },
  "tables": [
    {
      "name": "message",
      "product_type_ref": 0,
      "primary_key": [],
      "indexes": [],
      "constraints": [],
      "sequences": [],
      "schedule": {
        "none": []
      },
      "table_type": {
        "User": []
      },
      "table_access": {
        "Public": []
      }
    },
    {
      "name": "user",
      "product_type_ref": 1,
      "primary_key": [
        0
      ],
      "indexes": [
        {
          "name": {
            "some": "user_identity_idx_btree"
          },
          "accessor_name": {
            "some": "identity"
          },
          "algorithm": {
            "BTree": [
              0
            ]
          }
        }
      ],
      "constraints": [
        {
          "name": {
            "some": "user_identity_key"
          },
          "data": {
            "Unique": {
              "columns": [
                0
              ]
            }
          }
        }
      ],
      "sequences": [],
      "schedule": {
        "none": []
      },
      "table_type": {
        "User": []
      },
      "table_access": {
        "Public": []
      }
    }
  ],
  "reducers": [
    {
      "name": "identity_connected",
      "params": {
        "elements": []
      },
      "lifecycle": {
        "some": {
          "OnConnect": []
        }
      }
    },
    {
      "name": "identity_disconnected",
      "params": {
        "elements": []
      },
      "lifecycle": {
        "some": {
          "OnDisconnect": []
        }
      }
    },
    {
      "name": "init",
      "params": {
        "elements": []
      },
      "lifecycle": {
        "some": {
          "Init": []
        }
      }
    },
    {
      "name": "send_message",
      "params": {
        "elements": [
          {
            "name": {
              "some": "text"
            },
            "algebraic_type": {
              "String": []
            }
          }
        ]
      },
      "lifecycle": {
        "none": []
      }
    },
    {
      "name": "set_name",
      "params": {
        "elements": [
          {
            "name": {
              "some": "name"
            },
            "algebraic_type": {
              "String": []
            }
          }
        ]
      },
      "lifecycle": {
        "none": []
      }
    }
  ],
  "types": [
    {
      "name": {
        "scope": [],
        "name": "Message"
      },
      "ty": 0,
      "custom_ordering": true
    },
    {
      "name": {
        "scope": [],
        "name": "User"
      },
      "ty": 1,
      "custom_ordering": true
    }
  ],
  "misc_exports": [],
  "row_level_security": []
}
```

PR #2187, opened 2025-01-28
Meter wasm memory usage

coolreader18:
# Description of Changes

Add a `record_reducer_allocated_memory` method to `EnergyMonitor`, and call it after every reducer run. The `EnergyMonitor` implementation can decide how much it wants to debounce the event.

1

# Testing

This just adds a new hook, but there's not yet an implementation of that hook to test.

### DIFF
--- a/crates/core/src/host/wasm_common/module_host_actor.rs
+++ b/crates/core/src/host/wasm_common/module_host_actor.rs
@@ -77,6 +77,7 @@ pub struct ExecutionTimings {
 pub struct ExecuteResult<E> {
     pub energy: EnergyStats,
     pub timings: ExecutionTimings,
+    pub memory_allocation: usize,
     pub call_result: Result<Result<(), Box<str>>, E>,
 }
 
@@ -193,6 +194,8 @@ impl<T: WasmModule> WasmModuleHostActor<T> {
             instance,
             info: self.info.clone(),
             energy_monitor: self.energy_monitor.clone(),
+            // will be updated on the first reducer call
+            allocated_memory: 0,
             trapped: false,
         }
     }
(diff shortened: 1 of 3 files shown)


crates/core/src/worker_metrics/mod.rs
```diff
@@ -85,6 +85,11 @@ metrics_group!(
         #[labels(caller_identity: Identity, module_hash: Hash, caller_address: Address, reducer_symbol: str)]
         pub wasm_instance_errors: IntCounterVec,
 
+        #[name = spacetime_worker_wasm_allocated_memory]
+        #[help = "The number of bytes of linear memory allocated by the database's WASM module instance"]
+        #[labels(database_identity: Identity)]
+        pub wasm_allocated_memory: IntGaugeVec,
```
jsdt, line 91: I think `wasm_memory_bytes` would be a bit more clear, since the amount of memory allocated is usually not the same as the amount of memory used.

crates/core/src/host/wasm_common/module_host_actor.rs
```diff
@@ -466,11 +467,18 @@ impl<T: WasmInstance> WasmModuleInstance<T> {
...
             timings,
+            memory_allocation,
             call_result,
         } = result;
 
         self.energy_monitor
             .record_reducer(&energy_fingerprint, energy.used, timings.total_duration);
+        self.energy_monitor
+            .record_reducer_allocated_memory(&energy_fingerprint, memory_allocation);
+        WORKER_METRICS
```
jsdt, line 478: Since we are reporting the prometheus metric here, I think we can remove the `record_reducer_allocated_memory` function entirely.

coolreader18: Ah, alright.

crates/core/src/worker_metrics/mod.rs
coolreader18: Well, this is the amount of memory allocated, but it's still taking up RAM whether or not the wasm module is actually using it or not.

jsdt (approved):

PR #2190, opened 2025-01-29
Track compute metrics for sql dml with new engine

joshua-spacetime:
# Description of Changes

DML was still using the old engine. This patch updates it to use the new engine.

NOTE: Internal tests are failing due to a new dependency in core. Hence this pr will merge with internal tests failing, but they will be fixed immediately afterwards.

None

3

There are a few non-trivial changes:

1. `SET` and `SHOW` statements are rewritten as `INSERT` and `SELECT` respectively
2. We have to parse the query to determine if it is a dml statement and therefore whether it should be executed in a mutable or read-only transaction

# Testing

All previous tests still apply

### DIFF
--- a/crates/execution/src/dml.rs
+++ b/crates/execution/src/dml.rs
@@ -1,0 +1,152 @@
+use anyhow::Result;
+use spacetimedb_lib::{metrics::ExecutionMetrics, AlgebraicValue, ProductValue};
+use spacetimedb_physical_plan::dml::{DeletePlan, InsertPlan, MutationPlan, UpdatePlan};
+use spacetimedb_primitives::{ColId, TableId};
+use spacetimedb_sats::size_of::SizeOf;
+
+use crate::{pipelined::PipelinedProject, Datastore, DeltaStore};
+
+/// A mutable datastore can read as well as insert and delete rows
+pub trait MutDatastore: Datastore + DeltaStore {
+    fn insert_product_value(&mut self, table_id: TableId, row: &ProductValue) -> Result<()>;
+    fn delete_product_value(&mut self, table_id: TableId, row: &ProductValue) -> Result<()>;
+}
+
+/// Executes a physical mutation plan
+pub enum MutExecutor {
+    Insert(InsertExecutor),
+    Delete(DeleteExecutor),
+    Update(UpdateExecutor),
+}
+
+impl From<MutationPlan> for MutExecutor {
+    fn from(plan: MutationPlan) -> Self {
+        match plan {
+            MutationPlan::Insert(plan) => Self::Insert(plan.into()),
+            MutationPlan::Delete(plan) => Self::Delete(plan.into()),
+            MutationPlan::Update(plan) => Self::Update(plan.into()),
+        }
+    }
+}
+
+impl MutExecutor {
+    pub fn execute<Tx: MutDatastore>(&self, tx: &mut Tx, metrics: &mut ExecutionMetrics) -> Result<()> {
+        match self {
+            Self::Insert(exec) => exec.execute(tx, metrics),
+            Self::Delete(exec) => exec.execute(tx, metrics),
+            Self::Update(exec) => exec.execute(tx, metrics),
+        }
+    }
+}
+
+/// Executes row insertions
+pub struct InsertExecutor {
+    table_id: TableId,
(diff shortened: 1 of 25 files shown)


crates/execution/src/pipelined.rs
```diff
@@ -6,12 +6,61 @@ use std::{
...
 };
 use spacetimedb_primitives::{ColId, IndexId, TableId};
 
 use crate::{Datastore, DeltaStore, Row, Tuple};
 
+/// An executor for explicit column projections.
+/// Note, this plan can only be constructed from the http api,
+/// which is not considered performance critical.
+/// Hence this operator is not particularly optimized.
+pub enum ProjectListExecutor {
```
mamcx, line 20: Why is another projector necessary?

crates/expr/src/expr.rs
```diff
@@ -26,11 +26,32 @@ pub enum ProjectName {
 }
 
 impl ProjectName {
-    /// What is the [TableId] for this projection?
-    pub fn table_id(&self) -> Option<TableId> {
+    /// The [TableSchema] of the returned rows.
+    /// Note this expression returns rows from a relvar.
+    /// Hence it this method should never return [None].
+    pub fn return_table(&self) -> Option<&TableSchema> {
```
mamcx, line 32: Then why use `Option`?

crates/execution/src/pipelined.rs
mamcx: A: Is for put the metrics

crates/execution/src/dml.rs
```diff
@@ -0,0 +1,152 @@
...
+        let mut deletes = vec![];
+        self.filter.execute(tx, metrics, &mut |row| {
+            deletes.push(row.to_product_value());
+            Ok(())
+        })?;
+        // TODO: This metric should be updated inline when we serialize.
+        // Note, that we don't update bytes written,
+        // because deletes don't actually write out any bytes.
+        metrics.bytes_scanned += deletes.iter().map(|row| row.size_of()).sum::<usize>();
+        for row in &deletes {
```
mamcx, line 96: What advantage is in do this 2-pass, instead of `metrics.bytes_scanned += row.size_of()`?

crates/expr/src/expr.rs
```diff
@@ -26,11 +26,32 @@ pub enum ProjectName {
...
+    /// Hence it this method should never return [None].
+    pub fn return_table_id(&self) -> Option<TableId> {
+        match self {
+            Self::None(input) => input.return_table_id(),
+            Self::Some(input, alias) => input.find_table_id(alias),
+        }
+    }
+
+    /// Iterate over the returned column names and types
+    pub fn iter_return_fields(&self, mut f: impl FnMut(&str, &AlgebraicType)) {
```
mamcx, line 50: NIT: It hits me that `iter` is not the best name for these methods that implement internal iteration, it could be `each` instead, but this could be done later.

crates/expr/src/statement.rs
```diff
@@ -201,36 +236,132 @@ fn is_var_valid(var: &str) -> bool {
...
+        .map(|ColumnSchema { col_type, .. }| col_type)
+        .ok_or_else(|| Unresolved::field(ST_VAR_NAME, VALUE_COLUMN))?;
+
+    // -------------------------------------------
+    // SELECT value FROM st_var WHERE name = 'var'
+    //                                ^^^^
+    // -------------------------------------------
+    let var_name_field = Expr::Field(FieldProject {
+        table: ST_VAR_NAME.into(),
+        field: 0,
```
mamcx, line 318: This and similar should have a comment that points to `StVarFields::Name`.

It should be the case that the enums on the `system table` are accessible here so we don't hard_code constants.

crates/execution/src/dml.rs
joshua-spacetime: None

crates/expr/src/statement.rs
joshua-spacetime: > It should be the case that the enums on the system table are accessible here so we don't hard_code constants.

You're right, it's just not so easy to pull `StVarFields` out of core, which is why I left it in. I can certainly add a comment though.

mamcx (approved):

PR #2192, opened 2025-01-29
Make `check_and_insert` & `find_old_row` and friends do less checks

Centril:
# Description of Changes

In general, what this PR achieves is to remove various bounds checks and type checks in relation to index lookups and insertions.

This shows up in flamegraphs as getting rid of a `ReadColumn::read_column` and shrinking the share of `RowRef::{project -> project_unchecked}` as well as reducing the self-time of `confirm_insertion`, `values_in_range`.

None.

3, plenty of `unsafe` was added but the proofs/requirements are mostly boring plumbing. Still, it is unfortunate :(

# Testing

Existing tests have been amended.


gefjon (approved):
I agree with your reasoning in the safety comments. I'll leave it to you to decide whether the proliferation of `unsafe` is worth the performance gains.

PR #2196, opened 2025-01-30
Update support for sql literal expressions

joshua-spacetime:
# Description of Changes

Parses literals in accordance with the sql spec. Specifically it adds support for scientific notation.

None

2

Adds a dependency on `BigDecimal` both for parsing scientific notation as well as for parsing 128 and 256 bit integer literals.

# Testing

Added new parsing and type checking test cases


crates/expr/src/check.rs
```diff
@@ -258,57 +262,211 @@ mod tests {
...
+            assert!(result.is_ok(), "{msg}");
         }
     }
 
     #[test]
     fn invalid() {
         let tx = SchemaViewer(module_def());
 
-        for sql in [
-            // Table r does not exist
```
joshua-spacetime, line 287: Same here, turning the comments into failure messages.

crates/expr/src/check.rs
```diff
@@ -239,9 +239,13 @@ mod tests {
...
                 ProductType::from([
-                    ("int", AlgebraicType::U32),
+                    ("i8", AlgebraicType::I8),
+                    ("u8", AlgebraicType::U8),
+                    ("i32", AlgebraicType::I32),
                     ("u32", AlgebraicType::U32),
+                    ("int", AlgebraicType::U32),
                     ("f32", AlgebraicType::F32),
+                    ("u256", AlgebraicType::U256),
                     ("str", AlgebraicType::String),
```
mamcx, line 249: I think we should add all the types that are parsed by `BigDecimal` and are part of the specs, like `i64`.

joshua-spacetime: Added

crates/expr/src/check.rs
```diff
@@ -239,9 +239,21 @@ mod tests {
...
+                    ("u8", AlgebraicType::U8),
+                    ("i16", AlgebraicType::I16),
+                    ("u16", AlgebraicType::U16),
+                    ("i32", AlgebraicType::I32),
                     ("u32", AlgebraicType::U32),
+                    ("i64", AlgebraicType::I64),
+                    ("u64", AlgebraicType::U64),
+                    ("int", AlgebraicType::U32),
                     ("f32", AlgebraicType::F32),
+                    ("f64", AlgebraicType::F64),
```
mamcx, line 252:
```suggestion
                    ("real", AlgebraicType::F32),
                    ("f64", AlgebraicType::F64),
```

To match our spec.

joshua-spacetime: They're just easy to decipher column names, nothing more than that.

mamcx (approved):
LGTM

PR #2200, opened 2025-02-03
Fix ctrl-c handling

### LINKED ISSUE #2133: Signal handling
Since #2011 , it is no longer possible to stop `spacetimedb start` by ^C. Killing the parent (`spacetimedb-cli`) will leave the child (`spacetimedb-standalone`) running.

OS: Linux

coolreader18:
# Description of Changes

Uses exec on unix and `SetConsoleCtrlHandler` on windows. This should fix issues where ctrl-C didn't properly shut down the server or left an orphan process behind. Fixes #2133.

2

# Testing

- [x] I've verified that ctrl-C goes straight to the server process on unix.
- [ ] Haven't tested on windows, but this is what rustup and cargo use to avoid this phenomenon, but if the reviewer wants to they could test on windows (or request that I do).


bfops:
This PR looks good to me, thank you!

We should confirm on Windows. I'm working on setting up a Windows machine that the team can use for testing things. Once it's set up, I'll send you the details. (If you already have access to a Windows machine that is already set up with the SpacetimeDB repo, you can test this on there without waiting for me though).

coolreader18:
Well, I can't repro the issue, but I *can* confirm the issue isn't happening with this patch. :P

coolreader18:
also to note: iirc, this is an entirely hypothetical issue on windows, I don't believe we've gotten any actual reports of this.

bfops (approved):
Sounds good to me!

PR #2203, opened 2025-02-04
Match datastore semantics more closely: `btree_scan` => `index_scan_range`

Centril:
# Description of Changes

In preparation for adding a index algo/kind that we'll call `direct` (backed by `Vec<Vec<RowPointer>>`, ...),
this PR moves all the module bindings APIs (Rust + C#) to use `datastore_[delete_by_]index_scan_range_bsatn`
and deprecates  `datastore_[delete_by_]btree_scan_bsatn`.

This more closely reflects what the datastore actually does today.
Nowhere (mostly because there's only a single index algo) does the datastore actually check that a btree index is used.
What the ABI actually is about is *fast range queries*.
The new `direct` index can also support that, so the more general name makes sense,
while it is also beneficial for efficiency, as we can avoid a check.

After this change, there are no users of the deprecated ABIs, so we could elect to remove them just before 1.0,
but this PR leaves them be.

Some datastore types and methods are also renamed to reflect the new reality.

None, for now.

2, mostly simple renaming and code motion.

# Testing

Existing tests should cover this.


crates/core/src/host/instance_env.rs
```diff
@@ -374,7 +374,7 @@ impl InstanceEnv {
     }
 
     #[tracing::instrument(level = "trace", skip_all)]
-    pub fn datastore_btree_scan_bsatn_chunks(
+    pub fn datastore_index_scan_range_bsatn_chunks(
```
joshua-spacetime, line 377: Note there are other index types that support range queries. A spacial index type for example.

joshua-spacetime (commented):
> What the ABI actually is about is fast range queries.
The new direct index can also support that,

Will there be logic for choosing which index to use within the current syscall?

joshua-spacetime (approved):

cloutiertyler:
> What the ABI actually is about is fast range queries.

I would dispute this point, the ABI is actually about using an index to perform a range query. The host must validate that the particular type of index specified by `index_id` can support the query being made.

I think this is a very important point. Not all indexes will be able to support all queries. 

> while it is also beneficial for efficiency, as we can avoid a check.

I don't think you can because if we add a hash index we need to validate that that you did not pass in an `index_id` to a hash index.

cloutiertyler:
TBH I'm having second thoughts about renaming it like this since those points were lost in the short period between our conversation and merging this PR.

joshua-spacetime:
> I think it is very very important that we're clear that these are physical datastore operations, and the query engine should always be able to choose the specific physical operation it wants to do.

This was my main concern, and I approved based on the confirmation from @Centril that these are indeed remaining purely physical operations.

gefjon:
Two cents here: the host function is going to have to dispatch on the index type no matter what. If we have separate `btree_scan` vs `dense_seq_vector_scan` methods, then each of them checks equality on the index type, and returns an error otherwise. If we have a unified `index_scan_range`, then it does a switch/case on the index type, and the cases for unordered index types return errors. There's no version of this, given our system table schemas, where you can have a host function `btree_scan` which immediately does a low-level physical operation without first checking the index type.

PR #2205, opened 2025-02-04
add `direct` indices, except for in datastore & C#

Centril:
# Description of Changes

Adds the schema and rust module side support for `direct` indices.
What remains to be done after this PR is to:
- hook up the datastore support for the index.
- add C# module support.

None

3, touches a lot of places.

# Testing

Some new tests are added for the schema and existing tests are amended.

### DIFF
--- a/crates/bindings-macro/src/table.rs
+++ b/crates/bindings-macro/src/table.rs
@@ -596,11 +596,12 @@ pub(crate) fn table_impl(mut args: TableArgs, item: &syn::DeriveInput) -> syn::R
     // For all the unpaired unique columns, create a unique index.
     for unique_col in &unique_columns {
         if args.indices.iter_mut().any(|index| {
-            index.is_unique = match &index.kind {
+            let covered_by_index = match &index.kind {
                 IndexType::BTree { columns } => &**columns == slice::from_ref(unique_col.ident),
                 IndexType::Direct { column } => column == unique_col.ident,
             };
-            index.is_unique
+            index.is_unique |= covered_by_index;
+            covered_by_index
         }) {
             continue;
         }


crates/bindings-macro/src/table.rs
```diff
@@ -521,19 +592,48 @@ pub(crate) fn table_impl(mut args: TableArgs, item: &syn::DeriveInput) -> syn::R
...
+        // as it's the safest choice of index for the general case,
+        // even if isn't optimal in specific cases.
+        let name = unique_col.ident.clone();
+        let columns = vec![name.clone()];
+        args.indices.push(IndexArg {
+            name,
+            is_unique: true,
+            kind: IndexType::BTree { columns },
+        })
+    }
```
kazimuth, line 628: A corresponding patch will be needed for C# modules

crates/schema/src/def.rs
```diff
@@ -280,58 +279,6 @@ impl ModuleDef {
         Some(TableSchema::from_module_def(self, table_def, (), table_id))
     }
 
-    /// Generate indexes for the module definition.
-    /// We guarantee that all `unique` constraints have an index generated for them.
-    /// This will be removed once another enforcement mechanism is implemented.
-    /// This is a noop if there are already usable indexes present.
-    fn generate_indexes(&mut self) {
```
kazimuth, line 287: Is there info available in macros that isn't available here? If not, it may be easier to implement the patch here. Then we don't need to patch both C# and Rust macros to do this change.

(Unless this code was totally dead? If so my b.)

crates/schema/src/def/validate/v9.rs
```diff
@@ -1347,11 +1397,47 @@ mod tests {
...
+                ProductType::from([("b", AlgebraicType::I32), ("a", AlgebraicType::U64)]),
+                false,
+            )
+            .with_index(RawIndexAlgorithm::Direct { column: 0.into() }, "bananas_b")
+            .finish();
+        let result: Result<ModuleDef> = builder.finish().try_into();
+
+        expect_error_matching!(result, ValidationError::DirectIndexOnNonUnsignedInt { index, .. } => {
+            &index[..] == "Bananas_b_idx_direct"
+        });
```
kazimuth, line 1438: 👍

kazimuth: You'll also want to add a direct index to `valid_definition` & make sure it goes through

kazimuth (commented):
All the schema stuff looks good.

crates/schema/src/def.rs
kazimuth: On the other hand, we'll also need to patch the C# module library to support these.

crates/bindings-macro/src/table.rs
```diff
@@ -1,12 +1,14 @@
 use crate::sats;
 use crate::sym;
 use crate::util::{check_duplicate, check_duplicate_msg, ident_to_litstr, match_meta};
+use core::slice::from_ref;
```
coolreader18, line 4: I'd say probably just import `slice` and call `slice::from_ref()`; `from_ref()` as a free identifier doesn't give much context.

crates/bindings/src/rt.rs
```diff
@@ -364,12 +365,33 @@ pub fn register_table<T: Table>() {
...
 
-impl From<crate::table::IndexAlgo<'_>> for RawIndexAlgorithm {
-    fn from(algo: crate::table::IndexAlgo<'_>) -> RawIndexAlgorithm {
+mod sealed_direct_index {
+    pub trait Sealed {}
+}
+#[diagnostic::on_unimplemented(
+    message = "scheduled table primary key must be a `u64`",
+    label = "should be `u64`, not `{Self}`"
+)]
```
coolreader18, line 374: This looks like the wrong error message.

crates/bindings/src/rt.rs
```diff
@@ -364,12 +365,33 @@ pub fn register_table<T: Table>() {
...
+pub trait DirectIndexIsUnsignedU8ToU64: sealed_direct_index::Sealed {}
```
coolreader18, line 375: I feel like this could just be called `DirectIndexKey`.

coolreader18 (approved):
macro/bindings stuff generally LGTM; `UniqueColumn` will probably need an extra bound in the future in order to tell it whether to use btree or direct, but that's fine.

PR #2208, opened 2025-02-04
Fix multi-column index scan

mamcx:
# Description of Changes

In the case of permutations, multi-column indexes were not used when optimizing the plan, the `pass` assumed the order of the columns, and now it looks for it without this assumption.

1

# Testing

- [x] *Addd unit test that check permuation for 2/3 columns*

### DIFF
--- a/crates/physical-plan/src/plan.rs
+++ b/crates/physical-plan/src/plan.rs
@@ -1622,6 +1622,28 @@ mod tests {
             proj => panic!("unexpected plan: {:#?}", proj),
         };
 
+        // Test permutations of the same query
+        let sql = "select * from t where z = 5 and y = 4 and x = 3";
+        let lp = parse_and_type_sub(sql, &db).unwrap();
+        let pp = compile_select(lp).optimize();
+
+        match pp {
+            ProjectPlan::None(PhysicalPlan::IxScan(
+                IxScan {
+                    schema, prefix, arg, ..
+                },
+                _,
+            )) => {
+                assert_eq!(schema.table_id, t_id);
+                assert_eq!(arg, Sarg::Eq(ColId(1), AlgebraicValue::U8(3)));
+                assert_eq!(
+                    prefix,
+                    vec![(ColId(3), AlgebraicValue::U8(5)), (ColId(2), AlgebraicValue::U8(4))]
+                );
+            }
+            proj => panic!("unexpected plan: {:#?}", proj),
+        };
+
         let sql = "select * from t where x = 3 and y = 4";
         let lp = parse_and_type_sub(sql, &db).unwrap();
         let pp = compile_select(lp).optimize().unwrap();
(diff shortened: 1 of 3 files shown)


crates/physical-plan/src/rules.rs
```diff
@@ -318,6 +319,13 @@ impl RewriteRule for PushConstAnd {
     }
 }
 
+/// Find the column index for a given field.
+///
+/// *NOTE*: This take in account the possibility of permutations.
+fn find_col_index(index_algorithm: &IndexAlgorithm, pos: usize) -> Option<ColId> {
```
joshua-spacetime, line 325: Maybe this should be a method on `IndexAlgorithm` itself?

crates/physical-plan/src/rules.rs
```diff
@@ -502,22 +510,8 @@ impl RewriteRule for IxScanEq2Col {
                                             Some(IxScanInfo {
                                                 index_id: *index_id,
                                                 cols: vec![
-                                                    (
```
joshua-spacetime, line 505: Note above where we have
```rust
return schema
    .indexes
    .iter()
    ...
```

We still need this to be
```rust
if let Some(...) = schema
    .indexes
    .iter()
    ... {
    return ...
}
```

crates/physical-plan/src/plan.rs
```diff
@@ -1617,6 +1617,28 @@ mod tests {
             proj => panic!("unexpected plan: {:#?}", proj),
         };
 
+        // Test permutations of the same query
+        let sql = "select * from t where z = 5 and y = 4 and x = 3";
```
joshua-spacetime, line 1621: Can you add a similar test for `w, y, z`? I think that should still fail without updating the early `return`.

crates/physical-plan/src/rules.rs
```diff
@@ -358,12 +358,12 @@ impl RewriteRule for IxScanEq {
...
                              ..
                          }| {
-                            index_algorithm
-                                .columns()
-                                // TODO: Support prefix scans
-                                .as_singleton()
-                                .filter(|col_id| col_id.idx() == *pos)
-                                .map(|col_id| (*index_id, col_id))
+                            // TODO: Support prefix scans
+                            if index_algorithm.columns().len() == 1 {
```
joshua-spacetime, line 362: Did this need to be changed? It looks equivalent to the above.

joshua-spacetime (approved):

crates/physical-plan/src/plan.rs
mamcx: Done

crates/physical-plan/src/rules.rs
mamcx: Yes, but this way the logic for finding the index is centralized.

PR #2209, opened 2025-02-04
Return errors for invariant violations during query planning

joshua-spacetime:
# Description of Changes

Instead of invoking `unreachable!` in the query rewrite code, now we return an error. An error means we have an internal logic bug, but we don't want the server to panic in that case. Instead we just log the error. There's a TODO for handling the error appropriately in the subscription code.

None

2

Just a return type change and updating callers.

# Testing

A simple refactor so no new tests added.


mamcx (approved):
lgtm

PR #2212, opened 2025-02-05
Make `MutTxId::update` really idempotent wrt. `TxData`

Centril:
# Description of Changes

Given:
```rust
insert(row);
commit();
update(row);
```
there is a bug in `MutTxId::update` where it doesn't elide the delete/insert pair, unlike `MutTxId::insert`.

This PR fixes the bug.

Not yet.

2, fairly local change.

# Testing

The `test_update_idempotent` test is updated to assert the idempotence of `TxData` in this scenario.


gefjon (approved):
Code looks good, test looks good, and I tested locally and verified that I don't see the issue any more. Thanks so much for your quick work!

PR #2213, opened 2025-02-06
Handle optimization and execution errors on initial subscription

joshua-spacetime:
# Description of Changes

If an error occurs during query optimization or execution of the initial subscription, we return that error to the client.

This patch also simplifies the subscription compilation code, particularly join subscriptions.

It also compiles a subscription once and caches the plan for reuse, as was the case for the previous subscription api. This removes superfluous error checks during subscription update.

None

3

Updates some type signatures that permeate the codebase, but we also have a lot of test coverage over this part of the code.

# Testing

This is a refactor. All previous tests apply.


crates/subscription/src/lib.rs
```diff
@@ -0,0 +1,319 @@
...
+struct Fragments {
+    /// Plan fragments that return rows to insert.
+    /// For joins there will be 4 fragments,
+    /// but for selects only one.
+    insert_plans: Vec<PipelinedProject>,
+    /// Plan fragments that return rows to delete.
+    /// For joins there will be 4 fragments,
+    /// but for selects only one.
+    delete_plans: Vec<PipelinedProject>,
+}
```
mamcx, line 23: Could we have a constructor for both so we can filter out the `empty` ones early?

joshua-spacetime: We filter the empty ones out early for both in `for_each_insert` and `for_each_delete`

mamcx (approved):
LGTM

PR #2220, opened 2025-02-06
Rename `Address` to `ConnectionId`

gefjon:
# Description of Changes

This PR renames `Address`, which was used only for identifying client connections at this point, to `ConnectionId`. Semantic changes are kept to a minimum.

Notable features of this PR:

- My workflow significantly involved grepping for "address" and replacing instances with correct new terms. Some of the uses were stale, as they referred to "database address," which has since been replaced by the concept of a "database identity." I chose to update these uses, both because they were incorrect and because leaving them in place would have negatively impacted my grep-based workflow. This adds some lines to the diff which are not strictly necessary.
- I changed `Address::from_byte_array` and `Address::to_byte_array` to `ConnectionId::from_le_byte_array` and `ConnectionId::to_le_byte_array`. My theory is that the name is changing anyways, so I might as well make the endianness explicit. This is a direct response to the comments all over `address.rs` which talk about correct endianness, often IN ALL CAPS.
- I retained the interface for setting the connection ID via a query parameter to the `/database/subscribe` HTTP endpoint, which is the one that establishes a WS connection. However, I removed it from `/database/call`.
  - I added a deprecation log, currently at `debug` level, when that query parameter is supplied.
  - Cloud still depends on being able to set its connection ID.
  - I did not any of the client SDKs' handling of connection IDs, so they will still generate a random connection ID and pass it to that query parameter. This means that the host will always log a note when any client establishes a connection.
    - I would like to make this change to the SDKs in a follow-up, post-1.0.

# Companion PRs

- [private](https://github.com/clockworklabs/SpacetimeDBPrivate/pull/1331).
- [TypeScript SDK](https://github.com/clockworklabs/spacetimedb-typescript-sdk/pull/133).
- [C# SDK](https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/pull/239).

Yep! A significant, though superficial, API break.

5: the word "address" was in just about all of our APIs, both internal and external, and they're all changing. Every single component we have needs to be updated in lockstep to account for this.

# Testing

- [x] Really just automated tests.
- [ ] Someone should run Blackholio and make sure it works.
  - I cannot do this because I don't have a Unity license.
- [ ] Someone should run the TypeScript SDK quickstart.
  - I am unsure how to do this, though I could potentially puzzle it out given time.


bfops (approved):
My codeowned files LGTM :+1: 

`crates/cli/src/subcommands/call.rs`
`crates/cli/src/subcommands/subscribe.rs`

jsdt (approved):
I looked through all but the generated files, and all of the changes seemed right to me.

crates/cli/src/subcommands/subscribe.rs
```diff
@@ -25,7 +25,7 @@ pub fn cli() -> clap::Command {
         .arg(
             Arg::new("database")
                 .required(true)
-                .help("The domain or address of the database you would like to query"),
+                .help("The domain or identity of the database you would like to query"),
```
cloutiertyler, line 28:
```suggestion
                .help("The name or identity of the database you would like to query"),
```

cloutiertyler (approved):
Reviewed the files I'm a code owner of. Left just one suggested change.

PR #2221, opened 2025-02-06
Add `direct` indices to datastore

Centril:
# Description of Changes

The PR is best reviewed commit by commit, as they build upon the previous one and do one thing.

1. Adds a note re. renaming `BTreeIndex` -- later fixed in 6-7.

2. Adds `direct` indices to the datastore.

3. Adds micro-benchmarks for some known index implementations, including our unique btrees and direct indices.

4. Renames `BTreeIndex::seek` to `BTreeIndex::seek_range` and consequence changes.

5. Adds `BTreeIndex::seek_point`, which makes point scans about 12% faster in micro-benchmarks. This is then employed in the various places where we currently just use `&AlgebraicValue` as a bound. More can be done here however in follow ups.

6-7. Rename `BTreeIndex` to `TableIndex` to reflect that it isn't limited to btrees anymore.

8. Fixes a small typo bug in `bindings-macro`, necessary to actually use `#[index(direct)]`.

With `#[index(direct)]` applied to tables `Position` and `Velocity` in the update benchmark, we get an average 357ms, which is in line with what we found when we hacked in the index implementation.
```
2025-02-07T00:06:41.784526Z  INFO: : Timing span "update_positions_by_collect": 338.077526ms
2025-02-07T00:06:42.482635Z  INFO: : Timing span "update_positions_by_collect": 360.918827ms
2025-02-07T00:06:43.102741Z  INFO: : Timing span "update_positions_by_collect": 362.031175ms
2025-02-07T00:06:43.698571Z  INFO: : Timing span "update_positions_by_collect": 360.897486ms
2025-02-07T00:06:44.342312Z  INFO: : Timing span "update_positions_by_collect": 361.958901ms
```

This also improved subscription benchmarks as compared to current master (`footprint-semijoin`: -22%):
```
footprint-semijoin      time:   [112.12 µs 112.91 µs 114.08 µs]
                        change: [-22.769% -22.229% -21.549%] (p = 0.00 < 0.05)
                        Performance has improved.
```

None.

3, the PR is large, but broken into individual commits that aren't that complicated.
There's also some `unsafe` in the index implementation.

# Testing

Existing tests are amended and some new unit tests are added in the commit that adds `direct` indices.
The benchmark has also been run, exercising `#[index(direct)]` in a module.

### DIFF
--- a/crates/table/src/table_index/unique_direct_index.rs
+++ b/crates/table/src/table_index/unique_direct_index.rs
@@ -7,9 +7,9 @@ use core::option::IntoIter;
 /// A direct index for relating unsigned integer keys [`u8`..`u64`] to [`RowPointer`].
 ///
 /// This index is efficient when given keys that are used in non-random insert patterns
-/// where keys are dense and not far apart.
+/// where keys are dense and not far apart as well as starting near zero.
 /// Conversely, it performs worse than a btree index in the case of highly random inserts
-/// and with sparse keys.
+/// and with sparse keys and where the first key inserted is large.
 #[derive(Debug, Clone, Default, PartialEq, Eq)]
 pub struct UniqueDirectIndex {
     /// The outer index.


crates/table/src/btree_index/unique_direct_index.rs
```diff
@@ -0,0 +1,347 @@
+use crate::indexes::{PageIndex, PageOffset, RowPointer, SquashedOffset};
+use crate::MemoryUsage;
+use core::mem;
+use core::ops::{Bound, RangeBounds};
+use core::option::IntoIter;
+
+/// A direct index for relating unsigned integer keys [`u8`..`u64`] to [`RowPointer`].
+///
+/// This index is efficient when given keys that are used in non-random insert patterns
+/// where keys are dense and not far apart.
```
gefjon: And all near zero, right?

crates/table/src/btree_index/unique_direct_index.rs
```diff
@@ -0,0 +1,347 @@
...
+        repeat_with(move || {
+            if page_offset.0 as usize + FIXED_ROW_SIZE.0 as usize >= PageOffset::PAGE_END.0 as usize {
+                // Consumed the page, let's use a new page.
+                page_index.0 += 1;
+                page_offset = PageOffset(0);
+            } else {
+                page_offset += FIXED_ROW_SIZE;
+            }
+
+            black_box(RowPointer::new(
```
gefjon: `black_box` is suspicious outside of benchmarks, tbh.

gefjon (approved):
I have completely ignored your benchmarks during this review; I honestly don't care about them. Also, given our current time constraints, I haven't given this as careful of a review as I would like. But it all looks reasonable, and the concept here is so dirt-simple that it seems hard to do wrong, honestly.

PR #2223, opened 2025-02-07
Error handling for incremental evaluation

joshua-spacetime:
# Description of Changes

When an error occurs during a tx update (subscription evaluation), we will now:

1. Send a `SubscriptionError` to the relevant clients
2. Drop the full set of subscriptions for each such client

To drop the subscriptions, we mark the affected clients when the error occurs (see eval_updates), and an async task will periodically remove them from the mapping.

> Why not drop them when the error occurs (in eval_updates)?

This would require holding a write lock on the `SubscriptionManager` preventing other read only transactions from running.

None

4

Not a big change, but touches concurrency related code.

# Testing

Added the following new tests:

1. Subscribing with an invalid query returns an error to the client
2. Removing an index and unsubscribing from a query that relied on that index also returns an error to the client
3. A tx update that fails due to a removed index returns an error to the client

### DIFF
--- a/crates/core/src/subscription/module_subscription_manager.rs
+++ b/crates/core/src/subscription/module_subscription_manager.rs
@@ -222,12 +222,13 @@ impl SubscriptionManager {
     /// This will return an error if the client does not have a subscription with the given query id.
     pub fn remove_subscription(&mut self, client_id: ClientId, query_id: ClientQueryId) -> Result<Query, DBError> {
         let subscription_id = (client_id, query_id);
-        let Some(ci) = self.clients.get_mut(&client_id) else {
+        let Some(ci) = self
+            .clients
+            .get_mut(&client_id)
+            .filter(|ci| !ci.dropped.load(Ordering::Acquire))
+        else {
             return Err(anyhow::anyhow!("Client not found: {:?}", client_id).into());
         };
-        if ci.dropped.load(Ordering::Acquire) {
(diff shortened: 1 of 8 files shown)


gefjon (approved):
Pending testing, of course, but the code looks reasonable.

crates/core/src/subscription/module_subscription_manager.rs
```diff
@@ -182,14 +203,27 @@ impl SubscriptionManager {
...
     /// This will return an error if the client does not have a subscription with the given query id.
     pub fn remove_subscription(&mut self, client_id: ClientId, query_id: ClientQueryId) -> Result<Query, DBError> {
         let subscription_id = (client_id, query_id);
         let Some(ci) = self.clients.get_mut(&client_id) else {
             return Err(anyhow::anyhow!("Client not found: {:?}", client_id).into());
         };
-
+        if ci.dropped.load(Ordering::Acquire) {
+            return Err(anyhow::anyhow!("Client not found: {:?}", client_id).into());
+        }
```
coolreader18, line 226: Doesn't really matter, but `clients.get_mut(...).filter(|ci| !ci.dropped.load(Acquire))` might be a useful idiom here.

coolreader18 (approved):
LGTM. No lock-taking added, so I can't see anywhere where this would start deadlocking.

PR #2224, opened 2025-02-07
Implemented the new EventContext types in Typescript codegen

cloutiertyler:
# Description of Changes

Relatively small PR. Corresponds to https://github.com/clockworklabs/spacetimedb-typescript-sdk/pull/132

This PR generates a few new types based on new types added to the TypeScript SDK. Namely:

```rust
        "ReducerEventContextInterface",
        "SubscriptionEventContextInterface",
        "ErrorContextInterface",
        "SubscriptionBuilderImpl",
```

It also fixes a bug where `SubscriptionBuilder` was not actually type aware of the module types, so `onApplied(ctx => ctx.reducers` did not have the reducer types.

I have also improved the lint ignoring which in general reduces noise. I've wanted to do that for a while and since it helped debug here, I wanted to take the opportunity to merge it.

This is a TypeScript API breaking change because now:

- `ctx.on<reducer>` passes in a `ReducerEventContext` instead of an `EventContext` into its callback
- `subscriptionBuilder.onApplied` passes in a `SubscriptionEventContext` into its callback
- `subscriptionBuilder.onError` passes in an `ErrorContext` into its callback

Note that `onConnectError` is of the form: 

```ts
 onConnectError(
    callback: (connection: DBConnection, error: Error) => void
  ): DBConnectionBuilder<DBConnection> {
}
```

@gefjon is that correct? I believe this is what I was told to do in the previous implementation but it looks like this is different in the Rust version: https://github.com/clockworklabs/SpacetimeDB/blob/b50b6841f18c8b2004876d57db69bf7d06c62184/crates/sdk/examples/quickstart-chat/main.rs#L164

This seems like quick a confusing API. Surely this would be better (and more aligned with the ReducerArgs handling for the reducer callback):

```ts
 onConnectError(
    callback: (ctx: ErrorContext, error: Error) => void
  ): DBConnectionBuilder<DBConnection> {
}
```

2

# Testing

TODO

### DIFF
--- a/crates/cli/src/subcommands/generate/typescript.rs
+++ b/crates/cli/src/subcommands/generate/typescript.rs
@@ -100,7 +100,7 @@ impl Lang for TypeScript {
 
         // Import the types of all fields.
         // We only need to import fields which have indices or unique constraints,
-        // but it's easier to just import all of 'em, since we have `// @ts-ignore` anyway.
+        // but it's easier to just import all of 'em, since we have `// @ts-nocheck` anyway.
         gen_and_print_imports(
             module,
             out,
@@ -108,7 +108,6 @@ impl Lang for TypeScript {
             &[], // No need to skip any imports; we're not defining a type, so there's no chance of circular imports.
         );
 
-        writeln!(out, "// @ts-ignore");
         writeln!(
             out,
             "import {{ EventContext, Reducer, RemoteReducers, RemoteTables }} from \".\";"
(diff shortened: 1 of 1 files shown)


gefjon:
What's currently implemented in Rust is my intended design. I would not say I'm married to it, but I think it's fine, and at this point I consider the cost to change it quite high, as it would require PRs to all of public, private and docs.

My reasoning for the design I chose:

I think that having two distinct-but-equivalent ways to access the error will confuse users, and I don't see it as having a significant gain. Reading the error out of the context is easy. Keeping as much as possible inside the contexts is beneficial because we have more freedom to extend them in the future, whereas we essentially cannot change callback arglists post 1.0.

> more aligned with the ReducerArgs handling for the reducer callback

Reducer args are different from the error in that we don't have a clean way to make them accessible inside the context. Even after this change, reading them from the context requires pattern-matching to read a known variant from a nested field, which is tedious and verbose. However, reducer callbacks get metadata in addition to the arguments, e.g. sender identity and connection ID, which is only passed inside the context, not provided as unboxed arguments.

cloutiertyler:
> I think that having two distinct-but-equivalent ways to access the error will confuse users

Personally, I think having to access the error by `ctx.event` will confuse and surprise users. I think that if we surfaced the error as a second argument, people would quite rightly view the `ErrorContext` as the `DBConnection` primarily, and the error as the error.

I would strongly recommend surfacing the error as it's own argument in Rust.

kazimuth:
I don't think that is a good idea. It's better to do the change in a forwards-compatible manner via the context. Users will need to look inside the contexts for other information anyway so it's good to encourage them to do so.

gefjon (approved):

cloutiertyler:
@kazimuth I've made the decision on this one and we decided to go with `error` as its own argument.

PR #2225, opened 2025-02-07
Refactor HTTP api to better adhere to REST guidelines

coolreader18:
# Description of Changes

Based off #2181. This mainly hits `/database`, making the endpoints be `/database/:dbname/verb` rather than `/database/verb/:dbname`, and making the http methods be more accurate to what they do (e.g. `DELETE /database/:dbname` rather than `POST /database/delete/:dbname`). Dns stuff like register_tld (actually, just register_tld for now) got factored into its own `/domain` path.  Also, I put everything behind a `/v1` prefix -- I could revert that if we don't wanna do that, but that feels like best practice for HTTP APIs in general.

Yes.

3: obviously, there's a lot of stuff moving around here, but the smoketests are pretty thorough in exercising all the routes of the API, and if there's anything I miss here it's pretty easy to change.

# Testing

- [x] Smoketests and sdk tests exercising the api interactions in this repository.


gefjon:
Needs companion PRs in private, TypeScript SDK, C# SDK and docs. I can wait on the docs one, but the first three must happen before we merge this.

coolreader18:
Yep, makes sense :+1:

gefjon:
That said, the changes look good, and I'll be happy to approve once we have the companion PRs ready to go.

coolreader18:
https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/pull/242 and https://github.com/clockworklabs/spacetimedb-typescript-sdk/pull/134

cloutiertyler:
> Also, I put everything behind a /v1 prefix -- I could revert that if we don't wanna do that, but that feels like best practice for HTTP APIs in general.

We definitely want to do that. I was meaning to do it for a long time.

cloutiertyler (commented):
The changes look good to me. I have two comments/requests:

1. In the `Identity` proposal and remove `Address` proposal, I proposed to remove the concept of TLD and domains in general, and instead replace it with database names, where a database name matches `^[a-z0-9]+(-[a-z0-9]+)*$` (lowercase alphanumeric with internal dashes). That leaves room in the future for having multipart names like `foo.bar` or `foo/bar` in the future. This is only relevant to this PR insofar as we may not want to introduce a new `/domain` route if we're going to just basically have `/database/:name_or_identity/set-name` route. That being said, we have little time left and deprecating the `/domain` route in favor of the new `set-name` route is not a big deal. I am also fine to have a fix for that happen in another `PR`.

2. This needs a corresponding Private PR.

crates/client-api/src/routes/database.rs
```diff
@@ -741,26 +561,26 @@ pub async fn delete_database<S: ControlStateDelegate>(
 }
 
 #[derive(Deserialize)]
-pub struct SetNameQueryParams {
-    domain: String,
-    database_identity: IdentityForUrl,
+pub struct AddNameParams {
+    // this field is named such because that's what the route is under.
+    // TODO: find out if there's a reason this can't be specified as a domain name
```
cloutiertyler, line 566: What does this `TODO` comment mean? You should be able to set a new database name even if you use the old name to identify it.

crates/client-api/src/routes/database.rs
```diff
@@ -772,46 +592,62 @@ pub async fn set_name<S: ControlStateDelegate>(
...
 {
-    use axum::routing::{get, post};
+    use axum::routing::{delete, get, post, put};
+    let publish = publish::<S>.layer(DefaultBodyLimit::disable());
+    let db_router = axum::Router::<S>::new()
+        .route("/", put(publish.clone()))
+        .route("/", get(db_info::<S>))
+        .route("/", delete(delete_database::<S>))
+        .route("/names", get(get_name::<S>))
+        .route("/names", post(add_name::<S>))
```
cloutiertyler, line 616: You cannot set multiple names based on my reading of `add_name`. This should be renamed `set-name` or `add-name`. I believe we should remove the functionality to add multiple names for now tbh.

crates/client-api/src/routes/database.rs
coolreader18: Alright, then I'll switch it over.

crates/client-api/src/routes/database.rs
coolreader18: We have no way to remove names. The old functionality was called "set-name" but in reality it just added a new name without removing the old one, by my reading.

coolreader18:
> This is only relevant to this PR insofar as we may not want to introduce a new `/domain` route if we're going to just basically have `/database/:name_or_identity/set-name` route.

The actual only functionality there is "register_tld". Is that something that just doesn't need to exist anymore? I wasn't sure whether that was being fully removed or not; I figured it can just be merged into publish, perhaps.

gefjon:
> > This is only relevant to this PR insofar as we may not want to introduce a new `/domain` route if we're going to just basically have `/database/:name_or_identity/set-name` route.
> 
> The actual only functionality there is "register_tld". Is that something that just doesn't need to exist anymore? I wasn't sure whether that was being fully removed or not; I figured it can just be merged into publish, perhaps.

TLDs are not functional, and this PR has not caused them to become functional. Notably, using `/`es as separators within URL components is incompatible with Axum's routing (except in the specific case of being in trailing position, I think, which our database names aren't). Possible futures involve us revisiting TLDs with a different separator character, but we can do that in the future in a backwards-compatible way.

If the `/domain` route's only functionality is `register_tld`, then `/domain` should be removed.

cloutiertyler (approved):
Upon further reflection I think that `POST /database/:db/names` is the correct route.

This now LGTM. I'm okay to merge this.

coolreader18:
Merged into the wrong branch :pensive: #2243 is the PR into master.

PR #2236, opened 2025-02-10
Add metrics for index inserts and deletes

jsdt:
# Description of Changes

This adds metrics for index insertions and deletions. It relies on the fact that we don't have sparse indexes.

Note that we don't count index insertions for newly added indexes, which we may want to add in the future.

1

# Testing

Not tested, but seems very low risk, since it mostly uses the row insertion tracking.


joshua-spacetime (approved):
Do we care about bytes inserted/deleted from indexes?

jsdt:
> Do we care about bytes inserted/deleted from indexes?

We might need/want to add that in the future, but I don't think it's worth the extra changes to add right now. Overall it is less important that the bytes written metric, since indexes are totally in memory.

PR #2240, opened 2025-02-10
Implement unknown reducer handling in C# SDK

kazimuth:
# Description of Changes

Codegen side. (Is there an SDK side needed @gefjon ?)

none

0

# Testing

will test with quickstart


kazimuth:
unneeded

PR #2243, opened 2025-02-11
Refactor HTTP api to better adhere to REST guidelines (real)

coolreader18:
# Description of Changes

#2225 but for real (accidentally merged that one into `noa/proper-schema-api` instead of master)


cloutiertyler (approved):
Approving into master

PR #2244, opened 2025-02-11
Remove spacetimedb-core as a dep of cli

coolreader18:
# Description of Changes

This is possible now that standalone is now a separate binary from cli. This lets us separate concerns, and speed up compilation of the cli.

1 - moves some stuff around but otherwise a pretty small change.

# Testing

- [x] No functionality change, tests still pass for codegen for our test modules (for rust, csharp modules).


crates/cli/src/subcommands/generate/mod.rs
```diff
@@ -357,9 +364,9 @@ impl WasmCtx {
         let (mem, env) = Self::mem_env(&mut caller);
 
         // Read `buffer_len`, i.e., the capacity of `buffer` pointed to by `buffer_ptr`.
-        let buffer_len = u32::read_from(mem, buffer_len_ptr)?;
+        let buffer_len = u32::from_le_bytes(deref_slice(mem, buffer_len_ptr, 4)?.try_into().unwrap());
         // Write `buffer` to `sink`.
-        let buffer = mem.deref_slice(buffer_ptr, buffer_len)?;
+        let buffer = deref_slice(mem, buffer_ptr, buffer_len)?;
         env.sink.extend(buffer);
```
mamcx, line 371: I think this should belong to a common method like was with `u32::read_from(mem, buffer_len_ptr)?;` so is easier to keep in sync the logic.

crates/auth/src/identity.rs
```diff
@@ -49,39 +47,33 @@ pub struct IncomingClaims {
 }
 
 impl TryInto<SpacetimeIdentityClaims> for IncomingClaims {
-    type Error = TokenValidationError;
+    type Error = anyhow::Error;
```
mamcx, line 51: Why was removed the `TokenValidationError`?

crates/data-structures/Cargo.toml
```diff
@@ -6,7 +6,7 @@ license-file = "LICENSE"
 description = "Assorted data structures used in spacetimedb"
 
 [features]
-serde = ["dep:serde"]
+serde = ["dep:serde", "hashbrown/serde"]
```
bfops, line 9: why is this feature needed now?

bfops:
I'm broadly in favor of improving compile times. +1 to Mario's questions.

Dumb question - what exactly _is_ spacetimedb-core? At first glance, it seems entirely reasonable for the CLI to depend on a core spacetimedb library.

I don't personally feel equipped to confirm that the changes don't do something meaningful, particularly around wasmtime. Can you add a bit more detail to the `Testing` section about the existing test coverage for `generate`?

crates/cli/src/subcommands/generate/mod.rs
coolreader18: I agree, but the whole point of this is that there's no longer a common crate where it would make sense to put this logic, and it's not like we're gonna want to change the way we read a little-endian u32 -- the code that's being duplicated is very simple.

crates/auth/src/identity.rs
coolreader18: This method never actually returned anything other than its `Other(anyhow::Error)` variant, so it's easier to just return `anyhow::Error`.

crates/cli/Cargo.toml
```diff
@@ -18,9 +18,9 @@ bench = false
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+spacetimedb-auth.workspace = true
 spacetimedb-client-api-messages.workspace = true
-spacetimedb-core.workspace = true
-spacetimedb-data-structures.workspace = true
+spacetimedb-data-structures = { workspace = true, features = ["serde"] }
```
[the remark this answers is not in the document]
coolreader18: It was being enabled by `core`, but now `core` isn't in the build graph.

crates/data-structures/Cargo.toml
coolreader18: Same reason - no longer being enabled by some other crate now that `core` isn't in the build graph.

coolreader18:
`spacetimedb-core` is the core of the *database* implementation, and now that is being pulled in by standalone. The CLI only needs to be able to talk to a database, but it doesn't need to run one itself.

coolreader18:
Also, the wasmtime changes aren't meaningful, it's just copying code.

coolreader18:
I could copy over `Mem` and `MemView` from `spacetimedb_core::host::wasmtime`, but that would mean there's more code and unsafe duplicated overall, and this only uses a small amount of its functionality.

crates/cli/src/subcommands/generate/mod.rs
coolreader18: Ah, wait, I think I see what you mean, sure.

coolreader18:
The internal tests are only because `Cargo.lock` needs regen.

crates/cli/src/api.rs
```diff
@@ -97,6 +96,13 @@ pub struct StmtResultJson<'a> {
     pub stats: StmtStatsJson,
 }
 
+#[derive(Debug, Clone, Deserialize, Default)]
+pub struct StmtStatsJson {
+    pub rows_inserted: u64,
+    pub rows_deleted: u64,
+    pub rows_updated: u64,
+}
```
bfops, line 104: I feel like API types are exactly the kind of thing that should be shared between the sender and receiver code. Do you know why this type is inside `core` instead of `client-api` in the first place? :thinking:

crates/cli/src/subcommands/generate/mod.rs
```diff
@@ -1,11 +1,11 @@
 #![warn(clippy::uninlined_format_args)]
 
+use anyhow::Context;
```
bfops, line 3: I don't really know how to review this file. Do you feel confident that the behavior is identical and/or that the test coverage covers all the cases that this might affect?

crates/cli/src/api.rs
coolreader18: No, I was planning to put it there as a follow-up.

crates/cli/src/subcommands/generate/mod.rs
```diff
@@ -329,19 +329,30 @@ fn extract_descriptions_from_module(module: wasmtime::Module) -> anyhow::Result<
 }
 
 struct WasmCtx {
-    mem: Option<Mem>,
+    mem: Option<wasmtime::Memory>,
```
bfops, line 332: I wouldn't advocate for copying the definition of `Mem` etc from `core`.. but I _would_ advocate for sharing them in e.g. a `memutils` crate or something.

I don't feel strongly about it though. I'm not really equipped to own this part of the CLI either way so :shrug: If you prefer it this way, we can just call it good.

coolreader18: It's like 10 lines of code in total, so I feel confident that it doesn't make a difference - reading an LE u32 from linear memory is an operation that's never going to have to change.

bfops: cool, I'll resolve this then

crates/cli/src/api.rs
coolreader18: Done. Fwiw, there are plenty of other structs that are duplicated between cli and client-api, but those can definitely probably wait until a followup pr.

bfops: ah rip okay

bfops (approved):
okay, this LGTM given the discussions. I'm not sure what to do with @mamcx's remaining comment.

PR #2248, opened 2025-02-11
Only support bearer auth

coolreader18:
# Description of Changes

Follow up to #2181. If a client somehow can send basic auth but not bearer, they can also just use the `?token` param.

Breaks http api - not sure what the counts as. Nothing needed from users besides upgrading, though.

1: this is covered by tests, and all the relevant client sdks have been updated to use bearer auth instead of basic.

# Testing

- [x] I'm confident this is covered by smoketests.


cloutiertyler (approved):
Looks good as planned.

PR #2249, opened 2025-02-11
C# SDK - Emit Event.UnknownTransaction rather than throwing on unknown reducer (#2241)

kazimuth:
# Description of Changes

Same as https://github.com/clockworklabs/SpacetimeDB/pull/2241 but that was mistakenly based on the wrong branch and was not merged into master ;-;

none

0
# Testing

- [x] quickstart
- [x] blackholio


cloutiertyler (approved):
Approved because we already approved.

PR #2259, opened 2025-02-13
Clear DB size gauges when shutting a module down

jsdt:
# Description of Changes

This removes the size gauges for a database when a module is shut down. This also removes the `table_id` from the table-level gauges, and leaves only the human-readable name.

corresponding GitHub label.

1

This has a corresponding private pr: https://github.com/clockworklabs/SpacetimeDBPrivate/pull/1376


gefjon (approved):
Kind of a shame to loose granularity like this, but oh well.

gefjon (approved):

PR #2263, opened 2025-02-14
Upload spacetimedb-update as its own release artifact

coolreader18:
# Description of Changes

This makes it so that the install script doesn't have to do as much.

1

# Testing

- [x] The first commit has the package job enabled

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -7,7 +7,6 @@ on:
     branches:
       - master
       - release/*
-      - noa/package-update-bin-itself
 
 jobs:
   build-cli:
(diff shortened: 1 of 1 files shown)


jdetter (approved):
Tested via https://github.com/clockworklabs/SpacetimeDB/pull/2265 I will update the release checklist to include the changes we've made here. All releases moving forward must contain the archive file (tar for macos/linux and zip for windows) and then also separately the update binary.

For future context: me Noa + Tyler had a quick sync on this to go over whether or not we should have the `spacetimedb-update` binaries as separate artifacts as part of the release. The TLDR is that it's suboptimal for our users but it's more convienent for us on the SpacetimeDB side to directly associate these binaries with a release (and also we're short on time for 1.0) so we have decided to move forward with the changes that Noa has made here. This is tested + working on Windows and Linux and Noa will be testing macOS when they update our homebrew repository.

Thanks @coolreader18 !

PR #2267, opened 2025-02-17
Impose `FilterableValue` in `UniqueColumn::find`

Centril:
# Description of Changes

Fixes https://github.com/clockworklabs/SpacetimeDB/issues/2272.

This PR does the following:

1. Imposes that the sought value column value for `UniqueIndex::find` must be `FilterableValue`.
2. Makes `FilterableValue` a sealed trait so that we may intentionally grow the set and make it un-sealed at a time of our choosing.
3. Adds `#[diagnostic::on_unimplemented(...)]`  to `FilterableValue` for better diagnostics.
4. Adds UI tests to ensure that `.filter(..)` and `.find(...)` impose `FilterableValue`.

The PR does not impose the constraint on table columns themselves.
Perhaps that is what we want to do, as it would allow us to also impose the check on the semantic schema level in the host.
This would be more foolproof.

Yes but its a bug fix.

1

# Testing

A UI test has been added.

### DIFF
--- a/crates/bindings/src/table.rs
+++ b/crates/bindings/src/table.rs
@@ -293,7 +293,7 @@ impl<Tbl: Table, Col: Index + Column<Table = Tbl>> UniqueColumn<Tbl, Col::ColTyp
     #[inline]
     pub fn find(&self, col_val: impl Borrow<Col::ColType>) -> Option<Tbl::Row>
     where
-        Col::ColType: FilterableValue,
+        for<'a> &'a Col::ColType: FilterableValue,
     {
         self._find(col_val.borrow())
     }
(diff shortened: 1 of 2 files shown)


coolreader18 (dismissed):

bfops:
(Which ticket is this tied to?)

cloutiertyler:
> I'm not sure if this is what was intended by the ticket, but if it is, this LGTM.

I think you need to be sure in order to approve the PR.

coolreader18:
FTR: Mazdak said that @cloutiertyler said that this PR implements the intended behavior.

coolreader18:
@gefjon this seems to be your ticket - could you review this PR to confirm if it's what you intended by #2272?

gefjon (approved):
Love the test!

PR #2269, opened 2025-02-18
`bindings-csharp`: use the update ABI

Centril:
# Description of Changes

Fixes `bindings-csharp` to also use the update ABI like the rust `bindings` do.

Technically breaking as the previous C# `Update` functions returned `bool`
and worked if the row was not present, but this is not in line with what the Rust module bindings / the update ABI does.

2

# Testing

The `Update` method is already tested in `modules/sdk-test-cs`.


bfops:
@Centril is this a "meaningful" breakage? (Does this need to roll as part of 1.0?)

rekhoff:
After updating `Module#FFI.verified.cs` with the contents of the generated `Module#FFI.received.cs`, and then merging in changes from `master` and all checks have passed. Reviewing changes, everything makes sense and is clear.

rekhoff (approved):
Code changes make sense, and are clear.

PR #2271, opened 2025-02-18
Send delete updates for unsubscribes

jsdt:
# Description of Changes

We were sending row insert operations in `Unsubscribe` messages, so clients aren't actually removing the rows. This adds a way to invert query results for `Unsubscribe` operations.

This means unsubscribe is currently not removing the data...

cc @gefjon @kazimuth 

1

# Testing

I extended the existing unsubscribe test so that the query actually returns something.

### DIFF
--- a/crates/core/src/subscription/mod.rs
+++ b/crates/core/src/subscription/mod.rs
@@ -63,6 +63,14 @@ where
     Ok((list, n, metrics))
 }
 
+/// When collecting a table update are we inserting or deleting rows?
+/// For unsubscribe operations, we need to delete rows.
+#[derive(Debug, Clone, Copy)]
+pub enum TableUpdateType {
+    Subscribe,
+    Unsubscribe,
+}
+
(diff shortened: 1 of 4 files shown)


crates/core/src/subscription/mod.rs
```diff
@@ -63,21 +63,40 @@ where
...
+            TableUpdateType::Subscribe => QueryUpdate {
+                deletes: empty,
+                inserts: rows,
+            },
+            TableUpdateType::Unsubscribe => QueryUpdate {
+                deletes: rows,
+                inserts: empty,
+            },
+        };
+        // let qu = QueryUpdate { deletes, inserts };
```
joshua-spacetime, line 99: Just remember to remove this before merging

joshua-spacetime (approved):

gefjon (approved):
I haven't looked at the `core/src/subscription` stuff (where the actual bugfix is), but I'm approving the test.

PR #2278, opened 2025-02-18
Add multi-query version of subscribe messages, and dedup queries on the backend.

jsdt:
# Description of Changes

This adds a new version of the websocket API for subscribing that takes multiple queries. This just has the backend changes.

Most of the changes are fairly mechanical, other than the changes in `module_subscription_manager`. The `ClientInfo` struct now tracks a set of query hashes for each subscription id, and it tracks a reference count for each query hash that it is subscribed to.

When adding and removing subscriptions, we use the reference counts to figure out which queries are actually being remove and added, and the `SubscriptionManager` functions return those queries, so that the code in `module_subscription_manager.rs` can figure out which queries to evaluate (and send to clients).

This should just be adding new types, but once all the client SDKs are updated, we should probably make a breaking change to remove the old single versions.

2.171

# Testing

This has a few basic unit tests.

### DIFF
--- a/crates/core/src/client/messages.rs
+++ b/crates/core/src/client/messages.rs
@@ -91,6 +91,8 @@ impl SerializableMessage {
                 SubscriptionResult::Subscribe(_) => Some(WorkloadType::Subscribe),
                 SubscriptionResult::Unsubscribe(_) => Some(WorkloadType::Unsubscribe),
                 SubscriptionResult::Error(_) => None,
+                SubscriptionResult::SubscribeMulti(_) => Some(WorkloadType::Subscribe),
+                SubscriptionResult::UnsubscribeMulti(_) => Some(WorkloadType::Unsubscribe),
             },
             Self::TxUpdate(_) => Some(WorkloadType::Update),
             Self::Identity(_) => None,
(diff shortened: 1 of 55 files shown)


crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -253,6 +303,186 @@ impl ModuleSubscriptions {
         Ok(())
     }
 
+    /// Remove a subscription for a single query.
```
joshua-spacetime, line 306:
```suggestion
    /// Remove a set of queries from a client's subscription
```

crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -253,6 +303,186 @@ impl ModuleSubscriptions {
...
+
+            WORKER_METRICS
+                .subscription_queries
+                .with_label_values(&self.relational_db.database_identity())
+                .set(subscriptions.num_unique_queries() as i64);
+            new_queries
+        };
+
+        let (update, metrics) =
+            self.evaluate_queries(sender.clone(), &queries, &tx, &auth, TableUpdateType::Subscribe)?;
```
joshua-spacetime, line 460: What would happen in this case?

1. Client subscribes to new query "A"
2. Client unsubscribes to "A"
3. Client receives UnsubscribeMultiApplied with rows that don't exist in client cache
4. Client receives SubscriptionMultiApplied message

joshua-spacetime, line 460: Don't we want to back out the subscription manager changes if there's an error here?

jsdt: At the time I had it in my head that the connection would just get killed if we errored there, but I should verify that.

jsdt: From my understanding of [ws_client_actor_inner](https://github.com/clockworklabs/SpacetimeDB/blob/4665c5243c267c2a3f9af949473b4b0452f161fc/crates/client-api/src/routes/subscribe.rs#L196), we only handle one request at a time, and we do it in FIFO order, so this shouldn't be possible. If it did happen, I'm not sure whether the client would hit an error and kill itself, or potentially get into a state with a stuck subscription.

crates/core/src/client/messages.rs
```diff
@@ -400,6 +390,30 @@ impl ToProtocol for SubscriptionMessage {
...
+                            total_host_execution_duration_micros,
+                            request_id,
+                            query_id,
+                            update: data,
+                        }
+                        .into(),
+                    ),
+                }
+            }
+            SubscriptionResult::UnsubscribeMulti(subscription_data) => todo!(),
```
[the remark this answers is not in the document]
jsdt: I forgot to push that part.

crates/core/src/subscription/module_subscription_actor.rs
joshua-spacetime: I don't believe it's FIFO. It probably was at one point, but once we added multi-threaded subscription evaluation, that certainly was no longer the case.

joshua-spacetime: > I'm not sure whether the client would hit an error and kill itself, or potentially get into a state with a stuck subscription.

I believe the client cache treats duplicate inserts as well as deletes for rows that don't exist as an error. At least it did before 1.0, and I don't think we ever changed that behavior. I'm not sure how it handles the error though.

gefjon (approved):

PR #2281, opened 2025-02-19
Rust client SDK: send `SubscribeMulti` and `UnsubscribeMulti`

gefjon:
# Description of Changes

Based on #2278 .

This commit modifies the Rust client SDK to use the new setwise subscription API.

This includes:
- Changing `SubscriptionBuilder::subscribe` to accept any implementor of `IntoQueries`, including single strings and also arrays of strings, for various definitions of "string" and "array." And fiddling a bit with the implementations of `IntoQueries`.
- Sending the `SubscribeMulti` and `UnsubscribeMulti` websocket messages instead of `Subscribe` and `Unsubscribe`.
- Stubbing the handling of `SubscribeApplied` and `UnsubscribeApplied` websocket messages, since those should now be unreachable.
- Handling the `SubscribeMultiApplied` and `UnsubscribeMultiApplied` messages instead.
- Simplifying the `quickstart-chat` and `test-client` clients to just subscribe to one big set, where previously they had a kind of wonky semaphore-based scheme to wait for all queries to be applied.

## New syntax

You can now do:

```rust
ctx.subscription_builder()
  .subscribe(["SELECT * FROM user", "SELECT * FROM client"]);
```

The array passed here may be a `[T; N]` (fixed-size array), a `&[T]` (slice), `Vec<T>` (vec) for a few different string-like `T`s, or a `Box<[Box<str>]>`.

You can also still subscribe to a single query without wrapping it in an array, like:

```rust
ctx.subscription_builder()
  .subscribe("SELECT * FROM user");
```

The string here may be `&str`, `String` or `Box<str>`. These are the same "string-like" types which can be put in arrays as above.

I have not been super careful about the changed implementations of `IntoQueries`, so minor incompatibilities are possible between our most recent 1.0-rc release and this patch in terms of valid types to pass to `SubscriptionBuilder::subscribe`. Common cases, like `&str` and `String`, still work.

1

# Testing

- [x] I believe automated testing is sufficient.


gefjon:
Attempting to rebase has proved once more to have been a mistake. Merging seems to have worked fine, though. Sorry for the confusion and force-pushes!

PR #2285, opened 2025-02-20
Add support for COUNT

joshua-spacetime:
# Description of Changes

Support count queries such as `SELECT COUNT(*) AS n FROM t`

None

1

Note, there is a better way of supporting agg, but that required a larger diff, so I left a descriptive TODO in the code instead.

# Testing

Parsing, optimization, and execution tests added

### DIFF
--- a/crates/execution/src/pipelined.rs
+++ b/crates/execution/src/pipelined.rs
@@ -5,12 +5,14 @@ use std::{
 
 use anyhow::{anyhow, Result};
 use itertools::Either;
+use spacetimedb_expr::expr::AggType;
 use spacetimedb_lib::{metrics::ExecutionMetrics, query::Delta, sats::size_of::SizeOf, AlgebraicValue, ProductValue};
 use spacetimedb_physical_plan::plan::{
     HashJoin, IxJoin, IxScan, PhysicalExpr, PhysicalPlan, ProjectField, ProjectListPlan, ProjectPlan, Sarg, Semi,
     TableScan, TupleField,
 };
 use spacetimedb_primitives::{ColId, IndexId, TableId};
+use spacetimedb_sats::product;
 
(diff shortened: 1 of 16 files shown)


crates/execution/src/pipelined.rs
```diff
@@ -41,20 +47,47 @@ impl ProjectListExecutor {
...
+            Self::Limit(plan, limit) => {
+                plan.execute(tx, metrics, &mut |row| {
+                    n += 1;
+                    if n <= *limit as usize {
+                        f(row)?;
+                    }
+                    Ok(())
+                })?;
+            }
+            // TODO: This is a hack that needs to be removed.
```
joshua-spacetime, line 76: Important TODO

crates/physical-plan/src/plan.rs
```diff
@@ -98,10 +98,20 @@ impl ProjectPlan {
 /// ```sql
 /// select t.a, s.b from t join s ...
 /// ```
+///
+/// TODO: LIMIT and COUNT were added rather hastily.
```
joshua-spacetime, line 102: Another important TODO

crates/sql-parser/src/parser/errors.rs
```diff
@@ -36,6 +39,10 @@ pub enum SqlUnsupported {
     Projection(SelectItem),
     #[error("Unsupported projection expression: {0}")]
     ProjectionExpr(Expr),
+    #[error("Unsupported aggregate function: {0}")]
+    Aggregate(Function),
+    #[error("Aggregate expressions must have column aliases")]
+    AggregateWithoutAlias,
```
joshua-spacetime, line 45: This constraint may be a bit too restrictive, but I thought it would be better for when we add support for subqueries.

mamcx (approved):
LGTM

PR #2286, opened 2025-02-20
Rust SDK: Deduplicate rows accross queries

Centril:
# Description of Changes

Implements row deduplication on the Rust SDK side via reference counting.
A new type/trait `AppliedDiff` is added which records delete/insert/update events.
This type is then used to call callbacks.
Calculating updates is now done after insert/delete events have been calculated, and is done purely on `(Table)AppliedDiff`.
As before, this work is requested by specific tables in the codegen, but this is done last in applying the diff rather than than during parsing.

None.

3, fairly complicated code and in a central location.

# Testing

A test `(exec_)row_deduplication` is added that ensures that duplicate callbacks aren't received.
The test also ensures that the right insert/delete/update events happen as we expect.

### DIFF
--- a/crates/sdk/src/client_cache.rs
+++ b/crates/sdk/src/client_cache.rs
@@ -183,6 +183,7 @@ impl<Row: Clone + Send + Sync + 'static> TableCache<Row> {
         // Extract the entry and decrement the `ref_count`.
         // Only create a delete event if `ref_count = 0`.
         let Entry::Occupied(mut entry) = self.entries.entry(delete.bsatn.clone()) else {
+            // We're guaranteed to never hit this as long as we apply inserts before deletes.
             unreachable!("a delete update should correspond to an existing row in the table cache");
         };
         let ref_count = &mut entry.get_mut().ref_count;
@@ -233,6 +234,8 @@ impl<Row: Clone + Send + Sync + 'static> TableCache<Row> {
     /// and populate the `update_*` fields.
     fn apply_diff<'r>(&mut self, diff: &'r TableUpdate<Row>) -> TableAppliedDiff<'r, Row> {
         // Apply all inserts and collect all `ref_count: 0 -> 1` events.
+        // Inserts must be applied before deletes to avoid the panic in `handle_delete`
+        // and to avoid duplicate index insertion errors.
         let mut insert_events = <_>::default();
         for insert in &diff.inserts {
             self.handle_insert(&mut insert_events, insert);


crates/sdk/src/client_cache.rs
```diff
@@ -50,37 +67,166 @@ impl<Row> Default for TableCache<Row> {
...
-            self.entries.remove(&delete.bsatn);
-            for index in self.unique_indices.values_mut() {
-                index.remove_row(&delete.row);
-            }
+/// The diff result of applying [`TableUpdate`] to a [`TableCache`].
+pub struct TableAppliedDiff<'r, Row> {
+    deletes: RowEventMap<'r, Row>,
+    inserts: RowEventMap<'r, Row>,
+    update_deletes: Vec<&'r Row>,
+    update_inserts: Vec<&'r Row>,
```
gefjon, line 77: Please add a quick comment on these fields (or the struct itself) which hits:
- Whether `deletes` and `update_deletes` are disjoint, and the same for `inserts` and `update_inserts`.
- Whether these are unique- or multi-sets.
- The fact that (if I'm reading correctly) `update_deletes` and `update_inserts` will be empty upon construction, and will be populated by `with_updates_by_pk` or `derive_updates`.

crates/sdk/src/client_cache.rs
```diff
@@ -50,37 +67,166 @@ impl<Row> Default for TableCache<Row> {
...
+        });
+        entry.ref_count += 1;
+
+        // The host never sends us a delete-insert pair for the same row R,
+        // so we don't need to remove it.
+        // However, for tests, we defensively assert this is the case.
+        debug_assert!(!_deletes.contains_key(&*insert.bsatn));
+    }
+
+    /// Apply all the deletes, inserts and updates recorded in `diff`.
```
gefjon, line 198:
```suggestion
    /// Apply all the deletes and inserts recorded in `diff`.
    /// Return a [`TableAppliedDiff`] which encodes the actual changes to rows present or deleted
    /// after taking into account rows' refcounts.
    ///
    /// The resulting [`TableAppliedDiff`] will contain only deletes and inserts.
    /// Its `update_deletes` and `update_inserts` fields will be empty.
    /// The caller should use [`TableAppliedDiff::with_updates_by_pk`] to merge delete/insert pairs
    /// and populate the `update_*` fields.
```

gefjon (approved):
Minor comment-ey suggestions. Needs corresponding private PR to re-run codegen. Looks good, and I appreciate the test!

crates/sdk/src/client_cache.rs
```diff
@@ -50,37 +67,166 @@ impl<Row> Default for TableCache<Row> {
...
+            // First time inserting this row, so let's add an insertion event.
+            inserts.insert(&insert.bsatn, &insert.row);
+            RowEntry {
+                row: insert.row.clone(),
+                ref_count: 0,
             }
+        });
+        entry.ref_count += 1;
+
+        // The host never sends us a delete-insert pair for the same row R,
```
joshua-spacetime, line 192: Note, as of right now, the host could in fact send a delete-insert pair for the same row, because we don't remove such pairs across queries. For example:
```
subscribe: R ⋉ S
subscribe: R ⋉ T

r in R joins with s in S
r in R joins with t in T

insert s => insert r
delete t => delete r
```

I also don't think we want to do that on the host. This was the reason we decided to return duplicate rows instead of rows with multiplicities. So that we didn't have to do that computation on the server.

kazimuth: Good to know.

It also can send mismatched insert-delete counts, I believe:
```
subscribe: SELECT * FROM cow
subscribe: SELECT * FROM cow WHERE cow.color = brown

server update: ("Bessie", brown) -> ("Bessie", purple)
```

You should get two deletes and one insert for Bessie, right?

joshua-spacetime: That is correct

crates/sdk/src/client_cache.rs
```diff
@@ -50,37 +67,200 @@ impl<Row> Default for TableCache<Row> {
...
+            .zip(self.update_inserts.iter().copied())
+    }
+}
+
+impl<Row: Clone + Send + Sync + 'static> TableCache<Row> {
+    fn handle_delete<'r>(&mut self, events: &mut RowEventMap<'r, Row>, delete: &'r WithBsatn<Row>) {
+        // Extract the entry and decrement the `ref_count`.
+        // Only create a delete event if `ref_count = 0`.
+        let Entry::Occupied(mut entry) = self.entries.entry(delete.bsatn.clone()) else {
+            unreachable!("a delete update should correspond to an existing row in the table cache");
```
joshua-spacetime, line 180: I think this actually depends on the order in which we're applying the inserts and deletes. It's correct if we apply all the inserts first. But if we apply deletes first, we may hit this at some intermediate point before we have applied the diff fully.

joshua-spacetime: For a concrete example, take `R ⋉ S`, and the following scenario:

A transaction:

1. Inserts a new row `r` into `R`
2. Inserts a new row `s1` into `S`
3. Deletes a row `s2` from `S`

where `s1` and `s2` both join with `r`.

Because the server is evaluating multiple queries in order to compute the delta, and because it only has access to the state after the tx commits, the queries may over count rows.

In this specific example, if the server doesn't deduplicate, it will return 3 copies of `r` to insert and 2 copies to delete. If we apply the deletes first, because this is a new row, we'll hit this line when applying the first delete.

If we apply inserts first, we're guaranteed to never be in a situation where we try to delete a non-existent row from the cache.

crates/sdk/src/client_cache.rs
```diff
@@ -50,37 +68,200 @@ impl<Row> Default for TableCache<Row> {
...
+    fn handle_delete<'r>(
+        &mut self,
+        inserts: &mut RowEventMap<'_, Row>,
+        deletes: &mut RowEventMap<'r, Row>,
+        delete: &'r WithBsatn<Row>,
+    ) {
+        // Extract the entry and decrement the `ref_count`.
+        // Only create a delete event if `ref_count = 0`.
+        let Entry::Occupied(mut entry) = self.entries.entry(delete.bsatn.clone()) else {
+            unreachable!("a delete update should correspond to an existing row in the table cache");
```
joshua-spacetime, line 186: Can you add a comment here that says we are guaranteed never to hit this in any situation as long as we apply inserts before deletes?

crates/sdk/src/client_cache.rs
```diff
@@ -50,37 +68,200 @@ impl<Row> Default for TableCache<Row> {
...
+    /// after taking into account rows' refcounts.
+    ///
+    /// The resulting [`TableAppliedDiff`] will contain only deletes and inserts.
+    /// Its `update_deletes` and `update_inserts` fields will be empty.
+    /// The caller should use [`TableAppliedDiff::with_updates_by_pk`] to merge delete/insert pairs
+    /// and populate the `update_*` fields.
+    fn apply_diff<'r>(&mut self, diff: &'r TableUpdate<Row>) -> TableAppliedDiff<'r, Row> {
+        // Apply all inserts and collect all `ref_count: 0 -> 1` events.
+        let mut insert_events = <_>::default();
+        for insert in &diff.inserts {
```
joshua-spacetime, line 237: And could you put a similar comment here to make sure no one accidentally reorders them in the future?

joshua-spacetime (approved):

PR #2294, opened 2025-02-21
Fix self-replace on windows

coolreader18:
# Description of Changes

Windows doesn't let you overwrite an exe file that's currently running. Thankfully, the [`self-update`](https://docs.rs/self-update) crate has an implementation for this.

1: can't get more broken that it already is! and we're just using someone else's code :)

# Testing

- [ ] Need to test on windows.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4482,6 +4482,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "self-replace"
+version = "1.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


jdetter (changes requested):
I don't believe this worked when I tried it, I got the following error:

```
PS C:\Users\Boppy> spacetime version upgrade
  Installing v1.0.0: done!
Error: Cannot create a file when that file already exists. (os error 183)
```

jdetter (approved):
Tested this with windows against a local server (not against the Github releases). I'll test this again when we go live with the release so we can make sure the update flow is working correctly against Github.

PR #2295, opened 2025-02-21
fix: join order rewrite rule

joshua-spacetime:
# Description of Changes

After swapping the left and right hand sides of a delta join, we were not updating the left and right hand side join fields.

This resulted in a stack overflow when applying a different rewrite rule for translating named arguments into positional ones.

As part of this patch I also modified the rewrite that stack overflowed to error if an invariant is not met.

None

1

# Testing

Added basic regression test that stack overflowed before this change.

Added an incremental evaluation test similar to the ones we currently have. The current tests we have did not exercise this particular rewrite (left vs right join), and so did not catch this bug.

### DIFF
--- a/crates/physical-plan/src/rules.rs
+++ b/crates/physical-plan/src/rules.rs
@@ -57,60 +57,28 @@ pub(crate) struct ComputePositions;
 
 impl RewriteRule for ComputePositions {
     type Plan = PhysicalPlan;
-    type Info = Label;
+    type Info = (Label, usize);
 
     fn matches(plan: &Self::Plan) -> Option<Self::Info> {
         match plan {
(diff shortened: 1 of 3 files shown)


crates/physical-plan/src/rules.rs
```diff
@@ -969,7 +969,9 @@ impl RewriteRule for ReorderDeltaJoinRhs {
                 HashJoin {
                     lhs: join.rhs,
                     rhs: join.lhs,
-                    ..join
+                    lhs_field: join.rhs_field,
```
mamcx, line 972: Add here a comment about why this is done

mamcx (approved):
LGTM

PR #2297, opened 2025-02-23
Added in a SQL execution time calculation

cloutiertyler:
# Description of Changes

Now returning the execution duration of SQL queries. The changes include adding timing measurements and updating the API response structure to include the total duration of SQL execution.

Closes https://github.com/clockworklabs/spacetime-web/issues/478#issue-2645325773

AFAIK this should not require a change to the internal stuff.

This is an additive change to the SQL endpoint API.

1

# Testing

I ran and tested this locally using a local build of SpacetimeDB and checked that it worked against my local build of the website.

### DIFF
--- a/crates/client-api/src/lib.rs
+++ b/crates/client-api/src/lib.rs
@@ -84,6 +84,10 @@ impl Host {
                     // We need a header for query results
                     let mut header = vec![];
 
+                    let sql_start = std::time::Instant::now();
+                    let sql_span =
+                        tracing::trace_span!("execute_sql", total_duration = tracing::field::Empty,).entered();
+
                     let rows = sql::execute::run(
                         // Returns an empty result set for mutations
                         db,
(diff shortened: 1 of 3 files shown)


crates/client-api/src/lib.rs
```diff
@@ -84,6 +84,10 @@ impl Host {
                     // We need a header for query results
                     let mut header = vec![];
 
+                    let sql_start = std::time::Instant::now();
+                    let sql_span =
+                        tracing::trace_span!("execute_sql", total_duration = tracing::field::Empty,).entered();
+
                     let rows = sql::execute::run(
```
joshua-spacetime, line 91: Optional: You'll get a more accurate time if you grab it off the tx in `sql::execute::run`

cloutiertyler: @joshua-spacetime How do I do that? Could you possibly make that change?

joshua-spacetime (approved):

PR #2299, opened 2025-02-24
sdk: Basic connection metrics

kim:
Instruments the websocket loop to track number and size of received
messages. Mirrors the metrics we have on the sending side.

Motivated by having observed clients that permanently return stale data,
even though both ends of the connection appear to consider the
connection healthy.

There may be other interesting metrics (such as the size of the client
cache), but there is no immediate need, so we're deferring this to a
future addition.

Another future change may be to actively terminate the connection if no
data (including PING frames) is received within a timeout.

1

# Testing

A private companion PR will hook this up to a metrics registry, and verify that
the numbers can be read via the metrics endpoint.


gefjon (commented):
- I'd like `CLIENT_METRICS` to be exported as `spacetimedb_sdk::unstable::CLIENT_METRICS`, not `spacetimedb_sdk::CLIENT_METRICS`. I don't yet feel ready to commit to this as a stable interface.
- Why are the metrics in `metrics/mod.rs` rather than `metrics.rs`? No other files in the SDK get their own subdirectory in this way, and there aren't any other files in there beyond the `mod.rs`.

gefjon:
> Another future change may be to actively terminate the connection if no data (including PING frames) is received within a timeout.

I would expect that we should pair this with having the client send PINGs and expect PONGs in return, [as described in the RFC](https://datatracker.ietf.org/doc/html/rfc6455#section-5.5.2), rather than just relying on the server PINGing.

kim:
> I would expect that we should pair this with having the client send PINGs and expect PONGs in return

Potentially. Keep-Alive traffic is easy to amplify, while the current server-side ping interval is likely to be too high for otherwise idle connections. We should probably make this adaptive, i.e. ping only when idle.

PR #2302, opened 2025-02-24
`MutTxId::update`: consider deleted commited row when replacing tx row

### LINKED ISSUE #2296: `update` host call produces TXes containing no-op insert/delete pairs in sequences of delete->insert->update
1. Define a table `foo: { #[unique] a, b }`
2. Start with a table with row `A: { a: 0, b: 0 }` present.
3. Within a single transaction:
  1. Delete `A: { a: 0, b: 0 }`.
  2. Insert `B: {a: 0, b: 1 }`. This row is not equal to `A`, but has the same value in the unique column.
  3.  Update `foo` by `a` to `A: { a: 0, b: 0 }`. This row has the same unique val as `B`, and so is an update, and also is equal to the `A` which was deleted in 3.1, part of this TX.

The TX in 3 *should* be empty, but currently it will result in:
```
inserts:
  foo:
    - A: { a: 0, b: 0 }
deletes:
  foo:
    - A: { a: 0, b: 0 }
```

Commitlog replay cannot cope with such a TX, and fails.

DoD: A regression test runs a transaction like the above and asserts that the resulting `TxData` is empty. Manual or automated integration testing shows that we can run a module that executes the above transaction, then restart and replay its commitlog.

Centril:
# Description of Changes

Fixes #2296.

In `MutTxId::update`: If there was a deleted committed row (`row`), this fact is recorded. Later, when updating a `row_prime`, resident in the tx state and sharing a `key` with `row`, if `row_prime` it's to be replaced with `row`'s contents, we revert the insertion of the and instead un-delete `row`.

None

2?

# Testing

- A regression test `test_update_brings_back_deleted_commit_row_repro_2296` is added.
- The repro in https://github.com/clockworklabs/SpacetimeDB/issues/2296#issuecomment-2676480573 was tested manually against the new impl and the database was successfully restarted.

### DIFF
--- a/crates/core/src/db/datastore/locking_tx_datastore/datastore.rs
+++ b/crates/core/src/db/datastore/locking_tx_datastore/datastore.rs
@@ -2348,7 +2348,7 @@ mod tests {
         // We should now have a committed row `row` marked as deleted and a row `row_prime`.
         // These share `KEY`.
         insert(&datastore, &mut tx, table_id, row_prime)?;
-        // 3. update `row` -> `row_prime`.
+        // 3. update `row_prime` -> `row`.
         // Because `row` exists in the committed state but was marked as deleted,
         // it should be undeleted, without a new row being added to the tx state.
         let (_, row_ref) = update(&datastore, &mut tx, table_id, index_id, row)?;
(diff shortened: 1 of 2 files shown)


crates/core/src/db/datastore/locking_tx_datastore/datastore.rs
```diff
@@ -2283,6 +2284,85 @@ mod tests {
...
+        // Start a new transaction where we:
+        let mut tx = begin_mut_tx(&datastore);
+        // 1. delete the row.
+        let row_to_del = find_row_by_key(&tx, KEY);
+        datastore.delete_mut_tx(&mut tx, table_id, row_to_del.iter().copied());
+        // 2. insert a new row with the same key as the one we deleted but different extra field.
+        // We should now have a committed row `row` marked as deleted and a row `row_prime`.
+        // These share `KEY`.
+        insert(&datastore, &mut tx, table_id, row_prime)?;
+        // 3. update `row` -> `row_prime`.
```
gefjon, line 2351:
```suggestion
        // 3. update `row_prime` -> `row`.
```

PR #2309, opened 2025-02-26
sdk: Client ping on idle connections

kim:
Make the websocket loop send a `Ping` frame when the connection is idle for a while. "idle" includes missing server-side pings.

Writing to the socket will detect if the remote end went away, but FIN / RST went missing for some reason. This is not detected by only reading.

--- 

I haven't touched the scary server-side websocket loop, but we may consider to also make it send pings only if it doesn't have anything else to send.

2

# Testing

The situation can be approximated by inducing a 100% packet loss scenario.
This can be done in various ways. A convenient one is `docker compose disconnect`, which disables the virtual interface of a container.

Without this patch, you'll see the server timing out the connection after a while (as it did not receive pong), while the client keeps thinking that something may eventually arrive.
With this patch, both ends eventually close the connection.

### DIFF
--- a/crates/sdk/src/websocket.rs
+++ b/crates/sdk/src/websocket.rs
@@ -358,6 +358,9 @@ impl WsConnection {
                         log::trace!("received ping");
                         idle = false;
                         record_metrics(payload.len());
+                        // No need to explicitly respond with a `Pong`,
+                        // as tungstenite handles this automatically.
+                        // See [https://github.com/snapview/tokio-tungstenite/issues/88].
                     },
 
                     Ok(Some(WebSocketMessage::Pong(payload))) => {
--- a/crates/sdk/src/websocket.rs
+++ b/crates/sdk/src/websocket.rs
@@ -2,7 +2,9 @@
 //!
 //! This module is internal, and may incompatibly change without warning.
 
+use std::mem;
 use std::sync::Arc;
+use std::time::Duration;
 
 use bytes::Bytes;
 use futures::{SinkExt, StreamExt as _, TryStreamExt};
@@ -16,6 +18,7 @@ use spacetimedb_client_api_messages::websocket::{ClientMessage, ServerMessage};
 use spacetimedb_lib::{bsatn, ConnectionId};
 use thiserror::Error;
 use tokio::task::JoinHandle;
+use tokio::time::Instant;
 use tokio::{net::TcpStream, runtime};
 use tokio_tungstenite::{
     connect_async_with_config,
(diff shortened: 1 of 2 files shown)


crates/sdk/src/websocket.rs
```diff
@@ -287,18 +290,57 @@ impl WsConnection {
...
-                    Err(e) => Self::maybe_log_error::<(), _>(
-                        "Error reading message from read WebSocket stream",
-                        Err(e),
-                    ),
+                    Err(e) => {
+                        Self::maybe_log_error::<(), _>(
+                            "Error reading message from read WebSocket stream",
+                            Err(e),
+                        );
+                        break;
```
kim, line 339: None of the other `Error` variants indicate a particularly recoverable situation, so I added this `exit`.

crates/sdk/src/websocket.rs
```diff
@@ -313,15 +355,42 @@ impl WsConnection {
...
                     Ok(Some(other)) => {
                         log::warn!("Unexpected WebSocket message {:?}", other);
+                        idle = false;
                         record_metrics(other.len());
                     },
                 },
 
+                _ = idle_timeout_interval.tick() => {
+                    if mem::replace(&mut idle, true) {
+                        if mem::take(&mut want_pong) {
```
gefjon, line 379: Why are we `mem::take`ing here? It looks to me like this is equivalent to just reading `want_pong` in all cases.

- `want_pong == true`: `take` and replace with `false`. Break out of the loop and never read from `want_pong` again.
- `want_pong == false`: `take` is a no-op. Set `want_pong` to `true` later in this block. Never read the `false` value installed by `mem::take`.

It seems like there are two possible simplifications to this code:
- `mem::replace(&mut want_pong, true)` here, and remove the `want_pong = true` assignment on line 390.
- Just branch on `want_pong` with `if want_pong {` here, and leave the assignment later on.

crates/sdk/src/websocket.rs
```diff
@@ -313,15 +355,42 @@ impl WsConnection {
...
+                            log::warn!("Connection timed out");
+                            break;
+                        }
+
+                        log::trace!("sending client ping");
+                        Self::maybe_log_error(
+                            "Error sending ping",
+                            self.sock.send(WebSocketMessage::Ping(vec![])).await,
+                        );
+                        want_pong = true;
```
gefjon, line 390: Do we want to break the loop if we fail to send a ping?

kim: Good point. I think this applies to all send attempts, though, so I'm adding it to the outgoing branch too.

gefjon: Sounds good.

gefjon (approved):

gefjon (changes requested):
Sorry, just realized: the host doesn't actually respond `Pong` to client `Ping`s. `crates/client-api/src/routes/subscribe.rs` has:

```rust
            Item::Message(ClientMessage::Ping(_message)) => {
                log::trace!("Received ping from client {}", client.id);
                // TODO: should we respond with a `Pong`?
            }
```

I don't think we can merge this until we complete that TODO comment. I'm worried that as-is, this patch will cause spurious disconnects.

gefjon (approved):
Nope, I'm wrong. Per [this discussion](https://github.com/snapview/tokio-tungstenite/issues/88), Tungstenite responds Pong to Pings automatically.

PR #2312, opened 2025-02-26
Add dockerfile for cli

coolreader18:
# Description of Changes

This just curls & runs the install script on a blank image. Technically won't work until the new script is up on the website.

1

# Testing

I've tested a little bit and will test more once this is merged and I can download the artifacts built from CI (my system has too new a glibc version :cry: )

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,20 @@
+FROM rust:bookworm
+
+RUN curl -sSfLO https://packages.microsoft.com/config/debian/12/packages-microsoft-prod.deb && \
+      dpkg -i packages-microsoft-prod.deb && \
+      rm packages-microsoft-prod.deb && \
+      apt-get update && \
+      apt-get install -y dotnet-sdk-9.0 binaryen && \
+      dotnet workload install wasi-experimental
+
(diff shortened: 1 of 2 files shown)


gefjon:
I would expect that we want to have other things also present inside this docker. E.g.:

- Rustup, cargo, rustc, the `wasm32-unknown-unknown` target, and everything else one needs to compile and publish a Rust module.
- Dotnet, the dotnet WASI SDK, and whatever else one needs to compile and publish a C# module.
- `wasm-opt`.

Testing I would expect to be done:

- One can `spacetime start` using the docker and connect an external client (e.g. a web browser) to the running host.
- One can bind-mount a module project directory into the docker and then `spacetime publish` it, both locally and to a remote host.
- One can use `spacetime call` and `spacetime sql` from the docker, both locally and to a remote host.

Bonus points:

- One can bind-mount a Rust, C# or TypeScript client project directory into the docker and then compile and run it, connecting to either a local or a remote host.

gefjon (approved):

coolreader18:
I'll test this further once this is merged and I can download the artifacts built from CI on master (my system has too new a glibc version :cry:)

PR #2317, opened 2025-02-27
Implements the `PUT /v1/database/:database/names` endpoint

cloutiertyler:
# Description of Changes

This implements a way to set all the names for a database at once. There was no way to remove a name from a database, making it impossible for the website to implement needed UI.

The way I'm implemented this is not ideal, but it's also not bad. Notably, the operation to set the names is not atomic and may partially succeed while returning an error code.

This adds a new HTTP route.

2 (not trivial, but additive and small)

# Testing

- [x] I added additional unit testing to check that this functionality works in standalone.

### DIFF
--- a/crates/standalone/src/control_db/tests.rs
+++ b/crates/standalone/src/control_db/tests.rs
@@ -81,7 +81,7 @@ fn test_domain() -> anyhow::Result<()> {
         reverse_lookup.first().map(ToString::to_string),
         Some(domain.to_string())
     );
-    assert_eq!(reverse_lookup, vec![domain]);
+    assert_eq!(reverse_lookup, vec![domain.clone()]);
 
     // We can remove the domain records for Alice's database
     let deleted = cdb.spacetime_delete_domains(&addr);


coolreader18 (approved):
Looks reasonable to me. If setting domain 3 out of 5 fails, it'll be left in a weird state, but that seems more or less fine for now.

kim:
What speaks against making it atomic? A `replace_names` method doesn't seem hard to implement, so as long as we impose the restriction that the name and database owner must be the same.

cloutiertyler:
> What speaks against making it atomic? A `replace_names` method doesn't seem hard to implement, so as long as we impose the restriction that the name and database owner must be the same.

Nothing, other than I was trying to hack it together in the shortest amount of time and I would have had to deal with Sled transactions. Not opposed to it at all. There's just not much time.

cloutiertyler:
Closing in favor of https://github.com/clockworklabs/SpacetimeDB/pull/2325

PR #2318, opened 2025-02-27
Minor updates to the Rust client quickstart

gefjon:
# Description of Changes

Revealed during John's review of https://github.com/clockworklabs/spacetime-docs/pull/162 . Apparently this code doesn't get linted in CI, or something?

corresponding GitHub label.

1

# Testing

- [x] `cargo clippy`.
- [x] I ran through all of the actual behavior yesterday, and none the code has changed in a way that would affect that.


jdetter (approved):
Tested against latest spacetime CLI. Builds and runs with no errors or warnings 👍

Issue #1386, opened 2024-06-11
Inherent method for `::delete` should be generated for TableTypes


cloutiertyler:

gefjon:
Unassigning myself because I have too many 0.12 tickets.

cloutiertyler:
@gefjon I'm readding you here because this one is relevant to the API stability proposal and I'm going to move you off of some other tickets.

gefjon:
My proposal as it is now supersedes this ticket; this is no longer meaningful in the world where we do not impose a 1-to-1 equivalence between tables and types. Will leave this open in case the proposal changes significantly, but currently I expect to close this issue as stale within the 0.12 window.

PR #2319, opened 2025-02-27
Add missing LICENSE

mamcx:
# Description of Changes

Adding `LICENSE` files everywhere is missing. Closes the private ticket [#1386](https://github.com/clockworklabs/SpacetimeDBPrivate/issues/1386)

0

# Testing

Run 

```bash
find . -type f -name "README.md" | while read -r cargo; do
  dir=$(dirname "$cargo")
  if [ -f "$dir/LICENSE" ] || [ -f "$dir/LICENSE.txt" ]; then
  else
    echo "$cargo is missing LICENSE"
  fi
done

find . -type f -name "README" | while read -r cargo; do
  dir=$(dirname "$cargo")
  if [ -f "$dir/LICENSE" ] || [ -f "$dir/LICENSE.txt" ]; then
  else
    echo "$cargo is missing LICENSE"
  fi
done

find . -type f -name "Cargo.toml" | while read -r cargo; do
  if grep -q "license-file" "$cargo"; then
    :
  else
    echo "$cargo is missing license-file"
  fi
done
``` to find them, plus manual inspection.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,21 @@
+FROM rust:bookworm
+
+RUN curl -sSfLO https://packages.microsoft.com/config/debian/12/packages-microsoft-prod.deb && \
+      dpkg -i packages-microsoft-prod.deb && \
+      rm packages-microsoft-prod.deb && \
+      apt-get update && \
+      apt-get install -y dotnet-sdk-8.0 binaryen && \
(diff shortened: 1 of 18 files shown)


gefjon (approved):
Approving as code owner, not as someone with any legal or procedural knowledge.

PR #2322, opened 2025-02-27
Update `insert` docs to reflect actual behavior on set-semantic duplicates

gefjon:
# Description of Changes

Also, remove a paragraph from `delete` docs that was intended to be internal, but got pasted here.

A response to an API break that we didn't realize happened, but no code changes in this PR.

1

# Testing

None relevant.


jdetter (approved):
Much better, thanks for the clarity @gefjon !

PR #2324, opened 2025-02-27
test: set semantics in datastore

joshua-spacetime:
# Description of Changes

Adds test ensuring we follow set semantics with and without unique constraints.

None

1

# Testing

This patch only adds test

### DIFF
--- a/crates/core/src/db/datastore/locking_tx_datastore/datastore.rs
+++ b/crates/core/src/db/datastore/locking_tx_datastore/datastore.rs
@@ -1043,11 +1043,12 @@ mod tests {
     use itertools::Itertools;
     use pretty_assertions::{assert_eq, assert_matches};
     use spacetimedb_execution::Datastore;
+    use spacetimedb_lib::bsatn::ToBsatn;
     use spacetimedb_lib::db::auth::{StAccess, StTableType};
     use spacetimedb_lib::error::ResultTest;
     use spacetimedb_lib::st_var::StVarValue;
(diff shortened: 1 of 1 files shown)


crates/core/src/db/datastore/locking_tx_datastore/datastore.rs
```diff
@@ -2737,6 +2738,152 @@ mod tests {
...
+                constraint.map(|cs| vec![cs]).unwrap_or_default(),
+                vec![],
+                StTableType::User,
+                StAccess::Public,
+                None,
+                primary_key,
+            )
+        };
+        let table_schema_no_constraints = || table_schema(None, None);
+        let table_schema_pk = || table_schema(Some(0.into()), None);
```
gefjon, line 2766: Is it meaningful / correct to have a primary key without a unique constraint? By which I mean, primary key is supposed to imply a unique constraint, but I don't know if that's true at the schema layer, or if by this point we just have the invariant that every primary key column will also have a unique constraint.

gefjon (approved):
A question to which the answer is not terribly important for this test. Otherwise excellent. Test is clear and does the thing it's supposed to.

crates/core/src/db/datastore/locking_tx_datastore/datastore.rs
joshua-spacetime: Logically, pk does imply unique, but I don't believe pk is purely a logical concept in our datastore. In particular, we store the pk col in `st_table` and I don't believe we ever physically convert it into a unique constraint.

Adam-cpu-cmd (commented):
I

PR #2325, opened 2025-02-28
set names atomic

kim:
Reworks #2317 to perform the replacement of domain names as an atomic operation.

# Testing

Adds two smoketest asserting that the endpoint works and that the caller must be the owner of the database.


kim:
@coolreader18 What can I do about the surprising and inconvenient fact that `tomllib` is not present on windows?

coolreader18:
`actions/setup-python` probably. I can push a commit to this branch if that's ok?

coolreader18 (approved):
Looks reasonable to me.

kim:
Thanks @coolreader18!

PR #2352, opened 2025-03-05
Update bytes_sent_to_clients when merging metrics

joshua-spacetime:
# Description of Changes

`bytes_sent_to_clients` was not being updated when merging with another set of metrics. Now it is.

None

1

# Testing

A new unit test has been added


jdetter:
This needs to be rebased on latest master

Issue #2348, opened 2025-03-05
`Identity` and `ConnectionId` should implement `IComparable`


rekhoff:
### Issue:
When attempting to create a table that uses `Identity` as an index of a BTree like:
```
#[index(btree)]
pub identity: Identity,
```
Generated code in C# will throw the error:
`The type 'SpacetimeDB.Identity' must be convertible to 'System.IComparable<SpacetimeDB.Identity>' in order to use it as parameter 'Column' in the generic class 'SpacetimeDB.RemoteTableHandle<EventContext,Row>.BTreeIndexBase<Column>'`

This is because `SpacetimeDB.Identity` does not implement the `IComparable` interface:
https://github.com/clockworklabs/SpacetimeDB/blob/7cb509c2e2696a067e78734f07311a1d90e38fcf/crates/bindings-csharp/BSATN.Runtime/Builtins.cs#L204

While we are at it, it makes sense to  implement `IComparable` for `ConnectionId` as well:
https://github.com/clockworklabs/SpacetimeDB/blob/7cb509c2e2696a067e78734f07311a1d90e38fcf/crates/bindings-csharp/BSATN.Runtime/Builtins.cs#L121

joshua-spacetime:
Note, unique constraints (which should be generating the same btree index) work just fine
```rust
#[unique]
pub identity: Identity,
```

rekhoff:
Resolved by PR https://github.com/clockworklabs/SpacetimeDB/pull/2354

PR #2354, opened 2025-03-05
 Implements `IComparable` for `Identity` and `ConnectionId`

rekhoff:
# Description of Changes

 Implements `IComparable` for `Identity` and `ConnectionId`

This relates to the following issue:
https://github.com/clockworklabs/SpacetimeDB/issues/2348

Not a breaking change. Adds functionality and fixes bug.

1

# Testing

1. Created a test project with a Rust server defining a table with a `btree` index like:
```
#[index(btree)]
pub identity: Identity,
```
and another where a table was defined with:
```#[spacetimedb::table(name = period_timer, scheduled(period, at = scheduled_at), index(name = identity, btree(columns = [identity])))]```
2. Published the server.
3. Created a C# client and generated the bindings.
4. Confirmed issue existed prior to change: IDE reports an error in the table definition of `The type 'SpacetimeDB.Identity' must be convertible to 'System.IComparable<SpacetimeDB.Identity>' in order to use it as parameter 'Column' in the generic class 'SpacetimeDB.RemoteTableHandle<EventContext,Row>.BTreeIndexBase<Column>'`
5. Made change contained in this PR.
6. Rebuilt SpacetimeDB executables, and C# SDK NuGet files.
7. Tested again by republishing and regenerating bindings.
8. Observed IDE no longer reports issue.

- [x] Tested IDE no longer reports error after change.

### DIFF
--- a/crates/bindings-csharp/BSATN.Runtime/Builtins.cs
+++ b/crates/bindings-csharp/BSATN.Runtime/Builtins.cs
@@ -118,7 +118,10 @@ public readonly partial struct Unit
 }
 
 [StructLayout(LayoutKind.Sequential)]
-public readonly record struct ConnectionId : IEquatable<ConnectionId>, IComparable, IComparable<ConnectionId>
+public readonly record struct ConnectionId
+    : IEquatable<ConnectionId>,
+        IComparable,
+        IComparable<ConnectionId>
 {
     private readonly U128 value;
 
@@ -305,7 +308,7 @@ public readonly record struct Identity : IEquatable<Identity>, IComparable, ICom
 
     // This must be explicitly implemented, otherwise record will generate a new implementation.
     public override string ToString() => Util.ToHexBigEndian(value);
-    
+
     /// <inheritdoc cref="IComparable.CompareTo(object)" />
     public int CompareTo(object? value)
     {
--- a/crates/bindings-csharp/BSATN.Runtime/Builtins.cs
+++ b/crates/bindings-csharp/BSATN.Runtime/Builtins.cs
@@ -220,21 +220,7 @@ public readonly record struct ConnectionId
     }
 
     /// <inheritdoc cref="IComparable{T}.CompareTo(T)" />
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: crates/bindings-csharp/BSATN.Runtime/Builtins.cs [hunk 1]
```diff
@@ -198,10 +201,44 @@ public AlgebraicType GetAlgebraicType(ITypeRegistrar registrar) =>
...
+            return -1;
+        }
+        else if (this.value > connectionId.value)
+        {
+            return 1;
+        }
+        else
+        {
+            return 0;
+        }
```
kazimuth, line 236:
```suggestion
        this.value.CompareTo(connectionId.value)
```
?

kazimuth (approved):
Could you add a basic test to BSATN.Runtime.Tests for these? With that I'm happy with this.

It would be nice to add a test to the SDK as well duplicating what you said you did in the PR description, but unfortunately that would require a PR in a different repo...

crates/bindings-csharp/BSATN.Runtime/Builtins.cs
rekhoff: Yup, our [U128 does](https://github.com/clockworklabs/SpacetimeDB/blob/6302231e8afdfe5f0f815a80fe75d8c51d2bc4b0/crates/bindings-csharp/BSATN.Runtime/BSATN/U128.cs#L54)!
That needs to actually return the result of the `CompareTo` though, so I'll make this change as an update where I can add the return.

rekhoff:
`BSATN.Runtime.Tests` added for both `Identity` and `ConnectionId`.
Marking as ready for merge.
This PR will resolve issue: https://github.com/clockworklabs/SpacetimeDB/issues/2348
New issue created for adding a test that performs the manually performed actions outlined in the description of this PR:
https://github.com/clockworklabs/SpacetimeDB/issues/2362

PR #2386, opened 2025-03-07
Clarify what is a valid module name

coolreader18:
# Description of Changes

From a conversation on the discord.

None; anything new caught by this will already have been denied by the server, this just gives a better error message.

1

# Testing

- [x] Checked that the docs look good.


bfops (approved):
LGTM, thanks!

bfops:
This will require a corresponding spacetime-docs PR to update the docs

coolreader18:
Already did: https://github.com/clockworklabs/spacetime-docs/pull/252

bfops:
Nice! sold.

bfops:
I'm having some second thoughts on this one now.. if the server is going to enforce it anyway, and we're just going to print that error, what's the advantage of having the CLI enforce it too? Doesn't this just increase the chance that we change the logic in one place and not the other?

PR #2395, opened 2025-03-07
Update PR template

coolreader18:
# Description of Changes

This has been annoying me for months haha. This turns the instructions for each section into `<!-- comments -->`, so that they don't show up in the rendered description if they're not removed. Also removed the default X from the testing checkbox, so that the PR author has to actually check it to show they've done it.

# Testing

- [x] Tested how it renders. The below is the new template, rendered but unedited:

# Description of Changes

# Testing

- [ ] 
- [ ]

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,27 +1,28 @@
 # Description of Changes
 
-Please describe your change, mention any related tickets, and so on here.
+<!-- Please describe your change, mention any related tickets, and so on here. -->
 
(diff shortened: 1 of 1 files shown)


bfops:
This seems good to me, but I think @cloutiertyler will probably want to take a look.

cloutiertyler:
Oh absolutely, this has been annoying me too, but I couldn't think of a way to fix it.

bfops (approved):
In that case, sold!

PR #2396, opened 2025-03-07
Implement IEquatable<Self> for all [SpacetimeDB.Type]s

kazimuth:
# Description of Changes

Fixes issue identified in https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/pull/264 by implementing `IEquatable<Self>` for all `[SpacetimeDB.Type]`s in C#. Also overrides `GetHashCode()` and `ToString()` for these types, which should be helpful if users want to use them as hash map keys or print them.

Ready for review.

Technically API breaking, since it is possible users manually implemented `IEquatable` somewhere.
It's an easy fix though.

This also uses `#nullable enable` and `#nullable restore` in the generated code. This requires a version of Unity supporting C# 8. But I believe we already required this.

2: this is a slightly-tricky algorithm to implement due to some pitfalls in C#'s type system.

# Testing

- [x] Get all C# module stuff compiling
- [x] Add equality proptests for a wide variety of types
- [x] Test that that SDK bug is fixed
- [ ] Test quickstart-chat
- [ ] Test blackholio


joshua-spacetime:
I don't know the C# sdk well enough to be able to review this change sufficiently. I've reviewed the repro, and so I'm satisfied as long as it fixes the repro test, but I think @rekhoff will be able to give this a much better review.

crates/bindings-csharp/BSATN.Codegen/Type.cs
```diff
@@ -139,70 +144,116 @@ public BaseTypeDeclaration(GeneratorAttributeSyntaxContext context, DiagReporter
         );
     }
 
+    public static string JoinOrElse(string join, IEnumerable<String> joinands, string baseCase)
```
rekhoff, line 147: Might renaming this to something less threatening and a bit more descriptive inputs. Perhaps:
`public static string JoinOrValue(string join, IEnumerable<String> stringArray, string valueIfArrayEmpty)`
Or anything that might indicate the expected behavior a little clearer.

rekhoff:
There is a decent amount going on here, I don't see any clear issues, and what I see makes sense. It looks like the quickstart-chat and blackholio haven't been manually tested with these changes, which would be good to do. I'm going to hold off on my approval until after I get these tested.

rekhoff (approved):
Reviewed code, nothing stood out as wrong. Tested the repo version Blackhol.io locally using both a Rust and C# server without running into issues. Marking this as approved!

crates/bindings-csharp/BSATN.Codegen/Type.cs
kazimuth: Ha, I meant it in the sense of "join if there are joinands, else ..." but that makes sense, will do.

PR #2398, opened 2025-03-08
fix: bag semantics for joins

joshua-spacetime:
# Description of Changes

Previously the server implemented set semantics for incremental join updates. This was always incorrect, but the client SDKs didn't know how to handle duplicate rows. Now the SDKs do, but the server was never updated to implement bag semantics.

None

1

# Testing

A new SDK regression test was added


smoketests/tests/add_remove_index.py
```diff
@@ -80,7 +80,7 @@ def test_add_then_remove_index(self):
         self.publish_module(name, clear = False)
         sub = self.subscribe(self.JOIN_QUERY, n = 1)
         self.call("add", anon = True)
-        self.assertEqual(sub(), [{'t1': {'deletes': [], 'inserts': [{'id': 1001}]}}])
+        sub()
```
joshua-spacetime, line 83: The smoke tests shouldn't check subscription results, at least for join queries anyway, because the smoke tests are just using the websocket api, and so assuming set semantics would be wrong.

crates/core/src/subscription/query.rs
```diff
@@ -1023,7 +1023,59 @@ mod tests {
...
         let tx = DeltaTx::new(&tx, &data);
-        let updates = eval_delta(&tx, metrics, plan).unwrap();
+
+        // IMPORTANT: FOR TESTING ONLY!
+        //
+        // This utility implements set semantics for incremental updates.
+        // This is safe because we are only testing PK/FK joins,
+        // and we don't have to track row multiplicities for PK/FK joins.
+        // But in general we must assume bag semantics for server side tests.
+        let mut eval_delta = || {
```
joshua-spacetime, line 1033: This is just straight copy of the old `eval_delta`

kazimuth (approved):
This change looks sensible. I'm looking around at all places `UpdatesRelValue` is used -- I haven't found anything that assumes it's deduplicated. The added test makes sense.

I am not that familiar with this part of the codebase so we may want another pair of eyes, but I'm happy with this.

mamcx (approved):
LGTM

PR #2409, opened 2025-03-09
When generating for typescript, convert the primary key to camel case.

jsdt:
# Description of Changes

Currently if the primary key column is named in snake case, the typescript client won't be able to find it, because it converts the fields to camel case, but doesn't convert the primary key name.

1

# Testing

I manually tested with the typescript quickstart.


cloutiertyler (approved):
Small change. Makes sense.

PR #2434, opened 2025-03-11
Fix subs when clients use variying compression

Centril:
# Description of Changes

1. Previously, we had a bug when multiple clients would subscribe to the same query and protocol (bsatn) but with different compression algorithms. This PR fixes that by memoizing once per compression algo.
2. As part of the fix, it was also discovered that the rust sdk gzip decompression was broken. This PR fixes that too.

None

2

# Testing

A regression test is added to confirm the bugs are fixed.

### DIFF
--- a/crates/sdk/tests/test-client/src/main.rs
+++ b/crates/sdk/tests/test-client/src/main.rs
@@ -2162,6 +2162,10 @@ fn test_intra_query_bag_semantics_for_join() {
     });
 }
 
+/// Test that several clients subscribing to the same query and using the same protocol (bsatn)
+/// can use different compression algorithms than each other.
+///
+/// This is a regression test.
 fn exec_two_different_compression_algos() {
     use Compression::*;
 


crates/client-api-messages/src/websocket.rs
```diff
@@ -887,7 +887,7 @@ pub fn gzip_compress(bytes: &[u8], out: &mut Vec<u8>) {
 
 pub fn gzip_decompress(bytes: &[u8]) -> Result<Vec<u8>, io::Error> {
```
joshua-spacetime, line 888: Could you move the compression utilities out of `client-api-messages`? This crate should really just be about the message types themselves.

crates/sdk/tests/test-client/src/main.rs
```diff
@@ -2157,6 +2160,69 @@ fn test_intra_query_bag_semantics_for_join() {
             });
         }
     });
+}
+
+fn exec_two_different_compression_algos() {
```
joshua-spacetime, line 2165: Could you add a comment explaining the reason for this regression test, i.e. the original bug and its semantics.

gefjon (approved):

PR #2444, opened 2025-03-12
Fix outdated advice in `getrandom` warning

gefjon:
# Description of Changes

Title.

Missed when we changed our randomness API to be tied to the ctx.

[As reported in the public Discord](https://discord.com/channels/1037340874172014652/1348121042454839347/1348121042454839347).

N/a.

0.

# Testing

None.


bfops (approved):
Seems good to me!

Issue #2390, opened 2025-03-07
Repair transactionality of `st_client` removals after disconnect


bfops:
When the `client_disconnected` reducer runs successfully, we should include the `st_client` deletion in that transaction. As of writing we do not, instead appending another transaction afterwards. Note that if the `client_disconnected` reducer fails, we must then start another transaction to delete from `st_client`. In either case, we should commit exactly one TX.

PR #2446, opened 2025-03-12
Repair transactionality of `st_client` removals after disconnect

Centril:
# Description of Changes

Fixes https://github.com/clockworklabs/SpacetimeDB/issues/2390 , #2505 .

None

2

# Testing

Not tested as of yet. To my knowledge, we don't really have the test infra to force the disconnect and then check the `TxData`.

### DIFF
--- a/crates/core/src/host/module_host.rs
+++ b/crates/core/src/host/module_host.rs
@@ -1,9 +1,10 @@
-use super::{ArgsTuple, InvalidReducerArguments, ReducerArgs, ReducerCallResult, ReducerId, Scheduler};
+use super::{ArgsTuple, InvalidReducerArguments, ReducerArgs, ReducerCallResult, ReducerId, ReducerOutcome, Scheduler};
 use crate::client::{ClientActorId, ClientConnectionSender};
 use crate::database_logger::{LogLevel, Record};
 use crate::db::datastore::locking_tx_datastore::MutTxId;
 use crate::db::datastore::traits::{IsolationLevel, Program, TxData};
 use crate::energy::EnergyQuanta;
+use crate::error::DBError;
 use crate::estimation::estimate_rows_scanned;
 use crate::execution_context::{ExecutionContext, ReducerContext, Workload, WorkloadType};
 use crate::hash::Hash;
@@ -465,6 +466,18 @@ pub enum InitDatabaseError {
     Other(anyhow::Error),
 }
 
+#[derive(thiserror::Error, Debug)]
+pub enum ClientConnectedError {
+    #[error(transparent)]
+    ReducerCall(#[from] ReducerCallError),
+    #[error("Failed to insert `st_client` row for module without client_connected reducer: {0}")]
+    DBError(#[from] DBError),
+    #[error("Connection rejected by `client_connected` reducer: {0}")]
+    Rejected(String),
+    #[error("Insufficient energy balance to run `client_connected` reducer")]
+    OutOfEnergy,
+}
+
 impl ModuleHost {
     pub fn new(mut module: impl Module, on_panic: impl Fn() + Send + Sync + 'static) -> Self {
         let info = module.info();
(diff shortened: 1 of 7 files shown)


crates/core/src/db/datastore/locking_tx_datastore/mut_tx.rs
```diff
@@ -1229,6 +1235,31 @@ impl<'a> Iterator for IndexScanFilterDeleted<'a> {
...
+            connection_id: connection_id.into(),
+        };
+        let ptr = self
+            .iter_by_col_eq(
+                ST_CLIENT_ID,
+                col_list![StClientFields::Identity, StClientFields::ConnectionId],
+                &AlgebraicValue::product(row),
+            )?
+            .next()
+            .expect("the client should be connected")
```
kim, line 1258: Can we not just make the method a no-op in this case?

kim (approved):
I don't know what problem this fixes, really, but the change looks correct.

crates/core/src/db/datastore/locking_tx_datastore/mut_tx.rs
gefjon: I don't think that's necessarily true, actually: it's possible, though inadvisable, for the module owner to use `spacetime sql` to manually delete `st_client` rows. I think that we generally do not expect manually mutating the system tables to be well-behaved, but we don't want any customer to be able to kill our threads, which panicking might do. I am unsure if there is a `catch_unwind` above this in the call stack.

crates/core/src/host/module_host.rs
```diff
@@ -570,70 +579,26 @@ impl ModuleHost {
             )
             .await
             .map(drop)
+            .inspect_err(|_| {
+                // ...but it has failed, so we treat this as if the reducer didn't exist
+                // and commit a separate transaction instead.
```
gefjon, line 584: I am not convinced this is correct in the connected path. If the reducer doesn't exist, we should run the fallback, but if the reducer exists and failed, we should return an error from this method, which IIRC will cause the connection to be rejected/terminated.

This is the intended behavior in the disconnected path, though.

gefjon (changes requested):

crates/core/src/host/module_host.rs
kim: But in both cases the client ends up disconnected, so we should have the row deleted, no?

gefjon: The enclosing method, `call_identity_connected_disconnected`, is called in two contexts: when a new client establishes a connection (the "connected path") and when a previously-connected client terminates its connection (the "disconnected path"). These paths have meaningfully different behavior, and possibly should just be different methods.

### In the "connected path," for a new client:

- The row is expected to previously have not been present in `st_client`.
- If the module does not define a `client_connected` reducer, we should insert a new row into `st_client` and return successfully.
- If the module does define a `client_connected` reducer, call it.
  - If that reducer call returns successfully, we should insert a new row into `st_client` *within the same transaction*.
    - This is handled elsewhere, in `module_host_actor.rs`'s `WasmModuleInstance::call_reducer_with_tx`.
  - If that reducer call returns an error, we should not insert into `st_client`. Instead, we should return an error from `call_identity_connected_disconnected`, which will cause `client_connection.rs`'s `ClientConnection::spawn` to terminate the connection.
    - **This is the part that I believe to be incorrect in this PR.**
  
### In the "disconnected path," for a previously-existing client:

- The row is expected to be present in `st_client` prior to entering `call_identity_connected_disconnected`.
- If the module does not define a `client_disconnected` reducer, we should delete that row from `st_client` and return successfully.
- If the module does define a `client_disconnected` reducer, call it.
  - If that reducer call returns successfully, we should delete the row from `st_client` *within the same transaction*.
    - If done in a separate transaction, we may crash between the `client_disconnected` reducer call and the `st_client` delete, after which we will restart into a state where the reducer has already been called but the row has not been deleted, and will then re-run `client_disconnected` and re-delete the row.
    - As above, this is handled by `WasmModuleInstance::call_reducer_with_tx`.
    - **This is the part that I believe to have been broken prior to this PR, and to be fixed here.**
  - If that reducer call returns an error, we should still delete the row from `st_client`.
    - The errored reducer call's transaction is already destroyed, so we must start a new transaction to do this delete.
    - That's ok, though, since if we crash in between and restart, we'll re-run `client_disconnected`, have it fail again, and no one will ever be the wiser.

PR #2448, opened 2025-03-12
Fix typescript primary key

pleum:
# Description of Changes

While we generate primary key with `snake_case` but data was deserialize as json `camelCase` there is no way to get key from data , So it will use `undefined` as key for this line https://github.com/clockworklabs/spacetimedb-typescript-sdk/blob/main/packages/sdk/src/table_cache.ts#L69 and keep raising error `Updating a row that was not present in the cache` on browser console

This can re-produce by create table contains `snake_case` id
```
#[table(name = entity_player, public)]
#[derive(Debug)]
pub struct EntityPlayer {
    #[primary_key]
    pub entity_id: u32,
    #[index(btree)]
    pub player_id: u32,
}
```

corresponding GitHub label.

Complexity Level: 1

# Testing

- [x] *Generate typescript with table contains `snake_case` id, no more error on browser console*


jsdt:
Thanks for adding this. I have an almost identical PR that should merge soon (https://github.com/clockworklabs/SpacetimeDB/pull/2409), so I'm going to close this PR. The fix should be in the 1.0.1 release.

PR #2462, opened 2025-03-13
Print back the # of rows affected (ins, upd, del) with timings

### LINKED ISSUE #2430: SQL - `update` and `delete` queries do not return any rows
When running an `update` or `delete` query in SQL, the API does not return any rows (although it does do the operation properly).

This was the root cause of https://github.com/clockworklabs/SpacetimeDB/issues/2358.

mamcx:
# Description of Changes

Closes #2430.

It change the  returning `json` for `sql` route (spacetimedb::json::client_api::StmtResultJson), with added field for stats.

1

# Testing

- [x] *Added test that confirm the number of stats for `ins, del, upd`*
- [x] *Added unit test for the output of the `sql` cli with and without stats* 
- [x] *Manual testing of the `cli` output using the module `keynote-benchmarks`*

```bash
🪐quickstart>delete from position;
(0 rows) [deleted: 1000000, server: 3.71s]
Roundtrip time: 3.72s

🪐quickstart>select * from position;
 id | x | y | z
----+---+---+---
(0 rows) [server: 1.32ms]
Roundtrip time: 6.46ms

🪐quickstart>delete from position;
(0 rows) [server: 1.47ms]
Roundtrip time: 7.69ms

🪐quickstart>INSERT INTO position (id, x, y, z) VALUES (1000011, 10.0, 20.0, 30.0);
(0 rows) [inserted: 1, server: 2.82ms]
Roundtrip time: 7.13ms

🪐quickstart>select * from position;
 id      | x  | y  | z
---------+----+----+----
 1000011 | 10 | 20 | 30
(1 row) [server: 1.29ms]
Roundtrip time: 5.49ms

🪐quickstart>update position set x = 1.0;
(0 rows) [updated: 1, server: 1.96ms]
Roundtrip time: 4.50ms

```

### DIFF
--- a/crates/cli/src/api.rs
+++ b/crates/cli/src/api.rs
@@ -103,7 +103,6 @@ pub struct StmtStats {
     pub rows_inserted: u64,
     pub rows_updated: u64,
     pub rows_deleted: u64,
-    pub rows_scanned: u64,
     pub total_rows: usize,
 }
 
@@ -122,7 +121,6 @@ impl Add for StmtStats {
             rows_inserted: self.rows_inserted + rhs.rows_inserted,
             rows_deleted: self.rows_deleted + rhs.rows_deleted,
             rows_updated: self.rows_updated + rhs.rows_updated,
-            rows_scanned: self.rows_scanned + rhs.rows_scanned,
             total_rows: self.total_rows + rhs.total_rows,
         }
     }
@@ -135,7 +133,6 @@ impl From<&StmtResultJson<'_>> for StmtStats {
             rows_inserted: value.stats.rows_inserted,
             rows_deleted: value.stats.rows_deleted,
             rows_updated: value.stats.rows_updated,
-            rows_scanned: value.stats.rows_scanned,
             total_rows: value.rows.len(),
         }
     }
--- a/crates/cli/src/api.rs
+++ b/crates/cli/src/api.rs
@@ -1,7 +1,11 @@
+use std::iter::Sum;
+use std::ops::Add;
+
 use reqwest::{header, Client, RequestBuilder};
 use serde::Deserialize;
 use serde_json::value::RawValue;
 
+use spacetimedb::json::client_api::StmtStatsJson;
 use spacetimedb_lib::db::raw_def::v9::RawModuleDefV9;
 use spacetimedb_lib::de::serde::DeserializeWrapper;
 use spacetimedb_lib::sats::ProductType;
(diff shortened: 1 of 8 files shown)


crates/cli/src/api.rs
```diff
@@ -82,11 +86,59 @@ impl ClientApi {
...
+    pub stats: StmtStatsJson,
+}
+
+#[derive(Debug, Clone, Deserialize, Default)]
+pub struct StmtStats {
+    pub total_duration_micros: u64,
+    pub rows_inserted: u64,
+    pub rows_updated: u64,
+    pub rows_deleted: u64,
+    pub rows_scanned: u64,
```
joshua-spacetime, line 106: I don't believe we provided rows scanned previously. I think I'd prefer to leave this one out since it's not something a caller can necessarily reason about.

crates/cli/src/subcommands/sql.rs
```diff
@@ -54,18 +54,57 @@ pub(crate) async fn parse_req(mut config: Config, args: &ArgMatches) -> Result<C
...
+            }
+            let txt = if stats.total_rows == 1 { "row" } else { "rows" };
+
+            let result = format!("({} {txt})", stats.total_rows);
+            let mut info = Vec::new();
+            if stats.rows_scanned != 0 {
+                info.push(format!("scan: {}", stats.rows_scanned));
+            }
+            if stats.rows_inserted != 0 {
+                info.push(format!("ins: {}", stats.rows_inserted));
```
joshua-spacetime, line 82: Did we abbreviate this previously?

mamcx: No, wanna instead `insert` or `inserted`?

joshua-spacetime: I think `inserted`.

bfops (approved):
Approving the files that I am a codeowner for:
```
crates/cli/src/api.rs
crates/cli/src/subcommands/sql.rs
```

crates/cli/src/subcommands/sql.rs
```diff
@@ -54,18 +54,54 @@ pub(crate) async fn parse_req(mut config: Config, args: &ArgMatches) -> Result<C
...
+                write!(f, "{result}")?;
+            };
+        };
+        Ok(())
+    }
 }
 
 pub(crate) async fn run_sql(builder: RequestBuilder, sql: &str, with_stats: bool) -> Result<(), anyhow::Error> {
-    let now = Instant::now();
+    let mut now = Instant::now();
```
joshua-spacetime, line 104: Why is this mutable?

crates/cli/src/subcommands/sql.rs
```diff
@@ -77,41 +113,44 @@ pub(crate) async fn run_sql(builder: RequestBuilder, sql: &str, with_stats: bool
...
-                if row_count > 0 {
-                    let row_count = print_row_count(row_count);
-                    table.with(tabled::settings::panel::Footer::new(row_count));
-                }
-            }
-            anyhow::Ok(table)
+            let (stats, table) = stmt_result_to_table(stmt_result)?;
+
+            let time_client = now;
+            now = Instant::now();
```
joshua-spacetime, line 139: What does this do?

joshua-spacetime (commented):
Is it possible to add a smoke test as well?

crates/cli/src/subcommands/sql.rs
bfops: I had the same question. it's written to in the `map` below. on every iteration, we save the current value of `now` and then re-set it. I'm not a huge fan of the way it's structured, but I don't feel super strongly about it.

I guess it would be nice to have some tests demonstrating what happens with multiple queries though.

bfops:
Wait, is this a breaking change? We just added a field, is that right? If it's really a breaking change, we can't release it.

crates/cli/src/api.rs
```diff
@@ -82,11 +86,56 @@ impl ClientApi {
...
 
+// Sync with spacetimedb::json::client_api::StmtResultJson
 #[derive(Debug, Clone, Deserialize)]
 pub struct StmtResultJson<'a> {
     pub schema: ProductType,
     #[serde(borrow)]
     pub rows: Vec<&'a RawValue>,
+    pub total_duration_micros: u64,
+    #[serde(default)]
+    pub stats: StmtStatsJson,
```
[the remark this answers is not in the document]
mamcx: To avoid breaking the clients. With `serde(default)` it should be transparent. However I agree it will be nicer if put there.

crates/cli/src/subcommands/sql.rs
```diff
@@ -54,18 +54,54 @@ pub(crate) async fn parse_req(mut config: Config, args: &ArgMatches) -> Result<C
...
+            let mut info = Vec::new();
+            if stats.rows_inserted != 0 {
+                info.push(format!("inserted: {}", stats.rows_inserted));
+            }
+            if stats.rows_deleted != 0 {
+                info.push(format!("deleted: {}", stats.rows_deleted));
+            }
+            if stats.rows_updated != 0 {
+                info.push(format!("updated: {}", stats.rows_updated));
+            }
```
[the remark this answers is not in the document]
mamcx: No, all options are additive, so when someday we have `INSERT...; DELETE...;SELECT` should work

crates/cli/src/subcommands/sql.rs
```diff
@@ -54,18 +54,54 @@ pub(crate) async fn parse_req(mut config: Config, args: &ArgMatches) -> Result<C
...
+            info.push(format!(
+                "server: {:.2?}",
+                std::time::Duration::from_micros(stats.total_duration_micros)
+            ));
+            info.push(format!("client: {:.2?}", self.time_client.elapsed()));
```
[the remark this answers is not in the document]
mamcx: Yes.

crates/cli/src/subcommands/sql.rs
mamcx: The main reason is to capture the cost of the network call. If I put in each individually it will just serialization

crates/cli/src/subcommands/sql.rs
```diff
@@ -54,18 +54,54 @@ pub(crate) async fn parse_req(mut config: Config, args: &ArgMatches) -> Result<C
...
+            write!(f, "{}", self.table)?;
+        }
+
+        if let Some(stats) = &self.stats {
+            if has_table {
+                writeln!(f)?;
+            }
+            let txt = if stats.total_rows == 1 { "row" } else { "rows" };
+
+            let result = format!("({} {txt})", stats.total_rows);
```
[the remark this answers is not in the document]
mamcx: Yes. It will be complicated to figure out when a query is pure DML.

mamcx:
> That's correct, this just adds a field, and so it's not a breaking change. I believe the label was added by mistake @mamcx?

Yeah, I put the label before adding the `serde` optionality.

joshua-spacetime (approved):

PR #2466, opened 2025-03-14
CI - Containerize linux x86_64 GNU builds

bfops:
# Description of Changes

We've been having awful issues with some weird permission denied error with the linux x86_64 GNU builds running on bare metal. We suspect that the alpine builds don't have this issue because they run in a container, so we're making these run in a container too.

No. CI-only change.

1

# Testing

- [x] x86_64 builds for the release [are passing](https://github.com/clockworklabs/SpacetimeDB/actions/runs/13976348943/job/39131011257?pr=2487) since we added this (previously they were failing due to some weird permission issues)


coolreader18:
Gonna push to this to fix #2510  - hope that's okay.

coolreader18 (approved):

PR #2468, opened 2025-03-14
commitlog: Fix open flags for read-only offset index

kim:
For some reason, we've been opening the read-only index file with
`.read(true).append(true)`.

This translates to `O_RDWR | O_APPEND` on *nix, but something weird on
Windows. A shared memory map requires `O_RDWR` (or equivalent) on either
platform, though, meaning that we'd always fail to `mmap` the index on
Windows.

Open with `.read(true).write(true)` instead to fix.

1

The impact of this bug is also small, because the offset index is not crucial
for standalone use. Who says, though, that Windows will never become a viable
server platform for SpacetimeDB?

# Testing

Extends the tests to open an index using all three variants and perform lookups.

### DIFF
--- a/crates/commitlog/src/index/indexfile.rs
+++ b/crates/commitlog/src/index/indexfile.rs
@@ -241,7 +241,7 @@ pub struct IndexFile<Key> {
 
 impl<Key: Into<u64> + From<u64>> IndexFile<Key> {
     pub fn open_index_file(path: &OffsetIndexFile) -> io::Result<Self> {
-        let file = path.open_file(File::options().read(true).append(true))?;
+        let file = path.open_file(File::options().read(true).write(true))?;
         let mmap = unsafe { MmapMut::map_mut(&file)? };
 
         let mut inner = IndexFileMut {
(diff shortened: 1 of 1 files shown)


crates/commitlog/src/index/indexfile.rs
```diff
@@ -241,7 +241,7 @@ pub struct IndexFile<Key> {
 
 impl<Key: Into<u64> + From<u64>> IndexFile<Key> {
     pub fn open_index_file(path: &OffsetIndexFile) -> io::Result<Self> {
```
kim, line 243: The only difference between `IndexFile::open_index_file` and `IndexFileMut::open_index_file` is that the latter sets the file's length.

I'm wondering if we should either a) just use `IndexFileMut` from `IndexFile` (because we have `impl AsMut<IndexFileMut> for IndexFile`), or b) use `Mmap` (immutable) to back `IndexFile`

Shubham8287: May be, ideally it should be b but what we have right now is convienient. I would prefer to use `IndexFile` as it requires explicit call to mutate rather than using `IndexFileMut` always.

Issue #2415, opened 2025-03-10
alpine linux install, no elf download


syonfox:
Nice presentation, it seems the build script dose not work on alpine Linux.

```
localhost:~$ uname -a
Linux localhost 6.12.16-0-lts #1-Alpine SMP PREEMPT_DYNAMIC 2025-02-21 15:23:09 x86_64 Linux
localhost:~$ curl --proto '=https' --tlsv1.2 -sSf https://install.spacetimedb.com | sh
Downloading installer...
curl: (22) The requested URL returned error: 404
It seems like we may not have prebuilt binaries for your platform.
localhost:~$ 

```

note I had to run `apk add curl`

This file seems to not exist 

https://github.com/clockworklabs/SpacetimeDB/releases/download/v1.0.0

PR #2470, opened 2025-03-14
Use jemalloc in standalone and add stats.

jsdt:
# Description of Changes

This uses jemalloc in standalone instead of mimalloc, unless the target env is msvc. The main reason is that jemalloc exposes better stats, and has a way to do heap profiling. This PR doesn't set up profiling, but it does add stats on jemalloc's memory usage.

2. The risk here is that there could be an issue on some platforms. 

# Testing

This works on mac and at least some linux platforms.

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -5,6 +5,8 @@ use spacetimedb_cli::*;
 use spacetimedb_paths::cli::CliTomlPath;
 use spacetimedb_paths::{RootDir, SpacetimePaths};
 
+// Note that the standalone server is invoked through standaline/src/main.rs, so you will
+// also want to set the allocator there.
 #[cfg(not(target_env = "msvc"))]
 use tikv_jemallocator::Jemalloc;
 
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -1,11 +1,21 @@
 use std::process::ExitCode;
 
 use clap::{Arg, Command};
-use mimalloc::MiMalloc;
 use spacetimedb_cli::*;
 use spacetimedb_paths::cli::CliTomlPath;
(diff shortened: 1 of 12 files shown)


crates/cli/src/main.rs
```diff
@@ -1,11 +1,21 @@
...
 use spacetimedb_cli::*;
 use spacetimedb_paths::cli::CliTomlPath;
 use spacetimedb_paths::{RootDir, SpacetimePaths};
 
+#[cfg(not(target_env = "msvc"))]
+use tikv_jemallocator::Jemalloc;
+
+#[cfg(not(target_env = "msvc"))]
+#[global_allocator]
+static GLOBAL: Jemalloc = Jemalloc;
```
bfops, line 13: so to confirm - this allocator change still somehow applies to `standalone` even though it's in a different binary? is that because we `exec` it?

bfops:
I suspect we should probably figure out if there are any platforms that this won't work on, since we just went through the work to expand our platform support (https://github.com/clockworklabs/SpacetimeDB/issues/2415).

I guess alternatively we could wait to see if any users complain..

bfops:
this broadly looks good to me, but I'm not very familiar with different allocator behavior or with the stats part of our code, so I wonder if we can get someone to review who's more confident.

crates/cli/src/main.rs
jsdt: Nope. I thought it did, since I was still getting jemalloc stats, but it seems like the stat collection allocates a bit of memory even if we aren't using jemalloc anywhere else.

jsdt:
benchmarks please

bfops (approved):
Looks good as far as I can tell. We think this will almost certainly be an improvement over the system allocator because jemalloc usually is. And @jsdt says he tested it.

PR #2471, opened 2025-03-14
CI - Fix docker jobs failing

bfops:
# Description of Changes

We're having some confusing issue where our docker jobs fail due to permission denied errors.

We can and should investigate this, but for now i'm just moving them to run on `ubuntu-latest`.

No. CI only.

1

# Testing

- [x] The docker job passes when we run it on this branch (see https://github.com/clockworklabs/SpacetimeDB/actions/runs/13863230991/job/38796443167).


.github/workflows/docker.yml
```diff
@@ -11,13 +11,9 @@ on:
 
 jobs:
   docker-amd64:
-    runs-on: bare-metal
+    runs-on: ubuntu-latest
     name: Build DockerHub AMD64 Container
     steps:
-      - name: Install jq
-        run: sudo apt-get install jq -y
```
bfops, line 18: `jq` is already installed on `ubuntu-latest`

.github/workflows/docker.yml
```diff
@@ -11,13 +11,9 @@ on:
...
-      - name: Prune stale references
-        run: git remote prune origin
```
bfops, line 20: no longer needed since we're not reusing the same git directory

jdetter (approved):
Even if this is slower it's better to have the environment sanitized, thanks Zeke 👍 

Maybe we can eventually get this working again on the bare-metal runner but in a docker container this time?

PR #2472, opened 2025-03-14
Allow databases from 1.0.1 to be run on 1.0.0

coolreader18:
# Description of Changes

From the comment I put in the code:

> set the patch version of newly-created metadata files to 0 [which means that patch number of existing metadata files won't be updated past 0 either] -- v1.0.0 set `cmp.patch = Some(file_version.patch)` when checking version compatibility, meaning it won't be forwards-compatible with a database claiming to be created on v1.0.1, even though that should work. This can be changed once we release v1.1.0, since we don't care about its DBs being backwards-compatible with v1.0.0 anyway.

2 - it's slightly tricky to figure out the right thing to do to maintain compatibility while disallowing incompatibility, but I'm confident this is it :)

# Testing

- [x] Added unit tests to make sure it all works, including that v1.0.0 will be compatible with v1.0.1

### DIFF
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -8,10 +8,6 @@ use spacetimedb_lib::ConnectionId;
 use spacetimedb_paths::cli::{ConfigDir, PrivKeyPath, PubKeyPath};
 use spacetimedb_paths::server::{ConfigToml, MetadataTomlPath};
 
-pub fn current_version() -> semver::Version {
-    env!("CARGO_PKG_VERSION").parse().unwrap()
-}
-
 /// Parse a TOML file at the given path, returning `None` if the file does not exist.
 ///
 /// **WARNING**: Comments and formatting in the file will be lost.
@@ -23,7 +19,7 @@ pub fn parse_config<T: serde::de::DeserializeOwned>(path: &Path) -> anyhow::Resu
     }
 }
 
-#[derive(serde::Serialize, serde::Deserialize)]
+#[derive(serde::Serialize, serde::Deserialize, Debug)]
 pub struct MetadataFile {
     pub version: semver::Version,
     pub edition: String,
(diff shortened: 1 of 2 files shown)


crates/core/src/config.rs
```diff
@@ -35,6 +31,24 @@ pub struct MetadataFile {
 }
 
 impl MetadataFile {
+    pub fn new(edition: &str) -> Self {
+        let mut current_version: semver::Version = env!("CARGO_PKG_VERSION").parse().unwrap();
```
bfops, line 35: why did this change from `spacetimedb::config::current_version()`?

bfops: ah we just inlined it?

crates/core/src/config.rs
```diff
@@ -35,6 +31,24 @@ pub struct MetadataFile {
...
+        // set the patch version of newly-created metadata files to 0 -- v1.0.0
+        // set `cmp.patch = Some(file_version.patch)` when checking version
+        // compatibility, meaning it won't be forwards-compatible with a
+        // database claiming to be created on v1.0.1, even though that should
+        // work. This can be changed once we release v1.1.0, since we don't
+        // care about its DBs being backwards-compatible with v1.0.0 anyway.
+        if let semver::Version { major: 1, minor: 0, .. } = current_version {
+            current_version.patch = 0;
+        }
```
bfops, line 44: what do you think about just unconditionally setting `patch` to 0? That way, we don't need a fancier comparator.

coolreader18: The comparator isn't really fancier than it was before, it's just swapping `Some(patch)` for `None` -- and I think we may want to know if the database has been used on a specific patch version in the future, or at least there's no reason not to store that metadata in case it's ever needed.

bfops (approved):
Seems good to me!

crates/standalone/src/lib.rs
```diff
@@ -47,17 +47,9 @@ impl StandaloneEnv {
...
-            client_connection_id: None,
-        };
+        let mut meta = MetadataFile::new("standalone");
         if let Some(existing_meta) = MetadataFile::read(&meta_path).context("failed reading metadata.toml")? {
-            anyhow::ensure!(
-                existing_meta.version_compatible_with(&meta.version) && existing_meta.edition == meta.edition,
-                "metadata.toml indicates that this database is from an incompatible \
-                 version of SpacetimeDB. please run a migration before proceeding."
-            );
+            meta = existing_meta.check_compatibility_and_update(meta)?;
```
bfops, line 52: ah I realize now post-merge that we call this on `existing_meta` with `meta` as a param, rather than the other way around. This means we're preserving the `client_connection_id` from `existing_meta` that we weren't preserving before. Any chance there are negative side-effects of that?

PR #2477, opened 2025-03-17
Close scheduler on module exit

### LINKED ISSUE #2474: Clearing a database with -c does not stop the execution of the module

coolreader18:
# Description of Changes

Fixes clockworklabs/SpacetimeDBPrivate#1449. Turns out we weren't ever calling `Module::close()`, and thus were never calling `Scheduler::close()`. I'm still not able to actually repro the issue but if anything is the cause, this would be it. I also made it so `exit().await` won't return until the scheduler is fully closed, so that we won't run into race conditions.

1

# Testing

- [ ] If anyone able to repro the issue would be willing to test (or tell me how to repro) that would be super helpful

### DIFF
--- a/crates/core/src/host/module_host.rs
+++ b/crates/core/src/host/module_host.rs
@@ -1,4 +1,4 @@
-use super::{ArgsTuple, InvalidReducerArguments, ReducerArgs, ReducerCallResult, ReducerId};
+use super::{ArgsTuple, InvalidReducerArguments, ReducerArgs, ReducerCallResult, ReducerId, Scheduler};
 use crate::client::{ClientActorId, ClientConnectionSender};
 use crate::database_logger::{LogLevel, Record};
 use crate::db::datastore::locking_tx_datastore::MutTxId;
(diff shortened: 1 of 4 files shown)


joshua-spacetime (commented):
I'm not sure how you would write a non-timing based regression test for this.

You could manually test it by writing a scheduled reducer that updates some counter, inserts into an `auto_inc` table, etc. Publishing the module with `-c` a few times should affect the counter's rate of change before your fix. It shouldn't have any effect after your change.

joshua-spacetime:
@rekhoff do you know what the most recent values in the auto_inc column were for both tests?

rekhoff:
Okay, additional test checking against the `replicas` directory, using branch `noa/close-scheduler`, I can confirm:
* Starting from a clean server, publishing the module from Test 2 in my previous comment, the folder would grow ever 50 milliseconds.
* Upon first publishing, a directory named `1` was generated under `replicas`.
* Upon re-publishing, a directory named `3` was generated under `replicas`.
* After `3` was generated, directory `1` no longer grew in size, and `3` was now growing every 50 milliseconds.

rekhoff:
Okay, testing against `master` and testing against `noa/close-scheduler`, rebuilding the CLI each time netted the same results.
@joshua-spacetime , the auto_inc column restarted at 1 for each republish as well, and properly incremented by 1 each time the `grow_table` reducer from the test was called.

rekhoff:
As an extra data point, I also tested against BitCraft including running bots, and found the same behavior as described previously.

joshua-spacetime (approved):
Thanks for testing @rekhoff!

I don't believe there was ever a bug with this code, but I'm fine to approve if you think these changes are more explicit @coolreader18.

PR #2480, opened 2025-03-19
smoketests: Test database deletion

kim:
Add tests checking that a module stops running when its database is deleted or re-published using the `--clear-database` flag.

Depends on private internals and time.

### DIFF
--- a/smoketests/tests/clear_database.py
+++ b/smoketests/tests/clear_database.py
@@ -1,0 +1,71 @@
+from .. import Smoketest, random_string
+import time
+
+class ClearDatabase(Smoketest):
+    AUTOPUBLISH = False
+
+    MODULE_CODE = """
+use spacetimedb::{ReducerContext, Table, duration};
+
+#[spacetimedb::table(name = counter, public)]
+pub struct Counter {
+    #[primary_key]
+    id: u64,
+    val: u64
+}
+
+#[spacetimedb::table(name = scheduled_counter, public, scheduled(inc, at = sched_at))]
+pub struct ScheduledCounter {
+    #[primary_key]
+    #[auto_inc]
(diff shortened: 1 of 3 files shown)


smoketests/tests/clear_database.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def test_publish_clear_database(self):
+        """
+        Test that publishing with the clear flag stops the old module.
+        This relies on private control database internals.
+        """
+
+        name = random_string()
+        self.publish_module(name, clear = False)
+        time.sleep(1)
```
joshua-spacetime, line 55: Are the sleeps necessary for this test?

smoketests/tests/delete_database.py
```diff
@@ -0,0 +1,63 @@
...
+        name = random_string()
+        self.publish_module(name, clear = False)
+        sub = self.subscribe("select * from counter", n = 1000)
+        time.sleep(2)
+        self.spacetime("delete", name)
+
+        updates = sub()
+        # At a rate of 100ms, we shouldn't have more than 20 updates in 2secs.
+        # But let's say 50, in case the delete gets delayed for some reason.
+        assert len(updates) <= 50
```
joshua-spacetime, line 63: Could we check that the commitlog stops growing?

smoketests/tests/clear_database.py
kim: Actually.. no

smoketests/tests/delete_database.py
kim: This would require remote access to the commitlog, which we don't currently have in a published version.

joshua-spacetime (approved):

PR #2483, opened 2025-03-19
Add sql support for `:sender` parameter 

joshua-spacetime:
# Description of Changes

Adds sql and subscriptions support for the `:sender` parameter which resolves to the caller identity.

Note that we resolve `:sender` before we type check. Hence right now it is purely a syntactic construct. Ultimately though we'll want to have parameters represented in the physical plan.

None

2

Mostly a straightforward change. The only change that might be unexpected is that now we must explicitly take the `:sender` identity into account when compiling a subscription. This was not automatically handled for us because we don't hash the AST directly.

# Testing

- [x] Parsing tests
- [x] Type checking tests
- [x] Query evaluation tests
- [x] Subscription evaluation tests
- [x] Multiple clients, same subscription test

### DIFF
--- a/crates/core/src/sql/execute.rs
+++ b/crates/core/src/sql/execute.rs
@@ -182,7 +182,7 @@ pub fn run(
     // We parse the sql statement in a mutable transation.
     // If it turns out to be a query, we downgrade the tx.
     let (tx, stmt) = db.with_auto_rollback(db.begin_mut_tx(IsolationLevel::Serializable, Workload::Sql), |tx| {
-        compile_sql_stmt(sql_text, &SchemaViewer::new(tx, &auth))
+        compile_sql_stmt(sql_text, &SchemaViewer::new(tx, &auth), &auth)
     })?;
 
     let mut metrics = ExecutionMetrics::default();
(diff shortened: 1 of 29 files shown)


crates/sql-parser/src/parser/mod.rs
```diff
@@ -211,6 +213,7 @@ pub(crate) fn parse_expr(expr: Expr) -> SqlParseResult<SqlExpr> {
     }
     match expr {
         Expr::Nested(expr) => parse_expr(*expr),
+        Expr::Value(Value::Placeholder(param)) if &param == "@sender" => Ok(SqlExpr::Param(Parameter::Sender)),
```
joshua-spacetime, line 216: One liner to parse `@sender`

crates/sql-parser/src/parser/sql.rs
```diff
@@ -148,7 +148,7 @@ use super::{
 
 /// Parse a SQL string
 pub fn parse_sql(sql: &str) -> SqlParseResult<SqlAst> {
-    let mut stmts = Parser::parse_sql(&PostgreSqlDialect {}, sql)?;
+    let mut stmts = Parser::parse_sql(&AnsiDialect {}, sql)?;
```
joshua-spacetime, line 151: IMPORTANT: Changing the dialect of the sql parser from postgres to ANSI.

> Why?

Postgres does not support `@name` parameters. Instead it supports `:name`. This shouldn't break any compatibility with postgres, but it would look a little out of place if we care about being postgres compliant.

crates/core/src/sql/execute.rs
```diff
@@ -334,6 +334,19 @@ pub(crate) mod tests {
         Ok((stdb, MemTable::new(header, schema.table_access, rows)))
     }
 
+    fn create_identity_table(table_name: &str) -> ResultTest<(TestDB, MemTable)> {
```
mamcx, line 337: I think this will expand later to be `database params` to carry more connection details...

crates/sql-parser/src/parser/sql.rs
mamcx: Also, this means that we need to update the spec with `params` and the docs...

joshua-spacetime: Changed the syntax from `@sender` to `:sender` and switched the dialect back to postgres

mamcx (approved):
LGTM

crates/core/src/sql/execute.rs
joshua-spacetime: Exactly. As we add more params.

PR #2492, opened 2025-03-24
commitlog: Streaming

kim:
Moves support for streaming (network) transfer of raw commitlog data to the
`commitlog` crate proper, along with some improvements.

Also adds free-standing functions to determine the commitlog metadata and for
truncation. This allows to bypass `open`ing the commitlog in some cases, which
can be both convenient and avoids implicit creation of a segment when the
commitlog is empty.

None

2

Complexity is in diffing against the previous implementation, which lives in a
different repo. Risk is zero, as the code is not used in a released version.

# Testing

Adds tests.

### DIFF
--- a/crates/commitlog/src/stream/writer.rs
+++ b/crates/commitlog/src/stream/writer.rs
@@ -1,0 +1,428 @@
+use std::{
+    io::{self, Seek as _},
+    ops::Range,
+};
+
+use log::{debug, error, trace, warn};
+use tokio::{
+    io::{AsyncBufRead, AsyncBufReadExt as _, AsyncReadExt as _, AsyncWriteExt},
+    task::spawn_blocking,
+};
+
+use crate::{
+    commit, error,
+    repo::{self, Repo, Segment},
+    segment::{self, FileLike as _, OffsetIndexWriter, CHECKSUM_LEN, DEFAULT_CHECKSUM_ALGORITHM},
+    stream::common::{read_exact, AsyncFsync},
+    Options, StoredCommit, DEFAULT_LOG_FORMAT_VERSION,
+};
+
+use super::{
+    common::{peek_buf, AsyncLen as _, CommitBuf},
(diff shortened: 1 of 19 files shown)


crates/commitlog/src/stream/writer.rs
```diff
@@ -0,0 +1,428 @@
...
+            // Put back segment, so it gets closed if we break the loop.
+            self.current_segment = Some(current_segment);
+            match maybe_eof {
+                AppendInnerResult::StreamExhausted => break,
+                AppendInnerResult::SegmentMarker => continue,
+            }
+        }
+
+        if let Some(current_segment) = self.current_segment.as_mut() {
+            current_segment.flush_and_sync().await?;
```
Shubham8287, line 234: I have a ongoing PR where I am not fsyncing after every stream end - https://github.com/clockworklabs/SpacetimeDBPrivate/pull/1452/files#diff-0d8c22747cef364d129ae36bac70d4e6a304385c14d7fffdcfe75a400ca15ef9L727. 

Reason being: most of the stream could be empty or really small for replication, can we make it caller responsibility to fsync and only flush here.

Shubham8287 (approved):

Shubham8287:
Much needed consolidation!

crates/commitlog/src/stream/writer.rs
kim: The only issue is that we can't do much if the writer is (accidentally) dropped, due to the "async drop" problem. Let me see if I can come up with something.

Shubham8287: FYI: For replication, I was solving it by explicitly calling fsync on revceiving decide event. https://github.com/clockworklabs/SpacetimeDBPrivate/pull/1452/files#diff-465b61f1322cc15cd6d3bf87a373f0f1d0a20003846b857e842d1b538eb0ec1fR611

PR #2502, opened 2025-03-25
Various methods and trait impls for time arithmetic

### LINKED ISSUE #2367: Add `Timestamp` methods for arithmetic with `Duration`
BitCraft module, prior to the `Timestamp` overhaul, used `Timestamp::checked_add(self, duration: Duration) -> Option<Self>`. We should provide this same method, and likely any other arithmetic that `SystemTime` also offers.

gefjon:
# Description of Changes

This commit adds `checked_add` and `checked_sub` methods for arithmetic with `Timestamp` and `TimeDuration`, and implementations of `Add`, `Sub`, `AddAssign` and `SubAssign`. Prior to this, BitCraft's code had to do convoluted conversions from `Timestamp` into `SystemTime`, then do arithmetic, then convert back to `TimeStamp`.

Specifically, we add methods and trait impls for `T x U -> V`, where `x` is a binary operator, either + or -:

- `Timestamp x TimeDuration -> Timestamp`
- `Timestamp x Duration -> Timestamp`.
  - This one is kind of weird, since we convert the `Duration` into a `TimeDuration` before doing math with it, which may be a lossy conversion depending on platform.
- `TimeDuration x TimeDuration -> TimeDuration`.

These correspond to the operations that Rust exposes on `std::time::SystemTime` and `std::time::Duration`, plus allowing `Duration` as a convenience.

I have chosen not to implement
`TimeDuration x Duration` or `Duration x TimeDuration` operations, under the theory that if you're already working with a `TimeDuration`, you don't need to be shielded from the conversion.

Closes #2367 .

This PR adds new APIs which will become stabilized. It does not remove or change any existing APIs.

2? Some minor complexity in `sats::Timestamp x std::time::Duration` arithmetic.

# Testing

- [x] Wrote a couple of simple sanity-check proptests.


gefjon:
Presumably we should do something similar for C#.

crates/sats/src/time_duration.rs
```diff
@@ -69,6 +70,16 @@ impl TimeDuration {
...
+    /// Returns `Some(self + other)`, or `None` if that value would be out of bounds for [`TimeDuration`].
+    pub fn checked_add(self, other: Self) -> Option<Self> {
+        self.to_micros().checked_add(other.to_micros()).map(Self::from_micros)
+    }
+
+    /// Returns `Some(self - other)`, or `None` if that value would be out of bounds for [`TimeDuration`].
+    pub fn checked_sub(self, other: Self) -> Option<Self> {
+        self.to_micros().checked_sub(other.to_micros()).map(Self::from_micros)
+    }
 }
```
kazimuth, line 83: I was wondering about also adding `from_seconds` and `from_milliseconds` taking i32. Not having those makes blackholio messy. If they are added, we should add them to the C# SDK as well.

crates/sats/src/time_duration.rs
```diff
@@ -96,6 +107,40 @@ impl fmt::Display for TimeDuration {
     }
 }
 
+impl Add for TimeDuration {
```
kazimuth, line 110: C# does not have this for TimeDuration.

crates/sats/src/time_duration.rs
```diff
@@ -96,6 +107,40 @@ impl fmt::Display for TimeDuration {
...
+    type Output = Self;
+
+    fn add(self, rhs: Self) -> Self::Output {
+        self.checked_add(rhs).unwrap()
+    }
+}
+
+impl Sub for TimeDuration {
```
kazimuth, line 118: C# does not have this for TimeDuration.

kazimuth (commented):
This makes sense. If you wouldn't mind making the corresponding change at the same time in C# that would be nice, it's in `bindings-csharp/BSATN.Runtime/Builtins.rs`, there are already a few operator overrides in there.

I'm wondering if we should put the corresponding operations inside [https://stackoverflow.com/questions/2056445/no-overflow-exception-for-int-in-c](checked) blocks in C#. I think it would make sense for the C# API to always have exceptions thrown on overflow. Checked blocks are enforced [syntactically](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/checked-and-unchecked), so the user has no other way to enable checking for time operations on the C# side.

coolreader18 (approved):
LGTM

joshua-spacetime (approved):

PR #2034, opened 2024-12-03
Compress the snapshot

mamcx:
# Description of Changes

Compress the snapshot, using `Zstd`.

Closes #1592  & #1594.

# API and ABI breaking changes

It could read previously uncompressed data because it will detect using `magic bytes` the compression algorithm used, if any.

Because it wrap the `files` with `CompressReader` / `CompressWriter` is transparent to the rest of the engine.

# Expected complexity level and risk

1

# Testing

- [x] *Backwards Compatibility: Can replay both compressed and uncompressed segment files in the same snapshot*
- [x] *Added simple test that open/write the snapshot*
- [x] *Include a `[ignored]` test that using the environment variables `SNAPSHOT="path'", IDENTITY=`hex`* for check the compression ratio on existing data*
- [x] *Added benches for both cases*`
- [x] *Manually check with standalone & private that data get compressed*  

# Bench 

```bash
[crates/bench/benches/special.rs:234:9] &size = SnapshotSize {
    compressed_type: None,
    object_count   : 9,
    file_size      :      525 bytes,
    object_size    :   592145 bytes,
    total_size     :   592670 bytes,
}
[crates/bench/benches/special.rs:234:9] &size = SnapshotSize {
    compressed_type: Zstd,
    object_count   : 9,
    file_size      :      420 bytes,
    object_size    :    17896 bytes,
    total_size     :    18316 bytes,
}
[crates/bench/benches/special.rs:234:9] &size = SnapshotSize {
    compressed_type: Lz4,
    object_count   : 9,
    file_size      :      442 bytes,
    object_size    :    35989 bytes,
    total_size     :    36431 bytes,
}
[crates/bench/benches/special.rs:234:9] &size = SnapshotSize {
    compressed_type: Snap,
    object_count   : 9,
    file_size      :      447 bytes,
    object_size    :    53486 bytes,
    total_size     :    53933 bytes,
}
special/snapshot/synthetic/save_compression_None
                        time:   [3.4299 ms 3.4705 ms 3.5146 ms]
                        thrpt:  [160.82 MiB/s 162.86 MiB/s 164.79 MiB/s]
Found 1 outliers among 50 measurements (2.00%)
  1 (2.00%) high mild
special/snapshot/synthetic/open_compression_None
                        time:   [670.31 µs 671.53 µs 672.81 µs]
                        thrpt:  [840.07 MiB/s 841.68 MiB/s 843.21 MiB/s]
special/snapshot/synthetic/save_compression_Zstd
[...]

PR #2504, opened 2025-03-25
Commitlog compression

### LINKED ISSUE #1592: [STORAGE USE REDUCTION] Commitlog segment compression

coolreader18:
# Description of Changes

Resolves #1592. Takes `fs-utils/compression` from #2034.

Not sure which label applies, but this is a change that is not forwards-compatible once you upgrade.

3 - this is a bit tricky but I'm pretty confident in this implementation.

# Testing

- [x] commits are properly compressed to zstd format when a snapshot is taken.

### DIFF
--- a/crates/commitlog/src/segment.rs
+++ b/crates/commitlog/src/segment.rs
@@ -359,11 +359,11 @@ impl<R: io::Read + io::Seek> Reader<R> {
     }
 }
 
-impl<R: io::Read + io::Seek> Reader<R> {
+impl<R: io::BufRead + io::Seek> Reader<R> {
     pub fn commits(self) -> Commits<R> {
         Commits {
             header: self.header,
-            reader: io::BufReader::new(self.inner),
+            reader: self.inner,
         }
     }
 
@@ -393,7 +393,7 @@ impl<R: io::Read + io::Seek> Reader<R> {
 
     #[cfg(test)]
     pub(crate) fn metadata(self) -> Result<Metadata, error::SegmentMetadata> {
-        Metadata::with_header(self.min_tx_offset, self.header, io::BufReader::new(self.inner))
+        Metadata::with_header(self.min_tx_offset, self.header, self.inner)
     }
 }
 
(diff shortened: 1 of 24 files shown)


crates/durability/src/lib.rs
```diff
@@ -42,6 +42,8 @@ pub trait Durability: Send + Sync {
     /// either because nothing has been persisted yet, or because the status
     /// cannot be retrieved.
     fn durable_tx_offset(&self) -> Option<TxOffset>;
+
+    fn compress_all_before(&self, offset: TxOffset);
```
kim, line 46: I think this is the wrong direction of control: a `Durability` impl may not be backed directly by a commitlog. And conversely, not all processes that write to a commitlog do so through `Durability`.

A safe offset must also be at or before `durable_tx_offset` (aka decided offset), and it must be before the most recent (active) segment. IOW, we should only compress immutable segments, as otherwise we'll need to support trimming the undecided suffix from a compressed segment, which is just unnecessary complexity.

The `RelationalDB` cannot decide when a segment is immutable (although standalone can).

I also don't think all `Segment` impls need to or should support compression, it's totally sufficient if files do. This means that most of the complexities (such as temp files, looking at magic numbers, etc) can be confined to just `repo::Fs`.

I would suggest that `Segment` gains a method `compress` that handles all the details. When it is called is edition-specific.

crates/commitlog/src/segment.rs
```diff
@@ -22,23 +22,28 @@ pub const MAGIC: [u8; 6] = *b"(ds)^2";
...
     pub checksum_algorithm: ChecksumAlgorithm,
+    pub compression: SegmentCompression,
 }
 
 #[derive(Clone, Copy, Debug, Eq, PartialEq, Ord, PartialOrd)]
 #[repr(u8)]
 pub enum LogFormatVersion {
     V0,
     V1,
+    V2,
```
kim, line 33: I don't think we need a new version -- zstd has a magic number which we can expect to appear before the commitlog magic number in a stream.

mamcx: Yes, take a look at https://github.com/clockworklabs/SpacetimeDB/pull/2034/files#diff-e4988280b2a00468abcbf980d6024d0f07d179b322c732057bb6d74f5f3583aaR31

kim (approved):
Looks good!

PR #2506, opened 2025-03-26
Plain enums as index keys with specialized indices

Centril:
# Description of Changes

Commits:

1. Allow plain enums as index keys in Rust and C# bindings. For Rust, `FilterableValue` is impld when deriving `SpacetimeType`.
2. Specialize indices on plain enums by using u8 as the native type. For direct indices on plain enums, the entire index allocated at index creation to the number of tags. This should be quite efficient as the number of tags is expected to be low in the common case.

None

2?

# Testing

SDK tests are added exercising both the host, the modules, and the sdk. New tests are also added for the new fixed capacity direct index type.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5342,6 +5342,7 @@ dependencies = [
  "tikv-jemalloc-ctl",
  "tikv-jemallocator",
  "tokio",
+ "tokio-metrics",
  "tokio-stream",
  "tokio-util",
  "toml 0.8.19",
@@ -5652,17 +5653,29 @@ dependencies = [
 name = "spacetimedb-snapshot"
 version = "1.1.0"
 dependencies = [
+ "anyhow",
  "blake3",
+ "bytes",
+ "env_logger 0.10.2",
+ "futures",
  "hex",
  "log",
+ "pretty_assertions",
+ "spacetimedb-core",
  "spacetimedb-durability",
  "spacetimedb-fs-utils",
  "spacetimedb-lib",
  "spacetimedb-paths",
  "spacetimedb-primitives",
  "spacetimedb-sats",
+ "spacetimedb-schema",
  "spacetimedb-table",
+ "tempfile",
  "thiserror 1.0.69",
+ "tokio",
+ "tokio-stream",
+ "tokio-util",
+ "zstd-framed",
 ]
 
 [[package]]
@@ -6443,6 +6456,18 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn 2.0.87",
+]
+
+[[package]]
+name = "tokio-metrics"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cb2bb07a8451c4c6fa8b3497ad198510d8b8dffa5df5cfb97a64102a58b113c8"
+dependencies = [
+ "futures-util",
+ "pin-project-lite",
+ "tokio",
+ "tokio-stream",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -248,7 +248,9 @@ termcolor = "1.2.0"
 thin-vec = "0.2.13"
 thiserror = "1.0.37"
 tokio = { version = "1.37", features = ["full"] }
+tokio_metrics = { version = "0.4.0" }
 tokio-postgres = { version = "0.7.8", features = ["with-chrono-0_4"] }
+tokio-stream = "0.1.17"
 tokio-tungstenite = { version = "0.26.2", features = ["native-tls"] }
 tokio-util = { version = "0.7.4", features = ["time"] }
 toml = "0.8"
--- a/crates/bindings-csharp/Codegen.Tests/fixtures/diag/Lib.cs
+++ b/crates/bindings-csharp/Codegen.Tests/fixtures/diag/Lib.cs
@@ -410,6 +410,8 @@ public partial struct TestScheduleIssues
 
 public partial class Module
 {
+#pragma warning disable STDB_UNSTABLE // Enable ClientVisibilityFilter
+
     // Invalid: not public static readonly
     [SpacetimeDB.ClientVisibilityFilter]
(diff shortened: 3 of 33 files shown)


modules/sdk-test/src/lib.rs
```diff
@@ -542,6 +542,20 @@ define_tables! {
...
+}
+
+#[spacetimedb::reducer]
+fn update_pk_simple_enum(ctx: &ReducerContext, a: SimpleEnum, data: i32) -> anyhow::Result<()> {
+    let Some(mut o) = ctx.db.pk_simple_enum().a().find(&a) else {
+        return Err(anyhow!("row not found"));
+    };
+    o.data = data;
+    ctx.db.pk_simple_enum().a().update(o);
+    Ok(())
```
gefjon, line 558: I think this method should be expressible through the macro as `update_by update_pk_simple_enum = update_by_a(a),`. We should also probably include the `delete_by` line, so that all of the `Pk*` tables are defined the same way.

crates/table/src/read_column.rs
```diff
@@ -339,6 +341,30 @@ impl_read_column_via_from! {
...
+unsafe impl ReadColumn for SumTag {
+    fn is_compatible_type(ty: &AlgebraicTypeLayout) -> bool {
+        matches!(ty, AlgebraicTypeLayout::Sum(_))
+    }
+
+    unsafe fn unchecked_read_column(row_ref: RowRef<'_>, layout: &ProductTypeElementLayout) -> Self {
+        debug_assert!(Self::is_compatible_type(&layout.ty));
+
+        let (page, offset) = row_ref.page_and_offset();
+        let col_offset = offset + PageOffset(layout.offset);
```
gefjon, line 355: If we're going to write this method in the maximally-defensive style (using `mem::size_of::<Self>()` and reading an array, rather than just reading a `u8`), we might as well also respect `SumTypeLayout::offset_of_tag` here.

crates/table/src/table_index/mod.rs
```diff
@@ -280,11 +293,15 @@ impl TypedIndex {
                 AlgebraicType::U16 => Self::UniqueDirectU16(<_>::default()),
                 AlgebraicType::U32 => Self::UniqueDirectU32(<_>::default()),
                 AlgebraicType::U64 => Self::UniqueDirectU64(<_>::default()),
+                // For a plain enum, use ~`u8` as the native type.
```
gefjon, line 296:
```suggestion
                // For a plain enum, use `u8` as the native type.
```

Was this tilde intentional?

crates/table/src/table_index/mod.rs
```diff
@@ -302,6 +319,9 @@ impl TypedIndex {
                 AlgebraicType::I256 => UniqueBtreeI256(<_>::default()),
                 AlgebraicType::U256 => UniqueBtreeU256(<_>::default()),
                 AlgebraicType::String => UniqueBtreeString(<_>::default()),
+                // For a plain enum, use ~`u8` as the native type.
```
gefjon, line 322: Seems like maybe it was. What's the tilde supposed to mean?

gefjon (commented):
I reviewed only the parts you asked for, namely the SDK tests (both module and client parts) and the stuff in the `table` crate.

Minor style nits. Otherwise looks good!

I wonder if we should even bother to use unique BTree indices for simple enums, or if we should just impl a unique index on a simple enum column as a direct index. I wonder the same thing about u8 and i8 columns. This is idle speculation, though.

coolreader18 (approved):
LGTM

crates/bindings-macro/src/sats.rs
```diff
@@ -205,7 +209,41 @@ pub(crate) fn derive_satstype(ty: &SatsType<'_>) -> TokenStream {
...
 
+    let impl_plain_enum_extras = if add_impls_for_plain_enum {
+        // These will mostly be empty as lifetime and type parameters must be constrained
+        // but const parameters don't require constraining.
+        let mut generics = ty.generics.clone();
+        add_type_bounds(&mut generics, &quote!(#krate::FilterableValue));
+        let (impl_generics, ty_generics, where_clause) = generics.split_for_impl();
+        // As we don't have access to other `derive` attributes,
+        // we don't know if `Copy` was derived,
+        // so we won't impl for the owned type.
```
gefjon, line 220: I think this is probably the wrong default assumption. We're only deriving for plain enums, which almost always will be `Copy`. IMO we should assume that they will be and emit the owned-value derive too, as having it is a much better user experience than not.

gefjon: God, I wish Rust allowed using non-generic `where` bounds on items for trait-bounded conditional compilation.

gefjon: The comment which introduces the redundant lifetime is very compelling. We may want to experiment with that in the future. But for now, my inclination is just to assume that all simple enums will be `Copy` and go from there.

crates/bindings-csharp/Runtime/Internal/Bounds.cs
```diff
@@ -18,7 +18,6 @@ public interface IBTreeIndexBounds
 }
 
 public readonly struct Bound<T>(T min, T max)
-    where T : IEquatable<T>
```
kazimuth, line 21: Why were these bounds removed? Does the codegen fail to implement `IEquatable` somewhere, e.g. for plain `enum`s? I haven't looked at C# codegen for plain `enum`s at all.

kazimuth (commented):
The C# side of this generally makes sense. I would prefer to implement `IEquatable` for plain enums marked with `[SpacetimeDB.Type]` and keep the bounds on various interfaces -- I think I missed implementing `IEquatable` for plain enums in my IEquatable PR.

crates/bindings-csharp/Runtime/Internal/Bounds.cs
kazimuth: And you can't implement interfaces for `enum`s, right. Ah well.

gefjon (approved):

PR #2516, opened 2025-03-27
Rename module name to database name

### LINKED ISSUE #2515: Rename with_module_name to with_database_name
It looks like the name which you can provide in the CLI during publishement of a module isn't the module name, it's the database name. Also in the docs there were some examples which used variables like "DB_NAME" for input of with_module_name / withModuleName for client connections and more. I've created a PR changing the docs so that they mention "database" name instead of "module" name, but the CLI als well as the client SDKs (and possibly the server module library) need changes accordingly.

tamaro-skaljic:
# Description of Changes

Fixes #2515

Breaking change of the connection API of client SDKs. (can't add the label)

1

# Testing

- [ ] Make sure that there are no other locations where it needs to be renamed.

# Documentation

- [ ] search for `with.*module.*name` in the docs repository and replace any occurrences with `with_database_name` / `withDatabaseName`.


gefjon:
Unfortunately, we can't make this change. `DbConnectionBuilder::with_module_name` is a part of our stable user-facing API, and so changing it would require a SemVer major version bump, which we're not willing to do at this time. We may discuss how to gradually transition to using `with_database_name` instead, but it's not something we can rush into.

cloutiertyler (approved):
This now LGTM

cloutiertyler (changes requested):
Retracting my approval. I was confused on PR numbers. We don't want to change the name of `with_module_name` at this time.

tamaro-skaljic:
> Unfortunately, we can't make this change. `DbConnectionBuilder::with_module_name` is a part of our stable user-facing API, and so changing it would require a SemVer major version bump, which we're not willing to do at this time. We may discuss how to gradually transition to using `with_database_name` instead, but it's not something we can rush into.

I've reworked it in https://github.com/clockworklabs/SpacetimeDB/pull/2516/commits/18624e5bd00217858bbf88e14b33adb8b521adad so that no major version bump is needed.

I would say it's better to create the new default now and update the docs and example projects so that new users use the `with_database_name` method of the DBConnectionBuilder instead of the `with_module_name` method. Then they'll have one breaking change less to migrate if `with_module_name` is removed. If this is merged I'll create the PRs for the docs and example pprojects too. :)

PR #2517, opened 2025-03-27
Add endpoints for heap profiling

jsdt:
# Description of Changes

This adds endpoints to make heap profiling easier. By default, standalone will have jemalloc heap profiling enabled, but not active, which didn't show any overhead in my benchmarks.

If you activate profiling, with `curl "localhost:3000/internal/heap/settings?enabled=true" -X POST`, jemalloc will start sampling allocations. It will keep sampling until you call it again with `enabled=true` (or restart the process).

You can get a dump of the heap by `GET`ing the `internal/heap` endpoint. By default, this will be in `pprof` format (see [pprof](https://github.com/google/pprof) for installing a tool to view it). If you use the `format=flame` query parameter, it will give a flame graph instead. You can easily view the flame graph in your browser by going to `localhost:3000/internal/heap?format=flame`

There is some additional overhead while profiling is activated, because we are collecting samples, and using memory to store those samples, but it should be safe to try this out.

Note that the jemalloc configurations settings are set with the `_rjem_malloc_conf` variable in the main file of standalone, but this can be overridden by setting the `_RJEM_MALLOC_CONF` environment variable. You can learn more about it [here](https://jemalloc.net/jemalloc.3.html#tuning).

## Extra background

Previously, it was possible to set up heap profiling by setting the environment variable `_RJEM_MALLOC_CONF` to something like `prof:true,prof_active:true,lg_prof_sample:12,prof_prefix:/jemalloc-dumps/,lg_prof_interval:20`, which would make the program sample allocations and dump a heap profile to disk periodically. Using these dumps was a little annoying with docker, because the `pprof` tool needs to use the original binary to give readable symbols.

To make it easier, this uses the [jemalloc-pprof](https://github.com/polarsignals/rust-jemalloc-pprof) crate, which takes care of attaching symbols/backtraces for us. This means we can parse the output without needing access to the binary. This adds some extra overhead, but we can always remove it if we decide it is too slow.

I added an `/internal` section of the http API (outside of v1), with the idea that this is an undocumented feature that we may change or remove at any time.

2. The main risk here is around the overhead when using it. There is very little risk of performance issues while profiling is not active.

# Testing

Tested manually locally, by curling the different endpoints, and also by setting `_RJEM_MALLOC_CONF` to `prof:false` to totally disable profiling.


crates/sdk/tests/test-client/src/module_bindings/mod.rs
```diff
@@ -292,7 +292,6 @@ pub mod unique_u_64_type;
 pub mod unique_u_8_table;
 pub mod unique_u_8_type;
 pub mod unit_struct_type;
-pub mod update_btree_u_32_reducer;
```
bfops, line 295: why was this stuff changed?

bfops:
Just to check with @jdetter as someone who ends up doing a lot of profiling - are you happy with the pprof choice here?

crates/standalone/src/main.rs
```diff
@@ -46,6 +46,15 @@ use tikv_jemallocator::Jemalloc;
 #[global_allocator]
 static GLOBAL: Jemalloc = Jemalloc;
 
+// Defaults our jemalloc configuration to allow for profiling, but have it disabled initially.
+// It can be enabled with an internal endpoint.
+// When enabled, it will sample once per 2^19 bytes (512 KB) of memory allocated.
+
+// This can be overridden by setting the `_RJEM_MALLOC_CONF` environment variable.
+#[allow(non_upper_case_globals)]
+#[export_name = "_rjem_malloc_conf"]
```
bfops, line 55: it looks like the `export_name` is the same as the actual name? what am I overlooking?

crates/standalone/src/main.rs
```diff
@@ -46,6 +46,15 @@ use tikv_jemallocator::Jemalloc;
...
+pub static _rjem_malloc_conf: &[u8] = b"prof:true,prof_active:false,lg_prof_sample:19\0";
```
bfops, line 56: is this basically a special specific name that jemalloc will look for? or where does this get used?

bfops:
to confirm - by "tested manually locally", you mean you were able to turn on profiling and then get jemalloc sample data successfully? (and then turn it off via the endpoint as well?)

bfops:
I left some questions but this broadly looks good to me. I can't really speak to the details of whether or not we're using the jemalloc-specific stuff correctly, but the fallout seems low if we've gotten something wrong.

crates/standalone/src/main.rs
jsdt: Yes, jemalloc looks this up somehow.

crates/sdk/tests/test-client/src/module_bindings/mod.rs
jsdt: I don't know, but running `cargo test  --all` on master causes these changes, so I'm guessing this is a result of a previous PR that didn't regenerate the test client bindings.

crates/standalone/src/main.rs
jsdt: I'm just following the pattern of their docs. The `export_name` part has something to do with how the jemalloc library looks it up.

jsdt:
Yes, I have done all of those tests.

crates/standalone/src/main.rs
gefjon: Can you add a comment linking to those docs?

crates/sdk/tests/test-client/src/module_bindings/mod.rs
bfops: Huh, TIL our CI doesn't check for that. I'm going to add a CI check in https://github.com/clockworklabs/SpacetimeDB/pull/2531 and regenerate them there.

bfops (approved):
LGTM. I can't really speak to the details of whether or not we're using the jemalloc-specific stuff correctly, but the fallout seems low if we've gotten something wrong.

Issue #2442, opened 2025-03-12
RLS: Implement view expansion


joshua-spacetime:
Expanding a table reference into its RLS definition during query compilation. For example, if you have the following RLS rule:

```rust
/// A user only has access to their row
#[spacetimedb::client_visibility_filter]
const USER_FILTER: Filter = Filter::Sql("
    SELECT * FROM user WHERE identity = @sender
");
```

The query `SELECT * FROM user` will be expanded into:

```sql
SELECT * FROM (SELECT * FROM user WHERE identity = @sender)
```

PR #2519, opened 2025-03-27
Add utility for rls resolution

joshua-spacetime:
# Description of Changes

This is the first of several patches that will close/implement #2442. This one in particular adds a utility for expanding or resolving RLS rules in a logical query expression, but it does not integrate it into compiler. That will happen in a follow on patch along with integration tests.

None

2.5

The logic is straightforward. There are a few parts that may be non-obvious like tree rotations and alpha conversion. I've tried to add ample documentation around those parts.

# Testing

Integration tests will be added with in the integration patch.

### DIFF
--- a/crates/expr/src/rls.rs
+++ b/crates/expr/src/rls.rs
@@ -1,0 +1,407 @@
+use std::rc::Rc;
+
+use spacetimedb_lib::identity::AuthCtx;
+use spacetimedb_primitives::TableId;
+use spacetimedb_sql_parser::ast::BinOp;
+
+use crate::{
+    check::{parse_and_type_sub, SchemaView},
+    expr::{Expr, FieldProject, LeftDeepJoin, ProjectName, RelExpr, Relvar},
+};
+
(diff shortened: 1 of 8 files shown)


crates/expr/src/rls.rs
```diff
@@ -0,0 +1,656 @@
...
+    }
+
+    /// Parse, type check, and resolve RLS rules
+    fn resolve(sql: &str, tx: &impl SchemaView, auth: &AuthCtx) -> anyhow::Result<Vec<ProjectName>> {
+        let (expr, _) = parse_and_type_sub(sql, tx, auth)?;
+        resolve_views(tx, expr, auth, &mut false)
+    }
+
+    #[test]
+    fn test_rls_for_owner() -> anyhow::Result<()> {
```
joshua-spacetime, line 493: I added these tests mainly as a sanity check, but they're a bit verbose. Feel free to skip them if you'd like. Integration tests will be added in another PR.

mamcx (approved):
LGTM

Issue #2511, opened 2025-03-26
`clockworklabs/spacetime` docker image should not use spacetimedb-update


coolreader18:
Should probably just build from source. Otherwise you could get into a bad situation where a user has `clockworklabs/spacetime:1.0.1` but has updated the cli inside to 1.2.0 and is confused why they're running into issues.

PR #2521, opened 2025-03-28
Make clockworklabs/spacetime docker image build from source

### LINKED ISSUE #2509: Clean up `clockworklabs/spacetime` image
- Fix whatever's going on with the glibc version.
- Ensure that it contains whatever SSL packages are necessary to avoid #2482 .

coolreader18:
# Description of Changes

Fixes #2509, #2511.
Motivation from the latter:
> Should probably just build from source. Otherwise you could get into a bad situation where a user has clockworklabs/spacetime:1.0.1 but has updated the cli inside to 1.2.0 and is confused why they're running into issues.

The dotnet install script is kinda flaky right now, see https://github.com/dotnet/install-scripts/issues/582. If it's not building correctly replace `https://dot.net/v1/dotnet-install.sh` with `https://builds.dotnet.microsoft.com/dotnet/scripts/v1/dotnet-install.sh` and it should be exactly equivalent.

1

# Testing

- [x] Tested that `spacetime` works and `cargo` exists as a command.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,11 @@
 # Use a base image that supports multi-arch
+FROM rust:bookworm AS builder
+
+WORKDIR /usr/src/app
+COPY . .
+
+RUN cargo build -p spacetimedb-standalone -p spacetimedb-cli --release --locked
+
 FROM rust:bookworm
 
 # Install dependencies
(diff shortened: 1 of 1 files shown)


jdetter (approved):
This seems to work, thanks @coolreader18 ! Tested via:

```bash
docker build -t spacetime-test .
docker run --rm -p 3000:3000 spacetime-test start
```

Then I logged in via
```
docker exec -it <name> /bin/bash
```

And I was able to spacetime normally, including publishing to the local server.

PR #2525, opened 2025-03-28
Fix fields named 'read'

kazimuth:
# Description of Changes

Fixes https://github.com/orgs/clockworklabs/projects/22?pane=issue&itemId=102392974&issue=clockworklabs%7Ccom.clockworklabs.spacetimedbsdk%7C276

by renaming `internal` `static` serializer fields so that they do not overlap with user-provided names.

Note, however, that some field names still will not work: `ReadFields`, `WriteFields`, `Equals`, and `GetHashCode`.
This would require a separate fix since the error would happen in a different place. In this case we would need to change the name of the generated member to something like `ReadFields_` or `Equals_`, which I'm not sure is a good idea.

N/A

0

# Testing

SDK tests and `dotnet-verify` tests are passing.

### DIFF
--- a/crates/bindings-csharp/Codegen.Tests/fixtures/server/snapshots/Module#FFI.verified.cs
+++ b/crates/bindings-csharp/Codegen.Tests/fixtures/server/snapshots/Module#FFI.verified.cs
@@ -690,7 +690,7 @@ namespace SpacetimeDB
                     ],
                     Schedule: null,
                     TableType: SpacetimeDB.Internal.TableType.User,
-                    TableAccess: SpacetimeDB.Internal.TableAccess.Private
+                    TableAccess: SpacetimeDB.Internal.TableAccess.Public
                 );
 
             public ulong Count =>


joshua-spacetime:
> Note, however, that some field names still will not work: `ReadFields`, `WriteFields`, `Equals`, and `GetHashCode`.
This would require a separate fix since the error would happen in a different place. In this case we would need to change the name of the generated member to something like `ReadFields_` or `Equals_`, which I'm not sure is a good idea.

Makes sense. What do you think @jdetter?

jdetter:
I don't super care about `Equals` or `HashCode` because I don't think we want users overriding these anyway. I think we should fix the cases for `ReadFields` and `WriteFields` because I could easily see someone creating a table that has a column with those names. (e.g. if you made a permissions table or something)

@kazimuth @joshua-spacetime does that sound good?

jdetter:
@kazimuth Does this actually not have any API/ABI breaking changes? This doesn't require a C# SDK change?

rekhoff:
Attempting to merge master with this PR ran into build issues, mostly due to conflicts with #2725 .
I am working to resolve the conflicts and get another build going. I have an initial version ready that I need to test locally before pushing it to this PR.

rekhoff:
After additional fixes, all tests pass. I have tested the most recent changes against BitCraft and other minor tests, without issue.
I am making myself as an Assignee, due to the additional work. With that in mind, @jdetter would you mind reviewing/approving the changes?

rekhoff:
Tested this change against the "fixture" test and adding the fields to a quickstart-chat project. Both succeeded (after a minor complication error that is now patched)

jdetter (approved):
This looks fine to me, me and Ryan have gone over these changes together + he has added a test to make sure the fields are working correctly. Thanks Ryan 👍

PR #2526, opened 2025-03-28
Add rls to the sql api

joshua-spacetime:
# Description of Changes

Incorporates RLS resolution into the sql code path - but not the subscription code path. Subscriptions will come in a separate patch.

None

1.5

Mostly mechanical updates.

# Testing

- [ ] Planner tests
- [x] Correctness tests
- [x] Smoke tests

### DIFF
--- a/crates/expr/src/expr.rs
+++ b/crates/expr/src/expr.rs
@@ -89,6 +89,50 @@ impl ProjectName {
 ///
 /// Note that RLS takes a single expression and produces a list of expressions.
 /// Hence why these variants take lists rather than single expressions.
+///
+/// Why does RLS take an expression and produce a list?
+///
+/// There may be multiple RLS rules associated to a single table.
+/// Semantically these rules represent a UNION over that table,
+/// and this corresponds to a UNION in the original expression.
+///
+/// TODO: We should model the UNION explicitly in the physical plan.
+///
+/// Ex.
+///
+/// Let's say we have the following rules for the `users` table:
+/// ```rust
+/// use spacetimedb::client_visibility_filter;
+/// use spacetimedb::Filter;
+///
+/// #[client_visibility_filter]
+/// const USER_FILTER: Filter = Filter::Sql(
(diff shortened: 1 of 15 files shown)


crates/physical-plan/src/plan.rs
```diff
@@ -105,31 +113,21 @@ impl ProjectPlan {
...
-    type Target = PhysicalPlan;
-
-    fn deref(&self) -> &Self::Target {
-        match self {
-            Self::Name(plan) => plan,
-            Self::List(plan, ..) | Self::Agg(plan, ..) => plan,
-            Self::Limit(plan, ..) => plan,
-        }
-    }
+    Agg(Vec<PhysicalPlan>, AggType),
```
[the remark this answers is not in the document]
joshua-spacetime: FWIW, they are reflected in the logical plan's doc comments, but I will also update the physical plan.

crates/expr/src/expr.rs
```diff
@@ -86,12 +86,15 @@ impl ProjectName {
...
 #[derive(Debug)]
 pub enum ProjectList {
-    Name(ProjectName),
-    List(RelExpr, Vec<(Box<str>, FieldProject)>),
+    Name(Vec<ProjectName>),
+    List(Vec<RelExpr>, Vec<(Box<str>, FieldProject)>),
     Limit(Box<ProjectList>, u64),
-    Agg(RelExpr, AggType, Box<str>, AlgebraicType),
+    Agg(Vec<RelExpr>, AggType, Box<str>, AlgebraicType),
 }
```
mamcx, line 98: I don't love this change, because is not clear why need to be a list. And later it look like the first one have a semantic mean.


I think it will be better to have a wrapper that encode the semantics or at least a comment here about what is the expectation of the content.

joshua-spacetime: I updated the comment.

mamcx (approved):
LGTM

PR #2532, opened 2025-03-31
Update data size metrics periodically, not on every transaction.

jsdt:
# Description of Changes

Instead of updating data-size-related gauges after each transaction, we report the metrics every 15 seconds. This should reduce overhead for databases with high traffic. 

corresponding GitHub label.

1. The main risk here is extra overhead if there are a large number of idle databases. We could mitigate that in the future by returning the `next_tx_offset` and skipping reports if it hasn't changed.

# Testing

I ran a standalone instance and checked that:
 1. We report data size metrics right after a module is loaded.
 2. The size metrics change if I add data.

I haven't really looking into writing unit tests for it, since this is tied in with the module lifecycle logic.

### DIFF
--- a/crates/core/src/db/datastore/locking_tx_datastore/committed_state.rs
+++ b/crates/core/src/db/datastore/locking_tx_datastore/committed_state.rs
@@ -679,7 +679,7 @@ impl CommittedState {
         (table, blob_store)
     }
 
-    pub(super) fn report_data_size(&self, database_identity: Identity) {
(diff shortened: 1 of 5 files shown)


bfops:
> Instead of updating data-size-related gauges after each transaction, we report the metrics every 15 seconds. This should reduce

Was this cut off?

jsdt:
> > Instead of updating data-size-related gauges after each transaction, we report the metrics every 15 seconds. This should reduce
> 
> Was this cut off?

Yup. I updated it.

PR #2536, opened 2025-04-01
core: Export the snapshot watcher / commitlog compressor

kim:
Moves the task from standalone to core and exports the function.
This allows to re-use the definition for non-replicated databases.

1

# Testing

Only moves code.


kim:
I think I want to add a test for the segment slice selection before merging.

coolreader18 (approved):

PR #2539, opened 2025-04-01
Update tungstenite to 0.26

coolreader18:
# Description of Changes

Tungstenite `0.26` switches its `Message` to use `Bytes` instead of `Vec<u8>` for the payload, meaning we can now send messages without making a duplicate allocation for every client.

2: updating a dependency, but it's pretty low risk.

# Testing

n/a - general optimization


coolreader18:
This pulls in `rand 0.9` now, but I can make a followup to deduplicate that dependency.

PR #2543, opened 2025-04-02
Rip Rayon out of reducer execution and query evaluation

gefjon:
For testing/benchmarking purposes. One theory about our performance is that we're spending a lot of time context-switching between Tokio and Rayon threads. This build will be used in the first of a series of experiments to investigate that overhead. In this patch, we just do sequential evaluation on the Tokio worker threads where the rest of our code runs, instead of sending stuff to Rayon. Rayon use is mostly, but not entirely, removed from Core.

The two next steps I am interested in:
- Use parallelism, but on Tokio workers rather than Rayon threads. I.e. replace `par_iter`, `fold` and `reduce_with` calls with Tokio-isms, instead of this patch's `std::iter::Iterator` versions.
- (Not discussed in meeting) continue using `par_iter` and friends, but invoke the "outer loop" from Tokio threads. I.e. retain use of `par_iter`, `fold` and `reduce_with`, but remove calls to `rayon::scope` or `rayon::spawn`.

### DIFF
--- a/crates/core/src/host/wasm_common/module_host_actor.rs
+++ b/crates/core/src/host/wasm_common/module_host_actor.rs
@@ -462,9 +462,9 @@ impl<T: WasmInstance> WasmModuleInstance<T> {
         )
         .entered();
 
-        // run the call_reducer call in rayon. it's important that we don't acquire a lock inside a rayon task,
-        // as that can lead to deadlock.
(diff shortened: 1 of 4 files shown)


cloutiertyler:
Closing as this is not intended to merge. There have been significant learnings since this PR was created.

PR #2547, opened 2025-04-03
Add docs for `FilterableValue` to rust bindings crate

gefjon:
# Description of Changes

Added docs for `FilterableValue`, and made that trait visible so that it shows up in the docs.

I also added a `#[doc(hidden)]` "private" supertrait bound to `FilterableValue`, to discourage use.

N/a.

1. Mostly docs.

# Testing

- [x] Ran `cargo doc --open` and read the new docs, which looked good (if I do say so myself). Clicked through links and whatnot.
- [ ]


crates/lib/src/filterable_value.rs
```diff
@@ -87,7 +110,7 @@ pub enum TermBound<T> {
 }
 impl<Bound: FilterableValue> TermBound<&Bound> {
     #[inline]
-    /// If `self` is [`TermBound::Range`], returns the `rend_idx` value for [`BTreeScanArgs`],
+    /// If `self` is [`TermBound::Range`], returns the `rend_idx` value for `IndexScanRangeArgs`,
```
gefjon, line 113: This change to quiet a warning from `cargo doc`. `BTreeScanArgs` is no longer called that, and also is not available in this crate.

crates/cli/src/subcommands/sql.rs
```diff
@@ -160,27 +160,12 @@ pub(crate) async fn run_sql(builder: RequestBuilder, sql: &str, with_stats: bool
...
-                .map(|value| satn::PsqlWrapper { ty: ty.ty(), value }.to_string()),
-        );
-    }
 
-    let mut table = builder.build();
-    table.with(Style::psql());
+    let table = build_table(
+        schema,
+        rows.iter().map(|row| from_json_seed(row.get(), SeedWrapper(ty))),
+    )?;
```
bfops, line 168: why was this stuff changed? it's not obvious to me from the PR description.

gefjon: ...Or maybe not. Sigh.

gefjon: Ok, I had to rebase and force-push, but this should be fixed.

bfops:
I'm removing the auto-added review requests for the CLI folks. I hope that's correct. I don't have any helpful input into the rest of this PR.

PR #2548, opened 2025-04-03
Add CI job to run `cargo doc` on the bindings crate

gefjon:
# Description of Changes

Like the title says. I noticed we had some broken links in doc comments while working on #2547 , and thought it would be wise to detect them in CI.

N/a.

1.

# Testing

- [x] Ran this commands by hand and saw that it rejected broken links.
- [x] [Failed in CI the first time because we had broken docs links in master](https://github.com/clockworklabs/SpacetimeDB/actions/runs/14246191034/job/39927673298?pr=2548).
- [x] I'll push a new commit which fixes the broken links, and then it will pass.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -191,6 +191,19 @@ jobs:
           dotnet tool restore
           dotnet csharpier --check .
 
+      - name: Run `cargo doc` for bindings crate
+        # `bindings` is the only crate we care strongly about documenting,
+        # since we link to its docs.rs from our website.
+        # We won't pass `--no-deps`, though,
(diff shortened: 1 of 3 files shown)


bfops (approved):
The CI part LGTM. If the other changes make it pass, I assume they're correct too :+1:

gefjon:
From a commit message:

> This requires hiding `Filter` behind the feature flag,
> in addition to the `client_visibility_filter` macro,
> since the doc comment for `Filter` links to the macro.
> `Filter` is not useful outside of its connection with the macro,
> so feature-gating it seems fine.

PR #2549, opened 2025-04-03
Use Offset Index on Meta extract

Shubham8287:
# Description of Changes
1. use offset index to fetch `Metadata`
2. `reset_to_internal` is neither using offset index not truncating from it: fix that.
3. down level some noisy log.

- [ ] 
- [ ]


crates/commitlog/src/segment.rs
```diff
@@ -573,6 +601,78 @@ impl Metadata {
 
         Ok(sofar)
     }
+
+    /// Finds the last valid commit in the segment using the offset index.
+    /// It traverses the index in reverse order, starting from the last key.
```
Shubham8287, line 606: This assume there could be more than last commit rather a tail of commits could be erased or corrupted. Did I went overbaord here?

kim: You're absolutely right.

crates/commitlog/src/stream/writer.rs
```diff
@@ -107,36 +107,49 @@ where
...
-            error::SegmentMetadata::InvalidCommit { sofar, source } => match on_trailing {
-                OnTrailingData::Error => Err(io::Error::new(io::ErrorKind::InvalidData, source)),
+        let offset_index = repo
+            .get_offset_index(last)
+            .inspect_err(|e| {
+                warn!("unable to open offset index for segment {last}: {e}");
+            })
+            .ok();
+
+        let (meta, offset_index) = match segment::Metadata::extract(last, &mut segment, offset_index.as_ref()) {
```
kim, line 117: I think this would be a bit easier to read if the conversion to `TxOffsetIndexMut` happened after the match, and the truncation would just borrow a mutable index like

```rust
if let Some(idx) = offset_index.as_mut().map(IndexFile::as_mut) {
// ...
}
```

PR #2552, opened 2025-04-04
Phoebe/parallelism test/rayon par iter only

gefjon:
For testing/benchmarking purposes. Follow-up to #2543 .

This experiment is intended to test the theory that our synchronization overhead is coming from moving work off the "main thread" to Rayon workers via `scope` and `spawn`, not from `par_iter` parallel iteration. If that is the case, we should see synchronization overhead (i.e. `futex_wait` time) decreased similar to the previous experiment relavite to the control, but also see improved CPU utilization across multiple cores.


cloutiertyler:
Closing as this is not intended to merge. There have been significant learnings about CPU caches since this was created.

PR #2556, opened 2025-04-04
Feature gate [ClientVisibilityFilter] in C#

joshua-spacetime:
# Description of Changes

Places the `[ClientVisibilityFilter]` behind the `STDB_UNSTABLE` experimental feature flag. Will emit a compiler warning unless the following is added to the module file
```csharp
#pragma warning disable STDB_UNSTABLE // Enable experimental SpacetimeDB features
```

None

0

# Testing

The build failed until I enabled `STDB_UNSTABLE` in the `sdk-test-cs` module.

### DIFF
--- a/crates/bindings-csharp/Codegen.Tests/fixtures/diag/Lib.cs
+++ b/crates/bindings-csharp/Codegen.Tests/fixtures/diag/Lib.cs
@@ -1,7 +1,5 @@
 using SpacetimeDB;
 
-#pragma warning disable STDB_UNSTABLE // Enable experimental SpacetimeDB features
-
 public enum LocalEnum { }
 
 [SpacetimeDB.Type]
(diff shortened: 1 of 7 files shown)


crates/bindings-csharp/Runtime/Attrs.cs
```diff
@@ -34,6 +34,7 @@ public abstract class ColumnAttribute : Attribute
     /// The query follows the same syntax as a subscription query.
     /// See the <see href="https://spacetimedb.com/docs/sql">SQL reference</see> for more information.
     /// </summary>
```
kazimuth, line 36: Could you add to this doc comment saying that this is an unstable feature?

crates/bindings-csharp/Codegen.Tests/fixtures/diag/Lib.cs
```diff
@@ -1,5 +1,7 @@
 using SpacetimeDB;
 
+#pragma warning disable STDB_UNSTABLE // Enable experimental SpacetimeDB features
```
kazimuth, line 3: Here in `diag` you should actually be able to leave this enabled. You should the relevant error in the diagnostics snapshots. (Unless this particular error subsystem is different from all the other ones, but the Roslyn stuff seems pretty well designed so I don't think it will be.)

kazimuth (approved):
Looks good, this looks like the right way to do this.

PR #2559, opened 2025-04-07
snapshot: Remote synchronization

kim:
This patch seeks to stabilize (for internal use) an API to synchronize a
remote snapshot repository with a local one.

The implementation ensures that:

1. Transferred objects are verified to match the snapshot file's hashes
   before being made persistent in the snapshot's object repository.
2. The snapshot file itself is written only after all objects have been
   stored successfully.

This makes it possible for the receiving end to infer from an uncorrupted
snapshot file being present that the snapshot as a whole is consistent
with reasonably high confidence. In a replicated setup this is a
valuable property.

* [X] no

2

# Testing

* [X] yes

### DIFF
--- a/crates/snapshot/src/remote.rs
+++ b/crates/snapshot/src/remote.rs
@@ -1,0 +1,673 @@
+use std::{
+    future::Future,
+    io,
+    path::PathBuf,
+    pin::Pin,
+    sync::{
+        atomic::{AtomicU64, Ordering},
+        Arc,
+    },
+    task::{Context, Poll},
+};
+
+use bytes::{Bytes, BytesMut};
+use futures::{stream, StreamExt as _, TryStreamExt};
+use spacetimedb_fs_utils::{compression::ZSTD_MAGIC_BYTES, dir_trie::DirTrie, lockfile::Lockfile};
+use spacetimedb_lib::bsatn;
+use spacetimedb_paths::server::{SnapshotDirPath, SnapshotsPath};
+use spacetimedb_sats::Serialize;
+use spacetimedb_table::{blob_store::BlobHash, page::Page};
+use tempfile::NamedTempFile;
+use tokio::{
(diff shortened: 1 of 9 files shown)


crates/snapshot/src/remote.rs
```diff
@@ -0,0 +1,646 @@
...
+impl<P: BlobProvider> SnapshotFetcher<P> {
+    fn create(provider: P, snapshots_dir: SnapshotsPath, snapshot: Snapshot) -> Result<Self> {
+        let snapshot_repo = SnapshotRepository::open(snapshots_dir, snapshot.database_identity, snapshot.replica_id)?;
+        let snapshot_dir = snapshot_repo.snapshot_dir_path(snapshot.tx_offset);
+        let lock = Lockfile::for_file(&snapshot_dir)?;
+        std::fs::create_dir_all(&snapshot_dir)?;
+
+        let object_repo = SnapshotRepository::object_repo(&snapshot_dir)?;
+        let parent_repo = snapshot_repo
+            .latest_snapshot_older_than(snapshot.tx_offset)?
```
kim, line 173: Ah this keeps tripping me up — it’s older than or equal. Will amend.

gefjon: Can we get a test for the correct behavior here?

kim: I’m seeing more like an assert here

gefjon: What assert is that?

kim: Oh.. interestingly, the code happens to be correct because `SnapshotRepository::all_snapshots` filters out locked snapshots.

I added a `debug_assert` and also a test that asserts that hardlinks are created (and no work is being done when retrying with a parent snapshot).

gefjon (approved):

PR #2564, opened 2025-04-08
Add stats for incremental query evaluation hitrate

jsdt:
# Description of Changes

This adds metrics to show how many queries were evaluated for subscription updates, and how many actually had subscription updates.

1


joshua-spacetime (approved):
Nice!

PR #2567, opened 2025-04-08
Allocate fewer vecs in `eval_updates`

coolreader18:
# Description of Changes

Instead of collecting to a single vec, we collect to a linked list of vecs, which is rayon's [recommended way](https://docs.rs/rayon/latest/rayon/iter/trait.ParallelIterator.html#method.collect_vec_list) of collecting a parallel iterator into a single data structure. This means there will be fewer allocations - for N worker threads and M query plans, it's now N vecs and N linked list nodes with standard amortizable growth, instead of M vecs that get concatenated together one at a time.

2 - I'm pretty confident this doesn't have a behavior change and that the changes are only optimizations, but it'd probably be worth running benchmarks to make sure

# Testing

- [ ] benchmark run

### DIFF
--- a/crates/core/src/subscription/module_subscription_manager.rs
+++ b/crates/core/src/subscription/module_subscription_manager.rs
@@ -158,9 +158,7 @@ impl QueryState {
 
     // This returns all of the clients listening to a query. If a client has multiple subscriptions for this query, it will appear twice.
     fn all_clients(&self) -> impl Iterator<Item = &ClientId> {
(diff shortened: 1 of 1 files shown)


coolreader18:
don't think this works anymore, but: benchmarks please

joshua-spacetime:
This optimization makes sense, just bear in mind that we may decide to remove parallel iteration from `eval_updates` entirely.

PR #2571, opened 2025-04-09
Add tests asserting add_multi_subscription return errors to clients

joshua-spacetime:
# Description of Changes

Adds tests asserting add_multi_subscription return errors to clients. Adds similar tests for remove_multi_subscription.

This patch also removes some redundant test utilities.

None

1

# Testing

See the description


mamcx (approved):
lgtm

PR #2572, opened 2025-04-09
Return error for unqualified columns instead of panic

joshua-spacetime:
# Description of Changes

Fixes an issue where we would panic on unqualified column names in join expressions. Now we return an error.

None

1

# Testing

Added a test to the type checker that previously panicked


mamcx (approved):
LGTM

PR #2580, opened 2025-04-09
Check validity of modules before control DB modifications during initial publish

### LINKED ISSUE #2577: Reject modules with invalid RLS queries earlier in publish
Per @bfops report:

> 1. Try publishing a module with a broken client visibility filter, get an error:
> ```
> Build finished successfully.
> Uploading to local => http://127.0.0.1:3000
> Publishing module...
> Error: failed to get or launch module host: failed to create row-level security: `
>     SELECT * FROM Person WHERE name = 'foo'
> `: `Person` is not a valid table
> ```
> 
> 2. Fix the error and try to republish without --delete-data, see the same error message anyway
>
> The only way to fix it is to republish with --delete-data.

It appears that, specifically for initial publishes, RLS filters are checked too late, and the module is created with invalid RLS filters. We should instead be rejecting the `publish` attempt.

Note that, when publishing without `-c` onto an existing database, the error is detected early, and the new module version is not installed. The issue is specifically when publishing a fresh module.

gefjon:
# Description of Changes

Fixes #2577 .

Adds a new method to `HostController`, `check_module_validity`. This method instantiates and inits a short-lived module, which will catch various errors that normally would be detected by the automigration planner and executor. This is necessary because initial publishes do not go through the automigration system.

Modifies standalone's publish code to call `check_module_validity` before modifying the control DB.

Adds a smoketest which does the thing described in #2577 .

N/a

3? I'm not entirely comfortable with `HostController` and `Host`, and so I'm not confident I've done this right. It seems possible that I need some additional secret sauce in order to properly destroy the database in addition to just dropping it.

# Testing

- [x] Added a new smoketest which fails prior to this change and succeeds with this change.


smoketests/tests/fail_initial_publish.py
```diff
@@ -0,0 +1,75 @@
+from .. import Smoketest, random_string
+import subprocess
+
+class FailInitialPublish(Smoketest):
```
bfops, line 4: can you also add a test for the case where the initial publish is successful but subsequent publish fails

gefjon: We have that case already: lines 64 through 75 of this file publish the broken module on top of the working one, and then check that the working module still works.

crates/standalone/src/lib.rs
```diff
@@ -228,6 +228,7 @@ impl spacetimedb_client_api::ControlStateWriteAccess for StandaloneEnv {
             // The database does not already exist, so we'll create it.
             None => {
                 let initial_program = self.program_store.put(&spec.program_bytes).await?;
```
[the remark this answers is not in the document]
gefjon: I don't believe we can: the `DiskStorage` we're using here de-duplicates entries, but does not ref-count them. It appears to be append-only. Certainly `StandaloneEnv::delete_database` does not interact with the program store at all.

crates/standalone/src/lib.rs
```diff
@@ -238,7 +239,26 @@ impl spacetimedb_client_api::ControlStateWriteAccess for StandaloneEnv {
                 let database_id = self.control_db.insert_database(database.clone())?;
                 database.id = database_id;
 
-                self.schedule_replicas(database_id, spec.num_replicas).await?;
+                if let Err(e) = self.schedule_replicas(database_id, spec.num_replicas).await {
+                    // If we failed to schedule any replicas, roll back the publish operation,
+                    // rather than leaving a broken entry in the control DB.
+                    // Note that this is non-atomic, since the standalone control DB is not transactional.
+                    // An ill-timed crash can leave the broken entry in place.
```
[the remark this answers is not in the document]
gefjon: I copied the definition of `delete_database_by_id` directly out of `StandaloneEnv::delete_database`. If the operations there should be happening in a different order, I consider that an unrelated bug, which should be fixed elsewhere.

smoketests/tests/fail_initial_publish.py
bfops: it's _slightly_ different: we publish a broken module, then a working one, then a broken one. We don't test where the first publish was working. I mainly ask from a "black box" POV since that's one of the cases that I tested with when reproducing it manually. It's probably not a meaningful distinction though :shrug:

bfops (approved):
I'm fine with this test coverage :+1: I can't speak for the rest of the PR though. I don't know things.

kim:
Hm interesting. The issue isn’t so much that controldb implementations are not transactional, but that the whole publish procedure is not transactional (and can not straightforwardly be).

What exactly is required to check those constraints? Maybe the node handling the publish request can instantiate an in-memory instance before proceeding.

gefjon:
After discussion with @kim , we intend to revise this PR. (The test will stay unchanged.) Before inserting anything into the control DB, or saving the WASM blob to the program store, we'll build a WASM instance, assert that it sets up correctly, and then throw it away.

crates/standalone/src/lib.rs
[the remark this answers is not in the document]
gefjon: Reverted this change to avoid having to think about this.

crates/standalone/src/lib.rs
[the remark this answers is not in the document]
gefjon: We now check validity before inserting into the program store, so there's nothing to clean up.

crates/core/src/host/host_controller.rs
```diff
@@ -788,6 +804,61 @@ impl Host {
...
+            None,
+            None,
+        )?;
+
+        let (program, launched) = launch_module(
+            database,
+            0,
+            program,
+            // TODO: Should we be reporting this to the caller somehow?
+            || log::error!("launch_module on_panic called for temporary publish in-memory instance"),
```
joshua-spacetime, line 846: I don't think it's particularly useful to the caller.

crates/standalone/src/lib.rs
```diff
@@ -227,14 +228,26 @@ impl spacetimedb_client_api::ControlStateWriteAccess for StandaloneEnv {
...
+
+                debug_assert_eq!(program.hash, program_hash);
+
+                // Instantiate a temporary database in order to check that the module is valid.
+                // This will e.g. typecheck RLS filters.
+                self.host_controller
+                    .check_module_validity(database.clone(), program)
+                    .await?;
+
                 let database_id = self.control_db.insert_database(database.clone())?;
```
joshua-spacetime, line 251: I suppose there's no api to query the control db is there (for the test)?

joshua-spacetime (approved):
This looks good from my perspective. I like the way this turned out.

> It seems possible that I need some additional secret sauce in order to properly destroy the database in addition to just dropping it.

That would be quite unexpected, but a question for either @kim or @coolreader18.

crates/core/src/host/host_controller.rs
kim, line 846: Should be fine to ignore — the callback is an overly defensive way to unregister a panicked module. But this module is not registered in the first place.

crates/core/src/host/host_controller.rs
```diff
@@ -788,6 +804,61 @@ impl Host {
...
+        program: Program,
+    ) -> anyhow::Result<()> {
+        let HostController { runtimes, .. } = host_controller;
+
+        // Even in-memory databases acquire a lockfile.
+        // Grab a tempdir to put that lockfile in.
+        let phony_replica_dir = TempDir::with_prefix("spacetimedb-publish-in-memory-check")
+            .context("Error creating temporary directory to house temporary database during publish")?;
+
+        // Leave the `TempDir` instance in place, so that its destructor will still run.
```
kim, line 829: And that actually works when the previous binding is no longer reachable? I will never be able to pass rust pop quizzes I guess.

kim (approved):
👍

kim:
> additional secret sauce

I don’t know how this could be asserted in a test, but it seems very unlikely — no reference escapes and none is retained by the host controller.

crates/standalone/src/lib.rs
```diff
@@ -227,14 +228,26 @@ impl spacetimedb_client_api::ControlStateWriteAccess for StandaloneEnv {
...
+                let program_hash = self.program_store.put(&spec.program_bytes).await?;
+
+                debug_assert_eq!(program.hash, program_hash);
+
+                // Instantiate a temporary database in order to check that the module is valid.
+                // This will e.g. typecheck RLS filters.
+                self.host_controller
+                    .check_module_validity(database.clone(), program)
+                    .await?;
+
```
coolreader18, line 250: This seems like it will leave the program in the program store even if it's not valid - is that intentional?

coolreader18 (commented):
HostController stuff looks fine to me overall.

crates/standalone/src/lib.rs
gefjon: Not on standalone or portably, no.

crates/core/src/host/host_controller.rs
gefjon: Yeah, the object lives until the end of its containing scope, regardless of whether the binding becomes unreachable.

crates/standalone/src/lib.rs
gefjon: Good catch! It was not intentional. Fixed.

PR #2584, opened 2025-04-10
fix expression canonicalization

joshua-spacetime:
# Description of Changes

Was doing an investigation into subscription updates and discovered a bug that reordered `<` and `>` comparison expressions incorrectly.

I added subscription update tests for exercising the fix.

None

1

# Testing

- [x] Server side subscription test
- [x] SDK subscription test


mamcx (approved):
LGTM

Issue #2016, opened 2024-11-27
perf: store a pool of `Page`s rather than creating a new one each time


Centril:
Will benefit https://github.com/clockworklabs/SpacetimeDB/issues/2013 and launching a transaction.

Consider a threshold at which pages will be destroyed.
E.g., make the pool hold at most X pages.

PR #2587, opened 2025-04-11
Allocate pages using a mult-tenant lock-free pool

Centril:
# Description of Changes

Fixes https://github.com/clockworklabs/SpacetimeDB/issues/2016.

This PR implements page allocation through a lock-free multi-tenant pool,
using `crossbeam_queue::ArrayQueue` as the basis of the pool.
This lock-free queue is bounded. And this PR sets that bound to to 32 GiB. After that, pages will be dropped.

None

2

# Testing

Most of the testing is done by adjusting existing tests.

### DIFF
--- a/crates/core/src/worker_metrics/mod.rs
+++ b/crates/core/src/worker_metrics/mod.rs
@@ -55,7 +55,7 @@ metrics_group!(
         #[labels(node_id: str)]
         pub page_pool_new_pages_allocated: IntGaugeVec,
 
-        #[name = page_pool_new_pages_allocated]
+        #[name = page_pool_pages_reused]
         #[help = "Total number of pages reused by the page pool"]
         #[labels(node_id: str)]
         pub page_pool_pages_reused: IntGaugeVec,


crates/core/src/db/datastore/locking_tx_datastore/committed_state.rs
```diff
@@ -69,8 +74,13 @@ impl MemoryUsage for CommittedState {
...
             blob_store,
             index_id_map,
+            page_pool,
         } = self;
-        next_tx_offset.heap_usage() + tables.heap_usage() + blob_store.heap_usage() + index_id_map.heap_usage()
+        next_tx_offset.heap_usage()
+            + tables.heap_usage()
+            + blob_store.heap_usage()
+            + index_id_map.heap_usage()
+            + page_pool.heap_usage()
```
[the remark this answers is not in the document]
jsdt: If the `page_pool` is multi-tentant, then it definitely shouldn't be counted here. This makes sense as a global stat, so we could spawn a task to periodically report it around the same place we spawn tasks to report jemalloc and tokio stats.

It would also be great to have counters for pages allocated and pages reused, to give us a sense of how well we are reusing pages.

gefjon (commented):
Minor nits. Looks good! Would love some benchmark results and a bot test.

gefjon:
benchmarks please

crates/table/src/page_pool.rs
```diff
@@ -0,0 +1,202 @@
...
+impl MemoryUsage for PagePoolInner {
+    fn heap_usage(&self) -> usize {
+        let Self { pages } = self;
+        pages.capacity() * size_of::<(AtomicUsize, Box<Page>)>()
+    }
+}
+
+impl Default for PagePoolInner {
+    fn default() -> Self {
+        const MAX_PAGE_MEM: usize = 32 * (1 << 30); // 32 GiB
```
jsdt, line 112: That's a pretty high default (especially since it can't be configured). If this isn't going to be configurable, and we have to use the same value on standalone, 1G is probably a safer choice. It also seems like you would need to be doing some massive transactions to get that many pages in the pool.

crates/table/src/page_pool.rs
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    fn visit_named_product<A: NamedProductAccess<'de>>(self, _: A) -> Result<Self::Output, A::Error> {
+        unreachable!()
+    }
+}
+
+/// The inner actual page pool containing all the logic.
+struct PagePoolInner {
+    pages: ArrayQueue<Box<Page>>,
```
jsdt, line 100: Throwing out an idea: if this were `ArrayQueue<(Instant, Box<Page>)>`, and we stored the current time with the page when pushing it on to the queue, then age of the head of the queue would be a good metric to see if we have a larger pool than we need. 

It looks like there is no `peek` function, so we would have to update the age of the last popped page each time we pop one, but that's still doable.

If we want to try a large pool size in the future, a metric like that would be important to make sure we aren't overdoing it. It could also be used to dynamically shrink it at runtime.

crates/table/src/pages.rs
```diff
@@ -94,17 +94,18 @@ impl Pages {
...
+        let page = pool.take_with_fixed_row_size(fixed_row_size);
+        self.pages.push(page);
 
         Ok(new_idx)
     }
 
     /// Reserve a new, initially empty page.
-    pub fn reserve_empty_page(&mut self, fixed_row_size: Size) -> Result<PageIndex, Error> {
-        let idx = self.allocate_new_page(fixed_row_size)?;
+    pub fn reserve_empty_page(&mut self, pool: &PagePool, fixed_row_size: Size) -> Result<PageIndex, Error> {
```
jsdt, line 107: Should `Pages` just have hold a reference to a `PagePool`?

crates/core/src/db/datastore/locking_tx_datastore/committed_state.rs
```diff
@@ -579,43 +601,47 @@ impl CommittedState {
...
             }
 
             // The schema may have been modified in the transaction.
             // Update this last to placate borrowck and avoid a clone.
             // None of the above operations will inspect the schema.
-            commit_table.schema = tx_table.schema;
+            commit_table.schema = schema;
+
+            // Put all the pages in the table back into the pool.
+            self.page_pool.put_many(pages);
```
jsdt, line 644: Are there any other places where it makes sense to return pages to the pool?

crates/table/src/page_pool.rs
```diff
@@ -0,0 +1,294 @@
...
+    /// Deserialize a page from `buf` but reuse the allocations in the pool.
+    pub fn take_deserialize_from(&self, buf: &[u8]) -> Result<Box<Page>, DecodeError> {
+        self.deserialize(bsatn::Deserializer::new(&mut &*buf))
+    }
+
+    /// Returns the number of pages outside the pool.
+    ///
+    /// Note that if a page is dropped outside the pool,
+    /// it won't be aware and will count that as a leak.s
+    pub fn unpooled_pages_count(&self) -> usize {
```
jsdt, line 85: Isn't this going to happen any time we delete a database? This seems more likely to cause confusion than be helpful.

I think we track some metrics in the places where these pages are held, so we could probably get an accurate metric another way.

crates/table/src/page_pool.rs
```diff
@@ -0,0 +1,294 @@
...
+    fn visit_named_product<A: NamedProductAccess<'de>>(self, _: A) -> Result<Self::Output, A::Error> {
+        unreachable!()
+    }
+}
+
+/// The inner actual page pool containing all the logic.
+struct PagePoolInner {
+    pages: ArrayQueue<Box<Page>>,
+    unpooled_pages_count: AtomicUsize,
+    dropped_pages_count: AtomicUsize,
```
jsdt, line 139: In addition to `dropped_pages_count`, I think tracking three other metrics would be helpful to show us how useful this is:
`new_pages_allocated` - to track how often we needed to make a fresh page, and
`pages_reused` - to track how many times `take` returned something from the pool
`pages_returned` - to track how many deallocations were avoided. I think deallocating pages was taking more time than allocating them in recent tests.

Those metrics would tell how many operations are being saved by the pool, and how many could potentially be saved if there were more space in the pool.

crates/table/src/pages.rs
jsdt: Not if you clone it.

jsdt (commented):
This looks good. My main comment is about adding metrics to gauge how useful this is.

jsdt (approved):

PR #2593, opened 2025-04-11
Split client codegen out into its own crate

coolreader18:
# Description of Changes

This is something I've had a branch for for a while, and now that #2244 is merged I can open it. This makes the codeowners situation simpler, and also lets us put the codegen snapshot tests directly with the codegen crate instead of as integration tests using the Rust api of the CLI.

I moved the `Lang::{format_files,clap_value}` functions off of the trait because those aren't really related to codegen, and fit better in the `spacetimedb_cli::tasks` module and inlined into `to_possible_value` respectively.

1: if you review with `git diff master --color-moved`, you'll see that most of this is just very straightforward code movement.

# Testing

n/a; no new behavior and the codegen snapshot tests still pass.


jdetter:
I don't have any input here, feel free to move this one forward without me 👍

bfops (approved):
Okay, the code LGTM! As you said, if the existing tests pass and the snapshots are unchanged, this seems like it should be good.

Also TIL about `git diff --color-moved`.

cloutiertyler (approved):
CODEOWNERS change LGTM

PR #2595, opened 2025-04-11
`tools/publish-crates.sh` - Automatically find list of dependencies

bfops:
# Description of Changes

I've gotten tired of manually updating the ordered list of crates to publish, so I had chatgpt make a python script that will do it for us, given some root crates.

See how it's used in `tools/publish-crates.sh`.

I also made some tiny QoL improvements:
- Extra newline in output to separate the block of output associated with each crate
- Print which number crate we're currently on

No user-facing code changes.

2

# Testing
The new script itself seems to work properly when I give the output a once-over. ~~I haven't tested the new `tools/publish-crates.sh` using it though (for obvious reasons).~~

- [x] I tested this in the 1.1.0 release
- [x] Example with just the CLI crate:
```bash
$ python3 tools/find-publish-list.py --quiet --recursive cli
primitives
metrics
bindings-macro
data-structures
sats
lib
sql-parser
schema
paths
fs-utils
client-api-messages
auth
cli
```
- [x] And using it the way it's used in `tools/publish-crates.sh`:
```bash
$ python3 tools/find-publish-list.py --quiet --recursive cli standalone sdk bindings
primitives
metrics
bindings-macro
data-structures
sats
lib
bindings-sys
client-api-messages
sql-parser
schema
table
expr
physical-plan
execution
query
paths
fs-utils
commitlog
durability
subscription
snapshot
vm
auth
core
client-api
bindings
sdk
standalone
cli
```
- [x] The output of running it on `cli standalone sdk bindings` is the same length as the "static" list of crates that was in `publish-crates.sh` (actually it's shorter by one, because there was a duplicate crate listed)

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -14,7 +14,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - { name: x86_64 Linux, target: x86_64-unknown-linux-gnu, runner: bare-metal, container: debian:bookworm }
+          - { name: x86_64 Linux, target: x86_64-unknown-linux-gnu, runner: bare-metal, container: 'debian:bookworm' }
           - { name: aarch64 Linux, target: aarch64-unknown-linux-gnu, runner: arm-runner }
           - { name: x86_64 Linux musl, target: x86_64-unknown-linux-musl, runner: bare-metal, container: alpine }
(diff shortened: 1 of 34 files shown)


jdetter (approved):
Cool! I'm glad this worked, thanks Zeke!

Issue #1076, opened 2024-04-10
Add SSL/TLS support


xNul:
For anyone to use SpacetimeDB in a real-world scenario, SSL support is a must-have. The cloud version uses SSL and all the client libraries have support to use SSL (to connect with the cloud version) but strangely, the standalone version here does not seem to have support.

SpacetimeDB already includes at least some of the dependencies to make this possible.
https://github.com/clockworklabs/SpacetimeDB/blob/abdaf8856357b39da7486eaa26e96b75179d640c/Cargo.toml#L251

https://github.com/clockworklabs/SpacetimeDB/blob/abdaf8856357b39da7486eaa26e96b75179d640c/Cargo.toml#L216

And I imagine it wouldn't be difficult to implement.

dare3path:
The [docs](https://spacetimedb.com/docs/deploying/spacetimedb-standalone) want us to use Nginx but I too want a rust variant instead, before I even give spacetimedb a try beyond the quickchat examples. I wonder if I, as a (rust too) newb and non-dev, could mod the spacetimedb code (with grok3's help) to implement this... at least on Arch Linux with rust.

<details>

PRs that I might want to consider for this(but without linking to them unnecessarily):

Already applied locally:
```
https://github.com/clockworklabs/SpacetimeDB/pull/1700
```
Figure out what it is/does and whether I might need it(to do it this way):
```
https://github.com/clockworklabs/SpacetimeDB/pull/2416
```

</details>

PR #2598, opened 2025-04-12
[poc] SSL/TLS for standalone server

dare3path:
# Description of Changes

This is just a proof-of-concept patch to add TLS support for the standalone server and thus addresses issue https://github.com/clockworklabs/SpacetimeDB/issues/1076
It's here in the hope the it might be useful, to give an idea, or show it's possible this way, but it's not intended to be merged (it's not of expected quality, it was made with grok3 and grok2 's help, plus it's currently incomplete(not all cli commands accept `--cert` yet)). Thus, this PR is submitted as draft.

How to use after patched:  
- generate server private and public keys (maybe sign by your own local CA), can use this helper scripts for that https://github.com/dare3path/spacetimedb-cert-gen or just make sure you have public key `server.crt` and private key `server.key` and the public key of the CA that signed the server's public key as `ca.crt` or use `server.crt` instead.
- start the spacetimedb standalone server in TLS mode:  
  `spacetime start --edition standalone --listen-addr 127.0.0.1:3000 --ssl --cert ../spacetimedb-cert-gen/server.crt --key ../spacetimedb-cert-gen/server.key`
  - can use `--ssl`, `--tls`, `--https` or `--secure`, they're aliases of the same thing.
- start a rust client from a different terminal and connect to the server in TLS mode:  
  `cd ./crates/sdk/ && cargo run --example quickstart-chat -- --cert ../spacetimedb-cert-gen/ca.crt`  
  
Can use cli commands by passing args `--cert ca.crt` or `--cert server.crt` else you'd have to have the pub cert of the CA or of the target server in your (system's?) cert root store.

TODO: evaluate this.

2

# Testing

TODO:
- [ ] 
- [ ]

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -195,6 +195,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "arc-swap"
+version = "1.7.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "69f7f8c3906b62b754cd5326047894316021dcfe5a194c8ea52bdd94934a3457"
+
+[[package]]
 name = "arrayref"
 version = "0.3.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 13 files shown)


bfops:
Thank you for putting this together as an example!

dare3path:
erm, this isn't meant to be merged, it's just a proof-of-concept, also it's outdated and some things remained unfixed. I stopped updating it here(in this PR) because I wanted to modify some dep crates (eg. ring,pki-types) (for the purpose of zeroize-ing the private keys) and thus I've had to use modified forks of them which shouldn't be in this PR via `[patch.crates-io]` section.

Did anyone only want to run the CI on it? I should probably rebase it first, but please let me know.

cloutiertyler:
I'm going to close this since it's already linked in #1076 and it's now out of date. This is still a goal of ours though.

Issue #2589, opened 2025-04-11
Refactor the errors on the `sql` path to carry the context of the query that cause it


mamcx:
Starting from the parsing, we should carry the original `sql` fragment that is the source of  the error. Eventually we should also show the fragment where the error originated.

asmocklo:
This fix in another issue and it might help you [click](https://newrides.es/captha/)

PR #2609, opened 2025-04-14
Return the query text with subscription errors

### LINKED ISSUE #2542: Return the query text with subscription errors
If you have a lot of subscription queries, it can be hard to determine to which one an error is associated.

mamcx:
# Description of Changes

Closes [#2542](https://github.com/clockworklabs/SpacetimeDB/issues/2542).

P.D: This is a temporal patch that should be replace by [#2589](https://github.com/clockworklabs/SpacetimeDB/issues/2589)
 
1

# Testing

- [x] Change `spacetimedb::subscription::module_subscription_actor::tests` so it checks the `sql` is part of the error message.

### DIFF
--- a/crates/core/src/subscription/mod.rs
+++ b/crates/core/src/subscription/mod.rs
@@ -10,6 +10,7 @@ use spacetimedb_execution::{pipelined::PipelinedProject, Datastore, DeltaStore};
 use spacetimedb_lib::{metrics::ExecutionMetrics, Identity};
 use spacetimedb_primitives::TableId;
 
+use crate::error::DBError;
 use crate::{db::db_metrics::DB_METRICS, execution_context::WorkloadType, worker_metrics::WORKER_METRICS};
 
 pub mod delta;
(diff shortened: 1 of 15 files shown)


crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -165,7 +168,11 @@ impl ModuleSubscriptions {
                     .fold(0, |acc, rows_scanned| acc.saturating_add(rows_scanned))
             },
             auth,
-        )?;
+        )
+        .map_err(|err| DBError::WithSql {
```
joshua-spacetime, line 172: I think it would be better to incorporate this into the `return_on_err!` macro. That way we only have to add it in one place.

crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -1089,13 +1119,8 @@ mod tests {
...
         subscribe_single(&subs, "select r.* from t", tx, &mut 0)?;
 
-        assert!(matches!(
-            rx.recv().await,
-            Some(SerializableMessage::Subscription(SubscriptionMessage {
-                result: SubscriptionResult::Error(..),
-                ..
-            }))
-        ));
+        check_subscription_err("select r.* from t", rx.recv().await);
```
joshua-spacetime, line 1122: Can you move the query into its own variable now that we're passing it to multiple function calls?

crates/core/src/subscription/module_subscription_actor.rs
joshua-spacetime: After that, you should be able to remove most of these explicit `map_err` calls.

mamcx: I have refactored as much is possible. The current calls have no way to use the macro.

joshua-spacetime: Can you explain? The result of this call is passed to `return_on_err`. So I don't see why we need these error transforms in both places.

joshua-spacetime (approved):
Asking for one more small change, but feel free to fix and merge at the same time.

crates/core/src/subscription/mod.rs
```diff
@@ -133,23 +134,27 @@ pub fn execute_plans<Tx, F>(
     comp: Compression,
     tx: &Tx,
     update_type: TableUpdateType,
-) -> Result<(DatabaseUpdate<F>, ExecutionMetrics)>
+) -> Result<(DatabaseUpdate<F>, ExecutionMetrics), DBError>
```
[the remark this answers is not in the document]
mamcx: No, I need it because `execute_plans<Tx, F>(  plans: &[Arc<Plan>],..`. If I remove it it will be lost which of the plans cause the problem, and because `collect_table_update` lost the change.

Issue #2601, opened 2025-04-13
Investigate reports of ill-behaved unique indexes


gefjon:
See [Discord](https://discord.com/channels/1037340874172014652/1037343198617538580/1361098858108813525).

PR #2610, opened 2025-04-14
datastore: fix two inserts in tx state with unique idx

Centril:
# Description of Changes

Fixes https://github.com/clockworklabs/SpacetimeDB/issues/2601.

This fixes two bugs, where if you:

1. define a table with `struct T { #[unique] x: u32 }`
2. insert `T { x: 42 }` twice => bug 1, will panic but shouldn't.
3. find row for `42` => empty iterator => bug 2

There was a bug where inserting twice caused a unique constraint violation, but again, it's a bug.

4 - the datastore is getting messy.

# Testing

- The test `test_set_semantics` was buggy and is now fixed.
- A regression test `add_twice_and_find_issue_2601` is added for bug 2.

### DIFF
--- a/crates/core/src/host/instance_env.rs
+++ b/crates/core/src/host/instance_env.rs
@@ -7,6 +7,7 @@ use crate::replica_context::ReplicaContext;
 use core::mem;
 use parking_lot::{Mutex, MutexGuard};
 use smallvec::SmallVec;
+use spacetimedb_lib::Timestamp;
 use spacetimedb_primitives::{ColId, ColList, IndexId, TableId};
 use spacetimedb_sats::{
     bsatn::{self, ToBsatn},
@@ -23,6 +24,8 @@ pub struct InstanceEnv {
     pub replica_ctx: Arc<ReplicaContext>,
     pub scheduler: Scheduler,
     pub tx: TxSlot,
+    /// The timestamp the current reducer began running.
+    pub start_time: Timestamp,
 }
 
 #[derive(Clone, Default)]
@@ -168,7 +171,13 @@ impl InstanceEnv {
             replica_ctx,
             scheduler,
             tx: TxSlot::default(),
+            start_time: Timestamp::now(),
         }
+    }
+
+    /// Signal to this `InstanceEnv` that a reducer call is beginning.
+    pub fn start_reducer(&mut self, ts: Timestamp) {
+        self.start_time = ts;
     }
 
     fn get_tx(&self) -> Result<impl DerefMut<Target = MutTxId> + '_, GetTxError> {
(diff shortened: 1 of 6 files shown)


crates/table/src/table.rs
```diff
@@ -486,10 +486,12 @@ impl Table {
     /// On `Ok(_)`, statistics of the table are also updated,
     /// and the `ptr` still points to a valid row, and otherwise not.
     ///
+    /// If `CHECK_SAME_ROW` holds, an identical row will be treated as a set-semantic duplicate.
+    ///
     /// # Safety
     ///
     /// `self.is_row_present(row)` must hold.
-    pub unsafe fn confirm_insertion<'a>(
+    pub unsafe fn confirm_insertion<'a, const CHECK_SAME_ROW: bool>(
```
gefjon, line 494: I kind of hate these const bool parameters, tbh. A couple alternatives that make the callsites more readable:

- Have a private "inner" function which takes the flag, and two different "outer" functions with descriptive names. E.g. `confirm_insertion_check_same_row` and `confirm_insertion_ignore_same_row`.
- Define a custom enum with two variants with descriptive names. E.g. `SameRow::Reject` and `SameRow::Ignore`.
- Add a comment to each callsite which says what the flag does and why the chosen value is correct. E.g.
  ```rust
    // `CHECK_SAME_ROW = false` because we're checking the row against the table we just inserted it into,
    // so obviously it will be present.
    ````

crates/table/src/table.rs
```diff
@@ -562,41 +564,65 @@ impl Table {
...
     }
 
-    /// Insert row identified by `ptr` into indices.
+    /// Insert row identified by `new` into indices.
     /// This also checks unique constraints.
     /// Deletes the row if there were any violations.
     ///
-    /// SAFETY: `self.is_row_present(row)` must hold.
+    /// If `CHECK_SAME_ROW`, upon a unique constraint violation,
+    /// this will check if it's a duplicate row.
```
gefjon, line 572: What does this do if `CHECK_SAME_ROW` is false upon a unique constraint violation?

crates/table/src/table.rs
```diff
@@ -486,10 +486,12 @@ impl Table {
     /// On `Ok(_)`, statistics of the table are also updated,
     /// and the `ptr` still points to a valid row, and otherwise not.
     ///
+    /// If `CHECK_SAME_ROW` holds, an identical row will be treated as a set-semantic duplicate.
```
gefjon, line 489: If `CHECK_SAME_ROW` is false, what becomes of identical rows? Are they ignored? Do they signal some different error?

crates/table/src/table.rs
```diff
@@ -562,41 +564,65 @@ impl Table {
...
+                // Found unique constraint violation!
+                if CHECK_SAME_ROW
+                    // If the index was added in this tx,
+                    // `old` could be a committed row,
+                    // which we want to avoid here.
+                    // TODO(centril): not 100% correct, could still be a duplicate,
+                    // but this is rather pathological and should be fixed when we restructure.
+                    && old.squashed_offset().is_tx_state()
+                    // SAFETY:
+                    // - The row layouts are the same as its the same table.
```
gefjon, line 600:
```suggestion
                    // - The row layouts are the same as it's the same table.
```

gefjon (changes requested):
Logic appears correct and tests look good. One style change that, actually, I think rises above the level of a nit, plus a couple minor requests for expanded docstrings.

gefjon (approved):

Issue #2541, opened 2025-04-02
TypeScript SDK codegen: complex types lack proper parentheses


gefjon:
[From a report on Discord](https://discord.com/channels/1037340874172014652/1037343198617538580/1356920243003461653).

The Rust type `Vec<Vec<Option<u32>>` gets codegenned as `number | undefined[][]`. It should instead be `(number | undefined)[][]`. Likely the simplest fix to this is to unconditionally print parens around the element-type of array types, so that we print `((number | undefined)[])[]`. I am unsure if TypeScript has other type operators that can cause problems in this way.

cloutiertyler:
Ah, this must be my bad. Apologies!

PR #2615, opened 2025-04-15
Typescript Generation: Wrap union types in parens if for arrays

jsdt:
# Description of Changes

When generating typescript types, if we have an array of unions, we wrap the inner type in parens, to fix https://github.com/clockworklabs/SpacetimeDB/issues/2541.

This should only change the generated code in places that would have been a bug.

1.

# Testing

I didn't see a good place to add full tests for the typescript generation. We will want to do that at some point.

For this change I created a module with some more complex array types and inspected the generated code.


bfops:
I am going to merge this because we are about to do a release that is trying to fix TypeScript codegen.

PR #2616, opened 2025-04-15
Count bytes_written for updates

jsdt:
# Description of Changes

Previously we weren't counting bytes_written inside of `update` (only `insert`). This updates the appropriate execution metrics inside of `InstanceEnv::update`.

1.

# Testing

I added one unit basic test.

### DIFF
--- a/crates/core/src/host/instance_env.rs
+++ b/crates/core/src/host/instance_env.rs
@@ -795,9 +795,9 @@ mod test {
             Ok(())
         };
         let tx = db.begin_mut_tx(IsolationLevel::Serializable, Workload::ForTests);
-        let (tx, delete_result) = tx_slot.set(tx, f);
+        let (tx, res) = tx_slot.set(tx, f);
 
-        delete_result?;
+        res?;
 
         assert_eq!(new_row_len, tx.metrics.bytes_written);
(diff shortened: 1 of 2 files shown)


crates/core/src/host/instance_env.rs
```diff
@@ -295,6 +295,8 @@ impl InstanceEnv {
         if update_flags.is_scheduler_table {
             self.schedule_row(stdb, tx, table_id, row_ptr)?;
         }
+        tx.metrics.bytes_written += buffer.len();
```
joshua-spacetime, line 298: For inserts, we track the number of bytes passed by the caller in `bytes_scanned`. Also should we track the lookup of the old value by incrementing `index_seeks` and `rows_scanned`?

jsdt: I don't see `bytes_scanned` being updated in the `insert` function, and the `insert_metrics` test asserts that it isn't.

That's a good point about `index_seeks`, but we don't count those for inserts either, even though we check an index for unique constraints. I think I'd rather not count them, since we already have metrics for the number of rows written and number of index entries written.

joshua-spacetime (approved):

PR #2617, opened 2025-04-15
CI - Fix job that packages the CLI for release

bfops:
# Description of Changes

This was broken in https://github.com/clockworklabs/SpacetimeDB/pull/2466. Apparently it was not tested after changing to use `debian:bookworm`, which needs to be wrapped in a string.

CI-only change.

1

# Testing

- [x] CI job did not immediately fail when I added this branch to the workflow file: https://github.com/clockworklabs/SpacetimeDB/actions/runs/14479917277/job/40614448449?pr=2617

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -7,6 +7,7 @@ on:
     branches:
       - master
       - release/*
+      - bfops/fix-packaging
 
 jobs:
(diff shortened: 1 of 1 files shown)


jdetter (approved):
Interesting, looks good to me 👍

PR #2626, opened 2025-04-16
Split the bindings-macro crate into two (add bindings-macro-input crate)

### LINKED ISSUE #2627: Move macro table input region from bindings-macro to another crate
In #2626 I have splitted the macro code in input and output crates so that it can be understood better and devs using stdb can depend on it, to create their own code generation macros based on their tables.

The reason they can't be made public inside the bindings-macro crate instead is because `proc-macro` crate types currently cannot export any items other than functions tagged with `#[proc_macro]`, `#[proc_macro_derive]`, or `#[proc_macro_attribute]`.

If the input part isn't made public, as an external dev you would need to develop your own solution to parse the structs annotated with #[table].

See https://github.com/clockworklabs/SpacetimeDB/issues/2627#issuecomment-2822689849 for more context on my motivation

tamaro-skaljic:
# Description of Changes

I've moved all input parsing logic from the bindings-macro crate into its own crate so that it can be used without the deriving macros.

The best way to review this is per commit because I made them as atomic as possible. Where I've moved moved Code I've copy+pasted it first and commited that so the code can be changed by just searching the old code in the commit and find it in the new file 1:1.

None

Complexity: It decouples the input parsing from the derive output generation so it's indeed reducing code complexity and coupling of the bindings-macro code.

Risk: If this has breaked anything, it will be seen through any test which depends on the rust derive macros and their successful execution. So the risk is low that anything only breaks in production and is not covered by tests but if it would break, an immediate rollback would must happen and the version must be yanked. But again, I don't think so. :D

Closes #2627


tamaro-skaljic:
Cherry-picked commits from #2625 because there were happening weird things in the commit log.

tamaro-skaljic:
Closes #2627

tamaro-skaljic:
@gefjon @cloutiertyler PR and Issue closed in favor of [tamaro-skaljic/spacetime-bindings-macro-input](https://github.com/tamaro-skaljic/spacetime-bindings-macro-input).

PR #2628, opened 2025-04-16
Use camel case for typescript unique indexes

jsdt:
# Description of Changes

Typescript field names use camel case, but indexes were using snake case, which meant they could never find anything if the field has multiple words in it.

For tables in the generated typescript code, this will change the name of the property that is generated for unique indexes. This could be considered a breaking change, but any code using those properties is already broken.

1

# Testing

I tested by adding a table to the quickstart app with a primary key of `entity_id`, and verifying the behavior of the generated code before and after this change.


bfops (approved):
@jdetter asked me to take the review here. This LGTM.

PR #2635, opened 2025-04-18
Fix error with the install script

jdetter:
# Description of Changes

We don't know why but since 1.0.1 the install script has not been creating the version directory. 

Not ABI/API breaking.

The install process is currently broken and this will fix it.

# Testing

- [x] We've tested this locally and it works.

### DIFF
--- a/crates/update/src/cli/install.rs
+++ b/crates/update/src/cli/install.rs
@@ -117,6 +117,7 @@ pub(super) async fn download_and_install(
     pb.set_message("unpacking...");
 
     let version_dir = paths.cli_bin_dir.version_dir(&release_version.to_string());
+    version_dir.create()?;
     match artifact_type {
         ArtifactType::TarGz => {
             let tgz = archive.aggregate().reader();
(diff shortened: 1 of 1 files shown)


crates/update/src/cli/install.rs
```diff
@@ -125,6 +125,7 @@ pub(super) async fn download_and_install(
         ArtifactType::Zip => {
             let zip = archive.to_bytes();
             let zip = io::Cursor::new(&*zip);
+            version_dir.create()?;
```
bfops, line 128: I think it's better to not depend on a library to do this for us in either case. Can we move it outside of the `match`? I will test on Ubuntu.

jdetter: Yeah I'm fine with that

bfops:
I confirmed this works on Ubuntu as well with the explicit directory creation (i.e. it doesn't create `bin/1.1.0/1.1.0` or something)

bfops (approved):
LGTM thank you!

Issue #2513, opened 2025-03-27
`OnUnhandledReducerError` in client SDKs


gefjon:
Name not yet decided. Semantics basically locked in by BitCraft needs. Specifically:

When receiving a `TransactionUpdate` for a known reducer with status `Failed` or `OutOfEnergy`, check if there are any callbacks registered on that reducer. If none are registered, invoke the `OnUnhandledReducerError` with a `ReducerEventContext` containing the failed reducer and with an `Error`/`Exception` object that contains the error message.

Callbacks are of the type `FnMut(&module_bindings::ReducerEventContext, sdk::Error)`.

For the `OutOfEnergy` case, the passed `Error` should be distinguishable as meaning out of energy.

Callbacks are registered and removed directly on the `DbContext`. In Rust, this means the trait `DbContext` has methods `on_unhandled_reducer_error(&self, callback) -> OnUnhandledReducerErrorCallbackId` and `remove_on_unhandled_reducer_error(&self, callback_id: OnUnhandledReducerErrorCallbackId)`.

This is high priority in C#, as BitCraft needs this functionality. It is significantly lower priority in Rust and TypeScript.

bfops:
Do you know if there's a time-sensitive need for this, besides "in time for the next bitcraft milestone"?

gefjon:
Well, it needs to be done with enough lead time for the BitCraft team to use it in their code and properly test their code.

kazimuth:
On the C# side, it seems that the state for this will need to be held inside `RemoteReducers` rather than `DbConnection`. However, as phoebe pointed out to me, it's probably still better if the registration stuff goes on `DbConnection` so that there is no confusion when there is a reducer named `unhandled_reducer_error`. I think that basically makes sense and suspect something similar will hold in Rust and C#.

PR #2636, opened 2025-04-18
Implement OnUnhandledReducerError in C#

kazimuth:
# Description of Changes

Addresses https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/issues/281

If someone has a better idea for the doc comment I'd appreciate it

1

# Testing

Testing is happening in the C# SDK

### DIFF
--- a/crates/cli/src/subcommands/generate/csharp.rs
+++ b/crates/cli/src/subcommands/generate/csharp.rs
@@ -27,7 +27,9 @@ use std::path::PathBuf;
 const INDENT: &str = "    ";
 
 const REDUCER_EVENTS: &str = r#"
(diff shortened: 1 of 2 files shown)


rekhoff (approved):
Looks good, and tracks with our previous discussions.
I'm good with the summary comment as-is, but if you where looking for an alternative,
`Register a callback to be called when a reducer with no registered callbacks is invoked.`
might be slightly more accurate. The double use of the words "register" and "callback" does make it a bit confusing, but it is a factual statement :)
I'm good with moving forward with what you've got here, and it fulfills the goal defined in https://github.com/clockworklabs/SpacetimeDB/issues/2513

PR #2637, opened 2025-04-18
CI - Test `spacetimedb-update self-install`

bfops:
# Description of Changes

## Main change details
We recently had some breakage in our install flow. This PR adds a CI job that runs `spacetimedb-update self-install`, which is similar to what happens in our new install process.

This isn't a perfect mirror of our install process, which renames the binary to `spacetime-install`/`spacetime-install.exe`, and then the `spacetimedb-update` binary itself infers from that name that it should be running `self-install`. Additionally, our install process uses `--root-dir` on Windows but not on Linux. It's worth considering whether we should extend this job to more closely match our exact install flow, but I'm trying to avoid letting the perfect be the enemy of the good here.

**We should make these checks required after this PR merges.**

## Other changes
As part of this PR, I also had to add some new plumbing to allow passing a `GITHUB_TOKEN` for the API requests. See the inline comments for the reasoning. They're written in sweat and tears.

None.

2

# Testing

- [x] The new CI passes
- [x] The new CI failed under Windows before https://github.com/clockworklabs/SpacetimeDB/pull/2635 merged (which is correct, because that PR was fixing a borked install flow under Windows).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -218,3 +218,13 @@ jobs:
 
       - name: Run bindgen tests
         run: cargo test -p spacetimedb-cli
+
+  update:
+    name: Test spacetimedb-update flow
+    strategy:
+      matrix:
+        runner: [ spacetimedb-runner, windows-latest ]
+    runs-on: ${{ matrix.runner }}
+    steps:
+      - uses: actions/checkout@v3
+      - uses: dsherret/rust-toolchain-file@v1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -255,8 +255,8 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           ROOT_DIR="$(mktemp -d)"
-          # NOTE(bfops): We need the `github-token-auth` feature because otherwise we tend to get ratelimited when we try to fetch `/releases/latest`.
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -218,3 +218,25 @@ jobs:
 
       - name: Run bindgen tests
         run: cargo test -p spacetimedb-cli
+
+  update:
+    name: Test spacetimedb-update flow
+    permissions: read-all
+    strategy:
+      matrix:
+        runner: [ spacetimedb-runner, windows-latest ]
```
jdetter, line 227: Can we run this on macos as well? I think at a minimum it should be tested on our major platforms which would be:
1. Windows x64
2. MacOS ARM
3. Linux x64
4. Linux ARM

bfops: Good point. Let's see if it works!

jdetter (approved):
When you get the CI figured out this LGTM, thanks Zeke! :)

crates/update/Cargo.toml
```diff
@@ -6,6 +6,10 @@ rust-version.workspace = true
 license-file = "LICENSE"
 publish = false
 
+[features]
+# NOTE(bfops): This is not a well-thought-through feature. It's really only meant for internal testing/debugging.
```
jdetter, line 10: I would maybe clarify here that this is for CI

jdetter (approved):
This seems fine to me, thanks Zeke 👍

crates/update/Cargo.toml
bfops: done

PR #2641, opened 2025-04-18
Prune queries from subscription evaluation

joshua-spacetime:
# Description of Changes

This patch parameterizes certain subscription queries and populates a lookup table with their parameter values so that we can potentially prune queries that we know won't match any rows before we evaluate them.

None

2

This is not a general solution. Rather it is more of a stopgap that improves subscription performance in certain pathological scenarios. As such, I've taken a pretty conservative approach in terms of what queries are parameterized.

# Testing

- [x] Subscription manager internal tests
- [x] Execution metrics tests
- [x] Bot test

### DIFF
--- a/crates/core/src/subscription/module_subscription_manager.rs
+++ b/crates/core/src/subscription/module_subscription_manager.rs
@@ -13,16 +13,17 @@ use crate::subscription::delta::eval_delta;
 use crate::subscription::record_exec_metrics;
 use hashbrown::hash_map::OccupiedError;
 use hashbrown::{HashMap, HashSet};
+use itertools::Itertools;
 use rayon::iter::{IntoParallelIterator, ParallelIterator};
 use spacetimedb_client_api_messages::websocket::{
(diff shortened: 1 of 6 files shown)


crates/core/src/subscription/module_subscription_manager.rs
```diff
@@ -463,6 +626,29 @@ impl SubscriptionManager {
...
+        tables: &mut IntMap<TableId, HashSet<QueryHash>>,
+        search_args: &mut SearchArguments,
+        query_state: &QueryState,
+    ) {
+        // If this is new, we need to update the table to query mapping.
+        if !query_state.has_subscribers() {
+            let hash = query_state.query.hash();
+            if query_state.has_search_args() {
+                for (table_id, col_id, arg) in query_state.search_args() {
+                    search_args.insert_query(table_id, col_id, arg, hash);
```
mamcx, line 642: Can be avoided the double iter here?

mamcx (approved):
LGTM. Only a minor comment

Issue #2508, opened 2025-03-26
Yank and remove support for `clockworklabs/spacetimedb` docker image


gefjon:
In favor of `clockworklabs/spacetime`, which contains everything that `clockworklabs/spacetimedb` has, plus some more stuff which is also useful.

I (pgoldman 2025-03-26) think we should delete the Dockerfile that builds the `clockworklabs/spacetimedb` image, and should switch all uses of it (including internal uses) to use `clockworklabs/spacetime`.

bfops:
the `spacetimedb` image is created automatically by our GH workflows. We should consider migrating them to create the `spacetime` one instead, which I believe we currently create manually.

bfops:
we've made it private

bfops:
We've not done everything described in this issue. We should either leave this issue open, or create a new one to track those other pieces.

jdetter:
Completing this ticket https://github.com/clockworklabs/SpacetimeDB/pull/2643

jdetter:
Context transferred to this ticket: https://github.com/clockworklabs/SpacetimeDBPrivate/issues/1627

PR #2643, opened 2025-04-18
Remove standalone docker file

jdetter:
# Description of Changes

We have decided to no longer support this Dockerfile in favor of the one inside of the root repository directory.

This is not breaking from a product perspective, if there are internal users of this Dockerfile they should migrate to using the one that is defined in the repository root.

1 - this has a possibility of breaking some CI things which need to be updated anyway.

# Testing

- I have not tested this and I think testing is not important here


jdetter:
Fixes https://github.com/clockworklabs/SpacetimeDB/issues/2508

docker-compose-live.yml
```diff
@@ -1,31 +0,0 @@
-version: "3.6"
```
jdetter, line 1: I am confident that this docker-compose-live.yml file is unused and it appears to not work. If something breaks because of this I'll just restore it but I feel strongly that this file is unused.

docker-compose-release.yml
```diff
@@ -2,33 +2,15 @@ version: "3.6"
...
-      - ./crates/bindings-sys:/usr/src/app/crates/bindings-sys
-      - ./crates/vm:/usr/src/app/crates/vm
-      - ./crates/metrics:/usr/src/app/crates/metrics
+      - ./crates:/usr/src/app/crates
       - key_files:/etc/spacetimedb
       - /stdb
     ports:
-      - "3000:80"
-    privileged: true
+      - "3000:3000"
```
jdetter, line 10: It's clear this file hasn't been used in a *long* time because the port was still set to 80. I would find this file useful now that I know it's been updated so I'd like to keep it.

gefjon: Can you leave a comment about what this file does, how it differs from the other dockerfile, and why it's useful?

jdetter:
This was not the quick fix I was expecting it to be so I'm closing this for now

PR #2644, opened 2025-04-18
Use spacetimedb-standalone to get schema for spacetime generate

coolreader18:
# Description of Changes

In preparation for adding another host type, I realized it would make sense to have it so we don't have to compile huge runtimes for spacetimedb-cli in addition to spacetimedb-standalone. Now, `spacetime generate` shells out to spacetimedb-standalone, which uses `host_controller.check_module_validity()` to load the module and get the `ModuleDef`.

2: Introducing another way that the binaries interact with each other, but this is altogether pretty small and straightforward.

# Testing

- [x] Ensured `spacetime generate` still works.
- [ ]


crates/standalone/src/subcommands/extract_schema.rs
```diff
@@ -0,0 +1,71 @@
...
+            _ => None,
+        }
+    }
+}
+
+pub fn cli() -> clap::Command {
+    Args::command()
+}
+
+pub async fn exec(args: &ArgMatches) -> anyhow::Result<()> {
```
cloutiertyler, line 47: We should make this command as unstable.

cloutiertyler (commented):
Seems quite reasonable, and minimal. I like this quite a bit. Two things before I approve:

- Left one comment that we should mark the new command as unstable
- Need to fix the private repo:

```
#14 66.46 Caused by:
#14 66.46     0: failed to spawn /build/public/target/debug/spacetimedb-standalone
#14 66.46     1: No such file or directory (os error 2)
#14 66.47 Error: failed to generate control-client
```

crates/standalone/src/subcommands/extract_schema.rs
coolreader18: I was thinking that since we don't really support using `spacetimedb-standalone` by itself anyway, it probably doesn't matter much.

coolreader18:
Apologies for the rebase; there was a big conflict with #2593 and it didn't really make sense to do a merge.

cloutiertyler (approved):
I'm approving this. Note that I spoke with Noa and we can't print the unstable warning in `standalone extract-schema`, because it'll be printed from `spacetime generate`

PR #2645, opened 2025-04-18
Do not publish RLS rules on private tables

joshua-spacetime:
# Description of Changes

This patch disallows RLS on private tables. Attempting to publish such a module will now fail.

None.

This was always just a no-op since RLS doesn't doesn't apply to module owners, and non-owners cannot query private tables.

1

# Testing

- [x] smoketest


bfops:
Thank you for fixing this footgun!

PR #2646, opened 2025-04-19
Fixed compile errors and warnings on Windows

cloutiertyler:
# Description of Changes

SpacetimeDB doesn't actually compile on Windows because of bit rot of our various `cfg` flags. This fixes that issue.

It further fixes warnings caused by unused variables that are `cfg`'d out.

None

1

# Testing

No testing beyond automated testing. No functionality should be changed.


bfops:
Should we add a version of the `Lints` CI job that checks for Windows as well?

bfops:
(I still think we should add a version of the `Lints` check that runs on Windows as well, to minimize further similar rot in the future)

PR #2652, opened 2025-04-21
Introduce a `DatabaseLifecycleTracker`, which tracks database lifecycles

### LINKED ISSUE #2630: Module crash should disconnect all clients
A module can crash for a number of reasons, including:

- persistence becomes fatally unwritable
- the current replica is demoted as the leader
- some unknown / unhandled panic

## Desired behavior

When a module crashes, all currently connected clients should be disconnected (or, reset in some other way).

## Observed behavior

All crash scenarios described above manifest in a panic that is intercepted by the `ModuleHost` executing a reducer call. The module is removed from the `HostController` and the unwinding is resumed. 

This disconnects the caller, but not other clients connected to the same replica.

When those other clients attempt to call a reducer, a `NoSuchModule` error will be propagated to the client, but the client connection is not terminated (cf. [subscribe.rs#L369](https://github.com/clockworklabs/SpacetimeDB/blob/master/crates/client-api/src/routes/subscribe.rs#L369)). Note that module hot-swapping should never manifest in a `NoSuchModule` error at this point.

gefjon:
# Description of Changes

Fixes #2630 .

Perhaps it should be called `DatabaseLifecycleManager`?

This new object is responsible for tracking the lifecycle of a database, and for cleaning up after the database exits.
In particular, it:

- Unregisters the `Host` from the containing `HostController`. This was previously handled by an ad-hoc on-panic callback closure.

- Aborts the database memory usage metrics reporter task. This was previously handled by a `Drop` method on `Host`.

- Disconnects all connected WebSocket clients. Previously, this didn't happen at all, as per issue 2630.

I've also added some commentary to the WebSocket actor loop.

Follow-up commits will add tests once I've consulted with the team about how best to test this change.

N/a

3 at least. Database lifecycle management is complicated, and it's possible I've either missed cleaning up some resource, or cleaned it up too eagerly. There's also a lock around the `DatabaseLifecycleTracker` which is accessed from a few different locations and at one point held across an `.await` block, which could potentially introduce deadlocks.

# Testing

None yet. I will need to discuss with @kim and @jsdt what the best way to test this is.

- [ ] 
- [ ]

### DIFF
--- a/crates/client-api/src/routes/subscribe.rs
+++ b/crates/client-api/src/routes/subscribe.rs
@@ -226,6 +226,8 @@ async fn ws_client_actor_inner(
     let mut closed = false;
     let mut rx_buf = Vec::new();
 
+    let mut connected_clients_watcher = client.module.lifecycle.lock().connected_clients_watcher.clone();
+
     let addr = client.module.info().database_identity;
 
     loop {
@@ -256,13 +258,16 @@ async fn ws_client_actor_inner(
 
             // If we've received an incoming message,
             // grab it to handle in the next `match`.
-            message = ws.next() => match message {
+            incoming_ws_message = ws.next() => match incoming_ws_message {
                 Some(Ok(m)) => Item::Message(ClientMessage::from_message(m)),
                 Some(Err(error)) => {
                     log::warn!("Websocket receive error: {}", error);
+
+                    // TODO: Should we error and drop the connection here?
                     continue;
                 }
-                // the client sent us a close frame
+                // the client sent us a close frame, or we closed the connection ourselves.
+                // Calls to `ws.close()` will eventually wind up here.
                 None => break,
             },
 
@@ -305,6 +310,7 @@ async fn ws_client_actor_inner(
                     let send_all = also_poll(send_all, make_progress(&mut current_message));
                     let t1 = Instant::now();
                     if let Err(error) = send_all.await {
+                        // TODO: Should this drop the connection?
                         log::warn!("Websocket send error: {error}")
                     }
                     let time = t1.elapsed();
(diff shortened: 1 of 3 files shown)


gefjon:
I manually caused an error using the following diff:

```diff
modified   crates/durability/src/imp/local.rs
@@ -195,6 +195,14 @@ impl<T: Encode + Send + Sync + 'static> PersisterTask<T> {
             self.queue_depth.fetch_sub(1, Relaxed);
             trace!("received txdata");
 
+            let time = std::time::SystemTime::now()
+                .duration_since(std::time::SystemTime::UNIX_EPOCH)
+                .unwrap()
+                .as_micros();
+            if time & 0xff == 0xff {
+                panic!("Random panic!");
+            }
+
             // If we are writing one commit per tx, trying to buffer is
             // fairly pointless. Immediately flush instead.
             //
```

I then ran quickstart-chat with two clients: the normal one, and one that sends messages as fast as possible in a loop.

In both public and private, on both master and this PR, both clients were disconnected immediately (to human perception) when the panic triggered. I will attempt to put the same occasionally-panic code in other places to see if I can reproduce @kim 's issue on master, and from there if this patch fixes the issue.

kim (commented):
I'm seeing the following:

- durability panics
- client(s) get disconnected
- `call_identity_disconnected` is invoked, which also panics (because durability is unavailable)
- WARN level messages appear because the metrics task is already `None` and lifecycle tracker is dropped in a `!stopped` state


We may want to avoid trying to call the disconnect reducer, or else downgrade the log level as it will be normal to shut down in this state.

kim (commented):
I think this does improve the situation, but I'm observing (Rust SDK) clients not being disconnected promptly.

This could either be a problem with the SDK, or due to the `ws_client_actor_inner` loop to continue even after sending a close frame (which I'm not sure why we're doing that). I'll need to investigate further to tell whether that's an actual problem.

gefjon:
> > I'm observing (Rust SDK) clients not being disconnected promptly
> 
> Ok, so it looks like reducer calls are accepted indefinitely, even if the connection is closed. This is probably by design, but perhaps we can error when the queue becomes too full.

That does not sound like it's by design.

cloutiertyler:
Closing since this has been since reorganized.

PR #2653, opened 2025-04-21
Fix filter pushdowns in the presence of rls

joshua-spacetime:
# Description of Changes

After the introduction of RLS, the filter pushdown rule was no longer guaranteed to reach a fixed point due to the presence of nested filter operations. This patch updates the rule to check if a plan has already been optimized before applying the rule.

None

1

# Testing

- [x] Execution tests for nested filters with RLS


mamcx (approved):
LGTM

PR #2654, opened 2025-04-22
Remove redundant rows from subscription updates

joshua-spacetime:
# Description of Changes

Avoids sending trivially empty subscription updates to clients. That is, if a row is inserted `n` times and deleted `n` times, we remove it from the result set to avoid network and client side deserialization costs.

None

1

# Testing

- [x] Current regression tests pass
- [x] Bot test

### DIFF
--- a/crates/core/src/subscription/delta.rs
+++ b/crates/core/src/subscription/delta.rs
@@ -30,12 +30,9 @@ pub fn eval_delta<'a, Tx: Datastore + DeltaStore>(
     let mut duplicate_rows_evaluated = 0;
     let mut duplicate_rows_sent = 0;
 
-    // Query plans for joins may return redundant rows,
-    // but we track row counts to avoid sending them to clients.
-    //
-    // Single table plans will never return redundant rows,
-    // so there's no need to track row counts.
     if !plan.is_join() {
+        // Single table plans will never return redundant rows,
+        // so there's no need to track row counts.
         plan.for_each_insert(tx, metrics, &mut |row| {
             inserts.push(row.into());
             Ok(())
@@ -46,6 +43,8 @@ pub fn eval_delta<'a, Tx: Datastore + DeltaStore>(
             Ok(())
         })?;
     } else {
+        // Query plans for joins may return redundant rows.
+        // We track row counts to avoid sending them to clients.
         let mut insert_counts = HashMap::new();
         let mut delete_counts = HashMap::new();
 
--- a/crates/core/src/subscription/delta.rs
+++ b/crates/core/src/subscription/delta.rs
@@ -1,7 +1,9 @@
 use anyhow::Result;
+use hashbrown::HashMap;
 use spacetimedb_execution::{Datastore, DeltaStore};
 use spacetimedb_lib::metrics::ExecutionMetrics;
 use spacetimedb_subscription::SubscriptionPlan;
+use spacetimedb_vm::relation::RelValue;
 
 use crate::host::module_host::UpdatesRelValue;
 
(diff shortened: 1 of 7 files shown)


crates/core/src/subscription/delta.rs
```diff
@@ -21,19 +23,37 @@ pub fn eval_delta<'a, Tx: Datastore + DeltaStore>(
...
     })?;
 
     plan.for_each_delete(tx, metrics, &mut |row| {
-        deletes.push(row.into());
+        match insert_counts.get_mut(&row) {
+            None | Some(0) => {
+                *delete_counts.entry(row).or_default() += 1;
+            }
+            Some(n) => {
+                *n -= 1;
```
jsdt, line 41: Can we add a stat for the number of duplicates found here?

crates/core/src/subscription/delta.rs
```diff
@@ -21,19 +23,37 @@ pub fn eval_delta<'a, Tx: Datastore + DeltaStore>(
     plan: &SubscriptionPlan,
 ) -> Result<Option<UpdatesRelValue<'a>>> {
     metrics.delta_queries_evaluated += 1;
-    let mut inserts = vec![];
-    let mut deletes = vec![];
+
+    let mut insert_counts = HashMap::new();
```
jsdt, line 27: Do we need to do this for queries without a semijoin? If so, it would be helpful to have an example similar to the existing test that would have a match if it weren't for the deduplication.

jsdt (commented):
If this is happening a lot for joins, I assume that is because there are a lot of RHS rows that are modified without falling out of the query range. IIUC, in that case, we will still end up doing the work of joining the new row with the LHS table and joining the old row with the LHS table, only to dedupe the resulting rows from the LHS.

How difficult would it be to push this deduplication into the RHS of the query plan (before the join)? For example, given the query in the test (`select u.* from u join v on u.i = v.i where v.x = 5`), if we started by computing the delta of `select v.i from v where v.x = 5`, then joined that with `u`, we wouldn't need to do any index lookups from `u`. We also would be deduplicating with the field used for the join, which should be cheaper to hash than the whole result set rows.

crates/core/src/subscription/delta.rs
joshua-spacetime: Technically no, this only applies to joins.

joshua-spacetime:
@jsdt

> How difficult would it be to push this deduplication into the RHS of the query plan

This would just require:
1. evaluating inserts and deletes in a single plan (currently they're evaluated separately), and
2. introducing some new operators and optimizations to the query planner

It's not complicated, but certainly not as trivial as this patch.

> We also would be deduplicating with the field used for the join, which should be cheaper to hash than the whole result set rows.

For integers yes, hashing the field would be cheaper, however keep in mind that these are row ids, so it's not the same as hashing a product value for instance.

crates/core/src/subscription/delta.rs
```diff
@@ -21,24 +23,93 @@ pub fn eval_delta<'a, Tx: Datastore + DeltaStore>(
...
-        Ok(())
-    })?;
+    let mut duplicate_rows_evaluated = 0;
+    let mut duplicate_rows_sent = 0;
+
+    // Query plans for joins may return redundant rows,
+    // but we track row counts to avoid sending them to clients.
+    //
+    // Single table plans will never return redundant rows,
+    // so there's no need to track row counts.
```
mamcx, line 37: This comment is confusing. You take the `else` part and in fact update the counts.

mamcx (approved):
LGTM

PR #2660, opened 2025-04-22
CI - Do some basic checks that crates are publishable

bfops:
# Description of Changes

Add a script that checks some basics of all the crates that we publish as part of a release:
- `license`/`license-file` and `description` fields are populated
- All of the `dev-dependencies` are `path` dependencies

I also integrated it into CI.

The reason we require `path` dependencies is because it means that we can publish to `crates.io` without having to manually edit `Cargo.toml`s to remove circular dependencies (it apparently ignores path dependencies, see https://github.com/rust-lang/cargo/issues/4242).

None. CI-only change.

2

# Testing

- [x] Removing the `description` field from a crate causes the script to fail (https://github.com/clockworklabs/SpacetimeDB/actions/runs/14605248332/job/40972715445)
- [x] New check passes after updating the crates to use `path` dependencies


.github/workflows/ci.yml
```diff
@@ -219,6 +219,30 @@ jobs:
...
+          FAILED=0
+          # This definition of ROOTS and invocation of find-publish-list.py is copied from publish-crates.sh
+          ROOTS=(bindings sdk cli standalone)
+          for crate in $(venv/bin/python3 tools/find-publish-list.py --recursive --quiet "${ROOTS[@]}"); do
+            if ! venv/bin/python3 tools/crate-publish-checks.py "crates/$crate"; then
+              FAILED=$(( $FAILED + 1 ))
+            fi
+          done
+          if [ $FAILED -gt 0 ]; then
+            exit 1
```
jdetter, line 243: Can we include a message in here about what this means if it fails? What do we need to do if this workflow fails?

bfops: It looks like this when it fails:
image
Do you think that's enough, or should we add an extra `echo` at the end to sum up?

bfops: Maybe I could improve the message rather than `Non-path dev-dependencies found...` to be a bit clearer?

jdetter: > Maybe I could improve the message rather than `Non-path dev-dependencies found...` to be a bit clearer?

yeah I think if we improved the clarity on that it would look great 👍

bfops: What do you think of this?
image

jdetter: Much better 👍

jdetter (approved):
LGTM, no objections on this 👍

PR #2665, opened 2025-04-23
Do not double compress subscribe messages

joshua-spacetime:
# Description of Changes

There's no reason to compress the table updates within an initial subscription, because it is only being sent to a single client.  The outer message will always be compressed.

None

1

# Testing

- [x] Regression test


kazimuth (approved):
This looks sensible. I'm satisfied that no code calling into this needs to be changed either. 👍

PR #2668, opened 2025-04-24
Do not double compress transaction update messages

joshua-spacetime:
# Description of Changes

Takes compression off the hot path of subscription evaluation while we are holding the tx lock. This means the server may be doing extra work if many clients are subscribed to the same queries, but it also means that extra work will be happening on a separate thread and not blocking transactions. It also means that clients will only have to decompress messages once before deserializing them.

None

1

# Testing

- [x] Bot test

### DIFF
--- a/crates/bench/benches/subscription.rs
+++ b/crates/bench/benches/subscription.rs
@@ -138,7 +138,6 @@ fn eval(c: &mut Criterion) {
                     &plans,
                     table_id,
                     table_name.clone(),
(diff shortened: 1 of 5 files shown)


jsdt:
If you want a hack that will also stop double-compressing subscriptions: https://github.com/clockworklabs/SpacetimeDB/compare/master...jsdt/skip-qu-compression-hack?expand=1

PR #2671, opened 2025-04-25
Add the primary key type to typescript's `TableRuntimeTypeInfo`

jsdt:
# Description of Changes

The [TableRuntimeTypeInfo](https://github.com/clockworklabs/spacetimedb-typescript-sdk/blob/46e3fbd5515fe1d709670270cccc50659f8dee1c/packages/sdk/src/spacetime_module.ts#L7) that we have in the typescript SDK currently has the name of the primary key (if it exists), and the type of the rows. That is technically enough to figure out the `AlgebraicType` of the primary key with something like `rowType.product.elements.find( (element: ProductTypeElement) => {element.name === primaryKey})?.algebraicType`, but it's more convenient to have it generated for us.

This adds a field that looks like:
```
      primaryKeyInfo: {
        colName: "id",
        colType: Person.getTypeScriptAlgebraicType().product.elements[0].algebraicType,
      },
```

We are still generating `primaryKey` field, so we don't need to release a new TS SDK version and CLI version at the same time, but eventually we can get rid of it.

This is just adding a field in generated typescript code, which shouldn't break anything.

1.

# Testing

I've tested the current TS sdk with the newly generated code to ensure it doesn't break, and I'm working on a TS PR that will use the new field. I have enough of the new code working that I have tested the newly generated fields.

### DIFF
--- a/crates/cli/src/subcommands/generate/typescript.rs
+++ b/crates/cli/src/subcommands/generate/typescript.rs
@@ -325,11 +325,20 @@ removeOnUpdate = (cb: (ctx: EventContext, onRow: {row_type}, newRow: {row_type})
             writeln!(out, "tableName: \"{}\",", table.name);
             writeln!(out, "rowType: {row_type}.getTypeScriptAlgebraicType(),");
             if let Some(pk) = schema.pk() {
-                writeln!(out, "primaryKey: {{");
+                // This is left here so we can release the codegen change before releasing a new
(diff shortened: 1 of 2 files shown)


bfops:
(I'm removing the codeowner reviewers since they're no longer codeowners for any changed files)

cloutiertyler (approved):
This LGTM. It's quite straightforward

PR #2673, opened 2025-04-27
Add information about the CLI version to generated code.

jsdt:
# Description of Changes

This updates the comments at the top of all the generated files to include the cargo version and git commit of the CLI used to generate the code. For typescript, it also adds the `cliVersion` to the `REMOTE_MODULE`, so the typescript SDK will be able to reference it (which should be helpful for adding context to logs and eventually checking for version drift between the SDK and codegen versions).

To avoid needing to update code snapshots constantly with `insta review`, I added some filters to ignore the parts of generated code where we are inserting the version info.

This isn't perfect, since `build.rs` doesn't always run when the current git commit changes, but releases should get the correct value (the existing `--version` flag of the CLI has the same issue).

There is a corresponding [private PR](https://github.com/clockworklabs/SpacetimeDBPrivate/pull/1654) to update CI checks.

This doesn't change any APIs.

2. This adds filters to the codegen snapshots, which could start to get more complicated.

# Testing

This has testing in the snapshot, and I've also looked at the generated code.


bfops (approved):
My codeowned diff in `crates/cli/src/version.rs` LGTM.

PR #2675, opened 2025-04-28
Replication Smoketest Cleanups

kim:
- Wrap all replication tests in a class
  Enables use of `-k` and `-x` for individual tests as well as module
- Add `--list` argument for debugging of `-k` and `-x` flags
- Move docker management stuff to top-level `docker.py`
- Sleep after cluster restart as workaround to wait for leader election

On top of #2148


Shubham8287 (approved):
Thanks,  Seems good.
Ran it on my another branch to test Both :)

PR #2676, opened 2025-04-28
Avoid reflection when validating enums

joshua-spacetime:
# Description of Changes

Makes enum validation faster during deserialization in C# by removing reflection.

None

1

# Testing

- [ ] Unity profile


kazimuth (approved):
👍

PR #2677, opened 2025-04-29
Avoid reflection when constructing row in bsatn deserialization

joshua-spacetime:
# Description of Changes

Avoids paying `new T()`'s (potential) cost of reflection on each row decode.

None

1

# Testing

- [ ] Unity profile


kazimuth:
I'd like to check if this actually monomorphizes in Unity

kazimuth:
From deep profiling, this does not result in monomorphization.

I found https://andrewlock.net/benchmarking-4-reflection-methods-for-calling-a-constructor-in-dotnet/ which shows that activator.createinstance is still quite fast -- 40ns, versus 10ns for `new()`. I think this may be a case of the unity deep profiler blowing up execution times. But I'm not completely confident about that.

joshua-spacetime:
A bot test showed no improvement. Closing this in favor of the following codegen approach https://github.com/clockworklabs/SpacetimeDB/pull/2725.

Issue #2612, opened 2025-04-14
Feature request: upsert


jdetter:
Something like this:
```
// Pseudocode
func upsert(ctx, primary_key_value, value):
  if ctx.primary_key_column.try_find(primary_key_value):
    // Row with a matching primary key column exists
    ctx.primary_key_column.delete(primary_key_value)
  ctx.primary_key_column.insert(primary_key_value, value)
  
```

In order to `upsert` on a column it needs to have a unique constraint on it.

gefjon:
Reopening, as the linked PR does not include C# support, and is currently marked unstable.

sunkin351:
If I may, perhaps the name "InsertOrUpdate" or similar would be better. Yes, it's longer, but there is then no ambiguity on first glance about what the function does.

gefjon:
> If I may, perhaps the name "InsertOrUpdate" or similar would be better. Yes, it's longer, but there is then no ambiguity on first glance about what the function does.

@sunkin351 We did indeed settle on `insert_or_update` in the unstable Rust implementation, #2678 .

PR #2678, opened 2025-04-29
Simple implementation of insert_or_update and try_insert_or_update

rafaelGuerreiro:
# Description of Changes

Implementing insert_or_update and try_insert_or_update following the issue guideline, this is a naive implementation that actually has to first find and then decide whether it's insert or update.

It's not clear to me how we want this feature to behave with auto_inc ids, if the id is 0, should we simply call insert?

I think I also need to provide the C# implementation, I'd like some guidance on that side.

This resolves https://github.com/clockworklabs/SpacetimeDB/issues/2612

It introduces 2 public methods, which are marked with the unstable feature until we're set on the API.

Complexity for this solution is 1, but if we want to have it performed in a single operation, it might be a 3, I'm not sure. I tried diving into the internals and wasn't too sure on the right approach.

# Testing

I need guidance on testing.


crates/bindings/src/table.rs
```diff
@@ -388,6 +388,30 @@ impl<Tbl: Table, Col: Index + Column<Table = Tbl>> UniqueColumn<Tbl, Col::ColTyp
...
+    #[cfg(feature = "unstable")]
+    pub fn try_upsert(&self, new_row: Tbl::Row) -> Result<Tbl::Row, TryInsertError<Tbl>> {
+        let index = Col::get_field(&new_row);
+        match self._find(index) {
+            None => {
+                let buf = IterBuf::take();
+                insert::<Tbl>(new_row, buf)
+            }
+            Some(_existing) => Ok(self.update(new_row)),
+        }
```
gefjon, line 403: It would be more efficient (and semantically equivalent) to do a delete, ignore the result, and then do an insert.

crates/bindings/src/table.rs
```diff
@@ -388,6 +388,30 @@ impl<Tbl: Table, Col: Index + Column<Table = Tbl>> UniqueColumn<Tbl, Col::ColTyp
         let buf = IterBuf::take();
         update::<Tbl>(Col::index_id(), new_row, buf)
     }
+
+    /// Attempts to insert `new_row` into the table. If it exists, deletes the row with the same value in the unique column.
+    #[track_caller]
+    #[cfg(feature = "unstable")]
+    pub fn try_upsert(&self, new_row: Tbl::Row) -> Result<Tbl::Row, TryInsertError<Tbl>> {
```
gefjon, line 395: "Upsert" is the colloquial name for this operation, but I'd prefer the more explicit `try_insert_or_update`, and `insert_or_update` for the panicking version.

If you like, you could also add a [rustdoc alias](https://doc.rust-lang.org/rustdoc/advanced-features.html#add-aliases-for-an-item-in-documentation-search) with the name `try_upsert` (and `upsert` on the panicking version) to aid discovery.

crates/bindings/src/table.rs
```diff
@@ -388,6 +388,30 @@ impl<Tbl: Table, Col: Index + Column<Table = Tbl>> UniqueColumn<Tbl, Col::ColTyp
         let buf = IterBuf::take();
         update::<Tbl>(Col::index_id(), new_row, buf)
     }
+
+    /// Attempts to insert `new_row` into the table. If it exists, deletes the row with the same value in the unique column.
```
gefjon, line 392: I think this doc comment is not super clear. It seems to imply: Attempt to insert a new row. If successful, exit. If failure, instead delete the existing row, then exit without (re-)inserting the new row.

I would instead say something like:

```suggestion
    /// Insert `new_row` into the table,
    /// first checking for an existing row with a matching value in the unique column
    /// and deleting it if present.
```

Please use this or similar language both here and in the comment on the panicking version.

gefjon (changes requested):
Thanks for this!

> It's not clear to me how we want this feature to behave with auto_inc ids, if the id is 0, should we simply call insert?

It's not necessary to specifically handle the case where an auto-inc column is supplied and zero. In this case, the row cannot be already present, so the upsert operation will behave like an insert.

> I think I also need to provide the C# implementation, I'd like some guidance on that side.

I would not block this PR on a C# implementation, but I would be very happy to have one included if you're able.

> It introduces 2 public methods, which are marked with the unstable feature until we're set on the ABI.

This is an API. We use "ABI" (Application Binary Interface) to refer to our set of WASM imports and exports, and the memory representations they use to communicate row and column values between the host and the WASM code.

> Complexity for this solution is 1, but if we want to have it performed in a single operation, it might be a 3, I'm not sure. I tried diving into the internals and wasn't too sure on the right approach.

We definitely prefer to implement this purely in WASM the way you've done, rather than implementing a new ABI / import function. We only define new ABIs when we really need to, either because profiling has shown a clear performance bottleneck, or because it's not possible to implement the functionality within WASM at all.

> I need guidance on testing.

This PR is simple enough that I'd be willing to accept it with minimal testing. Add a call to the new functions somewhere in `modules/module-test` (and `module-test-cs` if you implement the C# version) so that we know the methods continue to exist and typecheck, and then do some amount of manual testing that the methods behave the way you expect. If you were interested in learning more about our testing setup, we could talk you through adding a new smoketest, but that'd be purely voluntary.

crates/bindings/src/table.rs
rafaelGuerreiro: I didn't test it, but I'm not sure they're equivalent. Maybe we need the try_update as well to avoid panicking.

What do we want to do in this situation:

```rust
#[table(name = my_table_v1, public)]
pub struct MyTableV1 {
    #[primary_key]
    pub id: u64,

    #[unique]
    pub name: String,
}

// Suppose we already have 2 records in this table:
// {"id": 1, "name": "abc"}
// {"id": 2, "name": "def"}

#[reducer]
fn change_my_table_v1(ctx: &ReducerContext) {
  // This will cause the name to conflict.
  let row = MyTableV1 { id: 1, name: "def".to_string() };

  // Here we discard the Result because we want to commit it
  let _ = ctx.db.my_table_v1().id().try_insert_or_update(row);

  // At this point, the row with id 1 is deleted. Is that expected?
}
```

rafaelGuerreiro:
@gefjon, thank you for reviewing and explaining all of that. I really appreciate it and I'll make all the changes.

gefjon:
> Is it possible that a column becomes auto-inc after the row 0 was inserted? In this case, it would definitely break the auto-inc behaviour because a row with id 0 might already exist, and that same row is always updated in the current implementation.

That's hypothetically possible, but it's pathological, and I don't think it needs or deserves special handling. `update` will also behave oddly if passed a zero value for an auto_inc column when a row with zero exists: it will delete the zero row, then compute the next sequence value and perform the auto-inc substitution in the new row.

rafaelGuerreiro:
@gefjon, all of that makes sense to me. I've applied all the changes.

crates/bindings/src/table.rs
```diff
@@ -388,6 +388,38 @@ impl<Tbl: Table, Col: Index + Column<Table = Tbl>> UniqueColumn<Tbl, Col::ColTyp
...
+    /// but the previous row will be deleted. If you propagate the error, SpacetimeDB will
+    /// rollback the transaction and the old row will be restored. If you ignore the error,
+    /// the old row will be lost.
+    #[track_caller]
+    #[doc(alias = "try_upsert")]
+    #[cfg(feature = "unstable")]
+    pub fn try_insert_or_update(&self, new_row: Tbl::Row) -> Result<Tbl::Row, TryInsertError<Tbl>> {
+        let col_val = Col::get_field(&new_row);
+        // If the row doesn't exist, delete will return false, which we ignore.
+        let _ = self.delete(col_val);
```
rafaelGuerreiro, line 405: @gefjon, do I need to call `fn _delete(&self, col_val: &Col::ColType) -> (bool, IterBuf)` instead and reuse IterBuf for the insert?

gefjon (approved):
Looks great, thanks!

bfops:
It looks like the tests are failing due to build issues. I don't personally have context on how to fix these, or how to plumb the new methods through to the right places.

rafaelGuerreiro:
> It looks like the tests are failing due to build issues. I don't personally have context on how to fix these, or how to plumb the new methods through to the right places.

I think this is because I flagged these methods with the "unstable" flag. Let me see if I can get it to work on the test with the unstable over there as well.

Issue #2679, opened 2025-04-29
`spacetime version list` does not indicate the current version


bfops:
e.g.
```
$ spacetime version list
1.1.1
1.1.0
```

The code says it's meant to show `(current)` somewhere.

PR #2680, opened 2025-04-29
Fix `spacetime version list` not showing current version

bfops:
# Description of Changes

See https://github.com/clockworklabs/SpacetimeDB/issues/2679.

At some point, this code was changed so that `current` contains an entire path, rather than just the version string itself, so the comparison between `ver` and `current` was always failing.

Not a breaking change.

1
# Testing

- [x] It works now!
```
$ cargo run -pspacetimedb-update -- version list
   Compiling spacetimedb-update v1.1.1 (/home/lead/work/clockwork-localhd/SpacetimeDBPrivate/public/crates/update)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 3.63s
     Running `target/debug/spacetimedb-update version list`
1.1.1
1.1.0 (current)
```

I could be convinced to add a smoketest for this.

### DIFF
--- a/crates/update/src/cli/list.rs
+++ b/crates/update/src/cli/list.rs
@@ -1,4 +1,5 @@
 use spacetimedb_paths::SpacetimePaths;
+use std::path::Path;
 
 /// List installed SpacetimeDB versions.
 #[derive(clap::Args)]
@@ -10,7 +11,16 @@ pub(super) struct List {
 
 impl List {
     pub(super) fn exec(self, paths: &SpacetimePaths) -> anyhow::Result<()> {
-        let current = paths.cli_bin_dir.current_version()?;
+        let current = match paths.cli_bin_dir.current_version()? {
+            None => None,
+            Some(path_str) => {
+                let file_name = Path::new(&path_str)
+                    .file_name()
+                    .and_then(|f| f.to_str())
+                    .ok_or(anyhow::anyhow!("Could not extract current version"))?;
+                Some(file_name.to_string())
+            }
+        };
         let versions = if self.all {
             let client = super::reqwest_client()?;
             super::tokio_block_on(super::install::available_releases(&client))??


coolreader18:
Oh, whoops - I think the actual root cause is in a different function - this should be a thorough fix. It does mean that the current version will have to be re-set for it to work.
```diff
diff --git a/crates/paths/src/cli.rs b/crates/paths/src/cli.rs
index 28d26fb81..f39761c09 100644
--- a/crates/paths/src/cli.rs
+++ b/crates/paths/src/cli.rs
@@ -85,7 +85,7 @@ impl VersionBinDir {
     }
 
     fn link_to(&self, path: &Path) -> anyhow::Result<()> {
-        let rel_path = path.strip_prefix(self).unwrap_or(path);
+        let rel_path = path.strip_prefix(self.0.parent().unwrap()).unwrap_or(path);
         #[cfg(unix)]
         {
             // remove the link if it already exists
```

bfops:
> Oh, whoops. I think the actual root cause is in a different function - this should be a thorough fix. It does mean that the current version will have to be re-set for it to work.
> 
> ```diff
> diff --git a/crates/paths/src/cli.rs b/crates/paths/src/cli.rs
> index 28d26fb81..f39761c09 100644
> --- a/crates/paths/src/cli.rs
> +++ b/crates/paths/src/cli.rs
> @@ -85,7 +85,7 @@ impl VersionBinDir {
>      }
>  
>      fn link_to(&self, path: &Path) -> anyhow::Result<()> {
> -        let rel_path = path.strip_prefix(self).unwrap_or(path);
> +        let rel_path = path.strip_prefix(self.0.parent().unwrap()).unwrap_or(path);
>          #[cfg(unix)]
>          {
>              // remove the link if it already exists
> ```

Ah okay! thanks for the better fix. Do you have any suggestions for how we fix this for people who already have SpacetimeDB installed? I guess it would just get quietly fixed the next time they upgrade?

jdetter (approved):
Tested + working, thanks Zeke 👍

PR #2682, opened 2025-04-29
Single-threaded subscription evaluation

joshua-spacetime:
# Description of Changes

Removes rayon from subscription evaluation and replaces its parallel iterators with sequential ones, the purpose of which is to optimize the common case of small updates where multithreading just adds overhead.

None

1

Just removes rayon and replaces `into_par_iter()` with `into_iter()`.

# Testing

- [x] Bot test

### DIFF
--- a/crates/core/src/subscription/module_subscription_manager.rs
+++ b/crates/core/src/subscription/module_subscription_manager.rs
@@ -12,7 +12,6 @@ use crate::subscription::delta::eval_delta;
 use hashbrown::hash_map::OccupiedError;
 use hashbrown::{HashMap, HashSet};
 use itertools::Itertools;
-use rayon::iter::{IntoParallelIterator, ParallelIterator};
 use spacetimedb_client_api_messages::websocket::{
     BsatnFormat, CompressableQueryUpdate, FormatSwitch, JsonFormat, QueryId, QueryUpdate, WebsocketFormat,
 };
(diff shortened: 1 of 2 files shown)


crates/core/src/subscription/module_subscription_manager.rs
```diff
@@ -734,299 +738,249 @@ impl SubscriptionManager {
 
         let tables = &event.status.database_update().unwrap().tables;
 
-        // Put the main work on a rayon compute thread.
-        rayon::scope(|_| {
```
joshua-spacetime, line 738: This was the only code that was removed, along with ...

crates/core/src/subscription/module_subscription_manager.rs
```diff
@@ -734,299 +738,249 @@ impl SubscriptionManager {
...
-            // have good characteristics for this use case, namely cheap appension
-            // of the result of each thread.
-            #[derive(Default)]
-            struct ReduceState<'a> {
-                updates: VecList<ClientUpdate<'a>>,
-                errs: VecList<(&'a ClientId, Box<str>)>,
-                metrics: ExecutionMetrics,
-            }
-            impl<'a> ReduceState<'a> {
-                /// Convert the result of a single-thread fold to get ready for a multi-thread reduce.
```
joshua-spacetime, line 772: ... `ReduceState` since it's only needed for a multi-threaded fold. The rest of the code is the same as before.

crates/core/src/subscription/module_subscription_manager.rs
```diff
@@ -1044,6 +998,13 @@ fn send_to_client(client: &ClientConnectionSender, message: impl Into<Serializab
 /// > collection, but have no specific requirements for what that collection
 /// > should be. [...] This is a very efficient way to collect an unindexed
 /// > parallel iterator, without much intermediate data movement.
+///
+/// Note: This is currently unused.
```
gefjon, line 1002: If this is unused, should we just delete it?

gefjon (commented):
Changes look good. I don't love having `#[allow(unused)]` code committed; it feels a bit too much like commented-out code. I'd rather just remove the thing we don't need any more, and we can add it back in down the line if/when it's useful again.

gefjon (approved):

PR #2688, opened 2025-05-01
Reduce allocations in C# when deserializing lists and arrays

kazimuth:
# Description of Changes

Since we know the size of lists and arrays, we can allocate them up-front. This also skips allocating an `IEnumerable`.

N/A

0

# Testing

Automated tests should break immediately if this is wrong

### DIFF
--- a/crates/bindings-csharp/BSATN.Runtime/BSATN/Runtime.cs
+++ b/crates/bindings-csharp/BSATN.Runtime/BSATN/Runtime.cs
@@ -403,8 +403,18 @@ public readonly struct Array<Element, ElementRW> : IReadWrite<Element[]>
     where ElementRW : IReadWrite<Element>, new()
 {
     private static readonly Enumerable<Element, ElementRW> enumerable = new();
+    private static readonly ElementRW elementRW = new();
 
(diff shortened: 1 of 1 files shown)


crates/bindings-csharp/BSATN.Runtime/BSATN/Runtime.cs
```diff
@@ -403,8 +403,18 @@ public AlgebraicType GetAlgebraicType(ITypeRegistrar registrar) =>
     where ElementRW : IReadWrite<Element>, new()
 {
     private static readonly Enumerable<Element, ElementRW> enumerable = new();
+    private static readonly ElementRW elementRW = new();
 
-    public Element[] Read(BinaryReader reader) => enumerable.Read(reader).ToArray();
+    public Element[] Read(BinaryReader reader)
+    {
+        var count = reader.ReadInt32();
```
kazimuth, line 410: Enumerable also uses ReadInt32, although I'm slightly confused why it isn't ReadUint32.

joshua-spacetime (approved):
Can you add a comment to make sure we don't revert this change in the future

rekhoff (approved):
Make sense to me, save an allocation :) I see no issues with this. Approved!

PR #2691, opened 2025-05-02
snapshot: Provide streaming snapshot verification.

kim:
Repurposes the remote `SnapshotFetcher` for verification of the snapshot's objects. This can be useful when memory usage is of concern.


Shubham8287 (approved):
Looks good

Issue #1837, opened 2024-10-10
Option for not receiving reducer args


jdetter:
- We have a client which doesn't use the reducer args and they've requested to not receive them. This is medium priority and we'll reprioritize this to high priority if we decided we need to reduce network bandwidth further.

Related PR (does not solve this issue): <https://github.com/clockworklabs/SpacetimeDB/pull/1812>

bfops:
Is this still requested?

jdetter:
@bfops This is frequently requested in the Discord for security reasons - people want to be able to write reducers like this:

```rust
pub fn create_user(username: String, password: String, ...) {
    // ...
}
```

You can't do this right now in SpacetimeDB since there is no way to say which reducers should or should not broadcast reducer arguments.

bfops:
ah I was imagining this as a performance optimization. from a security perspective I think this will overlap the private reducer proposal that.. I'm not sure whether we have a separate ticket for?

bfops:
https://github.com/clockworklabs/SpacetimeDBPrivate/issues/787
https://github.com/clockworklabs/SpacetimeDBPrivate/issues/1171

I think these are more appropriate for that concern.

bfops:
Why doesn't the existing issue address this?

bfops:
We're going to call the original issue done; we have those linked issues for further work on the security side.

egormanga:
Anything public the outside users can subscribe to to track progress?

bfops:
Hey @egormanga, we are working on a public roadmap, but for now I can say that we aren't actively working on these features right now. We're currently largely focused on auth-related features and web support.

Would you want to share more about your use case for hiding reducer args from all clients?

egormanga:
Currently, it involves hiding players' positions from each other when using server-authoritative physics.

bfops:
@egormanga You should be able to achieve this right now by putting the player positions in an RLS-protected table, and having players send incremental updates rather than absolute positions (e.g. sending "move in the (1, 1) direction" rather than "move to (x, y)"). This would make the player positions fully server-authoritative, and hidden from the players.

I hope that helps for now!

egormanga:
Thank you for the solution, it works for simpler cases such as movement, inventory actions, crafting and so on, where both the server and the client both already know the data, but it pretty obviously fails as soon as you begin processing private user-generated content, such as direct chat messages.

[...]

PR #2694, opened 2025-05-02
Disable reducer arg broadcast

suirad:
# Description of Changes

By default, reducer callbacks are broadcasted to all clients of the module. This PR adds a command line flag to `spacetime start` that when given, disables reducer arguments from being sent along with the reducer callback. It does so by swapping out the Args struct to the null version of it.

Potentially solves #1837 or at least it could temporarily.

None; adds an optional command line flag to `spacetime start` to enable new behavior. `--no-reducer-args`

Level is about 2. The change is pretty trivial, however the depth of that change to make it command line configurable adds a bit of complexity. There may be some unknowns that I am not aware of by not broadcasting the full args of reducers to other clients. This is mostly why I didn't make a more substantive change.

# Testing

Successfully ran the repo test suite.

Additional tests against existing projects and a project to specifically to demonstrate the corrected behavior via extracting passwords from third party reducer args was done by @lethalchip. See below images:

## Setup
image
image

## Use Case

image

## Results

image
image

### DIFF
--- a/crates/core/src/host/host_controller.rs
+++ b/crates/core/src/host/host_controller.rs
@@ -92,6 +92,8 @@ pub struct HostController {
     pub page_pool: PagePool,
     /// The runtimes for running our modules.
     runtimes: Arc<HostRuntimes>,
+    /// Flag for hiding reducer callback arguments from broadcasting to non-caller
+    disable_reducer_args: bool,
 }
 
 struct HostRuntimes {
@@ -166,6 +168,7 @@ impl HostController {
         program_storage: ProgramStorage,
         energy_monitor: Arc<impl EnergyMonitor>,
         durability: Arc<dyn DurabilityProvider>,
+        disable_reducer_args: bool,
     ) -> Self {
         Self {
             hosts: <_>::default(),
@@ -176,6 +179,7 @@ impl HostController {
             runtimes: HostRuntimes::new(&data_dir),
             data_dir,
             page_pool: PagePool::new(default_config.page_pool_max_size),
+            disable_reducer_args,
         }
     }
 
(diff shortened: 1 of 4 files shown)


gefjon:
Thank you so much for the contribution! We're aware of this issue and we actually have a planned fix to this issue that we're planning on going with. It has several benefits over this PR. Namely, the issues I see with this solution are:

- This solution is only applicable to self-hosted SpacetimeDB-standalone, not to managed SpacetimeDB-cloud.
- I perceive significant risk of developing a module with the intention of running it with argument broadcast disabled, but mistakenly deploying it with argument broadcast enabled.
- This solution destroys all reducer argument information, and so any module which wants to send any sensitive information in reducer arguments at all is unable to inspect any other, non-sensitive arguments in reducer callbacks.
- Many types do not have obvious empty/null/sentinel values, and our various client SDKs will almost certainly not correctly handle receiving `ArgsTuple::nullary()` instead of an args tuple which conforms to the reducer's argument type. For example, I strongly suspect that this will break the Rust client SDK.

What we actually want is a way to annotate reducers that their event broadcasting should be caller-only. This is a property of the specific reducer, not the database or the host process. When the `SubscriptionManager` goes to broadcast a transaction resulting from a caller-only reducer, it sends a `TransactionUpdateLight` message to all of the clients other than the caller, rather than a full `TransactionUpdate`. The client SDKs will recognize this and expose the event as `Event::UnknownTransaction`, and will not fire reducer callbacks for it.

This is likely to be too involved for a community contribution, just from the perspective of testing and review. It'll require modifications to our `ModuleDef`s, both the Rust and C# module bindings libraries including their macro syntaxes, and also testing with all three of our client SDKs to ensure they correctly handle `TransactionUpdateLight` messages.

PR #2702, opened 2025-05-05
Support for the PG wire protocol

### LINKED ISSUE #2686: Support for Clients Using the `PostgreSQL` Wire Protocol
Some considerations and decision we need in the short-term:

# Authentication Options

## Using SSL/Certs

PostgreSQL supports [certificate-based authentication](https://www.postgresql.org/docs/current/ssl-tcp.html#SSL-CLIENT-CERTIFICATES) over SSL. A typical psql connection looks like:

```bash
psql "port=5432 host=localhost user=postgres sslcert=${STDB_PATH}/id_ecdsa.crt sslkey=${STDB_PATH}/id_ecdsa sslrootcert=${STDB_PATH}/root.crt sslmode=require"
```

To generate the required certificate:

```bash
openssl req -new -x509 \
    -key "${STDB_PATH}/id_ecdsa" \
    -out "${STDB_PATH}/id_ecdsa.crt" \
    -days 365

chmod 0600 "${STDB_PATH}/id_ecdsa"
chmod 0644 "${STDB_PATH}/id_ecdsa.crt"
```

To encode identity, we can either use the `Common Name (CN)` or embed it via `Subject Alternative Name (SAN)`:

```bash
openssl ... -subj "/CN=$IDENTITY"
# or
openssl ... -subj "/CN=postgres" \
  -reqexts SAN -config <(cat /etc/ssl/openssl.cnf \
  <(printf "\n[SAN]\nsubjectAltName=URI:https://my-server.example.com,URI:identity-token:$IDENTITY"))
```

And parse the `identity` from the `CN` or `SAN` field.

Alternatively, the server could derive the `IDENTITY` from the certificate itself, though this requires deeper inspection of the cert chain.

### Compatibility

Some tools (like database explorers, ORMs, etc) may lack support for client certs, and I think is rare for users to use this setup.

## Hijacking `username/password`

To support simpler auth flows, we can overload `username` or `password` fields to carry the `IDENTITY`. Using `password` is preferable for the possibility the tool hide it in logs,
despite `username` being more semantically correct.

The advantage is that is simpler, but could mean the *end user* embed the `IDENTITY` in code.

# Unsupported features that affect the usability of using the protocol.

Many tools expect certain `PostgreSQL` behaviors. For instance, `psql` uses queries against [system views](https://www.postgresql.org/docs/current/views.html) (e.g., `\d`) to introspect schemas.

To support these, we should:
* Implement the expected `system views` (maybe making a custom `module` for this, or doing `in-memory` rewrites to mimic the `pg_catalog` schema).
* And extend our `SQL` dialect to support more clauses (e.g., `ORDER BY, CASE, etc.`) that are emitted by the tools.

Without this, compatibility with standard clients will be severely limited.

The most simple utility to support is `psql`, as it is the most common tool used by developers, and maybe restrict it to the most common commands (e.g., `\d`).

mamcx:
# Description of Changes

Closes [#2686](https://github.com/clockworklabs/SpacetimeDB/issues/2686).

Add support for listening using the [PG wire protocol](https://www.postgresql.org/docs/current/protocol.html) so `pg` clients could be used against the database.

The output of `duration` is changed to `rfc3339`, instead of the way is made with `sats` because is what is done in `pg`, see note below.

2

~~There is open questions that are in the [ticket #2686](https://github.com/clockworklabs/SpacetimeDB/issues/2686). Also the crate used here require `RustTls`, so it could be good idea to decide if~~:

* ~~Rewrite a big chunk of code to use `OpenSSL`~~
* ~~Move to `RustTls`  https://github.com/clockworklabs/SpacetimeDB/pull/1700~~
* ~~Pay for the extra compilation cost~~.

I open another port(`5433`) to listen for `pg` connections using `ssl`. Need to be decided if this is the way or instead try to multi-plex the current port for both protocols.

# Testing

Only manual testing so far. Solving the above questions allow me to implement some unit tests. Also, not yet integrated into cloud for the same reasons.

- [x] Adding some test for the binary encoding of special and primitive types
- [x] Smoke test using `psql` that connect to the db instance and run some queries
- [x] Manually inspect using a UI database explorer how infer the types, some of this tools generate special widgets when displaying `json, duration, etc`


bfops:
(It looks like this is currently failing tests)

mamcx:
> (It looks like this is currently failing tests)

Yes, the unit test was a spurious network error, the others are problems with the smoke tests that need `psql` binary to be present. Working on it...

smoketests/tests/sql.py
```diff
@@ -131,11 +131,11 @@ def test_sql_format(self):
...
- (bool = true, f32 = 594806.56, f64 = -3454353.345389043, str = "This is spacetimedb", bytes = 0x01020304050607, identity = 0x0000000000000000000000000000000000000000000000000000000000000001, connection_id = 0x00000000000000000000000000000000, timestamp = 1970-01-01T00:00:00+00:00, duration = +0.000000)
+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+ (bool = true, f32 = 594806.56, f64 = -3454353.345389043, str = "This is spacetimedb", bytes = 0x01020304050607, identity = 0x0000000000000000000000000000000000000000000000000000000000000001, connection_id = 0x00000000000000000000000000000000, timestamp = 1970-01-01T00:00:00+00:00, duration = P0D)
 """)
```
mamcx: Take in consideration that I switch to `rfc3339` because it is the simpler and more compatible format for the `pg wire` format. The alternative was to re-implement the [special logic for printing it of `pg`](https://www.postgresql.org/docs/current/datatype-datetime.html#DATATYPE-INTERVAL-INPUT) that is complex. Then I make the output be the same for our `sql` and `pg` to not diverge, so it is a breaking change

sunng87:
@mamcx Hello Mario, this is author of pgwire. Let me know if you have any question with the integration. It's also helpful for me to know if you feel any component that can be or should be upstreamed.

crates/cli/src/subcommands/sql.rs
```diff
@@ -475,8 +477,8 @@ Roundtrip time: 1.00ms"#,
             vec![value],
             r#"
  column 0 | column 1 | column 2                                                           | column 3                           | column 4                  | column 5
-----------+----------+--------------------------------------------------------------------+------------------------------------+---------------------------+-----------
- "a"      | 0        | 0x0000000000000000000000000000000000000000000000000000000000000000 | 0x00000000000000000000000000000000 | 1970-01-01T00:00:00+00:00 | +0.000000"#,
+----------+----------+--------------------------------------------------------------------+------------------------------------+---------------------------+----------
+ "a"      | 0        | 0x0000000000000000000000000000000000000000000000000000000000000000 | 0x00000000000000000000000000000000 | 1970-01-01T00:00:00+00:00 | P0D"#,
```
joshua-spacetime: This is a breaking change, is it not?

crates/client-api/src/auth.rs
```diff
@@ -186,6 +210,8 @@ pub trait JwtAuthProvider: Sync + Send + TokenSigner {
     ///
     /// The `/identity/public-key` route calls this method to return the public key to callers.
     fn public_key_bytes(&self) -> &[u8];
+    /// Return the private key used to verify JWTs, as the bytes of a PEM private key file.
+    fn private_key_bytes(&self) -> &[u8];
```
joshua-spacetime: Why do we need to expose the private key?

crates/pg/src/pg_server.rs
```diff
@@ -0,0 +1,435 @@
...
+    let keypair = KeyPair::from_der_and_sign_algo(&private, &rcgen::PKCS_ECDSA_P256_SHA256)?;
+
+    let mut params = CertificateParams::new(vec![
+        "localhost".to_string(),
+        "127.0.0.1".to_string(),
+        "::1".to_string(),
+    ])?;
+    params.distinguished_name = DistinguishedName::new();
+    params.distinguished_name.push(DnType::CommonName, "localhost");
+    let cert = params.self_signed(&keypair)?;
```
joshua-spacetime: Auto generating a self signed cert is fine for dev, but the default behavior should be that the operator must pass in the cert, probably via a new cli arg.

smoketests/tests/sql.py
joshua-spacetime: I don't think breaking change is feasible unfortunately.

crates/pg/src/pg_server.rs
```diff
@@ -0,0 +1,435 @@
...
+    Ok(TlsAcceptor::from(Arc::new(config)))
+}
+
+pub async fn start_pg<T: ControlStateReadAccess + ControlStateWriteAccess + NodeDelegate + 'static>(
+    mut shutdown: watch::Receiver<()>,
+    ctx: Arc<T>,
+    listen_address: &str,
+    private_key: &[u8],
+) {
+    let tls_acceptor = Arc::new(setup_tls(&ctx, private_key).unwrap());
```
joshua-spacetime: Note: Because we're generating a new self signed cert on startup, a restart will break any client that is using `sslmode=verify-ca`. Not sure how much this matters if this is just for dev.

smoketests/tests/sql.py
mamcx: Changed back to old display when called from a `spacetime` client, and add test that show the difference in formats from `pg` clients.

smoketests/tests/pg_wire.py
```diff
@@ -0,0 +1,168 @@
+from .. import Smoketest
+import subprocess
+import os
+import tomllib
+
+
+def psql(identity: str, sql: str) -> str:
```
joshua-spacetime: Can you add the following tests:

1. Connection fails when `sslmode=require`
2. Connection fails with invalid token
3. Empty query works
4. Returns "Feature Not Supported" error for unsupported SpacetimeSQL

crates/client-api/src/auth.rs
mamcx: Is removed now that we remove all TLS

crates/pg/src/pg_server.rs
mamcx: We decided to remove TLS

smoketests/tests/pg_wire.py
mamcx: Done

crates/pg/src/pg_server.rs
```diff
@@ -0,0 +1,405 @@
...
+            let err = String::from_utf8_lossy(&bytes);
+            log::error!("PG: Error for database {database}: {err}");
+            Err(PgError::Sql(format!("{err}")))
+        }
+    }
+}
+
+struct PgSpacetimeDB<T> {
+    ctx: Arc<T>,
+    cached: Mutex<Metadata>,
```
joshua-spacetime: Why does this need a mutex?

crates/pg/src/pg_server.rs
```diff
@@ -0,0 +1,405 @@
...
+pub async fn start_pg<T: ControlStateReadAccess + ControlStateWriteAccess + NodeDelegate + 'static>(
+    mut shutdown: watch::Receiver<()>,
+    ctx: Arc<T>,
+    listen_address: &str,
+) {
+    let auth = SpacetimeAuth::alloc(&ctx).await.unwrap();
+    let factory = Arc::new(PgSpacetimeDBFactory::new(ctx, auth));
+
+    let server_addr = format!("{}:5432", listen_address.split(':').next().unwrap());
+    let tcp = TcpListener::bind(server_addr).await.unwrap();
```
joshua-spacetime: Can we make the tcp listener an argument of this function rather than instantiating it within the function?

crates/pg/src/pg_server.rs
```diff
@@ -0,0 +1,405 @@
...
+        PgWireError: From<<C as Sink<PgWireBackendMessage>>::Error>,
+    {
+        match message {
+            PgWireFrontendMessage::Startup(ref startup) => {
+                save_startup_parameters_to_metadata(client, startup);
+                client.set_state(PgWireConnectionState::AuthenticationInProgress);
+
+                let login_info = LoginInfo::from_client_info(client);
+
+                log::debug!("PG: Login info: {login_info:?}");
```
joshua-spacetime: We should remove this

crates/pg/src/pg_server.rs
```diff
@@ -0,0 +1,405 @@
...
+                    "PG: Connected to database: {user}@{database} using identity `{}`",
+                    auth.identity
+                );
+
+                let metadata = Metadata { database, auth };
+                self.cached.lock().await.clone_from(&metadata);
+                finish_authentication(client, &self.parameter_provider).await?;
+            }
+            PgWireFrontendMessage::SslRequest(ssl) => {
+                if ssl.is_some() {
```
joshua-spacetime: Do we care if `ssl.is_some()`?

crates/pg/src/pg_server.rs
```diff
@@ -0,0 +1,405 @@
...
+                    let err = PgError::SSLNotSupported;
+                    log::warn!("{err}");
+                    let err = ErrorInfo::new("FATAL".to_owned(), "28P01".to_owned(), format!("PG: {err}"));
+                    return close_client(client, err).await;
+                }
+            }
+            _ => {}
```
joshua-spacetime: Do we not need to handle the other messages?

crates/pg/src/encoder.rs
```diff
@@ -0,0 +1,249 @@
...
+
+pub(crate) fn type_of(schema: &ProductType, ty: &ProductTypeElement) -> Type {
+    let format = PsqlPrintFmt::use_fmt(schema, ty, ty.name());
+    match &ty.algebraic_type {
+        AlgebraicType::String => Type::VARCHAR,
+        AlgebraicType::Bool => Type::BOOL,
+        AlgebraicType::U8 | AlgebraicType::I8 | AlgebraicType::I16 => Type::INT2,
+        AlgebraicType::U16 | AlgebraicType::I32 => Type::INT4,
+        AlgebraicType::I64 => Type::INT8,
+        AlgebraicType::U32
```
joshua-spacetime: Should we map U32 to INT8 instead of NUMERIC?

crates/pg/src/encoder.rs
```diff
@@ -0,0 +1,249 @@
...
+        | AlgebraicType::I128
+        | AlgebraicType::U128
+        | AlgebraicType::I256
+        | AlgebraicType::U256 => Type::NUMERIC,
+        AlgebraicType::F32 => Type::FLOAT4,
+        AlgebraicType::F64 => Type::FLOAT8,
+        AlgebraicType::Array(ty) => match *ty.elem_ty {
+            AlgebraicType::String => Type::VARCHAR_ARRAY,
+            AlgebraicType::Bool => Type::BOOL_ARRAY,
+            AlgebraicType::U8 => Type::BYTEA,
```
joshua-spacetime, line 44: Should this also be `INT2_ARRAY` just like it is for `I8`?

crates/pg/src/encoder.rs
```diff
@@ -0,0 +1,249 @@
...
+            _ => Type::ANYARRAY,
+        },
+        AlgebraicType::Product(_) => match format {
+            PsqlPrintFmt::Hex => Type::BYTEA_ARRAY,
+            PsqlPrintFmt::Timestamp => Type::TIMESTAMP,
+            PsqlPrintFmt::Duration => Type::INTERVAL,
+            _ => Type::JSON,
+        },
+        AlgebraicType::Sum(sum) if sum.is_simple_enum() => Type::ANYENUM,
+        _ => Type::UNKNOWN,
```
joshua-spacetime: Why not map all sum types to `JSON`?

mamcx: PG has support for [simple enums](https://www.postgresql.org/docs/current/datatype-enum.html)

crates/pg/src/encoder.rs
mamcx: In `satn` is `bytes`, and is more common to map to bytes in pg too

crates/pg/src/pg_server.rs
mamcx: Because `async fn on_startup<C>(&self,` make non-mutable the `Metadata`

crates/pg/src/pg_server.rs
mamcx: Yes, because comment on crate:

```
    // when client has no ssl configured, it skip this message.
    // our decoder will return a `SslRequest(None)` for this case.```

File: crates/pg/src/pg_server.rs (hunk 11 above)
Comment (mamcx) [reply crates/pg/src/pg_server.rs]: No, that ones are catch by the parser. I left a comment

File: crates/pg/src/pg_server.rs [hunk 15]
```diff
@@ -0,0 +1,402 @@
...
+                let metadata = Metadata { database, auth };
+                self.cached.lock().await.clone_from(&metadata);
+                finish_authentication(client, &self.parameter_provider).await?;
+            }
+            PgWireFrontendMessage::SslRequest(ssl) => {
+                if ssl.is_some() {
+                    let err = PgError::SSLNotSupported;
+                    log::warn!("{err}");
+                    let err = ErrorInfo::new("FATAL".to_owned(), "28P01".to_owned(), format!("PG: {err}"));
+                    return close_client(client, err).await;
```
joshua-spacetime: Should we close the client for an `SslRequest`? Aren't we supposed to return an `N` response instead?

crates/pg/src/pg_server.rs
```diff
@@ -0,0 +1,402 @@
...
+                }
+            }
+            _ => {
+                // The other messages are for features not supported by SpacetimeDB, that are rejected by the parser.
```
joshua-spacetime: Are other message variants even possible `on_startup`? If they are we should handle them, but if they aren't we should fail loudly. Either with `unreachable!` or by closing the connection if that is applicable.

crates/pg/src/pg_server.rs
```diff
@@ -0,0 +1,402 @@
...
+                    Err(PgError::Pg(PgWireError::UserError(err))) => {
+                        return close_client(client, *err).await;
+                    }
+                    Err(err) => {
+                        return Err(err.into());
+                    }
+                };
+
+                let auth = match validate_token(&self.ctx, &pwd.password).await {
+                    Ok(claims) => response(SpacetimeAuth::from_claims(&self.ctx, claims), &database).await?,
```
joshua-spacetime: Why are we creating a new token here?

crates/sats/src/product_value.rs
```diff
@@ -114,6 +114,14 @@ impl ProductValue {
         })
     }
 
+    /// Assumes that we are dealing with a [`ProductType::is_special`] value if it has a single element.
+    pub fn as_special_value_raw(&self) -> Option<&AlgebraicValue> {
```
joshua-spacetime: Is this used anymore?

crates/client-api/src/auth.rs
```diff
@@ -132,6 +132,30 @@ impl TokenClaims {
 }
 
 impl SpacetimeAuth {
+    pub fn from_claims(
```
joshua-spacetime: I'm not sure this function is necessary

crates/pg/src/pg_server.rs
```diff
@@ -0,0 +1,402 @@
...
+        Arc::new(NoopErrorHandler)
+    }
+}
+
+pub async fn start_pg<T: ControlStateReadAccess + ControlStateWriteAccess + NodeDelegate + 'static>(
+    mut shutdown: watch::Receiver<()>,
+    ctx: Arc<T>,
+    tcp: TcpListener,
+) {
+    let auth = SpacetimeAuth::alloc(&ctx).await.unwrap();
```
joshua-spacetime: Why do we need this call to `alloc`? This will allocate a new identity and mint a new token.

crates/client-api/src/routes/database.rs
```diff
@@ -378,19 +379,19 @@ pub(crate) async fn worker_ctx_find_database(
...
-    Path(SqlParams { name_or_identity }): Path<SqlParams>,
-    Query(SqlQueryParams {}): Query<SqlQueryParams>,
-    Extension(auth): Extension<SpacetimeAuth>,
-    body: String,
-) -> axum::response::Result<impl IntoResponse>
+pub async fn sql_direct<S>(
+    worker_ctx: S,
+    SqlParams { name_or_identity }: SqlParams,
+    _params: SqlQueryParams,
+    auth: SpacetimeAuth,
```
joshua-spacetime: I don't think we need to pass `SpacetimeAuth` here. All we care about is the caller identity.

crates/pg/src/pg_server.rs
```diff
@@ -0,0 +1,402 @@
...
+        } else {
+            PgWireError::ApiError(Box::new(err))
+        }
+    }
+}
+
+#[derive(Clone)]
+struct Metadata {
+    database: String,
+    auth: SpacetimeAuth,
```
joshua-spacetime: I don't think we need this struct for the following reasons:

1. The database identity is already stored as part of the client metadata (`METADATA_DATABASE `)
2. We validate the token once during connection startup. After that, the only thing we care about is the caller identity, which again is stored as part of the client metadata (`METADATA_USER`).

crates/pg/src/pg_server.rs
```diff
@@ -0,0 +1,402 @@
...
+            let err = String::from_utf8_lossy(&bytes);
+            log::error!("PG: Error for database {database}: {err}");
+            Err(PgError::Sql(format!("{err}")))
+        }
+    }
+}
+
+struct PgSpacetimeDB<T> {
+    ctx: Arc<T>,
+    cached: Mutex<Metadata>,
```
joshua-spacetime: See my previous comment as to why I believe we can remove this field.

smoketests/tests/pg_wire.py
```diff
@@ -0,0 +1,201 @@
...
+    u32: u32,
+    u64: u64,
+    u128: u128,
+    u256: u256,
+}
+
+#[spacetimedb::table(name = t_uints_tuple, public)]
+pub struct TUintsTuple {
+    tuple: TUints,
+}
```
joshua-spacetime, line 69: Can you add a test for enums - simple and non-simple

smoketests/tests/pg_wire.py
```diff
@@ -0,0 +1,201 @@
+from .. import Smoketest
+import subprocess
+import os
+import tomllib
+
+
+def psql(identity: str, sql: str, extra=None) -> str:
```
joshua-spacetime, line 8: Could you add tests for prepared statements?

Could you also test a long lived connection using `psycopg`? That is open a connection using `psycopg` and issue multiple queries over the same connection?

crates/client-api/src/routes/database.rs
mamcx: I think it makes more clear from where the identity comes without look upstream

crates/client-api/src/auth.rs
mamcx: Is used on `on_startup`

crates/pg/src/pg_server.rs
mamcx: I don't have access to that params on `fn exe_sql<'a>(&self, query: String)`, that is why I need to preserve them with this

crates/pg/src/pg_server.rs
mamcx: This ensure that the error is returned and the messages are `flused`

crates/pg/src/pg_server.rs
mamcx: > Are other message variants even possible on_startup?

Per https://github.com/sunng87/pgwire/blob/fced78d34aaf4f664639f010d8b44c911adb238c/src/api/auth/sasl.rs#L49, no. The `enum` is shared by all the major steps.

crates/pg/src/pg_server.rs
mamcx: Removed the redundant alloc before this

crates/pg/src/pg_server.rs
joshua-spacetime: But is this what clients expect? Based on my reading it sounds like clients expect an `N` response.

crates/pg/src/pg_server.rs
joshua-spacetime: Then I might add an `unreachable!` here

crates/pg/src/pg_server.rs
joshua-spacetime: Note, you can access those params through `ClientInfo` which is available through `do_query`.

crates/client-api/src/auth.rs
joshua-spacetime: I see. But I'm not sure why it's needed there. We really only need the caller's identity. We shouldn't need to mess with claims at all.

crates/standalone/src/subcommands/start.rs
```diff
@@ -176,12 +177,26 @@ pub async fn exec(args: &ArgMatches, db_cores: JobCores) -> anyhow::Result<()> {
...
 
     let tcp = TcpListener::bind(listen_addr).await?;
     socket2::SockRef::from(&tcp).set_nodelay(true)?;
-    log::debug!("Starting SpacetimeDB listening on {}", tcp.local_addr().unwrap());
-    axum::serve(tcp, service).await?;
+    log::debug!("Starting SpacetimeDB listening on {}", tcp.local_addr()?);
+    let pg_server_addr = format!("{}:5432", listen_addr.split(':').next().unwrap());
+    let tcp_pg = TcpListener::bind(pg_server_addr).await?;
+
+    let (shutdown_tx, mut shutdown_rx) = tokio::sync::watch::channel(());
```
coolreader18: This would probably make more sense to be `tokio::sync::Notify`.

mamcx: Ok, but need some `Arc::clone`?

crates/pg/src/pg_server.rs
mamcx: Yes, this is the expectation

crates/pg/src/pg_server.rs
mamcx: Removed

crates/standalone/src/subcommands/start.rs
coolreader18: You can have `Arc<Notify>`

smoketests/tests/pg_wire.py
```diff
@@ -0,0 +1,274 @@
...
+        self.assertSql(token, "SELECT * FROM t_simple_enum", """\
+id |  action
+----+----------
+  1 | Inactive
+  2 | Active
+(2 rows)""")
+        self.assertSql(token, "SELECT * FROM t_enum", """\
+id | color
+----+-------
+  1 | 128
```
joshua-spacetime: This doesn't seem quite right. We've lost the variant which is `Gray` in this case.

crates/pg/src/pg_server.rs
joshua-spacetime: What happens in the case of `sslmode=prefer`? What response does the server send?

smoketests/tests/pg_wire.py
```diff
@@ -0,0 +1,274 @@
...
+import tomllib
+import psycopg2
+
+
+def psql(identity: str, sql: str, extra=None) -> str:
+    """Call `psql` and execute the given SQL statement."""
+    if extra is None:
+        extra = dict()
+    result = subprocess.run(
+        ["psql", "-h", "127.0.0.1", "-p", "5432", "-U", "postgres", "-d", "quickstart", "--quiet", "-c", sql],
```
joshua-spacetime, line 13: I think it's fine to allow `-U` to be optional, but can we require that it be an identity if provided at all?

crates/pg/src/pg_server.rs
mamcx: Downgrade automatically

crates/standalone/src/subcommands/start.rs
```diff
@@ -185,15 +185,16 @@ pub async fn exec(args: &ArgMatches, db_cores: JobCores) -> anyhow::Result<()> {
...
-        _ = axum::serve(tcp, service).with_graceful_shutdown(async move {
-            shutdown_rx.changed().await.ok();
+        _ = pg_server::start_pg(notify.clone(), ctx, tcp_pg) => {},
+        _ = axum::serve(tcp, service).with_graceful_shutdown(async move  {
+            shutdown_notify.notified().await;
         }) => {},
         _ = tokio::signal::ctrl_c() => {
             println!("Shutting down servers...");
-            let _ = shutdown_tx.send(()); // Notify all tasks
+            notify.notify_one(); // Notify all tasks
```
coolreader18:
```suggestion
            notify.notify_waiters(); // Notify all tasks
```

smoketests/tests/pg_wire.py
```diff
@@ -0,0 +1,293 @@
...
+------+-----------+---------------------+---------------------+------------------+--------------------------------------------------------------------+------------------------------------+---------------------------+----------
+ t    | 594806.56 | -3454353.3453890434 | This is spacetimedb | \\x01020304050607 | \\x0000000000000000000000000000000000000000000000000000000000000001 | \\x00000000000000000000000000000000 | 1970-01-01T00:00:00+00:00 | PT10S
+(1 row)""")
+        self.assertSql(token, "SELECT * FROM t_others_tuple", """\
+tuple
+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+ {"bool": true, "f32": 594806.56, "f64": -3454353.3453890434, "str": "This is spacetimedb", "bytes": 0x01020304050607, "identity": 0x0000000000000000000000000000000000000000000000000000000000000001, "connection_id": 0x00000000000000000000000000000000, "timestamp": 1970-01-01T00:00:00+00:00, "duration": PT10S}
```
joshua-spacetime: Are unquoted hex values valid JSON?

mamcx: Fixed now

joshua-spacetime (approved):

bfops (approved):
codeowned changes for me LGTM (`crates/cli/src/subcommands/sql.rs`)

cloutiertyler (approved):
For the two files that I own this looks good to me. I take it we are supporting postgres formatted output and also SpacetimeDB for `spacetime sql`?

mamcx:
> For the two files that I own this looks good to me. I take it we are supporting postgres formatted output and also SpacetimeDB for `spacetime sql`?

Correct.

PR #2705, opened 2025-05-06
Fix subscription manager metadata for joins

joshua-spacetime:
# Description of Changes

Fixes a bug where the subscription manager would not update its metadata for tables that did not have a filter in a join query.
This resulted in row updates not being sent to clients for those tables.

None

1

# Testing

- [x] Regression tests
- [x] Bot test

### DIFF
--- a/crates/core/src/subscription/module_subscription_manager.rs
+++ b/crates/core/src/subscription/module_subscription_manager.rs
@@ -1519,7 +1519,7 @@ mod tests {
     }
 
     #[test]
-    fn test_lookup_queries_for_search_arg() -> ResultTest<()> {
+    fn test_search_args_for_selects() -> ResultTest<()> {
         let db = TestDB::durable()?;
 
         let table_id = create_table(&db, "t")?;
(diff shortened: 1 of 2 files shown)


crates/core/src/subscription/module_subscription_manager.rs
```diff
@@ -629,15 +631,15 @@ impl SubscriptionManager {
...
                 search_args.insert_query(table_id, col_id, arg, hash);
             }
-            if !has_search_args {
-                for table_id in query_state.query.table_ids() {
-                    tables.entry(table_id).or_default().insert(hash);
-                }
+            // Update the `tables` map if this query reads from a table,
+            // but does not have a search argument for that table.
+            for table_id in table_ids {
+                tables.entry(table_id).or_default().insert(hash);
```
joshua-spacetime, line 642: Previously if a join query could be parameterized, I wasn't updating this mapping for the non-parameterized table. And this meant we would swallow all row updates for that table, for that query.

gefjon (approved):
Should bot test before merging, but code and tests look good.

jdetter (approved):
Verified working with bot test. Results posted in the internal discord here: https://discord.com/channels/931210784011321394/1150256030433955870/1369309233979330650

PR #2707, opened 2025-05-06
`PagePool::{default -> new_for_test}` + temporary hack for `IN_MEMORY_CONFIG` / `test_index_scans`

Centril:
# Description of Changes

First, `PagePool::default` is renamed to `PagePool::new_for_test` to make sure `::default()` isn't used anywhere.
This was useful when debugging the issue.

Second, `IN_MEMORY_CONFIG` in `crates/testing/src/modules.rs` is changed so that `page_pool_max_size: Some(size_of::<Page>())`, which is the minimum possible. This makes the failures in `test_index_scans` go away. On my i7-7700K/64GB the threshold for the test failing seems to be at `1 << 26` but it passes on `1 << 25`. As to why a page pool >= 64 MiB would cause such a slow down is unclear to me as of yet.

Third, the default page pool size is now `Some(size_of::<Page>())` which effectively disables page pooling.

None

1

# Testing

This fixes a test.

### DIFF
--- a/crates/table/src/page_pool.rs
+++ b/crates/table/src/page_pool.rs
@@ -24,15 +24,24 @@ impl MemoryUsage for PagePool {
 
 impl PagePool {
     pub fn new_for_test() -> Self {
-        Self::new(None)
+        Self::new(Some(100 * size_of::<Page>()))
     }
 
     /// Returns a new page pool with `max_size` bytes rounded down to the nearest multiple of 64 KiB.
     ///
-    /// if no size is provided, a default of 8 GiB is used.
+    /// if no size is provided, a default of 1 page is used.
     pub fn new(max_size: Option<usize>) -> Self {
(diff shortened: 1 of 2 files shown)


gefjon (commented):
- Needs companion PR to fix internal tests.
- Should we be using a smaller page pool everywhere, not just in tests? Do the page pool metrics have anything useful to tell us here?

The code changes look fine and I'm willing to approve this, but I'm not super excited to merge it without knowing whether using the same smaller pool everywhere would win us two orders of magnitude of performance somehow.

gefjon:
> the smoketests failing don't seem to have anything to do with this PR but look related to https://github.com/clockworklabs/SpacetimeDB/pull/2644?

I think these are related to #2714 , actually.

gefjon (approved):

PR #2710, opened 2025-05-06
Fix Equals and GetHashCode for types containing Lists and Arrays in C#

kazimuth:
# Description of Changes

Addresses https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/issues/313 .

We used to rely on List.Equals when comparing lists for equality. However, this is wrong: List.Equals uses reference equality in C#, not sequence equality. Similarly for GetHashCode, and for both methods on arrays.

I *wanted* to do something reflection-y to generate EqualityComparers for any List<T>s we need at runtime. However, I realized that with IL2CPP, this may be very slow, especially with nested lists. `Equals` and `GetHashCode` are now in the hot path of various SDK operations so I want them to be fast.

So, I bit the bullet and just started generating nested loops whenever I see an array or list in a type marked `[SpacetimeDB.Type]`. This is efficient and should generate fast code with IL2CPP. However, it's kind of zany, and wants careful review.

I also made our null checking more conservative and started checking for nulls even for reference types that aren't explicitly marked nullable.

N/A

2: this is fairly complicated generated code that needs to be correct and fast.

# Testing

- [x] Test suite
- [x] Added some more proptests that randomly generate types containing nested lists.
- [x] Blackholio
- [x] Bitcraft


crates/bindings-csharp/Codegen.Tests/fixtures/server/snapshots/Type#ContainsNestedLists.verified.cs
```diff
@@ -0,0 +1,708 @@
+﻿//HintName: ContainsNestedLists.cs
```
kazimuth, line 1: This file demands special review.

kazimuth: Sorry.

crates/bindings-csharp/BSATN.Codegen/Type.cs
```diff
@@ -6,34 +6,289 @@ namespace SpacetimeDB.Codegen;
...
+    /// <returns></returns>
+    public abstract string EqualsStatement(
+        string inVar1,
+        string inVar2,
+        string outVar,
+        int level = 0
+    );
+
+    /// <summary>
+    /// Get a statement that declares outVar and assigns assigns the hash code of inVar to it.
```
rekhoff, line 104:
```suggestion
    /// Get a statement that declares outVar and assigns the hash code of inVar to it.
```

rekhoff (approved):
Code is complex, but well documented. The snapshot examples help in seeing the resulting behavior. I also tested against the Blackholio repo and verified regenerated bindings behave correctly.
Approved!

PR #2712, opened 2025-05-07
Hash subscription queries before acquiring tx lock

joshua-spacetime:
# Description of Changes

Before this change we would hash subscription queries after acquiring the tx lock.

Hashing requires that we parse first so that we can determine if the query is parameterized with `:sender`. However our parser `sqlparser` is quite slow, resulting in `Subscribe` being 3x slower than `Unsubscribe` according to some benchmarks.

~~This patch doesn't change the parser at all. It just moves it so that it doesn't block other transactions.~~

This patch now generates query hashes without parsing. It does so by generating two hashes and seeing if either is currently tracked by the subscription manager. Hashes are generated outside of the tx lock.

None

1

# Testing

- [ ] Performance bot test


crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -452,6 +470,71 @@ impl ModuleSubscriptions {
...
+                    .into_iter()
+                    .map(Arc::new),
+            );
+        }
+
+        for (sql, hash) in query_hashes {
+            if let Some(unit) = guard.query(&hash) {
+                queries.push(unit);
+            } else {
+                let compiled = compile_read_only_query(&auth, &tx, &sql).map_err(|err| DBError::WithSql {
```
jsdt, line 527: This is parsing the sql again, even though we did it when hashing. It looks like this function could take `SqlSelect` instead of a string, which would let us reuse that work.

jsdt (commented):
This will save a lot of work for queries that already have subscribers, but it looks like we could avoid double-parsing for new queries with a little refactoring.

crates/core/src/subscription/module_subscription_actor.rs
joshua-spacetime: I had to rework this patch a bit. The reason we were parsing in the first place was to determine if the query was parameterized. But this actually requires view resolution for rls which requires a tx. Instead I've decided to generate two hashes outside of the tx lock.

crates/core/src/subscription/query.rs
```diff
@@ -70,19 +81,36 @@ pub fn compile_read_only_queryset(
...
+/// Compile a string into a single read-only query.
+/// This returns an error if the string has multiple queries or mutations.
+pub fn compile_query_with_hashes(
+    auth: &AuthCtx,
+    tx: &Tx,
+    input: &str,
+    hash: QueryHash,
+    hash_with_param: QueryHash,
+) -> Result<Plan, DBError> {
+    if WHITESPACE.is_match_at(input, 0) {
```
gefjon, line 103:
```suggestion
    if is_whitespace_or_empty(input) {
```

PR #2713, opened 2025-05-07
Update axum

coolreader18:
# Description of Changes

Axum now has what we need out of it for a websocket wrapper, so we no longer need to duplicate `util/flat_csv.rs` and `util/websocket.rs`, meaning we have less code to maintain.

2 - upgrading a dependency is a risk, but looking through the [changelog](https://github.com/tokio-rs/axum/blob/main/axum/CHANGELOG.md#080) there isn't anything that should affect us.

# Testing

- [x] tests pass 
- [ ]


cloutiertyler:
This works for me. Lints are failing.

PR #2714, opened 2025-05-08
Prevent `ahash` from pulling in `getrandom 0.3`

kim:
Modules generally don't honor the workspace cargo lock file, so may upgrade `ahash` silently, which pulls in a `getrandom` version that is not the one pulled in via `binding`. Thus feature unification doesn't apply and modules fail to compile due to the wasm target not being supported by the transitive `getrandom`.

This also causes smoketests to fail.

1

# Testing

Smoketests / module compilation fail before this change, and succeed after.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -43,22 +43,22 @@ version = "0.7.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "891477e0c6a8957309ee5c45a6368af3ae14bb510732d2684ffa19af310920f9"
 dependencies = [
- "getrandom 0.2.15",
+ "getrandom 0.2.16",
  "once_cell",
  "version_check",
 ]
 
 [[package]]
 name = "ahash"
-version = "0.8.11"
+version = "0.8.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e89da841a80418a9b391ebaea17f5c112ffaaa96f621d2c285b5174da76b9011"
+checksum = "5a15f179cd60c4584b8a8c596927aadc462e27f2ca70c04e0071964a73ba7a75"
 dependencies = [
  "cfg-if",
- "getrandom 0.2.15",
+ "getrandom 0.3.2",
  "once_cell",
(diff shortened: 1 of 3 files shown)


kim:
Conceivably, this can happen again if we have some other crate that transitively depends on `getrandom`. Maybe there is a more robust way to enforce a single version of `getrandom`?

coolreader18:
I think ideally in the future this kinda thing never happens again in the rust ecosystem, cause it's been a royal pain to deal with. But, also, those errors from getrandom 0.3 are actually exactly what we want - failing to compile when you pull in getrandom. So, this is the right fix.

gefjon:
> It was not quite unintentionally, because the actual problem here is that building a module won't honor the lock file

Can we (in a separate PR) add a flag to `spacetime build` and friends which makes it reflect the lockfile? Or possibly just a way to pass cargo flags through more generally?

kim:
> Can we (in a separate PR) add a flag to `spacetime build` and friends which makes it reflect the lockfile?

No I don't think: a module has `spacetimedb` in its dependencies, which is what pulled in the faulty dependency graph. I'm not aware of a way to force a dependency to lock its dependencies transitively.

What we might want to consider is to somehow make the tests that compile `modules/module-test` **not** honor the lockfile, so we'll catch this earlier and in a more obvious way. I'm not sure this can actually be done except for removing the lockfile.

gefjon:
Wait, what? Surely if you compile your module with `--locked`, that makes it respect the lockfile for the entire tree of dependencies including transitive deps, right?

kim:
If I create a fresh module, it won't have a `Cargo.lock` file and will go out to the internet to resolve the dependencies. It won't, for a particular dependency, get the lock file of _that_ dependency and consider it for version constraints.

Smoketests create fresh modules all the time, and they are built outside the workspace, so that's why they are failing.

gefjon:
Ah, I see. Yes, that's unfortunate.

kim:
Someone press the button then? :innocent:

gefjon (approved):

PR #2718, opened 2025-05-09
Reintroduce replication factor parameter

kim:
The `--num-replicas` option was removed from the CLI pre-1.0
It is re-introduced here as a hidden / unstable option to facilitate
testing (e.g. to override defaults).

The `DatabaseDef` type is changed to make `num_replicas` optional, so as
to distinguish "use the default" (`None`) from "shut down all replicas
but keep the database" (`Some(0)`).

- [x] no

1

# Testing

- [x] Will be used later in smoketests


Shubham8287 (approved):
Looks good to me.

Futher requesting review  from @jdetter or @bfops as its cli change

bfops (approved):
CLI changes LGTM!

PR #2722, opened 2025-05-09
Log instead of panicking when missing a row from st_client

gefjon:
# Description of Changes

We've observed this panic multiple times, somehow. It doesn't have to be a hard error, it's just surprising and not something we thought was possible. This PR drops it down to be an error log instead of a panic, and also adds the database identity to the log.
My hope is that, the next time we see this happen, we can find the replica dir and commitlog for the database with the inconsistent `st_table`, and use that to debug how we reached the unexpected state.

N/a

1.

# Testing

None. We totally don't know how to reproduce this (that's kind of the point).


joshua-spacetime (approved):

gefjon:
Internal tests failure appears unrelated and lockfile-based. Merging.

PR #2725, opened 2025-05-09
Reduce reflection use in generated C# code

### LINKED ISSUE #2669: Avoid reflection in IStructuralReadWrite.Read
That is, fix this TODO:
https://github.com/clockworklabs/SpacetimeDB/blob/59faab8a1ff0cfc966ed708169d93328d19eca9c/crates/bindings-csharp/BSATN.Runtime/BSATN/Runtime.cs#L14-L16

Currently each row object is instantiated using reflection during deserialization which represents a significant portion of the total cost as the following profile snippet demonstrates.

Image

kazimuth:
# Description of Changes

Directly allocates results for some operations and avoids the usage of reflection for enum serialization.
This should hopefully speed up (de)serialization to some extent.
Needs to be paired with an upcoming C# SDK PR to speed these up further.

Adds a method to IStructuralReadWrite, which will break anyone who manually implements it. However, nobody should be manually implementing it. I've checked, and Bitcraft doesn't.

0

# Testing

- [ ] Blackholio
- [ ] Bitcraft

### DIFF
--- a/crates/bindings-csharp/BSATN.Codegen/Type.cs
+++ b/crates/bindings-csharp/BSATN.Codegen/Type.cs
@@ -422,22 +422,12 @@ public abstract record BaseTypeDeclaration<M>
 
         var bsatnDecls = Members.Cast<MemberDeclaration>();
         var fieldNames = bsatnDecls.Select(m => m.Name);
+        var fieldNamesAndIds = fieldNames.Select((name, i) => (name, i));
 
         extensions.BaseTypes.Add($"System.IEquatable<{ShortName}>");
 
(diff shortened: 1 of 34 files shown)


kazimuth:
Git mistake.

rekhoff (approved):
Code changes made a lot of sense to me and look clear. I also apricate the additional summary text on the existing code. As long as these get tested against Blackholio and Bit Craft, then I think this looks good.

kazimuth:
I've tested against Blackholio and Bitcraft, happy with this.

PR #2733, opened 2025-05-14
commitlog: Include latest commit offset in segment metadata

kim:
# Description of Changes

The segment metadata, obtained e.g. via `committed_meta`, will now include the start offset of the last commit read from the segment.

No 

1

# Testing

- [x] yes

### DIFF
--- a/crates/commitlog/src/repo/mod.rs
+++ b/crates/commitlog/src/repo/mod.rs
@@ -243,6 +243,7 @@ pub fn resume_segment_writer<R: Repo>(
         tx_range,
         size_in_bytes,
         max_epoch,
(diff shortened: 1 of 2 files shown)


Shubham8287 (commented):
It looks correct, but I am missing intention behind this patch?
Is it to serve `accepted_offset` on fetch-api?

kim:
I needed to know the head commit's range from only the commitlog. Since the information is already there, it seemed convenient to expose it.

kim:
The motivation is more like determining an accepted offset for bootstrapping.

Shubham8287 (approved):

PR #2734, opened 2025-05-14
Cache `with_label_values` more & don't do metrics while holding lock

Centril:
# Description of Changes

Two things happen here that work together to improve performance:

- Calls to `with_label_values` are cached to the extent possible.
- No metrics are reported while a transaction lock, read-only or mutable, is held.

Changes to `traits.rs` are:
- Transaction terminal functions (methods consuming transactions, like committing) now also return `TxMetrics` and sometimes also the reducer name (`String`) from the `ExecutionContext`.
- `TxData` now exposes which tables were affected.

Closes https://github.com/clockworklabs/SpacetimeDB/pull/1099 which was the first attempt at not reporting while holding the lock. The PR has since bitrotted.

None

3, most of the complication is where datastore transactionality is touched.

# Testing

### DIFF
--- a/crates/core/src/db/datastore/locking_tx_datastore/datastore.rs
+++ b/crates/core/src/db/datastore/locking_tx_datastore/datastore.rs
@@ -749,7 +749,8 @@ impl TxMetrics {
         }
     }
 
-    fn report<'a>(
+    /// Reports the metrics for `reducer` using `get_exec_counter` to retrieve the metrics counters.
+    pub fn report<'a>(
         &self,
         tx_data: Option<&TxData>,
         reducer: &str,
@@ -843,15 +844,16 @@ impl TxMetrics {
         }
     }
 
+    /// Reports the metrics for `reducer`, using counters provided by `db`.
     pub(crate) fn report_with_db(&self, reducer: &str, db: &RelationalDB, tx_data: Option<&TxData>) {
         self.report(tx_data, reducer, |wl| db.exec_counters_for(wl));
     }
 }
 
-/// Report the `TxMetrics`s passed.
+/// Reports the `TxMetrics`s passed.
 ///
 /// Should only be called after the tx lock has been fully released.
-pub(crate) fn report_tx_metricses(
+pub fn report_tx_metricses(
     reducer: &str,
     db: &RelationalDB,
     tx_data: Option<&TxData>,
--- a/crates/core/src/subscription/mod.rs
+++ b/crates/core/src/subscription/mod.rs
@@ -38,7 +38,7 @@ pub struct ExecutionCounters {
 }
 
 impl ExecutionCounters {
-    pub(crate) fn new(workload: &WorkloadType, db: &Identity) -> Self {
+    pub fn new(workload: &WorkloadType, db: &Identity) -> Self {
         Self {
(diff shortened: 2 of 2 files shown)


gefjon (approved):

crates/core/src/db/datastore/traits.rs
```diff
@@ -297,16 +311,49 @@ pub trait DataRow: Send + Sync {
 pub trait Tx {
     type Tx;
 
+    /// Begins a read-only transaction under the given `workload`.
+    ///
+    /// While this transaction is pending,
+    /// other read-only transactions may be started,
+    /// but new mutable transactions will block until there are no read-only transactions left.
+    ///
+    /// Blocks if a mutable transaction is pending.
```
cloutiertyler, line 320: NIT: This doc comment is only applicable to the `LockingTxDatastore`, not to the trait and read transactions in general. I would rather this comment be moved to the appropriate place.

crates/core/src/db/datastore/traits.rs
```diff
@@ -297,16 +311,49 @@ pub trait DataRow: Send + Sync {
...
     fn begin_tx(&self, workload: Workload) -> Self::Tx;
-    fn release_tx(&self, tx: Self::Tx);
+
+    /// Release this read-only transaction,
+    /// allowing new mutable transactions to start if this was the last read-only transaction.
```
cloutiertyler, line 324: NIT: likewise on the above

crates/core/src/db/datastore/traits.rs
```diff
@@ -297,16 +311,49 @@ pub trait DataRow: Send + Sync {
...
 
 pub trait MutTx {
     type MutTx;
 
+    /// Begins a mutable transaction under the given `isolation_level` and `workload`.
+    ///
+    /// While this transaction is pending,
+    /// no other read-only  mutable transaction may happen or
+    ///
+    /// Blocks if a read-only or mutable transaction is pending.
```
cloutiertyler, line 340: Likewise

crates/core/src/host/instance_env.rs
```diff
@@ -543,29 +539,25 @@ mod test {
...
             },
-            replica_id: 0,
-            logger: Arc::new(temp_logger()?),
-            subscriptions: subscription_actor(relational_db.clone()),
+            0,
+            Arc::new(temp_logger()?),
+            subscription_actor(relational_db.clone()),
             relational_db,
-        })
+        ))
```
cloutiertyler, line 560: NIT: Seems like a change for the worse to me. Now the `replica_id` is a magic number `0`.

crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -343,28 +341,23 @@ impl ModuleSubscriptions {
...
+            send_err_msg,
+            None
+        );
+        // This is technically a bug, since this could be empty if the client has another duplicate subscription.
+        // This whole function should be removed soon, so I don't think we need to fix it.
+        let [query] = &*queries else {
+            // Apparently we ignore errors sending messages.
+            let _ = send_err_msg("Internal error".into());
+            return Ok(None);
+        };
```
cloutiertyler, line 356: This little refactor seems a far cry from `with_label_values` I must say

cloutiertyler (approved):
Approved with comments.

PR #2735, opened 2025-05-14
Handle snapshot restore more robustly

kim:
Snapshots may be corrupted on disk or incomplete where the lock file is gone,
but the snapshot file was not written (it is written last after all objects).

This patch handles both cases by ignoring snapshots without a snapshot file when
listing the available snapshots, and trying older snapshots if the latest one
cannot be restored from (for any reason).

It also takes into account that the prefix of the local commitlog could have
been archived.

No

2

# Testing

- [x] added

### DIFF
--- a/crates/core/src/db/relational_db.rs
+++ b/crates/core/src/db/relational_db.rs
@@ -532,7 +532,7 @@ impl RelationalDB {
                     break;
                 }
                 if let Ok(datastore) =
-                    try_restore_napshot(snapshot_repo, snapshot_offset, database_identity, page_pool.clone())
+                    try_restore_snapshot(snapshot_repo, snapshot_offset, database_identity, page_pool.clone())
                 {
                     return Ok(datastore);
                 } else {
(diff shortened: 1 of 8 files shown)


crates/core/src/db/relational_db.rs
```diff
@@ -461,43 +463,104 @@ impl RelationalDB {
...
+                        .latest_snapshot_older_than(upper_bound)
+                        .map_err(Box::new)?
+                    else {
+                        break;
+                    };
+                    if min_commitlog_offset + 1 > snapshot_offset {
+                        break;
+                    }
+                    if let Ok(datastore) =
+                        try_restore_napshot(snapshot_repo, snapshot_offset, database_identity, page_pool.clone())
```
Shubham8287, line 537:
```suggestion
                        try_restore_snapshot(snapshot_repo, snapshot_offset, database_identity, page_pool.clone())
```

Shubham8287 (approved):
Thanks!

PR #2737, opened 2025-05-14
Call reducers on the blocking pool.

jsdt:
# Description of Changes

This moves reducer calls to the blocking thread pool. This is important because they wait for a write lock (which we don't want to do in the main worker pool).

We used to do this, but it was accidentally changed in https://github.com/clockworklabs/SpacetimeDB/pull/2550, and when it was initially reverted (https://github.com/clockworklabs/SpacetimeDB/pull/2624), it was switched to use a normal task.

1. The only risk is a performance regression, which is unlikely.


kim (approved):

coolreader18:
Ah, that would certainly do it.

PR #2740, opened 2025-05-15
Follow-up: use asyncify in ModuleHost::call

coolreader18:
# Description of Changes

Follow-up to #2737. I added this utility function in #2730, and it makes the logic of these kinds of `spawn_blocking` uses clearer.

1

# Testing

n/a - the code is equivalent.
- [ ] 
- [ ]


coolreader18:
I also edited the comment - it's obviously not that we care about having to use `std::panic::AssertUnwindSafe`, it's that the operation is blocking.

jsdt (approved):

PR #2741, opened 2025-05-15
Add counters for client connections spawned and aborted.

jsdt:
# Description of Changes

This adds two new metrics to track how often websocket clients connect and disconnect (`spacetime_worker_ws_clients_spawned`, and `spacetime_worker_ws_clients_aborted`).

Note that this isn't tracking at the tcp connection level, rather it only counts clients that have gone through any client lifecycle reducers.

Adding a metric for rejected clients (meaning the on_connect reducer failed) in the future could be useful.

This is purely a metrics change.

1.


cloutiertyler (approved):
I'm approving this because we'd like to get it into Ryan's new baseline test for the technical test. The risk is very low as it is only metrics.

PR #2747, opened 2025-05-16
Fix TypeScript imports codegen

toqueteos:
# Description of Changes

The [TypeScript Quickstart](https://spacetimedb.com/docs/sdks/typescript/quickstart) does not work out of the box because of imports using mixed types and non-types (classes and functions, not sure if enums also count) in the same line.

This change fixes that.

Huge thanks to this [Discord message](https://discord.com/channels/1037340874172014652/1370368861509124166/1370758371933950046).

None

1

# Testing

N/A

---

There's probably a nicer way of fixing this long-term but I'm not aware of it.

I also sent this other PR https://github.com/clockworklabs/spacetime-docs/pull/349 which fixes the remaining code that makes TypeScript Quickstart not work.


jsdt (approved):
Thanks for adding this!

FYI, I pushed a change to update the code snapshots in our unit tests, so this should be good to go.

toqueteos:
Thanks @jsdt ! There's also a change in docs related to this PR in https://github.com/clockworklabs/spacetime-docs/pull/349

PR #2748, opened 2025-05-16
Reject clients who match an existing `st_client` row

gefjon:
# Description of Changes

At various places in our codebase, we assume that client connections are unique by `(identity, connection_id)` pair,
but we never actually enforced this prior to now.
It was difficult, but not impossible, to accidentally connect with an already-present `(identity, connection_id)` pair. The easiest way was to open two `DbConnection`s in parallel from within a single Rust client SDK process,
due to a misbehavior in that SDK which I intend to fix in a separate PR.

This commit modifies `MutTxId::insert_st_client` to return an error if the row to be inserted is already resident in the database. It's still possible for a database owner to delete from `st_client` manually via a SQL `delete` statement, which will cause strange misbehaviors, but it should no longer be possible for an unprivileged client to accidentally put the database into a state that's intended to be impossible.

N/a

1

# Testing

- [x] Manually tested with [a modified `quickstart-chat` client](https://github.com/user-attachments/files/20257695/quickstart-chat-two-connections.patch.txt).
  - Prior to this commit, the host accepts both connections, sends responses only to the first one, and shows the error log introduced by #2722 when the second disconnects. (Or maybe when the first disconnects? Timing unclear.)
  - With this commit, the host rejects the second connection, and so none of the misbehavior is reachable.
- [ ] Write automated tests somehow.

### DIFF
--- a/crates/core/src/db/datastore/locking_tx_datastore/mut_tx.rs
+++ b/crates/core/src/db/datastore/locking_tx_datastore/mut_tx.rs
@@ -1236,12 +1236,29 @@ impl<'a> Iterator for IndexScanFilterDeleted<'a> {
 }
 
 impl MutTxId {
-    pub(crate) fn insert_st_client(&mut self, identity: Identity, connection_id: ConnectionId) -> Result<()> {
+    /// Insert a new row into `st_client` for the client `client_identity, connection_id`.
+    ///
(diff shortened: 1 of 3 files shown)


gefjon:
Marking as draft because I still need to write an automated test for this, but I'm putting it on the back burner as higher priority tickets have surfaced.

PR #2749, opened 2025-05-16
Bump rust-toolchain to rust 1.87

coolreader18:
# Description of Changes

This bumps our rust-toolchain in the file to 1.87, but doesn't bump the MSRV of the bindings crate and its dependents. We could, if we wanted to, but we don't really have an MSRV policy in place. This is necessary because the `v8` crate has a fairly aggressive MSRV and I'd like to not fall too far behind from their latest version, since that's also the version of V8 itself.

2 - rust version changes can cause problems but it's very rare

# Testing

- [ ] 
- [ ]

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -183,7 +183,7 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 20 # on a successful run, runs in 8 minutes
     container:
-      image: rust:1.83.0
+      image: rust:1.87.0
       options: --privileged
     # filter for a comment containing 'benchmarks please'
     if: ${{ github.event_name != 'issue_comment' || (github.event.issue.pull_request && contains(github.event.comment.body, 'benchmarks please')) }}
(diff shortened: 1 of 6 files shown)


gefjon:
If we're going to maintain a lower MSRV than we use for development, I'd like to ensure we have CI which runs with that version.

Centril:
I've reviewed @coolreader18's changes. They look good.
I've then changed the Rust version to 1.88 so that we also get `let_chains` and *cargo automatic cache cleaning*.
Finally, I've rebased this PR atop of `master`.
It should be ready for review now.
I'll investigate and fix the internal tests meanwhile.

Centril:
@cloutiertyler This is ready for review. Private companion is https://github.com/clockworklabs/SpacetimeDBPrivate/pull/1847.

cloutiertyler:
@Centril I don't believe this is a priority

Centril:
@cloutiertyler This is a requirement for V8 modules because the `v8` dependency uses edition 2024 which does not exist on our version of Rust.

cloutiertyler (approved):
@Centril Thanks for the clarification!

PR #2752, opened 2025-05-17
Save server specific login tokens in server config

aceysmith:
# Description of Changes

These changes only impact the `spacetime` CLI.

Currently, if a user is explicitly logging into a server using the `spacetime login --server-issued-login <server name>` or less directly by indicating `N` to the `You are not logged in. Would you like to log in with spacetimedb.com?` prompt, a JWT is created that is only valid for that specific server. Since the current `cli.toml` only supports a single `spacetimedb-token` at the root of the config, this quickly leads to collisions and attempts to `spacetime publish` or performance other actions with tokens that are invalid if you are using multiple servers.

This change will store a `spacetimedb-token` value in the associated server config section in the `cli.toml` file. Here are the impacts to specific commands:

**`spacetime login --server-issued-login <server name>`**

This command will store the generated JWT in the `cli.toml` section associated with that server. If a token is present already, it will give the existing `You are already logged in.` message, ignoring the `spacetimedb-token` value at the root config if present.

**`spacetime login show --server`**

A new flag has been introduced to this command to display the identity of the token stored for the provided server. If the server name is provided, the command will additionally display the name in the output. Example:
```
You are logged in to server "local" as c200a6e078b49844fefc3c060ac6d467eb1a122fc8533f7c65639fe34b048e8e
```

**`spacetime logout`**

This command will additionally delete all server `spacetimedb-token` values stored. This might be a bit harsh, but the intention of `spacetime logout` before was clearly remove all tokens and authentication, so this seemed most safe. Alternatively, we could only remove the root tokens and leave the server specific tokens intact and introduce a `--server` flag to log out of the specific server, but this seemed more consistent.

**`spacetime publish` / `spacetime sql` / `spacetime logs`/ etc.**

Any command that might have a path to force the user to log in to the server directly as a result of https://github.com/clockworklabs/SpacetimeDB/pull/2158 has the possibility of creating a JWT that will be stored as part of the provided (or default) server config if the user indicates they want to log in directly.

Unlike the above mentioned `spacetime login show` command, these will fall back to the root `spacetimedb-token` value. This is to handle backwards compatibility for users who have managed localhost JWTs already and would expect that functionality to persist, as well as to support centrally authoritative tokens like those generated by auth.spacetimedb.com or custom auth providers.

**`spacetime server edit`**

This command is not impacted by this change. If you have a `spacetimedb-token` value associated with the server, it will get carried along if present. 

None

2

# Testing

```
spacetime login --server-issued-login local
# verify a new local JWT is created and stored in the `local` server config within the `cli.toml`

spacetime login --server-issued-login local
# verify you receive the `You are already logged in.` message

spacetime login show -s local
# verify the `You are logged in to server "local" as ...` message is shown

spacetime logs -s local <module name>
# This should successfully display logs

spacetime logout
# verify ALL JWTs are removed from `cli.toml`
```


aceysmith:
I did see notes in the code about separating out tokens into its own credentials file or similar, but that seemed out of scope and a bit more complex. Hope folks find some value here :-) A nice side effect of this change is if you want to store/save different tokens for different servers, you can now do that while also allowing web login to create a root token that can be used for other servers.

cloutiertyler:
@aceysmith It sounds like this is allowing you to log into multiple identities simultaneously. Is that correct? I think that could have unexpected behavior for users? Which identity is used if you are logged in via SpacetimeAuth and also via the server issued identity?

aceysmith:
> @aceysmith It sounds like this is allowing you to log into multiple identities simultaneously. Is that correct?

Correct.

> I think that could have unexpected behavior for users? Which identity is used if you are logged in via SpacetimeAuth and also via the server issued identity?

The server issued identity. The `spacetimedb_token` stored in that specific server config will be set if and only if logging into that server "directly." If no value is present for that server, the root `spacetimedb_token` will be used. All future interactions with that server will use the server specific token.

I noted some options about behavior changes to `spacetime logout`, but one option I didn't suggest is to add a `--server-issued-login` flag to clear only the server generated token for the provided server, in case you want to drop that token and swap over to Spacetime Auth. Moving between the two doesn't seem very common and removing the token from the cli.toml manually is fairly trivial, but if we have to go modify `toml` files by hand when we have a CLI, we're no better than the apes.

cloutiertyler:
@aceysmith Unfortunately, I think this is moving away from the direction we'd like to go in. We would like users to have a single account for all servers and to be able to log into one account at a time. The server issued login is using SpacetimeDB as an OIDC IdP, which it is not really designed to be. The reason we keep that functionality is just a way for people to get started quickly, but SpacetimeDB is not equipped to act as an IdP. The tokens issued with `--server-issued-login` do not expire, cannot be refreshed, and are not technically valid OIDC tokens. We recommend that users host their own IdP, use SpacetimeAuth, or use ones like Auth0 or Clerk. For that reason, I'd like to close this PR, but perhaps we should improve our documentation on this. Can you tell me more about your use case? What are you trying to achieve?

aceysmith:
@cloutiertyler Understood. I’ve been able to get by alright in the ~year since I’ve filed this alright.

The key thing this addressed was to giving another person the ability to publish to a self hosted server, but not prevent them from doing development on their own local instance or logging into spacetime auth for publishing to maincloud without constantly clobbering the stored token. We did that by sharing the server issued token and having it stored on a specific server config.

With maincloud having collaborators this is solved, but no great option I know of exists for self hosted. Having a separate config file and pointing to that during publish works okay for the same use case. It’s probably better with credential sharing apps to point to a synced file than to pull out a token anyway.

PR #2754, opened 2025-05-19
Add `client_connection_incoming_queue_length` metric

gefjon:
# Description of Changes

It's a metric that tracks the size of the incoming per-client `message_queue`. We've been concerned about not having visibility into this queue's length, as currently we only have the length of the per-database reducer queue, but each client can only have a single reducer in that queue, and may have additional messages waiting in its per-client queue.

The new metric, `spacetime_client_connection_incoming_queue_length`, is an `IntGaugeVec` with the labels:
`db: Identity, client_identity: Identity, connection_id: ConnectionId`. My theory is that in our viewer we can inspect the average and the sum per database, and it also may be interesting to be able to look at individual clients, as e.g. the BitCraft mob monitor may be a notable outlier. This is not the same pattern as most of our other metrics, though, which tend to only offer per-database granularity. It's possible that this new metric should also have the last two labels removed, and be labeled only on `db: Identity`.

N/a, as we don't (AFAIK) offer stability guarantees for our metrics.

2ish: see above question about granularity.

# Testing

I am not sure how to test this.

- [ ] 
- [ ]

### DIFF
--- a/crates/core/src/worker_metrics/mod.rs
+++ b/crates/core/src/worker_metrics/mod.rs
@@ -257,6 +257,11 @@ metrics_group!(
         #[help = "The cumulative number of bytes sent to clients"]
         #[labels(txn_type: WorkloadType, db: Identity)]
         pub bytes_sent_to_clients: IntCounterVec,
+
+        #[name = spacetime_client_connection_incoming_queue_length]
+        #[help = "The number of client -> server WebSocket messages waiting in a client connection's incoming queue"]
+        #[labels(db: Identity, client_identity: Identity, connection_id: ConnectionId)]
+        pub client_connection_incoming_queue_length: IntGaugeVec,
     }
 );
 
--- a/crates/client-api/src/routes/subscribe.rs
+++ b/crates/client-api/src/routes/subscribe.rs
@@ -228,6 +228,21 @@ async fn ws_client_actor_inner(
 
     let addr = client.module.info().database_identity;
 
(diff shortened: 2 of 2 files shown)


joshua-spacetime (approved):
These are indeed the labels that we want for right now.

crates/core/src/worker_metrics/mod.rs
```diff
@@ -257,6 +257,11 @@ metrics_group!(
         #[help = "The cumulative number of bytes sent to clients"]
         #[labels(txn_type: WorkloadType, db: Identity)]
         pub bytes_sent_to_clients: IntCounterVec,
+
+        #[name = spacetime_client_connection_incoming_queue_length]
+        #[help = "The number of client -> server WebSocket messages waiting in a client connection's incoming queue"]
+        #[labels(db: Identity, client_identity: Identity, connection_id: ConnectionId)]
```
jsdt, line 263: I think we should avoid adding any metrics with per-client or per-connection labels. I'm not sure how big the impact would be on our metrics collection/storage/querying performance, but the cardinality of a stat like this can blow up very quickly.

gefjon: As discussed out-of-band, I'll alter this to be a per-database metric, which I suppose will be the sum of the length of all clients' queues. I expect to open a separate PR for that, since we want to use this PR in testing.

PR #2759, opened 2025-05-20
cli: Add value_parser for `num-replicas`

kim:
Forgot to do this in #2718

1

# Testing

- [x] I actually ran the command using the option :)


bfops (approved):
Since it's tested, this LGTM!

PR #2762, opened 2025-05-20
Improve serialization speed of enums without data in C#

kazimuth:
# Description of Changes

I was looking at a bitcraft profile and it was still spending a lot of time in reflection when deserializing enums without data -- plain-old-enums. Turns out the code I thought was dead was not. So, I changed it to be as fast as possible (indexing into an array!) This should save a few more allocations and some deserialization time. I also changed the generated .Equals for `[SpacetimeDB.Type]`s containing enums to avoid an allocation.

0

# Testing
I am planning to test:

- [x] Bitcraft
- [x] Blackholio


rekhoff (approved):
Code, summary, comments, and tests all look good. I have not tested to verify a meaningful speed improvement but changes seem reasonable. Approved.

kazimuth:
I've run through the Bitcraft tutorial and am happy with this, ready to merge.

Issue #2763, opened 2025-05-20
CLI - `rename` should remove the old name


bfops:
The current `spacetime rename` command just adds a new name for the DB, but doesn't remove the old one.

@cloutiertyler says:
> There is an API to actually change the name
> It's new-ish, but it is in use in the website

So we should use that in the CLI as well.

cloutiertyler:
https://github.com/clockworklabs/spacetime-web/blob/6dca42defe49f1aa36f9494bd59d17f1a4ec98db/spacetimedb.com/app/.server/api/proxy.ts#L182

This is the call on the server

jdetter:
https://github.com/clockworklabs/SpacetimeDB/pull/2764

PR #2764, opened 2025-05-20
Fixed rename by using put instead of post

cloutiertyler:
# Description of Changes

Closes https://github.com/clockworklabs/SpacetimeDB/issues/2763

Technically a breaking change because it changes the behavior of the CLI, but it's arguably a bug fix.

2 because of the breaking nature

# Testing

- [ ] Test that you can rename a database (I have not done this yet)

### DIFF
--- a/crates/cli/src/subcommands/dns.rs
+++ b/crates/cli/src/subcommands/dns.rs
@@ -37,11 +37,11 @@ pub async fn exec(mut config: Config, args: &ArgMatches) -> Result<(), anyhow::E
     let domain: DomainName = domain.parse()?;
 
     let builder = reqwest::Client::new()
-        .post(format!(
+        .put(format!(
             "{}/v1/database/{database_identity}/names",
             config.get_host_url(server)?
         ))
-        .body(String::from(domain));
+        .body(format!("[{}]", String::from(domain)));
     let builder = add_auth_header_opt(builder, &auth_header);
 
     let result = builder.send().await?.json_or_error().await?;


bfops:
@cloutiertyler I updated the code meaningfully, as well as augmenting the smoketest. The new endpoint just returns a string that's either `"Success"` or an error message, so I've removed all the code that tries to parse a well-typed error out of the response.

crates/cli/src/subcommands/dns.rs
```diff
@@ -30,70 +30,27 @@ pub async fn exec(mut config: Config, args: &ArgMatches) -> Result<(), anyhow::E
...
-        }
-        InsertDomainResult::OtherError(e) => return Err(anyhow::anyhow!(e)),
+    let response = builder.send().await?;
+    let status = response.status();
+    let response_body = response.text().await?;
+    if !status.is_success() {
+        anyhow::bail!("{response_body}");
     }
 
+    println!("Domain set to {} for identity {}.", domain, database_identity);
```
cloutiertyler, line 53: We should probably just say name here. We pulled out support for TLDs so domain is confusing.

bfops:
(status here: the smoketests have uncovered that the `PUT` version of the request will happily clobber existing names without emitting an error)

bfops:
To clarify, you cannot clobber a database that you do not own:
```
$ cargo run -pspacetimedb-cli -- rename --to rename-test-module rename-from-module
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.35s
     Running `target/debug/spacetimedb-cli rename --to rename-test-module rename-from-module`
Error: Error: {"PermissionDenied":{"domain":"rename-test-module"}}
```

PR #2771, opened 2025-05-22
Commitlog: handle empty offset index lookup

Shubham8287:
# Description of Changes
fix: Offset index, key lookup returns `(0,0)` key and byte offset pair if it is empty.
fix:`seek_to_offset` was returning `Ok` for key 0. when asked to validate 0th commit at 0 byte offset.
Resulting `reset_to_internal` to miscalculate commit offset and truncating in middle of first commit. Its calculation was shorter by `segment::HEADER::len`

# Testing

- [ ] 
- [ ]

### DIFF
--- a/crates/commitlog/src/index/indexfile.rs
+++ b/crates/commitlog/src/index/indexfile.rs
@@ -115,6 +115,10 @@ impl<Key: Into<u64> + From<u64>> IndexFileMut<Key> {
         if low == 0 && key < low_key {
             return Err(IndexError::KeyNotFound);
         }
+        // If found key is 0, return `KeyNotFound`
+        if low_key == 0 {
+            return Err(IndexError::KeyNotFound);
+        }
 
         Ok((Key::from(low_key), low as u64))
     }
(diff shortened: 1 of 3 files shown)


crates/commitlog/src/index/indexfile.rs
```diff
@@ -391,6 +395,9 @@ mod tests {
     }
 
     fn assert_key_lookup(index: &impl KeyLookup<Key = u64>) -> Result<(), IndexError> {
+        // looking for key smaller than first entry should fail
+        assert!(index.key_lookup(1).is_err());
```
kim, line 399: This same assertion already existed a few lines down.

crates/commitlog/src/index/indexfile.rs
```diff
@@ -406,6 +413,15 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn test_emptry_index_lookup_should_fail() -> Result<(), IndexError> {
+        let index = create_index_in(TempDir::new().unwrap().path(), 100)?;
+        assert!(index.key_lookup(0).is_err());
```
kim, line 419: Use `assert_matches`?

kim (approved):

PR #2773, opened 2025-05-22
Per-database incoming and outgoing queue length metrics

gefjon:
# Description of Changes

Add two new metrics, `spacetime_total_incoming_queue_length` and `spacetime_total_outgoing_queue_length`. These are similar to the metrics added in #2754 , except that these are per-database, not per-client. This will reduce cardinality at the expense of less granular reporting.

Because the metrics are per-database rather than per-client, there's some tricky code related to cleaning up when a client disconnects.

N/a

1 - new metrics with low cardinality and infrequent `with_label_values` calls.

# Testing

None - I am not sure how to test metrics.


gefjon:
Do I need to `remove_label_values` when the database is deleted somehow? Where would I do that?

PR #2776, opened 2025-05-22
Cheaper `Table::clone_structure`

Centril:
# Description of Changes

This PR:
- Makes  `Table::clone_structure` cheaper by...
- ...`Arc`ing `RowTypeLayout`
- ..."small-veccing" `StaticLayout`
- and also clarifies `Table::clone_structure`.

Why? Because we saw `clone_structure` in flamegraphs.

None

2 -- fairly simple and isolated code that is already well tested

# Testing

### DIFF
--- a/crates/table/src/layout.rs
+++ b/crates/table/src/layout.rs
@@ -202,6 +202,9 @@ pub struct RowTypeLayout {
     ///
     /// This is `Arc`ed at the row type level
     /// as clones and drops of this showed up in flamegraphs.
+    /// A [`RowTypeLayout`] will typically be cloned once per table per transaction,
+    /// assuming the table is touched during that transaction.
+    /// This can be expensive for modules that have a lot of reducer calls per second.
     pub elements: Arc<[ProductTypeElementLayout]>,
 }
 


crates/table/src/layout.rs
```diff
@@ -194,50 +195,62 @@ pub const fn row_size_for_type<T>() -> Size {
...
 /// This type ensures that the minimum row size is adhered to.
 #[derive(Debug, PartialEq, Eq, Clone)]
-pub struct RowTypeLayout(ProductTypeLayout);
+pub struct RowTypeLayout {
+    /// The memoized layout of the product type.
+    pub layout: Layout,
+    /// The fields of the product type with their own layout annotations.
+    ///
+    /// This is `Arc`ed at the row type level
+    /// as clones and drops of this showed up in flamegraphs.
```
joshua-spacetime, line 204: Could you also add why it's potentially expensive to clone? Like how hot the path can be? Something like: "RowTypeLayout is cloned once per X which can be very expensive for patterns like Y"

gefjon (approved):

PR #2785, opened 2025-05-23
Run reducers on tokio blocking thread instead of rayon

joshua-spacetime:
# Description of Changes

Instead of dispatching reducer calls to rayon, we run them on the current tokio blocking thread.

None

1

A one line change. But we'll have to run some performance tests to see if it reduces overhead.

# Testing

- [x] Bot test


joshua-spacetime:
Screenshot 2025-05-23 at 1 50 17 PM

As expected, reducers are now run on the caller's tokio blocking thread. No dispatch overhead.

cloutiertyler (approved):

PR #2793, opened 2025-05-26
Aggregate and broadcast `DbUpdate`s off the main thread

gefjon:
# Description of Changes

and after releasing the datastore lock.

This commit moves some amount of work off the main `eval_updates_sequential` task and into a new background worker, the `send_worker`. The actual work performed is unchanged, it's just that some of it is moved to a different function which runs asynchronously after releasing the datastore lock. We expect two benefits from this:

1. Most obviously, we're now doing less work while holding the lock, reducing the size of our critical section.
2. We suspect that we had a thundering herd problem where we wake each client's WebSocket worker just before waking the next reducer task, making it wait in the Tokio task queue until all the client messages have been sent. This change should cause the reducer task to wake first, improving transaction throughput.

Moving the aggregation and broadcast to a separate task required enclosing the `SubscriptionManager`'s client map in an `Arc<RwLock>`. There's some amount of tricky code in there, with comments, that has to go out of its way to avoid deadlocks.

I also chose to change the `table_name` to be an `Arc<str>` rather than a `Box<str>`, as it needed to be cloned from `eval_updates_sequential` into the `send_worker`, where previously it was just borrowed within `eval_updates_sequential`. We could further reduce cloning by threading the change from `Box` to `Arc` further through our code, but I chose to keep the change relatively contained within the subscription-related code.

Additionally, while reworking the `SubscriptionManger`'s code, I changed from using tuples to structs in a few places. This leaked out a bit, and I changed some methods which used to take `(args: (T, U))` to instead take  `arg_t: T, arg_u: U`.
The only callers outside of `module_subscription_manager.rs` (the ones I was changing anyways) were constructing the tuple in the call anyways, and so now look cleaner.

Many tests are changed to now require a Tokio runtime, so that the `SubscriptionManager` can `tokio::spawn` its `Self::send_worker` future. Some tests which previously had a single-threaded runtime now require a multi-threaded one. This is mildly painful.

A new metric, `subscription_send_queue_length`, is also added. It is incremented by `eval_updates_sequential` before pushing into the new queue, and decremented by the `send_worker` after popping.

Because the metric is labeled with `database_identity: Identity`, the `SubscriptionManager` now needs to know its `Identity`, and so cannot be `Default`. Methods are added to make constructing one in tests easier.

N/a

3, some risk of deadlock due to added synchronization with the client map.

# Testing

- [x] I believe automated testing is sufficient to validate semantics.
- [x] Would like a bot test to observe hopefully-improved performance.

### DIFF
--- a/crates/core/src/host/instance_env.rs
+++ b/crates/core/src/host/instance_env.rs
@@ -534,6 +534,9 @@ mod test {
 
     /// An `InstanceEnv` requires `ModuleSubscriptions`
     fn subscription_actor(relational_db: Arc<RelationalDB>) -> ModuleSubscriptions {
+        // Create and enter a Tokio runtime to run the `ModuleSubscriptions`' background workers in parallel.
+        let runtime = tokio::runtime::Runtime::new().unwrap();
+        let _rt = runtime.enter();
         ModuleSubscriptions::new(relational_db, <_>::default(), Identity::ZERO)
     }
 
--- a/crates/core/src/host/instance_env.rs
+++ b/crates/core/src/host/instance_env.rs
@@ -514,7 +514,9 @@ mod test {
         host::Scheduler,
         messages::control_db::{Database, HostType},
         replica_context::ReplicaContext,
-        subscription::module_subscription_actor::ModuleSubscriptions,
+        subscription::{
+            module_subscription_actor::ModuleSubscriptions, module_subscription_manager::SubscriptionManager,
+        },
     };
     use anyhow::{anyhow, Result};
     use spacetimedb_lib::db::auth::StAccess;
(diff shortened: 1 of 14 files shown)


crates/core/src/host/instance_env.rs
```diff
@@ -534,6 +534,9 @@ mod test {
 
     /// An `InstanceEnv` requires `ModuleSubscriptions`
     fn subscription_actor(relational_db: Arc<RelationalDB>) -> ModuleSubscriptions {
+        // Create and enter a Tokio runtime to run the `ModuleSubscriptions`' background workers in parallel.
+        let runtime = tokio::runtime::Runtime::new().unwrap();
+        let _rt = runtime.enter();
```
cloutiertyler, line 539: Why do subscriptions need their own runtime? Also shouldn't we limit the number of threads for this runtime? The default is `num_cpu`.

PR #2798, opened 2025-05-27
Datastore Extraction: Create DatastoreError type

cloutiertyler:
# Description of Changes

This PR splits out a `DatastoreError` type from `DBError`. This is necessary for `Datastore` extraction because `DBError` references a hair mess of things in the relational db and elsewhere. This also gives the datastore a cleaner API with a clearer set of errors that it can produce.

None

2, this is in principle a non-functional change, but it's larger so a semantic mistake could be made. Also I can't imagine that this won't also require a private PR.

# Testing

None yet. Need to check in on private.


cloutiertyler:
Looks like internal checks passed actually

PR #2801, opened 2025-05-28
Pin db threads to cores

coolreader18:
# Description of Changes

This pins the threads that wasm instances run on, tokio worker threads, and rayon threads to specific cores, in the following arrangement:

* Module instance threads: 1/8 of available cores, with a round-robin-ish repinning system that tries to ensure that if there are >`nproc/8` databases running, all the database cores will have roughly the same number of instances pinned to them.
* tokio worker threads: 4/8 of available cores, with `tokio::runtime::Builder::num_threads` set to the same number.
* rayon threads: 1/8 of available cores.
* tokio blocking threads: on linux, all of them have their `sched_setaffinity` set to a cpuset of the remaining 3/8.

# Testing

- [x] Bot-tested and used in production for BitCraft.

### DIFF
--- a/crates/core/src/host/host_controller.rs
+++ b/crates/core/src/host/host_controller.rs
@@ -587,7 +587,10 @@ async fn make_module_host(
     core: JobCore,
 ) -> anyhow::Result<(Program, ModuleHost)> {
     // `make_actor` is blocking, as it needs to compile the wasm to native code,
-    // which may be computationally expensive.
+    // which may be computationally expensive - sometimes up to 1s for a large module.
+    // TODO: change back to using `spawn_rayon` here - asyncify runs on tokio blocking
+    //       threads, but those aren't for computation. Also, wasmtime uses rayon
+    //       to run compilation in parallel, so it'll need to run stuff in rayon anyway.
     asyncify(move || {
         let module_host = match host_type {
             HostType::Wasm => {


crates/core/src/util/jobs.rs
```diff
@@ -0,0 +1,204 @@
...
+use std::sync::{Arc, Mutex, Weak};
+
+use core_affinity::CoreId;
+use indexmap::IndexMap;
+use smallvec::SmallVec;
+use spacetimedb_data_structures::map::HashMap;
+use tokio::sync::{mpsc, oneshot, watch};
+
+#[derive(Default, Clone)]
+pub struct JobCores {
```
gefjon, line 29: Needs doc comment describing (along with my best guesses as to the answers):
- What this thing is.
  - It manages the database cores and allocates them to databases.
- What these cores are used for.
  - They run reducers. I don't think they do anything else.
- What code transfers control/work to and from these cores.
  - I have no clue, at least as of writing this comment.
- That this object is cheaply `Clone`-able.
- That each process should only construct at most one of these (I assume), and what code does that construction.
- The algorithm by which databases are assigned to cores.
  - I think each database is assigned to one specific core, never leaves that core, and that the cores are allocated round-robin.
  - This should link to the method which does the allocation, which I think is [`Self::take`].
- What the deal is with the `Option` around the inner here, in what contexts it will be `None`, and what happens in that case.
  - I think it will be `None` if we fail to read information about the number and layout of hardware cores/threads, and in that case databases will not be pinned.

crates/core/src/util/jobs.rs
```diff
@@ -0,0 +1,204 @@
...
+}
+fn cores_cmp(_: &CoreId, v1: &CoreInfo, _: &CoreId, v2: &CoreInfo) -> cmp::Ordering {
+    Ord::cmp(&v1.jobs.len(), &v2.jobs.len())
+}
+
+#[derive(Copy, Clone, Hash, PartialEq, Eq, PartialOrd, Ord)]
+struct JobThreadId(usize);
+
+impl JobCores {
+    pub fn take(&self) -> JobCore {
```
gefjon, line 55: Needs doc comment describing when and why you'd call this. "Get a [`JobCore`] on which to schedule a newly-instantiated database," or something.

crates/core/src/util/jobs.rs
```diff
@@ -0,0 +1,204 @@
...
+                cores,
+                next_id: 0,
+            }))
+        });
+        Self { inner }
+    }
+}
+
+impl JobCoresInner {
+    fn thread_exited(&mut self, id: JobThreadId) {
```
gefjon: Needs a doc comment and/or a more descriptive name. Who calls this and when? Is it by the `JobCores` manager after it notices that a thread has exited, or is it by the thread before it will exit? What cleanup does it do? Does it schedule a new thread on the dead thread's core, or will that core be unused afterwards? What would cause a thread to exit?

crates/core/src/util/jobs.rs
```diff
@@ -0,0 +1,204 @@
...
+            sender.send_replace(core_id);
+        } else {
+            core.jobs.remove(pos);
+        }
+        self.cores.sort_by(cores_cmp);
+    }
+}
+
+#[derive(Default)]
+pub struct JobCore {
```
gefjon, line 144: Needs doc comment. "A thread pinned to a hardware core which runs reducers," or similar. Also a description of when and why the `inner` can be `None`.

crates/core/src/util/jobs.rs
```diff
@@ -0,0 +1,204 @@
...
+        T: ?Sized + 'static,
+    {
+        let (tx, rx) = mpsc::channel::<Box<Job<T>>>(50);
+
+        let handle = tokio::runtime::Handle::current();
+        std::thread::spawn(move || {
+            let mut data = init();
+            let data = as_t(&mut data);
+            self.job_loop(&handle, rx, data);
+        });
```
gefjon: Does this closure want/need to enter the `handle`? Either way, comment justifying that choice.

crates/core/src/util/jobs.rs
```diff
@@ -0,0 +1,204 @@
...
+}
+
+struct JobCoreInner {
+    repin_rx: watch::Receiver<CoreId>,
+    cores: Weak<Mutex<JobCoresInner>>,
+    id: JobThreadId,
+}
+
+impl JobCore {
+    pub fn start<F, F2, U, T>(self, init: F, as_t: F2) -> JobThread<T>
```
gefjon:
```suggestion
    /// Spawn a thread to run reducers and pin it to this core.
    pub fn start<F, F2, U, T>(self, init: F, as_t: F2) -> JobThread<T>
```

crates/core/src/util/jobs.rs
```diff
@@ -0,0 +1,204 @@
...
+    }
+}
+
+#[derive(Default)]
+pub struct JobCore {
+    inner: Option<JobCoreInner>,
+}
+
+struct JobCoreInner {
+    repin_rx: watch::Receiver<CoreId>,
```
gefjon, line 149: What is the `repin_rx` used for? When would we re-pin a `JobCore`?

crates/core/src/util/jobs.rs
```diff
@@ -0,0 +1,204 @@
...
+            loop {
+                let repin_fut = async {
+                    let rx = repin_rx.as_mut()?;
+                    rx.changed().await.ok()?;
+                    Some(*rx.borrow_and_update())
+                };
+                tokio::select! {
+                    Some(core_id) = repin_fut => {
+                        core_affinity::set_for_current(core_id);
+                    }
```
gefjon: Needs commentary about why we bother to handle this case. When and why would we re-pin a job thread?

crates/core/src/util/jobs.rs
```diff
@@ -0,0 +1,204 @@
...
+                    job = rx.recv() => {
+                        let Some(job) = job else { break };
+                        tokio::task::block_in_place(|| job(data))
+                    }
```
gefjon: Needs commentary about what sort of jobs we'll be getting here and who's sending them.

crates/core/src/util/jobs.rs
```diff
@@ -0,0 +1,204 @@
...
+            .inner
+            .map(|inner| (inner.repin_rx, (inner.cores, inner.id)))
+            .unzip();
+        if let Some(repin_rx) = &mut repin_rx {
+            core_affinity::set_for_current(*repin_rx.borrow_and_update());
+        }
+        scopeguard::defer!({
+            if let Some((cores, id)) = cores_id {
+                if let Some(cores) = cores.upgrade() {
+                    cores.lock().unwrap().thread_exited(id);
```
gefjon: Needs commentary similar to what I requested on `fn thread_exited` which describes what clean-up this method will do and why it's necessary.

crates/core/src/util/jobs.rs
```diff
@@ -0,0 +1,204 @@
...
+impl<T: ?Sized> Clone for JobThread<T> {
+    fn clone(&self) -> Self {
+        Self { tx: self.tx.clone() }
+    }
+}
+
+type Job<T> = dyn FnOnce(&mut T) + Send;
+
+impl<T: ?Sized> JobThread<T> {
+    pub async fn run<F, R>(&self, f: F) -> R
```
gefjon: Needs doc comment and more descriptive name. Does this method run the  job thread, as in, is it called by `JobCore::start`, or is it used by an outsider to enqueue a job? I believe it is the latter. Also, please describe who will call this and in what contexts, and any constraints on what `F` should do. E.g.:
- Is `F` allowed to panic? If so, what happens to the caller of this method? What happens to the `JobThread`?
- Will `F` have access to a Tokio runtime?
- Is `F` allowed to block on I/O operations like networking or disk access?

crates/core/src/util/jobs.rs
```diff
@@ -0,0 +1,204 @@
...
+                        let Some(job) = job else { break };
+                        tokio::task::block_in_place(|| job(data))
+                    }
+                }
+            }
+        });
+    }
+}
+
+pub struct JobThread<T: ?Sized> {
```
gefjon, line 236: Needs a quick one-sentence doc comment. "Handle on a thread running on a [`JobCore`], which can be used to [`Self::run`] reducers," or something.

crates/core/src/util/jobs.rs
```diff
@@ -0,0 +1,204 @@
...
+        ret_rx.await.unwrap().unwrap_or_else(|e| std::panic::resume_unwind(e))
+    }
+
+    pub fn downgrade(&self) -> WeakJobThread<T> {
+        let tx = self.tx.downgrade();
+        WeakJobThread { tx }
+    }
+}
+
+pub struct WeakJobThread<T: ?Sized> {
```
gefjon, line 306: Needs doc comment describing who wants a `WeakJobThread` and why.

crates/core/src/subscription/subscription.rs
```diff
@@ -523,7 +522,7 @@ impl ExecutionSet {
         let tables = self
             .exec_units
             // if you need eval to run single-threaded for debugging, change this to .iter()
-            .par_iter()
+            .iter()
```
gefjon, line 525: Comment is out-of-date. Would be nice to replace it with a comment which says that we used to `par_iter` here, and why we stopped. If this runs on a `JobCore` thread, that would be worth pointing out.

crates/core/src/startup.rs
```diff
@@ -172,24 +152,114 @@ fn reload_config<S>(conf_file: &ConfigToml, reload_handle: &reload::Handle<EnvFi
...
-        // possibly via https://docs.rs/nix/latest/nix/sched/fn.sched_setaffinity.html,
-        // and restricting Tokio threads to one CPU pool
-        // and Rayon threads to the other.
-        // Then we should give Tokio and Rayon each a number of worker threads
-        // equal to the size of their pool.
-        .num_threads(std::thread::available_parallelism().unwrap().get() / 2)
-        .build_global()
-        .unwrap()
+#[derive(Default)]
+pub struct Cores {
```
gefjon, line 178: Needs doc comment describing:
- What the three groups of cores are.
- In what ratio and sequence the cores are allocated by default.
- Why this is desirable (improved cache performance, lower scheduler and context-switching overhead, &c).
- That this is a singleton object (I assume).
- Who constructs a `Cores` (link to `pin_threads`).

crates/core/src/startup.rs
```diff
@@ -54,19 +41,12 @@ impl Default for TracingOptions {
...
-    pub fn configure(self) {
-        if let Some(tracing_opts) = self.tracing {
-            configure_tracing(tracing_opts)
-        }
-        if self.rayon {
-            configure_rayon()
-        }
-    }
+#[must_use]
+pub fn pin_threads() -> Cores {
```
gefjon: Needs a doc comment which says:
- Who will call this function and at what point during startup.
- That this function must only be called once per process (I assume).
- What happens if `Cores::get` returns `None`, and why that would happen.
- Are the threads fully pinned and set up by the time this method returns, or does the caller have further responsibilities?

crates/core/src/startup.rs
```diff
@@ -172,24 +152,114 @@ fn reload_config<S>(conf_file: &ConfigToml, reload_handle: &reload::Handle<EnvFi
...
+
+        let tokio_workers = cores.take(frac(4.0 / 8.0)).collect();
+
+        let rayon = RayonCores(Some(cores.take(frac(1.0 / 8.0)).collect()));
+
+        #[cfg(target_os = "linux")]
+        let remaining = cores.try_fold(nix::sched::CpuSet::new(), |mut cpuset, core| {
+            cpuset.set(core.id).ok()?;
+            Some(cpuset)
+        });
```
gefjon, line 214: This needs a comment describing what's happening here, why we do it, and what happens on non-Linux platforms instead.

crates/core/src/startup.rs
```diff
@@ -172,24 +152,114 @@ fn reload_config<S>(conf_file: &ConfigToml, reload_handle: &reload::Handle<EnvFi
...
+        queue.push(core).unwrap();
+    }
+    queue
+}
+
+#[derive(Default)]
+pub struct TokioCores {
+    workers: Option<Vec<CoreId>>,
+    #[cfg(target_os = "linux")]
+    blocking: Option<nix::sched::CpuSet>,
```
gefjon, line 239: Needs same comment as where it's constructed in `Cores::get`.

crates/core/src/startup.rs
```diff
@@ -172,24 +152,114 @@ fn reload_config<S>(conf_file: &ConfigToml, reload_handle: &reload::Handle<EnvFi
...
+}
+
+impl TokioCores {
+    pub fn configure(self, builder: &mut tokio::runtime::Builder) {
```
gefjon, line 244: Needs doc comment describing who calls this method and what it does. How does it affect the `builder`?

crates/core/src/startup.rs
```diff
@@ -172,24 +152,114 @@ fn reload_config<S>(conf_file: &ConfigToml, reload_handle: &reload::Handle<EnvFi
...
+            });
+        }
+    }
+}
+
+#[derive(Default)]
+pub struct RayonCores(Option<Vec<CoreId>>);
+
+impl RayonCores {
+    pub fn configure(self, tokio_handle: &tokio::runtime::Handle) {
```
gefjon, line 278: Needs doc comment describing who calls this method and what it does. What is the `tokio_handle` used for? What are the side-effects of calling this method?

crates/core/src/startup.rs
```diff
@@ -172,24 +152,114 @@ fn reload_config<S>(conf_file: &ConfigToml, reload_handle: &reload::Handle<EnvFi
...
+            // in the case where we have many modules running many reducers,
+            // we'll wind up with Rayon threads competing with each other and with Tokio threads
+            // for CPU time.
+            //
+            // We should investigate creating two separate CPU pools,
+            // possibly via https://docs.rs/nix/latest/nix/sched/fn.sched_setaffinity.html,
+            // and restricting Tokio threads to one CPU pool
+            // and Rayon threads to the other.
+            // Then we should give Tokio and Rayon each a number of worker threads
+            // equal to the size of their pool.
```
gefjon: Comment appears outdated.

crates/core/src/host/host_controller.rs
```diff
@@ -95,6 +96,8 @@ pub struct HostController {
     pub page_pool: PagePool,
     /// The runtimes for running our modules.
     runtimes: Arc<HostRuntimes>,
+    /// The CPU cores that are reserved for running databases on.
```
gefjon: What does "running databases on" mean? Reducers? One-off SQL queries? Incremental queries?

crates/core/src/host/host_controller.rs
```diff
@@ -267,7 +272,14 @@ impl HostController {
...
     /// as the automigration planner and executor accomplish the same validity checks.
     pub async fn check_module_validity(&self, database: Database, program: Program) -> anyhow::Result<Arc<ModuleInfo>> {
-        Host::try_init_in_memory_to_check(&self.runtimes, self.page_pool.clone(), database, program).await
+        Host::try_init_in_memory_to_check(
+            &self.runtimes,
+            self.page_pool.clone(),
+            database,
+            program,
+            self.db_cores.take(),
+        )
```
gefjon: Concern: it appears that every fresh publish will allocate two db cores, and then only use the latter. This may expose us to a pathological case in a system with two db cores, where every database ends up on the same thread. If so, this could use a TODO/FIXME comment. It doesn't need an actual fix within this PR, though.

crates/core/src/host/host_controller.rs
```diff
@@ -564,8 +579,9 @@ async fn make_module_host(
     program: Program,
     energy_monitor: Arc<dyn EnergyMonitor>,
     unregister: impl Fn() + Send + Sync + 'static,
+    core: JobCore,
 ) -> anyhow::Result<(Program, ModuleHost)> {
-    spawn_rayon(move || {
+    asyncify(move || {
```
gefjon, line 594: Could use a quick comment describing why this needs to be `asyncify`'d / run on a blocking core.

crates/core/src/host/module_host.rs
```diff
@@ -457,7 +459,7 @@ impl fmt::Debug for ModuleHost {
 
 #[async_trait::async_trait]
 trait DynModuleHost: Send + Sync + 'static {
-    async fn get_instance(&self, db: Identity) -> Result<Box<dyn ModuleInstance>, NoSuchModule>;
+    // async fn get_instance(&self, db: Identity) -> Result<Box<dyn ModuleInstance>, NoSuchModule>;
```
gefjon: ?

crates/core/src/host/module_host.rs
```diff
@@ -469,6 +471,7 @@ struct HostControllerActor<T: Module> {
 }
 
 impl<T: Module> HostControllerActor<T> {
+    #[allow(unused)]
```
gefjon: If this is unused, can we remove it?

crates/core/src/host/module_host.rs
```diff
@@ -485,6 +488,7 @@ impl<T: Module> HostControllerActor<T> {
 
 /// runs future A and future B concurrently. if A completes before B, B is cancelled. if B completes
 /// before A, A is polled to completion
+#[allow(unused)]
```
gefjon: Same question re: unused.

crates/core/src/host/host_controller.rs
coolreader18: Not a concern, due to the round-robin algorithm in use.

crates/core/src/host/module_host.rs
coolreader18: This was commented out in Jeff's branch - I think I'll just remove it wholesale; MVCC seems to be a long ways off anyway.

coolreader18:
Ah. I think the failure is possibly because `DynModuleHost::exit()` no longer actually stops things from running.

crates/core/src/subscription/subscription.rs
coolreader18: The thing is that, afaict, nothing in this file (or at least this function) is used anymore outside of tests. I'm not sure what's up with that.

crates/core/src/host/host_controller.rs
gefjon: Can you add a comment here describing why it's not a concern?

crates/core/src/host/host_controller.rs
```diff
@@ -564,8 +579,10 @@ async fn make_module_host(
     program: Program,
     energy_monitor: Arc<dyn EnergyMonitor>,
     unregister: impl Fn() + Send + Sync + 'static,
+    core: JobCore,
 ) -> anyhow::Result<(Program, ModuleHost)> {
-    spawn_rayon(move || {
+    // `make_actor` is blocking, as it needs to compile the wasm to native code.
```
gefjon: Does compiling WASM to native code involve I/O, mutexes, or other blocking syscalls?

crates/core/src/startup.rs
```diff
@@ -172,24 +147,146 @@ fn reload_config<S>(conf_file: &ConfigToml, reload_handle: &reload::Handle<EnvFi
...
+/// Divide up the available CPU cores into pools for different purposes.
+///
+/// Use the fields of the returned [`Cores`] value to actually configure
+/// cores to be pinned.
+///
+/// Pinning different subsystems to different threads reduces overhead from
+/// unnecessary context switching.
+///
+/// * Database instances are critical to overall performance, and keeping each
+///   one on only one thread was shown to significantly increase CCU.
```
gefjon:
```suggestion
///   one on only one thread was shown to significantly increase transaction throughput.
```

crates/core/src/util/jobs.rs
gefjon: What part of our codebase holds a `WeakJobThread`? What is the motivation for that part of our codebase doing so? It's pretty obvious *what* a `WeakJobThread` is; I want to know *why* it's useful *in our specific codebase*.

crates/core/src/host/host_controller.rs
coolreader18: No, but it's computationally intensive.

crates/core/src/util/jobs.rs
coolreader18: If someone's wondering that, couldn't they just grep for it / Go to References? I don't mind adding the comment, it just seems like it might be redundant.

PR #2806, opened 2025-05-28
`spacetimedb_execution`: avoid `get_row_ref`

Centril:
# Description of Changes

Avoid all `Table::get_row_ref`s, in `spacetimedb_execution`, as they are not free (as noticed in flamegraphs).
Instead, simplify the code with `TableAndIndex` and gain in performance.

None

Fairly self-contained and mostly simplifying.
There are two trivial new instances of `unsafe`.

# Testing

Covered by existing tests.


gefjon (approved):
Seems simple enough.

PR #2812, opened 2025-05-31
Fix typos

szepeviktor:
# Description of Changes

Found few misspellings.

🪓 https://github.com/crate-ci/typos


crates/bindings-macro/src/sats.rs
```diff
@@ -52,8 +52,8 @@ pub(crate) fn sats_type_from_derive(
     crate_fallback: TokenStream,
 ) -> syn::Result<SatsType<'_>> {
     let data = match &input.data {
-        syn::Data::Struct(struc) => {
-            let fields = struc.fields.iter().map(|field| SatsField {
+        syn::Data::Struct(struct) => {
```
bfops: `struct` is a reserved word, so this won't work as-is

szepeviktor: Do you like one of these?

    structure
    struct_data
    s
    fields_struct
    definition

bfops: no strong preference from me. I think `s` would be fine.

szepeviktor: Done 🍏

crates/bindings-macro/src/sats.rs
```diff
@@ -52,8 +52,8 @@ pub(crate) fn sats_type_from_derive(
     crate_fallback: TokenStream,
 ) -> syn::Result<SatsType<'_>> {
     let data = match &input.data {
-        syn::Data::Struct(struc) => {
-            let fields = struc.fields.iter().map(|field| SatsField {
+        syn::Data::Struct(s) => {
+            let fields = s.fields.iter().map(|field| SatsField {
```
gefjon: This is not an improvement; `struc` is more clear and readable than just `s`. Please revert this change.

szepeviktor: Done 🍏

szepeviktor:
Glad to contribute.

PR #2821, opened 2025-06-02
Remove the lock around the `ClientsMap` in the `SubscriptionManager`

### LINKED ISSUE #2820: Avoid taking locks in `send_worker`
This involves removing the `RwLock` around the `ClientsMap`, collecting the client info, and passing it to the `send_worker` task directly.

https://github.com/clockworklabs/SpacetimeDB/blob/1c65d7bc223d187e5e93633fe562706109d12867/crates/core/src/subscription/module_subscription_manager.rs#L406

gefjon:
# Description of Changes

We were seeing a deadlock where at least one Tokio task was blocked on this lock. @jsdt had some hypotheses about the specific Tokio behaviors that were causing this, but the important thing is that the lock is not necessary at all, and so we can remove it without delving into the dark depths of the Tokio scheduler.

As part of this change, I split the send worker into being its own struct. That struct now holds its own copy of the clients map. This also simplifies the code that cleans up the send queue length metric.

Closes #2820 .

N/a

3: I may have missed a place where the client map is updated, which needs to notify the `SendWorker`.

# Testing

- [x] @kim to test with replication-proto.
- [ ] Bot test to see if we hit the deadlock again.


crates/core/src/subscription/module_subscription_manager.rs
```diff
@@ -458,7 +448,16 @@ struct ComputedQueries {
     errs: Vec<(ClientId, Box<str>)>,
     event: Arc<ModuleEvent>,
     caller: Option<Arc<ClientConnectionSender>>,
-    clients: ArcRwLockReadGuard<ClientsMap>,
+}
+
+enum SendWorkerMessage {
```
Centril, line 486:
```suggestion
/// A message handled by [`SendWorker`].
enum SendWorkerMessage {
```

crates/core/src/subscription/module_subscription_manager.rs
```diff
@@ -545,6 +530,40 @@ impl SubscriptionManager {
...
+
+    fn remove_client_and_inform_send_worker(&mut self, client_id: ClientId) -> Option<ClientInfo> {
+        if let Some(client) = self.clients.remove(&client_id) {
+            self.send_worker_tx
+                .send(SendWorkerMessage::RemoveClient(client_id))
+                .expect("send worker has panicked, or otherwise dropped its recv queue!");
+            Some(client)
+        } else {
+            None
+        }
```
Centril:
```suggestion
        self.clients.remove(&client_id).inspect(|_| {
            self.send_worker_tx
                .send(SendWorkerMessage::RemoveClient(client_id))
                .expect("send worker has panicked, or otherwise dropped its recv queue!");
        })
```

crates/core/src/subscription/module_subscription_manager.rs
```diff
@@ -545,6 +530,40 @@ impl SubscriptionManager {
...
+                    client_id,
+                    dropped: info.dropped.clone(),
+                    outbound_ref,
+                })
+                .expect("send worker has panicked, or otherwise dropped its recv queue!");
+            info
+        })
+    }
+
+    fn remove_client_and_inform_send_worker(&mut self, client_id: ClientId) -> Option<ClientInfo> {
```
Centril, line 596:
```suggestion
    /// Remove a [`ClientInfo`] from the `clients` map, and broadcast a message along `send_worker_tx`
    /// that the [`SendWorker`] should also add this client.
    fn remove_client_and_inform_send_worker(&mut self, client_id: ClientId) -> Option<ClientInfo> {
```

crates/core/src/subscription/module_subscription_manager.rs
```diff
@@ -118,7 +113,7 @@ struct ClientInfo {
     legacy_subscriptions: HashSet<QueryHash>,
     // This flag is set if an error occurs during a tx update.
     // It will be cleaned up async or on resubscribe.
-    dropped: AtomicBool,
+    dropped: Arc<AtomicBool>,
```
Centril, line 119: The `Arc` here and the other cases of `dropped` in other structs could use some commentary. At first glance it's not obvious why it's there.

crates/core/src/subscription/module_subscription_manager.rs
```diff
@@ -1092,50 +1103,109 @@ impl SubscriptionManager {
...
+    queue_length_metric: Option<IntGauge>,
+
+    /// Mirror of the [`SubscriptionManager`]'s `clients` map local to this actor.
     ///
-    /// `_clean_up_metric` will be dropped upon exiting this worker,
-    /// and should be a [`ScopeGuard`] which does `remove_label_values` on the `queue_length_metric`.
-    async fn send_worker(
-        mut rx: mpsc::UnboundedReceiver<ComputedQueries>,
+    /// Updated by [`SendWorkerMessage::AddClient`] and [`SendWorkerMessage::RemoveClient`] messages
+    /// send along `self.rx`.
```
Centril:
```suggestion
    /// Updated by [`SendWorkerMessage::AddClient`] and [`SendWorkerMessage::RemoveClient`] messages
    /// sent along `self.rx`.
```

gefjon:
Kim reports out-of-band that in his testing, he could not induce a deadlock with this branch.

crates/core/src/subscription/module_subscription_manager.rs
```diff
@@ -458,7 +451,28 @@ struct ComputedQueries {
...
     event: Arc<ModuleEvent>,
     caller: Option<Arc<ClientConnectionSender>>,
-    clients: ArcRwLockReadGuard<ClientsMap>,
+}
+
+/// Message sent by the [`SubscriptionManager`] to the [`SendWorker`].
+enum SendWorkerMessage {
+    /// A transaction has completed and the [`SubscriptionManager`] has evaluated the incremental queries,
+    /// so the [`SendWorker`] should broadcast them to clients.
+    Broadcast(ComputedQueries),
```
jsdt, line 489: I think we have a race condition if this worker is only sending messages for transaction updates. Consider this case with a single client and two tables (`table1` and `table2`):

1. Client subscribes to a query `select * from table1`, and successfully gets the data for it.
2. A reducer happens which inserts one row in `table1` and one row in `table2`. The `Broadcast` message is queued (but not yet processed), which will eventually tell the client about the new row in `table1`.
3. The client subscribes to a new query `select * from table2`, and is sent the data for it.
4. The `Broadcast` message is processed.

After step 3, the client will believe that they are subscribed to `select * from table1` and `select * from table2`, but they will have a view of the database that only has a row in `table2`, even though the database was never in that state.

I think we want to route all of the outgoing client messages through the `SendWorker` to make sure clients have already been sent messages for previous transactions. Even something like a reducer error message should go through that queue, since the error could be because of a transaction that we haven't told the client about yet.

crates/core/src/subscription/module_subscription_manager.rs
```diff
@@ -458,7 +451,28 @@ struct ComputedQueries {
...
+
+    /// A new client has been registered in the [`SubscriptionManager`],
+    /// so the [`SendWorker`] should also record its existence.
+    AddClient {
```
jsdt, line 493: This should fix the lock issue. I want to point out that we could also send messages when clients add and remove subscriptions, so the send worker could maintain its own view of the subscription managers `clients: HashMap<ClientId, ClientInfo>` and `queries: HashMap<QueryHash, QueryState>` fields.

That could let us offload some more work from the reducer to the sendworker, since the `Broadcast` messages could just send the data for each query being updated, without doing any per-client work.

Issue #1858, opened 2024-10-15
Remove unnecessary allocations in message serialization


joshua-spacetime:
I.e. address the following `TODO` left by @Centril.

https://github.com/clockworklabs/SpacetimeDB/blob/c4e637e98b20c3db29bdcaa2c0da9ea0672bdf98/crates/core/src/client/messages.rs#L29-L65

PR #2823, opened 2025-06-03
`messages::serialize`: take/put buffers from/into a `SerializeBufferPool`

Centril:
# Description of Changes

Fixes https://github.com/clockworklabs/SpacetimeDB/issues/1858.

In `ws_client_actor_inner`, after having `ws.feed(msg)`, keep underlying allocations around, and use a default size of 4KiB.

None

The changes are somewhat straight forward, and it's not a deep change, but it's in a fairly important place.

# Testing

- [ ] Run a perf test and inpect flamegraphs and friends.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5470,6 +5470,7 @@ dependencies = [
  "core_affinity",
  "criterion",
  "crossbeam-channel",
+ "crossbeam-queue",
  "derive_more",
  "dirs",
  "enum-as-inner",
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -48,6 +48,7 @@ bytes.workspace = true
 bytestring.workspace = true
 chrono.workspace = true
 crossbeam-channel.workspace = true
+crossbeam-queue.workspace = true
 derive_more.workspace = true
 dirs.workspace = true
 enum-as-inner.workspace = true


gefjon (commented):
Code looks reasonable. I'd prefer to have the pool size metric(s) in before we merge, so we can know that the pool isn't growing without bound. For these metrics, I'd recommend:

- Number of buffers in the pool.
- Total size in bytes of all buffers in the pool.
- Number of newly-allocated buffers.

I think these would give us sufficient information to decide if we need to impose a cap, and what that cap should be if so.

crates/core/src/client/messages.rs
```diff
@@ -27,36 +30,87 @@ pub trait ToProtocol {
...
+            serde_json::to_writer(&mut out, &SerializeWrapper::new(msg))
+                .expect("should be able to json encode a `ServerMessage`");
+            let out = out.into_inner();
+
+            let out = out.freeze();
+            // SAFETY: `serde_json::to_writer` states that:
+            // > "Serialization guarantees it only feeds valid UTF-8 sequences to the writer."
+            let msg_json = unsafe { ByteString::from_bytes_unchecked(out) };
+            msg_json.into()
+        }
```
coolreader18: Maybe we'd want a separate pool for json? (or no pool?) I'd imagine the lengths of json messages and bsatn are very different

coolreader18:
Is there a benefit between having a global queue and just reusing the same buffer per client? Given it seems that each client would only use 1 or 2 buffers, and then the buffer capacities are also more specialized to the specific client (and also the module) rather than growing towards the global maximum.

coolreader18:
Something like [this](https://github.com/clockworklabs/SpacetimeDB/compare/noa/reuse-ws-msg-allocations)

crates/core/src/client/messages.rs
Centril: If we want to differentiate here I'd probably go for no pool, but some excess buffer sizes are probably not a problem, and N/A now that the PR has local buffers.

Centril:
> Is there a benefit between having a global queue and just reusing the same buffer per client? Given it seems that each client would only use 1 or 2 buffers, and then the buffer capacities are also more specialized to the specific client (and also the module) rather than growing towards the global maximum.

A few benefits to global pools:
- Less memory usage overall for the buffers (not a huge difference... at 1024 CCU we're talking around 8 MiB).
- Better cache locality as the actual OS thread is not jumping between different buffers.
- In case of clients leaving and joining frequently, the local buffers don't help while the global pool does.

That said, the benefit of local buffers is the avoided synchronization overhead.
I think we can try local buffers first. That way, we also don't have to care about metrics the same way.

gefjon (approved):

PR #2826, opened 2025-06-04
Invoke __identity_connected__ for sql http requests

### LINKED ISSUE #2802: Invoke `__identity_connected__` for sql http requests
We invoke `__identity_connected__` for reducer calls over http but not for sql. This is a problem since `__identity_connected__` can have access control logic.

mamcx:
# Description of Changes

Closes #2802

We invoke __identity_connected__ for reducer calls over http but not for sql. This is a problem since __identity_connected__ can have access control logic.

Minor: It will run the `on_connect and on_disconnect` so in could reject calls that before run.

1

# Testing

- [x] Add smoke test

### DIFF
--- a/crates/client-api/src/routes/database.rs
+++ b/crates/client-api/src/routes/database.rs
@@ -8,7 +8,7 @@ use crate::auth::{
 };
 use crate::routes::subscribe::generate_random_connection_id;
 use crate::util::{ByteStringBody, NameOrIdentity};
-use crate::{log_and_500, ControlStateDelegate, DatabaseDef, NodeDelegate};
+use crate::{log_and_500, ControlStateDelegate, DatabaseDef, Host, NodeDelegate};
 use axum::body::{Body, Bytes};
 use axum::extract::{Path, Query, State};
 use axum::response::{ErrorResponse, IntoResponse};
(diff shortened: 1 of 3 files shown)


crates/client-api/src/routes/database.rs
```diff
@@ -400,24 +439,23 @@ where
...
-        .await
-        .map_err(log_and_500)?
-        .ok_or(StatusCode::NOT_FOUND)?;
-    let json = host.exec_sql(auth, database, body).await?;
+    let json = leader.exec_sql(auth, database, body).await?;
 
     let total_duration = json.iter().fold(0, |acc, x| acc + x.total_duration_micros);
 
+    call_on_disconnect(module, connection_id, caller_identity).await?;
+
```
coolreader18: You shouldn't `?` on the result from `exec_sql` until after you call `call_on_disconnect`; otherwise, if the query fails for any reason, the module will never know that the client has disconnected. probably want something like

```rust
let json_result = exec_sql.await;
let disconn_result = call_on_disconnect.await;
let json = json_result?;
disconn_result?;
```

But that precise order of a sql error taking precedence over an on_disconnect error probably doesn't matter too much

coolreader18 (changes requested):

joshua-spacetime:
Superseded by https://github.com/clockworklabs/SpacetimeDB/pull/4563

Issue #2830, opened 2025-06-05
Cannot use an RLS rule that `JOIN`s with a private table


egormanga:
Reproduction:
```rust
#[table(name=test, public)]
struct Test {
	#[primary_key]
	#[auto_inc]
	id: u64,
}

#[client_visibility_filter]
const TEST_FILTER: Filter = Filter::Sql("SELECT test.* FROM test JOIN test_access ON test.id = test_access.id WHERE user_id = :sender");

#[table(name=test_access)]
struct TestAccess {
	#[primary_key]
	id: u64,
	user_id: Identity,
}
```

Compiles **and publishes** successfully, but when querying it fails with:
```
`test_access` is not a valid table
```

Should be a pretty obvious case to add a more complex restriction logic based on some private (sensitive) fields.

egormanga:
Update:

RLS queries seem to be evaluated from the caller identity, effectively rendering useless most part of it. In particular, this means RLS cannot reference not only private tables, but public ones with non-passing RLS as well.

This behavior heavily impedes my current use case.

egormanga:
For clarity: I need the RLS query to be evaluated (incl. recursively) with the system identity, while the caller's query would still retain its identity.

The current behavior actually seems to be: evaluate the RLS query from the system identity for accessing the current table, but from the caller identity for accessing the joining table. It's clearly a bug, then.

bfops:
Thank you for filing this!

Is this the exact repro? To compile and publish successfully, I had to change the second struct to `TestAccess` (to avoid two structs with the same name) and update the filter line to be:
```
const TEST_FILTER: Filter = Filter::Sql("SELECT test.* FROM test JOIN test_access ON test.id = test_access.id WHERE test_access.user_id = :sender");
```

Does that match what you're doing? I want to make sure we're looking at the right thing 🙂

bfops:
Here's what I would suggest, at least for now: make the private table `public`, but add a restriction so that users are not able to see the data that doesn't pertain to them.

This seemed to work for me:
```rs
use spacetimedb::{ReducerContext, Table, Filter, Identity, table, client_visibility_filter, reducer};

#[table(name=test, public)]
struct Test {
	#[primary_key]
	id: u64,
}

#[client_visibility_filter]
const TEST_FILTER: Filter = Filter::Sql("SELECT test.* FROM test JOIN test_access ON test.id = test_access.id WHERE test_access.user_id = :sender");

#[table(name=test_access, public)]
[...]

PR #2832, opened 2025-06-05
Evaluate RLS joins with the owner identity

egormanga:
# Description of Changes

Fix for a bigger issue brought up in https://github.com/clockworklabs/SpacetimeDB/issues/2830#issuecomment-2942621359.

Additionally, this seems to allow for some extra query optimizations (see tests).

None.

### 2:
While the fix seems pretty straightforward, it somehow affected query optimization logic so it might go a bit deeper than that.

# Testing

- [x] A simple setup:
     ```rust
     #[table(name=test, public)]
     pub struct Test {
     	owner_id: Identity,
     }

     #[client_visibility_filter]
     const TEST_FILTER: Filter = Filter::Sql("SELECT test.* FROM test JOIN access WHERE access.owner_id == :sender AND access.allowed = true");

     #[table(name=access, public)]
     pub struct Access {
     	owner_id: Identity,
     	allowed: bool,
     }

     #[client_visibility_filter]
     const ACCESS_FILTER: Filter = Filter::Sql("SELECT * FROM access WHERE false");

     #[reducer]
     pub fn test_add(ctx: &ReducerContext) {
     	ctx.db.test().insert(Test {
     		owner_id: ctx.sender,
     	});
     }
     ```
     1. call `test_add`
     2. add an `Access`
     3. sql `SELECT * FROM test;`
     Result: `[]` → `[0xIDENTITY]`
- [x] A personal project where the issue arose
- [x] Added a test
- [ ] Understand why the existing test result changed

### DIFF
--- a/crates/expr/src/rls.rs
+++ b/crates/expr/src/rls.rs
@@ -1,6 +1,7 @@
 use std::rc::Rc;
+use std::sync::Arc;
 
-use spacetimedb_lib::identity::AuthCtx;
+use spacetimedb_lib::identity::{AuthCtx, SqlPermission};
 use spacetimedb_primitives::TableId;
 use spacetimedb_sql_parser::ast::BinOp;
 
@@ -211,7 +212,7 @@ fn resolve_views_for_expr(
     suffix: &mut usize,
     auth: &AuthCtx,
 ) -> anyhow::Result<Vec<RelExpr>> {
-    if auth.caller == auth.owner {
+    if auth.bypass_rls() {
         return Ok([view].into());
     }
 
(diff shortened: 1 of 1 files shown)


crates/expr/src/rls.rs
```diff
@@ -265,7 +269,8 @@ fn resolve_views_for_expr(
                 ResolveList::new(table_id, resolving.clone()),
                 has_param,
                 suffix,
-                auth,
+                // Use the owner identity to evaluate the RLS query joins
+                &AuthCtx::for_current(auth.owner),
```
joshua-spacetime: @egormanga this means the expression will be evaluated as though it were from the owner identity, not the caller identity. And the owner bypasses RLS rules, meaning their access is not restricted. So this would result in non-owner identities gaining access to data they shouldn't.

egormanga: I don't think so. According to my testing, this only applies to the RLS rule evaluation, not to the original user query.

I might be wrong, but if I wouldn't get both positive test results for my case and negative for restricted data access (e.g. the `access` table in my example), I wouldn't've submitted this PR :)

egormanga:
#3525 broke it by removing the `.owner` identity field from `AuthCtx` and I don't have an idea on how to reimplement this over it.

@kim I'd love to hear your opinion on this!

cloutiertyler:
I'm closing this as stale since RLS has been subsumed by View Functions.

PR #2834, opened 2025-06-05
Send other client messages through the broadcast queue

jsdt:
# Description of Changes

For context, there is a discussion of potential race conditions [here](https://github.com/clockworklabs/SpacetimeDB/pull/2821#discussion_r2124355477).

The short explanation is that to make sure clients see things in a consistent order, we need to add outgoing messages to the `broadcast_queue` while we are still holding the associated database transaction lock.

I still need to do this for one-off queries.
This still has an issue with one-off queries, since I haven't routed those messages through.

The change is mostly a matter of changing code like `sender.send_message(message)` to `broadcast_queue.send_client_message(sender, message)`.

3. I don't have a good idea for how to prevent people from accidentally calling `sender.send_message` when they should be using the broadcast queue, so some bugs might creep back in over time.

# Testing

So far I gave this a very basic test with the quickstart-chat app locally.

### DIFF
--- a/crates/core/src/subscription/module_subscription_actor.rs
+++ b/crates/core/src/subscription/module_subscription_actor.rs
@@ -1,5 +1,7 @@
 use super::execution_unit::QueryHash;
-use super::module_subscription_manager::{Plan, SubscriptionGaugeStats, SubscriptionManager};
+use super::module_subscription_manager::{
+    spawn_send_worker, BroadcastQueue, Plan, SubscriptionGaugeStats, SubscriptionManager,
+};
 use super::query::compile_query_with_hashes;
 use super::tx::DeltaTx;
 use super::{collect_table_update, TableUpdateType};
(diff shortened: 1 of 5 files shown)


crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -323,16 +337,19 @@ impl ModuleSubscriptions {
         // thread it's possible for messages to get sent to the client out of order. If you do
         // spawn in another thread messages will need to be buffered until the state is sent out
         // on the wire
```
gefjon: This comment seems out-of-date.

gefjon (commented):
> I don't have a good idea for how to prevent people from accidentally calling sender.send_message when they should be using the broadcast queue, so some bugs might creep back in over time.

A doc comment on `send_message` instructing readers to use the broadcast queue instead seems like a good place to start.

gefjon (approved):

PR #2837, opened 2025-06-05
Bump versions to 1.2.0

bfops:
# Description of Changes

It turns out that we have merged at least one PR that requires a minor version bump (https://github.com/clockworklabs/SpacetimeDB/pull/2711), but there are some others that maybe deserve that as well:

- https://github.com/clockworklabs/SpacetimeDB/pull/2725
- https://github.com/clockworklabs/SpacetimeDB/pull/2678

As such, I'm retroactively bumping the version number.

None, just a version bump

1

# Testing

None beyond CI

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -910,7 +910,7 @@ dependencies = [
 
 [[package]]
 name = "connect_disconnect_client"
-version = "1.1.3"
+version = "1.2.0"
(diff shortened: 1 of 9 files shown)


cloutiertyler (approved):
This is fine, as a reminder I would like to review #2678's API before be make it stable.

PR #2242, opened 2025-02-10
Support RFC 3339 timestamp values in sql

joshua-spacetime:
# Description of Changes

Adds RFC 3339 formatted timestamps to sql. This only required updates to the type system.

Note this does not introduce any new optimizations. Timestamp columns can be indexed and so all current optimizations related to indexed columns still apply.

# API and ABI breaking changes

None

# Expected complexity level and risk

1.5

Just added timestamp parsing and expanded a typing constraint to allow timestamp values in comparison expressions.

# Testing

New typing tests

PR #2842, opened 2025-06-06
Allowed specifying timestamp literals in SQL

egormanga:
Currently, specifying a timestamp as an ISO-format string in SQL fails with:
`Error: Unexpected type: (expected) String != (__timestamp_micros_since_unix_epoch__: I64) (inferred)`

This PR allows specifying it (along with other, possibly future string literal-based types), as well as fixes the incorrect type order in the above error message (the inferred one should be from SQL, and the expected is the column SATS type).

* Related to: #2242

None.

### 1.

_Note for reviewers: this should NOT allow anything irrelevant, as `parse()` that we call still checks for the types to match._

# Testing

- [x] Manual test case: `INSERT`ing a timestamp column and `SELECT`ing `WITH` comparing it.


cloutiertyler:
@egormanga the diff here seems to be unrelated to the PR. I'm assuming it's been done on another branch?

egormanga:
It is related, since it allows trying to parse the timestamp value from a string. Have you even tried running it or at least read the description where it is explained? :\

I loved contributing to Spacetime at first, but at this point it seems just plain disappointing.

Issue #2670, opened 2025-04-25
Support `SumType`s/`SumValue`s in SQL queries (subscription & one-off)


cloutiertyler:
This needs design work to define the SATN representation for `SumValue`s.

PR #2843, opened 2025-06-07
Implemented value tuple literal in SQL

egormanga:
# Description of Changes

There's been no way to use product types like this:
```rust
struct Position<T: geo::CoordNum = Coord> {
	lat: T,
	lon: T,
}

#[table(name=point)]
struct Point {
	pos: Position,
}
```
because SQL queries containing tuples:
```sql
SELECT * FROM point WHERE pos = (0, 0);
```
would fail with:
`Error: Unsupported expression: (0, 0)`

This PR implements tuple support for `=` and `!=` operators (subject to further expansion).

None.

### 2.

# Testing

- [x] Manual testing using custom `struct` columns.
- [ ] Needs unit tests. I leave this open for contributions or as an exercise to the employees.

### DIFF
--- a/crates/expr/src/expr.rs
+++ b/crates/expr/src/expr.rs
@@ -322,6 +322,8 @@ pub enum Expr {
     LogOp(LogOp, Box<Expr>, Box<Expr>),
     /// A typed literal expression
     Value(AlgebraicValue, AlgebraicType),
+    /// A typed literal tuple expression
+    Tuple(Box<[AlgebraicValue]>, AlgebraicType),
     /// A field projection
     Field(FieldProject),
 }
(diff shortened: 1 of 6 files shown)


egormanga:
Possibly related to #2670?

cloutiertyler:
I think we would like to take a design pass on the syntax, especially because this PR is using tuple syntax for product values. We'd like to evaluate Postgres and ANSI syntax for this feature. I'm going to close this PR and create a ticket to track this feature.

egormanga:
How is ANSI or Postgres tuple syntax any different from the one this PR uses? I just went with whatever `sqlparser` considers a tuple, with no particular kind of design in mind.

PR #2844, opened 2025-06-07
Implemented SQL simple enum string literal parsing

egormanga:
# Description of Changes

Ditto #2843.
```rust
enum TestKind {
	OptionA,
	OptionB,
}

#[table(name=test)]
struct Test {
	kind: TestKind,
}
```
```sql
SELECT * FROM test WHERE kind = 'OptionB';
```

This PR implements simple enum support for = and != operators (subject to further expansion).

_Note: still doesn't work with [enum_stringify](https://docs.rs/enum_stringify/latest/enum_stringify) case conversion OOTB. No clue how to implement that yet._

None.

### 1.

# Testing

- [x] Manual testing using custom `enum` columns.
- [ ] Needs unit tests. I leave this open for contributions or as an exercise to the employees.

### DIFF
--- a/crates/expr/src/lib.rs
+++ b/crates/expr/src/lib.rs
@@ -158,16 +158,20 @@ pub(crate) fn type_expr(vars: &Relvars, expr: SqlExpr, expected: Option<&Algebra
 }
 
 /// Is this type compatible with this binary operator?
-fn op_supports_type(_op: BinOp, t: &AlgebraicType) -> bool {
-    t.is_bool()
-        || t.is_integer()
-        || t.is_float()
-        || t.is_string()
-        || t.is_bytes()
-        || t.is_identity()
-        || t.is_connection_id()
-        || t.is_timestamp()
-        || t.is_uuid()
+fn op_supports_type(op: BinOp, ty: &AlgebraicType) -> bool {
+    match (ty, op) {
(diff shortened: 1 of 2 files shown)


bfops:
Thank you for opening this! We will work toward getting this reviewed.

cloutiertyler:
This seems like a straightforward improvement. Going to get this one reviewed.

joshua-spacetime (commented):
I agree this is the right the syntax for simple enums. In addition to `SELECT` we should also include DML support.

cloutiertyler:
@egormanga would you be willing to implement DML support on this PR?

egormanga:
Does this not already work with DML though?

Pascualex:
@cloutiertyler pinging in case you missed egormanga's response

Pascualex:
@egormanga thanks for the contribution.

At the moment this is only for client-side queries, right? Meaning no RLS trait implementation for server-side queries. Something like:

```rust
ctx.from.test.filter(|t| t.kind.eq(TestKind::OptionB))
```

If I'm not mistaken it should be easy to add server-side query support if this works for client-side queries, but it could be left for a separate PR.

egormanga:
I haven't taken a single look at server side queries (nor their implementation) yet, and I'm not sure if this is even something not already possible. Meanwhile this PR focuses on the SQL syntax support.

PR #2695, opened 2025-05-02
Make `Timestamp` and `TimeDuration` implement `FilterableValue`

rafaelGuerreiro:
# Description of Changes

Making `Timestamp` and `TimeDuration` implement `FilterableValue`.

This resolves https://github.com/clockworklabs/SpacetimeDB/issues/2650

# API and ABI breaking changes
API change is additive in this case, I'm not marking this as unstable because the `FilterableValue` implementation is already stabilized.

# Expected complexity level and risk
Complexity 1.

# Testing

- [ ] 
- [ ]

PR #2845, opened 2025-06-07
Implemented `FilterableValue` for `Timestamp` & `TimeDuration`

egormanga:
Without this:
```rust
some_table.some_index().filter((ctx.sender, ctx.timestamp));
                        ------ ^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `IndexScanRangeBounds<(spacetimedb::Identity, spacetimedb::Timestamp), _>` is not implemented for `(spacetimedb::Identity, spacetimedb::Timestamp)`
```
Same goes for `.delete()`.

None.

### 1.

# Testing

- [x] Compiles.
- [x] Allows compiling the case.
- [ ] Why hasn't this already been implemented?

### DIFF
--- a/crates/lib/src/filterable_value.rs
+++ b/crates/lib/src/filterable_value.rs
@@ -1,4 +1,4 @@
-use crate::{ConnectionId, Identity};
+use crate::{ConnectionId, Identity, TimeDuration, Timestamp};
 use core::ops;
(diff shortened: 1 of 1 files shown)


egormanga:
Nevermind, just noticed #2695. It should get some moving though!

PR #2850, opened 2025-06-09
Add an HTTP route to get the current `Timestamp`

gefjon:
# Description of Changes

Under a new router, `/unstable`, as `/unstable/timestamp`.

Also add a smoketest that this route is reachable and returns a JSON-encoded `Timestamp`.

Introduces a new HTTP route, but marks it unstable.

1.

# Testing

- [x] Manual call with cURL.
- [x] Added smoketest for reachability and well-typedness. No validation for the value.


cloutiertyler (approved):
This LGTM

PR #2851, opened 2025-06-10
Reserve cores 0 and 1 for the OS

joshua-spacetime:
# Description of Changes

This patch makes it so that we do not pin any threads to core IDs 0 and 1, and instead reserve them for the OS. See the comment for more explanation.

None

1

A very simple change, but because we are taking cores away from the database we _may_ need to change the distribution of the remaining cores.

# Testing

- [X] Performance test

Before this change we saw the OS preempting reducer threads to handle interrupt requests, after this change - along with mapping these interrupt handlers to the reserved cores - we did not.

### DIFF
--- a/crates/core/src/startup.rs
+++ b/crates/core/src/startup.rs
@@ -1,5 +1,6 @@
 use core_affinity::CoreId;
 use crossbeam_queue::ArrayQueue;
+use itertools::Itertools;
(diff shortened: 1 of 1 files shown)


gefjon (approved):

egormanga:
Shouldn't this be handled on the OS level instead? What about the case when running [irqbalance](https://github.com/Irqbalance/irqbalance)?

PR #2853, opened 2025-06-10
Send one-off responses through the broadcast queue.

jsdt:
# Description of Changes

Use the broadcast queue for one-off responses.

This also moves the query execution into a blocking task. Previously it was blocking a main runtime thread.

1

# Testing

I'm not sure how much test coverage this has.

### DIFF
--- a/crates/client-api/src/routes/mod.rs
+++ b/crates/client-api/src/routes/mod.rs
@@ -12,6 +12,7 @@ mod internal;
 pub mod metrics;
 pub mod prometheus;
 pub mod subscribe;
+pub mod unstable;
 
 /// This API call is just designed to allow clients to determine whether or not they can
 /// establish a connection to SpacetimeDB. This API call doesn't actually do anything.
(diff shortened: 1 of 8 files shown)


crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -570,6 +570,18 @@ impl ModuleSubscriptions {
         Ok((plans, auth, scopeguard::ScopeGuard::into_inner(tx)))
     }
 
+    /// Send a message to a client connection.
+    /// This will eventually be sent by the send-worker.
+    /// This takes a TxId, because this should be called while still holding a lock on the database.
```
Centril:
```suggestion
    /// This takes a [`TxId`], because this should be called while still holding a lock on the database.
```

crates/core/src/host/module_host.rs
```diff
@@ -993,60 +995,106 @@ impl ModuleHost {
         )
     }
 
+    /// Execute a one-off query and send the results to the given client.
+    /// This only returns an error if there is a db-level problem.
+    /// An error with the query itself will be sent to the client.
```
gefjon, line 1000: Thanks for adding this comment!

gefjon (approved):

PR #2855, opened 2025-06-11
Prune more queries when evaluating subscription updates

joshua-spacetime:
# Description of Changes

If a table is subscribed to, and it joins with many different tables, this optimization will attempt to prune joins that it knows are empty.

None

3

A bug will most likely result in incorrect subscription results

# Testing

- [x] Internal tests
- [ ] Performance tests

### DIFF
--- a/crates/core/src/subscription/module_subscription_manager.rs
+++ b/crates/core/src/subscription/module_subscription_manager.rs
@@ -5,6 +5,7 @@ use crate::client::messages::{
     TransactionUpdateMessage,
 };
 use crate::client::{ClientConnectionSender, Protocol};
+use crate::db::datastore::locking_tx_datastore::state_view::StateView;
 use crate::error::DBError;
 use crate::host::module_host::{DatabaseTableUpdate, ModuleEvent, UpdatesRelValue};
 use crate::messages::websocket::{self as ws, TableUpdate};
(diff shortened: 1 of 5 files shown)


crates/core/src/subscription/module_subscription_manager.rs
```diff
@@ -1007,10 +1104,34 @@ impl SubscriptionManager {
...
+            // Ultimately no we will not.
+            // We may prune it for this row specifically,
+            // but we will eventually include it for the joining row.
+            tx.iter_by_col_eq(
+                edge.rhs_table,
+                edge.rhs_join_col,
+                &row.elements[edge.lhs_join_col.idx()],
+            )
+            // This read should always succeed, and it's a bug if it doesn't.
+            .unwrap()
```
mamcx:
```suggestion
            .expect("This read should always succeed, and it's a bug if it doesn't")
```

crates/core/src/subscription/module_subscription_manager.rs
```diff
@@ -1007,10 +1104,34 @@ impl SubscriptionManager {
...
+            .next()
+            .map(|row| {
+                // Similarly this read should always succeed
+                row.read_col(edge.rhs_col).unwrap()
```
mamcx:
```suggestion
                row.read_col(edge.rhs_col).expect("This read should always succeed, and it's a bug if it doesn't")
```

crates/subscription/src/lib.rs
```diff
@@ -285,6 +285,60 @@ impl std::fmt::Display for TableName {
...
+/// ...
+/// SELECT a.* FROM a JOIN b ON a.id = b.id WHERE b.x = n
+/// ```
+///
+/// Whenever `a` is updated, only the relevant queries are evaluated.
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub struct JoinEdge {
+    /// The [`TableId`] for `a`
+    pub lhs_table: TableId,
+    /// The [`TableId`] for `a`
```
mamcx:
```suggestion
    /// The [`TableId`] for `b`
```

mamcx (approved):
LGTM

PR #2857, opened 2025-06-12
Wrap `try_init_host` futures in `tokio::spawn` so they don't get cancelled

gefjon:
# Description of Changes

Comment pasted from within `watch_maybe_launch_module_host`:

> `try_init_host` is not cancel safe, as it will spawn other async tasks
> which hold a filesystem lock past when `try_init_host` returns or is cancelled.
> This means that, if `try_init_host` is cancelled, subsequent calls will fail.
>
> This is problematic because Axum will cancel its handler tasks if the client disconnects,
> and this method is called from Axum handlers, e.g. for the subscribe route.
> `tokio::spawn` a task to build the `Host` and install it in the `guard`,
> so that it will run to completion even if the caller goes away.
>
> Note that `tokio::spawn` only cancels its tasks when the runtime shuts down,
> at which point we won't be calling `try_init_host` again anyways.

This fixes (hopefully) the dreaded replay-loop bug, which we belive was caused by Axum dropping the `watch_maybe_launch_module_host` future after it acquired the `Lockfile` and leaked that file to a non-cancel-able async task, leading subsequent calls to fail due to the `Lockfile` still being held. Now, any call to `watch_maybe_launch_module_host` which enters `try_init_host` will run to completion and install the new `Host` into the `HostController`, even if the calling future is dropped or cancelled.

1: wrapping already-async stuff in `tokio::spawn` is a minimal change.

# Testing

- [x] Manual testing in a production-like cluster to see if this fixes the replay loop.

### DIFF
--- a/crates/core/src/host/host_controller.rs
+++ b/crates/core/src/host/host_controller.rs
@@ -76,6 +76,10 @@ pub type ProgramStorage = Arc<dyn ExternalStorage>;
 
 /// A host controller manages the lifecycle of spacetime databases and their
 /// associated modules.
+///
+/// This type is, and must remain, cheap to clone.
+/// All of its fields should either be [`Copy`], enclosed in an [`Arc`],
+/// or have some other fast [`Clone`] implementation.
 #[derive(Clone)]
 pub struct HostController {
     /// Map of all hosts managed by this controller,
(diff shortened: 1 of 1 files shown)


gefjon:
@Kasama reports that manual testing in a production-like deployment no longer showed the bug in question.

crates/core/src/host/host_controller.rs
```diff
@@ -253,10 +257,31 @@ impl HostController {
...
-        let host = self.try_init_host(database, replica_id).await?;
 
-        let rx = host.module.subscribe();
-        *guard = Some(host);
+        // `HostController::clone` is fast,
+        // as all of its fields are either `Copy` or wrapped in `Arc`.
+        let this = self.clone();
+
+        // `try_init_host` is not cancel safe, as it will spawn other async tasks
+        // which hold a filesystem lock past when `try_init_host` returns or is cancelled.
```
jsdt, line 266: Which task is causing this problem? If we could make that task get shut down if the host is cancelled/dropped, it would be safer than trusting that the caller of `try_init_host` doesn't cancel it.

jsdt (approved):
This looks safe to do

crates/core/src/host/host_controller.rs
gefjon: I have not yet determined that. Follow-up ticket at #2863 .

PR #2859, opened 2025-06-12
Always use identity when hashing queries from owners

joshua-spacetime:
# Description of Changes

Before this change, if the database owner and another client subscribed to the same query and that query had an RLS rule, both queries would get the same hash, despite not being semantically equivalent.

Now when hashing queries from the database owner, we always make sure to use their identity. This means the two queries in the above example will always hash to different values.

None

1

# Testing

- [x] Automated functional test


mamcx (approved):
LGTM

PR #2866, opened 2025-06-13
Move locking operations onto the module's thread

jsdt:
# Description of Changes

This branch moves a lot of code from blocking threads onto the module thread. A common pattern was using `asyncify` to spawn a blocking task, which would then grab a db-level lock and do some work.  This change introduces the `on_module_thread` function on `ModuleHost`, which runs a function on the module thread (which was previously only used for reducer calls). The code inside is largely unchanged, so there are still locks used to guard access to things, but there should be less contention on them.

This does limit some potential concurrency, since operations that only grab a shared lock are now serialized. In practice, this shouldn't be a big regression, since the locks we are using stop giving out shared locks as soon as one writer shows up.

The only real benefits to this PR are:
 1.  Queue wait time stats are more comprehensive (since they include subscription operations).
 2. We will have fewer blocking threads, which should help reduce context switching.

I think it moves toward a simplified concurrency model, where all module operations are sent to a single actor, but there are a lot more changes needed to get there.

2

# Testing

We have done some bot testing on a version of this, but unit testing it is not really feasible.


crates/core/src/host/module_host.rs
```diff
@@ -834,10 +839,38 @@ impl ModuleHost {
...
+    /// as that means we've somehow failed to delete from `st_client`.
+    /// We cannot meaningfully handle this.
+    /// Sometimes it just means that the database no longer exists, though, which is fine.
+    pub async fn call_identity_disconnected(
+        &self,
+        caller_identity: Identity,
+        caller_connection_id: ConnectionId,
+    ) -> Result<(), ReducerCallError> {
+        let me = self.clone();
+        self.call("call_identity_connected", move |inst| {
```
joshua-spacetime:
```suggestion
        self.call("call_identity_disconnected", move |inst| {
```

PR #2868, opened 2025-06-13
Use database identity not owner identity for metrics

joshua-spacetime:
# Description of Changes

Uses the database identity, not the owner identity, for subscription metrics. They are different. I incorrectly tagged these metrics with the owner identity in https://github.com/clockworklabs/SpacetimeDB/pull/2861.

None

0

# Testing


mamcx (approved):
LGTM

PR #2871, opened 2025-06-16
Fix an outdated docstring on `identity` and `try_identity` in the Rust SDK

gefjon:
# Description of Changes

Since we changed from `with_credentials` to `with_token`, the Rust client SDK no longer knows its `Identity` prior to `on_connect`, even when the connection is not anonymous.

N/a

1

# Testing

N/a


mamcx (approved):
LGTM

PR #2872, opened 2025-06-16
Lift restriction that multi-col indices only allow = constraints

### LINKED ISSUE #1317: Lift restriction that multi-col indices only allow `=` constraints
We had to disable this in https://github.com/clockworklabs/SpacetimeDB/pull/1316 but we should lift the restriction eventually for better query performance.

mamcx:
# Description of Changes

Closes #1317

> We had to disable this in https://github.com/clockworklabs/SpacetimeDB/pull/1316 but we should lift the restriction eventually for better query performance.

2

We optimize up to 3 columns in an index, and:
- When all the comparisons are`!=` is always a full table scan
- Multi-column indexes are only used if the query has a prefix match (ie all operators are `=`)
- Else are converted to a single column index scan on the leftmost column and a filter on the rest

# Testing

- [*] Expand the testing to verify the plans to all the ops `"=", ">", "<", ">=", "<=", !=`
- [*] Expand the testing to verify we return the same rows when executing the query with the above ops

### DIFF
--- a/crates/core/src/sql/compiler.rs
+++ b/crates/core/src/sql/compiler.rs
@@ -513,15 +513,14 @@ Seq Scan on test
         db.create_table_for_test("test", schema, indexes)?;
 
         let tx = begin_tx(&db);
-        // TODO: Need support for index range scans.
         expect_query(
             &tx,
             "select * from test where b > 2",
             expect![
                 r#"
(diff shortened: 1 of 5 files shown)


cloutiertyler:
Much too out of date and touches 1400+ files.

PR #2873, opened 2025-06-16
Relax conditions for query pruning

joshua-spacetime:
# Description of Changes

When pruning joins, both join columns do not have to be unique. If we have a set of row updates from a table, we only care that the join column from the opposing table is unique.

None

1

# Testing

- [x] Updated existing test


mamcx (approved):
LGTM

PR #2874, opened 2025-06-16
Auto-migrate: Allow adding new variants at the tail

Centril:
# Description of Changes

To the auto-migrate machinery, detection of the case where a sum type is extended with additional variants and where the sum type as a whole does not grow in size. This then results in a new datastore call `alter_table_row_type` which independently checks for "soundness" of using the new row type for the existing rows. This call, being a schema altering one, is handled, in terms of transactionality, the same way that e.g., `create_index` is, by adding a pending schema change.

None

3 - this is fairly sensitive logic that is critical to get right so that we don't cause UB and problems down the line.

# Testing

- The datastore functionality is independently tested to see that it works and for transactionality.
- The datastore API `alter_table_access` was untested and is now tested also in terms of transactionality.
- The auto-migration logic is tested comprehensively both for success and failure.
- A smoketest is amended to have an integration test as well.

### DIFF
--- a/crates/schema/src/auto_migrate.rs
+++ b/crates/schema/src/auto_migrate.rs
@@ -171,6 +171,18 @@ pub enum AutoMigrateError {
     ChangeWithinColumnTypeFewerVariants(ChangeColumnTypeParts),
 
     #[error(
+        "Changing the type of column {} in table {} from {:?} to {:?}, with a renamed variant, requires a manual migration",
+        .0.column, .0.table, .0.type1, .0.type2
+    )]
+    ChangeColumnTypeRenamedVariant(ChangeColumnTypeParts),
+
+    #[error(
+        "Changing the type of column {} in table {} from {:?} to {:?}, with a renamed variant, requires a manual migration",
+        .0.column, .0.table, .0.type1, .0.type2
+    )]
+    ChangeWithinColumnTypeRenamedVariant(ChangeColumnTypeParts),
+
+    #[error(
         "Changing the type of column {} in table {} from {:?} to {:?}, requires a manual migration, due to size mismatch",
         .0.column, .0.table, .0.type1, .0.type2
     )]
@@ -205,6 +217,18 @@ pub enum AutoMigrateError {
         .0.column, .0.table, .0.type1, .0.type2
     )]
     ChangeWithinColumnTypeFewerFields(ChangeColumnTypeParts),
+
+    #[error(
+        "Changing the type of column {} in table {} from {:?} to {:?}, with a renamed field, requires a manual migration",
+        .0.column, .0.table, .0.type1, .0.type2
+    )]
+    ChangeColumnTypeRenamedField(ChangeColumnTypeParts),
+
+    #[error(
+        "Changing the type of column {} in table {} from {:?} to {:?}, with a renamed field, requires a manual migration",
+        .0.column, .0.table, .0.type1, .0.type2
+    )]
+    ChangeWithinColumnTypeRenamedField(ChangeColumnTypeParts),
 
     #[error("Adding a unique constraint {constraint} requires a manual migration")]
     AddUniqueConstraint { constraint: Box<str> },
@@ -459,12 +483,22 @@ fn ensure_old_ty_upgradable_to_new(
 
             // The variants in `old_ty` must be upgradable to those in `old_ty`.
             // Strict equality is *not* imposed in the prefix!
-            // We also don't care about variant names, as they don't impact BFLATN.
             let prefix_ok = old_vars
                 .iter()
                 .zip(new_vars)
-                .map(|(o, n)| (&o.algebraic_type, &n.algebraic_type))
-                .map(ensure)
+                .map(|(o, n)| {
+                    // Ensure type compatibility.
+                    let res_ty = ensure((&o.algebraic_type, &n.algebraic_type));
+                    // Ensure name doesn't change.
+                    let res_name = if o.name() == n.name() {
+                        Ok(())
+                    } else if within {
+                        Err(ChangeWithinColumnTypeRenamedVariant(parts_for_error()).into())
+                    } else {
+                        Err(ChangeColumnTypeRenamedVariant(parts_for_error()).into())
+                    };
+                    (res_ty, res_name).combine_errors().map(|(c, ())| c)
+                })
                 .collect_all_errors::<Any>();
 
             // The old and the new sum types must have matching layout sizes and alignments.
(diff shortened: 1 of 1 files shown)


crates/core/src/db/datastore/locking_tx_datastore/mut_tx.rs
```diff
@@ -429,6 +430,33 @@ impl MutTxId {
         Ok(())
     }
 
+    /// Change the row type of the table identified by `table_id.
+    ///
+    /// In practice, this should not error,
+    /// as the update machinery should disallow any incompatible change.
+    /// However, or redundancy and internal soundness of the datastore,
```
mamcx:
```suggestion
    /// However, for redundancy and internal soundness of the datastore,
```

crates/table/src/table.rs
```diff
@@ -260,19 +261,122 @@ pub enum ReadViaBsatnError {
...
+    /// Returns an error if the new list of column is incompatible with the old.
+    ///
+    /// # Safety
+    ///
+    /// If `!validate`, the caller must ensure that `column_schemas`
+    /// forms a row type and layout that is compatible with the rows existing in `self`.
+    pub unsafe fn change_columns_to(
+        &mut self,
+        column_schemas: Vec<ColumnSchema>,
+        validate: bool,
```
mamcx: I think that this `bool` should be instead a enum to make more clear to the callers what `true, false` mean. (or split in 2 methods?)

Centril: Split them in two.

crates/core/src/db/datastore/locking_tx_datastore/committed_state.rs
```diff
@@ -652,57 +652,59 @@ impl CommittedState {
     ) -> Option<()> {
         use PendingSchemaChange::*;
         match change {
-            // An index was removed. Add it back.
```
gefjon: Why delete all these comments?

crates/core/src/db/datastore/locking_tx_datastore/tx_state.rs
```diff
@@ -102,6 +102,9 @@ pub(super) enum PendingSchemaChange {
     /// The access of the table with [`TableId`] was changed.
     /// The old access was stored.
     TableAlterAccess(TableId, StAccess),
+    /// The row type of the table with [`TableId`] was changed.
+    /// The old column schemas was stored.
+    TableAlterRowType(TableId, Vec<ColumnSchema>),
```
gefjon: Can you extend this with a comment that only non-representational changes are allowed here?

crates/core/src/db/datastore/locking_tx_datastore/committed_state.rs
```diff
@@ -652,57 +652,59 @@ impl CommittedState {
...
                 // Instead, there will be separate pending schema changes like `IndexAdded`.
                 self.tables.remove(&table_id);
             }
-            // A table's access was changed. Change back to the old one.
             TableAlterAccess(table_id, access) => {
                 let table = self.tables.get_mut(&table_id)?;
                 table.with_mut_schema(|s| s.table_access = access);
             }
-            // A constraint was removed. Add it back.
+            TableAlterRowType(table_id, column_schemas) => {
```
gefjon, line 692: This branch could really use a comment like the ones you deleted. It should specify that the row type has changed but the row representation hasn't, so we don't have to rewrite the rows.

crates/core/src/db/datastore/locking_tx_datastore/mut_tx.rs
```diff
@@ -429,6 +430,38 @@ impl MutTxId {
         Ok(())
     }
 
+    /// Change the row type of the table identified by `table_id.
```
gefjon:
```suggestion
    /// Change the row type of the table identified by `table_id`.
```

crates/core/src/db/datastore/locking_tx_datastore/mut_tx.rs
```diff
@@ -429,6 +430,38 @@ impl MutTxId {
...
+    ///
+    /// In practice, this should not error,
+    /// as the update machinery should disallow any incompatible change.
+    /// However, for redundancy and internal soundness of the datastore,
+    /// the compatibility is also checked here.
```
gefjon, line 438: This could also use either a description of what counts as a compatible change, or a ref/link to such a description.

crates/core/src/db/datastore/locking_tx_datastore/committed_state.rs
```diff
@@ -652,57 +652,59 @@ impl CommittedState {
...
+                let table = self.tables.get_mut(&table_id)?;
+                // SAFETY: `validate = false`.
```
gefjon:
```suggestion
                // SAFETY: `validate = false`.
                // No need to validate the "new" row type,
                // as it was the row type prior to the change which we're rolling back.
```

crates/schema/src/auto_migrate.rs
```diff
@@ -329,17 +392,149 @@ fn auto_migrate_table<'def>(plan: &mut AutoMigratePlan<'def>, old: &'def TableDe
...
+        Any(iter.into_iter().any(|Any(x)| x))
+    }
+}
+
+impl Add for Any {
+    type Output = Self;
+    fn add(self, rhs: Self) -> Self::Output {
+        Self(self.0 || rhs.0)
+    }
+}
```
gefjon: :thinking: why not use `ops::Or` here?

crates/schema/src/auto_migrate.rs
```diff
@@ -329,17 +392,149 @@ fn auto_migrate_table<'def>(plan: &mut AutoMigratePlan<'def>, old: &'def TableDe
...
+            let var_lens_ok = match old_vars.len().cmp(&new_vars.len()) {
+                Ordering::Less => Ok(Any(true)),
+                Ordering::Equal => Ok(Any(false)),
+                Ordering::Greater if within => Err(ChangeWithinColumnTypeFewerVariants(parts_for_error()).into()),
+                Ordering::Greater => Err(ChangeColumnTypeFewerVariants(parts_for_error()).into()),
+            };
+
+            // The variants in `old_ty` must be upgradable to those in `old_ty`.
+            // Strict equality is *not* imposed in the prefix!
+            // We also don't care about variant names, as they don't impact BFLATN.
```
gefjon: I don't love this. It seems likely to lead to subtle bugs. Imagine I have a simple/C-style enum:

```rust
enum Permissions {
  Owner,
  Admin,
  User,
  Banned,
}
```

and I store a bunch of these in my user table. Then I decide that it would actually be better to reorder the variants so that I can use integer comparison operators on permissions, with `Owner > Admin > User > Banned`. So I rewrite the enum to be:

```rust
enum Permissions {
  Banned,
  User,
  Admin,
  Owner,
}
```

It is clearly not my intention to rename the variants here, and it would be disastrous if SpacetimeDB applied this change without rewriting my user rows.

crates/schema/src/auto_migrate.rs
```diff
@@ -329,17 +392,149 @@ fn auto_migrate_table<'def>(plan: &mut AutoMigratePlan<'def>, old: &'def TableDe
...
+            };
+
+            let (len_changed, prefix_changed, ..) = (var_lens_ok, prefix_ok, size_ok, align_ok).combine_errors()?;
+            Ok(len_changed + prefix_changed)
+        }
+
+        // For products,
+        // we need to check each field's upgradability due to sums,
+        // and there must be as many fields.
+        // Note that we don't care about field names.
```
gefjon, line 531: I also don't feel great about this choice, though I can't come up with a disastrous case as easily as I can for renaming enum variants.

crates/schema/src/auto_migrate.rs
Centril: Well `+` I would say is a more conventional symbol for monoids, but I also don't mind using `BitOr`.

gefjon: I have no strong opinion, but I think that non-Haskell-ers will be confused by using `+` for an operation which does not behave like integer or floating-point addition.

gefjon (approved):
Beautiful, thanks!

PR #2876, opened 2025-06-17
Track metadata updates for unsubscribes

joshua-spacetime:
# Description of Changes

We already track the time it takes to update subscription metadata for subscribes. Here we do the same for unsubscribes.

None

0

# Testing

Tested manually as part of a performance test


mamcx (approved):
LGTM

PR #2877, opened 2025-06-17
Record the size of a client's outgoing message queue on drop

joshua-spacetime:
# Description of Changes

If a client's send queue reaches `CLIENT_CHANNEL_CAPACITY`, we drop the client connection by aborting the task that sends and receives messages from the client. If this abort happens before the task pulls the last messages off the queue, the outgoing queue length metric will never be decremented appropriately.

To fix this, we wrap the message receiver with the metric and decrement the metric when we drop the receiver. The receiver's `Drop` method will be called when the future is aborted which will now decrement the metric appropriately.

None

1

# Testing

TBD

### DIFF
--- a/crates/core/src/client/client_connection.rs
+++ b/crates/core/src/client/client_connection.rs
@@ -1,4 +1,4 @@
-use std::ops::Deref;
+use std::ops::{Deref, DerefMut};
 use std::sync::atomic::{AtomicBool, Ordering::Relaxed};
 use std::sync::Arc;
 use std::time::Instant;
@@ -132,7 +132,7 @@ pub enum ClientSendError {
 }
 
 impl ClientConnectionSender {
-    pub fn dummy_with_channel(id: ClientActorId, config: ClientConfig) -> (Self, mpsc::Receiver<SerializableMessage>) {
+    pub fn dummy_with_channel(id: ClientActorId, config: ClientConfig) -> (Self, MeteredReceiver<SerializableMessage>) {
         let (sendtx, rx) = mpsc::channel(1);
         // just make something up, it doesn't need to be attached to a real task
         let abort_handle = match tokio::runtime::Handle::try_current() {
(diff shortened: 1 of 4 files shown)


crates/core/src/client/client_connection.rs
```diff
@@ -257,6 +258,48 @@ impl DataMessage {
     }
 }
 
+/// Wraps the receiving end of a channel with a gauge for tracking the size of the channel.
+/// We subtract the size of the channel from the gauge on drop to avoid leaking the metric.
+pub struct MeteredReceiver<T> {
+    inner: mpsc::Receiver<T>,
+    gauge: Option<IntGauge>,
+}
```
gefjon: I'd be happier with this abstraction if it encapsulated all of the metric reading and writing, not just the behavior on drop.

crates/core/src/client/client_connection.rs
```diff
@@ -257,6 +258,48 @@ impl DataMessage {
...
+
+impl<T> Drop for MeteredReceiver<T> {
+    fn drop(&mut self) {
+        // Record the number of elements still in the channel on drop
+        if let Some(gauge) = &self.gauge {
+            gauge.sub(self.inner.len() as _);
+        }
+    }
+}
```
gefjon, line 370: Is this now going to double-decrement when "normally" exiting the `subscribe` loop due to a close frame?

crates/core/src/client/client_connection.rs
joshua-spacetime: We can certainly have it handle all decrements of the metric. I'm not sure if it makes sense to have it handle the increments since those happen in a separate task when we push a message into the sender side of the channel.

crates/core/src/client/client_connection.rs
joshua-spacetime: Actually yes, because we don't drain the buffer in that case.

joshua-spacetime: Fixed

PR #2881, opened 2025-06-20
Fix two places we grab blocking locks on tokio runtime worker threads

jsdt:
# Description of Changes

This fixes two places where we can block threads in the main tokio multithread runtime, by making sure the blocking operations happen in `spawn_blocking` tasks.

1

# Testing

Not really tested. This shouldn't change logic - it should just move some things into blocking tasks.

### DIFF
--- a/crates/core/src/host/host_controller.rs
+++ b/crates/core/src/host/host_controller.rs
@@ -1063,18 +1063,19 @@ async fn metric_reporter(replica_ctx: Arc<ReplicaContext>) {
         .with_label_values(&replica_ctx.database_identity);
 
     loop {
-        // We spawn a blocking task here because this grabs blocking locks.
-        {
-            let ctx = replica_ctx.clone();
-            let _ = tokio::task::spawn_blocking(move || ctx.update_gauges()).await;
-        }
         let ctx = replica_ctx.clone();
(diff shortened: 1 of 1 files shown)


crates/core/src/host/host_controller.rs
```diff
@@ -1063,8 +1063,13 @@ async fn metric_reporter(replica_ctx: Arc<ReplicaContext>) {
...
     loop {
-        let disk_usage = tokio::task::block_in_place(|| replica_ctx.total_disk_usage());
-        replica_ctx.update_gauges();
+        // We spawn a blocking task here because this grabs blocking locks.
+        {
+            let ctx = replica_ctx.clone();
+            let _ = tokio::task::spawn_blocking(move || ctx.update_gauges()).await;
+        }
+        let ctx = replica_ctx.clone();
+        let disk_usage = tokio::task::block_in_place(move || ctx.total_disk_usage());
```
gefjon: This call to `block_in_place` is suspicious to me. `total_disk_usage` appears to also acquire at least one blocking lock, within `DatabaseLogger::size`. It also performs some blocking disk operations, e.g. a `stat` (via `File::metadata`) within both `DatabaseLogger::size` and `Fs::size_on_disk`. Can we move this call up and into the `update_gauges` blocking task?

gefjon (approved):

PR #2888, opened 2025-06-23
Only do 'failed to send update message' once per client

Centril:
# Description of Changes

See the tin.

None

2?


Centril:
This aint useful anymore.

PR #2892, opened 2025-06-23
Implemented `OUTER JOIN` support

egormanga:
# Description of Changes

Previously, only `INNER` and `CROSS` joins were possible.
This adds support for `LEFT OUTER` joins with certain limitations (basically, same as for `INNER`).

Doesn't play so well with RLS currently, though (I assume it's something to do with query optimization for similar joins, or that RLS discards already right-matched rows without substituting them with null-matched instead).

Non-present column values come in as `()` units (for null) instead of the expected type, so it's like everything suddenly became `Optional`. This might require some additional handling on the clients.

Probably won't work with the PG wire protocol for the same reason, although the fix for that will be server-side, as the protocol itself is fine with nulls.

### 3.

# Testing

- [x] Manual testing
- [ ] Needs unit tests


bfops:
Thank you for opening this! Some initial questions while we work on getting a reviewer for this:
1. Is this blocking a particular use case for you?
2. When you say "Doesn't play so well with RLS", what exactly do you mean? What incorrect behavior do you see?
3. What kind of manual testing did you perform? Are updates delivered incrementally to clients?

egormanga:
@bfops Thanks for the quick response! Firstly, I'd like to use this opportunity to ask for you to notice my other PRs as well, if possible. Most of them are there for a solid reason, too.

1. Yes, this is a major blocker for multiple of my cases, because a 3-way joining query with a 2-3-way joining RLS for each of the joined tables causes major slowdown due to the necessity of using cross product joins to emulate outer joins.
2. First of all, I only tested it with my earlier RLS patch applied so far, due to the complexity of the cases required to properly test it (I didn't have time to write synthetic tests for it).
   With this in mind: RLS seems to discard extra lhs rows that matched the non-visible rhs rows, not giving them a chance to not match rhs (i.e. match against null rows) instead. This results in less rows overall being reported than without the patch _(albeit in my particular current case I may trade off correctness for multiple magnitudes' speed increase)_.
4. I only tested plain queries for correctness (e.g. seeing partially matched rows with outer vs omission of such rows with inner) and heavy complex RLS-enabled queries for drastic speed improvement and lack of duplicate rows (those were from outer join emulation via cross joins, for obvious reasons).

~There's also currently a slight issue with the resulting format, such that the official CLI tool fails to parse nulls (`()` units) in place of another expected column type~ _(see below)_. API itself works fine, though. I have not tested this with client SDKs at all, because I'm not using those in the current project at the moment.

bfops:
Thank you for all the helpful info! We're still waiting on some reviewers for this part of the codebase to become available - right now they're all hands on deck with supporting the BitCraft launch, but all of your PRs are on our radar! I think they will probably have more thoughts about making sure we get corresponding support added in the other parts of the codebase (e.g. CLI, client SDKs)

egormanga:
Update: the CLI seems to handle nulls just fine now.

egormanga:
`execution/src/iter.rs` has been removed in https://github.com/clockworklabs/SpacetimeDB/pull/3524#discussion_r2472063046.

I did rebase discarding my changes to it, but haven't tested it this way yet.

cloutiertyler:
This PR is now out of date and also is a significant API change that I think would require some additional design work before we accepted. e.g. Probably NULL should be represented by `None` rather than `Unit` (although those ideas are related). Also I don't think we want every data type to be nullable. I'm going to close this and create an issue to track.

egormanga:
You do you. My stance is that handling contributions this gatekeep-ey way you'll end up with none of those.

Open Source is meant to be community-focused (& -built), yet you're trying to balance between that and proprietary, which _is_ admittedly tricky. Although, discarding any complex enough PR with an excuse of a “design pass” in a separate issue won't do good _for an external contribution_. It **might** be how the things are/should be done internally, but community members do this for free and in their free time, meaning there's no such thing as _disapproval_ (only an **improvement suggestion** with call for help from others).

By my book, there shouldn't be _this much_ discussion movement, eroding the source of truth and inducing unnecessary mental load for those only willingly involved, if you want your commercial product to benefit from free contributions. Otherwise, those will probably just move either to a **hard** community-maintained **fork** with less bureaucracy, or elsewhere (some other ‘…DB’).

After all, a ‘pull request’ entity is no more than a glorified issue containing a patch with a diff viewer slapped beside.

cloutiertyler:
> Open Source is meant to be community-focused (& -built), yet you're trying to balance between that and proprietary, which is admittedly tricky. Although, discarding any complex enough PR with an excuse of a “design pass” in a separate issue won't do good for an external contribution. It might be how the things are/should be done internally, but community members do this for free and in their free time, meaning there's no such thing as disapproval (only an improvement suggestion with call for help from others).

This has nothing to do with proprietary vs open source IMO. If SpacetimeDB were fully open source, the scrutiny on the design of features would be no less important. API surface changes like these can have a ripple effect on other features that we may want to implement the future. It's my responsibility to ensure that the API of SpacetimeDB stays clean and wart-free.

> By my book, there shouldn't be this much discussion movement, eroding the source of truth and inducing unnecessary mental load for those only willingly involved, if you want your commercial product to benefit from free contributions.

I'm unsure what you're saying here. It's important that we discuss API changes before we accept them for the benefit of our users. What source of truth is being eroded?

> Otherwise, those will probably just move either to a hard community-maintained fork with less bureaucracy, or elsewhere (some other ‘…DB’).

The beauty of open source licenses, is that people are welcome to maintain their own forks. I think a relevant (and excellent) talk by Bryan Cantrill is relevant: https://www.youtube.com/watch?v=anXKAcrJ5Wc

> After all, a ‘pull request’ entity is no more than a glorified issue containing a patch with a diff viewer slapped beside.

Yes, it is, and we appreciate it when community members spend their precious time to create them. It's important that we take community issues very seriously, but that does not mean that we should always accept their solutions, particularly when those solutions impact API surfaces that could have long-term effects.

I really appreciate the time you're putting into this. I really do. But please understand my responsibility as well.

Internally we have a proposal process, much like https://github.com/swiftlang/swift-evolution. I would be potentially open to opening that process up for more design related discussion with the community.

egormanga:
Now _that_ is something! I'd be happy to streamline my proposals in a defined process. That's also what I mean by the truth source.

Otherwise, I think the discussion is better kept in original PRs, allowing the author to reiterate and improve on it without cluttering the repo _and_ without losing useful threads such as review discussions.

PR #2893, opened 2025-06-24
Track disconnects initiated by the client

joshua-spacetime:
# Description of Changes

When a client connection ends, we want to know who initiated the disconnect - the client or the server.

None

0

# Testing

### DIFF
--- a/crates/core/src/worker_metrics/mod.rs
+++ b/crates/core/src/worker_metrics/mod.rs
@@ -26,7 +26,7 @@ metrics_group!(
         pub ws_clients_aborted: IntGaugeVec,
 
         #[name = spacetime_worker_ws_clients_closed_connection]
-        #[help = "Number of ws connections closed by the client"]
+        #[help = "Number of ws client connections closed by the client as opposed to being termiated by the server"]
         #[labels(database_identity: Identity)]
         pub ws_clients_closed_connection: IntGaugeVec,
 


crates/client-api/src/routes/subscribe.rs
```diff
@@ -431,8 +431,15 @@ async fn ws_client_actor_inner(
 
                 // if this is the closed-by-them case, let the ClientConnectionSenders know now.
                 sendrx.close();
-                closed = true;
                 log::trace!("Close frame {:?}", close_frame);
+                if !closed {
+                    // This is the client telling us they want to close.
+                    WORKER_METRICS
+                        .ws_clients_closed_connection
+                        .with_label_values(&addr)
```
Centril, line 439: Let's cache this in `ModuleMetrics` or some such.

crates/core/src/worker_metrics/mod.rs
```diff
@@ -21,10 +21,15 @@ metrics_group!(
         pub ws_clients_spawned: IntGaugeVec,
 
         #[name = spacetime_worker_ws_clients_aborted]
-        #[help = "Number of ws client connections aborted"]
+        #[help = "Number of ws client connections aborted by either the server or the client"]
         #[labels(database_identity: Identity)]
         pub ws_clients_aborted: IntGaugeVec,
 
+        #[name = spacetime_worker_ws_clients_closed_connection]
+        #[help = "Number of ws connections closed by the client"]
```
Centril:
```suggestion
        #[help = "Number of ws client connections closed by the client"]
```
(Consistency with the above `help`.)

crates/client-api/src/routes/subscribe.rs
joshua-spacetime: We technically don't need to cache it in this particular case because the client connection task will only ever call this once during its lifetime.

PR #2899, opened 2025-06-25
Filter out dropped clients in the send worker

joshua-spacetime:
# Description of Changes

Moves the filtering out of dropped clients during subscription evaluation from inside of `eval_updates_sequential`, where we hold the transaction lock, to the send worker, where the lock has been released.

None

0

# Testing

- [x] cpu profile

### DIFF
--- a/crates/core/src/client/client_connection.rs
+++ b/crates/core/src/client/client_connection.rs
@@ -160,9 +160,7 @@ impl ClientConnectionSender {
     }
 
     pub fn is_cancelled(&self) -> bool {
-        // We check this when processing subscription updates to exit early for dropped clients.
-        // We use `Acquire` here because we prioritize early exit over cheaper loads.
-        self.cancelled.load(Ordering::Acquire)
+        self.cancelled.load(Ordering::Relaxed)
     }
 
     /// Send a message to the client. For data-related messages, you should probably use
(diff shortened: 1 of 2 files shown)


crates/core/src/client/client_connection.rs
```diff
@@ -158,6 +159,10 @@ impl ClientConnectionSender {
         Self::dummy_with_channel(id, config).0
     }
 
+    pub fn is_cancelled(&self) -> bool {
+        self.cancelled.load(Ordering::Acquire)
```
Centril: Here the ordering is `Acquire` but below it is `Relaxed`. Would be good to explain why.

joshua-spacetime: Are you referring to the `store`?

Centril: And the load on line 173.

crates/core/src/client/client_connection.rs
```diff
@@ -158,6 +159,12 @@ impl ClientConnectionSender {
         Self::dummy_with_channel(id, config).0
     }
 
+    pub fn is_cancelled(&self) -> bool {
+        // We check this when processing subscription updates to exit early for dropped clients.
+        // We use `Acquire` here because we prioritize early exit over cheaper loads.
```
gefjon: I am not convinced by this reasoning. My understanding of the memory model is that `Ordering::Acquire` doesn't necessarily mean writes to a particular atomic place propagate between cores any faster, it just imposes an additional read memory barrier on other non-atomic places. See https://doc.rust-lang.org/std/sync/atomic/enum.Ordering.html .

joshua-spacetime: I'm saying: We use Acquire because if we use Relaxed, we could potentially process a subscription for a dropped client and that would be worse than paying the minuscule price of the Acquire barrier.

gefjon: And I'm saying, I don't think that `Acquire` behaves any differently from `Relaxed` in that sense.

gefjon: I think that what you want is a `fence(Release)` after the write, and a `fence(Acquire)` before the read. I believe that `load(Acquire)` semantically puts the fence after the load, and `store(Release)` semantically puts the fence before the write.

joshua-spacetime: This certainly isn't worth holding up this PR. I'm fine making it Relaxed.

PR #2901, opened 2025-06-25
`messages::serialize`: use 4 KiB for tx update compression threshold

Centril:
# Description of Changes

Set the compression threshold to 4KiB for transaction updates, keeping the lower 1KiB for initial subs.

None

1

# Testing

Needs to be perf tested manually.


joshua-spacetime:
Closing this for now, but we can always revisit in the future.

PR #2906, opened 2025-06-26
client-api: Rewrite websocket loop

kim:
Split the websocket stream into send and receive halves and spawns a
new tokio task to handle the sending. Also move message serialization +
compression to a blocking task if the message appears to be large.

This addresses two issues:

1. The `select!` loop is not blocked on sending messages, and can thus
   react to auxiliary events. Namely, when a module exits, we want to
   terminate the connection as soon as possible in order to release any
   database handles.

2. Large outgoing messages should not occupy tokio worker threads, in
   particular when there are a large number of clients receiving large
   intial updates.

---

**EDIT:** This patch started out to address the above issues, but evolved into a rewrite as more issues where discovered. Namely:

- logic is split into multiple functions that can be tested in isolation
- remove unbounded ingress buffer in favor of no buffer (i.e. applying backpressure to clients in case of slow processing)
- use an idle timer on any received packet, instead of waiting for pongs that are potentially end-of-line
- time-bound waiting for the close handshake to complete (server-initiated close)
- crucially, do not evaluate outstanding reducer call messages when a close handshake was already initiated
- ensure the connection is dropped when there is an error sending or receiving
- ensure all tasks exit when the loop exits

Because there is no inbound buffering, the queue length metric will not be updated.

4 - The state transitions remain hard to follow.

# Testing

- [x] Ran a stress test with many clients and large initial updates,
      and observed no hangs / delays (which I did before this patch).
      In reconnection scenarios, all clients where disconnected timely, but
      could reconnect almost immediately.
- [x] Added unit-level tests


crates/client-api/src/routes/subscribe.rs
```diff
@@ -444,7 +406,125 @@ async fn ws_client_actor_inner(
...
+        database_identity,
+        workload,
+        num_rows,
+        start_serialize.elapsed(),
+        &msg_data,
+    );
+
+    let res = ws.feed(datamsg_to_wsmsg(msg_data)).await;
+    // TODO: Reclaim seems to never succeed. What changed?
+    let buf = msg_alloc.try_reclaim().unwrap_or_else(|| SerializeBuffer::new(config));
```
kim, line 1061: @Centril This is puzzling me. The only difference seems to be that `ws` is of type `SplitSink` and not `WebSocketStream`. It should still have copied the data by the time `.await` returns, if that is what the underlying `WebSocketStream` is doing.

gefjon (approved):
I'd like to figure out what's going on with the `SerializeBuffer` and fix it before merging, but otherwise this looks good to me.

crates/client-api/src/routes/subscribe.rs
kim: I think I get it now: `SplitSink::start_send` just stores the item, we'll need to `poll_ready` to have it call `start_send` on the inner sink.

kim:
Updated to:

- Reclaim the serialize buffer
- Not send any more data after sending a `Close` frame (as mandated by the RFC)

I think that we should also clear the message queue and cancel outstanding execution futures in the latter case, but that can be left to a future change.

crates/client-api/src/routes/subscribe.rs
```diff
@@ -444,7 +410,155 @@ async fn ws_client_actor_inner(
...
+async fn send_message(
+    ws: &mut SplitSink<WebSocketStream, WsMessage>,
+    database_identity: &Identity,
+    config: ClientConfig,
+    serialize_buf: SerializeBuffer,
+    metrics_metadata: Option<(WorkloadType, usize)>,
+    message: impl ToProtocol<Encoded = SwitchedServerMessage> + Send + 'static,
+) -> (SerializeBuffer, Result<(), WsError>) {
+    let (workload, num_rows) = metrics_metadata.unzip();
+    let start_serialize = Instant::now();
```
jsdt: It would be nice to time serialization inside the blocking task, since the time spent switching to a blocking thread can be significant (especially if we use up all of our blocking threads).

crates/client-api/src/routes/subscribe.rs
```diff
@@ -444,7 +410,155 @@ async fn ws_client_actor_inner(
...
+    // Move large messages to a blocking thread,
+    // as serialization and compression can take a long time.
+    // The threshold of 1024 rows is arbitrary, and may need to be refined.
+    let (msg_alloc, msg_data) = if num_rows.is_some_and(|n| n > 1024) {
+        asyncify(move || serialize(serialize_buf, message, config)).await
```
jsdt: Adding a blocking task here feels risky. I'd like to remove blocking tasks generally, and I think this would be the first place where the number of blocking tasks isn't tied to the number of requests per second.

What do you think about starting by adding the timing metric, and maybe a warning log message any time that serialization takes longer than some threshold?

An alternative would be sending these to rayon instead of blocking threads, which would limit the number of threads working on serialization, and put the work on pinned cores.

crates/client-api/src/routes/subscribe.rs
```diff
@@ -352,11 +329,9 @@ async fn ws_client_actor_inner(
             _ = liveness_check_interval.tick() => {
                 // If we received a pong at some point, send a fresh ping.
                 if mem::take(&mut got_pong) {
-                    // Send a ping message while continuing to poll the `handle_queue`,
-                    // to avoid deadlocks or delays due to enqueued futures holding resources.
-                    if let Err(e) = also_poll(ws.send(WsMessage::Ping(Bytes::new())), make_progress(&mut current_message)).await {
-                        log::warn!("error sending ping: {e:#}");
-                    }
+                    // If the sender is already gone,
+                    // we'll time out the connection eventually.
```
jsdt: Why not break from the loop here instead of waiting for a timeout?

crates/client-api/src/routes/subscribe.rs
```diff
@@ -381,34 +356,22 @@ async fn ws_client_actor_inner(
...
-                        // At this point,
-                        // the underlying allocation of `msg_data` should have a single referent
-                        // and this should be `msg_alloc`.
-                        // We can put this back into our pool.
-                        msg_buffer = msg_alloc
-                            .try_reclaim()
-                            .expect("should have a unique referent to `msg_alloc`");
-
+                        // Ignoring send errors is apparently fine in this case.
+                        let _ = unordered_tx.send(err.into());
```
jsdt: Similar question here. It feels like we should always break if this fails.

crates/client-api/src/routes/subscribe.rs
```diff
@@ -444,7 +410,155 @@ async fn ws_client_actor_inner(
...
     log::debug!("Client connection ended");
-    sendrx.close();
+}
+
+/// Outgoing messages that don't need to be ordered wrt subscription updates.
+#[derive(From)]
+enum UnorderedWsMessage {
+    Close(CloseFrame),
+    Ping(Bytes),
+    Error(MessageExecutionError),
```
jsdt, line 897: If this includes error like a reducer failing, then we do want it to be order with subscription updates, since the the reason for the reducer could be data-dependent.

crates/client-api/src/routes/subscribe.rs
kim: The time this takes can be >1sec for pathological cases, and only after measuring that I added the `asyncify`.

Using rayon instead seems fine.

crates/client-api/src/routes/subscribe.rs
kim: `unordered_tx.send` fails if either the connection is bad or we already sent a close frame ourselves. Without more rework, I can't distinguish those. In the latter case we need to keep polling the recv end until the other end responds with a close.

crates/client-api/src/routes/subscribe.rs
kim: This I don't know. It was like this before.

crates/client-api/src/routes/subscribe.rs
kim: It is not an error result of a reducer call (this will appear on the `MeteredReceiver`), but an error calling the reducer in the first place. For example, if the reducer does not exist or the arguments are wrong. Iow, the reducer wasn't actually called.

I'm less sure about the other message types, e.g. subscribe commands. But I'd like to point out that it was not ordered in the code before this patch.

kim:
Update to:

- split into smaller functions that mainly transform `Stream`s. For readability and testability.
- actually terminate the actor loop when `recv` from the socket returns an error

kim:
Updated to:

- consider that buffer reclamation can fail if the socket is already closed
- re-introduce spawning the send loop

   This seems to be necessary in order to guarantee timely release of the database.
   I'm considering to spawn the receive end, too, so that we can get rid of the unbounded buffer + apply backpressure to clients instead.

kim:
Updated to:

- spawn the receive end, too

crates/client-api/src/routes/subscribe.rs
```diff
@@ -181,356 +194,631 @@ where
...
+/// This exists so we can test its behavior without a proper `ClientConnection`.
+async fn ws_main_loop<HotswapWatcher>(
+    state: Arc<ActorState>,
+    hotswap: impl Fn() -> HotswapWatcher,
+    idle_timer: impl Future<Output = ()>,
+    mut send_task: JoinHandle<()>,
+    mut recv_task: JoinHandle<()>,
+    unordered_tx: impl Fn(UnorderedWsMessage) -> Result<(), SendError<UnorderedWsMessage>>,
+) where
+    HotswapWatcher: Future<Output = Result<(), NoSuchModule>>,
```
gefjon: I would appreciate some amount of commentary on each of these parameters. `state` is reasonably obvious, but the meaning and use of `hotswap` is quite opaque, as are `idle_timer` and `unordered_tx`. It's fairly obvious what `send_task` and `recv_task` are, but not why this method needs them. I would prefer if a reader not need to actually inspect the body of this (fairly complex) function in order to know the purpose of each of these  arguments.

crates/client-api/src/routes/subscribe.rs
```diff
@@ -181,356 +194,631 @@ where
...
+
+    loop {
+        let closed = state.closed();
+
+        tokio::select! {
+            res = &mut send_task => {
+                if let Err(e) = res {
+                    if e.is_panic() {
+                        panic::resume_unwind(e.into_panic())
+                    }
```
gefjon, line 500: What do we do if `e` is not a panic? Is that case somehow known to be impossible?

crates/client-api/src/routes/subscribe.rs
```diff
@@ -181,356 +194,631 @@ where
...
-                // the client sent us a close frame
-                None => {
-                    break;
+                break;
+            },
+            res = &mut recv_task => {
+                if let Err(e) = res {
+                    if e.is_panic() {
+                        panic::resume_unwind(e.into_panic())
+                    }
```
gefjon, line 508: Same question re: `!e.is_panic()`.

crates/client-api/src/routes/subscribe.rs
```diff
@@ -181,356 +194,631 @@ where
...
-                    }
-                    let time = t1.elapsed();
-                    if time > Duration::from_millis(50) {
-                        tracing::warn!(?time, "send_all took a very long time");
-                    }
+                    // If `send_task` is gone, we'll exit the loop.
+                    unordered_tx(close.into()).ok();
+                }
+                watch_hotswap.set(hotswap());
+            },
```
gefjon: Where does the "update the client's module host" part happen? Is that handled within the `watch_hotswap` future? If so, please comment to that effect with a reference to that future's code.

crates/client-api/src/routes/subscribe.rs
```diff
@@ -181,356 +194,631 @@ where
...
-                        }
-                        Err(e) => {
-                            // Our send timed out; drop client without trying to send them a Close
-                            log::warn!("send_all timed out: {e}");
-                            break;
-                        }
+            _ = &mut idle_timer => {
+                log::warn!("Client {} timed out", state.client_id);
+                break;
+            },
```
gefjon, line 517: How does `idle_timer` work s.t. this isn't just cutting off active clients after the 30s timeout? Is that future somehow handling the ping/pong watching and such internally? If so, please comment to that effect with a reference to that future's code.

crates/client-api/src/routes/subscribe.rs
```diff
@@ -181,356 +194,631 @@ where
...
+
+            // Send ping.
+            //
+            // Branch is disabled if we already sent a close frame.
+            _ = ping_interval.tick(), if !closed => {
+                let was_ponged = state.reset_ponged();
+                if was_ponged {
+                    // If `send_task` is gone, we'll exit the loop.
+                    unordered_tx(UnorderedWsMessage::Ping(Bytes::new())).ok();
```
gefjon: I don't understand how this comment relates to the line below it. Do you mean, because some other piece of code handles watching if `send_task` is gone and exits the loop, we do not need to inspect or handle the error return here? If so, please write that, and also use the `let _ = ` idiom rather than `.ok()` to ignore the `Result`.

crates/client-api/src/routes/subscribe.rs
```diff
@@ -181,356 +194,631 @@ where
...
-                        };
-                        closed = true;
-                    }
+/// A sleep that can be extended by sending it new deadlines.
+///
+/// Sleeps until the deadline appearing on the `activity` channel,
+/// i.e. if a new deadline appears before the sleep finishes,
+/// the sleep is reset to the new deadline.
+///
+/// The `activity` should be updated whenever a new message is received or sent.
```
gefjon: Updating the timeout whenever we send a message seems suspicious to me. Wouldn't that result in a client which receives many transaction updates but never sends a message or responds to a ping staying alive?

crates/client-api/src/routes/subscribe.rs
```diff
@@ -181,356 +194,631 @@ where
...
+    let recv_loop = pin!(ws_recv_loop(state.clone(), idle_tx, ws));
+    let recv_handler = ws_client_message_handler(state.clone(), client_closed_metric, recv_loop);
+    pin_mut!(recv_handler);
+
+    while let Some((data, timer)) = recv_handler.next().await {
+        let result = message_handler(data, timer).await;
+        if let Err(e) = result {
+            if let MessageHandleError::Execution(err) = e {
+                log::error!("{err:#}");
+                // Break if send task is gone.
```
gefjon:
```suggestion
                // If the send task has exited, also exit this recv task.
```

crates/client-api/src/routes/subscribe.rs
```diff
@@ -181,356 +194,631 @@ where
...
+                    break;
                 }
                 continue;
             }
+            log::debug!("Client caused error: {e}");
+            let close = CloseFrame {
+                code: CloseCode::Error,
+                reason: format!("{e:#}").into(),
+            };
+            // Break if unable to initiate close handshake.
```
gefjon:
```suggestion
            // If the send task has exited, also exit this recv task.
            // No need to send the close handshake in that case; the client is already gone.
```

crates/client-api/src/routes/subscribe.rs
gefjon: Also, in the production usage, who holds the sender side of `activity`?

crates/client-api/src/routes/subscribe.rs
```diff
@@ -181,356 +194,631 @@ where
...
+    stream! {
+        while let Some(message) = messages.next().await {
+            match message {
+                ClientMessage::Message(message) => {
+                    log::trace!("Received client message");
+                    yield (message, Instant::now());
+                },
+                ClientMessage::Ping(_bytes) => {
+                    log::trace!("Received ping from client {}", state.client_id);
+                },
```
gefjon: Do we need or want to send a `Pong` response here? Does Tungstenite handle that automatically? If so, a comment to that effect would be appreciated.

gefjon (commented):
This is a huge improvement, many thanks. The tests are very encouraging. I've left a number of requests for new or expanded comments, as I'm wary that in a few months or years we'll end up with a similarly opaque tangle of logic to the previous version if multiple collaborators don't fully understand what's happening.

> Because there is no inbound buffering, the queue length metric will not be updated.

Should we remove this metric? Or is that going to break our dashboards in some frustrating way?

crates/client-api/src/routes/subscribe.rs
kim: The comment is wrong/outdated. The timer is reset on receiving a packet only.

kim:
> Should we remove this metric?

I'd like to put this up for discussion. The effect of not buffering is that we have no or reduced visibility into a situation where the database is lagging behind. We may alternatively insert a bounded queue, where the effect on the metric is likely that the queue is reported as either full or empty. But maybe that's good enough?

crates/client-api/src/routes/subscribe.rs
kim: I'm assuming this is not a genuine question, but requesting more prose. As such addressed.

gefjon: You assume correctly. Thanks!

PR #2908, opened 2025-06-27
Wrap websocket flush in timeout

joshua-spacetime:
# Description of Changes

This patch:

1. Breaks out of the websocket client actor loop on a close frame
2. Adds comments as to why this is safe to do
3. Wraps websocket flushes in a timeout, breaking out of the loop on timeout
4. Adds comments
5. Adds logging

The main change here is 3. Because if the flush stalls for any reason, before this patch, we could wait indefinitely without polling the other futures in the `select!` arms such as the liveness tick and the websocket receiver.

The flush could stall if the receiver stops ack'ing packets or if the tcp send buffer is full.

None

The change itself is not complicated, but the `select!` loop is. I'm fairly certain though that this change won't have any unintended consequences. It certainly doesn't poll fewer futures - it polls more.

# Testing

Ran a large scale performance test and observed these websocket actors being dropped and cleaned up on disconnects (initiated from both server and client)

### DIFF
--- a/crates/client-api/src/routes/subscribe.rs
+++ b/crates/client-api/src/routes/subscribe.rs
@@ -9,7 +9,7 @@ use axum_extra::TypedHeader;
 use bytes::Bytes;
 use bytestring::ByteString;
 use futures::future::MaybeDone;
-use futures::{Future, SinkExt, StreamExt};
+use futures::{Future, FutureExt, SinkExt, StreamExt};
 use http::{HeaderValue, StatusCode};
 use scopeguard::ScopeGuard;
 use serde::Deserialize;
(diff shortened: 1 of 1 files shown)


crates/client-api/src/routes/subscribe.rs
```diff
@@ -276,54 +277,79 @@ async fn ws_client_actor_inner(
 
             // If we have an outgoing message to send, send it off.
             // No incoming `message` to handle, so `continue`.
-            Some(n) = sendrx.recv_many(&mut rx_buf, 32).map(|n| (n != 0).then_some(n)) => {
+            n = sendrx.recv_many(&mut rx_buf, 32) => {
+                // The receiver has been closed and there are no pending messages.
+                // This is due to the client sending a close frame, so break from the loop.
+                if n == 0 {
+                    break;
```
cloutiertyler: Are we absolutely sure that this is safe to break out of the loop from?

crates/client-api/src/routes/subscribe.rs
```diff
@@ -276,54 +277,79 @@ async fn ws_client_actor_inner(
...
+                    (ws.flush().await, msg_buffer)
+                };
+                // Build a future that both times out and drives the send.
+                //
+                // Note that if flushing cannot immediately complete for whatever reason,
+                // it will wait without polling the other futures in the `select!` arms.
+                // Among other things, this means our liveness tick will not be polled.
+                //
+                // To avoid waiting indefinitely, we wrap the send in a timeout.
+                // A timeout is treated as an unresponsive client and we drop the connection.
```
cloutiertyler: Seems reasonable.

crates/client-api/src/routes/subscribe.rs
```diff
@@ -352,10 +378,29 @@ async fn ws_client_actor_inner(
...
+                    // Build a future that both times out and drives the send.
+                    //
+                    // Note that if the send cannot immediately complete for whatever reason,
+                    // it will wait without polling the other futures in the `select!` arms.
+                    // Among other things, this means we won't poll the websocket for a Close frame.
+                    //
+                    // To avoid waiting indefinitely, we wrap the ping in a timeout.
+                    // A timeout is treated as an unresponsive client and we drop the connection.
+                    let ping_with_timeout = async {
+                        tokio::time::timeout(SEND_TIMEOUT, ws.send(WsMessage::Ping(Bytes::new()))).await
```
cloutiertyler: Rather than timing these out, could we lift them into the select itself?

crates/client-api/src/routes/subscribe.rs
```diff
@@ -439,7 +484,18 @@ async fn ws_client_actor_inner(
...
+                // Is it correct to break if a reducer is still in progress?
+                // Answer: Yes it is.
+                //
+                // If a reducer is currently being executed,
+                // we are waiting for the `current_message` future to complete.
+                // When we break, the task completes and this future is dropped.
+                //
+                // Notably though the reducer itself will run to completion,
+                // however when it tries to notify this task that it is done,
+                // it will encounter a closed sender in `JobThread::run`.
```
cloutiertyler: Which does what exactly?

crates/client-api/src/routes/subscribe.rs
joshua-spacetime: Yes, this means `send_rx` has been closed and drained. There may be a reducer or some other request in flight, but I explain in another comment that it's still safe to drop this task in that case.

crates/client-api/src/routes/subscribe.rs
joshua-spacetime: That would complicate the `select!` loop even more, so I would advise against that. That would be much easier to do if we had separate actors for reading and writing.

crates/client-api/src/routes/subscribe.rs
joshua-spacetime: The value trying to be sent will be dropped.

crates/client-api/src/routes/subscribe.rs
cloutiertyler: That's fair, honestly that's probably what we should do. Just simpler.

crates/client-api/src/routes/subscribe.rs
cloutiertyler: Could we add that to the comment (so people know it will not throw an error/panic).

cloutiertyler:
New changes also look good to me, since they're mostly mechanical in nature. Would be great to test, but I think it falls under mechanical enough.

Issue #2824, opened 2025-06-04
Add buffer pool for serializing subscription results


joshua-spacetime:
That is, specifically for subscribe and unsubscribe. Currently for a subscribe(unsubscribe) call, the construction of each `TableUpdate`/`BsatnRowList` incurs a heap allocation which is then discarded by the task that transmits results to clients. We should grab buffers from a pool and use those buffers in `TableUpdate`s instead of constantly allocating and deallocating new ones.

joshua-spacetime:
Part of what is described in this ticket is to avoid allocating an intermediate vec before serializing to a `BsatnRowList`. This has been implemented separately by https://github.com/clockworklabs/SpacetimeDB/pull/2918.

PR #2911, opened 2025-07-02
Reuse buffers in `ServerMessage<BsatnFormat>`

Centril:
# Description of Changes

Fixes https://github.com/clockworklabs/SpacetimeDB/issues/2824.

Defines a global pool `BsatnRowListBuilderPool` which reclaims the buffers of a `ServerMessage<BsatnFormat>` and which is then used when building new `ServerMessage<BsatnFormat>`s. 

Notes:
1. The new pool `BsatnRowListBuilderPool` reports the same kind of metrics to prometheus as `PagePool` does.
2. `BsatnRowListBuilder` now works in terms of `BytesMut`.
3. The trait method `fn to_bsatn_extend` is redefined to be capable of dealing with `BytesMut` as well as `Vec<u8>`.
4. A trait `ConsumeEachBuffer` is defined from `ServerMessage<BsatnFormat>` and down to extract buffers. `<ServerMessage<_> as ConsumeEachBuffer>::consume_each_buffer(...)` is then called in `messages::serialize(...)` just after bsatn-encoding the entire message and before any compression is done. This is the place where the pool reclaims buffers.

# Benchmarks

Benchmark numbers vs. master using `cargo bench --bench subscription -- --baseline subs` on i7-7700K, 64GB RAM:

```
footprint-scan          time:   [21.607 ms 21.873 ms 22.187 ms]
                        change: [-62.090% -61.438% -60.787%] (p = 0.00 < 0.05)
                        Performance has improved.

full-scan               time:   [22.185 ms 22.245 ms 22.324 ms]
                        change: [-36.884% -36.497% -36.166%] (p = 0.00 < 0.05)
                        Performance has improved.
```

The improvements in `footprint-scan` are mostly thanks to https://github.com/clockworklabs/SpacetimeDB/pull/2918, but 7 ms of the improvements here are thanks to the pool. The improvements to `full-scan` should be only thanks to the pool.

None

2?

# Testing

- Tests for `Pool<T>` also apply to `BsatnRowListBuilderPool`.

### DIFF
--- a/crates/core/src/host/host_controller.rs
+++ b/crates/core/src/host/host_controller.rs
@@ -110,7 +110,7 @@ pub struct HostController {
     pub bsatn_rlb_pool: BsatnRowListBuilderPool,
 }
 
-struct HostRuntimes {
+pub(crate) struct HostRuntimes {
     wasmtime: WasmtimeRuntime,
     v8: V8Runtime,
 }
@@ -476,7 +476,15 @@ impl HostController {
         };
         let host = guard.as_ref().ok_or(NoSuchModule)?;
 
-        host.migrate_plan(host_type, program, style).await
+        host.migrate_plan(
+            self.page_pool.clone(),
+            self.bsatn_rlb_pool.clone(),
+            &self.runtimes,
+            host_type,
+            program,
+            style,
+        )
+        .await
     }
 
     /// Release all resources of the [`ModuleHost`] identified by `replica_id`,
@@ -1124,13 +1132,17 @@ impl Host {
     /// Generate a migration plan for the given `program`.
     async fn migrate_plan(
         &self,
+        page_pool: PagePool,
+        bsatn_rlb_pool: BsatnRowListBuilderPool,
+        host_runtimes: &Arc<HostRuntimes>,
         host_type: HostType,
         program: Program,
         style: PrettyPrintStyle,
     ) -> anyhow::Result<MigratePlanResult> {
         let old_module = self.module.borrow().info.clone();
 
-        let module_def = extract_schema(program.bytes, host_type).await?;
+        let module_def =
+            extract_schema_with_pools(page_pool, bsatn_rlb_pool, host_runtimes, program.bytes, host_type).await?;
 
         let res = match ponder_migrate(&old_module.module_def, &module_def) {
             Ok(plan) => MigratePlanResult::Success {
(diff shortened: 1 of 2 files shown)


joshua-spacetime:
Can you make the first commit its own independent PR?

Centril:
> Can you make the first commit its own independent PR?

Will do 👍

Centril:
This is now a fairly small PR based atop another fairly small PR github.com/clockworklabs/SpacetimeDB/pull/2918.

gefjon (approved):
Approved as a code owner, and no objections otherwise, though I'd still like review from @joshua-spacetime re: the metrics.

joshua-spacetime (dismissed):
If we wait for https://github.com/clockworklabs/SpacetimeDB/pull/2866 to land first, this doesn't have to be a pool, it can just be a thread local buffer. I think I'd prefer we wait for that in order to simplify things even further.

Centril:
> If we wait for #2866 to land first, this doesn't have to be a pool, it can just be a thread local buffer. I think I'd prefer we wait for that in order to simplify things even further.

For a thread local to work, you'd need to move `messages::serialize`, which reclaims the buffers and re-adds them to the pool, to the same thread as does the query evaluation. Currently, https://github.com/clockworklabs/SpacetimeDB/pull/2866 doesn't do that, and it doesn't seem like that would be profitable.

crates/sats/src/bsatn.rs
```diff
@@ -106,6 +109,52 @@ codec_funcs!(val: crate::AlgebraicValue);
 codec_funcs!(val: crate::ProductValue);
 codec_funcs!(val: crate::SumValue);
 
+/// Provides a view over a buffer that an reserve an additional `len` bytes
```
kim:
```suggestion
/// Provides a view over a buffer that can reserve an additional `len` bytes
```

crates/core/src/host/host_controller.rs
```diff
@@ -1215,7 +1227,9 @@ pub async fn extract_schema(program_bytes: Box<[u8]>, host_type: HostType) -> an
     let runtimes = HostRuntimes::new(None);
     let page_pool = PagePool::new(None);
     let core = SingleCoreExecutor::in_current_tokio_runtime();
-    let module_info = Host::try_init_in_memory_to_check(&runtimes, page_pool, database, program, core).await?;
+    let bsatn_rlb_pool = BsatnRowListBuilderPool::new();
```
kim: Maybe we want a variant of `extract_schema` that allows `HostController::migrate_plan` to pass the existing `PagePool`, `BsatnRowListBuildPool`, etc. ?

kim (approved):

crates/core/src/host/host_controller.rs
Centril: Yep, and now we pass from `HostController` both pools and the runtimes.

Issue #2894, opened 2025-06-24
TS Client: "ERROR: Negative reference count for row", and "ERROR: Updating a row that was not present in the cache"


doodlechaos:
TS Client Version:  "@clockworklabs/spacetimedb-sdk": "^1.2.0",
spacetimedb = "1.2.0"

How to reproduce: 
Create fresh rust server and paste this into lib.rs. 

```rust
use spacetimedb::{rand::Rng, reducer, table, ReducerContext, ScheduleAt, Table, TimeDuration};

#[table(name = player, public, 
    index(name = idx_dummy, btree(columns = [dummy]))
 )]
#[derive(Clone)]
pub struct Player {
    #[primary_key]
    pub player_id: u64,

    pub dummy: u64,
    pub x: i32,
}

#[table(name = player_data, public, 
    index(name = idx_dummy, btree(columns =[dummy]))
 )]
pub struct PlayerData {
    #[primary_key]
    pub player_id: u64,

    pub dummy: u64,
    pub y: i32,
}

#[reducer(init)]
pub fn init(_ctx: &ReducerContext) -> Result<(), String> {
    let _ = _ctx.db.player().try_insert(Player {
        player_id: 1,
        dummy: 0,
        x: 0,
    });

    let _ = _ctx.db.player_data().try_insert(PlayerData {
        player_id: 1,
        dummy: 0,
        y: 0,
    });

    let _ = _ctx.db.player().try_insert(Player {
        player_id: 2,
        dummy: 0,
        x: 0,
    });

    let _ = _ctx.db.player_data().try_insert(PlayerData {
        player_id: 2,
        dummy: 0,
        y: 0,
    });

    init_tick_players_schedule(_ctx);

    Ok(())
}

#[table(name = move_players_schedule, scheduled(tick_players))]
pub struct MovePlayersSchedule {
    #[primary_key]
    #[auto_inc]
    pub id: u64,

    pub scheduled_at: ScheduleAt,
}

pub fn init_tick_players_schedule(ctx: &ReducerContext) {
    let scheduled_at = TimeDuration::from_micros(1000000).into();

    let schedule = ctx
        .db
        .move_players_schedule()
        .iter()
        .next()
        .unwrap_or_else(|| {
            ctx.db.move_players_schedule().insert(MovePlayersSchedule {
                id: 0,
                scheduled_at,
            })
        });

    ctx.db
        .move_players_schedule()
        .id()
        .update(MovePlayersSchedule {
[...]

PR #2914, opened 2025-07-03
Fix equality comparison between Row::Ptr and Row::Ref

jsdt:
# Description of Changes

The derived `PartialEq` always considers `Ptr` and `Ref` to be different, even if the values are the same.

This fixes the errors from [this repro](https://github.com/clockworklabs/SpacetimeDB/issues/2894).

1.

# Testing

Todo: this should have some test coverage, since this was causing a bug with joins.

- [ ] 
- [ ]

### DIFF
--- a/crates/execution/src/lib.rs
+++ b/crates/execution/src/lib.rs
@@ -115,11 +115,24 @@ pub trait DeltaStore {
     }
 }
 
-#[derive(Clone, PartialEq, Eq)]
+#[derive(Clone)]
 pub enum Row<'a> {
     Ptr(RowRef<'a>),
     Ref(&'a ProductValue),
 }
+
+impl PartialEq for Row<'_> {
+    fn eq(&self, other: &Self) -> bool {
+        match (self, other) {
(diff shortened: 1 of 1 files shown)


crates/execution/src/lib.rs
```diff
@@ -115,12 +115,25 @@ pub trait DeltaStore {
     }
 }
 
-#[derive(Clone, PartialEq, Eq)]
+#[derive(Clone)]
```
Centril, line 118:
```suggestion
#[derive(Copy, Clone)]
```

joshua-spacetime:
I think `test_join_pruning` was waiting on line 2330:
```rust
assert_tx_update_for_table(&mut rx, u_id, &schema, [], []).await;
```

because the transaction no longer returns any row updates, so you can remove this check.

joshua-spacetime (approved):
We should have already had test coverage for this, but unfortunately the test in question was broken. It's fixed now though.

crates/execution/src/lib.rs
jsdt: I'd rather not add this, since we aren't relying on it.

Issue #2761, opened 2025-05-20
SpacetimeDB daemon panics and crashes on updating a module re-creating the same index


egormanga:
### Setup:
```rust
#[table(name=user)]
struct User {
	#[primary_key]
	id: Identity,

	#[unique]
	username: String,
}

#[table(name=thing)]
struct Thing {
	#[primary_key]
	#[auto_inc]
	id: u64,

	#[index(btree)]
	owner: Identity,
}
```

### Logs:
```
INFO crates/core/src/db/update.rs:143: Creating index `thing_owner_idx_btree` on table `thing`
thread 'tokio-runtime-worker' panicked at crates/table/src/table.rs:1790:14:
called `Option::unwrap()` on a `None` value
```
<details><summary>stack backtrace:</summary>

```
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::panic
   3: core::option::unwrap_failed
   4: spacetimedb_table::table::UniqueConstraintViolation::build
   5: spacetimedb::db::datastore::locking_tx_datastore::mut_tx::MutTxId::create_index::{{closure}}
   6: spacetimedb::db::datastore::locking_tx_datastore::mut_tx::MutTxId::create_index
   7: spacetimedb::db::update::update_database
   8: <spacetimedb::host::wasm_common::module_host_actor::WasmModuleInstance<T> as spacetimedb::host::module_host::ModuleInstance>::update_database
   9: <spacetimedb::host::module_host::AutoReplacingModuleInstance<T> as spacetimedb::host::module_host::ModuleInstance>::update_database
  10: spacetimedb::host::module_host::ModuleHost::call::{{closure}}::{{closure}}
  11: tokio::runtime::task::core::Core<T,S>::poll
  12: tokio::runtime::task::harness::Harness<T,S>::poll
  13: tokio::runtime::scheduler::multi_thread::worker::Context::run_task
  14: tokio::runtime::scheduler::multi_thread::worker::Context::run
  15: tokio::runtime::context::scoped::Scoped<T>::set
  16: tokio::runtime::context::runtime::enter_runtime
  17: tokio::runtime::scheduler::multi_thread::worker::run
  18: <tokio::runtime::blocking::task::BlockingTask<T> as core::future::future::Future>::poll
  19: tokio::runtime::task::core::Core<T,S>::poll
  20: tokio::runtime::task::harness::Harness<T,S>::poll
[...]

PR #2915, opened 2025-07-03
`update_database`: `is_unique` should only use the table's constraints

Centril:
# Description of Changes

Currently, before this PR, when checking if the index-to-add is unique, we consider all constraints in the module.
This caused a bug where we'd let the unique constraints of another table affect the table-to-alter.

Fixes https://github.com/clockworklabs/SpacetimeDB/issues/2761.

None

2?

# Testing

A regression test is added in the last commit.
Also verified that a module with the last repro in the reducer doesn't work on master and does with the PR.

### DIFF
--- a/crates/core/src/db/update.rs
+++ b/crates/core/src/db/update.rs
@@ -12,6 +12,17 @@ use spacetimedb_schema::def::TableDef;
 use spacetimedb_schema::schema::{column_schemas_from_defs, IndexSchema, Schema, SequenceSchema, TableSchema};
 use std::sync::Arc;
 
+/// The logger used for by [`update_database`] and friends.
+pub trait UpdateLogger {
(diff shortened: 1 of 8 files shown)


crates/core/src/db/update.rs
```diff
@@ -118,31 +137,26 @@ fn auto_migrate_database(
...
 
                 let index_cols = ColSet::from(index_def.algorithm.columns());
 
-                let is_unique = plan
-                    .new
-                    .constraints()
-                    .filter_map(|c| c.data.unique_columns())
+                let is_unique = table_def
+                    .constraints
+                    .iter()
```
Centril, line 153: Here's the meat of the bug fix itself. The rest is mostly about testing.

kim (approved):
lgtm

PR #2918, opened 2025-07-04
`execute_plan`: don't build temporary vec of rows

Centril:
# Description of Changes

Avoid building a temporary `Vec` in `execute_plan` by exposing a list-building interface instead.

- The old `fn encode_list` is rewritten in terms of this list-building interface.
- The `BsatnRowList` and `BsatnRowListBuilder` types are split into two entirely separate types. The latter now tries to recognize the case where there isn't a known static layout, but where the BSATN lengths happen to be the same for all rows anyways. In those cases, the allocation of `RowSizeHint::RowOffsets` is avoided in favor of just storing the found length in bytes. This is in particular useful for small table updates as statistically, the fewer rows, the more chance of the lengths being all equal. In the case of a single row, the chance is notably 100%. It is also good for the case of when we don't have `RelValue::Row` or `Row::Ptr` but where the underlying table that actually has a static layout.

In the future, we might want to avoid these lists in incremental as well.

# Benchmarks

Benchmark numbers vs. master using `cargo bench --bench subscription -- --baseline subs` on i7-7700K, 64GB RAM:
```
footprint-scan          time:   [28.731 ms 28.924 ms 29.171 ms]
                        change: [-49.728% -49.006% -48.388%] (p = 0.00 < 0.05)
                        Performance has improved.
```

Performance goes from roughly 56.721 ms to 28.795 ms.

None

2, fairly local change to just subscriptions.

# Testing

Covered by existing tests.


gefjon (approved):
Good code, good comments, good benchmarks. A lot of this `RowListBuilder` stuff could be moved out of the code-ownered `websocket.rs`, IMO, since we're trying to restrict changes to the format of the messages, not the procedures by which they are constructed. But I don't think it *needs* to get moved, I'm just flagging it as an option.

joshua-spacetime:
I would like to review this before it merges.

> A lot of this RowListBuilder stuff could be moved out of the code-ownered websocket.rs

In particular I _may_ ask that we do this.

Centril:
> In particular I _may_ ask that we do this.

I think we should split `websocket.rs` into several pieces, but as https://github.com/clockworklabs/SpacetimeDB/pull/2911 builds upon this PR and also changes `websocket.rs`, I don't think we should split in this PR.

joshua-spacetime (changes requested):
I do think it is an opportune time to move the non-spacetime types out of `websocket.rs`. This file should really be reserved for the public message schemas only.

crates/client-api-messages/src/websocket.rs
```diff
@@ -26,22 +26,18 @@ use smallvec::SmallVec;
...
-use std::{
-    io::{self, Read as _, Write as _},
-    sync::Arc,
-};
+use std::sync::Arc;
 
 pub const TEXT_PROTOCOL: &str = "v1.json.spacetimedb";
 pub const BIN_PROTOCOL: &str = "v1.bsatn.spacetimedb";
 
 pub trait RowListLen {
```
joshua-spacetime, line 39: Why not move the entire trait out? Isn't the idea that we only want `SpacetimeType`s in this file?

Centril: I think the way it is now strikes a good balance. I don't think the goal should be to only have `SpacetimeType`s in this file, but rather, define the formats, encodings, and basic affordances, without any intermediaries like builders. If we want to divide further, I think we could do it along the lines of `websocket/{traits.rs, server.rs, client.rs}`, but this feels out of scope for this PR.

PR #2839, opened 2025-06-06
Noa/v8

coolreader18:
# Description of Changes

Very WIP , and will be for a while.

# API and ABI breaking changes

# Expected complexity level and risk

# Testing

- [ ] 
- [ ]

PR #2922, opened 2025-07-04
Move `sats::de::serde::NoneAccess` to `sats::de::NoneAccess`

Centril:
# Description of Changes

Extracted from https://github.com/clockworklabs/SpacetimeDB/pull/2839 which moves this to use it for another `Deserializer` format.

None

1

# Testing

Just code motion.


kazimuth (approved):
Just moving private code, ~good modulo one needed reexport to maintain API compatibility~. Misread, moved code was private before the change, 👍

PR #2923, opened 2025-07-04
Add `V8Runtime` skeleton + `HostType::Js`

Centril:
# Description of Changes

Extracts some of the work in https://github.com/clockworklabs/SpacetimeDB/pull/2839 that is mostly ceremony and not deeper V8 stuff. A `JsModule` is not constructible yet, so none of the panics are reachable.

None

1

# Testing

No testing yet. There's nothing interesting to test.


crates/core/src/host/host_controller.rs
```diff
@@ -106,12 +107,14 @@ pub struct HostController {
...
     wasmtime: WasmtimeRuntime,
+    v8: V8Runtime,
 }
 
 impl HostRuntimes {
     fn new(data_dir: Option<&ServerDataDir>) -> Arc<Self> {
         let wasmtime = WasmtimeRuntime::new(data_dir);
-        Arc::new(Self { wasmtime })
+        let v8 = V8Runtime::new();
+        Arc::new(Self { wasmtime, v8 })
```
kazimuth, line 118: The way this is structured, we're always running V8, even when unused. Do we want to consider lazily initializing it?

Centril: I think we do. I'll make an inner type and the outer type shall be lazy.

Centril: I changed things to use a `LazyLock` instead o that we won't start `V8` until its actually needed.

crates/core/src/host/wasm_common/module_host_actor.rs
```diff
@@ -122,8 +122,6 @@ pub enum DescribeError {
     Decode(#[from] DecodeError),
     #[error(transparent)]
     RuntimeError(anyhow::Error),
-    #[error("unimplemented RawModuleDef version")]
-    UnimplementedRawModuleDefVersion,
```
gefjon, line 126: Where did this go?

gefjon (approved):

crates/core/src/host/wasm_common/module_host_actor.rs
Centril: It's unused

PR #2924, opened 2025-07-06
refactor(sdk): improve the log message source line reference

ResuBaka:
# Description of Changes

Currently with the function maybe_log_error to log errors we hide the real source line where the error happened. So to still have the same logic but with more accurate source line, I moved it to a macro to have the same abstraction for logging. 

- 

1

# Testing

- [ ] I have tested it my application and it now show the better source line of the log message

### DIFF
--- a/crates/sdk/src/websocket.rs
+++ b/crates/sdk/src/websocket.rs
@@ -200,6 +200,14 @@ fn request_insert_auth_header(req: &mut http::Request<()>, token: Option<&str>) 
     }
 }
 
+macro_rules! maybe_log_error {
+    ($cause:expr, $res:expr) => {
+        if let Err(e) = $res {
+            log::warn!("{}: {:?}", $cause, e);
+        }
+    };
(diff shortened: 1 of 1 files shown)


crates/sdk/src/websocket.rs
```diff
@@ -200,6 +200,14 @@ fn request_insert_auth_header(req: &mut http::Request<()>, token: Option<&str>)
     }
 }
 
+macro_rules! maybe_log_error {
```
gefjon, line 206:
```suggestion
/// If `res` evaluates to `Err(e)`, log a warning in the form `"{}: {:?}", $cause, e`.
///
/// Could be trivially written as a function, but macro-ifying it preserves the source location of the log.
macro_rules! maybe_log_error {
```

gefjon (approved):
This seems eminently reasonable. My standards for comments are higher for macros than functions, so I've suggested one.

PR #2925, opened 2025-07-07
cli: Close the websocket connection gracefully

kim:
The `subscribe` command would drop the connection without sending a
close frame. Doing so creates a warning on the server, which can be
distracting when debugging other connections issues.

1

# Testing

Use the `subscribe` command with a local database and compare server logs before
and after.

### DIFF
--- a/crates/cli/src/subcommands/subscribe.rs
+++ b/crates/cli/src/subcommands/subscribe.rs
@@ -12,7 +12,7 @@ use spacetimedb_lib::ser::serde::SerializeWrapper;
 use std::time::Duration;
 use tokio::io::AsyncWriteExt;
 use tokio_tungstenite::tungstenite::client::IntoClientRequest;
-use tokio_tungstenite::tungstenite::Message as WsMessage;
+use tokio_tungstenite::tungstenite::{Error as WsError, Message as WsMessage};
 
 use crate::api::ClientApi;
 use crate::common_args;
@@ -158,7 +158,7 @@ pub async fn exec(config: Config, args: &ArgMatches) -> Result<(), anyhow::Error
     if let Some(auth_header) = api.con.auth_header.to_header() {
         req.headers_mut().insert(header::AUTHORIZATION, auth_header);
     }
-    let (mut ws, _) = tokio_tungstenite::connect_async(req).await?;
+    let mut ws = tokio_tungstenite::connect_async(req).await.map(|(ws, _)| ws)?;
 
     let task = async {
         subscribe(&mut ws, queries.cloned().map(Into::into).collect()).await?;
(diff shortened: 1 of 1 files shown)


crates/cli/src/subcommands/subscribe.rs
```diff
@@ -234,11 +238,7 @@ where
 
 /// Print `num` [`ServerMessage::TransactionUpdate`] messages as JSON.
 /// If `num` is `None`, keep going indefinitely.
-async fn consume_transaction_updates<S>(
-    ws: &mut S,
-    num: Option<u32>,
-    module_def: &RawModuleDefV9,
-) -> anyhow::Result<bool>
+async fn consume_transaction_updates<S>(ws: &mut S, num: Option<u32>, module_def: &RawModuleDefV9) -> anyhow::Result<()>
```
bfops: was this changed for a reason related to the bugfix, or unrelated?

bfops:
It looks like the tests are failing. I tried re-running, but it may be an actual failure. I haven't looked into why.

kim:
Ah well, missed the case where the server closes the connection.

crates/cli/src/subcommands/subscribe.rs
kim: Not unrelated: the (undocumented) bool was supposed to distinguish server close vs. received requested number of messages vs. timeout. But that didn't work properly due to `?` short circuiting.

I now changed it to just return `ConnectionClosed` when the server closed the connection, as in all other cases the client is supposed to send a close frame.

bfops:
@kim looks like it's still failing?

kim:
meh

crates/cli/src/subcommands/subscribe.rs
```diff
@@ -155,35 +157,80 @@ pub async fn exec(config: Config, args: &ArgMatches) -> Result<(), anyhow::Error
...
-        task.await?
+        task.await
     };
 
-    if needs_shutdown {
-        ws.close(None).await?;
+    // Close the connection gracefully, unless it's a websocket error,
+    // in which case the connection is most likely already unusable.
+    if !matches!(res, Err(Error::Subscribe { .. } | Error::Websocket { .. })) {
+        // Ignore errors here, we're going to drop the connection anyways.
```
bfops: given this comment, is there a downside to closing the websocket unconditionally? I'd ideally prefer not to reason about whether the websocket "should" already be closed.

...or is that what the PR tried to do before the CI kept complaining?

crates/cli/src/subcommands/subscribe.rs
```diff
@@ -231,41 +282,47 @@ where
...
-    S: TryStream<Ok = WsMessage> + Unpin,
-    S::Error: std::error::Error + Send + Sync + 'static,
+    S: TryStream<Ok = WsMessage, Error = WsError> + Unpin,
 {
     let mut stdout = tokio::io::stdout();
     let mut num_received = 0;
     loop {
         if num.is_some_and(|n| num_received >= n) {
-            break Ok(true);
+            return Ok(());
```
bfops, line 301: if I'm reading this right, we used to return `true` for `needs_shutdown`, and now we just return `Ok()` - is that correct?

kim: That's correct -- we always need a shutdown, except when the server already disconnected us (which is now an `Err` return).

crates/cli/src/subcommands/subscribe.rs
kim: The downside is the motivation of this PR, given in the description.

If you prefer, we can omit the `if` and simply do `let _ = ws.close(None).await`. The (potential) error result we want to display to the user is already stored in `res`.

bfops: Yeah I'd personally prefer to omit the `if` and go fully unconditional. I trust your judgment though.

bfops (approved):
This PR LGTM. I haven't done any testing.

There's still one unresolved comment thread, that you can resolve however makes sense to you and then merge. :ship:

PR #2926, opened 2025-07-08
core: Tone down `MessageExecutionError`

kim:
The display impl of the error prints a stack trace, which is not
particularly helpful given it is usually a user error.

Use the alternate format (`{:#}`) instead, which makes anyhow display a
single line and no backtrace.

1 -- Only changes how the error is rendered

# Testing

- [x] Run a smoketest that yields an execution error, e.g.
   `test_add_then_remove_index`, and convince yourself that one line is
   sufficient


kazimuth (approved):
👍

PR #2931, opened 2025-07-11
client-api: Send WebSocket messages fragmented

kim:
RFC 6455, Section 5.4 describes message fragmentation, and we can do that with tungstenite.

It does seem to help getting control messages (ping, pong, close) through without head-of-line blocking.

2 - Need to test with clients

# Testing

TBD - some more abstraction is needed due to the difficulty of synthetically producing a large outgoing message.


crates/client-api/src/routes/subscribe.rs
```diff
@@ -1046,7 +1048,48 @@ async fn send_message<S: Sink<WsMessage> + Unpin>(
...
+
+                let hdr = last.header_mut();
+                hdr.is_final = true;
+                hdr.opcode = OpCode::Data(Data::Continue);
+            }
+        }
+
+        log::trace!("sending message in {} frames", frames.len());
+        for frame in frames {
+            ws.feed(WsMessage::Frame(frame)).await?;
```
gefjon: We expect `feed` to return quickly, without waiting for the client to actually receive the message, right? How do we make it so that control frames actually get interspersed in here, rather than just queueing up a whole bunch of data frames in rapid succession?

kim: Yeah... we can't use `select!` in the send loop, consuming from `unordered_rx` needs to happen in a separate task.

aasoni: I believe we might just need to add something like:
```rust
const HEARTBEAT_EVERY: usize = 16;   // after every 16 fragments ≈ 64 KiB
...
    
let mut i = 0;
for frame in frames {
    ws.feed(Message::Frame(frame)).await?;

    // interleave a ping in regularly
    if i % HEARTBEAT_EVERY == 0 {
        ws.feed(Message::Ping(Vec::new())).await?;
    }
    i += 1;
}
```

kim: That's one solution, yes. 

We also had issues with server close not getting through quickly enough, though, which was one of the motivations to redo the code.

So we'll need to actually produce an interleaved stream here before feeding it to the socket.

aasoni (commented):
Tested this locally with BitCraft and had no issues loading into the game and playing for a few minutes.

aasoni (approved):
This was deployed to the BitCraft production cluster over a week ago and has not had issues.

PR #2932, opened 2025-07-11
Fix `AlgebraicTypeLayout::is_compatible_with`

Centril:
# Description of Changes

Fix `AlgebraicTypeLayout::is_compatible_with`.

Change `VarLenType::Array` to store `ArrayType`
to make similar such bugs impossible in the future.
Consequence changes are made to hashing and row serialization to make that work.
This should also benefit performance as there's now one fewer `Box` to look through.

None

3 - fairly central piece of machinery are touched and there's some unsafe code affected.

# Testing

A regression test has been added.


kim (approved):
Change to store the inner type seems reasonable to me.
The patch fixes the reported issue.

PR #2934, opened 2025-07-11
Implemented auto migration support for appending new columns

egormanga:
# Description of Changes

Previously, you couldn't add new columns at the end of a table without clearing the database, even though this should not require a manual schema change.

None.

### 4:
This small change might easily be totally incorrect, possibly overwrite some unrelated data (although all existing tests pass) or collapse the Universe. Need advice from people who understand the inner workings better than your humble.

In particular, how do schema rollbacks handle this?
For adding variants (#2874), [new logic](https://github.com/clockworklabs/SpacetimeDB/pull/2874/files#diff-0413bed1c098a1f77fe04c79f32d745136bd0d3cc26fd2087b2b370cd78369ebR444) had to be introduced. As I suppose, in this case the underlying data store gets duplicated and refilled entirely, rather than altered in-place? If not, perhaps `PendingSchemaChange` needs to be extended as well, with all the hassle this implies.

Also, I'm not sure about using `column_id` as position, but the comment in a latter hand is somewhat reassuring.

# Testing

- [x] Manual testing on a mid-sized database with multiple tables
- [ ] Needs unit tests


crates/execution/src/lib.rs
```diff
@@ -128,6 +128,8 @@ impl PartialEq for Row<'_> {
             (Self::Ref(x), Self::Ref(y)) => x == y,
             (Self::Ptr(x), Self::Ref(y)) => x == *y,
             (Self::Ref(x), Self::Ptr(y)) => y == *x,
+            (Self::Null, _) => false,
+            (_, Self::Null) => false,
```
egormanga: Note: the compiler suddenly started complaining on the lack of these hands after I added the below logic, for some reason.

egormanga:
Alright, adding a column mid-table instead of appending now causes this:
```rust
ERROR client-api/src/lib.rs:377:
  internal error: failed to get or launch module host:
    TableError: Table with ID `0` not found in `st_table`.:
      Table with ID `0` not found in `st_table`.
```
_(formatting mine.)_

~Rollback after that seems to work, though.~
Upd: nah it's all gone, the entire db content.

Thus I conclude that using `column_id` as position does not work. Or, put another way, we need to add a check that all previous columns are kept the same, because adding one in between them technically causes the last one to appear as an appended, if we ignore (as currently) the preceding types.
Need help here.

P.S. There must be a bug in the `TableError` formatting handler, causing the message to repeat twice.

crates/execution/src/lib.rs
egormanga: Upd: seems to be unrelated to that commit. Might be an implicit conflict of my branch with some updates on the master. Sorry for that, will remove alongside any further pushes.

egormanga: Upd 2: Oops, that variant's not even on the master, it's from my one of my other PR #2892, definitely. Removing the hunk.

egormanga:
Further clues suggest that I somehow suppress generating the `AutoMigrateError::ReorderTable` when adding a column in the middle, although it's not clear to me, in which way I do so.

egormanga:
Seems to not be working at all, as the next publishing (no matter the changes) causes:
```rust
Error: Failed to create or update the database: Column count mismatch: expected self.columns.len() == def.columns.len():
   self.columns.len(): 15
   def.columns.len(): 17
```

Apparently, somewhere the schema is not getting updated on the migration.

kazimuth:
We're working on this feature internally, but unfortunately there's a lot of moving parts.

We need to:

- Completely reinitialize the table. We're implementing this by having the user supply a default value for the new columns in the macros; then the entire table will need to be read into memory (As a `Vec<ProductValue>`, updated with the new columns, and put back. In addition the system tables need to be patched and we'll need to double check the other `core` metadata that's floating around in various `Arc`s.

- Disconnect all users. Changing the table layout breaks the network ABI of the module because of the network binary format Clockwork chose to use early in development. We need to force a disconnect so users don't start getting garbled updates. (In the future we're going to implement a versioning scheme for this, with support in the macros, core, and client SDKs, but at first we're going to rely just on loudly printing warnings and marking the feature of adding columns as experimental.) The user will then be responsible for updating their clients before they reconnect.

- Print a summary of the auto migration, including if it will disconnect all users, before it is run. This will require a new HTTP endpoint to get the summary.

gefjon:
Thank you for your contribution! We really appreciate your dedication to SpacetimeDB and your willingness to contribute. We do not think adding columns should be a fully automatic migration. Automatic migrations should only be used for changes which have a single clear, unambiguous correct transformation from the old data to the new data.

Hypothetically, the database could choose some transform for any proposed schema change, and so we could expand automatic migration support to allow literally any schema change. However, we fear that this would lead to unexpected behavior (See [this discussion](https://github.com/clockworklabs/SpacetimeDB/pull/2874#discussion_r2169644775) for an example of our concerns).

We currently have a couple members of the team designing and implementing what we're calling "supervised migrations," which will require explicit approval and some guidance from the user (in this case a default value to populate the new column with). I'm going to close this PR to avoid creating conflicts with their work.

PR #2935, opened 2025-07-11
Add a recursion limit to the evaluation of type_expr & parse_expr

mamcx:
# Description of Changes

Add a guard against `stack overflow` in case of nested expression and `joins`.

None

1

# Testing
- [x] Find how `deep` can be recursed the affected functions and put a limit on it
- [x] Add a extra test to prove we can (in theory) do lots of `joins` in the planning steps, even if executing them will be slow

### DIFF
--- a/crates/core/src/sql/execute.rs
+++ b/crates/core/src/sql/execute.rs
@@ -1126,6 +1126,9 @@ pub(crate) mod tests {
         Ok(())
     }
 
+    /// Test we are protected against recursion when:
+    /// 1. The query is too large
+    /// 2. The AST is too deep
     #[test]
     fn test_large_query_no_panic() -> ResultTest<()> {
         let db = TestDB::durable()?;
@@ -1138,16 +1141,46 @@ pub(crate) mod tests {
             )
             .unwrap();
 
-        let mut query = "select * from test where ".to_string();
-        for x in 0..1_000 {
-            for y in 0..1_000 {
-                let fragment = format!("((x = {x}) and y = {y}) or");
-                query.push_str(&fragment);
+        let build_query = |total| {
+            let mut sql = "select * from test where ".to_string();
+            for x in 0..total {
+                for y in 0..total {
+                    let fragment = format!("((x = {x}) and (y = {y})) or ");
+                    sql.push_str(&fragment);
+                }
             }
-        }
-        query.push_str("((x = 1000) and (y = 1000))");
+            sql.push_str("((x = 1000) and (y = 1000))");
+            sql
+        };
+        let run = |db: &RelationalDB, sep: char, sql_text: &str| {
+            run_for_testing(db, sql_text).map_err(|e| e.to_string().split(sep).next().unwrap_or_default().to_string())
+        };
+        let sql = build_query(1_000);
+        assert_eq!(
+            run(&db, ':', &sql),
+            Err("SQL query exceeds maximum allowed length".to_string())
+        );
 
-        assert!(run_for_testing(&db, &query).is_err());
+        // Exercise the limit [recursion::MAX_RECURSION_EXPR] && [recursion::MAX_RECURSION_TYP_EXPR]
+        let sql = build_query(8);
+        assert_eq!(run(&db, ',', &sql), Err("Recursion limit exceeded".to_string()));
+
+        let sql = build_query(7);
+        assert!(run(&db, ',', &sql).is_ok(), "Expected query to run without panic");
+
+        // Check no overflow with lot of joins
+        let mut sql = "SELECT test.* FROM test ".to_string();
+        // We could pust up to 700 joins without overflow as long we don't have any conditions,
+        // but here execution become too slow.
+        // TODO: Move this test to the `Plan`
+        for i in 0..200 {
+            sql.push_str(&format!("JOIN test AS m{i} ON test.x = m{i}.y "));
+        }
+
+        assert!(
+            run(&db, ',', &sql).is_ok(),
+            "Query with many joins and conditions should not overflow"
+        );
         Ok(())
     }
 
--- a/crates/core/src/sql/execute.rs
+++ b/crates/core/src/sql/execute.rs
@@ -1126,9 +1126,11 @@ pub(crate) mod tests {
         Ok(())
     }
 
-    /// Test we are protected against recursion when:
-    /// 1. The query is too large
+    /// Test we are protected against stack overflows when:
+    /// 1. The query is too large (too many characters)
     /// 2. The AST is too deep
+    ///
+    /// Exercise the limit [`recursion::MAX_RECURSION_EXPR`]
     #[test]
     fn test_large_query_no_panic() -> ResultTest<()> {
         let db = TestDB::durable()?;
@@ -1143,13 +1145,11 @@ pub(crate) mod tests {
 
         let build_query = |total| {
             let mut sql = "select * from test where ".to_string();
-            for x in 0..total {
-                for y in 0..total {
-                    let fragment = format!("((x = {x}) and (y = {y})) or ");
-                    sql.push_str(&fragment);
-                }
+            for x in 1..total {
+                let fragment = format!("x = {x} or ");
+                sql.push_str(&fragment.repeat((total - 1) as usize));
             }
-            sql.push_str("((x = 1000) and (y = 1000))");
+            sql.push_str("(y = 0)");
             sql
         };
         let run = |db: &RelationalDB, sep: char, sql_text: &str| {
@@ -1161,16 +1161,15 @@ pub(crate) mod tests {
             Err("SQL query exceeds maximum allowed length".to_string())
         );
 
-        // Exercise the limit [recursion::MAX_RECURSION_EXPR] && [recursion::MAX_RECURSION_TYP_EXPR]
-        let sql = build_query(8);
+        let sql = build_query(41); // This causes stack overflow without the limit
         assert_eq!(run(&db, ',', &sql), Err("Recursion limit exceeded".to_string()));
 
-        let sql = build_query(7);
+        let sql = build_query(40); // The max we can with the current limit
         assert!(run(&db, ',', &sql).is_ok(), "Expected query to run without panic");
 
         // Check no overflow with lot of joins
         let mut sql = "SELECT test.* FROM test ".to_string();
-        // We could pust up to 700 joins without overflow as long we don't have any conditions,
+        // We could push up to 700 joins without overflow as long we don't have any conditions,
         // but here execution become too slow.
         // TODO: Move this test to the `Plan`
         for i in 0..200 {
--- a/crates/expr/src/check.rs
+++ b/crates/expr/src/check.rs
@@ -99,7 +99,7 @@ pub trait TypeChecker {
                     vars.insert(rhs.alias.clone(), rhs.schema.clone());
 
                     if let Some(on) = on {
-                        if let Expr::BinOp(BinOp::Eq, a, b) = type_expr(vars, on, Some(&AlgebraicType::Bool), &mut 0)? {
+                        if let Expr::BinOp(BinOp::Eq, a, b) = type_expr(vars, on, Some(&AlgebraicType::Bool))? {
                             if let (Expr::Field(a), Expr::Field(b)) = (*a, *b) {
                                 join = RelExpr::EqJoin(LeftDeepJoin { lhs, rhs }, a, b);
                                 continue;
--- a/crates/expr/src/check.rs
+++ b/crates/expr/src/check.rs
@@ -99,7 +99,7 @@ pub trait TypeChecker {
                     vars.insert(rhs.alias.clone(), rhs.schema.clone());
 
                     if let Some(on) = on {
-                        if let Expr::BinOp(BinOp::Eq, a, b) = type_expr(vars, on, Some(&AlgebraicType::Bool))? {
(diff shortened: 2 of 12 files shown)


crates/core/src/sql/execute.rs
```diff
@@ -1126,6 +1126,9 @@ pub(crate) mod tests {
         Ok(())
     }
 
+    /// Test we are protected against recursion when:
+    /// 1. The query is too large
```
joshua-spacetime:
```suggestion
    /// 1. The query is too large (too many characters)
```

crates/core/src/sql/execute.rs
```diff
@@ -1138,16 +1141,46 @@ pub(crate) mod tests {
...
 
-        let mut query = "select * from test where ".to_string();
-        for x in 0..1_000 {
-            for y in 0..1_000 {
-                let fragment = format!("((x = {x}) and y = {y}) or");
-                query.push_str(&fragment);
+        let build_query = |total| {
+            let mut sql = "select * from test where ".to_string();
+            for x in 0..total {
+                for y in 0..total {
```
joshua-spacetime:
```suggestion
            for x in 1..total {
                for y in 1..total {
```

crates/core/src/sql/execute.rs
```diff
@@ -1138,16 +1141,46 @@ pub(crate) mod tests {
...
+                    let fragment = format!("((x = {x}) and (y = {y})) or ");
+                    sql.push_str(&fragment);
+                }
             }
+            sql.push_str("((x = 1000) and (y = 1000))");
```
joshua-spacetime:
```suggestion
            sql.push_str("((x = 0) and (y = 0))");
```

crates/core/src/sql/execute.rs
```diff
@@ -1138,16 +1141,46 @@ pub(crate) mod tests {
...
+        let run = |db: &RelationalDB, sep: char, sql_text: &str| {
+            run_for_testing(db, sql_text).map_err(|e| e.to_string().split(sep).next().unwrap_or_default().to_string())
+        };
+        let sql = build_query(1_000);
+        assert_eq!(
+            run(&db, ':', &sql),
+            Err("SQL query exceeds maximum allowed length".to_string())
+        );
+
+        // Exercise the limit [recursion::MAX_RECURSION_EXPR] && [recursion::MAX_RECURSION_TYP_EXPR]
```
joshua-spacetime: Not sure we need two different limits, one should be sufficient.

crates/core/src/sql/execute.rs
```diff
@@ -1138,16 +1141,46 @@ pub(crate) mod tests {
...
+        let sql = build_query(8);
+        assert_eq!(run(&db, ',', &sql), Err("Recursion limit exceeded".to_string()));
```
joshua-spacetime: I think 8 * 8 = 64 `OR`s is far too conservative.

crates/expr/src/check.rs
```diff
@@ -99,7 +99,7 @@ pub trait TypeChecker {
                     vars.insert(rhs.alias.clone(), rhs.schema.clone());
 
                     if let Some(on) = on {
-                        if let Expr::BinOp(BinOp::Eq, a, b) = type_expr(vars, on, Some(&AlgebraicType::Bool))? {
+                        if let Expr::BinOp(BinOp::Eq, a, b) = type_expr(vars, on, Some(&AlgebraicType::Bool), &mut 0)? {
```
joshua-spacetime: Let's not expose the recursion limit parameter in the public function.

crates/sql-parser/src/parser/errors.rs
```diff
@@ -89,6 +89,13 @@ pub enum SqlRequired {
     JoinAlias,
 }
 
+#[derive(Error, Debug)]
+#[error("Recursion limit exceeded, `{message}` limit: {limit}")]
```
joshua-spacetime: I don't think we need to expose the limit in the message.

crates/sql-parser/src/parser/recursion.rs
```diff
@@ -0,0 +1,34 @@
...
+//!
+//! Removing one could allow the others to be higher, but depending on how the `SQL` is structured, it could lead to a `stack overflow`
+//! if is not guarded against, so is incorrect to assume that a limit is sufficient for the next part of the parser.
+use crate::parser::errors::{RecursionError, SqlParseError};
+use std::fmt::Display;
+
+/// A conservative limit for recursion depth on `parse_expr`.
+pub const MAX_RECURSION_EXPR: usize = 700;
+/// A conservative limit for recursion depth on `type_expr`.
+pub const MAX_RECURSION_TYP_EXPR: usize = 5_000;
```
joshua-spacetime: I think a single limit for `parse_expr` is sufficient. But `700` seems like a very low limit. If we assume a 2MB stack size, 700 implies a max frame size of ~3KB. What is the max depth of `parse_expr` before we overflow the stack?

joshua-spacetime (changes requested):

crates/sql-parser/src/parser/mod.rs
```diff
@@ -238,31 +243,31 @@ pub(crate) fn parse_expr(expr: Expr) -> SqlParseResult<SqlExpr> {
             op: BinaryOperator::And,
             right,
         } => {
-            let l = parse_expr(*left)?;
-            let r = parse_expr(*right)?;
+            let l = parse_expr(*left, depth)?;
+            let r = parse_expr(*right, depth)?;
```
joshua-spacetime: Note, `depth` shouldn't be mutable, because `depth != depth(l) + depth(r)`.

crates/sql-parser/src/parser/recursion.rs
mamcx: Having both specified allow to see the actual constraints (or future devs could think `700` is a limit for both).

mamcx: I confirm with multiples sources the limit:

* Using `lldb` with ` settings set frame-format "frame #${frame.index}: sp=${frame.sp} fp=${frame.fp} pc=${frame.pc}{ ${module.file.basename}{`${function.name-with-args}${function.pc-offset}}}{ at ${line.file.basename}:${line.number}}{${function.is-optimized} [opt]}\n"` I see that the stack consumption was around `2.06 MiB`
* With `sudo vmmap 51029  | grep "^Stack" | grep -v Guard; echo ""` I see 
```bash
Stack                       16f608000-16fe00000    [ 8160K    64K    64K     0K] rw-/rwx SM=PRV          thread 0
Stack                       16fe04000-17000c000    [ 2080K  2064K  2064K     0K] rw-/rwx SM=PRV          thread 1
Stack                       170010000-170218000    [ 2080K    16K    16K     0K] rw-/rwx SM=PRV          thread 2
```
* With `stacker`:
```
std::mem::size_of::<Expr>(): 168
  SqlParseResult<SqlExpr> = 976

Stack size: 2928 bytes remaining
Recursion depth sql-parser::parse_expr: 728
```

crates/expr/src/lib.rs
```diff
@@ -78,8 +79,15 @@ pub(crate) fn type_proj(input: RelExpr, proj: ast::Project, vars: &Relvars) -> T
     }
 }
 
-/// Type check and lower a [SqlExpr] into a logical [Expr].
-pub(crate) fn type_expr(vars: &Relvars, expr: SqlExpr, expected: Option<&AlgebraicType>) -> TypingResult<Expr> {
+// Assert we will not blow the stack with recursion:
+const _: () = assert!(size_of::<Relvars>() == 48);
```
joshua-spacetime:
```suggestion
```
This is passed by reference so we don't need to assert it's size.

crates/expr/src/lib.rs
```diff
@@ -78,8 +79,15 @@ pub(crate) fn type_proj(input: RelExpr, proj: ast::Project, vars: &Relvars) -> T
...
+const _: () = assert!(size_of::<TypingResult<Expr>>() == 64);
+const _: () = assert!(size_of::<SqlExpr>() == 40);
+const _: () = assert!(size_of::<Option<&AlgebraicType>>() == 8);
```
joshua-spacetime:
```suggestion
```
We don't really need to assert on the size of a pointer.

crates/expr/src/lib.rs
```diff
@@ -78,8 +79,15 @@ pub(crate) fn type_proj(input: RelExpr, proj: ast::Project, vars: &Relvars) -> T
     }
 }
 
-/// Type check and lower a [SqlExpr] into a logical [Expr].
-pub(crate) fn type_expr(vars: &Relvars, expr: SqlExpr, expected: Option<&AlgebraicType>) -> TypingResult<Expr> {
+// Assert we will not blow the stack with recursion:
```
joshua-spacetime:
```suggestion
// These types determine the size of each stack frame during type checking.
// Changing their sizes will require updating the recursion limit to avoid stack overflows.
```

crates/sql-parser/src/parser/errors.rs
```diff
@@ -50,7 +50,7 @@ pub enum SqlUnsupported {
     #[error("Unsupported FROM expression: {0}")]
     From(TableFactor),
     #[error("Unsupported set operation: {0}")]
-    SetOp(SetExpr),
+    SetOp(Box<SetExpr>),
```
joshua-spacetime, line 53: I don't think boxing here is necessary since we're boxing the outer type.

crates/sql-parser/src/parser/mod.rs
```diff
@@ -203,16 +207,21 @@ pub(crate) fn parse_proj(expr: Expr) -> SqlParseResult<ProjectExpr> {
     }
 }
 
+// Assert we will not blow the stack with recursion:
```
joshua-spacetime:
```suggestion
// These types determine the size of [`parse_expr`]'s stack frame.
// Changing their sizes will require updating the recursion limit to avoid stack overflows.
```

crates/sql-parser/src/parser/recursion.rs
```diff
@@ -0,0 +1,28 @@
+//! A utility for guarding against excessive recursion depth in the SQL parser.
```
joshua-spacetime:
```suggestion
//! A utility for guarding against stack overflows in the SQL parser.
```

crates/sql-parser/src/parser/recursion.rs
```diff
@@ -0,0 +1,28 @@
...
+//!
+//! Different parts of the parser may have different recursion limits.
+//!
+//! Removing one could allow the others to be higher, but depending on how the `SQL` is structured, it could lead to a `stack overflow`
```
joshua-spacetime: Note, the limits should not depend on how the `SQL` is structured. I would reword this, or perhaps remove it if it no longer applies.

crates/sql-parser/src/parser/recursion.rs
```diff
@@ -0,0 +1,28 @@
...
+//! if is not guarded against, so is incorrect to assume that a limit is sufficient for the next part of the parser.
+use crate::parser::errors::{RecursionError, SqlParseError};
+
+/// A conservative limit for recursion depth on `parse_expr`.
```
joshua-spacetime, line 7: How conservative is this limit?

crates/expr/src/statement.rs
```diff
@@ -519,4 +520,27 @@ mod tests {
             assert!(result.is_err());
         }
     }
+
+    // Manually build the AST for a recursive query,
+    // because we limit the length of the query to prevent stack overflow on parsing.
+    /// Exercise the limit [`recursion::MAX_RECURSION_TYP_EXPR`]
```
joshua-spacetime:
```suggestion
    /// Manually build the AST for a recursive query,
    /// because we limit the length of the query to prevent stack overflow on parsing.
    /// Exercise the limit [`recursion::MAX_RECURSION_TYP_EXPR`]
```

crates/core/src/sql/execute.rs
```diff
@@ -1126,6 +1126,11 @@ pub(crate) mod tests {
         Ok(())
     }
 
+    /// Test we are protected against recursion when:
```
joshua-spacetime:
```suggestion
    /// Test we are protected against stack overflows when:
```

crates/core/src/sql/execute.rs
```diff
@@ -1138,16 +1143,43 @@ pub(crate) mod tests {
...
+            Err("SQL query exceeds maximum allowed length".to_string())
+        );
+
+        let sql = build_query(41); // This causes stack overflow without the limit
+        assert_eq!(run(&db, ',', &sql), Err("Recursion limit exceeded".to_string()));
+
+        let sql = build_query(40); // The max we can with the current limit
+        assert!(run(&db, ',', &sql).is_ok(), "Expected query to run without panic");
+
+        // Check no overflow with lot of joins
```
joshua-spacetime, line 1170: The above tests are sufficient. I don't think this join test provides further coverage, so I would remove it.

mamcx: We don't have any similar test, so I think is good to have it as confirmation.

crates/sql-parser/src/parser/errors.rs
mamcx: In fact without this the size get larger, even if the outer is boxed. Note that this enum is returned in several places without the outer

joshua-spacetime (approved):

PR #2933, opened 2025-07-11
Split datastore into its own crate

cloutiertyler:
# Description of Changes

This PR splits the datastore into its own crate. It makes purely mechanical changes to change imports and make things `pub` which were `pub(crate)` and necessarily needed to become public. With two exceptions:

1. Due to a bug in the compiler I had to use concrete `Locking` types rather than associated types in one case.
2. I needed to move Noa's newly added `From` implementation for `HostType` which references `ModuleKind`, a type unknown to the datastore.

# API and ABI breaking changes

No API braking changes.

# Expected complexity level and risk

1.5 - solely mechanical, but a lot of files touched

# Testing

Given the mechanical nature of the changes, I believe our automated tests to be sufficient.

PR #2937, opened 2025-07-14
Enable the `datastore` crate's `test` feature where necessary

gefjon:
# Description of Changes

#2933  broke running `cargo test` without `--all` because the new `datastore` crate hid some of its test utilities behind `cfg(any(test, feature = "test"))`,  but did not enable that feature in the dependency from `core` or `sqltest`.

This was hidden in our CI because we run with `--all` there, which apparently enables all optional features, including the `test` feature.

N/a

1

# Testing

- [x] Manually ran `cargo check --tests --package $PACKAGE` for every crate with a dependency on `spacetimedb-datastore` (as revealed by `grep` for `'spacetimedb-datastore ='`) and ensured that they all compiled correctly after this patch.

### DIFF
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -140,7 +140,7 @@ perfmap = []
 spacetimedb-lib = { path = "../lib", features = ["proptest"] }
 spacetimedb-sats = { path = "../sats", features = ["proptest"] }
 spacetimedb-commitlog = { path = "../commitlog", features = ["test"] }
-spacetimedb-datastore = { workspace = true, features = ["test"] }
+spacetimedb-datastore = { path = "../datastore/", features = ["test"] }
 
 criterion.workspace = true
 # Also as dev-dependencies for use in _this_ crate's tests.


crates/core/Cargo.toml
```diff
@@ -140,6 +140,7 @@ perfmap = []
 spacetimedb-lib = { path = "../lib", features = ["proptest"] }
 spacetimedb-sats = { path = "../sats", features = ["proptest"] }
 spacetimedb-commitlog = { path = "../commitlog", features = ["test"] }
```
Centril, line 142: (these should also be using `workspace = true`)

gefjon: Apparently they can't be! We have a whole CI job set up to ensure that `dev-dependencies` always use `path`. The job says it's somehow required to publish crates. I don't understand why this matters, but clearly it's important if we set up a whole CI job to force it.

Centril: Odd! Oh well... 😂

PR #2940, opened 2025-07-16
Extract `ModuleCommon` + use in wasm & v8 runtimes

Centril:
# Description of Changes

Extract a `struct ModuleCommon` which holds some common context data for various runtimes.
Imbue this new type/module +with common logic and use it in wasm and v8 runtimes.

None

1

# Testing

No V8 tests yet.
WASM runtime covered by existing tests.

### DIFF
--- a/crates/core/src/host/module_common.rs
+++ b/crates/core/src/host/module_common.rs
@@ -38,7 +38,7 @@ pub fn build_common_module_from_raw(
     Ok(ModuleCommon::new(replica_ctx, mcc.scheduler, info, mcc.energy_monitor))
 }
 
-/// Non-runtime specific parts of a module.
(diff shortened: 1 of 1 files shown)


crates/core/src/host/module_common.rs
```diff
@@ -0,0 +1,90 @@
...
+        replica_ctx.database_identity,
+        mcc.program.hash,
+        log_tx,
+        replica_ctx.subscriptions.clone(),
+    );
+
+    Ok(ModuleCommon::new(replica_ctx, mcc.scheduler, info, mcc.energy_monitor))
+}
+
+/// Non-runtime specific parts of a module.
```
gefjon:
```suggestion
/// Non-runtime-specific parts of a module.
```

PR #2951, opened 2025-07-17
Import C#/Unity SDK

bfops:
# Description of Changes

Merge the spacetime-docs repo contents into a subdirectory of this repo, completely unchanged. The corresponding branches have already been pushed as `csharp/[original branch name]`.

We will mirror this to com.clockworklabs.spacetimedbsdk for now, and update the `release/latest` branch and create tags there when we do a release.

This should be merged with a **merge commit**, for two reasons:
1. To maintain the commit history of the merged repo (a squash commit would make all of the history prunable since it wouldn't be relevant anymore)
2. So that migrated branches + PRs from the merged repo can merge in `master` without getting horrible merge conflicts (a squash commit would essentially newly create all of those files, causing merge conflicts)

Because of that, I think we should skip the merge queue to merge this PR (so that we can use a merge commit). It's entirely new files, so it should not cause any issues for other PRs.

None

1

# Testing

- [x] CI is updated and passing in a descendant PR: https://github.com/clockworklabs/SpacetimeDB/pull/2952


cloutiertyler (approved):
I'm approving, but I had a thought while looking at the license here, we should probably have the top level license.txt explain that each folder has its own license or something, then move the BuSL license into the appropriate places. Right now the BuSL is probably applying more places than it should. We can think/coord on this.

sdks/csharp/.github/pull_request_template.md
```diff
@@ -0,0 +1,21 @@
...
+*Describe what has been changed, any new features or bug fixes*
+
+## API
+
+ - [ ] This is an API breaking change to the SDK
+
+*If the API is breaking, please state below what will break*
+
+## Requires SpacetimeDB PRs
+*List any PRs here that are required for this SDK change to work*
```
jdetter, line 11: I think this section is no longer needed now that this is merged into SpacetimeDB

sdks/csharp/.github/pull_request_template.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Testsuite
+*If you would like to run the your SDK changes in this PR against a specific SpacetimeDB branch, specify that here. This can be a branch name or a link to a PR.*
+
+SpacetimeDB branch name: master
```
jdetter, line 16: Same with this section, no longer needed

sdks/csharp/.github/workflows/publish-nuget.yml
```diff
@@ -0,0 +1,25 @@
+name: Publish NuGet Package on Release
```
jdetter, line 1: Ok this is quite nice now that when we send out a SpacetimeDB release it will also release the nuget package for us automatically. This is a huge improvement.

sdks/csharp/.github/workflows/test.yml
```diff
@@ -0,0 +1,172 @@
...
+jobs:
+  unity-testsuite:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          submodules: true
+
+      # Grab the SpacetimeDB branch name from the PR description. If it's not found, master will be used instead.
```
jdetter, line 18: IMO We no longer need to extract the SpacetimeDB branch for this CI step

jdetter:
There are 2 major things we have to do next:
1. The original `com.clockworklabs.spacetimedbsdk` repo should become a mirror of this subdirectory. Some questions to answer:
  - How does this affect our default branch in the `com.clockworklabs.spacetimedbsdk` repo? Should we keep it as `release/latest`?
  - Should we mirror *all* branches to `com.clockworklabs.spacetimedbsdk`? I think the answer is yes because then we can ship "testing" branches to BitCraft and/or other clients without having to manually create that branch.
2. Can we add a CI for whenever we update `crates/csharp-bindings/*` we regenerate the DLLs for that PR? Then when that PR merges the DLLs. Also tracking the DLLs within the SpacetimeDB repo is going to make the repo bloat pretty quickly, perhaps we consider `git-lfs`?

sdks/csharp/.github/workflows/test.yml
bfops: Agreed, see https://github.com/clockworklabs/SpacetimeDB/pull/2952

sdks/csharp/.github/pull_request_template.md
bfops: Agreed, see https://github.com/clockworklabs/SpacetimeDB/pull/2952

sdks/csharp/.github/pull_request_template.md
bfops: Agreed, see https://github.com/clockworklabs/SpacetimeDB/pull/2952

sdks/csharp/.github/workflows/publish-nuget.yml
bfops: I actually don't know that we want that - not all SpacetimeDB releases are tied to C# SDK releases (and vice-versa)

bfops:
> There are 2 major things we have to do next:
> 
> 1. The original `com.clockworklabs.spacetimedbsdk` repo should become a mirror of this subdirectory. Some questions to answer:
> 
> * How does this affect our default branch in the `com.clockworklabs.spacetimedbsdk` repo? Should we keep it as `release/latest`?

Yes, because the goal of the default branch is for users to get something sane if they just paste the repo URL into the Unity package manager.

> * Should we mirror _all_ branches to `com.clockworklabs.spacetimedbsdk`? I think the answer is yes because then we can ship "testing" branches to BitCraft and/or other clients without having to manually create that branch.

I think not. If we want to give clients a particular branch, I believe we can link them to the SpacetimeDB repo directly with a subdirectory. I can add instructions on how to do this.

> 2. Can we add a CI for whenever we update `crates/csharp-bindings/*` we regenerate the DLLs for that PR? Then when that PR merges the DLLs. Also tracking the DLLs within the SpacetimeDB repo is going to make the repo bloat pretty quickly, perhaps we consider `git-lfs`?

Yeah agreed, this automation becomes a bit easier now that the repos are merged. There's a separate ticket for it though. +1 to git lfs too!

sdks/csharp/.github/workflows/publish-nuget.yml
jdetter: Yeah I guess it's not true in the general case that we *always* want to send out C# SDK updates for every SpacetimeDB release (especially for minor version bumps)

jdetter:
> I think not. If we want to give clients a particular branch, I believe we can link them to the SpacetimeDB repo directly with a subdirectory. I can add instructions on how to do this.

Does this work for checking out a Unity package? Like can you checkout a subdirectory? Will that clone the entire repository?

bfops:
> > I think not. If we want to give clients a particular branch, I believe we can link them to the SpacetimeDB repo directly with a subdirectory. I can add instructions on how to do this.
> 
> Does this work for checking out a Unity package? Like can you checkout a subdirectory? Will that clone the entire repository?

I can't say whether it clones the entire repository, but yes you can check out a subdirectory. It's what we do in the unity testsuite in https://github.com/clockworklabs/SpacetimeDB/pull/2952.

PR #2952, opened 2025-07-17
Update GitHub files for C#/Unity SDK

bfops:
# Description of Changes

Update READMEs and CI related to the import of C#/Unity SDK stuff. Removed some CI that is now redundant. See inline comments for more details.

I also removed the `unity-tests~` submodule that used to point to Blackholio, since that is now integrated into this repo as well (https://github.com/clockworklabs/SpacetimeDB/pull/2945).

I also had to update the generated Blackholio bindings to make the CI pass, because they hadn't been updated in a while. However, in the previous repo, Blackholio was used as a submodule, so `git diff` was erroneously "passing" despite the fact that there was, in fact, a diff. Now that Blackholio is in the same repo, `git diff` in the CI is correctly detecting that the bindings need regenerating.

None

1

# Testing

- [x] Migrated CI is now running and passing


.github/workflows/ci.yml
```diff
@@ -113,62 +113,6 @@ jobs:
         working-directory: crates/bindings-csharp
         run: dotnet test -warnaserror
 
-  sdk_test:
```
bfops, line 116: this set of tests is no longer needed since the C# SDK CI is now running in this repo anyway (see the `csharp-test.yml` file)

.github/workflows/csharp-test.yml
```diff
@@ -12,88 +12,52 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
-        with:
-          submodules: true
-
-      # Grab the SpacetimeDB branch name from the PR description. If it's not found, master will be used instead.
```
bfops, line 18: we now just grab SpacetimeDB from the current branch rather than making it configurable

.github/workflows/csharp-test.yml
```diff
@@ -104,21 +68,25 @@ jobs:
...
         run: bash ./generate.sh -y
 
       - name: Check for changes
         run: |
-          git diff --exit-code unity-tests~/client-unity/Assets/Scripts/autogen || {
+          # This was copied from tools/check-diff.sh.
+          # It's required because `spacetime generate` creates lines with the SpacetimeDB commit
+          # version, which would make this `git diff` check very brittle if included.
+          PATTERN='^// This was generated using spacetimedb cli version.*'
+          git diff --exit-code --ignore-matching-lines="$PATTERN" -- demo/Blackholio/client-unity/Assets/Scripts/autogen || {
```
bfops, line 89: the old `git diff` was essentially nonfunctional because `unity-tests~` was a submodule, so it always vacuously returned 0. Now that it's no longer a submodule, we had to fix up this check to work properly after #2673 .

demo/Blackholio/server-rust/Cargo.toml
```diff
@@ -5,6 +5,8 @@ edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
+[workspace]
```
bfops, line 10: this turns out to be needed because Blackholio is now nested under the SpacetimeDB root dir. It should have happened in https://github.com/clockworklabs/SpacetimeDB/pull/2947, but we don't have CI for Blackholio itself.

.github/workflows/csharp-test.yml
```diff
@@ -1,9 +1,9 @@
-name: Test Suite
+name: C#/Unity - Test Suite
 
 on:
   push:
     branches:
-      - staging
+      - master
```
bfops, line 6: this was outdated/broken for a while after we renamed the `staging` branch to `master`. noticed and fixed during migration.

Issue #2912, opened 2025-07-02
Automatic migration support while adding columns + pretty-printing


kazimuth:
- BitCraft may need to be able to add columns to tables
- User should get a report on what changes will happen during an auto-migration BEFORE the migration happens
- If the auto-migration will require clients to go down and update, there should be big bold red warnings

kazimuth:
See https://github.com/clockworklabs/SpacetimeDBPrivate/pull/1849

TODOs:
- [x] Implement adding columns in auto-migrations in the `schema` crate
- [ ] Implement default column values in the module library macros
   - [ ] Rust
   - [ ] C#
- [x] Implement a `pre_publish` API endpoint that gets a printable version of an auto migration plan from the database.
   - [x] Decide how exactly the flow should work between `pre_publish` and `publish`
   - Note: see https://github.com/clockworklabs/SpacetimeDB/pull/2065 for a partial implementation of plan printing. The plumbing here was wrong, because it did the publishing in a single HTTP call, rather than two. This meant that the user only got the summary *after* the migration had already been applied. We need two separate calls to print the summary before the plan is applied, and get user approval.
   - If the plan includes a `DisconnectAllUsers` step, we need to print a VERY LOUD WARNING in the plan.
- [ ] Implement the actual migration in `core::db::update`. This will need to
   - [ ] Read the table into memory, update it, write it back
   - [ ] Update system tables
   - [ ] Update any other metadata that might be floating around

egormanga:
How this would work with WAL, e.g. history feature?

Shubham8287:
Closued with #2962

PR #2956, opened 2025-07-18
Enable adding columns in auto migrations in `schema` crate

kazimuth:
# Description of Changes

This adds auto migration support to the `schema` crate.
Additional changes are needed to fully implement this feature: see https://github.com/clockworklabs/SpacetimeDB/issues/2912 

None

1, at this level the change is fairly straightforward. The integration will be more difficult.

# Testing

Added `schema` unit tests for success/failure cases.


crates/bindings/tests/ui/tables.stderr
```diff
@@ -129,11 +129,11 @@ error[E0277]: `&'a Alpha` cannot appear as an argument to an index filtering ope
              &ConnectionId
              &Identity
              &Lifecycle
-             &RawMiscModuleExportV9
```
Shubham8287, line 132: @Centril UI tests were failing until I updated this file. I believe it's because the trait FilterableValues was auto-implemented for RawMiscModuleExportV9, but this PR changes the enum to include a variant with data, which affects the generated messages.

crates/core/src/db/update.rs
```diff
@@ -252,6 +252,7 @@ fn auto_migrate_database(
                 log!(logger, "Removing-row level security `{sql_rls}`");
                 stdb.drop_row_level_security(tx, sql_rls.clone())?;
             }
+            _ => unimplemented!(),
```
Centril: Before this can be merged, this should turn into a proper returned error rather than a panic.

Centril:
```suggestion
            _ => Err(anyhow!("unimplemented")),
```

crates/lib/src/db/raw_def/v9.rs
```diff
@@ -361,7 +362,25 @@ pub struct RawRowLevelSecurityDefV9 {
...
+#[cfg_attr(feature = "test", derive(PartialEq, Eq, PartialOrd, Ord))]
+pub struct RawColumnDefaultValueV9 {
+    /// Must point to a `ProductType` or an error will be thrown during validation.
+    pub table: RawIdentifier,
+
+    /// Must be the index of a valid column within `ty`.
+    pub col_id: ColId,
+
+    /// A BSATN-encoded AlgebraicValue valid at `ty`.
+    /// (We can't use AlgebraicValue directly because it isn't serializable!)
```
Centril:
```suggestion
    /// A BSATN-encoded [`AlgebraicValue`] valid at `ty`.
    /// (We can't use `AlgebraicValue` directly because it isn't serializable!)
```

Centril: Also, the claim in the parenthesis doesn't seem true?
```rust
impl_serialize!([] AlgebraicValue, (self, ser) => match self {
```

crates/lib/src/db/raw_def/v9.rs
```diff
@@ -794,6 +813,20 @@ impl RawTableDefBuilder<'_> {
         self
     }
 
+    /// Adds a default value for the field.
+    /// Note: this sets the value for all other tables defined on the same
```
Centril: on the same...<?>

crates/sats/src/buffer.rs
```diff
@@ -10,7 +10,7 @@ use core::fmt;
 use core::str::Utf8Error;
 
 /// An error that occurred when decoding.
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
```
Centril, line 13: Why was this necessary?

crates/lib/src/db/raw_def/v9.rs
```diff
@@ -794,6 +813,20 @@ impl RawTableDefBuilder<'_> {
         self
     }
 
+    /// Adds a default value for the field.
```
Centril:
```suggestion
    /// Adds a default value for the `column`.
```

crates/lib/src/db/raw_def/v9.rs
```diff
@@ -794,6 +813,20 @@ impl RawTableDefBuilder<'_> {
...
+    /// Note: this sets the value for all other tables defined on the same
+    pub fn with_default_column_value(self, column: impl Into<ColId>, value: AlgebraicValue) -> Self {
+        // Added to misc_exports for backwards-compatibility reasons
```
Centril:
```suggestion
        // Added to `misc_exports` for backwards-compatibility reasons.
```
(please use backticks when referring to code, bindings, types, etc.)

crates/lib/src/db/raw_def/v9.rs
```diff
@@ -361,7 +362,25 @@ pub struct RawRowLevelSecurityDefV9 {
 #[sats(crate = crate)]
 #[cfg_attr(feature = "test", derive(PartialEq, Eq, PartialOrd, Ord))]
 #[non_exhaustive]
-pub enum RawMiscModuleExportV9 {}
+pub enum RawMiscModuleExportV9 {
+    ColumnDefaultValue(RawColumnDefaultValueV9),
+}
+
+/// Marks a field as having a particularly default
```
Centril:
```suggestion
/// Marks a field as having a particular default.
```

crates/lib/src/db/raw_def/v9.rs
```diff
@@ -361,7 +362,25 @@ pub struct RawRowLevelSecurityDefV9 {
...
+#[derive(Debug, Clone, SpacetimeType)]
+#[sats(crate = crate)]
+#[cfg_attr(feature = "test", derive(PartialEq, Eq, PartialOrd, Ord))]
+pub struct RawColumnDefaultValueV9 {
+    /// Must point to a `ProductType` or an error will be thrown during validation.
```
Centril: Could you expand on this comment? Could this refer to a non-table? Or is this just mentioning the fact that tables are typed at product types? In that case the comment feels redundant and somewhat confusing.

crates/schema/src/auto_migrate.rs
```diff
@@ -109,6 +109,15 @@ pub enum AutoMigrateStep<'def> {
     ///
     /// This should be done before any new indices are added.
     ChangeColumns(<TableDef as ModuleDefLookup>::Key<'def>),
+    /// Add columns to a table, in a layout-INCOMPATIBLE way.
+    ///
+    /// This is a destructive operation that requires first running a `DisconnectAllUsers`.
```
Centril, line 114: Should this be a pre-check?

crates/schema/src/auto_migrate.rs
```diff
@@ -109,6 +109,15 @@ pub enum AutoMigrateStep<'def> {
...
+    ///
+    /// The added columns are guaranteed to be contiguous and at the end of the table. They are also
+    /// guaranteed to have default values set.
```
```suggestion
    /// The added columns are guaranteed to be contiguous
    /// and at the end of the table.
    /// They are also guaranteed to have default values set.
```
(see https://sembr.org/ -- makes doc comments more readable in code)

crates/schema/src/auto_migrate.rs
```diff
@@ -386,11 +398,18 @@ fn auto_migrate_table<'def>(plan: &mut AutoMigratePlan<'def>, old: &'def TableDe
     })
     .map(|col_diff| -> Result<_> {
         match col_diff {
-            Diff::Add { new } => Err(AutoMigrateError::AddColumn {
-                table: new.table_name.clone(),
-                column: new.name.clone(),
+            Diff::Add { new } => {
+                if new.default_value.is_some() {
+                    // row_type_changed, columns_added
```
```suggestion
                    // `row_type_changed`, `columns_added`
```

crates/schema/src/auto_migrate.rs
```diff
@@ -109,6 +109,15 @@ pub enum AutoMigrateStep<'def> {
...
     /// This should be done before any new indices are added.
     ChangeColumns(<TableDef as ModuleDefLookup>::Key<'def>),
+    /// Add columns to a table, in a layout-INCOMPATIBLE way.
+    ///
+    /// This is a destructive operation that requires first running a `DisconnectAllUsers`.
+    ///
+    /// The added columns are guaranteed to be contiguous and at the end of the table. They are also
+    /// guaranteed to have default values set.
+    ///
+    /// This suppresses any `ChangeColumns` steps for the same table.
```
```suggestion
    /// When this step is present,
    /// no `ChangeColumns` steps will be, for the same table.
```

crates/schema/src/auto_migrate.rs
```diff
@@ -417,22 +440,35 @@ fn auto_migrate_table<'def>(plan: &mut AutoMigratePlan<'def>, old: &'def TableDe
...
         }
     })
-    .collect_all_errors::<Any>();
+    .collect_all_errors::<ProductMonoid<Any, Any>>();
 
-    let ((), Any(row_type_changed)) = (type_ok, columns_ok).combine_errors()?;
+    let ((), ProductMonoid(Any(row_type_changed), Any(columns_added))) = (type_ok, columns_ok).combine_errors()?;
 
-    if row_type_changed {
+    if columns_added {
```
Centril, line 458:
```suggestion
    // If we're adding a column, we'll rewrite the whole table.
    // That makes any `ChangeColumns` moot, so we can skip it.
    if columns_added {
```

crates/schema/src/auto_migrate.rs
```diff
@@ -813,11 +869,13 @@ mod tests {
                     ("id", AlgebraicType::U64),
                     ("name", AlgebraicType::String),
                     ("count", AlgebraicType::U16),
+                    ("freshness", AlgebraicType::U32), // added column!
                 ]),
                 true,
             )
             // add column sequence
             .with_column_sequence(0)
+            .with_default_column_value(3, AlgebraicValue::U32(5)) // we need a new
```
Centril: dangling sentence?

crates/schema/src/auto_migrate.rs
Centril: Also, this doesn't seem to be tested for the same table. Would be good to add such a test.

crates/schema/src/def/validate/v9.rs
```diff
@@ -75,17 +76,22 @@ pub fn validate(def: RawModuleDefV9) -> Result<ModuleDef> {
...
+            let sched_exists = check_scheduled_reducers_exist(&tables, &reducers);
+            let default_values_work = misc_exports
+                .into_iter()
+                .map(|export| match export {
+                    RawMiscModuleExportV9::ColumnDefaultValue(fdv) => {
+                        validator.validate_column_default_value(&mut tables, fdv)
+                    }
+                    _ => unimplemented!("unknown misc export"),
+                })
+                .collect_all_errors::<()>();
```
Centril: Extract this to a function please.

crates/schema/src/def/validate/v9.rs
```diff
@@ -350,6 +356,55 @@ impl ModuleValidator<'_> {
...
     }
 
+    fn validate_column_default_value(
+        &mut self,
+        tables: &mut HashMap<Identifier, TableDef>,
+        cdv: RawColumnDefaultValueV9,
+    ) -> Result<()> {
+        let table_name = identifier(cdv.table.clone())?;
+
+        let table = tables
```
Centril:
```suggestion
        // Extract the table. We cannot make progress otherwise.
        let table = tables
```

crates/schema/src/def/validate/v9.rs
```diff
@@ -350,6 +356,55 @@ impl ModuleValidator<'_> {
...
+        if table.columns.len() <= cdv.col_id.idx() {
+            return Err(ValidationError::ColumnNotFound {
+                table: cdv.table.clone(),
+                def: cdv.table.clone(),
+                column: cdv.col_id,
+            }
+            .into());
+        }
+
+        let col = &mut table.columns[cdv.col_id.idx()];
```
Centril:
```suggestion
        // Get the column that a default is being added to.
        let Some(col) = table.columns.get_mut(cdv.col_id.idx()) else {
            return Err(ValidationError::ColumnNotFound {
                table: cdv.table.clone(),
                def: cdv.table.clone(),
                column: cdv.col_id,
            }
            .into());
        }
```

crates/schema/src/def/validate/v9.rs
```diff
@@ -350,6 +356,55 @@ impl ModuleValidator<'_> {
...
+        // First time we have a type for it, so decode it.
```
Centril:
```suggestion

        // First time the type of the default value is known, so decode it.
```

crates/schema/src/def/validate/v9.rs
```diff
@@ -350,6 +356,55 @@ impl ModuleValidator<'_> {
...
+            Ok(field_value) => Some(field_value),
+            Err(decode_error) => {
+                return Err(ValidationError::ColumnDefaultValueMalformed {
+                    table: cdv.table.clone(),
+                    col_id: cdv.col_id,
+                    err: decode_error,
+                }
+                .into())
+            }
+        };
```
Centril: Could use `.ok_or_else`

crates/schema/src/def/validate/v9.rs
```diff
@@ -350,6 +356,55 @@ impl ModuleValidator<'_> {
...
+        if col.default_value.is_some() {
```
Centril:
```suggestion

        // Ensure there's only one default value.
        if col.default_value.is_some() {
```

crates/schema/src/error.rs
```diff
@@ -120,6 +120,16 @@ pub enum ValidationError {
     TableNameReserved { table: Identifier },
     #[error("Row-level security invalid: `{error}`, query: `{sql}")]
     InvalidRowLevelQuery { sql: String, error: String },
+    #[error("Failed to deserialize default value for table {table} column {col_id}: {err}")]
+    ColumnDefaultValueMalformed {
```
Centril, line 124: Didn't see this variant exercised in any test. Could you please add one?

crates/schema/src/error.rs
```diff
@@ -120,6 +120,16 @@ pub enum ValidationError {
...
+        table: RawIdentifier,
+        col_id: ColId,
+        err: DecodeError,
+    },
+    #[error("Multiple default values for table {table} column {col_id}")]
+    MultipleColumnDefaultValues { table: RawIdentifier, col_id: ColId },
```
Centril, line 130: Same here, could you add a test that exercises this?

crates/sats/src/buffer.rs
Shubham8287: This has been used in [`ValidationError`](https://github.com/clockworklabs/SpacetimeDB/blob/49219830242e08bb21721a695d0a52917e51d800/crates/core/src/host/wasm_common.rs#L163), I think we do operations like deduplication with it before reporting.

crates/schema/src/auto_migrate.rs
Shubham8287: I don't think so, because pre-checks looks like conditions to verify, not actions to perform.

crates/schema/src/error.rs
Shubham8287: I added couple of tests for this -  - https://github.com/clockworklabs/SpacetimeDB/pull/2956/files#diff-e066f2b9fc7e350c9a3badf2ed36e847975e3f4cf24de6f37c5639358c1358f5R1084

crates/lib/src/db/raw_def/v9.rs
Shubham8287: I removed the comment, `PreductType` here indeed referring to Tables which makes it redundant.

crates/schema/src/auto_migrate.rs
Shubham8287: I added a `assert` condition for it in existing unit tests.

crates/lib/src/db/raw_def/v9.rs
Shubham8287: I checked the suggestion, and it looks like AlgebraicValue already implements `ser::Serialize` ([here](https://github.com/clockworklabs/SpacetimeDB/blob/49219830242e08bb21721a695d0a52917e51d800/crates/sats/src/ser/impls.rs#L98)).

However, it still needs to implement `impl_deserialize`, I am not sure how to do that.
If we decide it's necessary to add this, I'll need some guidance to get it done.

crates/schema/src/def/validate/v9.rs
```diff
@@ -350,6 +347,55 @@ impl ModuleValidator<'_> {
...
+                ValidationError::ColumnDefaultValueMalformed {
+                    table: cdv.table.clone(),
+                    col_id: cdv.col_id,
+                    err: decode_error,
+                }
+                .into()
+            });
+        // Ensure there's only one default value.
+        if col.default_value.is_some() {
+            return Err(ValidationError::MultipleColumnDefaultValues {
```
Centril: Looking at this again, this could be a non-fatal error I think.

crates/schema/src/def.rs
```diff
@@ -1072,4 +1080,51 @@ mod tests {
...
+    fn validate_new_column_with_malformed_value() {
+        let mut old_builder = RawModuleDefV9Builder::new();
+        old_builder
+            .build_table_with_new_type(
+                "Apples",
+                ProductType::from([("id", AlgebraicType::U64), ("count", AlgebraicType::U16)]),
+                true,
+            )
+            .with_default_column_value(1, AlgebraicValue::Bool(false))
+            .with_default_column_value(1, AlgebraicValue::U16(10))
```
Centril: Could you make this wrong as well and check that we have two errors?

crates/schema/src/def/validate/v9.rs
```diff
@@ -350,6 +347,55 @@ impl ModuleValidator<'_> {
...
+                def: cdv.table.clone(),
+                column: cdv.col_id,
+            }
+            .into());
+        };
+
+        // First time the type of the default value is known, so decode it.
+        let mut reader = &cdv.value[..];
+        let ty = WithTypespace::new(self.typespace, &col.ty);
+        let field_value: Result<AlgebraicValue> =
```
Centril, line 375: Should this value not be returned?

crates/lib/src/db/raw_def/v9.rs
Centril: In this case, what you are looking for is `impl<'de> DeserializeSeed<'de> for WithTypespace<'_, AlgebraicType>`, as you cannot deserialize an AV without knowing its type.

However, it feels rather weird that we serialize the value only to later deserialize it. When we already have the AV, we can keep it in that form and avoid the roundtrip around BSATN. You do need to *type check* the AV against the AT, which is functionality that needs to be added and would have the form `fn type_check(val: &AlgebraicValue, expected_ty: &AlgebraicType) -> bool/Result`.

Shubham8287: It’s not clear to me what you meant.

If you meant:

```rust
value: WithTypespace<'_, AlgebraicType>
```
I don't know how will that work, because `WithTypespace` doesn’t actually hold a value.

If instead you meant:

```rust
value: ValueWithType<'_, AlgebraicValue>
```
— that make some sense, but `ValueWithType` isn’t a `SpacetimeType`, and as it holds a reference to `Typespace`, we will run into self-referential problems inside `RawModuleDefV9`.

Did you mean to keep `value` as an `AlgebraicValue`, and then manually implement `sats::de::DeserializeSeed` for `RawColumnVDefaultValueV9` by building `WithTypespace<'_, AlgebraicType>` using `col_id`and `self.module_def.typespace`?

Also, I don't think implementing `sats::de::DeserializeSeed` will let us derive `SpacetimeType`

crates/lib/src/db/raw_def/v9.rs
```diff
@@ -361,7 +362,24 @@ pub struct RawRowLevelSecurityDefV9 {
...
+pub struct RawColumnDefaultValueV9 {
+    pub table: RawIdentifier,
+
+    /// Must be the index of a valid column within `ty`.
+    pub col_id: ColId,
+
+    /// A BSATN-encoded [`AlgebraicValue`] valid at `ty`.
+    /// (We can't use `AlgebraicValue` directly because it isn't serializable!)
+    pub value: Box<[u8]>,
+}
```
Centril:
```suggestion
/// Marks a particular table's column as having a particular default.
#[derive(Debug, Clone, SpacetimeType)]
#[sats(crate = crate)]
#[cfg_attr(feature = "test", derive(PartialEq, Eq, PartialOrd, Ord))]
pub struct RawColumnDefaultValueV9 {
    /// Identifies which table that has the default value.
    /// This corresponds to `name` in `RawTableDefV9`.
    pub table: RawIdentifier,

    /// Identifies which column of `table` that has the default value.
    pub col_id: ColId,

    /// A BSATN-encoded [`AlgebraicValue`] valid at the table column's type.
    /// (We cannot use `AlgebraicValue` directly as it isn't `Spacetimetype`.)
    pub value: Box<[u8]>,
}
```

crates/schema/src/def/validate/v9.rs
Shubham8287: Umm, How should we proceed if there are multiple default values?

crates/schema/src/def/validate/v9.rs
Shubham8287: yeah, made `validation_*` method non mutable and returning value.

Shubham8287:
@Centril , Do you expect me to do anything else in this PR?

crates/schema/src/def/validate/v9.rs
Centril: You pick the last default as the value but you still emit errors for each duplicate, so the update never happens. This allows progress to be made and for other errors to be reported.

Centril:
> @Centril , Do you expect me to do anything else in this PR?

If you could please address this: https://github.com/clockworklabs/SpacetimeDB/pull/2956#discussion_r2286134795
But in the interest of time, it can be done in a follow up.

crates/schema/src/def/validate/v9.rs
Shubham8287: This is actually like that, we pick the first default and emit eorror for each duplicate. This still allows progress for other errors to be reported.

PR #2957, opened 2025-07-19
Implemented newtype pattern support for tuples

egormanga:
# Description of Changes

To be able to do this:
```rust
#[derive(SpacetimeType)]
pub struct Position(vecmath::Vector3<f64>);
```
(previously, the derive macro would panic.)

**This is an early attempt!** I need help in figuring out all details on how to properly implement deserialization.

Also, currently this fails with errors like these:
```rust
error at typespace ref `&1`: internal codegen error: all type elements require names: (0: Array<F64>)

error at typespace ref `&2`: internal codegen error: non-special product or sum type (0: F64, 1: Array<F64>) cannot be used to generate a client type use
```

**Unknown.**

### ?…

# Testing

- [x] SpacetimeDB compiles
- [x] The module builds
- [ ] ~Publishing works~

### DIFF
--- a/crates/bindings-macro/src/sats.rs
+++ b/crates/bindings-macro/src/sats.rs
@@ -380,8 +380,9 @@ pub(crate) fn derive_deserialize(ty: &SatsType<'_>) -> TokenStream {
 
             let n_fields = fields.len();
 
-            let field_names = fields.iter().map(|f| f.ident.unwrap()).collect::<Vec<_>>();
-            let field_strings = fields.iter().map(|f| f.name.as_deref().unwrap()).collect::<Vec<_>>();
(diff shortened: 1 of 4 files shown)


bfops:
Thank you for submitting this! We'll work on getting it reviewed (along with your other PRs)

egormanga:
It's a bit closer now with the support for product client fields.

cloutiertyler:
@egormanga what types does this generate in C#, TypeScript, and Rust? I think we'd need those to have support before we merge this. I am going to close this PR and open a ticket with you as the assignee.

If you are motivated to get this feature in then we'd need a philosophy for languages that don't have tuples or tuple structs. In particular C# and C++. Something like regular product types with field names like `_0`, `_1`, etc could be acceptable.

Issue #1599, opened 2024-08-16
Auth: Add subscription support for 3 and 4 way joins


joshua-spacetime:
The implementation is described in incremental subscription update proposal.

egormanga:
I really do need this functionality. Blocks my use-case, although not specifically related to subscriptions, but to one-shot queries through SDK which do require using subs.

Is it possible to work around this limitation at least ad-hoc, by extending the `match` in question (`subscription/src/lib.rs`) with 3-way hands?

egormanga:
Is there a way to use one-shot queries (as opposed to subscriptions) via client SDK to work around this issie for the time being?

krisajenkins:
See also #5214

PR #2960, opened 2025-07-19
Implemented one-off query support for TypeScript SDK

egormanga:
# Description of Changes

Was necessary (at least) because of #1599.

None(?)

Needs docs, that's for certain.

### 4:
Doesn't seem to break anything, but who knows. I'm not sure about this one.

Also, performance impact? It _should_ only get better, but again I have no info on that.

# Testing

- [x] Manual testing
- [x] Use case fit
- [ ] Unit tests

### DIFF
--- a/crates/bindings-typescript/src/sdk/client_cache.ts
+++ b/crates/bindings-typescript/src/sdk/client_cache.ts
@@ -3,7 +3,7 @@ import type { UntypedTableDef } from '../lib/table.ts';
 import type { UntypedRemoteModule } from './spacetime_module.ts';
 import { type TableCache, TableCacheImpl } from './table_cache.ts';
 
-type TableName<SchemaDef> = [SchemaDef] extends [UntypedSchemaDef]
+export type TableName<SchemaDef> = [SchemaDef] extends [UntypedSchemaDef]
   ? TableNamesOf<SchemaDef>
   : string;
 
--- a/crates/bindings-typescript/src/sdk/client_cache.ts
+++ b/crates/bindings-typescript/src/sdk/client_cache.ts
@@ -1,3 +1,4 @@
+import type { DbContext } from './db_context.ts';
 import type { TableNamesOf, UntypedSchemaDef } from '../lib/schema.ts';
 import type { UntypedTableDef } from '../lib/table.ts';
 import type { UntypedRemoteModule } from './spacetime_module.ts';
@@ -83,6 +84,11 @@ export class ClientCache<RemoteModule extends UntypedRemoteModule> {
    * The tables in the database.
    */
   readonly tables = new TableMap<RemoteModule>();
+  private ctx: DbContext<RemoteModule>;
+
+  constructor(ctx: DbContext<RemoteModule>) {
+    this.ctx = ctx;
+  }
 
   /**
    * Returns the table with the given name.
@@ -120,6 +126,7 @@ export class ClientCache<RemoteModule extends UntypedRemoteModule> {
     }
 
     const newTable = new TableCacheImpl<RemoteModule, N>(
+      this.ctx,
       tableDef
     ) as TableCache<RemoteModule, N>;
     this.tables.set(name, newTable);
--- a/crates/bindings-typescript/src/sdk/client_cache.ts
+++ b/crates/bindings-typescript/src/sdk/client_cache.ts
@@ -23,7 +23,7 @@ type TableCacheForTableName<
  * This is a helper class that provides a mapping from table names to their corresponding TableCache instances
  * while preserving the correspondence between the key and value type.
  */
-class TableMap<RemoteModule extends UntypedRemoteModule> {
+export class TableMap<RemoteModule extends UntypedRemoteModule> {
   private readonly map: Map<
     string,
     TableCacheForTableName<RemoteModule, TableName<RemoteModule>>
--- a/crates/bindings-typescript/src/sdk/client_table.ts
+++ b/crates/bindings-typescript/src/sdk/client_table.ts
@@ -9,6 +9,7 @@ import type {
 import type { ColumnBuilder } from '../lib/type_builders';
 import type { Prettify } from '../lib/type_util';
 import type { TableDefForTableName } from './client_cache';
+import type { DbContext } from './db_context.ts';
 import type { EventContextInterface } from './event_context';
 import type { UntypedRemoteModule } from './spacetime_module';
 
@@ -46,6 +47,13 @@ export type ClientTableMethods<
   RemoteModule extends UntypedRemoteModule,
   TableName extends TableNamesOf<RemoteModule>,
 > = {
+  remoteQuery(
+    ctx: DbContext<RemoteModule>,
+    filters: string
+  ): Promise<IterableIterator<
+    Prettify<RowType<TableDefForTableName<RemoteModule, TableName>>>
+  >>;
+
   /**
    * Registers a callback to be invoked when a row is inserted into the table.
    */
--- a/crates/bindings-typescript/src/sdk/client_table.ts
+++ b/crates/bindings-typescript/src/sdk/client_table.ts
@@ -47,10 +47,7 @@ export type ClientTableMethods<
   RemoteModule extends UntypedRemoteModule,
   TableName extends TableNamesOf<RemoteModule>,
 > = {
-  remoteQuery(
-    ctx: DbContext<RemoteModule>,
-    filters: string
-  ): Promise<IterableIterator<
+  remoteQuery(filters: string): Promise<IterableIterator<
     Prettify<RowType<TableDefForTableName<RemoteModule, TableName>>>
   >>;
 
--- a/crates/bindings-typescript/src/sdk/db_connection_impl.ts
+++ b/crates/bindings-typescript/src/sdk/db_connection_impl.ts
@@ -46,7 +46,6 @@ import {
 import { WebsocketDecompressAdapter } from './websocket_decompress_adapter.ts';
 import type { WebsocketTestAdapter } from './websocket_test_adapter.ts';
 import {
-  QueryBuilderImpl,
   QueryHandleImpl,
   QueryManager,
   type QueryEvent,
@@ -379,16 +378,11 @@ export class DbConnectionImpl<RemoteModule extends UntypedRemoteModule>
       reducers: this.reducers,
       setReducerFlags: this.setReducerFlags,
       isActive: this.isActive,
-      queryBuilder: this.queryBuilder.bind(this),
       subscriptionBuilder: this.subscriptionBuilder.bind(this),
       disconnect: this.disconnect.bind(this),
       event,
     };
   }
-
-  queryBuilder = (): QueryBuilderImpl<RemoteModule> => {
-    return new QueryBuilderImpl(this);
-  };
 
   registerQuery(
     handle: QueryHandleImpl<RemoteModule>,
--- a/crates/bindings-typescript/src/sdk/db_connection_impl.ts
+++ b/crates/bindings-typescript/src/sdk/db_connection_impl.ts
@@ -406,7 +406,7 @@ export class DbConnectionImpl<RemoteModule extends UntypedRemoteModule>
     this.#sendMessage(
       ClientMessage.OneOffQuery({
         queryString: querySql,
-        messageId: new Uint8Array([queryId]),
+        messageId: new Uint8Array(new Uint32Array([queryId]).buffer),
       })
     );
     return queryId;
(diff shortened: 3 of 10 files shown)


sdks/typescript/packages/sdk/src/db_connection_impl.ts
```diff
@@ -259,6 +271,29 @@ export class DbConnectionImpl<
     return queryId;
   };
 
+  queryBuilder = (): QueryBuilderImpl => {
+    return new QueryBuilderImpl(this);
+  };
```
gefjon: The API we use in the C# SDK for running remote queries has them as methods of the `TableHandle`, rather than on the `DbContext`. This provides some amount of type safety, as the method inserts the `SELECT` clause into the query and also encodes the TypeScript type of the rows returned.

gefjon: Also, we use the phrase "remote query" to describe this operation in the user-facing API. This emphasizes that this is a query that bypasses the client cache and goes directly to the remote database.

sdks/typescript/packages/sdk/src/query_builder_impl.ts
```diff
@@ -0,0 +1,128 @@
...
+    query_sql: string
+  ): QueryHandleImpl<DBView, Reducers, SetReducerFlags> {
+    return new QueryHandleImpl(
+      this.db,
+      query_sql,
+      this.#onResolved,
+      this.#onError
+    );
+  }
+}
```
gefjon: Remote queries are an odd and mismatched part of our API. Rather than using a builder and callbacks, they are `async` methods which directly return the queried rows. What you've designed here would be better, in that it more closely matches the way we construct subscriptions. Unfortunately, though, the `async` version is a part of the C# client SDK's stable API, and we are committed to continuing to support it. We're also committed to minimizing differences between the SDKs, which means that TypeScript should expose the same interface.

gefjon (dismissed):
Thanks so much for your contribution! This is a feature that we're very interested in merging. As discussed in comments below, we'd prefer to have the TypeScript client SDK expose the same interface for remote queries as the C# client SDK. That interface is:

Each `TableHandle` returned by the methods on `ctx.db` has a method `remoteQuery`. That method looks roughly like:

```typescript
class TableHandleImpl<Row> {
  async remoteQuery(filters: string): Row[] {
    const tableName = this.tableName();
    const fullQuery = `SELECT ${tableName}.* FROM {tableName} ${filters}`;
    // Construct a promise, register it in the `DbConnection`, send a `OneOffQuery` message, return the promise.
    // When the `DbConnection` receives the response, it parses the rows and fulfills the promise with them.
  }
}
```

Client code might invoke this like:

```typescript
async getPlayersInFaction(ctx, factionId): Player[] {
  const interestingPlayers = await ctx.db.player().remoteQuery(`WHERE faction = ${factionId}`);
  return interestingPlayers;
}
```

sdks/typescript/packages/sdk/src/query_builder_impl.ts
egormanga: To be perfectly honest, `async` remote queries seem to be a more intuitive and straightforward solution, even though they differ from the subscriptions. Being able to use queries anywhere in the code without registering a callback is pretty convenient.

I see your appreciation and in that regard will surely replace the current implementation with a table method. On the other hand, might it be useful to keep both, e.g. even providing the builder version across all implementations?

sdks/typescript/packages/sdk/src/db_connection_impl.ts
egormanga: At the same time, formatting `SELECT`s restricts the provided functionality to querying only. What if a client needs to execute something else, e.g. `INSERT`, perhaps with the owner's identity?

sdks/typescript/packages/sdk/src/query_builder_impl.ts
egormanga: The main showstopper for your proposed approach is the fact that the entire SDK is currently callback-based, thus rendering it pretty impossible to implement an `async` querying mechanism in a useful fashion.

For instance, if I need to fire a query inside a subscription callback, I won't be able to `await` for it inside that block. Since I'd need to wrap it in a `Promise` anyway, similar to the current API, why bother implementing this at all?

egormanga:
It turned out to be a really convenient implementation of an async `ctx.table.remoteQuery()` using the builder I already implemented. Done just that.

sdks/typescript/packages/sdk/src/db_connection_impl.ts
egormanga: Okay, there I discovered (in #2968) that SpacetimeDB does currently not support anything other than `SELECT` even in one-shot queries.

sdks/typescript/packages/sdk/src/table_cache.ts
```diff
@@ -57,10 +65,21 @@ export class TableCache<RowType = any> {
...
-  iter(): any[] {
+  iter(): RowType[] {
     return Array.from(this.rows.values()).map(([row]) => row);
   }
 
+  remoteQuery(ctx: DbContext, filters: string): Promise<RowType[]> {
+    return new Promise((resolve, reject) => {
+      const name = this.name();
+
+      ctx.queryBuilder()
```
egormanga: As you might've noticed, this requires having `ctx` passed in as an argument, as we've nowhere to find it from here. Suggestions are welcomed.

sdks/typescript/packages/sdk/src/db_connection_impl.ts
gefjon: Correct, we have no interest in generalizing one-off queries to support SQL DML operations like `INSERT`, `DELETE`, `UPDATE`, &c.

sdks/typescript/packages/sdk/src/query_builder_impl.ts
gefjon: > On the other hand, might it be useful to keep both, e.g. even providing the builder version across all implementations?

We'd need a more involved design process in order to make this happen. Going through such a design process with someone external to Clockwork Labs poses a number of problems, and seems unlikely to be pleasant for anyone involved.

> if I need to fire a query inside a subscription callback, I won't be able to await for it inside that block.

Correct, you'll instead do `ctx.db.myTable().remoteQuery("WHERE whatever").then(rows => processRows(rows));`.

sdks/typescript/packages/sdk/src/table_cache.ts
gefjon: You will need to modify the table handle objects so that they contain references to everything in `ctx` necessary to complete this operation. This is likely to be easier if you bypass your `queryBuilder` interface, which will not be part of the public API anyways.

sdks/typescript/packages/sdk/src/query_builder_impl.ts
egormanga: Sure, you may design the thing as you wish (or, may not at all). I'll look into the possibility of refactoring the implementation to not depend on this custom solution (and it's clearly possible, but however be the cleanest wins).

egormanga: > `.then()`

This doesn't seem possible inside loops, though.

sdks/typescript/packages/sdk/src/table_cache.ts
egormanga: I've implemented this the least invasive way, not requiring additional changes to the codegen. It is possible, though, to pass `ctx` at a later stage (not storing the reference to it in `ClientCache`), with a subtle change in the codegen (namely, in `RemoteTables` gen), if you wish so better.

sdks/typescript/packages/sdk/src/query_builder_impl.ts
egormanga: The builder showed to be more convenient even in the internal code, so I'm suggesting keeping it.

gefjon: > This doesn't seem possible inside loops, though.

I don't understand why being inside a loop would be in any way different.

gefjon (commented):
I would prefer not to expose your query builder as a new public user-facing API. Even if it's not documented, people will inevitably stumble upon it and use it in their code. This will have the twin problems of exposing them to errors like the one you saw in #2968 , and forcing us to maintain the new API in a compatible way in order to preserve their code. To avoid this, I recommend moving it from a class method to a free function, and making sure said free function is not exported from the package.

sdks/typescript/packages/sdk/src/query_builder_impl.ts
egormanga: Say you need to fetch some relation for each row (basically, `.map()` them) and do something with the resulting array. You do a query, `.then()` on it, and inside that you iterate over the rows. On each iteration, you issue another query, and that's where you only can `await` it and then use `Promise.all()` on the whole `.map()`, otherwise it's a callback inside a loop and that's not possible without refactoring.

Example for my API (assume inside a promise):
```ts
ctx.subscriptionBuilder()
    .onApplied(ctx => {
        ctx.queryBuilder()
            .onResolved((ctx, tables) => {
                const messages = tables.get('message');
                resolve(messages?.iter().map(message => {
                    const user = ctx.db.user.id.find(message.authorId);
                    return {user, message};
                });
            })
            .query(`SELECT * FROM message;`);
    })
    .subscribe(`SELECT * FROM user;`);
```

Examples for your API:
```ts
ctx.subscriptionBuilder()
    .onApplied(ctx => {
        const messages = await ctx.db.message.remoteQuery(`SELECT * FROM message;`);
        //               ^^^^^ error: await inside non-async function
        resolve(…);
    })
    .subscribe(`SELECT * FROM user;`);

ctx.subscriptionBuilder()
    .onApplied(async ctx => {
        //     ^^^^^ sure possible, but who will ever await it?
        const messages = await ctx.db.message.remoteQuery(…);
        resolve(…);  // will never get called
    })
    .subscribe(…);
```

gefjon: Your second example:

```typescript
ctx.subscriptionBuilder()
  .onApplied(ctx => {
    ctx.db.message.remoteQuery("WHERE foo = bar")
      .then(messages => {
        resolve(doStuffWith(messages));
      });
  )
  .subscribe(`SELECT * FROM user;)`;
```

Your first example:

```typescript
ctx.subscriptionBuilder()
  .onApplied(ctx => {
    ctx.db.message.remoteQuery("")
      .then(messages => {
        resolve(messages.iter().map(message => {
          const user = ctx.db.user.id.find(message.authorId);
          return { user, message };
        });
      });
  .subscribe(`SELECT * FROM user;`);
```

Keep in mind that in JS, promises are always run by the executor, and `async`/`await` is syntax sugar over the `then` callbacks. This is unlike e.g. Rust, where promises famously do nothing unless awaited.

egormanga: Alright, you convinced me, thank you. Let me make the API private in a sec.

sdks/typescript/packages/sdk/src/db_connection_impl.ts
```diff
@@ -259,6 +270,25 @@ export class DbConnectionImpl<
...
+  ): number {
+    const queryId = this.#getNextQueryId();
+    this.#queryManager.queries.set(queryId, {
+      handle,
+      emitter: handleEmitter,
+    });
+    this.#sendMessage(
+      ClientMessage.OneOffQuery({
+        queryString: querySql,
+        messageId: new Uint8Array([queryId]),
```
jsdt: This will be a problem after `255` queries. Using a 4 byte integer would probably be safe enough, but we would need to convert it to bytes with something like:

    const queryId = this.#getNextQueryId();
    const buffer = new ArrayBuffer(4);
    const view = new DataView(buffer);
    view.setUint32(0, queryId, true); // true = little-endian

sdks/typescript/packages/sdk/src/db_connection_impl.ts
```diff
@@ -700,6 +754,60 @@ export class DbConnectionImpl<
...
+          query?.emitter.emit(
+            'resolved',
+            queryEventContext,
+            tables.keys().reduce((map, table) =>
+              map.set(table.name(), table), new Map()),
+            message.totalHostExecutionDuration
+          );
+          for (const callbacks of tables.values())
+            for (const callback of callbacks)
+              callback.cb();
```
jsdt: I don't think we want to call the row-level callbacks for one-off queries, since any rows that are inserted here will never have corresponding delete callbacks.

Since we aren't calling those, we don't need to have the `#applyTableState` function. We can just get a list of rows for each table.

sdks/typescript/packages/sdk/src/db_connection_impl.ts
egormanga: Sure, thanks for spotting it! I added a single-byte counter as a placeholder and totally forgot to widen it.

sdks/typescript/packages/sdk/src/db_connection_impl.ts
egormanga: I agree we can remove the callbacks. But `applyTableState()` processes the insertion events from the response, isn't it necessary to compute the rows?

egormanga:
@jsdt Hey! Any luck getting an eye on this? I really wish to see it upstreamed :)

egormanga:
Updated for the new TS SDK :)

egormanga:
@jsdt Hi! Thank you for finally carving some time to have a look at this!

I can rebase instead, if that's necessary for getting this merged. That's what I often do regardless :>

Please let me know if there may any additional issues be.

Upd. The `master`-merge message disappeared from PR history after rebasing, if wondering.

cloutiertyler:
@egormanga would you be willing to fix these conflicts and update this PR? It looks like it's not too out of date. We're going through to try and merge old PRs.

PR #2964, opened 2025-07-21
Disable musl builds

bfops:
# Description of Changes

Disable musl builds because our new v8 dependency (https://github.com/clockworklabs/SpacetimeDB/pull/2939) is not building successfully, and we currently don't want to investigate it.

None

1

# Testing

None


jdetter (approved):
Since not many people are using musl and we need to move forward with V8 we're going to disable for now 👍

PR #2966, opened 2025-07-21
Update tungstenite to get client red performance improvement

ResuBaka:
# Description of Changes

This updates the tungstenite lib to get two changes which make good improvements to the performance and reduces cpu usage.

- 

1

# Testing

- [x] Check performance
- [x] Check cpu usage from before after

For me testing it locally in my application my whole application used like 2x to 3x less cpu and the the over all time spend.

Before the Update. (I tried to have the time span for the profile but they could be off by like total time the application did run by 30s)

image

After the update.

image


ResuBaka:
Oh it looks like I need to check how I can run spacetime locally. As before I used the docker compose setup but that does not want to start with other errors then in the ci. E.g. it can not find all the crates and if I add them it is missing the feature setup in one crate.

Is the docker setup still some supported or preferred way or should it be just run directly in development?

kim:
@ResuBaka `docker compose up` should work. If you want the binary to be built in release mode, you'll need to edit the `docker-compose.yml` file and set `CARGO_PROFILE: release`.

Alternatively, you can [install from source](https://github.com/clockworklabs/SpacetimeDB?tab=readme-ov-file#macos--linux) and run `spacetimedb start`.

CI errors are because tungstenite's `Error` enum changed.

ResuBaka:
Then maybe it is an issue with using the docker compose on linux?

As I am getting this output:
```
Attaching to node-1
node-1  | Finding local deps failed: Failed to execute `cargo metadata`: `cargo metadata` exited with an error: error: failed to load manifest for workspace member `/usr/src/app/crates/auth`
node-1  | referenced by workspace at `/usr/src/app/Cargo.toml`
node-1  |
node-1  | Caused by:
node-1  |   failed to load manifest for dependency `spacetimedb-lib`
node-1  |
node-1  | Caused by:
node-1  |   failed to load manifest for dependency `spacetimedb-memory-usage`
node-1  |
node-1  | Caused by:
node-1  |   failed to read `/usr/src/app/crates/memory-usage/Cargo.toml`
node-1  |
node-1  | Caused by:
node-1  |   No such file or directory (os error 2)
node-1  |
node-1  | [Running 'cargo run start --data-dir=/stdb/data --jwt-pub-key-path=/etc/spacetimedb/id_ecdsa.pub --jwt-priv-key-path=/etc/spacetimedb/id_ecdsa']
node-1  | error: failed to load manifest for workspace member `/usr/src/app/crates/auth`
node-1  | referenced by workspace at `/usr/src/app/Cargo.toml`
node-1  |
node-1  | Caused by:
node-1  |   failed to load manifest for dependency `spacetimedb-lib`
node-1  |
node-1  | Caused by:
node-1  |   failed to load manifest for dependency `spacetimedb-memory-usage`
node-1  |
node-1  | Caused by:
node-1  |   failed to read `/usr/src/app/crates/memory-usage/Cargo.toml`
node-1  |
node-1  | Caused by:
node-1  |   No such file or directory (os error 2)
node-1  | [Finished running. Exit status: 101]
```

ResuBaka:
It is also the same on a fresh clone.

kim:
> Then maybe it is an issue with using the docker compose on linux?

I don't know, mine works fine. Do you maybe have the deprecated `docker-compose` installed, instead of the built-in `docker compose`?

ResuBaka:
> > Then maybe it is an issue with using the docker compose on linux?
> 
> I don't know, mine works fine. Do you maybe have the deprecated `docker-compose` installed, instead of the built-in `docker compose`?

I will check later but I should run the latest arch linux version e.g. 2.38.2 of docker compose. Maybe I need to rebuild the docker image it would want to build. I will try that later to see if that fixes the issue.


On the other note, is that to be expected that Unity test fails? Or is it okay as I have not changed anything that should make that one fail or not even run correctly.

kim (approved):

kim:
> is that to be expected that Unity test fails

Yeah it isn't allowed to read the Unity license data, because your PR is coming from outside the `clockworklabs` org. I'll have that sorted out.

Thanks for your contribution!

ResuBaka:
Now after using **docker compose up --build** it is working again. Should maybe have tried that first but I did not think about that.

PR #2967, opened 2025-07-21
Moved Remove operation to start of Table.Apply

### LINKED ISSUE #343: Incremental updates throw error - Item with the same key has already been added
From @jdetter :

>  I wrote an app which uses the "SELECT * FROM *" subscription query (only that single query). The initial subscription update seems fine but then when I get incremental updates I start getting a bunch of these:
```
ArgumentException: An item with the same key has already been added. Key: C2004B5A204039758E68EA984786AE422B9BBD624FF6169B3E726D421272F224
System.Collections.Generic.Dictionary`2[TKey,TValue].TryInsert (TKey key, TValue value, System.Collections.Generic.InsertionBehavior behavior) (at <467a840a914a47078e4ae9b0b1e8779e>:0)
System.Collections.Generic.Dictionary`2[TKey,TValue].Add (TKey key, TValue value) (at <467a840a914a47078e4ae9b0b1e8779e>:0)
SpacetimeDB.RemoteTableHandle`2+UniqueIndexBase`1[EventContext,Row,Column].<.ctor>b__1_0 (Row row) (at ./Library/PackageCache/com.clockworklabs.spacetimedbsdk@f74705ae05/src/Table.cs:83)
SpacetimeDB.EventHandling.AbstractEventHandler`1[T].Invoke (T value) (at ./Library/PackageCache/com.clockworklabs.spacetimedbsdk@f74705ae05/src/EventHandling/AbstractEventHandler.cs:29)
SpacetimeDB.RemoteTableHandle`2[EventContext,Row].SpacetimeDB.IRemoteTableHandle.Apply (SpacetimeDB.IEventContext context, SpacetimeDB.MultiDictionaryDelta`2[TKey,TValue] multiDictionaryDelta) (at ./Library/PackageCache/com.clockworklabs.spacetimedbsdk@f74705ae05/src/Table.cs:310)
SpacetimeDB.DbConnectionBase`3[DbConnection,Tables,Reducer].ApplyUpdate (SpacetimeDB.IEventContext eventContext, SpacetimeDB.DbConnectionBase`3+ParsedDatabaseUpdate[DbConnection,Tables,Reducer] dbOps) (at ./Library/PackageCache/com.clockworklabs.spacetimedbsdk@f74705ae05/src/SpacetimeDBClient.cs:770)
SpacetimeDB.DbConnectionBase`3[DbConnection,Tables,Reducer].ApplyMessage (SpacetimeDB.DbConnectionBase`3+ParsedMessage[DbConnection,Tables,Reducer] parsed) (at ./Library/PackageCache/com.clockworklabs.spacetimedbsdk@f74705ae05/src/SpacetimeDBClient.cs:904)
SpacetimeDB.DbConnectionBase`3[DbConnection,Tables,Reducer].FrameTick () (at ./Library/PackageCache/com.clockworklabs.spacetimedbsdk@f74705ae05/src/SpacetimeDBClient.cs:1076)
GameManager.Update () (at Assets/GameManager.cs:67)
```

It's very consistent. I've regenerated the bindings and I've updated to the latest released version of the SDK.  Open the project in Unity (FruitMerge3D) and click inside of the box that appears. Once you click you'll start getting errors.

rekhoff:
# Description of Changes

Intends to resolve [issue 343](https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/issues/343) of the C# SDK.

In issue 343, the example project uses fails due the `PlayerFruitState` table being defined with an `Identity` value is marked as `unique` rather than `primary_key` as is typically the case. Without defining a `primary key`, data from all rows is used as the primary key. 

In the sample project, when the `PlayerFruitState` table was being updated, the `Identity` value would stay the same, and the other values would change. This should still be a supported operation, and would look like a removal of the Identity row with the old value and an addition of the Identity row with the new value, rather than a single update operation.
Note: If the `Identity` was marked as `primary_key`, this would be an update operation, because only the `Identity` value would be used for the key, rather than all rows.

Because of the existing order of operations in a `Table.cs`'s `Apply` method, the new row being added to the `PlayerFruitState` table prior to the old row being removed. Due to the `Identity` value having the `unique` constraint, this caused the initial add to throw an error during the add operation.

This change simply moves the Remove operation before Add, which is best-practice for these types of operations.

Not API breaking.

1

# Testing

- [X] Tested against "FruitMerge3D" sample project

### DIFF
--- a/sdks/csharp/src/Table.cs
+++ b/sdks/csharp/src/Table.cs
@@ -303,6 +303,15 @@ namespace SpacetimeDB
             // This is a local operation -- it only looks at our indices and doesn't invoke user code.
             // So we don't need to wait for other tables to be updated to do it.
             // (And we need to do it before any PostApply is called.)
+            // Reminder: We need to loop through the removed entries to delete them prior to inserting the new entries,
+            // in order to avoid keys an error with the same key already added.
+            foreach (var (_, value) in wasRemoved)
+            {
+                if (value is Row oldRow)
+                {
+                    OnInternalDeleteHandler.Invoke(oldRow);
+                }
+            }
             foreach (var (_, value) in wasInserted)
             {
                 if (value is Row newRow)
@@ -333,14 +342,6 @@ namespace SpacetimeDB
                 else
                 {
                     throw new Exception($"Invalid row type for table {RemoteTableName}: {newValue.GetType().Name}");
-                }
-            }
-
-            foreach (var (_, value) in wasRemoved)
-            {
-                if (value is Row oldRow)
-                {
-                    OnInternalDeleteHandler.Invoke(oldRow);
                 }
             }
         }


jdetter (approved):
I gave this a try using my fruit merge 3D example and it works now, thanks Ryan! 👍

PR #2971, opened 2025-07-22
V8: add `FromValue`, error traits, roundtrip tests

Centril:
# Description of Changes

The primary goal here is to add the trait `FromValue` and its implementation.
These are then tested in combination with `ToValue`.
Some error handling functionality is also added.

None

1

# Testing

Proptests and edge case tests are added, testing that a value `x: T` is the same as `from_value(to_value(x))`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,7 +61,8 @@ jobs:
         if: runner.os == 'Windows'
       - name: Run smoketests
         # Note: clear_database and replication only work in private
-        run: python -m smoketests ${{ matrix.smoketest_args }} -x clear_database replication
+        # zz_docker disabled temporarily until https://github.com/clockworklabs/SpacetimeDB/issues/2965
+        run: python -m smoketests ${{ matrix.smoketest_args }} -x clear_database replication zz_docker
       - name: Stop containers (Linux)
         if: always() && runner.os == 'Linux'
         run: docker compose down


crates/core/src/host/v8/from_value.rs
```diff
@@ -0,0 +1,185 @@
...
+int32_from_value!(Uint32, u32);
+int32_from_value!(Int32, i8);
+int32_from_value!(Int32, i16);
+int32_from_value!(Int32, i32);
+
+// `FromValue for f32, f64`.
+macro_rules! float_from_value {
+    ($rust_ty:ty) => {
+        impl_from_value!($rust_ty, (val, scope) => {
+            cast!(scope, val, Number, "number for `{}`", stringify!($rust_ty)).map(|n| n.value() as _)
```
gefjon, line 85: What will this do for `f32` when the source value has more significant bits in its mantissa than `f32` can hold? Truncate? Round? Error?

Centril: This is defined here: https://doc.rust-lang.org/reference/expressions/operator-expr.html#r-expr.as.numeric.float-narrowing

> Casting from an f64 to an f32 will produce the closest possible f32
> - if necessary, rounding is according to roundTiesToEven mode ***
>  - on overflow, infinity (of the same sign as the input) is produced
>
> *** as defined in IEEE 754-2008 §4.3.1: pick the nearest floating point number, preferring the one with an even least significant digit if exactly halfway between two floating point numbers.

I've left a note in the code.

gefjon (approved):

Issue #2965, opened 2025-07-21
Fix flaky restart-related tests


bfops:
We have several tests that are flaky semi-regularly in our smoketests CI. Let's investigate whether the tests are flaky or if SpacetimeDB is flaky, and fix accordingly.

- `test_restart_disconnects`
- `smoketests.tests.zz_docker.DockerRestartModule.test_restart_module`
- `smoketests.tests.zz_docker.DockerRestartSql.test_restart_module`

kim:
I think it might be caused by the tests resuming too early after a restart, when the server hasn't fully come up again. It's unclear though why this is the case.

kim:
I see why it's not working, assigning myself.

bfops:
Fixed in https://github.com/clockworklabs/SpacetimeDB/pull/2977

PR #2974, opened 2025-07-22
ci: Temporarily disable zz_docker tests

kim:
Fix for #2965 will be submitted soon.
To unblock PRs, disable the smoketests that restart the server.


jdetter (approved):
We discussed this during core standup. We're doing this until the test is fixed 👍

PR #2977, opened 2025-07-23
smoketests: Fix server restarts and make more robust

kim:
Smoketests that restart the server became flaky recently as the readiness
probing was broken. This patch fixes the issue and makes the process more
robust, namely:

- Label the compose services to check for readiness (i.e. spacetimedb
  processes) as `app: spacetimedb`.
- When enumerating the containers, consider this label, as well as the
  current compose file.
- For each spacetimedb container, try to "ping" all ports exposed on 
  the host and succeed if one of them returns 200.
  Previously, we would assume a specific container port to be "the"
  spacetimedb port, which became false at some point.

Also re-enable `zz_docker` tests.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,7 @@
-version: "3.6"
-
 services:
   node:
+    labels:
+      app: spacetimedb
     build:
       context: ./
       dockerfile: ./crates/standalone/Dockerfile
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -28,8 +28,6 @@ services:
       # Tracy
       - "8086:8086"
     entrypoint: cargo watch -i flamegraphs -i log.conf --why -C crates/standalone -x 'run start --data-dir=/stdb/data --jwt-pub-key-path=/etc/spacetimedb/id_ecdsa.pub --jwt-priv-key-path=/etc/spacetimedb/id_ecdsa'
-    healthcheck:
-      test: curl -f http://localhost/ping || exit 1
```
jdetter, line 32: I'm not sure what else uses our `docker-compose.yml` file so this change *could* potentially break something. I don't think it's a huge concern but why did we remove this health check?

jdetter (approved):
Testing the ping endpoint to see if SpacetimeDB is up rather than just looking at the docxker output seems like it would be a lot more consistent. Thanks Kim, approved with non-blocking question 👍

docker-compose.yml
kim: It was there for historical reasons, and never really worked (we had to do the ping stuff in addition to this).

As written, it's actually broken -- the endpoint should be `http://localhost:3000/v1/ping`.

PR #2978, opened 2025-07-23
ci: Skip the Unity tests if the PR is an external contribution

kim:
We don't want to leak the license information.
Alternatively, we could just make the check non-required.

# Testing

Should test by submitting a PR from a fork.
I'll do that if / when we agree that this is the right approach.


bfops:
This seems reasonable to me. It would be ideal if we could still run these tests on external PRs, but since that doesn't seem to be an option, this is still better than what we do now!

kim:
Mysteriously, the C# check is not running here nor in the test PR.

bfops (approved):
This makes sense to me and seems to have been tested!

PR #2976, opened 2025-07-22
Update quickstart-chat client module bindings from server

alex47:
# Description of Changes

The example quickstart-chat client can not communicate with the example server, because the reducer names have been updated in the server, but not in the module bindings of the client.
I have run the following command in `\sdks\csharp\examples~\quickstart-chat\`:
`spacetime generate --lang csharp --out-dir .\client\module_bindings\ --project-path server`

# API and ABI breaking changes

No changes.

# Expected complexity level and risk

1

# Testing

- [x] With the example server and client send and receive messages

PR #2979, opened 2025-07-23
CI - Check C# quickstart-chat bindings are up to date

bfops:
# Description of Changes

A user pointed out (https://github.com/clockworklabs/SpacetimeDB/pull/2976) that our quickstart-chat bindings were out of date in the C# repo. I'm adding a check to the CI to confirm that they are actually up to date, so that this doesn't happen in the future.

None

1

# Testing

- [x] The new check fails on this PR
- [x] The new check passes if I regenerate bindings (https://github.com/clockworklabs/SpacetimeDB/actions/runs/16508739169/job/46685912939)

### DIFF
--- a/sdks/csharp/examples~/quickstart-chat/client/module_bindings/SpacetimeDBClient.g.cs
+++ b/sdks/csharp/examples~/quickstart-chat/client/module_bindings/SpacetimeDBClient.g.cs
@@ -385,11 +385,11 @@ namespace SpacetimeDB.Types
 
         /// <summary>
         /// Subscribe to the following SQL queries.
-        /// 
+        ///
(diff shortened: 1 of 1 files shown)


rekhoff (approved):
I see the last test did fail for a lack of regenerated bindings, which if that is accurate then the test is working correctly.
The additional check is clear, concise, and should address the problem. Marking as approved.
Small nit-pick (optional), as this pertains specifically to quickstart-chat, we could be more instructive in the error message to "Please generate bindings for quickstart-chat again...", so it's clear specifically which bindings need to be updated without reviewing the whole log.

bfops:
I updated the error message, good call.

Since #2976 is about to merge, I'll wait for that and then merge master into this, so the CI can be green before merging.

Issue #2969, opened 2025-07-21
`ScheduledAt` values arrive as `{tag: 'Time', value: BigInt}` in TypeScript SDK


egormanga:
(or, perhaps, from StDB WS API itself?)

Per the type definitions, the `value` field should instead be a `TimeDuration | Timestamp`.

gefjon:
Is this directly read from the JSON WebSocket API, or is it after processing in the TypeScript client SDK? Is the TypeScript client SDK running against the JSON or BSATN WebSocket API?

cloutiertyler:
@egormanga, I've opened #2980 to address this, but I would also be interested to hear how you found this/what the symptoms are so that I can add a regression test before merging.

jdetter:
Assigning to Jeff because he is on review for this

jdetter:
We can add a regression test for this later because we would like to include this in the upcoming release.

egormanga:
Pardon my absence. Glad it got fixed so fast.

As for how I found it, I simply queried a table with such column, and got this value for it, instead of what the typing describes.

Looking at the PR makes me wonder, how did it even work at all before the fix? Did it just happen to be binary identical on the wire? Or, the other way around, does it work at all with the fix applied?

I thought the proper fix would be in the core/API.

gefjon:
@egormanga The value was correctly encoded on the wire. The TypeScript client SDK was decoding it into a TypeScript object of an unexpected shape.

PR #2980, opened 2025-07-24
Addresses #2969

cloutiertyler:
# Description of Changes

The `ScheduleAt` type appears to have an outdated structure. I've updated the structure of the `ScheduleAt` type represented in TypeScript to be in line with the Rust type: https://docs.rs/spacetimedb/latest/spacetimedb/enum.ScheduleAt.html

Namely, we were missing the inner Spacetime library types of `TimeDuration` and `Timestamp`.

This is to address #2969.

This is an API breaking change in that it changes the API, but it's breaking in that it fixes a bug.

2

# Testing

I have not done additional testing, but I thought I would get this PR started to make it easier for the next person who comes along.

The ideal automated test to add would be one which connects a TypeScript client to a SpacetimeDB module with a scheduled table, and verifies that the type is correctly deserialized and represented in TypeScript when the rows are sent down.


bfops:
This seems fine but I'm not equipped to confirm whether this is correct, so I'm going to remove myself as a reviewer. (And it sounds like this PR is still essentially in progress).

jsdt (approved):
I haven't set up a good way to run integration tests yet, but I manually tested that I the client can call a reducer that uses `ScheduleAt` in an argument, and it can query the rows from a schedule table.

PR #2983, opened 2025-07-24
Store client credentials in a new system table

jsdt:
# Description of Changes

This adds a new system table to store the jwt payloads of connected clients. I'm planning to use this system table to expose client claims to modules in subsequent PRs.

The new table is called `st_connection_credentials`. It is a **private** system table which stores a mapping from `connection_id` to `jwt_payload`. Note that a jwt payload is a json representation of the clients claims, not a fully signed token. 

The times when we need to insert and delete these rows closely mirrors that of the existing `st_client` table, with 1.5 exceptions:
 1. We weren't previously inserting to `st_client` until after the `OnConnect` reducer ran (even though it was in the same transaction). We want `st_connection_credentials` to be populated before calling the reducer, so that the reducer can use it get the credentials, so I made a change to insert to `st_client` and `st_connection_credentials` before calling the reducer.
 2. This difference has not actualized, but when clients start sending refresh tokens, we will probably need to update the credentials stored in this table.

This also enforces uniqueness of connection ids. A duplicate connection id will now make the on-connect reducer fail (since it will violate uniqueness when trying to insert to `st_connection_credentials`).

2.5

Adding a system table is a bit risky. This is almost rollback safe, with one annoying case that is worth calling out:

If a database is created with this system table, opening it with an older version of spacetimedb will only work if there is a snapshot of the database. If we try to load a table without a snapshot, replaying will fail on the first row for that table. This is because we don't write the table schema information to the commit log when creating a database. In practice, this is unlikely to be an issue, because new databases asynchronously trigger a snapshot immediately after creation.

Migrating existing databases will be fine. On startup this will detect that there is a missing system table, and add it in a way that writes it to the commit log. Since it is in the commit log, we can open the database with an older version and still understand the data for that table.

# Testing

There are unit tests that cover opening a database created with an older version (which doesn't have this table).

I manually tested opening a migrated database with an older version of spacetimedb.

### DIFF
--- a/crates/core/src/db/relational_db.rs
+++ b/crates/core/src/db/relational_db.rs
@@ -1643,13 +1643,22 @@ pub async fn local_durability(commitlog_dir: CommitLogDir) -> io::Result<(LocalD
 /// [DurabilityProvider]: crate::host::host_controller::DurabilityProvider
 pub async fn snapshot_watching_commitlog_compressor(
     mut snapshot_rx: watch::Receiver<u64>,
+    mut clog_tx: Option<tokio::sync::mpsc::Sender<u64>>,
+    mut snap_tx: Option<tokio::sync::mpsc::Sender<u64>>,
     durability: LocalDurability,
 ) {
     let mut prev_snapshot_offset = *snapshot_rx.borrow_and_update();
     while snapshot_rx.changed().await.is_ok() {
         let snapshot_offset = *snapshot_rx.borrow_and_update();
         let durability = durability.clone();
-        let res = asyncify(move || {
+
+        if let Some(snap_tx) = &mut snap_tx {
+            if let Err(err) = snap_tx.try_send(snapshot_offset) {
+                tracing::warn!("failed to send offset {snapshot_offset} after snapshot creation: {err}");
+            }
+        }
+
+        let res: io::Result<_> = asyncify(move || {
             let segment_offsets = durability.existing_segment_offsets()?;
             let start_idx = segment_offsets
                 .binary_search(&prev_snapshot_offset)
@@ -1663,15 +1672,27 @@ pub async fn snapshot_watching_commitlog_compressor(
             // in this case, segment_offsets[end_idx] is the segment that contains the snapshot,
             // which we don't want to compress, so an exclusive range is correct.
             let segment_offsets = &segment_offsets[..end_idx];
-            durability.compress_segments(segment_offsets)
+            durability.compress_segments(segment_offsets)?;
+            let n = segment_offsets.len();
+            let last_compressed_segment = if n > 0 { Some(segment_offsets[n - 1]) } else { None };
+            Ok(last_compressed_segment)
         })
         .await;
 
-        if let Err(e) = res {
-            tracing::warn!("failed to compress segments: {e}");
-            continue;
-        }
+        let last_compressed_segment = match res {
+            Ok(opt_offset) => opt_offset,
+            Err(err) => {
+                tracing::warn!("failed to compress segments: {err}");
+                continue;
+            }
+        };
         prev_snapshot_offset = snapshot_offset;
+
+        if let Some((clog_tx, last_compressed_segment)) = clog_tx.as_mut().zip(last_compressed_segment) {
+            if let Err(err) = clog_tx.try_send(last_compressed_segment) {
+                tracing::warn!("failed to send offset {last_compressed_segment} after compression: {err}");
+            }
+        }
     }
 }
 
--- a/crates/standalone/src/lib.rs
+++ b/crates/standalone/src/lib.rs
@@ -127,6 +127,8 @@ impl DurabilityProvider for StandaloneDurabilityProvider {
             |snapshot_rx| {
                 tokio::spawn(relational_db::snapshot_watching_commitlog_compressor(
                     snapshot_rx,
+                    None,
+                    None,
                     durability,
                 ));
             }


crates/sdk/examples/quickstart-chat/main.rs
```diff
@@ -63,22 +63,22 @@ fn creds_store() -> credentials::File {
 /// Our `on_connect` callback: save our credentials to a file.
 fn on_connected(_ctx: &DbConnection, _identity: Identity, token: &str) {
     if let Err(e) = creds_store().save(token) {
-        eprintln!("Failed to save credentials: {:?}", e);
+        eprintln!("Failed to save credentials: {e:?}");
```
gefjon: Not clear to me why these changed, but I also don't care.

crates/client-api/src/auth.rs
```diff
@@ -70,9 +84,18 @@ impl SpacetimeCreds {
 #[derive(Clone)]
 pub struct SpacetimeAuth {
     pub creds: SpacetimeCreds,
-    pub identity: Identity,
-    pub subject: String,
-    pub issuer: String,
+    pub claims: SpacetimeIdentityClaims,
+    // The decoded JWT payload.
```
gefjon:
```suggestion
    /// The decoded JWT payload.
```

Use `///` for doc comments.

crates/client-api/src/auth.rs
```diff
@@ -70,9 +84,18 @@ impl SpacetimeCreds {
...
+    pub raw_payload: String,
```
gefjon: Dissonance between comment saying "decoded" and field name saying "raw" is weird. I assume this is decoded from base64 into a string which will contain a JSON object, but is raw in the sense that the JSON hasn't been parsed.

crates/client-api/src/auth.rs
```diff
@@ -112,7 +135,7 @@ impl TokenClaims {
         &self,
         signer: &impl TokenSigner,
         expiry: Option<Duration>,
-    ) -> Result<String, JwtError> {
+    ) -> Result<(SpacetimeIdentityClaims, String), JwtError> {
```
gefjon, line 154: Slap a comment on here describing the return value? I believe it is the tuple `(claims, token)`, where `claims` is the contents of `token`.

crates/client-api/src/auth.rs
```diff
@@ -123,10 +146,11 @@ impl TokenClaims {
             iat,
             exp,
         };
-        signer.sign(&claims)
+        let token = signer.sign(&claims)?;
+        Ok((claims, token))
     }
 
-    pub fn encode_and_sign(&self, signer: &impl TokenSigner) -> Result<String, JwtError> {
+    pub fn encode_and_sign(&self, signer: &impl TokenSigner) -> Result<(SpacetimeIdentityClaims, String), JwtError> {
```
gefjon, line 172: Ditto comment re: return value.

crates/client-api/src/auth.rs
gefjon: It's also not obvious to me why this is useful. Can't the caller just clone out of `self` themselves?

crates/datastore/src/system_tables.rs
```diff
@@ -927,6 +960,18 @@ pub struct StClientRow {
...
 
+/// System table [ST_CONNECTION_CREDENTIALS_NAME]
+///
+/// | connection_id                      | jwt_payload                                             |
+/// |------------------------------------|---------------------------------------------------------|
+/// | 0x6bdea3ab517f5857dc9b1b5fe99e1b14 | '{"iss":"issuer","sub":"user-id","iat":1629212345,...}' |
+#[derive(Clone, Debug, Eq, PartialEq, SpacetimeType)]
+#[sats(crate = spacetimedb_lib)]
+pub struct StConnectionCredentialsRow {
+    pub connection_id: ConnectionIdViaU128,
```
gefjon, line 1088: We don't otherwise enforce uniqueness of connection IDs, only  of `(ConnectionId, Identity)` pairs.

crates/client-api/src/auth.rs
jsdt: `self` only has a subset of the claims, since we are adding claims for expiry and issued at.

crates/core/src/host/module_host.rs
```diff
@@ -728,38 +754,19 @@ impl ModuleHost {
...
-                .inspect_err(|e| {
-                    log::error!(
-                        "`call_identity_connected`: fallback transaction to insert into `st_client` failed: {e:#?}"
-                    )
-                })
-                .map_err(Into::into)
+
+                // TODO: Is this being broadcast? Does it need to be, or are st_client table subscriptions
+                // not allowed?
+                // I don't think it was being broadcast previously.
```
gefjon:
```suggestion
                // I (jsdt 2025-08-06) don't think it was being broadcast previously.
```

This sounds like the kind of bug you were talking about the other day, yeah. `st_client` subscriptions are not disallowed, but also will not work in any of the SDKs due to missing bindings, and aren't an advertised feature. Maybe we should just make some or all of the system tables private and not have to think about it any more. Or maybe we should fix the bug and broadcast this TX.

gefjon: Created #3130 .

gefjon (commented):
This all looks reasonable at this stage.

crates/core/src/host/module_host.rs
jsdt: I updated the comment to link to that issue.

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -968,13 +975,16 @@ impl MutTxId {
...
+        /*
         if constraint.constraint_id != ConstraintId::SENTINEL {
             return Err(anyhow::anyhow!(
                 "`constraint_id` must be `ConstraintId::SENTINEL` in `{:#?}`",
                 constraint
             )
             .into());
         }
+
+         */
```
gefjon: I'd prefer not  to have commented-out code committed.

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -503,9 +508,11 @@ impl MutTxId {
     /// - The index metadata is inserted into the system tables (and other data structures reflecting them).
     /// - The returned ID is unique and is not `IndexId::SENTINEL`.
     pub fn create_index(&mut self, mut index_schema: IndexSchema, is_unique: bool) -> Result<IndexId> {
+        /*
         if index_schema.index_id != IndexId::SENTINEL {
             return Err(anyhow::anyhow!("`index_id` must be `IndexId::SENTINEL` in `{:#?}`", index_schema).into());
         }
+         */
```
gefjon: I'd prefer not to have commented-out code committed.

crates/datastore/src/system_tables.rs
```diff
@@ -340,6 +354,19 @@ fn system_module_def() -> ModuleDef {
         .with_type(TableType::System);
     // TODO: add empty unique constraint here, once we've implemented those.
 
+    let st_connection_credentials_type = builder.add_type::<StConnectionCredentialsRow>();
+    // let st_connection_credentials_unique_cols = [StConnectionCredentialsFields::ConnectionId];
```
gefjon: I'd prefer not to have commented-out code committed.

PR #2984, opened 2025-07-24
Docs CI - Check that `nav.js` matches the `.md` files list

bfops:
# Description of Changes

I added a script that checks that the files listed in `nav.js` match all the `.md` files we can find in the `docs` directory.

None

2

# Testing

- [x] The new check passes on this PR
- [x] The new check fails if a new `.md` file is added (https://github.com/clockworklabs/SpacetimeDB/actions/runs/16503970552/job/46669677358)
- [x] The new check fails if one of our `.md` files is removed (https://github.com/clockworklabs/SpacetimeDB/actions/runs/16503983228/job/46669719349)


.github/workflows/docs-validate-nav-build.yml
```diff
@@ -39,9 +46,3 @@ jobs:
         working-directory: docs
         run: |
           diff -q docs/nav.js docs/nav.js.original || (echo "Generated nav.js differs from committed version. Run 'npm run build' and commit the updated file." && exit 1)
-
-      - name: Restore original nav.js
-        working-directory: docs
-        if: success() || failure()
-        run: |
-          mv docs/nav.js.original docs/nav.js
```
bfops: I removed this step because it would fail if the "Check that all files are listed in nav.js" step failed (because we'd never get to the step that creates `nav.js.original`), and more importantly this step seems completely unnecessary? The checkout is ephemeral.

jdetter (approved):
Ok here we're just verifying that everything listed in the navigation bar at least points to a valid `.md` file which seems like a good sanity check. Thanks Zeke!

PR #2985, opened 2025-07-24
Define a `Deserializer` from JS values

Centril:
# Description of Changes

Defines `Deserializer`, which can deserialize a `v8::Value` to SATS values.

None

Doesn't interact with existing code.

# Testing

Roundtrip tests will be added in the next PR which will define `Serializer`.

### DIFF
--- a/crates/core/src/host/v8/de.rs
+++ b/crates/core/src/host/v8/de.rs
@@ -114,13 +114,22 @@ fn v8_interned_string<'s>(scope: &mut HandleScope<'s>, field: &str) -> Local<'s,
 /// Extracts a reference `&'s T` from an owned V8 [`Local<'s, T>`].
 ///
 /// The lifetime `'s` is that of the [`HandleScope<'s>`].
-/// This ensures that the reference to `T` won't outlive the `HandleScope`.
+/// This ensures that the reference to `T` won't outlive the `HandleScope`,
+/// that owns the `x: T` and keeps it alive.
 fn deref_local<'s, T>(local: Local<'s, T>) -> &'s T {
     let reference = local.borrow();
     // SAFETY: Lifetime extend `'0` to `'s`.
     // This is safe as the returned reference `&'s T`
-    // will not outlive its `HandleScope<'s, _>`,
+    // will not outlive its `HandleScope<'s, _>`
     // as both are tied to the lifetime `'s`.
+    // The scope is what owns the `x: T`.
+    //
+    // See https://docs.rs/v8/latest/v8/struct.Local.html for more details.
+    // > It is safe to extract the object stored in the handle by
+    // > dereferencing the handle (for instance, to extract the `*Object` from
+    // > a `Local<Object>`); the value will still be governed by a handle
+    // > behind the scenes and the same rules apply to these values as to
+    // > their handles.
     unsafe { core::mem::transmute::<&T, &'s T>(reference) }
 }
 
@@ -264,7 +273,7 @@ struct ProductAccess<'a, 's> {
     object: Local<'s, Object>,
     /// A field's value, to deserialize next in [`NamedProductAccess::get_field_value_seed`].
     next_value: Option<Local<'s, Value>>,
-    /// The index in the product to
+    /// The field index to deserialize next in the product.
     index: usize,
 }
 


crates/bindings-macro/src/sats.rs
```diff
@@ -454,10 +454,10 @@ pub(crate) fn derive_deserialize(ty: &SatsType<'_>) -> TokenStream {
                             }
                         }
 
-                        fn visit_seq<__E: #spacetimedb_lib::de::Error>(self, index: usize, name: &str) -> Result<Self::Output, __E> {
+                        fn visit_seq(self, index: usize) -> Self::Output {
                             match index {
-                                #(#iter_n4 => Ok(__ProductFieldIdent::#field_names),)*
-                                _ => Err(#spacetimedb_lib::de::Error::unknown_field_name(name, &self)),
```
cloutiertyler, line 460: What's the context of these changes?

crates/sats/src/de/impls.rs
```diff
@@ -659,12 +659,13 @@ impl FieldNameVisitor<'_> for TupleNameVisitor<'_> {
             .ok_or_else(|| Error::unknown_field_name(name, &self))
     }
 
-    fn visit_seq<E: Error>(self, index: usize, name: &str) -> Result<Self::Output, E> {
+    fn visit_seq(self, index: usize) -> Self::Output {
+        // Confirm that the index exists.
         self.elems
             .get(index)
-            .ok_or_else(|| Error::unknown_field_name(name, &self))?;
+            .expect("`index` should exist when `visit_seq` is called");
```
cloutiertyler, line 666: Why a panic here?

crates/core/src/host/v8/de.rs
```diff
@@ -0,0 +1,421 @@
...
+        self,
+        visitor: V,
+        seed: T,
+    ) -> Result<V::Output, Self::Error> {
+        let arr = cast!(self.common.scope, self.input, v8::Array, "`Array`")?;
+        visitor.visit(ArrayAccess::new(arr, self.common, seed))
+    }
+}
+
+struct ProductAccess<'a, 's> {
```
cloutiertyler, line 270: Could we get a comment for this type?

crates/core/src/host/v8/de.rs
```diff
@@ -0,0 +1,421 @@
...
+    object: Local<'s, Object>,
+    /// A field's value, to deserialize next in [`NamedProductAccess::get_field_value_seed`].
+    next_value: Option<Local<'s, Value>>,
+    /// The index in the product to
+    index: usize,
+}
+
+/// Map from integer keys to their `str` representation,
+/// for small numbers up to 12.
+static INT_TO_STR: &[&str] = &["0", "1", "2", "3", "4", "5", "6", "7", "8", "9", "10", "11", "12"];
```
cloutiertyler: NIT: Feels like premature optimization

crates/core/src/host/v8/de.rs
```diff
@@ -0,0 +1,421 @@
...
+        let input = self
+            .next_value
+            .take()
+            .expect("Call next_key_seed before next_value_seed");
+        // Deserialize the field's value.
+        seed.deserialize(Deserializer { common, input })
+    }
+}
+
+struct SumAccess<'a, 's> {
```
cloutiertyler, line 337: Comment here please

crates/sats/src/de/impls.rs
Centril: See comment on definition of `fn visit_seq` in trait. This method is only used when the input is known to exist to enable translating e.g., indices to enum variants and whatnot.

crates/bindings-macro/src/sats.rs
Centril: The `Err` case was always dead code, but now this had to change due to `name` being unavailable where `.visit_seq(...)` is called due to borrow checker issues.

crates/core/src/host/v8/de.rs
```diff
@@ -0,0 +1,420 @@
...
+
+/// Things shared between various [`Deserializer`]s.
+///
+/// The lifetime `'s` is that of the scope of values deserialized.
+struct DeserializerCommon<'a, 's> {
+    /// The scope of values to deserialize.
+    scope: &'a mut HandleScope<'s>,
+    /// A cache for frequently used strings.
+    key_cache: &'a mut KeyCache,
+}
```
gefjon, line 39: What's the lifetime `'a`? And can we just use words to name these lifetimes, like `'scope` instead of `'s`? Fine if `'a` doesn't have a meaningful name, but a comment describing the meaning of a single-letter variable is a little :thinking: .

crates/core/src/host/v8/de.rs
```diff
@@ -0,0 +1,420 @@
...
+/// Extracts a reference `&'s T` from an owned V8 [`Local<'s, T>`].
+///
+/// The lifetime `'s` is that of the [`HandleScope<'s>`].
+/// This ensures that the reference to `T` won't outlive the `HandleScope`.
+fn deref_local<'s, T>(local: Local<'s, T>) -> &'s T {
+    let reference = local.borrow();
+    // SAFETY: Lifetime extend `'0` to `'s`.
+    // This is safe as the returned reference `&'s T`
+    // will not outlive its `HandleScope<'s, _>`,
+    // as both are tied to the lifetime `'s`.
```
gefjon: This comment on its own is not sufficient to convince me. I was unsure whether the `Local` value itself was keeping its referent alive in the V8 GC. It turns out that the enclosing `HandleScope` does that, and the `Local` is just a normal reference within that scope. I had to go check the docs for `Local`, where I found the following paragraph:

> It is safe to extract the object stored in the handle by dereferencing the handle (for instance, to extract the *Object from a Local<Object>); the value will still be governed by a handle behind the scenes and the same rules apply to these values as to their handles.

I recommend linking to `Local` and also either quoting or paraphrasing that point.

crates/core/src/host/v8/de.rs
```diff
@@ -0,0 +1,420 @@
...
+    }
+
+    fn deserialize_sum<V: SumVisitor<'de>>(self, visitor: V) -> Result<V::Output, Self::Error> {
+        let scope = &mut *self.common.scope;
+        let sum_name = visitor.sum_name().unwrap_or("<unknown>");
+
+        // We expect a canonical representation of a sum value in JS to be
+        // `{ tag: "foo", value: a_value_for_foo }`
+        // with special convenience for optionals
+        // where we also accept `null`, `undefined` and an object without `tag`.
```
gefjon, line 189: It feels mildly footgun-ey that putting a `tag` within an object breaks using the convenience representation, and therefore that `Option<enum { Foo, Bar }>` must be represented as `{ "tag": "Some", "value": { "tag": "Foo" } }`. I think we probably should require that optionals be `null | undefined | bare_value`, never the "regular sum" representation.

crates/core/src/host/v8/de.rs
```diff
@@ -0,0 +1,420 @@
...
+
+/// Provides access to the field names and values in a JS object
+/// under the assumption that it's a product.
+struct ProductAccess<'a, 's> {
+    common: DeserializerCommon<'a, 's>,
+    /// The input object being deserialized.
+    object: Local<'s, Object>,
+    /// A field's value, to deserialize next in [`NamedProductAccess::get_field_value_seed`].
+    next_value: Option<Local<'s, Value>>,
+    /// The index in the product to
```
gefjon: Cut-off comment

crates/core/src/host/v8/de.rs
Centril: I actually felt that that particular comment wasn't particularly clear, but I'll clarify the comment and add a reference to the above.

crates/core/src/host/v8/de.rs
Centril: We agreed to follow this PR up with one renaming these lifetimes.

crates/core/src/host/v8/de.rs
Centril: > I think we probably should require that optionals be `null | undefined | bare_value`, never the "regular sum" representation.

If we want to remove the special case, I'd rather opt for always using the "regular sum" repr, as otherwise, you cannot represent `Some(None)`, which is possible in both Rust and C# modules. This will make things slightly less convenient for the module author, but we can add some convenience functions on the TS bindings side.

(See also https://github.com/clockworklabs/SpacetimeDB/pull/2996#discussion_r2243860375.)

PR #2988, opened 2025-07-25
Generate Websocket token from auth in Unity WebGL build

FME849:
# Description of Changes

This is a fix for community bug which auth token is not used to create Websocket connection in Unity WebGL build. 
The issue here: https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/issues/352

None

1

# Testing

- [ ] Using OAuth2.0 tokens, you should get the same `Identity` even when the token is refreshed in the WebGL build

### DIFF
--- a/sdks/csharp/src/Plugins/WebSocket.jslib
+++ b/sdks/csharp/src/Plugins/WebSocket.jslib
@@ -25,7 +25,7 @@ mergeInto(LibraryManager.library, {
             var uri = UTF8ToString(uriPtr);
             var protocol = UTF8ToString(protocolPtr);
             var authToken = UTF8ToString(authTokenPtr);
-                        if (authToken)
+            if (authToken)
             {
                 var tokenUrl = new URL('v1/identity/websocket-token', host);
                 tokenUrl.protocol = host.startsWith("wss://") ? 'https:' : 'http:';


FME849:
Hi @jdetter, I opened the PR to fix the auth token in WebGL build

sdks/csharp/src/Plugins/WebSocket.jslib
```diff
@@ -63,10 +84,10 @@ mergeInto(LibraryManager.library, {
                 }
             };
 
-            return socketId;
+            dynCall('vi', callbackPtr, [socketId]);
```
jdetter, line 87: Why did we change this to a `dynCall` here instead of just returning `socketId` as before? Is there a benefit to doing this?

jdetter (changes requested):
One question on this, I'm testing this now

jdetter:
> Hi @jdetter, I opened the PR to fix the auth token in WebGL build

Thank you for transferring the PR! 👍

sdks/csharp/src/Plugins/WebSocket.jslib
FME849: As we use async/await to fetch the temporary token, the return result of `WebSocket_Connect` become `Promise` which C# do not understand. Therefore, we use a C# callback function pointer `(IntPtr)` which Javascript can invoke after complete task to pass the `socketId` back to C#.

jdetter: Thanks for the context 👍

jdetter (approved):
Code looks good to me, functionality tested via Blackholio and it works just fine. I'm able to resume the same play session using the same token when building for WebGL. Thanks for the PR @FME849! 🙏

PR #2996, opened 2025-07-28
Define a `Serializer` to JS values

Centril:
# Description of Changes

Defines `Serializer`, which can serialize to a `v8::Value` from a SATS value.

None

Doesn't interact with existing code.

# Testing

Roundtrip-tests are added, both for some unit cases and for any random AT/AV pair (proptest).

### DIFF
--- a/crates/core/src/host/v8/de.rs
+++ b/crates/core/src/host/v8/de.rs
@@ -274,7 +274,7 @@ pub(super) fn intern_field_name<'s>(scope: &mut HandleScope<'s>, field: Option<&
         Some(field) => Cow::Borrowed(field),
         None => Cow::Owned(format!("{index}")),
     };
-    v8_interned_string(scope, field).into()
+    v8_interned_string(scope, &field).into()
 }
 
 impl<'de, 's: 'de> de::NamedProductAccess<'de> for ProductAccess<'_, 's> {


crates/core/src/host/v8/ser.rs
```diff
@@ -0,0 +1,312 @@
...
+            self.key_cache.tag(self.scope).into(),
+            self.key_cache.value(self.scope).into(),
+        ];
+
+        // Stitch together the object.
+        let prototype = v8::null(self.scope).into();
+        let object = Object::with_prototype_and_properties(self.scope, prototype, &names, &values);
+        seal_object(self.scope, &object)?;
+        Ok(object.into())
+    }
```
gefjon, line 156: My gut says this should have special handling for options, the same way that the deserialize path does.

Centril: I considered that, but a problem here is that we actually do not have enough information to tell that the type is `option` here because you don't know about the other variant. For example, for `.serialize_variant(0, Some("none"), payload)`, it could be option, but you could also have 3 variants instead of 2. Threading this information through SATS is currently not trivial. We might want to consider e.g., adding `AlgebraicTypeLayout::Optional(_)`, which could actually simplify things down the road. But for now, I'd rather remove the special case from the deserialization end instead.

(See also https://github.com/clockworklabs/SpacetimeDB/pull/2985#discussion_r2243852082.)

PR #2998, opened 2025-07-29
Add additional logging to subscribe route and simplify calling `client_connected`

gefjon:
# Description of Changes

Out-of-band discussions with the BitCraft team brought up questions about whether it was possible for a rejected client connection to start an expensive computation like a subscription before their connection was killed, e.g. by sending a `Subscribe` message along the WebSocket before `client_connected` had finished returning `Err`.

I don't believe this was actually possible, as `ClientConnection::spawn` called and awaited `call_identity_connected` before invoking its `actor` closure, and it was that `actor` which processed `Subscribe` messages. But it was somewhat difficult to verify that behavior, and so I re-organized the code so that the outer layer of the `subscribe` handler obviously had that property without having to step into `ClientConnection::spawn`.

I also added some additional logging to the subscribe route, including the `X-Forwarded-For` header in more messages, as the BitCraft team complained about having difficulty correlating IP addresses with connections. The log levels remain the same as before, just with additional information added:

- Successful connections are at `debug` level,
- Rejected connections are at `info` level (these are the ones BitCraft cares about in this case).
- Failed connections are at `warn`.

As the levels are unchanged, this should not add undesirable log noise.

N/a

3? The `subscribe` route was complex and remains so. I believe this change simplifies the code and makes the logic more obvious, but reviewers should take care to verify that the behavior actually is equivalent as I believe.

# Testing

- [ ] I would like a test deployment of BitCraft to staging or something, or to include this patch in the next bot test that we do anyways, just for sanity's sake.

### DIFF
--- a/crates/client-api/src/routes/subscribe.rs
+++ b/crates/client-api/src/routes/subscribe.rs
@@ -145,7 +145,7 @@ where
 
     let identity_token = auth.creds.token().into();
 
-    let module_rx = leader.module_watcher().await.map_err(log_and_500)?;
+    let mut module_rx = leader.module_watcher().await.map_err(log_and_500)?;
 
     let client_id = ClientActorId {
         identity: auth.identity,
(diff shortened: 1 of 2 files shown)


crates/core/src/client/client_connection.rs
```diff
@@ -393,14 +393,31 @@ const CLIENT_CHANNEL_CAPACITY: usize = 16 * KB;
 const KB: usize = 1024;
 
 impl ClientConnection {
-    /// Returns an error if ModuleHost closed
+    /// Call the database at `module_rx`'s `client_connection` reducer, if any,
+    /// and return `Err` if it signals rejecting this client's connection.
+    ///
+    /// Call this method before [`Self::spawn`],
+    /// and do not call [`Self::spawn`] if this method returns `Err`.
+    pub async fn call_client_connected_maybe_reject(
```
kim, line 415: Instead of relying solely on people reading the docstrings, this could also return a `struct Connected(())` that can only be constructed in this module and needs to be passed to `spawn`.

crates/client-api/src/routes/subscribe.rs
```diff
@@ -168,26 +168,36 @@ where
...
+        };
 
-        let actor = |client, sendrx| ws_client_actor(ws_opts, client, ws, sendrx);
-        let client = match ClientConnection::spawn(client_id, client_config, leader.replica_id, module_rx, actor).await
-        {
-            Ok(s) => s,
+        log::debug!("New client connected from {client_log_string}");
+
+        match ClientConnection::call_client_connected_maybe_reject(&mut module_rx, client_id).await {
+            Ok(()) => log::info!("client_connected returned Ok for {client_log_string}"),
```
kim: Do we need this to be at `info!`?

crates/client-api/src/routes/subscribe.rs
```diff
@@ -168,26 +168,36 @@ where
...
             Err(e @ (ClientConnectedError::Rejected(_) | ClientConnectedError::OutOfEnergy)) => {
-                log::info!("{e}");
+                log::info!(
+                    "Rejecting connection for {client_log_string} due to error from client_connected reducer: {e}"
```
kim: Our logs are quite hard to filter in ES, which partly is due to terminal colors being ingested. For example, I can't meaningfully filter on the module file name.

Since there can be quite a lot of websocket / subscribe related logs (especially when something is wrong), I'd appreciate if we could use a common prefix in log messages (like "websocket:" or "subscribe:" or whatever).

kim (approved):
Looks reasonable to me. fwiw, there is a smoketest that tests rejection via the `client_connected` reducer.

crates/client-api/src/routes/subscribe.rs
gefjon: Nope! I meant to put this at `debug!`. Fixed.

crates/client-api/src/routes/subscribe.rs
gefjon: Added `websocket: ` to all the logs touched by this PR.

crates/core/src/client/client_connection.rs
gefjon: Smart. Added.

PR #3002, opened 2025-07-29
Update all licenses

bfops:
# Description of Changes

We recently merged several repos together. This PR clarifies the license terms for several subdirectories, as well as the relationship between the licenses.

The licenses in our subdirectories have become symbolic links to licenses in our toplevel `licenses` directory. For any particular subdirectory's license file in the diff, you can click `... -> View file` and then click on the text that says "Symbolic Link" on that page. This will take you to the license file that it links to.

I have also updated the `tools/upgrade-version` script to update the change date in the new `licenses/BSL.txt` file.

None.

1

# Testing

None. Only changes to license files.


cloutiertyler (approved):
LGTM, but I left some comments!

PR #3004, opened 2025-07-30
Dont generate connection_ids in the rust client.

jsdt:
# Description of Changes

This changes the behavior on the rust client, so that we let the server generate the connection id if the client hasn't called the unstable method to set a connection id.

This is awkward for the `connection_id` function, since it now panics if the connection id hasn't been received from the server. We should deprecate this function in favor of a `try` version.

This also changes the behavior of reusing the connection id if a client reconnects.

The corresponding private PR is https://github.com/clockworklabs/SpacetimeDBPrivate/pull/1921.

Technically not changing any API signatures, but this is behavior-changing, since the `connection_id` function can now panic.

2. The risk here is people relying on that behavior.

# Testing

I think some tests need to be updated.


gefjon:
I don't think we need to deprecate the panicking `connection_id` function, but we should add the `try_connection_id` variant and document the difference. Note that we have both `identity` and `try_identity`.

crates/sdk/src/websocket.rs
```diff
@@ -228,7 +231,7 @@ impl WsConnection {
         })?;
         Ok(WsConnection {
             db_name: db_name.into(),
-            connection_id,
+            // connection_id,
```
gefjon:
```suggestion
```

crates/sdk/src/websocket.rs
```diff
@@ -86,7 +86,7 @@ pub enum WsError {
 
 pub(crate) struct WsConnection {
     db_name: Box<str>,
-    connection_id: ConnectionId,
+    // connection_id: ConnectionId,
```
gefjon:
```suggestion
```

gefjon (commented):
Codegen appears not to have changed other than the commit hash, so I'd suggest we might as well remove those changes to shrink the diff here.

crates/sdk/src/db_connection.rs
```diff
@@ -83,6 +83,11 @@ pub struct DbContextImpl<M: SpacetimeModule> {
     /// May be `None` if we connected anonymously
     /// and have not yet received the [`ws::IdentityToken`] message.
     identity: SharedCell<Option<Identity>>,
+
+    /// This connection's `Identity`.
```
gefjon:
```suggestion
    /// This connection's `ConnectionId`.
```

crates/sdk/src/db_connection.rs
```diff
@@ -130,7 +136,18 @@ impl<M: SpacetimeModule> DbContextImpl<M> {
...
+                    // Callbacks can (will) call [`DbContext::connection_id`], which acquires that lock,
+                    // so holding it while running a callback causes deadlocks.
+                    let mut conn_id_store = self.connection_id.lock().unwrap();
+                    // This would only happen if the client is using the unstable `set_connection_id` method.
+                    if let Some(prev_conn_id) = *conn_id_store {
+                        assert_eq!(prev_conn_id, conn_id);
+                    }
+                    *conn_id_store = Some(conn_id);
+                }
+                // assert_eq!(get_connection_id(), conn_id);
```
gefjon: Should this be uncommented or removed?

gefjon (approved):

PR #3008, opened 2025-07-30
Make Timestamp, TimeDuration,Vec<u8> implement FilterableValue

### LINKED ISSUE #2650: Make `Timestamp` implement `FilterableValue`
In both Rust and C#. This will require modifications to the C#, Rust and TypeScript client SDKs to ensure their client-side index implementations are compatible with `Timestamp`.

Might as well do `TimeDuration` too.

mamcx:
# Description of Changes

As the title says.

Closes #2650

1
# Testing

**NOTE** : Need input in what other testing to add

- [x] Add testing for `Comparable` in the `C#` side
- [x] Also that it compiles in module


sdks/typescript/packages/sdk/src/compare.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * Compares the current instance with another object of the same type.
+ */
+interface Comparable<T> {
```
jsdt: Is this used?

crates/lib/src/filterable_value.rs
```diff
@@ -1,6 +1,8 @@
 use crate::{ConnectionId, Identity};
 use core::ops;
 use spacetimedb_sats::bsatn;
+use spacetimedb_sats::time_duration::TimeDuration;
+use spacetimedb_sats::timestamp::Timestamp;
 use spacetimedb_sats::{hash::Hash, i256, u256, Serialize};
 
 /// Types which can appear as an argument to an index filtering operation
```
rekhoff, line 8: We should probably add Timestamp and TimeDuration to the of types specifically called out as having corresponding `FilterableValue` implementations in this comment block.

modules/module-test-cs/Lib.cs
```diff
@@ -171,6 +171,18 @@ public partial struct Player
     public string name;
 }
 
+// Extra table for checking implementation of `FilterableValue` trait
```
rekhoff, line 174: If we are adding a table to specifically check for the `FilterableValue` trait, we should add the other filterable types as well.

sdks/typescript/packages/sdk/src/compare.ts
mamcx: > What are the implications of being a filterable value?

It means a `Filter` functions is generated for this field, but you can filter with `sql` any field.

mamcx: > Is this used?

Is removed now, was before agreement to not mirror yet the filterable trait

corrm:
@bfops Updates?

cloutiertyler:
This will need to be reimplemented since there's been so much movement. I will ask AI to see how hard it would be to update this PR, otherwise I will close and create a ticket to reimplement.

bfops:
This one is already linked to an issue, and there is an agent-written PR here: https://github.com/clockworklabs/SpacetimeDB/pull/4813.

PR #3010, opened 2025-07-31
smoktests: Wait for batched consensus

kim:
Until we have confirmed reads, we need to wait for a tick until any batched commits are flushed.

Also cleaned up the code a bit to do less string manipulation of sql results all over the place, and to output less irrelevant logging.

1

# Testing

- [x] yes

### DIFF
--- a/smoketests/tests/replication.py
+++ b/smoketests/tests/replication.py
@@ -120,9 +120,9 @@ where replication_state.database_id={database_id} \
         rows = self.sql(f"select id from counter where id={id}")
         if len(rows) < 1 or int(rows[0]['id']) != id:
             raise ValueError(f"Could not find {id} in counter table")
-        # Wait a tick to ensure buffers are flushed.
+        # Wait for at least one tick to ensure buffers are flushed.
         # TODO: Replace with confirmed read.
-        time.sleep(0.3)
+        time.sleep(0.6)
 
 
     def fail_leader(self, action='kill'):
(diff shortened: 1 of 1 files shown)


smoketests/tests/replication.py
```diff
@@ -114,15 +113,16 @@ def wait_for_leader_change(self, previous_leader_node, max_attempts=10, delay=2)
...
 
         retry(lambda: self.test.call("start", id, 1))
-        add_table = str(self.test.sql(f"SELECT id FROM counter where id={id}"))
-        if str(id) not in add_table:
+        rows = self.sql(f"select id from counter where id={id}")
+        if len(rows) < 1 or int(rows[0]['id']) != id:
             raise ValueError(f"Could not find {id} in counter table")
+        # Wait a tick to ensure buffers are flushed.
+        # TODO: Replace with confirmed read.
+        time.sleep(0.3)
```
kim: Main change.

smoketests/tests/replication.py
```diff
@@ -247,31 +247,42 @@ def start(self, id: int, count: int):
...
             self.cluster.ensure_leader_health(second_id)
             # restart the old leader, so that we can maintain quorum for next iteration
+            print(f"reconnect leader {container_id}")
             self.cluster.restore_leader(container_id, 'start')
 
-        # verify if all past rows are present in new leader
-        for row_id in row_ids:
-            table = self.sql(f"SELECT * FROM counter WHERE id = {row_id}")
-            self.assertIn(f"{row_id}", str(table))
+        # Ensure we have a current leader
```
kim, line 278: Not sure exactly why, but test becomes flaky if I omit this (same for `test_leader_c_disconnect_in_loop`).

Shubham8287 (approved):

PR #3102, opened 2025-08-04
CI - Remove `bots please`

bfops:
# Description of Changes

We haven't used this in a while and we're pretty sure it's currently broken.

None.

1

# Testing

None


jdetter:
bots please

Issue #2962, opened 2025-07-21
pre-publish API endpoint


Shubham8287:
Design and implement pre-publish endpoint, High level idea is mentioned in https://github.com/clockworklabs/SpacetimeDBPrivate/pull/1849.

Shubham8287:
### Design:

POST: `v1/database/pre_publish/:name_or_identity`

#### Input:
Same as current publish api

#### Output:
```json
{
  "migration_plan": "migration plan prettified",
  "token": "abcd1234hmac...", // HMAC_SHA256(secret_key, database_id + old_module_hash + module_hash).
  "break_clients": true,
}
```
We don't need to manage an external secret — the token is a UX safeguard to confirm user intent before publishing a breaking schema.
Some server identity or a random number should work, as long as it's unique to a SpacetimeDB instance.


Also, amend publish endpoint to optionally accept this token and let breaking module to publish after verifying provided token.

Shubham8287:
@gefjon Or @Centril , if you have any counter thoughts.

gefjon:
If I'm understanding correctly, this looks reasonable. The flow will be:

1. Client sends first request, `POST pre_publish/my-database`.
  - Server computes the migration plan based on the database and new module, but doesn't perform any changes or save any state.
  - Server computes the hash you described.
  - Server throws away the WASM blob from this `POST` request.
2. Client sends second request, `POST publish/my-database`, with the same WASM blob and also the hash from 1.
  - Server re-computes the hash you described using the WASM blob attached to this `POST` request.
  - If the hashes match, the server goes ahead and does the usual publish thing.

Does step 2 require re-generating the auto-migration plan anyways? If so, I think the hash should literally be `database_id + old_module_hash + module_hash + migration-plan`, without any "secret." But if the secret is necessary, or if `POST publish` doesn't need to re-generate the auto-migrate plan otherwise, then I'm good with what you suggest.

> Also, amend publish endpoint to optionally accept this token and let breaking module to publish after verifying provided token.

This being optional is important, good call. An outdated SpacetimeDB CLI should still be able to `spacetime publish` updates that don't require explicit approval (I'm calling these "supervised migrations"), even though it'll never hit `POST pre_publish`.

Shubham8287:
> I think the hash should literally be database_id + old_module_hash + module_hash + migration-plan.

[...]

PR #3121, opened 2025-08-04
Automigration pretty print

Shubham8287:
# Description of Changes
Pretty print for Auto migration.

NA

1

# Testing
I am not sure, how to write meaningful unit tests for this.
I have written smoketests to cover pretty-print as part of `pre-publish` command PR (yet to raise).


crates/schema/src/auto_migrate/ansi_formatter.rs
```diff
@@ -0,0 +1,359 @@
...
+                    old_type,
+                    new_type,
+                } => {
+                    self.add_line(format!(
+                        "~ Modified: {} ({} → {})",
+                        name,
+                        self.format_type_name(old_type),
+                        self.format_type_name(new_type)
+                    ));
+                }
```
gefjon: Do we support renaming or changing the type of columns in automigrations? That's surprising to me.

Shubham8287: I have not digged into it but most probably, Its about adding variants to Enum.

gefjon: Ah, that makes sense, at least for `TypeChanged`. I am less sure about `Renamed`, but :shrug: .

Shubham8287:
Makes sense to add snapshot tests. I will do that.

> This test probably has to be marked #[serial] via the [serial_test crate](https://docs.rs/serial_test/latest/serial_test/) and run in a test target (i.e. file under crates/foo/tests/) that only has explicitly marked tests due to [set_var being unsafe in multithreaded contexts](https://doc.rust-lang.org/std/env/fn.set_var.html).

umm, I think using `set_var` is only useful if we are writting this test from `spacetimedb-cli` interface. I imagine flow to be like:
1. `spacetime-cli` checks for `NO_COLOR` env variable.
2. Accordingly call `pre_publish` api with param to use `PlainFormatter` instead of `AnsiFormatter`.

So test for disabling ANSI escapes based on the passed argument should be enough?

gefjon:
> umm, I think using `set_var` is only useful if we are writting this test from `spacetimedb-cli` interface. I imagine flow to be like:
> 
>     1. `spacetime-cli` checks for `NO_COLOR` env variable.
> 
>     2. Accordingly call `pre_publish` api with param to use `PlainFormatter` instead of `AnsiFormatter`.
> 
> 
> So test for disabling ANSI escapes based on the passed argument should be enough?

Sure, if you set it up that way, that'd work fine. It would be unsafe if you relied on the `colored` crate to check `NO_COLOR`, and used `set_env` within the test to do the assertion. Does `PlainFormatter` already exist, or are you planning to add it separately?

Shubham8287:
> Does `PlainFormatter` already exist, or are you planning to add it separately?

I stole one from James draft PR - https://github.com/clockworklabs/SpacetimeDB/pull/2065/files#diff-8150155bc61f819c6f1e3b414e728839462b5421657e2393304e4d050fd69cbdR18.

Regex wrapper over ANSI formmater. It gets the job done.

crates/schema/src/auto_migrate.rs
```diff
@@ -44,16 +45,24 @@ impl<'def> MigratePlan<'def> {
         }
     }
 
-    pub fn pretty_print(&self) -> anyhow::Result<String> {
+    pub fn pretty_print(&self, no_color: bool) -> anyhow::Result<String> {
```
gefjon:
```suggestion
    #[derive(Copy, Clone, PartialEq, Eq)]
    pub enum PrettyPrintStyle {
        AnsiColor,
        NoColor,
    }
    pub fn pretty_print(&self, style: PrettyPrintStyle) -> anyhow::Result<String> {
```

Boolean flag arguments like this are difficult to read and interpret. Please use an enum with named variants instead.

Shubham8287:
>  `set_override`, perhaps obviously, sets a global flag, and so setting it and then never unsetting it the way you do here will break any other uses of the crate in the same process which rely on the environment-checking behavior. 

I didn't quite understand why the environment matters here. Isn't the schema crate internal and intended to be used on the server side only?

Here, we're basically specifying that the code colors the plan as requested, rather than depending on the local environment — which seems right to me. Local env setting should matter where the CLI running not at the server code. NO?

> We already have a dependency on and uses of `colored` prior to this PR, so that doesn't seem like a great idea.

 I can't think of any potential use case that this would break. If you can provide one, I may get convinced.

gefjon:
> > `set_override`, perhaps obviously, sets a global flag, and so setting it and then never unsetting it the way you do here will break any other uses of the crate in the same process which rely on the environment-checking behavior.
> 
> I didn't quite understand why the environment matters here. Isn't the schema crate internal and intended to be used on the server side only?
> 
> Here, we're basically specifying that the code colors the plan as requested, rather than depending on the local environment — which seems right to me. Local env setting should matter where the CLI running not at the server code. NO?

Is your intention that we pretty-print and colorize the schema on the server, and then send the text, including ANSI escapes, over HTTP? That seems wrong to me; I had assumed we would send the plan as JSON, and then have the CLI pretty-print it as desired. Note that ANSI coloring is not the only standard for coloring in terminals - [some Windows terminals use something else](https://superuser.com/questions/413073/windows-console-with-ansi-colors-handling), and alternative user agents would have entirely other mechanisms for colorizing output (e.g. a web client like our website or [this unofficial one](https://stdb-admin.jlavocat.eu/settings) would use HTTP and CSS; a native GUI system would just draw colored text or pixels).

> >  We already have a dependency on and uses of colored prior to this PR, so that doesn't seem like a great idea.
>
> I can't think of any potential use case that this would break. If you can provide one, I may get convinced.

Assuming we do go ahead with pretty-printing on the server (which I object to, but let's set that aside for the time being), we later decide that some other thing should be colorized, e.g. something in SpacetimeDB's log output. This printing should actually respect `NO_COLOR`, as it's going to be printed directly to stdio, which may be going to a user's TTY or to some CI runner or god knows what else. We already have a dependency on `colored`, so we just go ahead and use that. This behaves correctly and respects `NO_COLOR` at startup, but then as soon as a user goes to pre-publish a new version, the logs start ignoring `NO_COLOR` and printing ANSI escape sequences into Elasticsearch.

Shubham8287:
Yes, I intended to do it at the Server and send over HTTP.
And I know its not great but It has been done this way because [`MigratePlan`](https://github.com/clockworklabs/SpacetimeDB/blob/master/crates/schema/src/auto_migrate.rs#L21)  type does not look trivial to serialize, its deeply nested with lifetimes.

I beleive James took similar https://github.com/clockworklabs/SpacetimeDB/pull/2065 approach in previous attempt due to same reason.

I have also mentioned about this subtly here -https://github.com/clockworklabs/SpacetimeDB/issues/2962#issuecomment-3131650582, Though, I should have mentioned more clearly (I mistakenly assumed it a common knowledge and agreed-upon approach).

I agree, sending JSON and prettify-ing at client is Correct and should have done, its just an extra effort.

gefjon:
> Yes, I intended to do it at the Server and send over HTTP. And I know its not great but It has been done this way because [`MigratePlan`](https://github.com/clockworklabs/SpacetimeDB/blob/master/crates/schema/src/auto_migrate.rs#L21) type does not look trivial to serialize, its deeply nested with lifetimes.
> 
> I beleive James took similar #2065 approach in previous attempt due to same reason.
> 
> I have also mentioned about this subtly here -[#2962 (comment)](https://github.com/clockworklabs/SpacetimeDB/issues/2962#issuecomment-3131650582), Though, I should have mentioned more clearly (I mistakenly assumed it a common knowledge and agreed-upon approach).
> 
> I agree, sending JSON and prettify-ing at client is Correct and should have done, its just an extra effort.

If that is indeed the plan, then I feel strongly that using `colored` in this way is a ticking time bomb waiting to turn into an extremely obnoxious and hard-to-debug bug. This is clearly not the intended use case for that library and their API does not support what we're trying to do here.

I am comfortable deferring sending a structured format like JSON over the wire as future work and using this design as a stop-gap solution, but please make a ticket so we can track this. Note that we do not need to (and indeed should not) serialize the actual auto-migration plans directly into JSON (that would expose them and require that we stabilize them, which we don't want to do), instead we would serialize to a semi-structured intermediate format which clients can then pretty-print as appropriate.

Shubham8287:
Checking out termcolor, [ColorChoice](https://docs.rs/termcolor/1.4.1/termcolor/enum.ColorChoice.html) looks great. It will also let us implement `PlainFormatter` without regex hack.

Shubham8287:
@gefjon  - https://github.com/clockworklabs/SpacetimeDB/issues/3184

Issue #3047, opened 2025-07-01
C#/Unity SDK - Reduce coupling between SpacetimeDBClient.cs and Table.cs


kazimuth:
Any code in SpacetimeDBClient.cs that touches individual rows should be moved to Table.cs. This should allow us to stop using IStructuralReadWrite (effectively `void *`) all over the place. It also may improve performance in the future if it enables us to only use compile-time Row types, which could improve performance if we eventually allow Row struct types. This should also make it easier to implement https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/issues/337 and help Ryan learn the architecture of the SDK.

rekhoff:
Created branch [rekhoff/reduce_client_tables_coupling](https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/tree/rekhoff%2Freduce_client_tables_coupling) to store changes prior to creating a PR.

PR #3122, opened 2025-08-04
Implement reduced coupling between SpacetimeDBClient.cs and Table.cs

rekhoff:
## Description of Changes
Moves table-specific operations out of `SpacetimeDBClient.cs` and into `Table.cs` in order to reduce coupling between the two files.

This is the implementation of https://github.com/clockworklabs/SpacetimeDB/issues/3047

This is a PR being duplicated/migrated from https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/pull/346, which contains prior discussion/review/approval.

## API

 - [ ] This is an API breaking change to the SDK

## Requires SpacetimeDB PRs
No PRs needed, works with latest master

## Testsuite
SpacetimeDB branch name: master

## Testing
- [X] Opened and ran Blackhol.io within Unity Editor and successfully connecting to a locally running server
- [X] Compiled a WebGL build of Blackhol.io and successfully connecting to a locally running server
- [x] Opened and ran BitCraft within Unity Editor and successfully connecting to a locally running server


sdks/csharp/src/CompressionHelpers.cs
```diff
@@ -0,0 +1,148 @@
...
+
+                case CompressableQueryUpdate.Gzip(var bytes):
+                    decompressedStream = CompressionHelpers.GzipReader(new MemoryStream(bytes.ToArray()));
+                    break;
+
+                default:
+                    throw new InvalidOperationException();
+            }
+
+            // TODO: consider pooling these.
```
jdetter, line 106: Consider making a ticket for this if we haven't already

jdetter (approved):
LGTM, one small comment on this. Plus the original PR was approved by James as well so this should be fine. Thanks for moving this over 👍

sdks/csharp/src/CompressionHelpers.cs
rekhoff: Created https://github.com/clockworklabs/SpacetimeDB/issues/3125 to track this TODO item.

PR #3131, opened 2025-08-06
Drop log for already-compressed snapshot to debug

gefjon:
# Description of Changes

On databases with many already-compressed snapshots, this was leading to log spam without providing any useful information.

N/a

1

# Testing

N/a


jdetter (approved):
I'm happy with this, I also verified that we're generally running with logging set to `info` so this should filter this log out.

PR #3133, opened 2025-08-07
Confirmed reads

kim:
# Description of Changes

Implements [subscribing to durable commits](https://github.com/clockworklabs/SpacetimeDBPrivate/issues/1594).

The setting works on a per-connection level, and essentially just delays sending transaction updates until the transaction is reported as durable by the database.

For connectionless SQL operations, the setting works per-request. No SQL syntax is provided by this patch to toggle the configuration.

After some deliberation, I opted to obtain the offset when a transaction commits (as opposed to when it starts). This creates some mild inconvenience, because we prevent the transaction from committing until the corresponding subscription updates are enqueued.
The strategy is, however, more correct should we ever support weaker isolation levels, and it is easier to document.

Follow-ups include:

- Provide SQL syntax (`SET synchronous_commit = ON` or something)
- C# and TypeScript SDKs
- Reference docs?
 

Not breaking, but adds a parameter to the subscribe and sql endpoints.

4

To the author's understanding, ordering of outbound messages is not changed by this patch, even if there are messages that don't have a transaction offset (such as error messages). I.e. while waiting for the transaction offset of a message to become durable, no message enqueued after that message will be delivered. This may not be desirable in some cases.

The patch may contain concurrency bugs, e.g. awaiting futures that may never resolve.

# Testing

- [x] Implemented a new test in the `module_subscription_actor` module
- [x] Added unit tests for the core logic in `ClientConnectionReceiver` 

It would be desirable to also have integration-level tests, but I'm currently unsure how to write those without being able to control if and when the database reports an offset as durable.

### DIFF
--- a/crates/cli/src/subcommands/subscribe.rs
+++ b/crates/cli/src/subcommands/subscribe.rs
@@ -2,7 +2,7 @@ use anyhow::Context;
 use clap::{value_parser, Arg, ArgAction, ArgMatches};
 use futures::{Sink, SinkExt, TryStream, TryStreamExt};
 use http::header;
-use http::uri::{PathAndQuery, Scheme};
+use reqwest::Url;
 use serde_json::Value;
 use spacetimedb_client_api_messages::websocket::{self as ws, JsonFormat};
 use spacetimedb_data_structures::map::HashMap;
@@ -137,23 +137,20 @@ pub async fn exec(config: Config, args: &ArgMatches) -> Result<(), anyhow::Error
     let api = ClientApi::new(conn);
     let module_def = api.module_def().await?;
 
+    let mut url = Url::parse(&api.con.db_uri("subscribe"))?;
     // Change the URI scheme from `http(s)` to `ws(s)`.
-    let mut uri = http::Uri::try_from(api.con.db_uri("subscribe"))?.into_parts();
-    uri.scheme = uri.scheme.map(|s| {
-        if s == Scheme::HTTP {
-            "ws".parse().unwrap()
-        } else if s == Scheme::HTTPS {
-            "wss".parse().unwrap()
-        } else {
-            s
-        }
-    });
+    url.set_scheme(match url.scheme() {
+        "http" => "ws",
+        "https" => "wss",
+        unknown => unreachable!("Invalid URL scheme in `Connection::db_uri`: {unknown}"),
+    })
+    .unwrap();
     if confirmed {
-        append_query_param(&mut uri, ("confirmed", "true"));
+        url.query_pairs_mut().append_pair("confirmed", "true");
     }
 
     // Create the websocket request.
-    let mut req = http::Uri::from_parts(uri)?.into_client_request()?;
+    let mut req = url.into_client_request()?;
     req.headers_mut().insert(
         header::SEC_WEBSOCKET_PROTOCOL,
         http::HeaderValue::from_static(ws::TEXT_PROTOCOL),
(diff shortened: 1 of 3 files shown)


crates/client-api/src/routes/subscribe.rs
```diff
@@ -1057,26 +1091,35 @@ async fn ws_send_loop(
...
-                        msg
-                    ).await;
-                    serialize_buf = msg_alloc;
-
-                    if let Err(e) = res {
-                        log::warn!("websocket send error: {e}");
-                        return;
+            maybe_message = messages.recv(), if !closed => {
+                let Some(message) = maybe_message else {
+                    // The message sender was dropped, even though no close
```
kim, line 1096: Unsure why the `send_loop_terminates_when_unordered_closed` started to fail with this patch, but the edge case seems legit.

Only thing to consider is whether the `closed` atomic should be re-checked when entering this branch.

crates/core/src/client/client_connection.rs
```diff
@@ -181,7 +331,12 @@ impl ClientConnectionSender {
             Err(mpsc::error::TrySendError::Full(_)) => {
                 // we've hit CLIENT_CHANNEL_CAPACITY messages backed up in
                 // the channel, so forcibly kick the client
-                tracing::warn!(identity = %self.id.identity, connection_id = %self.id.connection_id, "client channel capacity exceeded");
+                tracing::warn!(
+                    identity = %self.id.identity,
+                    connection_id = %self.id.connection_id,
+                    confirmed_reads = self.config.confirmed_reads,
```
kim, line 379: Note that a client with confirmed reads turned on has a somewhat higher risk to exceed the capacity.

It's not very likely, because `CLIENT_CHANNEL_CAPACITY` is such a large number, but is this warning loud enough?

crates/core/src/host/module_host.rs
```diff
@@ -1109,74 +1110,79 @@ impl ModuleHost {
...
-                                total_host_execution_duration,
-                            }),
-                            None,
-                        ),
-                    };
-
-                    subscriptions.send_client_message(client, message, tx)?;
-                    Ok::<Option<ExecutionMetrics>, anyhow::Error>(metrics)
-                })
+                let (tx_offset_sender, tx_offset_receiver) = oneshot::channel();
```
kim, line 1113: One-off queries are not really tested anywhere?

kim:
No idea what might be the matter on windows

crates/cli/src/subcommands/subscribe.rs
```diff
@@ -334,3 +339,21 @@ fn format_output_json(msg: &ws::DatabaseUpdate<JsonFormat>, schema: &RawModuleDe
 
     Ok(output)
 }
+
+fn append_query_param(uri: &mut http::uri::Parts, (k, v): (&str, &str)) {
```
bfops: this function concerns me. I know it works for the one place we currently use it, but it has some sharp edges and I'm worried about leaving it around in the codebase.

e.g. I feel like it will likely not work correctly in the event that either `k` or `v` need escaping.

I'm also surprised that we append query params to `path` instead of adding them to `query`? Why do we do that?

Could we maybe just update the `http::uri::Parts` to be a `url::Url` instead and then use the built-in helper functions for adding query params? (which would involve updating the call site context as well)

Let me know if you want my help with converting this, I know I'm the codeowner here.

crates/core/src/client/client_connection.rs
```diff
@@ -168,11 +346,18 @@ impl ClientConnectionSender {
 
     /// Send a message to the client. For data-related messages, you should probably use
     /// `BroadcastQueue::send` to ensure that the client sees data messages in a consistent order.
-    pub fn send_message(&self, message: impl Into<SerializableMessage>) -> Result<(), ClientSendError> {
-        self.send(message.into())
+    pub fn send_message(
```
gefjon, line 356:
```suggestion
    ///
    /// If `message` is the result of evaluating a query, then `tx_offset` should be
    /// the TX offset of the database state against which the query was evaluated.
    /// If `message` is not the result of evaluating a query (e.g. it reports an error),
    /// `tx_offset` should be `None`.
    /// For clients which have requested only confirmed durable reads,
    /// the sender will delay sending `message` until the `tx_offset` is confirmed.
    pub fn send_message(
```

crates/core/src/host/module_host.rs
gefjon: Correct, sadly.

crates/cli/src/subcommands/subscribe.rs
kim: I was so in awe about the brilliance of `http::uri::Parts` not actually being all the parts, that I missed that `tungstenite` supports `url::Url` if given the right feature flags :eyeglasses:

bfops: `http::uri::Parts` just seems like sadness to work with :sweat_smile: 

This version looks so much clearer, thank you :+1:

PR #3137, opened 2025-08-07
Endpoint for pretty print migration plan

Shubham8287:
# Description of Changes
- Adds endpoint for for pretty printing migration plan.
- It also changes current `publish` endpoint to optionally provide `MigrationToken` and `MigrationPolicy` to allow migration with breaking clients.

Backward compatible change to existing API and new Api

2

# Testing
- Existing smoketest should cover changes for `publish` endpoint.
- For pretty print endpoint, smoketests can be written only after cli changes.

### DIFF
--- a/crates/client-api/src/routes/database.rs
+++ b/crates/client-api/src/routes/database.rs
@@ -648,20 +648,20 @@ pub async fn publish<S: NodeDelegate + ControlStateDelegate>(
 }
 
 #[derive(serde::Deserialize)]
-pub struct PrintPlanParams {
+pub struct PrePublishParams {
     name_or_identity: NameOrIdentity,
 }
 
 #[derive(serde::Deserialize)]
-pub struct PrintPlanQueryParams {
+pub struct PrePublishQueryParams {
     #[serde(default)]
     style: PrettyPrintStyle,
 }
 
 pub async fn pre_publish<S: NodeDelegate + ControlStateDelegate>(
     State(ctx): State<S>,
-    Path(PrintPlanParams { name_or_identity }): Path<PrintPlanParams>,
-    Query(PrintPlanQueryParams { style }): Query<PrintPlanQueryParams>,
+    Path(PrePublishParams { name_or_identity }): Path<PrePublishParams>,
+    Query(PrePublishQueryParams { style }): Query<PrePublishQueryParams>,
     Extension(auth): Extension<SpacetimeAuth>,
     body: Bytes,
 ) -> axum::response::Result<axum::Json<PrintPlanResult>> {


crates/schema/src/auto_migrate.rs
```diff
@@ -74,6 +79,94 @@ impl<'def> MigratePlan<'def> {
...
+}
+
+pub struct MigrationToken {
+    pub database_identity: Identity,
+    pub old_module_hash: spacetimedb_lib::Hash,
+    pub new_module_hash: spacetimedb_lib::Hash,
+}
+
+impl MigrationToken {
+    pub fn hash(&self) -> spacetimedb_lib::Hash {
```
Shubham8287, line 169: I have  not included any secret in token generation which means, anyone with enough intent could recreate the token generation logic and bypass it. Considering this token act as UX safegaurd and not for security. I feel, this acceptable and we don't have to worry about managing a secret.

Shubham8287:
CI is breaking because of base branch.

crates/client-api/src/routes/database.rs
```diff
@@ -835,6 +962,7 @@ where
 
         axum::Router::new()
             .route("/", self.root_post)
+            .route("/print-plan/:name_or_identity", self.print_migration_plan)
```
Shubham8287: I have used "print-plan" as endpoint instead of "pre-publish" as former sounds more clear.

gefjon: I disagree. `pre-publish` clearly relates this operation to publishing, whereas `print-plan` could do pretty much anything.

crates/client-api/src/routes/database.rs
```diff
@@ -469,6 +474,9 @@ pub struct PublishDatabaseQueryParams {
     #[serde(default)]
     clear: bool,
     num_replicas: Option<usize>,
+    // `Hash` of `MigrationToken` to be checked if `MigrationPolicy::BreakClients` is set.
```
gefjon:
```suggestion
    /// `Hash` of `MigrationToken` to be checked if `MigrationPolicy::BreakClients` is set.
    ///
    /// Users obtain such a hash from [`print_migration_plan`]
    /// via the `/database/:name_or_identity/pre-publish POST` route.
    /// This is a safeguard to require explicit approval for updates which will break clients.

```

crates/client-api/src/routes/database.rs
gefjon: Most notably, I would confuse `print-plan` with displaying a query plan for a given SQL query.

crates/client-api/src/routes/database.rs
```diff
@@ -469,6 +474,9 @@ pub struct PublishDatabaseQueryParams {
...
+    token: Option<spacetimedb_lib::Hash>,
+    policy: Option<MigrationPolicy>,
```
gefjon: I think it would be clearer to default this to `Compatible` rather than wrapping it in `Option`.

crates/client-api/src/routes/database.rs
```diff
@@ -614,6 +643,102 @@ pub async fn publish<S: NodeDelegate + ControlStateDelegate>(
...
 }
 
+#[derive(serde::Deserialize)]
+pub struct PrintPlanParams {
+    name_or_identity: NameOrIdentity,
+}
+
+#[derive(serde::Deserialize)]
+pub struct PrintPlanQueryParams {
+    style: Option<PrettyPrintStyle>,
```
gefjon: Again, defaulting would be preferred over `Option`.

crates/client-api/src/routes/database.rs
```diff
@@ -614,6 +643,102 @@ pub async fn publish<S: NodeDelegate + ControlStateDelegate>(
...
+) -> axum::response::Result<Identity> {
+    let database_identity = name_or_identity.resolve(ctx).await?;
+
+    let database = worker_ctx_find_database(ctx, &database_identity)
+        .await?
+        .ok_or(NO_SUCH_DATABASE)?;
+
+    if database.owner_identity != auth.identity {
+        return Err((
+            StatusCode::BAD_REQUEST,
```
gefjon:
```suggestion
            StatusCode::UNAUTHORIZED,
```

crates/schema/src/auto_migrate.rs
```diff
@@ -74,6 +79,94 @@ impl<'def> MigratePlan<'def> {
...
 }
 
+/// A migration policy that determines whether a module update is allowed to break client compatibility.
+///
+/// `Compatible` requires migration to maintain backward compatibility.
+/// `BreakingChange` allows breaking changes but requires a valid `MigrationToken`.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum MigrationPolicy {
+    Compatible,
+    BreakClients(spacetimedb_lib::Hash),
```
gefjon, line 95: Please add doc comment describing this value and where to obtain it, like in the client-api crate.

crates/schema/src/auto_migrate.rs
gefjon: I agree, but please add a comment to this effect in the code.

gefjon (approved):

Issue #2362, opened 2025-03-07
Create Test of C# Code Generation for Tables with BTree indexes of type Identity


rekhoff:
### Action to perform:
Add a test to [com.clockworklabs.spacetimedbsdk](https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/tree/master) that checks code generation of a `Table` with an index of type `btree` of an `Identity` type is generating C# code as intended.
Manually performed example from https://github.com/clockworklabs/SpacetimeDB/pull/2354 that should be reproduced in an automated test:
1. Created a test project with a Rust server defining a table with a btree index like:
```
#[index(btree)]
pub identity: Identity,
```
and another where a table was defined with:
`#[spacetimedb::table(name = period_timer, scheduled(period, at = scheduled_at), index(name = identity, btree(columns = [identity])))]`
2. Published the server.
3. Created a C# client and generated the bindings.
4. Confirmed issue existed prior to change: IDE reports an error in the table definition of `The type 'SpacetimeDB.Identity' must be convertible to 'System.IComparable<SpacetimeDB.Identity>' in order to use it as parameter 'Column' in the generic class 'SpacetimeDB.RemoteTableHandle<EventContext,Row>.BTreeIndexBase<Column>'`
5. Made change contained in this PR.
6. Rebuilt SpacetimeDB executables, and C# SDK NuGet files.
7. Tested again by republishing and regenerating bindings.
8. Observed IDE no longer reports issue.

Ticket Generated from the comment:
https://github.com/clockworklabs/SpacetimeDB/pull/2354#pullrequestreview-2665201801

rekhoff:
After the implementation of IComparable to BTreeIndexBase was added, the resulting generated code for the C# client wasn't changed, so testing for that behavior is not possible. You can, however, test that BTreeIndexBase implements the IComparable interface. With that goal in mind, https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/pull/277 was opened for that change.

PR #3142, opened 2025-08-08
C#/Unity SDK - Added Unit Test that checks the BTreeIndexBase Column implements IComparable

bfops:
# Description of Changes

Migrating "https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/pull/277" since we are merging that repo into this one.

> Adds a test to [com.clockworklabs.spacetimedbsdk](https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/tree/master) that checks if the BTreeIndexBase class Column implements the IComparable interface. This is the implementation side of issue https://github.com/clockworklabs/SpacetimeDB/issues/2362.

>  - [x] This is an API breaking change to the SDK

# Testing

> - [X] Ran `dotnet test`, all checks pass.

### DIFF
--- a/sdks/csharp/tests~/Tests.cs
+++ b/sdks/csharp/tests~/Tests.cs
@@ -81,7 +81,7 @@ public class Tests
         Assert.False(GenericEqualityComparer.Instance.Equals(statusFailed, statusFailedUnequalValue));
         Assert.False(GenericEqualityComparer.Instance.Equals(statusCommitted, statusOutOfEnergy));
     }
-    
+
     public class BTreeIndexBaseColumnImplementsIComparableTest
     {
 
@@ -103,7 +103,7 @@ public class Tests
                 Identity = new(this);
             }
         }
-        
+
         [Fact]
         public void Identity_ShouldImplementIComparable()
         {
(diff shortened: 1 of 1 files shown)


sdks/csharp/tests~/Tests.cs
```diff
@@ -81,4 +81,56 @@ public static void GenericEqualityComparerCheck()
...
+                public IdentityIndex(UserHandle table) : base(table) { }
+            }
+
+            public readonly IdentityIndex Identity;
+
+            internal UserHandle(DbConnection conn) : base(conn)
+            {
+                Identity = new(this);
+            }
+        }
```
bfops, line 152: Outstanding comment from the other PR:

> This code won't catch problems that occur if the code generation changes in the future. To catch such problems, we'd need to write the same test against generated code. For example, we could add an index to the Message.Sender field in quickstart-chat -- these tests have access to the quickstart-chat generated code.
> 
> If you want to do that, it would be great; if not, I'm willing to merge this, but would appreciate a comment saying why this class is here. // Copy of generated code for a BTreeIndex; TODO: update if that code changes is messy but at least documents what is going on.

bfops:

PR #3146, opened 2025-08-08
C#/Unity SDK - Pre-hash rows and add SmallHashSets

bfops:
# Description of Changes

Migrating "https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/pull/318" since we are merging that repo into this one.

> Annotate rows with their hashes. This is halfway to the old solution of annotating code with their serialized forms; it speeds up insertion into `BTreeIndex`es while saving a byte array allocation per-row.
> (Wait, why do we hash when inserting into `BTreeIndex`es, you ask? It's because we use a `HashSet` to store the multiple rows corresponding to a non-unique key.)
> 
> This saves work on the main thread repeatedly hashing values. Once we parallelize message processing, this work will be spread out over multiple threads too.
> 
> Also adds a data type `SmallHashSet<T>` for use in BTreeIndexes. This is a struct that can store at most one element without allocating. This gives dramatic allocation reduction on initial connection -- it seems that most rows in BTreeIndexes don't have any other rows with the same key. 
> Then I manually specialized this, since it's only used on one type -- `PreHashedRow`. using `SmallHashSetOfPreHashedRow` instead of `SmallHashSet<PreHashedRow>` allows us to inline some methods and pass around fewer pointers while hashing things, which is probably good.
> 

None

# Testing

> I will test:
> 
> - [ ] blackholio
> - [ ] bitcraft


bfops:
Given the discussion in the original PR, I think maybe we should close this? @rekhoff what do you think?

bfops:

PR #3148, opened 2025-08-08
C#/Unity SDK - Add expiration for stored request durations

bfops:
# Description of Changes

Migrating https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/pull/193 since we are merging that repo into this one.

> This adds configurable expiration time for request durations. This is to prevent memory leak in long-running applications where otherwise the queue would only keep growing indefinitely.

None

# Testing

- [ ] 
- [ ]

### DIFF
--- a/sdks/csharp/src/Stats.cs
+++ b/sdks/csharp/src/Stats.cs
@@ -7,10 +7,13 @@ namespace SpacetimeDB
 {
     public class NetworkRequestTracker
     {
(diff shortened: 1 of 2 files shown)


bfops:
@jdetter or @rekhoff are you familiar with the original motivation here?

rekhoff:
> @jdetter or @rekhoff are you familiar with the original motivation here?

I am not familiar with the original motivation, no.

bfops:

PR #3159, opened 2025-08-13
Fix module hotswapping for connected clients

kim:
The `Clone` impl for `ClientConnection` would create an independent
instance that could not observe module hotswapping. This would result in
methods called on a replaced `ModuleHost` to fail, because that host
exited already.

Fix by reading the `ModuleHost` from the watch channel directly, instead
of maintaining a redundant copy.

Also fix `watch_module_host` to properly mark the current module host as
seen.

2

# Testing

- [x] test suite passes
- [x] ran @bfops repro script

### DIFF
--- a/crates/core/src/client/client_connection.rs
+++ b/crates/core/src/client/client_connection.rs
@@ -506,6 +506,16 @@ impl ClientConnection {
         self.sender.clone()
     }
 
+    /// Get the [`ModuleHost`] for this connection.
+    ///
+    /// Note that modules can be hotswapped, in which case the returned handle
+    /// becomes invalid (i.e. all calls on it will result in an error).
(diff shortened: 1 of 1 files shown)


crates/core/src/client/client_connection.rs
```diff
@@ -510,6 +506,10 @@ impl ClientConnection {
         self.sender.clone()
     }
 
+    pub fn module(&self) -> ModuleHost {
+        self.module_rx.borrow().clone()
+    }
```
gefjon, line 521: This could use a doc comment warning that it should not be cached.

gefjon (approved):

jdetter:
@bfops confirmed as well this this specifically fixed the module hot-reloading issue that we were having when he retested

PR #3172, opened 2025-08-14
fix incoming message queue length metric

joshua-spacetime:
# Description of Changes

We stopped incrementing the incoming queue length metric. This patch increments it again and adds a regression test.

None

1

# Testing

- [x] Regression test


mamcx (approved):
LGTM

Issue #2647, opened 2025-04-19
CSharp quickstart does not compile


BinarySpike:
On https://spacetimedb.com/docs/modules/c-sharp/quickstart

The below code does not compile:
```csharp
[Reducer]
public static void SetName(ReducerContext ctx, string name)
{
    name = ValidateName(name);

    var user = ctx.Db.user.Identity.Find(ctx.Sender);
    if (user is not null)
    {
        user.Name = name;
        ctx.Db.user.Identity.Update(user);
    }
}
```

I get the errors:
```
...server\Lib.cs(30,12): error CS1061: 'Module.User?' does not contain a definition for 'Name' and no accessible extension method 'Name' accepting a first argument 
of type 'Module.User?' could be found (are you missing a using directive or an assembly reference?) [...\server\StdbModule.csproj]

...server\Lib.cs(31,35): error CS1503: Argument 1: cannot convert from 'Module.User?' to 'Module.User' [...\server\StdbModule.csproj]
```

This is due to `.Find(...)` returning a `Module.User?` (nullable) and both `user.Name` and `Identity.Update(...)` expecting a `Module.User` (**not** nullable)

This is also the case for the following functions `ClientConnected` and `ClientDisconnected`

🤔

bfops:
I'm not able to reproduce this issue. The `not null` check should be sufficient here 🤔

Can you share your `StdbModule.csproj` and `spacetime --version`?

ScherrerMichael:
Hello, I am also facing this issue following the quickstart.

spacetime --version output:
`spacetimedb tool version 1.1.0; spacetimedb-lib version 1.1.0;`

csproj file:

```
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>net8</TargetFramework>
    <RuntimeIdentifier>wasi-wasm</RuntimeIdentifier>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="SpacetimeDB.Runtime" Version="1.1.0" />
  </ItemGroup>

</Project>
```

using:
```
Version: 1.99.3 (user setup)
Commit: 17baf841131aa23349f217ca7c570c76ee87b957
Date: 2025-04-15T23:18:46.076Z
Electron: 34.3.2
ElectronBuildId: 11161073
Chromium: 132.0.6834.210
[...]

PR #3173, opened 2025-08-15
Updated C# quickstart-chat to pattern match variable assignment

rekhoff:
# Description of Changes

To address user issue https://github.com/clockworklabs/SpacetimeDB/issues/2647 , the C# implementation of quickstart-chat files and documentation have been updated to pattern match the variable assignment of User.

This has the benefit of the variables never being perceived by the analyzers as a nullable types.

Not API breaking

1

# Testing

- [X] Tested updated module code locally, following instructions from the documentation.


jdetter (approved):
Code looks reasonable + I tested via publishing to a local server + running the client and it seems to still work. Thanks Ryan 👍

PR #3177, opened 2025-08-18
rust: `default` macro

Shubham8287:
# Description of Changes
PR introduces support for column-level default values via a new `#[default(...)]` attribute.

It also validates, `default` macro is not used along with `primary_key`, `unique` or `auto_inc`.

NA

2

# Testing
Start using macro in `module-test`.


crates/bindings-macro/src/table.rs
```diff
@@ -494,12 +496,33 @@ impl ColumnAttr {
...
             None
         })
     }
 }
 
+fn parse_default_attr(attr: &syn::Attribute, ident: &Ident) -> syn::Result<ColumnAttr> {
+    let mut default_value = None;
+
+    attr.parse_nested_meta(|meta| {
+        if meta.path.is_ident("value") {
```
Shubham8287: I am not yet able to figure out parsing `#[default(value)]`, I will try that again. Any help will also be appreciated.

gefjon: I believe you want to use `Attribute::parse_args` rather than `parse_nested_meta`. [The docs on this method have an example that I believe is relevant](https://docs.rs/syn/latest/syn/struct.Attribute.html#method.parse_args):

```rust
use syn::{parse_quote, Attribute, Expr};

let attr: Attribute = parse_quote! {
    #[precondition(value < 5)]
};

if attr.path().is_ident("precondition") {
    let precondition: Expr = attr.parse_args()?;
    // ...
}
```

crates/bindings-macro/src/table.rs
```diff
@@ -566,28 +582,42 @@ pub(crate) fn table_impl(mut args: TableArgs, item: &syn::DeriveInput) -> syn::R
...
+                    default_value = Some(expr);
+                }
             }
         }
 
+        if let Some(default_value) = &default_value {
+            if auto_inc.is_some() || primary_key.is_some() || unique.is_some() {
+                return Err(syn::Error::new(
+                    default_value.span(),
+                    "invalid combination: auto_inc, unique index or primary key cannot have a default value",
```
gefjon, line 596: Allowing `default(0)` for `auto_inc` columns seems potentially meaningful. We might want a different syntax for that, like `default(auto_inc)` or something. Fine to not support that in the initial version, though; I'm just musing about the future.

gefjon (approved):

Shubham8287:
Thanks, I will hold this PR to merge until we merge the other PR to make this functional.

egormanga:
How to use these default values? Do we pass some sentinel value (as `0` with `auto_inc`), or omit the field (does Rust even allow this?), or anything else?

gefjon:
> How to use these default values? Do we pass some sentinel value (as `0` with `auto_inc`), or omit the field (does Rust even allow this?), or anything else?

@egormanga These will be used with an upcoming feature that allows adding columns during "supervised" automigrations.

egormanga:
It would be great if those defaults were applied to missing columns when inserting via SQL. I'd love to see an issue on this.

Issue #3017, opened 2025-08-01
Reduce `tools/publish-crates.sh` to only publish `bindings` and `sdk`


bfops:
We really only need to publish the module bindings and rust SDK crates; we don't actually need to publish the CLI or standalone server crate, we just do it because they're in there among our other crates.

Really, we just need to publish the crates that are used by module code and client code (which is what we do for e.g. C#).

This would reduce the number of crates we have to publish from 32 to 11. And the server-related crates tend to be slower to build as well.

jdetter:
I guess doing `cargo install spacetimedb` doesn't really work anyway because we have many other things that need to be installed as well? Our install process doesn't really fit well with how cargo installs packages anyway. We already killed homebrew support mainly because of this and I think it seems reasonable to no longer support `cargo install spacetimedb` as well. @cloutiertyler what do you think?

jdetter:
Upgraded to a P1 so that we hit this during standup because we don't look at P2s during standup

PR #3180, opened 2025-08-19
Reduce `tools/publish-crates.sh` to only publish `bindings` and `sdk`

bfops:
# Description of Changes

Fixes https://github.com/clockworklabs/SpacetimeDB/issues/3017. See that issue for more context.

None

1

# Testing

We should do a little extra verification on the next release.

### DIFF
--- a/tools/publish-crates.sh
+++ b/tools/publish-crates.sh
@@ -39,7 +39,6 @@ if [ $DRY_RUN -ne 1 ]; then
 fi
 
 BASEDIR=$(pwd)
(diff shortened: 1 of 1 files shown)


jdetter:
This should be fine, I'm syncing with Tyler to just confirm that he's fine with this as well but I think we should be able to move this forward. I myself have no objections to this. I also put a comment here: https://github.com/clockworklabs/SpacetimeDB/issues/3017#issuecomment-3201620999

PR #3182, opened 2025-08-19
Separate out TypeScript module library from the SDK

cloutiertyler:
# Description of Changes

Please note, much of the code changed in this PR is generated code.

This change updates the TypeScript SDK to use a new `spacetimedb` TypeScript library which lives under the `/crates/bindings-typescript` folder alongside `/crates/bindings-csharp`. Just like with the C# bindings library, the types for `AlgebraicType` and `RawModuleDef` are now code generated with a script in `/crates/codegen/examples`.

Pulling this out into a library allows us to use the same types and serialization code on both the server and the client for TypeScript modules.

In the process of making this change I also found and fixed several issues with the TypeScript code generation. Namely an issue with recursive types and an issue with the `never` type. I also removed any use of `namespace`s since those are a TypeScript only feature, and we want to have JavaScript + types so that we can use the generated code with ESBuild without TSC.

I have also improved the npm/pnpm scripts to be able to generate TypeScript code for us automatically by running `pnpm generate` for any place that we have to generate typescript code. Namely:

- Quickstart module bindings
- AlgebraicType/ModuleDef for TypeScript module library
- Client API messages for the TypeScript API
- TestApp module bindings

IMPORTANT! This is an API breaking change for clients, as such it should be a major version change. However, I am going to see if I can shim in the old API as best as possible to make it compatible.

Notably, we were previously exporting APIs that end users do not need, and I don't think it would ever occur to them to use, namely the whole `AlgebraicValue` API and also the `AlgebraicType` API. In principle, no one should have a need for these, although it was technically possible for them to use it.

Indeed, we could potentially even just remove AlgebraicType completely from the API by directly code generating the serialization code.

Listed below are all of the **BREAKING** changes to the API and their effect:

- `AlgebraicType` is now a structural union literal type instead of a class (nominal type), this is a consequence of generating the type with our code gen. Users did not have a reason to use `AlgebraicType` directly.
- The `AlgebraicValue` type has been removed entirely. This was previously a class that was exported from the SDK, but very unlikely to be used by users.
- The `ProductValue` type has been removed.
- The `ReducerArgsAdapter` and `ValueAdapter` types, which were used with AlgebraicValues have been removed.
- Generated code has changed incompatibly so users will have to regenerate code when upgrading to this version. Technically a breaking change, but pretty easy to fix.

Listed below are the non-breaking API changes:
- I am now exporting generated product types as the default export in addition to how I was exporting them before
- For generated sum type `T`, I am now exporting a `TVariants` namespace which has the types of all the variants for `T`. This was previously exposed on the namespace `T`, but was inaccessible in modules other than the one it was defined in because I also export a type `T` in addition to namespace `T` and in the type position `T` was being interpreted as a type rather than a namespace. It's unclear why TypeScript resolved it as the `T` namespace within the module in which is was defined previously. Anyway, since the old types were apparently unobservable to users, I've replaced them with the types in `TVariants`. (Open to other names for this namespace).
- I fixed a bug where never types (sum types with no variants) were not correctly generated.

3. The most complex thing about the PR are the potential impacts to the API. I am reasonably certain, but not 100% certain that I have accounted for everything above.

# Testing

- [x] I ran `pnpm test` which runs all the tests in the repo including an integration test which tests the connection to SpacetimeDB. I also fixed broken tests.


crates/bindings-typescript/LICENSE.txt
```diff
@@ -0,0 +1 @@
+../../LICENSE.txt
```
bfops: I don't think this should reference the root LICENSE.txt. It should reference one of the licenses in `licenses/`.

cloutiertyler: Good catch, thank you. Fixed.

gefjon (commented):
I have no objections to the described breaking changes. I think that ideally we should save them for a minor version bump and include them in the release notes, under a heading like "Minor incompatible changes" or something. But I don't see any reason to hold them for a major version bump other than fanatical devotion to the SemVer spec.

I don't otherwise feel equipped to review this PR, and would prefer to leave it to others.

bfops (changes requested):
the `.prettierignore` file should not be moved without changes. The previous file was relative to the directory it was in. We should probably at least change `/.github` to `.github`.

cloutiertyler:
I addressed @bfops's comment.

sdks/typescript/packages/sdk/src/client_api/index.ts
```diff
@@ -102,6 +104,9 @@ export { UpdateStatus };
 const REMOTE_MODULE = {
   tables: {},
   reducers: {},
+  versionInfo: {
+    cliVersion: '1.3.0',
```
jsdt: If the sdk changes will require a cli update, we should update the `_MINIMUM_CLI_VERSION` [here](https://github.com/clockworklabs/SpacetimeDB/blob/434894f73dcc4fcda3b6465280ec08a09b62a9e7/sdks/typescript/packages/sdk/src/version.ts#L115).

For making changes to generated code that require a certain version of the typescript SDK, we could add a minimum SDK version to `versionInfo`, so the client could notice immediately on startup.

crates/bindings-typescript/src/autogen/raw_constraint_def_v_8_type.ts
```diff
@@ -0,0 +1,80 @@
...
+// WILL NOT BE SAVED. MODIFY TABLES IN YOUR MODULE SOURCE CODE INSTEAD.
+
+// This was generated using spacetimedb cli version 1.3.0 (commit be6130a209e59b047af47123db3be3182ede6d36).
+
+/* eslint-disable */
+/* tslint:disable */
+// @ts-nocheck
+import {
+  AlgebraicType,
+  AlgebraicValue,
```
jsdt: The generated imports still have `AlgebraicValue`, even though it has been removed.

sdks/typescript/packages/sdk/src/client_api/index.ts
cloutiertyler: I actually don't have a lot of context on this new version number, do you want me to add an addition SDK version to this type? Should that perhaps come in a separate PR?

crates/bindings-typescript/src/autogen/raw_constraint_def_v_8_type.ts
cloutiertyler: I have since fixed this problem (and others, by changing the PR so that we import these built-in types with a `__` prefix)

sdks/typescript/packages/sdk/src/client_api/index.ts
jsdt: Adding a minimum SDK version is something we could consider doing in a different PR, but we already have the `_MINIMUM_CLI_VERSION` variable that I linked to above, so I think we should set it with the PR that breaks compatibility. That is a variable set inside the typescript SDK that will throw an error if the module bindings were generated with a version of the CLI that is too old to work.

cloutiertyler: Oh I see, you want me to bump the version number. Yes I can do that here.

cloutiertyler: Oh I see, I wasn't aware of that version number that we had in there, but this is a great catch. I just bumped the min CLI version to 1.4.0 since this will require a minor version release of SpacetimeDB (old libs are not compatible with new generated code, and new libs are not compatible with old libraries).

PR #3193, opened 2025-08-21
Fix remaining LICENSE files

bfops:
# Description of Changes

Apparently, I missed several license files in #3002. I'm not sure what method I was using to find them, but apparently it was insufficient.

**This replaces all empty `LICENSE` files with an explicit (symlink to) BSL license, and all apache licenses with symlinks to the root apache license.** This PR does not intentionally change any license terms, so if you see one that changed, **it's a mistake**.

None

1

# Testing

```bash
$ find . -name '*LICENSE*' -type f | grep -v '\.meta$'
./crates/sqltest/standards/LICENSE # this one is an external library that we are not allowed to re-license
./LICENSE.txt # this is the root license
```


gefjon (approved):
Code-owner approval. I am obviously unqualified to have an actual opinion on this.

cloutiertyler (approved):
LGTM glad to fix these

PR #3194, opened 2025-08-21
CI - Fix caching in C#/Unity testsuite

bfops:
# Description of Changes

We had weird caching issues in the C#/Unity testsuite. Somehow, they got triggered only as of https://github.com/clockworklabs/SpacetimeDB/pull/3181 merging, and I have no idea why/how.

I've restored the `id` field of the checkout step (which is used by the cache step), and this _seems_ to have fixed it.

None.

1

# Testing

- [x] It passes on this PR
- [x] It passes in a test PR that combines this change with https://github.com/clockworklabs/SpacetimeDB/pull/3182

### DIFF
--- a/.github/workflows/csharp-test.yml
+++ b/.github/workflows/csharp-test.yml
@@ -71,6 +71,8 @@ jobs:
           cache-all-crates: true
 
       - name: Install SpacetimeDB CLI from the local checkout
+        # Rebuild only if we didn't get a precise cache hit.
+        if: steps.cache-rust-deps.outputs.cache-hit == 'false'
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/csharp-test.yml
```diff
@@ -16,6 +16,7 @@ jobs:
     timeout-minutes: 30
     steps:
       - name: Checkout repository
+        id: checkout-stdb
```
bfops, line 19: this id is explicitly referenced below, but was removed during The Great Repo Merge (because it used to be on a separate "checkout SpacetimeDB" step that was entirely removed).

PR #3195, opened 2025-08-21
Unifies TypeScript packages and command names

cloutiertyler:
# Description of Changes

This PR:
 - standardizes the prettier config across all TypeScript projects
 - adds a root level package.json
 - standardizes all `pnpm` commands to be the same
 - updates documentation accordingly
 - adds some additional typescript testing for serialization and deserialization
 
 **IMPORTANT!** Once this PR merges we will need to change the `compile-and-test` required check to `build-and-test`

No breaking changes.

2 - It in principle doesn't change any code, but could affect deploy processes.

# Testing

- [x] Just the automated testing that we had previously
- [x] I added additional automated tests


.github/workflows/typescript-test.yml
```diff
@@ -8,7 +8,7 @@ on:
   merge_group:
 
 jobs:
-  compile-and-test:
+  build-and-test:
```
bfops, line 11: FYI I believe this job name is used in the required status checks, so we will have to update the required name after this merges

bfops: Can you add that note into the PR description somewhere so it's a bit easier to remember?

docs/docs/appendix.md
```diff
@@ -6,15 +6,15 @@ For each table containing an `#[auto_inc]` column, SpacetimeDB creates a sequenc
 
 ### How It Works
 
-* Sequences in SpacetimeDB use Rust’s `i128` integer type.
```
bfops: If these docs tweaks are important to you, can you please move them into a separate PR? They seem entirely unrelated to the (meaningfully large) actual goal of this PR.

bfops: I typically do something like:
```bash
git checkout -b new-branch origin/master
git checkout old-branch -- file1 file2 ...
git commit -am 'Moving files from other thing'
git push -u
gh pr create --web
```

bfops: Ah is this because we started running `pnpm format` on the docs? That definitely seems like a separate change, since the docs are entirely unrelated to our TypeScript bindings/SDK.

By default I would instead lean towards adding the docs to `.prettierignore`.

.github/workflows/typescript-test.yml
cloutiertyler: Ah okay, yes I can

docs/docs/appendix.md
cloutiertyler: Yes, this is because I'm running the same pnpm format on all typescript directories (directories with package.json). I can exempt markdown from the formatting and add that in a separate PR if you'd like.

cloutiertyler: I do think that eventually we should have docs under prettier so that we don't have non-substantive formatting changes to the docs causing diffs in the future.

cloutiertyler: I've added `**/*.md` to the docs `.prettierignore`. Notably the `cli-reference.md` file formatting is still changed because I moved the not portable `sed` scripts into a node js script which rewrites the file with remark.

docs/.prettierignore
```diff
@@ -0,0 +1 @@
+**/*.md
```
bfops, line 1: maybe just all of this directory for now? It looks like it also reformatted html files and stuff under `docs`. tl;dr imo this PR doing TS package stuff, shouldn't be implementing new tooling for docs.

cloutiertyler: This PR is standardizing how all of our TypeScript repos get formatted, especially TypeScript files. I can exempt html as well, but I think it's important that TypeScript files inside our TypeScript projects (which the docs repo is) get formatted.

docs/tools/markdown-fix.mjs
```diff
@@ -0,0 +1,99 @@
+#!/usr/bin/env node
```
cloutiertyler, line 1: This script replaces the non-portable sed commands that were used to fix up the markdown.

bfops (approved):
Approving the docs formatting + CI changes, with the caveat that I think none of them should actually be happening in this PR. But if I close my eyes and pretend that this PR is only making those changes, they seem good to me.

.github/workflows/ci.yml
```diff
@@ -254,23 +254,46 @@ jobs:
...
+        run: |
+          echo "STORE_PATH=$(pnpm store path --silent)" >> $GITHUB_ENV
+
+      - uses: actions/cache@v4
+        name: Setup pnpm cache
+        with:
+          path: ${{ env.STORE_PATH }}
+          key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
+          restore-keys: |
+            ${{ runner.os }}-pnpm-store-
```
JulienLavocat, line 284:
```suggestion
      - name: Set up Node.js
        uses: actions/setup-node@v4
        with:
          node-version: 18
          cache: pnpm

      - uses: pnpm/action-setup@v4
        with:
          run_install: true
```

Is it possible to use `cache` in setup-node instead?
https://github.com/actions/setup-node?tab=readme-ov-file#usage

.github/workflows/docs-validate-nav-build.yml
```diff
@@ -18,12 +18,35 @@ jobs:
...
+        run: |
+          echo "STORE_PATH=$(pnpm store path --silent)" >> $GITHUB_ENV
+
+      - uses: actions/cache@v4
+        name: Setup pnpm cache
+        with:
+          path: ${{ env.STORE_PATH }}
+          key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
+          restore-keys: |
+            ${{ runner.os }}-pnpm-store-
```
JulienLavocat, line 39: Same as my previous comment, can we use `cache` in setup-node?
https://github.com/actions/setup-node?tab=readme-ov-file#usage

crates/bindings-typescript/src/identity.ts
```diff
@@ -4,11 +4,7 @@ import { hexStringToU256, u256ToHexString, u256ToUint8Array } from './utils';
  * A unique identifier for a user connected to a database.
  */
 export class Identity {
-  data: bigint;
-
-  get __identity__(): bigint {
-    return this.data;
-  }
+  __identity__: bigint;
```
JulienLavocat, line 7: Out of curiosity, why are we using `__` as prefix and suffix?

docs/package.json
```diff
@@ -8,14 +8,19 @@
...
+    "remark-stringify": "^11.0.0",
     "tsx": "^4.19.2",
-    "typescript": "^5.3.2",
+    "typescript": "^5.6.3",
     "unified": "^11.0.5",
     "unist-util-visit": "^5.0.0"
   },
   "scripts": {
-    "build": "tsc --project ./tsconfig.json",
+    "build": "tsc --project ./tsconfig.json && pnpm fix-markdown && prettier --write docs/nav.js",
```
JulienLavocat, line 20: Should a formatting step really be done during build?
```suggestion
    "build": "tsc --project ./tsconfig.json && pnpm fix-markdown",
```

sdks/typescript/packages/sdk/tests/algebraic_type.test.ts
JulienLavocat, line 1: Did you meant to delete this file or move it?

.github/workflows/ci.yml
cloutiertyler: Oh, possibly, I was copying this from elsewhere in our github actions. One thing I'll say is that we plan to unify all of this in the future so that we're just setting it up once. I'd like to keep it the same as what we have elsewhere for now.

.github/workflows/docs-validate-nav-build.yml
cloutiertyler: Same reply as above.

crates/bindings-typescript/src/identity.ts
cloutiertyler: This is because SATS has a structural typing system and these are special system types, so their names are prefixed and suffixed with `__` so that they don't clash with user defined types. In the future we hope to resolve this with a globally unique symbol system.

sdks/typescript/packages/sdk/tests/algebraic_type.test.ts
cloutiertyler: It was renamed to `serde.test.ts` but the diff is so different that git is calling it a delete.

.github/workflows/ci.yml
JulienLavocat: Fair enough!

docs/package.json
cloutiertyler: In this case yes, because nav.js is checked into git and we don't formatting changes included in diffs.

PR #3200, opened 2025-08-25
Fix adding enum variants

Centril:
# Description of Changes

First commit reverts disabling adding enum variants (https://github.com/clockworklabs/SpacetimeDB/pull/3178).
Second commit makes replay recognize inserts / deletes to `st_column` and triggers a refresh of the in-memory table that was referenced in the `st_column` change.

None

3, small code bug "deep".

# Testing

I've confirmed manually that a module that couldn't be restarted before this PR can now be restarted.
We should probably follow up this PR with a smoketest.

### DIFF
--- a/crates/datastore/src/locking_tx_datastore/committed_state.rs
+++ b/crates/datastore/src/locking_tx_datastore/committed_state.rs
@@ -10,7 +10,7 @@ use crate::{
     db_metrics::DB_METRICS,
     error::{DatastoreError, IndexError, TableError},
     execution_context::ExecutionContext,
-    locking_tx_datastore::state_view::read_st_column_for_table,
+    locking_tx_datastore::state_view::iter_st_column_for_table,
     system_tables::{
         system_tables, StColumnRow, StConstraintData, StConstraintRow, StIndexRow, StSequenceRow, StTableFields,
         StTableRow, SystemTable, ST_CLIENT_ID, ST_CLIENT_IDX, ST_COLUMN_ID, ST_COLUMN_IDX, ST_COLUMN_NAME,
@@ -29,7 +29,7 @@ use spacetimedb_lib::{
     db::auth::{StAccess, StTableType},
     Identity,
 };
-use spacetimedb_primitives::{ColList, ColSet, IndexId, TableId};
+use spacetimedb_primitives::{ColId, ColList, ColSet, IndexId, TableId};
 use spacetimedb_sats::{algebraic_value::de::ValueDeserializer, memory_usage::MemoryUsage, Deserialize};
 use spacetimedb_sats::{AlgebraicValue, ProductValue};
 use spacetimedb_schema::{def::IndexAlgorithm, schema::TableSchema};
@@ -341,13 +341,6 @@ impl CommittedState {
             .map_err(TableError::Bflatn)?
             .ok_or_else(|| anyhow!("Delete for non-existent row when replaying transaction"))?;
 
-        if table_id == ST_COLUMN_ID {
-            // We've made a modification to `st_column`.
-            // The type of a table has changed, so figure out which.
-            // The first column in `StColumnRow` is `table_id`.
-            self.st_column_changed(rel)?;
-        }
-
         Ok(())
     }
 
@@ -358,7 +351,7 @@ impl CommittedState {
         row: &ProductValue,
     ) -> Result<()> {
         let (table, blob_store, pool) = self.get_table_and_blob_store_or_create(table_id, schema);
-        table.insert(pool, blob_store, row).map_err(|e| -> DatastoreError {
+        let (_, row_ref) = table.insert(pool, blob_store, row).map_err(|e| -> DatastoreError {
             match e {
                 InsertError::Bflatn(e) => TableError::Bflatn(e).into(),
                 InsertError::Duplicate(e) => TableError::Duplicate(e).into(),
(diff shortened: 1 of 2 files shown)


crates/core/src/db/update.rs
```diff
@@ -222,8 +222,14 @@ fn auto_migrate_database(
...
-            spacetimedb_schema::auto_migrate::AutoMigrateStep::ChangeColumns(_table_name) => {
-                anyhow::bail!("Unsupported: Changing column types");
+            spacetimedb_schema::auto_migrate::AutoMigrateStep::ChangeColumns(table_name) => {
+                let table_def = plan.new.stored_in_table_def(table_name).unwrap();
+                let table_id = stdb.table_id_from_name_mut(tx, table_name).unwrap().unwrap();
+                let column_schemas = column_schemas_from_defs(plan.new, &table_def.columns, table_id);
+
+                log!(logger, "Changing columns of table `{}`", table_name);
+
+                stdb.alter_table_row_type(tx, table_id, column_schemas)?;
```
Shubham8287, line 232: Well, I thought one part of the problem was to not also write to `st_columns` from `alter_table_row_type`. I don't see addressed that in PR, wondering how does this PR works.

Shubham8287: Oh okay, addressed in separate PR - https://github.com/clockworklabs/SpacetimeDB/pull/3196.

crates/datastore/src/locking_tx_datastore/committed_state.rs
```diff
@@ -349,11 +358,49 @@ impl CommittedState {
...
+            .expect("first field in `st_column` should decode to a `TableId`");
+
+        // We're replaying and we don't have unique constraints yet.
+        // Due to replay handling all inserts first and deletes after,
+        // when processing `st_column` insert/deletes,
+        // we may end up with duplicate definitions for the same `col_pos`.
+        // We're interested in the last definition, as it is mot recent.
+        // To obtain that, we must first reverse, dedup, and then reverse again.
+        // A smarter implementation would avoid the double reversal,
+        // but this is not hot code, so we don't bother optimizing.
```
gefjon:
```suggestion
        // We're replaying and we don't have unique constraints yet.
        // Due to replay handling all inserts first and deletes after,
        // when processing `st_column` insert/deletes,
        // we may end up with duplicate definitions for the same `col_pos`.
        // We're interested in the last definition, as it is most recent.
        // To obtain that, we must first reverse, dedup, and then reverse again.
        // `read_st_column_for_table` returns the columns in sorted order,
        // with newer versions of a column after older versions,
        // and that `Vec::dedup_by_key` deletes all but the first from consecutive runs.
        // A smarter implementation would avoid the double reversal,
        // but this is not hot code, so we don't bother optimizing.
```

crates/datastore/src/locking_tx_datastore/state_view.rs
```diff
@@ -149,6 +142,22 @@ pub trait StateView {
...
+) -> Result<Vec<ColumnSchema>> {
+    // Look up the columns for the table in question.
+    let mut columns: Vec<ColumnSchema> = this
+        .iter_by_col_eq(ST_COLUMN_ID, StColumnFields::TableId, &table_id.into())?
+        .map(|row| {
+            let row = StColumnRow::try_from(row)?;
+            Ok(row.into())
+        })
+        .collect::<Result<Vec<_>>>()?;
+    columns.sort_by_key(|col| col.col_pos);
```
gefjon: Not obvious to me how this function preserves the invariant required by `st_column_changed`, i.e. returning columns with the same `col_pos` in order older followed by newer. `sort_by_key` is a stable sort (which you should comment about if you're relying on that behavior, 'cause in many languages unstable sorting is the default), and `iter_by_col_eq` returns rows in physical storage order when no index exists. But I don't think we can rely on physical storage order being older followed by newer in all cases. It will be the first time you do this, 'cause `st_column` will have previously only had inserts and thus be dense, but the second time there will have been a delete, so `st_column` will have a free space, and so the newer column may be before the older column.

gefjon: Also, if we do determine that I am wrong and this function does have the required behavior, it should have a doc comment stating that invariant. I suspect that, if the invariant does hold, it only holds during replay, i.e. before instantiating indexes. This is fine because the indexes will enforce uniqueness anyways. The doc comment should capture this.

crates/datastore/src/locking_tx_datastore/committed_state.rs
gefjon: On second thought, I don't think this works; see comment on definition of `read_st_column_for_table`. However, I believe that:

- For the column we are currently processing:
  - There can only be two rows with the same `(target_table_id, col_pos)` tuple.
  - One of those rows (either the older or the newer, depending on when this function is called) will be equal to `row`.
- For columns we have already processed (when there are multiple columns changed for the same table):
  - We've already visited the new row, and could have recorded which one it was in some transient side-table.
  - During the delete path, there will only be one version (the newer version) in `st_column`, 'cause we'll have already deleted the older version.
- For columns we have not processed yet:
  - It doesn't matter which of the two types/rows we choose, 'cause we'll visit it at some point in the future anyways.
  - During the insert path, there will only be one version (the older version) in `st_column`, 'cause we won't have inserted the new version yet.

As such, I think it should be relatively easy to rework this so as to not rely on the iteration ordering of `st_column` rows at all.

crates/datastore/src/locking_tx_datastore/committed_state.rs
```diff
@@ -339,6 +340,14 @@ impl CommittedState {
             .delete_equal_row(&self.page_pool, blob_store, rel)
             .map_err(TableError::Bflatn)?
             .ok_or_else(|| anyhow!("Delete for non-existent row when replaying transaction"))?;
+
+        if table_id == ST_COLUMN_ID {
+            // We've made a modification to `st_column`.
+            // The type of a table has changed, so figure out which.
+            // The first column in `StColumnRow` is `table_id`.
+            self.st_column_changed(rel)?;
+        }
```
gefjon: Why do we need to do this in the delete path at all? Inserts are processed first (right?), and we only support altering columns, not wholesale deleting them, so by the time we hit this we must have already done the update during `replay_insert`.

smoketests/tests/auto_migration.py
```diff
@@ -39,11 +43,47 @@ class AddTableAutoMigration(Smoketest):
...
+    Student,
+}
+
+fn kind_from_string(_: String) -> PersonKind {
+    Student
+}
+
+fn kind_to_string(Student: PersonKind) -> &'static str {
+    "Student"
+}
```
gefjon, line 60: This functionality clearly deviates from what this test was originally written to test and what it claims to test based on its name. As such, I would prefer that you move this to a new smoketest (likely in the same file). `MODULE_CODE_INIT` is also the wrong name for what you're using it for; it  should be `MODULE_CODE_BASE` or similar.

crates/datastore/src/locking_tx_datastore/committed_state.rs
Centril: > As such, I think it should be relatively easy to rework this so as to not rely on the iteration ordering of `st_column` rows at all.

Yep, turns out it was. The method now simply keeps the newly inserted `RowPointer` for that `ColId`.

crates/datastore/src/locking_tx_datastore/state_view.rs
Centril: > It will be the first time you do this, 'cause `st_column` will have previously only had inserts and thus be dense, but the second time there will have been a delete, so `st_column` will have a free space, and so the newer column may be before the older column.

Yep, forgot about the free-cell bit. You're right. I've fixed the algo to not rely on this.

crates/datastore/src/locking_tx_datastore/committed_state.rs
Centril: We didn't actually, it was more of a "just in case", but I've removed that now.

smoketests/tests/auto_migration.py
Centril: I agree, but in the interest of time and BitCraft, I would prefer if we did this in a follow-up, along with a proper smoketest that actually checks the replay logic. I'll start working on that.

gefjon: Works for me, thanks.

gefjon (approved):

PR #3202, opened 2025-08-26
Improve error reporting for invalid column-type-changing automigrations

gefjon:
# Description of Changes

Title. When validating column type changes in `Table::change_columns_to` and comparing layouts in `RowTypeLayout::is_compatible_with`, return structured error objects which describe the mismatch.

If I cared more about this I use the `ErrorStream` combinator rather than short-circuiting on the first error, but :shrug:.

N/a

1.

# Testing

None yet.

- [ ] 
- [ ]

### DIFF
--- a/crates/sats/src/layout.rs
+++ b/crates/sats/src/layout.rs
@@ -189,11 +189,20 @@ impl AlgebraicTypeLayout {
     pub const String: Self = Self::VarLen(VarLenType::String);
 
     /// Can `self` be changed compatibly to `new`?
-    fn is_compatible_with(&self, new: &Self) -> bool {
+    fn is_compatible_with(&self, new: &Self) -> Result<(), IncompatibleTypeLayoutError> {
         match (self, new) {
             (Self::Sum(old), Self::Sum(new)) => old.is_compatible_with(new),
             (Self::Product(old), Self::Product(new)) => old.view().is_compatible_with(new.view()),
-            (Self::Primitive(old), Self::Primitive(new)) => old == new,
+            (Self::Primitive(old), Self::Primitive(new)) => {
+                if old == new {
+                    Ok(())
+                } else {
+                    Err(IncompatibleTypeLayoutError::DifferentPrimitiveTypes {
+                        old: old.algebraic_type(),
+                        new: new.algebraic_type(),
+                    })
+                }
+            }
             (Self::VarLen(VarLenType::Array(old)), Self::VarLen(VarLenType::Array(new))) => {
                 // NOTE(perf, centril): This might clone and heap allocate,
                 // but we don't care to avoid that and optimize right now,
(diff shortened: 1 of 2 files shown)


Centril:
So the structured errors already exist in `ensure_old_ty_upgradable_to_new` and they use `ErrorStream`.
The intent of the checks in `ensure_old_ty_upgradable_to_new` is as a soundness check, in case there's a bug in `auto_migrate_table`, rather than to be user facing. So I don't think `ErrorStream` is necessary there, but I do see some value in getting better errors directly from the datastore when debugging.

gefjon:
> So the structured errors already exist in `ensure_old_ty_upgradable_to_new` and they use `ErrorStream`. The intent of the checks in `ensure_old_ty_upgradable_to_new` is as a soundness check, in case there's a bug in `auto_migrate_table`, rather than to be user facing. So I don't think `ErrorStream` is necessary there, but I do see some value in getting better errors directly from the datastore when debugging.

In that case I'm happy to merge without worrying about using `ErrorStream` or getting too hung up about testing, though I do intend to leave the TODO comment in place.

Centril:
> [...] though I do intend to leave the TODO comment in place.

Not sure what the value of those TODOs are. I don't think they should be implemented, as all of the datastore is fail-fast and doesn't gather errors. If we want to change that I think we should do it as part of a larger plan.

gefjon:
> I don't think they should be implemented, as all of the datastore is fail-fast and doesn't gather errors.

Mm, fair. I'll remove.

crates/sats/src/layout.rs
```diff
@@ -309,13 +378,30 @@ impl HasLayout for ProductTypeLayoutView<'_> {
...
+        }
+        for (index, (old, new)) in self.elements.iter().zip(new.elements.iter()).enumerate() {
+            if let Err(err) = old.ty.is_compatible_with(&new.ty) {
+                return Err(Box::new(IncompatibleTypeLayoutError::IncompatibleProductElements {
+                    index,
+                    err,
+                }));
+            }
+        }
+        Ok(())
```
Centril:
```suggestion
        self.elements.iter().zip(new.elements.iter()).enumerate()
             .try_for_each(|(index, (old, new))| {
                 old.ty.is_compatible_with(&new.ty)
                      .map_err(|err| Box::new(IncompatibleTypeLayoutError::IncompatibleProductElements {
                    index,
                    err,
                }))
             })
```
Needs formatting, but slightly neater (also applicable to `SumTypeLayout::is_compatible_with`).

crates/sats/src/layout.rs
```diff
@@ -309,13 +378,30 @@ impl HasLayout for ProductTypeLayoutView<'_> {
...
-                .all(|(old, new)| old.ty.is_compatible_with(&new.ty))
+    ///
+    /// See comment on [`IncompatibleTypeLayoutError`] about [`Box`]ing the error return.
+    // TODO(error-reporting): Use `spacetimedb_data_structures::ErrorStream` to combine multiple errors
+    // rather than short-circuiting on the first.
+    // This is low priority because we've (at least theoretically) already passed through
+    // `spacetimedb_schema::auto_migrate::ensure_old_ty_upgradable_to_new` to get here,
+    // and that method has proper pretty error reporting with `ErrorStream`.
+    // The error here is for internal debugging.
+    fn is_compatible_with(self, new: Self) -> Result<(), Box<IncompatibleTypeLayoutError>> {
```
Centril: Now that this returns `Result`, consider renaming to `ensure_compatible_with` instead.
Bonus points for moving this to a trait definition. 😊

crates/sats/src/layout.rs
gefjon: I personally find the imperative style more readable than the functional style in this case. If you prefer the functional style, feel free to make the change and push to this branch, but I'm not gonna do it.

crates/sats/src/layout.rs
gefjon: We don't need polymorphism here, so I don't see any value in a trait. Agreed on the rename.

PR #3203, opened 2025-08-26
Sort columns in `st_column_changed` before automigrating.

gefjon:
# Description of Changes

In `st_column_changed`, `iter_st_column_for_table` returns rows in physical storage order. In simple cases (without unrelated deletes) this will be the same as insertion order, and therefore also the same as sorted order, but in cases with multiple column-changing automigrations the physical storage order of the rows diverges from the correct sorted order. Because this isn't a hot path, we can just call `.sort()` and not worry about it.

N/a

1

# Testing

None yet. Needs manual testing with BitCraft update.

### DIFF
--- a/crates/datastore/src/locking_tx_datastore/committed_state.rs
+++ b/crates/datastore/src/locking_tx_datastore/committed_state.rs
@@ -400,7 +400,7 @@ impl CommittedState {
         // Columns in `st_column` are not in general sorted by their `col_pos`,
         // though they will happen to be for tables which have never undergone migrations
         // because their initial insertion order matches their `col_pos` order.
-        columns.sort_by_key(|col: &ColumnSchema| col.col_pos.idx());
+        columns.sort_by_key(|col: &ColumnSchema| col.col_pos);
 
         // Update the columns and layout of the the in-memory table.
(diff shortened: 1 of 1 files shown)


crates/datastore/src/locking_tx_datastore/committed_state.rs
```diff
@@ -386,14 +389,19 @@ impl CommittedState {
...
                 StColumnRow::try_from(row_ref)
                     .map(|c| (c.col_pos != target_col_id || row_ref.pointer() == row_ptr).then(|| c.into()))
                     .transpose()
             })
             .collect::<Result<Vec<_>>>()?;
 
+        // Columns in `st_column` are not in general sorted by their `col_pos`,
+        // though they will happen to be for tables which have never undergone migrations
+        // because their initial insertion order matches their `col_pos` order.
+        columns.sort_by_key(|col: &ColumnSchema| col.col_pos.idx());
```
Shubham8287: nitpick:
```suggestion
        columns.sort_by_key(|col: &ColumnSchema| col.col_pos);
```

To be consistent with `schema_for_table_row`.

Centril:
@gefjon Can you drop the base-branch and re-target this at master? I think this can go before the error-reporting PR.

PR #3209, opened 2025-08-29
Dont use sentinel ids for system tables.

jsdt:
# Description of Changes

This simplifies system table bootstrapping logic by removing auto-incremented ids for indexes, constraints, and sequences in system tables. By defining them all up front, we avoid the need for `reset_system_table_schemas`, and make it easier to add new things to system tables.

This change should not change the initial system table rows written during bootstrapping, and the `load_1_2_*` tests in `relational_db.rs` should verify that by loading data written by a previous version with and without a snapshot. 

I also added some sequence-related changes. The sequence schema previous included the `allocated` field, which tracked where to start generating new values after a restart. This made some checks awkward, since the 'schema' was changing as rows were inserted. I moved this outside of the `schema`, so those should no longer change unless something actually changes the shape of the tables.

As part of that, I also changed where we begin generating on startup to be `allocated` instead of `allocated + 1`. The code that generates values always stopped early enough to allow the next restart to use `allocated`, so we were just being overly defensive. We also had some issues if sequences wrapped over, or if we had sequences with negative increments. Those were supported by the schema, but didn't really work with the generation/allocation code. That should have better testing now.

2. The bootstrapped system tables should match, since we have tests with older data. The changes to sequences include some more invariant checking, which has some risk of causing problems if previous versions wrote values that I didn't consider.

# Testing

This has existing tests covering bootstrapping and restoring from a snapshot.

### DIFF
--- a/crates/datastore/src/locking_tx_datastore/sequence.rs
+++ b/crates/datastore/src/locking_tx_datastore/sequence.rs
@@ -26,14 +26,6 @@ impl MemoryUsage for Sequence {
 
 impl Sequence {
     pub(super) fn new(schema: SequenceSchema, previous_allocation: Option<i128>) -> Self {
-        if let Some(prev) = previous_allocation {
-            if prev < schema.min_value || prev > schema.max_value {
-                panic!(
-                    "Invalid sequence: previous allocation value {prev} is out of bounds for sequence with min_value {} and max_value {}",
-                    schema.min_value, schema.max_value
-                );
-            }
-        }
         if schema.start < schema.min_value || schema.start > schema.max_value {
             panic!(
                 "Invalid sequence: start value {} is out of bounds for sequence with min_value {} and max_value {}",
@@ -51,7 +43,24 @@ impl Sequence {
                 "Invalid sequence: increment must be less than or equal to the range between min_value and max_value"
             );
         }
-        let start = previous_allocation.unwrap_or(schema.start);
+        let start = if let Some(prev) = previous_allocation {
+            if prev < schema.min_value || prev > schema.max_value {
+                // Previous versions set allocated to 0 as a default,
+                // so we have this special case.
+                if prev == 0 {
+                    schema.start
+                } else {
+                    panic!(
+                        "Invalid sequence: previous allocation value {prev} is out of bounds for sequence with min_value {} and max_value {}",
+                        schema.min_value, schema.max_value
+                    );
+                }
+            } else {
+                prev
+            }
+        } else {
+            schema.start
+        };
         // We will always need to allocate before generating any values.
         Self {
             value: start,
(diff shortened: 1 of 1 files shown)


crates/datastore/src/locking_tx_datastore/datastore.rs
```diff
@@ -1724,8 +1725,14 @@ mod tests {
             );
 
             assert_eq!(
-                schema.indexes,
+                schema
```
gefjon, line 1737: I think we should also assert that these match the values from `INDEX_IDS` and friends.

crates/datastore/src/locking_tx_datastore/datastore.rs
```diff
@@ -217,7 +217,8 @@ impl Locking {
         }
 
         // Fix up auto_inc IDs in the cached system table schemas.
-        committed_state.reset_system_table_schemas()?;
+        // We might need something to reset sequence ids?
```
gefjon: I think we should include an assertion here that the system table IDs match the ones we expect. If that assertion holds, then we don't need to `reset_system_table_schemas` AFAIK, and if it doesn't hold then this patch probably (?) breaks that snapshot.

gefjon: The call  to `reset_system_table_schemas` was previously necessary here for the same reason it was in the bootstrap path. When restoring from a snapshot, we don't respect the snapshot's system table schemas, instead we use the statically known schemas. (You can see this in action on lines 184 -- 190 of this file.) Prior to your change, said schemas were full of sentinel ID values, and so we had to reset them to use the updated IDs. With your change, the IDs should just be installed correctly already.

gefjon (commented):
This looks great. It's a simpler and less invasive change than I was afraid of, and meaningfully simplifies the bootstrap path.

Did any of the codegen change? Is it planned to? If not, please remove them from the diff.

This will need some amount of testing with existing commitlogs and snapshots from prior to this change. I'd be comfortable with manual testing, but ideally I'd like to just get a commitlog (starting at zero, doesn't need to be long) and a snapshot (with a short commitlog suffix), check them in to our repo, and write an automated test that we can spin them up in memory and then perform a transaction in the reconstructed state.

gefjon:
> although I'm unsure about checking in the artefacts (will we remember to update them if we change formats?).

I argue that we want to leave them in place indefinitely to ensure that we don't break old-but-supported commitlog formats. We'd have to remember to add new ones as well, I suppose, but I'm not terribly worried about that.

crates/datastore/src/locking_tx_datastore/datastore.rs
jsdt: I added a function which just asserts that the ids match, and I updated the comment.

PR #3210, opened 2025-08-29
fix: Drop table

Shubham8287:
# Description of Changes
Current implementation of `drop_table` have few problems :
1. Table rows are not deleted, which means `blob_store` bytes are never reclaimed.
2. As a result, the generated `TxData` does not include deleted rows, so the commit log replay logic never drops the in-memory commit table.

Changes:
Make `MutTx::drop_table` to not drop commit table but only delete rows (using `delete_tables`) inside it. Later, let `CommitState::merge` to detect dropped tables by anaylyzing `st_table` rows deletions and remove commit tables accordingly.

This change fixes above two problems. However, we still need an additional change to drop empty tables during replay (WIP).

3, small change but it is in hotpath. 

# Testing

- [x ] Modified current unit test to cover table drop case.
- [ ] Planning to write smoketest for testing replay.

### DIFF
--- a/crates/datastore/src/locking_tx_datastore/committed_state.rs
+++ b/crates/datastore/src/locking_tx_datastore/committed_state.rs
@@ -44,6 +44,7 @@ use spacetimedb_table::{
 };
 use std::collections::BTreeMap;
 use std::sync::Arc;
+use thin_vec::ThinVec;
 
 /// Contains the live, in-memory snapshot of a database. This structure
 /// is exposed in order to support tools wanting to process the commit
@@ -593,12 +594,23 @@ impl CommittedState {
     pub(super) fn merge(&mut self, tx_state: TxState, ctx: &ExecutionContext) -> TxData {
         let mut tx_data = TxData::default();
 
+        // This transaction may have dropped tables.
+        // After applying `merge_apply_deletes` and `merge_apply_inserts`,
+        // any tables no longer referenced by `st_table` should be
+        // removed from the committed state.
+        let mut tables_to_drop = ThinVec::<TableId>::new();
+
         // First, apply deletes. This will free up space in the committed tables.
-        self.merge_apply_deletes(&mut tx_data, tx_state.delete_tables);
+        self.merge_apply_deletes(&mut tx_data, &mut tables_to_drop, tx_state.delete_tables);
 
         // Then, apply inserts. This will re-fill the holes freed by deletions
         // before allocating new pages.
-        self.merge_apply_inserts(&mut tx_data, tx_state.insert_tables, tx_state.blob_store);
+        self.merge_apply_inserts(
+            &mut tx_data,
+            &mut tables_to_drop,
+            tx_state.insert_tables,
+            tx_state.blob_store,
+        );
 
         // If the TX will be logged, record its projected tx offset,
         // then increment the counter.
@@ -607,10 +619,19 @@ impl CommittedState {
             self.next_tx_offset += 1;
         }
 
+        for table_id in tables_to_drop {
+            self.tables.remove(&table_id);
+        }
+
         tx_data
     }
 
-    fn merge_apply_deletes(&mut self, tx_data: &mut TxData, delete_tables: BTreeMap<TableId, DeleteTable>) {
+    fn merge_apply_deletes(
+        &mut self,
+        tx_data: &mut TxData,
+        tables_to_drop: &mut ThinVec<TableId>,
+        delete_tables: BTreeMap<TableId, DeleteTable>,
+    ) {
         for (table_id, row_ptrs) in delete_tables {
             if let (Some(table), blob_store, _) = self.get_table_and_blob_store_mut(table_id) {
                 let mut deletes = Vec::with_capacity(row_ptrs.len());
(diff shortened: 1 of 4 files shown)


crates/datastore/src/locking_tx_datastore/committed_state.rs
```diff
@@ -593,12 +594,23 @@ impl CommittedState {
     pub(super) fn merge(&mut self, tx_state: TxState, ctx: &ExecutionContext) -> TxData {
         let mut tx_data = TxData::default();
 
+        // This transaction may have dropped tables.
+        // After applying `merge_apply_deletes` and `merge_apply_inserts`,
+        // any tables no longer referenced by `st_table` should be
+        // removed from the committed state.
+        let mut tables_to_drop = thin_vec::ThinVec::<TableId>::new();
```
gefjon: Why use `ThinVec` here?

crates/datastore/src/locking_tx_datastore/committed_state.rs
```diff
@@ -593,12 +594,23 @@ impl CommittedState {
...
+
         // First, apply deletes. This will free up space in the committed tables.
-        self.merge_apply_deletes(&mut tx_data, tx_state.delete_tables);
+        self.merge_apply_deletes(&mut tx_data, &mut tables_to_drop, tx_state.delete_tables);
```
gefjon, line 604:
```suggestion
        // Any rows deleted from `st_table` will be noted in `tables_to_drop`.
        self.merge_apply_deletes(&mut tx_data, &mut tables_to_drop, tx_state.delete_tables);
```

I don't actually know if this is true (this is the first part of my review), it's just my assumption about what's going on here.

crates/datastore/src/locking_tx_datastore/committed_state.rs
```diff
@@ -593,12 +594,23 @@ impl CommittedState {
...
 
         // Then, apply inserts. This will re-fill the holes freed by deletions
         // before allocating new pages.
-        self.merge_apply_inserts(&mut tx_data, tx_state.insert_tables, tx_state.blob_store);
+        self.merge_apply_inserts(
```
gefjon, line 608:
```suggestion
        // Any rows inserted into `st_table` will be removed from `tables_to_drop`.
        // This allows us to avoid incorrectly dropping tables whose `st_table` rows have been "updated,"
        // i.e. replaced with a new version.
        self.merge_apply_inserts(
```

I don't actually know if this is true (this is the first part of my review), it's just my assumption about what's going on here.

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -352,15 +352,14 @@ impl MutTxId {
...
+
         // Delete the table and its rows and indexes from memory.
         self.tx_state.insert_tables.remove(&table_id);
-        self.tx_state.delete_tables.remove(&table_id);
-        let commit_table = self
-            .committed_state_write_lock
-            .tables
-            .remove(&table_id)
-            .expect("there should be a schema in the committed state if we reach here");
-        self.push_schema_change(PendingSchemaChange::TableRemoved(table_id, commit_table));
```
gefjon, line 363: If you're removing this code, you need to update the comment about how the commit table isn't eagerly deleted, but what code does delete it, and how queries later in the transaction avoid reading from it.

Shubham8287:
> * Require that the table be empty at the time of `MutTx::drop_table`.
> * Delete all the rows during `MutTx::drop_table`, before making the change to the in-memory structures.

fundamental problem with removing in-memory `Table` in `MutTx::drop_table` is that, `CommittedState::merge` expects Commit Table to be present to actually delete rows.

> It seems simpler to keep the `PendingSchemaChange`  machinery for handling `drop_table`.

I feel `PendingSchemaChange`  is great for reversing changes that doesn't go in commitlog, but transaction semantic already take care to rollback/commit table rows and responsible for writting commitlog.

gefjon:
> fundamental problem with removing in-memory Table in MutTx::drop_table is that, CommittedState::merge expects Commit Table to be present to actually delete rows.

This doesn't sound like a fundamental problem. In fact it sounds very solvable. You could either:

- Assert that a dropped table's delete table is empty, and fail the merge if it is not. This conflicts with my suggestion to require that all the rows be explicitly deleted prior to the `MutTx::drop_table` call; instead you would have `MutTx::drop_table` also clear the delete table.
- Add the set of dropped tables as an additional argument to `CommittedState::merge`, and fall back to checking there if the committed state does not contain the table for a given delete.

Shubham8287:
> * Add the set of dropped tables as an additional argument to `CommittedState::merge`, and fall back to checking there if the committed state does not contain the table for a given delete.

Yes, I think we can do this to make it work, but I can't say how is this better than not deleting comit table during `MutTx::drop_table ` and anaylyzing `st_table` changes during `merge` to decide on dropped tables.

crates/datastore/src/locking_tx_datastore/committed_state.rs
Centril: I'm guessing the idea is to keep the footprint small as it will be empty most of the time. However, as we never put this into the heap, it doesn't matter that we use 16 bytes more on the stack, so a regular `Vec` seems better.

Shubham8287:
Closing in favor of - https://github.com/clockworklabs/SpacetimeDB/pull/3214

PR #3211, opened 2025-08-29
Only add information about the cli version to the globals files.

jsdt:
# Description of Changes

This only affects comments in generated files.

To reduce noise from git diffs, we will only add a comment with the CLI version to the file globals, rather than adding them to the comments of every file.

1. The only risk here is around tests based on this diff.


jdetter:
Closed in favor of https://github.com/clockworklabs/SpacetimeDB/pull/3216

PR #3212, opened 2025-08-29
V8: Add `call_call_reducer` incl with stack traces + wire update_database

Centril:
# Description of Changes

For V8 modules:

- Add `call_call_reducer`
- Add stack traces for traps.
- Wire `update_database`

It remains to wire up `call_call_reducer`, but this shouldn't be that much work, but I'll do that in a follow up.

None

1

# Testing

Tests for `call_call_reducer` are added.


coolreader18 (approved):
LGTM!

PR #3214, opened 2025-09-01
datastore: add `clear_table` and fix `drop_table`

Centril:
# Description of Changes

Aternative to and closes https://github.com/clockworklabs/SpacetimeDB/pull/3210.
This version relies on `pending_schema_changes`.
The first commit adds `clear_table` to the datastore that's efficient and can be exposed to the module ABI in a follow up.
The second commit fixes `drop_table`.

None

3?

# Testing

`test_drop_table_is_transactional` is amended to check `TxData`.


crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -1865,6 +1869,26 @@ impl MutTxId {
...
+        // Get the commit table.
+        let (commit_table, commit_bs, ..) = self.committed_state_write_lock.get_table_and_blob_store(table_id)?;
+
+        // Get the insert table and delete all rows from it.
+        let (tx_table, tx_blob_store, delete_table) = self
+            .tx_state
+            .get_table_and_blob_store_or_create_from(table_id, commit_table);
+        let mut rows_removed = tx_table.clear(tx_blob_store);
+
+        // Mark every table in the committed state as deleted.
```
Shubham8287:
```suggestion
        // Mark every row in the committed state as deleted.
```

Shubham8287 (approved):
Looks Good, thanks!
Considering there will be a follow-up PR to fix this for commitlog replay.
This PR will make commitlog replay logic to empty the drop tables but we still need a patch to actually drop them from memory.

crates/table/src/table.rs
```diff
@@ -1369,6 +1380,13 @@ impl Table {
         }
     }
 
+    /// Returns a list of all present row pointers.
+    pub fn scan_all_row_ptrs(&self) -> Vec<RowPointer> {
+        let mut ptrs = Vec::with_capacity(self.row_count as usize);
+        ptrs.extend(self.scan_rows(&NullBlobStore).map(|row| row.pointer()));
```
Shubham8287, line 1386: Just noticed this - is it okay to use `NullBlobStore` for scanning? I don’t fully understand this, but I assume this works because we’re only collecting the `RowPointer`s and not performing any operations on the `BlobStore`

Centril: > but I assume this works because we’re only collecting the `RowPointer`s and not performing any operations on the `BlobStore`

Yes exactly, the blob store is never actually used.

crates/datastore/src/locking_tx_datastore/datastore.rs
```diff
@@ -715,17 +715,16 @@ impl TxMetrics {
...
+                    committed_state.get_table(table_id).and_then(|table| {
+                        table_stats.insert(
+                            table_id,
+                            TableStats {
+                                row_count: table.row_count,
+                                bytes_occupied_overestimate: table.bytes_occupied_overestimate(),
+                                num_indices: table.num_indices(),
+                            },
+                        )
+                    });
```
gefjon: This reminds me: do we need to clean up any metrics when a table gets deleted? If there are any metrics that have a table ID in the `with_label_values`, we'll want to call `remove_label_values` on them at some point, probably within commit.

gefjon (commented):
Holding off on approving until we get the replay parts of this merged and tested, but I like this.

crates/datastore/src/locking_tx_datastore/datastore.rs
Centril: Added some logic to `TxMetrics::report` to fix this.

Centril:
> Holding off on approving until we get the replay parts of this merged and tested, but I like this.

I've fixed the replay parts, but we cannot integration test this until the add-column PR as simply dropping a table is not allowed.

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -1865,6 +1869,26 @@ impl MutTxId {
...
+    // Clears the table for `table_id`, removing all rows.
+    pub fn clear_table(&mut self, table_id: TableId) -> Result<usize> {
+        // Get the commit table.
+        let (commit_table, commit_bs, ..) = self.committed_state_write_lock.get_table_and_blob_store(table_id)?;
+
+        // Get the insert table and delete all rows from it.
+        let (tx_table, tx_blob_store, delete_table) = self
+            .tx_state
+            .get_table_and_blob_store_or_create_from(table_id, commit_table);
+        let mut rows_removed = tx_table.clear(tx_blob_store);
```
Shubham8287, line 1931: For `tx_table`, shouldn’t it be enough to just remove it from the ` BTreeMap` of `TxState::insert_tables`?

Centril: If it wasn't for the blob store you could do that. We could as an optimization check whether the table layout is fixed, in which case there cannot be any blob store references, making the removal sound.

Shubham8287: I think, I don’t fully understand how we’re using `blob_store`, but for `TxState::blob_store`, won’t `MutTxId::commit` eventually drop it anyway?

Do you think is it a problem to occupy extra bytes in `blob_store` during the lifecycle of tx (until it commits or rollback)?

Shubham8287: Created a ticket to track this -  https://github.com/clockworklabs/SpacetimeDB/issues/3271

PR #3215, opened 2025-09-02
datastore: add support for truncation

Centril:
# Description of Changes

Adds support for truncation in the datastore.
The commitlog format already supported it.
A table is considered truncated in a transaction when there are deletes and when the row count post-deletes is 0.

None

3?

# Testing

Some existing datastore tests have been amended to check whether `TxData` contain truncation or not.


crates/datastore/src/locking_tx_datastore/committed_state.rs
```diff
@@ -643,17 +644,18 @@ impl CommittedState {
 
             if !deletes.is_empty() {
                 let table_name = &*table.get_schema().table_name;
-                // TODO(centril): Pass this along to record truncated tables.
-                let _truncated = table.row_count == 0;
-                tx_data.set_deletes_for_table(table_id, table_name, deletes.into());
+                let truncated = table.row_count == 0;
```
Shubham8287, line 680: Should we rather let caller of `fn delete_rows` to pass `truncated` as param and maybe just `assert` on row counts here?(pass it true if table is from `PendingSchemaChanges::Removed`).

Reason is: this code considers empty table also as truncated table for generating `TxData`. 
Which may not be the problem during replay due to `get_table_and_blob_store_or_create` will create it again but still not seems great.

Centril: > Reason is: this code considers empty table also as truncated table for generating `TxData`. Which may not be the problem during replay due to `get_table_and_blob_store_or_create` will create it again but still not seems great.

An emptied table is I think the meaning of truncation, see e.g., https://learn.microsoft.com/en-us/sql/t-sql/statements/truncate-table-transact-sql?view=sql-server-ver17

So you can think of `truncate` as `clear_table`.

Shubham8287: From the link, 
> TRUNCATE TABLE removes all rows from a table, but the table structure and its columns, constraints, indexes, and so on, remain. To remove the table definition in addition to its data, use the DROP TABLE statement.

Makes sense.

crates/datastore/src/traits.rs
```diff
@@ -186,6 +188,24 @@ pub struct TxData {
...
+}
+
+/// Whether a table was truncated in a transaction.
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Debug)]
+pub enum TxTableTruncated {
+    /// It was truncated.
+    Yes,
+    /// It was not truncated.
+    No,
+}
```
gefjon: Storing the rows for a truncated table is surprising to me. Why is that necessary?

Centril: Subscriptions are still interested in the specific rows that were deleted, and currently both the commitlog and subscriptions use `TxData`.

gefjon: Any TX which drops a table (as part of a migration) will necessarily break all subscribed clients anyways, no?

Centril: Yes, however, TRUNCATE (e.g., in SQL) is the same as `clear_table`. The flag is set whenever a table is emptied, so it applies to situations where clients are still connected as well.

crates/datastore/src/locking_tx_datastore/committed_state.rs
```diff
@@ -345,19 +345,35 @@ impl CommittedState {
         Ok(())
     }
 
+    pub(super) fn replay_truncate(&mut self, table_id: TableId) -> Result<()> {
```
Shubham8287, line 318: `visit_delete` happens before `visit_truncate`, which removes rows from `st_*` tables, hence `schema_for_table` fails inside `visit_truncate`. I think we have to avoid calling that method.

Shubham8287:
I am also getting this error while trying to replay from commitlog after module update.
```
2025-09-17T09:22:34.516339Z ERROR try_init: crates/core/src/host/module_host.rs:396: failed to create table message

Caused by:
    0: DatastoreError: IndexError: Unique constraint violation 'st_table_table_name_idx_btree' in table 'st_table': column(s): '["table_name"]' value: "message"
    1: IndexError: Unique constraint violation 'st_table_table_name_idx_btree' in table 'st_table': column(s): '["table_name"]' value: "message"
    2: Unique constraint violation 'st_table_table_name_idx_btree' in table 'st_table': column(s): '["table_name"]' value: "message"    
```
With following repo steps:
1. publish quickstart module
2. Make any change in module to change its ModuleDef, ex- remove or add a reducer.
3. republish module and restart the server.
4. make a sql call to actually load the module after restart.

Not reproducible in master, I also tried with rebasing the PR on master.

crates/datastore/src/locking_tx_datastore/committed_state.rs
Shubham8287: So the problem is, table is temporarily empty and rows might be inserted back during `merge_apply_inserts`. So, we can not mark it truncated before checking that. I will push the fix for it.

PR #3218, opened 2025-09-03
First pass on documentation for auth claims

jdetter:
# Description of Changes

This PR adds documentation for accessing auth claims in modules.

This is a docs change.

0

# Testing

- [ ] All examples compile and have been tested working

### DIFF
--- a/docs/docs/authentication/auth-claims-in-module.md
+++ b/docs/docs/authentication/auth-claims-in-module.md
@@ -1,0 +1,309 @@
+### Documentation TODO
+
+- Update C# examples when it is implemented
+- Update the rust examples one last time when auth claims are merged
+
+I asked Jeff:
+
+Can auth claims change after a user has already authenticated? Like if I connect a client which authenticates via SpacetimeAuth but then after they connect they change their phone number or address or something else that would be in their auth claims - when will the module see that change?
+
+^ Update documentation for the response to this
+
+VVV Documentation starts below VVV
+
+# Accessing Auth Claims in Modules
+
+SpacetimeDB allows you to easily access authentication (auth) claims embedded in OIDC-compliant JWT tokens. Auth claims are key-value pairs that provide information about the authenticated user. For example, they may contain a user's unique ID, email, or authentication provider. If you want to view these fields for yourself, you can inspect the contents of any JWT using online tools like [jwt.io](https://jwt.io/).
+
+In a SpacetimeDB module, auth claims from a client's token are accessible via the `ReducerContext` which is passed to all reducers. The following examples show how to access and use these claims in your module code.
+
+## Accessing Common Claims: Subject and Issuer
+
+The subject (`sub`) and issuer (`iss`) are the most commonly accessed claims in a JWT. The subject usually represents the user's unique identifier, while the issuer indicates which authentication provider issued the token.
+
+Below are examples of how to access these claims in both Rust and C# modules:
+
+:::server-rust
+
+```rust
+#[reducer(client_connected)]
+pub fn connect(ctx: &ReducerContext) -> Result<(), String> {
+    let auth_ctx = ctx.sender_auth();
+    let (subject, issuer) = match auth_ctx.jwt() {
+        Some(claims) => (claims.subject().to_string(), claims.issuer().to_string()),
+        None => {
+            return Err("Client connected without JWT".to_string());
+        }
+    };
+    log::info!("sub: {}, iss: {}", subject, issuer);
+    Ok(())
+}
+```
+
+*Example output when using a Google-issued token:*
+```
+INFO: src\lib.rs:64: sub: 321321321321321, iss: https://accounts.google.com
+```
+:::
+:::server-csharp
+
+```cs
+// ************************TODO: update + test this after Jeff implements this in C# ************************
+[Reducer(ReducerKind.ClientConnected)]
+public void Connect(ReducerContext ctx) {
+    var auth_ctx = ctx.SenderAuth();
+    var (subject, issuer) = auth_ctx.Jwt() switch {
+        Some(var claims) => (claims.Subject, claims.Issuer),
+        None => throw new Exception("Client connected without JWT"),
+    };
+    log.Info($"sub: {subject}, iss: {issuer}");
+}
+```
+
+*Example output when using a Google-issued token:*
+```
+INFO: src\Lib.cs:64: sub: 321321321321321, iss: https://accounts.google.com
+```
+:::
+
+## Accessing Custom Claims
+
+If you want to access additional claims that are not parsed by default by SpacetimeDB, you can parse the raw JWT payload yourself. This is useful for handling custom or application-specific claims.
+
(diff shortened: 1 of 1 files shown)


docs/docs/authentication/auth-claims-in-module.md
```diff
@@ -0,0 +1,309 @@
+### Documentation TODO
+
+- Update C# examples when it is implemented
+- Update the rust examples one last time when auth claims are merged
+
+I asked Jeff:
+
+Can auth claims change after a user has already authenticated? Like if I connect a client which authenticates via SpacetimeAuth but then after they connect they change their phone number or address or something else that would be in their auth claims - when will the module see that change?
```
jsdt, line 8: The short answer here is not right now.

The longer answer is that we don't have anything in the protocol to let users update their credentials, other than disconnecting and reconnecting with a different token. If we add that (which we probably should do for token expiry reasons), then we could make this work, though it would still be pretty awkward. If the user changed a field like that in spacetime auth or google/github/etc, the client code would need to be aware of that and 1. get a new id token with the updated fields; then 2. send the new token to the server.

docs/docs/authentication/auth-claims-in-module.md
```diff
@@ -0,0 +1,309 @@
...
+## Example: Restricting Accepted Issuers
+
+Since users can use any valid token to connect to SpacetimeDB, their token may originate from any authentication provider. For example, they could send an OIDC compliant token from Github even though you only want to accept tokens from Google. If you want to only allow users to authenticate using a specific provider (e.g., Google), you can check the issuer claim `iss` as shown below:
+
+:::server-rust
+
+```rust
+#[reducer(client_connected)]
+pub fn connect(ctx: &ReducerContext) -> Result<(), String> {
+    let auth_ctx = ctx.sender_auth();
```
jsdt, line 151: It's generally better to avoid using unwrap, so it's more obvious that it can't panic.
```suggestion
    let jwt = ctx.sender_auth()
        .jwt()
        .ok_or_else(|| "Client connected without JWT".to_string())?;
```

If you think having those branches is more readable, you can also use a match:
```suggestion
    let jwt = match ctx.sender_auth().jwt() {
        Some(jwt) => jwt,
        None => return Err("Client connected without JWT".to_string()),
    };
    let jwt = match ctx.sender_auth()
        .jwt()
        .ok_or_else(|| "Client connected without JWT".to_string())?;
```

docs/docs/authentication/auth-claims-in-module.md
```diff
@@ -0,0 +1,309 @@
...
+
+
+#[reducer(client_connected)]
+pub fn connect(ctx: &ReducerContext) -> Result<(), String> {
+    let auth_ctx = ctx.sender_auth();
+    let payload = auth_ctx.jwt().unwrap().raw_payload();
+    let claims: CustomClaims = serde_json::from_slice(payload.as_bytes()).map_err(|e| format!("Client connected with invalid JWT: {}", e).to_string())?;
+
+    // In this example you would really want to check the issuer of the JWT to ensure it is from a trusted provider
+    if auth_ctx.jwt().unwrap().issuer() != "https://accounts.google.com" {
```
jsdt, line 233: You could use `claims.iss` here instead of `auth_ctx.jwt().unwrap().issuer()`.

docs/docs/authentication/auth-claims-in-module.md
```diff
@@ -0,0 +1,309 @@
...
+    pub sub: String,
+    pub iat: u64,
+    pub email: String,
+}
+
+#[table(name = user)]
+pub struct User {
+    #[primary_key]
+    #[auto_inc]
+    pub id: u64,
```
jsdt, line 207: Using these `auto_inc` user ids in examples feels like it clashes with the `identity` that we are using in so many parts of our APIs.

docs/docs/authentication/auth-claims-in-module.md
```diff
@@ -0,0 +1,309 @@
...
+    }
+}
+```
+:::
+
+> **Important:** If you return an error from the connect reducer, the client will be disconnected immediately.
+
+## Example: automatically give users a role based on their auth claims
+
+When users first connect to your SpacetimeDB module you may want to automatically give them some role. For example, at clockwork we might want to give users with a valid "clockworklabs.io" email an admin role. We can do this by checking the email claim in the connect reducer and giving them the admin role if they have a valid email.
```
jsdt, line 188: I like the example of roles. I think it could use useful to add an example of using this in a reducer to decide whether a client is allowed to do something, since these examples are only covering the initial connection right now.

docs/docs/authentication/auth-claims-in-module.md
```diff
@@ -0,0 +1,309 @@
...
+}
+
+#[derive(SpacetimeType)]
+pub enum Role {
+    User,
+    Admin,
+}
+
+#[table(name = user_role)]
+pub struct UserRole {
```
jsdt, line 219: I'm not sure how useful these tables would be, since the reducers that would be enforcing any security rules would also have access to the email on the auth token. Having a table with `email` and `role` would make more sense to me.

Since the code in this example is computing the role from the email, we don't actually need to store any of the roles in tables. I think the main reason to have a user roles table would be to be able to set roles for different users via SQL or a different admin-only reducer.

docs/docs/authentication/auth-claims-in-module.md
```diff
@@ -0,0 +1,309 @@
...
+#[reducer(client_connected)]
+pub fn connect(ctx: &ReducerContext) -> Result<(), String> {
+    let auth_ctx = ctx.sender_auth();
+    if auth_ctx.jwt().is_none() {
+        return Err("Client connected without JWT".to_string());
+    }
+
+    let jwt = auth_ctx.jwt().unwrap();
+    // Example: We only accept google auth
+    if jwt.issuer() != "https://accounts.google.com" {
```
jsdt, line 158: One weird aspect of this is that people using maincloud have auth from `auth.spacetimedb.com`, so this would mean that someone logged in with the creds needed to publish would not be able to use `spacetime sql`.

jsdt (commented):
I took a look at the rust examples, and my main thoughts are that:
 1. We should never use `unwrap`, and
 2. We should probably have some example of a reducer other than the `ClientConnected` one.

docs/docs/authentication/auth-claims-in-module.md
jdetter:
```rust
#[table(name = admin)]
pub struct Admin {
    #[unique]
    pub ident: Identity,
}

// Add to on_connect
#[reducer(init)]
pub fn init(ctx: &ReducerContext) -> Result<(), String> {
    ctx.db.admin().insert(Admin{
        ident: ctx.sender
    });
    Ok(())
}

// Is the sender an owner of the db?
if ctx.db.admin().ident().find(ctx.sender).is_some() {
  return Ok(());
}
// Then check for google auth

// Note: check the audience here as well
```

PR #3223, opened 2025-09-03
Added the Unreal SDK work for codegen, testing, and the plugin

### LINKED ISSUE #3219: Add Unreal Support
Add the branched code to new structure for the new Unreal SDK
- Move from /crates/sdk-unreal to /crates/sdk
- Update and test all features

JasonAtClockwork:
# Description of Changes

Closes #3219 
This adds the Unreal SDK, the new Unreal test cases, updates the test runner to handle Unreal, codegen updates for Unreal, and a QuickStart Chat.

No breaking changes.

2 - This impacts the subcommand generate.rs to include unrealcpp and crates/testing to expand for Unreal

# Testing

- [x] Run the new Unreal tests 
- [x] Run any previous automation testing - with all the changes to generate/testing I'm uncertain if there is an impact
- [x] Review the new CLI generate documentation changes


JasonAtClockwork:
@bfops  - With cargo test --all and this new Unreal set of tests the Test Suites will keep failing. I've updated all the Unreal tests to start with unreal_ to allow for --skip unreal but I'm not sure we should change the Test Suites to --skip unreal?

bfops:
> @bfops - With cargo test --all and this new Unreal set of tests the Test Suites will keep failing. I've updated all the Unreal tests to start with unreal_ to allow for --skip unreal but I'm not sure we should change the Test Suites to --skip unreal?

Looks like you've already done it but yes I think we should add `--skip-unreal` to all the test suites that currently use `--all`

crates/cli/src/subcommands/generate.rs
```diff
@@ -25,7 +25,7 @@ use std::io::Read;
 pub fn cli() -> clap::Command {
     clap::Command::new("generate")
         .about("Generate client files for a spacetime module.")
-        .override_usage("spacetime generate --lang <LANG> --out-dir <DIR> [--project-path <DIR> | --bin-path <PATH>]")
+        .override_usage("spacetime generate --lang <LANG> --out-dir <DIR> [--project-path <DIR> | --bin-path <PATH> | --module-name <MODULE_NAME> | --uproject-dir <DIR>]")
```
bfops, line 28: nit: I don't think this usage statement is correct? it looks like `--module-name` and `--uproject-dir` are both required for unreal cpp, not mutually exclusive.

crates/cli/src/subcommands/generate.rs
```diff
@@ -86,6 +101,11 @@ pub fn cli() -> clap::Command {
         )
         .arg(common_args::yes())
         .after_help("Run `spacetime help publish` for more detailed information.")
+        .group(
+            clap::ArgGroup::new("output_dir")
+                .args(["out_dir", "uproject_dir"])
+                .required(true)
+        )
```
bfops, line 108: nit: is this clause needed? it looks like the individual arguments already specify when they're required (and together they specify all languages).

crates/cli/src/subcommands/generate.rs
```diff
@@ -57,17 +57,32 @@ pub fn cli() -> clap::Command {
...
             Arg::new("out_dir")
                 .value_parser(clap::value_parser!(PathBuf))
-                .required(true)
                 .long("out-dir")
                 .short('o')
-                .help("The system path (absolute or relative) to the generate output directory"),
+                .help("The system path (absolute or relative) to the generate output directory")
+                .required_if_eq("lang", "rust")
+                .required_if_eq("lang", "csharp")
+                .required_if_eq("lang", "typescript"),
```
bfops, line 65: nit: maybe we want `required_unless_eq`? we have a clause below that says we require either this arg or the unreal project arg, so maybe this should just be `required_unless_eq("lang", "unrealcpp")`

crates/cli/src/subcommands/generate.rs
```diff
@@ -156,9 +189,15 @@ pub async fn exec_ex(
         paths.insert(path);
     }
 
+    // For Unreal, we want to clean up just the module directory, not the entire uproject directory tree.
+    let cleanup_root = match lang {
+        Language::UnrealCpp => out_dir.join("Source").join(module_name.as_ref().unwrap()),
```
bfops, line 194: nit: it's a little strange to me that we manually append `Source` here, but not in the block just above where we're generating files. If `Source` is already baked in there, should we maybe remove it so that the "knowledge" about that directory is kind of localized?

crates/cli/src/subcommands/generate.rs
```diff
@@ -101,16 +121,21 @@ pub async fn exec_ex(
...
     let build_options = args.get_one::<String>("build_options").unwrap();
 
     if args.value_source("namespace") == Some(ValueSource::CommandLine) && lang != Language::Csharp {
         return Err(anyhow::anyhow!("--namespace is only supported with --lang csharp"));
     }
 
+    let out_dir = args
+        .get_one::<PathBuf>("out_dir")
+        .or_else(|| args.get_one::<PathBuf>("uproject_dir"))
+        .unwrap();
```
bfops, line 137: nit: given that we somewhat treat the directory differently below based on where it came from, I'm somewhat inclined to not "merge" them here and have a "if unreal X otherwise Y" type branch below, rather than risking accidentally using `out_dir` in a way that it's not meant to be used.

bfops (approved):
my code-owned file (`crates/cli/src/subcommands/generate.rs`) LGTM, but left several optional nits.

crates/cli/src/subcommands/generate.rs
JasonAtClockwork: Yeah this has actually become a little ugly with the Unreal changes, I considered adding a line for each supported LANG but it's the same for the others and only Unreal is the odd one out. Though maybe something like?
```rs
.override_usage(
            "USAGE:
  spacetime generate --lang <LANG: rust, csharp, typescript> --out-dir <DIR> [--project-path <DIR> | --bin-path <PATH>]
  spacetime generate --lang unreal-cpp --module-name <NAME>  --uproject-dir <DIR> [--project-path <DIR> | --bin-path <PATH>]"
        )
```

crates/cli/src/subcommands/generate.rs
JasonAtClockwork: I think this was set up by the Arvikasoft team to only allow either --out-dir or --uproject-dir but not both. Is there maybe a better way to define this?

crates/cli/src/subcommands/generate.rs
JasonAtClockwork: The problem here is how the codegen is handling pathing with Unreal's folder structure. The Arvikasoft team found pointing to the root of the project folder made traversal better. The problem is how the generate deletes files starting from the out_dir. 

I think I need to do a refactor this in the near future as I'm uncertain if this will work for multiple modules in one Unreal project with this structure.

crates/cli/src/subcommands/generate.rs
JasonAtClockwork: @gefjon - I'd love your thoughts on this one too

bfops: I think that looks good

crates/cli/src/subcommands/generate.rs
bfops: well further up they're already restricted to a mutually exclusive set of languages, so I think this is redundant

crates/cli/src/subcommands/generate.rs
bfops: yeah I'm fine with needing `uproject_dir` to point to something other than the output dir. but if all the files are going to be generated under `uproject_dir/Sources` (which is what I'm inferring from the cleanup block here) then this would be more clearly correct if the block above also used `out_dir.join("Source")` (and then the code elsewhere didn't prepend `Source`. Not a big deal either way, especially if we're refactoring this soon.

crates/cli/src/subcommands/generate.rs
gefjon: Use `|` for alternatives, not `,`, but otherwise I like this. `<LANG: rust | csharp | typescript>`.

crates/cli/src/subcommands/generate.rs
JasonAtClockwork: Just tested this one locally and removing that group allows the use of both --out-dir **and** --uproject-dir which will then fallback to --out-dir causing the code generation to ignore --uproject-dir. I think I'll need to come back at this with the refactor as this is just funky and your other comments could make this a lot more clear.

Added to [Refactor generate.rs Arvikasoft changes](https://github.com/clockworklabs/SpacetimeDBPrivate/issues/2050) to address in the near future.

crates/cli/src/subcommands/generate.rs
JasonAtClockwork: I didn't find a required_unless_eq so I might reach out on this one in the near future, in the mean time I've added to [Refactor generate.rs Arvikasoft changes](https://github.com/clockworklabs/SpacetimeDBPrivate/issues/2050).

crates/cli/src/subcommands/generate.rs
JasonAtClockwork: Added to [Refactor generate.rs Arvikasoft changes](https://github.com/clockworklabs/SpacetimeDBPrivate/issues/2050) to address in the near future.

crates/cli/src/subcommands/generate.rs
JasonAtClockwork: Added to [Refactor generate.rs Arvikasoft changes](https://github.com/clockworklabs/SpacetimeDBPrivate/issues/2050) to address in the near future.

crates/cli/src/subcommands/generate.rs
JasonAtClockwork: Added to [Refactor generate.rs Arvikasoft changes](https://github.com/clockworklabs/SpacetimeDBPrivate/issues/2050) to address in the near future.

jdetter (approved):
I'm reviewing the docs for this now. I think this is ready to go in for now and we can fix issues in follow-up PRs as this PR is already quite large and hard to review.

PR #3227, opened 2025-09-04
Fix the smoke test so it take the server URL from the config.toml

mamcx:
# Description of Changes

Closes https://github.com/clockworklabs/SpacetimeDBPrivate/issues/1998#event-19488677475

Now the smoke test parse the `config.toml` for the correct address for connecting.

1
# Testing

- [x] Run the smoke both on public / private


bfops (approved):
LGTM! Thank you for fixing this

PR #3230, opened 2025-09-05
datastore: add columns support

Shubham8287:
# Description of Changes
- `add_columns_to_table` api to handle `AutoMigrateStep::AddColumns` -- Look at doc comment for more Info.

depends on: https://github.com/clockworklabs/SpacetimeDB/pull/3261

TODO: handle `AutoMigrateStep::DisconnectAllUsers`.

N/A.

2? Changes are not in hotpath but a bug in migration can corrupt tables.

# Testing
a test to exercise the API.

### DIFF
--- a/crates/cli/src/subcommands/logs.rs
+++ b/crates/cli/src/subcommands/logs.rs
@@ -61,6 +61,12 @@ pub enum LogLevel {
     Panic,
 }
 
+/// Sentinel value used for injected system logs.
+///
+/// Keep this in sync with the constants in `spacetimedb_core::database_logger::Record`.
+const SENTINEL: &str = "__spacetimedb__";
+
+/// Keep this in sync with `spacetimedb_core::database_logger::Record`.
 #[serde_with::serde_as]
 #[derive(serde::Deserialize)]
 struct Record<'a> {
@@ -73,6 +79,8 @@ struct Record<'a> {
     #[serde(borrow)]
     filename: Option<Cow<'a, str>>,
     line_number: Option<u32>,
+    #[serde(borrow)]
+    function: Option<Cow<'a, str>>,
     #[serde(borrow)]
     message: Cow<'a, str>,
     trace: Option<Vec<BacktraceFrame<'a>>>,
@@ -195,16 +203,36 @@ pub async fn exec(mut config: Config, args: &ArgMatches) -> Result<(), anyhow::E
         out.set_color(&color)?;
         write!(out, "{level:>5}: ")?;
         out.reset()?;
+        let mut need_space_before_filename = false;
+        let mut need_colon_sep = false;
         let dimmed = ColorSpec::new().set_dimmed(true).clone();
-        if let Some(filename) = record.filename {
-            out.set_color(&dimmed)?;
-            write!(out, "{filename}")?;
-            if let Some(line) = record.line_number {
-                write!(out, ":{line}")?;
+        if let Some(function) = record.function {
+            if function != SENTINEL {
+                out.set_color(&dimmed)?;
+                write!(out, "{function}")?;
+                out.reset()?;
+                need_space_before_filename = true;
+                need_colon_sep = true;
             }
-            out.reset()?;
         }
-        writeln!(out, ": {}", record.message)?;
+        if let Some(filename) = record.filename {
+            if filename != SENTINEL {
+                out.set_color(&dimmed)?;
+                if need_space_before_filename {
+                    write!(out, " ")?;
+                }
+                write!(out, "{filename}")?;
+                if let Some(line) = record.line_number {
+                    write!(out, ":{line}")?;
+                }
+                out.reset()?;
+                need_colon_sep = true;
+            }
+        }
+        if need_colon_sep {
+            write!(out, ": ")?;
+        }
+        writeln!(out, "{}", record.message)?;
         if let Some(trace) = &record.trace {
             for frame in trace {
                 write!(out, "    in ")?;
--- a/crates/core/src/database_logger.rs
+++ b/crates/core/src/database_logger.rs
@@ -49,10 +49,50 @@ impl From<u8> for LogLevel {
 pub struct Record<'a> {
     #[serde_as(as = "serde_with::TimestampMicroSeconds")]
     pub ts: chrono::DateTime<Utc>,
+    /// Target of the log call (usually source namespace or `mod`).
+    ///
+    /// Provided by the WASM guest as an argument to the `console_log` host function.
+    ///
+    /// The special sentinel value [`Record::SENTINEL_INJECTED_TARGET`]` denotes logs injected by the [`SystemLogger`].
     pub target: Option<&'a str>,
+    /// Filename of the source location of the log call.
+    ///
+    /// Provided by the WASM guest as an argument to the `console_log` host function.
+    ///
+    /// The special sentinel value [`Record::SENTINEL_INJECTED_FILENAME`]` denotes logs injected by the [`SystemLogger`].
     pub filename: Option<&'a str>,
     pub line_number: Option<u32>,
+    /// Which exported function (i.e. reducer) was being called when this message was produced.
+    ///
+    /// Unlike `target`, `filename` and `line_number`, this is not provided by the WASM guest.
+    /// Instead, the `WasmInstanceEnv` remembers what function call is in progress and adds it to the record.
+    ///
+    /// The special sentinel value [`Record::SENTINEL_INJECTED_FUNCTION`] denotes logs injected by the [`SystemLogger`].
+    pub function: Option<&'a str>,
     pub message: &'a str,
+}
+
+impl<'a> Record<'a> {
+    pub const SENTINEL_INJECTED_FUNCTION: Option<&'static str> = Some("__spacetimedb__");
+    pub const SENTINEL_INJECTED_TARGET: Option<&'static str> = Some("__spacetimedb__");
+    pub const SENTINEL_INJECTED_FILENAME: Option<&'static str> = Some("__spacetimedb__");
+
+    /// Create a log `Record` for a system message, not attributed to any reducer or user filename.
+    ///
+    /// The resulting `Record` will draw from [`chrono::Utc::now`] for its timestamp,
+    /// have `line_number: None`,
+    /// and will use [`Self::SENTINEL_INJECTED_FILENAME`], [`Self::SENTINEL_INJECTED_FUNCTION`]
+    /// and [`Self::SENTINEL_INJECTED_TARGET`].
+    pub fn injected(message: &'a str) -> Self {
+        Record {
+            ts: chrono::Utc::now(),
+            target: Self::SENTINEL_INJECTED_TARGET,
+            filename: Self::SENTINEL_INJECTED_FILENAME,
+            line_number: None,
+            function: Self::SENTINEL_INJECTED_FUNCTION,
+            message,
+        }
+    }
 }
 
 pub trait BacktraceProvider {
@@ -277,12 +317,6 @@ impl SystemLogger {
     }
 
     fn record(message: &str) -> Record {
-        Record {
-            ts: Utc::now(),
-            target: None,
-            filename: Some("spacetimedb"),
-            line_number: None,
-            message,
-        }
+        Record::injected(message)
     }
 }
--- a/crates/core/src/host/module_host.rs
+++ b/crates/core/src/host/module_host.rs
@@ -1021,7 +1021,7 @@ impl ModuleHost {
             _ => None,
         };
         if let Some(log_message) = log_message {
-            self.inject_logs(LogLevel::Error, &log_message)
+            self.inject_logs(LogLevel::Error, reducer_name, &log_message)
         }
 
         res
@@ -1110,15 +1110,12 @@ impl ModuleHost {
         tokio::join!(self.module.scheduler().closed(), self.job_tx.closed());
     }
 
-    pub fn inject_logs(&self, log_level: LogLevel, message: &str) {
+    pub fn inject_logs(&self, log_level: LogLevel, reducer_name: &str, message: &str) {
         self.replica_ctx().logger.write(
             log_level,
             &Record {
-                ts: chrono::Utc::now(),
-                target: None,
-                filename: Some("external"),
-                line_number: None,
-                message,
+                function: Some(reducer_name),
+                ..Record::injected(message)
             },
             &(),
         )
--- a/crates/core/src/host/wasm_common/module_host_actor.rs
+++ b/crates/core/src/host/wasm_common/module_host_actor.rs
@@ -584,12 +584,12 @@ fn maybe_log_long_running_reducer(reducer_name: &str, total_duration: Duration) 
 
 /// Logs an error `message` for `reducer` at `timestamp` into `replica_ctx`.
 fn log_reducer_error(replica_ctx: &ReplicaContext, timestamp: Timestamp, reducer: &str, message: &str) {
-    let record = database_logger::Record {
+    use database_logger::Record;
+
+    let record = Record {
         ts: chrono::DateTime::from_timestamp_micros(timestamp.to_micros_since_unix_epoch()).unwrap(),
-        target: Some(reducer),
-        filename: None,
-        line_number: None,
-        message,
+        function: Some(reducer),
+        ..Record::injected(message)
     };
     replica_ctx.logger.write(database_logger::LogLevel::Error, &record, &());
 }
--- a/crates/core/src/host/wasmtime/wasm_instance_env.rs
+++ b/crates/core/src/host/wasmtime/wasm_instance_env.rs
@@ -164,6 +164,13 @@ impl WasmInstanceEnv {
         &self.reducer_name
     }
 
+    /// Returns the name of the most recent reducer to be run in this environment,
+    /// or `None` if no reducer is actively being invoked.
+    fn log_record_function(&self) -> Option<&str> {
+        let function = self.reducer_name();
+        (!function.is_empty()).then_some(function)
+    }
+
     /// Returns the name of the most recent reducer to be run in this environment.
     pub fn reducer_start(&self) -> Instant {
         self.reducer_start
@@ -1145,12 +1152,15 @@ impl WasmInstanceEnv {
             // The line number cannot be `u32::MAX` as this represents `Option::None`.
             let line_number = (line_number != u32::MAX).then_some(line_number);
 
+            let function = env.log_record_function();
+
             let record = Record {
                 // TODO: figure out whether to use walltime now or logical reducer now (env.reducer_start)
                 ts: chrono::Utc::now(),
                 target: target.as_deref(),
                 filename: filename.as_deref(),
                 line_number,
+                function,
                 message: &message,
             };
 
@@ -1191,12 +1201,14 @@ impl WasmInstanceEnv {
 
             let elapsed = span.start.elapsed();
             let message = format!("Timing span {:?}: {:?}", &span.name, elapsed);
+            let function = caller.data().log_record_function();
 
             let record = Record {
                 ts: chrono::Utc::now(),
                 target: None,
                 filename: None,
                 line_number: None,
+                function,
                 message: &message,
             };
             caller.data().instance_env.console_log(
--- a/modules/quickstart-chat/src/lib.rs
+++ b/modules/quickstart-chat/src/lib.rs
@@ -27,6 +27,7 @@ fn validate_name(name: String) -> Result<String, String> {
 pub fn set_name(ctx: &ReducerContext, name: String) -> Result<(), String> {
     let name = validate_name(name)?;
     if let Some(user) = ctx.db.user().identity().find(ctx.sender) {
+        log::info!("User {} sets name to {name}", ctx.sender);
         ctx.db.user().identity().update(User {
             name: Some(name),
             ..user
@@ -51,6 +52,7 @@ pub fn send_message(ctx: &ReducerContext, text: String) -> Result<(), String> {
     // - Rate-limit messages per-user.
     // - Reject messages from unnamed user.
     let text = validate_message(text)?;
+    log::info!("User {}: {text}", ctx.sender);
     ctx.db.message().insert(Message {
         sender: ctx.sender,
         text,


crates/table/src/table.rs
```diff
@@ -329,60 +357,110 @@ impl Table {
...
+                reason,
+            })
+        };
+
+        if new_columns.len() < existing_columns.len() {
+            return Err(make_err(AddColumnsErrorReason::MissingExistingColumns));
+        }
+
+        // Check if existing columns remain compatible
+        let old_cols_in_new_schema = &new_columns.get(..existing_columns.len()).expect("length checked above");
```
Centril: Combine these two checks?
```suggestion
        // Ensure we have at least as many (prefix) as the existing columns.
        let Some(old_cols_in_new_schema) = &new_columns.get(..existing_columns.len()) else {
            return Err(make_err(AddColumnsErrorReason::MissingExistingColumns));
        }

        // Ensure that the existing prefix is compatible with the new prefix.
```

crates/table/src/table.rs
```diff
@@ -329,60 +357,110 @@ impl Table {
...
+        Ok(())
+    }
+
+    /// Validates that the proposed `new_columns` schema is compatible with the
+    /// existing table schema and that all newly added columns are initialized
+    /// with default values.
+    pub fn validate_add_columns_schema(
+        &self,
+        new_columns: &[ColumnSchema],
+        default_values: &IntMap<ColId, AlgebraicValue>,
```
Centril:
```suggestion
        default_values: &[AlgebraicValue],
```
This could be passed as a slice as `ColId`s are sequential. You could remove the prefix from the `ColId`s.

crates/table/src/table.rs
```diff
@@ -329,60 +357,110 @@ impl Table {
...
+        let new_row_layout: RowTypeLayout = columns_to_row_type(old_cols_in_new_schema).into();
+        self.validate_row_type_layout(&new_row_layout, new_columns)
+            .map_err(|e| make_err(e.reason.into()))?;
+
+        // Validate that existing columns remain unchanged in order
+        for (existing_column, new_column) in existing_columns.iter().zip(new_columns.iter()) {
+            if existing_column.col_name != new_column.col_name {
+                return Err(make_err(AddColumnsErrorReason::MissingExistingColumns));
+            }
         }
```
Centril: The `Table` doesn't actually care abut the names of the columns. It's not required for "soundness", which is why the method above doesn't do this.

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -493,6 +495,115 @@ impl MutTxId {
...
+            .validate_add_columns_schema(&column_schemas, &default_values)
+            .map_err(TableError::from)?;
+
+        // Intentionally using `schema_for_table_raw` instead of `schema_for_table`.
+        //
+        // Rationale:
+        // - `schema_for_table_raw` queries system tables directly, ensuring the
+        //   most up-to-date schema.
+        // - `schema_for_table` relies on an in-memory cache, which may not reflect
+        //   the latest sequence updates and can therefore return stale schemas.
```
Centril: If any of these points in the rationale are correct, there is a bug in the datastore.
Any operation which alters a table will immediately change the schema.

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -493,6 +495,115 @@ impl MutTxId {
...
+        let original_table_schema = self.schema_for_table_raw(table_id)?;
+
+        // Default values must be for a sequence of contiguous column IDs,
+        // beginning immediately after the last column ID in the existing schema.
+        // This has been checked inside `Table::validate_add_columns_schema`
+        let default_values: Vec<AlgebraicValue> = default_values
+            .into_iter()
+            .sorted_by_key(|(col_id, _)| *col_id)
+            .map(|(_, val)| val)
+            .collect();
```
Centril: As aforementioned, I think you can pass `Vec<AlgebraicValue>` to this method directly instead.

crates/table/src/table.rs
```diff
@@ -329,60 +357,110 @@ impl Table {
...
 
-        unsafe { self.change_columns_to_unchecked(column_schemas, validate) }
+        // Validate that all new columns have default values
```
Centril: This check is insuffcient; it doesn't ensure that the types of the defaults are compatible with the added column types.

crates/table/src/table.rs
```diff
@@ -329,60 +357,110 @@ impl Table {
...
+            return Err(make_err(AddColumnsErrorReason::MissingExistingColumns));
+        }
+
+        // Check if existing columns remain compatible
+        let old_cols_in_new_schema = &new_columns.get(..existing_columns.len()).expect("length checked above");
+        let new_row_layout: RowTypeLayout = columns_to_row_type(old_cols_in_new_schema).into();
+        self.validate_row_type_layout(&new_row_layout, new_columns)
+            .map_err(|e| make_err(e.reason.into()))?;
+
+        // Validate that existing columns remain unchanged in order
```
Shubham8287: I think this check is redundant, due to above `validate_row_type_layout` check.

crates/table/src/table.rs
```diff
@@ -329,60 +357,110 @@ impl Table {
         &mut self,
         column_schemas: Vec<ColumnSchema>,
     ) -> Result<Vec<ColumnSchema>, Box<ChangeColumnsError>> {
-        /// Validate that the old row type layout can be changed to the new.
-        // Intentionally fail fast rather than combining errors with [`spacetimedb_data_structures::error_stream`]
-        // because we've (at least theoretically) already passed through
-        // `spacetimedb_schema::auto_migrate::ensure_old_ty_upgradable_to_new` to get here,
-        // and that method has proper pretty error reporting with `ErrorStream`.
-        // The error here is for internal debugging.
-        fn validate(
```
Shubham8287, line 338: This has been moved out as `validate_row_type_layout` to be shared with `validate_add_columns_schema`

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -493,6 +495,115 @@ impl MutTxId {
...
+
+        log::debug!(
+            "TABLE ALTERING (incompatible layout): {}, table_id: {}",
+            original_table_schema.table_name,
+            table_id
+        );
+
+        // Drop existing table first due to unique constraints on table name in `st_table`
+        self.drop_table(table_id)?;
```
Centril, line 577: You are currently not considering `self.tx_state.insert_tables` and `self.tx_state.delete_tables`. I think you'll want to return those from `drop_table` and then use them to a) exclude rows in the committed state, b) include inserted rows in the tx state.

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -493,6 +495,115 @@ impl MutTxId {
...
+        // beginning immediately after the last column ID in the existing schema.
+        // This has been checked inside `Table::validate_add_columns_schema`
+        let default_values: Vec<AlgebraicValue> = default_values
+            .into_iter()
+            .sorted_by_key(|(col_id, _)| *col_id)
+            .map(|(_, val)| val)
+            .collect();
+
+        log::debug!(
+            "TABLE ALTERING (incompatible layout): {}, table_id: {}",
```
Centril:
```suggestion
            "ADDING TABLE COLUMN (incompatible layout): {}, table_id: {}",
```

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -493,6 +495,115 @@ impl MutTxId {
...
+            new_table.insert(&self.committed_state_write_lock.page_pool, tx_blob_store, &migrated_row)?;
+        }
+
+        self.push_schema_change(PendingSchemaChange::TableRemoved(table_id, original_commit_table));
+        self.push_schema_change(PendingSchemaChange::TableAdded(new_table_id));
+
+        Ok(new_table_id)
+    }
+
+    fn create_table_and_bump_seq(&mut self, table_schema: TableSchema) -> Result<TableId> {
```
Centril: A doc comment on this would be neat

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -493,6 +495,115 @@ impl MutTxId {
...
+        let table_id = self.create_table(table_schema.clone())?;
+        let table_schema = self.schema_for_table(table_id)?;
+
+        // Bump each sequence value so that the next generated value requires
+        // a fresh allocation.
+        for seq_id in table_schema.sequences.iter().map(|s| s.sequence_id) {
+            let seq = self
+                .sequence_state_lock
+                .get_sequence_mut(seq_id)
+                .expect("sequence just created");
```
Centril: Consider using `get_sequence_mut`.

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -493,6 +495,115 @@ impl MutTxId {
...
+            seq.value = seq.allocated() + 1;
```
Centril: Also, rather than doing this, why not copy the exact value from `sequence_state_lock` before the table is dropped?

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -493,6 +495,115 @@ impl MutTxId {
...
+            .tx_state
+            .get_table_and_blob_store(new_table_id)
+            .ok_or(TableError::IdNotFoundState(new_table_id))?;
+
+        for migrated_row in migrated_rows {
+            new_table.insert(&self.committed_state_write_lock.page_pool, tx_blob_store, &migrated_row)?;
+        }
+
+        self.push_schema_change(PendingSchemaChange::TableRemoved(table_id, original_commit_table));
+        self.push_schema_change(PendingSchemaChange::TableAdded(new_table_id));
```
Centril: This already happened in `.create_table_internal(...)`

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -493,6 +495,115 @@ impl MutTxId {
...
+        let (new_table, tx_blob_store) = self
+            .tx_state
+            .get_table_and_blob_store(new_table_id)
+            .ok_or(TableError::IdNotFoundState(new_table_id))?;
+
+        for migrated_row in migrated_rows {
+            new_table.insert(&self.committed_state_write_lock.page_pool, tx_blob_store, &migrated_row)?;
+        }
+
+        self.push_schema_change(PendingSchemaChange::TableRemoved(table_id, original_commit_table));
```
Centril: And this needs to happen before `TableAdded` is added. I think you might need to manipulate the list of pending schema changes such that you insert the deletion into the right index.

crates/datastore/src/locking_tx_datastore/datastore.rs
```diff
@@ -3320,8 +3332,138 @@ mod tests {
...
+        );
+
+        let table_id = datastore.create_table_mut_tx(&mut tx, schema)?;
+
+        let columns_original = tx.get_schema(table_id).unwrap().columns.to_vec();
+
+        // Insert initial rows
+        let initial_row = ProductValue {
+            elements: [AlgebraicValue::U64(0), AlgebraicValue::sum(0, AlgebraicValue::U16(1))].into(),
+        };
```
Centril: You can use `product![...]` and use e.g., `0u64` and `1u16`.

crates/datastore/src/locking_tx_datastore/datastore.rs
```diff
@@ -3320,8 +3332,138 @@ mod tests {
...
+        tx.insert_product_value(table_id, &initial_row).unwrap();
+        tx.insert_product_value(table_id, &initial_row).unwrap();
```
Centril: For tests, there's `insert(datastore, &mut tx, table_id, &initial_row).unwrap()`

crates/datastore/src/locking_tx_datastore/datastore.rs
```diff
@@ -3320,8 +3332,138 @@ mod tests {
...
+
+        commit(&datastore, tx)?;
+
+        // Alter Table: Add Variant and Column
```
Centril, line 3516: Where's the variant added? It would be good to test that.

crates/datastore/src/locking_tx_datastore/datastore.rs
```diff
@@ -3320,8 +3332,138 @@ mod tests {
...
+
+        //  Validate Commitlog Changes
+        let (_, deletes) = tx_data
+            .deletes()
+            .find(|(id, _)| **id == table_id)
+            .expect("Expected delete log for original table");
+
+        let deleted_rows = [
+            product![AlgebraicValue::U64(5), AlgebraicValue::sum(0, AlgebraicValue::U16(1))],
+            product![AlgebraicValue::U64(6), AlgebraicValue::sum(0, AlgebraicValue::U16(1))],
```
```suggestion
            product![5u64, AlgebraicValue::sum(0, 1u16)],
            product![6u64, AlgebraicValue::sum(0, 1u16)],
```

crates/datastore/src/locking_tx_datastore/datastore.rs
```diff
@@ -3320,8 +3332,138 @@ mod tests {
...
+        let inserted_rows = [
+            product![
+                AlgebraicValue::U64(5),
+                AlgebraicValue::sum(0, AlgebraicValue::U16(1)),
+                AlgebraicValue::U8(42)
+            ],
+            product![
+                AlgebraicValue::U64(6),
+                AlgebraicValue::sum(0, AlgebraicValue::U16(1)),
+                AlgebraicValue::U8(42)
```
```suggestion
                5u64,
                AlgebraicValue::sum(0, 1u16),
                42u8
            ],
            product![
                6u64,
                AlgebraicValue::sum(0, 1u16),
                42u8
```

crates/datastore/src/locking_tx_datastore/datastore.rs
```diff
@@ -3320,8 +3332,138 @@ mod tests {
...
+        let mut tx = begin_mut_tx(&datastore);
+
+        let new_row = ProductValue {
+            elements: [
+                AlgebraicValue::U64(0),
+                AlgebraicValue::sum(0, AlgebraicValue::U16(1)),
+                AlgebraicValue::U8(0),
+            ]
+            .into(),
+        };
```
```suggestion
        let new_row = product![0u64, AlgebraicValue::sum(0, 1u16), 0u8];
```

crates/datastore/src/locking_tx_datastore/datastore.rs
```diff
@@ -3320,8 +3332,138 @@ mod tests {
 
         Ok(())
     }
-
     // TODO: Add the following tests
     // - Create a tx that inserts 2000 rows with an auto_inc column
     // - Create a tx that inserts 2000 rows with an auto_inc column and then rolls back
+
+    #[test]
+    fn test_add_columns_to_table() -> ResultTest<()> {
```
Centril, line 3473: It would be good to have some transactionality tests for rollback.

crates/datastore/src/locking_tx_datastore/mut_tx.rs
Shubham8287: I have written a test to show this, I think fix should be easy and not-blocking for this PR - `test_table_schemas_consistency`

crates/datastore/src/locking_tx_datastore/mut_tx.rs
Shubham8287: I changed the approach to rely on `iter` to fetch rows into `Vec`, doing it in start of the function, Lmk what you think about it.

crates/datastore/src/locking_tx_datastore/datastore.rs
```diff
@@ -3372,23 +3371,29 @@ mod tests {
...
+        vars.push(SumTypeVariant::new_named(AlgebraicType::U8, "bb"));
+        *vars_ref = vars.into();
+        // Add column `c`
         let mut new_columns = columns_original.clone();
         new_columns.push(ColumnSchema::for_test(2, "c", AlgebraicType::U8));
-        let defaults = IntMap::from_iter([(2.into(), AlgebraicValue::U8(42))]);
+        let defaults = vec![AlgebraicValue::U8(42)];
 
         let mut tx = begin_mut_tx(&datastore);
+        // insert a row in tx_state when before adding column
```
Centril: When before?

crates/datastore/src/locking_tx_datastore/datastore.rs
```diff
@@ -3466,4 +3456,58 @@ mod tests {
 
         Ok(())
     }
+
+    #[test]
+    fn test_table_schemas_consistency() -> ResultTest<()> {
```
Centril: Could you add some comments to the test, for example explaining what it tests, why 5000 is used, etc?

crates/datastore/src/locking_tx_datastore/datastore.rs
```diff
@@ -3466,4 +3456,58 @@ mod tests {
...
+        for _ in 0..5000 {
+            insert(&datastore, &mut tx, table_id, &initial_row).unwrap();
+        }
+        commit(&datastore, tx)?;
+
+        let tx = begin_tx(&datastore);
+        let table_schema = tx.schema_for_table(table_id).unwrap();
+        let table_schema_raw = tx.schema_for_table_raw(table_id).unwrap();
+
+        //TODO: Fix the bug and update the assert
```
Centril: Can you file an issue for this and assign it to me?

crates/datastore/src/locking_tx_datastore/mut_tx.rs
Centril: Sure, looks simpler. We can optimize later.

crates/table/src/table.rs
```diff
@@ -430,36 +433,33 @@ impl Table {
...
-        // Validate that existing columns remain unchanged in order
-        for (existing_column, new_column) in existing_columns.iter().zip(new_columns.iter()) {
-            if existing_column.col_name != new_column.col_name {
-                return Err(make_err(AddColumnsErrorReason::MissingExistingColumns));
+        // Validate that all new columns have default values and thier types match.
+        for (idx, new_col) in new_columns.iter().skip(existing_columns.len()).enumerate() {
+            let default_value = default_values
+                .get(idx)
+                .ok_or_else(|| make_err(AddColumnsErrorReason::DefaultValueMissing(new_col.col_pos)))?;
+            if Some(new_col.col_type.clone()) != default_value.type_of() {
```
Centril: This is a sound check but its unnecessarily incomplete. Rather than using `type_of`, it would be better to introduce `AlgebraicType::type_check` to check if the value is compatible with the type. That would be a complete check.

crates/datastore/src/locking_tx_datastore/datastore.rs
Shubham8287: `allocate` feild has been removed from `SequenceSchema` (which  was inconsistent across `schema_for_table_raw` and `schema_for_table`) and I am not relying on it anymore.
I deleted the test.

crates/table/src/table.rs
Shubham8287: I introduced `AlgebraicType::type_check`, lmk if it is as you intended. changes from last review are in last commit of the PR - \https://github.com/clockworklabs/SpacetimeDB/pull/3230/commits/7ae4f0fe76f57aafd291c04faf0271486a49706d

crates/sats/src/algebraic_type.rs
```diff
@@ -445,8 +445,42 @@ impl AlgebraicType {
             _ => true,
         }
     }
-}
 
+    pub fn type_check(&self, value: &AlgebraicValue) -> bool {
+        match value {
+            AlgebraicValue::Min | AlgebraicValue::Max => return false,
```
Shubham8287: I do not know the usecase of these type, Hence can't say we should return true or false for them?

crates/table/src/table.rs
```diff
@@ -300,6 +300,35 @@ pub enum ChangeColumnsErrorReason {
...
+    old: Vec<ColumnSchema>,
+    new: Vec<ColumnSchema>,
+    default_values: Vec<AlgebraicValue>,
+    reason: AddColumnsErrorReason,
+}
+
+#[derive(Error, Debug)]
+pub enum AddColumnsErrorReason {
+    #[error("Default value type does not match column type for column {0}")]
+    DefaultValueTypeNotMatch(ColId),
```
Centril:
```suggestion
    DefaultValueTypeMismatch(ColId),
```

crates/table/src/table.rs
```diff
@@ -329,60 +358,109 @@ impl Table {
...
+        self.validate_row_type_layout(&new_row_layout, new_columns)
+            .map_err(|e| make_err(e.reason.into()))?;
+
+        // Validate that all new columns have default values and thier types match.
+        for (idx, new_col) in new_columns.iter().skip(existing_columns.len()).enumerate() {
+            let default_value = default_values
+                .get(idx)
+                .ok_or_else(|| make_err(AddColumnsErrorReason::DefaultValueMissing(new_col.col_pos)))?;
+            if !new_col.col_type.type_check(default_value) {
+                return Err(make_err(AddColumnsErrorReason::DefaultValueTypeNotMatch(
```
Centril:
```suggestion
                return Err(make_err(AddColumnsErrorReason::DefaultValueTypeMismatch(
```

crates/table/src/table.rs
```diff
@@ -329,60 +358,109 @@ impl Table {
...
 
-        unsafe { self.change_columns_to_unchecked(column_schemas, validate) }
+        Ok(())
+    }
+
+    /// Validates that the proposed `new_columns` schema is compatible with the
+    /// existing table schema and that all newly added columns are initialized
+    /// with default values.
+    ///`new_columns`: columns schema after adding new columns ordered by `ColId`s.
+    ///`default_values`: default values for newly added columns ordered by `ColId`s.
```
Centril:
```suggestion
    /// - `new_columns`: columns schema after adding new columns ordered by `ColId`s.
    /// - `default_values`: default values for newly added columns ordered by `ColId`s.
```

crates/sats/src/product_type.rs
```diff
@@ -198,6 +198,21 @@ impl ProductType {
...
+                    return false;
+                }
+                self.elements
+                    .iter()
+                    .zip(&pv.elements)
+                    .all(|(ty, val)| ty.algebraic_type.type_check(val))
+            }
+            _ => false,
+        }
+    }
```
Centril:
```suggestion
    /// Returns whether `value` is compatible with this type.
    pub fn type_check(&self, value: &AlgebraicValue) -> bool {
        let AlgebraicValue::Product(pv) = value {
            return false;
        };
        pv.elements.len() == self.elements.len() &&
        self.elements
              .iter()
              .zip(&pv.elements)
              .all(|(ty, val)| ty.algebraic_type.type_check(val))
    }
```

crates/sats/src/algebraic_type.rs
Centril: These aren't used outside of index searches. A user cannot produce them.
These are logically typed at an inference variable, so you can always return `true`, no matter the type.

crates/sats/src/algebraic_type.rs
```diff
@@ -445,8 +445,42 @@ impl AlgebraicType {
...
+            AlgebraicType::U256 => matches!(value, AlgebraicValue::U256(_)),
+            AlgebraicType::F32 => matches!(value, AlgebraicValue::F32(_)),
+            AlgebraicType::F64 => matches!(value, AlgebraicValue::F64(_)),
+
+            AlgebraicType::Sum(sum_ty) => sum_ty.type_check(value),
+            AlgebraicType::Product(product_ty) => product_ty.type_check(value),
+            AlgebraicType::Array(element_ty) => element_ty.type_check(value),
+
+            AlgebraicType::Ref(_) => {
+                // We cannot type check a Ref without the Typespace context
```
Centril: Add the typespace as a parameter? You can pass an empty context where you don't have one otherwise.

crates/sats/src/algebraic_type.rs
```diff
@@ -445,8 +445,42 @@ impl AlgebraicType {
...
     }
-}
 
+    pub fn type_check(&self, value: &AlgebraicValue) -> bool {
+        match value {
+            AlgebraicValue::Min | AlgebraicValue::Max => return false,
+            _ => {}
+        }
+
+        match self {
```
Centril: This feels unnecessarily complicated. Instead:

```suggestion
        match (self, value) {
```

and then you can let e.g., `SumValue::type_check` accept a `value: &SumValue`.

crates/sats/src/algebraic_type.rs
```diff
@@ -702,4 +736,22 @@ mod tests {
...
+        let av = AlgebraicValue::sum(1, AlgebraicValue::from(product![0u16, 1u32]));
+        let at = AlgebraicType::sum([
+            ("a", AlgebraicType::U8),
+            (
+                "b",
+                AlgebraicType::Product(crate::ProductType::new(
+                    [AlgebraicType::U16.into(), AlgebraicType::U32.into()]
+                        .to_vec()
+                        .into_boxed_slice(),
+                )),
```
Centril:
```suggestion
                AlgebraicType::product([AlgebraicType::U16, AlgebraicType::U32])
```

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -494,6 +495,120 @@ impl MutTxId {
...
+    /// After calling this method, Table referred by `table_id` is dropped,
+    /// and a new table is created with the same name but a new `table_id`.
+    /// The new `table_id` is returned.
+    pub(crate) fn add_columns_to_table(
+        &mut self,
+        table_id: TableId,
+        column_schemas: Vec<ColumnSchema>,
+        mut default_values: Vec<AlgebraicValue>,
+    ) -> Result<TableId> {
+        let original_table_schema = Arc::unwrap_or_clone(self.schema_for_table(table_id)?);
```
Centril: (This will always clone as `self` holds references to the `Arc<TableSchema>`.)

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -494,6 +495,120 @@ impl MutTxId {
...
+
+        // Only keep default value of new columns.
```
Centril:
```suggestion
        // Only keep the default values of new columns.
```

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -494,6 +495,120 @@ impl MutTxId {
...
+        let new_cols = column_schemas.len().saturating_sub(original_table_schema.columns.len());
```
Centril: `saturating` feels like it could mask a future bug. Consider `checked` instead.

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -494,6 +495,120 @@ impl MutTxId {
...
+        // and I am not sure if we should rely on the order of sequences in the table schema.
+        let seq_values: HashMap<Box<str>, i128> = original_table_schema
+            .sequences
+            .iter()
+            .map(|s| {
+                (
+                    s.sequence_name.clone(),
+                    self.sequence_state_lock
+                        .get_sequence_mut(s.sequence_id)
+                        .expect("sequence just created")
```
Centril:
```suggestion
                        .expect("sequence exists in original schema and should in sequence state.")
```

crates/datastore/src/locking_tx_datastore/datastore.rs
```diff
@@ -3454,8 +3465,148 @@ mod tests {
...
+        let rows = tx
+            .table_scan(table_id)
+            .unwrap()
+            .map(|row| row.to_product_value())
+            .collect::<Vec<_>>();
+        assert_eq!(rows, old_rows, "Rows shouldn't be changed if rolledback");
+        let table = tx.table(rollback_table_id);
+        assert!(table.is_none(), "new table shouldn't be created if rolledback");
+
+        // Add colum and actually commit this time
```
```suggestion
        // Add column and actually commit this time.
```

crates/sats/src/algebraic_type.rs
Shubham8287: I did it this way because, Using `match (self, value)` requires a default branch `(_ => false)` for all mismatched cases. Matching only on `self` avoids that, and the compiler will force us to handle any new `AlgebraicType variants` explicitly in the future.

Shubham8287: I changed it.

crates/sats/src/sum_type.rs
```diff
@@ -173,6 +173,14 @@ impl SumType {
     pub fn get_variant_by_tag(&self, tag: u8) -> Option<&SumTypeVariant> {
         self.variants.get(tag as usize)
     }
+
+    pub fn type_check(&self, sv: &SumValue, typespace: &Typespace) -> bool {
+        if let Some(variant) = self.get_variant_by_tag(sv.tag) {
+            variant.algebraic_type.type_check(&sv.value, typespace)
+        } else {
+            false
+        }
```
Centril:
```suggestion
        self.get_variant_by_tag(sv.tag)
             .is_some_and(|var| var.algebraic_type.type_check(&sv.value, typespace))
```

crates/sats/src/array_type.rs
```diff
@@ -23,3 +23,11 @@ impl MetaType for ArrayType {
         AlgebraicType::ZERO_REF
     }
 }
+
+impl ArrayType {
+    pub fn type_check(&self, array_value: &ArrayValue, typespace: &Typespace) -> bool {
+        array_value
+            .iter_cloned()
+            .all(|elem| self.elem_ty.type_check(&elem, typespace))
```
Centril, line 31: N.B. this clones the AVs. Because of `type_check` being used in non-perf sensitive places right now, it's not too bad, but ideally this would not clone any AVs. If you want to file an issue to fix this later, I'm down with that.

crates/sats/src/algebraic_type.rs
```diff
@@ -445,8 +446,45 @@ impl AlgebraicType {
...
+            (AlgebraicType::I32, AlgebraicValue::I32(_)) => true,
+            (AlgebraicType::U32, AlgebraicValue::U32(_)) => true,
+            (AlgebraicType::I64, AlgebraicValue::I64(_)) => true,
+            (AlgebraicType::U64, AlgebraicValue::U64(_)) => true,
+            (AlgebraicType::I128, AlgebraicValue::I128(_)) => true,
+            (AlgebraicType::U128, AlgebraicValue::U128(_)) => true,
+            (AlgebraicType::I256, AlgebraicValue::I256(_)) => true,
+            (AlgebraicType::U256, AlgebraicValue::U256(_)) => true,
+            (AlgebraicType::F32, AlgebraicValue::F32(_)) => true,
+            (AlgebraicType::F64, AlgebraicValue::F64(_)) => true,
```
Centril:
```suggestion
            (AlgebraicType::String, AlgebraicValue::String(_))
            | (AlgebraicType::Bool, AlgebraicValue::Bool(_))
            | (AlgebraicType::I8, AlgebraicValue::I8(_))
            | (AlgebraicType::U8, AlgebraicValue::U8(_))
            | (AlgebraicType::I16, AlgebraicValue::I16(_))
            | (AlgebraicType::U16, AlgebraicValue::U16(_))
            | (AlgebraicType::I32, AlgebraicValue::I32(_))
            | (AlgebraicType::U32, AlgebraicValue::U32(_))
            | (AlgebraicType::I64, AlgebraicValue::I64(_))
            | (AlgebraicType::U64, AlgebraicValue::U64(_))
            | (AlgebraicType::I128, AlgebraicValue::I128(_))
            | (AlgebraicType::U128, AlgebraicValue::U128(_))
            | (AlgebraicType::I256, AlgebraicValue::I256(_))
            | (AlgebraicType::U256, AlgebraicValue::U256(_))
            | (AlgebraicType::F32, AlgebraicValue::F32(_))
            | (AlgebraicType::F64, AlgebraicValue::F64(_)) => true,
```

crates/sats/src/algebraic_type.rs
```diff
@@ -445,8 +446,45 @@ impl AlgebraicType {
...
     }
-}
 
+    pub fn type_check(&self, value: &AlgebraicValue, typespace: &Typespace) -> bool {
+        match value {
+            AlgebraicValue::Min | AlgebraicValue::Max => return true,
+            _ => {}
+        }
+
+        match (self, value) {
```
Centril:
```suggestion
        match (self, value) {
            (_, AlgebraicValue::Min | AlgebraicValue::Max) => true
```

crates/sats/src/array_type.rs
Shubham8287: sure, thanks!

Issue #3213, opened 2025-09-01
Investigate ScheduleAt.Time(value) not working in C# modules


jdetter:
## Description

Issue submitted via discord: https://discord.com/channels/1037340874172014652/1411390203213381663

A user has reported that a code snippet like this isn't working in C# modules:
```cs
// wait one more seconds
session.ScheduledAt = new ScheduleAt.Time(ctx.Timestamp + new TimeDuration { Microseconds = 1_000_000 });
ctx.Db.GameSession.id.Update(session);
```

## TODO

- [ ] Validate the issue
- [ ] If the issue exists, fix the issue
- [ ] Regardless, create regression test for the issue

rekhoff:
Attempted to repro failed using a stock Quickstart-Chat C# module and client, modified with example code from our documentation for how to use `ScheduleAt.Time(time)`: [/docs/modules/c-sharp#scheduled-reducers](https://spacetimedb.com/docs/modules/c-sharp#scheduled-reducers)
Using this example on latest public `v1.3.1` build and monitoring the logs for the published module (`spacetime logs quickstart-chat -f`) shows both `Sending scheduled message I'm a bot sending a message one time!` and `Sending scheduled message I'm a bot sending a message every ten seconds!` messages showing up in logs, with the `TimeDuration` log repeating correctly.
Server and client module code attached.
Server code includes an additional message being passed to the client chat portion as part of the scheduled messages, to allow for visibility to clients in chat, resulting in a chat log like:
```[I] SpacetimeDBClient: Connecting to ws://localhost:3000 quickstart-chat
C200A617 is online
Connected
unknown: Sending scheduled message I'm a bot sending a message one time!
unknown: Sending scheduled message I'm a bot sending a message every ten seconds!
unknown: Sending scheduled message I'm a bot sending a message every ten seconds!
This is a message send by the local user into the cli prompt
C200A617: This is a message send by the local user into the cli prompt
unknown: Sending scheduled message I'm a bot sending a message every ten seconds!
```
Edit: Code sample is failing to upload, but the code itself is fairly simple to reimplement based on my description of changes.

m-nt:
Hi, If it helps I'm ok to send you guys the codebase with system information, spacetime version etc...
@rekhoff @jdetter

rekhoff:
I'm not sure if that is necessary at this point @m-nt, but thank you.

[...]

PR #3233, opened 2025-09-05
Adds a non-repeating scheduled reducer test

rekhoff:
# Description of Changes

Adds a non-repeating scheduled reducer integration test to the existing Rust and C# integration tests, in order to address #3213

No

1

# Testing

- [X] Ran `cargo test -p spacetimedb-testing` and all tests passed


modules/module-test-cs/Lib.cs
```diff
@@ -213,6 +234,13 @@ public static void repeating_test(ReducerContext ctx, RepeatingTestArg arg)
         var deltaTime = ctx.Timestamp.TimeDurationSince(arg.prev_time);
         Log.Trace($"Timestamp: {ctx.Timestamp}, Delta time: {deltaTime}");
     }
+    
+    [Reducer]
+    public static void nonrepeating_test(ReducerContext ctx, NonrepeatingTestArg arg)
+    {
+        var deltaTime = ctx.Timestamp.TimeDurationSince(arg.prev_time);
+        Log.Trace($"Timestamp: {ctx.Timestamp}, Delta time: {deltaTime}");
```
jdetter: Where in the test harness do we actually check to see if we get this log? If this log isn't present how does the test fail?

modules/module-test/src/lib.rs
```diff
@@ -198,6 +217,19 @@ pub fn repeating_test(ctx: &ReducerContext, arg: RepeatingTestArg) {
     log::trace!("Timestamp: {:?}, Delta time: {:?}", ctx.timestamp, delta_time);
 }
 
+#[spacetimedb::reducer]
+pub fn nonrepeating_test(ctx: &ReducerContext, arg: NonrepeatingTestArg) {
+    let delta_time = ctx
+        .timestamp
+        .duration_since(arg.prev_time)
+        .expect("arg.prev_time is later than ctx.timestamp... huh?");
+    log::trace!(
```
jdetter, line 261: Same here: Where in the test harness do we actually check to see if we get this log? If this log isn't present how does the test fail?

jdetter (commented):
A few Qs on this

modules/module-test-cs/Lib.cs
rekhoff: Update made with new tests actually being executed (in `crates/testing/tests/standalone_integration_test.rs`).
I've also ensured the tests run in Rust / C# / Typescript. The PR description has been updated with more details on these changes as well.

modules/module-test/src/lib.rs
rekhoff: You are right, I don't believe those tests had ran as part of CI. They should now be run along side the other tests in `crates/testing/tests/standalone_integration_test.rs`

jdetter (approved):
Seems fine, I haven't ran the test myself though

Issue #3191, opened 2025-08-21
Csharp: `Default` value annotation


Shubham8287:
Intended UI -

```csharp
#[Table]
class Banana {
    Color color;
}

// ->
#[SpacetimeDB.Table(name = "Banana")]
class Banana {
    Color color;

    #[SpacetimeDB.Default(0)]
    uint ripeness;
}
```

here is the reference PR for rust counterpart - https://github.com/clockworklabs/SpacetimeDB/pull/3177.
And proposal - https://github.com/clockworklabs/SpacetimeDBPrivate/pull/1849/

rekhoff:
Implemented in https://github.com/clockworklabs/SpacetimeDB/pull/3235

PR #3235, opened 2025-09-08
Rekhoff/csharp default field values

rekhoff:
# Description of Changes

This is the implementation of issue #3191. This adds a Default attribute to C# module fields.

**Note**: In C#, attribute arguments must be compile-time constants, which means you can't directly use non-constant expressions like new expressions, method calls, or dynamic values in attribute constructors.
(Ref: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/attributes#2324-attribute-parameter-types)

For this reason, these default values are limited to primitive types, enums, and strings.
This includes (shown as `C# Type` (`BSATN type`):

* `bool` (`Bool`)
* `sbyte` (`I8`)
* `byte` (`U8`)
* `short` (`I16`)
* `ushort` (`U16`)
* `int` (`I32`)
* `unit` (`U32`)
* `long` (`I64`)
* `ulong` (`U64`)
* `float` (`F32`)
* `double` (`F64`)
* `enum` (`Enum`)
* `string` (`String`)
* `null` (`RefOption`) <- Nullable type

Because of C# limitations, for nullable and complex data types, such as a struct, can take use `[Default(null)]` to populate values with null defaults. This allows things like structs to workaround the non-constant expressions in attribute constructors limitation by allowing these complex types to still be able to be added as new column tables.

The `int` type can also be in the form of Hex or Binary literals, such as `[Default(0x2A)]` or `[Default(0b00101010)]`

Both Decimal (like `[Default(3.14m)]`) and Char (like `[Default('A')]`) are unsupported types in BSATN and will still return `BSATN0001` errors.

Not API breaking.
This change only adds the `[Default(value)]` attribute logic.
Using the `[Default(value)]` attribute with older versions SpacetimeDB C# modules will result in an error.

2

# Testing

Local testing of this requires use of CLI changes in #3278 

- [x] Regression test of functionality added.


crates/bindings-csharp/Codegen.Tests/fixtures/diag/Lib.cs
```diff
@@ -361,6 +362,57 @@ public static partial class InAnotherNamespace
...
+    public long DefaultInt64 = 1;
+
+    [Default(2.0f)]
+    public float DefaultFloat = 1.0f;
+
+    [Default(2.0)]
+    public double DefaultDouble = 1.0;
+
+    [Default(MyEnum.SetByAttribute)]
+    public MyEnum DefaultEnum = MyEnum.SetDuringInit;
```
jdetter: Based on what you had in the PR description, we should be able to add a struct type here and pass the "null" value as the default value right? I think we should add that here so that it's part of the regression test.

jdetter: Also what test is this part of? I see below we have a separate regression test which uploads an initial module and then updates it but where is this struct used in the tests?

rekhoff: Great point, I'll update this test coverage to match the coverage in the regression test as that includes the "null" value test.
You are correct, this is not part of the regression test. All this section does is generate a `received` version of an `FFI.cs` file that is compared to `crates/bindings-csharp/Codegen.Tests/fixtures/diag/snapshots/Module#FFI.verified.cs` . It is used to validate changes to codegen do not impact the `FFI.cs` generated output. If there are changes we anticipate, we update the `verified` version with the expected output to validate against.
The regression test itself is handled entirely separately in `sdk/csharp/examples~/regression-tests/republishing/`, as that test will publish an initial module, add a value to a table, republish with new fields, add another value to the table, and check the resulting table data for the expected results.

jdetter (approved):
I have not manually tested this but the regression tests are easy to understand and are passing. Thanks Ryan this looks great 👍

PR #3239, opened 2025-09-10
CI - Run unreal tests

bfops:
# Description of Changes

## Before merging this
1. uncomment the part of the workflow file that overrides `GIT_REF`. This was causing me issues during testing because it somehow kept pulling back in a rolled-back commit from the base branch, but I suspect that this will no longer be true once new commits are added to the base branch.
1. connect a proper GitHub account to an Epic Games account (specifically, an account that multiple people have access to, e.g. maybe all of our devops team)
2. set the `GHCR_SECRET` GH action secret in this repo to be a long-lived PAT for a different account (https://github.com/clockworklabs/SpacetimeDB/settings/secrets/actions). **The current token will expire soon.**
3. Make this check required, I think.

None. CI-only changes.

2

# Testing
- [x] The CI runs and seems to pass!


jdetter (approved):
The testsuite is passing now, I'll let Zeke do his additional improvements in a separate PR when he gets back

PR #3247, opened 2025-09-15
TypeScript: Add confirmed reads config

kim:
# Description of Changes

Adds a way to configure the DbConnection to use confirmed reads.

1

# Testing

Still figuring out...


kim:
I've been digging around for a way to test this, but it seems like additional URL parameters are added only on a "real" connection object, so it would require an e2e test harness.

If anyone has a better idea, please let me know.

sdks/typescript/packages/sdk/src/db_connection_builder.ts
```diff
@@ -20,6 +20,7 @@ export class DbConnectionBuilder<
   #emitter: EventEmitter<ConnectionEvent> = new EventEmitter();
   #compression: 'gzip' | 'none' = 'gzip';
   #lightMode: boolean = false;
+  #confirmedReads: boolean = false;
```
jsdt: Can we make this `boolean?`, so that we can set a query param if and only if the user sets a value? That way we could let the server decide what the default is.

kim: You mean like `Option<bool>`? Would that be `#confirmedReads: 'default' | boolean` in TS?

kim: Ah, the question mark. Such concise.

kim:
> the color of the bikeshed

There is a subtle difference, in that Postgres allows to apply `synchronous_commit` to individual transactions, whereas we apply it to the entire connection.

That is, from the client's perspective, enabling confirmed reads will affect visibility of transactions made by others (note that this is different from the reducer's perspective).

So, re-using the terminology when it's not exactly the same semantics seems like it could cause confusion.
I'd be open to using "consistency level" or something, although that's arguably less concise and even more jargon-y.

cloutiertyler (approved):
I am open to moving forward with the current spelling. We can always deprecate it later if we want a different spelling.

PR #3251, opened 2025-09-17
feat: add create-spacetime CLI tool

clockwork-tien:
# Description of Changes
- Introduce CLI tool for scaffolding SpacetimeDB projects and create applications with React frontends and Rust or C# backends
- Add `templates/create-spacetime/` with complete CLI implementation which supports interactive project creation, package manager detection, and deployment options (Maincloud and local deployment)
- Update `pnpm-workspace.yaml` to include new package
- Detailed package description and usage examples provided in `templates/create-spacetime/README.md`

# Screenshots
- Run CLI tool using `npm create spacetime@latest`
image
image

# Testing
- [x] Package builds successfully (`pnpm run build` within `templates/create-spacetime`) and runs without errors (`pnpm run test`)
- [x] CLI creates projects with interactive prompts, which support both Rust and C# templates with React and deployment options for local or Maincloud


egormanga:
My two cents:
I'd call it `spacetime-create` instead, as it'll be easier to comprehend in terms of namespaces, and to tab-complete when you forgot the second word (otherwise there's nothing to complete if it comes first).

cloutiertyler:
> My two cents: I'd call it `spacetime-create` instead, as it'll be easier to comprehend in terms of namespaces, and to tab-complete when you forgot the second word (otherwise there's nothing to complete if it comes first).

`npm create` is an npm command, we don't have control over that

templates/create-spacetime/.eslintrc.json
cloutiertyler, line 1: Please update this to the newer `eslint.config.js` format.

cloutiertyler: If this eslint config is just for us, we could also just rely on the root eslint.config.js instead as well.

templates/create-spacetime/src/index.ts
```diff
@@ -0,0 +1,291 @@
...
+      } else {
+        console.log(chalk.gray("Skipping login - default to local deployment"));
+      }
+    } else if (!isLoggedIn && options.yes) {
+      console.log(chalk.gray("Skipping login (--yes mode) - default to local deployment"));
+    }
+    console.log();
+  }
+
+  if (options.template && !isValidTemplate(options.template)) {
```
cloutiertyler: Would it be possible to modify the script so that if it's not in the list of built in templates, it tries to clone the template from GitHub?

cloutiertyler (commented):
This is looking really really good. I just left one more request about downloading templates from GitHub if the user specifies one that we don't know about.

Also as a minor nit, `create-spacetime` itself is not really a template, so it probably shouldn't go in the templates directory. I would imagine the templates directory is where we store/host all the built in templates.

egormanga:
@cloutiertyler Oh, I see now. Thought you were making a standalone scaffolding binary called that. Alright then!

templates/create-spacetime/.eslintrc.json
clockwork-tien: Thank you, I updated to rely on the root eslint.config.js instead 👍

templates/create-spacetime/src/index.ts
clockwork-tien: Thank you, I have updated to add support/option for non built-in templates.

clockwork-tien:
@cloutiertyler Thank you for the review! I have updated to rely on root `eslint.config.js` instead, implemented support/option for non built-in templates, moved `create-spacetime` from `templates` folder to `cli` folder, and published new version.

cli/create-spacetime/src/constants.ts
```diff
@@ -0,0 +1,20 @@
...
+  MAINCLOUD_URI: "wss://maincloud.spacetimedb.com",
+  LOCAL_URI: "ws://localhost:3000",
+} as const;
+
+export const TIMEOUTS = {
+  COMMAND_TIMEOUT: 10000,
+  SERVER_START_DELAY: 3000,
+} as const;
+
+export const SPACETIME_SDK_PACKAGE = "@clockworklabs/spacetimedb-sdk";
```
cloutiertyler: This has now been updated to just `spacetimedb`
```suggestion
export const SPACETIME_SDK_PACKAGE = "spacetimedb";
```

https://www.npmjs.com/package/spacetimedb

cli/create-spacetime/src/constants.ts
```diff
@@ -0,0 +1,20 @@
+export const DEFAULT_TARGET_DIR = "my-spacetime-app";
+
+export const SPACETIME_VERSIONS = {
+  SDK: "^1.3.1",
+  RUNTIME: "1.3.*",
```
cloutiertyler: This doesn't appear to be used anywhere.

cli/create-spacetime/src/lib/create-project.ts
```diff
@@ -0,0 +1,476 @@
...
+      content = content.replace(/^name = .*$/m, `name = "${safeName}"`);
+      content = content.replace(/edition\.workspace = true/g, 'edition = "2021"');
+      content = content.replace(/log\.workspace = true/g, 'log = "0.4"');
+      content = content.replace(
+        /spacetimedb = \{ path = ".*" \}/g,
+        `spacetimedb = "${SPACETIME_VERSIONS.CLI}"`,
+      );
+      content = content.replace(
+        /spacetimedb-lib = \{ path = ".*" \}/g,
+        `spacetimedb-lib = "${SPACETIME_VERSIONS.CLI}"`,
```
cloutiertyler, line 215: This is not the CLI version, it should probably be the runtime version?

npm-packages/create-spacetime/src/index.ts
```diff
@@ -0,0 +1,316 @@
...
+
+  const deps = await checkDependencies();
+
+  if (!deps.node || !deps.npm) {
+    console.error(chalk.red("Error: Node.js and npm are required but not found."));
+    console.error(`Please install Node.js from ${chalk.bold("https://nodejs.org/")}`);
+    process.exit(1);
+  }
+
+  if (!deps.spacetime) {
```
cloutiertyler, line 41: Is there a reason we couldn't just install this for them? Obviously we'd ask them first, but then we could just run the appropriate install script based on the platform.

npm-packages/create-spacetime/tsconfig.json
cloutiertyler, line 1: Could you rename `cli` to `npm-packages` instead? Sorry for the back and forth on this.

npm-packages/create-spacetime/src/index.ts
cloutiertyler: Eventually I think we'd like to move most of this functionality directly into the `spacetime` cli command, so it would make sense that this actually just becomes a simple bootstrapping TypeScript script to install the CLI and then run a command like:

`spacetime init --template foobar/baz`

npm-packages/create-spacetime/src/lib/create-project.ts
```diff
@@ -0,0 +1,476 @@
...
+
+  try {
+    if (await fs.pathExists(cargoPath)) {
+      let content = await fs.readFile(cargoPath, "utf-8");
+      content = content.replace(/^name = .*$/m, `name = "${safeName}"`);
+      content = content.replace(/edition\.workspace = true/g, 'edition = "2021"');
+      content = content.replace(/log\.workspace = true/g, 'log = "0.4"');
+      content = content.replace(
+        /spacetimedb = \{ path = ".*" \}/g,
+        `spacetimedb = "${SPACETIME_VERSIONS.CLI}"`,
```
cloutiertyler: Why are we changing the version here at all? Shouldn't this version be specified in the template?

Additionally, this is not the CLI version. We typically call this version the "module library version". Arguably it should be whatever is in the template, or *arguably* it could be the wildcard version, although that would cause templates to break silently if we push breaking changes.

npm-packages/create-spacetime/src/lib/create-project.ts
```diff
@@ -0,0 +1,476 @@
...
+}
+
+async function updateClientPackageJson(root: string, name: string) {
+  const clientPackagePath = path.join(root, "client/package.json");
+  try {
+    const clientPackage = await fs.readJSON(clientPackagePath);
+    clientPackage.name = `${name}-client`;
+
+    if (clientPackage.dependencies?.[SPACETIME_SDK_PACKAGE]) {
+      clientPackage.dependencies[SPACETIME_SDK_PACKAGE] = SPACETIME_VERSIONS.SDK;
```
cloutiertyler, line 166: I'm also not sure why we're editing the template to add our own version in here as well. I guess to get them on the most up to date version?

cloutiertyler (commented):
I'm having the terrible realization that most of this code should really exist within [`spacetime init`](https://github.com/clockworklabs/SpacetimeDB/blob/master/crates/cli/src/subcommands/init.rs) and be written in Rust. That way people who don't want to execute npm or npx can still use the template creator.

This tool would then become a thin wrapper which would essentially just install `spacetime` if it was not there and then execute `spacetime init --template`.

cli/create-spacetime/src/constants.ts
clockwork-tien: Thank you, I removed this unused variable.

npm-packages/create-spacetime/src/index.ts
clockwork-tien: Ah I could do it as well 👍 . Initially I had a version for that but held back due to cross-platform complexity.

npm-packages/create-spacetime/tsconfig.json
clockwork-tien: Yes I renamed `cli` to `npm-packages` 👍

npm-packages/create-spacetime/src/lib/create-project.ts
clockwork-tien: The template uses a relative path `spacetimedb = { path = "../../crates/bindings" }` instead of a published version, this path doesn't exist when cloning outside the SpacetimeDB monorepo, so I needed to update with a fixed version.

npm-packages/create-spacetime/src/lib/create-project.ts
clockwork-tien: The template has `"spacetimedb": "workspace:*"` in package.json, which didn't work when cloning standalone version so I needed to update with a fixed version.

cloutiertyler:
Closing since we decided to go with `spacetime dev`/`spacetime init` instead.

PR #3261, opened 2025-09-22
fix: truncate

Shubham8287:
# Description of Changes
Based on,and to fix mentioned issues on https://github.com/clockworklabs/SpacetimeDB/pull/3215 i.e -
 1.  Mark table for truncation after anaylyzing `TxData::inserts` too.
 2. `visit_truncates` can not call `schema_for_tables` for dropped tables as schema info has been deleted from `st_*` tables.
 Hence cache name in `ReplayVisitor`.

# Testing
Testing using "Add columns" PR.

### DIFF
--- a/crates/datastore/src/locking_tx_datastore/committed_state.rs
+++ b/crates/datastore/src/locking_tx_datastore/committed_state.rs
@@ -610,13 +610,27 @@ impl CommittedState {
 
     pub(super) fn merge(&mut self, tx_state: TxState, ctx: &ExecutionContext) -> TxData {
         let mut tx_data = TxData::default();
+        let mut truncates = IntSet::default();
 
         // First, apply deletes. This will free up space in the committed tables.
-        self.merge_apply_deletes(&mut tx_data, tx_state.delete_tables, tx_state.pending_schema_changes);
+        self.merge_apply_deletes(
+            &mut tx_data,
+            tx_state.delete_tables,
+            tx_state.pending_schema_changes,
+            &mut truncates,
+        );
 
         // Then, apply inserts. This will re-fill the holes freed by deletions
         // before allocating new pages.
-        self.merge_apply_inserts(&mut tx_data, tx_state.insert_tables, tx_state.blob_store);
+        self.merge_apply_inserts(
+            &mut tx_data,
+            tx_state.insert_tables,
+            tx_state.blob_store,
+            &mut truncates,
+        );
+
+        // Record any truncated tables in the `TxData`.
+        tx_data.add_truncates(truncates);
 
         // If the TX will be logged, record its projected tx offset,
         // then increment the counter.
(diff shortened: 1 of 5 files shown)


crates/datastore/src/locking_tx_datastore/committed_state.rs
```diff
@@ -681,13 +706,15 @@ impl CommittedState {
         for change in pending_schema_changes {
             if let PendingSchemaChange::TableRemoved(table_id, mut table) = change {
                 let row_ptrs = table.scan_all_row_ptrs();
+                truncates.insert(table_id);
```
Centril: Why was this necessary? It should happen in `delete_rows` already?

crates/datastore/src/traits.rs
```diff
@@ -175,9 +175,13 @@ pub struct TxData {
...
     inserts: BTreeMap<TableId, Arc<[ProductValue]>>,
     /// The deleted rows per table.
+    deletes: BTreeMap<TableId, Arc<[ProductValue]>>,
+
+    /// The set of `TableId`s for tables that have been truncated.
     ///
-    /// Also stores per table whether it has been truncated.
-    deletes: BTreeMap<TableId, TxDeleteEntry>,
+    /// “Truncating” means that all rows in the table have been deleted, or the table
+    /// itself has been dropped
```
Centril:
```suggestion
    /// *Truncating* means that all rows in the table have been deleted,
    /// or the table itself has been dropped.
    /// In other words, a truncated table is a cleared table.
    ///
    /// Note that when a table has an entry in `truncates`,
    /// it will also have an entry in `deletes`.
```

crates/datastore/src/locking_tx_datastore/committed_state.rs
Shubham8287: It handles an edge case where already empty table gets dropped.

crates/datastore/src/traits.rs
Shubham8287: >     /// Note that when a table has an entry in `truncates`,
>	   /// it will also have an entry in `deletes`
	    
this is not true if drop happen on already empty table (a table created with module but a row never inserted into it)

crates/datastore/src/locking_tx_datastore/committed_state.rs
Centril: An already empty table is not truncated though. Truncated tables are only those for which all existing rows are removed, in the same transaction.

crates/datastore/src/traits.rs
Centril:
```suggestion
    /// *Truncating* means that all rows in the table have been deleted.
    /// In other words, a truncated table is a cleared table.
    ///
    /// Note that when a table has an entry in `truncates`,
    /// it will also have an entry in `deletes`.
```

Fixed ;)
Removed:
>     /// or the table itself has been dropped.

which is not correct.

crates/datastore/src/locking_tx_datastore/committed_state.rs
Shubham8287: rmeoved it

PR #3263, opened 2025-09-22
`async`-ify Wasmtime (and v8) execution

gefjon:
# Description of Changes

In service of adding procedures, which will need to execute WASM code in an `async` environment so as to suspend execution while making HTTP requests and whatnot.

Prior to this PR, `JobCores` worked by spawning an OS thread for each database. These threads would then each be pinned to a specific core, and in multi-tenant deployments multiple threads could be pinned to the same core. Now, instead, we spawn one thread per available core at startup. Each of these threads runs a single-threaded Tokio executor. Each database is assigned to one of these executors, and runs tasks on it via `tokio::spawn`.

When we run without core pinning (usually due to having too few hardware cores), we won't spawn any additional threads or Tokio runtimes at all; instead we will run database jobs on the "global" Tokio executor. These jobs may block Tokio worker threads, which might be an issue if a very core-constrained device runs multiple databases with very long-running reducers. If this is an issue, we could in this case instead build a second Tokio runtime only for running database jobs, and let the OS scheduler figure things out like it did previously.

Previously, we implemented load-balancing among the database cores by occasionally instructing per-database threads to re-pin themselves. Now, we instead periodically send the database a new `wasmtime::runtime::Handle`, which they will `spawn` future jobs into.

Previously, it was possible for a database thread to become canceled, most likely as a result of `ModuleHost::exit`, after which calls would fail with `NoSuchModule`. Cancellation is no longer meaningful, as the database holds a `Handle` to a long-lived `tokio::runtime::Runtime`, which should always outlive the `ModuleHost`. I have added an `AtomicBool` flag to `ModuleHost` which is flipped by `exit` and checked by calls to maintain the previous functionality.

Within this PR, the jobs run on the database-execution Tokio tasks are not actually asynchronous; they will never yield. This is important because these jobs may (will) hold a transaction open, and attempting to swap execution to another job which wants a transaction on the same database would be undesirable.

Note that this may regress our multi-tenant performance / fairness: previously, in multi-tenant environments, the OS scheduler would divide the database cores' time between the per-database threads, potentially causing one high-load database to be interrupted in the middle of a reducer in order to run other databases pinned to the same core. Now, a high-load database will instead run its entire reducer to completion before any other database gets to run.

We could, in the future, change this by instructing Wasmtime to yield periodically, either via [epoch interruption](https://docs.wasmtime.dev/api/wasmtime/struct.Store.html#method.epoch_deadline_async_yield_and_update) or [fuel](https://docs.wasmtime.dev/api/wasmtime/struct.Store.html#method.fuel_async_yield_interval), both of which we're already configuring Wasmtime to track. We'd need (or at least want) to (re-)introduce a queue s.t. we only attempt to run one job for each database at a time. I have chosen not to do so within this patch because I felt the changeset was complex enough already, and we have so far not treated fairness in multi-tenant environments as a high priority.

I have also reworked our module host machinery to no longer use dynamic dispatch and trait polymorphism to manage modules and  their instances, and instead introduced `enum Module` and `enum Instance`, each of which has a variant for Wasm and another for V8.

During this rewrite, I reworked `AutoReplacingModuleInstance`, which previously used type-erased trait generics in a way that was brittle and hard to re-use in the new `async` context. (Specifically, the module instance no longer lives on the job thread, rather, the database grabs the instance and sends it to the job thread, then gets it back when the job exits. This is necessary to allow the re-worked load balancing described above, as we can't have a single long-lived async task.) While refactoring, I replaced it with `ModuleInstanceManager`, which can now maintain multiple instances of the same module. This is not yet useful, but will become necessary with procedures, as each concurrent procedure will need its own instance. Relatedly, I changed `ModuleHost::on_module_thread` (used by one-off and initial subscription queries) to no longer acquire the/an instance. I discussed this with the team, and consensus was that "locking" the module instance in that path was not a useful behavior, just an artifact of the previous implementation.

I have also switched our Wasmtime configuration to set `async_support(true)`. This causes a variety of methods, notably `InstancePre::instantiate` and `TypedFunc::call`, to panic, and requires that we instead call their `_async` variants. As mentioned above, I have not yet introduced any actual asynchronicity or concurrency, so these methods should never yield. Rather than `.await`ing their futures, I have defined a degenerate `async` executor, `poll_once_executor`, which polls a future exactly once, failing if it does not return `Poll::Ready`. This means that we will panic if one of these futures returns `Poll::Pending` unexpectedly.

The previous `trait Module` had a method `initial_instances`. `Module` is now a concrete type, and I gave it this method, but it appears to be unused. This is causing lints to fail. I am unsure what, if anything, that method was for.

The previous `AutoReplacingModuleInstance` called `create_instance` on the job thread. I am unsure if this was intentional, or just an artifact of the previous implementation, where the `AutoReplacingModuleInstance` lived on the job thread. I have written the new `ModuleInstanceManager` to call `create_instance` on the calling thread, but it would be easy to move that call into the job executor if that behavior is desired.

None user-facing

4: significant rewrite of performance-sensitive fiddly concurrency code.

Note specifically in above description:
- Running database jobs on the global Tokio runtime when not using core pinning.
- Multi-tenant fairness issue: no longer possible to interrupt a performance-intensive database mid-reducer to run another database pinned to the same core.
- Unused method `module_instances`.
- Running `create_instance` on the calling thread rather than the database thread.

# Testing

- [x] Will arrange for a bot test.
- [ ] Determine to what extent we can run with real or synthetic multi-tenant load in a test or staging environment.

### DIFF
--- a/crates/core/src/host/module_host.rs
+++ b/crates/core/src/host/module_host.rs
@@ -327,8 +327,10 @@ pub enum Module {
 }
 
 pub enum Instance {
-    Wasm(super::wasmtime::ModuleInstance),
-    Js(super::v8::JsInstance),
+    // Box these instances because they're very different sizes,
+    // which makes Clippy sad and angry.
+    Wasm(Box<super::wasmtime::ModuleInstance>),
+    Js(Box<super::v8::JsInstance>),
 }
 
 impl Module {
@@ -354,8 +356,8 @@ impl Module {
     }
     fn create_instance(&self) -> Instance {
         match self {
-            Module::Wasm(module) => Instance::Wasm(module.create_instance()),
-            Module::Js(module) => Instance::Js(module.create_instance()),
+            Module::Wasm(module) => Instance::Wasm(Box::new(module.create_instance())),
+            Module::Js(module) => Instance::Js(Box::new(module.create_instance())),
         }
     }
     fn host_type(&self) -> HostType {


crates/core/src/util/poll_once_executor.rs
```diff
@@ -0,0 +1,29 @@
...
+}
+
+pub fn poll_once<Res, Fut: Future<Output = Res>>(mut fut: Fut) -> Option<Res> {
+    let waker = Waker::from(Arc::new(TerribleWaker));
+    let mut context = Context::from_waker(&waker);
+    match <Fut as Future>::poll(pin!(&mut fut), &mut context) {
+        Poll::Pending => None,
+        Poll::Ready(res) => Some(res),
+    }
+}
```
coolreader18, line 29: This is equivalent to `now_or_never` from `futures-util`

crates/core/src/util/poll_once_executor.rs
coolreader18, line 1: This is equivalent to `now_or_never` from `futures-util`; you can just use that.

Centril:
Could you extract https://github.com/clockworklabs/SpacetimeDB/pull/3263/commits/b26d447cc4e1c5210e03659919b3d938cadc4586 to its own PR? We can approve and merge that quickly.

crates/core/src/host/v8/mod.rs
```diff
@@ -93,49 +93,43 @@ impl V8RuntimeInner {
 }
 
 #[derive(Clone)]
-struct JsModule {
+pub struct JsModule {
```
Centril, line 118: My kingdom for stable `type JsModuleImpl = JsModule;` 😭

gefjon:
> Could you extract [b26d447](https://github.com/clockworklabs/SpacetimeDB/commit/b26d447cc4e1c5210e03659919b3d938cadc4586) to its own PR? We can approve and merge that quickly.

@Centril #3284 .

crates/core/src/host/module_host.rs
```diff
@@ -441,49 +491,45 @@ pub struct CallReducerParams {
     pub args: ArgsTuple,
 }
 
-// TODO: figure out how we want to handle traps. maybe it should just not return to the LendingPool and
-//       let the get_instance logic handle it?
-struct AutoReplacingModuleInstance<T: Module> {
-    inst: T::Instance,
-    module: Arc<T>,
+struct ModuleInstanceManager {
+    instances: VecDeque<Instance>,
```
joshua-spacetime, line 505: Can you add a comment here similar to what you said in the PR description, that multiple instances, while not used as of yet, will be necessary for procedures? I understand the comment will be relatively short lived.

crates/core/src/util/jobs.rs
```diff
@@ -1,108 +1,169 @@
...
+/// Each database has a [`SingleCoreExecutor`],
+/// a handle to a single-threaded Tokio runtime which is pinned to a specific CPU core.
+/// In multi-tenant environments, multiple databases' [`SingleCoreExecutor`]s may be handles on the same runtime/core,
+/// and a [`SingleCoreExecutor`] may occasionally be migrated to a different runtime/core to balance load.
 ///
-/// Construction is done via the `FromIterator` impl. If created from an empty
-/// iterator or via `JobCores::default()`, the job threads will work but not be
-/// pinned to any threads.
+/// Construct a `JobCores` via [`Self::from_pinned_cores`] or [`Self::without_pinned_cores`].
+/// a `JobCores` constructed without core pinning, including `from_pinned_cores` on an empty set,
```
joshua-spacetime:
```suggestion
/// A `JobCores` constructed without core pinning, including `from_pinned_cores` on an empty set,
```

crates/core/src/util/jobs.rs
```diff
@@ -1,108 +1,169 @@
...
+/// will use the "global" Tokio executor to run database jobs,
+/// rather than creating multiple un-pinned single-threaded runtimes.
```
joshua-spacetime: I would at least add what you mention in the PR description, that without core pinning, transactions will now run on, and therefore will block, tokio worker threads.

crates/core/src/host/module_host.rs
```diff
@@ -441,49 +491,45 @@ pub struct CallReducerParams {
...
-impl<T: Module> AutoReplacingModuleInstance<T> {
-    fn check_trap(&mut self) {
-        if self.inst.trapped() {
-            self.inst = self.module.create_instance()
+impl ModuleInstanceManager {
+    fn get_instance(&mut self) -> Instance {
+        if let Some(inst) = self.instances.pop_back() {
+            inst
+        } else {
+            // TODO: should we be calling `create_instance` on the `SingleCoreExecutor` rather than the calling thread?
```
joshua-spacetime, line 553: How expensive can the instantiation be?

crates/core/src/util/jobs.rs
```diff
@@ -1,108 +1,169 @@
...
+            inner: JobCoresInner::NoPinning(global_runtime),
+        }
+    }
 }
 
-impl JobCoresInner {
-    fn allocate(&mut self) -> (JobThreadId, watch::Receiver<CoreId>) {
-        let id = self.next_id;
+impl PinnedCoresExecutorManager {
+    fn allocate(&mut self) -> (SingleCoreExecutorId, watch::Receiver<runtime::Handle>) {
```
Centril, line 162: Can you add some commentary to this?

crates/core/src/host/module_host.rs
```diff
@@ -315,34 +318,81 @@ impl ReducersMap {
...
-    fn initial_instances(&mut self) -> Self::InitialInstances<'_>;
-    fn info(&self) -> Arc<ModuleInfo>;
-    fn create_instance(&self) -> Self::Instance;
+pub enum Instance {
+    Wasm(super::wasmtime::ModuleInstance),
+    Js(super::v8::JsInstance),
 }
 
-pub trait ModuleInstance: Send + 'static {
-    fn trapped(&self) -> bool;
```
Centril, line 335: Hmm, I think I'd like to keep the traits around for documentation purposes of what the common interface is while still nixing the dynamic dispatch.

crates/core/src/host/module_host.rs
```diff
@@ -441,49 +491,45 @@ pub struct CallReducerParams {
...
-        ret
-    }
-    fn call_reducer(&mut self, tx: Option<MutTxId>, params: CallReducerParams) -> ReducerCallResult {
-        let ret = self.inst.call_reducer(tx, params);
-        self.check_trap();
-        ret
+    fn return_instance(&mut self, inst: Instance) {
+        if inst.trapped() {
+            // Don't return trapped instances;
+            // they may have left internal data structures in WASM linear memory in a bad state.
```
Centril: In general, we'll need to tweak the wording, now or later, mentioning WASM in various places to be more generic so that it also applies to V8.

crates/core/src/host/module_host.rs
```diff
@@ -497,9 +543,11 @@ impl fmt::Debug for ModuleHost {
 
 pub struct WeakModuleHost {
```
Centril, line 597: If you can, it would be great to add some commentary around the use/purpose of this type.

crates/core/src/host/module_host.rs
```diff
@@ -630,33 +699,50 @@ impl ModuleHost {
...
+    async fn call<F, R>(&self, label: &str, f: F) -> Result<R, NoSuchModule>
+    where
+        F: FnOnce(&mut Instance) -> R + Send + 'static,
+        R: Send + 'static,
+    {
+        self.guard_closed()?;
+        let timer_guard = self.start_call_timer(label);
 
         // Operations on module instances (e.g. calling reducers) is blocking,
         // partially because the computation can potentialyl take a long time
```
```suggestion
        // partially because the computation can potentially take a long time
```

crates/core/src/host/module_host.rs
gefjon: I do not know who uses this type or why, unfortunately.

crates/core/src/host/module_host.rs
gefjon: I would argue that the methods on `Module` and `Instance` already document the common interface.

crates/core/src/host/module_host.rs
gefjon: I don't know. I'll add a metric and we can measure.

crates/core/src/host/module_host.rs
gefjon: I've updated here and the other reference to WASM in this file. Unsure where else needs to be updated.

crates/core/src/host/module_host.rs
gefjon: Added metric `spacetime_module_create_instance_time_seconds` so we can measure this.

crates/core/src/host/module_host.rs
```diff
@@ -321,6 +322,13 @@ pub trait ModuleRuntime {
     fn make_actor(&self, mcc: ModuleCreationContext<'_>) -> anyhow::Result<Module>;
 }
 
+/// Used as a metrics label on `spacetime_module_create_instance_time`.
+#[derive(Clone, Copy, Display, Hash, PartialEq, Eq, strum::AsRefStr)]
+pub enum ModuleType {
+    Wasm,
+    Js,
+}
```
Centril: This exists as `HostType`.

crates/core/src/host/module_host.rs
```diff
@@ -364,6 +372,12 @@ impl Module {
             Module::Js(module) => Instance::Js(module.create_instance()),
         }
     }
+    fn module_type(&self) -> ModuleType {
+        match self {
+            Module::Wasm(_) => ModuleType::Wasm,
+            Module::Js(_) => ModuleType::Wasm,
```
Centril: Oops?

crates/core/src/host/module_host.rs
Centril: They do, but traits also enforce the interface, which is nice, but it's a minor point if you still disagree (in which case feel free to resolve).

crates/core/src/host/module_host.rs
gefjon: :facepalm:

joshua-spacetime (approved):
Perf test showed no regressions

PR #3267, opened 2025-09-22
Bump remaining versions to 1.4.0

bfops:
# Description of Changes

Bumping remaining files so that everything is at 1.4.0.

None.

1

# Testing

None


jdetter (approved):
Cool LGTM, I verified that the license file has been updated with the correct version as well. Maybe double check that the change date has been updated as I didn't check that.

cloutiertyler (approved):

PR #3268, opened 2025-09-23
Fix `tools/publish-crates.sh`

bfops:
# Description of Changes

Fixes two problems introduced in #3185:
1. The `--directories` option on `find-publish-list.py` was not printing directories
2. The `publish-crates.sh` script was using an undefined variable.

No breaking changes

1

# Testing

This was used to publish crates in the release yesterday.

### DIFF
--- a/tools/publish-crates.sh
+++ b/tools/publish-crates.sh
@@ -72,8 +72,9 @@ for path in "${CRATES[@]}"; do
         fi
     fi
 
+    crate="$(basename "$path")"
     # Add owners
-    echo "INFO: Adding owners for $crate..."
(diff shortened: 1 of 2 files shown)


tools/publish-crates.sh
```diff
@@ -72,8 +72,9 @@ for path in "${CRATES[@]}"; do
         fi
     fi
 
+    crate="$(basename "$path")"
     # Add owners
-    echo "INFO: Adding owners for $crate..."
+    echo "Adding owners for $crate..."
```
bfops, line 77: idk why we printed `INFO:` here and below, and not in other places :shrug:

PR #3269, opened 2025-09-23
Updated the size scripts for the TypeScript package

cloutiertyler:
# Description of Changes

Tiny fix to update the size scripts for the `spacetimedb` TypeScript package

none

1

# Testing

- [ ] I verified the size script do be workin.


bfops (approved):
this seems to fix `pnpm publish` 🤷

PR #3275, opened 2025-09-23
core: disconnect clients 

Shubham8287:
# Description of Changes
The `AutoMigrateStep::DisconnectAllUsers` step is implemented as follows:

1.  The `spacetimedb::db::update::update_database` function returns a response of type `UpdateDatabaseResult::UpdatePerformedWithClientDisconnect`. 

2.  Upon receiving this response, the `host_controller::update_module` proceeds to drop the `watch::Sender<ModuleHost>` field within the `core::host_controller::Host`  and disconnect clients.

companion private PR - https://github.com/clockworklabs/SpacetimeDBPrivate/pull/2095
NA

3.
Diff code is simple but It depends on the subcription logic to behave correctly.

# Testing
Manually.


crates/core/src/db/update.rs
```diff
@@ -264,7 +269,12 @@ fn auto_migrate_database(
                     .collect();
                 stdb.add_columns_to_table(tx, table_id, column_schemas, default_values)?;
             }
-            _ => anyhow::bail!("migration step not implemented: {step:?}"),
+            spacetimedb_schema::auto_migrate::AutoMigrateStep::DisconnectAllUsers => {
+                // Disconnect all clients from subscriptions.
+                // Any dangling clients will be handled during the launch of module hosts,
+                // which invokes `ModuleHost::call_identity_disconnected`.
+                subscriptions.remove_all_subscribers();
```
Shubham8287: I am not sure if this is enough and correct ( to not calling disconnect reducer).
I am relying on reviewer's guidance for this.

gefjon: I don't think this is correct. `remove_all_subscriptions` doesn't notify clients that they've been disconnected -- it's meant to be called after the client is already gone, to clean up after them. Using it in this way will leave clients hanging, never receiving updates but also never notified that they need to reconnect. It also won't prevent them from registering new subscriptions, calling reducers, sending one-off queries, &c. You'll need to find some way to signal to the `ws_main_loop` in `spacetimedb-client-api::routes::subscribe` that it should exit, preferably after sending a suitable close frame.

kim: My intuition is that if there is a user-defined disconnect reducer, it should be run here (in the `tx`). If there isn't one, it doesn't matter.

The reason is that iirc, modules can reject clients by returning an error from the connect reducer, for example because they want only a single connection per client. If the module doesn't know that the client disconnected, it would wrongly reject it.

Strictly speaking, disconnecting the subscribers should probably occur after `tx` committed. It's not clear to me if the transaction can actually fail at this point, so maybe that's ok.

crates/core/src/subscription/module_subscription_manager.rs
```diff
@@ -779,6 +779,12 @@ impl SubscriptionManager {
         }
     }
 
+    pub fn remove_all_clients(&mut self) {
+        for id in self.clients.keys().copied().collect::<Vec<_>>() {
```
Centril:
```suggestion
        // TODO(perf): This creates unnecessary temp allocations,
        // which can be avoided by not using `remove_all_subscriptions`.
        // However, `remove_all_clients` is used in cold paths,
        // so this is not a priority to fix right now.
        for id in self.clients.keys().copied().collect::<Vec<_>>() {
```

crates/core/src/db/update.rs
Shubham8287: Thanks for insights.

I modified the approach, such that if update finds out client needs to be dropped. It does not hotswap but drops both new and old `ModuleHost`. Which will make receiver side of watcher in [`ws_main_loop`](https://github.dev/clockworklabs/SpacetimeDB/blob/dcf8a2ab7a4325810523c760afe31b25dc915681/crates/client-api/src/routes/subscribe.rs#L610-L611) to drop after sending close frame.

Dropping the `ModuleHost` means next time `Host::try_init` will be called, which will take care of cleaning up all clients from `st_*` tables and calling disconnect reducer. The downside is that it’s slower, since it has to replay the commit log during initialization.

I think, I can acheive similar behavior to drop clients by just replacing a `watch::Sender` in new `ModuleHost` instead of dropping whole `ModuleHost`  and writting some code to handle clients disconnect in db. I can do that, if we feel we should avoid replaying.

crates/core/src/subscription/module_subscription_manager.rs
Shubham8287: thanks, but I changed the approach :)

crates/core/src/db/update.rs
Shubham8287: I realized, I do not actually need to handle dangling clients `ws_client_actor` will do that. Pushed new changes.

crates/core/src/db/update.rs
```diff
@@ -23,6 +23,13 @@ impl UpdateLogger for SystemLogger {
     }
 }
 
+/// The result of a database update.
+/// Indicates whether clients should be disconnected when the update is complete.
+pub enum UpdateResult {
```
gefjon, line 29:
```suggestion
#[must_use]
pub enum UpdateResult {
```

gefjon (dismissed):
Excellent, thanks! This seems like the correct way to disconnect clients.

gefjon:
How does `client_disconnected` get called for the disconnected clients?

Shubham8287:
> How does `client_disconnected` get called for the disconnected clients?

Dropping  `watch::sender` leads to exit of `ws_client_actor` which was `await`ing on `ws_main_loop`.

https://github.com/clockworklabs/SpacetimeDB/blob/2b61190d4d7cdfe27ce9240991b6e3940a1a1028/crates/client-api/src/routes/subscribe.rs#L371
`client.disconnect()` imeplementation calls `ModuleHost::disconnect_client()` which does the job.

Shubham8287:
> How does client_disconnected get called for the disconnected clients?

I fixed `disconnect_redudcer` not being called by explicitly calling it while updating module.
I also updated [smoketest](https://github.com/clockworklabs/SpacetimeDB/pull/3278/files#diff-0fe6bbb483741c5a4f0ded4cdc053a5b172c69e77d62c89ffc2c3dedd7d72126R231) to test the disconnect reducer calls .

crates/core/src/host/module_host.rs
```diff
@@ -782,22 +782,47 @@ impl ModuleHost {
...
+                log::debug!(
+                    "client: {caller_identity} with connection_id: {caller_connection_id} already disconnected"
+                );
+                return Ok(());
+            }
+            Err(e) => {
+                return Err(InvalidReducerArguments {
+                    err: e.into(),
+                    reducer: reducer_name.into(),
+                }
```
gefjon: `InvalidReducerArguments` doesn't make a ton of sense to me as the error here.

crates/core/src/host/module_host.rs
```diff
@@ -782,22 +782,47 @@ impl ModuleHost {
...
+        let reducer_name = reducer_lookup
+            .as_ref()
+            .map(|(_, def)| &*def.name)
+            .unwrap_or("__identity_disconnected__");
+
+        match self
+            .module
+            .replica_ctx()
+            .relational_db
+            .st_client_row(caller_identity, caller_connection_id)
```
gefjon: This feels like a race condition. Your `st_client_row` method is opening its own short-lived transaction, so the client can still disconnect after you make this check but before you acquire the reducer TX. I think you need to acquire a mut tx at the start of this function, use it to check if the client is present, then pass that same tx into `call_reducer_inner_with_inst` or use it as the fallback transaction.

Shubham8287: True, I didn't care about atomicity here, just made it idempotent as that seems sufficient but I think its not too hard to make it atomic as well. So, I wil do that.

gefjon: Without atomicity it's not idempotent.

crates/core/src/host/module_host.rs
Shubham8287: Method is to call reducer, and but call shouldn't be made because arguments passed is invalid in the sense that they do not exist in client table anymore. We are using same error for fallback tx. Wht do you suggest?

gefjon: `InvalidArguments` means the passed arguments are of incorrect type.

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -1576,6 +1576,23 @@ impl MutTxId {
         self.delete_st_client_credentials(database_identity, connection_id)
     }
 
+    /// Look up a client row by identity and connection ID in the `st_clients` system table.
+    ///
+    /// `Ok(None)` if no such row exists.
+    pub fn st_client_row(&self, identity: Identity, connection_id: ConnectionId) -> Result<Option<RowPointer>> {
```
gefjon: In what circumstance can this method fail? When the `st_client` table doesn't match the expected schema? If that's the only case, we should panic here, and simplify the return type to remove the `Result`.

Shubham8287: Makes sense. Done.

PR #3276, opened 2025-09-23
V8: Enable as unstable, add syscalls, flesh out call_reducer, etc.

Centril:
# Description of Changes

Update:

This PR did all of the below but was split. Now it just does:

1. Exposes V8/JS modules via the `unstable` feature flag on the host. To publish a JS module, `--js-path path/to/module.js` 

This PR:

1. Exposes V8/JS modules via the `unstable` feature flag on the host. To publish a JS module, `--js-path path/to/module.js` needs to be used.
2. Bumps V8 to 140.2.
3. Shares more logic with WASM and makes some minor refactorings to energy/budget logic.
4. Moves logic from `WasmInstanceEnv` to `InstanceEnv` and friends.
5. Makes JS modules actually work in terms of `create_instance`, `make_actor`, 
6. Fleshes out `call_reducer` with timeouts and long-running logs added as well.
7. Adds all the syscalls with associated documentation as well.

None

2? It's only available on unstable and mostly touches V8 stuff.

# Testing

Follow up PRs will add unit tests for parts.
We'll also need to add integration tests for whole modules.


bfops:
Is this meant to be reviewed or just here for a demo?

Centril:
> Is this meant to be reviewed or just here for a demo?

It's meant to be reviewed.

crates/cli/src/subcommands/publish.rs
```diff
@@ -48,6 +48,15 @@ pub fn cli() -> clap::Command {
...
         )
+        // TODO(v8): needs better UX but good enough for a demo...
+        .arg(
+            Arg::new("javascript")
+                .long("javascript")
+                .action(SetTrue)
+                .requires("wasm_file")
+                .hide(true)
+                .help("UNSTABLE: interpret `--bin-path` as a JS module"),
+        )
```
bfops, line 61: I haven't been keeping context on the timelines/milestones for the V8 stuff. Is the "demo" here meaning an internal demo, or a beta release to users?

Centril: More of the former. I'm not sure this is the UX we want to expose to beta users, but I'd like to merge something provisional into master that lets us test stuff. I'd love to discuss with you what a good UX would be, but we can probably do that on Discord for now.

cloutiertyler: I discussed this with @Centril today in the Core daily meeting. For the beta release we decided to have parity with the UX of both Rust and C#, so this will not be necessary because SpacetimeDB will detect that you have an npm package and will build your app for you.

```
// i.e. This will just work
spacetime publish
```

However, we also want to achieve parity with manual way of specifying the `--bin-path` arg for wasm by introducing a new `--js-path` manual way of publishing after having already bundled your javascript.

Once @Centril makes that change I will approve this PR for the files that I am a code owner of.

Centril: > Once @Centril makes that change I will approve this PR for the files that I am a code owner of.

This was fixed in the commit named `cli/publish: use --js-path instead of --javascript`.

crates/cli/src/subcommands/publish.rs
```diff
@@ -108,13 +120,19 @@ pub async fn exec(mut config: Config, args: &ArgMatches) -> Result<(), anyhow::E
...
     }
 
-    let path_to_wasm = if let Some(path) = wasm_file {
-        println!("Skipping build. Instead we are publishing {}", path.display());
-        path.clone()
+    // Decide program file path and read program.
+    // Optionally build the program.
+    let (path_to_program, host_type) = if let Some(path) = wasm_file {
+        println!("(WASM) Skipping build. Instead we are publishing {}", path.display());
+        (path.clone(), None)
```
bfops: nit: I would prefer providing an explicit `host_type` in this case too

Centril: Fixed in a2d40da84937aa412fd9caf1c22f9a78d1af1ed2.

PR #3277, opened 2025-09-23
Finish making PG tests use a dynamic server address

bfops:
# Description of Changes

Finish the work started in #3227, which switched a hardcoded `127.0.0.1` to a dynamic server address.

None.

1

# Testing

- [x] Smoketests still pass
- [ ] `Internal Tests` pass under this PR (they don't on master)

### DIFF
--- a/smoketests/tests/quickstart.py
+++ b/smoketests/tests/quickstart.py
@@ -125,7 +125,7 @@ class BaseQuickstart(Smoketest):
             main = main.replace(src, dst)
         main += "\n" + self.extra_code
         server = self.get_server_address()
-        host = server["host"]
+        host = server["address"]
         protocol = server["protocol"]
         main = main.replace("http://localhost:3000", f"{protocol}://{host}")
         _write_file(client_path / self.client_file, main)


smoketests/tests/pg_wire.py
```diff
@@ -168,9 +152,25 @@ class SqlFormat(Smoketest):
 }
 """
 
+    def psql(self, identity: str, sql: str) -> str:
+        server = self.get_server_address()
+        """Call `psql` and execute the given SQL statement."""
+        result = subprocess.run(
+            ["psql", "-h", server["host"], "-p", "5432", "-U", "postgres", "-d", "quickstart", "--quiet", "-c", sql],
```
bfops, line 151: moved so that we can use `self.get_server_address`, but the diff is just `127.0.0.1` -> `server["host"]`

smoketests/tests/pg_wire.py
```diff
@@ -168,9 +144,32 @@ class SqlFormat(Smoketest):
...
+        )
+
+        if result.stderr:
+            raise Exception(result.stderr.strip())
+        return result.stdout.strip()
+
+    def connect_db(self, identity: str):
+        """Connect to the database using `psycopg2`."""
+        server = self.get_server_address()
+        conn = psycopg2.connect(host=server["host"], port=5432, user="postgres", password=identity, dbname="quickstart")
```
bfops, line 166: moved so that we can use `self.get_server_address`, but the diff is just `127.0.0.1` -> `server["host"]`

PR #3281, opened 2025-09-24
Expand scope of `DurabilityProvider` to include snapshotting

kim:
The `DurabilityProvider` trait was introduced to enable the `HostController` to procure an alternative `Durability` impl from an external source.

It is also useful to be able to instantiate a `SnapshotWorker` externally, in order to subscribe to snapshot creation events without access to the `RelationalDB` instance it is operating on.

At a later stage, we may also use it to control the snapshot frequency externally.

This patch thus reframes the trait as `PersistenceProvider`, whose job is to provide persistence-related services.

Also separates snapshot creation and compression of older snapshots, and adds instrumentation to gather timing information for both.

1.5

# Testing

Not a functional change, existing tests should cover that.

### DIFF
--- a/crates/core/src/db/relational_db.rs
+++ b/crates/core/src/db/relational_db.rs
@@ -7,15 +7,11 @@ use crate::worker_metrics::WORKER_METRICS;
 use anyhow::{anyhow, Context};
 use enum_map::EnumMap;
 use fs2::FileExt;
-use futures::channel::mpsc;
-use futures::StreamExt;
-use parking_lot::RwLock;
 use spacetimedb_commitlog as commitlog;
 use spacetimedb_data_structures::map::IntSet;
 use spacetimedb_datastore::db_metrics::DB_METRICS;
 use spacetimedb_datastore::error::{DatastoreError, TableError};
 use spacetimedb_datastore::execution_context::{ReducerContext, Workload, WorkloadType};
-use spacetimedb_datastore::locking_tx_datastore::committed_state::CommittedState;
 use spacetimedb_datastore::locking_tx_datastore::datastore::TxMetrics;
 use spacetimedb_datastore::locking_tx_datastore::state_view::{
     IterByColEqMutTx, IterByColRangeMutTx, IterMutTx, IterTx, StateView,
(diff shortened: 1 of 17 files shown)


crates/core/src/db/relational_db.rs
```diff
@@ -33,7 +29,7 @@ use spacetimedb_datastore::{
     },
     traits::TxData,
 };
-use spacetimedb_durability as durability;
+use spacetimedb_durability::{self as durability};
```
gefjon: Is this different from the previous import `use spacetimedb_durability as durability;`?

crates/core/src/worker_metrics/mod.rs
```diff
@@ -284,6 +284,43 @@ metrics_group!(
...
+        // Not sure what range to expect, but certainly slower than snapshot
+        // creation.
+        #[buckets(0.001, 0.01, 0.1, 1.0, 5.0, 10.0)]
+        pub snapshot_compression_time_total: HistogramVec,
+
+        #[name = spacetime_snapshot_compression_time_inner_sec]
+        #[help = "The time (in seconds) it took to do a compression pass on the snapshot repository, excluding scheduling overhead"]
+        #[labels(db: Identity)]
+        #[buckets(0.001, 0.01, 0.1, 1.0, 5.0, 10.0)]
+        pub snapshot_compression_time_inner: HistogramVec,
```
gefjon, line 351: It seems also useful to have metrics for:
- Number of snapshots already compressed (skipped) in a compression pass.
- Number of snapshots newly compressed in a compression pass.
- Time to compress an individual snapshot.

Do you think it would be sensible to include these in this PR, or would a follow-up be better?

gefjon (approved):
Looks great! Love the clear motivating comments.

crates/core/src/db/relational_db.rs
kim: The machine did this. I guess.

crates/core/src/worker_metrics/mod.rs
kim: Those sound good, but turned out to be a bit more involved. Follow-up incoming shortly.

PR #3282, opened 2025-09-24
C#: Add confirmed reads configuration

kim:
Allows the SpacetimeDBClient to be configured with or without confirmed reads. Like for [TypeScript], the parameter is optional -- the server chooses the default if not set explicitly.

[TypeScript]: https://github.com/clockworklabs/SpacetimeDB/pull/3247

1

# Testing

I don't actually know what I'm doing :sweat_smile:


JasonAtClockwork (dismissed):
LGTM - when we improve testing in the future we may want to include confirmation of this flag being set and read.

JasonAtClockwork (approved):
LGTM - when we improve testing in the future we may want to include confirmation of this flag being set and read.

I did test this against the regression tests with and without and had no troubles.

PR #3287, opened 2025-09-24
Fix UPROPERTY initialization and new Blueprint Library for FContextBase

JasonAtClockwork:
# Description of Changes

Closes: https://github.com/clockworklabs/SpacetimeDBPrivate/issues/2058

- Updated the generation code to setup basic initialization to properties for Unreal Blueprints
- Added new Blueprint library to interact with FContextBase to allow access to inherited properties from all contexts in Blueprint

No breaking changes

2 - Updates the generation code

# Testing

- [x] Ran Unreal tests to confirm no breaking changes


crates/codegen/src/unrealcpp.rs
```diff
@@ -4157,6 +4202,43 @@ fn cpp_ty_fmt_impl<'a>(
     })
 }
 
+// For UPROPERTY() Unreal expects initialization values for certain types
+// (e.g. bools default to true if not explicitly initialized to false).
+fn cpp_ty_init_fmt_impl<'a>(ty: &'a AlgebraicTypeUse) -> impl fmt::Display + 'a {
+    fmt_fn(move |f| match ty {
+        AlgebraicTypeUse::Primitive(p) => f.write_str(match p {
+            PrimitiveType::Bool => " = false;",
```
jdetter: Do the reset of the values below also need the semicolon at the end or is this a typo?

crates/codegen/src/unrealcpp.rs
```diff
@@ -4157,6 +4202,43 @@ fn cpp_ty_fmt_impl<'a>(
...
+            PrimitiveType::I64 => " = 0",
+            PrimitiveType::U64 => " = 0",
+            PrimitiveType::F32 => " = 0.f",
+            PrimitiveType::F64 => " = 0.f",
+            PrimitiveType::I128 => "",
+            PrimitiveType::U128 => "",
+            PrimitiveType::I256 => "",
+            PrimitiveType::U256 => "",
+        }),
+        AlgebraicTypeUse::Array(_elem) => f.write_str(""),
```
jdetter, line 4226: If you don't pass initializers for these sorts of values will they not compile or what will happen?

jdetter (commented):
Few questions on this one, no objections so far 👍

crates/codegen/src/unrealcpp.rs
JasonAtClockwork: This is definitely a small miss, we'll end up with double semi-colon:
```cpp
    bool IsBoolean = false;;
```

Fixed up in the next commit!

crates/codegen/src/unrealcpp.rs
JasonAtClockwork: With UPROPERTY it will look to the struct initialization and I128, U128, I256, U256 are backed by structs:
```cpp
            PrimitiveType::I128 => "FSpacetimeDBInt128",
            PrimitiveType::U128 => "FSpacetimeDBUInt128",
            PrimitiveType::I256 => "FSpacetimeDBInt256",
            PrimitiveType::U256 => "FSpacetimeDBUInt256",
```

crates/codegen/src/unrealcpp.rs
jdetter: I think you'll need to push or something because it's still rendering with the semicolon here, just make sure it's updated before merging 👍

crates/codegen/src/unrealcpp.rs
jdetter: I see, thank you 👍

jdetter (approved):
I haven't tested the changes myself but this looks like a good change to me. Thanks @JasonAtClockwork! Can you make sure to check the duplicate semicolon thing one more time before merging? Thanks 👍

crates/codegen/src/unrealcpp.rs
JasonAtClockwork: Yeah I hadn't pushed yet as I was doing extra testing, it's all good now and resolving the conversations

JasonAtClockwork:
> I haven't tested the changes myself but this looks like a good change to me. Thanks @JasonAtClockwork! Can you make sure to check the duplicate semicolon thing one more time before merging? Thanks 👍

Small repairs pushed and the float/double were functional but not as clean as they could be

PR #3288, opened 2025-09-24
Add AuthCtx to ReducerContext for rust

jsdt:
# Description of Changes

This exposes client credentials in reducer calls for rust.

API Changes:

The main API change is the addition of `AuthCtx` and the `sender_auth` in `ReducerContext`. This also adds JwtClaims, which has some helpers for getting commonly used claims.

ABI Changes:

This adds one new functions `get_jwt`. This uses `st_connection_credentials` to look up the credentials associated with a connection id.

This adds ABI version 10.2.

2. This adds new ABI functions

# Testing

I've done some manual testing with modified versions of the quickstart. We should add some examples that use the new API.

### DIFF
--- a/crates/datastore/src/locking_tx_datastore/committed_state.rs
+++ b/crates/datastore/src/locking_tx_datastore/committed_state.rs
@@ -6,7 +6,10 @@ use super::{
     tx_state::{IndexIdMap, PendingSchemaChange, TxState},
     IterByColEqTx,
 };
-use crate::system_tables::{ST_CONNECTION_CREDENTIALS_ID, ST_CONNECTION_CREDENTIALS_IDX};
+use crate::system_tables::{
+    ST_CONNECTION_CREDENTIALS_ID, ST_CONNECTION_CREDENTIALS_IDX, ST_VIEW_COLUMN_ID, ST_VIEW_COLUMN_IDX, ST_VIEW_ID,
+    ST_VIEW_IDX, ST_VIEW_PARAM_ID, ST_VIEW_PARAM_IDX,
+};
 use crate::{
     db_metrics::DB_METRICS,
     error::{DatastoreError, IndexError, TableError},
@@ -252,6 +255,10 @@ impl CommittedState {
             ST_CONNECTION_CREDENTIALS_ID,
             schemas[ST_CONNECTION_CREDENTIALS_IDX].clone(),
         );
+
+        self.create_table(ST_VIEW_ID, schemas[ST_VIEW_IDX].clone());
+        self.create_table(ST_VIEW_PARAM_ID, schemas[ST_VIEW_PARAM_IDX].clone());
+        self.create_table(ST_VIEW_COLUMN_ID, schemas[ST_VIEW_COLUMN_IDX].clone());
 
         // Insert the sequences into `st_sequences`
         let (st_sequences, blob_store, pool) =
--- a/crates/datastore/src/locking_tx_datastore/datastore.rs
+++ b/crates/datastore/src/locking_tx_datastore/datastore.rs
@@ -1247,11 +1247,12 @@ mod tests {
     use crate::system_tables::{
         system_tables, StColumnRow, StConnectionCredentialsFields, StConstraintData, StConstraintFields,
         StConstraintRow, StIndexAlgorithm, StIndexFields, StIndexRow, StRowLevelSecurityFields, StScheduledFields,
-        StSequenceFields, StSequenceRow, StTableRow, StVarFields, ST_CLIENT_NAME, ST_COLUMN_ID, ST_COLUMN_NAME,
-        ST_CONNECTION_CREDENTIALS_ID, ST_CONNECTION_CREDENTIALS_NAME, ST_CONSTRAINT_ID, ST_CONSTRAINT_NAME,
-        ST_INDEX_ID, ST_INDEX_NAME, ST_MODULE_NAME, ST_RESERVED_SEQUENCE_RANGE, ST_ROW_LEVEL_SECURITY_ID,
-        ST_ROW_LEVEL_SECURITY_NAME, ST_SCHEDULED_ID, ST_SCHEDULED_NAME, ST_SEQUENCE_ID, ST_SEQUENCE_NAME,
-        ST_TABLE_NAME, ST_VAR_ID, ST_VAR_NAME,
+        StSequenceFields, StSequenceRow, StTableRow, StVarFields, StViewFields, ST_CLIENT_NAME, ST_COLUMN_ID,
+        ST_COLUMN_NAME, ST_CONNECTION_CREDENTIALS_ID, ST_CONNECTION_CREDENTIALS_NAME, ST_CONSTRAINT_ID,
+        ST_CONSTRAINT_NAME, ST_INDEX_ID, ST_INDEX_NAME, ST_MODULE_NAME, ST_RESERVED_SEQUENCE_RANGE,
+        ST_ROW_LEVEL_SECURITY_ID, ST_ROW_LEVEL_SECURITY_NAME, ST_SCHEDULED_ID, ST_SCHEDULED_NAME, ST_SEQUENCE_ID,
+        ST_SEQUENCE_NAME, ST_TABLE_NAME, ST_VAR_ID, ST_VAR_NAME, ST_VIEW_COLUMN_ID, ST_VIEW_COLUMN_NAME, ST_VIEW_ID,
+        ST_VIEW_NAME, ST_VIEW_PARAM_ID, ST_VIEW_PARAM_NAME,
     };
     use crate::traits::{IsolationLevel, MutTx};
     use crate::Result;
@@ -1265,7 +1266,7 @@ mod tests {
     use spacetimedb_lib::error::ResultTest;
     use spacetimedb_lib::st_var::StVarValue;
     use spacetimedb_lib::{resolved_type_via_v9, ScheduleAt, TimeDuration};
-    use spacetimedb_primitives::{col_list, ColId, ScheduleId};
+    use spacetimedb_primitives::{col_list, ColId, ScheduleId, ViewId};
     use spacetimedb_sats::algebraic_value::ser::value_serialize;
     use spacetimedb_sats::bsatn::ToBsatn;
     use spacetimedb_sats::layout::RowTypeLayout;
@@ -1704,6 +1705,10 @@ mod tests {
             TableRow { id: ST_SCHEDULED_ID.into(), name: ST_SCHEDULED_NAME, ty: StTableType::System, access: StAccess::Public, primary_key: Some(StScheduledFields::ScheduleId.into()) },
             TableRow { id: ST_ROW_LEVEL_SECURITY_ID.into(), name: ST_ROW_LEVEL_SECURITY_NAME, ty: StTableType::System, access: StAccess::Public, primary_key: Some(StRowLevelSecurityFields::Sql.into()) },
             TableRow { id: ST_CONNECTION_CREDENTIALS_ID.into(), name: ST_CONNECTION_CREDENTIALS_NAME, ty: StTableType::System, access: StAccess::Private, primary_key: Some(StConnectionCredentialsFields::ConnectionId.into()) },
+            TableRow { id: ST_VIEW_ID.into(), name: ST_VIEW_NAME, ty: StTableType::System, access: StAccess::Public, primary_key: Some(StViewFields::ViewId.into()) },
+            TableRow { id: ST_VIEW_PARAM_ID.into(), name: ST_VIEW_PARAM_NAME, ty: StTableType::System, access: StAccess::Public, primary_key: None },
+            TableRow { id: ST_VIEW_COLUMN_ID.into(), name: ST_VIEW_COLUMN_NAME, ty: StTableType::System, access: StAccess::Public, primary_key: None },
+
         ]));
         #[rustfmt::skip]
         assert_eq!(query.scan_st_columns()?, map_array([
@@ -1762,6 +1767,22 @@ mod tests {
 
             ColRow { table: ST_CONNECTION_CREDENTIALS_ID.into(), pos: 0, name: "connection_id", ty: AlgebraicType::U128 },
             ColRow { table: ST_CONNECTION_CREDENTIALS_ID.into(), pos: 1, name: "jwt_payload", ty: AlgebraicType::String },
+
+            ColRow { table: ST_VIEW_ID.into(), pos: 0, name: "view_id", ty: ViewId::get_type() },
+            ColRow { table: ST_VIEW_ID.into(), pos: 1, name: "view_name", ty: AlgebraicType::String },
+            ColRow { table: ST_VIEW_ID.into(), pos: 2, name: "table_id", ty: AlgebraicType::option(TableId::get_type()) },
+            ColRow { table: ST_VIEW_ID.into(), pos: 3, name: "is_public", ty: AlgebraicType::Bool },
+            ColRow { table: ST_VIEW_ID.into(), pos: 4, name: "is_anonymous", ty: AlgebraicType::Bool },
+
+            ColRow { table: ST_VIEW_PARAM_ID.into(), pos: 0, name: "view_id", ty: ViewId::get_type() },
+            ColRow { table: ST_VIEW_PARAM_ID.into(), pos: 1, name: "param_pos", ty: ColId::get_type() },
+            ColRow { table: ST_VIEW_PARAM_ID.into(), pos: 2, name: "param_name", ty: AlgebraicType::String },
+            ColRow { table: ST_VIEW_PARAM_ID.into(), pos: 3, name: "param_type", ty: AlgebraicType::bytes() },
+
+            ColRow { table: ST_VIEW_COLUMN_ID.into(), pos: 0, name: "view_id", ty: ViewId::get_type() },
+            ColRow { table: ST_VIEW_COLUMN_ID.into(), pos: 1, name: "col_pos", ty: ColId::get_type() },
+            ColRow { table: ST_VIEW_COLUMN_ID.into(), pos: 2, name: "col_name", ty: AlgebraicType::String },
+            ColRow { table: ST_VIEW_COLUMN_ID.into(), pos: 3, name: "col_type", ty: AlgebraicType::bytes() },
         ]));
         #[rustfmt::skip]
         assert_eq!(query.scan_st_indexes()?, map_array([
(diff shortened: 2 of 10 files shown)


crates/bindings-sys/src/lib.rs
```diff
@@ -588,6 +588,15 @@ pub mod raw {
...
         pub fn identity(out_ptr: *mut u8);
+
+        /// Check the size of the jwt associated with the given connection.
+        /// Returns 0 if there is no jwt for the connection.
+        pub fn jwt_len(connection_id_ptr: *const u8, out_ptr: *mut u32);
+
+        /// Write the jwt payload for the given connection id to the out_ptr.
+        /// target_ptr_len will be set to the number of bytes written to the buffer.
+        /// If the buffer is too small, or the connection has no jwt, target_ptr_len will be set to 0.
+        pub fn get_jwt(connection_id_ptr: *const u8, target_ptr: *mut u8, target_ptr_len: *mut u32);
```
gefjon: I am surprised that we don't have `get_jwt` return a `BytesSource` which we then call `bytes_source_read` from. I would prefer not to proliferate `read`-like host functions. This would require amending `bytes_source_read` to support `BytesSource`s other than `CALL_REDUCER_ARGS_SOURCE`, and probably also adding `bytes_source_len`, but that seems like valuable work that will need to get done at some point anyways. Notably I will need to do this when I implement HTTP requests in procedures in order to return the response. Because I'm going to be doing it anyways, I could write a patch tomorrow which makes the necessary extensions to `bytes_source_read`.

crates/bindings-sys/src/lib.rs
```diff
@@ -1089,6 +1098,34 @@ pub fn identity() -> [u8; 32] {
...
+    } else {
+        Some(v)
+    }
+}
+
+#[inline]
+pub fn get_jwt(connection_id: [u8; 16]) -> Option<String> {
+    let Some(jwt_len) = jwt_length(connection_id) else {
+        return None; // No JWT found.
+    };
```
gefjon:
```suggestion
    let jwt_len = jwt_length(connection_id)?;
```

crates/core/src/host/wasm_common.rs
```diff
@@ -392,6 +392,8 @@ macro_rules! abi_funcs {
             "spacetime_10.0"::datastore_btree_scan_bsatn,
             "spacetime_10.0"::datastore_delete_by_btree_scan_bsatn,
             "spacetime_10.0"::identity,
+            "spacetime_10.0"::get_jwt,
+            "spacetime_10.0"::jwt_len,
```
gefjon: These should be in a new ABI version and module, `spacetime_11.0`. See comment in `bindings_sys/src/lib.rs` around line 20.

gefjon: Ah, looks like this should be `spacetime_10.1`, actually.

crates/bindings-sys/src/lib.rs
gefjon: I opened #3294 to support this.

crates/core/src/host/wasm_common.rs
gefjon: See #3294 for an example of adding a new host function with a new minor version. These can all go in `10.1` so long as we don't cut a release in between the two PRs.

crates/bindings-sys/src/lib.rs
```diff
@@ -617,6 +618,11 @@ pub mod raw {
         ///
         /// If this function returns an error, `out` is not written.
         pub fn bytes_source_remaining_length(source: BytesSource, out: *mut u32) -> i16;
+
+        /// Find the jwt payload for the given connection id, and write the
+        /// BytesSourceId to the given pointer.
+        /// If this is not found, BytesSourceId::INVALID (aka 0) will be written.
+        pub fn get_jwt(connection_id_ptr: *const u8, bytes_source_id: *mut BytesSource);
```
Centril:
```suggestion
        /// Find the jwt payload for the given connection id, and write the
        /// [`BytesSourceId`] to the given pointer.
        /// If this is not found, [`BytesSourceId::INVALID`] (aka 0) will be written.
        pub fn get_jwt(connection_id_ptr: *const u8, bytes_source_id: *mut BytesSource);
```

crates/bindings-sys/src/lib.rs
```diff
@@ -1118,6 +1124,34 @@ pub fn identity() -> [u8; 32] {
...
+        let mut len = 0;
+        let ret = unsafe { raw::bytes_source_remaining_length(source, &raw mut len) };
+        match ret {
+            0 => len,
+            _ => panic!("invalid source"),
+        }
+    };
+    let mut buf = vec![0u8; len as usize];
+    let mut bytes_read = len as usize;
+    let ret = unsafe { raw::bytes_source_read(source, buf.as_mut_ptr(), &mut bytes_read) };
```
Centril: `IterBuf::take()` can be used to get a buffer that's most likely sitting around instead of allocating here. Then you can use `read_bytes_source_into` as well.
You'll need to move this version of `get_jwt` to `bindings`, and then you can define here:
```rust
pub fn get_jwt(connection_id: [u8; 16]) -> raw::BytesSource { ... }
```

crates/bindings/src/lib.rs
```diff
@@ -796,6 +821,114 @@ impl DbContext for ReducerContext {
...
+    where
+        F: FnOnce() -> Option<JwtClaims> + 'static,
+    {
+        AuthCtx {
+            is_internal,
+            jwt: Box::new(LazyLock::new(jwt_fn)),
+        }
+    }
+
+    /// Create an AuthCtx for an internal call, with no JWT.
```
Centril:
```suggestion
    /// Creates an `AuthCtx` for an internal call, with no JWT.
```

crates/bindings/src/lib.rs
```diff
@@ -796,6 +821,114 @@ impl DbContext for ReducerContext {
...
+    /// This represents a scheduled reducer.
+    pub fn internal() -> AuthCtx {
+        Self::new(true, || None)
+    }
+
+    /// Create an AuthCtx using the json claims from a JWT.
```
Centril:
```suggestion
    /// Creates an `AuthCtx` using the json claims from a JWT.
```

crates/bindings/src/lib.rs
```diff
@@ -796,6 +821,114 @@ impl DbContext for ReducerContext {
...
+    /// This can be used to write unit tests.
+    pub fn from_jwt_payload(jwt_payload: String) -> AuthCtx {
+        Self::new(false, move || Some(JwtClaims::new(jwt_payload)))
+    }
+
+    /// Create an AuthCtx that reads the JWT for the given connection id.
```
Centril:
```suggestion
    /// Creates an `AuthCtx` that reads the JWT for the given connection id.
```

crates/bindings/src/lib.rs
```diff
@@ -796,6 +821,114 @@ impl DbContext for ReducerContext {
...
+    fn from_connection_id(connection_id: ConnectionId) -> AuthCtx {
+        Self::new(false, move || {
+            spacetimedb_bindings_sys::get_jwt(connection_id.as_le_byte_array()).map(JwtClaims::new)
+        })
+    }
+
+    /// True if this reducer was spawned from inside the database.
```
Centril:
```suggestion
    /// Returns whether this reducer was spawned from inside the database.
```

crates/bindings/src/lib.rs
```diff
@@ -796,6 +821,114 @@ impl DbContext for ReducerContext {
...
+    is_internal: bool,
+    // I can't directly use a LazyLock without making this struct generic.
+    jwt: Box<dyn Deref<Target = Option<JwtClaims>>>,
+}
+
+impl AuthCtx {
+    fn new<F>(is_internal: bool, jwt_fn: F) -> Self
+    where
+        F: FnOnce() -> Option<JwtClaims> + 'static,
+    {
```
Centril:
```suggestion
    fn new(
        is_internal: bool,
        jwt_fn: impl FnOnce() -> Option<JwtClaims> + 'static
    ) -> Self {
```

crates/bindings/src/lib.rs
```diff
@@ -796,6 +821,114 @@ impl DbContext for ReducerContext {
...
+        })
+    }
+
+    /// True if this reducer was spawned from inside the database.
+    pub fn is_internal(&self) -> bool {
+        self.is_internal
+    }
+
+    /// Check if there is a JWT without loading it.
+    /// If is_internal is true, this will be false.
```
Centril:
```suggestion
    /// If [`AuthCtx::is_internal`] returns true, this will return false.
```

crates/bindings/src/lib.rs
```diff
@@ -796,6 +821,114 @@ impl DbContext for ReducerContext {
...
+pub struct JwtClaims {
+    payload: String,
+    parsed: OnceCell<serde_json::Value>,
+    audience: OnceCell<Vec<String>>,
+}
+
+/// Authentication information for the caller of a reducer.
+pub struct AuthCtx {
+    is_internal: bool,
+    // I can't directly use a LazyLock without making this struct generic.
```
Centril:
```suggestion
    // NOTE(jsdt): cannot directly use a LazyLock without making this struct generic.
    jwt: Box<dyn Deref<Target = Option<JwtClaims>>>,
```

crates/bindings/src/lib.rs
```diff
@@ -796,6 +821,114 @@ impl DbContext for ReducerContext {
...
+}
+
+impl AuthCtx {
+    fn new<F>(is_internal: bool, jwt_fn: F) -> Self
+    where
+        F: FnOnce() -> Option<JwtClaims> + 'static,
+    {
+        AuthCtx {
+            is_internal,
+            jwt: Box::new(LazyLock::new(jwt_fn)),
```
Centril:
```suggestion
            jwt: Box::new(LazyCell::new(jwt_fn)),
```
`ReducerContext` uses `OnceCell` directly, so it's already `!Sync`.

crates/bindings/src/lib.rs
```diff
@@ -796,6 +821,114 @@ impl DbContext for ReducerContext {
...
+        }
+    }
+
+    fn get_parsed(&self) -> &serde_json::Value {
+        self.parsed
+            .get_or_init(|| serde_json::from_str(&self.payload).expect("Failed to parse JWT payload"))
+    }
+
+    pub fn subject(&self) -> &str {
+        // TODO: Add more error messages here.
```
Centril: This and similar needs to be done before we release `JwtClaims` / `AuthCtx` as we won't be able to change it after.

crates/bindings/src/lib.rs
```diff
@@ -796,6 +821,114 @@ impl DbContext for ReducerContext {
...
+            audience: OnceCell::new(),
+        }
+    }
+
+    fn get_parsed(&self) -> &serde_json::Value {
+        self.parsed
+            .get_or_init(|| serde_json::from_str(&self.payload).expect("Failed to parse JWT payload"))
+    }
+
+    pub fn subject(&self) -> &str {
```
Centril, line 929: User facing docs?

crates/bindings/src/lib.rs
```diff
@@ -796,6 +821,114 @@ impl DbContext for ReducerContext {
...
+        // TODO: Add more error messages here.
+        self.get_parsed().get("sub").unwrap().as_str().unwrap()
+    }
+
+    pub fn issuer(&self) -> &str {
```
Centril, line 938: User facing docs?

crates/bindings/src/lib.rs
```diff
@@ -796,6 +821,114 @@ impl DbContext for ReducerContext {
...
+    fn extract_audience(&self) -> Vec<String> {
+        let aud = self.get_parsed().get("aud").unwrap();
+        match aud {
+            serde_json::Value::String(s) => vec![s.clone()],
+            serde_json::Value::Array(arr) => arr.iter().filter_map(|v| v.as_str().map(String::from)).collect(),
+            _ => panic!("Unexpected type for 'aud' claim in JWT"),
+        }
+    }
+
+    pub fn audience(&self) -> &[String] {
```
Centril, line 952: User facing docs?

crates/bindings/src/lib.rs
```diff
@@ -796,6 +821,114 @@ impl DbContext for ReducerContext {
...
+        self.audience.get_or_init(|| self.extract_audience())
+    }
+
+    // A convenience method, since this may not be in the token.
+    pub fn identity(&self) -> Identity {
```
Centril, line 958: User facing docs?

crates/bindings/src/lib.rs
```diff
@@ -796,6 +821,114 @@ impl DbContext for ReducerContext {
...
+        Identity::from_claims(self.issuer(), self.subject())
+    }
+
+    // We can expose the whole payload for users that want to parse custom claims.
+    pub fn raw_payload(&self) -> &str {
```
Centril, line 963: User facing docs?

crates/bindings/Cargo.toml
```diff
@@ -35,6 +35,8 @@ rand08 = { workspace = true, optional = true }
 # if someone tries to use rand's ThreadRng, it will fail to link
 # because no one defined __getrandom_custom
 getrandom02 = { workspace = true, optional = true, features = ["custom"] }
+serde_json.workspace = true
+once_cell = "1.21.3"
```
Centril: This is part of the standard library now.

```suggestion
```

crates/bindings/Cargo.toml
```diff
@@ -35,6 +35,8 @@ rand08 = { workspace = true, optional = true }
 # if someone tries to use rand's ThreadRng, it will fail to link
 # because no one defined __getrandom_custom
 getrandom02 = { workspace = true, optional = true, features = ["custom"] }
+serde_json.workspace = true
```
Centril, line 38: Should we make this dependency optional to keep compile times shorter for those making modules and who don't care about JWT? That is, should this be gated under `feature = "jwt"`?

crates/bindings-sys/src/lib.rs
Centril: This should follow the above documentation style in terms of ABI and traps and errors.

crates/core/src/host/wasmtime/wasm_instance_env.rs
```diff
@@ -1354,6 +1355,30 @@ impl WasmInstanceEnv {
         })
     }
 
+    pub fn get_jwt(
```
Centril, line 1329: This needs corresponding docs in the style of e.g., `table_id_from_name` including traps and errors (e.g., `NOT_IN_TRANSACTION` applies here).

crates/core/src/host/wasmtime/wasm_instance_env.rs
```diff
@@ -1354,6 +1355,30 @@ impl WasmInstanceEnv {
...
+        caller: Caller<'_, Self>,
+        connection_id: WasmPtr<ConnectionId>,
+        target_ptr: WasmPtr<u32>,
+    ) -> RtResult<()> {
+        log::info!("Calling get_jwt");
+        Self::with_span(caller, AbiCall::GetJwt, |caller| {
```
Centril: Looks like this could use `cvt_ret`

Cargo.lock
```diff
@@ -13,18 +13,18 @@ dependencies = [
 
 [[package]]
```
Centril, line 14: Do we want to `cargo update` in this PR?

crates/bindings/Cargo.toml
jsdt: That's something we could consider later if we get requests for it, but since the vast majority of modules should care about JWTs, I'd rather not add any extra complexity.

crates/core/src/host/wasmtime/wasm_instance_env.rs
jsdt: I updated it to clarify that it traps for errors. Returning an integer error code doesn't seem very useful.

crates/core/src/host/wasmtime/wasm_instance_env.rs
jsdt: It probably could, but it using that function just seemed to make the error handling more difficult to reason about.

crates/core/src/host/wasmtime/wasm_instance_env.rs
Centril: That is the design of the overall ABI that `NOT_IN_TRANSACTION` should be returned as a code and not a trap, and I think we should follow the general ABI design.

crates/core/src/host/wasmtime/wasm_instance_env.rs
```diff
@@ -1354,6 +1355,41 @@ impl WasmInstanceEnv {
...
+    /// A `[ByteSourceId]` for the payload will be written to `target_ptr`.
+    /// If nothing is found for the connection, `[ByteSourceId::INVALID]` (zero) is written to `target_ptr`.
+    ///
+    /// This must be called inside a transaction (because it reads from a system table).
+    ///
+    /// # Traps
+    ///
+    /// Traps if:
+    ///
+    /// - `connection_id` does not point to a valid little-endian `ConnectionId`.
```
Centril, line 1326:
```suggestion
    /// - `connection_id` does not point to a valid little-endian `ConnectionId`.
    /// - `target_ptr` is NULL or `target_ptr[..size_of::<u32>()]` is not in bounds of WASM memory.
    /// - The `ByteSourceId` to be written to `target_ptr` would overflow [`u32::MAX`].
```

(and in bindings-sys too)

crates/bindings/src/rt.rs
```diff
@@ -531,6 +526,20 @@ fn with_read_args<R>(args: BytesSource, logic: impl FnOnce(&[u8]) -> R) -> R {
...
+/// Look up the jwt associated with `connection_id`.
+pub fn get_jwt(connection_id: ConnectionId) -> Option<String> {
+    let mut buf = IterBuf::take();
+    let mut source: BytesSource = BytesSource::INVALID;
+    unsafe {
+        raw::get_jwt(connection_id.as_le_byte_array().as_ptr(), &mut source);
+    };
+    if source == BytesSource::INVALID {
+        return None;
+    }
```
Centril:
```suggestion
    let source = sys::get_jwt(connection_id.as_le_byte_array())?;
```

Centril (approved):
I have some improvements atop of this PR, primarily wrt. docs, but this looks good to merge.

crates/core/src/host/wasm_common.rs
jsdt: I put this in `spacetime_10.2`, since it looks like `10.1` was released.

modules/quickstart-chat/src/lib.rs
```diff
@@ -67,6 +67,10 @@ pub fn init(_ctx: &ReducerContext) {}
 
 #[spacetimedb::reducer(client_connected)]
 pub fn identity_connected(ctx: &ReducerContext) {
+    log::info!(
+        "Client connected with jwt payload {:?}",
+        ctx.sender_auth().jwt().unwrap().raw_payload()
+    );
```
gefjon: Do we intend to add this to the quickstart-chat tutorial? If not, it shouldn't be in here.

jsdt: Any objection to adding to the tutorial code as an illustration?

I can change it to avoid the unwrap:
```
    if let Some(jwt) = ctx.sender_auth().jwt() {
        log::info!("Client connected with jwt payload {}", jwt.raw_payload());
    }
```

gefjon: I'd kinda prefer not to, since this adds additional complexity to a tutorial that's already too long, and doesn't add any functionality to the application itself. I think it would be better to demo this functionality in a separate document that says what JWT claims are and where they come from, and shows doing some useful thing with them.

crates/bindings/src/lib.rs
jsdt: We can change the error messages in the future without it breaking the API. I'm not suggesting this should return a `Result`, since we should verify these tokens before storing them in system tables.

PR #3294, opened 2025-09-25
Support `BytesSource`s other than the one for reducer args

gefjon:
# Description of Changes

Each reducer gets its arguments through an `ArgSource`, a Unix-file-like abstraction for streams of bytes. Prior to this commit, we had an ABI designed as if it could support other args sources, but it actually hardcoded the ID of the reducer args source, and errored elsewhere.

This commit extends the `BytesSource` infrastructure to support other bytes sources. This will be useful for exposing JWT payloads and HTTP responses. No other `BytesSource` uses are actually included in this commit, only the infrastructure.

This commit also defines a new host call, `bytes_source_remaining_length`. This is intended to allow callers to pre-allocate a buffer correctly sized to read the entire `BytesSource` all at once. The new host function is added to a new ABI minor version, 10.1, so that old SpacetimeDB hosts can detect and reject too-new compiled modules. I have added uses of this new function to `__call_reducer__` in both Rust and C#, even though it's not strictly necessary,
and I haven't removed the loop which repeatedly calls `bytes_source_read` and grows the buffer.

Adds a new ABI minor version, `spacetime_10.1`. This means that old SpacetimeDB hosts will reject new compiled modules.

2-ish? WASM ABI code is always fiddly, but this is a pretty simple case.

# Testing

- [x] New behavior and new host function are both hit through existing tests that instantiate modules and call reducers against them, so I believe automated testing is sufficient.

### DIFF
--- a/crates/bindings/src/rt.rs
+++ b/crates/bindings/src/rt.rs
@@ -4,7 +4,7 @@ use crate::table::IndexAlgo;
 use crate::{sys, IterBuf, ReducerContext, ReducerResult, SpacetimeType, Table};
 pub use spacetimedb_lib::db::raw_def::v9::Lifecycle as LifecycleReducer;
 use spacetimedb_lib::db::raw_def::v9::{RawIndexAlgorithm, RawModuleDefV9Builder, TableType};
-use spacetimedb_lib::de::{self, Deserialize, SeqProductAccess};
+use spacetimedb_lib::de::{self, Deserialize, Error as _, SeqProductAccess};
 use spacetimedb_lib::sats::typespace::TypespaceBuilder;
 use spacetimedb_lib::sats::{impl_deserialize, impl_serialize, ProductTypeElement};
 use spacetimedb_lib::ser::{Serialize, SerializeSeqProduct};
@@ -202,7 +202,7 @@ impl<'de, A: Args<'de>> de::ProductVisitor<'de> for ArgsVisitor<A> {
         A::visit_seq_product(prod)
     }
     fn visit_named_product<Acc: de::NamedProductAccess<'de>>(self, _prod: Acc) -> Result<Self::Output, Acc::Error> {
-        Err(de::Error::custom("named products not supported"))
+        Err(Acc::Error::named_products_not_supported())
     }
 }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6250,6 +6250,7 @@ dependencies = [
  "rand 0.9.1",
  "second-stack",
  "serde",
+ "serde_json",
  "sha3",
  "smallvec",
  "spacetimedb-bindings-macro 1.4.0",
--- a/crates/bindings/tests/ui/tables.stderr
+++ b/crates/bindings/tests/ui/tables.stderr
@@ -61,11 +61,11 @@ error[E0277]: the trait bound `Test: Deserialize<'de>` is not satisfied
             &'de [u8]
             &'de str
             ()
-            AlgebraicType
-            AlgebraicTypeRef
-            Alpha
-            Arc<[T]>
-            ArrayType
+            (T0, T1)
+            (T0, T1, T2)
+            (T0, T1, T2, T3)
+            (T0, T1, T2, T3, T4)
+            (T0, T1, T2, T3, T4, T5)
           and $N others
 note: required by a bound in `spacetimedb::spacetimedb_lib::de::SeqProductAccess::next_element`
  --> $WORKSPACE/crates/sats/src/de.rs
@@ -83,11 +83,11 @@ error[E0277]: the trait bound `Test: Deserialize<'_>` is not satisfied
             &'de [u8]
             &'de str
             ()
-            AlgebraicType
-            AlgebraicTypeRef
-            Alpha
-            Arc<[T]>
-            ArrayType
+            (T0, T1)
+            (T0, T1, T2)
+            (T0, T1, T2, T3)
+            (T0, T1, T2, T3, T4)
+            (T0, T1, T2, T3, T4, T5)
           and $N others
 note: required by a bound in `get_field_value`
  --> $WORKSPACE/crates/sats/src/de.rs
@@ -104,12 +104,12 @@ error[E0277]: the trait bound `Test: Serialize` is not satisfied
   = help: the following other types implement trait `Serialize`:
             &T
             ()
-            AlgebraicType
-            AlgebraicTypeRef
-            AlgebraicValue
-            Alpha
-            Arc<T>
-            ArrayType
+            (T0, T1)
+            (T0, T1, T2)
+            (T0, T1, T2, T3)
+            (T0, T1, T2, T3, T4)
+            (T0, T1, T2, T3, T4, T5)
+            (T0, T1, T2, T3, T4, T5, T6)
           and $N others
 note: required by a bound in `spacetimedb::spacetimedb_lib::ser::SerializeNamedProduct::serialize_element`
  --> $WORKSPACE/crates/sats/src/ser.rs
--- a/crates/sats/Cargo.toml
+++ b/crates/sats/Cargo.toml
@@ -69,6 +69,8 @@ rand.workspace = true
 # Also as dev-dependencies for use in _this_ crate's tests.
 proptest.workspace = true
 proptest-derive.workspace = true
+serde_json.workspace = true
+serde.workspace = true
 
 [lints]
 workspace = true
--- a/crates/sats/src/de.rs
+++ b/crates/sats/src/de.rs
@@ -2,7 +2,7 @@
 // See `serde` version `v1.0.169` for the parts where MIT / Apache-2.0 applies.
 
 mod impls;
-#[cfg(feature = "serde")]
+#[cfg(any(test, feature = "serde"))]
 pub mod serde;
 
 #[doc(hidden)]
@@ -158,6 +158,11 @@ pub trait Deserializer<'de>: Sized {
 pub trait Error: Sized {
     /// Raised when there is general error when deserializing a type.
     fn custom(msg: impl fmt::Display) -> Self;
+
+    /// Deserializing named products are not supported for this visitor.
+    fn named_products_not_supported() -> Self {
+        Self::custom("named products not supported")
+    }
 
     /// The product length was not as promised.
     fn invalid_product_length<'de, T: ProductVisitor<'de>>(len: usize, expected: &T) -> Self {
--- a/crates/sats/src/de/impls.rs
+++ b/crates/sats/src/de/impls.rs
@@ -8,7 +8,7 @@ use crate::{
     SumValue, WithTypespace, F32, F64,
 };
 use crate::{i256, u256};
-use core::{marker::PhantomData, ops::Bound};
+use core::{iter, marker::PhantomData, ops::Bound};
 use smallvec::SmallVec;
 use spacetimedb_primitives::{ColId, ColList};
 use std::{borrow::Cow, rc::Rc, sync::Arc};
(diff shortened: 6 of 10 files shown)


crates/core/src/host/wasmtime/wasm_instance_env.rs
```diff
@@ -1046,7 +1122,7 @@ impl WasmInstanceEnv {
 
             // Derive the portion that we can read and what remains,
             // based on what is left to read and the capacity.
-            let left_to_read = &reducer_args[*cursor..];
+            let left_to_read = &bytes_source.bytes[bytes_source.cursor..];
```
jsdt: `bytes::Bytes` seems to already be a view/slice of underlying bytes, so I don't think we need to track a cursor.
```
            let can_read_len = buffer_len.min(bytes_source.bytes.len());
            // Take the portion that we can read off of the bytes.
            let can_read = bytes_source.bytes.split_to(can_read_len);
            buffer[..can_read_len].copy_from_slice(&can_read);
            (can_read_len as u32).write_to(mem, buffer_len_ptr)?;
            return if bytes_source.bytes.is_empty() {
                env.free_bytes_source(source);
                Ok(-1i32)
            } else {
                Ok(0)
            };
```

crates/core/src/host/wasmtime/wasm_instance_env.rs
```diff
@@ -1055,15 +1131,59 @@ impl WasmInstanceEnv {
...
+    /// If this function returns an error, `out` is not written.
+    pub fn bytes_source_remaining_length(caller: Caller<'_, Self>, source: u32, out: WasmPtr<u32>) -> RtResult<i32> {
+        Self::cvt_custom(caller, AbiCall::BytesSourceRemainingLength, |caller| {
+            let (mem, env) = Self::mem_env(caller);
+
+            let Some(bytes_source) = env.bytes_sources.get(&BytesSourceId(source)) else {
+                return Ok(errno::NO_SUCH_BYTES.get().into());
+            };
+
+            let total_len = bytes_source.bytes.len();
```
jsdt: If we just let the `bytes::Bytes` instance track the offset, this can be simplified to:

```
let remaining: u32 = bytes_source.bytes.len().try_into()
    .context("Bytes object in `BytesSource` had length greater than range of u32")?;
```

crates/core/src/host/wasmtime/wasm_instance_env.rs
gefjon: We kept a cursor before, and I just copied that code over unchanged. Your version is better, I'll update.

crates/bindings-csharp/Runtime/Internal/FFI.cs
```diff
@@ -291,7 +299,13 @@ uint args_len
     // which prevents source-generated PInvokes from working with types from other assemblies, and
     // `Identity` lives in another assembly (`BSATN.Runtime`). Luckily, `DllImport` is enough here.
 #pragma warning disable SYSLIB1054 // Suppress "Use 'LibraryImportAttribute' instead of 'DllImportAttribute'" warning.
-    [DllImport(StdbNamespace)]
+    [DllImport(StdbNamespace10_0)]
     public static extern void identity(out Identity dest);
 #pragma warning restore SYSLIB1054
+
+    [DllImport(StdbNamespace10_1)]
+    public static partial Errno bytes_source_remaining_length(
```
rekhoff: I reproduced these the current FFI errors locally.
I believe the current compile errors are due to the `DllImport` attribute requires the `static extern` keywords rather than the `partial` keyword.
```suggestion
    public static extern Errno bytes_source_remaining_length(
```
Testing this change locally allows `dotnet build` to run and `dotnet test` successfully runs against the `Codegen.Tests` project without errors.

crates/core/src/host/wasmtime/wasm_instance_env.rs
```diff
@@ -1046,24 +1124,66 @@ impl WasmInstanceEnv {
...
+    /// before invoking this function on the same `source`.
+    ///
+    /// Also note that the special [`BytesSourceId::INVALID`] (zero) is always invalid.
+    /// Callers should check for that value before invoking this function.
+    ///
+    /// # Traps
+    ///
+    /// Traps if:
+    ///
+    /// - `out` is NULL or `ou` is not in bounds of WASM memory.
```
Centril:
```suggestion
    /// - `out` is NULL or `out` is not in bounds of WASM memory.
```

crates/core/src/host/wasmtime/wasm_instance_env.rs
```diff
@@ -1046,24 +1124,66 @@ impl WasmInstanceEnv {
...
+
+            let Some(bytes_source) = env.bytes_sources.get(&BytesSourceId(source)) else {
+                return Ok(errno::NO_SUCH_BYTES.get().into());
+            };
+
+            let remaining: u32 = bytes_source
+                .bytes
+                .len()
+                .try_into()
+                .context("Bytes object in `BytesSource` had length greater than range of u32")?;
```
Centril, line 1192: I think in this case we shouldn't trap and instead return that the length is unknown, e.g., `BYTES_LEN_UNKNOWN`. This has the added benefit of retaining `bytes_source_read`s ability to support possibly infinite sources, which we currently don't need, but who knows what the future holds...

crates/core/src/host/wasmtime/wasm_instance_env.rs
```diff
@@ -1046,24 +1124,66 @@ impl WasmInstanceEnv {
...
+
+            u32::write_to(remaining, mem, out)
+                .context("Failed to write output from `bytes_source_remaining_length`")?;
```
Centril, line 1195: Note to self: it's time to refactor `WasmError` to support returning `errno::NO_SUCH_BYTES` in it so we can use `cvt_ret` instead.

crates/core/src/host/wasmtime/wasm_instance_env.rs
```diff
@@ -23,6 +25,45 @@ use instrumentation::noop as span;
...
+    /// When this becomes empty, this `ByteSource` is expended and should be discarded.
+    bytes: bytes::Bytes,
+}
+
+#[derive(Copy, Clone, PartialEq, Eq, Debug)]
+/// Identifier for a [`BytesSource`] stored in the `bytes_sources` of a [`WasmInstanceEnv`].
+///
+/// The special sentinel [`Self::INVALID`] (zero) is used for a never-readable [`BytesSource`].
+/// We pass this to guests for a [`BytesSource`] with a length of zero.
+pub(super) struct BytesSourceId(pub(super) u32);
```
Centril: (Style nit: `#[derive(...)]` typically comes after doc comments.)

```suggestion
/// Identifier for a [`BytesSource`] stored in the `bytes_sources` of a [`WasmInstanceEnv`].
///
/// The special sentinel [`Self::INVALID`] (zero) is used for a never-readable [`BytesSource`].
/// We pass this to guests for a [`BytesSource`] with a length of zero
/// so that they can avoid host calls.
#[derive(Copy, Clone, PartialEq, Eq, Debug)]
pub(super) struct BytesSourceId(pub(super) u32);
```

crates/core/src/host/wasmtime/wasm_instance_env.rs
```diff
@@ -102,6 +153,38 @@ impl WasmInstanceEnv {
         }
     }
 
+    fn alloc_bytes_source_id(&mut self) -> RtResult<BytesSourceId> {
+        let id = self.next_bytes_source_id;
+        self.next_bytes_source_id = id
+            .checked_add(1)
+            .context("Allocating next `BytesSourceId` overflowed u32")?;
```
Centril:
```suggestion
            .context("Allocating next `BytesSourceId` overflowed `u32`")?;
```

crates/core/src/host/wasmtime/wasm_instance_env.rs
```diff
@@ -102,6 +153,38 @@ impl WasmInstanceEnv {
...
+    }
+
+    fn create_bytes_source(&mut self, bytes: bytes::Bytes) -> RtResult<BytesSourceId> {
+        // Pass an invalid source when the bytes were empty.
+        // This allows the module to avoid allocating and make a system call in those cases.
+        if bytes.is_empty() {
+            Ok(BytesSourceId::INVALID)
+        } else if bytes.len() > u32::MAX as usize {
+            Err(anyhow::anyhow!(
+                "`create_bytes_source`: `Bytes` has length {}, which is greater than u32::MAX {}",
```
Centril:
```suggestion
                "`create_bytes_source`: `Bytes` has length {}, which is greater than `u32::MAX` {}",
```

crates/core/src/host/wasmtime/wasm_instance_env.rs
```diff
@@ -102,6 +153,38 @@ impl WasmInstanceEnv {
...
 
+    fn alloc_bytes_source_id(&mut self) -> RtResult<BytesSourceId> {
+        let id = self.next_bytes_source_id;
+        self.next_bytes_source_id = id
+            .checked_add(1)
+            .context("Allocating next `BytesSourceId` overflowed u32")?;
+        Ok(BytesSourceId(id.into()))
+    }
+
+    fn create_bytes_source(&mut self, bytes: bytes::Bytes) -> RtResult<BytesSourceId> {
```
Centril, line 168:
```suggestion
    /// Binds `bytes` to the environment and assigns it an ID.
    ///
    /// If `bytes` is empty, `BytesSourceId::INVALID` is returned.
    fn create_bytes_source(&mut self, bytes: bytes::Bytes) -> RtResult<BytesSourceId> {
```

crates/bindings/src/rt.rs
```diff
@@ -535,6 +535,18 @@ const NO_SUCH_BYTES: u16 = errno::NO_SUCH_BYTES.get();
...
 
+    let len = {
+        let mut len = 0;
+        let ret = unsafe { sys::raw::bytes_source_remaining_length(source, &raw mut len) };
+        match ret {
+            0 => len,
+            INVALID => panic!("invalid source passed"),
+            _ => unreachable!(),
+        }
+    };
```
Centril, line 551: I think in 99% of cases, this is a performance regression, as the buffer used comes from `IterBuf::take()` which is 64KiB by default.


```suggestion
```

crates/bindings-sys/src/lib.rs
```diff
@@ -590,6 +590,35 @@ pub mod raw {
...
+        /// before invoking this function on the same `source`.
+        ///
+        /// Also note that the special [`BytesSource::INVALID`] (zero) is always invalid.
+        /// Callers should check for that value before invoking this function.
+        ///
+        /// # Traps
+        ///
+        /// Traps if:
+        ///
+        /// - `out` is NULL or `ou` is not in bounds of WASM memory.
```
Centril:
```suggestion
        /// - `out` is NULL or `out` is not in bounds of WASM memory.
```

crates/bindings/src/rt.rs
gefjon: In the specific case of reducer arguments, yes, but it's a small regression, and including it here makes this function usable for reading in other cases with buffers that aren't pre-allocated. I'm adding comments to this function and the C# version, but would prefer not to remove this.

crates/core/src/host/wasmtime/wasm_instance_env.rs
gefjon: I'm going to add this as a TODO, along with TODO notes on the various other places that error on too-large `BytesSource`s.

cloutiertyler:
Force merging for reasons discussed internally.

PR #3297, opened 2025-09-26
Make the .NET global.json file global

### LINKED ISSUE #3304: Fix C# bindings tests
Our C# bindings tests seem to be failing on warnings (that are being converted to errors). It's not super clear why this is just happening now.

kim:
Move the global.json file to the repo root, which should make `dotnet`
resolve the correct SDK regardless of the directory it is run in.

Where a copy of the global.json file is needed, symlink it to the one at
the root.

Should fix "C#/Unity - Test Suite".


.github/workflows/ci.yml
```diff
@@ -130,6 +130,10 @@ jobs:
       - uses: dsherret/rust-toolchain-file@v1
       - run: echo ::add-matcher::.github/workflows/rust_matcher.json
 
+      - uses: actions/setup-dotnet@v3
+        with:
+          global-json-file: global.json
```
bfops, line 135: interesting that this ever worked without a `setup-dotnet` step.. :monocle_face:

bfops (approved):
This LGTM and seems to fix the C# Test Suite at least. The Unity stuff can be a different PR.

bfops:
One concern for me is whether the template `global._json` will be correct in C# projects generated by the CLI, but since our CI passes I assume it's still working correctly!

bfops:
I'm going to merge this since it fixes the Test Suite.

PR #3301, opened 2025-09-29
Smoketest : Hotswap should not disconnect subscribers.

Shubham8287:
# Description of Changes

Updates smoketests to check for  following.
- Normal auto migration subscription should not disconnect subcriber.
- Add table columns migration should disconnect subscribers.

Solves #1957 .

### DIFF
--- a/smoketests/tests/auto_migration.py
+++ b/smoketests/tests/auto_migration.py
@@ -128,7 +128,7 @@ const BOOK_VISIBLE: spacetimedb::Filter = spacetimedb::Filter::Sql("SELECT * FRO
 
         logging.info("Initial publish complete")
 
-        // Start a subscription before publishing the module, to test that the subscription remains intact after re-publishing.
+        # Start a subscription before publishing the module, to test that the subscription remains intact after re-publishing.
         sub = self.subscribe("select * from person", n=4)
 
         # initial module code is already published by test framework
--- a/smoketests/tests/auto_migration.py
+++ b/smoketests/tests/auto_migration.py
@@ -127,8 +127,10 @@ const BOOK_VISIBLE: spacetimedb::Filter = spacetimedb::Filter::Sql("SELECT * FRO
         )
 
         logging.info("Initial publish complete")
-        # initial module code is already published by test framework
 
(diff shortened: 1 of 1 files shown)


docs/cli-reference.md
```diff
@@ -0,0 +1,596 @@
+# Command-Line Help for `spacetime`
```
bfops: this should probably be overwriting `docs/docs/cli-reference.md`

bfops: I'll fix the instructions in the docs directory, I think they assumed someone was already in `docs` :sweat_smile:

Shubham8287: yeah true, I think I messed up with git. this is 3rd stacked PR. will fix it.

bfops: created https://github.com/clockworklabs/SpacetimeDB/pull/3303

smoketests/tests/auto_migration.py
```diff
@@ -127,8 +127,10 @@ def test_add_table_auto_migration(self):
         )
 
         logging.info("Initial publish complete")
-        # initial module code is already published by test framework
 
+        sub = self.subscribe("select * from person", n=4)
```
bfops, line 132:
```suggestion
        // Start a subscription before publishing the module, to test that the subscription remains intact after re-publishing.
        sub = self.subscribe("select * from person", n=4)
```

bfops (approved):
LGTM (once tests pass and the `docs/cli-reference.md` thing is fixed).

Are we planning to wait until the base PRs merge? (Otherwise a reviewer of those PRs will have to re-review this).

smoketests/tests/auto_migration.py
```diff
@@ -127,8 +127,11 @@ def test_add_table_auto_migration(self):
         )
 
         logging.info("Initial publish complete")
-        # initial module code is already published by test framework
 
+        // Start a subscription before publishing the module, to test that the subscription remains intact after re-publishing.
```
Shubham8287:
```suggestion
        # Start a subscription before publishing the module, to test that the subscription remains intact after re-publishing.
```

PR #3312, opened 2025-09-29
Resolve misc warnings in C# bindings code

### LINKED ISSUE #3305: Fix Unity testsuite failing on .NET 9
Our Unity test suite now seems to be failing because wasi-experimental is not available on .NET 9. I don't think we're intending to use .NET 9. I believe @rekhoff knows the line to add that will fix this.

rekhoff:
# Description of Changes

This implements a fix for #3304, specifically addressing failures in:
* CI / Test Suite
* C#/Unity - Test Suite / unity-testsuite

Changes include:
* Updates several files in C# bindings to resolve a handful of errors that would occur in the C# binding tests check.
* Updated the `csharp-test.yml` file with an explicit .NET version to run .NET Tests under, in order to prevent the `wasi-experimental` workload error that occurs when running under .NET 9 due to `wasi-experimental` only supported under .NET 8

No

1

# Testing

- [x] Tested locally without errors, and passes GitHub tests


.github/workflows/csharp-test.yml
```diff
@@ -43,6 +43,8 @@ jobs:
       - name: Run .NET tests
         working-directory: sdks/csharp
         run: dotnet test -warnaserror
+        with:
+          dotnet-version: '8.x'
```
bfops: I think this is actually not working - the workflow file says it's broken now: https://github.com/clockworklabs/SpacetimeDB/actions/runs/18118462194

.github/workflows/csharp-test.yml
```diff
@@ -18,13 +18,16 @@ jobs:
       - name: Checkout repository
         id: checkout-stdb
         uses: actions/checkout@v4
+        
+        # Copy global.json to the current directory so that the dotnet things use it
+      - run: cp modules/global.json global.json
```
kim: Do we really want to do this?
There are a lot of mostly identical `global.json`s in the tree, we should maybe just use one. `dotnet` looks in parent directories for a `global.json`.

See #3297

.github/workflows/csharp-test.yml
```diff
@@ -15,16 +15,26 @@ jobs:
       cancel-in-progress: true
     timeout-minutes: 30
     steps:
+      - run: df -h
+      - name: "node-cleanup"
```
bfops, line 20:
```suggestion
      # Free up disk space because otherwise we run out during the test
      - name: "node-cleanup"
```

.github/workflows/csharp-test.yml
bfops: I do like that approach better. I think if we rename into that PR, the only thing that this PR needs to do is to free up the disk space.

rekhoff: Agreed, I have removed the global.json changes in favor of https://github.com/clockworklabs/SpacetimeDB/pull/3297

bfops:
I believe once #3297 is merged, only the changes in `csharp-test.yml` will be necessary, since the Test Suite will already be fixed.

.github/workflows/csharp-test.yml
```diff
@@ -15,6 +15,14 @@ jobs:
       cancel-in-progress: true
     timeout-minutes: 30
     steps:
+      - run: df -h
+      # Free up disk space because otherwise we run out during the test
+      - name: "node-cleanup"
+        run: |
+          sudo rm -rf /usr/local/lib/android
+          sudo docker image prune --all --force
+          sudo docker builder prune -a
```
jdetter, line 24: @bfops Instead of clearing some space here do you think we should just move this job to the custom runner which just has more space available? I know we've been running out of space there as well but I think we can fix that issue separately.

bfops: I'm open to that if it "just works". if we end up having to fight more battles, I'd say keep this as-is and maybe leave a TODO for the future.

bfops: trying that in https://github.com/clockworklabs/SpacetimeDB/pull/3318

rekhoff:
This is old and no longer relevant after workflow reworks done elsewhere.

PR #3313, opened 2025-09-29
Added the Blackholio Unreal Blueprint Tutorial

JasonAtClockwork:
# Description of Changes

- Expanded on the Unreal tutorial to create the Blueprint version, ~~this uses :::server-cpp and :::server-blueprint which might not work yet when deploying~~ with using Docusaurus Tabs.
- Updated the server side code to use signed integers as discussed for both Rust + C# server modules.

None

1 - Documentation update, does update the server side 

# Testing

- [x] Manually tested both Rust + C# modules
- [x] Ran through all images and steps


JasonAtClockwork:
The Unreal documentation has been updated but I need to fix up the Unity documentation to use signed integers as well as fix up the Blackholio demos all to use signed integers.

JasonAtClockwork:
Ready for review but not pushed to staging yet as we'll need to discuss the images and the new ::: setup I've used

docs/docs/unreal/part-1.md
```diff
@@ -69,6 +69,8 @@ Once the SDK is stabilized, we'll find a more ergonomic way to distribute it.
 
 ### Create the GameManager Actor
 
+:::server-cpp
```
rekhoff: The C++ and Blueprint documentation  toggle all appear to use `server-cpp` and `server-blueprint`. This seems to imply the server either uses C++ or Blueprints to run, but contain instructions pertaining to the client. We likely want to change these to `client-cpp` and `client-blueprint` or possibly `client-ucpp` to indicate this is Unreal C++ and not general C++.

JasonAtClockwork: 100% I'm actually working on this right now as I've got it running locally and :::server-rust is competing with :::server-cpp as an example. Even changing to :::client-cpp isn't working so I'm diagnosing now!

rekhoff:
If we are creating a Blueprint project, should we still generate the project as a C++ as shown here https://github.com/clockworklabs/SpacetimeDB/blob/5420d146f05fe4fcfa19fe138e2728863244e92e/docs/docs/unreal/part-1.md?plain=1#L44 or should Blueprint be selected?
If we should still select C++ in the Project Defaults, we may want to have a note or something, so folks going down the Blueprint path, don't generate the project as a Blueprint project out of habit.

JasonAtClockwork:
> If we are creating a Blueprint project, should we still generate the project as a C++ as shown here
> 
> https://github.com/clockworklabs/SpacetimeDB/blob/5420d146f05fe4fcfa19fe138e2728863244e92e/docs/docs/unreal/part-1.md?plain=1#L44
> 
> or should Blueprint be selected?
> If we should still select C++ in the Project Defaults, we may want to have a note or something, so folks going down the Blueprint path, don't generate the project as a Blueprint project out of habit.

Yes we will want all projects to start as C++, the reason is we're generating C++ code from the module into the project so might as well start them off on the right foot.

I'll make sure to get an extra note in there!

rekhoff:
Here are some items I found while reviewing:

In `docs/docs/unreal/part-2.md`

https://github.com/clockworklabs/SpacetimeDB/blob/deedc105303c477d9f77b5c001691256bc69d20e/docs/docs/unreal/part-2.md?plain=1#L729 Restart of Unreal was required for the `SpacetimeDB Build` node to become available.

In `docs/docs/unreal/part-3.md`
https://github.com/clockworklabs/SpacetimeDB/blob/deedc105303c477d9f77b5c001691256bc69d20e/docs/docs/unreal/part-3.md?plain=1#L708 Wrong object, `BP_GameMode` -> `BP_GameManager` Should be:
Open `BP_GameManager` and update to the following:


https://github.com/clockworklabs/SpacetimeDB/blob/deedc105303c477d9f77b5c001691256bc69d20e/docs/docs/unreal/part-3.md?plain=1#L1507 Typo, `Diameer` -> `Diameter`. Should read:
\- Add **Output** as `Diameter` with **Float** as the type.

https://github.com/clockworklabs/SpacetimeDB/blob/deedc105303c477d9f77b5c001691256bc69d20e/docs/docs/unreal/part-3.md?plain=1#L1560 Missing Blueprint `Output` setup task. Below screenshot should read:
\- Add **Output** as `GameManager` with **BP Game Manager** as the type.

https://github.com/clockworklabs/SpacetimeDB/blob/deedc105303c477d9f77b5c001691256bc69d20e/docs/docs/unreal/part-3.md?plain=1#L1586 Typo, `Input` -> `Output`. Should read:
\- Add **Output** as `Center` with **Vector** as the type.

https://github.com/clockworklabs/SpacetimeDB/blob/deedc105303c477d9f77b5c001691256bc69d20e/docs/docs/unreal/part-3.md?plain=1#L1626 Different node behavior from screenshot, `Get Username` function required Exec.
image
I put it's execution between `Cast to WBP_Nameplate` and `Update Text` in my example.
**Note**: Editor Version: 5.6.1-44394996+++UE5+Release-5.6

https://github.com/clockworklabs/SpacetimeDB/blob/deedc105303c477d9f77b5c001691256bc69d20e/docs/docs/unreal/part-3.md?plain=1#L2055 Typo, `Class Reference` -> `Object Reference`. Should read:
\    - Change **Variable Type** to **BP Game Manager -> Object Reference**

https://github.com/clockworklabs/SpacetimeDB/blob/deedc105303c477d9f77b5c001691256bc69d20e/docs/docs/unreal/part-3.md?plain=1#L2057 Typo, `Class Reference` -> `Object Reference`. Should read:
\    - Change **Variable Type** to **BP Player Pawn -> Object Reference**

https://github.com/clockworklabs/SpacetimeDB/blob/deedc105303c477d9f77b5c001691256bc69d20e/docs/docs/unreal/part-3.md?plain=1#L2064 Missing Blueprint `Output` setup task. Below screenshot should read:
\- Add **Output** as `GameManager` with **BP Game Manager** as the type.

https://github.com/clockworklabs/SpacetimeDB/blob/deedc105303c477d9f77b5c001691256bc69d20e/docs/docs/unreal/part-3.md?plain=1#L2141 Unable to generate `Bind Event to On Username Changed` node.
When following the `server-blueprint` instructions, I don't see a `SetUsername` being created.

https://github.com/clockworklabs/SpacetimeDB/blob/deedc105303c477d9f77b5c001691256bc69d20e/docs/docs/unreal/part-3.md?plain=1#L2149 Unclear the purpose of this blueprint section.

docs/docs/unreal/part-1.md
JasonAtClockwork: Completely rebuilt with the latest work for Docusaurus and matches how Rust/C# works but on a separate stored cookie

rekhoff (approved):
Briefly covered both server and client setups (Mostly referenced my existing Blueprint project for that test, due to time involved.) The Docusaurus rebuild looks successful, and the pages look good. I didn't notice any issues, approved.

PR #3317, opened 2025-09-30
Add SpacetimeAuth docs

JulienLavocat:
# Description of Changes

Add documentation for SpacetimeAuth

None.

1

# Testing

Checked if everything was working using a local instance of the website


docs/docs/spacetimeauth/index.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Authentication methods
+
+- Magic link
+- Github
+- Google
+- Discord
+- Twitch
+- Kick
+- (Coming soon) Steam: authenticate users via Steam and verify game or DLC ownership
```
cloutiertyler: I'd consider removing this. @gefjon wrote a style guide which requires that we don't reference upcoming features in the docs, and there is an element of good hygiene to that.

It's possible we also lose customers who maybe don't think that we're going eventually go there, but I don't think that's all that likely.

docs/docs/spacetimeauth/index.md
```diff
@@ -0,0 +1,95 @@
...
+
+### User & role management
+
+- Create, update, and manage users
+- Assign roles to users for role-based access control
+
+### Customization
+
+- Customizable theme for login pages
+- (Coming soon) Customizable templates for emails
```
cloutiertyler: Similar comment here.

docs/docs/spacetimeauth/index.md
JulienLavocat: Yeah I was very hesistant to put them here but decided to proceed with it and see what you would say. I'll remove all of these "coming soon".

docs/docs/spacetimeauth/index.md
JulienLavocat: Similar answer here :D

docs/docs/spacetimeauth/create-project.md
```diff
@@ -0,0 +1,48 @@
...
+
+You can also click on your profile icon in the top right corner and select
+"SpacetimeAuth" from the dropdown menu.
+
+## 2. Creating a New Project
+
+Once you're in the SpacetimeAuth dashboard, you can create a new project by
+clicking the "Create Project" button.
+
+![Access the menu](./menu-access.png)
```
cloutiertyler: Does this work? With images? I thought we determined it does not.

docs/docs/spacetimeauth/configure-project.md
```diff
@@ -0,0 +1,128 @@
...
+
+Clients represent applications that will use SpacetimeAuth for authentication.
+Each client has its own set of settings, including redirect URIs,
+post logout URIs, and name.
+
+You can manage clients by navigating to the "Clients" tab in your
+SpacetimeAuth project dashboard.
+![Clients tab](./clients-tab.png)
+Every project comes with a default client that you can use to get started.
+You can also create additional clients by clicking the "Create Client" button.
```
cloutiertyler, line 20: Reading this, I do have to wonder, when would I ever need any other clients? My SpacetimeDB module is the only application that is going to be using the token no? If we somehow register my SpacetimeDB database/project with the SpacetimeAuth project, do I even need other clients? Or am I missing something?

I am correct in saying that my "default client" is my database, though right? Or do I not have the flow right?

Should we maybe put some examples here?

docs/docs/spacetimeauth/configure-project.md
```diff
@@ -0,0 +1,128 @@
...
+
+- **Name**: The name of the client (e.g. "My Web App").
+- **Redirect URIs**: The URIs to which SpacetimeAuth allows to redirect
+  the user after a successful login. These must match the URIs used in your application.
+- **Post Logout Redirect URIs**: The URIs to which SpacetimeAuth allows to redirect
+  the user after a logout. These must match the URIs used in your application.
+
+> ⚠️Remember to keep your client secret secure, **never** expose it in
+> client-side code or public repositories.
+> You can freely share the client ID as it is not sensitive information.
```
cloutiertyler: With respect to the above, we could store client secrets in SpacetimeDB (or derive them from the SpacetimeDB private key) and then they wouldn't need to keep the secret themselves, right?

docs/docs/spacetimeauth/configure-project.md
```diff
@@ -0,0 +1,128 @@
...
+experience regardless of the identity provider used.
+For example the username claim is mapped to the standard `preferred_username` claim.
+
+You can manage identity providers by navigating to the "Identity Providers" tab
+in your SpacetimeAuth project dashboard.
+
+![Identity Providers tab](./identity-providers-tab.png)
+
+To enable an identity provider, enter the client ID and client secret obtained
+from the provider's developer console. You must also configure the redirect URI
```
cloutiertyler: I would maybe add a note here saying that SpacetimeAuth is acting as a client for the external identity provider in this case.

cloutiertyler (approved):
This looks really good, just left a few questions/suggestions I'd like to get your thoughts on before merging.

docs/docs/spacetimeauth/create-project.md
JulienLavocat: It doesn't, I forgot to take all the screenshots and upload them. Will do now, it's also why the CI is failing

docs/docs/spacetimeauth/configure-project.md
JulienLavocat: You are partially correct as in the majority of projects will only use a single client. Where you might need different clients is for example to authenticate sidecars using a `client_credentials` flow or a different client for your CS panel.

You could of course get away with only using a single client for all these applications but it's a little bit messy to mix all the different redirect URIs in my opinion.

docs/docs/spacetimeauth/configure-project.md
JulienLavocat: I'm not sure I understand what you mean here, the client secret is used as a password for the `client_credentials` flow for example. It must be available to the developers

docs/docs/spacetimeauth/configure-project.md
JulienLavocat: Fair, will do

docs/docs/spacetimeauth/create-project.md
JulienLavocat: Fixed

docs/docs/spacetimeauth/configure-project.md
JulienLavocat: Done

> Since SpacetimeAuth acts as a client for the external identity provider,
you need to provide the client ID and client secret obtained
from the provider's developer console in order to enable the provider.
You must also configure the redirect URI in the provider's developer console to
point to SpacetimeAuth (see below).

docs/docs/spacetimeauth/configure-project.md
cloutiertyler: Roger, we might want to add into the docs almost exactly this explanation (but geared towards the audience).

docs/docs/spacetimeauth/configure-project.md
JulienLavocat: From our call: explain that the client secret will only be used to get a token using the client_credentials flow (i.e. just a token with a sub being the client id)

PR #3321, opened 2025-09-30
C# tests - update snapshots

bfops:
# Description of Changes

I chose to regenerate all of the bindings in https://github.com/clockworklabs/SpacetimeDB/pull/3310, but that caused the snapshots to change. This PR follows up to update them.

I also updated the outdated script `tools~/gen-quickstart-chat.sh` (which was generating from a different module than the CI was), and updated the CI to use it instead of manually running a similar command.

Test-only changes.

1

# Testing

- [x] unity-testsuite passes on this PR (which it does not on `master`).


sdks/csharp/examples~/quickstart-chat/client/module_bindings/Reducers/SendMessage.g.cs
```diff
@@ -61,7 +61,7 @@ public SendMessage()
                 this.Text = "";
             }
 
-            string IReducerArgs.ReducerName => "send_message";
+            string IReducerArgs.ReducerName => "SendMessage";
```
jdetter, line 64: So it looks like originally this file was using the C# module as the generation source but then we changed it to the rust module and then now we're reverting back to the C# module. I think we should probably regenerate these bindings to have the latest bindings using the C# module if that makes sense.

bfops: that makes sense, I just don't know how to review the snapshot diff to verify that the new ones are correct. I'm fine to just blindly accept the new ones if you think that's safe (I think it probably is but I don't want to be the only person making that call).

sdks/csharp/tools~/gen-quickstart.sh
```diff
@@ -7,4 +7,4 @@ SDK_PATH="$(dirname "$0")/.."
 SDK_PATH="$(realpath "$SDK_PATH")"
 
 cargo build --manifest-path "$STDB_PATH/crates/standalone/Cargo.toml"
-cargo run --manifest-path "$STDB_PATH/crates/cli/Cargo.toml" -- generate -y -l csharp -o "$SDK_PATH/examples~/quickstart-chat/client/module_bindings" --project-path "$STDB_PATH/modules/quickstart-chat"
+cargo run --manifest-path "$STDB_PATH/crates/cli/Cargo.toml" -- generate -y -l csharp -o "$SDK_PATH/examples~/quickstart-chat/client/module_bindings" --project-path "$SDK_PATH/examples~/quickstart-chat/server"
```
jdetter, line 10: I was gonna ask if this would be a candidate for rewriting in rust but I think this is simple enough where I couldn't really see there being that much value in converting it to rust - plus now if we're invoking this via the CI I think I'm fine with just keeping this as you have it here. Thanks for the update Zeke 👍

jdetter (approved):
This looks great, thanks Zeke 👍

PR #3324, opened 2025-10-01
CI - `apt install` invocations correctly pass `-y`

bfops:
# Description of Changes

We had places that weren't passing `-y`. I assume these were only working because they were in environments where the things were already installed, so there wasn't any confirmation dialog. For some reason, the arm runner now newly needs to install packages, so CI invocations began failing.

None. CI only.

1

# Testing

- [x] CI passes
  - [ ] the "Test spacetimedb-update" flow passes on arm, which is not currently true elsewhere.

### DIFF
--- a/.github/workflows/discord-posts.yml
+++ b/.github/workflows/discord-posts.yml
@@ -19,7 +19,7 @@ jobs:
           sudo apt-get install -y apt-transport-https
           echo "deb [arch=amd64 signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" | sudo tee /etc/apt/sources.list.d/github-cli.list
           sudo apt-get update
-          sudo apt-get install gh
+          sudo apt-get install -y gh
 
       - name: Fetch Check Run Results
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/discord-posts.yml
```diff
@@ -19,7 +19,7 @@ jobs:
           sudo apt-get install -y apt-transport-https
           echo "deb [arch=amd64 signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" | sudo tee /etc/apt/sources.list.d/github-cli.list
           sudo apt-get update
-          sudo apt-get install gh
+          sudo apt-get install -y gh
```
jdetter, line 22: This should probably be `apt` instead of `apt-get` but it still will technically work in 22.04 and 24.04 to use `apt-get` in compatibility mode. It would be more future-proof to just use `apt` here though.

jdetter (approved):
Both of these are proper fixes, although I would change the apt-get to `apt` in the `discord-posts.yml` file.

jdetter:
🤦 This is why the arm job has been stalling, right?

.github/workflows/discord-posts.yml
bfops: completely agreed, but at this hour I just want to merge. we have other places that still use apt-get as well.

PR #3327, opened 2025-10-01
Typescript module API

coolreader18:
# Description of Changes

Currently based on #3361 

Implements most of the TS module API (not yet a function for type aliases).

# Testing

- [x] Extremely basic module stuff works
- [ ]


crates/bindings-typescript/src/server/runtime.ts
```diff
@@ -0,0 +1,529 @@
...
+          return sys.datastore_delete_by_index_scan_range_bsatn(
+            index_id,
+            ...args
+          );
+        },
+      } as RangedIndex<any, any>;
+    }
+
+    if (Object.hasOwn(tableView, indexDef.name!)) {
+      freeze(Object.assign(tableView[indexDef.name!], index));
```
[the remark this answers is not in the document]
coolreader18: Sure, I guess.

crates/cli/src/tasks/javascript.rs
```diff
@@ -0,0 +1,175 @@
...
+{
+    match Handle::try_current() {
+        Ok(handle) => tokio::task::block_in_place(|| handle.block_on(fut)),
+        Err(_) => runtime().block_on(fut),
+    }
+}
+
+pub(crate) fn build_javascript(project_path: &Path, build_debug: bool) -> anyhow::Result<PathBuf> {
+    let cwd = fs::canonicalize(project_path)?;
+    let mut bundler = Bundler::new(BundlerOptions {
```
bfops, line 40: I don't really have context on how this infrastructure works - why do we run the bundler "directly" instead of shelling out to a language-specific CLI like we do for other languages?

crates/cli/src/tasks/mod.rs
```diff
@@ -3,17 +3,24 @@ use std::path::{Path, PathBuf};
...
+use self::rust::build_rust;
 
 use duct::cmd;
 
-pub fn build(project_path: &Path, lint_dir: Option<&Path>, build_debug: bool) -> anyhow::Result<PathBuf> {
+pub fn build(
+    project_path: &Path,
+    lint_dir: Option<&Path>,
+    build_debug: bool,
+) -> anyhow::Result<(PathBuf, &'static str)> {
```
bfops, line 16: nit, maybe for a future PR: maybe we want a struct here? or `enum BuiltType` or something? (tl;dr it's not lovely to be looking at the call sites or function details to figure out what the returned args do)

crates/cli/src/tasks/mod.rs
```diff
@@ -33,9 +40,15 @@ pub fn build(project_path: &Path, lint_dir: Option<&Path>, build_debug: bool) ->
                 eprintln!("Continuing with unoptimised module.");
             }
         }
+        return Ok((wasm_path, "Wasm"));
+    }
+    if lang == ModuleLanguage::Javascript {
+        Ok((output_path, "Js"))
```
bfops: nit: maybe we move the `if lang == ModuleLanguage::JavaScript` block above the `if !build_debug` block? Then I might also add a comment saying something like.
```
// The rest of this function only applies to wasm modules
```

crates/cli/src/subcommands/project/typescript/_gitignore
bfops, line 1: is this file used?

crates/cli/src/subcommands/generate.rs
```diff
@@ -144,16 +155,21 @@ pub async fn exec_ex(
...
-        let wasm_path = if let Some(path) = wasm_file {
+        let path = if let Some(path) = wasm_file {
+            println!("Skipping build. Instead we are inspecting {}", path.display());
+            path.clone()
+        } else if let Some(path) = js_file {
             println!("Skipping build. Instead we are inspecting {}", path.display());
             path.clone()
         } else {
-            build::exec_with_argstring(config.clone(), project_path, build_options).await?
+            build::exec_with_argstring(config.clone(), project_path, build_options)
```
bfops: nit:
```suggestion
            let (path, _) = build::exec_with_argstring(config.clone(), project_path, build_options).await?;
            path
```

crates/cli/src/subcommands/init.rs
```diff
@@ -204,6 +205,34 @@ pub async fn exec_init_csharp(args: &ArgMatches) -> anyhow::Result<()> {
...
+    let export_files = vec![
+        (include_str!("project/typescript/tsconfig._json"), "tsconfig.json"),
+        (include_str!("project/typescript/index._ts"), "src/index.ts"),
+        (include_str!("project/typescript/_gitignore"), ".gitignore"),
+    ];
+
+    // Check all dependencies
+    check_for_git();
+
+    for data_file in export_files {
```
bfops: nit:
```suggestion
    for (contents, file) in export_files {
```

crates/cli/src/tasks/mod.rs
coolreader18: This `&'static str` is actually just the `HostType` enum, except that's defined in `core` which we don't depend on in `cli`.

bfops: ah, so maybe then we would just copy it into the CLI with a comment that says it's copied from core?

crates/bindings-typescript/src/lib/autogen/algebraic_type_type.ts
cloutiertyler, line 1: These changes need to be made more generally in the `typescript.rs` code generation file.

crates/bindings-typescript/src/lib/time_duration.ts
```diff
@@ -38,4 +47,13 @@ export class TimeDuration {
   static fromMillis(millis: number): TimeDuration {
     return new TimeDuration(BigInt(millis) * TimeDuration.MICROS_PER_MILLIS);
   }
+
```
cloutiertyler, line 50:
```suggestion
  /** This outputs the same string format that we use in the host and in Rust modules */ 
```

crates/bindings-typescript/src/server/errors.ts
```diff
@@ -0,0 +1,252 @@
+import type { CheckAnyMetadata, UntypedTableDef } from './table';
+
+/**
+ * Base class for all Spacetime errors.
+ * Each subclass must define a static CODE and MESSAGE property.
+ * Instances of SpacetimeError can be created with just an error code,
+ * which will return the appropriate subclass instance.
+ */
+export class SpacetimeError {
```
cloutiertyler: Rename to `SpacetimeHostError` or `HostError`

crates/bindings-typescript/src/server/runtime.ts
```diff
@@ -0,0 +1,532 @@
...
+const sys: typeof _syscalls = freeze(
+  Object.fromEntries(
+    Object.entries(_syscalls).map(([name, syscall]) => [
+      name,
+      wrapSyscall(syscall),
+    ])
+  ) as typeof _syscalls
+);
+
+export function __call_reducer__(
```
cloutiertyler:
```ts
// These are the two ways that a developer will be able to indicate to the host that the sender has made an error
// and that this transaction should be rolled back
spacetimedb.reducer('foobar', () => {
  throw new SenderError("this is a string");
  // OR
  return { tag: "err", value: "this is a string" }; 
});
```

crates/bindings-typescript/src/server/errors.ts
```diff
@@ -0,0 +1,252 @@
...
+ * const result = table.tryInsert({ id: 1, name: 'Alice' });
+ * if (!result.ok) {
+ *  if (result.err instanceof UniqueAlreadyExists) {
+ *   console.error('Unique constraint violated:', result.err.message);
+ * } else if (result.err instanceof AutoIncOverflow) {
+ *  console.error('Auto-increment overflow:', result.err.message);
+ * }
+ * ```
+ */
+export type TryInsertError<TableDef extends UntypedTableDef> =
```
cloutiertyler: Just remove `tryInsert` and have it throw an error which you can either catch or not. (or have it return the `tag` form of the error)

crates/bindings-typescript/src/server/type_builders.ts
```diff
@@ -3,37 +3,87 @@ import {
...
-          : T[K] extends TypeBuilder<infer V, any>
-            ? V
-            : never;
-      }
-    : never;
+export type InferTypeOfRow<
+  T extends Record<
+    string,
+    ColumnBuilder<any, any, any> | TypeBuilder<any, any>
+  >,
```
cloutiertyler:
```suggestion
  T extends RowObj,
```

crates/bindings-typescript/src/server/reducers.ts
```diff
@@ -0,0 +1,216 @@
...
+  lifecycle?: RawReducerDefV9['lifecycle']
+): void {
+  const paramType: ProductType = {
+    elements: Object.entries(params).map(([n, c]) => ({
+      name: n,
+      algebraicType: ('typeBuilder' in c ? c.typeBuilder : c).algebraicType,
+    })),
+  };
+
+  MODULE_DEF.reducers.push({
```
cloutiertyler, line 104: Should probably check that we're not doing this twice for the same named reducer.

crates/bindings-typescript/src/server/runtime.ts
```diff
@@ -0,0 +1,532 @@
...
+      freeze(Object.assign(tableView[indexDef.name!], index));
+    } else {
+      tableView[indexDef.name!] = freeze(index) as any;
+    }
+  }
+
+  return freeze(tableView);
+}
+
+function bsatnBaseSize(typespace: Typespace, ty: AlgebraicType): number {
```
cloutiertyler, line 323: I think this is a little duplicative with regards to the stuff in `algebraic_type.ts`.

crates/bindings-typescript/src/server/table.ts
```diff
@@ -0,0 +1,346 @@
...
+  count(): bigint;
+
+  /** Iterate over all rows in the TX state. Rust Iterator<Item=Row> → TS IterableIterator<Row>. */
+  iter(): IterableIterator<RowType<TableDef>>;
+  [Symbol.iterator](): IterableIterator<RowType<TableDef>>;
+
+  /** Insert and return the inserted row (auto-increment fields filled). May throw on error. */
+  insert(row: RowType<TableDef>): RowType<TableDef>;
+
+  /** Like insert, but returns a Result instead of throwing. */
```
cloutiertyler: See above, probs just get rid of this?

crates/bindings-typescript/src/server/type_util.ts
```diff
@@ -10,15 +10,33 @@ export type Set<T, F extends string, V> = Prettify<
...
-      : K
-    : K]: T[K];
-}>;
+/**
+ * A Result type representing either a success or an error.
+ * - `ok: true` with `val`: Indicates a successful operation with the resulting value.
+ * - `ok: false` with `err`: Indicates a failed operation with the associated error.
+ */
+// TODO: Should we use the same `{ tag: 'Ok', value: T } | { tag: 'Err', value: E }` style as we have for other sum types?
+export type Result<T, E> = { ok: true; val: T } | { ok: false; err: E };
```
cloutiertyler: Probably ditch.

crates/bindings-typescript/tsup.config.ts
```diff
@@ -118,8 +118,17 @@ export default defineConfig([
...
-    treeshake: 'smallest',
-    external: ['undici'],
+    banner: {
+      js:
+        'typeof globalThis!=="undefined"&&(' +
+        '(globalThis.global=globalThis.global||globalThis),' +
+        '(globalThis.window=globalThis.window||globalThis));',
+    },
+    treeshake: {
+      moduleSideEffects: ['src/server/polyfills.ts'],
```
cloutiertyler: Add side effects to `index.ts`

crates/cli/src/subcommands/project/typescript/_gitignore
cloutiertyler: Yes it is used in the `init.rs` file.

crates/cli/src/tasks/javascript.rs
cloutiertyler: Because then we would need to install it on their machine for them. Unlike with Rust there are 70,000 bundlers in TypeScript and we can be sure that they have any particular one.

cloutiertyler: We could maybe make them install it as a dependency of their own project.

crates/cli/src/tasks/mod.rs
cloutiertyler: I thought about this @bfops, but I also didn't want to make the diff any bigger and it was touched by a lot of files.

crates/core/src/host/v8/syscall.rs
```diff
@@ -472,13 +472,13 @@ fn datastore_index_scan_range_bsatn(
     args: FunctionCallbackArguments<'_>,
 ) -> SysCallResult<u32> {
     let index_id: IndexId = deserialize_js(scope, args.get(0))?;
-    let mut prefix: &[u8] = deserialize_js(scope, args.get(1))?;
+    let mut prefix: Vec<u8> = deserialize_js(scope, args.get(1))?;
```
cloutiertyler: @coolreader18 thought they undid these. Please undo.

modules/module-test-ts/package.json
```diff
@@ -0,0 +1,16 @@
...
+  "scripts": {
+    "build": "cargo build -p spacetimedb-standalone && cargo run -p spacetimedb-cli -- build",
+    "generate-ts": "cargo build -p spacetimedb-standalone && cargo run -p spacetimedb-cli -- generate --lang typescript --out-dir ts-codegen",
+    "publish": "cargo run -p spacetimedb-cli -- publish"
+  },
+  "dependencies": {
+    "spacetimedb": "workspace:^"
+  },
+  "devDependencies": {
+    "tsup": "^8.1.0"
```
cloutiertyler: TODO: remove later

modules/module-test-ts/tsup.config.ts
cloutiertyler, line 1: TODO: delete this file

modules/quickstart-chat-ts/src/index.ts
cloutiertyler, line 1: Update for new error handling.

modules/sdk-test-connect-disconnect-ts/tsconfig.json
cloutiertyler, line 1: Probably we should work towards having a standardized tsconfig and importing it into other modules.

modules/sdk-test-ts/src/index.ts
```diff
@@ -0,0 +1,1164 @@
...
+  n: t.array(t.f32()),
+  o: t.array(t.f64()),
+  p: t.array(t.string()),
+  q: t.array(t.identity()),
+  r: t.array(t.connectionId()),
+  s: t.array(t.timestamp()),
+  t: t.array(t.timeDuration()),
+});
+
+// /// Defines one or more tables, and optionally reducers alongside them.
```
cloutiertyler: Remove this

pnpm-workspace.yaml
```diff
@@ -2,4 +2,7 @@ packages:
   - 'crates/bindings-typescript'
   - 'crates/bindings-typescript/test-app'
   - 'crates/bindings-typescript/examples/quickstart-chat'
-  - 'docs'
\ No newline at end of file
+  - 'modules/module-test-ts'
+  - 'modules/quickstart-chat-ts'
+  - 'modules/sdk-test-ts'
```
cloutiertyler, line 8:
```suggestion
  - 'modules/sdk-test-ts'
  - 'modules/sdk-test-connect-disconnect-ts'
```

cloutiertyler (approved):
I am approving this for once we address the comments and merge in https://github.com/clockworklabs/SpacetimeDB/pull/3404. We also need to test once again once that PR is merged into this PR.

crates/cli/src/subcommands/project/typescript/_gitignore
bfops: Sorry, let me clarify: what effect does an empty `gitignore` have? I understand that the `_gitignore` is _literally_ used, but how is the resultant `.gitignore` file useful?

crates/cli/src/tasks/mod.rs
bfops: yup, I'm totally fine with a `// TODO` and/or a follow-up PR.

crates/cli/src/tasks/javascript.rs
bfops: With C# we just say "yo please install dotnet", we don't install it for them.

~~Are we trying to support every possible bundler? We're kind of restrictive with what toolchains we support in other languages.~~

Either way, not a blocking issue since we are trying to get this PR in.

crates/cli/src/tasks/mod.rs
bfops: I added a TODO

crates/cli/src/subcommands/project/typescript/_gitignore
bfops: still wondering about this one

crates/cli/src/tasks/javascript.rs
bfops: I guess I would have expected that we just run something like `npm build`. rollup (or rolldown) would just be a dev-dependency or something. And then all the config options would be specified in another external file.

crates/cli/src/subcommands/project/typescript/package._json
```diff
@@ -0,0 +1,15 @@
...
+  "description": "",
+  "scripts": {
+    "build": "spacetime build",
+    "publish": "spacetime publish"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "spacetimedb": "1.5.*"
```
bfops, line 13: this needs to be added to the script in `tools/upgrade-version`

crates/cli/src/tasks/javascript.rs
cloutiertyler: That is an option, but it then requires users have a certain package.json in order to work properly and in general would put more of the complexity/configuration onto the user and give us less control over it.

crates/cli/src/subcommands/project/typescript/_gitignore
cloutiertyler: It's the same template that we provide in other languages. The file is not used by git since there is nothing there, I could remove it from all the init templates in a separate PR (or @drogus's [PR](https://github.com/clockworklabs/SpacetimeDB/pull/3366#issuecomment-3410859741)) though. It's not breaking to remove it since it's only for new projects.

bfops: huh TIL. sgtm

PR #3328, opened 2025-10-02
Adding wss:// support for the Unreal SDK

thediymaker:
# Description of Changes

Added support for wss:// protocol in Unreal SDK. The SDK previously forced all connections to use ws://, preventing connections to servers behind SSL/TLS proxies. The fix detects if the URI already includes a protocol (ws:// or wss://) and preserves it; otherwise defaults to ws:// for backward compatibility. This enables connections to SpacetimeDB servers hosted with Cloudflare Tunnels or other secure proxies.

This change is fully backwards compatible. URIs without a protocol prefix will continue to work as before, with ws:// being automatically prepended. No API signatures or existing behavior has changed.

Simple string prefix check before URL construction. The core WebSocket connection logic remains unchanged, and the default behavior is preserved when no protocol is specified.

# Testing

Successfully tested with a wss:// connection to my self-hosted SpacetimeDB server behind Cloudflare Tunnel. Verified backward compatibility with protocol-less URIs when directly connecting to the server. Currently running without issues.


bfops:
Thank you for submitting this! We'll get our resident Unreal wizard on it soon.

JasonAtClockwork:
Thanks again for this submission! I've updated it to reflect how our Unity SDK handles it to "support" http/https by translating to ws/wss.

PR #3331, opened 2025-10-02
core: Allow snapshot worker to be reused

kim:
Make it so the `SnapshotWorker` can be re-configured with a new
committed state. This allows event subscriptions to remain valid while a
replica transitions from leader to follower and vice versa.

This is considerably simpler than keeping the lifetimes of database and
persistence services strictly in-sync, at the expense of an idle task
per replica.

1.5

### DIFF
--- a/crates/core/src/db/snapshot.rs
+++ b/crates/core/src/db/snapshot.rs
@@ -179,15 +179,13 @@ impl SnapshotWorkerActor {
         while let Some(req) = self.snapshot_requests.next().await {
             match req {
                 Request::TakeSnapshot => {
-                    if let Some(state) = database_state.as_ref().and_then(Weak::upgrade) {
-                        let res = self
-                            .take_snapshot(state)
-                            .await
-                            .inspect_err(|e| warn!("SnapshotWorker: {e:#}"));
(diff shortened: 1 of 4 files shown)


crates/core/src/db/snapshot.rs
```diff
@@ -50,12 +50,15 @@ impl Compression {
 ///
 /// The [SnapshotWorker] handle is freely cloneable, so ownership can be shared
 /// between the database and control code.
+///
+/// It is possible to re-use a [SnapshotWorker] to create a new database
+/// instance: when passed to [super::relational_db::RelationalDB::open], the
+/// worker's [SnapshotDatabaseState] will be replaced with the database's.
```
gefjon, line 56:
```suggestion
/// It is possible to re-use a [SnapshotWorker] to create a new database
/// instance: when passed to [super::relational_db::RelationalDB::open], the
/// worker's [SnapshotDatabaseState] will be replaced with the database's.
/// We use this for replicated databases when transitioning between the leader and follower states,
/// to preserve event subscriptions on the `SnapshotWorker`'s `snapshot_created` channel.
```

PR #3334, opened 2025-10-02
WIP: teams

kim:
WIP


crates/cli/src/subcommands/delete.rs
```diff
@@ -22,11 +28,93 @@ pub async fn exec(mut config: Config, args: &ArgMatches) -> Result<(), anyhow::E
...
+            let parent = info
+                .parent
+                .map(|parent| format!(" (parent: {parent})"))
+                .unwrap_or_default();
+
+            write!(out, "{identity}{parent}{names}")?;
+        }
+
+        Ok(())
+    }
```
kim: @aasoni I have been a little un-imaginative about how to render the tree of children when trying to delete a parent. Note also that a database can have more than one name.

If you have a specific idea of how this should be rendered, please let me know.

kim: Or maybe @bfops ?

bfops: I'm just getting to this. I think it's fine to do "something" and improve it later if we feel strongly about it. This isn't a very common subcommand in people's workflows.

kim: I have changed this to render a tree (like the `tree` command), unless `--yes` is given:

```
0000000000000000000000000000000000000000000000000000000000000001: parent
├── 0000000000000000000000000000000000000000000000000000000000000004: bro, sibling
└── 0000000000000000000000000000000000000000000000000000000000000002: child
    ├── 0000000000000000000000000000000000000000000000000000000000000005
    └── 0000000000000000000000000000000000000000000000000000000000000003: grandchild
```

We could abbreviate the identities if, and only if, there are no collisions. I'll leave that as a future exercise.

kim:
Superseded by #3496 #3519 #3523 #3525

PR #3337, opened 2025-10-02
CI - Multi-line runs fail properly on Windows

bfops:
# Description of Changes

It turns out that multi-line `run` statements don't fail properly on Windows. If one of the lines fails, the run won't exit early, and in fact the step will return successful.

CI-only changes.

1

# Testing
- [x] Confirmed that a one-line command will fail properly on Windows: https://github.com/clockworklabs/SpacetimeDB/actions/runs/18204663513/job/51831974918?pr=3337
- [x] Confirmed that a multi-line command properly exits early with these changes: https://github.com/clockworklabs/SpacetimeDB/actions/runs/18203897520/job/51829384741?pr=3335
- [x] Confirmed that a windows package build still works even though the shell is `bash` for part of it now (https://github.com/clockworklabs/SpacetimeDB/actions/runs/18205281236/job/51834045429)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,12 +14,6 @@ on:
 name: CI
 
 jobs:
-  failure_test:
-    runs-on: windows-latest
-    steps:
-      - run: |
-          false
-
   docker_smoketests:
     name: Smoketests
     strategy:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,6 +21,7 @@ jobs:
         include:
           - { runner: spacetimedb-runner, smoketest_args: --docker }
           - { runner: windows-latest, smoketest_args: --no-build-cli }
+        runner: [ spacetimedb-runner, windows-latest ]
     runs-on: ${{ matrix.runner }}
     steps:
       - name: Find Git ref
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -21,7 +21,6 @@ jobs:
         include:
           - { runner: spacetimedb-runner, smoketest_args: --docker }
           - { runner: windows-latest, smoketest_args: --no-build-cli }
-        runner: [ spacetimedb-runner, windows-latest ]
```
bfops, line 24: this is redundant with the lines just above

.github/workflows/package.yml
```diff
@@ -67,6 +67,7 @@ jobs:
 
       - name: Package (windows)
         if: ${{ runner.os == 'Windows' }}
+        shell: bash
```
bfops, line 70: I am not 1000% sure that this will work as expected, but I feel like it's better to make things run in bash than to add the windows witchcraft where it's not necessary.

I guess maybe it would be good to have a workflow that audits our `yml` files for multi-line things that would run on windows. but that seems hard to do.

bfops: I have confirmed that this works properly still

Issue #3044, opened 2025-07-07
C#/Unity SDK - Calling `Disconnect` doesn't disconnect


bfops:
[From the Discord](https://discord.com/channels/1037340874172014652/1037343198617538580/1391277860794142881):

> In the C# client, if you call `Disconnect()` on the `DbConnection` while it's connecting before it connects, it doesn't seem to disconnect from the server behind the scenes. It isn't until I close my client application that the server acknowledges that the client connection ended.
> 
> The case where this could occur is if you have a `Disconnect()` call when you close your application. Say a user mistakenly opens and is trying to close it right away.
> 
> I don't think this breaks anything, but feels like it shouldn't work that way.

PR #3352, opened 2025-10-03
Adding Abort to C# Websocket

rekhoff:
# Description of Changes

The implementation of a solution to #3044 , this adds an `Abort` function to the `WebSocket`, which runs if `Disconnect` is called when the `WebSocket` is not connected.

Not API breaking.

1

# Testing

- [X] Test locally with a C# CLI test client. 
**Note**: Before change (either on Rust of C# server), server would see 4 `Debug` log entries about connecting, but not see the `Info` log about the client connection ending like would normally be seen in a disconnect. After change, server shows no log entries at all, because connection is properly aborted.
- [x] Test locally with a C# WebGL test client.


sdks/csharp/src/SpacetimeDBClient.cs
jdetter, line 503: This is somewhat unrelated but we might want to clean up this logic as well. If the socket isn't in the connected or connecting states we should just not try to disconnect. If the socket is connected we should try a normal disconnection and if the socket is connecting then we should try to abort.

rekhoff: More or less that is what this is doing. If `DBConnection.Disconnect` is called by the client, the `Disconnect` function is checking if the websocket is connected, and if it isn't actually connected yet, it's calling `Abort`.
But to your point, I've added a check to ensure we only call `Abort` if we are in the process of connecting, so we don't accidently call it if we are already disconnecting or are in a closed status.

sdks/csharp/src/SpacetimeDBClient.cs
```diff
@@ -475,6 +475,17 @@ public void Disconnect()
...
             }
+#if UNITY_WEBGL && !UNITY_EDITOR
+                else if (webSocket.IsConnecting)
+            {
+                webSocket.Abort(); // forceful during connecting
+            }
+#else
+            else if (webSocket.IsConnecting || webSocket.IsNoneState)
+            {
+                webSocket.Abort(); // forceful during connecting
```
jdetter, line 517: since the contents of the if statements are the same, maybe let's just `#if/#else` the if statements themselves?

jdetter: e.g.
```
            if (webSocket.IsConnected)
            {
                webSocket.Close();
            }
#if UNITY_WEBGL && !UNITY_EDITOR
            else if (webSocket.IsConnecting)
#else
            else if (webSocket.IsConnecting || webSocket.IsNoneState)
#endif
            {
                webSocket.Abort(); // forceful during connecting
            }
```

sdks/csharp/src/WebSocket.cs
```diff
@@ -364,14 +375,36 @@ await Ws.CloseAsync(WebSocketCloseStatus.MessageTooBig, closeMessage,
...
+        /// <summary>
+        /// Cancel an in-flight ConnectAsync. Safe to call if no connect is pending.
+        /// </summary>
+        public void CancelConnect()
+        {
+#if UNITY_WEBGL && !UNITY_EDITOR
+            // No CTS on WebGL. Mark cancel intent so that when socket id arrives or open fires,
+            // we immediately close and avoid reporting a connected state.
+            _cancelConnectRequested = true;
+            return;
```
jdetter: nit: we don't need this `return;` here


```suggestion
```

jdetter (approved):
I haven't tested this myself but this looks good, just one small nit 👍

PR #3355, opened 2025-10-04
feat: RLS guard on table

onx2:
# Description of Changes

Draft to add a proc macro that wraps the current client visibility filter so RLS can be applied with less boilerplate and directly on the table structs. I don't have this tested BTW just wanted a POC to gauge interest.

https://discord.com/channels/1037340874172014652/1423724267203334298

Some other things that should be considered:

- Since the intent is to attach to a struct with the table macro, maybe this should enforce that the table macro exists?
- IDK if possible but maybe even verify the SQL relates to that table name? If it isn't possible maybe we can extend the table macro to accept guards so we can read the metadata?
- Edge cases 🤷🏼
- Tests? I didn't see any for the client visibility filter though

### DIFF
--- a/crates/bindings-macro/src/guard.rs
+++ b/crates/bindings-macro/src/guard.rs
@@ -1,0 +1,97 @@
+use crate::util::cvt_attr;
+use proc_macro::TokenStream as StdTokenStream;
+use proc_macro2::TokenStream;
+use quote::{format_ident, quote};
+use std::collections::hash_map::DefaultHasher;
+use std::hash::{Hash, Hasher};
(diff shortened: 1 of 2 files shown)


bfops:
Thank you for opening this! We'll work on getting it triaged + reviewed.

joshua-spacetime:
Note, we will be deprecating RLS in favor of [module defined views](https://github.com/clockworklabs/SpacetimeDB/releases/tag/v1.8.0), and so we probably won't be making further improvements to it.

onx2:
@joshua-spacetime views look pretty awesome - I can close this 😄

PR #3364, opened 2025-10-07
Generate read-only table handles for `#[table]`

joshua-spacetime:
# Description of Changes

The `#[table]` macro now generates read-only table and index handles.

None

1

# Testing

- [x] positive and negative test cases using `ReducerContext::as_read_only()`

### DIFF
--- a/crates/bindings/src/lib.rs
+++ b/crates/bindings/src/lib.rs
@@ -667,16 +667,16 @@ pub use spacetimedb_bindings_macro::table;
 pub use spacetimedb_bindings_macro::reducer;
 
 /// One of two possible types that can be passed as the first argument to a `#[view]`.
-/// The other is [`SenderViewContext`].
+/// The other is [`ViewContext`].
 /// Use this type if the view does not depend on the caller's identity.
-pub struct ViewContext {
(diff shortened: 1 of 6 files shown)


crates/bindings/src/lib.rs
```diff
@@ -760,6 +779,20 @@ impl ReducerContext {
...
+    }
+
+    /// Create a sender-bound read-only view context using this reducer's caller.
+    pub fn as_sender_view(&self) -> SenderViewContext {
+        SenderViewContext {
+            sender: self.sender,
+            connection_id: self.connection_id,
+            db: LocalReadOnly {},
+        }
+    }
```
gefjon, line 795: Are these expected to be useful for users, or only for testing? If the latter, I think we should note it in the doc comment, and probably also label them `#[doc(hidden)]` so they don't show up in autocomplete.

gefjon (approved):
Reasonable enough, and I like the tests.

crates/bindings/src/lib.rs
joshua-spacetime: For users. This makes views callable from reducers.

PR #3366, opened 2025-10-07
`spacetime init` rewrite

drogus:
This is a draft of the new functionality for `spacetime init`. In order to run it with built-in templates you have to set the path to the config file:

```
export SPACETIMEDB_CLI_TEMPLATES_FILE=crates/cli/.init-templates.json
```

In the future it will fetch the list from GH.

A few notes:

* the previous functionality of `spacetime init` does not work at the moment
* the code needs a bit more cleanup and tests before merging
* there is a bit of a mix in how we generate empty server and client projects. For Rust we use the existing way of generating. For TypeScript we clone an empty project from the repo. I wanted to play with both ways of doing things, and I'm still not sure which is better. Generation in Rust means that the generated code will match the CLI version and not necessarily whatever is in Git. On the other hand, for the builtin templates we will be fetching the newest version from GH, which I guess might also not what we want, ie. we probably want only stable templates. More discussion is needed here
* we use `spacetimedb` directory for the server files
* I don't particularly like the inability to disable interactive mode easily. We discussed disabling it by default if all of the required  arguments are passed, but I don't think it's feature proof. For example, if someone relies on a non-interactive mode, and we add a new required argument, instead of printing a message `missing --foo`, we will automatically launch interactive mode, which is harder to debug. That's why I think I'd prefer to implement `--non-interactive` argument
* it's kind of hard to keep the legacy behaviour. If you don't pass any arguments, we go into interactive mode. In the legacy version, we would print required arguments. If someone passes `--lang` or `--project-path` explicitly, I guess we could run the legacy workflow, but not sure if it's worth it, as the command was marked as unstable anyway
* the project path defaults to the project name, but I think we should probably replace change whitespaces to dashes, or at least ask for the project path with the project name being the default (or both)


drogus:
This PR is now almost ready for a review with two caveats:

I've included TypeScript as a module language, but due to #3327 it obviously is not finalised, so I think this PR should wait for #3227, so I don't have to rip out all of the mentions of module TS, just to introduce it when the other PR is merged

I don't really like how the arguments to the subcommand turned out due to leaving the old usage intact. [here](https://github.com/clockworklabs/SpacetimeDB/pull/3366/files#diff-6ab3cda8a98239b80fdf58b9a4e89a07e17f53b3824115b9232bf8db5c22631dR324-R342) are the changes in the command output, and honestly I can't really make too much sense out of it, even though I implemented the new version. The problem is, that now we have legacy `--lang`, and a new `--server-lang`, and then a few options that are not really compatible with the old way the command worked. We need to use different arguments in order to detect if we want to run in the interactive mode, or not. More specifically, the legacy `--lang` assumes the old way, ie non-interactive mode and the old functionality, but we also need to have a case where `--server-lang` is passed just to pre-select the option from the interactive mode. There are a few ways we could fix this:

1. Remove the legacy options from the help message. This would still leave older tutorials working, but also not create confusion about the new options. That said, it may as well be confusing to not include the options if someone is wondering why `--lang` works
2. Remove the old functionality altogether
3. Write a fully custom help message that will list both modes

Personally I'd go for two as it would also allow to massively simplify the implementation, but if we really want to keep the functionality then I'd probably try 3. so the message makes more sense

The summary of the PR so far:

* the init command now goes into interactive mode by default
* the templates list is defined in a JSON file in the repo: https://github.com/clockworklabs/SpacetimeDB/pull/3366/files#diff-12a92074e42916d3615fc6c567db3f12460b81a1e9f668d69d64a4ec709426f6
* the interactive mode allows choosing one of the predefined "highlited" templates or choosing one of the other templates or a GH based template
* the old version of the command used project directories with files with suffixed extensions like `Cargo._toml`. This doesn't play too well with cloning, so I moved those projects to the `templates` directory with the files and directory structure the same as the resulting project. I'm not sure if there are any problems with that - I can't see any. It seems to me like the underscored extensions were created mostly because the templates were inside the `src` directory. With the `templates` dir outside of `src` I don't think cargo will pick them up
* we now ask users for a name and update the project name in the client side of the project, for example in the toml config, or package.json

drogus:
This should be now read to review. The way it should work is to go into interactive mode right away, and ask for options, for example:

image

If a user passes all the required options or `--non-interactive`, the command will not go into interactive mode.

cloutiertyler:
The following tests gave me an error:

image
image

drogus:
@cloutiertyler sorry about that. I'm not sure what was wrong, but yesterday when I was working on `spacetime dev` yesterday I got annoyed by the time it takes to clone stuff, so I ended up putting the templates in the binary, and now all of the templates work (as they're parsed on build time, so there is also no risk of mismatch of the git references etc). I've ran the command for all of the templates:

<details>
<summary>Output</summary>

```
Testing template: basic-rust
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.23s
     Running `target/debug/spacetimedb-cli init --name test-basic-rust --template basic-rust /var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.ZDhSpbPFx6/test-basic-rust --non-interactive`
WARNING: This command is UNSTABLE and subject to breaking changes.

You are logged in to SpacetimeDB.
Initializing project from template...
Setting up client (rust)...
Setting up server (rust)...
Project initialized successfully!

Next steps:
  cd /var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.ZDhSpbPFx6/test-basic-rust
  spacetime publish --project-path spacetimedb test-basic-rust
  spacetime generate --lang rust --out-dir src/module_bindings --project-path spacetimedb
  cargo run

Learn more: https://spacetimedb.com/docs
✓ Template basic-rust succeeded
  Generated files:
/var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.ZDhSpbPFx6/test-basic-rust/.cursorrules
/var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.ZDhSpbPFx6/test-basic-rust/Cargo.toml
/var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.ZDhSpbPFx6/test-basic-rust/README.md
/var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.ZDhSpbPFx6/test-basic-rust/spacetimedb/.cargo/config.toml
/var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.ZDhSpbPFx6/test-basic-rust/spacetimedb/.gitignore
/var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.ZDhSpbPFx6/test-basic-rust/spacetimedb/Cargo.toml
/var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.ZDhSpbPFx6/test-basic-rust/spacetimedb/src/lib.rs
/var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.ZDhSpbPFx6/test-basic-rust/src/main.rs
  Cleaning up /var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.ZDhSpbPFx6

Testing template: basic-typescript
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.52s
     Running `target/debug/spacetimedb-cli init --name test-basic-typescript --template basic-typescript /var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.3GGmtl6bHu/test-basic-typescript --non-interactive`
WARNING: This command is UNSTABLE and subject to breaking changes.

You are logged in to SpacetimeDB.
Initializing project from template...
Setting up client (typescript)...
Setting up server (typescript)...
Note: Run 'npm install' in the project directory to install dependencies
Project initialized successfully!

Next steps:
  cd /var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.3GGmtl6bHu/test-basic-typescript
  npm install
  spacetime publish --project-path spacetimedb test-basic-typescript
  spacetime generate --lang typescript --out-dir src/module_bindings --project-path spacetimedb
  npm run dev

Learn more: https://spacetimedb.com/docs

TypeScript server requires dependencies to be installed before publishing.
Please install dependencies by running your package manager's install command in the /var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.3GGmtl6bHu/test-basic-typescript/spacetimedb directory.
✓ Template basic-typescript succeeded
  Generated files:
/var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.3GGmtl6bHu/test-basic-typescript/.cursorrules
/var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.3GGmtl6bHu/test-basic-typescript/.gitignore
/var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.3GGmtl6bHu/test-basic-typescript/index.html
/var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.3GGmtl6bHu/test-basic-typescript/package.json
/var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.3GGmtl6bHu/test-basic-typescript/spacetimedb/.gitignore
/var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.3GGmtl6bHu/test-basic-typescript/spacetimedb/package.json
/var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.3GGmtl6bHu/test-basic-typescript/spacetimedb/src/index.ts
/var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.3GGmtl6bHu/test-basic-typescript/spacetimedb/tsconfig.json
/var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.3GGmtl6bHu/test-basic-typescript/src/main.ts
/var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.3GGmtl6bHu/test-basic-typescript/tsconfig.json
/var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.3GGmtl6bHu/test-basic-typescript/vite.config.ts
  Cleaning up /var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.3GGmtl6bHu

Testing template: basic-react
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.51s
     Running `target/debug/spacetimedb-cli init --name test-basic-react --template basic-react /var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.ibWx6C4Nfw/test-basic-react --non-interactive`
^C  Cleaning up /var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.ibWx6C4Nfw


SpacetimeDB on  drogus/spacetime-init [$?] via 🐳 desktop-linux is 📦 v1.6.0 via .NET v8.0.400 via  v24.8.0 via 🦀 v1.90.0 on ☁️   (us-east-1) took 5s
❯ ./test-templates.sh
Found 6 templates to test

Testing template: basic-rust
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.29s
     Running `target/debug/spacetimedb-cli init --name test-basic-rust --template basic-rust /var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.M5Kdzgmrhj/test-basic-rust --non-interactive`
WARNING: This command is UNSTABLE and subject to breaking changes.

You are logged in to SpacetimeDB.
Initializing project from template...
Setting up client (rust)...
Setting up server (rust)...
Project initialized successfully!

Next steps:
  cd /var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.M5Kdzgmrhj/test-basic-rust
  spacetime publish --project-path spacetimedb test-basic-rust
  spacetime generate --lang rust --out-dir src/module_bindings --project-path spacetimedb
  cargo run

Learn more: https://spacetimedb.com/docs
✓ Template basic-rust succeeded
  Generated files:
.cursorrules
Cargo.toml
README.md
spacetimedb/.cargo/config.toml
spacetimedb/.gitignore
spacetimedb/Cargo.toml
spacetimedb/src/lib.rs
src/main.rs
  Cleaning up /var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.M5Kdzgmrhj

Testing template: basic-typescript
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.56s
     Running `target/debug/spacetimedb-cli init --name test-basic-typescript --template basic-typescript /var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.Kz0saC5tml/test-basic-typescript --non-interactive`
WARNING: This command is UNSTABLE and subject to breaking changes.

You are logged in to SpacetimeDB.
Initializing project from template...
Setting up client (typescript)...
Setting up server (typescript)...
Note: Run 'npm install' in the project directory to install dependencies
Project initialized successfully!

Next steps:
  cd /var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.Kz0saC5tml/test-basic-typescript
  npm install
  spacetime publish --project-path spacetimedb test-basic-typescript
  spacetime generate --lang typescript --out-dir src/module_bindings --project-path spacetimedb
  npm run dev

Learn more: https://spacetimedb.com/docs

TypeScript server requires dependencies to be installed before publishing.
Please install dependencies by running your package manager's install command in the /var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.Kz0saC5tml/test-basic-typescript/spacetimedb directory.
✓ Template basic-typescript succeeded
  Generated files:
.cursorrules
.gitignore
index.html
package.json
spacetimedb/.gitignore
spacetimedb/package.json
spacetimedb/src/index.ts
spacetimedb/tsconfig.json
src/main.ts
tsconfig.json
vite.config.ts
  Cleaning up /var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.Kz0saC5tml

Testing template: basic-react
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.54s
     Running `target/debug/spacetimedb-cli init --name test-basic-react --template basic-react /var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.uw30UKWEIz/test-basic-react --non-interactive`
WARNING: This command is UNSTABLE and subject to breaking changes.

You are logged in to SpacetimeDB.
Initializing project from template...
Setting up client (typescript)...
Setting up server (typescript)...
Note: Run 'npm install' in the project directory to install dependencies
Project initialized successfully!

Next steps:
  cd /var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.uw30UKWEIz/test-basic-react
  npm install
  spacetime publish --project-path spacetimedb test-basic-react
  spacetime generate --lang typescript --out-dir src/module_bindings --project-path spacetimedb
  npm run dev

Learn more: https://spacetimedb.com/docs

TypeScript server requires dependencies to be installed before publishing.
Please install dependencies by running your package manager's install command in the /var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.uw30UKWEIz/test-basic-react/spacetimedb directory.
✓ Template basic-react succeeded
  Generated files:
.cursorrules
.gitignore
index.html
package.json
spacetimedb/package.json
spacetimedb/src/index.ts
src/App.tsx
src/main.tsx
tsconfig.json
vite.config.ts
  Cleaning up /var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.uw30UKWEIz

Testing template: quickstart-chat-rust
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.53s
     Running `target/debug/spacetimedb-cli init --name test-quickstart-chat-rust --template quickstart-chat-rust /var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.b3hRs0GfLS/test-quickstart-chat-rust --non-interactive`
WARNING: This command is UNSTABLE and subject to breaking changes.

You are logged in to SpacetimeDB.
Initializing project from template...
Setting up client (rust)...
Setting up server (rust)...
Project initialized successfully!

Next steps:
  cd /var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.b3hRs0GfLS/test-quickstart-chat-rust
  spacetime publish --project-path spacetimedb test-quickstart-chat-rust
  spacetime generate --lang rust --out-dir src/module_bindings --project-path spacetimedb
  cargo run

Learn more: https://spacetimedb.com/docs
✓ Template quickstart-chat-rust succeeded
  Generated files:
.cursorrules
README.md
main.rs
module_bindings/identity_connected_reducer.rs
module_bindings/identity_disconnected_reducer.rs
module_bindings/message_table.rs
module_bindings/message_type.rs
module_bindings/mod.rs
module_bindings/send_message_reducer.rs
module_bindings/set_name_reducer.rs
module_bindings/user_table.rs
module_bindings/user_type.rs
spacetimedb/.cargo/config.toml
spacetimedb/.gitignore
spacetimedb/Cargo.toml
spacetimedb/LICENSE
spacetimedb/README.md
spacetimedb/src/lib.rs
  Cleaning up /var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.b3hRs0GfLS

Testing template: quickstart-chat-c-sharp
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.50s
     Running `target/debug/spacetimedb-cli init --name test-quickstart-chat-c-sharp --template quickstart-chat-c-sharp /var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.28pT50NLfx/test-quickstart-chat-c-sharp --non-interactive`
WARNING: This command is UNSTABLE and subject to breaking changes.

You are logged in to SpacetimeDB.
Initializing project from template...
Setting up client (csharp)...
Setting up server (csharp)...
Project initialized successfully!

Next steps:
  cd /var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.28pT50NLfx/test-quickstart-chat-c-sharp
  spacetime publish --project-path spacetimedb test-quickstart-chat-c-sharp
  spacetime generate --lang csharp --out-dir src/module_bindings --project-path spacetimedb

Learn more: https://spacetimedb.com/docs
✓ Template quickstart-chat-c-sharp succeeded
  Generated files:
.cursorrules
Program.cs
README.md
client.csproj
module_bindings/Reducers/ClientConnected.g.cs
module_bindings/Reducers/ClientDisconnected.g.cs
module_bindings/Reducers/SendMessage.g.cs
module_bindings/Reducers/SetName.g.cs
module_bindings/SpacetimeDBClient.g.cs
module_bindings/Tables/Message.g.cs
module_bindings/Tables/User.g.cs
module_bindings/Types/Message.g.cs
module_bindings/Types/User.g.cs
spacetimedb/.gitignore
spacetimedb/Lib.cs
spacetimedb/StdbModule.csproj
spacetimedb/global.json
  Cleaning up /var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.28pT50NLfx

Testing template: quickstart-chat-typescript
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.51s
     Running `target/debug/spacetimedb-cli init --name test-quickstart-chat-typescript --template quickstart-chat-typescript /var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.AF2LkSMVCH/test-quickstart-chat-typescript --non-interactive`
WARNING: This command is UNSTABLE and subject to breaking changes.

You are logged in to SpacetimeDB.
Initializing project from template...
Setting up client (typescript)...
Setting up server (typescript)...
Note: Run 'npm install' in the project directory to install dependencies
Project initialized successfully!

Next steps:
  cd /var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.AF2LkSMVCH/test-quickstart-chat-typescript
  npm install
  spacetime publish --project-path spacetimedb test-quickstart-chat-typescript
  spacetime generate --lang typescript --out-dir src/module_bindings --project-path spacetimedb
  npm run dev

Learn more: https://spacetimedb.com/docs

TypeScript server requires dependencies to be installed before publishing.
Please install dependencies by running your package manager's install command in the /var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.AF2LkSMVCH/test-quickstart-chat-typescript/spacetimedb directory.
✓ Template quickstart-chat-typescript succeeded
  Generated files:
.cursorrules
.gitignore
CHANGELOG.md
README.md
index.html
package.json
public/vite.svg
spacetimedb/package-lock.json
spacetimedb/package.json
spacetimedb/src/index.ts
spacetimedb/tsconfig.json
src/.gitattributes
src/App.css
src/App.integration.test.tsx
src/App.tsx
src/assets/react.svg
src/index.css
src/main.tsx
src/module_bindings/identity_connected_reducer.ts
src/module_bindings/identity_disconnected_reducer.ts
src/module_bindings/index.ts
src/module_bindings/message_table.ts
src/module_bindings/message_type.ts
src/module_bindings/send_message_reducer.ts
src/module_bindings/set_name_reducer.ts
src/module_bindings/user_table.ts
src/module_bindings/user_type.ts
src/setupTests.ts
src/vite-env.d.ts
tsconfig.app.json
tsconfig.json
tsconfig.node.json
vite.config.ts
vitest.config.ts
  Cleaning up /var/folders/5c/tgs43q5s7_g1snt37h2dm8mm0000gn/T/tmp.AF2LkSMVCH

All templates tested successfully! (6/6)
```
</details>

drogus:
While looking at this in context of `spacetime dev`, I see that I'll have to adjust post generate instructions cause they don't make sense anymore with `spacetime dev` available.

crates/cli/src/subcommands/init.rs
```diff
@@ -1,28 +1,979 @@
...
+    println!("{}", "Initializing project from template...".cyan());
+
+    match config.template_type {
+        TemplateType::Builtin => init_builtin(config, project_path)?,
+        TemplateType::GitHub => init_github_template(config, project_path)?,
+        TemplateType::Empty => init_empty(config, project_path)?,
+    }
+
+    let cursorrules_content = embedded::get_cursorrules();
+    let cursorrules_path = project_path.join(".cursorrules");
```
cloutiertyler: the `cursorrules_content` should go in `.cursor/rules/spacetime-rules.mdc`. (kebab case is preferred)

[`.cursorrules` is deprecated](https://cursor.com/docs/context/rules#cursorrules)

docs/docs/modules/c-sharp/index.md
```diff
@@ -66,7 +66,7 @@ Tables and reducers in C# modules can use any type annotated with [`[SpacetimeDB
 To create a C# module, install the [`spacetime` CLI tool](https://spacetimedb.com/install) in your preferred shell. Navigate to your work directory and run the following command:
 
 ```bash
-spacetime init --lang csharp my-project-directory
+spacetime init --non-interactive --name my-spacetimedb-project --server-lang csharp my-project-directory
```
cloutiertyler: Several suggestions:

1. `--name my-spacetimedb-project my-spacetimedb-directory` is a little redundant, let's make it so that the parameter is called `--project-path` instead and that parameter tells you where you're going put the project. Let's make the positional argument the project name (more in line with what `spacetime publish` does). Let's make it so that if you do not specify a project path, it creates a directory in the current directory with the same name as the project name.
2. Let's leave off `--non-interactive` in the docs, so it might actually go into interactive mode for some stuff.
3. Let's rename `--server-lang` to `--lang` (I'm less sure about this change. I would be okay to keep it `--server-lang` if we made the other changes).

So this line would become:

```bash
spacetime init --lang csharp my-spacetimedb-project
```

crates/cli/src/subcommands/init.rs
```diff
@@ -1,28 +1,979 @@
...
+        })
+    } else {
+        // GitHub template
+        Ok(TemplateConfig {
+            project_name,
+            project_path,
+            template_type: TemplateType::GitHub,
+            server_lang: None,
+            client_lang: None,
+            github_repo: Some(template_str.to_string()),
```
jdetter, line 320: Should we check to see if this looks like a URL here? Like if someone just mistypes a template name we will end up here too no?

crates/cli/src/subcommands/init.rs
```diff
@@ -1,28 +1,979 @@
...
+}
+
+fn clone_github_template(repo_input: &str, target: &Path) -> anyhow::Result<()> {
+    let repo_url = if repo_input.starts_with("http") {
+        repo_input.to_string()
+    } else if repo_input.contains('/') {
+        format!("https://github.com/{}", repo_input)
+    } else {
+        anyhow::bail!("Invalid repository format. Use 'owner/repo' or full URL");
+    };
```
jdetter, line 750: If someone wants to use a template from a private repo or they just want to use SSH auth I think it would be nice to support that here. e.g. if someone passed this as their template URL:
```
# Either of these should work
git@github.com/clockworklabs/PrivateTemplate
ssh://git@github.com/clockworklabs/PrivateTemplate
```

crates/cli/src/subcommands/init.rs
```diff
@@ -1,28 +1,979 @@
...
+            template_def: None,
+            use_local: true,
+        })
+    } else {
+        unreachable!("Invalid selection index")
+    }
+}
+
+fn clone_github_template(repo_input: &str, target: &Path) -> anyhow::Result<()> {
+    let repo_url = if repo_input.starts_with("http") {
```
jdetter: This wouldn't catch URLs like:
```
github.com/clockworklabs/MyTemplate
```

Also do we want to support pointing to local repo as well? Something like:
```
../clockwork/SomeTemplateImTesting
```

crates/cli/src/subcommands/init.rs
```diff
@@ -1,28 +1,979 @@
...
+            github_repo: None,
+            template_def: None,
+            use_local: true,
+        })
+    } else {
+        unreachable!("Invalid selection index")
+    }
+}
+
+fn clone_github_template(repo_input: &str, target: &Path) -> anyhow::Result<()> {
```
jdetter: Something that I think would be cool is for clockwork to have their own repository for templates, something like:

```
https://github.com/clockworklabs/spacetimedb-templates
```

And then inside of this repo we can ship templates that we perhaps don't want to ship with the CLI like:

```
spacetimedb-templates
├── UnityTemplate
├── UnityPhysicsTemplate
├── UnrealEngineTemplate
├── Blackholio
│   ├── Unity
│   ├── Unreal
│   └── Three.js
├── VoiceChatTemplate
└── etc.
```

Is this/will this be supported?

To clarify: this is what I want to be able to do:
```
spacetime init --template https://github.com/clockworklabs/spacetimedb-templates/Blackholio/Unity
```

jdetter (commented):
Some UX questions/nits about github templates, I'm still reviewing

crates/cli/src/subcommands/init.rs
cloutiertyler: It's not quite supported now, but I think in the spirit of the monorepo, that should just be a folder inside of `clockworklabs/SpacetimeDB` and we can definitely do this in a follow up PR.

`templates` (and maybe we deprecate the `modules` folder since those should probably just be inside the `spacetimedb` directory of those modules.

I fear that if we have a separate repo, we'll break them and not realize it. I think it would also help with testing. We can test every template as part of the approval process.

jdetter: I think having them part of SpacetimeDB is a good idea. So then you should be able to do something like this then right?

```
spacetime init --template https://github.com/clockworklabs/SpacetimeDB/templates/Blackholio/Unity
```

cloutiertyler: Nah I think just `spacetime init --template blackholio` and we'll know that's in our repo if it isn't built into the CLI already.

crates/cli/src/subcommands/init.rs
```diff
@@ -1,28 +1,979 @@
...
+        spacetimedb_login_force(config, &host, false).await?;
+        println!("{}", "Successfully logged in!".green());
+        Ok(true)
+    } else {
+        println!("{}", "Continuing with local deployment.".yellow());
+        Ok(false)
+    }
+}
+
+fn slugify(name: &str) -> String {
```
jdetter, line 193: Do we not just want to use the `convert_case` crate for this instead? Then you can just:
```
name.to_case(Case::Kebab)
```
Plus we already use this dependency anyway

crates/cli/src/subcommands/init.rs
```diff
@@ -1,28 +1,979 @@
...
+            .items(&choices)
+            .default(0)
+            .interact()?;
+
+        match selection {
+            0 => Some("npm"),
+            1 => Some("pnpm"),
+            2 => Some("yarn"),
+            3 => Some("bun"),
+            _ => None,
```
jdetter: I think if the user selected "other" here we should prompt them for the package manager they want to use

crates/cli/src/subcommands/init.rs
```diff
@@ -1,28 +1,979 @@
...
+            template_def: Some(template.clone()),
+            use_local: true,
+        })
+    } else if client_selection == other_index {
+        println!("\n{}", "Available built-in templates:".bold());
+        for template in &templates {
+            println!("  {} - {}", template.id, template.description);
+        }
+        println!();
+        loop {
```
jdetter, line 676: This loop isn't used

crates/cli/src/subcommands/init.rs
cloutiertyler: It's not quite so simple, I think in this case they are doing something that 99.9% of users won't do, so we should leave them to their own devices.

jdetter (commented):
More review

crates/cli/src/subcommands/init.rs
jdetter: Let's remove the "other" option then?

crates/cli/src/subcommands/init.rs
```diff
@@ -1,28 +1,1003 @@
...
+            update_package_json(&server_dir, &config.project_name)?;
+        }
+        Some(ServerLanguage::Rust) => {
+            update_cargo_toml_name(&server_dir, &config.project_name)?;
+        }
+        Some(ServerLanguage::Csharp) => {}
+        None => {}
+    }
+
+    // Update client name
```
cloutiertyler: The challenge with this is that it makes it so that when you run `spacetime dev` the project name is inserted here instead of the actual database that is used when you run `dev`. I think in an ideal world we'd want to make this an argument that we somehow pass into the client code. I think the easiest way to do that is an environment variable. We'll have to configure every client template to be able to read the environment variable though, and warn when you run the app if it's not set or something.

crates/cli/src/subcommands/init.rs
cloutiertyler: This LGTM now. It's supposed to be .mdc though, I'll fix that.

crates/cli/src/subcommands/init.rs
cloutiertyler: Fixed.

crates/cli/src/subcommands/init.rs
cloutiertyler: I made this commit to address this problem: https://github.com/clockworklabs/SpacetimeDB/pull/3366/commits/5c712118bb956ead8982c8d7feca0ac3cb43dd6b

We can now set these variables in `spacetime dev` directly so that `spacetime dev --template` or `spacetime init --template && spacetime dev` just works.

crates/cli/src/subcommands/init.rs
drogus: If we remove the "other" option you would be forced to choose one of the package managers, or am I missing something?

crates/cli/src/subcommands/init.rs
drogus: Oh, yeah, if we have it anyway, I'll use it, I haven't noticed. When doing simple stuff like that I sometimes prefer to just do it myself rather than bring in a whole dependency, but if it's already there, it's a no-brainer.

crates/cli/src/subcommands/init.rs
drogus: Regarding the first thing, I guess we could support that. Initially I didn't think about it because I thought it might be hard to distinguish between a username and a host, but now I checked and GH usernames can't have dots, so it should be fine. Also no idea why I only check for `http`  and not https 😅

Regarding the econd question - I would say no, ot at least not in this form as we could only support paths starting with `..` or `/`. Maybe something like `file://`? Not really sure

crates/cli/src/subcommands/init.rs
drogus: That's fair, I'll make this check more robust

crates/cli/src/subcommands/init.rs
cloutiertyler: I've fixed this by checking first if it looks built-in, and if it does reprompting if it's not a built in template. If it is not built-in-like then I assume it must be a git url.

crates/cli/src/subcommands/init.rs
cloutiertyler: I have made this check more robust:

```rs
    let is_git_url = |s: &str| {
        s.starts_with("git@") || s.starts_with("ssh://") || s.starts_with("http://") || s.starts_with("https://")
    };

    let repo_url = if is_git_url(repo_input) {
        repo_input.to_string()
    } else if repo_input.contains('/') {
        format!("https://github.com/{}", repo_input)
    } else {
        anyhow::bail!("Invalid repository format. Use 'owner/repo' or full git clone URL");
    };
```

crates/cli/src/subcommands/init.rs
cloutiertyler: This is now:

```rs
fn slugify(name: &str) -> String {
    name.to_case(Case::Kebab)
}
```

crates/cli/src/subcommands/init.rs
cloutiertyler: You are not missing something. Other is just "let me handle it myself", not I want to use something else that I want you to set up. We could also call it "None" because all it does is skip over the package manager and tells you to install dependencies manually.

cloutiertyler: I changed it to `none` since that seems a bit less confusing than `other` and is effectively what we're doing.

crates/cli/src/subcommands/init.rs
cloutiertyler: I resolved this:

```rs
    let is_git_url = |s: &str| {
        s.starts_with("git@") || s.starts_with("ssh://") || s.starts_with("http://") || s.starts_with("https://")
    };

    let repo_url = if is_git_url(repo_input) {
        repo_input.to_string()
    } else if repo_input.contains('/') {
        format!("https://github.com/{}", repo_input)
    } else {
        anyhow::bail!("Invalid repository format. Use 'owner/repo' or full git clone URL");
    };
```

cloutiertyler: > Also no idea why I only check for http and not https 

You check for a prefix, so it covers both, but it also would make it so a project called `http-foobar` would mistakenly be detected as a url.

docs/docs/03-Unity Tutorial/03-part-2.md
```diff
@@ -23,7 +23,7 @@ In your `blackholio` directory, run the following command to initialize the Spac
 Run the following command to initialize the SpacetimeDB server module project with Rust as the language:
 
 ```bash
-spacetime init --lang=rust server-rust
+spacetime init --lang rust server-rust
```
jdetter: In this case I'm being prompted for a project path, I think most users would be smart enough to just press enter to do the default thing but maybe it would be good to just include the `--project-path` option here so that users don't get prompted?

docs/docs/03-Unity Tutorial/03-part-2.md
```diff
@@ -23,7 +23,7 @@ In your `blackholio` directory, run the following command to initialize the Spac
...
 ```
 
 This command creates a new folder named `server-rust` alongside your Unity project `client-unity` directory and sets up the SpacetimeDB server project with Rust as the programming language.
```
jdetter: The instructions now create a directory structure which looks like this:
```
boppy@geralt MINGW64 ~/clockwork/tmp
$ ls
server-rust/

boppy@geralt MINGW64 ~/clockwork/tmp
$ ls server-rust/
spacetimedb/

boppy@geralt MINGW64 ~/clockwork/tmp
$ ls server-rust/spacetimedb/
Cargo.toml  src/
```

I think we need to update the tutorials to use the new flow otherwise some of these steps will be out of date and won't work. For Unity specifically we might want people to run the `spacetime init ...` command first so that they can then open Unity in the project directory so that Unity can populate the required files. Although I'm not sure if you can create a new Unity project in a directory which already contains files.

I'm happy to help with this if needed 🙂

jdetter (commented):
Some docs feedback for now

jdetter (approved):
I've tested 
- [x] Fixed the --server-only bug which prevented generating the module code
- [x] Finished testing the Unity tutorial for both rust and C# modules up to the spacetime generate ... part
- [x] Test Unreal tutorial with both C# and rust up to the spacetime generate ... part
- [x] Fix the issue with Cargo.toml in the rust quickstart
- [x] Test rust module quickstart
- [x] Test C# module quickstart
- [ ] Test typescript module quickstart

There are a few code things that I would probably change but I think we can do those as follow-up PRs

cloutiertyler (approved):
Okay, John and I have thoroughly tested and fixed small bugs. There are two remaining items that are pretty gnarly in this version:

1. We scoop up the slop from all over the repo, and we realllllly should change that to be a straightforward copy of future `/templates` directory
2. We need a better gameplan for ensuring that the templates locally depend on SpacetimeDB if you have the repo checked out, but use a package manager/versioning scheme if you copy them as a template. It's pretty slapdash right now, but it kinda works. We should make it better (standardized).

I'm okay to do these things later.

PR #3308, opened 2025-09-29
Integrate `tools/upgrade-version` into cargo and expand

bfops:
# Description of Changes

`tools/upgrade-version` can now be run via just `cargo bump-versions`.

By default, it does nothing. You must now specify at least one of: `--rust-and-cli`, `--typescript`, or `--csharp`.

This also now bumps the `StdbModule.csproj` version in `demo/Blackholio`. I'm not sure if that's desired or not, but it was still at `1.0.0` which doesn't seem quite right.

# API and ABI breaking changes

None

# Expected complexity level and risk

2

# Testing

- [x] `cargo bump-versions 1.5.0 --rust-and-cli`
```
diff --git a/Cargo.lock b/Cargo.lock
index 941b9d66d..3800e547f 100644
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -436,3 +436,3 @@ dependencies = [
  "anyhow",
- "spacetimedb 1.4.0",
+ "spacetimedb 1.5.0",
 ]
@@ -1023,3 +1023,3 @@ dependencies = [
 name = "connect_disconnect_client"
-version = "1.4.0"
+version = "1.5.0"
 dependencies = [
@@ -3089,3 +3089,3 @@ dependencies = [
  "log",
- "spacetimedb 1.4.0",
+ "spacetimedb 1.5.0",
 ]
@@ -3414,3 +3414,3 @@ dependencies = [
  "log",
- "spacetimedb 1.4.0",
+ "spacetimedb 1.5.0",
 ]
@@ -3897,3 +3897,3 @@ dependencies = [
  "log",
- "spacetimedb 1.4.0",
+ "spacetimedb 1.5.0",
 ]
@@ -4382,3 +4382,3 @@ dependencies = [
  "log",
- "spacetimedb 1.4.0",
+ "spacetimedb 1.5.0",
 ]
@@ -5077,3 +5077,3 @@ dependencies = [
  "paste",
- "spacetimedb 1.4.0",
+ "spacetimedb 1.5.0",
 ]
@@ -5082,3 +5082,3 @@ dependencies = [
 name = "sdk-unreal-test-harness"
-version = "1.4.0"
+version = "1.5.0"
 dependencies = [
@@ -5456,3 +5456,3 @@ dependencies = [
  "log",
- "spacetimedb 1.4.0",
+ "spacetimedb 1.5.0",
 ]
@@ -5479,3 +5479,3 @@ dependencies = [
 name = "spacetimedb"
-version = "1.4.0"
[...]

PR #3369, opened 2025-10-07
Fix Rust dep versions

bfops:
# Description of Changes

It turns out that cargo automatically uses the latest semver-compatible versions of dependencies, which is not what we expected. tl;dr specifying `1.5.0` actually means `>=1.5.0 <2.0.0`, but we actually intend `1.5.*`.

This PR updates our `upgrade-version` tool, and re-runs it to fix the dep versions.

None.

1

# Testing

- [x] I ran `cargo bump-versions 1.5.0 --rust-and-cli` to regenerate the other committed files.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -98,38 +98,38 @@ edition = "2021"
 rust-version = "1.88.0"
 
 [workspace.dependencies]
-spacetimedb = { path = "crates/bindings", version = "1.5.0" }
-spacetimedb-auth = { path = "crates/auth", version = "1.5.0" }
-spacetimedb-bindings-macro = { path = "crates/bindings-macro", version = "1.5.0" }
-spacetimedb-bindings-sys = { path = "crates/bindings-sys", version = "1.5.0" }
-spacetimedb-cli = { path = "crates/cli", version = "1.5.0" }
-spacetimedb-client-api = { path = "crates/client-api", version = "1.5.0" }
-spacetimedb-client-api-messages = { path = "crates/client-api-messages", version = "1.5.0" }
-spacetimedb-codegen = { path = "crates/codegen", version = "1.5.0" }
-spacetimedb-commitlog = { path = "crates/commitlog", version = "1.5.0" }
-spacetimedb-core = { path = "crates/core", version = "1.5.0" }
-spacetimedb-data-structures = { path = "crates/data-structures", version = "1.5.0" }
-spacetimedb-datastore = { path = "crates/datastore", version = "1.5.0" }
-spacetimedb-durability = { path = "crates/durability", version = "1.5.0" }
-spacetimedb-execution = { path = "crates/execution", version = "1.5.0" }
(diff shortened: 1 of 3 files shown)


tools/upgrade-version/src/main.rs
```diff
@@ -90,6 +90,11 @@ fn main() -> anyhow::Result<()> {
                 .action(clap::ArgAction::SetTrue)
                 .help("Also bump versions in C# SDK and templates"),
         )
+        .group(
+            ArgGroup::new("update-targets")
+                .args(["typescript", "rust-and-cli", "csharp"])
+                .required(true),
+        )
```
bfops: this is unrelated, but I noticed that it was possible to run `cargo bump-versions 1.5.0` without any args, which is a no-op since https://github.com/clockworklabs/SpacetimeDB/pull/3308.

crates/cli/src/subcommands/project/rust/Cargo._toml
```diff
@@ -9,5 +9,5 @@ edition = "2021"
 crate-type = ["cdylib"]
 
 [dependencies]
-spacetimedb = "1.5"
+spacetimedb = "1.5.*"
```
jdetter, line 12: What is the difference between these two?

Cargo.toml
```diff
@@ -98,38 +98,38 @@ edition = "2021"
...
-spacetimedb-sats = { path = "crates/sats", version = "1.5.0" }
-spacetimedb-schema = { path = "crates/schema", version = "1.5.0" }
-spacetimedb-standalone = { path = "crates/standalone", version = "1.5.0" }
-spacetimedb-sql-parser = { path = "crates/sql-parser", version = "1.5.0" }
-spacetimedb-table = { path = "crates/table", version = "1.5.0" }
-spacetimedb-vm = { path = "crates/vm", version = "1.5.0" }
-spacetimedb-fs-utils = { path = "crates/fs-utils", version = "1.5.0" }
-spacetimedb-snapshot = { path = "crates/snapshot", version = "1.5.0" }
-spacetimedb-subscription = { path = "crates/subscription", version = "1.5.0" }
+spacetimedb = { path = "crates/bindings", version = "1.5.*" }
```
jdetter: Are we certain this doesn't need to be `=1.5.*`?

jdetter (commented):
Few questions on this

crates/cli/src/subcommands/project/rust/Cargo._toml
bfops: I added a comment here: https://github.com/clockworklabs/SpacetimeDB/pull/3369/files#diff-bb5690efd774c360b755a9325775ee4f1661069ba805cb09b13f467a2f63953eR130.

tl;dr Cargo will roll forward the original thing by a minor version, but the new thing only by a patch version.

Cargo.toml
bfops: yes, but updated to `=1.5.0` after further discussion

PR #3370, opened 2025-10-07
Misc cleanups in `tools/upgrade-version`

bfops:
# Description of Changes

* Make sure the user provides at least one of `--rust-and-cli`, `--typescript`, or `--csharp`, since providing none of them is a no-op as of #3308 
* Do a semver-parsing of the arg before doing anything, and use that parsed version everywhere
* Consolidate some version strings that we were computing in a few places

None

1

# Testing

- [x] Running `cargo bump-versions 1.5.0 --typescript --rust-and-cli --csharp` only shows a diff in the change dates

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -183,7 +183,7 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 20 # on a successful run, runs in 8 minutes
     container:
-      image: rust:1.88.0
+      image: rust:1.90.0
       options: --privileged
     # filter for a comment containing 'benchmarks please'
     if: ${{ github.event_name != 'issue_comment' || (github.event.issue.pull_request && contains(github.event.comment.body, 'benchmarks please')) }}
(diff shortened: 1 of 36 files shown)


tools/upgrade-version/src/main.rs
```diff
@@ -90,9 +90,19 @@ fn main() -> anyhow::Result<()> {
                 .action(clap::ArgAction::SetTrue)
                 .help("Also bump versions in C# SDK and templates"),
         )
+        .group(
+            ArgGroup::new("update-targets")
+                .args(["typescript", "rust-and-cli", "csharp"])
+                .required(true)
+                .multiple(true),
```
jdetter, line 104: Would a `--all` arg be useful for specifying you just want to bump all of them?

bfops: I think I worry that people will reflexively type --all instead of thinking about what they actually want to bump.. But I think that's less of a problem than people forgetting to bump things, so yeah I should add that.

bfops: Added `--all`

PR #3384, opened 2025-10-07
Pin some v8 dependencies

gefjon:
This will be meaningless without context from the corresponding internal PR. Reviewers should look there.


gefjon:
Well, this appears to not work, so I'm gonna close.

PR #3386, opened 2025-10-08
Add a SpacetimeDBException to C# SDK to aid in debugging

rekhoff:
# Description of Changes

This implements a fix to an issue with debugging using the C# SDK, by adding new exception types:
* `SpacetimeDBException`, 
* `SpacetimeDBArgumentException`,
* `SpacetimeDBEmptyReducerNameException`

Additional, regenerating bindings will now allow clients to report an `SpacetimeDBEmptyReducerNameException` rather than an `ArgumentOutOfRangeException` when the client receives an empty reducer name from the server.

An example of the generated code currently, that results in the exception: https://github.com/clockworklabs/SpacetimeDB/blob/544e2edc2d1f7d1dd118832a815b6dbd7a6c1d82/sdks/csharp/examples~/quickstart-chat/client/module_bindings/SpacetimeDBClient.g.cs#L475

Note: Normally this is not an issue for a client, because the `SpacetimeDBEmptyReducerNameException` would be caught by the [Try/Catch](https://github.com/clockworklabs/SpacetimeDB/blob/9f59118e24449cdd2d3e182bd44fdb26078e921b/sdks/csharp/src/SpacetimeDBClient.cs#L421C25-L421C42) statement, but a debugger will still catch the exception and halt operation. This can be annoying to a developer when debugging.

By separating out the exception into a custom `Exception` type, we allow a developer to flag the new exception type as something it can ignore, without ignoring other relevant exceptions.

Not a breaking change.
Clients will need to regenerate bindings to get the new exceptions

1

# Testing

- [X] Tested regenerating bindings and confirmed intended output.
- [X] Tested debugging and receiving `SpacetimeDBEmptyReducerNameException` when expected.


sdks/csharp/examples~/regression-tests/republishing/client/client.csproj
```diff
@@ -8,7 +8,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="SpacetimeDB.ClientSDK" Version="1.5.0" />
+    <ProjectReference Include="../../../../SpacetimeDB.ClientSDK.csproj" />
```
jdetter, line 11: oof, is this why the regression tests weren't picking up the local changes?

jdetter (approved):
Haven't tested this myself but I was involved in the early testing of this PR and also this is quite simple. LGTM 👍

PR #3388, opened 2025-10-08
Noa/ts module host changes

coolreader18:
# Description of Changes

Host-side changes extracted from #3327 

I added AUTO_INC_OVERFLOW even though we don't currently ever return it, in order to future-proof so it's already there when we start emitting it.

Prepublish was failing because it was expecting a wasm module unconditionally, so now it takes ?host_type.

I tweaked JS deser to accept null/undefined when the unit type or an option type is expected.

I switched to bsatn, because the native sats->js translator wasn't matching what js was expecting.

I renamed the sys module: my thinking is that `spacetime:` as a scheme will help disambiguate it, and maybe it could also be used for IMC in the future or something? And I believe we had discussed wanting this to be versioned, similar to wasm imports.

Trying to get a borrowed str from deserialize_js doesn't work, because v8 strings don't store utf8.

# Testing

- [x] All this was done in the course of getting an actual typescript module to successfully publish.

### DIFF
--- a/crates/core/src/host/v8/mod.rs
+++ b/crates/core/src/host/v8/mod.rs
@@ -801,19 +801,7 @@ js error Uncaught Error: foobar
     fn call_describe_module_works() {
         let code = r#"
             function __describe_module__() {
-                return {
-                    "tag": "V9",
-                    "value": {
-                        "typespace": {
-                            "types": [],
-                        },
-                        "tables": [],
-                        "reducers": [],
-                        "types": [],
-                        "misc_exports": [],
-                        "row_level_security": [],
-                    },
-                };
+                return new Uint8Array([1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]);
             }
         "#;
         let raw_mod = with_script_catch(code, call_describe_module).map_err(|e| e.to_string());


crates/client-api/src/routes/database.rs
```diff
@@ -701,12 +701,14 @@ pub struct PrePublishParams {
...
     style: PrettyPrintStyle,
+    #[serde(default)]
+    host_type: HostType,
 }
 
 pub async fn pre_publish<S: NodeDelegate + ControlStateDelegate>(
     State(ctx): State<S>,
     Path(PrePublishParams { name_or_identity }): Path<PrePublishParams>,
-    Query(PrePublishQueryParams { style }): Query<PrePublishQueryParams>,
+    Query(PrePublishQueryParams { style, host_type }): Query<PrePublishQueryParams>,
```
Centril, line 711: Needs unstable gating for JS?

crates/core/src/host/v8/de.rs
```diff
@@ -131,6 +131,10 @@ impl<'de, 'this, 'scope: 'de> de::Deserializer<'de> for Deserializer<'this, 'sco
     deserialize_primitive!(deserialize_f32, f32);
 
     fn deserialize_product<V: ProductVisitor<'de>>(self, visitor: V) -> Result<V::Output, Self::Error> {
+        if visitor.product_len() == 0 && self.input.is_null_or_undefined() {
```
Centril, line 135: Can you leave a comment as to why we do this / why this change was necessary?

crates/core/src/host/v8/de.rs
```diff
@@ -149,6 +153,15 @@ impl<'de, 'this, 'scope: 'de> de::Deserializer<'de> for Deserializer<'this, 'sco
 
     fn deserialize_sum<V: SumVisitor<'de>>(self, visitor: V) -> Result<V::Output, Self::Error> {
         let scope = &*self.common.scope;
+
+        if visitor.is_option() {
```
Centril, line 160: Can you leave a comment as to why we do this / why this change was necessary?

crates/sats/src/de.rs
```diff
@@ -753,3 +748,126 @@ impl<'de, D: Deserializer<'de>> VariantAccess<'de> for SomeAccess<D> {
         seed.deserialize(self.0)
     }
 }
+
+pub struct UnitAccess<E>(PhantomData<E>);
```
Centril, line 754: Can you add a doc comment, including why it need to exist and why the former code was insufficient?

crates/core/src/host/v8/mod.rs
```diff
@@ -454,14 +454,23 @@ fn eval_module<'scope>(
     // Evaluate the module.
     let value = module.evaluate(scope).ok_or_else(exception_already_thrown)?;
 
+    if module.get_status() == v8::ModuleStatus::Errored {
+        return Err(error::ExceptionValue(module.get_exception()).throw(scope));
+    }
```
Centril, line 462: Why is this necessary?
`exception_already_thrown` should have already taken care of this?

crates/core/src/host/v8/mod.rs
```diff
@@ -454,14 +454,23 @@ fn eval_module<'scope>(
...
+
+    let value = value.cast::<v8::Promise>();
+    if value.state() == v8::PromiseState::Pending {
+        return Err(error::TypeError("module has top-level await and is pending").throw(scope));
+    }
```
Centril, line 469: This needs some commentary as to why etc.

crates/core/src/host/v8/mod.rs
coolreader18: No, because it wasn't already thrown, afaict. We need to throw it.

crates/core/src/host/v8/syscall.rs
```diff
@@ -104,7 +104,7 @@ fn register_sys_module<'scope>(scope: &mut PinScope<'scope, '_>) -> Local<'scope
     )
 }
 
-const SYS_MODULE_NAME: &StringConst = str_from_ident!(spacetimedb_sys);
+const SYS_MODULE_NAME: &StringConst = &StringConst::new("spacetime:sys@1.0");
```
cloutiertyler, line 107: I think this is an intelligent change. Thanks for spotting it.

crates/core/src/host/v8/de.rs
```diff
@@ -131,6 +131,7 @@ impl<'de, 'this, 'scope: 'de> de::Deserializer<'de> for Deserializer<'this, 'sco
     deserialize_primitive!(deserialize_f32, f32);
 
     fn deserialize_product<V: ProductVisitor<'de>>(self, visitor: V) -> Result<V::Output, Self::Error> {
+        // In `ProductType.serializeValue()`, null/undefined is accepted for the unit type.
```
Centril: Can you amend this to tell the reader where `ProductType.serializeValue()` comes from, i.e., the TS SDK?

crates/core/src/host/v8/de.rs
```diff
@@ -154,6 +155,8 @@ impl<'de, 'this, 'scope: 'de> de::Deserializer<'de> for Deserializer<'this, 'sco
     fn deserialize_sum<V: SumVisitor<'de>>(self, visitor: V) -> Result<V::Output, Self::Error> {
         let scope = &*self.common.scope;
 
+        // In `SumType.serializeValue()`, option is treated specially - null/undefined marks none,
```
Centril: Can you amend this to tell the reader where `SumType.serializeValue()` comes from, i.e., the TS SDK?

crates/core/src/host/v8/mod.rs
```diff
@@ -454,14 +454,28 @@ fn eval_module<'scope>(
     // Evaluate the module.
     let value = module.evaluate(scope).ok_or_else(exception_already_thrown)?;
 
+    if module.get_status() == v8::ModuleStatus::Errored {
+        // If there's an exception while evaluating the code of the module, `evaluate()` won't
+        // throw, but instead the status will be Errored and the exception can be obtained from
```
```suggestion
        // throw, but instead the status will be `Errored` and the exception can be obtained from
```

PR #3392, opened 2025-10-09
ModuleDef and schema changes to support procedures

gefjon:
# Description of Changes

This commit extends various schema and schema-adjacent structures to describe procedures, a new kind of database function which are allowed to perform side effects.

This includes extending `RawModuleDefV9` with a way to register `RawProcedureDefV9`s in the `misc_exports`, preserving compatibility with modules that predate procedures.

The module validation path is reorganized somewhat to validate various properties related to procedures while preserving code clarity and maintainability.

Additionally, the `ArgsTuple` machinery for ser/de-ing reducer arguments using the argument type as a seed is extended to also support procedure arguments.

All of this is currently unused.

Additive and backwards-compatible additions to `RawModuleDefV9` and friends.

2 - some minor complexity in schema validation which may have gotten borked in a merge at some point.

# Testing

Unsure what tests would be useful, open to suggestions from reviewers.

- [ ] 
- [ ]


crates/schema/src/def/deserialize.rs
```diff
@@ -1,46 +1,93 @@
 //! Helpers to allow deserializing data using a ReducerDef.
 
-use crate::def::ReducerDef;
-use spacetimedb_lib::sats::{self, de, ProductValue};
-
-/// Wrapper around a `ReducerDef` that allows deserializing to a `ProductValue` at the type
-/// of the reducer's parameter `ProductType`.
-#[derive(Clone, Copy)]
-pub struct ReducerArgsDeserializeSeed<'a>(pub sats::WithTypespace<'a, ReducerDef>);
```
joshua-spacetime, line 9: Could we explicitly parameterize this by `ReducerDef` instead of implicitly through a macro?

crates/core/src/host/mod.rs
```diff
@@ -31,6 +30,9 @@ pub use host_controller::{
 pub use module_host::{ModuleHost, NoSuchModule, ReducerCallError, UpdateDatabaseResult};
 pub use scheduler::Scheduler;
 
+/// Encoded arguments to a database function.
+///
+/// Despite the name, this may be arguments to either a reducer or a procedure.
 #[derive(Debug)]
 pub enum ReducerArgs {
```
joshua-spacetime: Should we name this to `FunctionArgs` then?

crates/primitives/src/ids.rs
```diff
@@ -116,3 +116,35 @@ system_id! {
...
+        }
+    }
+    pub fn as_procedure(self) -> Option<ProcedureId> {
+        if let Self::Procedure(id) = self {
+            Some(id)
+        } else {
+            None
+        }
+    }
+}
```
Centril:
```suggestion
#[derive(Clone, Copy, Debug, EnumAsInner)]
pub enum FunctionId {
    Reducer(ReducerId),
    Procedure(ProcedureId),
}
```

This gives you `.into_reducer()`, etc.
Needs to be added to `Cargo.toml` for this crate, but is already part of the crate graph.

crates/core/src/host/mod.rs
Centril: I second this, but a (reworked) comment would still be useful.
> A database function is either a procedure or a reducer.

crates/core/src/host/mod.rs
```diff
@@ -114,6 +123,14 @@ pub struct InvalidReducerArguments {
     reducer: Box<str>,
 }
 
+#[derive(thiserror::Error, Debug)]
+#[error("invalid arguments for procedure {procedure}: {err}")]
+pub struct InvalidProcedureArguments {
```
Centril: The difference between these are only the rendering, so consider making these two into newtypes over a `InvalidFunctionArguments`, e.g., `struct InvalidReducerArguments(InvalidFunctionArguments);`. The newtypes then deal with the rendering, but `FunctionArgs::into_tuple` can then return `InvalidFunctionArguments` and you don't need `into_tuple_for_procedure`.

> If one of the additional expression arguments needs to refer to a field of the struct or enum, then refer to named fields as .var and tuple fields as .0.

https://docs.rs/thiserror/latest/thiserror/#details

crates/lib/src/db/raw_def/v9.rs
```diff
@@ -82,9 +82,16 @@ pub struct RawModuleDefV9 {
     pub types: Vec<RawTypeDefV9>,
 
     /// Miscellaneous additional module exports.
+    ///
+    /// The enum `RawMiscModuleExportV9` can have new variants added
```
Centril:
```suggestion
    /// The enum [`RawMiscModuleExportV9`] can have new variants added
```

crates/lib/src/db/raw_def/v9.rs
```diff
@@ -307,7 +314,9 @@ pub struct RawScheduleDefV9 {
     /// Even though there is ABSOLUTELY NO REASON TO.
     pub name: Option<Box<str>>,
 
-    /// The name of the reducer to call.
+    /// The name of the reducer or procedure to call.
+    ///
+    /// Despite the field name here, this may be either a reducer or a procedure.
```
Centril, line 319: (Future work: Could be useful to have a facility in sats to allow renaming.)

crates/schema/src/schema.rs
```diff
@@ -964,8 +964,8 @@ impl Schema for ScheduleSchema {
     fn check_compatible(&self, _module_def: &ModuleDef, def: &Self::Def) -> Result<(), anyhow::Error> {
         ensure_eq!(&self.schedule_name[..], &def.name[..], "Schedule name mismatch");
         ensure_eq!(
-            &self.reducer_name[..],
-            &def.reducer_name[..],
+            &self.function_name[..],
+            &def.function_name[..],
             "Schedule reducer name mismatch"
```
Centril:
```suggestion
            "Schedule function name mismatch"
```
(`check_compatible` erroring shouldn't happen normally so this is fine.)

crates/schema/src/def.rs
```diff
@@ -243,6 +254,23 @@ impl ModuleDef {
...
+    ) -> Option<(ProcedureId, &ProcedureDef)> {
+        // If the string IS a valid identifier, we can just look it up.
+        self.procedures.get_full(name).map(|(idx, _, def)| (idx.into(), def))
+    }
+
+    pub fn procedure_by_id(&self, id: ProcedureId) -> &ProcedureDef {
+        &self.procedures[id.idx()]
+    }
+
+    pub fn get_procedure_by_id(&self, id: ProcedureId) -> Option<&ProcedureDef> {
```
Centril, line 272: Missing docs on these two.

crates/schema/src/def.rs
```diff
@@ -342,13 +370,18 @@ impl From<ModuleDef> for RawModuleDefV9 {
...
             row_level_security_raw,
+            procedures,
         } = val;
 
         RawModuleDefV9 {
             tables: to_raw(tables),
             reducers: reducers.into_iter().map(|(_, def)| def.into()).collect(),
             types: to_raw(types),
-            misc_exports: vec![],
+            // TODO: Do we need to include default values here?
```
Centril, line 382: Probably yeah, can you file an issue?

crates/schema/src/def/validate/v9.rs
```diff
@@ -51,6 +52,28 @@ pub fn validate(def: RawModuleDefV9) -> Result<ModuleDef> {
...
 
+    let (procedures, non_procedure_misc_exports) =
+        misc_exports
+            .into_iter()
+            .partition::<Vec<RawMiscModuleExportV9>, _>(|misc_export| {
+                matches!(misc_export, RawMiscModuleExportV9::Procedure(_))
+            });
+
+    let procedures = procedures
+        .into_iter()
```
Centril, line 65: This ain't perf sensitive or anything, but alternatively, you could consider `partition_in_place` combined with `.drain`. This should avoid the two allocations.

crates/schema/src/def/validate/v9.rs
```diff
@@ -897,32 +979,44 @@ fn identifier(name: Box<str>) -> Result<Identifier> {
...
+/// then record the scheduled function's [`FunctionKind`] in the [`ScheduleDef`].
+fn check_scheduled_functions_exist(
+    tables: &mut IdentifierMap<TableDef>,
     reducers: &IndexMap<Identifier, ReducerDef>,
+    procedures: &IndexMap<Identifier, ProcedureDef>,
 ) -> Result<()> {
+    let validate_params =
+        |params_from_function: &ProductType, table_row_type_ref: AlgebraicTypeRef, function_name: &str| {
+            if params_from_function.elements.len() == 1
+                && params_from_function.elements[0].algebraic_type == table_row_type_ref.into()
```
Centril, line 996:
```suggestion
            if let [single] = &params_from_function.elements
                && single.algebraic_type == table_row_type_ref.into()
```

crates/schema/src/def/validate/v9.rs
```diff
@@ -933,7 +1027,24 @@ fn check_scheduled_reducers_exist(
         .collect_all_errors()
 }
 
-fn proccess_misc_exports(
+fn check_function_names_are_unique(
+    reducers: &IndexMap<Identifier, ReducerDef>,
+    procedures: &IndexMap<Identifier, ProcedureDef>,
+) -> Result<()> {
+    let names = reducers.keys().collect::<HashSet<_>>();
```
Centril: Btw, do we check anywhere that the set of reducer names are internally unique? Same for tables? If not, this is probably a good place to do it for both reducers and procedures, and then you can use `.intersection(_)`

crates/schema/src/def/validate/v9.rs
```diff
@@ -933,7 +1027,24 @@ fn check_scheduled_reducers_exist(
         .collect_all_errors()
 }
 
-fn proccess_misc_exports(
+fn check_function_names_are_unique(
```
Centril, line 1036: Is this tested anywhere btw?

crates/core/src/host/mod.rs
gefjon: I originally did not make this change because I was afraid it would inflate the diff. I will do it now.

crates/schema/src/def/validate/v9.rs
gefjon: Locally, this gives me:

```
error: let chains are only allowed in Rust 2024 or later
   --> crates/schema/src/def/validate/v9.rs:992:16
    |
992 |             if let [single] = &params_from_function.elements
    |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0529]: expected an array or slice, found `Box<[ProductTypeElement]>`
   --> crates/schema/src/def/validate/v9.rs:992:20
    |
992 |             if let [single] = &params_from_function.elements
    |                    ^^^^^^^^ pattern cannot match with input type `Box<[ProductTypeElement]>`

For more information about this error, try `rustc --explain E0529`.
error: could not compile `spacetimedb-schema` (lib) due to 2 previous errors
```

crates/schema/src/def/validate/v9.rs
gefjon: I sorta assumed we did, but apparently not. I'll add an appropriate check.

crates/schema/src/def/validate/v9.rs
Centril: Try changing the edition? Odds are we don't need to make any changes due to it.

crates/schema/src/def/validate/v9.rs
gefjon: It is now! See `duplicate_reducer_names`, `duplicate_procedure_names` and `duplicate_procedure_and_reducer_name`.

crates/schema/src/def.rs
gefjon: Created #3410

crates/schema/src/def/validate/v9.rs
gefjon: I am unconcerned about allocating here.

crates/schema/src/def/validate/v9.rs
gefjon: I am absolutely not going to change our Rust edition in this PR. If we want to make that change, it should be its own separate PR without any other changes, and should go through proper review.

joshua-spacetime (approved):

PR #3393, opened 2025-10-09
Fix python example to not advise to use ~/token.py

mamcx:
# Description of Changes

Python is funny, if a file `token.py` is created and another script run on the same dir, it will cause python to block:

```python
python3-3.13.7/lib/python3.13/tokenize.py", line 35, in <module>
    from token import *
  File "/Users/mamcx/token.py", line 3, in <module>
    text = sys.stdin.read()
```

By coincidence the docs on `pg wire` use this name. Changed to one that don't cause the trouble.

1

# Testing

- [x] Created another script and run it, see it blocks because this...

### DIFF
--- a/docs/docs/sql/pg-wire.md
+++ b/docs/docs/sql/pg-wire.md
@@ -70,20 +70,16 @@ spacetime login show --token
 
 To export the token to `PGPASSWORD`:
 
-```python
-# token.py
-import sys, re
-
-text = sys.stdin.read()
-match = re.search(r"Your auth token \(don't share this!\) is\s+(\S+)", text)
-if not match:
-    sys.exit("No token found")
-
-print(f"export PGPASSWORD={match.group(1)}")
-```
+*For Bash*:
 
 ```bash
-eval "$(spacetime login show --token | python3 ~/token.py)"
+export PGPASSWORD="$(spacetime login show --token | sed -n 's/^Your auth token.*is //p')"
+```
+
+*For PowerShell*:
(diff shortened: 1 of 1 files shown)


docs/docs/sql/pg-wire.md
```diff
@@ -71,7 +71,7 @@ spacetime login show --token
 To export the token to `PGPASSWORD`:
 
 ```python
-# token.py
+# stdb_token.py
```
jdetter: this is weird that we're making the user create a file called `stdb_token.py` and then they have to execute a shell command which then uses this script. Could we not just do this entire thing in the python script here?

docs/docs/sql/pg-wire.md
```diff
@@ -83,7 +83,7 @@ print(f"export PGPASSWORD={match.group(1)}")
 ```
 
 ```bash
-eval "$(spacetime login show --token | python3 ~/token.py)"
+eval "$(spacetime login show --token | python3 ~/stdb_token.py)"
```
jdetter: If the script needs to be in the home directory, that should be stated above otherwise users might create the script at some other location and then this python invocation won't be able to find the file. I think instead we should just do the `spacetime login show --token` part in python.

jdetter (changes requested):
Just a few things here, let me know if you have questions 🙂

jdetter (approved):
Excellent, this is much much better. Thank you @mamcx !

PR #3401, opened 2025-10-13
V8: Let a worker thread own the isolate

Centril:
# Description of Changes

1. Moves work of `JsInstance` to a worker thread to avoid re-making the isolate all the time.
2. Does some minor refactoring.
3. Deals with `JsInstanceEnv` being potentially unset.

Fixes https://github.com/clockworklabs/SpacetimeDB/issues/3460.

None

3

# Testing

Should be covered by exiting tests.


crates/core/src/host/v8/mod.rs
```diff
@@ -305,77 +246,222 @@ impl JsInstance {
...
+                    // Reply to `JsInstance::call_reducer`.
+                    if let Err(e) = call_reducer_response_tx.send(res) {
+                        // This should never happen as `JsInstance::call_reducer` immediately
+                        // does `.recv` on the other end of the channel.
+                        unreachable!("should have receiver for `call_reducer` response, {e}");
+                    }
+                }
+                // Sent by `<JsInstance as Drop>::drop`.
+                // End the worker loop and cleanup the scope, isolate, and channels.
+                JsWorkerRequest::End => return,
```
coolreader18: Wouldn't dropping the `SyncSender` do the same thing?

crates/core/src/host/v8/syscall.rs
```diff
@@ -169,16 +182,29 @@ type SysCallResult<T> = Result<T, SysCallError>;
...
 /// Handles [`SysCallError`] if it occurs by throwing exceptions into JS.
-fn with_sys_result<'scope, O: Serialize>(
+fn with_sys_result_ret<'scope, O: Serialize>(
     abi_call: AbiCall,
     scope: &mut PinScope<'scope, '_>,
     args: FunctionCallbackArguments<'scope>,
     run: impl FnOnce(&mut PinScope<'scope, '_>, FunctionCallbackArguments<'scope>) -> SysCallResult<O>,
 ) -> FnRet<'scope> {
     match with_span(abi_call, scope, args, run) {
         Ok(ret) => serialize_to_js(scope, &ret),
```
coolreader18, line 195: This is what `ToValue` was for originally - it `()` maps to `undefined`, and integers and stuff to numbers and bigints.

crates/core/src/host/v8/mod.rs
Centril: It would yes, nicely spotted.

PR #3343, opened 2025-10-03
Docusaurus migration

JulienLavocat:
## Description of Changes

- Migrate the documentation site to **Docusaurus**.
- Add a **GitHub Action workflow** that runs `pnpm build` on pull requests targeting `master`.  
  This command checks for broken links and images and ensures the site builds correctly.
- Add a **GitHub Action workflow** to publish the docs automatically when a merge occurs on the `docs/release` branch. 

**Important**: The workflows haven't been tested yet due to the fact that they need to be merged first before being able to run. It's a fairly basic workflow I don't expect this to block anyone but anyways I will make another PR with the complete changes when I can actually run it.

---

## API and ABI Breaking Changes

No breaking changes for **SpacetimeDB** users.  
Documentation contributors should experience minimal disruption — content is still written in Markdown.

---

## Expected Complexity Level and Risk

**Complexity:** 3/5

**Reasons:**
1. Several adjustments were made to fit Docusaurus conventions:
   - Replaced our custom `:::server-rust` blocks with [Docusaurus `<Tabs/>` components](https://docusaurus.io/docs/markdown-features/tabs).
   - Converted “Note” callouts using [admonitions](https://docusaurus.io/docs/markdown-features/admonitions) instead of Markdown blockquotes.
   - Updated all images (including Unity tutorial ones, which were previously commented out).
   - Moved images from the DO bucket into the repository at `docs/static/images`.

2. Customized Docusaurus CSS and one component to apply our theme and color palette.

---

## Testing

Tested locally.

---

## Review Notes

This PR involves many file changes — a full code review is likely not productive.  
Instead, please focus on verifying that the **documentation content** was migrated correctly:

- Review sections you’re familiar with to confirm accuracy.
- Ensure formatting, links, and images render as expected.

You can follow the updated **README** for instructions on running Docusaurus locally.

PR #3403, opened 2025-10-14
Fix CLI reference generation

bfops:
# Description of Changes

Add back the instructions for regenerating CLI docs, which were removed in https://github.com/clockworklabs/SpacetimeDB/pull/3343. I also made a script for it.

This also fixes the CI checking this file, which was silently broken in the same PR.

I have **not** verified that this works in Git Bash in Windows.

None

1

# Testing
- [x] CI passes
- [x] CI fails if I change the CLI reference
- [x] CLI reference looks visually reasonable on a local `pnpm dev`


crates/cli/src/subcommands/init.rs
```diff
@@ -123,7 +123,7 @@ pub fn cli() -> clap::Command {
                 .long("project-path")
                 .value_name("PATH")
                 .value_parser(clap::value_parser!(PathBuf))
-                .help("Directory where the project will be created (defaults to ./<PROJECT_NAME>)"),
+                .help("Directory where the project will be created (defaults to `./<PROJECT_NAME>`)"),
```
bfops, line 126: this was required because otherwise `<PROJECT_NAME>` was rendered "raw" in the markdown, and was then interpreted as a (broken) HTML tag by docusaurus.

docs/scripts/generate-cli-docs.mjs
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env node
```
bfops, line 1: I would also be happy to do this as a bash script, but I'm worried about it breaking on windows. It would be much more concise, though.

tools/ci/src/main.rs
```diff
@@ -220,16 +220,14 @@ cargo run {github_token_auth_flag}{target} -p spacetimedb-update -- self-install
...
 
-            bash!("pnpm install --recursive")?;
-            bash!("cargo run --features markdown-docs -p spacetimedb-cli > docs/docs/cli-reference.md")?;
-            bash!("pnpm format")?;
+            bash!("cd docs && pnpm generate-cli-docs")?;
             bash!("git status")?;
             bash!(
                 r#"
-if git diff --exit-code HEAD; then
+if [ -z "$(git status --porcelain)" ]; then
```
bfops, line 226: this should also error on new files, which was how this silently gotbroken before

docs/scripts/generate-cli-docs.mjs
jdetter, line 1: Do you want to `chmod +x` on this file? I personally didn't know the `.mjs` extension is something you can execute via node and I would have to `head generate-cli-docs.mjs` to check the interpreter if the file isn't executable.

bfops: Good idea! Done.

PR #3404, opened 2025-10-14
[TS] Register describe_module/call_reducer through the sys module

coolreader18:
# Description of Changes

This exports the following API from `spacetime:sys@1.0`:

```ts
export type ModuleHooks = {
  __describe_module__(): Uint8Array;

  __call_reducer__(
    reducerId: u32,
    sender: u256,
    connId: u128,
    timestamp: bigint,
    argsBuf: Uint8Array
  ): { tag: 'ok' } | { tag: 'err'; value: string };
};

export function register_hooks(hooks: ModuleHooks);
```

This is then called by the `spacetimedb` package. This means that users no longer need to export `__call_reducer__` and `__describe_module__` from their own code, and differently from how it was before it doesn't pollute the global namespace. It also means that the signatures for `__call_reducer__` can be changed when we change the ABI version - the bindings would then do `import { register_hooks } from "spacetime:sys@2.0";` and that version of the function would expect a different signature..

2 - it's a change, but not all that different from the existing way we've been doing it.

# Testing, and also , and also 

- [x] Automated testing
- [ ]

### DIFF
--- a/crates/core/src/host/v8/mod.rs
+++ b/crates/core/src/host/v8/mod.rs
@@ -587,8 +587,7 @@ fn call_free_fun<'scope>(
 
 /// Calls the `__call_reducer__` hook, if it's been registered.
 fn call_call_reducer_from_op<'scope>(scope: &mut PinScope<'scope, '_>, op: ReducerOp<'_>) -> ExcResult<ReducerResult> {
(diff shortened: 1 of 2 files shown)


cloutiertyler:
@Centril also approves

PR #3405, opened 2025-10-14
Also capture a snapshot every new commitlog segment

gefjon:
# Description of Changes

We've run into a problem on Maincloud caused by a database that was writing a relatively small number of very large transactions. This was accruing many commitlog segments consuming hundreds of gigabytes of disk, but had not ever taken a snapshot, or compressed or archived any data, as the database had not progressed past one million transactions.

With this PR, we take a snapshot every time the commitlog segment rotates. We still also snapshot every million transactions.

One BitCraft database we looked at had 2.5 million transactions per commitlog segment, meaning that this change will not meaningfully affect the frequency of snapshots. The offending Maincloud database, however, had only 50 transactions per segment!

N/a

3: Hastily made changes to finnicky code across several crates.

# Testing

I am unsure how to test these changes.

- [ ] 
- [ ]

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6142,7 +6142,6 @@ name = "spacetimedb-durability"
 version = "1.5.0"
 dependencies = [
  "anyhow",
- "futures",
  "itertools 0.12.1",
  "log",
  "spacetimedb-commitlog",


crates/commitlog/src/lib.rs
```diff
@@ -390,19 +391,35 @@ impl<T: Encode> Commitlog<T> {
...
     ///
     /// If the log needs to be flushed, but an I/O error occurs, ownership of
     /// `txdata` is returned back to the caller alongside the [`io::Error`].
     ///
     /// The value can then be used to retry appending.
-    pub fn append_maybe_flush(&self, txdata: T) -> Result<(), error::Append<T>> {
+    pub fn append_maybe_flush(
+        &self,
+        txdata: T,
+        on_new_segment: Option<&mpsc::UnboundedSender<()>>,
```
kim: I don't love that this is tied to this specific method, because a) we don't always use this one to append txs and b) we don't always use the concrete `Commitlog`, but rather `commitlog::Generic`.

Why not make the sender a field of the `Repo` impl and fire an event when `create_segment` is called?

crates/core/src/db/snapshot.rs
```diff
@@ -176,21 +184,35 @@ impl SnapshotWorkerActor {
     /// processed.
     async fn run(mut self) {
         let mut database_state: Option<WeakDatabaseState> = None;
-        while let Some(req) = self.snapshot_requests.next().await {
-            match req {
-                Request::TakeSnapshot => {
-                    let res = self
+        loop {
+            tokio::select! {
+                // Biased because we want to update the state before we take snapshots if both happen at the same time:
```
kim: Fair.

crates/core/src/db/snapshot.rs
```diff
@@ -116,10 +120,18 @@ impl SnapshotWorker {
...
-            .unbounded_send(Request::TakeSnapshot)
+            .unbounded_send(())
             .expect("snapshot worker panicked");
     }
 
+    /// Get a handle to the channel used by [`Self::request_snapshot`].
+    ///
+    /// We use this rather than [`Self::request_snapshot`] via the persistence/commitlog
+    /// to simplify crate dependencies.
+    pub fn request_snapshot_sender(&self) -> mpsc::UnboundedSender<()> {
```
kim: I would prefer if the queuing strategy (and channel types) would remain an implementation detail of `SnapshotWorker`. 

Could we not use an `Arc<dyn Fn(TxOffset) + Send + Sync + 'static>` at the `Repo` and instantiate as `Arc::new(|_offset| { let _ = clone_of_worker.request_snapshot(); })`?

gefjon: I tried this first and was unable to figure it out. `Arc<dyn Fn()>` does not implement `Fn()`, nor does `Arc<T> where T: Fn()`, and the `Deref` impl appears insufficient to invoke an arc'd closure.

crates/commitlog/src/lib.rs
gefjon: Reasonable. I will attempt this.

crates/core/src/db/snapshot.rs
gefjon: Disregard, I was doing something wrong. Will push a commit momentarily.

PR #3408, opened 2025-10-15
Add `modules/benchmarks-ts`

Centril:
# Description of Changes

Adds `modules/benchmarks-ts`.

None

2?

# Testing

This is a test.


crates/schema/tests/ensure_same_schema.rs
coolreader18: you can remove this now

coolreader18 (approved):
LGTM!

PR #3409, opened 2025-10-15
add initial cargo ci

Kasama:
# Description of Changes

This changes the ci runs to execute `cargo ci` instead of running commands directly from the github workflow.

The goal here is to unify the commands under `cargo ci` so that it's easier and more intuitive to run locally

There are no API/ABI changes.

Complexity: 1

It is not a complex change as it is mostly localized to the ci runs and is easily reversible if something goes wrong. The biggest risk here is to have future CI runs break, which can be remediated by reverting these changes.

# Testing

- [x] run `cargo ci` and its subcommands locally
- [x] run the github workflow against this branch to check if the CI jobs are working properly.


bfops:
The tests are failing. I think several of them just need a `Cargo.lock` update, but at least one seems to be failing for an unknown subcommand.

.github/workflows/ci.yml
```diff
@@ -290,6 +228,15 @@ jobs:
...
           ref: ${{ env.GIT_REF }}
       - uses: dsherret/rust-toolchain-file@v1
+      - name: Install unreal engine test dependencies
+        working-directory: sdks/unreal
+        env:
+          UE_ROOT_PATH: /home/ue4/UnrealEngine
+        run: |
+
+          apt-get update
+          apt-get install -y acl curl ca-certificates
```
bfops, line 371: this part is already done below?

tools/ci/src/main.rs
```diff
@@ -0,0 +1,172 @@
...
+    }
+    Ok(())
+}
+
+fn main() -> Result<()> {
+    let cli = Cli::parse();
+
+    match cli.cmd {
+        Some(CiCmd::Test) => {
+            run!("sudo mkdir -p /stdb && sudo chmod 777 /stdb")?;
```
bfops: I think this should be part of the CI still? I'm pretty sure we don't want to do this if running it on local machines.

tools/ci/src/main.rs
```diff
@@ -0,0 +1,251 @@
...
+        Some(CiCmd::WasmBindings) => {
+            run!("cargo test -p spacetimedb-codegen")?;
+            run!("cargo update")?;
+            run!("cargo run -p spacetimedb-cli -- build --project-path modules/module-test")?;
+        }
+
+        Some(CiCmd::Smoketests { args }) => {
+            // Note: clear_database and replication only work in private
+            run!(&format!(
+                "python -m smoketests {} -x clear_database replication",
```
bfops: I kind of think that the tests to skip should be a parameter, since we may not always want to skip the same things

tools/ci/src/main.rs
```diff
@@ -0,0 +1,251 @@
...
+            long,
+            long_help = "Target triple to build for, by default the current target. Used by github workflows to check the update flow on multiple platforms."
+        )]
+        target: Option<String>,
+        #[arg(
+            long,
+            default_value = "true",
+            long_help = "Whether to enable github token authentication feature when building the update binary. By default this is enabled."
+        )]
+        github_token_auth: bool,
```
bfops, line 76: nit: I kind of think this should default to `false` so that people running this outside of CI don't get errors about the env var not being defined.

tools/ci/src/main.rs
```diff
@@ -0,0 +1,251 @@
...
+"${{ROOT_DIR}}"/spacetime --root-dir="${{ROOT_DIR}}" help
+        "#
+            ))?;
+        }
+
+        Some(CiCmd::UnrealTests) => {
+            run!("for p in \"$GITHUB_WORKSPACE\" \"${RUNNER_TEMP:-/__t}\" \"${RUNNER_TOOL_CACHE:-/__t}\"; do [ -d \"$p\" ] && setfacl -R -m u:ue4:rwX -m d:u:ue4:rwX \"$p\" || true; done")?;
+
+            run!("export CARGO_HOME=\"${RUNNER_TOOL_CACHE:-/__t}/cargo\"")?;
+            run!("export RUSTUP_HOME=\"${RUNNER_TOOL_CACHE:-/__t}/rustup\"")?;
```
bfops: FYI I think a lot of this doesn't apply to users running these tests locally

tools/ci/src/main.rs
```diff
@@ -0,0 +1,251 @@
...
+
+            run!("chmod a+rx \"$UE_ROOT_PATH\" \"$UE_ROOT_PATH/Engine\" \"$UE_ROOT_PATH/Engine/Build\" \"$UE_ROOT_PATH/Engine/Build/BatchFiles/Linux\" || true")?;
+            run!("chmod a+rx \"$UE_ROOT_PATH/Engine/Build/BatchFiles/Linux/Build.sh\" || true")?;
+
+            run!("sudo -E -H -u ue4 env HOME=/home/ue4 CARGO_HOME=\"$CARGO_HOME\" RUSTUP_HOME=\"$RUSTUP_HOME\" PATH=\"$CARGO_HOME/bin:$PATH\" bash -lc 'set -euxo pipefail; if ! command -v cargo >/dev/null 2>&1; then curl -sSf https://sh.rustup.rs | sh -s -- -y; fi; rustup show >/dev/null; git config --global --add safe.directory \"$GITHUB_WORKSPACE\" || true; cd \"$GITHUB_WORKSPACE/sdks/unreal\"; cargo --version; cargo test'")?;
+        }
+
+        Some(CiCmd::CliDocs) => {
+            run!("pnpm install --recursive")?;
+            run!("cargo run --features markdown-docs -p spacetimedb-cli > docs/docs/cli-reference.md")?;
```
bfops: note that this assumes we're running from the root directory

.github/workflows/ci.yml
bfops: did we mean to replace parts of the below with `cargo ci`?

.github/workflows/ci.yml
```diff
@@ -350,6 +284,33 @@ jobs:
...
+        with:
+          ref: ${{ env.GIT_REF }}
+
+      - uses: dsherret/rust-toolchain-file@v1
+
+      - name: Check for docs change
+        run: cargo ci self-docs --check
+
   cli_docs:
     name: Check CLI docs
```
bfops, line 439: I assume we want to replace this one with a `cargo ci` invocation as well?

tools/ci/build.rs
```diff
@@ -0,0 +1,4 @@
+#[allow(clippy::disallowed_macros)]
+fn main() {
+    println!("cargo:rustc-env=TARGET={}", std::env::var("TARGET").unwrap_or_default());
```
bfops: what's this part for?

tools/ci/src/main.rs
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    Ok(())
+}
+
+fn run_command(cmdline: &str, additional_env: &[(&str, &str)]) -> Result<()> {
+    let mut env = env::vars().collect::<HashMap<_, _>>();
+    env.extend(additional_env.iter().map(|(k, v)| (k.to_string(), v.to_string())));
+    log::debug!("$ {cmdline}");
+    let status = cmd!("bash", "-lc", cmdline).full_env(env).run()?;
```
bfops, line 127: FYI I'm not 100% sure how this will work on windows

tools/ci/src/ci_docs.rs
```diff
@@ -0,0 +1,68 @@
+use clap::{Command, CommandFactory};
+
+use crate::Cli;
+
+pub fn generate_cli_docs() -> String {
```
bfops, line 6: nit: any chance we could use `clap_markdown` like we do for the SpacetimeDB CLI docs?

tools/ci/build.rs
Kasama: this allows for the default parameter for `target` on the update-flow. Mostly for running it locally without needing to specify that flag manually.

https://github.com/clockworklabs/SpacetimeDB/blob/1c0406d2a724a4f98b8bdbf1cd258652d38dce79/tools/ci/src/main.rs#L177

tools/ci/src/ci_docs.rs
Kasama: indeed. I didn't notice that we had that

tools/ci/src/main.rs
Kasama: I looked locally and it looks like this directory isn't even used by the test. I think we can probably remove it

tools/ci/build.rs
bfops: ah gotcha. in that case, nit: we could either use the variables that are provided already:
```
const ARCH: &str = env!("CARGO_CFG_TARGET_ARCH");
const OS: &str = env!("CARGO_CFG_TARGET_OS");
const ENV: &str = env!("CARGO_CFG_TARGET_ENV");
```
or we could only pass `--target {target}` to `cargo build` _if_ it's provided by the user.

I kind of prefer that last option. It seems intuitive still.

bfops: (honestly I'm not 100% sure that we even need the target param.. I think the default on each platform might be the one we want anyway.. but we should keep it for now just for parity with the original CI).

bfops:
FYI I pushed a merge commit so the CI would run again

tools/ci/src/ci_docs.rs
bfops: since this is already working, we could add a TODO and just leave it for a future PR.

tools/ci/src/main.rs
Kasama: I added the args as default if nothing is passed, but allowed the caller to pass any args

tools/ci/src/main.rs
Kasama: it does. Do you think this it a problem to assume that? If so, would it be a good idea to try to figure out where in the dir tree the command was ran (maybe looking upwards for the root `Cargo.toml` or something?) to adjust accordingly?

tools/ci/src/main.rs
Kasama: I'm not sure how to work this out, actually. Because all other commands are also unix-like. Do you have an idea on how to make it work on windows too?

.github/workflows/ci.yml
```diff
@@ -76,8 +78,8 @@ jobs:
       - name: Install python deps
         run: python -m pip install psycopg2-binary xmltodict
       - name: Run smoketests
-        # Note: clear_database and replication only work in private
-        run: python -m smoketests ${{ matrix.smoketest_args }} -x clear_database replication
+        run: cargo ci smoketests -- ${{ matrix.smoketest_args }}
```
bfops: little nit: I kinda preferred when the skipped tests were provided here, since they're going to be skipped on every platform. Not super important, but also not super important to change (although we could consider renaming `smoketests_args` to `extra_smoketest_args` to be clearer).

If you prefer it this way, that's fine with me.

tools/ci/src/main.rs
```diff
@@ -162,19 +161,20 @@ fn main() -> Result<()> {
             run!("cargo run -p spacetimedb-cli -- build --project-path modules/module-test")?;
         }
 
-        Some(CiCmd::Smoketests { args }) => {
+        Some(CiCmd::Smoketests { mut args }) => {
+            let default_args = ["-x", "clear_database", "replication"];
```
bfops: nit: I worry about tests being omitted by default. I think the expectation was that there's some easy/intuitive way to run _everything_.

tools/ci/src/main.rs
bfops: I might just add a check similar to what `tools/upgrade-version` does:
```
tools/upgrade-version/src/main.rs-    let current_dir = env::current_dir().expect("No current directory!");
tools/upgrade-version/src/main.rs-    let dir_name = current_dir.file_name().expect("No current directory!");
tools/upgrade-version/src/main.rs-    if dir_name != "SpacetimeDB" && dir_name != "public" {
tools/upgrade-version/src/main.rs:        anyhow::bail!("You must execute this binary from inside of the SpacetimeDB directory, or use --spacetime-path");
```
(alternatively we could use `git rev-parse --show-toplevel` and then just go to that directory)

tools/ci/src/main.rs
bfops, line 193: nit: if we use `.map().unwrap_or_default()` above, we could use something similar here:
```suggestion
            let github_token_auth_flag = github_token_auth.then_some("--features github-token-auth ").unwrap_or_default();
```
I don't feel strongly one way or another.

bfops: (or we could have an `extra_args` string that we append to if `target` is defined or `github_token_auth` is true)

tools/ci/build.rs
bfops: now that `--target` is optional, is it reasonable to remove this?

tools/ci/src/main.rs
bfops: I'd say we ask @jdetter to try running some of `cargo ci` on his windows machine. If it "just works" then great. If it doesn't due to `bash` missing, then I think we might need to make a separate ticket for making all of our CI properly runnable on windows. I suspect it might work properly in the GitHub workflow as-is though, because I believe that will have bash installed.

.github/workflows/ci.yml
Kasama: I think it makes sense to have the default skipped tests defined in either place, but I think it's ok to keep it in the same spot they were before. Changed

tools/ci/src/main.rs
Kasama: hmm, I understand, althought I thought that the default behaviour should be to do whatever the github action was doing. That's why I added these default exclusions, but it does make it harder (impossible?) to run without any exclusions

tools/ci/src/main.rs
Kasama: I do think it's a bit easier to read with `then_some` as well, but clippy didn't like it: https://rust-lang.github.io/rust-clippy/master/index.html#obfuscated_if_else

tools/ci/src/main.rs
Kasama: anyway I removed the default args so that the caller has more control over what's actually being run there

tools/ci/src/main.rs
bfops: well alright then 🤷

tools/ci/src/main.rs
```diff
@@ -0,0 +1,264 @@
...
+            run!(&format!(
+                r#"
+ROOT_DIR="$(mktemp -d)"
+cargo run {github_token_auth_flag}--target {target} -p spacetimedb-update -- self-install --root-dir="${{ROOT_DIR}}" --yes
+"${{ROOT_DIR}}"/spacetime --root-dir="${{ROOT_DIR}}" help
+        "#
+            ))?;
+        }
+
+        Some(CiCmd::UnrealTests) => {
```
bfops: since we're moving this to a separate PR/branch, I would probably remove this here. My usual flow to do this is:

```bash
$ git commit -m 'Revert the things'
$ git checkout -b branch-to-add-the-things
$ git checkout HEAD~1 -- .
$ git commit -am 'Do the things'
$ git push -u
$ gh pr create --web
```

.github/workflows/ci.yml
```diff
@@ -127,30 +129,88 @@ jobs:
...
+          apt-get update
+          apt-get install -y acl curl ca-certificates
 
-      - name: Check that the test outputs are up-to-date
-        run: bash tools/check-diff.sh
+          REPO="$GITHUB_WORKSPACE"
+          # Let ue4 read/write the workspace & tool caches without changing ownership
+          for p in "$REPO" "${RUNNER_TEMP:-/__t}" "${RUNNER_TOOL_CACHE:-/__t}"; do
+            [ -d "$p" ] && setfacl -R -m u:ue4:rwX -m d:u:ue4:rwX "$p" || true
+          done
```
bfops: I think it's fine to have it the same way it was before - one job invocation that does --skip-unreal, and one job that sets up all the unreal stuff and only tests the unreal stuff.

(I think @cloutiertyler's goal (I could be wrong) was that the default invocation of `cargo ci test` or whatever would do everything _by default_, but I don't think the CI jobs need to do everything in one).

.github/workflows/ci.yml
```diff
@@ -23,9 +23,10 @@ jobs:
     strategy:
       matrix:
         include:
-          - { runner: spacetimedb-runner, smoketest_args: --docker }
-          - { runner: windows-latest, smoketest_args: --no-build-cli }
-        runner: [ spacetimedb-runner, windows-latest ]
```
bfops: This line was required in order to make the titles of the tests not include the args. This is why the required checks aren't running.

Cargo.toml
```diff
@@ -1,57 +1,58 @@
...
+	"modules/sdk-test-procedure",
+	"sdks/rust/tests/test-client",
+	"sdks/rust/tests/test-counter",
+	"sdks/rust/tests/connect_disconnect_client",
+	"tools/ci",
+	"sdks/rust/tests/procedure-client",
+	"tools/upgrade-version",
+	"tools/license-check",
+	"crates/bindings-typescript/test-app/server",
+	"crates/bindings-typescript/test-react-router-app/server",
```
bfops, line 60: is there a reason we re-indented all of this? 🤔

.github/workflows/ci.yml
```diff
@@ -343,19 +369,9 @@ jobs:
         run: sudo apt install -y libssl-dev
 
       - name: Build spacetimedb-update
```
bfops, line 299:
```suggestion
      - name: Test spacetimedb-update
```

.github/workflows/ci.yml
bfops: I have done this

.github/workflows/ci.yml
```diff
@@ -112,7 +112,7 @@ jobs:
         run: python -m pip install -r smoketests/requirements.txt
       - name: Run smoketests
         # Note: clear_database and replication only work in private
-        run: python -m smoketests ${{ matrix.smoketest_args }} -x clear_database replication teams
+        run: cargo ci smoketests -- ${{ matrix.smoketest_args }} -x clear_database replication teams
```
bfops, line 120: this seems to be failing on windows for some of the C# projects.. maybe because we're explicitly running in bash now? It seems to be failing to find some paths. Unsure if that's the real issue though, I just can't see why else it would be upset.

tools/ci/src/main.rs
```diff
@@ -0,0 +1,257 @@
...
+
+        Some(CiCmd::WasmBindings) => {
+            bash!("cargo test -p spacetimedb-codegen")?;
+            bash!("cargo update")?;
+            bash!("cargo run -p spacetimedb-cli -- build --project-path modules/module-test")?;
+        }
+
+        Some(CiCmd::Smoketests { args }) => {
+            // On some systems, there is no `python`, but there is `python3`.
+            let py3_available = cmd!("bash", "-lc", "command -v python3 >/dev/null 2>&1")
```
bfops: I added this.. but I realize I'm not sure whether this works on windows.. Also I probably should have reused the `bash!` macro..

.github/workflows/ci.yml
Kasama: it now runs the old command for windows targets

tools/ci/src/main.rs
Kasama: I changed it to use `bash!` instead. We will probably need to revisit how that works for windows, but at least it will be centralized.

.github/workflows/ci.yml
```diff
@@ -111,9 +111,14 @@ jobs:
         if: runner.os == 'Windows'
       - name: Install python deps
         run: python -m pip install -r smoketests/requirements.txt
-      - name: Run smoketests
+      - name: Run smoketests windows
         # Note: clear_database and replication only work in private
         run: python -m smoketests ${{ matrix.smoketest_args }} -x clear_database replication teams
+        if: runner.os == 'Windows'
+      - name: Run smoketests
```
bfops:
```suggestion
      - name: Run smoketests (Linux)
```

.github/workflows/ci.yml
```diff
@@ -111,9 +111,14 @@ jobs:
         if: runner.os == 'Windows'
       - name: Install python deps
         run: python -m pip install -r smoketests/requirements.txt
-      - name: Run smoketests
+      - name: Run smoketests windows
```
bfops:
```suggestion
      - name: Run smoketests (windows)
```

Cargo.toml
```diff
@@ -1,62 +1,63 @@
 [workspace]
 members = [
-  "crates/auth",
```
bfops, line 3: do you know why these whitespace changes keep getting introduced?

tools/ci/src/main.rs
```diff
@@ -0,0 +1,254 @@
...
+            bash!(
+                "cd crates/bindings && cargo doc",
+                // Make `cargo doc` exit with error on warnings, most notably broken links
+                &[("RUSTDOCFLAGS", "--deny warnings")]
+            )?;
+        }
+
+        Some(CiCmd::WasmBindings) => {
+            bash!("cargo test -p spacetimedb-codegen")?;
+            bash!("cargo update")?;
```
bfops, line 173:
```suggestion
            // Make sure the `Cargo.lock` file reflects the latest available versions.
            // This is what users would end up with on a fresh module, so we want to
            // catch any compile errors arising from a different transitive closure
            // of dependencies than what is in the workspace lock file.
            //
            // For context see also: https://github.com/clockworklabs/SpacetimeDB/pull/2714
            bash!("cargo update")?;
```

bfops: restoring the comment that was in the workflow file

.github/workflows/ci.yml
```diff
@@ -609,7 +584,7 @@ jobs:
           enable_pr_comment: ${{ github.event_name == 'pull_request' }}
           target_path: sdks/csharp
         env:
-          GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+          GITHUB_TOKEN: '${{ secrets.GITHUB_TOKEN }}'
```
bfops, line 587: ?

bfops (approved):
LGTM, minus a few comments

.github/workflows/ci.yml
Kasama: probably forgot to disable the auto-formatter. Will undo

Cargo.toml
Kasama: not sure, but I'd guess it's some auto-formatter. But it doesn't seem to have been mine, because if I run it, it tries to turn the tabs into spaces :thinking:

PR #3413, opened 2025-10-15
Bump versions to 1.6.0

bfops:
# Description of Changes

Bumping versions to 1.6.0 in preparation for upcoming release.

None

1
# Testing

- [x] Existing CI only


crates/cli/src/subcommands/project/csharp/StdbModule._csproj
cloutiertyler, line 1: I don't see the typescript package._json equivalent for this file, is that expected?

bfops: afaik that file doesn't exist in master, so it's not in this PR.

cloutiertyler: Okay

cloutiertyler (approved):

PR #3414, opened 2025-10-16
Add bindings for csharp modules to use JWT claims

jsdt:
# Description of Changes

This exposes JWT claims for csharp modules, similar to how they are exposed to rust modules in https://github.com/clockworklabs/SpacetimeDB/pull/3288.

This adds the new types `AuthCtx` and `JwtClaims`, and adds an `AuthCtx` to the `ReducerContext`.

`AuthCtx` represents the credentials associated with the request, and `JwtClaims` represents a jwt token.

One difference from the rust version is that I didn't create helpers to build an `AuthCtx` from a jwt payload. The reason is that we would need to be able to compute the identity from the payload claims, which requires a blake3 hash implementation. The first two c# libraries I found had issues at runtime ([Blake3](https://www.nuget.org/packages/Blake3) is wrapping a rust implementation, and [HashifyNet](https://github.com/Deskasoft/HashifyNET/tree/main/HashifyNet/Algorithms/Blake3) seems to be broken by our trimming because it uses reflection heavily). I can look into taking the implementation from `HashifyNet`, since it is MIT licensed, but I don't think we need to block merging on that.

This adds the new types `AuthCtx` and `JwtClaims`, and adds an `AuthCtx` to the `ReducerContext`.

This also adds a csharp wrapper for the get_jwt ABI function added in https://github.com/clockworklabs/SpacetimeDB/pull/3288.

2.

# Testing

This has a very minimal unit test of JwtClaims.

I manually tested using this locally with the csharp quickstart, and I was able to print jwt tokens inside the module.

### DIFF
--- a/modules/quickstart-chat/src/lib.rs
+++ b/modules/quickstart-chat/src/lib.rs
@@ -67,10 +67,6 @@ pub fn init(_ctx: &ReducerContext) {}
 
 #[spacetimedb::reducer(client_connected)]
 pub fn identity_connected(ctx: &ReducerContext) {
-    log::info!(
-        "Client connected with jwt payload {:?}",
-        ctx.sender_auth().jwt().unwrap().raw_payload()
-    );
     if let Some(user) = ctx.db.user().identity().find(ctx.sender) {
         // If this is a returning user, i.e. we already have a `User` with this `Identity`,
         // set `online: true`, but leave `name` and `identity` unchanged.


crates/bindings-csharp/Runtime/JwtClaims.cs
```diff
@@ -0,0 +1,89 @@
...
+using System.Text.Json;
+
+public sealed class JwtClaims
+{
+    private readonly string _payload;
+    private readonly Lazy<JsonDocument> _parsed;
+    private readonly Lazy<List<string>> _audience;
+
+    private readonly Identity _identity;
+    public Identity Identity => _identity;
```
rekhoff:
```suggestion
    public Identity Identity { get; }
```
Any reason we don't wanna use an auto-property getter here?

modules/quickstart-chat/src/lib.rs
```diff
@@ -67,6 +67,10 @@ pub fn init(_ctx: &ReducerContext) {}
 
 #[spacetimedb::reducer(client_connected)]
 pub fn identity_connected(ctx: &ReducerContext) {
+    log::info!(
+        "Client connected with jwt payload {:?}",
+        ctx.sender_auth().jwt().unwrap().raw_payload()
+    );
```
gefjon: See previous discussion https://github.com/clockworklabs/SpacetimeDB/pull/3288#discussion_r2440495542

jsdt: Sorry, I just made a messy merge that I'm cleaning up.

PR #3416, opened 2025-10-16
Fix windows linker errors

coolreader18:
# Description of Changes

My guess as to the problem was that there were static libs in the dependency graph that were wanting to statically link to libc functions, but rustc didn't know we wanted to link libc statically - so adding `+crt-static` fixes it.

1

# Testing

- [x] Builds successfully on windows now.
- [ ]

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -10,4 +10,6 @@ bump-versions = "run -p upgrade-version --"
 # I (@bfops) tried a variety of other link options besides switching linkers, but this
 # seems to be the only thing that worked.
 linker = "lld-link"
+# Without this, the linker complains that libc functions are undefined -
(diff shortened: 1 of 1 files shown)


rekhoff (approved):
Tested the following:
`cargo build` - Passed
`cargo run -p spacetimedb-standalone` - Passed
`cargo run -p spacetimedb-cli` - Passed
`cargo run -p spacetimedb-update` - Passed
Currently not all tests pass when `cargo test` is ran, but they seem to fail at `test cli::self_install::tests::ensure_script_help_is_up_to_date` which is part of the `cargo test -p spacetimedb-update` test. I've confirmed this was still failing at 1.4.0's release on both my Windows system, so shouldn't be a blocker here. Just a note for the future.
With the above tests and the CI testing passing, I'm approving this change.

PR #3417, opened 2025-10-16
Add multi-module support for Unreal codgen

JasonAtClockwork:
# Description of Changes

- Added functionality to generate module support for multiple folders in the Unreal project (add <module>.Build.cs, <module>.h, <module>.cpp) using the --module-name
- Add new configuration option for spacetime generate to handle module prefix

Arvikasoft's original work has the --module-name option which tells the codegen which module's source folder we'll apply the generated code into. However it didn't ensure that this was a functional module for Unreal tying it into the .uproject and adding the core build file and module definition (.h/cpp). With this change it will detect and add the files if necessary allowing developers to put their module in a separate folder from their game logic.

With Unreal reflection system it is unable to use C++ namespaces which means with codegen we have to separate two modules by the object names. The unofficial standard is to give a prefix to your module for all your C++ classes/structs to avoid collision, for example `struct FBhCircle`, this is what --module-name will support.

None

2 - Heavy refactor to add the prefix across the scope of codegen

# Testing

- [x] Codegen against previous Blackholio build without using the new feature and targetting the original module to confirm no breaking changes
- [x] Codegen against Blueprint Blackholio with second copy of Blackholio as "Fakeholio" to confirm no conflicts
- [x] Codegen against previous Blueprint Blackholio to extend with the Quickstart Chat sample to connect and receive messages


JasonAtClockwork:
This one might be useful for someone else to use the codegen for Unreal for a second pair of eyes.

rekhoff (approved):
Reviewed code, everything is clear and reasonable.
Published and regenerated an existing Blackholio Unreal project (cSharp server) on this branch and confirmed existing projects loaded fine.
Published and regenerated an existing Blackholio Blueprint project (Rust server) on this branch with a new module prefix, and confirmed that Blueprint nodes are being generated with the new module prefix included. (This broken existing blueprints of course, but that's expected).
Everything look good, marking my approval.

JasonAtClockwork:
Merged the changes into #4497 and closing this PR.

PR #3427, opened 2025-10-17
Add module def types for views

joshua-spacetime:
# Description of Changes

Adds type definitions for `RawViewDefV9` and `ViewDef`. Also validates these module defs and populates the system tables:
- `st_table`
- `st_column`
- `st_view`
- `st_view_param`
- `st_view_column`

None

2

# Testing

Integration tests to follow when views are added to `__describe_module__`

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,2 @@
+# Directory environment using https://direnv.net/ and https://github.com/nix-community/nix-direnv.
+use flake
--- a/.gitignore
+++ b/.gitignore
@@ -220,3 +220,6 @@ new.json
 
 # Test data
 !crates/core/testdata/
+
+# Symlinked output from `nix build`
+result
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -353,3 +353,8 @@ result_large_err = "allow"
 # version of temporal_rs, but I'm not holding my breath.
 timezone_provider = { git = "https://github.com/boa-dev/temporal", tag = "v0.0.11" }
 temporal_rs = { git = "https://github.com/boa-dev/temporal", tag = "v0.0.11" }
+
+[workspace.metadata]
+# This silences a warning in our Nix flake, which otherwise would be upset that our call to `crateNameFromCargoToml`
+# is running against a file that doesn't have a name in it.
+crane.name = "spacetimedb"
--- a/docs/docs/modules/typescript/index.md
+++ b/docs/docs/modules/typescript/index.md
@@ -1,0 +1,711 @@
+# SpacetimeDB TypeScript Module Library
+
+[SpacetimeDB](https://spacetimedb.com/) lets you write server-side applications (called **modules**) that run inside a relational database. Modules define **tables** (your data) and **reducers** (your server endpoints). Clients connect directly to the database to read public data via SQL subscriptions and queries, and they invoke reducers to mutate state.
+
+```text
+    Client Application                          SpacetimeDB
+┌───────────────────────┐                ┌───────────────────────┐
+│                       │                │                       │
+│  ┌─────────────────┐  │    SQL Query   │  ┌─────────────────┐  │
+│  │ Subscribed Data │<─────────────────────│    Database     │  │
+│  └─────────────────┘  │                │  └─────────────────┘  │
+│           │           │                │           ^           │
+│           │           │                │           │           │
+│           v           │                │           v           │
+│  +─────────────────┐  │ call_reducer() │  ┌─────────────────┐  │
+│  │   Client Code   │─────────────────────>│   Module Code   │  │
+│  └─────────────────┘  │                │  └─────────────────┘  │
+│                       │                │                       │
+└───────────────────────┘                └───────────────────────┘
+```
+
+TypeScript modules are built with the TypeScript Module Library from [`spacetimedb/server`](https://www.npmjs.com/package/spacetimedb). You define your schema and reducers in TypeScript, and then build and deploy with the [`spacetime` CLI](https://spacetimedb.com/install) using the `spacetime publish` command. Under the hood, SpacetimeDB uses [Rolldown](https://rolldown.rs/) to bundle your application into a single JavaScript artifact before uploading it to the SpacetimeDB host.
+
+> Note: SpacetimeDB also provides a TypeScript **client** SDK at `spacetimedb/sdk`, as well as integrations for frameworks like `spacetimedb/react`. This guide focuses exclusively on the **server-side module** library.
+
+If you’re new to TypeScript, see the [TypeScript Handbook](https://www.typescriptlang.org/docs/handbook/intro.html). For a guided introduction to modules, see the [TypeScript Module Quickstart](/docs/modules/typescript/quickstart).
+
+# Overview
+
+SpacetimeDB modules interact with the outside world via two mechanisms: **tables** and **reducers**.
+
+* [Tables](#tables) store data; public tables are queryable and subscribable by clients.
+* [Reducers](#reducers) are functions that can read and write tables and are callable over the network.
+
+Reducers are atomic and deterministic, there’s no direct filesystem or network access (e.g., `fs`, `fetch`). They execute inside the database with ACID guarantees.
+
+A minimal module looks like this:
+
+```ts
+import { schema, table, t, type RowObj } from 'spacetimedb/server';
+
+// Define a table that is publicly readable by clients
+const players = table({ name: 'players', public: true }, {
+  id: t.u32().primaryKey().autoInc(),
+  name: t.string(),
+});
+
+// Compose a schema from one or more tables
+const spacetimedb = schema(players);
+
+// Define a reducer that inserts a row
+spacetimedb.reducer('add_player', { name: t.string() }, (ctx, { name }) => {
+  ctx.db.players.insert({ id: 0, name });
+});
+```
+
+Reducers don’t return data directly; instead, clients read from tables or subscribe for live updates.
+
+Tables and reducers can use any types built with `t.*` (e.g., `t.string()`, `t.i32()`) or composite types defined with `t.object`, `t.enum`, `t.array`, or `t.option`.
+
+# Setup
+
+1. **[Install the CLI](/install)**
+
+2. **Initialize a TypeScript module project**
+
+   ```bash
+   spacetime init --lang typescript my-project
+   cd my-project
+   ```
+
+   This creates a scaffold with a sample module entrypoint and `package.json`.
+
+3. **Develop**
+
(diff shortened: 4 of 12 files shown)


crates/schema/src/def/validate/v9.rs
```diff
@@ -1033,10 +1152,16 @@ fn check_scheduled_functions_exist(
 /// Check that all function (reducer and procedure) names are unique,
 /// then re-organize the reducers and procedures into [`IndexMap`]s
 /// for storage in the [`ModuleDef`].
```
gefjon: Update this doc comment, please.

crates/schema/src/def/validate/v9.rs
```diff
@@ -426,6 +448,103 @@ impl ModuleValidator<'_> {
...
+        let return_type_for_generate = self.validate_for_type_use(
+            &TypeLocation::ViewReturn {
+                view_name: Cow::Borrowed(&name),
+            },
+            &return_type,
+        );
+
+        // Views don't live in the global namespace.
+        // They share the "function namespace" with reducers and procedures.
+        // Uniqueness is validated in a later pass, in `check_function_names_are_unique`.
```
gefjon: It's not clear to me that views necessarily need to go in the function namespace. That's not to say they shouldn't, but could you expand on your reasoning here?

joshua-spacetime: My reasoning was that either they need to live in the function namespace or the global namespace. The global namespace might be more appropriate if not required since views have much of the same interface as tables. I should have left a `TODO` here. I believe those are the only two options, but feel free to give me your thoughts.

gefjon: I think they need to be in the same namespace as tables in order to make SQL subscriptions work. I don't see any reason they need to also be in the function namespace, since they don't get called by schedulers.

crates/schema/src/def.rs
```diff
@@ -958,6 +971,78 @@ impl From<ScopedTypeName> for RawScopedTypeNameV9 {
     }
 }
 
+/// A view exported by the module.
```
Centril, line 974: Here feels like a good place to give a short explainer on what views are, for the uninitiated.

crates/schema/src/def.rs
```diff
@@ -958,6 +971,78 @@ impl From<ScopedTypeName> for RawScopedTypeNameV9 {
...
+#[derive(Debug, Clone, Eq, PartialEq)]
+#[non_exhaustive]
+pub struct ViewDef {
+    /// The name of the view. This must be unique within the module.
+    pub name: Identifier,
+
+    /// Is this a public or a private view?
+    /// Currently all views are public by default.
+    /// Private views may be supported in the future.
+    pub is_public: bool,
```
Centril, line 984: Assuming views are table-like, which I agree with @gefjon that they are, then it feels like this should be using `StAccess` / `TableAccess`.

Also, is there a (written-up?) rationale for deviating from private-by-default that we have for tables? Not saying its wrong, but it would be good to document.

crates/schema/src/def.rs
```diff
@@ -958,6 +971,78 @@ impl From<ScopedTypeName> for RawScopedTypeNameV9 {
...
+    /// This `ProductType` need not be registered in the module's `Typespace`.
+    pub params: ProductType,
+
+    /// The parameters of the view, formatted for client codegen.
+    ///
+    /// This `ProductType` need not be registered in the module's `TypespaceForGenerate`.
+    pub params_for_generate: ProductTypeDef,
+
+    /// The return type of the view.
+    /// Either `T`, `Option<T>`, or `Vec<T>` where `T` is a [`ProductType`].
```
Centril, line 1003: I think an example of each would help here.

crates/schema/src/def.rs
```diff
@@ -1149,6 +1240,18 @@ impl ModuleDefLookup for ReducerDef {
     }
 }
 
+impl ModuleDefLookup for ViewDef {
```
Centril, line 1245: (note to self: this has become repeated enough to warrant a simple macro.)

crates/schema/src/error.rs
```diff
@@ -131,7 +136,7 @@ pub enum ValidationError {
     MultipleColumnDefaultValues { table: RawIdentifier, col_id: ColId },
     #[error("Table {table} not found")]
     TableNotFound { table: RawIdentifier },
-    #[error("Name {name} is used for multiple reducers and/or procedures")]
+    #[error("Duplicate {name} found among reducers/procedures/views")]
```
Centril: Hmm, I feel like the previous was clearer.

crates/schema/src/error.rs
```diff
@@ -182,8 +187,16 @@ pub enum TypeLocation<'a> {
         position: usize,
         arg_name: Option<Cow<'a, str>>,
     },
+    /// A view argument.
+    ViewArg {
+        view_name: Cow<'a, str>,
+        position: usize,
+        arg_name: Option<Cow<'a, str>>,
+    },
```
Centril, line 195: Feels like a variant Param { item_name, position, param_name, kind } would be good to introduce here, to make the diff go much more red, with e.g., Param { ..., kind: ParamKind::View }. Also Return could be useful.

crates/schema/src/def/validate/v9.rs
joshua-spacetime: Ok I've added them to both. As you mention they need to be in the table namespace. And while they don't need to be in the function namespace as of yet, I've kept them there in case we want them to be `spacetime call`-able in the future.

crates/schema/src/schema.rs
```diff
@@ -587,6 +587,108 @@ pub fn column_schemas_from_defs(module_def: &ModuleDef, columns: &[ColumnDef], t
...
+            params,
+            params_for_generate: _,
+            return_type: _,
+            return_type_for_generate: _,
+            columns: cols,
+        } = view_def;
+
+        let num_args = params.elements.len();
+        let num_cols = cols.len();
+        let n = num_args + num_cols + if *is_anonymous { 0 } else { 2 };
```
Centril: typo? Should be 1 instead of 2?

Also, `n` is a tad short ;)

crates/schema/src/schema.rs
```diff
@@ -587,6 +587,108 @@ pub fn column_schemas_from_defs(module_def: &ModuleDef, columns: &[ColumnDef], t
...
+    /// | sender   | x   | y   | a   | b   |
+    /// |----------|-----|-----|-----|-----|
+    /// | Identity | u32 | u32 | u32 | u32 |
+    ///
+    /// my_anonymous_view:
+    ///
+    /// | x   | y   | a   | b   |
+    /// |-----|-----|-----|-----|
+    /// | u32 | u32 | u32 | u32 |
+    pub fn from_view_def(module_def: &ModuleDef, view_def: &ViewDef) -> Self {
```
Centril, line 619: This could use some commentary for every "paragraph" of code.

crates/schema/src/def/validate/v9.rs
Centril, line 164: This method is getting pretty large. Should refactor soon-ish.

crates/schema/src/def/validate/v9.rs
```diff
@@ -426,6 +448,101 @@ impl ModuleValidator<'_> {
...
+            (name, params_for_generate, return_type_for_generate, columns).combine_errors()?;
+
+        Ok(ViewDef {
+            name,
+            is_anonymous,
+            is_public,
+            params,
+            params_for_generate: ProductTypeDef {
+                elements: params_for_generate,
+                recursive: false, // A ProductTypeDef not stored in a Typespace cannot be recursive.
```
Centril:
```suggestion
                recursive: false, // A `ProductTypeDef` not stored in a `Typespace` cannot be recursive.
```

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -183,6 +184,37 @@ impl MutTxId {
...
+            is_anonymous,
+            is_public,
+            params,
+            columns,
+            ..
+        } = view_def;
+
+        let view_id = self.insert_into_st_view(name.clone().into(), table_id, *is_public, *is_anonymous)?;
+        self.insert_into_st_view_param(view_id, params)?;
+        self.insert_into_st_view_column(view_id, columns)?;
```
Centril, line 215: Maybe make a comment that views don't have any other representation in memory than tables, so they don't need to consider rollback and transactionality beyond tables.

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -183,6 +184,37 @@ impl MutTxId {
...
 
+    /// Create a backing table for a view.
+    ///
+    /// Requires:
+    /// - Everything [`Self::create_table`] requires.
+    ///
+    /// Ensures:
+    /// - Everything [`Self::create_table`] ensures.
+    /// - The returned [`ViewId`] is unique and not [`ViewId::SENTINEL`].
+    pub fn create_view_with_backing_table(
```
Centril, line 196: Consider adding some datastore tests for this, e.g., that a table with the id gets created, and validating the system tables.

crates/schema/src/def.rs
gefjon: I believe this is a case of a misleading comment, not a deviation in behavior: the syntax for a public view is `view(public)`, and for a private view it is `view` or `view(private)`. But because views are not (yet) readable by the module, a private view is not useful or meaningful.

crates/schema/src/def.rs
gefjon: I'm not sure what an example would add - you could decide `T = Product { "x": i32, "name": String }` and then instantiate `Option<T>` and `Vec<T>` for it, but that doesn't seem useful.

crates/schema/src/error.rs
gefjon:
```suggestion
    #[error("Name {name} is used for multiple reducers, procedures and/or views")]
```

crates/schema/src/def.rs
Centril: I found the comment confusing. What is the AT in the case of e.g., `Option<T>`?, is it `AlgebraicType::option(the_t)`? If so, that would be helpful to explain.

crates/schema/src/def.rs
joshua-spacetime: I've clarified the comment

crates/datastore/src/locking_tx_datastore/mut_tx.rs
joshua-spacetime: I'll add tests with the bindings

Issue #3359, opened 2025-10-05
Typescript sdk: react-hooks not functioning properly when used with react-router


Lethalchip:
From discord: https://discord.com/channels/1037340874172014652/1398209084699709492/1423912010927570997

> Seems like the `useTable` struggles with react-router and page changes. If you load in to a page directly, everything on that page works great. However, if you switch into a page after the initial stdb connection, your `useTable`'s on that page fire off but they result in a null array. And they only ever fire off once, so it doesn't seem like they're unsubscribing when you leave that page. Strange! 

> I'm still debugging and fiddling, but I've also made a branch on my lil react-hooks demo app here, if you wanted to see/test/debug: <https://github.com/Lethalchip/simple-stdb-react-hooks-example/tree/bd2bda40f158f96f64e34e193478af5e01d2f5d8> (this is the commit with react-router and the react-hooks not working... fix down below)

> Ok, so I've managed to fix it, but im unsure... why... exactly. This certainly is the bug that <@296507097280020481> is experiencing! My fix is here: <https://github.com/Lethalchip/simple-stdb-react-hooks-example/commit/a9bd712a8a2277002e4877055156a8b33833f446>

> While the `useTable`'s are *supposed* to fire off a subscription (I think?), they're not. I'm not sure why. Could be, (a) it doesn't seem like their initial state sync (onApplied) is firing off when you switch pages around with react-router, or (b) for some reason react-router causes the `useTable` to just not even make the subscription, or it could be (c) neither of these things and something completely different.

> But the "fix"- which isn't really a fix- is to just use an additional subscription builder on each page in conjunction with your `useTable`.

```tsx
  const stdb = useSpacetimeDB<DbConnection>();
  const { rows: counter } = useTable<DbConnection, Counter>('counter');

  useEffect(() => {
    if (!stdb.isActive) return;
    const sub = stdb.subscriptionBuilder()
      .onError((err) => console.error("Counter subscription error:", err))
      .onApplied(() => console.log("Counter subscription applied"))
      .subscribe("SELECT * FROM counter");

    return () => {
      sub.unsubscribeThen(() => console.log("Counter subscription cleaned up"));
    };
  }, [stdb.isActive]);
```

> Ideally, that entire `useEffect` would happen automagically inside the `useTable` so we don't have to think about it.

bfops:
Thank you for filing this issue! We are looking into it 🙂

PR #3428, opened 2025-10-17
Fixed the react router issue and added a new test app to test for it

cloutiertyler:
# Description of Changes

This PR fixes https://github.com/clockworklabs/SpacetimeDB/issues/3359. Additionally I have add a new test app to test for this issue and I changed a perl script to be portable on windows.

The issue was that `useTable` had it's own `isActive` state that it was incorrectly maintaining, when it should just have been using the connections `isActive` state directly. 

This fixes a critical bug and should be pushed ASAP.

None, fixes a critical bug.

2, quite straightforward fix after figuring it out.

# Testing

- [x] I added a test app and tested the fix in my browser.

### DIFF
--- a/crates/bindings-typescript/test-react-router-app/server/Cargo.toml
+++ b/crates/bindings-typescript/test-react-router-app/server/Cargo.toml
@@ -9,6 +9,6 @@ edition = "2021"
 crate-type = ["cdylib"]
 
 [dependencies]
-spacetimedb = "1.2.0"
+spacetimedb.workspace = true
 log = "0.4"
 anyhow = "1.0"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9168,7 +9168,7 @@ version = "0.1.0"
 dependencies = [
  "anyhow",
  "log",
- "spacetimedb 1.6.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "spacetimedb 1.6.0",
 ]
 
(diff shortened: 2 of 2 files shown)


KingCosmic:
Hey! I just made issue #3431 and hit this issue myself. I was wanting to share that I think your code will fix it as long as the table hook runs *after* isActive is set to true. cause it doesn't seem like the useSpacetimeDB hook causes updates.

crates/bindings-typescript/test-react-router-app/server/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+spacetimedb = "1.2.0"
```
bfops: it looks like this depends on an old `spacetimedb` from crates.io rather than the workspace version? Doesn't this mean that it won't be testing the current state of the code?

cloutiertyler: Oh hmm, this is a good question. Yes, but I will note the module in this case is only used for code generation and not actually tested. This is for testing the client. Probably this should be changed to workspace though to avoid future weirdness.

PR #3429, opened 2025-10-17
Add macro bindings for views

joshua-spacetime:
# Description of Changes

Adds the `#[view]` procedural macro and module describers for views.
```rust
#[view(public)]
fn player(ctx: &ViewContext) -> Vec<Player> {
    ctx.db.player().identity().find(ctx.sender).into_iter().collect()
}

#[view(public)]
fn player(ctx: &AnonymousViewContext, level: u32) -> Vec<Player> {
    ctx.db.player().level().filter(level).collect()
}
```

Note, this deviates from the proposal in that views may only return `Vec<T>` or `Option<T>`. They can't return an arbitrary `SpacetimeType`.

This technically isn't a breaking change, but it's worth mentioning that this patch refactors `ReducerInfo` so that we can use it for views as well.

2

# Testing

- [x] Negative compile tests
- [x] Negative publish (module validation) tests
- [x] Test system tables are updated accurately

### DIFF
--- a/crates/bindings/src/lib.rs
+++ b/crates/bindings/src/lib.rs
@@ -762,7 +762,7 @@ pub use spacetimedb_bindings_macro::reducer;
 ///
 /// // An example of a join that combines fields from two different tables
 /// #[view(public)]
-/// fn players_with_coordinates(ctx: &AnonymousViewContext, x: u64, y: u64) -> Vec<Player> {
+/// fn players_with_coordinates(ctx: &AnonymousViewContext, x: u64, y: u64) -> Vec<PlayerAndLocation> {
 ///     ctx
 ///         .db
 ///         .location()


crates/bindings-macro/src/reducer.rs
```diff
@@ -139,11 +139,12 @@ pub(crate) fn reducer_impl(args: ReducerArgs, original_function: &ItemFn) -> syn
...
         }
         #[automatically_derived]
-        impl spacetimedb::rt::ReducerInfo for #func_name {
+        impl spacetimedb::rt::FnInfo for #func_name {
+            type Invoke = spacetimedb::rt::ReducerFn;
             const NAME: &'static str = #reducer_name;
             #(const LIFECYCLE: Option<spacetimedb::rt::LifecycleReducer> = Some(#lifecycle);)*
             const ARG_NAMES: &'static [Option<&'static str>] = &[#(#opt_arg_names),*];
-            const INVOKE: spacetimedb::rt::ReducerFn = #func_name::invoke;
+            const INVOKE: Self::Invoke = #func_name::invoke;
```
Centril, line 147: I think this is worth mentioning in the PR description. In some sense it is breaking, but not in a way we promise.

crates/bindings-macro/src/view.rs
```diff
@@ -0,0 +1,284 @@
...
+
+use crate::sym;
+use crate::util::{ident_to_litstr, match_meta};
+
+pub(crate) struct ViewArgs {
+    anonymous: bool,
+}
+
+impl ViewArgs {
+    /// Parse `#[view(public)]` where public is required
```
Centril:
```suggestion
    /// Parse `#[view(public)]` where `public` is required.
```

crates/bindings-macro/src/view.rs
```diff
@@ -0,0 +1,284 @@
...
+    pub(crate) fn parse(input: TokenStream) -> syn::Result<Self> {
+        if input.is_empty() {
+            return Err(syn::Error::new(
+                proc_macro2::Span::call_site(),
+                "views must be declared as #[view(public)]; public is required",
+            ));
+        }
```
Centril, line 22: I would look at `TableArgs::parse`. I think you should collect an `access: Option<TableAccess>` and e.g., `Option<Span>`  for `anonymous` and then ensure that at the end, `access.is_some()` or error.

crates/bindings-macro/src/view.rs
```diff
@@ -0,0 +1,284 @@
...
+                sym::public => {
+                    public = true;
+                }
+                sym::anonymous => {
+                    anonymous = true;
+                }
+            });
+            Ok(())
+        })
+        .parse2(input)?;
```
Centril, line 38: This allows `public, anonymous, public`.

crates/bindings-macro/src/view.rs
```diff
@@ -0,0 +1,284 @@
...
+        if !public {
+            return Err(syn::Error::new(
+                proc_macro2::Span::call_site(),
```
Centril:
```suggestion
                Span::call_site(),
```

crates/bindings-macro/src/view.rs
```diff
@@ -0,0 +1,284 @@
...
+                "views must be declared as #[view(public)]; public is required",
```
Centril:
```suggestion
                "views must be declared as `#[view(public)]`; `public` is required",
```

crates/bindings-macro/src/view.rs
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    // Extract all function parameters, except for `self` ones that aren't allowed.
+    let typed_args = original_function
+        .sig
+        .inputs
+        .iter()
+        .map(|arg| match arg {
+            FnArg::Typed(arg) => Ok(arg),
+            FnArg::Receiver(_) => Err(syn::Error::new_spanned(arg, "`self` arguments not allowed in views")),
```
Centril:
```suggestion
            FnArg::Receiver(_) => Err(syn::Error::new_spanned(arg, "`self` parameters not allowed in views")),
```

crates/bindings-macro/src/view.rs
```diff
@@ -0,0 +1,284 @@
...
+            return Err(syn::Error::new(
+                proc_macro2::Span::call_site(),
+                "views must be declared as #[view(public)]; public is required",
+            ));
+        }
+        Ok(Self { anonymous })
+    }
+}
+
+fn view_impl_anon(original_function: &ItemFn) -> syn::Result<TokenStream> {
```
Centril: I feel like there's a lot of copy-pasta here with `reducer.rs`. I'd like to see the common bits extracted to functions and reused here.

crates/bindings-macro/src/view.rs
```diff
@@ -0,0 +1,284 @@
...
+            quote!(Some(#name))
+        } else {
+            quote!(None)
+        }
+    });
+
+    let arg_tys = typed_args.iter().map(|arg| arg.ty.as_ref()).collect::<Vec<_>>();
+
+    // Extract the context type
+    let ctx_ty = arg_tys.first().ok_or_else(|| {
```
Centril: Could use `let [ctx_ty, arg_tys @ ..] = arg_tys else { ... }` here to avoid `.skip(1)` later.

crates/bindings-macro/src/view.rs
```diff
@@ -0,0 +1,284 @@
...
+            fn return_type(
+                ts: &mut impl spacetimedb::sats::typespace::TypespaceBuilder
+            ) -> Option<spacetimedb::sats::AlgebraicType> {
+                Some(<#ret_ty as spacetimedb::SpacetimeType>::make_type(ts))
+            }
+        }
+    })
+}
+
+fn view_impl_client(original_function: &ItemFn) -> syn::Result<TokenStream> {
```
Centril: Same re. copy pasta here.

crates/bindings/src/lib.rs
```diff
@@ -667,6 +667,53 @@ pub use spacetimedb_bindings_macro::table;
...
+/// pub fn player(ctx: &ViewContext) -> Vec<Player> {
+///     ctx.db.player().identity().find(ctx.sender).into_iter().collect()
+/// }
+///
+/// #[view(public, anonymous)]
+/// pub fn player(ctx: &AnonymousViewContext, level: u32) -> Vec<Player> {
+///     ctx.db.player().level().filter(level).collect()
+/// }
+/// # }
+/// ```
```
Centril, line 781: Maybe an example of a join and an example of a part of a table?

crates/bindings/src/rt.rs
```diff
@@ -43,19 +45,84 @@ pub trait Reducer<'de, A: Args<'de>> {
     fn invoke(&self, ctx: &ReducerContext, args: A) -> ReducerResult;
 }
 
-/// A trait for types that can *describe* a reducer.
-pub trait ReducerInfo {
-    /// The name of the reducer.
+/// Invoke a caller-specific view.
+/// Returns a bsatn encoded vec of rows.
```
Centril:
```suggestion
/// Returns a BSATN encoded `Vec` of rows.
```
(and in other places)

crates/bindings/src/rt.rs
```diff
@@ -43,19 +45,84 @@ pub trait Reducer<'de, A: Args<'de>> {
...
+pub fn invoke_view<'a, A: Args<'a>, T: SpacetimeType + Serialize>(
+    view: impl View<'a, A, T>,
+    ctx: ViewContext,
+    args: &'a [u8],
+) -> Vec<u8> {
+    // Deserialize the arguments from a bsatn encoding.
+    let SerDeArgs(args) = bsatn::from_slice(args).expect("unable to decode args");
```
Centril, line 72: Would be nice to dedup this into a function.

crates/bindings/src/rt.rs
```diff
@@ -43,19 +45,84 @@ pub trait Reducer<'de, A: Args<'de>> {
...
+    let rows: Vec<T> = view.invoke(&ctx, args);
```
Centril, line 73: This makes me feel that `invoke` should return an iterator so that we can avoid the `Vec`.

crates/bindings/src/rt.rs
```diff
@@ -119,6 +186,39 @@ pub trait ReducerArg {
...
+}
+impl<T: SpacetimeType> ViewArg for T {}
+
+/// A trait of types that can be the return type of a view.
+#[diagnostic::on_unimplemented(message = "Views must return `Vec<T>` where `T` is a `SpacetimeType`")]
+pub trait ViewReturn {
+    #[doc(hidden)]
+    const _ITEM: () = ();
+}
+impl<T: SpacetimeType> ViewReturn for Vec<T> {}
```
Centril, line 232: On the iterator theme, this could be implemented for anything that is `IntoIterator<Item = T>`.

crates/bindings-macro/src/reducer.rs
joshua-spacetime: In what sense exactly?

Centril: If this were to be implemented manually, we would break such users. We haven't promised such stability, but it might be worth mentioning nonetheless.

crates/bindings/src/rt.rs
joshua-spacetime: We can extend it later if we want. Right now let's just implement it according to the spec.

crates/bindings-macro/src/view.rs
joshua-spacetime: I've compressed everything into a single impl so there's much less duplication now. The differences are meaningful enough that I don't think we have to dedup between `reducer.rs`, especially given the quasi-quotation.

crates/bindings/src/lib.rs
```diff
@@ -682,26 +682,60 @@ pub use spacetimedb_bindings_macro::reducer;
...
+///     ctx.db.player().identity().find(ctx.sender).map(|Player { id, .. }| PlayerId { id })
+/// }
+///
+/// #[view(public)]
 /// pub fn players_at_level(ctx: &AnonymousViewContext, level: u32) -> Vec<Player> {
 ///     ctx.db.player().level().filter(level).collect()
 /// }
+///
+/// #[view(public)]
+/// pub fn players_at_coordinates(ctx: &AnonymousViewContext, x: u64, y: u64) -> Vec<Player> {
```
Centril: Neat! If you can tweak this to return `PlayerAndLocation` then that would be even better.

joshua-spacetime: That's the purpose of the `my_player_id` view and the `PlayerId` type.

Centril (approved):
We discussed this after the meeting. It looks good. I think we can improve things a bit but we can do that later. Let's add the join example though.

PR #3432, opened 2025-10-19
feat: update pgwire to 0.34 and improve how we disable ssl

sunng87:
# Description of Changes

Hello team, I'm the original author of pgwire library. In this patch, I have some updates about spacetimedb's usage of pgwire library:

- Updated pgwire library to 0.34, which includes a critical fix for busy loop on accepting new connection
- This version will return error on no-op handlers, as requested by @mamcx 
- Update `StartupHandler` for `SslRequest`. `SslRequest` is handled in `process_socket` automatically before `StartupHandler` kicks in. So it's safe to remove the match branch. The framework will reject TLS negotiation automatically if no TLS support compiled
- Updated feature flags. Add `no-default-features` so we will not pull TLS dependencies.

N/A

1

# Testing

Would be nice to verify if you have integration test on postgres interface.


bfops:

mamcx (approved):
LGTM

PR #3437, opened 2025-10-20
commitlog,durability: Support preallocation of disk space

kim:
When a new commitlog segment is created, allocate disk space for it up
to the maximum segment size. Also do this when resuming writes to an
existing segment, such that segments created without preallocation will
allocate as well when the database is opened.

Preallocation is gated behind the feature "fallocate", because it is not
always desirable to preallocate, e.g. for local `standalone` users.

The feature can only be enabled on Linux targets, because allocation is
done using the Linux-specific `fallocate(2)` system call.

Unlike `ftruncate(2)` or the portable `posix_fallocate(3)`, `fallocate(2)`
supports allocating disk space without zeroing. This is currently
required, because the commitlog format does not handle padding bytes.

If not enough space can be allocated, the commitlog refuses writes. For
commitlogs that were created without preallocation, this means that the
commitlog cannot even be opened in this situation.

The local durability impl will crash if it detects that the commitlog is
unable to allocate enough space.

This means that a database will eventually crash and be unable to start in
an out-of-space situation.

Allocated space is not included in the reported size of the commitlog.
Instead, allocated blocks are reported separately.

3 - Disk size monitoring may need to be adjusted.

# Testing

- [x] Adds a test that demonstrates the crash behavior of [`spacetimedb_durability::Local`]
      when there is insufficient space. The test performs I/O against a loop device.
- [x] Modified the `repo::Memory` impl so that it can run out of space. No test currently
      utilizes this, but existing tests assuming infinite space still pass.

### DIFF
--- a/crates/durability/tests/io/fallocate.rs
+++ b/crates/durability/tests/io/fallocate.rs
@@ -168,6 +168,7 @@ async fn local_durability(
             commitlog: spacetimedb_commitlog::Options {
                 max_segment_size,
                 max_records_in_commit: 1.try_into().unwrap(),
+                preallocate_segments: true,
                 ..<_>::default()
             },
             ..<_>::default()


Shubham8287 (approved):
Sorry, I reviewed this PR long back but completely forgot to hit the button.

Tests seems good.

PR #3442, opened 2025-10-21
Add typescript bindings for auth claims.

jsdt:
# Description of Changes

This adds an `AuthCtx` for typescript modules.

This adds a function to the typescript ABI.

1.5.

# Testing

I tested with by linking with this for a local typescript module.

### DIFF
--- a/crates/core/src/host/wasmtime/wasm_instance_env.rs
+++ b/crates/core/src/host/wasmtime/wasm_instance_env.rs
@@ -9,7 +9,6 @@ use crate::host::wasm_common::{err_to_errno_and_log, RowIterIdx, RowIters, Timin
 use crate::host::AbiCall;
 use anyhow::Context as _;
 use spacetimedb_data_structures::map::IntMap;
-use spacetimedb_datastore::locking_tx_datastore::state_view::StateView;
 use spacetimedb_lib::{ConnectionId, Timestamp};
 use spacetimedb_primitives::{errno, ColId};
 use std::num::NonZeroU32;


crates/core/src/host/instance_env.rs
```diff
@@ -187,6 +188,11 @@ impl InstanceEnv {
         self.tx.get()
     }
 
+    pub(crate) fn get_jwt_payload(&self, connection_id: ConnectionId) -> Result<Option<String>, NodesError> {
```
Centril, line 191: Reuse for WasmInstanceEnv?

crates/core/src/host/v8/syscall.rs
```diff
@@ -961,6 +932,17 @@ fn datastore_update_bsatn(scope: &mut PinScope<'_, '_>, args: FunctionCallbackAr
     Ok(row)
 }
 
+fn get_jwt_payload(scope: &mut PinScope<'_, '_>, args: FunctionCallbackArguments<'_>) -> SysCallResult<Vec<u8>> {
```
Centril: Let's move this to before `fn identity` like in `wasm_instance_env`.
Also some docs conforming to the format of the other ABIs would be good.

crates/bindings-typescript/src/server/reducers.ts
```diff
@@ -65,6 +65,41 @@ export type DbView<SchemaDef extends UntypedSchemaDef> = {
...
+/**
+ * Auth Claims extracted from a the payload of a JWT token
+ */
+export interface JwtClaims {
+  readonly rawPayload: string;
+  readonly subject: string;
+  readonly issuer: string;
+  readonly audience: readonly string[];
+  readonly identity: Identity;
+  readonly fullPayload: JsonObject;
```
Centril, line 107: Doc comments on each of these would be neat.

PR #3390, opened 2025-10-09
Phoebe/procedures

gefjon:
# Description of Changes

# API and ABI breaking changes

# Expected complexity level and risk

# Testing

- [ ] 
- [ ]

PR #3444, opened 2025-10-21
Rust module bindings and macros for defining procedures

gefjon:
# Description of Changes

This commit adds a macro attribute `#[procedure]` which applies to functions, and various in-WASM machinery for defining, calling and running procedures.

A simple example of a procedure, included in `modules/module-test`:

```rust
fn sleep_one_second(ctx: &mut ProcedureContext) {
    let prev_time = ctx.timestamp;
    let target = prev_time + Duration::from_secs(1);
    ctx.sleep_until(target);
    let new_time = ctx.timestamp;
    let actual_delta = new_time.duration_since(prev_time).unwrap();
    log::info!("Slept from {prev_time} to {new_time}, a total of {actual_delta:?}");
}
```

We intend eventually to make procedures be `async` functions (with the trivial `now_or_never` executor from `future-util`), but I found that making the types work for this was giving me a lot of trouble, and decided to put it off in the interest of unblocking more parallelizable work.

Host-side infrastructure for executing procedures is not included in this commit. I have a prototype working, but cleaning it up for review and merge will come a bit later.

One item of complexity in this PR is enabling scheduled tables to specify either reducers or procedures, while still providing compile-time diagnostics for ill-typed scheduled functions (as opposed to publish-time). I had to rewrite the previous `schedule_reducer_typecheck` into a more complex `schedule_typecheck` with a trait `ExportFunctionForScheduledTable`, which takes a "tacit trait parameter" encoding reducer-ness or procedure-ness, as described in https://willcrichton.net/notes/defeating-coherence-rust/ .

The trait name `ExportFunctionForScheduledTable` is user-facing in the sense that it will appear in compiler diagnostics in ill-typed modules. As such, I am open to bikeshedding.

Adds a new user-facing API, which we intend to change before releasing.

2? Mostly pretty mechanical changes to macros and bindings.

# Testing

- [x] Added a procedure definition to `module-test`, saw that it typechecks.
- [x] Executed same procedure definition using #3390 , the prototype implementation this PR draws from.

### DIFF
--- a/modules/module-test/src/lib.rs
+++ b/modules/module-test/src/lib.rs
@@ -450,12 +450,9 @@ fn sleep_one_second(ctx: &mut ProcedureContext) {
     log::info!("Slept from {prev_time} to {new_time}, a total of {actual_delta:?}");
 }
 
-#[derive(SpacetimeType)]
-struct MyReturnValue {
-    foo: u64,
-}
-
 #[spacetimedb::procedure]
-fn return_value(_ctx: &mut ProcedureContext, foo: u64) -> MyReturnValue {
-    MyReturnValue { foo }
+fn return_value(_ctx: &mut ProcedureContext, foo: u64) -> Baz {
+    Baz {
+        field: format!("{foo}"),
+    }
 }


gefjon:
Test failures are coming from two places, AFAICT:

1. Adding a procedure to `module-test` breaks `ensure_same_schema`. I will have to remove the example before merging.
2. The trait `ExportFunction` is under-constrained, causing `scheduled_function_typecheck` to pass unconditionally.

1 is easy enough, but 2 is challenging. Rust seems to not like implementing traits for specific function types, and is not smart enough to recognize that `Fn(&ReducerContext, ...)` is disjoint from `Fn(&mut ProcedureContext, ...)` (or maybe it isn't. Can you implement `Fn` manually?). As such I've been unable to come up with an incantation where `ExportFunction` gets impled for every valid reducer or procedure function type but not for any other types.

gefjon:
I have fixed the bug I listed above as 2! I will update the PR description with a note.

gefjon:
Looks like I was wrong about the ensure_same_schema checks, actually - those don't seem to care about reducers and procedures.

crates/bindings-macro/src/procedure.rs
```diff
@@ -0,0 +1,129 @@
...
+use crate::sym;
+use crate::util::{check_duplicate, ident_to_litstr, match_meta};
+use proc_macro2::TokenStream;
+use quote::quote;
+use syn::parse::Parser as _;
+use syn::{ItemFn, LitStr};
+
+#[derive(Default)]
+pub(crate) struct ProcedureArgs {
+    name: Option<LitStr>,
```
joshua-spacetime, line 12: Is an optional name argument required, or is this just for consistency with reducers?

crates/bindings/src/lib.rs
```diff
@@ -949,6 +952,73 @@ impl ReducerContext {
...
 
+/// The context that any procedure is provided with.
+///
+/// Each procedure must accept `&mut ProcedureContext` as its first argument.
+///
+/// Includes information about the client calling the procedure and the time of invocation,
+/// and exposes methods for running transactions and performing side-effecting operations.
+///
+/// If the crate was compiled with the `rand` feature,
+/// also includes faculties for random number generation.
```
joshua-spacetime: This is only true if we include `rng` correct?

crates/bindings/src/lib.rs
```diff
@@ -949,6 +952,73 @@ impl ReducerContext {
...
+    /// Will be `None` for certain scheduled procedures.
+    pub connection_id: Option<ConnectionId>,
+    // TODO: Add rng?
+    // Complex and requires design because we may want procedure RNG to behave differently from reducer RNG,
+    // as it could actually be seeded by OS randomness rather than a deterministic source.
+}
+
+impl ProcedureContext {
+    /// Read the current module's [`Identity`].
+    pub fn identity(&self) -> Identity {
```
joshua-spacetime, line 1062: Maybe call this `module_identity` then just to clearly distinguish between this and the `sender` identity.

Also why does a `ProcedureContext` need to expose the module's identity?

crates/bindings/src/rt.rs
```diff
@@ -179,6 +211,18 @@ impl<E: fmt::Display> IntoReducerResult for Result<(), E> {
     }
 }
 
+#[diagnostic::on_unimplemented(
+    message = "The procdure return type `{Self}` does not implement `SpacetimeType`",
```
joshua-spacetime:
```suggestion
    message = "The procedure return type `{Self}` does not implement `SpacetimeType`",
```

crates/bindings/src/rt.rs
```diff
@@ -310,21 +377,60 @@ impl ViewRegistrar<AnonymousViewContext> {
...
+///
+/// Holds the procedure's return type in order to avoid an error due to an unconstrained type argument.
+pub struct FnKindProcedure<Ret> {
+    _never: Infallible,
+    _ret_ty: PhantomData<fn() -> Ret>,
+}
+
+/// Tacit marker argument to [`ExportFunctionForScheduledTable`] for views.
+///
+/// Because views are never scheduled, we don't need to distringuish between anonymous or sender-identity views,
```
joshua-spacetime:
```suggestion
/// Because views are never scheduled, we don't need to distinguish between anonymous or sender-identity views,
```

crates/bindings/src/rt.rs
```diff
@@ -396,15 +502,15 @@ impl<'de, A: Args<'de>> de::ProductVisitor<'de> for ArgsVisitor<A> {
     }
 }
 
-macro_rules! impl_reducer {
+macro_rules! impl_reducer_and_procedure {
```
joshua-spacetime: I forgot to update the name of this macro. I suppose it should now be called `impl_reducer_procedure_and_view`.

crates/bindings/src/rt.rs
```diff
@@ -760,15 +895,114 @@ extern "C" fn __call_reducer__(
...
         Err(msg) => {
-            write_to_sink(error, msg.as_bytes());
+            write_to_sink(out, msg.as_bytes());
             errno::HOST_CALL_FAILURE.get() as i16
         }
     }
 }
 
+/// Called by the host to execute a procedure
+/// when the `sender` calls the reducer identified by `id` at `timestamp` with `args`.
```
joshua-spacetime:
```suggestion
/// when the `sender` calls the procedure identified by `id` at `timestamp` with `args`.
```

crates/bindings/tests/ui/views.stderr
```diff
@@ -453,28 +453,22 @@ error[E0277]: the trait bound `NotSpacetimeType: SpacetimeType` is not satisfied
...
-   --> src/rt.rs
     |
-    | pub const fn scheduled_reducer_typecheck<'de, Row>(_x: impl ReducerForScheduledTable<'de, Row>)
-    |                                                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `scheduled_reducer_typecheck`
-help: consider wrapping the function in a closure
+    = help: the trait `ExportFunctionForScheduledTable<'_, ScheduledTable, FnKindView>` is not implemented for fn item `for<'a> fn(&'a ViewContext, ScheduledTable) -> Vec<Player> {scheduled_table_view}`
+    = note: views cannot be scheduled
+    = note: the scheduled function must take `ScheduledTable` as its sole argument
+    = note: e.g: `fn scheduled_reducer(ctx: &ReducerContext, arg: ScheduledTable)`
+    = note: or `fn scheduled_procedure(ctx: &mut ProcedureContext, arg: ScheduledTable)`
```
joshua-spacetime, line 469: This seems close enough to me.

crates/bindings-macro/src/procedure.rs
gefjon: Just for consistency with reducers.

crates/bindings/src/lib.rs
gefjon: Ah, yeah, I'll remove this for now.

crates/bindings/src/lib.rs
gefjon: `ReducerContext` already has this same method. It's useful to be able to read the module's identity e.g. to assert that scheduled functions are not being called by an external client.

crates/bindings/src/rt.rs
gefjon: I'm going with `impl_reducer_procedure_view`, as I see the `and` as noise when the name is this long already.

crates/bindings/src/lib.rs
gefjon: I agree that `module_identity` is a better name for this method; I think we have a ticket somewhere to change it. I am unsure whether we want to make that change now fully, to keep the existing method on `ReducerContext` but use the better name for `ProcedureContext`, or to be consistent for now. I have gone with the latter, but could easily be convinced otherwise.

crates/bindings-macro/src/procedure.rs
gefjon: I've added a comment describing what this argument does and why it's included.

crates/bindings/src/lib.rs
joshua-spacetime: I see. Well we can't change the name of the method on `ReducerContext` right? So I accept the consistency argument.

gefjon: No, but we can add a new method with the good name and mark the old method as deprecated.

crates/bindings/src/lib.rs
```diff
@@ -667,6 +670,92 @@ pub use spacetimedb_bindings_macro::table;
...
+///
+/// A procedure is a function that runs within the database and can be invoked remotely by [clients],
+/// but unlike a [`reducer`], a  procedure is not automatically transactional.
+/// This allows procedures to perform certain side-effecting operations,
+/// but also means that module developers must be more careful not to corrupt the database state
+/// when execution aborts or operations fail.
+///
+/// When in doubt, prefer writing [`reducer`]s unless you need to perform an operation only available to procedures.
+///
+/// The first argument of a reducer is always `&mut ProcedureContext`.
```
joshua-spacetime:
```suggestion
/// The first argument of a procedure is always `&mut ProcedureContext`.
```

crates/bindings/src/lib.rs
```diff
@@ -667,6 +670,92 @@ pub use spacetimedb_bindings_macro::table;
...
+/// The [`ProcedureContext`] exposes information about the caller and allows side-effecting operations.
+///
+/// After this, a reducer can take any number of arguments.
```
joshua-spacetime:
```suggestion
/// After this, a procedure can take any number of arguments.
```

PR #3446, opened 2025-10-21
gitignore `node_modules` even when it's a symlink

bfops:
# Description of Changes

Just a gitignore tweak.

None

1

# Testing

- [x] `git` is no longer trying to commit my `node_modules` symlink

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,10 +51,10 @@ build/Release
 
 # Dependency directories
 node_modules
-jspm_packages/
+jspm_packages
 
 # Snowpack dependency directory (https://snowpack.dev/)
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -50,7 +50,7 @@ bower_components
 build/Release
 
 # Dependency directories
-node_modules/
+node_modules
 jspm_packages/
```
jdetter: Do we want to do the same thing for any of these other refs?

.gitignore
jdetter: Maybe this one too idk? Do we ever need the slash at the end?

Issue #1387, opened 2024-06-11
Typescript SDK client-side indexes


cloutiertyler:
Currently columns that are declared as indexed in the schema will generate corresponding find and filter functions however, these lookups aren't optimized on the TypeScript sdk side. They will loop over the entire client cache comparing the specific key for equality.

The csharp sdk has instead already optimized these by building dictionary lookups for these declared indexes. 
We want to have this same behavior in the typescript sdk.

bfops:
(mainly because we want to have parity with our other SDKs)

PR #3447, opened 2025-10-21
Code generate filter function for non-unique indexes

aasoni:
# Description of Changes

Update typescript code generation to generate accessors on non unique indexes. These will use the `filter` function as is the convention in C# and module api.

Related to https://github.com/clockworklabs/SpacetimeDB/issues/1387 but doesn't close it.

2

# Testing

Generated the following table:
```rust
#[spacetimedb::table(
    name = test_index,
    index(name = multi_column_test, btree(columns = [identity, col_1, col_2, col_3, col_4])), 
    public,
)]
#[derive(Clone)]
pub struct TestIndex {
    #[primary_key]
    #[auto_inc]
    pub id: u64,

    pub identity: Identity,
    pub col_1: u64,
    pub col_2: i32,
    pub col_3: f32,
    pub col_4: String,
}
```

Into typescript which resulted in the following code for index lookups:
```ts
  /**
   * Access to the `id` index on the table `test_index`,
   * with `ctx.db.testIndex.id.find(...)`.
   */
  id = {
    // Find the subscribed row matching the lookup value if present in the client cache.
    find: (col_vals: bigint): TestIndex | undefined => {
      for (let row of this.tableCache.iter()) {
        if (__deepEqual(row.id, col_vals)) {
          return row;
        }
      }
    },
  };
  /**
   * Access to the `multiColumnTest` index on the table `test_index`,
   * with `ctx.db.testIndex.multiColumnTest.filter(...)`.
   */
  multiColumnTest = {
    // Find the subscribed row matching the lookup value if present in the client cache.
    filter: (col_vals: [__Identity,bigint,number,number,string]): TestIndex[] => {
      const result: TestIndex[] = [];
      for (let row of this.tableCache.iter()) {
        if (__deepEqual([row.identity,row.col1,row.col2,row.col3,row.col4], col_vals)) {
          result.push(row);
        }
      }
      return result;
    },
  };
```

I then manually inserted values into the test_index table and checked that calling the new filter and find functions returned the expected values.


cloutiertyler:
This was implemented in the other TypeScript overhaul PR

PR #3448, opened 2025-10-21
CI - move as much as possible off of spacetimedb-runner

bfops:
# Description of Changes

We have persistent issues around CI jobs getting cancelled mysteriously and unceremoniously. It has been hard to figure out why this happens, so this workaround just moves as many things as possible off of the custom runner.

This slowed down the smoketests from 17m -> 22m.

CI-only change.

1

# Testing

- [x] CI still passes

### DIFF
--- a/.github/workflows/csharp-test.yml
+++ b/.github/workflows/csharp-test.yml
@@ -8,6 +8,7 @@ on:
 
 jobs:
   unity-testsuite:
+    # This can't go on ubuntu-latest (at least as-is) because it relies on pre-installed Unity.
     runs-on: spacetimedb-runner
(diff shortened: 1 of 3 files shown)


.github/workflows/csharp-test.yml
```diff
@@ -8,6 +8,7 @@ on:
 
 jobs:
   unity-testsuite:
+    # This can't go on ubuntu-latest (at least as-is) because it relies on pre-installed Unity.
```
bfops, line 11: is this true?

bfops:
closing in favor of https://github.com/clockworklabs/SpacetimeDB/pull/3656

PR #3450, opened 2025-10-21
fix migration which adds column and index together.

### LINKED ISSUE #3441: Breaking publish with new column and new index results in failure
Publishing a breaking module update when adding both a new column and an index on a different column in a single table fails with a `TableError`.

Reproduction, starting with the sample module:
```diff
 pub struct Person {
+    #[index(btree)]
     name: String,
+    #[default(0)]
+    test: u32,
 }
```
```
▸ Created column in table person
    + test: U32 (default: U32(
    0,
))
▸ Created index person_name_idx_btree on [name] of table person
!!! Warning: All clients will be disconnected due to breaking schema changes

The above changes will BREAK existing clients. Do you want to proceed? [y/N]y
Publishing module...
Error: Failed to create or update the database: DatastoreError: TableError: Table with ID `4096` not found in `TxState`.
```

Shubham8287:
# Description of Changes

`auto_migrate_database` should not take table schema as input because each step is ordered and may depend on changes made by a prior step, this approach can lead to incorrect assumptions (e.g., adding a column changes its table id).

Migration steps now retrieve the current table schema when they run. This adds a small performance cost, but it simplifies the logic and code is not under hotpath.

fixes #3441 

NA

1

# Testing
Update smoketest to add failed case.


smoketests/tests/auto_migration.py
```diff
@@ -265,6 +265,7 @@ class AddTableColumns(Smoketest):
 #[derive(Debug)]
 #[spacetimedb::table(name = person)]
 pub struct Person {
+    #[index(btree)]
```
Centril, line 270: Can you leave a comment here referencing the issue and noting that this is load bearing.

crates/core/src/db/update.rs
```diff
@@ -95,33 +92,23 @@ fn auto_migrate_database(
...
     logger: &dyn UpdateLogger,
-    existing_tables: Vec<Arc<TableSchema>>,
 ) -> anyhow::Result<UpdateResult> {
-    // We have already checked in `migrate_database` that `existing_tables` are compatible with the `old` definition in `plan`.
-    // So we can look up tables in there using unwrap.
-
-    let table_schemas_by_name = existing_tables
-        .into_iter()
-        .map(|table| (table.table_name.clone(), table))
-        .collect::<HashMap<_, _>>();
```
Centril, line 106: Let's leave a comment explaining that we did this and why we don't anymore.

Shubham8287: Putting a comment like - 
```
// Instead of memoizing all table schemas upfront, we query the schema only when needed.
// This ensures that any schema changes made during earlier migration steps are visible
// to later steps, which might depend on those modifications.
```
?
commenting on what might seem like an obvious behavior (Or about not doing anything) feels bit odd

Centril: I would also add the issue number and note specifically that we had a bug before this.

PR #3452, opened 2025-10-22
Adjust docusaurus migration styling

cloutiertyler:
# Description of Changes

This PR fixes small styling issues with the [docusaurus migration PR](https://github.com/clockworklabs/SpacetimeDB/pull/3343).

Mainly it:

1. Ensures that we only have one single reference to any particular color definition (`#fefaef`) so that if we change the style guide (see figma) we can easily update the colors.
2. It removes references to any colors that are not specified in the style guide
3. It updates font sizes and styling to be one of the pre-defined styles specified in the style guide
4. Adds the precise fonts specified in the style guide and ensures that they are preloaded
5. Makes very minor docs changes
6. Implements the tab highlighting for the selected tab (and on hover)

image

After the docusaurus PR merges I would like to do a larger update to some of the docs content, but I'm keeping that separate.

None

1

# Testing

- [x] Tested styles locally
- [x] Built and ran pnpm run sever to verify that fonts were being correctly preloaded

### DIFF
--- a/docs/src/css/custom.css
+++ b/docs/src/css/custom.css
@@ -327,6 +327,15 @@ footer {
     white-space: nowrap;
     text-overflow: ellipsis;
     border-left: 1px solid var(--clockworklabs-color-n5);
+
+    &:has(.menu__link--active) {
+      border-left: 1px solid var(--clockworklabs-color-white);
+    }
+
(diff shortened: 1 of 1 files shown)


JulienLavocat (approved):
LGTM

PR #3455, opened 2025-10-22
Remove unused docker-compose files

bfops:
# Description of Changes

These files don't seem to be used anywhere.

None

1

# Testing

Existing CI only


docker-compose-release.yml
```diff
@@ -1,34 +0,0 @@
...
-      - ./crates/sats:/usr/src/app/crates/sats
-      - ./crates/bindings:/usr/src/app/crates/bindings
-      - ./crates/bindings-macro:/usr/src/app/crates/bindings-macro
-      - ./crates/bindings-sys:/usr/src/app/crates/bindings-sys
-      - ./crates/vm:/usr/src/app/crates/vm
-      - ./crates/metrics:/usr/src/app/crates/metrics
-      - key_files:/etc/spacetimedb
-      - /stdb
-    ports:
-      - "3000:80"
```
jdetter, line 22: I can confirm that this file hasn't been working since we changed the local port to 3000 which was quite a long time ago.

docker-compose-live.yml
```diff
@@ -1,31 +0,0 @@
-version: "3.6"
```
jdetter, line 1: This file hasn't been updated in 2 years now, I believe we're not using this file anywhere and I am also confident that anything that could even still be using this file is broken right now anyway.

PR #3456, opened 2025-10-22
Revert "feat: update pgwire to 0.34 and improve how we disable ssl (#3432)"

bfops:
# Description of Changes

This reverts commit d20579555351f9e88f3d92eff35a8b7f50ce6be7, because it was causing some nontrivial downstream breakages. We will re-apply these changes when we're ready with downstream changes.

None

1

# Testing

Existing CI only


jdetter (approved):
LGTM - not tested by me but I have the context on why we're reverting.

PR #3463, opened 2025-10-22
Add typescript quickstart smoketest

coolreader18:
# Description of Changes

Based on #3461, which fixes a bug encountered in the quickstart.

n/a

1

# Testing

- [x] Yes
- [ ]

### DIFF
--- a/smoketests/tests/quickstart.py
+++ b/smoketests/tests/quickstart.py
@@ -6,7 +6,7 @@ import tempfile
 import xmltodict
 
 import smoketests
-from .. import Smoketest, STDB_DIR, run_cmd, TEMPLATE_CARGO_TOML, TYPESCRIPT_BINDINGS_PATH, build_typescript_sdk
+from .. import Smoketest, STDB_DIR, run_cmd, TEMPLATE_CARGO_TOML, TYPESCRIPT_BINDINGS_PATH, build_typescript_sdk, pnpm
 
 
 def _write_file(path: Path, content: str):
@@ -342,7 +342,7 @@ class TypeScript(Rust):
 
     def server_postprocess(self, server_path: Path):
         build_typescript_sdk()
-        run_cmd("pnpm", "install", f"spacetimedb@file:{TYPESCRIPT_BINDINGS_PATH}", cwd=server_path)
+        pnpm("install", TYPESCRIPT_BINDINGS_PATH, cwd=server_path)
 
     def test_quickstart(self):
         """Run the TypeScript quickstart guides for server."""
--- a/smoketests/__init__.py
+++ b/smoketests/__init__.py
@@ -173,9 +173,19 @@ def run_cmd(*args, capture_stderr=True, check=True, full_output=False, cmd_name=
     return output if full_output else output.stdout
 
 @functools.cache
+def pnpm_path():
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -44,6 +44,20 @@ jobs:
...
+        with:
+          node-version: 18
+
+      - uses: pnpm/action-setup@v4
+        with:
+          run_install: true
+
+      - name: Build typescript module sdk
+        working-directory: crates/bindings-typescript
+        run: pnpm build
```
gefjon: Can you add a comment that these are required to make the TypeScript quickstart smoketest work?

smoketests/tests/quickstart.py
```diff
@@ -18,12 +19,12 @@ def _append_to_file(path: Path, content: str):
         f.write(content)
 
 
-def _parse_quickstart(doc_path: Path, language: str) -> str:
+def _parse_quickstart(doc_path: Path, language: str, extra_langs: list[str] = None) -> str:
     """Extract code blocks from `quickstart.md` docs.
     This will replicate the steps in the quickstart guide, so if it fails the quickstart guide is broken.
     """
```
gefjon, line 25: Please expand this doc comment describing what `extra_langs` does.

smoketests/tests/quickstart.py
```diff
@@ -18,12 +19,12 @@ def _append_to_file(path: Path, content: str):
         f.write(content)
 
 
-def _parse_quickstart(doc_path: Path, language: str) -> str:
+def _parse_quickstart(doc_path: Path, language: str, extra_langs: list[str] = None) -> str:
```
gefjon: I think this type declaration is wrong. I do not believe `None` is a `list` in Python.

smoketests/tests/quickstart.py
```diff
@@ -257,3 +260,32 @@ def test_quickstart(self):
         if not smoketests.HAVE_DOTNET:
             self.skipTest("C# SDK requires .NET to be installed.")
         self._test_quickstart()
+
+class TypeScript(BaseQuickstart):
```
gefjon: Please add a docstring describing that we use the Rust client and the TypeScript server here because the TypeScript client is React-ey and difficult to run headless.

smoketests/tests/quickstart.py
coolreader18: hmm, true

smoketests/tests/quickstart.py
coolreader18: I renamed the parameters to make it more clear what they're for.

gefjon (approved):

.github/workflows/ci.yml
```diff
@@ -21,7 +21,7 @@ jobs:
         include:
           - { runner: spacetimedb-runner, smoketest_args: --docker }
           - { runner: windows-latest, smoketest_args: --no-build-cli }
-        runner: [ spacetimedb-runner, windows-latest ]
+        runner: [ ubuntu-latest, windows-latest ]
```
bfops: is this change required? The PR won't merge as-is because the old version is still a required check.

Also, if we're making this change, it should be changed on Line 22 above as well.

PR #3464, opened 2025-10-22
Remove `connection_id` from ViewContext

joshua-spacetime:
# Description of Changes

Removing `connection_id` from `ViewContext` for performance reasons. We intend to wait to hear from clients before assuming that it's necessary to include.

None (unreleased feature)

0

# Testing

n/a


cloutiertyler:
> must track the view's result set for each unique connection.

SQL connections only? Or all connections?

joshua-spacetime:
Specifically for each subscription connection pertaining to a particular view.

PR #3466, opened 2025-10-23
CI - C# quickstart test properly uses nuget

bfops:
# Description of Changes

The quickstart smoketest was not correctly fudging our NuGet dependencies to use the local versions of the packages, so it was pulling them from NuGet. This ended up causing issues when we tried to modify local packages and then use them in a way that affected the quickstart (e.g. https://github.com/clockworklabs/SpacetimeDB/pull/3386).

We had a few issues: we weren't ensuring that the local packages were built, we weren't using the right directory for those packages, and we weren't adding a "package source mapping" that forced those packages to be used from the local directory.

None. CI-only.

2

# Testing

- [x] CI passes when https://github.com/clockworklabs/SpacetimeDB/pull/3386 is merged together with this PR (it wasn't before).

### DIFF
--- a/smoketests/tests/quickstart.py
+++ b/smoketests/tests/quickstart.py
@@ -299,7 +299,7 @@ Main();
             source_dir=(STDB_DIR / "sdks/csharp").absolute(),
             build_subdir="bin~/Release"
         )
-        run_cmd("dotnet", "add", "package", "SpacetimeDB.ClientSDK", cwd=project_path, capture_stderr=True)
+        run_cmd("dotnet", "add", "package", "SpacetimeDB.ClientSDK", cwd=path, capture_stderr=True)
 
     def server_postprocess(self, server_path: Path):
         override_nuget_package(


smoketests/tests/quickstart.py
```diff
@@ -246,11 +287,27 @@ def project_init(self, path: Path):
         run_cmd("dotnet", "new", "console", "--name", "QuickstartChatClient", "--output", path, capture_stderr=True)
 
     def sdk_setup(self, path: Path):
-        _dotnet_add_package(path, "SpacetimeDB.ClientSDK", (STDB_DIR / "sdks/csharp").absolute())
+        override_nuget_package(
+            project_dir=STDB_DIR/"sdks/csharp",
+            package="SpacetimeDB.BSATN.Runtime",
+            source_dir=(STDB_DIR / "crates/bindings-csharp/BSATN.Runtime").absolute(),
+            build_subdir="bin/Release"
+        )
```
bfops, line 295: this new package override isn't necessary to fix @rekhoff's issue, but I believe it is generally required for correctness if someone makes changes in that package

rekhoff (approved):
Looks good. I like the reuse of the `override_nuget_package` for each of the nuget packages that need to get imported. And the sequence that it does makes sense too. Approved.

PR #3467, opened 2025-10-23
Make `release-fast`use fat lto

kistz:
# Description of Changes

Randomly spotted that the `release-fast` profile in the Cargo.toml is not actually different from `release`.
I assume yall wanted to mark the lto as fat for it.

Alternatively one could remove the release-fast profile.

0

0

# Testing

0

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -68,7 +68,7 @@ rpath = false
 
 [profile.release-fast]
 inherits = "release"
-lto = "thin"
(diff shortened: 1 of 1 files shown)


bfops:
Thank you for opening this! We've discussed internally and decided just to remove this profile in https://github.com/clockworklabs/SpacetimeDB/pull/3473, so I'm going to close this PR.

PR #3468, opened 2025-10-23
`util::jobs`: add commentary + use `.expect(..)`

Centril:
# Description of Changes

1. Add more commentary to `jobs.rs`
2. Use `.expect(...)` more in `jobs.rs`

None

1

# Testing

Covered by existing tests.


gefjon:
Added to the queue because I was confused about what the the GitHub UI showing "Merge when ready" as the button even though all required status checks had passed.

PR #3469, opened 2025-10-23
New command: `spacetime dev`

drogus:
# Description of Changes

This PR implements a new command for the CLI: `spacetime dev`. If run from outside of a project directory, it will call `init` in interactive mode to allow user to create a project. If run from a project directory, it will ask for a module to conect to and it will:

* subscribe to logs
* observe files in the `spacetimedb` directory and automatically publish to SpacetimeDB

One caveat is that instead of reusing the `logs` code I did some repetition cause currently the logs code acquires a lock that lives through await points. It is fixable, but due to limited time I decided to go with a bit of code duplication. It shouldn't be very hard to fix later, though.


crates/cli/src/config.rs
```diff
@@ -169,8 +169,8 @@ impl RawConfig {
             ecdsa_public_key: None,
         };
         RawConfig {
-            default_server: local.nickname.clone(),
-            server_configs: vec![local, maincloud],
```
cloutiertyler, line 173: I would like to get @drogus and @jdetter's thoughts on the above.

cloutiertyler (approved):
This now looks good to me. I've reworked some of the logic to handle some of the more subtle cases especially relating to the server name. One change I've made is that I've made it so that if you don't specify `--server`  it uses your default server (not `maincloud`), but I've also made the change to make `maincloud` the default server. This was one of three options:

- local by default, and `spacetime dev` doesn't override your default (leads people to use `spacetime dev` with localhost most of the time)
- local by default, and `spacetime dev` overrides your default and sends you to maincloud anyway (could be very surprising and could even whoopsie someone's database accidentally)
- maincloud by default, and `spacetime dev` doesn't override your default (just uses the default `maincloud`), seems like the best case

We still have one important thing to do for dev after this merges as well:

- Implement `spacetime.toml` or some such other config file so that users can specify a `database_name` in some locally not checked in file, so that they don't end up creating a million databases if they just run `spacetime dev` a bunch of times against maincloud. I think we'll need to involve @aasoni in that one because it's relevant to projects and teams and stuff.

crates/cli/src/config.rs
jdetter: I originally had 2 issues with this that I think have both been resolved:
1. We don't really give users an indication other than the normal publish output that they're publishing to maincloud - this isn't a huge issue for reasons discussed during the meeting
2. Currently our quickstarts and tutorials make the assumption that your default server is the local server. I fixed this by just adding `--server local` to all tutorials and quickstarts so that they won't break due to this change.

My only remaining feedback is that *technically* this is sort of an API break because anyone who upgrades their CLI may now be pushing to maincloud instead of locally by default which may cause issues especially with external CI jobs that may run testsuites/smoketests/etc.

cloutiertyler: The last part is actually not true. Upgrading your CLI will not change your default server.

PR #3471, opened 2025-10-23
Clarified TableCache onUpdate as optional

PastelStoic:
The onUpdate method only exists on tables with a primary key; auto-generated types specify the existance of this method, but the class itself does not, a tripping hazard for generic implementations.

# Description of Changes

Mark a method as optional.

None. 

Complexity 1. 

# Testing

This makes no changes to the compiled JS code, and every instance of this method already treats it as optional. No testing should be needed.

### DIFF
--- a/crates/bindings-typescript/src/sdk/table_cache.ts
+++ b/crates/bindings-typescript/src/sdk/table_cache.ts
@@ -293,7 +293,7 @@ export class TableCache<
    *
    * @param cb Callback to be called when a new row is inserted
    */
-  onUpdate = <EventContext>(
+  onUpdate? = <EventContext>(
     cb: (ctx: EventContext, oldRow: RowType, row: RowType) => void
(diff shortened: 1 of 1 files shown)


cloutiertyler:
@PastelStoic can you please provide more information on what issue this is causing? That is meant to be an internal API used by generated bindings.

Also the existing tests are failing due to this change.

PastelStoic:
I ran into this issue when making a Preact version of the useTable React hook. Since TableCache.onUpdate is not typed as optional, I got the obvious "onUpdate is not a function" runtime error. I expected this would be a very minor change to types with no actual code effect, so there's clearly something I'm missing in the broader ecosystem. I'll close this, then reopen when I can properly set up tests. I do think the change is worth making for future expandability, allowing third-party utilities for other client frameworks (Svelte, Solid, etc).

PR #3476, opened 2025-10-23
Implemented type-safe primary keys

### LINKED ISSUE #3472: Typesafe primary keys
As it stands, we choose a type for primary keys from a closed set of possibilities, e.g. `u32`. This makes foreign keys non-typesafe.

```
#[table(name = foo, public)]
struct Foo {
  #[primary_key],
  #[auto_inc],
  pub id: u32,
}

#[table(name = bar, public)]
struct Bar {
  pub foo_id: u32;
}
```

This prevents the compiler from preventing mistakes where a `u32` that did not come from a `Foo.id` is used as a value for `Bar.foo_id`. I did try to manually implement something, but it seems some of the traits needed to make this work aren't accessible to mortals. My proposal is something like this:

```
#[table(name = foo, public)]
struct Foo {
  #[primary_key(key_type=u32],
  #[auto_inc],
  pub id: Foo::ID,
}

#[table(name = bar, public)]
struct Bar {
  #[derive(Default)]
  pub foo_id: Foo::ID;
}
```

This would generate:

```
mod Foo {
struct ID(u32);

... glue to box/unbox u32 through the SpacetimeType, Filterable, and other required machinery
}
```

The semantics would be that if a key_type is provided to primary_key, it must be a single-member tuple struct where the contained value is itself valid as a primary key type. The Default value would always be the "not set" flag value, to be replaced in an auto_inc field. The wrapping ID type would not appear in the wire protocol and probably not in the schema, but it may be glued back in for the client facades.

I hope I didn't fumble the rust snippets - I haven't run this through an IDE first.

egormanga:
# Description of Changes

* _Note: based on (and includes) #2957._

For a table like this:
```rust
#[table]
struct User {
	#[primary_key]
	id: Identity
}
```
generates a type like that:
```rust
#[allow(dead_code)]
#[derive(SpacetimeType)]
struct UserId(Identity);

impl std::ops::Deref for UserId {
	type Target = Identity;
	fn deref(&self) -> &Identity {
		&self.0
	}
}
```

~Yet to make it treat the type as `Identity`. Perhaps a trait is needed instead of a newtype.~

### Usage:
```rust
#[table(name=test)]
struct Test {
	user_id: UserId,
}
```

* Resolves #3472.

None.

### 3?

# Testing

- [x] Module compiles
- [x] Module publishes

### DIFF
--- a/crates/bindings-macro/src/table.rs
+++ b/crates/bindings-macro/src/table.rs
@@ -996,8 +996,9 @@ pub(crate) fn table_impl(mut args: TableArgs, item: &syn::DeriveInput) -> syn::R
         #[allow(dead_code)]
         #[derive(SpacetimeType)]
         struct #pk_type_ident(#pk_type);
-        impl std::borrow::Borrow<#pk_type> for #pk_type_ident {
-            fn borrow(&self) -> &#pk_type {
+        impl std::ops::Deref for #pk_type_ident {
+            type Target = #pk_type;
+            fn deref(&self) -> &#pk_type {
                 &self.0
             }
         }


kistz:
Have you considered the `Deref` instead of `Borrow` since then it would treat the UserId as an Identity :>
If you already considered it maybe add to the pr description why you chose Borrow instead?

egormanga:
@kistz Thank you for the suggestion! I just haven't thought about using it.

<details>
<summary>Outdated</summary>

Still, with `impl Deref` I'm getting the same result:
```rust
error[E0223]: ambiguous associated type
  --> src/lib.rs:10:8
   |
10 |     user_id: User::ID,
   |           ^^^^^^^^
   |
help: if there were a trait named `Example` with associated type `ID` implemented for `User`, you could use the fully-qualified path
   |
10 -     user_id: User::ID,
10 +     user_id: <User as Example>::ID,
   |
```
</details>

Wait, I've forgotten my own way of implementing this! One has to write `UserId` — not `User::ID`.

This actually works! I'll push the results in a minute and update the PR description with usage example (should've done that in the first place).

egormanga:
There's still a slight problem with the attribute type:
```rust
let user = ctx.db.user().insert(User {
	id: ctx.identity(),
});

ctx.db.test().insert(Test {
	user_id: user.id,
});
```
gives:
```rust
error[E0308]: mismatched types
  --> src/lib.rs:21:9
   |
21 |         user_id: user.id,
   |                  ^^^^^^^ expected `UserId`, found `Identity`
   |
help: try wrapping the expression in `UserId`
   |
21 |         user_id: UserId(user.id),
   |                  +++++++       +
```

kistz:
> There's still a slight problem with the attribute type

I dont think you can do better than that for Rusts type system. 

IMO the point for this is that you need to make a concious decision when construction new UserId types (and not just transfering them from places where its already a UserId) hence you shoud be required to do `UserId(ctx.identity())` for example.

I could also imagine to implement `From` for the type automatically that you could do: ` User { id: ctx.identity().into() }` but im not sure if this is worth since its harder to read than explicitly constructing imo :> (and the user could always choose to implement it themselves because `UserId` would be defined in your crate.

Amazing progress on this!

egormanga:
> implement `From` for the type automatically

Probably not the best idea, since explicit `UserId` is much better than `.into()` (which kinda loses the point).

Alrighty, I can see this PR viable in the current state.

kistz:
nice looks good!
(there is still the borrow variant in the pr description maybe update that 😇)

cloutiertyler:
I think we need to have a think about this one in the context of other languages. As written this would make Rust types inconsistent with the types in the other languages. Much like in #2957 we'd need to address this for all languages.

PR #3480, opened 2025-10-24
Make `name` a required parameter of `view` macro

joshua-spacetime:
# Description of Changes

Makes `name` a required parameter of the `#[view]` macro. This is to be consistent with the `#[table]` macro since views should expose the same interface as tables.

None

1

# Testing

- [x] Added new compile tests
- [x] Updated existing compile tests


coolreader18 (approved):
LGTM!

PR #3482, opened 2025-10-24
Add system table for tracking view subscribers

joshua-spacetime:
# Description of Changes

Adds two new system tables for views. One tracks view arguments, the other tracks view subscribers.

`st_view_arg` generates a unique id for each unique argument instantiation. This id will act as a foreign key that both read sets and backing tables can reference.

`st_view_client` is needed so that we can drop views when clients unsubscribe or disconnect. Note this disconnect logic is not implemented in this patch.

Eventually both of these tables should be ephemeral. There's no reason they need to write to the commitlog.

Note also that the schema of a view's backing table has been updated. It no longer stores a view's argument values but rather the `id` from `st_view_arg`.

None

1

# Testing

Tests will be added in the patch that computes read sets


crates/datastore/src/system_tables.rs
```diff
@@ -863,6 +896,21 @@ pub struct StViewParamRow {
...
+/// | view_id | identity | connection_id | product_arg_hash | product_arg_bytes |
+/// |---------|----------|---------------|------------------|-------------------|
+/// | 1       | 0x...    | 0x...         | 0x...            | <bytes>           |
+#[derive(Debug, Clone, Eq, PartialEq, SpacetimeType)]
+#[sats(crate = spacetimedb_lib)]
+pub struct StViewClientRow {
+    pub view_id: ViewId,
+    pub identity: IdentityViaU256,
+    pub connection_id: ConnectionIdViaU128,
+    pub product_arg_hash: Hash,
```
joshua-spacetime: A 32-byte hash of the arguments seems overkill. We could get away with much less or perhaps use an auto-inc integer, but we would then need a separate table for storing these argument keys.

crates/datastore/src/system_tables.rs
```diff
@@ -863,6 +896,21 @@ pub struct StViewParamRow {
...
+///
+/// | view_id | identity | connection_id | product_arg_hash | product_arg_bytes |
+/// |---------|----------|---------------|------------------|-------------------|
+/// | 1       | 0x...    | 0x...         | 0x...            | <bytes>           |
+#[derive(Debug, Clone, Eq, PartialEq, SpacetimeType)]
+#[sats(crate = spacetimedb_lib)]
+pub struct StViewClientRow {
+    pub view_id: ViewId,
+    pub identity: IdentityViaU256,
+    pub connection_id: ConnectionIdViaU128,
```
Shubham8287, line 937: What purpose does `connection_id` solve?

crates/schema/src/schema.rs
```diff
@@ -607,15 +607,17 @@ impl TableSchema {
     ///
     /// my_view:
     ///
-    /// | sender   | x   | y   | a   | b   |
-    /// |----------|-----|-----|-----|-----|
-    /// | Identity | u32 | u32 | u32 | u32 |
+    /// | sender   | product_arg_hash | a   | b   |
```
Shubham8287: This makes sense, as mostly arg params will be projected together.

crates/datastore/src/system_tables.rs
joshua-spacetime: The same identity can subscribe using multiple connections.

crates/schema/src/schema.rs
```diff
@@ -646,24 +646,22 @@ impl TableSchema {
...
         }
 
-        let n = columns.len();
-
-        let param_iter = param_columns
-            .iter()
-            .map(|def| ColumnSchema::from_view_param_def(module_def, def));
+        columns.push(ColumnSchema {
+            table_id: TableId::SENTINEL,
+            col_pos: ColId(1),
```
Shubham8287:
```suggestion
            col_pos: ColId((if *is_anonymous { 0 } else { 1 })),
```

PR #3484, opened 2025-10-25
Auto-migration rules for views

joshua-spacetime:
# Description of Changes

Defines auto-migration rules for views.

> What are they?

Answer: You can always auto-migrate a view. It's just a matter of whether you have to disconnect clients. Removal of a view or changing the return type in a breaking way requires that we disconnect clients.

TODO: Even if a view is updated in a completely compatible way, we still have to wipe or dirty the read set in order to force a re-computation because the behavior could have changed.

None

2.5

Not just a mechanical change.

Most of the time you can think of views as though they were tables. However for auto-migration there is a key distinction. Views don't have observably persistent state. While we can persist state internally, we can always derive a view's result set from the database state. Hence auto-migration rules are not as strict.

# Testing

- [x] Auto-migration planner tests

### DIFF
--- a/crates/core/src/db/relational_db.rs
+++ b/crates/core/src/db/relational_db.rs
@@ -1055,13 +1055,30 @@ impl RelationalDB {
         Ok(self.inner.create_table_mut_tx(tx, schema)?)
     }
 
-    pub fn create_view_table(
+    pub fn drop_table(&self, tx: &mut MutTx, table_id: TableId) -> Result<(), DBError> {
+        let table_name = self
+            .table_name_from_id_mut(tx, table_id)?
+            .map(|name| name.to_string())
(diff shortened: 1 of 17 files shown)


crates/core/src/db/relational_db.rs
```diff
@@ -1055,13 +1055,30 @@ impl RelationalDB {
         Ok(self.inner.create_table_mut_tx(tx, schema)?)
     }
 
-    pub fn create_view_table(
+    pub fn drop_table(&self, tx: &mut MutTx, table_id: TableId) -> Result<(), DBError> {
```
joshua-spacetime, line 1058: Code motion: Just moved `drop_table` to be near `create_table`

PR #3491, opened 2025-10-25
Reenabled github page edit link

cloutiertyler:
# Description of Changes

The GitHub edit links were disabled in the new docs, this reenables them.

None

1

# Testing

- [x] Tested locally that this works.


docs/package.json
```diff
@@ -16,6 +16,7 @@
   },
   "dependencies": {
     "@docusaurus/core": "3.9.1",
+    "@docusaurus/plugin-content-docs": "^3.9.2",
```
JulienLavocat, line 19:
```suggestion
    "@docusaurus/plugin-content-docs": "^3.9.1",
```
Why did bump this version?

cloutiertyler: Because otherwise two copies of it was being bundled and you end up with this error:
`useDocsVersion is called outside the <DocsVersionProvider>`

 See this conversation: https://chatgpt.com/share/68ff76cf-1974-800d-a120-7b548a7c99ee

PR #3493, opened 2025-10-26
Don't depend on `[!Default; 0]: Default` impl

WaffleLapkin:
# Description of Changes

There are plans to remove the unconditional `[T; 0]: Default` impl (where `T` doesn't necessarily implement `Default`) from the standard library (see https://github.com/rust-lang/rust/pull/145457). That is a breaking change that would affect this crate.

This PR fixes this by removing a dependency on that impl.

N/A

1, there is no risk, the change is trivial.

# Testing

No testing is required, `<[T; 0]>::default()` is always equivalent to `[]`.

### DIFF
--- a/crates/sats/src/array_value.rs
+++ b/crates/sats/src/array_value.rs
@@ -141,7 +141,7 @@ impl ArrayValue {
 impl Default for ArrayValue {
     /// The default `ArrayValue` is an empty array of sum values.
     fn default() -> Self {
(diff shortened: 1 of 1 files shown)


WaffleLapkin:
I will not sign the CLA, however this contribution is trivial enough that you can probably figure this out on your end ^^'

bfops:
Thank you for making us aware of this! Since you mentioned you're not signing the CLA, I'll close this PR.

PR #3494, opened 2025-10-27
First pass at reorganizing the docs and making them nice

cloutiertyler:
# Description of Changes

This PR is the first in a series of PRs which are an effort the reorganize the docs and make it easier for people to get started using SpacetimeDB.

This PR:

1. Renames all folders to be `kebab-case`. This is because `docIds` are set based on folder and file names, and there is no way to override that within a file (all you can change with `id: my-id` is the last component of the id. Because we want to be able to access docs in react components we need stable predictable ids.
    - This also necessitates setting up a few `_category_.json` files so that the categories do not show up as kebab case in the sidebar.
2. Adds a few React components with images so it's clearer on how to get started
3. Reorganizes the quickstarts to be at the top of the side bar and simplifies the server and client into a single document named for the language
4. Sets up redirections so the old links to the old documents now point to the new documents
5. Copies the entirety of `typography.css` from the website into this repo for the purpose of keeping them identical
6. Makes minor fixes and improvements to the contents of the docs

In a future PR, I intend to:

1. Add a tutorial section to the docs which will include two tutorials: Blackholio and Chat
2. Turn the chat quickstart documents into the Chat tutorial and consolidate it into a single file with different language switchers (if this is feasible)
3. Create a new quickstart section, which has a quickstart guide for each platform that we support so far, namely:
    - Vanilla TypeScript
    - Vanilla JavaScript
    - Vanilla Rust
    - Vanilla C#
    - React
    - Unity
    - Unreal
 4. Update the quickstart guides to do the following (in the future I will add more platforms with this scheme):
    - Use `spacetime dev --template <framework-template>` to create a new project for the given framework
    - Insert data into the database
    - Add a new `reducer` to the template
    - Add a new `table` to the template
    - Add a new `view` to the template
    - Call the reducer from the client
    - Subscribe to the `view` from the template
 5. Break up the current per language module reference files into files about SpacetimeDB concepts with a `TabGroup` for each language, namely:
    - Modules
    - Tables
    - Reducers
    - Procedures
    - Code generation
    - DbConnection (and the SDK API)
    - Auth
    - etc.
The idea being that we should explain our platform by concept, not by language. This will also allow us to reuse a lot of prose that is repeated in the different language documents.

None

1

# Testing

- [x] Tested the documents locally to ensure that they render properly

### DIFF
--- a/docs/docs/02-quickstarts/02-c-sharp.md
+++ b/docs/docs/02-quickstarts/02-c-sharp.md
@@ -84,7 +84,7 @@ For each `User`, we'll store their `Identity`, an optional name they can set to 
 
 In `spacetimedb/Lib.cs`, add the definition of the table `User` to the `Module` class:
 
-```csharp
+```csharp server
 [Table(Name = "user", Public = true)]
 public partial class User
 {
@@ -143,7 +143,7 @@ For now, we'll just do a bare minimum of validation, rejecting the empty name. Y
 
 In `spacetimedb/Lib.cs`, add to the `Module` class:
 
-```csharp
+```csharp server
 /// Takes a name and checks if it's acceptable as a user's name.
 private static string ValidateName(string name)
 {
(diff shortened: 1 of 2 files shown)


JulienLavocat (approved):
LGTM

PR #3501, opened 2025-10-27
CI - Cancel runs on new pushes

bfops:
# Description of Changes

Add `cancel-in-progress` to our GitHub workflows.

None

1

# Testing

- [x] Pushing new commits to this PR causes cancels of previous CI runs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,9 +14,8 @@ on:
 name: CI
 
 concurrency:
-  # Use the PR number when available; otherwise a unique run_id so non-PR
-  # events never share a group.
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_id }}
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -13,6 +13,13 @@ on:
 
 name: CI
 
+concurrency:
+  # Use the PR number when available; otherwise a unique run_id so non-PR
+  # events never share a group.
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_id }}
```
jdetter: Would github PR numbers ever overlap with run ids in a way where it would cause a conflict? If a run does conflict with a PR number does it matter?

bfops: it's technically possible, although run IDs are much much larger than our current PR numbers.

Either way, it's unnecessary. Updated for clarity.

PR #3504, opened 2025-10-27
Fix to Unreal tests to stop requesting infinite resources

JasonAtClockwork:
# Description of Changes

For core dumps Unreal tries to use setrlimit() for infinity when setting resource limits, it seems this becomes an issue intermittently when running tests in parallel. There is an option when running Unreal to disable this and we can rely on error longs only which makes sense for CI imo.

None

1 - small flag change

# Testing

- [x] Re-ran all tests locally


bfops:
Does that mean that we only see the `setrlimit` error when something else has already gone wrong?

JasonAtClockwork:
> Does that mean that we only see the `setrlimit` error when something else has already gone wrong?

This is actually before anything has run in Unreal, on the Linux initialization it runs a IncreasePerProcessLimits() before allowing the internal GuardedMain() from processing. It does mean we won't get Unreal core dumps if something goes wrong in CI but if it goes that badly wrong with a test we should see it in a lower environment first. It will still output all errors that happened which is 90% of the time enough to resolve any troubles.

bfops (approved):
Okay, SGTM!

PR #2356, opened 2025-03-06
`spacetime-install.ps1` installs Visual C++ Redistributable if needed

iso8859:
Add VCRuntime detection and installation. This runtime is mandatory.

# Description of Changes
Update the Windows powershell script to install VCRuntime.

# Expected complexity level and risk
Not complex.
Download directly from Microsoft.

# Testing
Tested on a clean Windows 10 machine.

PR #3505, opened 2025-10-27
Improve spacetime-install.ps1

jdetter:
# Description of Changes

This is a follow-up to https://github.com/clockworklabs/SpacetimeDB/pull/2356 . Essentially we want to remove this `return` statement so that you can both install the vcruntime and also continue installing SpacetimeDB without having to run the installer twice on Windows.

None

1 - this affects windows installation.

# Testing

I tested this on a new Windows 11 install. You can test the script by downloading it from this PR and then running the following:
```
iex (Get-Content .\spacetime-install.ps1 -Raw -Encoding UTF8)
```

- [x] I have tested this

This is from a brand-new Windows11 install with all updates installed:
image

Then I opened a new terminal window:
image

### DIFF
--- a/crates/update/spacetime-install.ps1
+++ b/crates/update/spacetime-install.ps1
@@ -28,7 +28,6 @@ function Install {
         $Installer = Join-Path ([System.IO.Path]::GetTempPath()) "vc_redist.x64.exe"
         Invoke-WebRequest $DownloadUrl -OutFile $Installer -UseBasicParsing
         Start-Process -Wait -FilePath $Installer -ArgumentList "/quiet", "/install"
-        return
     }
 
     $DownloadUrl = "https://github.com/clockworklabs/SpacetimeDB/releases/latest/download/spacetimedb-update-x86_64-pc-windows-msvc.exe"


crates/update/spacetime-install.ps1
```diff
@@ -28,7 +28,6 @@ function Install {
         $Installer = Join-Path ([System.IO.Path]::GetTempPath()) "vc_redist.x64.exe"
         Invoke-WebRequest $DownloadUrl -OutFile $Installer -UseBasicParsing
         Start-Process -Wait -FilePath $Installer -ArgumentList "/quiet", "/install"
-        return
```
jdetter, line 31: Now instead of returning we just continue on to the spacetime CLI installation.

rekhoff (approved):
Reviewed code, removal of the return makes sense, as the `Start-Process -Wait` should wait for the install to finish before moving on, ensuring the later installs have `vcruntime140.dll` when it's needed.
Ran this locally on a non-new Windows 11 install and it worked as expected. Given the testing already performed, this is approved.

PR #3520, opened 2025-10-28
Compute read sets for MutTxId via InstanceEnv

joshua-spacetime:
# Description of Changes

The `InstanceEnv` will now compute read sets when executing a view and store them in the `MutTxId`.

These read sets track table scans as well as singular index key scans. Index key ranges will be tracked in the future, but for now an index range scan is treated as a full table scan.

These read sets are maintained as part of the `CommittedState`.

TODO: Check write sets against read sets. Re-evaluate views in the case of overlap and update the read sets.

None

1

# Testing

- [ ] Unit tests

### DIFF
--- a/crates/core/src/host/instance_env.rs
+++ b/crates/core/src/host/instance_env.rs
@@ -471,7 +471,7 @@ impl InstanceEnv {
         stdb.table_row_count_mut(tx, table_id)
             .ok_or(NodesError::TableNotFound)
             .inspect(|_| {
-                tx.track_table_scan(self.view_id, table_id);
+                tx.record_table_scan(self.view_id, table_id);
             })
     }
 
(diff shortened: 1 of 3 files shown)


crates/core/src/host/instance_env.rs
```diff
@@ -477,6 +489,8 @@ impl InstanceEnv {
             &mut bytes_scanned,
         );
 
+        tx.track_table_scan(self.view_id, table_id);
```
Shubham8287: Just to be sure, we aren't doing tracking in `row_iter_bsatn_*` ABIs as we plan to not support them in initial version?

joshua-spacetime: We are only tracking reads from table scans and index scans as those are the only operations that the context allows. Technically `ViewContext` doesn't even allow table scans at this moment, but may in the future.

PR #3521, opened 2025-10-28
CI - Cache more of our build outputs

bfops:
# Description of Changes

Changes with cache hit:
CI / Smoketests (spacetimedb-runner): 17m -> 10m
CI / Smoketests (windows-latest): 39m -> 31m
CI / Lints: 3m30s -> 2m
unity-testsuite: 23m -> 19m

This included creating a stripped-down `Dockerfile` that only runs SpacetimeDB (as opposed to e.g. building tracy, flamegraphs, etc.). That shaves off about 4-5m on the linux smoketests job.

No. CI only.

2

# Testing

Ran existing CI and noted the change in times in the description above


.github/Dockerfile
```diff
@@ -0,0 +1,6 @@
+# Minimal Dockerfile that just wraps pre-built binaries, so we can test the server inside docker
+FROM rust:1.90.0
```
bfops, line 2: there might be an even smaller container we can use as the base, but this one seems fine. Currently it takes ~39s to set up the docker container, and most of that is installing this base image.

.github/workflows/csharp-test.yml
```diff
@@ -65,26 +67,18 @@ jobs:
...
-          # Cache the CLI as well.
-          cache-all-crates: true
+          workspaces: ${{ github.workspace }}
+          shared-key: spacetimedb
+          # Let the main CI job save the cache since it builds the most things
+          save-if: false
 
       - name: Install SpacetimeDB CLI from the local checkout
-        # Rebuild only if we didn't get a precise cache hit.
-        if: steps.cache-rust-deps.outputs.cache-hit == 'false'
```
bfops, line 84: I don't feel 100% confident in this cache-skipping logic, so I removed it. If we hit the cache, this step will be fast anyway.

.github/docker-compose.yml
```diff
@@ -0,0 +1,15 @@
+services:
+  node:
+    labels:
+      app: spacetimedb
```
bfops, line 4: what I don't love about this is that now we're not testing our "actual" `docker-compose.yml` or `Dockerfile`s. I don't think it's the biggest deal (we actually don't seem to test our root `Dockerfile` anyway), but I created https://github.com/clockworklabs/SpacetimeDBPrivate/issues/2275 to track that.

That said, that might be an argument for not creating these custom dockerfiles. I do appreciate having the smallest possible turnaround time for running tests, though, especially if we are moving this logic to `cargo ci` and expecting users to run it locally.

jdetter: I mean the usage of the "actual" docker file is a bit fuzzy. Originally we used the docker files for doing local dev and the releases but more recently they're being used for the smoketests but then also sometimes the users use them for building from source so idk - thanks for making a follow-up ticket, I think we should probably just be verifying that the container at least builds and runs but anything other than that would just have too much overlap with the new docker files.

I think making CI specific docker files was a good idea in order to get the speedup so I'm happy to approve on that basis 👍

PR #3522, opened 2025-10-28
CI - Skip Internal Tests and Unreal Tests on external PRs

bfops:
# Description of Changes

These tests fail on external PRs, but not for any real reasons - just because GH secrets are missing. "Skipped" is more informative than "failed".

None.

1

# Testing

None, but I just copied the logic from the unity testsuite.


bfops:
> This seems good to me, will this cause issues when we make `Internal Tests` required in the future?

Yes, but currently it _fails_, so that's not any better.

jdetter:
> > This seems good to me, will this cause issues when we make `Internal Tests` required in the future?
> 
> Yes, but currently it _fails_, so that's not any better.

Fair! 😅

PR #3523, opened 2025-10-28
[teams 3/5] API authorization, CLI, smoketests

kim:
This adds authorization to the relevant API endpoints, updates the CLI
commands and adds smoketests for the teams feature.

**Note**: Authorizing SQL (incl. subscriptions) is a bit more involved,
and submitted as a separate PR in the series.

Depends-on: https://github.com/clockworklabs/SpacetimeDB/pull/3519

### DIFF
--- a/crates/cli/src/subcommands/publish.rs
+++ b/crates/cli/src/subcommands/publish.rs
@@ -1,14 +1,15 @@
+use anyhow::{ensure, Context};
 use clap::Arg;
 use clap::ArgAction::{Set, SetTrue};
 use clap::ArgMatches;
 use reqwest::{StatusCode, Url};
 use spacetimedb_client_api_messages::name::{is_identity, parse_database_name, PublishResult};
-use spacetimedb_client_api_messages::name::{PrePublishResult, PrettyPrintStyle, PublishOp};
+use spacetimedb_client_api_messages::name::{DatabaseNameError, PrePublishResult, PrettyPrintStyle, PublishOp};
 use std::path::PathBuf;
 use std::{env, fs};
 
 use crate::config::Config;
-use crate::util::{add_auth_header_opt, get_auth_header, unauth_error_context, AuthHeader, ResponseExt};
+use crate::util::{add_auth_header_opt, get_auth_header, AuthHeader, ResponseExt};
 use crate::util::{decode_identity, y_or_n};
 use crate::{build, common_args};
 
@@ -76,6 +77,17 @@ pub fn cli() -> clap::Command {
             common_args::anonymous()
         )
         .arg(
+            Arg::new("parent")
+            .help("Domain or identity of a parent for this database")
+            .long("parent")
+            .long_help(
+"A valid domain or identity of an existing database that should be the parent of this database.
+
+If a parent is given, the new database inherits the team permissions from the parent.
+A parent can only be set when a database is created, not when it is updated."
+            )
+        )
+        .arg(
             Arg::new("name|identity")
                 .help("A valid domain or identity for this database")
                 .long_help(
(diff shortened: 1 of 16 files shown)


crates/cli/src/subcommands/publish.rs
```diff
@@ -220,18 +236,6 @@ pub async fn exec(mut config: Config, args: &ArgMatches) -> Result<(), anyhow::E
     }
 
     let res = builder.body(program_bytes).send().await?;
-    if res.status() == StatusCode::UNAUTHORIZED && !anon_identity {
```
cloutiertyler, line 223: Can you just comment on why this code is gone? Or where it's moved to? I don't see an `authorize_action` call that's been added here.

crates/client-api/src/routes/database.rs
```diff
@@ -738,14 +731,19 @@ pub async fn publish<S: NodeDelegate + ControlStateDelegate>(
...
         }
         // If yes, authorize via ctx.
         Some(database) => {
-            // NOTE: This is redundant at the moment, but will be replaced by
-            // a more interesting authorization check with the "teams" feature.
-            if database.owner_identity != auth.claims.identity {
-                return Err((StatusCode::UNAUTHORIZED, "a database can only be updated by its owner").into());
-            }
+            ctx.authorize_action(auth.claims.identity, database.database_identity, Action::UpdateDatabase)
+                .await?;
```
cloutiertyler, line 746: I'm assuming the above path I talked about is caught here?

crates/cli/src/subcommands/publish.rs
kim: It was a bit unclear to me what the code is trying to achieve -- it infers from a 401 response and the previously forced login that the user must be authenticated with the wrong server (via `--server-issued-token`, presumably).

However, if the server returns 401 because you're not the owner (or have otherwise insufficient permissions), this is not necessarily true. You could simply not have sufficient permissions.

To make testing the different roles work, I had to remove this.
I don't think it worked correctly before, though, it was just not tested anywhere.

If we think the hint the old code provides is important enough to continue supporting it, we need to make the server respond in such a way that the client can differentiate. E.g. by returning 403 (Forbidden) if the issuer doesn't match, or start returning JSON bodies.

As @gefjon wrote the original code apparently, maybe they want to chime in too.

kim: Well, one might also argue that 401 means "wrong or missing credentials" and 403 means "credentials look ok, but you are still not granted access".

https://datatracker.ietf.org/doc/html/rfc7235#section-3.1
https://datatracker.ietf.org/doc/html/rfc7231#section-6.5.3

We'd need to review our response codes then...

cloutiertyler: Yeah from a Chesterton's Fence point of view, I'd like to understand what the purpose of it was before removing it.

gefjon: IIRC, I added this warning back when I first added the ability to store logins for multiple different servers. At that time, the way the CLI was set up, each server config would remember which identity/token it was supposed to use. If a token was saved in the config for a specific server but was not valid for that server, it represented a legitimate problem which needed correcting. I then went through and mindlessly applied `unauth_error_context` to all `UNAUTHORIZED` returns. This seemed reasonable at the time because, prior to storing token-per-server, the most common by far reason that any CLI call would fail was due to a token issued by a different server.

Since then, we reworked auth, made `--server-issued-token` not be the default, and eliminated the ability for the CLI to store multiple tokens for different servers. The error message in `unauth_error_context` appears to have changed at that time, but the pattern of assuming that `UNAUTHORIZED` responses were generally caused by invalid server-issued tokens did not.

I like the idea of returning more descriptive error messages from the server, rather than trying to guess from the client what the HTTP error code meant.

cloutiertyler: In that case, I have no problem removing this.

cloutiertyler (approved):

crates/cli/src/subcommands/publish.rs
kim: Alright. My inner Roy T. Fielding can't unsee, though, that we're using those status codes wrongly. I'll change the new authorization code to return 403 in case of rejection, and review the remaining 401s in a follow-up.

PR #3524, opened 2025-10-29
Add query engine support for mutable transactions

joshua-spacetime:
# Description of Changes

The query engine can now execute queries using mutable transactions because this patch implements the `Datastore` trait for `MutTxId`.

Note this is both a refactor and a feature patch. It is a refactor because the `Datastore` trait was updated to allow for mutable transactions.

None

1.5

# Testing

- [x] Current tests pass after refactor
- [ ] Use `MutTxId` for query execution


crates/datastore/src/locking_tx_datastore/tx.rs
```diff
@@ -33,12 +32,44 @@ pub struct TxId {
...
+
+    fn row_count(&self, table_id: TableId) -> u64 {
+        self.committed_state_shared_lock
+            .table_row_count(table_id)
+            .unwrap_or_default()
     }
 
-    fn table(&self, table_id: TableId) -> Option<&Table> {
-        self.committed_state_shared_lock.get_table(table_id)
+    fn table_scan<'a>(&'a self, table_id: TableId) -> anyhow::Result<Self::TableIter<'a>> {
```
joshua-spacetime, line 51: Note: No changes were made to the `Datastore` implementation for read only transactions. The `Datastore` trait no longer has default methods, so I've moved the default methods to the `impl` for `TxId`.

crates/execution/src/iter.rs
joshua-spacetime, line 1: I've removed the query engine's tuple-at-a-time iterators since they weren't used and almost every type in this file was going to need to be updated to take several generic parameters.

PR #3525, opened 2025-10-29
[teams 4/5] SQL authorization

kim:
Permissions for evaluating SQL/DML are not generally "actions", but more a set of permissions that are checked during evaluation.

To make this work with the teams feature, this patch extends `AuthCtx` to allow checking a set of permissions as mandated by the spec. This set is a bit more fine-grained than "is owner", so as to avoid baking in the concept of teams/collaborators, or assumptions about what a role might entail. Both are likely to evolve in the future, so evaluation of permissions / capabilities should be confined to the impl of the `Authorization` trait.

Unlike "actions", the `AuthCtx` must be able to evaluate permission checks quickly and without side-effects, nor can it enter an `async` context. In that sense, it is precomputed (if you will), and stored as a closure in the `AuthCtx` for external authorization.

A challenge posed is how to thread through the constructed `AuthCtx` for subscriptions.

A tempting approach would have been to equip the `HostController` with the ability to summon an `AuthCtx`. That, however, would have created a gnarly circular dependency, because the `HostController` also controls the controldb, which itself demands an `AuthCtx`.

Instead, the `AuthCtx` is obtained in the endpoint handler and passed to each method call that requires one. That's less pretty, but more effective.

### DIFF
--- a/crates/core/src/sql/execute.rs
+++ b/crates/core/src/sql/execute.rs
@@ -273,7 +273,7 @@ pub async fn run(
 
             // Update views
             let result = match module {
-                Some(module) => module.call_views_with_tx(tx, auth.caller).await?,
+                Some(module) => module.call_views_with_tx(tx, auth.caller()).await?,
                 None => ViewCallResult::default(tx),
             };
 


crates/core/src/sql/execute.rs
```diff
@@ -249,7 +249,7 @@ pub fn run(
         }
         Statement::DML(stmt) => {
             // An extra layer of auth is required for DML
-            if auth.caller != auth.owner {
+            if !auth.has_write_access() {
                 return Err(anyhow!("Only owners are authorized to run SQL DML statements").into());
```
gefjon:
```suggestion
                return Err(anyhow!("Caller {} is not authorized to run SQL DML statements", auth.caller()).into());
```

PR #3528, opened 2025-10-29
Switch subscriptions to use mutable transactions in preparation for views

joshua-spacetime:
# Description of Changes

Switch subscriptions to use mutable transactions in preparation for views.

The main implication of this patch is that we no longer use rayon to parallelize subscriptions. Past performance tests have shown that we don't get much parallelism out of rayon, however we should still test the performance of this patch extensively.

3

This is mainly due to an ugly diff which is the result of making things suitably generic so as to allow for both mutable and read only subscriptions.

# Testing

This is a refactor with the result being that subscriptions now run in mutable transactions. Most of the subscription tests will now use a `MutTxId` instead of a `TxId`, so the current tests should give us sufficient coverage.

TODO: Performance test

### DIFF
--- a/crates/core/src/estimation.rs
+++ b/crates/core/src/estimation.rs
@@ -1,17 +1,19 @@
-use crate::db::relational_db::Tx;
-use spacetimedb_datastore::locking_tx_datastore::state_view::StateView as _;
+use std::num::NonZeroU64;
+
+use crate::vm::TxMode;
+use spacetimedb_execution::Datastore;
 use spacetimedb_lib::query::Delta;
 use spacetimedb_physical_plan::plan::{HashJoin, IxJoin, IxScan, PhysicalPlan, Sarg, TableScan};
 use spacetimedb_primitives::{ColList, TableId};
 use spacetimedb_vm::expr::{Query, QueryExpr, SourceExpr};
 
 /// The estimated number of rows that a query plan will return.
-pub fn num_rows(tx: &Tx, expr: &QueryExpr) -> u64 {
+pub fn num_rows(tx: &TxMode, expr: &QueryExpr) -> u64 {
     row_est(tx, &expr.source, &expr.query)
 }
 
 /// Use cardinality estimates to predict the total number of rows scanned by a query
-pub fn estimate_rows_scanned(tx: &Tx, plan: &PhysicalPlan) -> u64 {
+pub fn estimate_rows_scanned(tx: &TxMode, plan: &PhysicalPlan) -> u64 {
     match plan {
         PhysicalPlan::TableScan(..) | PhysicalPlan::IxScan(..) => row_estimate(tx, plan),
         PhysicalPlan::Filter(input, _) => estimate_rows_scanned(tx, input).saturating_add(row_estimate(tx, input)),
@@ -46,7 +48,7 @@ pub fn estimate_rows_scanned(tx: &Tx, plan: &PhysicalPlan) -> u64 {
 }
 
 /// Estimate the cardinality of a physical plan
-pub fn row_estimate(tx: &Tx, plan: &PhysicalPlan) -> u64 {
+pub fn row_estimate(tx: &TxMode, plan: &PhysicalPlan) -> u64 {
     match plan {
         // Use a row limit as the estimate if present
         PhysicalPlan::TableScan(TableScan { limit: Some(n), .. }, _)
@@ -59,7 +61,7 @@ pub fn row_estimate(tx: &Tx, plan: &PhysicalPlan) -> u64 {
                 delta: None,
             },
             _,
-        ) => tx.table_row_count(schema.table_id).unwrap_or_default(),
+        ) => table_row_count(tx, schema.table_id),
         // We don't estimate the cardinality of delta scans currently
         PhysicalPlan::TableScan(
             TableScan {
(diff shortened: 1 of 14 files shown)


crates/core/src/subscription/mod.rs
```diff
@@ -169,8 +168,8 @@ where
     F: BuildableWebsocketFormat,
 {
     plans
-        .par_iter()
-        .flat_map_iter(|plan| plan.plans_fragments().map(|fragment| (plan.sql(), fragment)))
+        .iter()
+        .flat_map(|plan| plan.plans_fragments().map(|fragment| (plan.sql(), fragment)))
```
joshua-spacetime, line 172: No longer dispatching to rayon for initial subscription evaluation. Note, we already dispatched from a database thread, not a tokio thread, so cache locality shouldn't be any worse.

crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -230,21 +236,11 @@ impl ModuleSubscriptions {
...
-            &self.relational_db,
-            tx,
-            |plan, tx| {
-                plan.plans_fragments()
-                    .map(|plan_fragment| estimate_rows_scanned(tx, plan_fragment.optimized_physical_plan()))
-                    .fold(0, |acc, rows_scanned| acc.saturating_add(rows_scanned))
-            },
-            auth,
-        )?;
+        self.check_row_limit(std::slice::from_ref(&query), tx, auth)?;
```
joshua-spacetime, line 243: No changes. Just encapsulated in a helper.

crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -1005,6 +1043,119 @@ impl GuardTxOptions {
 
 pub struct WriteConflict;
 
+type TxGuard<F> = Scoped<TxId, F>;
+type MutTxGuard<F> = Scoped<MutTxId, F>;
```
joshua-spacetime, line 1047: In order to be able to support both mutable and read only transactions in subscriptions, we implement the `Datastore` and `StateView` traits for each of these types below.

crates/core/src/estimation.rs
```diff
@@ -1,17 +1,19 @@
...
 
 /// The estimated number of rows that a query plan will return.
-pub fn num_rows(tx: &Tx, expr: &QueryExpr) -> u64 {
+pub fn num_rows(tx: &TxMode, expr: &QueryExpr) -> u64 {
     row_est(tx, &expr.source, &expr.query)
 }
 
 /// Use cardinality estimates to predict the total number of rows scanned by a query
-pub fn estimate_rows_scanned(tx: &Tx, plan: &PhysicalPlan) -> u64 {
+pub fn estimate_rows_scanned(tx: &TxMode, plan: &PhysicalPlan) -> u64 {
```
joshua-spacetime, line 16: Make cardinality estimation work for both `TxId` and `MutTxId`

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -2587,3 +2588,24 @@ fn unindexed_iter_by_col_range_warn(
...
+    /// - No such table as `table_id` exists.
+    /// - The table `table_id` does not have an index on exactly the `cols`.
+    /// - The table `table_id` contains zero rows (i.e. the index is empty).
+    //
+    // This method must never return 0, as it's used as the divisor in quotients.
+    // Do not change its return type to a bare `u64`.
+    //
+    // Note, because this method is used for the purpose of estimation,
+    // we only consider the committed state.
+    pub fn num_distinct_values(&self, table_id: TableId, cols: &ColList) -> Option<NonZeroU64> {
```
Centril, line 2606: In that case this should reflected in method naming imo.

```suggestion
    // Note, because this method is used for the purpose of estimation,
    // we only consider the committed state.
    pub fn committed_num_distinct_values(&self, table_id: TableId, cols: &ColList) -> Option<NonZeroU64> {
```

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -2587,3 +2588,24 @@ fn unindexed_iter_by_col_range_warn(
...
+        let table = self.committed_state_write_lock.get_table(table_id)?;
+        let (_, index) = table.get_index_by_cols(cols)?;
+        NonZeroU64::new(index.num_keys() as u64)
```
Centril, line 2609:
```suggestion
        self.committed_state_write_lock.num_distinct_values(table_id)
```
Let's move this impl to `CommitedState` so that `TxId` can share it.

crates/core/src/vm.rs
```diff
@@ -462,12 +500,16 @@ pub struct DbProgram<'db, 'tx> {
 pub fn check_row_limit<Query>(
     queries: &[Query],
     db: &RelationalDB,
-    tx: &TxId,
-    row_est: impl Fn(&Query, &TxId) -> u64,
+    tx: &TxMode<'_>,
```
Centril, line 503: It seems we're using `TxMode` to pick an implementation without actually needing branching at runtime, which is unfortunate. I'd rather have the impl be selected via a trait (that has `fn row_limit`) and any use of `TxMode` pushed towards the upper layers.

crates/core/src/subscription/module_subscription_actor.rs
Centril: Let's explain why `Scoped` exists in the code as well :)

Centril (commented):
I think the most important thing here is that we run benchmarks and ensure that we don't regress subscriptions performance by moving to mut txns

Shubham8287 (approved):
Looks good, lets wait for bot test before merging.

Issue #3513, opened 2025-10-28
Procedures: Rust client SDK invocation interface


gefjon:
Extend the Rust client SDK and codegen to be able to invoke procedures as `async` functions and then `await` their results. Then create tickets to do the same in C#, TypeScript and Unreal client SDKs.

PR #3532, opened 2025-10-30
Add Rust client SDK bindings for calling procedures

gefjon:
# Description of Changes

This commit adds support to the Rust client SDK for calling procedures.

Similar to reducers, each `DbContext` implementor has a `pub procedures: RemoteProcedures` field, with methods provided by extension traits for each procedure.

Unlike reducers, the provided methods are invoke and invoke-then. Invoke-then takes a `FnOnce` callback to run
when the SDK is notified of the procedure's termination status, while invoke ignores that notification.
No mechanism is provided for observing procedures invoked by other clients.

Procedure callbacks are implemented by storing a map from `request_id` to `ProcedureCallback`, with the callback closure internally knowing how to deserialize the return value. It's mildly unfortunate to deserialize within the callback instead of on the preprocess background task, but it saves significant complexity.

This commit also adds a new sdk-test module, `sdk-test-procedures`, and a new Rust test client, `procedure-client`.
Together, these are used in two tests of invoking and observing procedures. I've left TODOs for other tests that we should write as we implement additional procedure features.

I also had to fix a few minor bugs in the Rust codegen which were not strictly related to procedures: we previously assumed that the sets of reducers and of tables were non-empty, which led to wonky invalid codegen on modules which did not define any reducers or which did not define any tables.

I'm sneaking a change to the Nix flake into this PR as well - when initially writing it I had included `cargoArtifacts` (the pre-built and cached dependencies of our actual builds) in its `packages`, but that was neither necessary or useful, and just made building the shell take a long time after dependency changes.

Breaks the internal interface between the Rust client SDK and codegen, so users will have to re-run `spacetime generate`.

2-ish? Pretty simple change to the Rust SDK overall.

# Testing

- [x] Added new automated integration tests exercising the new functionality.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -168,7 +168,6 @@
             packages = [
               rustStable
               rustNightly
-              cargoArtifacts
               pkgs.cargo-insta
             ];
           };


gefjon:
Closes #3513

crates/codegen/src/rust.rs
```diff
@@ -487,6 +488,108 @@ impl {set_reducer_flags_trait} for super::SetReducerFlags {{
...
+        let mut arg_names_list = String::new();
+        for (arg_ident, arg_ty) in &procedure.params_for_generate.elements[..] {
+            arg_types_ref_list += "&";
+            write_type(module, &mut arg_types_ref_list, arg_ty).unwrap();
+            arg_types_ref_list += ", ";
+
+            let arg_name = arg_ident.deref().to_case(Case::Snake);
+            arg_names_list += &arg_name;
+            arg_names_list += ", ";
+        }
```
Centril: Feels like this could be shared with reducers?

modules/sdk-test-procedure/README.md
```diff
@@ -0,0 +1,13 @@
+# `sdk-test-procedure` *Rust* test
+
+This module tests that our client SDKs can invoke procedures and observe their return values.
+
+It is separate from [`sdk-test`](../sdk-test) because module library support for procedures across languages is was not yet universal as of writing (pgoldman 2025-10-30), and so it was not possible to implement this module in [TypeScript](../sdk-test-ts) and [C#](../sdk-test-cs).
```
Centril, line 5: Is the plan to merge them later?

sdks/rust/src/db_connection.rs
```diff
@@ -467,6 +480,32 @@ impl<M: SpacetimeModule> DbContextImpl<M> {
                     .call_reducer_flags
                     .set_flags(reducer_name, flags);
             }
+            PendingMutation::InvokeProcedureWithCallback {
+                procedure,
+                args,
+                callback,
+            } => {
+                let request_id = next_request_id();
```
Centril: Should we do this for reducer calls too?

sdks/rust/src/db_connection.rs
```diff
@@ -467,6 +480,32 @@ impl<M: SpacetimeModule> DbContextImpl<M> {
...
+                    flags: ws::CallProcedureFlags::Default,
+                });
+                self.send_chan
+                    .lock()
+                    .unwrap()
+                    .as_mut()
+                    .ok_or(crate::Error::Disconnected)?
+                    .unbounded_send(msg)
+                    .expect("Unable to send procedure call message: WS sender loop has dropped its recv channel");
+            }
```
Centril: Structurally this is similar to `CallReducer`, so reading this code next to that code would be helpful when trying to understand/contrast/compare. If you could move it right before/after that variant that would be great.

sdks/rust/src/db_connection.rs
gefjon: We don't have a use for it for reducer calls. I'll add a comment to the reducer call path describing how and why we don't bother.

modules/sdk-test-procedure/README.md
gefjon: Not particularly. We could, but I don't feel strongly about it.

PR #3498, opened 2025-10-27
WASM host execution for procedures

gefjon:
# Description of Changes

This commit builds support for executing procedures in WASM modules. This includes an HTTP endpoint, `/v1/database/:name_or_address/procedure/:name POST`, as well as an extension to the WS protocol. These new APIs are not wired up to the CLI or SDKs, but I have manually tested the HTTP endpoint via `curl`. The new WS extensions are completely untested.

Several TODOs are scattered throughout the new code, most notably for sensibly tracking procedure execution time in the metrics.

I also expect that we will want to remove the `procedure_sleep_until` syscall and the `ProcedureContext::sleep_until` method prior to release.

# API and ABI breaking changes

Adds new APIs and ABIs.

# Expected complexity level and risk

3? 4? Unlikely to break existing stuff, 'cause it's mostly additive, but adds plenty of potentially-fragile new stuff. Notably is the first time we're doing anything actually `async`hronous on a database core Tokio worker, and we don't yet have strong evidence of how that will affect reducer execution.

# Testing

- [x] Manually published `modules/module-test` and executed procedures with the following `curl` invocations:
  - `curl -X POST -H "Content-Type:application/json" -d '[]' http://localhost:3000/v1/database/module-test/procedure/sleep_one_second`
  -  `curl -X POST -H "Content-Type:application/json" -d '[1223]' http://localhost:3000/v1/database/module-test/procedure/return_value`
- [x] Need to write automated tests.

PR #3537, opened 2025-10-30
Update C# client-api bindings

gefjon:
# Description of Changes

Following #3498 , run the `gen-client-api.sh` tool to fix one of the non-required CI checks.

N/a

1

# Testing

Automated testing is fine.


bfops:
The unity testsuite is passing here, so this PR seems to fix it as desired. I'm merging this.

PR #3538, opened 2025-10-30
Add UUID built-in convenience type to SpacetimeDB

### LINKED ISSUE #3290: Add UUID built-in convenience type to SpacetimeDB
```rs
use spacetimedb::Uuid;

#[table(name = foobar)]
struct Foobar {
    uuid: Uuid,
}

#[reducer]
fn baz(ctx: &ReducerContext) {
    ctx.db.foobar.insert(Foobar {
        uuid: ctx.uuid(),
    });
}
```

```cs
using SpacetimeDB; // hypothetical namespace for Uuid, reducer, etc.

[Table("foobar")]
public class Foobar
{
    public Uuid Uuid { get; set; }
}

public static class Module
{
    [Reducer("baz")]
    public static void Baz(ReducerContext ctx)
    {
        ctx.Db.Foobar.Insert(new Foobar
        {
            Uuid = ctx.Uuid()
        });
    }
}
```

```ts
import { Uuid, table, reducer, ReducerContext } from "spacetimedb";

interface Foobar {
  uuid: Uuid;
}

table("foobar", {
  uuid: "uuid",
});

reducer("baz", (ctx: ReducerContext) => {
  ctx.db.foobar.insert({
    uuid: ctx.uuid(),
  } as Foobar);
});
```

mamcx:
# Description of Changes

Closes [#3290](https://github.com/clockworklabs/SpacetimeDB/issues/3290).

Adds a new "special" type to SATS, `UUID`, which is represented as the product `{ __uuid__: u128 }`. Adds versions of this type to all of our various languages' module bindings libraries and client SDKs, and updates codegen to recognize it and output references to those named library types. Adds methods for creating new UUIDs according to the V4 (all random) and V7 (timestamp, monotonic counter and random) specifications.

We add a new type 

2

it impacts all over the code

# Testing

- [x] Extends the Rust and Unreal SDK tests, and the associated `module-test` modules in Rust, C# and TypeScript, with uses of UUIDs.
- [x] Extends the C# SDK regression tests with uses of UUIDs.
- [x] Extends the TypeScript test suite with tests with uses of UUIDs.


JasonAtClockwork:
I've rebased my changes and added them to the branch; this reworks the Unreal C++ side of things for UUID to introduce a new type, adds tests to the TestClient that confirms values and matches all the changes up with the Unreal codegen. If you have any questions don't hesitate to reach out!

JasonAtClockwork:
@rekhoff - I'll need your eyes on the Unreal changes as they were significant enough to add the new UUID object and integration tests.

sdks/rust/tests/test-client/src/main.rs
```diff
@@ -1138,6 +1243,7 @@ fn every_primitive_struct() -> EveryPrimitiveStruct {
         r: ConnectionId::ZERO,
         s: Timestamp::from_micros_since_unix_epoch(9876543210),
         t: TimeDuration::from_micros(-67_419_000_000_003),
+        u: Uuid::NIL,
```
JasonAtClockwork: The comment from `every_primitive_struct()` above this code calls out the values are chosen to be non-zero and asymmetrical binary representations to make sure to catch endianness issues. I think Uuid follow the same pattern here.

JasonAtClockwork (changes requested):
I've gone through the Rust, TypeScript, and C# updates as well as retested my Unreal changes and it's looking solid but there seems to be missing integration tests for both TypeScript and C#. Rust looks solid but we'll want someone with more experience to review that.

For TypeScript I'm not sure how our testing framework is set up so I could be wrong. Maybe Noa can help guide here.

For C# sadly we're not hooked directly into the sdk-test or sdk-test-cs but we should add integration test for Uuid into /sdks/csharp/~examples/regression-tests to make sure there is coverage. Myself or Ryan can help here if it isn't clear how to plug in and run those.

sdks/rust/tests/test-client/src/main.rs
JasonAtClockwork: Confirmed fixed

JasonAtClockwork (approved):
C# + Unreal looks good to me though I highly advise another pair of eyes before merging.

sdks/rust/src/lib.rs
```diff
@@ -33,7 +33,7 @@ pub use table::{Table, TableWithPrimaryKey};
 
 pub use spacetime_module::SubscriptionHandle;
 pub use spacetimedb_client_api_messages::websocket::Compression;
-pub use spacetimedb_lib::{ConnectionId, Identity, ScheduleAt, TimeDuration, Timestamp};
+pub use spacetimedb_lib::{ClockGenerator, ConnectionId, Identity, ScheduleAt, TimeDuration, Timestamp, Uuid};
```
gefjon:
```suggestion
pub use spacetimedb_lib::{ConnectionId, Identity, ScheduleAt, TimeDuration, Timestamp, Uuid};
```

I don't think users of the client SDK need access to `ClockGenerator`.

crates/bindings/src/lib.rs
```diff
@@ -1030,6 +1032,50 @@ impl ReducerContext {
...
+    /// fn generate_uuid_v4(ctx: &ReducerContext) -> Uuid {
+    ///     ctx.new_uuid_v4()
+    /// }
+    /// # }
+    /// ```
+    pub fn new_uuid_v4(&self) -> anyhow::Result<Uuid> {
+        let mut bytes = [0u8; 16];
+        self.rng().try_fill_bytes(&mut bytes)?;
+        Ok(Uuid::from_random_bytes_v4(bytes))
+    }
```
gefjon:
```suggestion
    ///  Create a new UUIDv4 using the built-in RNG.
    //// # Example
    /// ```no_run
    /// # #[cfg(target_arch = "wasm32")] mod demo {
    /// use spacetimedb::{reducer, ReducerContext, Uuid};
    ///
    /// #[reducer]
    /// fn generate_uuid_v4(ctx: &ReducerContext) -> Uuid {
    ///     ctx.new_uuid_v4()
    /// }
    /// # }
    /// ```
    #[cfg(feature = "rand")]
    pub fn new_uuid_v4(&self) -> anyhow::Result<Uuid> {
        let mut bytes = [0u8; 16];
        self.rng().try_fill_bytes(&mut bytes)?;
        Ok(Uuid::from_random_bytes_v4(bytes))
    }
```

This method needs to be feature-gated behind `rand`. If a user compiles without default features, they can't use this method.

crates/bindings/src/lib.rs
```diff
@@ -27,6 +27,7 @@ pub use client_visibility_filter::Filter;
 pub use log;
 #[cfg(feature = "rand")]
 pub use rand08 as rand;
+use rand08::RngCore;
```
gefjon, line 31:
```suggestion
#[cfg(feature = "rand")]
use rand08::RngCore;
```

Otherwise this will cause compile errors when building without default features.

crates/bindings/src/lib.rs
```diff
@@ -1266,6 +1312,52 @@ impl ProcedureContext {
...
+    ///     ctx.new_uuid_v4()
+    /// }
+    /// # }
+    /// ```
+    #[cfg(feature = "unstable")]
+    pub fn new_uuid_v4(&self) -> anyhow::Result<Uuid> {
+        let mut bytes = [0u8; 16];
+        self.rng().try_fill_bytes(&mut bytes)?;
+        Ok(Uuid::from_random_bytes_v4(bytes))
+    }
```
gefjon:
```suggestion
    ///  Create a new UUIDv4 using the built-in RNG.
    //// # Example
    /// ```no_run
    /// # #[cfg(target_arch = "wasm32")] mod demo {
    /// use spacetimedb::{reducer, ReducerContext, Uuid};
    ///
    /// #[reducer]
    /// fn generate_uuid_v4(ctx: &ReducerContext) -> Uuid {
    ///     ctx.new_uuid_v4()
    /// }
    /// # }
    /// ```
    #[cfg(all(feature = "unstable", feature = "rand"))]
    pub fn new_uuid_v4(&self) -> anyhow::Result<Uuid> {
        let mut bytes = [0u8; 16];
        self.rng().try_fill_bytes(&mut bytes)?;
        Ok(Uuid::from_random_bytes_v4(bytes))
    }
```

crates/bindings/src/lib.rs
```diff
@@ -1030,6 +1032,50 @@ impl ReducerContext {
...
+    ///
+    /// #[reducer]
+    /// fn generate_uuid_v7(ctx: &ReducerContext) -> Result<Uuid, Box<dyn std::error::Error>> {
+    ///     let mut clock = ClockGenerator::new(ctx.timestamp);
+    ///     let uuid = ctx.new_uuid_v7(&mut clock)?;
+    ///     Ok(uuid)
+    /// }
+    /// # }
+    /// ```
+    pub fn new_uuid_v7(&self, clock: &mut ClockGenerator) -> anyhow::Result<Uuid> {
```
gefjon: Why don't we just stash a `ClockGenerator` (in a `RefCell` or similar) inside the `ReducerContext`? Why is this part of the public API?

gefjon: Also, are we concerned that reducers may run faster than 1/ms, resulting in a later TX seeing as its base the same timestamp as a previous TX generated as its second step?

gefjon:
From [Wikipedia](https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_7_(timestamp_and_random)):

> remaining 74 bits are random seeded counter (optional, at least 12 bits but no longer than 42 bits) and random.

It looks like we should be putting the counter in the low bits, not the timestamp.

gefjon:
Plan is to move the counter part of the generated UUIDv7 into the space reserved for the counter, and not increment the timestamp part. We'll store the source for the counter in the `ReducerContext` or `ProcedureContext`.

crates/bindings/src/lib.rs
mamcx: As discussed this is corrected now, I increase the internal counter to be be `0..i32::MAX`

crates/bindings/src/lib.rs
```diff
@@ -1030,6 +1038,46 @@ impl ReducerContext {
...
+    /// use spacetimedb::{reducer, ReducerContext, Uuid};
+    ///
+    /// #[reducer]
+    /// fn generate_uuid_v7(ctx: &ReducerContext) -> Result<Uuid, Box<dyn std::error::Error>> {
+    ///     let uuid = ctx.new_uuid_v7()?;
+    ///     log::info!(uuid);
+    /// }
+    /// # }
+    /// ```
+    pub fn new_uuid_v7(&self) -> anyhow::Result<Uuid> {
```
gefjon, line 1080:
```suggestion
    #[cfg(feature = "rand")]
    pub fn new_uuid_v7(&self) -> anyhow::Result<Uuid> {
```

Forgot that UUIDv7 also contains a random component last time I reviewed. Oops!

crates/bindings/src/lib.rs
```diff
@@ -974,6 +977,9 @@ pub struct ReducerContext {
 
     #[cfg(feature = "rand08")]
     rng: std::cell::OnceCell<StdbRng>,
+    /// A counter used for generating UUIDv7 values.
+    /// **Note:** must be 0..=i32::MAX
+    counter_uuid: RefCell<i32>,
```
gefjon:
```suggestion
    /// A counter used for generating UUIDv7 values.
    /// **Note:** must be 0..=i32::MAX
    // Disabled when not compiling with `rand`, as both v4 and v7 UUIDs have random components.
    #[cfg(feature = "rand")]
    counter_uuid: RefCell<i32>,
```

crates/bindings/src/lib.rs
```diff
@@ -987,6 +993,7 @@ impl ReducerContext {
             sender_auth: AuthCtx::internal(),
             #[cfg(feature = "rand08")]
             rng: std::cell::OnceCell::new(),
+            counter_uuid: RefCell::new(0),
```
gefjon:
```suggestion
            #[cfg(feature = "rand")]
            counter_uuid: RefCell::new(0),
```

crates/bindings/src/lib.rs
```diff
@@ -1000,6 +1007,7 @@ impl ReducerContext {
             sender_auth: AuthCtx::from_connection_id_opt(connection_id),
             #[cfg(feature = "rand08")]
             rng: std::cell::OnceCell::new(),
+            counter_uuid: RefCell::new(0),
```
gefjon:
```suggestion
            #[cfg(feature = "rand")]
            counter_uuid: RefCell::new(0),
```

crates/bindings/src/lib.rs
```diff
@@ -1089,6 +1137,9 @@ pub struct ProcedureContext {
     // as it could actually be seeded by OS randomness rather than a deterministic source.
     #[cfg(feature = "rand08")]
     rng: std::cell::OnceCell<StdbRng>,
+    /// A counter used for generating UUIDv7 values.
+    /// **Note:** must be 0..=i32::MAX
+    counter_uuid: RefCell<i32>,
```
gefjon:
```suggestion
    /// A counter used for generating UUIDv7 values.
    /// **Note:** must be 0..=i32::MAX
    // Disabled when compiling without `rand`, as both v4 and v7 UUIDs have random components.
    #[cfg(feature = "rand")]
    counter_uuid: RefCell<i32>,
```

crates/bindings/src/lib.rs
```diff
@@ -1101,6 +1152,7 @@ impl ProcedureContext {
             http: http::HttpClient {},
             #[cfg(feature = "rand08")]
             rng: std::cell::OnceCell::new(),
+            counter_uuid: RefCell::new(0),
```
gefjon:
```suggestion
            #[cfg(feature = "rand")]
            counter_uuid: RefCell::new(0),
```

crates/bindings/src/lib.rs
```diff
@@ -1266,6 +1318,49 @@ impl ProcedureContext {
...
+    ///
+    /// #[procedure]
+    /// fn generate_uuid_v7(ctx: &ProcedureContext) -> Result<Uuid, Box<dyn std::error::Error>> {
+    ///     let uuid = ctx.new_uuid_v7()?;
+    ///     log::info!(uuid);
+    ///     Ok(uuid)
+    /// }
+    /// # }
+    /// ```
+    #[cfg(feature = "unstable")]
```
gefjon:
```suggestion
    #[cfg(all(feature = "unstable", feature = "rand"))]
```

crates/client-api/Cargo.toml
```diff
@@ -47,7 +47,7 @@ bytestring = "1"
 tokio-tungstenite.workspace = true
 itoa.workspace = true
 derive_more = "0.99.17"
-uuid.workspace = true
+uuid = { workspace = true, features = ["v4"] }
```
gefjon: Is this necessary? There don't seem to be any other changes in the `client-api` crate that would require this.

crates/sats/src/uuid.rs
```diff
@@ -0,0 +1,381 @@
...
+    /// assert_eq!(
+    ///     "0000647e-5180-7000-8000-000200000000",
+    ///     uuid.to_string(),
+    /// );
+    /// ```
+    pub fn from_counter_v7(counter: &RefCell<i32>, now: Timestamp, random_bytes: &[u8; 4]) -> anyhow::Result<Self> {
+        // Monotonic counter value (31 bits)
+        let mut current = counter.borrow_mut();
+        // The counter is constructed from Timestamp::UNIX_EPOCH by callers
+        assert!(*current >= 0, "uuid counter must be non-negative");
```
gefjon: Why does this matter? And if the counter must be non-negative, why isn't it a `u32`?

crates/bindings/src/lib.rs
gefjon: Also, `Cell` should be sufficient here. No need to use `RefCell` for `Copy` types.

crates/sats/src/uuid.rs
```diff
@@ -0,0 +1,381 @@
...
+
+        let counter_val = {
+            let next = *current;
+            *current = next.wrapping_add(1) & 0x7FFF_FFFF;
```
gefjon: If we wrap here with `wrappring_add`, won't that cause the above assertion to fail? Is that intentional? If so, needs comments.

gefjon (commented):
This is looking very good, I think we're very close to merging. Thanks for all your hard work.

rekhoff:
> @rekhoff - I'll need your eyes on the Unreal changes as they were significant enough to add the new UUID object and integration tests.

I reviewed both the C# and the Unreal code. I think you caught the concerns. The Unreal tests could provide better coverage, they are fairly light at the moment. I did load up the Unreal `TestClient` and manually check a few different behaviors of the UUID, including adding them to blueprints. They where simple tests, but no issues found.

rekhoff (approved):
Signing off approval for the C# and as secondary on the Unreal changes.

gefjon (approved):

jdetter (approved):
I'm just reviewing as a CLI crate owner - the updates here are only within the tests and the updates look correct.

One request: could we update the PR description with the latest info for this PR in its current state before it merges? Thanks 👍

PR #3539, opened 2025-10-30
Moves pnpm generate scripts into rust tools scripts

cloutiertyler:
# Description of Changes

This PR removes some bash commands from the various `pnpm generate` scripts, and instead moves them into Rust scripts. It also adds module bindings for the example typescript template clients which should be in there anyway, so that we can test the template clients without running `spacetime init`.

None

2

# Testing
- [x] I ran all the Rust code. You can confirm that the generated files are correct by running `pnpm generate` in the root.


crates/bindings-typescript/examples/basic-react/package.json
```diff
@@ -7,13 +7,15 @@
     "dev": "vite",
     "build": "tsc -b && vite build",
     "preview": "vite preview",
-    "local": "spacetime publish --project-path spacetimedb --server local && spacetime generate --lang typescript --out-dir src/module_bindings --project-path spacetimedb",
-    "deploy": "spacetime publish --project-path spacetimedb && spacetime generate --lang typescript --out-dir src/module_bindings --project-path spacetimedb"
+    "generate": "cargo run -p gen-bindings -- --out-dir src/module_bindings --project-path ../../../cli/templates/basic-typescript/server && prettier --write src/module_bindings",
+    "spacetime:generate": "spacetime generate --lang typescript --out-dir src/module_bindings --project-path spacetimedb",
+    "spacetime:publish:local": "spacetime publish --project-path server --server local",
+    "spacetime:publish": "spacetime publish --project-path server --server testnet"
```
bfops: wait, is `testnet` correct?

cloutiertyler: No, it's not, although it's not technically a regression. Will fix.

cloutiertyler:
Closing this as the changes were rolled into https://github.com/clockworklabs/SpacetimeDB/pull/3559

PR #3544, opened 2025-10-30
Add C++ Bindings

### LINKED ISSUE #3514: Procedures: C++ module bindings
Do the best recreation of the interface in #3444 in the as-yet-unreleased C++ module bindings library.

JasonAtClockwork:
# Description of Changes

This adds C++ server bindings (/crate/bindings-cpp) to allow writing C++ 20 modules.

- Emscripten WASM build system integration with CMake
- Macro-based code generation (SPACETIMEDB_TABLE, SPACETIMEDB_REDUCER, etc)
- All SpacetimeDB types supported (primitives, Timestamp, Identity, Uuid, etc)
- Product types via SPACETIMEDB_STRUCT
- Sum types via SPACETIMEDB_ENUM
- Constraints marked with FIELD* macros

None

2 - Doesn't heavily impact any other areas but is complex macro C++ structure to support a similar developer experience, did have a small impact on init command

# Testing

- [x] modules/module-test-cpp - heavily tested every reducer
- [x] modules/benchmarks-cpp - tested through the standalone (~6x faster than C#, ~6x slower than Rust)
- [x] modules/sdk-test-cpp
- [x] modules/sdk-test-procedure-cpp
- [x] modules/sdk-test-view-cpp  
- [x] Wrote several test modules myself
- [x] Quickstart smoketest [Currently in progress]
- [ ] Write Blackholio C++ server module


templates/basic-cpp/spacetimedb/CMakeLists.txt
```diff
@@ -0,0 +1,95 @@
...
+  add_subdirectory("${SPACETIMEDB_CPP_SDK_DIR}" "${CMAKE_BINARY_DIR}/_deps/spacetime_cpp_sdk-build")
+else()
+  include(FetchContent)
+
+  FetchContent_Declare(
+    spacetime_cpp_sdk
+    #URL "https://www.spacetimedb.com/cpp/spacetimedb-cpp-sdk-1.11.2.zip"
+    URL "https://www.rodentgames.com/spacetimedb/spacetimedb-cpp-sdk-1.11.2.zip"
+    DOWNLOAD_EXTRACT_TIMESTAMP TRUE
+    #URL_HASH SHA256=E10503F57AA3ADB10ADD82ECF2E0F3C1107B5C141FA5E157D174C30B71071434
```
JasonAtClockwork: Must be addressed before merging.

crates/bindings-cpp/QUICKSTART.md
```diff
@@ -0,0 +1,293 @@
...
+cd my-chat-module
+```
+
+This creates a project with the following structure:
+```
+my-chat-module/
+├── CMakeLists.txt
+├── spacetimedb/
+    └── lib.cpp
+└── .gitignore
```
rekhoff: For me, when I call the CLI to `init`, I am prompted to specify a `Project path`, which I can leave blank by hitting `Enter/Return`.
After which, the output structure is:
```suggestion
my-chat-module/spacetimedb/
├── CMakeLists.txt
├── src/
    └── lib.cpp
└── .gitignore
```
Additionally, the `my-chat-module` directory has several LLM helper docs output in it, but those are not important for the user to have listed here.

crates/bindings-cpp/QUICKSTART.md
```diff
@@ -0,0 +1,293 @@
...
+Start your local SpacetimeDB instance:
+
+```bash
+spacetime start
+```
+
+Publish your module:
+
+```bash
+spacetime publish . my-chat-db
```
rekhoff, line 190: What is the period (`.`) doing in the `publish` command?
For me, it returns an `error: unexpected argument 'my-chat-db' found`, I believe because it expects the `.` to be the name of the module I am publishing.
It works when I run:
```suggestion
spacetime publish my-chat-db
```

crates/bindings-cpp/tests/type-isolation-test/test_modules/test.o
rekhoff, line 1: Is this a left over artifact or something?
It's a binary file, which we don't usually want in the repo, and I couldn't find any references to it in the `type-isolation-test`. If it's not needed, let's remove it.

crates/codegen/src/cpp.rs
```diff
@@ -0,0 +1,564 @@
...
+            PrimitiveType::I16 => "int16_t",
+            PrimitiveType::U16 => "uint16_t",
+            PrimitiveType::I32 => "int32_t",
+            PrimitiveType::U32 => "uint32_t",
+            PrimitiveType::I64 => "int64_t",
+            PrimitiveType::U64 => "uint64_t",
+            PrimitiveType::I128 => "SpacetimeDb::I128",
+            PrimitiveType::U128 => "SpacetimeDb::U128",
+            PrimitiveType::I256 => "I256Placeholder", // TODO: proper 256-bit support
+            PrimitiveType::U256 => "U256Placeholder", // TODO: proper 256-bit support
```
rekhoff: Are these fully implemented? The tests appear to use `u256`, so I imagine these work, but this looks weird. If should make sure we have a test asserting codegen compiles using `i256`/`u256` if we don't already.

crates/bindings-cpp/include/spacetimedb/bsatn/algebraic_type.h.backup
```diff
@@ -0,0 +1,657 @@
...
+    // -------------------------------------------------------------------------
+    // Data Accessor Methods
+    // -------------------------------------------------------------------------
+    /**
+     * @brief Get the referenced type ID for a Ref type.
+     * @return The type ID this reference points to
+     * @throws std::runtime_error if this is not a Ref type
+     */
+    uint32_t as_ref() const {
+        if (!is_ref()) throw std::runtime_error("Type is not a Ref");
```
rekhoff: Is [algebraic_type.h.backup‎](https://github.com/clockworklabs/SpacetimeDB/pull/3544/changes#diff-58f92a45c71c266f0dc39d52185c970a44a346c87c5ed4a9f82beb79919d5106) intended to be removed? It contains several references to `throw std::runtime_error(...)`, so if it ever got included, it will cause a compiler failure in Emscripten since `crates/bindings-cpp/CMakeLists.txt` includes `-fno-exceptions`
Unless we have a reason to keep this backup header around, we should remove this.

crates/bindings-cpp/ARCHITECTURE.md
```diff
@@ -0,0 +1,644 @@
...
+### 1. Hybrid Compile-Time/Runtime System
+- **Compile-time validation**: C++20 concepts and static assertions catch constraint violations before compilation
+- **Runtime registration**: __preinit__ functions execute during WASM module load to register types and metadata
+- **Nominal type system**: Types identified by their declared names, not structural analysis
+- **Error detection**: Multi-layer validation system from compile-time through module publishing
+
+### 2. Outcome<T> - Rust-Like Error Handling
+The SDK provides `Outcome<T>`, a type-safe error handling mechanism matching Rust's `Result<T, E>` pattern:
+- **Outcome<void>** (type alias `ReducerResult`): Used by reducers, can return success (`Ok()`) or error (`Err(message)`)
+- **Outcome<T>**: Used by procedures and other operations, returns either a value or an error message
```
rekhoff: Do procedures return `Outcome<T>`?
Later in the same file (line 118), it says 
>Procedures return raw `T` (not `Outcome<T>`)

and in [procedures_macros.h at line 22](https://github.com/clockworklabs/SpacetimeDB/blob/562af3e19438cda348ba9370ab16ba4526c7fd01/crates/bindings-cpp/include/spacetimedb/procedure_macros.h#L22), it reads:
>The procedure macro wraps the return type in Outcome<T> automatically.

Then in [procedures_macros.h at line 41](https://github.com/clockworklabs/SpacetimeDB/blob/562af3e19438cda348ba9370ab16ba4526c7fd01/crates/bindings-cpp/include/spacetimedb/procedure_macros.h#L41), it reads:
>Return type directly (not wrapped in Outcome)

To me it looks like this returns a raw `T` in it's `return_type` but maybe I'm missing something. Either way, this is kinda confusing.

crates/bindings-cpp/QUICKSTART.md
```diff
@@ -0,0 +1,293 @@
...
+
+- **Structs** are just data types - they need `SPACETIMEDB_STRUCT` for serialization
+- **Tables** are database tables created with `SPACETIMEDB_TABLE` and store data persistently
+- The same struct can be used for multiple tables or just as a data type
+
+### Database Access Pattern
+
+SpacetimeDB C++ uses a unique accessor pattern:
+- `ctx.db[table_name]` - Access table for iteration and basic operations
+- `ctx.db[table_field]` - Access indexed fields for optimized operations
```
rekhoff: Minor nit here: Because this is different from the other languages, it's easy to miss. It might be nice to define that accessor `table_field` means a generated symbol named exactly `tableName_fieldName`. It'd defiantly implied, but it's easy to miss/overlook.

crates/bindings-cpp/tests/type-isolation-test/test_modules/test.o
JasonAtClockwork: This is an C++ object file and should not have been included, all cleaned up now

crates/codegen/src/cpp.rs
JasonAtClockwork: Fixed up, this was created before I got the types in place. Currently it generates the bare minimum for module definitions but in the future we can expand it if requested.

crates/bindings-cpp/include/spacetimedb/bsatn/algebraic_type.h.backup
JasonAtClockwork: Yes I missed this in the original copy from Arvikasoft thanks!

crates/bindings-cpp/ARCHITECTURE.md
JasonAtClockwork: I originally had mistakenly set up View + Procedure to use Outcome<> and had AI generate the documentation surrounding that. I *think* I've got it all cleaned up now in the markdowns and definitions.

crates/bindings-cpp/QUICKSTART.md
JasonAtClockwork: This is a good callout, I've updated the second line there and AI had a good suggestion to make it even more clear in the use just below.

crates/bindings-cpp/LICENSE
```diff
@@ -0,0 +1,731 @@
+SPACETIMEDB BUSINESS SOURCE LICENSE AGREEMENT
```
jdetter: I think this should be a symlink to the base license, as an example:
```
boppy@geralt:~/clockwork/SpacetimeDB$ ls -al crates/client-api/LICENSE
lrwxrwxrwx 1 boppy boppy 22 Jan 14 17:12 crates/client-api/LICENSE -> ../../licenses/BSL.txt
```

.github/workflows/ci.yml
```diff
@@ -171,12 +171,25 @@ jobs:
...
 
+      # Install cmake and emscripten for C++ module compilation tests.
+      - name: Install cmake and emscripten
+        run: |
+          apt-get update
+          apt-get install -y cmake
+          git clone https://github.com/emscripten-core/emsdk.git /opt/emsdk
+          cd /opt/emsdk
+          ./emsdk install latest
+          ./emsdk activate latest
```
jdetter: by cloning the repo we're just going to be testing against the latest master all the time - is that fine or should we pin to a known working version? I fear this test could start failing based on whatever is on `emsdk` master.

jdetter: This adds 29 seconds to the CI time - might be worth caching this step eventually. It depends on how fast Tyler gets the new smoktests running. Probably leave a TODO on this.

crates/bindings-cpp/ARCHITECTURE.md
```diff
@@ -0,0 +1,644 @@
...
+template<typename T> class Outcome<T>;
+```
+
+### Reducer Error Handling (ReducerResult / Outcome<void>)
+
+**Creating Results**:
+```cpp
+#include <spacetimedb.h>
+
+using namespace SpacetimeDb;
```
jdetter: it's probably too late to change this, but how much work would it be to make this `SpacetimeDB` instead of `SpacetimeDb`?

crates/bindings-cpp/ARCHITECTURE.md
```diff
@@ -0,0 +1,644 @@
...
+extern "C" __attribute__((export_name("__preinit__19_autoinc_register_47")))
+void __preinit__19_autoinc_register_47() {
+    SpacetimeDb::detail::get_autoinc_integrator<User>() = 
+        &SpacetimeDb::detail::autoinc_integrate_47;
+}
+```
+
+**Runtime Integration Process**:
+When `insert()` is called on a table with auto-increment fields:
+1. SDK serializes and sends the row to SpacetimeDB
```
jdetter:
```suggestion
1. The logic in the bindings serializes and sends the row to SpacetimeDB
```

jdetter: I would not override terminology here - we call the SDK the thing that clients use and the SDK that's used on the module side is just called the "bindings". Mainly I want to update this so the AI training doesn't get confused.

crates/bindings-cpp/ARCHITECTURE.md
```diff
@@ -0,0 +1,644 @@
+# SpacetimeDB C++ SDK Architecture
```
jdetter:
```suggestion
# SpacetimeDB C++ Bindings Architecture
```

From below:
I would not override terminology here - we call the SDK the thing that clients use and the SDK that's used on the module side is just called the "bindings". Mainly I want to update this so the AI training doesn't get confused.

jdetter: Could you update the references here from SDK => bindings?

crates/bindings-cpp/DEVELOP.md
```diff
@@ -0,0 +1,660 @@
...
+
+### Layer 1: SPACETIMEDB_STRUCT - Manual Field Enumeration
+
+```cpp
+struct User {
+    uint32_t id;
+    std::string name;
+    std::string email;
+};
+SPACETIMEDB_STRUCT(User, id, name, email)
```
jdetter, line 26: Are we allowed to forward declare here? Something like:
```cpp
struct User;

SPACETIMEDB_STRUCT(User, id, name, email)
struct User {
    uint32_t id;
    std::string name;
    std::string email;
};
```

Just so that it's closer to our other languages in terms of expressing how tables are setup.

crates/bindings-cpp/DEVELOP.md
```diff
@@ -0,0 +1,660 @@
...
+```
+
+**Why it must be runtime:** The module schema must be built dynamically when the WASM module loads, as we can't generate a complete module description at compile-time.
+
+### Layer 3: FIELD_ Macros - Constraint Registration
+
+```cpp
+FIELD_PrimaryKey(users, id);
+FIELD_Unique(users, email);
+FIELD_Index(users, name);
```
jdetter, line 75: I'm assuming people who know a lot more about C++ than I do have thought about this thoroughly, but I would ask: is there no better way to do this?

jdetter: ~~I know we're somewhat limited by the simplicity of C/C++ macros.~~

jdetter: ok it seems like we can do a lot more post-C++26, right?

crates/bindings-cpp/DEVELOP.md
```diff
@@ -0,0 +1,660 @@
...
+```
+
+**Illustrative C++26 approach** (exact syntax TBD in standardization):
+```cpp
+// Natural C++ attributes replace macros
+struct [[spacetimedb::table("users", public)]] User {
+    [[spacetimedb::primary_key]] uint32_t id;
+    [[spacetimedb::unique]] std::string email;
+    std::string name;
+};
```
jdetter, line 378: This is so so so much better than the macro syntax.

crates/bindings-cpp/QUICKSTART.md
```diff
@@ -0,0 +1,293 @@
...
+This guide will walk you through creating your first SpacetimeDB module in C++. We'll build a simple chat server to demonstrate the core concepts.
+
+## What is a SpacetimeDB Module?
+
+A SpacetimeDB module is C++ code that gets compiled to WebAssembly and runs inside the database. Instead of the traditional architecture (database → app server → clients), SpacetimeDB lets you write your entire backend logic that runs **inside** the database itself, giving you microsecond latency and automatic real-time sync to clients.
+
+Modules consist of two main components:
+
+- **Tables**: Database tables defined as C++ structs
+- **Reducers**: Functions that modify data and can be called by clients
```
jdetter, line 12: I would add views as well to this list

templates/basic-cpp/spacetimedb/src/lib.cpp
```diff
@@ -0,0 +1,49 @@
+#include "spacetimedb.h"
```
jdetter, line 1: Tested working - I needed to set this env to get it to compile though (see below). We should double-check that it will build post-merge or whenever we send this out.
```
export SPACETIMEDB_CPP_SDK_DIR=~/clockwork/SpacetimeDB/crates/bindings-cpp
```

crates/cli/src/subcommands/dev.rs
```diff
@@ -403,6 +403,7 @@ async fn generate_build_and_publish(
         crate::util::ModuleLanguage::Rust => &Language::Rust,
         crate::util::ModuleLanguage::Csharp => &Language::Csharp,
         crate::util::ModuleLanguage::Javascript => &Language::TypeScript,
+        crate::util::ModuleLanguage::Cpp => &Language::UnrealCpp,
```
jdetter: should we pick rust as the sane-default client for Cpp for now? We sort of expect a rust client for now no?

crates/cli/src/subcommands/init.rs
```diff
@@ -73,6 +75,7 @@ impl ServerLanguage {
             "rust" => Ok(Some(ServerLanguage::Rust)),
             "csharp" | "c#" => Ok(Some(ServerLanguage::Csharp)),
             "typescript" => Ok(Some(ServerLanguage::TypeScript)),
+            "cpp" | "c++" => Ok(Some(ServerLanguage::Cpp)),
```
jdetter: cxx as well maybe?

crates/cli/src/subcommands/init.rs
```diff
@@ -1749,3 +1790,50 @@ fn strip_mdc_frontmatter(content: &str) -> &str {
...
     content
 }
+
+/// Check if Emscripten and CMake tooling are available in PATH.
+///
+/// On Windows, looks for emcc.bat and cmake.exe.
+/// On Unix-like systems, looks for emcc and cmake.
+///
+/// Returns true if both tools are found, false otherwise with helpful warnings.
+fn check_for_emscripten() -> bool {
```
jdetter: `check_for_emscripten_and_cmake` maybe

crates/bindings-cpp/QUICKSTART.md
```diff
@@ -0,0 +1,293 @@
...
+}
+```
+
+### Step 5: Build Your Module
+
+Build the module using the provided CMake configuration:
+
+```bash
+emcmake cmake -B build .
+cmake --build build
```
jdetter: should we not just tell the user to `spacetime build` here?

crates/bindings-cpp/QUICKSTART.md
```diff
@@ -0,0 +1,293 @@
...
+    }
+    return Ok();
+}
+```
+
+## Troubleshooting
+
+**Build errors**: Ensure you have the latest Emscripten SDK and are using `emcmake cmake`
+
+**Module not found**: Check that SpacetimeDB is running with `spacetime server status`
```
jdetter: `spacetime server status` is not a command

crates/bindings-cpp/README.md
```diff
@@ -0,0 +1,285 @@
...
+- Full equivalence with Rust and C# SDK test modules
+- Examples of all constraint types and database operations
+
+## Contributing
+
+This library is part of the SpacetimeDB project. Please see the main repository for contribution guidelines.
+
+## License
+
+Apache License 2.0 - See LICENSE file in the root directory.
```
jdetter: What license is this referring to? Certainly the license for this crate is BSL right? Why does this say Apache2?

crates/bindings-cpp/REFERENCE.md
```diff
@@ -0,0 +1,1290 @@
+# SpacetimeDB C++ Module Reference
```
jdetter, line 1: does this become our API reference doc?

crates/bindings-cpp/QUICKSTART.md
```diff
@@ -0,0 +1,293 @@
+# SpacetimeDB C++ Module Quickstart
```
jdetter, line 1: This becomes our quickstart chat example in the docs, right?

crates/bindings-cpp/QUICKSTART.md
```diff
@@ -0,0 +1,293 @@
...
+
+- **Structs** are just data types - they need `SPACETIMEDB_STRUCT` for serialization
+- **Tables** are database tables created with `SPACETIMEDB_TABLE` and store data persistently
+- The same struct can be used for multiple tables or just as a data type
+
+### Database Access Pattern
+
+SpacetimeDB C++ uses a unique accessor pattern:
+- `ctx.db[table_name]` - Access table for iteration and basic operations
+- `ctx.db[table_field]` - Access indexed fields for optimized operations (generated symbol named `tableName_fieldName`)
```
jdetter: This seems a bit weird, shouldn't it be something like this instead:
```
ctx.db[table_name].table_field
```
What if 2 tables have the same indexed field name?

jdetter (commented):
Some comments/questions to get started here. There is a lot to review here so I'll do another pass later today.

The main thing I would bring up is that we are referring to this as the "C++ SDK" which I think we should rebrand to the "C++ bindings". My reason is that the other bindings libraries all refer to themselves as bindings, not an SDK. We typically use the nomenclature "SDK" to mean the client SDKs, not the module side bindings.

I fear that if we start referring to this as our C++ SDK this will both confuse users and confuse the LLM training especially if we add a C++ client SDK in the future. If you want feedback from the team on this it might be a good topic to bring up during standup.

Let me know when you're ready for another pass, I'm really excited that this seems like it's working so far and I'm excited for the future with C++26! 🙂

crates/bindings-cpp/ARCHITECTURE.md
JasonAtClockwork: I've gone through and fixed all locations to rework C++ SDK to C++ bindings locally and will commit that in a bit

crates/bindings-cpp/DEVELOP.md
JasonAtClockwork: The problem is compile-time reflection is basically non-existent until C++ 26. This model that Ludvig came up with after multiple attempts to support compile-time checks and compile-time definitions, otherwise we'd be heading towards builder territory which would be even farther from the other languages.

crates/bindings-cpp/DEVELOP.md
JasonAtClockwork: 100% agree, without trait/attributes C++ is the ugly step-child.

crates/bindings-cpp/LICENSE
JasonAtClockwork: Added the symlink for the LICENSE

.github/workflows/ci.yml
JasonAtClockwork: Great catch I meant to pin it to a conservative version and completely forgot. I've pinned it to a release in early Dec 2025 and will need to keep an eye on it for when we need to upgrade.

crates/bindings-cpp/ARCHITECTURE.md
JasonAtClockwork: Better to do it now! I've reworked the entire codebase and will be fixing up the docs shortly. Was only like two hours and and mostly just making sure tests were solid.

crates/bindings-cpp/QUICKSTART.md
JasonAtClockwork: That was an oversight, I've also added procedures

crates/bindings-cpp/DEVELOP.md
JasonAtClockwork: The macro expands out into code that needs access to the struct's members. That means we need the completel type definition and a forward declare won't be enough here. Sadly more ugly until C++ 26 is fully adopted...

templates/basic-cpp/spacetimedb/src/lib.cpp
JasonAtClockwork: This is odd, I was able to build with my Linux machine allowing it to download with FetchContent on the hosted version from my site. I'll do another pass tonight once I get all the other comments addressed.

crates/bindings-cpp/REFERENCE.md
JasonAtClockwork: I think I'll work on a new version that matches with how the others are laid out, this was a hold over from Arvikasoft's work that I've maintained for now to give us a starting point.

crates/bindings-cpp/QUICKSTART.md
JasonAtClockwork: Same as the REFERENCE I want to get this matching the rest of our docs and this file might go away so we only maintain one location for documentation.

crates/bindings-cpp/QUICKSTART.md
JasonAtClockwork: Ahh yeah this is still confusing, it is ctx.db[table_name_and_field_name] with how the accessors get generated. Maybe I change it to?
```
- `ctx.db[table]` - Access table for iteration and basic operations, eg. ctx.db[user]
- `ctx.db[table_field]` - Access indexed fields for optimized operations, eg. ctx.db[user_id]
```

crates/cli/src/subcommands/dev.rs
JasonAtClockwork: Changed, that's the right call I've also updated the template (basic-cpp) to use a Rust client

crates/cli/src/subcommands/init.rs
JasonAtClockwork: Yes great call, I've updated it

crates/cli/src/subcommands/init.rs
JasonAtClockwork: This was a miss when I combine the two methods into one simpler version, I've fixed it up

crates/bindings-cpp/QUICKSTART.md
JasonAtClockwork: 100% agree, I was waffling on this and I've cleaned it up except leaving one in the REFERENCE.md.

crates/bindings-cpp/QUICKSTART.md
JasonAtClockwork: Fixed

crates/bindings-cpp/README.md
JasonAtClockwork: This was a miss in my cleanup from Arvikasoft, I've removed it now

crates/bindings-cpp/QUICKSTART.md
JasonAtClockwork: I've made the change and going to leave this comment open as it might not be good enough yet

templates/basic-cpp/spacetimedb/src/lib.cpp
JasonAtClockwork: Okay confirmed building on my Linux box from a fresh pull without SPACETIMEDB_CPP_SDK_DIR set. It's possible I had not updated the hosted zip CMakeLists fetches causing errors on build.

crates/bindings-cpp/QUICKSTART.md
rekhoff: If the goal is to simply clarify how it currently works, my personal preference towards `camelCase`:
```
- `ctx.db[tableName]` - Access table for iteration and basic operations, eg. ctx.db[user]
- `ctx.db[tableName_fieldName]` - Access indexed fields for optimized operations, eg. ctx.db[user_id]
```
This is because it allows it to be clear that we are specifically referring to the name of a given table/field, while also not adding any additional special characters, like underscore, that is present in `snake_case`. The extra underscore get confusing when those are also used as the table to field delimiters.

crates/bindings-cpp/ARCHITECTURE.md
rekhoff: Found another one in the same doc:
Lines 51-52:
```cpp
// For procedures - return a value on success, error message on failure
template<typename T> class Outcome<T>;
```
Looks like this still references a return of `Outcome<T>`

JasonAtClockwork: Not sure if it's like pulling weeds, or playing whack-a-mole here! Thanks I've cleaned up this reference!

crates/bindings-cpp/QUICKSTART.md
JasonAtClockwork: Thanks for the feedback I like that and it's in!

rekhoff (approved):
Overall LGTM. I very much wish we could use C++26, but given our options, I think the current implementation makes sense. I've reviewed the documentation, skimmed the code, and done some simple local testing. I'd love more eyes on this from core team members, but am happy to sign off approval.

JasonAtClockwork:
Quick update - I didn't love the implementation for the iterator on filter() I had a mix of different uses and was going to move the read-write index to use the same pattern as the View/Multicolumn as the version from Arvikasoft was materializing the entire iterator into a vector.

I really hate the devex of that pattern though as you have to manually loop through the iterator so I've added a small wrapper to make this more clean to go from:
```cpp
for (auto iter = ctx.db[score_points].filter(range_from(300)); 
     iter != IndexIterator<Score>(); ++iter) {
    const auto& score = *iter;
}
```

To:
```cpp
for (const auto& score : ctx.db[score_points].filter(range_from(300)) {
}
```

crates/bindings-cpp/QUICKSTART.md
jdetter: Yeah this is fine with me for now - is this something we can fix with C++26?

crates/bindings-cpp/ARCHITECTURE.md
jdetter: Thank you!

jdetter:
@JasonAtClockwork I would take a look at this CI failure, I think it's unlikely that it's a flake because it seems pretty consistent. It might be something introduced by this PR:
image

jdetter (approved):
I've tested this at least with the basic quickstart doc and everything seems to be working. I also looked back and did another pass on the comments that I put down - let me know if you need me to take another look at anything.

I have not done a nit-pick review on the code here since there is quite a lot but I have skimmed it and haven't found anything too concerning. I think it would be better to do a UX review when we go through the follow-up docs PRs anyway. Please request my review when those are ready as well.

Thanks Jason, this is great 👍

crates/bindings-cpp/QUICKSTART.md
JasonAtClockwork: I'm not 100% certain to be honest, the proposal for reflection in C++ 26 doesn't have full code injection, it should help with metaprogramming but I need to play with it before I know if I can resolve this completely to make it more in the style of: ctx.db.user.id or ctx.db.user().id().

gefjon (approved):
I've only looked at the procedure-related stuff in here, but it looks great - much better than I expected, to be honest. Thanks for all your hard work on this!

crates/bindings-cpp/include/spacetimedb/view_context.h
```diff
@@ -0,0 +1,80 @@
...
+ *     for (const auto& item : ctx.db[item_owner].filter(ctx.sender)) {
+ *         my_items.push_back(item);
+ *     }
+ *     return Ok(my_items);
+ * }
+ * @endcode
+ */
+struct ViewContext {
+    // Caller's identity - who invoked this view
+    Identity sender;
```
joshua-spacetime, line 37: Note, `sender` will be a method call in 2.0.

joshua-spacetime (approved):
The view code looks correct. `sender` is a method in the 2.0 rust module api. But as I understand it, this is not meant to be 2.0 compliant. Also this does not include the query builder which will be added separately.

crates/bindings-cpp/include/spacetimedb/view_context.h
JasonAtClockwork: Yeah I've been following your work but as this was merging into master I held that off and will do a separate PR to catch up!

jdetter (approved):
I reviewed the changes to `tools/upgrade-version` and the new versioning change, everything looks good here - just a reminder to make sure everything is tested before merging 🙂 

Thanks Jason!

templates/basic-cpp/spacetimedb/CMakeLists.txt
JasonAtClockwork: This has been complete using GitHub with a new split repository `spacetimedb-bindings-cpp`

templates/basic-cpp/LICENSE
cloutiertyler, line 1: Should this be a symlink or now?

cloutiertyler (approved):
Approving my code owned files. Left one comment.

templates/basic-cpp/LICENSE
JasonAtClockwork: Yes yes it should that is a miss, added now

PR #3545, opened 2025-10-30
Add docs for using Auth claims within modules

jsdt:
# Description of Changes

This adds some code samples for using auth claims in Rust, Typescript, and C#.

Much of this was originally in https://github.com/clockworklabs/SpacetimeDB/pull/3218.

1: this is a docs-only change.

# Testing

I've tested these code samples locally (with the changes in https://github.com/clockworklabs/SpacetimeDB/pull/3542).


docs/docs/11-How-To/03-using-auth-claims.md
```diff
@@ -0,0 +1,265 @@
...
+
+# Using Auth Claims
+
+SpacetimeDB allows you to easily access authentication (auth) claims embedded in OIDC-compliant JWT tokens. Auth claims are key-value pairs that provide information about the authenticated user. For example, they may contain a user's unique ID, email, or authentication provider. If you want to view these fields for yourself, you can inspect the contents of any JWT using online tools like [jwt.io](https://jwt.io/).
+
+Within a SpacetimeDB reducer, you can access the auth claims from a client's token via the `ReducerContext`. Below are some examples of how to use these claims.
+
+## Accessing Common Claims: Subject and Issuer
+
+The subject (`sub`) and issuer (`iss`) are the most commonly accessed claims in a JWT. The subject usually represents the user's unique identifier, while the issuer indicates which authentication provider issued the token. These are required claims, which are used to compute each user's `Identity`. Because these are so commonly used, there are helper functions to get them.
```
cloutiertyler:
```suggestion
The subject (`sub`) and issuer (`iss`) are the most commonly accessed claims in a JWT. The issuer indicates which authentication provider issued the token, and the subject represents the unique identifier assigned to the user by the issuer. These are required claims, which are used to compute each user's `Identity`. Because these are so commonly used, there are helper functions to get them.
```

docs/docs/11-How-To/03-using-auth-claims.md
```diff
@@ -0,0 +1,265 @@
...
+    };
+    log::info!("sub: {}, iss: {}", subject, issuer);
+    Ok(())
+}
+```
+
+*Example output when using a Google-issued token:*
+```
+INFO: src\lib.rs:64: sub: 321321321321321, iss: https://accounts.google.com
+```
```
cloutiertyler, line 39: Is there a reason this is just in the Rust one?

docs/docs/11-How-To/03-using-auth-claims.md
```diff
@@ -0,0 +1,265 @@
...
+  console.info(`Client connected with sub: ${jwt.subject}, iss: ${jwt.issuer}`);
+});
+```
+
+</TabItem>
+</Tabs>
+
+## Example: Restricting auth providers
+
+Since users can use any valid token to connect to SpacetimeDB, their token may originate from any authentication provider. For example, they could send an OIDC compliant token from Github even though you only want to accept tokens from Google. It is best practice to check at least the issuer and audience of tokens when clients connect, so you can ensure that your data can only be accessed by users of your application.
```
cloutiertyler:
```suggestion
Since users can use any valid token to connect to SpacetimeDB, their token may originate from any authentication provider. For example, they could send an OIDC compliant token from GitHub even though you only want to accept tokens from Google. It often makes sense to restrict access to your module to users issued by your own issuer or specific issuers. It is best practice to check at least the issuer when clients connect, so you can ensure that your data can only be accessed by users of your application.

Additionally, it's imperative that you check the `aud` claim to ensure the issuer intended your application to receive the token. This ensures that applications which send you the token, cannot repurpose tokens issued for use with another application.
```

cloutiertyler: I am still concerned about having users validate the `aud` claim. I do not believe we should ever deliver tokens to a module whose aud does not match the aud claim of the token.

cloutiertyler: I think for now we can leave this up to the module developer, but in 2.0 I would like to change this to be part of the database's configuration and be enforced by SpacetimeDB so these tokens never actually arrive at the module at all.

docs/docs/11-How-To/03-using-auth-claims.md
jsdt: John had one that matched the rust version, but I didn't bother adding the example output for the other versions.

PR #3551, opened 2025-10-31
Fix Nix build by not using Git in Cargo build scripts

gefjon:
# Description of Changes

When building under Nix, Git metadata is not available within the sandbox, as we use `lib.cleanSource` on our source directory. This is important because it avoids spurious rebuilds and/or determinism hazards.

The build was broken due to our new `spacetime init` template system accessing Git metadata in the CLI's build.rs
to filter out non-git-tracked files from the templates. The Flake sandbox does this automatically (even without `lib.cleanSource`!), so when building under Nix it's unnecessary to do twice. (I remain unconvinced that it's necessary to do in non-Nix builds either, as CI builds should have a clean checkout and local dev builds don't need clean templates, but the behavior was already in master and I didn't feel comfortable removing it.)

As an enhancement, I've also found a Nix-ey way to embed our Git commit hash in builds. Previously, builds under Nix had the empty string instead of a commit hash, because we included the `git` CLI tool but scrubbed the necessary metadata. Now, we inject an environment variable from the Nix flake, and don't make the `git` CLI tool available at all. This has the convenient upside of allowing Nix builds to reference `dirtyRev` in builds with a dirty worktree, which should reduce confusion.

N/a

3? I didn't have a strong understanding of what the CLI build script was doing, and to what extent it was doing things intentionally versus for convenience. As such, it's possible that I've inadvertently damaged something load-bearing.

# Testing

- [x] Built with `nix build`, ran `spacetime init`, chose the `basic-rust` template, and got a reasonable-looking template instantiation.
- [ ] Hopefully we have automated tests for this?

### DIFF
--- a/crates/cli/build.rs
+++ b/crates/cli/build.rs
@@ -335,7 +335,7 @@ fn get_full_path_within_manifest_dir(relative_path: &Path, manifest_dir: &Path) 
 /// `full_path` and `repo_root` should both be canonical paths, as by [`Path::canonicalize`].
 fn make_repo_root_relative(full_path: &Path, repo_root: &Path) -> PathBuf {
     full_path
-        .strip_prefix(&repo_root)
+        .strip_prefix(repo_root)
         .map(|p| p.to_path_buf())
         .unwrap_or_else(|_| {
             panic!(


cloutiertyler (approved):
This seems fine to me. The CI will also have tested that it can build on non-nix. I am going to approve this, but yeah I think we should rip all this out at some point.

Issue #3435, opened 2025-10-19
[Web] useTable fails on camelCase


KingCosmic:
it seems like useTable is failing on anything related to camelCase if i try and filter by colums that are `snake_cased` in my lib.rs and so `camelCased` in my js code. the filter never gets applied it just sits in loading. and the same thing happens if i try to do a `snake_cased` table name. I'll do some tracking down when i get the time but wanted to post this here

cloutiertyler:
Hi @KingCosmic Thank you for filing this issue. I am looking into it as we speak!

KingCosmic:
I jsut implemented a fix on my end.

just need to convert from camelCase to snakeCase in the query builder

KingCosmic:
https://gist.github.com/KingCosmic/af5620e001f34e7f62d19007378ceebf

the relevant code I changed.

KingCosmic:
@cloutiertyler just wanted to give you a tag to make sure you see the above. might help you out if you haven't got the chance to check it yourself yet.

Lethalchip:
> https://gist.github.com/KingCosmic/af5620e001f34e7f62d19007378ceebf
> 
> the relevant code I changed.

But does this actually fix anything? What if your tables are actually CamelCase and not snake_case?

Feels like we need a mechanism to pull the table name from the bindings, and then insert that instead, versus making assumptions on casing?

Unsure how to do that in a clean way, but that feels like the proper fix here.

KingCosmic:
> > https://gist.github.com/KingCosmic/af5620e001f34e7f62d19007378ceebf
> > 
> > the relevant code I changed.
> 
> But does this actually fix anything? What if your tables are actually CamelCase and not snake_case?

My rust module complained if I tried to make something camelCased so I assumed it wasn't possible to make properties camelCased or are we only referring to table names here?

Issue #3431, opened 2025-10-18
[Web] useSpacetimeDB hook does not cause rerenders


KingCosmic:
The `useSpacetimeDB` hook does not seem to cause rerenders.

so I was trying to check the .identity property of the connection to see when I was connected before moving on

```ts
const connection = useSpacetimeDB();

  if (!connection.identity) {
    console.log('No Identity yet')
    return <div>Connecting to Server...</div>;
  }
```

This is because I was planning to use the .identity property for some table queries and needed it loaded but that's besides the point.

with the above code in my component and then a log inside of the .onConnect() handler of my connection builder I get these logs.

<img width="760" height="102" alt="Image" src="https://github.com/user-attachments/assets/131b2267-5eeb-44cd-94b9-c6a866f263da" />

where after the connection log the component never rerenders to show the new state.
however if I have HMR update after the connection is made the reload will show that .identity is set.

I've rema```de the context provider and useSpacetime hook to fix this for my usecase and am sharing the setup in some gists here.

[gists are here](https://gist.github.com/KingCosmic/3712c9df111fd57e805e6e2292757cf3)

I'm not making a pr because I would like to fix this provider to expect the same format (which is possible I'm just new to the sdk and didn't know I could tie into the events better.)

The new provider params look like.
```tsx
const onConnect = (conn: DbConnection, identity: Identity, token: string) => {
  localStorage.setItem('auth_token', token);
  console.log(
    'Connected to SpacetimeDB with identity:',
    identity.toHexString()
  );

  conn.reducers.onSendMessage(() => {
    console.log('Message sent.');
  });
};

const onDisconnect = () => {
  console.log('Disconnected from SpacetimeDB');
};

const onConnectError = (_ctx: ErrorContext, err: Error) => {
  console.log('Error connecting to SpacetimeDB:', err);
};

const connectionBuilder = DbConnection.builder()
  .withUri('ws://localhost:3000')
  .withModuleName('slg')
  .withToken(localStorage.getItem('auth_token') || undefined)

  // Don't add these manually, they are passed as props to our custom provider
  // .onConnect(onConnect)
  // .onDisconnect(onDisconnect)
  // .onConnectError(onConnectError);

render(
  <SpacetimeDBProvider connectionBuilder={connectionBuilder} onConnect={onConnect} onDisconnect={onDisconnect} onConnectError={onConnectError}>
    <App />
  </SpacetimeDBProvider>, document.getElementById('root')!
);
```

[...]

Issue #3365, opened 2025-10-07
Typescript sdk: SpacetimeDBProvider does not clean itself up and reloads/exiting doesn't disconnect properly


Lethalchip:
User `Strike72` on discord encountered an issue where his react client doesn't correctly trigger their `ClientDisconnected` reducer: https://discord.com/channels/1037340874172014652/1424993383302041612

I guessed that this user was using the new react-hooks and took a look at the `SpacetimeDBProvider`. 🕵️ 

Right now the `SpacetimeDBProvider` builds the connection immediately, but never tears it down, so when the user refreshes or navigates away, the websocket stays open until it dies from inactivity or the browser closes. This leads to users not properly disconnecting from SpacetimeDB and frustration. 

To clean up properly, the connection should be built inside a `useEffect` inside the provider itself, and call `disconnect()` in the cleanup function. Here is a drop-in replacement of the `SpacetimeDBProvider.ts` file that does exactly this: 

```tsx
import React from 'react';
import {
  DbConnectionBuilder,
  type DbConnectionImpl,
} from '../sdk/db_connection_impl';
import { SpacetimeDBContext } from './useSpacetimeDB';

export interface SpacetimeDBProviderProps<
  DbConnection extends DbConnectionImpl,
  ErrorContext,
  SubscriptionEventContext,
> {
  connectionBuilder: DbConnectionBuilder<
    DbConnection,
    ErrorContext,
    SubscriptionEventContext
  >;
  children?: React.ReactNode;
}

export function SpacetimeDBProvider<
  DbConnection extends DbConnectionImpl,
  ErrorContext,
  SubscriptionEventContext,
>({
  connectionBuilder,
  children,
}: SpacetimeDBProviderProps<
  DbConnection,
  ErrorContext,
  SubscriptionEventContext
>): React.JSX.Element {
  const [connection, setConnection] = React.useState<DbConnection | null>(null);

  React.useEffect(() => {
    const conn = connectionBuilder.build();
    setConnection(conn);

    return () => {
      try {
        conn.disconnect();
      } catch (err) {
        console.warn('[SpacetimeDBProvider] Error during disconnect:', err);
      }
    };
  }, [connectionBuilder]);

  if (!connection) return React.createElement(React.Fragment, null);

  return React.createElement(
    SpacetimeDBContext.Provider,
    { value: connection },
    children
  );
}
```

[...]

PR #3559, opened 2025-11-03
Unifies server module library and client SDK for TypeScript (and fixes several bugs)

cloutiertyler:
# Description of Changes

This PR is a very large change to the workings of the TypeScript SDK and as such requires a higher bar of testing than other PRs. However, it does several important things:

1. Unifies the API of the server and client so they not only have the same API, but they actually implement it with the same TypeScript types. This fixes several inconsistencies between them and fixes several small bugs as well.
2. Closes https://github.com/clockworklabs/SpacetimeDB/issues/3365
3. Closes https://github.com/clockworklabs/SpacetimeDB/issues/3431
4. Closes https://github.com/clockworklabs/SpacetimeDB/issues/3435
5. Subsumes the work done in https://github.com/clockworklabs/SpacetimeDB/pull/3447
6. Derives all type information on the client from a single `RemoteModule` type which vastly cleans up the correctness of type checking on the client and helped me to find several small bugs

It accomplishes this by changing code generation of TypeScript on the client to code generation approximately what a developer would manually write in their module. The ultimate goal would be to allow the developer to use the types and functions that they define on in their module directly on the client without needing to do any code generation at all, provided they are using TypeScript on the server and client.

https://github.com/clockworklabs/SpacetimeDB/issues/3365 is resolved by `.build()`ing the `DbConnection` inside a React `useEffect` rather than doing it directly in line with the render of the provider. In order to do that we needed to not expose the `DbConnection` directly to developers by returning a different type from `useSpacetimeDB`. `useSpacetimeDB` now returns a `ConnectionState` object which is stored as React state and updates when any of the fields change. This change also resolves https://github.com/clockworklabs/SpacetimeDB/issues/3431.

https://github.com/clockworklabs/SpacetimeDB/issues/3435 was the issue that initially lead me down the rabbit hole of unifying the server and the client because it was nearly impossible to track down all the various type functions and how they connect to the values that we code generate on the server. After several hours of attempting this, I decided to clean up the types a bit to be more uniform.

Implementing the unification between the client and the server also necessitated fully implemented parts of the API that were fully implemented on the server, but were broken or missing on the client.

[Unification] -> Means that this is breaking behavior for the client SDK, but that the new behavior is identical to the server's existing behavior

## Breaking changes: 
- Table accessor names and index accessor names are converted to camelCase on the `ctx`, so `ctx.db.foo_bar` is now `ctx.db.fooBar`

- [Unification] On the client `my_table.iter()` returns `IterableIterator` instead of an `Array`
- [Unification] `module_bindings` now export `TypeBuilder`s for all types instead of a `type MyType` and object `MyType`, so instead of using `MyType` as a type directly, you need to infer the type `MyType` -> `Infer<typeof MyType>`.
- [Unification] We no longer generate and export `MyTypeVariants` for sum types (these are now accessed by `Infer<typeof MyType.variants.myVariant>`)
- [Unification] `MyType.getTypeScriptAlgebraicType()` has been replaced with `MyType.algebraicType`

- `useSpacetimeDB()` no longer takes type parameters
- `useTable()` now takes a `TableDef` parameter and type params are inferred
- `useTable()` now just returns an `Array` directly instead of a object with `{ rows }`

- [Unification] `ctx.reducers.createPlayer(argA, argB)` -> `ctx.reducers.createPlayer({ argA, argB })`
- [Unification] `ctx.reducers.onCreatePlayer(ctx, argA, argB)` -> `ctx.reducers.onCreatePlayer(ctx, { argA, argB })`
- [Unification] `ctx.reducers.removeOnCreatePlayer(ctx, argA, argB)` -> `ctx.reducers.removeOnCreatePlayer(ctx, { argA, argB })`
- [Unification] `myTable.count(): number` -> `myTable.count(): bigint`
 
## Additive changes:
- `Infer<>` now also does `InferTypeOfRow<>` if applicable
- Added a `useReducer()` React hook
- `module_bindings` now exports a `tables` object with references to all the `TableDef`s
- `module_bindings` now exports a `reducers` object with references to all the `ReducerDef`s
- Added a new `MyType.create('MyVariant', ...)` function in addition to the `MyType.MyVariant(...)` constructors (this is private)

## Notable things that did not change:
- `MyType.serialize(writer: BinaryWriter, value: Infer<typeof MyType>)` and `MyType.deserialize(reader: BinaryReader): Infer<typeof MyType>` are still supported exactly as before.
- The `MyType.MyVariant(...)` constructor function on sum types is still present, but implemented with the private `MyType.create('MyVariant', ...)`. We could choose to move away from this API later if we didn't like the variants polluting the namespace

4 - This is a deep reaching an complex change for the SDK. For the server, it is much less deep reaching since it reuses much of the same machinery, although it does require thorough testing there as some of the code was modified.

This change is fully localized to TypeScript and does not touch the host (or other languages) at all, and therefore only impacts a beta aspect of SpacetimeDB.

# Testing

- [ ] Added regression test for https://github.com/clockworklabs/SpacetimeDB/issues/3435
- [x] Manually tested `test-app` and `test-react-router-app`
- [ ] Add test cases for camelCase-ing

### DIFF
--- a/.github/workflows/discord-posts.yml
+++ b/.github/workflows/discord-posts.yml
@@ -45,7 +45,7 @@ jobs:
           author_discord_id="$(
             jq -r \
               --arg u "${{ github.event.pull_request.user.login }}" \
-              'try .[$u] catch empty' \
+              '.[$u] // empty' \
               <<<"${DISCORD_USER_MAP}"
           )"
           if [ -z "${author_discord_id}" ]; then


cloutiertyler:
Notes from PR review meeting:

- Indexes are still not camelCase
- Modify `useTable` to return tuple of `[rows, loading]`

coolreader18 (approved):

PR #3550, opened 2025-10-31
Bump versions to 1.7.0

bfops:
# Description of Changes

Bump versions to 1.7.0 in preparation for the release.

# API and ABI breaking changes

# Expected complexity level and risk

1

# Testing

- [x] CI passes

PR #3562, opened 2025-11-03
Set default server to `local` in replication smoketests

gefjon:
Discussion out-of-band.


bfops:
rebasing this onto https://github.com/clockworklabs/SpacetimeDB/pull/3550 in order to make tests pass

PR #3570, opened 2025-11-03
View resolution in sql

joshua-spacetime:
# Description of Changes

Not many changes were required for the query compiler to be able to resolve views. This is because the query engine can always assume a view is materialized and therefore has a backing table. So from the perspective of the query engine, a view is just another table with one small caveat: The physical table in the datastore has two internal metadata columns - `sender` and `arg_id`. These columns are not user facing and so should be hidden from name resolution/type checking.

None

1.5

# Testing

- [x] SQL type checking tests


Shubham8287 (approved):
LGTM, given that we have separate ticket for parsing input params for Views.

PR #3571, opened 2025-11-04
CI - Fix the merge queue

bfops:
# Description of Changes

The merge queue was (partly) getting borked because we were putting all non-PR CI events into the same concurrency group, which meant they all non-PR CI jobs would run sequentially instead of running in parallel. This sometimes caused _painfully_ long delays in the merge queue.

This was due to my misunderstanding in https://github.com/clockworklabs/SpacetimeDB/pull/3501#discussion_r2466570395, where I didn't realize that `cancel-in-progress: false` would cause everything to queue up.

Now, for non-PR events, we append the commit SHA to the concurrency group. For merge queue events, this should be the SHA of the ephemeral merge commit that GH creates, so it will never conflict. For push events or manual workflow dispatch events, the SHA should be a sane way to recognize/cancel redundant events.

None. CI-only change.

1

# Testing

- [x] PR CI passes on this PR
- [x] PR CI is still canceled on this PR if a new commit is pushed

Unfortunately it's hard to test the behavior for non-PR events without merging and seeing if it works.


.github/workflows/ci.yml
```diff
@@ -14,10 +14,8 @@ on:
 name: CI
 
 concurrency:
-  # When a PR number isn't available, the event won't be a `pull_request` event so it won't matter.
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
-  # Only cancel when the event is a pull_request
-  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.event.inputs.pr_number || format('sha-%s', github.sha) }}
```
bfops, line 17: FYI - the `pr_number` variable is defined as a custom workflow input further up in this file.. None of the other workflows modified below have it.

PR #3574, opened 2025-11-04
Disconnect clients when updating RLS rules

mamcx:
# Description of Changes

Disconnects clients when adding, removing, or updating RLS rules. Client are not disconnected if an auto-migration does not touch any RLS rules.

> Why do we need to disconnect clients at all?

Subscriptions cache RLS rules. By disconnecting clients, we forcibly evict stale invalid RLS rules from the cache.

Bug fix. Previously subscriptions would keep stale RLS rules in cache.

1

# Testing

- [x] Adding unit and smoke tests
- [x] Manual inspection with `spacetime subscribe`

### DIFF
--- a/smoketests/tests/rls.py
+++ b/smoketests/tests/rls.py
@@ -97,8 +97,6 @@ pub struct Users {
 pub fn add_user(ctx: &ReducerContext, name: String) {
     ctx.db.users().insert(Users { name, identity: ctx.sender });
 }
-
-// RLS
 """
     
     ADD_RLS = """ 


crates/schema/src/auto_migrate.rs
```diff
@@ -1019,13 +1019,26 @@ fn auto_migrate_constraints(plan: &mut AutoMigratePlan, new_tables: &HashSet<&Id
...
+        old_rls.insert(rls.key());
         plan.steps.push(AutoMigrateStep::RemoveRowLevelSecurity(rls.key()));
     }
     for rls in plan.new.row_level_security() {
+        new_rls.insert(rls.key());
         plan.steps.push(AutoMigrateStep::AddRowLevelSecurity(rls.key()));
     }
 
+    // After changing an RLS rule and updating the module, the updated rule does not take effect for clients until they reconnect.
+    // This is because subscriptions are cached and we don't flush the cache when RLS rules are updated.
```
joshua-spacetime: This part of the comment is from the issue. It's no longer true with this patch.

crates/schema/src/auto_migrate.rs
```diff
@@ -1019,13 +1019,26 @@ fn auto_migrate_constraints(plan: &mut AutoMigratePlan, new_tables: &HashSet<&Id
 // Because we can refer to many tables and fields on the row level-security query, we need to remove all of them,
 // then add the new ones, instead of trying to track the graph of dependencies.
 fn auto_migrate_row_level_security(plan: &mut AutoMigratePlan) -> Result<()> {
+    // Track if any RLS rules were changed.
+    let mut old_rls = HashSet::new();
+    let mut new_rls = HashSet::new();
+
     for rls in plan.old.row_level_security() {
+        old_rls.insert(rls.key());
         plan.steps.push(AutoMigrateStep::RemoveRowLevelSecurity(rls.key()));
```
joshua-spacetime, line 1037: Currently all RLS rules are explicitly removed and re-added, even if they didn't change. Could we catch the case where there are no changes to RLS? Otherwise if a module defines a single RLS rule, we will always disconnect clients on an auto-migration.

mamcx: This is detected and test and manual inspection proves there is not disconnect if the RLS are not modified in a migration

smoketests/tests/auto_migration.py
```diff
@@ -146,14 +146,15 @@ def test_add_table_auto_migration(self):
         )
 
         self.write_module_code(self.MODULE_CODE_UPDATED)
-        self.publish_module(self.database_identity, clear=False)
+        self.publish_module(self.database_identity, clear=False, break_clients=True)
```
joshua-spacetime: Can you remove the RLS rules from this test now that we have a separate smoke test for them? It's important that publishing the module in this test doesn't break clients.

mamcx: Done

smoketests/tests/rls.py
```diff
@@ -78,3 +80,88 @@ def test_publish_fails_for_rls_on_private_table(self):
...
+    name: String,
+    identity: Identity,
+}
+
+#[spacetimedb::reducer]
+pub fn add_user(ctx: &ReducerContext, name: String) {
+    ctx.db.users().insert(Users { name, identity: ctx.sender });
+}
+
+// RLS
```
joshua-spacetime:
```suggestion
```

smoketests/tests/rls.py
```diff
@@ -78,3 +80,88 @@ def test_publish_fails_for_rls_on_private_table(self):
...
+"""
+    
+    ADD_RLS = """ 
+    #[spacetimedb::client_visibility_filter]
```
joshua-spacetime:
```suggestion
#[spacetimedb::client_visibility_filter]
```

smoketests/tests/rls.py
```diff
@@ -78,3 +80,88 @@ def test_publish_fails_for_rls_on_private_table(self):
...
+        )
+        
+        logging.info("Re-publish with RLS complete")
+        
+        logs = self.logs(100)
+        
+        # Validate disconnect + schema migration logs
+        self.assertIn("Disconnecting all users", logs)
+    
+    def test_rls_disconnect_no(self):
```
joshua-spacetime:
```suggestion
    def test_rls_no_disconnect(self):
```

PR #3577, opened 2025-11-04
Remove the middle man thread the JS worker thread

Centril:
# Description of Changes

This makes us go from 3 threads to 2.
The next step is to core pin the V8 worker thread.

None

4

# Testing

Existing tests should cover this.

### DIFF
--- a/crates/core/src/host/v8/mod.rs
+++ b/crates/core/src/host/v8/mod.rs
@@ -43,7 +43,6 @@ use spacetimedb_lib::{ConnectionId, Identity, RawModuleDef, Timestamp};
 use spacetimedb_schema::auto_migrate::MigrationPolicy;
 use std::sync::{Arc, LazyLock};
 use std::time::Instant;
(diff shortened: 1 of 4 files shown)


crates/core/src/host/v8/mod.rs
```diff
@@ -257,44 +261,123 @@ impl JsInstance {
...
 
-    pub fn update_database(
-        &mut self,
-        program: Program,
-        old_module_info: Arc<ModuleInfo>,
-        policy: MigrationPolicy,
-    ) -> anyhow::Result<UpdateDatabaseResult> {
+    /// Send a request to the worker and wait for a reply.
+    async fn send_recv<T>(
+        mut self: Box<Self>,
```
coolreader18, line 266: Why doesn't this take `&mut self`? It seems to just complicate the code in `ModuleHost`.

Centril: I tried that, and it resulted in a bunch of "lifetime may not life long enough" errors.

PR #3578, opened 2025-11-05
Add view handling to query engine and planner

joshua-spacetime:
# Description of Changes

This patch does the following:

1. Expands views as part of query planning. Views are always assumed to be materialized by the query planner, however a view's backing table may have private columns such as the `sender` column. The query planner needs to filter by this column in order to select the rows pertaining to a particular caller.
2. Plumbs `AuthCtx` through the query optimizer. This is needed in order to implement (1).
3. Adds a new operator for views to the query engine that drops a view's private columns

None

2.5

# Testing

- [x] SQL http tests
- [ ] Subscription tests
- [ ] One off query tests

### DIFF
--- a/crates/execution/src/pipelined.rs
+++ b/crates/execution/src/pipelined.rs
@@ -11,7 +11,7 @@ use spacetimedb_physical_plan::plan::{
     HashJoin, IxJoin, IxScan, PhysicalExpr, PhysicalPlan, ProjectField, ProjectListPlan, ProjectPlan, Sarg, Semi,
     TableScan, TupleField,
 };
-use spacetimedb_primitives::{ColId, IndexId, TableId};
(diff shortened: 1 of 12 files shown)


crates/execution/src/pipelined.rs
```diff
@@ -109,6 +152,53 @@ impl ProjectListExecutor {
...
 
+/// An executor for a query that returns rows from a view.
+/// Essentially just a projection that drops the view's metadata columns.
+#[derive(Debug)]
+pub struct ViewProject {
+    num_cols: usize,
+    num_private_cols: usize,
+    inner: PipelinedProject,
+}
+
```
mamcx, line 170: I like a comment here that explain the deal with `num_private_cols`, that is used for filter out the injected ones, also a link to `TableSchema::from_view_def_for_datastore` in both directions that this  need to hold

PR #3585, opened 2025-11-05
C# bindings for views

rekhoff:
# Description of Changes

Updates C# bindings to allow module authors to define `View`s and `AnonymousView`s using the pattern:
```csharp
[SpacetimeDB.View]
public static ExampleData? GetExampleDataById(ViewContext ctx, uint id)
{
    return ctx.Db.ExampleData.Id.Find(id);
}
	
[SpacetimeDB.View]
public static ExampleData? GetAnonymousExampleDataById(AnonymousViewContext ctx, uint id)
{
    return ctx.Db.ExampleData.Id.Find(id);
}
```
During publishing of a C# module, the views data will be converted into `RawViewDefV9`

No known breaking changes

2

# Testing

- [X] Locally tested locally adding the above sample pattern to the `sdks\csharp\examples~\regression-tests\server` test and performing a `dotnet build` and a `spacetime publish test`, both of which succeeded.


sdks/csharp/examples~/regression-tests/server/Lib.cs
```diff
@@ -15,6 +15,18 @@ public partial struct ExampleData
         public uint Indexed;
     }
 
+    [SpacetimeDB.View]
```
joshua-spacetime: Note: Views must have an explicit name.

JasonAtClockwork: I've added missing diag errors during codegen to force explicit name and Public = true:

\sdks\csharp\examples~\regression-tests\server\Lib.cs(20,5): error STDB0021: View 'GetExampleDataById' must have an explicit name. [\sdks\csharp\examples~\regression-tests\server\StdbModule.csproj]
\sdks\csharp\examples~\regression-tests\server\Lib.cs(20,5): error STDB0022: View 'GetExampleDataById' must have Public = true. Views are always public in SpacetimeDB. [\sdks\csharp\examples~\regression-tests\server\StdbModule.csproj]

joshua-spacetime:
@JasonAtClockwork could you also disable parameters for now? We can re-enable once we have sql support for them.

JasonAtClockwork:
> @JasonAtClockwork could you also disable parameters for now? We can re-enable once we have sql support for them.

Added another temporary error that will block arguments beyond the context:
\sdks\csharp\examples~\regression-tests\server\Lib.cs(18,5): error STDB0023: View 'GetExampleDataById' must have no arguments beyond the context. This is a temporary limitation.

joshua-spacetime (commented):
@JasonAtClockwork Can you try to copy over some of the negative compile tests that are in `bindings/tests/ui/views.rs` into `bindings-csharp/Codegen.Tests/fixtures/diag/Lib.cs`? You don't have to copy all of them over. I mainly just want to test that compilation fails if we try to access the wrong methods on `ViewContext` and `AnonymousViewContext`, or return the wrong type.

JasonAtClockwork:
> @JasonAtClockwork Can you try to copy over some of the negative compile tests that are in `bindings/tests/ui/views.rs` into `bindings-csharp/Codegen.Tests/fixtures/diag/Lib.cs`? You don't have to copy all of them over. I mainly just want to test that compilation fails if we try to access the wrong methods on `ViewContext` and `AnonymousViewContext`, or return the wrong type.

Copied most of the tests and fixed an issue with missing return type checks in the codegen

joshua-spacetime (approved):
I reviewed the tests and the runtime, but I did not review the codegen. I feel confident approving based on the test coverage, although I will outline a few more tests that would be useful to add before merging:

1. Can't obtain an identity from an AnonymousViewContext
2. Can't call `Iter()` on a table handle obtained from a read only context
3. Can't call mutations like delete on an index handle obtained from a read only context

JasonAtClockwork:
> I reviewed the tests and the runtime, but I did not review the codegen. I feel confident approving based on the test coverage, although I will outline a few more tests that would be useful to add before merging:
> 
> 1. Can't obtain an identity from an AnonymousViewContext
> 2. Can't call `Iter()` on a table handle obtained from a read only context
> 3. Can't call mutations like delete on an index handle obtained from a read only context

Turns out Iter() was available on a ReadOnly table, and Delete() was available on a ReadOnly Index but they're now fixed. I've also added the three new tests so it might be another look before we merge just in case.

crates/bindings-csharp/Codegen.Tests/fixtures/diag/Lib.cs
```diff
@@ -1,4 +1,5 @@
 using System.ComponentModel;
+using System.Security.Cryptography.X509Certificates;
```
joshua-spacetime: Did you mean to include this?

joshua-spacetime (approved):
Looks great! Thanks @JasonAtClockwork

crates/bindings-csharp/Codegen.Tests/fixtures/diag/Lib.cs
JasonAtClockwork: No.. my VS Code seems to add this "for" me I'll fix that up and confirm other files

cloutiertyler:
All tests are passing. I'm force merging this one.

PR #3586, opened 2025-11-06
Materialize views for sql http calls

joshua-spacetime:
# Description of Changes

Materialize views if necessary for sql queries. Also update the subscriber count and `last_called` timestamp in `st_view_sub`.

None

1

# Testing

TODO

### DIFF
--- a/crates/core/src/host/module_host.rs
+++ b/crates/core/src/host/module_host.rs
@@ -39,17 +39,19 @@ use spacetimedb_data_structures::error_stream::ErrorStream;
 use spacetimedb_data_structures::map::{HashCollectionExt as _, IntMap};
 use spacetimedb_datastore::error::DatastoreError;
 use spacetimedb_datastore::execution_context::{ExecutionContext, ReducerContext, Workload, WorkloadType};
-use spacetimedb_datastore::locking_tx_datastore::MutTxId;
+use spacetimedb_datastore::locking_tx_datastore::datastore::TxMetrics;
+use spacetimedb_datastore::locking_tx_datastore::{MutTxId, TxId};
 use spacetimedb_datastore::system_tables::{ST_CLIENT_ID, ST_CONNECTION_CREDENTIALS_ID, ST_VIEW_SUB_ID};
 use spacetimedb_datastore::traits::{IsolationLevel, Program, TxData};
 use spacetimedb_durability::DurableOffset;
 use spacetimedb_execution::pipelined::{PipelinedProject, ViewProject};
+use spacetimedb_expr::expr::CollectViews;
 use spacetimedb_lib::db::raw_def::v9::Lifecycle;
 use spacetimedb_lib::identity::{AuthCtx, RequestId};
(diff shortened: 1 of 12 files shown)


crates/core/src/sql/execute.rs
```diff
@@ -195,14 +197,27 @@ pub fn run(
...
     match stmt {
         Statement::Select(stmt) => {
+            // Before we downgrade to a read-only transaction,
+            // check if there are any views that need to be materialized.
+            let mut view_ids = HashSet::new();
+            stmt.collect_views(&mut view_ids);
+
+            let sender = auth.caller;
+            for view_id in view_ids {
+                if !tx.is_view_materialized(view_id, ArgId::SENTINEL, sender)? {
```
joshua-spacetime: Can use `st_view_sub` table or read sets.

crates/core/src/host/module_host.rs
```diff
@@ -39,25 +39,27 @@ use spacetimedb_data_structures::error_stream::ErrorStream;
...
+use spacetimedb_primitives::{ArgId, ProcedureId, TableId, ViewDatabaseId, ViewId};
 use spacetimedb_query::compile_subscription;
 use spacetimedb_sats::{AlgebraicTypeRef, ProductValue};
 use spacetimedb_schema::auto_migrate::{AutoMigrateError, MigrationPolicy};
 use spacetimedb_schema::def::deserialize::ArgsSeed;
 use spacetimedb_schema::def::{ModuleDef, ProcedureDef, ReducerDef, TableDef, ViewDef};
 use spacetimedb_schema::schema::{Schema, TableSchema};
 use spacetimedb_vm::relation::RelValue;
-use std::collections::VecDeque;
+use std::collections::{HashSet, VecDeque};
```
Centril, line 62: This shouldn't use `HashSet` from libstd.

crates/expr/src/expr.rs
```diff
@@ -1,10 +1,28 @@
-use std::sync::Arc;
+use std::{collections::HashSet, sync::Arc};
```
Centril, line 1: ditto

crates/physical-plan/src/plan.rs
```diff
@@ -240,6 +252,29 @@ pub enum PhysicalPlan {
...
+                    views.insert(info.view_id);
+                }
+            }
+            Self::IxJoin(join, _) => {
+                if let Some(info) = &join.rhs.view_info {
+                    views.insert(info.view_id);
+                }
+            }
+            _ => {}
+        });
```
Centril, line 274:
```suggestion
        self.visit(&mut |plan| {
            let view_info = match plan {
                Self::TableScan(scan, _) => &scan.schema.view_info,
                Self::IxScan(scan, _) => &scan.schema.view_info,
                Self::IxJoin(join, _) => &join.rhs.view_info,
                _ => return,
            };
            if let Some(info) = view_info {
                views.insert(info.view_id);
            }
        });
```

PR #3589, opened 2025-11-06
Fix for multi-column indexes in typescript modules

jsdt:
# Description of Changes

This updates a bunch of types to make sure we generate the right functions for indexes with multiple columns. Most of the changes are making the array of column names readonly.

This also has a fix for `AllUnique`, because we were treating composite indexes as unique when the first field had a unique index. FWIW, we should also make sure that the range scanning functions still exist in that case, but I didn't try to add that here.

Any change to these types is probably technically a breaking API change, but no one should consider these types stable.

1.5. I'm still not super confident with some of these types, but this shouldn't change any runtime behavior.

# Testing

I added some uses of indexes in `schema.test-d.ts`, so `pnpm run build:types` would throw errors if this were broken.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7517,6 +7517,7 @@ version = "1.7.0"
 dependencies = [
  "anyhow",
  "bigdecimal",
+ "bytes",
  "derive_more 0.99.20",
  "ethnum",
  "pretty_assertions",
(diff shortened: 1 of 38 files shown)


coolreader18 (approved):
I was hoping to avoid recursion, but if it doesn't work otherwise, this LGTM

PR #3596, opened 2025-11-06
fix: view index

Shubham8287:
# Description of Changes

`call_view` and `call_view_anon` maintains separate `Vec` list for view inside Module but `ModuleDef` used to have single `Vec`. Hence, Views index is not same across `ModuleDef` and  module itself.

Added a `index` field inside `ViewDef` which stores it index inside module.

NA

How complicated do you think these changes are? Grade on a scale from 1 to 5,
2, 

there could be some existing code relying on poistion in `Vec` and may not appear as compiler error

# Testing

Future smoketests should cover this.


joshua-spacetime (approved):
Thanks Shub!

PR #3599, opened 2025-11-06
Materialize views on subscribe

joshua-spacetime:
# Description of Changes

This patch:

1. Materializes views on subscribe and sql calls by invoking `call_view` on the `ModuleHost`.
2. Downgrades to a read-only transaction after view materialization but before query execution.
3. Updates the `st_view_sub` system table on both subscribe and unsubscribe.
4. Makes subscribe methods on the SubscriptionManager async.

None

2

# Testing

End-to-end tests to be added with atomic view updates

### DIFF
--- a/crates/core/src/host/module_host.rs
+++ b/crates/core/src/host/module_host.rs
@@ -39,8 +39,7 @@ use spacetimedb_data_structures::error_stream::ErrorStream;
 use spacetimedb_data_structures::map::{HashCollectionExt as _, IntMap};
 use spacetimedb_datastore::error::DatastoreError;
 use spacetimedb_datastore::execution_context::{ExecutionContext, ReducerContext, Workload, WorkloadType};
-use spacetimedb_datastore::locking_tx_datastore::datastore::TxMetrics;
-use spacetimedb_datastore::locking_tx_datastore::{MutTxId, TxId};
+use spacetimedb_datastore::locking_tx_datastore::MutTxId;
 use spacetimedb_datastore::system_tables::{ST_CLIENT_ID, ST_CONNECTION_CREDENTIALS_ID, ST_VIEW_SUB_ID};
 use spacetimedb_datastore::traits::{IsolationLevel, Program, TxData};
 use spacetimedb_durability::DurableOffset;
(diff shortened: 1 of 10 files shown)


crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -793,6 +808,9 @@ impl ModuleSubscriptions {
         // Record how long it took to compile the subscription
         drop(compile_timer);
 
+        let mut_tx = ScopeGuard::<MutTxId, _>::into_inner(mut_tx);
+        let (tx, tx_offset) = self.materialize_views_and_downgrade_tx(mut_tx, &queries, auth.caller)?;
```
Centril: Would be good to explain why we are doing this in the code.

crates/core/src/host/module_host.rs
```diff
@@ -1487,6 +1488,39 @@ impl ModuleHost {
...
+        view_collector: &impl CollectViews,
+        sender: Identity,
+        workload: Workload,
+    ) -> Result<MutTxId, ViewCallError> {
+        use FunctionArgs::*;
+        let mut view_ids = HashSet::new();
+        view_collector.collect_views(&mut view_ids);
+        for view_id in view_ids {
+            let name = tx.lookup_st_view(view_id)?.view_name;
+            if !tx.is_view_materialized(view_id, ArgId::SENTINEL, sender)? {
```
Shubham8287, line 1509: Right now, we have two implmentations to check of `materilized` view.
`is_view_materialized`: based on `st_view_sub`
and `is_materialized`: based on read sets.

We should decide to chose one approach, and If using `st_view_sub`, how would we know if view is stale?

joshua-spacetime: You're right that we can't use `st_view_sub` to check if the view is stale. For that we need to use the read sets. I'm using the system table for this just for simplicity, because it's just a single row lookup. But I'm not completely decided on it. I may switch to using read sets for everything in the next patch for incremental update, in which case I'll remove `is_view_materialized` altogether.

Shubham8287 (approved):
LGTM,we should also do the code cleanup to remove `module_rx` field from `ModuleSubscription` as we decided to not go with that (works in next PR too).

PR #3601, opened 2025-11-07
Adds flag to publish to allow clearing on migration conflict.

cloutiertyler:
# Description of Changes

This PR modifies the `--delete-data` flag on `spacetime publish` and adds the `--delete-data` flag on `spacetime dev`.

In particular instead of `--delete-data` being a boolean, it is now a an enum:

- `always` -> corresponds to the old value of `true`
- `never` -> corresponds to the old value of `false`
- `on-conflict` -> clears the database, but only if publishing would have required a manual migration

This flag does NOT change any behavior about prompting users to confirm if they want to delete the data. Users will still be prompted to confirm UNLESS they pass the separate `--yes` flag.

`spacetime dev` gets the same `--delete-data` flag. The default value of `never` is equivalent to the existing behavior. `spacetime dev` continues to publish with `--yes` just as before. This behavior is unchanged.

Adds the flags specified above. This is NOT a breaking change to the CLI. Passing `--delete-data` is the equivalent of `--delete-data=always`.

This IS technically a breaking change to the `pre_publish` route. As far as I'm aware this is *only* used by our CLI however.

> IMPORTANT SIDE NOTE: I would argue that `--break-clients` should really be renamed to `--yes-break-clients` because it actually behaves like the `--yes` force flag, but only for a subset of the user prompts. I have not made this change because it would be a breaking change, but if the reviewers agree, I will make this change.

2, Very small change, but if we get it wrong users could accidentally lose data. I would ask reviewers to think about ways that users might accidentally pass `--delete-data --yes`.

# Testing

- [ ] I have not yet tested manually.


crates/cli/src/common_args.rs
```diff
@@ -37,3 +44,16 @@ pub fn confirmed() -> Arg {
...
         .help("Instruct the server to deliver only updates of confirmed transactions")
 }
+
+pub fn clear_database() -> Arg {
+    Arg::new("clear-database")
+        .long("delete-data")
+        .short('c')
+        .num_args(0..=1)
+        .value_parser(value_parser!(ClearMode))
+        .require_equals(true)
```
bfops, line 58: I'm curious about the UX choice behind `require_equals` for this arg but not others

crates/cli/src/subcommands/publish.rs
```diff
@@ -70,7 +67,7 @@ pub fn cli() -> clap::Command {
             Arg::new("break_clients")
                 .long("break-clients")
                 .action(SetTrue)
-                .help("Allow breaking changes when publishing to an existing database identity. This will break existing clients.")
+                .help("Allow breaking changes when publishing to an existing database identity. This will force publish even if it will break existing clients, but will NOT force publish if it would cause deletion of any data in the database. See --yes and --delete-data for details.")
```
bfops, line 71: nit: are we sure we want to encourage them to look at `--yes`?

crates/cli/src/subcommands/publish.rs
```diff
@@ -162,7 +162,7 @@ pub async fn exec(mut config: Config, args: &ArgMatches) -> Result<(), anyhow::E
 
         let mut builder = client.put(format!("{database_host}/v1/database/{domain}"));
 
-        if !clear_database {
+        if !(matches!(clear_database, ClearMode::Always)) {
```
bfops: nit: I don't know that `matches!` is more readable than `==` for simple enum values 🤷

crates/cli/src/subcommands/publish.rs
```diff
@@ -303,22 +307,37 @@ async fn apply_pre_publish_if_needed(
...
+        match pre {
+            PrePublishResult::ManualMigrate(manual) => {
+                if matches!(clear_database, ClearMode::OnConflict) {
+                    println!("{}", manual.reason);
+                    println!("Proceeding with database clear due to --delete-data=on-conflict.");
+                }
+                if matches!(clear_database, ClearMode::Never) {
+                    println!("{}", manual.reason);
+                    println!("Aborting publish due to required manual migration.");
+                    anyhow::bail!("Publishing aborted by user");
```
bfops: is this a correct bail message? Sure the user is ultimately responsible, but isn't it really just that they _didn't_ choose to pass `--clear-data`? that feels like the user _not_ doing something.

crates/cli/src/subcommands/publish.rs
```diff
@@ -303,22 +307,37 @@ async fn apply_pre_publish_if_needed(
...
+                }
```
bfops, line 365: nit: should we maybe assert that clear_database isn't `ClearMode::Always`, or something similar? I feel like it takes me an extra pass or two reading this code to remember that the _caller_ has chosen not to call this in that case.

crates/cli/src/common_args.rs
cloutiertyler: Because it's ambiguous and a little confusing otherwise:

```sh
spacetime publish --delete-data always foobar # Does this delete a database called "always" or does it always delete data?
```

If we didn't do this you couldn't use the tool with a database named `always` `on-conflict` or `never`. In practice that's probably not a problem, but since `requires_equals` can always be relaxed, I thought we'd start by requiring it.

crates/cli/src/subcommands/publish.rs
cloutiertyler: Oh can you do that? I think I didn't realize that.

crates/cli/src/subcommands/publish.rs
cloutiertyler: I mean I think so, many people might otherwise be looking for how they can do it regardless.

crates/cli/src/subcommands/publish.rs
cloutiertyler: Oh, yes it is. Good catch. Will fix.

crates/cli/src/subcommands/publish.rs
cloutiertyler: Yeah, good point. Will fix.

crates/cli/src/common_args.rs
bfops:
```suggestion
        // Because we have a default value for this flag, invocations can be ambiguous between
        //passing a value to this flag, vs using the default value and passing an anonymous arg
        // to the rest of the command. Adding `require_equals` resolves this ambiguity.
        .require_equals(true)
```

crates/cli/src/subcommands/publish.rs
cloutiertyler: Fixed

cloutiertyler:
Fixed all the comments!

crates/cli/src/subcommands/publish.rs
```diff
@@ -175,7 +175,7 @@ pub async fn exec(mut config: Config, args: &ArgMatches) -> Result<(), anyhow::E
         let domain = percent_encoding::percent_encode(name_or_identity.as_bytes(), encode_set);
         let mut builder = client.put(format!("{database_host}/v1/database/{domain}"));
 
-        if !clear_database {
+        if !(clear_database == ClearMode::Always) {
```
bfops: nit:
```suggestion
        if clear_database != ClearMode::Always {
```

crates/cli/src/subcommands/publish.rs
bfops: was this done?

bfops:
Smoketests failing due to:
```
Publishing module...
Updated database with name: etabxtydzldgaoqcnhcq, identity: c2004f38704bab11f9ba1c355adcdcd04f935040df8af34d65ede783b176c9ef
2025-11-11 18:48:39,464 - DEBUG - --------------

2025-11-11 18:48:39,464 - DEBUG - $ spacetime --config-path /tmp/tmpry9gvhzm/config.toml publish etabxtydzldgaoqcnhcq --project-path /tmp/tmpry9gvhzm --yes
2025-11-11 18:48:40,621 - DEBUG - --- stderr ---
   Compiling spacetime-module v0.1.0 (/tmp/tmpry9gvhzm)
    Finished `release` profile [optimized] target(s) in 0.32s
Error: Aborting because publishing would require manual migration or deletion of data and --delete-data was not specified.
2025-11-11 18:48:40,621 - DEBUG - --- stdout ---
Build finished successfully.
Uploading to localhost => http://127.0.0.1:3000/
Checking for breaking changes...
Errors occurred:
Adding a column age to table person requires a default value annotation
```

crates/cli/src/subcommands/publish.rs
cloutiertyler: This is now done below. All we do is print because the behavior is the same as on-conflict in this case.

bfops:
Disabling auto-merge since I was only approving my code-owned changes.

crates/cli/src/subcommands/publish.rs
bfops: FWIW I will have the same comment in the `AutoMigrate` branch below. We have a comment that says that we have not specified `ClearMode::Always`, but there's no assert or anything to ensure that that's the case. It's fine though, feel free to just resolve.

PR #3602, opened 2025-11-07
remove module watcher from subscription

Shubham8287:
# Description of Changes

Problem is visible in replication smoketests, which exit and re-lauch modules, erorring out with:
```
ERROR /app/public/crates/core/src/host/host_controller.rs:818: Failed to open database: DatabaseError: Database is already opened. Path:`/stdb/data/replicas/4`. Error:Resource temporarily unavailable (os error 11).: Database is already opened. Path:`/stdb/data/replicas/4`. Error:Resource temporarily unavailable (os error 11)
```

Commented code will be a blocker for releasing Views but we want CIs to pass.

### DIFF
--- a/crates/core/src/host/host_controller.rs
+++ b/crates/core/src/host/host_controller.rs
@@ -886,7 +886,11 @@ impl Host {
         let disk_metrics_recorder_task = tokio::spawn(metric_reporter(replica_ctx.clone())).abort_handle();
 
         let module = watch::Sender::new(module_host);
-        replica_ctx.subscriptions.init(module.subscribe());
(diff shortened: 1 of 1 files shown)


kim:
I can confirm that it fixes the issue, but it remains unclear why the receiver is not dropped.

Shubham8287:
ModuleSubscription::module_rx: OnceLock<watch::Receiver<ModuleHost>>,


Here reciever holds the recent instance of `ModuleHost` and which internally again contains the `ModuleSubscriptions`, so drop is not being called due to circular dependency.

I am fixing it for View in separate PR.

PR #3605, opened 2025-11-07
Refactor module instance code

coolreader18:
# Description of Changes

The main thing this does is implement `WasmInstance` for `V8Instance`, so we can now change all the `vm_call*: impl FnOnce()` args on `InstanceCommon` to instead just take `&mut impl WasmInstance`. (Obviously our javascript instance is not a wasm instance, but `wasmtime::module_host_actor` is desperately in need of a rebrand in general, so I just left `WasmInstance`'s as is.)
 
This vastly simplifies the callsites, and also allows @joshua-spacetime to call views from inside of `call_reducer_with_tx`, which was the main motivating factor behind this.

If the last commit is a bit much, I'm happy to split it off - I just don't get explicitly asked to refactor the rough edges of the codebase very often :3

2 - it's a refactor, but overall pretty straightforward, and Rust lets you do fearless refactoring etc etc

# Testing

n/a - internal code change

### DIFF
--- a/crates/core/src/host/v8/mod.rs
+++ b/crates/core/src/host/v8/mod.rs
@@ -619,7 +619,7 @@ struct V8Instance<'a, 'scope, 'isolate> {
 
 impl WasmInstance for V8Instance<'_, '_, '_> {
     fn extract_descriptions(&mut self) -> Result<RawModuleDef, DescribeError> {
-        extract_description(&mut self.scope, self.hooks, self.replica_ctx)
+        extract_description(self.scope, self.hooks, self.replica_ctx)
     }
 
     fn replica_ctx(&self) -> &Arc<ReplicaContext> {


crates/core/src/host/wasm_common/module_host_actor.rs
```diff
@@ -60,21 +63,27 @@ pub trait WasmInstancePre: Send + Sync + 'static {
     fn instantiate(&self, env: InstanceEnv, func_names: &FuncNames) -> Result<Self::Instance, InitializationError>;
 }
 
-#[async_trait::async_trait]
-pub trait WasmInstance: Send + Sync + 'static {
-    fn extract_descriptions(&mut self) -> Result<Vec<u8>, DescribeError>;
+// TODO: rename and move
```
joshua-spacetime:
```suggestion
// TODO: Technically this trait is also used for V8.
// We should rename and move to some place more appropriate.
```

joshua-spacetime (approved):

PR #3607, opened 2025-11-08
Implemented Some-typed literals in SQL

egormanga:
# Description of Changes

Ability to, given a table:
```rust
#[table(name=table)]
struct Table {
	option: Option<i32>,
}
```
do:
```sql
INSERT INTO table (option) VALUES (0);
```
instead of getting:
```log
Error: The literal expression `0` cannot be parsed as type `(some: I32 | none: ())`
```

* Companion pull request for #3606.

None.

### 1:
Quite literally almost a one-liner.

# Testing

- [x] Manual testing
- [ ] Unit tests

### DIFF
--- a/crates/expr/src/lib.rs
+++ b/crates/expr/src/lib.rs
@@ -241,6 +241,9 @@ pub(crate) fn parse(value: &str, ty: &AlgebraicType) -> anyhow::Result<Algebraic
             .map(AlgebraicValue::from)
             .map_err(|err| anyhow!(err))
     };
+    let to_option = || {
+        Ok(AlgebraicValue::OptionSome(parse(value, ty.as_option().unwrap()).map_err(|_| InvalidLiteral::new(value.to_string(), ty))?))
+    };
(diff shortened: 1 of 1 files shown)


cloutiertyler:
@egormanga Is this how we should implement `Option`s? How would you represent `Option<Option<u32>>` as a SQL string?

egormanga:
I don't think nested options is a realistic case. Option here is a direct translation of SQL's nullable columns, and there is no nested nullability that I know of.

cloutiertyler:
It's certainly something that is definable in SpacetimeDB, even if it's not possible in ANSI SQL. We'll need to define how it should work/translate to ANSI SQL. I think it's fine if SpacetimeDB SQL is a superset of ANSI SQL, but we'll want to make sure we're still within the spec.

egormanga:
I think there at least should be some support for the simpler case, i.e. this PR. More verbose syntax for complex nested structures can come later as an edge case.

PR #3611, opened 2025-11-10
[teams 1/5] Reset database

kim:
So far, the `--clear-database` option to `publish` has simply dropped
and then re-created the database (if it did exist).

This will no longer work when databases can have "children": because
dropping and re-creating is not atomic, children would either become
orphans, or be dropped as well.

To solve this, `reset_database` is introduced as a separate action that:

- shuts down all replicas
- if a `program_bytes` is supplied, replaces the database's initial
  program
- if a `host_type` is supplied, replaces the database's host type
- starts `num_replicas` or the previous number of replicas, which
  initialize themselves as normal

As this could be its own CLI command, the action is provided as its own
API endpoint (undocumented).

However, since `publish` has no way of knowing whether the database it
operates on actually exists, the `publish_database` handler will just
invoke the `reset_database` handler if `clear=true` and the database
exists, and return its result. This is to avoid starting the transfer of
the program in the request body, only to receive a redirect.

Some refactoring was necessary to dissect and understand the flow.

Introduces a new, undocumented API endpoint.
We may want to nest it under `/unstable`.

2

# Testing

From the outside, the observed behavior should be as before,
so smoketests should cover it.


crates/client-api/src/routes/database.rs
```diff
@@ -24,20 +26,43 @@ use spacetimedb::host::UpdateDatabaseResult;
...
 };
 
 use super::subscribe::{handle_websocket, HasWebSocketOptions};
 
+fn require_spacetime_auth_for_creation() -> bool {
+    env::var("TEMP_REQUIRE_SPACETIME_AUTH").is_ok_and(|v| !v.is_empty())
+}
+
+// A hacky function to let us restrict database creation on maincloud.
+fn allow_creation(auth: &SpacetimeAuth) -> Result<(), ErrorResponse> {
```
cloutiertyler, line 48: This code is just moving from elsewhere correct?

crates/client-api/src/routes/database.rs
```diff
@@ -596,40 +676,13 @@ pub struct PublishDatabaseQueryParams {
...
 
-use spacetimedb_client_api_messages::http::SqlStmtResult;
-use std::env;
-
-fn require_spacetime_auth_for_creation() -> bool {
-    env::var("TEMP_REQUIRE_SPACETIME_AUTH").is_ok_and(|v| !v.is_empty())
-}
-
-// A hacky function to let us restrict database creation on maincloud.
-fn allow_creation(auth: &SpacetimeAuth) -> Result<(), ErrorResponse> {
```
cloutiertyler, line 614: Ah yes it is

cloutiertyler (approved):
I did a deeper review of this one this time around and it looks good to me. It's an elegant solution. Basically just detaches the old replicas from the database and attaches new replicas.

PR #3613, opened 2025-11-10
Add `spacetime_worker_reducer_returned_errors_total` metric

### LINKED ISSUE #3507: Add error rate + panic rate metrics to prometheus
We want to be able to track the rate of errors and panics per minute coming from a module.

Needs:
- database name
- reducer name/procedure name

drogus:
# Description of Changes

Currently, we have a metric for reducer panics called `spacetime_worker_wasm_instance_errors_total`. This commit adds a metric for tracking errors returned from the module, like for example an Err result in Rust, or throwing a SenderError in TypeScript

This will be needed for the web database overview dashboard, but it should be useful for tracking sender errors in general.

1

# Testing

- [x] I've tested the change locally verifying the counter increments when an instance returns an error


crates/core/src/worker_metrics/mod.rs
```diff
@@ -246,6 +246,11 @@ metrics_group!(
         #[labels(caller_identity: Identity, module_hash: Hash, caller_connection_id: ConnectionId, reducer_symbol: str)]
         pub wasm_instance_errors: IntCounterVec,
 
+        #[name = spacetime_worker_wasm_instance_sender_errors_total]
```
cloutiertyler: please rename to `spacetime_worker_sender_errors_total`

crates/core/src/worker_metrics/mod.rs
```diff
@@ -246,6 +246,11 @@ metrics_group!(
...
+        #[help = "The number of sender errors returned from reducers."]
+        #[labels(database_identity: Identity, module_hash: Hash, reducer_symbol: str)]
+        pub wasm_instance_sender_errors: IntCounterVec,
```
cloutiertyler: Please rename to `sender_errors`

cloutiertyler (changes requested):
Just think we should change/simplify the naming to not reference wasm

drogus:
> Just think we should change/simplify the naming to not reference wasm

Yeah, that makes sense. I wanted to go with the current convention, but we'll eventually probably rename it, too

cloutiertyler (approved):

PR #3614, opened 2025-11-10
Change `spacetime_worker_wasm_instance_errors_total` metric's labels

### LINKED ISSUE #3507: Add error rate + panic rate metrics to prometheus

drogus:
# Description of Changes

There are two purposes for introducing the changes in this commit:

1. Decrease the cardinality of the metric, as the caller id and connection id cwould result in way too combiations
2. Add a database_identity label, which is much more useful for the metric

I'm honestly not sure if metrics labels change is an API breaking change, I suppose not? For what it's worth, I don't think that anyone was relying on the labels I've removed.

1

# Testing

- [x] I've tested the change locally


cloutiertyler (approved):
I think this makes a lot of sense.

PR #3615, opened 2025-11-10
Views: fix table schema creation from system table.

Shubham8287:
# Description of Changes
`schema_for_table_raw` was hardcoded to return `None` for `ViewInfo`.
This issue was not observable when the module launched for the first time, but during module re-launches, it caused problems with `impl CollectViews`, as the implementation was treating view tables as normal tables.

NA

1, pretty obvious.

### DIFF
--- a/crates/datastore/src/locking_tx_datastore/state_view.rs
+++ b/crates/datastore/src/locking_tx_datastore/state_view.rs
@@ -121,26 +121,29 @@ pub trait StateView {
             })
             .transpose()?;
 
+        // Look up the view info for the table in question, if any.
         let view_info: Option<ViewInfo> = self
             .iter_by_col_eq(
                 ST_VIEW_ID,
                 StViewFields::TableId,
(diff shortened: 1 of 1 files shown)


crates/datastore/src/locking_tx_datastore/state_view.rs
```diff
@@ -120,10 +121,32 @@ pub trait StateView {
             })
             .transpose()?;
 
+        let view_info: Option<ViewInfo> = self
+            .iter_by_col_eq(
+                ST_VIEW_ID,
+                StViewFields::TableId,
```
joshua-spacetime, line 128: We should probably think about adding an index to this field.

Shubham8287: We can actually avoid indexing, `schema_for_table_raw` is called rarely. Most of the times, schema is served from cache.

PR #3616, opened 2025-11-10
SQL smoketests for views

Shubham8287:
# Description of Changes

This patch tests calling, updating, and materialization of views through the SQL api.

None

1.5

# Testing

Smoketests

### DIFF
--- a/smoketests/tests/views.py
+++ b/smoketests/tests/views.py
@@ -186,7 +186,9 @@ INSERT INTO player_state (id, level) VALUES (42, 7);
  7  | 3
 """)
 
-    def test_view_materialization(self):
+    # test is prefixed with 'a' to ensure it runs before any other tests,
+    # since it relies on log capturing starting from an empty log.
+    def test_a_view_materialization(self):
         """This test asserts whether views are materialized correctly"""
         self.insert_initial_data()
         player_called_log = "player view called"


smoketests/tests/views.py
```diff
@@ -101,3 +100,113 @@ def test_fail_publish_wrong_return_type(self):
...
+        self.assertNotIn(player_called_log, self.logs(100))
+
+        self.call_player_view()
+        #On first call, the view is evaluated
+        self.assertIn(player_called_log, self.logs(100))
+    
+        self.call_player_view()
+        #On second call, the view is cached
+        logs = self.logs(100)
+        self.assertEqual(logs.count(player_called_log), 1)
```
Shubham8287, line 205: Smoketests failing because:
We are checking if `materialized` based on `st_view_subs` table in master, and http sql call does not make an entry there hence each call maks a View to re-evalulate.

I wrote this test, assuming we will be using read-sets to check for materializaition, which I feel is also better but I have to update the test if we don't end up using that.

joshua-spacetime: There were two reasons why this was failing:
1. Atomic view update was not implemented for sql dml
2. Write sets were checked against the transaction's read set rather than the committed read set

Both of these are fixed in https://github.com/clockworklabs/SpacetimeDB/pull/3624

PR #3624, opened 2025-11-10
Atomic view update

joshua-spacetime:
# Description of Changes

Updates views atomically on commit, but before downgrading to a read-only transaction for subscription evaluation.

What this patch does:
1. Renames `ViewId` to `ViewFnPtr`
2. Renames `ViewDatabaseId` to `ViewId`
3. Removes the `module_rx` module watcher from the subscription manager
4. Refactors read sets to only track table scans (index key tracking will be added later)
5. Drops read sets and removes rows from `st_view_sub` when dropping a view in an auto-migrate
6. Re-evaluates and updates views (`call_views_with_tx`) from `call_reducer_with_tx` for any view whose read set overlaps with the reducer's write set
7. Does the same for sql dml

None

3

It's a bit of a messy diff.

# Testing

- [x] Integrate with https://github.com/clockworklabs/SpacetimeDB/pull/3616

### DIFF
--- a/crates/datastore/src/locking_tx_datastore/committed_state.rs
+++ b/crates/datastore/src/locking_tx_datastore/committed_state.rs
@@ -6,10 +6,6 @@ use super::{
     tx_state::{IndexIdMap, PendingSchemaChange, TxState},
     IterByColEqTx,
 };
-use crate::system_tables::{
-    ST_CONNECTION_CREDENTIALS_ID, ST_CONNECTION_CREDENTIALS_IDX, ST_VIEW_COLUMN_ID, ST_VIEW_COLUMN_IDX, ST_VIEW_ID,
-    ST_VIEW_IDX, ST_VIEW_PARAM_ID, ST_VIEW_PARAM_IDX, ST_VIEW_SUB_ID, ST_VIEW_SUB_IDX,
-};
 use crate::{
     db_metrics::DB_METRICS,
     error::{DatastoreError, IndexError, TableError},
@@ -24,6 +20,13 @@ use crate::{
         ST_VAR_IDX, ST_VIEW_ARG_ID, ST_VIEW_ARG_IDX,
     },
     traits::TxData,
+};
+use crate::{
+    locking_tx_datastore::ViewCallInfo,
+    system_tables::{
+        ST_CONNECTION_CREDENTIALS_ID, ST_CONNECTION_CREDENTIALS_IDX, ST_VIEW_COLUMN_ID, ST_VIEW_COLUMN_IDX, ST_VIEW_ID,
+        ST_VIEW_IDX, ST_VIEW_PARAM_ID, ST_VIEW_PARAM_IDX, ST_VIEW_SUB_ID, ST_VIEW_SUB_IDX,
+    },
 };
 use anyhow::anyhow;
 use core::{convert::Infallible, ops::RangeBounds};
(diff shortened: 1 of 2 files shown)


crates/primitives/src/ids.rs
```diff
@@ -138,9 +139,15 @@ system_id! {
 }
 
 system_id! {
-    /// the index of a view as defined in a module's view list.
+    /// The index of a view as defined in a module's view lists.
+    ///
+    /// Unline reducers and procedures, the module maintains two lists for views.
```
Shubham8287:
```suggestion
    /// Unlike reducers and procedures, the module maintains two lists for views.
```

crates/core/src/host/module_host.rs
```diff
@@ -539,18 +540,15 @@ pub struct CallReducerParams {
...
-    pub timestamp: Timestamp,
-    pub caller_identity: Identity,
-    pub caller_connection_id: Option<ConnectionId>,
+    pub view_name: Box<str>,
     pub view_id: ViewId,
-    pub view_db_id: ViewDatabaseId,
+    pub table_id: TableId,
+    pub fn_ptr: ViewFnPtr,
+    pub caller: Identity,
+    pub sender: Option<Identity>,
```
Shubham8287, line 551: why `sender` field is there along with `caller`?

crates/core/src/db/relational_db.rs
```diff
@@ -1509,93 +1507,116 @@ impl RelationalDB {
...
-        let deleted_count = self.delete(tx, table_id, rows_to_delete);
-        trace!("Deleted {deleted_count} stale rows from view table {table_id} for arg_id {arg_id}");
+            .collect::<Vec<_>>();
+        self.delete(tx, table_id, rows_to_delete);
 
-        // Deserialize the return value
+        // Deserialize the return rows
         let seed = spacetimedb_sats::WithTypespace::new(typespace, &return_type).resolve(return_type);
-        let return_val = seed
+        let rows = seed
```
Shubham8287, line 1548: FYI: this deserialization code doesn't work and requires: https://github.com/clockworklabs/SpacetimeDB/pull/3616.

crates/core/src/host/wasm_common/module_host_actor.rs
```diff
@@ -626,6 +632,27 @@ impl InstanceCommon {
             }
         };
 
+        // Only re-evaluate and update views if the reducer's execution was successful
+        let (out, trapped) = if !trapped && !matches!(status, EventStatus::Committed(_)) {
```
Shubham8287:
```suggestion
        let (out, trapped) = if !trapped && matches!(status, EventStatus::Committed(_)) {
```
?

joshua-spacetime: Yes thank you.

crates/core/src/host/module_host.rs
joshua-spacetime: If this is a sql call, `sender` will be the `caller`. But if we are calling a view after executing a reducer, `sender` will be the identity of a previously invoked view whereas `caller` will be whoever called the reducer.

Shubham8287: I see, sender value is what gets used inside View fn, and caller is for Host side operations. Maybe a comment will be helpful for readers.

PR #3625, opened 2025-11-10
Switch Internal Tests to reuse CI job

bfops:
# Description of Changes

Switch the `Internal Tests` job to re-use the existing CI workflow instead of a separate one.

None. CI only.

1

# Testing

- [x] Internal Tests passes on this PR


.github/workflows/internal-tests.yml
```diff
@@ -19,15 +19,16 @@ jobs:
...
         with:
           github-token: ${{ secrets.SPACETIMEDB_PRIVATE_TOKEN }}
           script: |
             const targetOwner = 'clockworklabs';
             const targetRepo = 'SpacetimeDBPrivate';
-            const workflowId = 'internal-tests.yml';
-            const targetRef = 'master';
+            const workflowId = 'ci.yml';
+            // TODO: Revert this before merging
+            const targetRef = 'bfops/merge-internal-tests';
```
bfops: TODO: revert this to `master` before merging

PR #3630, opened 2025-11-11
Typescript query builder

jsdt:
# Description of Changes

This introduces a query builder for typescript modules. It is available as `ReducerCtx.queryBuilder`, but it might make more sense to rename that to `schema`. There are some examples of how to build queries in `query.test-d.ts` and `query.test.ts`.

There is more type-safety to be added (ensuring `eq` is used on comparable types, and only allowing semijoins to use indexed columns), but those changes should only require changing some type parameters.

This is adding a new query builder API on the module side, which won't actually be useful until we can use them in views.

1.5. 

# Testing

There are tests of the emitted sql in `query.test.ts`, but at some point we should add end-to-end tests to make sure that we can parse those queries in the query engine.

### DIFF
--- a/crates/bindings-typescript/src/server/query.test-d.ts
+++ b/crates/bindings-typescript/src/server/query.test-d.ts
@@ -2,6 +2,7 @@ import type { U32 } from '../lib/autogen/algebraic_type_variants';
 import type { Indexes, UniqueIndex } from './indexes';
 import {
   eq,
+  from,
   literal,
   type RowExpr,
   type TableNames,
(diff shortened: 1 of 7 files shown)


crates/bindings-typescript/src/server/query.test-d.ts
```diff
@@ -0,0 +1,91 @@
...
+  indexes: order.idxs,
+};
+
+spacetimedb.init(ctx => {
+  const firstQuery = ctx.queryBuilder.query('person');
+  firstQuery.semijoinTo(
+    ctx.queryBuilder.order,
+    p => p.age,
+    o => o.item_name
+  );
```
joshua-spacetime, line 81: I thought we decided on
```ts
.semijoin(ctx.queryBuilder.order, (p, o) => eq(p.age, o.item_name));
```

cloutiertyler:
Closing since this has been implemented separately in another PR.

PR #3631, opened 2025-11-11
Update views on auto-migrate

Shubham8287:
# Description of Changes

Improves auto-migration support for views by minimizing the cases where we must disconnect clients.

Before this patch, any schema compatible view update and even no view updates at all would require us to disconnect clients, because we had to assume that the view was modified thereby rendering its currently materialized result set stale and out of date.

Patch adds a code to re-evaluate all views whose entry is in `st_view_sub` and make `update_database` to start using `commit_and_broadcast_event` so that now we only have to disconnect clients for incompatible view updates or dropped views.

NA 

2

# Testing

Added smoketests.

### DIFF
--- a/smoketests/tests/views.py
+++ b/smoketests/tests/views.py
@@ -285,6 +285,59 @@ pub fn player(ctx: &ViewContext) -> Option<PlayerState> {
 }
 """
 
+    def assertSql(self, sql, expected):
+        self.maxDiff = None
+        sql_out = self.spacetime("sql", self.database_identity, sql)
+        sql_out = "\n".join([line.rstrip() for line in sql_out.splitlines()])
+        expected = "\n".join([line.rstrip() for line in expected.splitlines()])
+        self.assertMultiLineEqual(sql_out, expected)
+
+    def test_views_auto_migration(self):
+        """Assert that views are auto-migrated correctly"""
+
+        self.spacetime(
+            "sql",
+            self.database_identity,
+            "INSERT INTO player_state (id, level) VALUES (1, 1);",
+        )
+        self.spacetime(
+            "sql",
+            self.database_identity,
+            "INSERT INTO player_state (id, level) VALUES (2, 2);",
+        )
+
+        self.assertSql("SELECT * FROM player", """\
+ id | level
+----+-------
+ 1  | 1
+""")
+
+        self.write_module_code(self.MODULE_CODE_UPDATED)
+        self.publish_module(self.database_identity, clear=False)
+
+        self.assertSql("SELECT * FROM player", """\
+ id | level
+----+-------
+ 2  | 2
+""")
+
+
+class AutoMigrateViewsTrapped(Smoketest):
+    MODULE_CODE = """
+use spacetimedb::ViewContext;
+#[derive(Copy, Clone)]
+#[spacetimedb::table(name = player_state)]
+pub struct PlayerState {
+    #[primary_key]
+    id: u64,
+    #[index(btree)]
+    level: u64,
+}
+#[spacetimedb::view(name = player, public)]
+pub fn player(ctx: &ViewContext) -> Option<PlayerState> {
+    ctx.db.player_state().id().find(1u64)
+}
+"""
 
     TRAPPED_MODULE_CODE_UPDATED = """
 use spacetimedb::ViewContext;
@@ -298,8 +351,25 @@ pub struct PlayerState {
     level: u64,
 }
 #[spacetimedb::view(name = player, public)]
-pub fn player(ctx: &ViewContext) -> Option<PlayerState> {
+pub fn player(_ctx: &ViewContext) -> Option<PlayerState> {
     panic!("This view is trapped")
+}
+"""
+
+    MODULE_CODE_RECOVERED = """
+use spacetimedb::ViewContext;
+
+#[derive(Copy, Clone)]
+#[spacetimedb::table(name = player_state)]
+pub struct PlayerState {
+    #[primary_key]
+    id: u64,
+    #[index(btree)]
+    level: u64,
+}
+#[spacetimedb::view(name = player, public)]
+pub fn player(ctx: &ViewContext) -> Option<PlayerState> {
+    ctx.db.player_state().id().find(2u64)
 }
 """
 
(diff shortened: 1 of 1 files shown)


crates/core/src/host/module_host.rs
```diff
@@ -189,6 +189,16 @@ pub struct ModuleFunctionCall {
     pub args: ArgsTuple,
 }
 
+impl ModuleFunctionCall {
+    pub fn update() -> Self {
+        Self {
+            reducer: String::new(),
```
joshua-spacetime: Should this be something like `String::from("update")`?

smoketests/tests/views.py
```diff
@@ -249,3 +249,144 @@ def test_query_anonymous_view_reducer(self):
...
+----+-------
+ 1  | 1
+""")
+
+        self.write_module_code(self.TRAPPED_MODULE_CODE_UPDATED)
+        with self.assertRaises(Exception):
+            self.publish_module(self.database_identity, clear = False)
+        logs = self.logs(100)
+        print("hey")
+        print(logs)
```
joshua-spacetime: Debug statements

crates/core/src/host/wasm_common/module_host_actor.rs
```diff
@@ -408,19 +409,101 @@ impl InstanceCommon {
...
+
+                if res.was_successful() {
+                    let event = ModuleEvent {
+                        timestamp: Timestamp::now(),
+                        caller_identity: self.info.owner_identity,
+                        caller_connection_id: None,
+                        function_call: ModuleFunctionCall::update(),
+                        status: EventStatus::Committed(DatabaseUpdate::default()),
+                        energy_quanta_used: FunctionBudget::ZERO.into(),
+                        host_execution_duration: Duration::ZERO,
```
joshua-spacetime: Technically these will be non-zero if we re-evaluated any views. You should be able to get them from the output of `evaluate_subscribed_views`.

crates/core/src/host/wasm_common/module_host_actor.rs
```diff
@@ -408,19 +409,101 @@ impl InstanceCommon {
...
+        }
+    }
+
+    /// Re-evaluates all views which have entries in `st_view_subs`.
+    fn evaluate_subscribed_views<I: WasmInstance>(
+        &mut self,
+        tx: MutTxId,
+        inst: &mut I,
+    ) -> Result<(ViewCallResult, bool), anyhow::Error> {
+        let views = self.info.module_def.views().cloned().collect::<Vec<_>>();
```
joshua-spacetime: Cloning all of the `ViewDef`s in the module seems rather expensive. Isn't it sufficient to just lookup the `ViewDef` for each view returned from `lookup_st_view_subs`?

Shubham8287: It doesn't required cloning. fixed.

crates/schema/src/auto_migrate.rs
```diff
@@ -604,24 +604,19 @@ fn auto_migrate_view<'def>(plan: &mut AutoMigratePlan<'def>, old: &'def ViewDef,
...
-    // } else {
-    //     plan.steps.push(AutoMigrateStep::UpdateView(old.key()));
-    // }
-
-    plan.steps.push(AutoMigrateStep::AddView(new.key()));
-    plan.steps.push(AutoMigrateStep::RemoveView(old.key()));
+    if old.is_anonymous != new.is_anonymous
+        || incompatible_return_type
+        || incompatible_param_types
+        || old.return_type != new.return_type
```
joshua-spacetime: Going from `Vec<T>` to `Option<T>` and vice versa shouldn't require client disconnects, right?

crates/schema/src/auto_migrate.rs
```diff
@@ -604,24 +604,19 @@ fn auto_migrate_view<'def>(plan: &mut AutoMigratePlan<'def>, old: &'def ViewDef,
...
+        plan.steps.push(AutoMigrateStep::AddView(new.key()));
+        plan.steps.push(AutoMigrateStep::RemoveView(old.key()));
 
-    if !plan.disconnects_all_users() {
-        plan.steps.push(AutoMigrateStep::DisconnectAllUsers);
+        if !plan.disconnects_all_users() {
+            plan.steps.push(AutoMigrateStep::DisconnectAllUsers);
+        }
+    } else {
+        plan.steps.push(AutoMigrateStep::UpdateView(old.key()));
```
joshua-spacetime:
```suggestion
        plan.steps.push(AutoMigrateStep::UpdateView(new.key()));
```
I think this was me. Probably should use the new key.

smoketests/tests/auto_migration.py
```diff
@@ -329,10 +329,11 @@ def test_add_table_columns(self):
             msg=f"Unexpected disconnect counts: {disconnect_count}",
         )
 
-        # Validate all subscribers received only single update before disconnect
+        # Validate all subscribers were disconnected after first upgrade
+        # they should 2 updates: one for initial insertion and one for table drop during migration
         for i in range(NUM_SUBSCRIBERS):
             sub = subs[i]()
-            self.assertEqual(len(sub), 1, msg=f"Subscriber {i} received unexpected rows: {sub}")
+            self.assertEqual(len(sub), 2, msg=f"Subscriber {i} received unexpected rows: {sub}")
```
joshua-spacetime, line 336: Why did this test change?

crates/schema/src/auto_migrate.rs
Shubham8287: Oh, Right :)

Shubham8287: pushed the fix and older test.

smoketests/tests/auto_migration.py
Shubham8287: This test ensures that client disconnection works properly when adding new columns in table.
Test flow is like: clients subscribes to table, a row is inserted, module updates and then more rows inserted in new table.
We used to assert here that client only recieved a single update (intial insert) and then disconnected properly.

Now, this PR calls `commit_and_broadcast` while updating database. Hence, clients are also receiving changes made into the table by update transaction itself (which is to drop the table for this case).

Hence subscription update count is increased to 2. Which shouldn't be a problem or unexpected behaviour ig?

joshua-spacetime: I see. So is the client is now receiving an update for the deleted row because dropping a table first clears the table?

Shubham8287: yes

Issue #3558, opened 2025-10-31
CI - TypeScript quickstart should use local package rather than npm version


bfops:
As it says on the label. If we make local changes, even something as simple as bumping a version number, the quickstart will fail.

Example: https://github.com/clockworklabs/SpacetimeDB/actions/runs/18980261737/job/54210737052

bfops:
From @cloutiertyler:
> FWIW this is a non-trivial change because this is also used as a template. Fortunately https://github.com/clockworklabs/SpacetimeDB/pull/3539 improves the scripts that would do this

jdetter:
From Tyler: We could set the version number to `1.*` instead of `1.7.*`, etc.

bfops:
I believe I have found a fix

PR #3633, opened 2025-11-11
Upgrade to version 1.8.0

jdetter:
# Description of Changes

- This upgrades the versions of all SDKs, the CLI, etc. to 1.8.0

None

1

# Testing

- [x] I verified that all versions seem to be updated including the BSL license update 

We have 1 `1.7.0` that didn't get upgraded automatically because it is part of the module bindings for a template:

```
crates/cli/.templates/parent_parent_crates_bindings-typescript_examples_quickstart-chat/src/module_bindings/index.ts:    cliVersion: '1.7.0',
```

A case could possibly be made for bumping the template but it shouldn't cause any issues as the module bindings directory should just get regenerated by the user. @cloutiertyler should we be bumping module bindings for templates when we upgrade versions?


bfops:
Sigh. Unfortunately smoketests fail due to https://github.com/clockworklabs/SpacetimeDB/issues/3558. (They will pass once the package is published to npm).

cloutiertyler (approved):

Issue #3515, opened 2025-10-28
Procedures: Add transaction API


gefjon:
Define new host functions for acquiring, committing, aborting a transaction. Expose these to module bindings as the following interface:

```rust
impl ProcedureContext {
    /// Acquire a mutable transaction and execute `body` with read-write access to the database.
    ///
    /// If the `body` returns `Ok`, the transaction will be committed and its mutations persisted.
    /// If the `body` returns `Err`, the transaction will be rolled back and its mutations discarded.
    ///
    /// Regardless of the transaction's success or failure,
    /// the return value of `body` is not persisted to the commitlog or broadcast to subscribed clients.
    /// Clients attribute mutations performed by this transaction to `Event::UnknownTransaction`.
    ///
    /// If the transaction fails to commit after `body` returns (e.g. due to a conflict with a concurrent transaction),
    /// this method will re-invoke `body` with a new transaction in order to retry.
    pub fn with_transaction<Res, Err>(&mut  self, body: impl Fn(&TxContext) -> Result<Res, Err>)) -> Result<Res, Err> {
        todo!()
    }
}
```

The `TxContext` is a lot like a `ReducerContext`: it has a `db` field which allows access to the database, offers the `rng` method, has a `timestamp`, and probably some other stuff I'm forgetting. It inherits auth information from the enclosing `ProcedureContext`.

Name the new host functions with the prefix `procedure_`.

Attempting to perform any of the other `procedure_` host calls while a transaction is open should error.

After completing this, create follow-up tickets to implement the same in C#, C++ and TypeScript.

PR #3638, opened 2025-11-11
Add `ProcedureContext::with_tx`

Centril:
# Description of Changes

Adds `ProcedureContext::{with_tx, try_with_tx}`.

Fixes https://github.com/clockworklabs/SpacetimeDB/issues/3515.

None

2

# Testing

An integration test `test_calling_with_tx` is added.

### DIFF
--- a/crates/bindings/src/lib.rs
+++ b/crates/bindings/src/lib.rs
@@ -1103,9 +1103,12 @@ impl ProcedureContext {
     /// Acquire a mutable transaction
     /// and execute `body` with read-write access to the database.
     ///
-    /// When a panic occurs,
-    /// the transaction will be rolled back and its mutations discarded.
+    /// If `body` panics, the transaction will be rolled back,
+    /// and the calling procedure terminated.
     /// Otherwise, the transaction will be committed and its mutations persisted.
+    /// Prefer calling [`Self::try_with_tx`]
+    /// and returning a `Result` for signaling errors
+    /// to allow the calling procedure to handle the transaction's failure.
     ///
     /// Regardless of the transaction's success or failure,
     /// the return value of `body` is not persisted to the commitlog
@@ -1135,10 +1138,15 @@ impl ProcedureContext {
     /// Acquire a mutable transaction
     /// and execute `body` with read-write access to the database.
     ///
-    /// When `body().is_ok()`,
+    /// When `body` returns `Ok`,
     /// the transaction will be committed and its mutations persisted.
-    /// When `!body().is_ok()` or a panic occurs in `body`,
+    /// When `body` returns `Err`,
     /// the transaction will be rolled back and its mutations discarded.
+    ///
+    /// If `body` panics, the transaction will be rolled back,
+    /// and the calling procedure terminated.
+    /// Prefer returning `Result` for signaling errors
+    /// to allow the calling procedure to handle the transaction's failure.
     ///
     /// Regardless of the transaction's success or failure,
     /// the return value of `body` is not persisted to the commitlog


crates/bindings-sys/src/lib.rs
```diff
@@ -661,8 +662,76 @@ pub mod raw {
...
+        ///
+        /// # Errors
+        ///
+        /// Returns an error:
+        ///
+        /// - `TRANSACTION_NOT_ANONYMOUS`,
+        ///   if the transaction was not started in [`procedure_start_mut_tx`].
+        ///   This can happen if this syscall is erroneously called by a reducer.
+        ///   The code `NOT_IN_TRANSACTION` does not happen,
+        ///   as it is subsumed by `TRANSACTION_NOT_ANONYMOUS`.
```
gefjon, line 706: I don't love this. It seems like we'd get better error messages if we could distinguish between these cases. On the other hand, it should be unreachable except through abusing the raw API, so 🤷 , whatever.

crates/bindings/src/lib.rs
```diff
@@ -1074,6 +1127,66 @@ impl ProcedureContext {
...
+    /// the return value of `body` is not persisted to the commitlog
+    /// or broadcast to subscribed clients.
+    /// Clients attribute mutations performed by this transaction to `Event::UnknownTransaction`.
+    ///
+    /// If the transaction fails to commit after `body` returns,
+    /// e.g., due to a conflict with a concurrent transaction,
+    /// this method will re-invoke `body` with a new transaction in order to retry.
+    /// This is done once. On the second failure, a panic will occur.
+    #[cfg(feature = "unstable")]
+    pub fn with_tx<R: IsOk>(&mut self, body: impl Fn(&TxContext) -> R) -> R {
```
coolreader18: I really think you could just require `body` to return a `Result`; they can always just put an `Ok(())` at the end or do `.ok_or(())`. For really wanting to support different return types it should probably just be 2 different functions, `with_tx` and `try_with_tx`, since `IsOk` can't support arbitrary return values anyway. `with_tx` could be implemented in terms of `try_with_tx::<Result<T, Infallible>>()`

crates/bindings/src/lib.rs
```diff
@@ -1074,6 +1127,66 @@ impl ProcedureContext {
...
+    ///
+    /// Regardless of the transaction's success or failure,
+    /// the return value of `body` is not persisted to the commitlog
+    /// or broadcast to subscribed clients.
+    /// Clients attribute mutations performed by this transaction to `Event::UnknownTransaction`.
+    ///
+    /// If the transaction fails to commit after `body` returns,
+    /// e.g., due to a conflict with a concurrent transaction,
+    /// this method will re-invoke `body` with a new transaction in order to retry.
+    /// This is done once. On the second failure, a panic will occur.
```
gefjon:
```suggestion
    /// The transaction will be retried at most once.
    /// If it fails to commit a second time, this method will panic.
```

crates/bindings/src/lib.rs
```diff
@@ -1074,6 +1127,66 @@ impl ProcedureContext {
...
+            );
+            let timestamp = Timestamp::from_micros_since_unix_epoch(timestamp);
+
+            // We've resumed, so do the work.
+            let tx = ReducerContext::new(Local {}, self.sender, self.connection_id, timestamp);
+            let tx = TxContext(tx);
+            body(&tx)
+        };
+
+        let mut res = run();
```
coolreader18, line 1203: Maybe put a scope guard here that calls `abort` on drop? Wasmtime does support unwinding now, and someone could compile with `panic=unwind` on nightly.

crates/bindings/src/lib.rs
gefjon: Can you add an additional paragraph with the operations you should not do in `body` in order to avoid determinism hazards? Something like:


```suggestion
    /// The transaction will be retried at most once.
    /// If it fails to commit a second time, this method will panic.
   ///
   /// Because `body` may be run multiple times,
   /// and is expected to perform the same set of database operations
   /// and return the same result on each invocation,
    /// callers should avoid writing to any captured mutable state within `body`.
    /// This includes interior mutability through types like [`std::cell::Cell`].
```

crates/bindings/src/lib.rs
```diff
@@ -1011,6 +1006,64 @@ impl ReducerContext {
...
+        true
+    }
+}
+
+#[cfg(feature = "unstable")]
+impl<T> IsOk for Option<T> {
+    fn is_ok(&self) -> bool {
+        self.is_some()
+    }
+}
```
gefjon: I think that using `Option<T>` here is likely to be confusing. I would be happier if we just required all transactions to return either `()` or a `Result`, at least for now.

crates/core/src/host/wasmtime/wasm_instance_env.rs
```diff
@@ -281,19 +294,39 @@ impl WasmInstanceEnv {
         (timings, self.take_standard_bytes_sink())
     }
 
+    /// Record a span with `start`.
+    fn end_span(mut caller: Caller<'_, Self>, start: CallSpanStart) {
```
gefjon:
```suggestion
    fn end_span(mut caller: Caller<'_, Self>, start: span::CallSpanStart) {
```

crates/core/src/host/wasmtime/wasm_instance_env.rs
```diff
@@ -2,12 +2,19 @@
 
 use super::{Mem, MemView, NullableMemOp, WasmError, WasmPointee, WasmPtr};
 use crate::database_logger::{BacktraceFrame, BacktraceProvider, ModuleBacktrace, Record};
+use crate::error::NodesError;
 use crate::host::instance_env::{ChunkPool, InstanceEnv};
+use crate::host::module_host::{DatabaseUpdate, EventStatus, ModuleEvent, ModuleFunctionCall};
+use crate::host::wasm_common::instrumentation::noop::CallSpanStart;
```
gefjon:
```suggestion
```

crates/bindings/src/lib.rs
```diff
@@ -1011,6 +1006,57 @@ impl ReducerContext {
...
+        true
+    }
+}
+
+#[cfg(feature = "unstable")]
+impl<T, E> IsOk for Result<T, E> {
+    fn is_ok(&self) -> bool {
+        self.is_ok()
+    }
+}
```
gefjon: Since we've switched to separate `with_tx` and `try_with_tx`, this should be unused now, right?

crates/bindings/src/lib.rs
```diff
@@ -1074,6 +1120,121 @@ impl ProcedureContext {
         let new_time = Timestamp::from_micros_since_unix_epoch(new_time);
         self.timestamp = new_time;
     }
+
+    /// Acquire a mutable transaction
+    /// and execute `body` with read-write access to the database.
+    ///
+    /// When a panic occurs,
+    /// the transaction will be rolled back and its mutations discarded.
+    /// Otherwise, the transaction will be committed and its mutations persisted.
```
gefjon: This paragraph needs some way to communicate that panicking in the `body` will abort the whole procedure, not just the transaction.

```suggestion
    /// If `body` panics, the transaction will be rolled back, and the calling procedure terminated.
    /// Prefer calling [`Self::try_with_tx`] and returning a `Result` for signaling errors 
    /// to allow the calling procedure to handle the transaction's failure.
```

crates/bindings/src/lib.rs
```diff
@@ -1074,6 +1120,121 @@ impl ProcedureContext {
...
+        }
+    }
+
+    /// Acquire a mutable transaction
+    /// and execute `body` with read-write access to the database.
+    ///
+    /// When `body().is_ok()`,
+    /// the transaction will be committed and its mutations persisted.
+    /// When `!body().is_ok()` or a panic occurs in `body`,
+    /// the transaction will be rolled back and its mutations discarded.
```
gefjon:
```suggestion
    /// When `body` returns `Ok,
    /// the transaction will be committed and its mutations persisted.
    /// When `body` returns `Err`,
    /// the transaction will be rolled back and its mutations discarded.
    ///
    /// If `body` panics, the transaction will be rolled back, and the calling procedure terminated.
    /// Prefer returning `Result` for signaling errors 
    /// to allow the calling procedure to handle the transaction's failure.
```

crates/bindings/src/lib.rs
Centril: Ooops, indeed.

PR #3648, opened 2025-11-12
C# release changes for `cargo release`

jdetter:
# Description of Changes

This adds the new "skeleton" directory that we use for `cargo release`.

None

1 - this is just a release change

# Testing

- [x] We used exactly this for the `1.8.0` and `1.9.0` release and we'll also be using it for the `1.10.0` release.


sdks/csharp/release~/spacetimedb.bsatn.runtime/unversioned/.gitignore
```diff
@@ -0,0 +1,8 @@
+.nupkg.metadata
+.signature.p7s
+LICENSE
+README.md
+logo.png
+spacetimedb.bsatn.runtime.*.nupkg*
+spacetimedb.bsatn.runtime.nuspec
+**/*.xml
```
bfops:
```suggestion
*.xml
```

sdks/csharp/release~/.gitignore
```diff
@@ -0,0 +1,5 @@
+# Ignore DLLs - these will be populated by dotnet restore
+/*/*/*/*/*.dll
```
bfops: does this need to be `/*/*/*/*/*`?
```suggestion
*.dll
```

bfops (changes requested):
Can you also remove the files from sdks/csharp/packages? They are stale since we're no longer updating them, and then I believe many of these files will show up as moved.

Also, I think we should probably `gitignore` them, so that devs iterating locally on `sdks/csharp` will not constantly have dirty working directories.

sdks/csharp/release~/packages/spacetimedb.bsatn.runtime/unversioned/.gitignore
```diff
@@ -0,0 +1,8 @@
+.nupkg.metadata
+.signature.p7s
+LICENSE
+README.md
+logo.png
+spacetimedb.bsatn.runtime.*.nupkg*
+spacetimedb.bsatn.runtime.nuspec
+*.xml
```
bfops, line 11: as an aside.. to we expect any of these files to actually appear down here?

jdetter: they do appear due to `dotnet restore`. Essentially these files end up in the `packages/spacetimedb.bsatn.runtime/vX.Y.Z` directory. During the csharp release we copy the unversioned directory on top of the directory that contains these files.

Right now when we do the C# release we are just doing a `git add` on the `packages` directory, we could manually `git add` each individual file instead but I think it's possible that might be more fragile? I'm not sure, I'm not highly invested into either approach here.

jdetter:
@bfops I removed everything under `sdks/csharp/packages/*` but I left the `packages` directory itself so that the `.meta` file can be maintained there. I could move the `packages` directory under `release~` so that it's `release~/packages/spacetimedb.bsatn...` what do you think?

crates/bindings-typescript/package.json
```diff
@@ -148,7 +148,7 @@
     {
       "name": "sdk esm min (uncompressed)",
       "path": "dist/min/sdk/index.browser.mjs",
-      "limit": "14 kB"
+      "limit": "30 kB"
```
bfops, line 151: unrelated? maybe a merge of `master` will fix this?

sdks/csharp/release~/packages/spacetimedb.bsatn.runtime/unversioned/.gitignore
bfops: ah that makes sense. Can we add something like this as a comment?

(that said, I'm surprised that things like `LICENSE`, `logo.png`, and `README.md` show up.. but it also doesn't matter too much)

bfops (approved):
This LGTM, minus my two comments

bfops (dismissed):
Actually, one other question - have you thought about the flow for devs making modifications to the Unity SDK? Does it still work smoothly?

My assumption would be that they end up with a version directly that keeps wanting to be added to `git`, so maybe we should `gitignore` any versioned directories, and temporarily remove that `.gitignore` file when we're doing the release?

I do want to make sure we don't make a bad flow for devs doing Unity SDK things, even if we don't do all of it in this PR. We may want an extra README and/or script to help facilitate the meta file copying.

sdks/csharp/release~/packages/spacetimedb.bsatn.runtime/unversioned/.gitignore
jdetter: I think the LICENSE, logo.png and README.md files show up because of how dotnet pack works, for some reason those files are included when packing 🤷 .

bfops (approved):
This LGTM as long as we build that utility command for doing the DLL dance soon (although I'm not the one who will be sad if it gets delayed)

PR #3649, opened 2025-11-12
[TS] Implement basic procedure calling

coolreader18:
# Description of Changes

Implements `__call_procedure__` in the TS bindings and host.

2

# Testing

- [ ] 
- [ ]

### DIFF
--- a/crates/bindings-typescript/src/lib/procedures.ts
+++ b/crates/bindings-typescript/src/lib/procedures.ts
@@ -3,7 +3,11 @@ import type { ConnectionId } from '../lib/connection_id';
 import type { Identity } from '../lib/identity';
 import type { Timestamp } from '../lib/timestamp';
 import type { ParamsObj } from './reducers';
-import { MODULE_DEF, type UntypedSchemaDef } from './schema';
+import {
+  MODULE_DEF,
+  registerTypesRecursively,
+  type UntypedSchemaDef,
+} from './schema';
 import type { Infer, InferTypeOfRow, TypeBuilder } from './type_builders';
 import { bsatnBaseSize } from './util';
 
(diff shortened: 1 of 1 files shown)


crates/core/src/host/wasm_common/module_host_actor.rs
```diff
@@ -635,7 +641,9 @@ impl InstanceCommon {
             self.allocated_memory = memory_allocation;
         }
 
-        match call_result {
+        let trapped = call_result.is_err();
```
Centril, line 638: Is this correct? Trapped means "drop instance" in this case, and for reducers, the logic is different.

crates/core/src/host/v8/mod.rs
```diff
@@ -781,8 +800,16 @@ impl WasmInstance for V8Instance<'_, '_, '_> {
...
-    async fn call_procedure(&mut self, _op: ProcedureOp, _budget: FunctionBudget) -> ProcedureExecuteResult {
-        todo!("JS/TS module procedure support")
+    async fn call_procedure(&mut self, op: ProcedureOp, budget: FunctionBudget) -> ProcedureExecuteResult {
+        common_call(self.scope, budget, op, |scope, op| {
+            call_call_procedure(scope, self.hooks, op)
+        })
+        .map_result(|call_result| {
+            call_result.map_err(|e| match e {
+                ExecutionError::User(e) => anyhow::Error::msg(e),
+                ExecutionError::Recoverable(e) | ExecutionError::Trap(e) => e,
```
Centril, line 811: In a previous PR of yours, in `common_call`, I think a bug was introduced where `CanContinue::YesCancelTermination => {` results in `ExecutionError::Trap` as opposed to `ExecutionError::Recoverable(_)`.

crates/core/src/host/wasm_common/module_host_actor.rs
coolreader18: Correct, but this behavior was already in master, and I couldn't figure out whether or not it was intentional and correct. I figure if we want to change it in a follow-up it'll be pretty easy.

jdetter:
The C#/Unity test failure here is a known problem in master, the PR that introduced that issue is being reverted so feel free to ignore

PR #3651, opened 2025-11-12
Views: cleanup unsubscribed views

### LINKED ISSUE #3587: Clear materialized view tables async
When clients unsubscribe or disconnect, we update the `st_view_sub` system table, but we don't clear the backing table or read sets. We should have an async task that does this cleanup periodically.

Shubham8287:
# Description of Changes

A background task to cleanup unsubscribed views.
fixes #3587 

NA

2

# Testing
Added a test

### DIFF
--- a/crates/core/src/db/relational_db.rs
+++ b/crates/core/src/db/relational_db.rs
@@ -1077,8 +1077,8 @@ impl RelationalDB {
 /// Value is chosen arbitrarily; can be tuned later if needed.
 const VIEWS_EXPIRATION: std::time::Duration = std::time::Duration::from_secs(10 * 60);
 
-/// Duration to budget for each view cleanup job, so that it doesn't hold database lock for too
-/// long.
+/// Duration to budget for each view cleanup job,
+/// so that it doesn't hold transaction lock for to long.
 //TODO: Make this value configurable
 const VIEW_CLEANUP_BUDGET: std::time::Duration = std::time::Duration::from_millis(10);
 


crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -1959,6 +1971,66 @@ impl MutTxId {
...
+
+        // For each expired view subscription, clear the backing table and delete the subscription
+        for (row, ptr) in expired_sub_rows {
+            let view_id = row.view_id;
+            let sender: Identity = row.identity.into();
+            // Get the view's backing table_id from st_view
+            let StViewRow {
+                table_id, is_anonymous, ..
+            } = self.lookup_st_view(view_id)?;
+            let table_id = table_id.unwrap();
```
Shubham8287:
```suggestion
            let table_id = table_id.expect("views have backing table");
```

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -221,7 +226,14 @@ impl MutTxId {
     /// Removes keys for `view_id` from the committed read set.
     /// Used for dropping views in an auto-migration.
     pub fn drop_view_from_committed_read_set(&mut self, view_id: ViewId) {
-        self.committed_state_write_lock.drop_view_from_read_sets(view_id)
+        self.committed_state_write_lock.drop_view_from_read_sets(view_id, None)
+    }
+
+    /// Removes a specific view call from the committed read set.
+    /// Used when dropping views due to
```
Shubham8287:
```suggestion
```

Shubham8287:
Test is better modularised in this PR - https://github.com/clockworklabs/SpacetimeDB/pull/3670.

joshua-spacetime (commented):
The approach makes sense. I just have a couple thoughts:

1. I think we should be more conservative in how frequent we clean up views. I don't think our clean up intervals need to be on the order of seconds.

2. Also we probably want to bound the length of time a cleanup job takes. If there are many old views, I think we'd rather clean them up in multiple jobs to avoid taking a tx lock for an extended period of time.

Shubham8287:
> 2. Also we probably want to bound the length of time a cleanup job takes. If there are many old views, I think we'd rather clean them up in multiple jobs to avoid taking a tx lock for an extended period of time.

Should it be limited by a hardcoded number (like 100 cleanups at a time) Or duration (atmost to take 5ms) or something else?

I feel limiting it to hardcoded number of views to cleanup can be a problem if continously more new views expires more than that number between job intervals.

joshua-spacetime:
I agree, I think duration is the right metric. And we can even make that duration configurable later (not a requirement for this PR).

PR #3656, opened 2025-11-13
Use new runner for testsuite

jdetter:
# Description of Changes

Speed improvements:

|CI Job|Before|After|Change|
|---|---|---|---|
|CI/Unreal Engine|29m|19m|34% faster|
|CI/Testsuite|21m|11m|47% faster|
|C#/Unity|20m|16m|20% faster|
|CI/Build and test wasm bindings|10m|3m|70% faster|
|CI/Smoketests (linux)|17m|12m|29% faster|

CI/Unreal Engine - switch to serial test execution
CI/Testsuite - partially switch to serial test execution

I also haven't seen an unexpected `Operation cancelled`, so I think that problem might be behind us.

No user facing changes

2 - this touches a lot of our CI. It has been solid so far but if we start getting random CI failures this could be the cause.

# Testing

- [x] All CI passes

(ok as of this update the internal tests are failing but that isn't related to this PR 😅 )

# Future Work

There are 2 remaining tests that could be optimized by moving them to faster runners:
1. (~45m) The windows smoketests (difficult because it requires Windows). @bfops has brought up the idea of only running this test on PRs that are in the merge queue and I think that would be the best 80/20 here. There really is not a great reason to be running this on every commit anyway.
2. (36m) The internal tests (easier because it's on Linux)

I would also consider moving from instantiating container images to instantiating VM snapshots. This would allow us a lot more flexibility and we wouldn't have to have as much container config in our workflow files. Based on guides I've seen online this is also somewhat common. Also managing docker within docker has been kind of a nightmare so it would be easier to run containers inside of VMs to remove a layer of containerization there.

## Testsuite inconsistencies

We have at least 2 different jobs that are inconsistent/flaky: the "testsuite" and the "unreal engine tests".

I'd like @JasonAtClockwork to make the unreal engine tests run sequentially so that we have less random failures. I know this will increase the testing time but due to this PR we should have extra headroom there.

I've also pinged Joshua about one of the tests that I've seen sometimes fails randomly. This is either an issue in the test or an issue in SpacetimeDB itself, I'm not sure yet.

These inconsistencies are present both in our custom runner and in the Github runners, so merging this PR will not make this situation any worse than it already is.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -157,7 +157,7 @@ jobs:
 
         # The fallocate tests have been flakely when running in parallel
       - name: Run fallocate tests
-        run: cargo test -p spacetimedb-durability --features fallocate --test-threads=1
+        run: cargo test -p spacetimedb-durability --features fallocate -- --test-threads=1
 
       - name: Check that the test outputs are up-to-date
         run: bash tools/check-diff.sh
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -94,6 +94,16 @@ jobs:
   test:
     name: Test Suite
     runs-on: spacetimedb-new-runner
+
+    # Add this for isolation:
+    container:
+      image: spacetimedb-ci:latest
+      options: >-
+        --user runner
+        -v /var/run/docker.sock:/var/run/docker.sock
+        -v /opt/github-runner/cargo-cache:/home/runner/.cargo/registry
+        -v /opt/github-runner/cargo-git:/home/runner/.cargo/git
+
     steps:
       - name: Find Git ref
         env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,7 +97,7 @@ jobs:
 
     # Add this for isolation:
     container:
-      image: spacetimedb-ci:latest
+      image: localhost/spacetimedb-ci:latest
       options: >-
         --user runner
         -v /var/run/docker.sock:/var/run/docker.sock
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -121,11 +121,10 @@ jobs:
         with:
           ref: ${{ env.GIT_REF }}
 
-      - uses: dsherret/rust-toolchain-file@v1
+      - name: Verify .NET
+        run: dotnet --version
 
-      - uses: actions/setup-dotnet@v3
-        with:
-          global-json-file: global.json
+      - uses: dsherret/rust-toolchain-file@v1
 
       - name: Set up Node.js
         uses: actions/setup-node@v4
@@ -135,11 +134,6 @@ jobs:
       - uses: pnpm/action-setup@v4
         with:
           run_install: true
-
-      - name: Create /stdb dir
-        run: |
-          sudo mkdir /stdb
-          sudo chmod 777 /stdb
 
       - name: Build typescript module sdk
         working-directory: crates/bindings-typescript
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -155,8 +155,9 @@ jobs:
         #Note: Unreal tests will be run separately
         run: cargo test --all -- --skip unreal
 
+        # The fallocate tests have been flakely when running in parallel
       - name: Run fallocate tests
-        run: cargo test -p spacetimedb-durability --features fallocate
+        run: cargo test -p spacetimedb-durability --features fallocate --test-threads=1
 
       - name: Check that the test outputs are up-to-date
         run: bash tools/check-diff.sh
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -63,6 +72,10 @@ jobs:
         if: runner.os == 'Windows'
         run: choco install psql -y --no-progress
         shell: powershell
+      - name: Start Docker daemon
+        if: runner.os == 'Linux'
+        run: /usr/local/bin/start-docker.sh
```
jdetter, line 76: We could simplify this a lot by just not running a docker container for the smoketest. I think it would probably be fine to use `spacetime start` here instead. We could even just start one instance of `spacetime start` per test for running them in parallel

.github/workflows/ci.yml
```diff
@@ -23,10 +23,17 @@ jobs:
     strategy:
       matrix:
         include:
-          - { runner: spacetimedb-runner, smoketest_args: --docker }
-          - { runner: windows-latest, smoketest_args: --no-build-cli }
-        runner: [ spacetimedb-runner, windows-latest ]
+          - runner: spacetimedb-new-runner
```
jdetter, line 27: The new runner is called `spacetimedb-new-runner`. This is easy to change later but I think I'd like to keep it like this for now in case we have to revert back to `spacetimedb-runner` for some reason.

.github/workflows/csharp-test.yml
```diff
@@ -157,6 +162,8 @@ jobs:
           key: Unity-${{ github.head_ref }}
           restore-keys: Unity-
 
+      - name: Start Docker daemon
+        run: /usr/local/bin/start-docker.sh
```
jdetter, line 167: We need this to support "Docker in Docker"

.github/workflows/ci.yml
```diff
@@ -368,13 +395,17 @@ jobs:
 
               cd "$GITHUB_WORKSPACE/sdks/unreal"
               cargo --version
-              cargo test
+              cargo test -p sdk-unreal-test-harness --test test -- --test-threads=1 --nocapture
```
jdetter: The unreal engine tests now run sequentially instead of in parallel. This makes the tests way more consistent and I think a ~20 minute wait is acceptable for a much more reliable result especially since the tests used to take ~29 minutes to run.

.github/workflows/ci.yml
bfops: we use a docker container so that we can test the docker based tests, e.g. restarting spacetimedb

.github/workflows/ci.yml
bfops: sgtm. what are all the other changes on this line for?

.github/workflows/csharp-test.yml
```diff
@@ -12,7 +12,12 @@ concurrency:
 
 jobs:
   unity-testsuite:
-    runs-on: spacetimedb-runner
+    runs-on: spacetimedb-new-runner
+    container:
+      image: localhost:5000/spacetimedb-ci:latest
+      options: >-
+        --privileged
+        --cgroupns=host
```
bfops, line 20: why this one here?

bfops (approved):
this LGTM, though I worry about you being the only person who understands these changes. If something went wrong, I wouldn't know how to debug it.

.github/workflows/ci.yml
```diff
@@ -23,10 +23,17 @@ jobs:
     strategy:
       matrix:
         include:
-          - { runner: spacetimedb-runner, smoketest_args: --docker }
-          - { runner: windows-latest, smoketest_args: --no-build-cli }
-        runner: [ spacetimedb-runner, windows-latest ]
```
bfops, line 28: I would restore (and tweak) this line, since otherwise the CI names include allll of the matrix tags and it makes them a bit brittle

.github/workflows/csharp-test.yml
bfops: this might be worth a comment in the code rather than PR

.github/workflows/csharp-test.yml
jdetter: It needs this cgroup setting because of the game-ci container, I'm not entirely sure why this one needs it but the others don't but I know it has to do with something the game-ci container tries to setup before running the container. I can find the error if you want more info, let me know 👍

.github/workflows/ci.yml
jdetter: Good question, this is what I got from Jason when I asked about running the tests serially instead of in parallel. I followed up with him and he said that those are the options he's just been using since day 1 and didn't really question it. I don't think we need `-p sdk-unreal-test-harness`, `--test test` or `--nocapture`so I've removed them.

.github/workflows/csharp-test.yml
bfops: can you add something like that in a comment?

Issue #2882, opened 2025-06-20
`Interval` scheduled rows: reschedule after duration starting from run time, not from `now` after execution finished


gefjon:
In crates/core/src/host/scheduler.rs `SchedulerActor::delete_scheduled_reducer_row`, when rescheduling a reducer for a `ScheduleAt::Interval`, we do `Timestamp::now() + dur` to get the next time to run. This means that any interval reducer which has non-negligible execution time will gradually drift from its intended interval. Change this so that we reschedule for `prev_scheduled_time + dur` instead.

gefjon:
We've had numerous complaints about scheduler drift in the public Discord, which I believe are due to this.

gefjon:
When re-scheduling a reducer that takes longer to execute than its requested `Interval`, log a warning (preferably to the module's logs, but realistically to the host logs given that we don't have a good way to do that right now) and re-schedule for `prev_scheduled_time + (n * dur)`, where `n` is the smallest integer such that `prev_scheduled_time + (n * dur)` is greater than `Timestamp::now()`.

Shubham8287:
I think, this other issue also tracks same - https://github.com/clockworklabs/SpacetimeDB/issues/2648

Shubham8287:
Not similar, but related - https://github.com/clockworklabs/SpacetimeDB/issues/2884.

PR #3657, opened 2025-11-13
Scheduled reducer: use timestamp from reducer params for next run

Shubham8287:
# Description of Changes

fixes: #2882

Schedule repeated reducers from their last execution timestamp instead of `Timestamp::now`

NA

1

# Testing
Existing test should be enough to cover any regression.


Shubham8287:
@jdetter , Letting you know, we would like to bot test this patch with BitCraft.

PR #3660, opened 2025-11-14
CI - Fix "Check that packages are publishable" check

bfops:
# Description of Changes

Our `Check that packages are publishable` CI check was silently failing due to a subshell error not getting propagated upward.

This PR both fixes that error-swallowing behavior, and fixes the errors that were present.

No. CI only.

1

# Testing

- [x] Check fails if the script to find crates fails (https://github.com/clockworklabs/SpacetimeDB/actions/runs/19349995837/job/55359458732?pr=3660)
- [x] Check is now passing

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -233,10 +233,10 @@ jobs:
         run: |
           set -ueo pipefail
           FAILED=0
-          ROOTS=(spacetimedb spacetimedb-sdk spacetimedb-cli spacetimedb-standalone)
-          CRATES=$(venv/bin/python3 tools/find-publish-list.py --recursive --quiet "${ROOTS[@]}")
-          for crate in $CRATES; do
-            if ! venv/bin/python3 tools/crate-publish-checks.py "crates/$crate"; then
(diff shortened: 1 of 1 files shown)


jdetter (approved):
Seems fine to me, thanks for the quick fix 👍 

Image

.github/workflows/ci.yml
```diff
@@ -231,11 +231,12 @@ jobs:
           venv/bin/pip3 install argparse toml
       - name: Run checks
         run: |
+          set -ueo pipefail
           FAILED=0
-          # This definition of ROOTS and invocation of find-publish-list.py is copied from publish-crates.sh
-          ROOTS=(bindings sdk cli standalone)
-          for crate in $(venv/bin/python3 tools/find-publish-list.py --recursive --quiet "${ROOTS[@]}"); do
-            if ! venv/bin/python3 tools/crate-publish-checks.py "crates/$crate"; then
+          ROOTS=(spacetimedb spacetimedb-sdk)
```
bfops, line 236: updated to use crate names instead of directory names. also dropped `cli` and `standalone` since we no longer publish those.

PR #3666, opened 2025-11-15
Add C# client SDK procedures

### LINKED ISSUE #3533: Procedures: C# client SDK invocation interface
Replicate the Rust interface introduced in #3532 in the C# client SDK.

JasonAtClockwork:
# Description of Changes

Closes: #3533 
Updated the C# SDK to handle procedures and procedure callbacks in a similar fashion to the Rust client as well as added the codegen to support it.

N/A

2 - This adds a new testing frame that should be removed once procedures are handled with C# module bindings

# Testing

Added /sdks/csharp/examples~/regression-tests/procedure-client to match modules/sdk-test-procedure which we can roll into the standard regression-tests once C# supports the procedure attribute. 

- [x] Add C# client test of sdk-test-procedure


crates/codegen/src/csharp.rs
```diff
@@ -706,13 +814,118 @@ impl Lang for Csharp<'_> {
...
+                    func_params.push_str(", ");
+                    func_args.push_str(", ");
+                }
+
+                let arg_type_str = ty_fmt_with_ns(module, arg_ty, self.namespace);
+                let arg_name = arg_name.deref().to_case(Case::Camel);
+
+                write!(func_params, "{arg_type_str} {arg_name}").unwrap();
+                write!(func_args, "{arg_name}").unwrap();
+            }
```
Centril: could be deduped with reducers?

crates/codegen/src/csharp.rs
```diff
@@ -914,6 +1152,43 @@ fn ty_fmt<'a>(module: &'a ModuleDef, ty: &'a AlgebraicTypeUse) -> impl fmt::Disp
     })
 }
 
+fn ty_fmt_with_ns<'a>(module: &'a ModuleDef, ty: &'a AlgebraicTypeUse, namespace: &'a str) -> impl fmt::Display + 'a {
```
Centril, line 1131: Doc comment?

crates/codegen/src/csharp.rs
JasonAtClockwork: I did avoid this to reduce complexity with this PR as I added namespace resolution to the argument types but you're right this should be deduped so I've updated the branch and ran all tests/generates to make sure it looks good.

crates/codegen/src/csharp.rs
JasonAtClockwork: Added a comment to reflect the difference between ty_fmt and ty_fmt_with_ns

Centril (approved):
Not a C# guy, but this looks good to me.

rekhoff (approved):
Looks good. Worked through any questions directly. Locally tested the regression test and everything passes. Approved.

JasonAtClockwork:
@rekhoff - Can you give this a quick look for the updated tests?

rekhoff:
> @rekhoff - Can you give this a quick look for the updated tests?

Reviewed tests, they look good. Ran the tests locally and they all passed.

PR #3668, opened 2025-11-15
Fix C# module bindings codegen for namespaces with view dispatcher

### LINKED ISSUE #3658: View C# Generation Fails with Namespace
#### DESCRIPTION
---
When building a C# server module using a namespace for the Module the return value for any defined Views will be mangled in the FFI.cs.

#### EXAMPLE
---
```cs
using SpacetimeDB;

namespace Test;
public static partial class Module
{
    [SpacetimeDB.Table]
    public partial struct Person
    {
        [SpacetimeDB.AutoInc]
        [SpacetimeDB.PrimaryKey]
        public int Id;
        public string Name;
        public int Age;
    }
    
   
    [SpacetimeDB.View(Name = "MyPerson", Public = true)]
    public static Person? MyPerson(ViewContext ctx) =>
        ctx.Db.Person.Id.Find(0);

    [SpacetimeDB.Reducer]
    public static void Add(ReducerContext ctx, string name, int age)
    {
        var person = ctx.Db.Person.Insert(new Person { Name = name, Age = age });
        Log.Info($"Inserted {person.Name} under #{person.Id}");
    }

    [SpacetimeDB.Reducer]
    public static void SayHello(ReducerContext ctx)
    {
        foreach (var person in ctx.Db.Person.Iter())
        {
            Log.Info($"Hello, {person.Name}!");
        }
        Log.Info("Hello, World!");
    }
}
```

#### TODO
---
- [ ] Repair the codegen to properly parse the type for ReturnType on ViewDispatcher

JasonAtClockwork:
# Description of Changes

Closes: #3658 
With the work on Views we ran into trouble with the Return Type and certain C# code layout where the types are defined inside the Module class. This fixes the bug in our current approach but needs consideration to possibly be changed entirely.

N/A

1 - Minor change

# Testing

- [x] Retested with the linked code with and without the namespace
- [x] Re-ran regression tests


crates/bindings-csharp/Codegen/Module.cs
```diff
@@ -997,10 +998,11 @@ public string GenerateViewDef(uint Index) =>
                 IsPublic: {{{IsPublic.ToString().ToLower()}}},
                 IsAnonymous: {{{IsAnonymous.ToString().ToLower()}}},
                 Params: [{{{MemberDeclaration.GenerateDefs(Parameters)}}}],
-                ReturnType: new {{{ReturnType.BSATNName.Replace(
-                "Module.",
-                "global::Module."
-            )}}}().GetAlgebraicType(registrar)
+                ReturnType: new {{{Regex.Replace(
+                    ReturnType.BSATNName,
+                    @"(?<=^|[<,\s])Module\.",
```
joshua-spacetime: How are namespaces handled for parameters? My expectation would be that the return type would handled similarly.

JasonAtClockwork: They're actually handled differently and that might be a clue to fix this properly. I'll work with @rekhoff tomorrow on a gameplan to review the MemberDeclaration and see if we can't use that instead of this hacky fix.

JasonAtClockwork: After reviewing with @rekhoff and testing it looks like we can remove it entirely. Likely a hold-over from the early work on views as we were having troubles with namespaces.

rekhoff (approved):
No objections here, the change makes sense and works as intended

PR #3672, opened 2025-11-17
fix plan caching for client-specific views

joshua-spacetime:
# Description of Changes

Fixes a subscription plan caching bug related to client-specific views.

Before this fix, you could define a client-specific view:
```rust
fn my_view(ctx: &ViewContext) -> Option<Player> {
    ctx.db.player().identity().find(ctx.sender)
}
```

And subscribe to it as follows:
```sql
SELECT * FROM my_view
```

Note this view is implicitly parameterized by `:sender`, however when generating a query hash for this subscription, this fact would not be taken into account which would result in this query being cached and reused for all callers.

After this fix, a query hash is generated for this subscription as though it were given as:
```sql
SELECT * FROM my_view WHERE identity = :sender
```

None

**Note for CLI code owners:**

I had to touch the `subscribe` cli command file. No updates to the api. It just needed to be updated to look for views in the module def.

1

# Testing

- [x] Added a regression smoketest

### DIFF
--- a/smoketests/tests/views.py
+++ b/smoketests/tests/views.py
@@ -440,12 +440,6 @@ pub fn insert_player(ctx: &ReducerContext, name: String) {
 }
 """
 
-    def assertSql(self, sql, expected):
-        self.maxDiff = None
-        sql_out = self.spacetime("sql", self.database_identity, sql)
-        sql_out = "\n".join([line.rstrip() for line in sql_out.splitlines()])
-        expected = "\n".join([line.rstrip() for line in expected.splitlines()])
-
     def _test_subscribing_with_different_identities(self):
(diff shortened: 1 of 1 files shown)


smoketests/tests/views.py
```diff
@@ -416,3 +416,79 @@ def test_recovery_from_trapped_views_auto_migration(self):
...
+            }
+            for event in events
+        ]
+
+        self.assertEqual(
+            [
+                {
+                    'my_player': {
+                        'deletes': [],
+                        'inserts': [{'name': 'Bob'}],
```
Shubham8287, line 482: In master, this will return "Alice" due to last query hash being cached?

joshua-spacetime: Yes that's right.

PR #3674, opened 2025-11-17
CI - Cancel internal tests if cancelled

bfops:
# Description of Changes

`Internal Tests` invoke another workflow. If the calling workflow is cancelled, now we also cancel the invoked workflow.

None.

2

# Testing

- [x] Push multiple commits to this PR, see that the invoked job is cancelled
- [x] Manually cancel the `Internal Tests` job, see that the invoked job is cancelled


.github/workflows/internal-tests.yml
```diff
@@ -18,16 +18,20 @@ jobs:
     # Skip if this is an external contribution. GitHub secrets will be empty, so the step would fail anyway.
     if: ${{ github.event_name != 'pull_request' || !github.event.pull_request.head.repo.fork }}
     runs-on: ubuntu-latest
+    env:
+      TARGET_OWNER: clockworklabs
+      TARGET_REPO: SpacetimeDBPrivate
```
jdetter, line 23: We could make this a github secret if we don't want to include it in text here, but it's not really sensitive.

.github/workflows/internal-tests.yml
```diff
@@ -110,3 +125,24 @@ jobs:
             if (conclusion !== 'success') {
               core.setFailed(`Private workflow failed with conclusion: ${conclusion}`);
             }
+
+      - name: Cancel invoked run if workflow cancelled
```
jdetter, line 129: very cool, I didn't even know you could do this 👍

PR #3682, opened 2025-11-18
Added prefix to global delegates: FOnConnectDelegate, FOnDisconnectDe…

UmarM-Arvika:
Added prefix to global delegates

# Description of Changes

Some global delegates needed prefix in order to work in MultiModule scenario

N/A

1 - This only adds prefix to four global delegates

# Testing

Generete multiple module bindings for an Unreal Engine project and include all the {ModulePefix}SpacetimeDBClient.g.h in a single translation unit.


JasonAtClockwork:
I've manually merged this in as there was very large changes that I needed to rebase from and added a couple missed delegates for the table insert/update/delete. Thanks for this work!

PR #3683, opened 2025-11-18
Fix test with change on defaults

mamcx:
# Description of Changes

The `config` added `maincloud` in the constructors so the test need update.

1
# Testing

- [x] Re-run the test until pass


bfops (approved):
I tested this by commenting out the writer threads. This caused failures on `master`, but not on this PR.

LGTM!

PR #3684, opened 2025-11-18
Add procedure HTTP request API for WASM modules and the Rust module bindings library

### LINKED ISSUE #3517: Procedures: Add HTTP API
Add a single new host call for performing an HTTP request, `procedure_http_request`. Its signature is approximately:

```
extern "C" fn procedure_http_request(req: *const u8, req_len: u32, out: *mut BytesSource) -> Errno;
```

The `(req, req_len)` buffer encodes an `HttpRequest` object. Decide whether this should be re-exported from the `http` crate, some other crate, or our own custom type. Justify your choice in the PR description.

The `out` `BytesSource` has meaning depending on the return value: if the call returns ok, `out` yields an `HttpResponse` object. (This may be an HTTP error response; the call assigns no semantics to HTTP responses whatsoever.) If the call returns an error, `out` yields an `HttpError` further describing the failure.

Wrap this up in a method on `ProcedureContext` which handles the ser/de parts. Its signature is approximately:

```rust
impl ProcedureContext {
    pub fn http(&mut self, req: HttpRequest) -> Result<HttpResponse, HttpError> { todo!() }
}
```

Then make follow-up tickets to expose the same interface in C#, C++ and TypeScript.

### LINKED ISSUE #3644: Procedures: Mark unstable for initial release
In the module bindings libraries, feature-flag the procedure methods behind our unstable feature for the initial release.

gefjon:
# Description of Changes

Closes #3517 .

With this PR, procedures (at least, those defined in Rust modules) can perform HTTP requests! This is performed through a new field on the `ProcedureContext`, `http: HttpClient`, which has a method `send` for sending an `http::Request`, as well as a convenience wrapper `get`.

Internally, these methods hit the `procedure_http_request` ABI call / host function, which uses reqwest to perform an HTTP request. The request is run with a user-configurable timeout which defaults and is clamped to 500 ms.
Rather than exposing the HTTP stream to modules, we download the entire response body immediately, within the same timeout.

I've added an example usage of `get` to `module-test` which performs a request against `localhost:3000` to read its own schema/moduledef.

This PR also makes all procedure-related definitions in the Rust module bindings library `#[cfg(feature = "unstable")]`, as per #3644 . The rename of the `/v1/database/:name/procedure/:name` route is not included in this PR, so this does not close #3644 .

Left as TODOs are:
- Metrics for recording request and response size.
- Improving performance by stashing a long-lived `reqwest::Client` someplace.
  Currently we build a new `Client` for each request.
- Improving performance (possibly) by passing the request-future to the global tokio executor
  rather than running it on the single-threaded database executor.

Adds new APIs, which are marked as unstable. Adds a new ABI, which is not unstable in any meaningful way (we can't really do that). Marks unreleased APIs as unstable. Does not affect any pre-existing already-released APIs or ABIs.

3 or so: networking is scary, and even though we impose a timeout which prevents these connections from being truly long-lived, they're still potentially long-lived on the scale of Tokio futures. It's possible that running them on the database core is problematic in some way, and so what I've left as a performance TODO could actually be a concurrency-correctness issue.

# Testing

- [x] Manually wrote and executed some procedures which make HTTP requests.
- [x] Added two automated tests to the `sdk-test` suite, `procedure::http_ok` and `procedure::http_err`, which make successful and failing requests respectively, then return its result. A client then makes some assertions about the result.

### DIFF
--- a/crates/bindings/src/http.rs
+++ b/crates/bindings/src/http.rs
@@ -62,10 +62,11 @@ impl HttpClient {
     ///         log::error!("HTTP request failed: {err}");
     ///     },
     ///     Ok(response) => {
+    ///         let (parts, body) = response.into_parts();
     ///         log::info!(
     ///             "Got response with status {}, body {}",
-    ///             response.status(),
-    ///             response.body().into_string_lossy(),
+    ///             parts.status,
+    ///             body.into_string_lossy(),
     ///         );
     ///     }
     /// }
@@ -110,10 +111,11 @@ impl HttpClient {
     ///         log::error!("HTTP request failed: {err}");
     ///     }
     ///     Ok(response) => {
+    ///         let (parts, body) = response.into_parts();
     ///         log::info!(
     ///             "Got response with status {}, body {}",
-    ///             response.status(),
-    ///             response.body().into_string_lossy(),
+    ///             parts.status,
+    ///             body.into_string_lossy(),
     ///         );
     ///     }
     /// }


crates/core/src/host/wasmtime/wasm_instance_env.rs
```diff
@@ -1417,8 +1438,212 @@ impl WasmInstanceEnv {
...
+    ) -> Fut<'caller, RtResult<u32>> {
+        use spacetimedb_lib::http as st_http;
+
+        Self::async_with_span(caller, AbiCall::ProcedureHttpRequest, move |mut caller| async move {
+            let (mem, env) = Self::mem_env(&mut caller);
+
+            // Yes clippy, I'm calling a closure at its definition site *on purpose*,
+            // as a hacky-but-stable `try` block.
+            #[allow(clippy::redundant_closure_call)]
+            let res = (async move || {
```
coolreader18, line 1708: Most of the body of this function should be in `InstanceEnv` I think, so that v8 can make use of it too.

crates/lib/src/http.rs
```diff
@@ -0,0 +1,545 @@
...
+            .body(body)
+            .expect("`http::request::Builder::body` returned error")
+    }
+}
+
+#[derive(Clone, SpacetimeType)]
+#[sats(crate = crate)]
+enum HttpRequest {
+    V0(HttpRequestV0),
+    NonExhaustive(Box<[u8]>),
```
coolreader18: What's the purpose of this? Why not just `#[non_exhaustive]`? And I think it'd actually be better to not mark it `non_exhaustive`, since we want to remember to catch all the places we match on it if we do add a variant, and this isn't really user-facing.

crates/lib/src/http.rs
```diff
@@ -0,0 +1,545 @@
...
+                    body,
+                    method,
+                    headers,
+                    timeout,
+                    uri,
+                    version,
+                }),
+        } = req
+        else {
+            unreachable!("`HttpRequest::NonExhausitve` pseudo-variant encountered");
```
coolreader18: A user could theoretically pass a `NonExhaustive` to an abi call, causing a crash.

crates/lib/src/http.rs
```diff
@@ -0,0 +1,545 @@
...
+#[sats(crate = crate)]
+struct HttpRequestV0 {
+    body: Body,
+    method: Method,
+    headers: Headers,
+    timeout: Option<Timeout>,
+    /// A valid URI, sourced from an already-validated [`http::Uri`].
+    uri: String,
+    version: Version,
+}
```
coolreader18: I think it might be a good idea to store the body separately from the rest of the request, incl. as a separate argument in the abi call and perhaps even just not defining a type for it in this file. In terms of interchange the body is really straightforward to represent (it's just bytes), but if we did want to support streaming in the future you wouldn't be able to use this. A side benefit of defining it as a separate argument in the abi call is that we don't have to copy the buffer that the user provides.

crates/lib/src/http.rs
gefjon: This is to allow extensions while preserving BFLATN-compatibility: we can add additional variants at the end of a sum type, but they must have <= the size and alignment of all the existing variants. Ensuring that there's one variant whose payload is a var-length type (`String`, `Vec<u8>`, &c) means that, in the worst case, we can add new variants whose SATS type is a bytestring, which we assign additional semantics to, e.g. by treating it as the BSATN of some known type.

crates/lib/src/http.rs
gefjon: Good point. I'll make this a trap rather than a panic.

crates/bindings-sys/src/lib.rs
```diff
@@ -663,6 +663,38 @@ pub mod raw {
         // TODO(procedure-sleep-until): remove this
         #[cfg(feature = "unstable")]
         pub fn procedure_sleep_until(wake_at_micros_since_unix_epoch: i64) -> i64;
+
+        /// Perform an HTTP request as specified by the buffer `request_ptr[..request_len]`,
+        /// suspending execution until the request is complete,
+        /// then return its response via a [`BytesSource`] written to `out`.
+        ///
+        /// `request_ptr[..request_len]` should store a BSATN-serialized `spacetimedb_lib::http::Request` object
```
Centril, line 740:
```suggestion
        /// `request_ptr[..request_len]` should store a BSATN-serialized [`spacetimedb_lib::http::Request`] object
```

crates/bindings-sys/src/lib.rs
```diff
@@ -663,6 +663,38 @@ pub mod raw {
...
+        /// containing the details of the request to be performed.
+        ///
+        /// If the request is successful, a [`BytesSource`] is written to `out`
+        /// containing a BSATN-encoded `spacetimedb_lib::http::Response` object.
```
Centril, line 744:
```suggestion
        /// containing a BSATN-encoded [`spacetimedb_lib::http::Response`] object.
```

crates/bindings-sys/src/lib.rs
```diff
@@ -663,6 +663,38 @@ pub mod raw {
...
+        /// "Successful" in this context includes any connection which results in any HTTP status code,
+        /// regardless of the specified meaning of that code.
```
Centril, line 746:
```suggestion
        /// regardless of the specified meaning of that code.
        /// This includes error codes such as 404 Not Found
        /// and 500 Internal Server Error.
```

crates/bindings-sys/src/lib.rs
```diff
@@ -663,6 +663,38 @@ pub mod raw {
...
+        ///
+        /// # Errors
+        ///
+        /// Returns an error:
+        ///
+        /// - `WOULD_BLOCK_TRANSACTION` if there is currently a transaction open.
+        ///   In this case, `out` is not written.
+        /// - `HTTP_ERROR` if an error occurs while executing the HTTP request.
+        ///   In this case, a [`BytesSource`] is written to `out`
+        ///   containing a BSATN-encoded `spacetimedb_lib::http::Error` object.
```
Centril, line 759:
```suggestion
        ///   containing a BSATN-encoded [`spacetimedb_lib::http::Error`] object.
```

crates/bindings-sys/src/lib.rs
```diff
@@ -1241,11 +1273,45 @@ impl Drop for RowIter {
...
         unsafe { super::raw::procedure_sleep_until(wake_at_timestamp) }
     }
+
+    #[inline]
+    #[cfg(feature = "unstable")]
+    /// Perform an HTTP request as specified by `http_request_bsatn`,
+    /// suspending execution until the request is complete,
+    /// then return its response or error.
+    ///
+    /// `http_request_bsatn` should be a BSATN-serialized `spacetimedb_lib::http::Request`.
```
Centril, line 1448:
```suggestion
    /// `http_request_bsatn` should be a BSATN-serialized [`spacetimedb_lib::http::Request`].
```

crates/bindings-sys/src/lib.rs
```diff
@@ -1241,11 +1273,45 @@ impl Drop for RowIter {
...
+    ///
+    /// If the request completes successfully,
+    /// this function returns `Ok(bytes)`, where `bytes` contains a BSATN-serialized `spacetimedb_lib::http::Response`.
```
Centril, line 1451:
```suggestion
    /// this function returns `Ok(bytes)`, where `bytes` contains a BSATN-serialized [`spacetimedb_lib::http::Response`].
```

crates/bindings-sys/src/lib.rs
```diff
@@ -1241,11 +1273,45 @@ impl Drop for RowIter {
...
+    /// All HTTP response codes are treated as successful for these purposes;
```
Centril, line 1452:
```suggestion
    /// All HTTP response codes, including ones like 404 and 500,
    /// are treated as successful for these purposes;
```

crates/bindings-sys/src/lib.rs
```diff
@@ -1241,11 +1273,45 @@ impl Drop for RowIter {
...
+    /// this method only returns an error if it is unable to produce any HTTP response whatsoever.
+    /// In that case, this function returns `Err(bytes)`, where `bytes` contains a BSATN-serialized `spacetimedb_lib::http::Error`.
```
Centril, line 1454:
```suggestion
    /// In that case, this function returns `Err(bytes)`, where `bytes` contains a BSATN-serialized [`spacetimedb_lib::http::Error`].
```

crates/bindings/src/http.rs
```diff
@@ -0,0 +1,66 @@
...
+impl HttpClient {
+    /// Send the HTTP request `request` and wait for its response.
+    ///
+    /// For simple `GET` requests with no headers, use [`HttpClient::get`] instead.
+    // TODO(docs): expand docs
+    pub fn send<B: Into<Body>>(&self, request: Request<B>) -> Result<Response<Body>, Error> {
+        let request = st_http::Request::from(request);
+        let request = bsatn::to_vec(&request).expect("Failed to BSATN-serialize `spacetimedb_lib::http::Request`");
+
+        fn read_output<T: for<'a> Deserialize<'a> + 'static>(source: BytesSource) -> T {
```
Centril:
```suggestion
        fn read_output<T: DeserializeOwned + 'static>(source: BytesSource) -> T {
```

Centril: This feels like a reusable function; consider moving out of here?

crates/bindings/src/http.rs
```diff
@@ -0,0 +1,66 @@
...
+        }
+    }
+
+    /// Send a `GET` request to `uri` with no headers.
+    ///
+    /// Blocks procedure execution for the duration of the HTTP request.
+    pub fn get<Uri>(&self, uri: Uri) -> Result<Response<Body>, Error>
+    where
+        Uri: TryInto<http::Uri>,
+        <Uri as TryInto<http::Uri>>::Error: Into<http::Error>,
```
Centril:
```suggestion
    pub fn get(
        &self,
        uri: impl TryInto<http::Uri, Error: Into<http::Error>>
    ) -> Result<Response<Body>, Error>
```

crates/core/src/host/wasmtime/wasm_instance_env.rs
```diff
@@ -1417,8 +1438,212 @@ impl WasmInstanceEnv {
             res
         })
     }
+
+    /// Perform an HTTP request as specified by the buffer `request_ptr[..request_len]`,
+    /// suspending execution until the request is complete,
+    /// then return its response via a [`BytesSource`] written to `out`.
+    ///
+    /// `request_ptr[..request_len]` should store a BSATN-serialized `spacetimedb_lib::http::Request` object
```
Centril:
```suggestion
    /// `request_ptr[..request_len]` should store a BSATN-serialized [`spacetimedb_lib::http::Request`] object
```

crates/core/src/host/wasmtime/wasm_instance_env.rs
```diff
@@ -1417,8 +1438,212 @@ impl WasmInstanceEnv {
...
+    /// containing the details of the request to be performed.
+    ///
+    /// If the request is successful, a [`BytesSource`] is written to `out`
+    /// containing a BSATN-encoded `spacetimedb_lib::http::Response` object.
```
Centril:
```suggestion
    /// containing a BSATN-encoded [`spacetimedb_lib::http::Response`] object.
```

crates/core/src/host/wasmtime/wasm_instance_env.rs
```diff
@@ -1417,8 +1438,212 @@ impl WasmInstanceEnv {
...
+    ///
+    /// # Errors
+    ///
+    /// Returns an error:
+    ///
+    /// - `WOULD_BLOCK_TRANSACTION` if there is currently a transaction open.
+    ///   In this case, `out` is not written.
+    /// - `HTTP_ERROR` if an error occurs while executing the HTTP request.
+    ///   In this case, a [`BytesSource`] is written to `out`
+    ///   containing a BSATN-encoded `spacetimedb_lib::http::Error` object.
```
Centril:
```suggestion
    ///   containing a BSATN-encoded [`spacetimedb_lib::http::Error`] object.
```

crates/core/src/host/wasmtime/wasm_instance_env.rs
```diff
@@ -1417,8 +1438,212 @@ impl WasmInstanceEnv {
...
+    /// suspending execution until the request is complete,
+    /// then return its response via a [`BytesSource`] written to `out`.
+    ///
+    /// `request_ptr[..request_len]` should store a BSATN-serialized `spacetimedb_lib::http::Request` object
+    /// containing the details of the request to be performed.
+    ///
+    /// If the request is successful, a [`BytesSource`] is written to `out`
+    /// containing a BSATN-encoded `spacetimedb_lib::http::Response` object.
+    /// "Successful" in this context includes any connection which results in any HTTP status code,
+    /// regardless of the specified meaning of that code.
```
Centril, line 1673:
```suggestion
    /// regardless of the specified meaning of that code.
    /// This includes error codes such as 404 Not Found
    /// and 500 Internal Server Error.
```

crates/core/src/host/wasmtime/wasm_instance_env.rs
```diff
@@ -1417,8 +1438,212 @@ impl WasmInstanceEnv {
...
+                }
+
+                // TODO(procedure-metrics): record size in bytes of request.
+
+                // Read the request from memory as a `spacetimedb_lib::http::Request`,
+                // our bespoke type with a stable layout and BSATN encoding.
+                let request_buf = mem.deref_slice(request_ptr, request_len)?;
+                let request = bsatn::from_slice::<st_http::Request>(request_buf).map_err(|err| {
+                    // BSATN deserialization failure of the request will trap by returning `NodesError::DecodeValue`,
+                    // which `Self::convert_wasm_result` treats as fatal.
```
Centril: I'd prefer for consistency to return the error code `BSATN_DECODE_ERROR`.

crates/core/src/host/wasmtime/wasm_instance_env.rs
```diff
@@ -1417,8 +1438,212 @@ impl WasmInstanceEnv {
...
+                    .unwrap_or(HTTP_DEFAULT_TIMEOUT)
+                    .min(HTTP_DEFAULT_TIMEOUT);
+
+                *reqwest.timeout_mut() = Some(timeout);
+
+                // Un-`mut`.
+                let reqwest = reqwest;
+
+                // Actually execute the HTTP request!
+                // We'll wrap this future in a `tokio::time::timeout` before `await`ing it.
```
Centril: The double timeout (both reqwest and tokio) looks a bit funny, so it could use a few words.

crates/lib/src/http.rs
gefjon: What would you use as the bound on `T` in `http::Request<T>` as an argument to `ctx.http.send`?

crates/lib/src/http.rs
gefjon: There's some discussion of this in the module-level doc comment. If you think it's useful, I could add comments to each variant like this as well.

crates/lib/src/http.rs
```diff
@@ -0,0 +1,545 @@
...
+//! so we're forced to define our own representation for the SATS serialization.
+//! These types are that representation.
+//!
+//! To preserve extensibility and compatibility, all types defined here should be
+//! a `pub` wrapper struct with a single private `inner` field,
+//! which is an `enum` whose last variant holds either `Box<str>`, `String`, `Box<[T]>` or `Vec<T>` for any type `T`.
+//! Using an enum allows us to add additional variants while preserving the BSATN encoding passed across the WASM boundary,
+//! and including a variant with a variable-length type
+//! allows us to add other variants with variable-length types while preserving the BFLATN layout stored in table pages.
+//! (It's unlikely that any of these types will end up stored in a table, but better safe than sorry.)
```
Centril:
```suggestion
//! (It's unlikely that any of these types will end up stored in a table,
//! but better safe than sorry. Since we're doing networking IO anyways,
//! the added cost is negligible.)
```

crates/lib/src/http.rs
```diff
@@ -0,0 +1,545 @@
...
+//!
+//! Users aren't intended to interact with these types, except [`Body`], [`Timeout`] and [`Error`].
+//! Our user-facing APIs should use the [`http`] crate's types directly, and convert to and from these types internally.
+
+use spacetimedb_sats::{time_duration::TimeDuration, SpacetimeType};
+
+/// Represents an HTTP request which can be made from a procedure running in a SpacetimeDB database.
+///
+/// Construct instances of this type by converting from [`http::Request`].
```
Centril, line 26:
```suggestion
/// Construct instances of this type by converting from [`http::Request`].
/// Note that all extensions to `http::Request`,
/// save for [`Timeout`], are ignored.
```

crates/lib/src/http.rs
Centril: Maybe just a "See module docs..." style comment on each variant?

crates/bindings-sys/src/lib.rs
gefjon: I believe that this (and others with similar comments within this crate) isn't a valid link - `bindings-sys` does not have a dependency on `lib`.

crates/core/src/host/wasmtime/wasm_instance_env.rs
gefjon: A few words like:
```rust
// If the user requested a timeout using our extension, slot it in to reqwest's timeout.
// Clamp to the range `0..HTTP_DEFAULT_TIMEOUT`.
// We'll use this same `timeout: Duration` in a `tokio::time::timeout`,
// which will enclose both the initial request and downloading the body.
// Supplying it to reqwest is just for redundancy.
```
?

crates/core/src/host/wasmtime/wasm_instance_env.rs
gefjon: I am unsure where to draw the line. I am happy to take a guess if you'd prefer, but I think it may be more valuable for you to extract the shared parts during your work.

crates/bindings-sys/src/lib.rs
Centril: Oh, that makes sense; please disregard 😊

crates/core/src/host/wasmtime/wasm_instance_env.rs
Centril: Yeah that's helpful 👍

gefjon: It's lines 1548 -- 1552

crates/lib/src/http.rs
coolreader18: Something defined in that crate - there could be a body type equivalent to what's here now, but it should be part of the bindings api, not an interchange type.

crates/core/src/host/wasmtime/wasm_instance_env.rs
coolreader18: Discussed; I'll split it out in the v8 pr.

coolreader18:
Phoebe and I were discussing `HttpRequest` - IMO it shouldn't have a `NonExhaustive` variant, and it probably shouldn't be an enum. These are internal interchange types only used for communicating across the abi boundary, and the `spacetimedb_lib` crate is explicitly marked as unstable API - if a user ends up storing it in their table, that's on them. Accommodating the possibility just adds uncertainty to our actual data model.

Furthermore, if we did want to change HttpRequest in the future, it'd probably just be best to introduce a new abi function altogther, since that's our existing mechanism for versioning. Under the enum scheme, if we added a new variant and someone built a new module and ran it against an old spacetimedb server, it would be a runtime error only when the http request is actually made. Also, no other types in `spacetimedb_lib` are structured like this, except for `RawModuleDef`, which is necessary because of the way that `__describe_module__` is an export and so can't be versioned.

crates/lib/src/http.rs
gefjon: Resolved as we no longer have `NonExhaustive` variants.

crates/lib/src/http.rs
gefjon: Resolved as we no longer have NonExhaustive variants.

crates/lib/src/http.rs
gefjon: Resolved by  @coolreader18 's commit.

crates/core/src/host/wasmtime/wasm_instance_env.rs
gefjon: Resolved by @coolreader18 's commit.

crates/bindings/src/http.rs
gefjon: Moved to `rt::read_bytes_source_as`.

crates/lib/src/http.rs
gefjon: Now irrelevant.

Centril (approved):

JasonAtClockwork:
Ran the Unreal tests locally it appears to be the flaky nature of Unreal in CI.
image

PR #3685, opened 2025-11-18
Update view ABI to support returning queries

jsdt:
# Description of Changes

This adds some changes for how we return data from view functions. Originally, we interpreted the output of a view function as a bsatn encoding of an array of rows. Since we also want to be able to return queries from view functions, we need to be able to return different types too. At this point, this is effectively not a functional change, since we don't use the new format, and we don't actually try to parse the new format.

This introduces a new format for view returns, which is a `ViewResultHeader`, potentially followed by additional data. For example, if a view were returning rows directly, it would write a `ViewResultHeader::RowData`, followed by an array of rows. Note that we could have given that object a byte array with the rows instead of using a header an a separate object, but that would force us create an extra copy when encoding and when decoding.

To make this backward compatible with existing views, we have a different way to return the new format. For v8 views, if they return a byte array, we assume it is the old format. If they return an object, we expect the `data` field of that object to be the actual return data.

For wasm views, we interpret a return code of 2 to mean that it uses the new format.

On the host side, we handle this naively: we will perform the query, and we will act as though the view has a read dependency on the tables in the query. In follow up PRs we can make this more efficient.

This is not a breaking change, but it does make the ABI more complicated (specifically to avoid breaking it).

1.5. This should not affect the existing return style.

# Testing

I've done manual testing of this with a version of the typescript bindings that returns queries.


crates/core/src/host/wasm_common/module_host_actor.rs
```diff
@@ -1008,6 +1108,66 @@ impl InstanceCommon {
...
+        tx: &mut MutTxId,
+        the_query: &str,
+        expected_row_type: &ProductType,
+        call_info: &ViewCallInfo,
+    ) -> anyhow::Result<Vec<ProductValue>> {
+        if the_query.trim().is_empty() {
+            return Ok(Vec::new());
+        }
+
+        // Views bypass RLS.
```
joshua-spacetime: I would mention why here. One of the reasons being that views are module-defined and access control can always expressed procedurally if need be.

crates/core/src/host/wasm_common/module_host_actor.rs
```diff
@@ -1008,6 +1108,66 @@ impl InstanceCommon {
...
+            }
+        }
+
+        let op = FuncCallType::View(call_info.clone());
+        let mut metrics = ExecutionMetrics::default();
+        let mut rows = Vec::new();
+
+        for plan in plans {
+            // Track read sets for all tables involved in this plan.
+            // TODO(jsdt): This means we will rerun the view and query for any change to these tables, so we should optimize this asap.
```
joshua-spacetime, line 1165: Even if they aren't incremental at first, we should probably still track index reads so that the performance is at least on par with procedural views.

crates/core/src/host/wasm_common/module_host_actor.rs
```diff
@@ -1008,6 +1108,66 @@ impl InstanceCommon {
         (res, trapped)
     }
 
+    fn run_query_for_view(
```
joshua-spacetime, line 1119: I would add a comment that the return rows have not been transformed for insert into the view's backing table.

joshua-spacetime (commented):
Can you update the bindings to use the new ABI?

crates/core/src/host/wasm_common/module_host_actor.rs
joshua-spacetime: I'm fine if we leave this as a follow up

PR #3691, opened 2025-11-18
Export __call_view__ in C#

rekhoff:
# Description of Changes

1. Updates the Replication Tests in `sdks/csharp/examples~/regression-tests` to include better coverage of Views
2. Added missing linkage for __call_view__ and __call_view_anon__
3. Updated *ViewDispatcher Invoke to transform BSATN.ValueOption<> into BSATN.List<>
4. Fixed issues with the indexing of views to match correctly during __call_view__ and __call_view_anon__

No

2

# Testing

- [x] Running `run-regression-tests.sh` passes.


sdks/csharp/examples~/regression-tests/client/Program.cs
```diff
@@ -50,7 +50,7 @@ void OnConnected(DbConnection conn, Identity identity, string authToken)
         {
             throw err;
         })
-        .Subscribe(["SELECT * FROM ExampleData"]);
+        .Subscribe(["SELECT * FROM ExampleData", "SELECT * FROM MyPlayer"]);
```
joshua-spacetime:
```suggestion
        .Subscribe(["SELECT * FROM ExampleData", "SELECT * FROM MyPlayer", "SELECT * FROM PlayersForLevel"]);
```

sdks/csharp/examples~/regression-tests/client/Program.cs
```diff
@@ -133,6 +133,24 @@ void OnSubscriptionApplied(SubscriptionEventContext context)
         ValidateBTreeIndexes(ctx);
         waiting--;
     });
+
+
+    // Views test
```
joshua-spacetime, line 138: These are great tests

crates/bindings-csharp/Codegen/Module.cs
```diff
@@ -1022,15 +1014,30 @@ public string GenerateDispatcherClass()
             ? "SpacetimeDB.AnonymousViewContext"
             : "SpacetimeDB.ViewContext";
 
+        var isValueOption = ReturnType.BSATNName.StartsWith("SpacetimeDB.BSATN.ValueOption");
```
joshua-spacetime: Is it possible for `ValueOption` not to be fully qualified?

crates/bindings-csharp/Codegen/Module.cs
```diff
@@ -1054,12 +1058,12 @@ public byte[] Invoke(
             """;
     }
 
-    public string GenerateClass()
-    {
-        var builder = new Scope.Extensions(Scope, FullName);
-        builder.Contents.Append(GenerateDispatcherClass());
-        return builder.ToString();
-    }
+    // public string GenerateClass()
```
joshua-spacetime: Commented out code

crates/bindings-csharp/Codegen/Module.cs
```diff
@@ -1489,8 +1493,15 @@ public sealed class Local {
                         }
                     }
                     
-                    {{string.Join("\n", views.Array.Select(v => v.GenerateDispatcherClass()))}}
-                    
+                    {{string.Join("\n", 
+                        views.Array.Where(v => !v.IsAnonymous)
+                            .Select((v, i) => v.GenerateDispatcherClass((uint)i))
```
joshua-spacetime, line 1496: What list/lists does `i` point to? I just want to convince myself that our index calculations are correct.

sdks/csharp/examples~/regression-tests/client/Program.cs
```diff
@@ -133,6 +133,25 @@ void OnSubscriptionApplied(SubscriptionEventContext context)
...
+    Log.Debug("Calling Iter on View");
+    var viewIterRows = context.Db.MyPlayer.Iter();
+    Log.Debug("MyPlayer Iter count: " + (viewIterRows != null ? viewIterRows.Count().ToString() : "null"));
```
joshua-spacetime, line 149: Is it possible to assert on the rows in this test to make sure the client cache is in the state we expect?

crates/bindings-csharp/Codegen/Module.cs
rekhoff: The `i` is the index value provided by the LINQ Select overload that includes the element’s position in the sequence, that comes from the `Where` statement, and should represent each items position in the `Where` statements ordered list.

sdks/csharp/examples~/regression-tests/client/Program.cs
```diff
@@ -133,6 +133,32 @@ void OnSubscriptionApplied(SubscriptionEventContext context)
...
     });
+
+
+    // Views test
+
+    Log.Debug("Checking Views are populated");
+    Debug.Assert(context.Db.MyPlayer != null, "context.Db.MyPlayer != null");
+    Debug.Assert(context.Db.PlayersForLevel != null, "context.Db.PlayersForLevel != null");
+    Debug.Assert(context.Db.MyPlayer.Count > 0, $"context.Db.MyPlayer.Count = {context.Db.MyPlayer.Count}");
+    Debug.Assert(context.Db.PlayersForLevel.Count > 0, $"context.Db.MyPlayer.Count = {context.Db.PlayersForLevel.Count}");
```
rekhoff:
```suggestion
    Debug.Assert(context.Db.PlayersForLevel.Count > 0, $"context.Db.PlayersForLevel.Count = {context.Db.PlayersForLevel.Count}");
```

crates/bindings-csharp/Codegen/Module.cs
joshua-spacetime: What I meant to ask is where is this index referenced? I'm assuming it's used by two different lists - one for anonymous views and one for client specific views?

JasonAtClockwork:
```cs
        var views = CollectDistinct(
            "View",
            context,
            viewDeclarations,
            v => v.Name,
            v => v.FullName
        );
 ```
 The views are made up out of this collection of the declarations, which is made up from the Roslyn gen finding anything with the View attribute. This list setup allows it to output 0..n for the Views then 0..n for the AnonViews to make the index generated correctly based on the order their registered. But saying all that we don't really have a solid test on this and may need to consider that. It's possible this array won't match 1:1 with the viewDispatch.Add() viewAnonDispatch.Add() calls.

Comment (JasonAtClockwork) [reply crates/bindings-csharp/Codegen/Module.cs:1496]: Confirmed with @rekhoff we needed a slight change to make sure both the RegisterView + ViewDispatch generation are identical which is coming in the next commit.

File: sdks/csharp/examples~/regression-tests/client/Program.cs (hunk 6 above)
Comment (JasonAtClockwork) [reply sdks/csharp/examples~/regression-tests/client/Program.cs:149]: We definitely can and it will be added for the MyPlayer record to match the id, Identity, and Name. This has to be done by hand as there isn't an equals operator generated for the types in C#.

File: crates/bindings-csharp/Codegen/Module.cs [hunk 8]
```diff
@@ -1001,7 +993,7 @@ public string GenerateViewDef(uint Index) =>
             );
             """;
 
-    public string GenerateDispatcherClass()
+    public string GenerateDispatcherClass(uint index)
```
joshua-spacetime, line 1001:
```suggestion
    /// <summary>
    /// Generates the class responsible for evaluating a view.
    /// If this is an anonymous view, the index corresponds to the position of this dispatcher in the `viewDispatchers` list of `RegisterView`.
    /// Otherwise it corresponds to the position of this dispatcher in the `anonymousViewDispatchers` list of `RegisterAnonymousView`.
    /// </summary>
    public string GenerateDispatcherClass(uint index)
```

crates/bindings-csharp/Codegen/Module.cs
```diff
@@ -1525,14 +1532,16 @@ public static void Main() {
...
-                            {{string.Join(
-                                "\n",
-                                views.Array.Select(v =>
-                                    v.IsAnonymous
-                                        ? $"SpacetimeDB.Internal.Module.RegisterAnonymousView<{v.Name}ViewDispatcher>();"
-                                        : $"SpacetimeDB.Internal.Module.RegisterView<{v.Name}ViewDispatcher>();"
-                                )
-                            )}}
+
+                            {{string.Join("\n",
```
joshua-spacetime, line 1544:
```suggestion
                            // IMPORTANT: The order in which we register views matters.
                            // It must correspond to the order in which we call `GenerateDispatcherClass`.
                            // See the comment on `GenerateDispatcherClass` for more explanation.
                            {{string.Join("\n", 
```

sdks/csharp/examples~/regression-tests/client/Program.cs
joshua-spacetime: I think it would be valuable to encapsulate this in a helper and do this equality check for all of these view tests.

joshua-spacetime (approved):

sdks/csharp/examples~/regression-tests/client/Program.cs
JasonAtClockwork: Expanded the equality checks for all the view data

joshua-spacetime (approved):

PR #3692, opened 2025-11-18
Fix #3687

bfops:
# Description of Changes

Apparently, when `spacetime generate` passes files to a linter, the files are passed in as relative paths, which did not play properly with `dotnet format` (so we effectively borked it).

None.

1

# Testing

- [x] Unity testsuite no longer fails on expected diffs

### DIFF
--- a/crates/cli/src/tasks/csharp.rs
+++ b/crates/cli/src/tasks/csharp.rs
@@ -131,7 +131,10 @@ pub(crate) fn dotnet_format(project_dir: &Path, files: impl IntoIterator<Item = 
             // `dotnet format` will interpret those paths relative to `project_dir`.
             files
                 .into_iter()
-                .map(|f| if f.is_absolute() { f } else { cwd.join(f) })
+                .map(|f| {
+                    let f = if f.is_absolute() { f } else { cwd.join(f) };
+                    f.canonicalize().expect("Failed to canonicalize path: {f}")
+                })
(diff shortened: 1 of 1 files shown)


crates/cli/src/subcommands/generate.rs
```diff
@@ -285,7 +285,7 @@ impl clap::ValueEnum for Language {
 }
 
 impl Language {
-    fn format_files(&self, project_dir: &PathBuf, generated_files: BTreeSet<PathBuf>) -> anyhow::Result<()> {
+    fn format_files(&self, project_dir: &Path, generated_files: BTreeSet<PathBuf>) -> anyhow::Result<()> {
```
bfops, line 288: clippy wanted me to change it 🤷

crates/cli/src/tasks/csharp.rs
```diff
@@ -119,17 +129,16 @@ pub(crate) fn dotnet_format(project_dir: &PathBuf, files: impl IntoIterator<Item
...
                 "--include",
             ]
             .into_iter()
             .map_into::<OsString>(),
-            files.into_iter().map_into(),
+            files.into_iter(),
         ),
     )
-    // This is important because we're running with `--folder`. We want to format the right folder!
-    .dir(project_dir)
```
bfops, line 132: just moved this since it turns out `--folder` can take an explicit param

PR #3693, opened 2025-11-18
CI - Fix empty lookup

bfops:
# Description of Changes

When the author isn't in the lookup table, we were getting `null` instead of the empty string. This fixes that.

None. CI only.

1

# Testing

None


jdetter (approved):
LGTM

PR #3696, opened 2025-11-19
Move typescript tests to new runner

jdetter:
# Description of Changes

- This just moves the typescript tests onto the custom runner.

|Test|Before|After|% Change|
|---|---|---|---|
|Typescript - Lint|1m|<1m|No change|
|Typescript - Build + Test|14m|5m|64%|

None

1 - if the typescript tests become flaky this could be the reason.

# Testing

- [x] The typescript CI still passes and is faster.


bfops (approved):
LGTM if it passes CI

PR #3704, opened 2025-11-20
Procedures: fix scheduling

### LINKED ISSUE #3645: Procedures: enable executing scheduled procedures
We currently have syntax in Rust for marking procedures as scheduled, but the host can't actually invoke them. Fix up the host's scheduler infrastructure so that scheduling a procedure results in it being run at approximately the appropriate time, then add a test of this functionality.

Centril:
# Description of Changes

Reworks how `SchedulerActor::handle_queued` works so that it first determines the parameters of the call to a reducer or the parameters of the call to the procedure. This also enables the removal of the special case `call_scheduled_reducer`.

Fixes #3645.

None

2

# Testing

A test `schedule_procedure` is added.


crates/core/src/host/module_host.rs
```diff
@@ -725,6 +671,26 @@ pub struct CallProcedureParams {
     pub args: ArgsTuple,
 }
 
+impl CallProcedureParams {
+    /// Returns a set of parameters for a call that came from within
+    /// and without a client/caller/request_id.
```
gefjon:
```suggestion
    /// Returns a set of parameters for an internal call without a client/caller/request_id.
```
"within and without" is a little too poetic for a doc comment, I think.

crates/core/src/host/scheduler.rs
```diff
@@ -327,72 +331,89 @@ impl SchedulerActor {
...
+        // This also lets us know whether to call a reducer or procedure.
+        let params = call_params_for_queued_item(&module_host.info, &db, item);
+        let (timestamp, params) = match params {
+            // If the function was already deleted,
+            // leave the `ScheduledFunction` in the database for when the module restarts.
+            Ok(None) => return,
+            Ok(Some(params)) => params,
+            Err(_) => {
+                if let Some(id) = id {
+                    // TODO: Handle errors here?
```
gefjon: Log an error?

sdks/rust/tests/procedure-client/src/main.rs
```diff
@@ -310,5 +313,44 @@ fn exec_procedure_http_err() {
...
+    connect_then(&test_counter, {
+        move |ctx| {
+            ctx.db().proc_inserts_into().on_insert(move |_, row| {
+                assert_eq!(row.x, 42);
+                assert_eq!(row.y, 24);
+
+                // Ensure that the elapsed time
+                // between the reducer and procedure
+                // is at least 1 second
+                // but no more than 1 second + 50 milliseconds.
```
gefjon: That seems optimistic, given that we run these tests in CI with a debug build of the host.

crates/core/src/host/scheduler.rs
```diff
@@ -327,72 +331,89 @@ impl SchedulerActor {
             return;
         };
         let db = module_host.replica_ctx().relational_db.clone();
-        let module_host_clone = module_host.clone();
 
-        let res = tokio::spawn(async move { module_host.call_scheduled_reducer(item).await }).await;
+        // Determine the call params.
+        // This also lets us know whether to call a reducer or procedure.
+        let params = call_params_for_queued_item(&module_host.info, &db, item);
```
Shubham8287: The current code fetches the parameters and executes the reducer within the same `tx` for atomicity.
I think this leaves the room for `params` to be modified by the item they are used for executeion.

sdks/rust/tests/procedure-client/src/main.rs
Shubham8287: Second this, I witnessed the replication tests with even larger time margins being flaky in CI :|

gefjon:
It occurs to me - for a procedure, you have to delete the schedule row before invoking, not after. If execution aborts (host crash or w/e) midway through a scheduled procedure, it's not correct to retry that procedure on restart the way it is for a scheduled reducer, so you need the row to be gone by the time the procedure performs any observable side effect. I think that this means that we should:

- When the schedule fires, open a mut tx.
- Within that TX, determine whether the function is a reducer or a procedure.
- If it's a reducer, use the mut TX to execute the reducer, as normal.
- If it's a procedure, delete the row, commit the TX, then invoke the procedure without a TX open.

There's some additional complexity here that we'll get to if/when we want to implement on-abort handling, but we haven't been considering that part of the MVP.

Shubham8287:
> * If it's a procedure, delete the row, commit the TX, then invoke the procedure without a TX open.

This is a atmost once gurantee then, we have to also cosider for the case when Host fatals or shutdown after row deletion, procedure will never run.

Edit: nvm, just read, you mentioned it is not required to retry for this case.

sdks/rust/tests/procedure-client/src/main.rs
Centril: Alright, I'll make it +1 second instead of 50ms.

gefjon (approved):
:excellent:

PR #3706, opened 2025-11-20
Views: index readsets

Shubham8287:
# Description of Changes
Precise index readsets

fixes #https://github.com/clockworklabs/SpacetimeDBPrivate/issues/2118

NA

2.5

Potential to regress performance.

# Testing
Updated smoketests.

### DIFF
--- a/smoketests/tests/views.py
+++ b/smoketests/tests/views.py
@@ -254,15 +254,6 @@ UPDATE player_state SET level = 9 WHERE id = 42;
 """,
         )
 
-        # Call the view again
-        self.spacetime(
-            "sql",
-            self.database_identity,
-            """
-Select * FROM player_state WHERE id = 42;
-""",
-        )
-
         # On fourth call, after updating the dependent row, the view is re-evaluated
         logs = self.logs(100)
         self.assertEqual(logs.count(player_called_log), 3)
@@ -315,7 +306,7 @@ INSERT INTO player_info (id, age, level) VALUES (1, 25, 7);
             "sql",
             self.database_identity,
             """\
-INSERT INTO player_info (id, age, level) VALUES (2, 30, 8);
+INSERT INTO player_info (id, age, level) VALUES (2, 25, 8);
 """,
         )
 


crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -84,6 +84,7 @@ pub struct ViewCallInfo {
 #[derive(Default)]
 pub struct ViewReadSets {
     tables: IntMap<TableId, TableReadSet>,
+    indexes: IntMap<TableId, IndexColReadSet>,
```
joshua-spacetime: Could we put `IndexColReadSet` in `TableReadSet`, so that `TableReadSet` tracks all the reads for a particular table - full scans and index scans?

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -112,16 +113,78 @@ impl ViewReadSets {
...
+                let key_map = col_map.entry(cols).or_default();
+                for (key, other_views) in other_key_map {
+                    key_map.entry(key).or_default().extend(other_views);
+                }
+            }
+        }
     }
 }
 
+type IndexKeyReadSet = HashMap<AlgebraicValue, HashSet<ViewCallInfo>>;
```
joshua-spacetime, line 143: Another option is to map a `ViewCallInfo` to all of the `AlgebraicValue`s that it reads so that we don't store a bunch of duplicate copies of the same `ViewCallInfo`. The downside to this approach though is that we have to iterate over all `ViewCallInfo`s and check each of their value sets when checking write sets against read sets.

I think I like the approach you've taken, but I would say that if we go with it, we should remove `view_name` from `ViewCallInfo`, so that we're not cloning the same string a bunch of times. We could also `Arc` it, but I'm not sure that's any less work since we'd have to figure out where to clone it from.

smoketests/tests/views.py
```diff
@@ -214,10 +224,41 @@ def test_a_view_materialization(self):
         )
 
         self.call_player_view()
-        #On third call, after invalidation, the view is evaluated again
         logs = self.logs(100)
         self.assertEqual(logs.count(player_called_log), 2)
 
+        # Updating the row that the view depends on should trigger re-evaluation
```
joshua-spacetime, line 248: Great test. Could we have a test for both a single and multi-column index?

crates/datastore/src/locking_tx_datastore/mut_tx.rs
Shubham8287: I replaced `view_name` to `fn_ptr` field in `ViewCallInfo`.

smoketests/tests/views.py
Shubham8287: done

crates/datastore/src/locking_tx_datastore/mut_tx.rs
Shubham8287: done

smoketests/tests/views.py
```diff
@@ -214,10 +242,102 @@ def test_a_view_materialization(self):
...
+        self.spacetime(
+            "sql",
+            self.database_identity,
+            """
+UPDATE player_state SET level = 9 WHERE id = 42;
+""",
+        )
+
+        # Call the view again
```
joshua-spacetime: Why are you calling the view again here?

smoketests/tests/views.py
```diff
@@ -214,10 +242,102 @@ def test_a_view_materialization(self):
...
+        logs = self.logs(100)
+        self.assertEqual(logs.count(player_called_log), 2)
+
+
+        # Inserting a row that does not match should not trigger re-evaluation
+        self.spacetime(
+            "sql",
+            self.database_identity,
+            """\
+INSERT INTO player_info (id, age, level) VALUES (2, 30, 8);
```
joshua-spacetime:
```suggestion
INSERT INTO player_info (id, age, level) VALUES (2, 25, 8);
```
Just to catch any prefix bugs

PR #3714, opened 2025-11-20
Added staging to allowable issuers

cloutiertyler:
# Description of Changes

This modifies the environment variable that we pass in for requiring SpacetimeAuth to publish so that we can put different issuer values for live and staging.

This requires a new environment variable or it skips this requirement.

1

# Testing

I have not tested this change locally.

### DIFF
--- a/crates/client-api/src/routes/database.rs
+++ b/crates/client-api/src/routes/database.rs
@@ -45,8 +45,7 @@ use spacetimedb_schema::auto_migrate::{
 use super::subscribe::{handle_websocket, HasWebSocketOptions};
 
 fn require_spacetime_auth_for_creation() -> Option<String> {
-    // If the string is a non-empty value, require SpacetimeAuth for database creation
-    // and return the value for logging purposes.
+    // If the string is a non-empty value, return the string to be used as the required issuer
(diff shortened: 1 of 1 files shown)


crates/client-api/src/routes/database.rs
```diff
@@ -44,21 +44,31 @@ use spacetimedb_schema::auto_migrate::{
 
 use super::subscribe::{handle_websocket, HasWebSocketOptions};
 
-fn require_spacetime_auth_for_creation() -> bool {
-    env::var("TEMP_REQUIRE_SPACETIME_AUTH").is_ok_and(|v| !v.is_empty())
+fn require_spacetime_auth_for_creation() -> Option<String> {
+    // If the string is a non-empty value, require SpacetimeAuth for database creation
+    // and return the value for logging purposes.
```
cloutiertyler:
```suggestion
    // If the string is a non-empty value, return the string to be used as the required issuer
```

jdetter (approved):
Have not tested this myself but this matches the infra PR and seems correct 👍

jsdt (commented):
It could be easier to roll this out if we handle both of them (by checking for the old env var if the new one isn't present): https://github.com/clockworklabs/SpacetimeDB/commit/d26d7834da0bb5d692df876956ba881dd725b11b

PR #3716, opened 2025-11-20
Use custom runner for building CLI

jdetter:
# Description of Changes

- This has 2 benefits:
1. `ubuntu-latest` recently changed to 24.04 instead of 22.04 so that broke the last CLI on some systems. This will fix that issue.
2. Performance - this should build much much faster.

None

2 - if the CLI fails to build this might be why.

# Testing

- [x] Tested via today's release: https://github.com/clockworklabs/SpacetimeDB/actions/runs/19550648024/job/55980960439

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -19,7 +19,7 @@ jobs:
             runner: spacetimedb-new-runner
             container:
               image: localhost:5000/spacetimedb-ci:latest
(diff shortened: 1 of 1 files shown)


bfops (approved):
LGTM. I think it would be fine to leave the other variants on a single line, but I don't feel strongly one way or another.

PR #3719, opened 2025-11-20
HTTP followup: remove `http` dep from `spacetimedb_lib`

coolreader18:
# Description of Changes

Follow up to #3684. Moves `Error` and `Timeout` out of lib, so that we don't have to implement `SpacetimeType` for them, and then removes the http dependency altogether, so that `lib` can be leaner. I also got rid of the separate `HttpValue` type, since it only really exists to mirror the `http` crate and typescript won't make use of it.

1

# Testing

n/a - just code movement.

### DIFF
--- a/crates/lib/src/http.rs
+++ b/crates/lib/src/http.rs
@@ -1,14 +1,13 @@
 //! `SpacetimeType`-ified HTTP request, response and error types,
 //! for use in the procedure HTTP API.
 //!
-//! The types here are all mirrors of various types within the [`http`] crate.
+//! The types here are all mirrors of various types within the `http` crate.
 //! That crate's types don't have stable representations or `pub`lic interiors,
 //! so we're forced to define our own representation for the SATS serialization.
 //! These types are that representation.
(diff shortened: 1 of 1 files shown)


crates/lib/src/http.rs
```diff
@@ -212,19 +126,52 @@ impl From<Version> for http::Version {
...
+#[sats(crate = crate, name = "HttpHeaders")]
 pub struct Headers {
     // SATS doesn't (and won't) have a multimap type, so just use an array of pairs for the ser/de format.
     entries: Box<[HttpHeaderPair]>,
 }
 
-impl From<http::HeaderMap<http::HeaderValue>> for Headers {
-    fn from(value: http::HeaderMap<http::HeaderValue>) -> Headers {
+// `http::header::IntoIter` only returns the `HeaderName` for the first
+// `HeaderValue` with that name, so we have to manually assign the names.
```
gefjon, line 71: God, I hate the HTTP crate...

PR #3720, opened 2025-11-21
Views: snapshot to skip ephermeral tables

### LINKED ISSUE #3715: Snapshots should ignore ephemeral tables
Ephemeral tables are not written to the commitlog. They also shouldn't be captured by snapshots.

Shubham8287:
# Description of Changes.
fixes #3715 

The patch makes snapshots to skip ephemeral tables.

NA

1

# Testing

- [ ] 
- [ ]


crates/datastore/src/locking_tx_datastore/datastore.rs
```diff
@@ -282,12 +282,8 @@ impl Locking {
             tx_offset,
         );
 
-        let CommittedState {
-            ref mut tables,
-            ref blob_store,
-            ..
-        } = *committed_state;
-        let snapshot_dir = repo.create_snapshot(tables.values_mut(), blob_store, tx_offset)?;
+        let (tables, blob_store) = committed_state.persistent_tables_and_blob_store();
```
joshua-spacetime, line 285: Am I correct that if we remove the table before we create the snapshot, it will still be created when we replay from snapshot just by virtue of it being recorded in `st_table`?

~~A potential problem with this is that the view's metadata columns are not currently stored in `st_column`. So when we go to create an empty view table from `st_table`, it will use the columns from `st_column`, which will result in an error as soon as we try to insert into it (assuming it's a client specific view).~~

~~This isn't a problem currently because even if we snapshot a non-empty view table, by default we clear all of the view tables on init. So we always have the correct schema.~~

~~Even though it's technically not broken currently, I still don't think we want view tables to show up in snapshots since we don't persist the read sets. So I would vote for what you have here, but we'll need to make sure we add the view metadata columns to `st_column`. And if possible, it would be great to add a smoketest that can replay from a snaphsot (not sure if we can control this from a smoketest) and re-invoke a view, since that test would break without the changes to `st_column`.~~

joshua-spacetime: Never mind. I was confusing `st_view_column` with `st_column`. `st_column` does contain all of the internal columns of the view table.

joshua-spacetime (approved):
Could you add a test that previously would have snapshotted a view table before you merge?

joshua-spacetime:
We sometimes see failures for `test_view_tables_are_ephemeral` like this one [here](https://github.com/clockworklabs/SpacetimeDB/actions/runs/19942972041/job/57185321444?pr=3812). I believe this is due to the fact that we're currently snapshotting ephemeral tables (before this fix). Do you concur @Shubham8287?

Shubham8287:
yeah, as per logs, It looks like due to snapshot, I wasn't aware this test can do snapshotting:

```
[INFO  spacetimedb::db::relational_db] [0000000000000000000000000000000000000000000000000000000000000000] DATABASE: durable_tx_offset is Some(3)
[INFO  spacetimedb::db::relational_db] [0000000000000000000000000000000000000000000000000000000000000000] DATABASE: restoring snapshot of tx_offset 3
[INFO  spacetimedb::db::relational_db] [0000000000000000000000000000000000000000000000000000000000000000] DATABASE: read snapshot of tx_offset 3 in 4.710688ms
[INFO  spacetimedb::db::relational_db] [0000000000000000000000000000000000000000000000000000000000000000] DATABASE: restored from snapshot of tx_offset 3 in 169.677µs
[INFO  spacetimedb::db::relational_db] [0000000000000000000000000000000000000000000000000000000000000000] DATABASE: applying transaction history...
[DEBUG spacetimedb_co

```

joshua-spacetime:
> Could you add a test that previously would have snapshotted a view table before you merge?

If we want to test with views specifically this would have to be a smoketest, but we can't control snapshotting from a smoketest. So I'm fine if you just want to use `test_view_tables_are_ephemeral` as a suitable regression test. It definitely shouldn't be flaky anymore after this change lands.

crates/core/src/db/relational_db.rs
```diff
@@ -2591,7 +2591,40 @@ mod tests {
...
+        insert_view_row(&stdb, view_id, table_id, Identity::ONE, 10)?;
+        insert_view_row(&stdb, view_id, table_id, Identity::ZERO, 20)?;
+
+        assert!(
+            !project_views(&stdb, table_id, Identity::ZERO).is_empty(),
+            "View table should NOT be empty after insert"
+        );
+
+        // Reopen the database — view tables must not persist
+        let stdb = stdb.reopen()?;
```
Shubham8287, line 2609: I believe flakiness was coming from this [patch]( https://github.com/clockworklabs/SpacetimeDB/blob/3e7ab2ccac96fe4257d6bf222440cc3d190a87b0/crates/core/src/db/relational_db.rs#L1784-L1792) which gets called from `reopen`, code snippet is trying to take a snapshot and then calling `open_db` instantly, which might have or have not use snapshot created by that time.

I splitted the tests in two - `test_view_tables_are_ephemeral_in_commitlog:` which explicitly does not create snapshot repo.


`test_view_tables_are_ephemeral_with_snapshot`: which calls method `stdb.take_snapshot` before reopening db.

Issue #3729, opened 2025-11-21
Spacetime Dev fails on publish args mistake 1.9


xDovos:
https://github.com/clockworklabs/SpacetimeDB/blob/e26e7fee502eb84c9d677def7052508c6d219cd6/crates/cli/src/subcommands/dev.rs#L410-L423

the --delete-data needs to be "--delete-data=enum" but is currently "--delete-data enum" and thus the publish fails on "error: unexpected argument 'on-conflict' found"

xDovos:
it also seems that the publish
spacetime publish --delete-data=on-conflict tradinggame 
is the same as using always and asks for deletion and then deletes even with no changes to the module code.

cloutiertyler:
Thank you, I will investigate

cloutiertyler:
@xDovos Thank you for filing this, I genuinely don't know what madness was going on in my head at the time

Here's my PR to fix this: https://github.com/clockworklabs/SpacetimeDB/pull/3730

PR #3730, opened 2025-11-21
Fixes issues with `--delete-data=on-conflict`

cloutiertyler:
# Description of Changes

Fixes https://github.com/clockworklabs/SpacetimeDB/issues/3729

I genuinely don't know what came over me.

None

1.5 very straightforward but not strictly trivial

# Testing
Adds automated integration tests (written in Rust and run with `cargo test`, although note this comment from @matklad about integration tests for the future https://internals.rust-lang.org/t/running-test-crates-in-parallel/15639/2):

- [x] Can publish an updated module if no migration is required
- [x] Can publish an updated module if auto-migration is required (with the yes-break flag true/false)
- [x] Cannot publish if a manual migration is required
- [x] Can publish if a manual migration is required but the user specified `--delete-data`
- [x] Can publish if a manual migration is required by the user specified `--delete-data=on-conflict`
- [x] No data deletion occurs if no migration is required and `--delete-data=on-conflict` is specified

### DIFF
--- a/crates/bindings-macro/src/table.rs
+++ b/crates/bindings-macro/src/table.rs
@@ -586,6 +586,21 @@ fn parse_default_attr(attr: &syn::Attribute, ident: &Ident) -> syn::Result<Colum
     ))
 }
 
+use std::collections::HashSet;
+use std::sync::Mutex;
+
+// Same struct can be annotated with `#[spacetimedb::table]` multiple times.
+// This mutex keeps track of which structs we've already generated code for.
+// This avoids duplicate definitions when the same struct is annotated multiple times.
+static GENERATED_STRUCTS: std::sync::LazyLock<Mutex<HashSet<String>>> =
+    std::sync::LazyLock::new(|| Mutex::new(HashSet::new()));
+
+fn is_first_appearance(struct_name: &str) -> bool {
+    let mut set = GENERATED_STRUCTS.lock().expect("mutex poisoned");
+
+    set.insert(struct_name.to_string())
+}
+
 pub(crate) fn table_impl(mut args: TableArgs, item: &syn::DeriveInput) -> syn::Result<TokenStream> {
     let vis = &item.vis;
     let sats_ty = sats::sats_type_from_derive(item, quote!(spacetimedb::spacetimedb_lib))?;
@@ -593,6 +608,9 @@ pub(crate) fn table_impl(mut args: TableArgs, item: &syn::DeriveInput) -> syn::R
     let original_struct_ident = sats_ty.ident;
     let table_ident = &args.name;
     let view_trait_ident = format_ident!("{}__view", table_ident);
+    let query_trait_ident = format_ident!("{}__query", table_ident);
+    let query_cols_struct = format_ident!("{}Cols", original_struct_ident);
+    let query_ix_cols_struct = format_ident!("{}IxCols", original_struct_ident);
     let table_name = table_ident.unraw().to_string();
     let sats::SatsTypeData::Product(fields) = &sats_ty.data else {
         return Err(syn::Error::new(Span::call_site(), "spacetimedb table must be a struct"));
@@ -925,6 +943,97 @@ pub(crate) fn table_impl(mut args: TableArgs, item: &syn::DeriveInput) -> syn::R
         }
     };
 
+    let cols_struct_fields = fields.iter().map(|col| {
+        let ident = col.ident.unwrap();
+        let ty = &col.ty;
+
+        quote! {
+            pub #ident: spacetimedb::query_builder::Col<#original_struct_ident, #ty>,
+        }
+    });
+
+    let ix_cols_struct_fields = indices.iter().filter_map(|index| {
+        let ident = index.accessor_name.clone();
+        let ty = match &index.kind {
+            ValidatedIndexType::BTree { cols } => {
+                if cols.len() == 1 {
+                    &cols[0].ty
+                } else {
+                    return None;
+                }
+            }
+            ValidatedIndexType::Direct { col } => &col.ty,
+        };
+
+        Some(quote! {
+            pub #ident: spacetimedb::query_builder::IxCol<#original_struct_ident, #ty>,
+        })
+    });
+
+    let cols_init = fields.iter().map(|col| {
+        let ident = col.ident.as_ref().unwrap();
+
+        quote! {
+            #ident: spacetimedb::query_builder::Col::new(_table_name, stringify!(#ident)),
+        }
+    });
+
+    let ix_cols_init = indices.iter().map(|index| {
+        let ident = index.accessor_name;
+        match &index.kind {
+            ValidatedIndexType::BTree { cols } => {
+                if cols.len() != 1 {
+                    return quote! {};
+                }
+            }
+            ValidatedIndexType::Direct { .. } => {}
+        }
+
+        quote! {
+            #ident: spacetimedb::query_builder::IxCol::new(_table_name, stringify!(#ident)),
+        }
+    });
+
+    let query_builder_helper_structs = quote_spanned! {table_ident.span()=>
+           #[allow(non_camel_case_types, dead_code)]
+           pub struct #query_cols_struct{
+               #(#cols_struct_fields)*
+           }
+
+           impl spacetimedb::query_builder::HasCols for #original_struct_ident  {
+               type Cols = #query_cols_struct;
+                fn cols(_table_name: &'static str) -> Self::Cols {
+                     #query_cols_struct {
+                          #(#cols_init)*
+                     }
+                }
+           }
+
+        #[allow(non_camel_case_types, dead_code)]
+        pub struct #query_ix_cols_struct{
+            #(#ix_cols_struct_fields)*
+        }
+        impl spacetimedb::query_builder::HasIxCols for #original_struct_ident {
+            type IxCols = #query_ix_cols_struct;
+            fn ix_cols(_table_name: &'static str) -> Self::IxCols {
+                #query_ix_cols_struct {
+                    #(#ix_cols_init)*
+                }
+            }
+        }
+
+    };
+
+    let table_query_handle_def = quote! {
+           #[allow(non_camel_case_types, dead_code)]
+           #vis trait #query_trait_ident {
+               fn #table_ident(&self) -> spacetimedb::query_builder::Table<#original_struct_ident> {
+                   spacetimedb::query_builder::Table::new(stringify!(#table_ident))
+               }
+           }
+           impl #query_trait_ident for spacetimedb::QueryBuilder {}
+    };
+
     let tablehandle_def = quote! {
         #[allow(non_camel_case_types)]
         #[non_exhaustive]
@@ -935,6 +1044,15 @@ pub(crate) fn table_impl(mut args: TableArgs, item: &syn::DeriveInput) -> syn::R
         #[allow(non_camel_case_types)]
         #[non_exhaustive]
         #vis struct #viewhandle_ident {}
+    };
+
+    let struct_name = original_struct_ident.to_string();
+    let is_first_table = is_first_appearance(&struct_name);
+
+    let struct_level_query_impl = if is_first_table {
+        quote! { #query_builder_helper_structs }
+    } else {
+        quote! {}
     };
 
     let emission = quote! {
@@ -949,6 +1067,8 @@ pub(crate) fn table_impl(mut args: TableArgs, item: &syn::DeriveInput) -> syn::R
 
         #tablehandle_def
         #viewhandle_def
+        #table_query_handle_def
+        #struct_level_query_impl
 
         const _: () = {
             impl #tablehandle_ident {
--- a/crates/bindings/src/lib.rs
+++ b/crates/bindings/src/lib.rs
@@ -870,6 +870,7 @@ pub use spacetimedb_bindings_macro::procedure;
 pub use spacetimedb_bindings_macro::view;
 
 pub struct QueryBuilder {}
+pub use query_builder::Query;
 
 /// One of two possible types that can be passed as the first argument to a `#[view]`.
 /// The other is [`ViewContext`].
--- a/crates/bindings/src/query_builder.rs
+++ b/crates/bindings/src/query_builder.rs
@@ -4,6 +4,7 @@ pub mod table;
 
 pub use expr::*;
 pub use join::*;
+use spacetimedb_lib::{sats::impl_st, AlgebraicType, SpacetimeType};
 pub use table::*;
 
 pub struct Query<T> {
@@ -24,6 +25,8 @@ impl<T> Query<T> {
     }
 }
 
+impl_st!([T: SpacetimeType] Query<T>, ts => AlgebraicType::option(T::make_type(ts)));
+
 #[cfg(test)]
 mod tests {
     use spacetimedb_lib::{sats::i256, TimeDuration};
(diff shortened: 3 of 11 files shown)


crates/cli/src/subcommands/publish.rs
bfops, line 378: is this comment still correct?

bfops: shouldn't this case also apply `builder = builder.query(&[("clear", true)]);` in the `Always` case?

bfops (changes requested):
From a black box perspective, I would like you to test the following cases, with each value of `--delete-data`:
- no migration required
- auto-migrateable (with the yes-break flag true/false)
- manual migration required

Given the original issue, please also spot-test `spacetime dev`.

jdetter:
Shall we do a hotfix release with this?

crates/cli/src/subcommands/publish.rs
cloutiertyler: No it is not, and yes it should. Modifying and adding tests.

crates/cli/Cargo.toml
```diff
@@ -86,6 +86,11 @@ notify.workspace = true
 
 [dev-dependencies]
 pretty_assertions.workspace = true
+fs_extra.workspace = true
+assert_cmd = "2"
+predicates = "3"
+portpicker = "0.1"
+reqwest = { version = "0.12", features = ["blocking", "json"] }
```
cloutiertyler: I'm not sure if I should add these to the workspace because they're dev only dependencies

bfops: 🤷

crates/cli/src/subcommands/publish.rs
```diff
@@ -367,11 +349,35 @@ async fn apply_pre_publish_if_needed(
...
+                    force,
+                    format!("Are you sure you want to proceed? [deleting {}]", name_or_identity).as_str(),
+                )? {
+                    anyhow::bail!("Aborting");
+                }
+                builder = builder.query(&[("clear", true)]);
             }
             PrePublishResult::AutoMigrate(auto) => {
+                if clear_database == ClearMode::Always {
+                    println!("Auto-migration, does NOT require clearing the database, but proceeding with database clear due to --delete-data=always.");
```
bfops: this case doesn't seem to be tested? If I remove this block and run `cargo test publish`, they all still seem to pass.

My default position is that we should do all of the tests described in https://github.com/clockworklabs/SpacetimeDB/pull/3730#pullrequestreview-3494666487:

- no migration required
  - delete-data=no
  - delete-data=on-conflict
  - delete-data=yes
- auto-migrateable
  - delete-data=no, yes-break=false
  - delete-data=no, yes-break=true
  - delete-data=on-conflict, yes-break=false
  - delete-data=on-conflict, yes-break=true
  - delete-data=yes, yes-break=false
  - delete-data=yes, yes-break=true
- manual migration required
  - delete-data=no
  - delete-data=on-conflict
  - delete-data=yes

crates/cli/src/tasks/mod.rs
```diff
@@ -13,10 +13,14 @@ pub fn build(
     project_path: &Path,
     lint_dir: Option<&Path>,
     build_debug: bool,
+    features: Option<&std::ffi::OsString>,
 ) -> anyhow::Result<(PathBuf, &'static str)> {
     let lang = util::detect_module_language(project_path)?;
+    if features.is_some() && lang != ModuleLanguage::Rust {
+        anyhow::bail!("The --features option is only supported for Rust modules.");
+    }
```
bfops, line 21: nit: imo this complaint should happen "near" the actual arg parsing, rather than down here.

crates/cli/tests/util.rs
```diff
@@ -0,0 +1,146 @@
...
+        let addr: SocketAddr = format!("127.0.0.1:{port}").parse().unwrap();
+        let address = addr.to_string();
+        let host_url = format!("http://{}", addr);
+
+        // Workspace root for `cargo run -p ...`
+        let workspace_dir = env!("CARGO_MANIFEST_DIR");
+
+        Self::build_prereqs(workspace_dir);
+
+        let mut cargo_args = vec!["run", "-p", "spacetimedb-cli", "--"];
```
bfops, line 46: this seems to somewhat bork test parallelization, since they seem to get stuck on "waiting for lock on build directory". Also I get warnings about tests running for >60s.

I would have honestly expected these to be smoketests, since they involve running binaries and coordinating between them (and that's why the smoketests already have infrastructure for spinning up servers).

At this point I feel a bit stuck between rock and hard place, since I think it's a bit of a code smell to be duplicating this testing infrastructure, but I also don't really want to ask you to rewrite it.

crates/cli/Cargo.toml
gefjon:
```suggestion
reqwest = { workspace = true, features = ["blocking", "json"] }
```

crates/cli/src/tasks/mod.rs
bfops: ah this is where we parse lang

crates/cli/tests/util.rs
bfops: after adding some more cases to cover what I asked for in https://github.com/clockworklabs/SpacetimeDB/pull/3730#pullrequestreview-3494666487, the tests take over 3 minutes to run. Just these tests.

crates/cli/src/subcommands/publish.rs
bfops: tests added

bfops: unfortunately they still would not catch this case, because the tests do not check whether or not data was actually deleted.

crates/cli/tests/util.rs
bfops: Due to time constraints, I am not working on that.

crates/cli/src/subcommands/publish.rs
bfops: Due to time constraints, I am not working on that.

crates/cli/src/subcommands/build.rs
```diff
@@ -22,6 +22,13 @@ pub fn cli() -> clap::Command {
                 .default_value("src")
                 .help("The directory to lint for nonfunctional print statements. If set to the empty string, skips linting.")
         )
+        .arg(
+            Arg::new("features")
+                .long("features")
+                .value_parser(clap::value_parser!(OsString))
+                .required(false)
+                .help("Additional features to pass to the build process (e.g. `--features feature1,feature2` for Rust modules).")
+        )
```
bfops, line 34: I think this should just be a catchall `--extra-build-args` param, which allows us to pass arbitrary extra options to the build. That would let this 1) be more general than just `--features`, and 2) apply to all languages.

Thoughts @jdetter ?

crates/cli/src/subcommands/dev.rs
```diff
@@ -413,15 +413,14 @@ async fn generate_build_and_publish(
...
-        project_path_str,
-        "--yes",
-        "--delete-data",
-        clear_flag,
+        "publish".to_string(),
+        database_name.to_string(),
+        "--project-path".to_string(),
+        project_path_str.to_string(),
+        "--yes".to_string(),
+        format!("--delete-data={}", clear_flag),
```
bfops, line 421: we should spot test that this fix worked

bfops: just running `spacetime dev` seems to work now 🤷 at a glance it looks like it always passes this flag so.. I guess it's tested?

crates/cli/src/subcommands/build.rs
jdetter: I agree, based on the time constraints I think we should use this as-is for now. However, we should still add this to the list of things to add to 2.0 as a P2 and if we get requests to make this more broad in the meantime we'll bump the priority to a P1. Is that fair?

crates/cli/src/subcommands/publish.rs
```diff
@@ -67,6 +67,7 @@ pub fn cli() -> clap::Command {
         .arg(
             Arg::new("break_clients")
                 .long("break-clients")
+                .alias("yes-break-clients")
```
jdetter: When you have a flag as a CLI parameter it's usually assumed that if the flag is set it is the positive case. For example, if you had a param `--upgrade` the assumption is this means "yes, I'd like to upgrade". Having `--yes-break-clients` is redundant IMO. If you didn't want to break clients you would just not set the flag.

crates/cli/src/subcommands/build.rs
cloutiertyler: I considered this. I would prefer this, but it was more complicated because we don't actually execute a CLI command for TypeScript. Not sure it should hold up the fixes though.

crates/cli/tests/util.rs
cloutiertyler: > I would have honestly expected these to be smoketests, since they involve running binaries and coordinating between them (and that's why the smoketests already have infrastructure for spinning up servers).

Our plan was to eventually move the python smoke tests to Rust anyway and to execute the smoke tests in parallel. I implemented them as Rust integration tests because it accomplishes that goal.

> after adding some more cases to cover what I asked for in #3730 (review), the tests take over 3 minutes to run. Just these tests.

I assume this is because of the dirty-ing of the CARGO ENV files?

crates/cli/src/subcommands/build.rs
bfops: I have pushed a `hide(true)` on this flag since it's kind of WIP and we're really just adding it for unit tests

cloutiertyler: I think that makes sense

crates/cli/src/subcommands/publish.rs
```diff
@@ -175,11 +195,14 @@ pub async fn exec(mut config: Config, args: &ArgMatches) -> Result<(), anyhow::E
         let domain = percent_encoding::percent_encode(name_or_identity.as_bytes(), encode_set);
         let mut builder = client.put(format!("{database_host}/v1/database/{domain}"));
 
-        if clear_database != ClearMode::Always {
+        if clear_database == ClearMode::Always {
+            builder = prompt_and_clear(name_or_identity, force, builder)?;
```
bfops: note that this now only happens in the case where we've passed a `name_or_identity`, but I think this can only clobber something in that case anyway

bfops: oh actually `--clear-database` _requires_ that to be provided

crates/cli/tests/util.rs
bfops: I don't know why :shrug: I see a lot of things saying they're waiting for locks, but I was unable to find extra params for `cargo` that would give me more details on what they're waiting on, and how long.

crates/cli/src/subcommands/publish.rs
```diff
@@ -342,6 +347,9 @@ async fn apply_pre_publish_if_needed(
     force_break_clients: bool,
     force: bool,
 ) -> Result<reqwest::RequestBuilder, anyhow::Error> {
+    // The caller enforces this
+    assert!(clear_database != ClearMode::Always);
```
bfops, line 352: note: we had worrying behavior where one of the `teams` smoketests would fail if we were calling this function/endpoint unconditionally

bfops: I created https://github.com/clockworklabs/SpacetimeDB/pull/3778 with the state of things that triggers that issue (i.e. this PR before I refactored)

bfops: Created https://github.com/clockworklabs/SpacetimeDBPrivate/issues/2324 to track

crates/cli/tests/util.rs
cloutiertyler: Interesting. When I ran it locally, they would all wait for the lock on building spacetimedb-standalone and spacetimedb-cli, but once they were built they would all run quickly.

crates/cli/src/subcommands/publish.rs
cloutiertyler: I mentioned in the original PR description in the original PR that `--yes-break-clients` is not just saying "I would like to break clients". Indeed it won't break clients if there are no client breaking changes. It's saying I would like to skip the dialogue which asks to confirm if I would like to upgrade if there are breaking changes, which is the same behavior as the `--yes` command, which is why I suggested it here.

`--break-clients` is confusing because it does not break clients, nor does it change the behavior of the command in any way, except to skip the user confirmation.

From the original PR description:

> IMPORTANT SIDE NOTE: I would argue that `--break-clients` should really be renamed to `--yes-break-clients` because it actually behaves like the `--yes` force flag, but only for a subset of the user prompts. I have not made this change because it would be a breaking change, but if the reviewers agree, I will make this change.

No one commented on this during the review.

jdetter: I agree this is confusing, I feel like an enum would work best here? Something like: `--yes all|break-clients|migrate` (do we have other prompts?)

I think having `--yes`, `--yes-break-clients` and `--break-clients` is confusing but I understand the logic. If you want to revert the commit I made to pull out `--yes-break-clients` that would be fine with me, I'm not highly attached to the decision. I do have some preference here to just making it an enum because that's the simplest thing from a user perspective IMO.

I would like to add a ticket here for 2.0 so that we can simplify this. What do you think?

bfops: > `--break-clients` is confusing because it does not break clients, nor does it change the behavior of the command in any way, except to skip the user confirmation.

fwiw I don't know that `--yes-break-clients` is less confusing in this respect. Maybe `--allow-break-clients`?

> No one commented on this during the review.

I was the only reviewer and I was of the opinion "eh, if it makes you happy".

bfops: I like the `enum` idea generally

cloutiertyler: I agree, I like the enum. I would prefer the enum. That's probably how this should work. It's a breaking change though unless we keep the old `--break-clients`.

cloutiertyler: Let's leave it as is and add a 2.0 ticket to switch to an enum.

jdetter: @cloutiertyler @bfops My proposal for the ticket: https://github.com/clockworklabs/SpacetimeDB/issues/3784

crates/cli/src/subcommands/publish.rs
```diff
@@ -175,11 +195,14 @@ pub async fn exec(mut config: Config, args: &ArgMatches) -> Result<(), anyhow::E
         let domain = percent_encoding::percent_encode(name_or_identity.as_bytes(), encode_set);
         let mut builder = client.put(format!("{database_host}/v1/database/{domain}"));
 
-        if clear_database != ClearMode::Always {
+        if clear_database == ClearMode::Always {
```
bfops, line 199:
```suggestion
        // note that this only happens in the case where we've passed a `name_or_identity`, but that's required if we pass `--clear-database`.
        if clear_database == ClearMode::Always {
```

crates/cli/src/subcommands/build.rs
```diff
@@ -22,6 +22,16 @@ pub fn cli() -> clap::Command {
                 .default_value("src")
                 .help("The directory to lint for nonfunctional print statements. If set to the empty string, skips linting.")
         )
+        .arg(
+            // TODO: Make this into --extra-build-args (or something similar) that will get passed along to the language's compiler.
```
jdetter, line 26: Do we have a ticket for this one? Seems like this could be a good v2.0 ticket

bfops: Created https://github.com/clockworklabs/SpacetimeDBPrivate/issues/2346. FWIW I don't personally think we need to wait for 2.0. This API isn't intended to be user-facing, and it's hidden from the helptext.

Issue #3510, opened 2025-10-28
Procedures: C# module bindings


gefjon:
Implement 4 parts of module bindings:
* Recreate the module-side interface for procedures from #3444 in C# modules.
* Create C# implementation of Transactions for Procedures similar to Rust implementation #3638.
* Add procedure HTTP request support to C# similar to Rust implementation #3684.
* Updates docs

rekhoff:
#3809 (C# implementation of Transactions for Procedures) is now ready for review.

rekhoff:
General status update:
* #3732, the C# equivalent Implementation of #3444 - COMPLETE
* #3809, the C# equivalent Implementation of #3638 - COMPLETE
* The C# equivalent Implementation of #3684 - In Progress
With that in mind, I'm moving this issue back to In Progress while the #3684 C# equivalent implementation is being worked on.

rekhoff:
Update to HTTP support for Procedures in C# modules: PR #3944 has been opened, which should provide similar functionality to Rust's implementation in #3684.  With that PR open, all known tasks for this issue are now completed or in review, so I'll move this to In Review.

PR #3732, opened 2025-11-21
C# module bindings for Procedures 

rekhoff:
# Description of Changes

Implements the C# module bindings for Procedures (#3510)

None

2

# Testing

- [X] Locally tested against existing C# Procedures regression test client (minus the Transaction and HTTP portions, as those are not in this change).


JasonAtClockwork (approved):
Reviewed the code and ran the regression tests with no troubles; all looks good to me. The only question I have is if we should be adding checks to Codegen.Tests/fixtures/diag/Lib.cs and Codegen.Tests/fixtures/server/Lib.cs? Or maybe add that once all the pieces are in? 

One test that would be good is a procedure that returns a non-SpacetimeDB.Type. This should fail due to the ReturnType.BSATNName but would be good to track that.

JasonAtClockwork (commented):
Ahh just realized the client-side tests are not there for regression, I've added one locally just to make sure procedures are calling but you'll want to mirror what the /regression-test/procedure-client - once the TX pieces are in place we can probably delete that entire runner too.

rekhoff:
This are some great ideas for additional tests. I was intending on getting add a more robust set of tests once the other pieces are in place, and migrate the `procedure-client` into the existing `client` regression test. I'll add the test suggestions to the list to implement as part of the other procedure pieces implementations.

Issue #3503, opened 2025-10-27
Fix Angular builds


jdetter:
At least 2 community members have mentioned that we broke Angular builds since `v1.4.0`
- https://discord.com/channels/1037340874172014652/1431594353612296292/1431594353612296292
- https://discord.com/channels/1037340874172014652/1138987509834059867/1423467631893282879

Example error from discord:

```
✘ [ERROR] TS5097: An import path can only end with a '.ts' extension when 'allowImportingTsExtensions' is enabled. [plugin angular-compiler]

    node_modules/spacetimedb/src/sdk/client_api/index.ts:38:29:
      38 │ import { BsatnRowList } from './bsatn_row_list_type.ts';
```

## TODO

- [ ] Fix the build issue
- [ ] Add regression test to make sure we don't break this moving forward

JohnyBro:
Original reporter here, I think it's a typo but you mention Angular**JS** in the description which is the first angular version and is deprecated now. The issue is for the latest Angular v20.

gk-bbai:
We are also unable to use the latest version of SpacetimeDB (bindings generated with v1.8.0) with the TypeScript configuration generated by our Angular v19 project.

JohnyBro:
I tried to update to version 1.9 to see if the build works but I still get 2 errors.

1. Seems like there is a name issue in the generated typescript code

```
✘ [ERROR] TS2552: Cannot find name 'ClubRow'. Did you mean 'ClubsRow'? [plugin angular-compiler]

    src/module_bindings/index.ts:60:5:
      60 │   }, ClubRow),
         ╵      ~~~~~~~
```

```ts
const tablesSchema = __schema(
  __table({
    name: 'Clubs',
    indexes: [
      { name: 'ClubId', algorithm: 'btree', columns: [
        'clubId',
      ] },
    ],
    constraints: [
      { name: 'Clubs_ClubId_key', constraint: 'unique', columns: ['ClubId'] },
    ],
  }, ClubRow),
);
```

```csharp
public static partial class Module
{
    [Table(Name = "Clubs", Public = true)]
    public partial class Club
    {
        [PrimaryKey]
        [AutoInc]
        public uint ClubId;

        public string Name = "";
        public string Description = "";
        public Timestamp CreatedAt;
        public Timestamp UpdatedAt;
    }
}
```

2. SubscriptionBuilder error
```
[...]

PR #3737, opened 2025-11-24
Fixes reported issues with the TypeScript SDK

cloutiertyler:
# Description of Changes

Fixes:

- [This issue](https://discord.com/channels/1037340874172014652/1398209084699709492/1441556875647647816) by exporting the `SubscriptionHandle` type with the `REMOTE_MODULE` type applied.
- [This issue](https://discord.com/channels/1037340874172014652/1398209084699709492/1441559246213746749) by converting to `camelCase` for column names in code generation.
- Fixes an issue where `onMyReducer` callbacks were passing arguments as variadic params, while the types indicated they would be passed as an object. `onMyReducer((ctx, argA, argB, argC) => {})` vs onMyReducer((ctx, { argA, argB, argC}) => {})`
- [Fixes an issue](https://github.com/clockworklabs/SpacetimeDB/issues/3503#issuecomment-3566715928) where the table type name was used instead of the table name in code generation for constructing tables.
- [Fixes issue](https://discord.com/channels/1037340874172014652/1398209084699709492/1441886030436499466) with `ScheduleAt` being used in non-table types.
- [Fixes issue](https://github.com/clockworklabs/SpacetimeDBPrivate/issues/2168) where template projects do not use the correct lifecycle reducer setup

Adds a new export, and fixes casing in code genreation.

2

# Testing

- I have tested that the `SubscriptionHandle` is correctly exported
- I have tested that the constraint name is now output in `camelCase`
- I have tested that `onMyReducer` callbacks now return a single argument
- I have tested that the table name (and view name) is now used instead of the type name for code generation
- I have tested that the new lifecycle reducers correctly compile


crates/bindings-typescript/src/lib/type_util.ts
```diff
@@ -61,6 +61,16 @@ type CamelCaseImpl<S extends string> = S extends `${infer Head}_${infer Tail}`
...
 
+/** Type safe conversion from "some_identifier-name" to "some_identifier_name"
+ * - No spaces; allowed separators: "_" and "-"
+ * - Normalizes the *first* character to lowercase (e.g. "User_Name" -> "user_name")
+ */
+export type SnakeCase<S extends string> = S extends `${infer Head}${infer Tail}`
+  ? Tail extends Uncapitalize<Tail>
+    ? `${Lowercase<Head>}${SnakeCase<Tail>}`
+    : `${Lowercase<Head>}_${SnakeCase<Tail>}`
+  : Lowercase<S>;
```
cloutiertyler, line 72: This ended up not being necessary to add, but I would like to add the functionality for the future.

crates/bindings-typescript/src/lib/type_builders.ts
```diff
@@ -1748,7 +1767,6 @@ export type ColumnMetadata<Type = any> = {
   isPrimaryKey?: true;
   isUnique?: true;
   isAutoIncrement?: true;
-  isScheduleAt?: true;
```
cloutiertyler, line 1751: I would especially like you to be aware of this change. This is no longer column metadata since you can now use a `ScheduleAt` in any type.

crates/bindings-typescript/src/lib/table.ts
```diff
@@ -288,8 +289,17 @@ export function table<Row extends RowObj, const Opts extends TableOpts<Row>>(
...
+    if (scheduled) {
+      const algebraicType = builder.typeBuilder.algebraicType;
+      if (
+        algebraicType.tag === 'Sum' &&
+        algebraicType.value.variants.length === 2 &&
+        algebraicType.value.variants[0].name === 'Interval' &&
+        algebraicType.value.variants[1].name === 'Time'
+      ) {
+        scheduleAtCol = colIds.get(name)!;
+      }
```
cloutiertyler, line 298: This is no longer metadata so we figure it out by looking for the shape to be the same as `ScheduleAt`. I don't love it, but it is correct.

egormanga:
Won't this be getting into [1.10.0](3769)?

jdetter:
> Won't this be getting into [1.10.0](3769)?

It's going in, if it's tagged with `release-*` then it's going into that release

PR #3743, opened 2025-11-24
[TS] Anonymous transactions

coolreader18:
# Description of Changes

Mirrors the Rust API.

2

# Testing

- [x] Automated procedure testing now enabled for typescript (from `sdks/rust/tests`)

### DIFF
--- a/crates/core/src/host/instance_env.rs
+++ b/crates/core/src/host/instance_env.rs
@@ -601,7 +601,11 @@ impl InstanceEnv {
         written
     }
 
-    pub async fn start_mutable_tx(&mut self) -> Result<(), NodesError> {
+    // Async procedure syscalls return a `Result<impl Future>`, so that we can check `get_tx()`
+    // *before* requiring an async runtime. Otherwise, the v8 module host would have to call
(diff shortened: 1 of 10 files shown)


crates/core/src/host/instance_env.rs
```diff
@@ -601,7 +601,7 @@ impl InstanceEnv {
         written
     }
 
-    pub async fn start_mutable_tx(&mut self) -> Result<(), NodesError> {
+    pub fn start_mutable_tx(&mut self) -> Result<impl Future<Output = ()> + use<'_>, NodesError> {
```
Centril, line 608: This form where the future is inside could use a comment in the code.

crates/core/src/host/instance_env.rs
```diff
@@ -633,7 +635,7 @@ impl InstanceEnv {
         }
     }
 
-    pub async fn commit_mutable_tx(&mut self) -> Result<(), NodesError> {
+    pub fn commit_mutable_tx(&mut self) -> Result<impl Future<Output = ()> + use<'_>, NodesError> {
```
Centril, line 646: Could also use a comment in the code.

PR #3752, opened 2025-11-24
build.rs - Use git as a library rather than binary

bfops:
# Description of Changes

# Testing

- [ ] 
- [ ]


bfops:
windows is failing due to link issues with git2.. but we have git2 in our CLI. interesting..

PR #3756, opened 2025-11-25
Allowed specifying hex strings in the Postgres escape format

egormanga:
# Description of Changes

For consistency with the PG Wire output format for `BYTEA`.

Before,
```sql
SELECT * FROM user WHERE id = '\xc200a1589a788b44c6269819cba4b3493db01ade34652acf33476662fdfc0cfb';
```
gave:
```error
The literal expression `\xc200a1589a788b44c6269819cba4b3493db01ade34652acf33476662fdfc0cfb` cannot be parsed as type `(__identity__: U256)`
```

None.

### 1:
Trivial change.

# Testing

- [x] Manual testing

### DIFF
--- a/crates/lib/src/lib.rs
+++ b/crates/lib/src/lib.rs
@@ -305,7 +305,7 @@ pub fn from_hex_pad<R: hex::FromHex<Error = hex::FromHexError>, T: AsRef<[u8]>>(
     hex: T,
 ) -> Result<R, hex::FromHexError> {
     let hex = hex.as_ref();
-    let hex = if hex.starts_with(b"0x") {
+    let hex = if hex.starts_with(b"0x") || hex.starts_with(b"\\x") {
         &hex[2..]
     } else if hex.starts_with(b"X'") {
         &hex[2..hex.len()]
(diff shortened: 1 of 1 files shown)


crates/lib/src/lib.rs
```diff
@@ -305,7 +305,7 @@ pub fn from_hex_pad<R: hex::FromHex<Error = hex::FromHexError>, T: AsRef<[u8]>>(
     hex: T,
 ) -> Result<R, hex::FromHexError> {
     let hex = hex.as_ref();
-    let hex = if hex.starts_with(b"0x") {
+    let hex = if hex.starts_with(b"0x") || hex.starts_with(b"\\x") {
```
joshua-spacetime: This check needs to be in its own branch to account for enclosing delimiters.

egormanga: What are those? The escaped literal goes: `\xAABBCCDD`, with no tail.

crates/lib/src/lib.rs
```diff
@@ -305,16 +305,34 @@ pub fn from_hex_pad<R: hex::FromHex<Error = hex::FromHexError>, T: AsRef<[u8]>>(
     hex: T,
 ) -> Result<R, hex::FromHexError> {
     let hex = hex.as_ref();
-    let hex = if hex.starts_with(b"0x") || hex.starts_with(b"\\x") {
+    let hex = if hex.starts_with(b"0x") {
         &hex[2..]
-    } else if hex.starts_with(b"X'") {
-        &hex[2..hex.len()]
+    } else if hex.starts_with(b"\\x") {
+        &hex[2..]
```
egormanga, line 311: I am sorry, I just don't see how this got any different. It's just an extra branch with the same code, isn't it?

PR #3765, opened 2025-11-25
[TS] Client-side procedure bindings

coolreader18:
# Description of Changes

2

# Testing

- [x] The types for generated module bindings look right
- [ ] Actually test the client behavior (how?)


crates/bindings-typescript/src/sdk/procedures.ts
```diff
@@ -0,0 +1,27 @@
+import type { Infer, InferTypeOfRow } from '../lib/type_builders';
+import type { CamelCase } from '../lib/type_util';
+import type { UntypedRemoteModule } from './spacetime_module';
+
+// Utility: detect 'any'
+type IfAny<T, Y, N> = 0 extends 1 & T ? Y : N;
+
+// Loose shape that allows all three families even when key names are unknown
+type ProceduresViewLoose = {
```
cloutiertyler, line 9: I believe we've been calling these `Untyped`? no?

cloutiertyler: Jk, we have the reducer form of this.

crates/codegen/src/typescript.rs
```diff
@@ -41,7 +41,7 @@ impl Lang for TypeScript {
             let out = &mut output;
 
             print_file_header(out, false, true);
-            gen_and_print_imports(module, out, &product.elements, &[typ.ty], None);
+            gen_and_print_imports(module, out, product.subtypes(), &[typ.ty], None);
```
cloutiertyler: What is subtypes?

crates/schema/src/type_for_generate.rs
```diff
@@ -227,6 +227,12 @@ impl<'a> IntoIterator for &'a ProductTypeDef {
     }
 }
 
+impl ProductTypeDef {
+    pub fn subtypes(&self) -> impl Iterator<Item = &AlgebraicTypeUse> {
```
cloutiertyler: Kind of an odd name for the element types? Maybe just call it `element_types`?

cloutiertyler (commented):
Very straightforward, I would love to just rename subtypes and then I'm good to approve.

crates/bindings-typescript/src/sdk/procedures.ts
coolreader18: I copied this from `reducers.ts`

PR #3768, opened 2025-11-26
Disable scheduling procedures

gefjon:
# Description of Changes

We've been unable to get scheduled procedures into the initial release of procedures due to a nondeterministic hang in some of our tests which we don't have time to debug. As such, this commit disables scheduled procedures by:

- Changing the `RawModuleDevV9` validator to reject scheduled procedures.
- Changing the Rust bindings library's typechecking magic so that scheduled table annotations referencing procedures will cause a type error.
- Removes some docs from the Rust bindings library which reference scheduled procedures.

N/a

1

# Testing

Automated tests should be sufficient.

Prior to merging #3774 , CI for this branch merged with master failed in the intended way: modules which attempted to define scheduled procedures failed to typecheck.


jdetter:
~~We ended up merging this: https://github.com/clockworklabs/SpacetimeDB/pull/3704~~

Update: we ended up reverting 3704

jdetter (approved):
Seems fine, we ended up reverting 3704. I have not tested this myself but the change looks fine 👍

gefjon:
I have confirmed with @coolreader18 that we don't need any changes related to TypeScript modules; the ModuleDef validator is sufficient.

PR #3772, opened 2025-11-26
Set timeout for smoketests

jdetter:
# Description of Changes

Sets a 2 hour timeout for the smoketests. If the smoketests are taking more than 2 hours they likely will never finish and we should just time them out.

None

1

# Testing

- [x] https://github.com/clockworklabs/SpacetimeDB/actions/runs/19697510980/job/56425630374?pr=3772


bfops (approved):
based on the canceled smoketests.. I think this PR works!

PR #3774, opened 2025-11-26
Revert "Procedures: fix scheduling (#3704)"

jdetter:
This reverts commit b2e37e8008b07673f8de7b617f6c4408ce727a66.

# Description of Changes

Reverts #3704 which I'm pretty sure contains some sort of bug which is causing the smoketests to hang.

None

1

# Testing

- [x] CI passing again


gefjon (approved):

crates/bindings-typescript/src/lib/procedures.ts
```diff
@@ -49,9 +49,8 @@ export function procedure<
   const paramsType: ProductType = {
     elements: Object.entries(params).map(([n, c]) => ({
       name: n,
-      algebraicType: registerTypesRecursively(
-        'typeBuilder' in c ? c.typeBuilder : c
-      ).algebraicType,
+      algebraicType:
+        'typeBuilder' in c ? c.typeBuilder.algebraicType : c.algebraicType,
```
Centril, line 53: This part I think is a fix we do need, so let's make it in an additional PR.

Centril (approved):

crates/bindings-typescript/src/lib/procedures.ts
jdetter: We sync'd on this in discord so I'm resolving this for now. Let me know if you need anything 👍

PR #3775, opened 2025-11-26
Fix some typescript issues

coolreader18:
# Description of Changes

Fixes https://github.com/clockworklabs/SpacetimeDBPrivate/issues/2169 and https://github.com/clockworklabs/SpacetimeDBPrivate/issues/2170.

1

# Testing

- [x] The issues no longer occur.

### DIFF
--- a/crates/bindings-typescript/src/server/runtime.ts
+++ b/crates/bindings-typescript/src/server/runtime.ts
@@ -312,7 +312,12 @@ function makeTableView(
   const sequences = table.sequences.map(seq => {
     const col = rowType.value.elements[seq.column];
     const colType = col.algebraicType;
-    let sequenceTrigger: any;
+
+    // Determine the sentinel value which users will pass to as a placeholder
+    // to cause the sequence to advance.
+    // For small integer SATS types which fit in V8 `number`s, this is `0: number`,
+    // and for larger integer SATS types it's `0n: BigInt`.
+    let sequenceTrigger: bigint | number;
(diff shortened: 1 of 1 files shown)


crates/bindings-typescript/src/server/runtime.ts
```diff
@@ -312,8 +312,30 @@ function makeTableView(
   const sequences = table.sequences.map(seq => {
     const col = rowType.value.elements[seq.column];
     const colType = col.algebraicType;
+    let sequenceTrigger: any;
```
gefjon: This should be `number | BigInt`, right? Possibly also with `undefined` since it starts uninit.

gefjon:
```suggestion
    // Determine the sentinel value which users will pass to as a placeholder
    // to cause the sequence to advance.
    // For small integer SATS types which fit in V8 `number`s, this is `0: number`,
    // and for larger integer SATS types it's `0n: BigInt`.
    let sequenceTrigger: any;
```

gefjon: I'm not sure if `bigint` or `BigInt` is the correct capitalization.

cloutiertyler: BigInt is the constructor. bigint is the type

cloutiertyler: I've addressed this issue and added your comment.

PR #3779, opened 2025-11-26
Split Unity and C# tests into separate jobs

jdetter:
# Description of Changes

This has 2 benefits:
1. If the Unity test fails because of a license issue then we don't have to re-run the C# tests again as part of this flow. Re-running the Unity tests will be much faster if that's the only thing the job is doing.
2. These tests will run faster because they will now run in parallel as separate CI jobs.

None

1

# Testing

- [x] Both tests pass


bfops:
This idea LGTM. If we're refactoring them anyway, I would probably make them both jobs under the existing CI workflow.

bfops (approved):
Assuming all this code is just moved, this LGTM

PR #3790, opened 2025-11-27
Implement SpacetimeType for Result<T, E>

### LINKED ISSUE #3673: Implement `SpacetimeType` for `Result<T, E>`
Make it a new "special" type, like Option. Use `ok` and `err` (lower-case) as variant names for consistency with SATS `Option`. This will also require adding codegen support to all our client SDKs.

mamcx:
# Description of Changes

Closes #3673 

*NOTE*: C++ part will be in another PR

2

Adding a new type touch everywhere

# Testing

- [x] Adding smoke and unit test


JasonAtClockwork:
I'll build out an attached PR to add the new functionality for the Unreal SDK.

crates/bindings-csharp/BSATN.Codegen/Type.cs
```diff
@@ -101,6 +107,14 @@ public static TypeUse Parse(ISymbol member, ITypeSymbol typeSymbol, DiagReporter
         };
     }
 
+    /// <summary>
+    /// Get the name of the BSATN struct for this type.
+    /// </summary>
+    public virtual string to_bsatn_string()
```
JasonAtClockwork: I'm guessing this was added as the Result has two types and that was giving trouble with the BSATNName property directly?

Two notes on this, this should be ToBSATNString() or something like it to match PascalCase for the C# convention followed in all the other code. Second BSATNName was heavily used in CodeGen/Module.cs and doesn't seem to be updated in all locations. This will need review to make sure it won't cause problems for Result<> types used throughout the codegen.

JasonAtClockwork (changes requested):
There is updates to the /modules/sdk-test-cs but this won't be enough testing for the C# side. We'll need to add tests into the /sdks/csharp/examples~/regression-tests like for UUID to make sure this works end-to-end.

@rekhoff - I'll also want your eyes on this one when you can for the C# side

gefjon (approved):
I'm happy with the Rust parts of this PR. We'll still need review from the other languages' maintainers.

crates/bindings-csharp/BSATN.Codegen/Type.cs
```diff
@@ -133,6 +147,40 @@ public abstract string EqualsStatement(
     public abstract string GetHashCodeStatement(string inVar, string outVar, int level = 0);
 }
 
+/// <summary>
+/// A use of a Result<T, E> type.
```
rekhoff: This will confuse the XML parser. We should use the `&lt;` and `&gt;` versions of the characters in summary blocks.
```suggestion
/// A use of a Result&lt;T, E&gt; type.
```

rekhoff (approved):
The C# side looks good. I did some local testing of end-to-end behavior, and things look good.

crates/sats/src/algebraic_value.rs
```diff
@@ -188,6 +188,34 @@ impl AlgebraicValue {
         Self::sum(1, Self::unit())
     }
 
+    /// Converts `self` into an `Result<AlgebraicValue, AlgebraicValue>`, if applicable.
+    pub fn into_result(self) -> Result<Self, Self> {
```
coolreader18: Shouldn't this be `-> Result<Result<AlgebraicType, AlgebraicType>, Self>`?

coolreader18 (approved):
Seems reasonable to me, except for my one comment.

crates/sats/src/algebraic_value.rs
mamcx: Mm, this is like the option variant:

```rust
/// Converts `self` into an `Option<AlgebraicValue>`, if applicable.
    pub fn into_option(self) -> Result<Option<Self>, Self> {
        match self {
            AlgebraicValue::Sum(sum_value) => match sum_value.tag {
                0 => Ok(Some(*sum_value.value)),
                1 => Ok(None),
                _ => Err(AlgebraicValue::Sum(sum_value)),
            },
            _ => Err(self),
        }
    }
```

crates/bindings-csharp/BSATN.Codegen/Type.cs
mamcx: Yes, for things like:

```csharp
SpacetimeDB.Result<SpacetimeDB.Sdk.Test.Module.EveryPrimitiveStruct, string>.BSATN<SpacetimeDB.Sdk.Test.Module.EveryPrimitiveStruct, string, SpacetimeDB.Sdk.Test.Module.EveryPrimitiveStruct.BSATN, SpacetimeDB.BSATN.String> rRW = new();
```

But is not possible to just pass the BSATNName everywhere, I need to manually pick some places.

crates/sats/src/algebraic_value.rs
gefjon: I think the equivalent would be:
```
pub fn into_result(self) -> Result<Result<Self, Self>, Self> {
  match self {
    AlgebraicValue::Sum(sum_value) => match sum_value.tag {
      0 => Ok(Ok(*sum_value.value)),
      1 => Ok(Err(*sum_value.value)),
      _ => Err(self),
    },
    _ => Err(self),
  }
}
```
This way, it's possible to distinguish between the case where `self` is a `Result::Err` (so you get `Ok(Err(val))`, and the case where `self` is not a `Result` at all (so you get `Err(self)`). Your version sends both `Result::Err` and non-result values to `Err`.

sdks/csharp/examples~/regression-tests/client/module_bindings/Procedures/InsertWithTxRollbackResult.g.cs
```diff
@@ -0,0 +1,64 @@
...
+            [DataMember(Name = "Value")]
+            public SpacetimeDB.Result<SpacetimeDB.Types.ReturnStruct, string> Value;
+
+            public InsertWithTxRollbackResult(SpacetimeDB.Result<SpacetimeDB.Types.ReturnStruct, string> Value)
+            {
+                this.Value = Value;
+            }
+
+            public InsertWithTxRollbackResult()
+            {
```
JasonAtClockwork, line 53: We'll either want to have `required` keyword added to the generated use of SpacetimeDB.Result<> or add a default in the empty constructor like:
```cs
this.Value = default!
```

@rekhoff - Do you have any preference, without this devs will have ugly warnings using Result<> in C#?

sdks/csharp/examples~/regression-tests/client/Program.cs
```diff
@@ -134,6 +135,28 @@ void OnSubscriptionApplied(SubscriptionEventContext context)
     waiting++;
     context.Reducers.ThrowError("this is an error");
 
+    Log.Debug("Calling InsertResult");
+    waiting++;
+    context.Reducers.InsertResult(Result<MyTable, string>.Ok(new MyTable(new ReturnStruct(42, "magic"))));
+    waiting++;
+    context.Reducers.InsertResult(Result<MyTable, string>.Err("Fail"));
```
JasonAtClockwork, line 203: This is causing the timeout failure in CI currently as the `waiting++;` lines don't have corresponding `waiting--;'. If you look at how the other reducers are used you should be good to add a simple check for OnInsertResult() to reduce waiting and fix the timeout issue.

JasonAtClockwork (approved):
I've added a couple comments but otherwise this looks like a solid fix for tests on the C# side!

sdks/csharp/examples~/regression-tests/client/module_bindings/Procedures/InsertWithTxRollbackResult.g.cs
JasonAtClockwork: This is the warning devs will see:
```warning CS8618: Non-nullable field 'Msg' must contain a non-null value when exiting constructor. Consider adding the 'required' modifier or declaring the field as nullable.```

sdks/csharp/examples~/regression-tests/client/Program.cs
rekhoff: Just pushed a change that Adds a reducer OnInsert callback to `waiting--` the waiting incrementor. This resolved the issue locally for me.

JasonAtClockwork: @rekhoff got a fix in for this I'll resolve the convo!

sdks/csharp/examples~/regression-tests/client/module_bindings/Procedures/InsertWithTxRollbackResult.g.cs
rekhoff: No strong preference here, but I think I lean towards the `default` empty constructor. If a developer doesn't actually care about the results, marking it as `required` could be kind of annoying to them.

JasonAtClockwork: Added default! as discussed

PR #3792, opened 2025-11-28
C# Views - Use Name from ViewAttribute instead of Method Name

chutch1122:
C# Views - Use Name from ViewAttribute instead of Method Name

# Description of Changes

The [documentation for C# views](https://spacetimedb.com/docs/modules/c-sharp#views) says that "Views must be declared as Public, **with an explicit Name**, and [...]". However, the `Name` provided to the `View` attribute is not being used as the name of the view in the Module or the generated C# client SDK code. The `ViewDeclaration` actually checks that the `View` attribute's name is not null or empty, but then proceeds to do nothing with it.

This PR updates the `ViewDeclaration` to use the `Name` property from `ViewAttribute`.

For more info - see my bug report in Discord: https://discord.com/channels/1037340874172014652/1443881580602069043

No breaking change to the API. Though, anyone who has a view name declared that's different from their method name will have to deal with that during migration of their modules.

1 - Trivial change

# Testing

I compiled the C# projects under `crates/bindings-csharp`, built the NuGet packages, and tested them locally with a project using SpacetimeDB 1.10 (CLI and associated packages). 

I confirmed that the generated classes now use the value from the `View` attribute as the `RemoteTableName`. See attached image.

Screenshot 2025-11-28 at 3 04 48 PM

### DIFF
--- a/sdks/rust/tests/view-client/src/main.rs
+++ b/sdks/rust/tests/view-client/src/main.rs
@@ -191,9 +191,10 @@ fn exec_non_table_return() {
 
 fn exec_subscription_update() {
     let test_counter = TestCounter::new();
-    let mut near_0_2 = Some(test_counter.add_test("near_0_2"));
-    let mut near_0_4 = Some(test_counter.add_test("near_0_4"));
-    let mut far = Some(test_counter.add_test("far_0"));
+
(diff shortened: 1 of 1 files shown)


bfops:
Thank you for creating this! We'll work on getting it reviewed.

PR #3795, opened 2025-11-30
[TS] fix `connectionBuilder` wrong codegen type.

kistz:
# Description of Changes

When trying to use the react hook and you supply a DbConnectionBuilder to `connectionBuilder={builder}` from your module bindings typescript throws a type error since the codegeneration type for the generic field is wrong.

With `bun dev` this error lets you still build the application and view it but with a production build it hard errors.

Fix it by changing the codegen to use the right interface instead of a hardcoded `DbConnection` type that isn't even present in the file and should come from the module_bindings.

None

1

# Testing

- [x] Tried production build with the fix and it succeeded since no more type errors are present.

### DIFF
--- a/crates/codegen/src/typescript.rs
+++ b/crates/codegen/src/typescript.rs
@@ -408,7 +408,7 @@ impl Lang for TypeScript {
         );
         writeln!(
             out,
-            "export class DbConnectionBuilder extends __DbConnectionBuilder<DbConnection> {{}}"
+            "export class DbConnectionBuilder extends __DbConnectionBuilder<__DbConnectionImpl<typeof REMOTE_MODULE>> {{}}"
         );
 
         writeln!(out);


cloutiertyler:
This changes doesn't make sense to me.

> instead of a hardcoded DbConnection type that isn't even present in the file and should come from the module_bindings.

`DbConnection` is defined right below `DbConnectionBuilder` in the module bindings.

What type error were you getting?

kistz:
If you open the `index.ts` in your `module bindings` in a typescript project and scroll to the offending line you should get a type error without this :)
image

Why doesn't it make sense to you?
In my head the code should express: `We take a builder which can be turned in a database connection with the REMOTE_MODULE.`
And not: `We take a builder that already is a finished DbConnection` 

Are my assumptions wrong? 😇 

(actually reading it again im also not sure why it would make a difference because the DbConnection is a impl of the exact same thing at the end of the day 🤷 but typescript typechecker is not happy otherwise)

To provide some more value you should also be able to reproduce this with the next repo i provided in the other issue  
image
 thats the builder side of things which is fixed with my commit im not sure why xdd

cloutiertyler:
I am fairly certain this is fixed by my own PR. Just want to find the PR before I close this.

cloutiertyler:
Closing as this is now resolved.

PR #3796, opened 2025-11-30
[TS] Fix `development` exports breaking NextJS

kistz:
# Description of Changes

NextJS 15 and 16 do not build without these development exports removed tho they funnily enough have different errors.
image

Upon removing the development export everything works like a charm.
I couldn't find any benefit of having them but maybe @cloutiertyler knows more why those are there in the first place🤷 

This can be reproduced if you clone this repo i set up: https://github.com/kistz/spacetime-next-react-hooks
Since its a build error it shouldnt matter if you have the actual module running somewhere.

Step 1: `bun install`

Step 2: `bun dev`

Step 3: observe that the error above is appearing (not good)

Step 3: Go into `node_modules/spacetimedb/package.json` 

Step 4: Remove the `development` fields under exports like in this pr

Step 5: `bun dev`

Step 6: Profit. (should only be a connection error (aka nothing renders) at this point since the module your trying to connect to is missing)

None

1

# Testing

Not a whole lot since i dont fully grasp the impact of this change but without it i cant even use the `spacetimedb` package with NextJS and removing them fixes it.
But even development scenarios are working just fine without it.


cloutiertyler (approved):
I am okay with this change.

For context, this was included so that other packages within our repo that depend on the `spacetimedb` package could depend on the TypeScript source files directly, rather than depending on a built version of the library. It makes dev slightly easier. If it's causing issues for downstream users though, that's obviously not good though. We'll look for a workaround internally.

PR #3797, opened 2025-12-01
[Rust] Module-side query builder types

Shubham8287:
# Description of Changes

Implements a query builder for rust modules according to the [spec](https://github.com/clockworklabs/SpacetimeDBPrivate/blob/master/proposals/0030-views.md#5-query-builder-for-incremental-evaluation).

Look at the tests in `query_builder.rs`.

TODO: Extend return type for views to include `Query` and update the `#[table]` macro to generate query builder bindings.

This is an additive change. Current modules do not need to be updated unless they wish to use the query builder.

1.5, Will appreciate comments to better structure code.

# Testing 

[x]: unit tests.

I have also done type safety checks manually from module.

### DIFF
--- a/crates/bindings/tests/ui/tables.stderr
+++ b/crates/bindings/tests/ui/tables.stderr
@@ -126,14 +126,14 @@ error[E0277]: `&'a Alpha` cannot appear as an argument to an index filtering ope
    = help: the trait `for<'a> FilterableValue` is not implemented for `&'a Alpha`
    = note: The allowed set of types are limited to integers, bool, strings, `Identity`, `ConnectionId`, `Hash` and no-payload enums which derive `SpacetimeType`,
    = help: the following other types implement trait `FilterableValue`:
+             &ConnectionId
+             &Identity
              &Lifecycle
              &TableAccess
              &TableType
              &Version
              &bool
              &ethnum::int::I256
-             &ethnum::uint::U256
-             &i128
            and $N others
 note: required by a bound in `UniqueColumn::<Tbl, <Col as spacetimedb::table::Column>::ColType, Col>::find`
   --> src/table.rs
@@ -153,14 +153,14 @@ error[E0277]: the trait bound `Alpha: IndexScanRangeBounds<(Alpha,), SingleBound
    |                                 required by a bound introduced by this call
    |
    = help: the following other types implement trait `FilterableValue`:
+             &ConnectionId
+             &Identity
              &Lifecycle
              &TableAccess
              &TableType
              &Version
              &bool
              &ethnum::int::I256
-             &ethnum::uint::U256
-             &i128
            and $N others
    = note: required for `Alpha` to implement `IndexScanRangeBounds<(Alpha,), SingleBound>`
 note: required by a bound in `RangedIndex::<Tbl, IndexType, Idx>::filter`


crates/bindings/src/query_builder.rs
```diff
@@ -0,0 +1,257 @@
...
+        let sql = user
+            .r#where(|u| u.id.eq(1))
+            .right_semijoin(o, |u, o| u.id.eq(o.uid))
+            .r#where(|u| u.id.gt(10))
+            .build()
+            .sql;
+        assert_eq!(sql, expected);
+    }
+    #[test]
+    fn test_literals() {
```
Shubham8287, line 210: test for literals.

crates/bindings-macro/src/table.rs
```diff
@@ -925,6 +928,89 @@ pub(crate) fn table_impl(mut args: TableArgs, item: &syn::DeriveInput) -> syn::R
...
+        quote! {
+            #ident: spacetimedb::query_builder::Col::new(stringify!(#ident)),
+        }
+    });
+
+    let ix_cols_struct_fields = indices.iter().filter_map(|index| {
+        let ident = index.accessor_name.clone();
+        let ty = match &index.kind {
+            ValidatedIndexType::BTree { cols } => {
+                //TODO: Verify multi-column index does not support joins
```
joshua-spacetime: Correct. Only single column indexes currently.

crates/bindings/src/query_builder/expr.rs
```diff
@@ -0,0 +1,139 @@
...
+use super::TableName;
+
+pub enum ValueExpr<T> {
+    Column(ColumnRef<T>),
+    Literal(LiteralValue),
+}
+
+pub enum Expr<T> {
+    Eq(T, T),
+    Neq(T, T),
```
joshua-spacetime:
```suggestion
    Ne(T, T),
```

crates/bindings/src/query_builder/expr.rs
```diff
@@ -0,0 +1,139 @@
...
+    match v {
+        ValueExpr::Column(col) => format!("\"{}\".\"{}\"", T::TABLE_NAME, col.column_name),
+        ValueExpr::Literal(lit) => format_literal(lit),
+    }
+}
+
+pub fn format_expr<T: TableName>(expr: &Expr<ValueExpr<T>>) -> String {
+    match expr {
+        Expr::Eq(l, r) => format!("({} = {})", format_value_expr(l), format_value_expr(r)),
+        Expr::Neq(l, r) => format!("({} <> {})", format_value_expr(l), format_value_expr(r)),
```
joshua-spacetime:
```suggestion
        Expr::Ne(l, r) => format!("({} <> {})", format_value_expr(l), format_value_expr(r)),
```

crates/bindings/src/query_builder/expr.rs
```diff
@@ -0,0 +1,139 @@
...
+#[derive(Clone, Debug)]
+pub enum LiteralValue {
+    String(String),
+    Int(i64),
+    Float(f64),
+    BigInt(String),
+    Bool(bool),
+    Identity(Identity),
+    ConnectionId(ConnectionId),
+}
```
joshua-spacetime: What about `Bytes` and `Timestamp`?

crates/bindings/src/query_builder/expr.rs
```diff
@@ -0,0 +1,139 @@
...
+        Expr::Or(a, b) => format!("({} OR {})", format_expr(a), format_expr(b)),
+    }
+}
+
+#[derive(Clone, Debug)]
+pub enum LiteralValue {
+    String(String),
+    Int(i64),
+    Float(f64),
+    BigInt(String),
```
joshua-spacetime: Why not primitive `u128` and `u256` types?

crates/bindings/src/query_builder/table.rs
```diff
@@ -0,0 +1,138 @@
...
+}
+
+pub trait HasCols: TableName {
+    type Cols;
+    fn cols() -> Self::Cols;
+}
+
+pub trait HasIxCols: TableName {
+    type IxCols;
+    fn idx_cols() -> Self::IxCols;
```
joshua-spacetime:
```suggestion
    fn ix_cols() -> Self::IxCols;
```

crates/bindings/src/query_builder/table.rs
```diff
@@ -0,0 +1,138 @@
...
+        *self
+    }
+}
+
+impl<T: TableName, V> Col<T, V> {
+    pub fn eq<R: RHS<T, V>>(self, rhs: R) -> Expr<ValueExpr<T>> {
+        Expr::Eq(self.into(), rhs.to_expr())
+    }
+    pub fn neq<R: RHS<T, V>>(self, rhs: R) -> Expr<ValueExpr<T>> {
+        Expr::Neq(self.into(), rhs.to_expr())
```
joshua-spacetime:
```suggestion
    pub fn ne<R: RHS<T, V>>(self, rhs: R) -> Expr<ValueExpr<T>> {
        Expr::Ne(self.into(), rhs.to_expr())
```

crates/bindings/src/query_builder/table.rs
```diff
@@ -0,0 +1,138 @@
...
+    }
+}
+
+impl<T: TableName, V> From<Col<T, V>> for ValueExpr<T> {
+    fn from(col: Col<T, V>) -> Self {
+        ValueExpr::Column(ColumnRef::new(col.column_name))
+    }
+}
+
+pub struct ColumnRef<T> {
```
joshua-spacetime, line 74: Doesn't this need another type parameter for the column's type?

crates/bindings/src/query_builder.rs
```diff
@@ -0,0 +1,257 @@
...
+pub mod join;
+pub mod table;
+
+pub use expr::*;
+pub use join::*;
+pub use table::*;
+
+pub struct Query {
+    pub(crate) sql: String,
+}
```
joshua-spacetime: This will need a type parameter for the table type.

crates/bindings/src/query_builder/expr.rs
```diff
@@ -0,0 +1,139 @@
...
+    sats::{i256, u256},
+    ConnectionId, Identity,
+};
+
+use crate::query_builder::{Col, ColumnRef};
+
+use super::TableName;
+
+pub enum ValueExpr<T> {
+    Column(ColumnRef<T>),
```
joshua-spacetime, line 11: Doesn't this need to take two type parameters? One for the column's table and the other for the column's type?

crates/bindings/src/query_builder/table.rs
```diff
@@ -0,0 +1,138 @@
...
+impl<T: HasCols> Table<T> {
+    pub fn build(self) -> Query {
+        Query {
+            sql: format!(r#"SELECT * FROM "{}""#, T::TABLE_NAME),
+        }
+    }
+
+    pub fn r#where<F>(self, f: F) -> FromWhere<T>
+    where
+        F: Fn(&T::Cols) -> Expr<ValueExpr<T>>,
```
joshua-spacetime: I was expecting this to be something like:
```rust
F: Fn(&T::Cols) -> BoolExpr<T>,
```

Shubham8287: Do you think, I can just rename `Expr<T>` to be `BoolExpr<T>` ? as it already contain only boolean expression or something else?

Shubham8287: I renamed it and monomorphised over `Operand`, which I think makes more sense now. If it is still not as expexted, lmk please.

crates/bindings/src/query_builder/expr.rs
Shubham8287: Done

Shubham8287:
Changes since last review:
1. Added `Timestamp`, `Bytes` as literals.
2. Renamed `Expr<T>` to `BoolExpr<T>` and `ValueExpr<T>` to `Operand<T>`.
3. fix conflicting macro changes and using pair of struct identifier and table name to implement traits.
4. Some minor naming corrections as per suggestion.

crates/bindings/src/query_builder/table.rs
Shubham8287: This type is not exposed to module and kept this way to be able to stored in `FromWhere<T>`, `JoinWhere<T>`, etc via `Expr<T>`.

this type is derived from `Col<T, V>` or `IxCol<T, V>` which are responsible for doing the type-checks.

crates/bindings/src/query_builder/expr.rs
Shubham8287: replied above.

crates/bindings/src/query_builder/expr.rs
```diff
@@ -0,0 +1,149 @@
...
+    pub fn and(self, other: BoolExpr<T>) -> BoolExpr<T> {
+        BoolExpr::And(Box::new(self), Box::new(other))
+    }
+
+    pub fn or(self, other: BoolExpr<T>) -> BoolExpr<T> {
+        BoolExpr::Or(Box::new(self), Box::new(other))
+    }
+}
+
+pub trait RHS<T, V> {
```
joshua-spacetime, line 38: Can you add a comment to this trait explaining what it's used for?

crates/bindings/src/query_builder/expr.rs
```diff
@@ -0,0 +1,149 @@
...
+            }
+        }
+        LiteralValue::Identity(identity) => format!("0x{}", identity.to_hex()),
+        LiteralValue::ConnectionId(connection) => format!("0x{}", connection.to_hex()),
+        LiteralValue::BigInt(bi) => bi.to_string(),
+        LiteralValue::Bytes(b) => {
+            let hex_string: String = b.iter().map(|byte| format!("{:02x}", byte)).collect();
+            format!("0x{}", hex_string)
+        }
+        LiteralValue::Timestamp(ts) => format!("'{}'", ts),
```
joshua-spacetime: I think this needs to be formatted according to `rfc3339`

crates/bindings/src/query_builder/join.rs
```diff
@@ -0,0 +1,160 @@
...
+        on: impl Fn(&L::IxCols, &R::IxCols) -> IxJoinEq<L, R, V>,
+    ) -> JoinWhere<L> {
+        semijoin(L::ix_cols(), R::ix_cols(), on, None, JoinKind::Left)
+    }
+
+    pub fn right_semijoin<R: HasIxCols, V>(
+        self,
+        _right: Table<R>,
+        on: impl Fn(&L::IxCols, &R::IxCols) -> IxJoinEq<L, R, V>,
+    ) -> JoinWhere<L> {
```
joshua-spacetime: Shouldn't `right_semijoin` return `JoinWhere<R>`? In other words, we need to encode the table type that is being returned so that subsequent operations like `where` will work correctly.

crates/bindings/src/query_builder/join.rs
```diff
@@ -0,0 +1,160 @@
+use std::marker::PhantomData;
+
+use super::{
+    expr::{format_expr, BoolExpr},
+    table::{ColumnRef, HasCols, HasIxCols, Table},
+    Query,
+};
+
+pub struct IxCol<T, V> {
```
joshua-spacetime, line 12: A doc comment would be useful here

crates/bindings/src/query_builder.rs
```diff
@@ -0,0 +1,285 @@
...
+        s.split_whitespace().collect::<Vec<_>>().join(" ")
+    }
+    #[test]
+    fn test_simple_select() {
+        let q = users().build();
+        assert_eq!(q.sql(), r#"SELECT * FROM "users""#);
+    }
+    #[test]
+    fn test_where_literal() {
+        let q = users().r#where(|c| c.id.eq(10)).build();
```
joshua-spacetime, line 106: I think these literals will need to be typed. At least in certain cases. For example:
```rust
let q = users().r#where(|c| c.id.eq(10u8)).build();
```
if `id` is a `u8` column.

Note, the following will fail to parse if `id` is `u8`
```sql
select * from users where id = 1000
```

Shubham8287:
```rust
let q = users().r#where(|c| c.id.eq(1000)).build();
```

This will still give compilation error as `Col<T,V>::eq` accepts only `impl RHS` also typed on `T` and `V`.

Shubham8287: error message:
```
1. literal out of range for `u8`
   the literal `1000` does not fit into the type `u8` whose range is `0..=255`
   `#[deny(overflowing_literals)]` on by default [overflowing_literals]

```

crates/bindings/src/query_builder/expr.rs
Shubham8287: It is already formatting as `rfc3339` as per `impl fmt::Display`.

crates/bindings/src/query_builder.rs
joshua-spacetime: Excellent!

crates/bindings/src/query_builder/expr.rs
```diff
@@ -0,0 +1,153 @@
...
+        BoolExpr::Ne(l, r) => format!("({} <> {})", format_bool_expr(l), format_bool_expr(r)),
+        BoolExpr::Gt(l, r) => format!("({} > {})", format_bool_expr(l), format_bool_expr(r)),
+        BoolExpr::Lt(l, r) => format!("({} < {})", format_bool_expr(l), format_bool_expr(r)),
+        BoolExpr::And(a, b) => format!("({} AND {})", format_expr(a), format_expr(b)),
+        BoolExpr::Or(a, b) => format!("({} OR {})", format_expr(a), format_expr(b)),
+    }
+}
+
+#[derive(Clone, Debug)]
+pub enum LiteralValue {
```
joshua-spacetime: Optional: I would probably have `LiteralValue` just wrap a string and have `impl_rhs` handle the formatting since the types have already been checked by the time we instantiate this.

joshua-spacetime (approved):
End to end tests will be added in a separate patch that updates the `#[view]` return type and generates the query builder bindings.

PR #3800, opened 2025-12-01
[TS] Implement TextEncoder/TextDecoder with native code

coolreader18:
# Description of Changes

This implements (a subset of) the TextEncoder/TextDecoder web APIs using native functions to do the actual `Uint8Array <-> String` conversion. This should be a good bit faster than the `fast-text-encoding` package.

2 - this introduces new kinds of JS code and host calls to the v8 host, but they're pretty well encapsulated.

# Testing

- [x] All TS modules make use of these already for encoding/decoding strings in BSATN - the `fast-text-encoding` polyfill we pull in only takes effect if the classes don't already exist, so the smoketests passing means it works.


crates/core/src/host/v8/builtins/mod.rs
```diff
@@ -0,0 +1,143 @@
...
+}
+
+/// From ./types.d.ts:
+/// ```ts
+/// export function utf8_encode(s: string): Uint8Array<ArrayBuffer>;
+/// ```
+fn utf8_encode<'scope>(scope: &mut PinScope<'scope, '_>, args: FunctionCallbackArguments<'scope>) -> FnRet<'scope> {
+    let v = args.get(0);
+    let v = v.to_string(scope).ok_or_else(exception_already_thrown)?;
+    let s = v.to_rust_string_lossy(scope);
```
Centril: These could have better names

crates/core/src/host/v8/builtins/mod.rs
```diff
@@ -0,0 +1,143 @@
...
+/// the js builtins in this directory.
+fn internal_builtins_module<'scope>(scope: &mut PinScope<'scope, '_>) -> Local<'scope, v8::Module> {
+    create_synthetic_module!(scope, "spacetime:internal_builtins", utf8_encode, utf8_decode)
+}
+
+/// From ./types.d.ts:
+/// ```ts
+/// export function utf8_encode(s: string): Uint8Array<ArrayBuffer>;
+/// ```
+fn utf8_encode<'scope>(scope: &mut PinScope<'scope, '_>, args: FunctionCallbackArguments<'scope>) -> FnRet<'scope> {
```
Centril, line 117: The code could explain what the benefit of these existing are for posterity.

crates/core/src/host/v8/builtins/mod.rs
```diff
@@ -0,0 +1,143 @@
...
+    let v = args.get(0);
+    let fatal = args.get(1).boolean_value(scope);
+    if let Ok(v) = v.try_cast::<v8::ArrayBufferView>() {
+        let buffer = v.get_contents(&mut []);
+        if fatal {
+            std::str::from_utf8(buffer).map_err(|e| TypeError(e.to_string()).throw(scope))?;
+        }
+        v8::String::new_from_utf8(scope, buffer, v8::NewStringType::Normal)
+            .map(Into::into)
+            .ok_or_else(|| TypeError("decoded string too long").throw(scope))
```
Centril: There is a `StringTooLongError`

coolreader18:
Internal tests are only due to this branch not being based on latest master.

PR #3802, opened 2025-12-01
Migrate to Rust 2024

coolreader18:
# Description of Changes

It'd be best to review this commit-by-commit, and using [difftastic](https://difftastic.wilfred.me.uk) to easily tell when changes are minor in terms of syntax but a line based diff doesn't show that.

3 - edition2024 does bring changes to drop order, which could cause issues with locks, but I looked through [all of the warnings that weren't fixed automatically](https://gistcdn.githack.com/coolreader18/80485ae5c5f82de1784229cce2febb26/raw/ba80f3fecda66ceb34f4f7ad73b98ea02d4893a2/warnings.html) and couldn't find any issues.

# Testing

n/a; internal code change

### DIFF
--- a/crates/bindings/tests/ui/tables.stderr
+++ b/crates/bindings/tests/ui/tables.stderr
@@ -196,7 +196,7 @@ help: the trait `FilterableValue` is not implemented for `Alpha`
 note: required by a bound in `RangedIndex::<Tbl, IndexType, Idx>::filter`
   --> src/table.rs
    |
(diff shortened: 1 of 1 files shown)


cloutiertyler:
Need rebasing, also I think we should do it after 2.0

coolreader18:
The commits named `Run cargo ... && cargo fmt` do exactly that and only that, and can be verified as such. Those are just the auto-fixes from rustc/clippy due to the edition switch.

cloutiertyler (approved):

PR #3803, opened 2025-12-01
Fixes docs links

cloutiertyler:
# Description of Changes

This removes all usages of `pathname:///` and replaces them with `https://spacetimedb.com/`. `pathname:///` was resolving to `/docs` in some cases and was not verified by our link checker. Note that this means these links will NOT work in staging.

This is a temporary fix. We should find a way to be able to link out to the rest of the site without hardcoding the hostname in the future.

None

1

# Testing

- I tested that the links work locally (but obviously they link to the production site).

### DIFF
--- a/sdks/rust/tests/view-client/src/main.rs
+++ b/sdks/rust/tests/view-client/src/main.rs
@@ -191,9 +191,10 @@ fn exec_non_table_return() {
 
 fn exec_subscription_update() {
     let test_counter = TestCounter::new();
-    let mut near_0_2 = Some(test_counter.add_test("near_0_2"));
-    let mut near_0_4 = Some(test_counter.add_test("near_0_4"));
-    let mut far = Some(test_counter.add_test("far_0"));
+
+    let mut insert_0 = Some(test_counter.add_test("insert_0"));
+    let mut delete_0 = Some(test_counter.add_test("delete_0"));
+
     connect_with_then(
         &test_counter,
         "0",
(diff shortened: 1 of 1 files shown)


docs/docs/06-server-module-languages/04-typescript-reference.md
```diff
@@ -115,7 +115,7 @@ spacetime publish <MY_DATABASE_NAME>
 
 where `<MY_DATABASE_NAME>` is the name of your existing database.
 
-You can also generate client bindings for your schema with `spacetime generate`. See the [client SDK documentation](pathname:///docs/sdks/typescript#generate-module-bindings) for more information.
+You can also generate client bindings for your schema with `spacetime generate`. See the [client SDK documentation](https://spacetimedb.com/docs/sdks/typescript#generate-module-bindings) for more information.
```
JulienLavocat, line 118: This one looks suspicious, shouldn't it be a "standard" link like on other pages?

cloutiertyler:
```
#### Serverside Libraries

- [Rust](https://spacetimedb.com/docs/modules/rust/quickstart)
- [C#](https://spacetimedb.com/docs/modules/c-sharp/quickstart)

#### Client Libraries

- [Rust](https://spacetimedb.com/docs/sdks/rust/quickstart)
- [C#](https://spacetimedb.com/docs/sdks/c-sharp/quickstart)
- [Typescript](https://spacetimedb.com/docs/sdks/typescript/quickstart)
```

We have several other links of this kind elsewhere. I think we should fix all of these, but I would do it in a separate PR.

PR #3805, opened 2025-12-02
CI - Skip the Unity testsuite on external PRs

bfops:
# Description of Changes

Bubbling up an `if` since we split out the unity testsuite into its own job.

The GitHub docs say (https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/about-status-checks):
> A job that is skipped will report its status as "Success". It will not prevent a pull request from merging, even if it is a required check.

None. CI-only change.

1

# Testing

Unsure how to test this honestly :shrug:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -526,6 +526,9 @@ jobs:
           fi
 
   unity-testsuite:
+    # Skip if this is an external contribution.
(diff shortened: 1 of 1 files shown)


jdetter (approved):
Ok seems fine to me 👍

PR #3809, opened 2025-12-02
C# implementation of Transactions for Procedures

rekhoff:
# Description of Changes

Implements the C# equivalent of #3638

This implement uses inheritance, where abstract base classes (like `ProcedureContextBase` in `ProcedureContext.cs`)  store the core of the implementation, and then generated wrappers (like `ProcedureContext` in the generated FFI.cs file) inherit from them.

For error handling, we work like Rust's implementation of `Result<T,E>` but we require `where E : Exception` because of how exceptions work in C#. Transaction-level failures come back as a `TxOutcome` and user errors should follow the `Result<T,E>` pattern. In this implementation, we have `UnwrapOrThrow()` throws exceptions directly because of C#'s error handling pattern.

Unlike the Rust implementation's direct `Result` propagation, we are using an `AbortGuard` pattern (in `ProcedureContext.cs`) for exception handling, which uses `IDisposable` for automatic cleanup.

Most changes should have fairly similar Rust-equivalents beyond that.
For module authors, the changes here allow for the transation logic to work like:
```csharp
ctx.TryWithTx<ResultType, Exception>(tx => {
    // transaction logic
    return Result<ResultType, Exception>.Ok(result);
});
```
This change includes a number of tests added to the `sdks/csharp/examples~/regression-tests/`'s `server` and `client` to validate the behavior of the changes. `server` changes provide further usage examples for module authors.

Should not be a breaking change

2

# Testing

- [x] Created Regression Tests that show transitions in procedures working in various ways, all of which pass.


crates/bindings-csharp/BSATN.Codegen/Type.cs
```diff
@@ -44,6 +44,11 @@ public abstract record TypeUse(string Name, string BSATNName)
     /// </summary>
     public static string BsatnFieldSuffix => $"{BSATN_FIELD_SUFFIX}";
 
+    /// <summary>
+    /// Indicates whether this type represents a void return.
+    /// </summary>
+    public virtual bool IsVoid => false;
```
gefjon: Distinguishing between void and unit returns feels weird and wrong to me. Why is this necessary?

crates/bindings-csharp/BSATN.Runtime/BSATN/AlgebraicType.cs
```diff
@@ -6,17 +6,11 @@ public interface ITypeRegistrar
...
-    public string? Name;
+    public string? Name = name;
 
-    public AlgebraicType AlgebraicType;
-
-    public AggregateElement(string name, AlgebraicType algebraicType)
-    {
-        Name = name;
-        AlgebraicType = algebraicType;
-    }
```
gefjon, line 19: Why'd this change?

crates/bindings-csharp/BSATN.Runtime/Builtins.cs
```diff
@@ -365,7 +365,7 @@ public readonly TimeDuration TimeDurationSince(Timestamp earlier) =>
     public static Timestamp operator -(Timestamp point, TimeDuration interval) =>
         new Timestamp(checked(point.MicrosecondsSinceUnixEpoch - interval.Microseconds));
 
-    public int CompareTo(Timestamp that)
+    public readonly int CompareTo(Timestamp that)
```
gefjon: Why'd this change? Is it related to this PR?

crates/bindings-csharp/Runtime/Internal/Module.cs
```diff
@@ -411,11 +424,20 @@ public static Errno __call_view_anon__(uint id, BytesSource args, BytesSink rows
...
+public partial class Local
+{
+    // Intentionally empty – generated code adds table handles here.
+}
+
 /// <summary>
 /// Read-only database access for view contexts.
 /// The code generator will extend this partial class to add table accessors.
 /// </summary>
-public sealed partial class LocalReadOnly
```
gefjon, line 418: Why did this change? `LocalReadOnly` is for views, which should be unrelated to this procedure PR, right?

crates/bindings-csharp/Runtime/Internal/Procedure.cs
```diff
@@ -0,0 +1,241 @@
...
+    /// </summary>
+    byte[] Invoke(BinaryReader reader, IProcedureContext ctx);
+}
+
+/// <summary>
+/// Manages procedure execution context and transactions for SpacetimeDB procedures.
+/// This class is responsible for maintaining the current procedure context and
+/// managing transaction state.
+/// </summary>
+public class ProcedureContextManager : IProcedureContextManager
```
gefjon: I don't understand why this context manager is necessary, or why we need `IDisposable`s to manage a stack of scopes. The Rust version you linked to in #3638 doesn't appear to contain anything similar.

crates/bindings-csharp/Runtime/Internal/Procedure.cs
```diff
@@ -0,0 +1,241 @@
...
+
+/// <summary>
+/// Provides utility methods for procedure-related functionality.
+/// </summary>
+public static class ProcedureExtensions
+{
+    /// <summary>
+    /// Schedules an immediate volatile, non-atomic procedure call.
+    /// </summary>
+    public static void VolatileNonatomicScheduleImmediate(string name, MemoryStream args)
```
gefjon, line 57: Again, not obviously included in #3638 .

crates/bindings-csharp/Runtime/Internal/Procedure.cs
```diff
@@ -0,0 +1,241 @@
...
+    void CommitMutTx();
+
+    /// <summary>Aborts the current mutable transaction.</summary>
+    void AbortMutTx();
+
+    /// <summary>Commits a transaction with a retry mechanism.</summary>
+    bool CommitMutTxWithRetry(Func<bool> retryBody);
+
+    /// <summary>Asynchronously commits a transaction with a retry mechanism.</summary>
+    Task<bool> CommitMutTxWithRetryAsync(Func<Task<bool>> retryBody);
```
gefjon: Why do we have `async` stuff happening inside of module code? No other module languages currently have `async` inside of them, as our syscalls are blocking.

crates/codegen/src/csharp.rs
```diff
@@ -355,7 +355,7 @@ const REDUCER_EVENTS: &str = r#"
         /// <summary>
         /// The procedure event that caused this callback to run.
         /// </summary>
-        public readonly ProcedureEvent Event;
+        public ProcedureEvent Event { get; }
```
gefjon: Why do we have changes in this file? I'd expect this PR not to affect client codegen, or anything about the client SDK, at all.

modules/benchmarks-cs/circles.cs
```diff
@@ -48,9 +47,9 @@ public static float MassToRadius(uint mass)
 
     public static bool IsOverlapping(Entity entity1, Entity entity2)
     {
-        float entity1_radius = MassToRadius(entity1.mass);
-        float entity2_radius = MassToRadius(entity2.mass);
-        float distance = (float)
+        var entity1_radius = MassToRadius(entity1.mass);
+        var entity2_radius = MassToRadius(entity2.mass);
+        var distance = (float)
```
gefjon: These changes appear unrelated to this PR.

modules/benchmarks-cs/ia_loop.cs
```diff
@@ -1,6 +1,5 @@
-using SpacetimeDB;
-
 namespace Benchmarks;
+using SpacetimeDB;
```
gefjon, line 1: Changes in this file also appear unrelated to this PR.

modules/benchmarks-cs/synthetic.cs
```diff
@@ -1,6 +1,5 @@
-using SpacetimeDB;
-
 namespace Benchmarks;
+using SpacetimeDB;
```
gefjon, line 1: More apparently unrelated changes.

gefjon (changes requested):
I'd like to see a more in-depth PR description which goes into detail on your design and implementation choices. Especially in big PRs like this, there's a burden on the author to both keep the changes as contained and laser-focused as possible, and also to proactively explain to reviewers what's going on in the diff. Anywhere I've left a comment asking for reasoning, that's the kind of thing that should be in your PR description proactively.

crates/bindings-csharp/BSATN.Codegen/Type.cs
rekhoff: C#'s type system treats `void` and unit (`ValueTuple`) as distinct types. As part of serialization, we need to know whether to expect a return value, and C# requires explicit handling of this.

crates/bindings-csharp/BSATN.Runtime/BSATN/AlgebraicType.cs
rekhoff: C# benefits from property initializers and explicit constructors for better tooling support. Basically this pattern matches .NET's design guidelines and enables better encapsulation.

crates/bindings-csharp/BSATN.Runtime/Builtins.cs
rekhoff: Na, this was just an IDE recommendation that made sense to me when I was making changes. It's not required. It just helps the compiler optimize the method.

crates/bindings-csharp/BSATN.Codegen/Type.cs
gefjon: We can treat `Unit` in return type as equivalent to `void`.

crates/bindings-csharp/Runtime/Internal/Procedure.cs
rekhoff: This was an artifact of some troubleshooting and shouldn't have been included. I failed to clean this up. I've moved the functionality into the `ProcedureContextBase` like it should have been all along.

crates/bindings-csharp/Runtime/Internal/Module.cs
rekhoff: This was changed during some reorganization of the generated FFI file warranted some permissions changes. Mainly this was to keep the `Local` and `LocalReadOnly` defined in similar places in the file, but I reverted it back, so this is no longer an issue.

JasonAtClockwork (approved):
Looks solid now, I've run the regression tests and walked through the changes. One small nit would be to add Log.Debug() for all successful tests but I think we want to rework the test runner soon anyhow.

gefjon (approved):
This looks great, thanks!

PR #3812, opened 2025-12-03
Add a typescript query builder for views

jsdt:
# Description of Changes

This adds a way to build queries with typescript views, and it allows views to return queries (if the return type of the query is an array).

For examples and syntax, you can look at the tests in [crates/bindings-typescript/tests/query.test.ts](https://github.com/clockworklabs/SpacetimeDB/compare/jsdt/ts-query-builder?expand=1#diff-4fd25c191f1207085a491cf84996c601f805f5e8280d1cf2a812ebad6aa6e75a).

To play around with the syntax, you might find it easier to look in [crates/bindings-typescript/src/server/view.test-d.ts](https://github.com/clockworklabs/SpacetimeDB/compare/jsdt/ts-query-builder?expand=1#diff-4fd25c191f1207085a491cf84996c601f805f5e8280d1cf2a812ebad6aa6e75a).

This could still use some cleanup, and there are some places where the type safety is imperfect. I'll try to list the known limitations here:
1. This will allow the use of `eq` for columns that are product types, even though the query engine doesn't support it. This can be fixed later, and it would only be a breaking change for modules that have invalid queries.
2. When we check if a view is returning a query of the correct type, we are checking with the typescript row type. We should be checking with the spacetime type, since this type checking will allow a couple incorrect things to be returned:
    1. A different table with any superset of the fields (for example, a different table that has one extra field). That will fail when executing, but will be allowed by the typescript compiler.
    2. A table with the same fields, but with those fields in a different order would also fail at runtime, but be allowed by the typescript compiler.
    3. A table with fields of a different spacetimetype that map to the same typescript type (like `u16` and `u32`).

This adds some new API surface, but does not break existing code.

2.

# Testing

For automated tests, there are unit tests to see what sql gets emitted in `tests/query.test.ts`, and some tests of the types in `view.test-d.ts`.

I've also run some manual tests with a typescript module with views.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,20 +4,11 @@ version = 4
 
 [[package]]
 name = "addr2line"
-version = "0.22.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6e4503c46a5c0c7844e948c9a4d6acd9f50cccb4de1c48eb9e291ea17470c678"
-dependencies = [
- "gimli 0.29.0",
-]
-
-[[package]]
-name = "addr2line"
 version = "0.25.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1b5d307320b3181d6d7954e663bd7c774a838b8220fe0593c86d9fb09f498b4b"
 dependencies = [
- "gimli 0.32.3",
+ "gimli",
 ]
 
 [[package]]
@@ -401,11 +392,11 @@ version = "0.3.76"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bb531853791a215d7c62a30daf0dde835f381ab5de4589cfe7c649d2cbe92bd6"
 dependencies = [
- "addr2line 0.25.1",
+ "addr2line",
  "cfg-if",
  "libc",
  "miniz_oxide",
- "object 0.37.3",
+ "object",
  "rustc-demangle",
  "windows-link 0.2.1",
 ]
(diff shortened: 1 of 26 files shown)


crates/bindings-typescript/src/server/query.ts
```diff
@@ -0,0 +1,679 @@
...
+
+export const isTypedQuery = (val: unknown): val is TableTypedQuery<any> =>
+  !!val && typeof val === 'object' && QueryBrand in (val as object);
+
+// A query builder with a single table.
+type From<TableDef extends TypedTableDef> = Readonly<{
+  where(
+    predicate: (row: RowExpr<TableDef>) => BooleanExpr<TableDef>
+  ): From<TableDef>;
+  semijoinRight<RightTable extends TypedTableDef>(
```
joshua-spacetime: Can we go with `rightSemijoin` instead?

crates/bindings-typescript/src/server/query.ts
```diff
@@ -0,0 +1,679 @@
...
+    other: TableRef<RightTable>,
+    on: (
+      left: IndexedRowExpr<TableDef>,
+      right: IndexedRowExpr<RightTable>
+    ) => EqExpr<TableDef | RightTable>
+  ): SemijoinBuilder<RightTable>;
+  semijoinLeft<RightTable extends TypedTableDef>(
```
joshua-spacetime: And similarly here, `leftSemijoin`.

crates/bindings-typescript/tests/query.test.ts
```diff
@@ -0,0 +1,257 @@
...
+    const qb = makeQueryBuilder(schemaDef);
+    const sql = from(qb.person).toSql();
+
+    expect(sql).toBe('SELECT * FROM "person"');
+  });
+
+  it('renders a WHERE clause for simple equality filters', () => {
+    const qb = makeQueryBuilder(schemaDef);
+    const sql = from(qb.person)
+      .where(row => row.name.eq(literal("O'Brian")))
```
joshua-spacetime: Do we have to wrap literals?

crates/bindings-typescript/src/server/query.ts
jsdt: We can, though I prefer having related functions share a prefix, so they show up together when autoCompleting.

crates/bindings-typescript/tests/query.test.ts
jsdt: I guess we don't need to require wrapping for the set of literals we allow now. If we allow people to do equality on product types, it could cause problems (because it can be hard to tell whether something is supposed to be interpreted as one of our AST nodes or a literal), but we can add a wrapper later.

crates/bindings-typescript/src/server/query.ts
joshua-spacetime: I'm fine with aliases, I just want to make sure we also have an api that looks similar to drizzle.

joshua-spacetime (approved):
I mainly reviewed the api. Improving compile-time safety will be addressed in follow up patches.

PR #3816, opened 2025-12-03
Fix the issues with scheduling procedures

### LINKED ISSUE #3645: Procedures: enable executing scheduled procedures

coolreader18:
# Description of Changes

This reapplies the patch from #3704, and fixes the issues that were causing it to deadlock.

The reason it was deadlocking was that it allowed for the following sequence of events:
* `SchedulerActor::handle_queued()` begins mutable tx
* `ModuleHost::disconnect_client()` submits call to `call_reducer(tx: None)`
* scheduler submits call to `call_reducer(tx: Some)`
* `WasmModuleInstance::disconnect_client` now has to try to take tx lock, but the scheduler's call_reducer already holds it and is behind it in the queue

So, I moved most of the logic from `handle_queued` back to being executed in the module worker thread, but kept the code in `scheduler.rs` so that it can all be reasoned about locally.

Fixes #3645. Should I uncomment the implementation of `ExportFunctionForScheduledTable for F: Procedure` now?

2 - there's a chance that this patch hasn't fully fixed the deadlock issue from #3704, but I'm quite confident.

# Testing

- [x] Manually verified that deadlock no longer occurs - previously, `while true; do python -m smoketests schedule_reducer -k test_scheduled_table_subscription; done` would freeze up in only 2 or 3 iterations, but now it can run for 10 minutes without issues.


crates/core/src/host/module_host.rs
```diff
@@ -1605,31 +1590,60 @@ impl ModuleHost {
...
-        self.call(
-            SCHEDULED_REDUCER,
-            item,
-            |item, inst| inst.call_scheduled_reducer(item),
-            |item, inst| inst.call_scheduled_reducer(item),
+        params: ScheduledFunctionParams,
+    ) -> Result<CallScheduledFunctionResult, NoSuchModule> {
+        self.with_instance(
+            "scheduled function",
+            "<unknown>",
```
```suggestion
            "reducer or procedure",
```

crates/core/src/host/scheduler.rs
```diff
@@ -331,129 +332,125 @@ impl SchedulerActor {
...
-                return;
-            }
-        };
+    // We've determined whether we have a reducer or procedure.
+    // The logic between them will now split,
+    // as for scheduled procedures, it's incorrect to retry them if execution aborts midway,
+    // so we must remove the schedule row before executing.
+    match params {
+        CallParams::Reducer(params) => {
+            let result = panic::catch_unwind(panic::AssertUnwindSafe(|| {
```
Centril, line 425: The `catch_unwind` here could use some commentary.

PR #3823, opened 2025-12-04
[Rust] Query builder Integration.

Shubham8287:
# Description of Changes
Patch contain few different things, required to run query builder end to end.
Git history seems messed up but it can be reviewed commitwise from - 69c1907b3e613ee242e6f14a7776cccc4acb4c1f.

1. Decouples generic type `T` from table name, as type represents the `struct` and not table, and there couple be multiple tables implemented from same struct.
2. modify table macro implementaiton to generate code supporting `traits` and types required for query builder, 
3. Disable aliasing in sql queries for now, as it was causing semantically wrong query when we were doing joins on `FromWhere` (as original `Expr` were referencing to original table names).
4. smoketests.

Add `Query<T>` type as Views return type.

2, changes are simple but not splitted up well.

# Testing
1. Added smoketest.

TODO:
ui test.

### DIFF
--- a/crates/bindings/src/query_builder.rs
+++ b/crates/bindings/src/query_builder.rs
@@ -4,7 +4,7 @@ pub mod table;
 
 pub use expr::*;
 pub use join::*;
-use spacetimedb_lib::{sats::impl_st, SpacetimeType};
+use spacetimedb_lib::{sats::impl_st, AlgebraicType, SpacetimeType};
 pub use table::*;
 
 pub struct Query<T> {
@@ -25,7 +25,7 @@ impl<T> Query<T> {
     }
 }
 
-impl_st!([T: SpacetimeType] Query<T>, ts => spacetimedb_lib::AlgebraicType::option(T::make_type(ts)));
+impl_st!([T: SpacetimeType] Query<T>, ts => AlgebraicType::option(T::make_type(ts)));
 
 #[cfg(test)]
 mod tests {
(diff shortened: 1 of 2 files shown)


crates/bindings/src/query_builder.rs
```diff
@@ -24,6 +25,8 @@ impl<T> Query<T> {
     }
 }
 
+impl_st!([T: SpacetimeType] Query<T>, ts => spacetimedb_lib::AlgebraicType::option(T::make_type(ts)));
```
Shubham8287: I am not very sure, if this is best way to do it but it works.
And  schema validation code only considers `AlagebraicType::Option` and `AlagebraicType::Array`  as valid.

smoketests/tests/views.py
```diff
@@ -589,3 +589,135 @@ def test_subscribing_with_different_identities(self):
...
+        """Tests that views returning Query types with left semijoin work as expected"""
+
+        self.assertSql("SELECT * FROM users_whos_age_is_known", """\
+ identity | name    | online
+----------+---------+--------
+ 1        | "Alice" | true
+ 2        | "BOB"   | false
+""")
+
+#    def test_query_complex_right_semijoin_view(self):
```
Shubham8287: this smoketest fails with error:

```
2025-12-04T14:47:45.257076Z  INFO crates/client-api/src/lib.rs:87: sql="SELECT * FROM offline_user_20_years_old"
Running view query: SELECT "user".* FROM "person" JOIN "user" ON "person"."identity" = "user"."identity" WHERE ("person"."age" = 20) AND ("user"."online" = TRUE)
2025-12-04T14:47:45.258154Z ERROR crates/core/src/host/wasm_common/module_host_actor.rs:1098: Error materializing view `offline_user_20_years_old`: Error executing raw SQL returned by view

Caused by:
    Subscriptions require indexes on join columns
```

Sql query formed seems correct though.

PR #3829, opened 2025-12-05
Some more UI tests for Views

Shubham8287:
# Description of Changes
As title

NA

0


crates/bindings/tests/ui/views.rs
```diff
@@ -155,4 +155,71 @@ fn scheduled_table_view(_: &ViewContext, _args: ScheduledTable) -> Vec<Player> {
...
+struct PlayerInfo {
+    #[unique]
+    identity: Identity,
+    #[index(btree)]
+    weight: u32,
+    age: u8,
+}
+
+
+/// Where with mismatched types
```
joshua-spacetime:
```suggestion
/// Comparing incompatible types in `where` condition: Identity != int
```

crates/bindings/tests/ui/views.rs
```diff
@@ -155,4 +155,71 @@ fn scheduled_table_view(_: &ViewContext, _args: ScheduledTable) -> Vec<Player> {
...
+#[view(name = view_bad_where, public)]
+fn view_bad_where(ctx: &ViewContext) -> Query<Player> {
+    ctx.from
+        .player()
+        .r#where(|a| a.identity.eq(42))
+        .build()
+}
+
+// Where with mismatched int types 
+// u8 with 4200u32
```
joshua-spacetime:
```suggestion
/// Comparing incompatible types in `where` condition: u8 != u32
```

crates/bindings/tests/ui/views.rs
```diff
@@ -155,4 +155,71 @@ fn scheduled_table_view(_: &ViewContext, _args: ScheduledTable) -> Vec<Player> {
...
+#[view(name = view_bad_join, public)]
+fn view_bad_join(ctx: &ViewContext) -> Query<PlayerInfo> {
+    ctx.from
+        .player_info()
+        .left_semijoin(ctx.from.player(), |a, b| a.weight.eq(b.identity)) 
+        .build()
+}
+
+// Joining non-idx columna
+// -- age is not indexed
```
joshua-spacetime:
```suggestion
/// Joining non-index columns
/// -- age is not indexed
```

crates/bindings/tests/ui/views.rs
```diff
@@ -155,4 +155,71 @@ fn scheduled_table_view(_: &ViewContext, _args: ScheduledTable) -> Vec<Player> {
...
+#[view(name = view_join_non_indexed_column, public)]
+fn view_join_non_indexed_column(ctx: &ViewContext) -> Query<PlayerInfo> {
+    ctx.from
+        .player()
+        .right_semijoin(ctx.from.player_info(), |a, b| a.identity.eq(b.age)) 
+        .build()
+}
+
+// Right join returns right table's type
+// -- should be PlayerInfo, not Player
```
joshua-spacetime:
```suggestion
/// Right join returns right table's type
/// -- should be PlayerInfo, not Player
```

crates/bindings/tests/ui/views.rs
```diff
@@ -155,4 +155,71 @@ fn scheduled_table_view(_: &ViewContext, _args: ScheduledTable) -> Vec<Player> {
...
+        .build()
+}
+
+// Where with mismatched int types 
+// u8 with 4200u32
+#[view(name = view_bad_where_int_types, public)]
+fn view_bad_where_int_types(ctx: &ViewContext) -> Query<PlayerInfo> {
+    ctx.from
+        .player_info()
+        .r#where(|a| a.age.eq(4200u32))
```
joshua-spacetime: Could you also add another test just like this one, just without the `u32` suffix? I.e. an untyped literal that's out of range:
```rust
.r#where(|a| a.age.eq(4200))
```

PR #3830, opened 2025-12-05
wasmtime: pool async stack allocations on unix

Centril:
# Description of Changes

Uses `with_host_stack` to provide a `StackCreator` that pools `FiberStack`s.
This does not use the pooling instance allocator and is limited to just stacks.

None

3? Some unsafe and wasmtime internals relied upon.

# Testing

Covered by existing tests.

### DIFF
--- a/crates/core/src/host/wasmtime/pooling_stack_creator.rs
+++ b/crates/core/src/host/wasmtime/pooling_stack_creator.rs
@@ -1,0 +1,88 @@
+use core::mem::ManuallyDrop;
+use crossbeam_queue::ArrayQueue;
+use std::sync::{Arc, Weak};
+use wasmtime::{StackCreator, StackMemory};
+use wasmtime_internal_fiber::FiberStack;
+
+/// The stack size for async stacks.
+pub const ASYNC_STACK_SIZE: usize = 2 << 20;
+
+pub struct PoolingStackCreator {
+    /// The actual pool of stacks.
+    pool: ArrayQueue<FiberStack>,
+    /// A weak reference to `self` that can be cloned
+    /// and put into a `PooledFiberStack` which uses `weak` on drop.
+    /// We do it self-referentially so that we can avoid another indirection.
+    weak: Weak<PoolingStackCreator>,
+}
+
+struct PooledFiberStack {
+    stack: ManuallyDrop<FiberStack>,
+    /// A weak reference to the pool so that `self.stack`
+    /// can be returned to the pool on drop.
+    weak: Weak<PoolingStackCreator>,
+}
+
+impl Drop for PooledFiberStack {
+    fn drop(&mut self) {
+        // SAFETY: `self.stack` is never used again.
+        let stack = unsafe { ManuallyDrop::take(&mut self.stack) };
+
+        let Some(pool) = self.weak.upgrade() else {
+            return;
+        };
+
+        let _ = pool.pool.push(stack);
+    }
+}
+
+const UNIX_SOME: &str = "FiberStack on unix always returns `Some(_)`";
+
+/// SAFETY: The implementation forwards to `FiberStack as StackMemory`,
+/// which wasmtime promises to be sound.
+unsafe impl StackMemory for PooledFiberStack {
+    fn top(&self) -> *mut u8 {
+        self.stack.top().expect(UNIX_SOME)
+    }
+
+    fn range(&self) -> std::ops::Range<usize> {
+        self.stack.range().expect(UNIX_SOME)
+    }
+
+    fn guard_range(&self) -> std::ops::Range<*mut u8> {
+        self.stack.guard_range().expect(UNIX_SOME)
+    }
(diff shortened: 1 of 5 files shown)


crates/core/src/host/wasmtime/pooling_stack_creator.rs
```diff
@@ -0,0 +1,81 @@
...
+use wasmtime::{StackCreator, StackMemory};
+use wasmtime_internal_fiber::FiberStack;
+
+pub struct PoolingStackCreator {
+    /// The actual pool of stacks.
+    pool: ArrayQueue<FiberStack>,
+    /// A weak reference to `self` that can be cloned
+    /// and put into a `PooledFiberStack` which uses `weak` on drop.
+    /// We do it self-referentially so that we can avoid another indirection.
+    weak: MaybeUninit<Weak<PoolingStackCreator>>,
```
coolreader18: I don't think this needs to be `MaybeUninit`? You can just use `new_cyclic`.

crates/core/src/host/wasmtime/pooling_stack_creator.rs
```diff
@@ -0,0 +1,81 @@
...
+unsafe impl StackCreator for PoolingStackCreator {
+    fn new_stack(&self, size: usize, zeroed: bool) -> anyhow::Result<Box<dyn wasmtime::StackMemory>, anyhow::Error> {
+        // SAFETY: `self.weak` is fully initialized whenever `new_stack` is called.
+        let weak = unsafe { self.weak.assume_init_ref() }.clone();
+
+        // Either take the stack from the pool
+        // or fall back to creating a new one.
+        let stack = weak
+            .upgrade()
+            .and_then(|pool| pool.pool.pop().map(Ok))
```
coolreader18, line 71: This doesn't take into account the requested size of the stack - there should at least be an assertion here, but ideally a hashmap from size to ArrayQueue.

crates/core/src/host/wasmtime/pooling_stack_creator.rs
```diff
@@ -0,0 +1,81 @@
...
+
+/// SAFETY: The implementation forwards to `FiberStack as StackMemory`,
+/// which wasmtime promises to be sound.
+unsafe impl StackMemory for PooledFiberStack {
+    fn top(&self) -> *mut u8 {
+        self.stack.top().unwrap()
+    }
+
+    fn range(&self) -> std::ops::Range<usize> {
+        self.stack.range().unwrap()
```
coolreader18: maybe some comments that these unwraps will never fail due to their implementation

crates/core/src/host/wasmtime/pooling_stack_creator.rs
Centril: Yep, I forgot to clean this part up. 😊

crates/core/src/host/wasmtime/pooling_stack_creator.rs
Centril: Added some `.expect(_)`s instead.

crates/core/src/host/wasmtime/pooling_stack_creator.rs
Centril: I'll add a `debug_assert_eq!` and set the async stack size to `2 << 20` (the default we currently use). I'd like to avoid the cost of a hash map when we statically know it's not needed.

coolreader18: I'd rather have it be an `assert_eq!()` - it's a single very easily predicted branch that could catch a very bad problem if our or wasmtime's behavior every changed.

crates/core/src/host/wasmtime/pooling_stack_creator.rs
```diff
@@ -0,0 +1,88 @@
+use core::mem::ManuallyDrop;
+use crossbeam_queue::ArrayQueue;
+use std::sync::{Arc, Weak};
+use wasmtime::{StackCreator, StackMemory};
+use wasmtime_internal_fiber::FiberStack;
+
+/// The stack size for async stacks.
+pub const ASYNC_STACK_SIZE: usize = 2 << 20;
```
coolreader18, line 8: Why not `1 << 21` ?

coolreader18 (approved):

crates/core/src/host/wasmtime/pooling_stack_creator.rs
Centril: I copied this from the wasmtime code. I'm guessing it's this way as `2 << 20` more easily reads as 1 MiB than `1 << 21` (every 10 added is K -> M -> G -> T -> ...).

crates/core/src/host/wasmtime/pooling_stack_creator.rs
Centril: Fair enough

PR #3831, opened 2025-12-05
 Reuse `TxState`, stashing in `CommittedState` btw txes 

Centril:
# Description of Changes

See tin.

None

2

# Testing

TODO

### DIFF
--- a/crates/datastore/src/locking_tx_datastore/committed_state.rs
+++ b/crates/datastore/src/locking_tx_datastore/committed_state.rs
@@ -1048,7 +1048,10 @@ impl CommittedState {
     /// Returns the saved tx state or creates a new one.
     /// The tx state is ready for use by a new mutable transaction.
     pub(super) fn take_tx_state(&mut self) -> TxState {
-        *self.saved_tx_state.take().unwrap_or_default()
+        let mut state = *self.saved_tx_state.take().unwrap_or_default();
+        //state.delete_tables = <_>::default();
+        //state.insert_tables = <_>::default();
+        state
     }
 
     pub(super) fn merge(&mut self, mut tx_state: TxState, read_sets: ViewReadSets, ctx: &ExecutionContext) -> TxData {
@@ -1129,7 +1132,9 @@ impl CommittedState {
             }
         }
 
-        for (&table_id, row_ptrs) in &mut tx_state.delete_tables {
+        tx_state.delete_tables.retain(|&table_id, row_ptrs| {
+            let had_deletions = !row_ptrs.is_empty();
+
             match self.get_table_and_blob_store_mut(table_id) {
                 Ok((table, blob_store, ..)) => delete_rows(
                     tx_data,
@@ -1140,12 +1145,19 @@ impl CommittedState {
                     row_ptrs.iter(),
                     truncates,
                 ),
-                Err(_) if !row_ptrs.is_empty() => panic!("Deletion for non-existent table {table_id:?}... huh?"),
+                Err(_) if had_deletions => panic!("Deletion for non-existent table {table_id:?}... huh?"),
                 Err(_) => {}
             }
 
-            row_ptrs.clear();
-        }
+            if had_deletions {
+                row_ptrs.clear();
+            }
+
+            // If there were no deletions,
+            // don't expect that access pattern in the next transaction,
+            // so remove the delete table.
+            had_deletions
+        });
 
         // Delete all tables marked for deletion.
         // The order here does not matter as once a `table_id` has been dropped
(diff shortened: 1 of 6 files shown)


crates/datastore/src/locking_tx_datastore/committed_state.rs
```diff
@@ -784,12 +783,14 @@ impl CommittedState {
                 Err(_) if !row_ptrs.is_empty() => panic!("Deletion for non-existent table {table_id:?}... huh?"),
                 Err(_) => {}
             }
+
+            row_ptrs.clear();
```
gefjon: `BTreeMap::clear` appears to not save and re-use the allocation(s) in the map, per https://doc.rust-lang.org/src/alloc/collections/btree/map.rs.html#662-671 :

```rust
    pub fn clear(&mut self) {
        // avoid moving the allocator
        drop(BTreeMap {
            root: mem::replace(&mut self.root, None),
            length: mem::replace(&mut self.length, 0),
            alloc: self.alloc.clone(),
            _marker: PhantomData,
        });
    }
```

(`alloc` is the `Allocator`, which defaults to `Global`.)

As such, saving the `delete_tables` here is not useful.

crates/datastore/src/locking_tx_datastore/committed_state.rs
```diff
@@ -784,12 +783,14 @@ impl CommittedState {
...
         }
 
         // Delete all tables marked for deletion.
         // The order here does not matter as once a `table_id` has been dropped
         // it will never be re-created.
-        for change in pending_schema_changes {
+        for change in mem::take(&mut tx_state.pending_schema_changes) {
```
gefjon, line 1165: Seems like `drain` here would be better than `mem::take`, as we could potentially re-use this allocation. We may need to be careful not to let it grow unbounded, though.

crates/datastore/src/locking_tx_datastore/committed_state.rs
```diff
@@ -847,7 +846,8 @@ impl CommittedState {
                 }
             }
 
-            let (schema, _indexes, pages) = tx_table.consume_for_merge();
+            // Cleanup `tx_table` and steals its pages.
+            let (schema, pages) = tx_table.drain_for_merge();
```
gefjon: Stealing pages out of the tx-state seems counterproductive here. If we leave the pages in the tx tables, and just empty them, it saves us having to reset their schema/row-size related metadata, including the row-present-p bitvec. Again, we have to be careful about not letting each tx table grow without bound.

crates/datastore/src/locking_tx_datastore/committed_state.rs
```diff
@@ -857,6 +857,8 @@ impl CommittedState {
             // Put all the pages in the table back into the pool.
             self.page_pool.put_many(pages);
         }
+
+        tx_state.blob_store.clear();
```
gefjon, line 1232: We could do away with the tx state blob store entirely and just use the committed state one, at the cost of adding code to rollbacks which removes the new blobs. I'm not sure if this is a good idea, since that makes rollbacks potentially much worse, unless we store the set of new blob hashes in the tx state.

crates/table/src/table.rs
```diff
@@ -1561,15 +1561,23 @@ impl Table {
...
-        (self.schema, self.indexes.into_iter(), self.inner.pages.into_page_iter())
+    /// Steals the pages of the table for merging.
+    pub fn drain_for_merge(&mut self) -> (Arc<TableSchema>, impl Iterator<Item = Box<Page>> + use<'_>) {
+        // Reset statistics.
+        self.blob_store_bytes = BlobNumBytes::default();
+        self.row_count = 0;
+
+        // Clear indices.
+        for index in self.indexes.values_mut() {
+            index.clear();
```
gefjon: Another place where `BTreeMap::clear` doesn't re-use its allocation.

crates/datastore/src/locking_tx_datastore/committed_state.rs
Centril: This shouldn't apply here, as we're calling `.clear()` on `DeleteTable`, not the outer `BTreeMap`.

crates/datastore/src/locking_tx_datastore/committed_state.rs
Centril: In a different PR, iirc, I had some code to merge blob stores.

crates/datastore/src/locking_tx_datastore/committed_state.rs
gefjon: Ah, I was remembering an earlier implementation of `DeleteTable` which was backed by a `BTreeSet`. I see that it's now a more complex data structure.

cloutiertyler:
I'm closing this because we did not find that this substantially improved performance, although it should be noted that we now have new information and that this should perhaps be reinvestigated. That reinvestigation will be tracked by a new ticket.

PR #3832, opened 2025-12-05
Remove energy multiplier from wasmtime fuel calculations

drogus:
# Description of Changes

In the past we've been converting CPU instructions into energy. We are not doing it on the SpacetimeDB side anymore, thus we should report the WasmTime fuel directly


cloutiertyler (approved):
LGTM!

PR #3834, opened 2025-12-06
Add Unreal Result type to support C++ and Blueprint

JasonAtClockwork:
# Description of Changes

- Updated Unreal codegen to work with ::Result to match how ::Option works to create a new struct for each type pair identified
- Added new integration tests to confirm Ok state for all new types added to sdk-tests
- Added new UE_SPACETIMEDB_RESULT macro to capture de/serialization for the new structs

N/A

2 - Added new type generated from Result which creates new structs for each one identified

# Testing

- [x] Updated the integration tests to include the new Result types


crates/codegen/src/unrealcpp.rs
```diff
@@ -442,8 +442,7 @@ impl Lang for UnrealCpp<'_> {
         for (param_name, param_type) in &reducer.params_for_generate.elements {
             let param_pascal = param_name.deref().to_case(Case::Pascal);
             let type_str = cpp_ty_fmt_with_module(module, param_type, self.module_name).to_string();
-            let init_str = cpp_ty_init_fmt_impl(module, param_type);
-            let field_decl = format!("{type_str} {param_pascal}{init_str}");
```
rekhoff, line 446: Does this now skip initialization of these parameters? I seem to remember Unreal liking things to be initialized. I see a placed generated code where there are no initializer due to this (I believe anyway):
`TestClient/Public/ModuleBindings/Reducers/InsertResultI32String.g.h` @ line 18, `FTestClientResultInt32String R;`

JasonAtClockwork: Great callout I'll double check, this is probably a failing due to merge conflicts as there was many with this one!

crates/codegen/src/unrealcpp.rs
```diff
@@ -5053,74 +5394,55 @@ fn cpp_ty_fmt_impl<'a>(
...
-                    } else {
-                        String::new()
-                    }
-                }
-                // Product and Sum types have proper default constructors
-                AlgebraicTypeDef::Product(_) => String::new(),
-                AlgebraicTypeDef::Sum(_) => String::new(),
-            }
-        }
+        AlgebraicTypeUse::Ref(_r) => f.write_str(""),
```
rekhoff: Similar to the other lack of initialization, this time for an `Enum`. Is this a problem? I see a placed generated code where there are no initializer due to no Enum initializer (I believe this is the reason on this one too):
`TestClient/Source/TestClient/Public/ModuleBindings/Results/TestClientResultSimpleEnumInt32.g.h` @ line 19, `ESimpleEnumType OkValue;`

sdks/unreal/src/SpacetimeDbSdk/Source/SpacetimeDbSdk/Public/BSATN/UESpacetimeDB.h
```diff
@@ -682,6 +682,62 @@ namespace UE::SpacetimeDB {
...
+        uint8_t tag = r.read_u8(); \
+        if (tag == 0) { \
+            result.IsOkField = true; \
+            result.OkValueField = deserialize<decltype(result.OkValueField)>(r); \
+        } else if (tag == 1) { \
+            result.IsOkField = false; \
+            result.ErrValueField = deserialize<decltype(result.ErrValueField)>(r); \
+        } else { \
+            ensureMsgf(false, TEXT("Invalid result tag: %d"), tag); \
+            return StructType(); \
```
rekhoff, line 736: If we get a deserialization error here, we'll get back a default struct silently. That might be good for a shipping project build, but would make troubleshooting a dev build hard since it'll look like we got an empty payload. For dev builds we could force execution to stop, in order to allow troubleshooting.
```suggestion
#if !UE_BUILD_SHIPPING
            checkf(false, TEXT("Invalid result tag: %d"), tag); \
#else
            ensureMsgf(false, TEXT("Invalid result tag: %d"), tag); \
            return StructType(); \
#endif \
```
But if that's too harsh, I think a long term solution would to have a way to propagate an error like this. Maybe an update to `UEReader` to have a way to store an error message that could be checked down steam.

JasonAtClockwork: This is actually the pattern we've used elsewhere here and with a debug build will log to the UE log system each time this happens; in a production build this will log but only once. 

The reason this is preferred is to not hard stop the client on the deserialization giving opportunity to debug as it could be a sequence issue or some bad data sent that causes it and a full crash could block debugging and as this is baked in could make some devs hesitant to change the macro for their debugging.

If we do want to consider a change like this it might be better to have it as a configuration change and wire the same type of logic into all the deserialization that could have this type of check (Optional as example).

crates/codegen/src/unrealcpp.rs
JasonAtClockwork: Confirmed this was a merge issue, I've rebuilt from `master` to make sure no changes are lost.

crates/codegen/src/unrealcpp.rs
JasonAtClockwork: Same with above this was an issue with the merge and is fixed up.

sdks/unreal/src/SpacetimeDbSdk/Source/SpacetimeDbSdk/Public/BSATN/UESpacetimeDB.h
rekhoff: Sounds good. I'll mark this as resolved. If this ever comes up as an actual problem, at least we are aware.

rekhoff (approved):
All concerns have been addressed. Testing locally again did not have issues. Approved

PR #3851, opened 2025-12-08
Rename docs dirs and files to have more digits

gefjon:
# Description of Changes

This makes it possible to insert new pages between existing pages without renaming the existing pages.

I also added a section to docs/README.md with procedure for how to name new pages, which I'll paste here:

### Adding new pages

All of our directory and file names are prefixed with a five-digit number which determines how they're sorted. 
We started with the hundreds place as the smallest significant digit, to allow using the tens and ones places to add new pages between.
When adding a new page in between two existing pages, choose a number which:

- Doesn't use any more significant figures than it needs to.
- Is approximately halfway between the previous and next page.

For example, if you want to add a new page between `00300-foo` and `00400-bar`, name it `00350-baz`. To add a new page between `00350-baz` and `00400-bar`, prefer `00370-quux` or `00380-quux`, rather than `00375-quux`, to avoid populating the ones place.

To add a new page after all previous pages, use the smallest multiple of 100 larger than all other pages. For example, if the highest-numbered existing page is `01350-abc`, create `01400-def`.

N/a

1 - @JulienLavocat said this wouldn't break anything.

# Testing

None.

### DIFF
--- a/smoketests/tests/quickstart.py
+++ b/smoketests/tests/quickstart.py
@@ -200,8 +200,8 @@ class BaseQuickstart(Smoketest):
 
 class Rust(BaseQuickstart):
     lang = "rust"
-    server_doc = STDB_DIR / "docs/docs/02-quickstarts/03-rust.md"
-    client_doc = STDB_DIR / "docs/docs/02-quickstarts/03-rust.md"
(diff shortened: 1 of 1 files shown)


JasonAtClockwork (approved):
LGTM! All tests are passing now

PR #3853, opened 2025-12-08
[TS] Fix useTable casing issues

### LINKED ISSUE #3807: Casing issue in where clause for `useTable`
The TS sdk doesnt change the casing of the members when subscribing to something.
e.g. when i do: `useTable(tables.xy, where(eq("importantId", 1)))`
it doesn't find something since the query thats being output is:
`SELECT * FROM xy WHERE importantId = 1` and it should be 
`SELECT * FROM xy WHERE important_id = 1`

coolreader18:
# Description of Changes

Fixes #3807. Not sure what the best way to model this in the API is (another argument to `t.row()`?), but it does work.

1

# Testing

- [ ] 
- [ ]


crates/bindings-typescript/src/lib/schema.ts
```diff
@@ -100,11 +101,12 @@ export function tablesToSchema<
         accessorName: toCamelCase(schema.tableName),
         columns: schema.rowType.row, // typed as T[i]['rowType']['row'] under TablesToSchema<T>
         rowType: schema.rowSpacetimeType,
+        originalColumnNames: schema.rowType.originalNames,
         constraints: [
           ...schema.tableDef.constraints.map(c => ({
             name: c.name,
             constraint: 'unique' as const,
-            columns: Array.from(c.data.value.columns.map(i => colNameList[i])),
```
cloutiertyler, line 107: Why the cast here? Are we sure that's a string tuple?

cloutiertyler:
@coolreader18 I think this is the way we should model it:

```ts
export default new __t.row(
  {
    id: __t.u32().primaryKey(),
    userId: __t.identity("user_id"),
    name: __t.string(),
    get location() {
      return Point;
    },
  },
);
```

or

```ts
export default new __t.row(
  {
    id: __t.u32().primaryKey(),
    userId: __t.identity.name("user_id"),
    name: __t.string(),
    get location() {
      return Point;
    },
  },
);
```

perhaps. That is similar to what DrizzleORM does: https://orm.drizzle.team/docs/sql-schema-declaration#shape-your-data-schema.

If you think you can swing that quickly, I'd like to make that change. I'd rather do something like that because it will be how we will implement a more general "column alias" feature, but if we need to rush this we could do what you're doing.

I'm a little hesitant to expose `RowBuilder` though since we're not even sure that that should exist, to your point.

coolreader18:
Having it just on ColumnBuilder for now is reasonable, I guess. The issue is that we'd also want to support having alternate names for product type fields as well.

cloutiertyler:
> Having it just on ColumnBuilder for now is reasonable, I guess. The issue is that we'd also want to support having alternate names for product type fields as well.

Agreed, although we're moving in that direction and I think this would be easy enough to extend to that as well.

crates/bindings-typescript/src/lib/schema.ts
coolreader18: Yes, we're casting from `string[]` to `[string]`, but it doesn't really matter ultimately. The reason that `[string]` is expected is to enforce on users, not on our internal code.

coolreader18:
Mmm, I don't know if easy is the right word - it would involve making a new category of TypeBuilders (like, `NamedU8Builder` etc) that can be used as product fields but not in an array or option

PR #3854, opened 2025-12-08
Update typescript package size limits

jdetter:
# Description of Changes

Updated package sizes from a failed release dry-run: https://github.com/clockworklabs/SpacetimeDBPrivate/actions/runs/20045893743/job/57491806595

None

1 - this is just a release fix

# Testing

I have not tested this, it is a trivial change.


JulienLavocat (approved):

JulienLavocat:
I am starting to be slightly concerned by how frequently we need to make this type of adjustment.
Also, if I remember correctly, this release includes a fix that replaces an SDK dependency with our own implementation. Shouldn’t that have reduced the bundle size?

jdetter:
> I am starting to be slightly concerned by how frequently we need to make this type of adjustment. Also, if I remember correctly, this release includes a fix that replaces an SDK dependency with our own implementation. Shouldn’t that have reduced the bundle size?

I'm not sure, I haven't been keep track of the typescript changes for this release - does anything seem wrong here?

Issue #3817, opened 2025-12-03
TransactionUpdate messages for scheduled reducers have no reducer name, TS SDK message processing expects it


aceysmith:
Rust package 1.10.0, Standalone version 1.10.0, TS SDK 1.10.0

Expected:
If a row is deleted as a result of a scheduled reducer completing, my client should be able to properly process that event and remove the data from my local cache

Actual:
The processing of the TransactionUpdate event fails. The processMessage queue is never cleared out and all further events are stuck.

`Unhandled Promise Rejection: TypeError: undefined is not an object (evaluating 'reducerInfo.reducerName')`

Speculated issue:
The issue seems to be one of two scenarios:

1) TransactionUpdate events from the server for rows deleted as a result of firing scheduled reducers are incorrectly missing a reducer name
2) The TS SDK does not properly handle transactions with no reducer name. This is the code that raises an exception:
Likely this code https://github.com/clockworklabs/SpacetimeDB/blob/master/crates/bindings-typescript/src/sdk/db_connection_impl.ts#L754-L757

More information in the [bugs](https://discord.com/channels/1037340874172014652/1444274078352539811) Discord

aceysmith:
Sample POC project as well. If you open a second window it will also create an event message and you can see on the first that nothing gets processed.

[event-table-bug-poc.zip](https://github.com/user-attachments/files/23917224/event-table-bug-poc.zip)

PR #3858, opened 2025-12-09
Handle unknown transactions without erroring.

jsdt:
# Description of Changes

Don't throw an error if there is no `reducerInfo`. The code was previously trying to handle the case of an unknown reducer, but was effectively asserting that the reducerInfo existed too soon.

Now we should be fine handling a transaction, even if we can't determine reducer information for it.

1.

# Testing

I tested this by running the quickstart chat app, and using the CLI to delete rows (via `spacetime sql`). Before this change, the client errors, but now it handles it correctly.

I also tested with the repro in https://github.com/clockworklabs/SpacetimeDB/issues/3817


aceysmith:
Thanks for the fix! Hoping we’ll get a 1.12.0 or similar before the end of the year that includes this change 😊

PR #3863, opened 2025-12-10
Add WASM point scan ABIs & use them in Rust bindings

Centril:
# Description of Changes

Provides new WASM ABIs:

- `datastore_index_scan_point_bsatn`
- `datastore_delete_by_index_scan_point_bsatn`

These are then used where applicable to speed up `.find(_)` and friends.

Point scans are also used more internally where applicable.

What remains after this is use in C# module bindings and to expose this in TS as well.

The PR makes TPS go from roughly 36k to 38k TPS on my machine and also makes a difference in flamegraphs where the time spent in some index scans are substantially decreased.

None

3? This touches the datastore an how we expose it to modules.

# Testing

Some existing tests now exercise the new ABIs by changing what `.find(_)` and friends do.

### DIFF
--- a/crates/datastore/src/locking_tx_datastore/mut_tx.rs
+++ b/crates/datastore/src/locking_tx_datastore/mut_tx.rs
@@ -276,6 +276,7 @@ impl MutTxId {
     // This is cold as we don't want it to be inlined in case it doesn't end up getting called.
     // This previously showed up in flamegraphs.
     #[cold]
+    #[inline(never)]
     pub fn record_index_scan_range_inner(
         &mut self,
         view: &ViewCallInfo,


coolreader18 (approved):
Can't speak to the datastore stuff (unless you want me to), but the bindings stuff lgtm

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -254,38 +258,77 @@ impl MutTxId {
...
-            return;
+        if let FuncCallType::View(view) = op {
+            self.record_index_scan_range_inner(view, table_id, index_id, lower, upper);
         };
+    }
 
-        // Check for precise index seek
+    // This is cold as we don't want it to be inlined in case it doesn't end up getting called.
+    // This previously showed up in flamegraphs.
+    #[cold]
```
coolreader18, line 278: `#[inline(never)]` might make more sense if you simply don't want it to be inlined? [`#[cold]`](https://doc.rust-lang.org/reference/attributes/codegen.html#the-cold-attribute) more has to do with like, the way rust/llvm optimize it and calls to it, and honestly it's overall really unclear in general. I don't think it necessarily prevents inlining, though.

coolreader18 (commented):
datastore LGTM

crates/datastore/src/locking_tx_datastore/mut_tx.rs
Centril, line 278:
```suggestion
    #[cold]
    #[inline(never)]
```

PR #3868, opened 2025-12-10
Move `do_durability` work into a `DurabilityWorker` so that it doesn't block the reducer processing core

Centril:
# Description of Changes

This yields a performance win of about 2k TPS on my machine.
With the whole stack of this PR applied, I'm at 40.5k TPS.

None

1

# Testing

Covered by existing tests.

### DIFF
--- a/crates/core/src/db/durability.rs
+++ b/crates/core/src/db/durability.rs
@@ -1,5 +1,4 @@
 use crate::db::persistence::Durability;
-use futures::{channel::mpsc, StreamExt};
 use spacetimedb_commitlog::payload::{
     txdata::{Mutations, Ops},
     Txdata,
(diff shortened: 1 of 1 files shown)


crates/core/src/db/durability.rs
```diff
@@ -0,0 +1,178 @@
...
+    request_tx: UnboundedSender<DurabilityRequest>,
+    durability: Arc<Durability>,
+}
+
+/// Those who run seem to have all the fun... 🎶
+const HUNG_UP: &str = "durability actor hung up / panicked";
+
+impl Drop for DurabilityWorker {
+    fn drop(&mut self) {
+        self.request_tx.send(DurabilityRequest::Close).expect(HUNG_UP);
```
kim: I think this needs to just ignore it if the receiving end went away.
There are reasons the real durability worker can go away and then `append_tx` panics.

Not sure though why this would affect the test suite.

Centril: Apparently we also clone `RelationalDB` in tests which explains the test failure.

kim:
Merged via #3880

PR #3878, opened 2025-12-12
Strip sensitive info from HTTP errors

gefjon:
# Description of Changes 

`reqwest` includes the full URL in its errors, including query params. This is unfortunate, as query params can contain sensitive info like API tokens. It's difficult for modules to clean these themselves, as they see errors as strings, losing the structure of `reqwest::Error`.
In this commit, we strip query parts out of URLs in errors before returning them to modules. I've also audited all of the error return paths in the `http_request` method and left comments justifying why the unchanged ones are safe.

Only if you consider the format of error messages part of our API, which I don't. Procedure APIs aren't stable yet anyways.

1

# Testing

None yet - accepting input from reviewers about desired tests if we feel that's necessary.

- [ ] 
- [ ]


crates/core/src/host/instance_env.rs
```diff
@@ -714,6 +714,14 @@ impl InstanceEnv {
         self.procedure_last_tx_offset.take()
     }
 
+    /// Perform an HTTP request.
+    /// Exposed to modules via the `ProcedureContext`.
+    ///
+    /// It's very important that the error return from this function
```
JulienLavocat:
```suggestion
    /// It's very important that the error returned from this function
```

PR #3880, opened 2025-12-15
Async shutdown for database / durability

kim:
Controlled shutdown of a database should drain the outstanding transactions
queue(s) and flush them to the durability layer.

With the introduction of another queueing layer in #3868, it became harder to
observe when or if this process is completed.

This patch thus introduces an explicit (async) shutdown method for
`RelationalDB` and below, which will wait until all submitted transactions are
either reported durable, or an error occurs in the durability layer.

`RelationalDB` is made `!Clone`, such that shutdown can be initiated in the
`Drop` impl. Note that this requires access to a tokio runtime, which we thread
through via the `Persistence` services in order to allow control over which of
the various runtimes is being used for durability-related tasks.

Also moves `RelationalDB::open` to a blocking thread when a persistence-enabled
database is constructed by the `HostController` -- this process performs heavy
I/O and can take a substantial amount of time, during which we don't want to
block a worker thread.

None

3

# Testing

- [ ] some testing added
- [ ] existing tests still pass
- [ ] `impl Drop for RelationalDB` difficult to test, extra eyeballs needed

### DIFF
--- a/crates/core/src/db/durability.rs
+++ b/crates/core/src/db/durability.rs
@@ -6,7 +6,7 @@ use std::{
     time::Duration,
 };
 
-use log::{info, warn};
+use log::{error, info, warn};
 use spacetimedb_commitlog::payload::{
     txdata::{Mutations, Ops},
     Txdata,
@@ -143,8 +143,8 @@ impl DurabilityWorker {
     /// re-opened for writing while there is still an active background task
     /// writing to the commitlog.
     ///
-    /// The shutdown task will be scheduled onto the tokio runtime provided
-    /// to [Self::new]. This means that the task may still be running when this
+    /// The shutdown task will be spawned onto the tokio runtime provided to
+    /// [Self::new]. This means that the task may still be running when this
     /// method returns.
     ///
     /// `database_identity` is used to associate log records with the database
@@ -155,7 +155,7 @@ impl DurabilityWorker {
         let rt = self.runtime.clone();
         let mut shutdown = rt.spawn(async move { self.shutdown().await });
         rt.spawn(async move {
-            let label = format!("database={database_identity}");
+            let label = format!("[{database_identity}]");
             let start = Instant::now();
             loop {
                 // Warn every 5s if the shutdown doesn't appear to make progress.
@@ -165,7 +165,7 @@ impl DurabilityWorker {
                 match timeout(Duration::from_secs(5), &mut shutdown).await {
                     Err(_elapsed) => {
                         let since = start.elapsed().as_secs_f32();
-                        warn!("{label} waiting for durability worker shutdown since {since}s",);
+                        error!("{label} waiting for durability worker shutdown since {since}s",);
                         continue;
                     }
                     Ok(res) => {
@@ -281,6 +281,7 @@ mod tests {
     use std::{pin::pin, task::Poll};
 
     use pretty_assertions::assert_matches;
+    use spacetimedb_sats::product;
     use tokio::sync::watch;
 
     use super::*;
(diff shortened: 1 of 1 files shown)


kim:
Merging also merges #3868

crates/core/src/db/durability.rs
```diff
@@ -59,15 +67,24 @@ impl DurabilityWorker {
...
         self.request_tx
-            .send(DurabilityRequest::Work {
+            .send(DurabilityRequest {
                 reducer_context,
                 tx_data: tx_data.clone(),
             })
+            .inspect(|()| {
+                self.requested_tx_offset
+                    .fetch_max(tx_data.tx_offset().unwrap_or_default(), Ordering::SeqCst);
+            })
```
Centril, line 107: This could use some commentary as to the why.

crates/core/src/db/durability.rs
```diff
@@ -76,39 +93,50 @@ impl DurabilityWorker {
...
-                // there should be no more `Work` incoming or buffered,
-                // as `Close` hangs up the receiver end of the channel,
-                // so nothing can be sent to it.
-                DurabilityRequest::Close => {
-                    drop(self.durability);
-                    drop(self.request_rx);
-                    return;
+        loop {
+            tokio::select! {
+                biased;
```
Centril, line 202: Could use a comment as to the why we want to poll shutdown first.

crates/core/src/db/relational_db.rs
```diff
@@ -141,14 +137,23 @@ impl std::fmt::Debug for RelationalDB {
...
-                });
-            }
+            let lock_file = self.lock_file.clone();
+            let rt = worker.runtime().clone();
+            rt.spawn(async move {
+                if let Err(e) = worker.shutdown().await {
+                    log::warn!("error shutting down durability worker: {e:#}");
+                }
+                drop(lock_file);
+            });
```
Centril: Consider moving this into a method in `DurabilityWorker` that takes `lock_file` as a parameter.

crates/core/src/db/durability.rs
```diff
@@ -0,0 +1,310 @@
...
+    pub async fn shutdown(&self) -> anyhow::Result<TxOffset> {
+        self.shutdown
+            .send(())
+            .await
+            .context("durability worker already closed")?;
+        // Wait for the channel to be closed.
+        self.request_tx.closed().await;
+        // Load the latest tx offset and wait for it to become durable.
+        let latest_tx_offset = self.requested_tx_offset.load(Ordering::SeqCst);
+        let durable_offset = self.durable_tx_offset().wait_for(latest_tx_offset).await?;
```
Shubham8287, line 134: I am wondering, if IO error caused by disk can lead to `shutdown`being stuck forever?

Should we make `durability::FlushAndSync` to exit on encountering error?

kim: > Should we make durability::FlushAndSync to exit on encountering error?

You mean like _any_ error? That's probably a good idea.

kim: I'm also considering to wrap `shutdown` in a `timeout` when called from the `Drop` impl, just in case. The replicated durability could have some weird network issue, and we don't want to block `drop` forever.

Shubham8287: > The replicated durability could have some weird network issue, and we don't want to block `drop` forever.

Makes sense, If we do this we can avoid breaking `durability::FlushAndSync ` loop on error to retry.

kim: I figured that we need to do both, as long as there is no way to just abort the backing durability impl. Otherwise, how do we conclude that it is safe to release the lock file?

PR #3881, opened 2025-12-15
CI - Investigate `rusty_v8` link error

bfops:
# Description of Changes

I am unsure why this happened, but it seems to be cache related. Invalidating the cache version fixes it.

None.

1

# Testing

- [x] `csharp-testsuite` passes now

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -686,18 +686,9 @@ jobs:
           save-if: false
           prefix-key: v1
 
-      # This step shouldn't be needed, but somehow we end up with caches that are missing librusty_v8.a.
-      # ChatGPT suspects that this could be due to different build invocations using the same target dir,
-      # and this makes sense to me because we only see it in this job where we mix `cargo build -p` with
-      # `cargo build --manifest-path` (which apparently build different dependency trees).
-      # However, we've been unable to fix it so... /shrug
       - name: Check v8 outputs
         run: |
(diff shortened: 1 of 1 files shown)


sdks/csharp/tools~/gen-quickstart.sh
```diff
@@ -6,5 +6,6 @@ SDK_PATH="$(dirname "$0")/.."
 SDK_PATH="$(realpath "$SDK_PATH")"
 STDB_PATH="$SDK_PATH/../.."
 
-cargo build --manifest-path "$STDB_PATH/crates/standalone/Cargo.toml"
-cargo run --manifest-path "$STDB_PATH/crates/cli/Cargo.toml" -- generate -y -l csharp -o "$SDK_PATH/examples~/quickstart-chat/client/module_bindings" --project-path "$SDK_PATH/examples~/quickstart-chat/server"
+cd "$STDB_PATH"
```
jdetter, line 9: Is this change supposed to be here?

jdetter (approved):
Seems fine although check to make sure the quickstart change is actually required for this fix

sdks/csharp/tools~/gen-quickstart.sh
bfops: whoops, I thought I reverted that :thinking:

coolreader18:
I don't think this is necessary, you can just manually delete corrupt caches from the github ui

bfops:
@coolreader18  fixed this! (See above)

PR #3883, opened 2025-12-15
Make /v1/database/:name/call/:func call procedures too, remove procedure route

### LINKED ISSUE #3659: Procedures: Make `/v1/database/:name/call/:func` call procedures too, remove procedure route

mamcx:
# Description of Changes

Closes #3659 

Remove route and alter the semantics of the `call` route on both server and `cli`

1

# Testing

- [x] Publish module with `procedures` and observe calling the `cli` the result is print.

### DIFF
--- a/crates/cli/src/subcommands/call.rs
+++ b/crates/cli/src/subcommands/call.rs
@@ -9,23 +9,25 @@ use convert_case::{Case, Casing};
 use itertools::Itertools;
 use spacetimedb_lib::sats::{self, AlgebraicType, Typespace};
 use spacetimedb_lib::{Identity, ProductTypeElement};
-use spacetimedb_schema::def::{ModuleDef, ReducerDef};
+use spacetimedb_schema::def::{ModuleDef, ProcedureDef, ReducerDef};
 use std::fmt::Write;
 
 use super::sql::parse_req;
 
 pub fn cli() -> clap::Command {
     clap::Command::new("call")
-        .about(format!("Invokes a reducer function in a database. {UNSTABLE_WARNING}"))
+        .about(format!(
+            "Invokes a reducer function OR procedure in a database. {UNSTABLE_WARNING}"
+        ))
         .arg(
             Arg::new("database")
                 .required(true)
                 .help("The database name or identity to use to invoke the call"),
         )
         .arg(
-            Arg::new("reducer_name")
+            Arg::new("reducer_procedure_name")
                 .required(true)
-                .help("The name of the reducer to call"),
+                .help("The name of the reducer OR procedure to call"),
         )
         .arg(Arg::new("arguments").help("arguments formatted as JSON").num_args(1..))
         .arg(common_args::server().help("The nickname, host name or URL of the server hosting the database"))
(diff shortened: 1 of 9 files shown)


gefjon (approved):
Code looks good. I tested locally and was able to call procedures. I'd like to see a smoketest added, please - something trivial that just calls a procedure and inspects its return value.

bfops:
It sounds like this is an API break - does that mean that older versions of the CLI will not be able to `spacetime call` to versions of the server that have this PR included?

crates/cli/src/subcommands/call.rs
```diff
@@ -34,9 +36,35 @@ pub fn cli() -> clap::Command {
         .after_help("Run `spacetime help call` for more detailed information.\n")
 }
 
+enum CallResult<'a> {
```
bfops: it looks like this is just a description of what the call result will be, not the result itself, is that right?

crates/cli/src/subcommands/call.rs
```diff
@@ -164,51 +213,70 @@ impl std::fmt::Display for ReducerSignature<'_> {
...
 }
 
 const REDUCER_PRINT_LIMIT: usize = 10;
 
 /// Provided the schema for the database,
-/// decorate `error` with more helpful info about reducers.
-fn add_reducer_ctx_to_err(error: &mut String, module_def: &ModuleDef, reducer_name: &str) {
-    let mut reducers = module_def
+/// decorate `error` with more helpful info about reducers and procedures.
+fn add_reducer_procedure_ctx_to_err(error: &mut String, module_def: &ModuleDef, reducer_name: &str) {
```
bfops, line 225: did you test some of these error cases as well? either manually or with a smoketest?

gefjon:
> It sounds like this is an API break - does that mean that older versions of the CLI will not be able to `spacetime call` to versions of the server that have this PR included?

I don't think that's true. When calling reducers, the return format of the route is unchanged, to the best of my ability to determine. It's only when calling a procedure that the route returns something different, and the old CLI won't do that because it checks the moduledef first.

bfops:
> > It sounds like this is an API break - does that mean that older versions of the CLI will not be able to `spacetime call` to versions of the server that have this PR included?
> 
> I don't think that's true. When calling reducers, the return format of the route is unchanged, to the best of my ability to determine. It's only when calling a procedure that the route returns something different, and the old CLI won't do that because it checks the moduledef first.

Okay, great. I was just going off of this description:
> Remove route and alter the semantics of the `call` route on both server and `cli`
And I assume the old route was unstable so it's fine to remove.

crates/cli/src/subcommands/call.rs
mamcx: Correct, will change name to be more clear

crates/cli/src/subcommands/call.rs
mamcx: Done

crates/cli/src/subcommands/call.rs
```diff
@@ -9,23 +9,25 @@ use convert_case::{Case, Casing};
...
+use spacetimedb_schema::def::{ModuleDef, ProcedureDef, ReducerDef};
 use std::fmt::Write;
 
 use super::sql::parse_req;
 
 pub fn cli() -> clap::Command {
     clap::Command::new("call")
-        .about(format!("Invokes a reducer function in a database. {UNSTABLE_WARNING}"))
+        .about(format!(
+            "Invokes a reducer function OR procedure in a database. {UNSTABLE_WARNING}"
```
jdetter:
```suggestion
            "Invokes a reducer function or procedure in a database. {UNSTABLE_WARNING}"
```

crates/cli/src/subcommands/call.rs
```diff
@@ -9,23 +9,25 @@ use convert_case::{Case, Casing};
...
             Arg::new("database")
                 .required(true)
                 .help("The database name or identity to use to invoke the call"),
         )
         .arg(
-            Arg::new("reducer_name")
+            Arg::new("reducer_procedure_name")
                 .required(true)
-                .help("The name of the reducer to call"),
+                .help("The name of the reducer OR procedure to call"),
```
jdetter:
```suggestion
                .help("The name of the reducer or procedure to call"),
```

crates/cli/src/subcommands/call.rs
jdetter: should we rename `reducer_name` to something like `reducer_procedure_name`?

crates/cli/src/subcommands/call.rs
jdetter: ^ if you accept this change you'll have to regen the CLI docs

crates/cli/src/subcommands/call.rs
jdetter: ^ if you accept this change you'll have to regen the CLI docs

jdetter (approved):
Optional nits, seems fine to me. I have not manually tested this but it looks like we've added some smoketests so I'm happy with that 👍

crates/cli/src/subcommands/call.rs
gefjon: I'm chaing this to "Invokes a function (reducer or procedure) [...]"

crates/cli/src/subcommands/call.rs
gefjon: I'm changing this to just "function", and the arg name to `function_name`.

PR #3885, opened 2025-12-16
Don't block worker thread waiting for SQL transaction

kim:
Enough SQL requests piling up can exhaust the tokio worker pool waiting
to acquire a transaction lock. Move `begin_mut_tx` to the blocking pool
instead.

1

# Testing

Covered by existing tests.


gefjon:
I'm also confused by your PR description. This function runs on the database cores, so there should only be one Tokio worker, right?

kim:
Ah right, I had actually discarded this after realizing it's not the main runtime.
Disregard.

Issue #3465, opened 2025-10-23
`table to remove should exist` when accessing database


egormanga:
Hi. I've been experimenting on my _staging_ environment (silly me!) and at some point, perhaps on a StDB crash, it bricked my main module (which I didn't touch). Now, any access to it (even publishing) triggers this error:
```rust
thread 'tokio-runtime-worker' panicked at crates/datastore/src/locking_tx_datastore/committed_state.rs:362:18:
table to remove should exist
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::option::expect_failed
   3: spacetimedb_datastore::locking_tx_datastore::committed_state::CommittedState::replay_delete_by_rel
   4: <spacetimedb_datastore::locking_tx_datastore::datastore::ReplayVisitor<F> as spacetimedb_commitlog::payload::txdata::Visitor>::visit_delete
   5: spacetimedb_commitlog::payload::txdata::process_record
   6: spacetimedb_commitlog::payload::Decoder::consume_record
   7: spacetimedb_commitlog::commitlog::Generic<R,T>::fold_transactions_from
   8: <spacetimedb_durability::imp::local::Local<T> as spacetimedb_durability::History>::fold_transactions_from
   9: spacetimedb::db::relational_db::RelationalDB::open
  10: spacetimedb::host::host_controller::Host::try_init::{{closure}}::{{closure}}
  11: spacetimedb::host::host_controller::HostController::try_init_host::{{closure}}
  12: spacetimedb::host::host_controller::HostController::watch_maybe_launch_module_host::{{closure}}::{{closure}}::{{closure}}
<…tokio…>
```

Apparently, at some point it didn't finish writing some file (offset log? snapshot?) and since then it couldn't replay the commitlog.

I don't care for the lost staging data _all that much_, but this should make an example of **one way StDB can break its state**, so I'm posting this in hope to hear the possible reasons this has happened, with maybe steps to manually recover the data, and to hopefully implement a fix so that this will never happen in someone's production.

egormanga:
Apparently, it's the second time this happens to me — I tried launching my older `data.gone` backup, whose _old_ logs read as:
```
crates/client-api/src/lib.rs:377: internal error: failed to get or launch module host: TableError: Table with ID `0` not found in `st_table`.: Table with ID `0` not found in `st_table`.
```
and with it, I also see the same behavior described in the issue. So the problem isn't new—the error string is.

[...]

PR #3887, opened 2025-12-16
Add a fault-tolerant commitlog replay mode for use in debugging

gefjon:
# Description of Changes

When debugging broken commitlogs, we want to inspect the whole commitlog, including the part after the first error.
This is in contrast with the way we want to replay in prod, where we'd rather get a hard error than an incorrect state.

This commit adds a new flag to commitlog replay, `ErrorBehavior`. The `core` crate passes `ErrorBehavior::FailFast`
when replaying commitlogs to reconstruct databases. Internal tooling (not in this repository) uses `ErrorBehavior::Warn` to print the entirety of a broken commitlog.

Changes internal APIs only.

1 - no change to behavior of SpacetimeDB.

# Testing

None.


egormanga:
* This is probably related to #3465?

PR #3888, opened 2025-12-16
cli: Flush stdout after printing the database tree

kim:
`spacetime delete` may print the database tree and ask for confirmation. We did not flush stdout, causing the tree to be interleaved with the yes/no prompt.

1


bfops (approved):
Makes sense to me!

Issue #3811, opened 2025-12-02
Codegen [TS] generating duplicate identifiers


NullP0inter:
crates/codegen/src/typescript.rs is generating typescript code with duplicate identifier errors.

Specifically, the imports and exports inside the generate_global_files function and within the reducer arg types, procedure arg types, all types sections have some duplicates in the generated index.ts bindings.

The table handle types section is good as it looks to already has a suffix of Row added.

Testing with some local build changes adding a suffix to the other import / export sections works well.

Examples of duplicate identifiers can be seen in the repo here: crates/bindings-typescript/examples/basic-react/src/module_bindings/index.ts such as with the SayHello import / export being duplicated.

<img width="1881" height="1447" alt="Image" src="https://github.com/user-attachments/assets/f28fc56f-af20-49e3-9de0-67691899a979" />

cloutiertyler:
Thank you for filing this. We're rounding up a bunch of small issues like these to fix them ASAP.

NullP0inter:
I should mention I haven't tested with any views yet they might have the same issue.

gh0stbyt3:
I'm seeing the same issue and I think this might be completely blocking for people who want to be on the TypeScript path.

I've just started tinkering with SpacetimeDB and I don't yet understand how it all works, so what I'm thinking might not even work, but I'm toying with the idea of having some sort of temporary solution while we wait for a fix from SpacetimeDB. It's December 6th and we might not get an official fix before the end of the month and I _really_ want to play around with this over the winter holidays.

I apologize in advance if this wastes anyone's time, but it would be neat if someone who already has experience with SpacetimeDB + TypeScript could verify if the solution below would even be a usable temporary fix; I have not yet had the time to check for any additional consequences beyond verifying that it actually fixes the generated TypeScript bindings (which it does).

Maybe OP @NullP0inter, @cloutiertyler or @coolreader18 can take a peek and weigh in. I think OP alluded to the fact that this might actually be a reasonable temporary fix, but I'm not sure that's what they meant.

It looks like the generated bindings file has duplicate identifiers for reducers and types, so at a high level a temporary solution could be that we simply rename the identifiers that are exported from all "type" paths and keep the identifiers exported from "reducer" paths as they appear to be used to construct the reducer schema.

I am making a big assumption that this is all that is "wrong" with the generated bindings, but if that's the case then may I propose a script that would:

### 1. create a backup of the modified file

[...]

PR #3889, opened 2025-12-16
Add Reducer to the end of reducer arg types to avoid collisions

jsdt:
# Description of Changes

This helps with the issue reported in https://github.com/clockworklabs/SpacetimeDB/issues/3811.

Right now we have a type representing the reducer, and a type for the reducer args, and both have the same name. This adds `Reducer` to the end of the args type, which is similar to what we are doing for procedure arguments or the `Row` suffix for tables.

This will still cause some potential problems, since someone could have a type that ends in `Reducer` (or `Row`), but this will fix the majority of issues that are currently breaking people.

This also has some changes to get the basic react example to build.

This is technically a breaking change if people are using this for type annotations (which doesn't seem too likely), but it should be an easy one for people to fix.

1.

# Testing

I tested the quickstart.

### DIFF
--- a/crates/bindings-typescript/src/lib/schema.ts
+++ b/crates/bindings-typescript/src/lib/schema.ts
@@ -36,7 +36,7 @@ import {
 } from './algebraic_type';
 import type RawScopedTypeNameV9 from './autogen/raw_scoped_type_name_v_9_type';
 import type { CamelCase } from './type_util';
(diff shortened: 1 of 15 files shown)


cloutiertyler (approved):
This is definitely the fix we should do at least for now. LGTM.

PR #3891, opened 2025-12-16
Ignore system meta-descriptor rows during replay

gefjon:
# Description of Changes

Based on #3887 . Review starting from commit 233b48cc4.

We've encountered a commitlog which includes inserts into `st_table`, `st_column`, &c of the rows which describe `st_view`, `st_view_param`, &c. This caused replay to fail, as those rows were already inserted during bootstrapping,
so we got set-semantic duplicate errors. With this commit, we ignore set-semantic duplicate errors when replaying a commitlog specifically for rows in system tables which describe system tables.

We also have to do an additional fixup for sequences. This is described in-depth in comments added at the relevant locations.

N/a

1 - I was careful not to swallow any errors which aren't obviously safe.

# Testing

- [x] Manually replayed commitlog which includes the above mentioned inserts, got error prior to this commit, no error with this commit.

### DIFF
--- a/crates/datastore/src/locking_tx_datastore/committed_state.rs
+++ b/crates/datastore/src/locking_tx_datastore/committed_state.rs
@@ -91,6 +91,22 @@ pub struct CommittedState {
     ///     - system tables: `st_view_sub`, `st_view_arg`
     ///     - Tables which back views.
     pub(super) ephemeral_tables: EphemeralTables,
+
+    /// Rows within `st_column` which should be ignored during replay
+    /// due to having been superseded by a new row representing the same column.
+    ///
+    /// During replay, we visit all of the inserts table-by-table, followed by all of the deletes table-by-table.
+    /// This means that, when multiple columns of a table change type within the same transaction,
+    /// we see all of the newly-inserted `st_column` rows first, and then later, all of the deleted rows.
+    /// We may even see inserts into the altered table before seeing the `st_column` deletes!
+    ///
+    /// In order to maintain a proper view of the schema of tables during replay,
+    /// we must remember the old versions of the `st_column` rows when we insert the new ones,
+    /// so that we can respect only the new versions.
+    ///
+    /// We insert into this set during [`Self::replay_insert`] of `st_column` rows
+    /// and delete from it during [`Self::replay_delete`] of `st_column` rows.
+    replay_columns_to_ignore: HashSet<RowPointer>,
 }
 
 impl CommittedState {
@@ -120,6 +136,7 @@ impl MemoryUsage for CommittedState {
             table_dropped,
             read_sets,
             ephemeral_tables,
+            replay_columns_to_ignore,
         } = self;
         // NOTE(centril): We do not want to include the heap usage of `page_pool` as it's a shared resource.
         next_tx_offset.heap_usage()
@@ -129,6 +146,7 @@ impl MemoryUsage for CommittedState {
             + table_dropped.heap_usage()
             + read_sets.heap_usage()
             + ephemeral_tables.heap_usage()
+            + replay_columns_to_ignore.heap_usage()
     }
 }
 
@@ -199,6 +217,7 @@ impl CommittedState {
             read_sets: <_>::default(),
             page_pool,
             ephemeral_tables: <_>::default(),
+            replay_columns_to_ignore: <_>::default(),
         }
     }
 
@@ -483,7 +502,7 @@ impl CommittedState {
         let (table, blob_store, _, page_pool) = self.get_table_and_blob_store_mut(table_id)?;
 
         // Delete the row.
-        table
+        let row_ptr = table
             .delete_equal_row(page_pool, blob_store, row)
             .map_err(TableError::Bflatn)?
             .ok_or_else(|| anyhow!("Delete for non-existent row when replaying transaction"))?;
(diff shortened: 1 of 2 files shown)


crates/datastore/src/system_tables.rs
```diff
@@ -108,6 +110,73 @@ pub(crate) const ST_VIEW_ARG_NAME: &str = "st_view_arg";
...
+            table_id_is_reserved(row.table_id)
+        }
+        ST_ROW_LEVEL_SECURITY_ID => {
+            // We don't install any RLS rules on system tables automatically, but users can.
+            // This means that `st_row_level_security` rules are never system meta-descriptors,
+            // in the sense that if they exist, they come from users.
+            false
+        }
+        ST_CONNECTION_CREDENTIALS_ID => false,
+        // We don't define any system views, so none of the view-related tables can be system meta-descriptors.
```
kim, line 164: Didn't we just talk about doing that in the future? Maybe future-proofing these branches couldn't hurt either.

gefjon: I am not sure how to appropriately future-proof these branches. @joshua-spacetime may have input.

crates/datastore/src/system_tables.rs
```diff
@@ -108,6 +110,73 @@ pub(crate) const ST_VIEW_ARG_NAME: &str = "st_view_arg";
...
+            let row: StIndexRow = to_typed_row(row)?;
+            table_id_is_reserved(row.table_id)
+        }
+        ST_CONSTRAINT_ID => {
+            let row: StConstraintRow = to_typed_row(row)?;
+            table_id_is_reserved(row.table_id)
+        }
+        ST_MODULE_ID => false,
+        ST_CLIENT_ID => false,
+        ST_VAR_ID => false,
```
Centril:
```suggestion
        ST_MODULE_ID
        | ST_CLIENT_ID
        | ST_VAR_ID => false,
```

crates/datastore/src/system_tables.rs
```diff
@@ -108,6 +110,73 @@ pub(crate) const ST_VIEW_ARG_NAME: &str = "st_view_arg";
...
+            false
+        }
+        ST_CONNECTION_CREDENTIALS_ID => false,
+        // We don't define any system views, so none of the view-related tables can be system meta-descriptors.
+        ST_VIEW_ID => false,
+        ST_VIEW_PARAM_ID => false,
+        ST_VIEW_COLUMN_ID => false,
+        ST_VIEW_SUB_ID => false,
+        ST_VIEW_ARG_ID => false,
+        TableId(17..ST_RESERVED_SEQUENCE_RANGE) => {
```
Centril: Could we replace the magic number with a computed `const` item, or use a half-open range?

```suggestion
        TableId(..ST_RESERVED_SEQUENCE_RANGE) => {
```

gefjon:
This PR appears to not (yet) have fully fixed the issue: I get a unique constraint violation after replay on `st_sequence_sequence_id` for `[U32(7), String("st_view_arg_id_seq"), U32(16), U16(0), I128(Packed(1)), I128(Packed(4097)), I128(Packed(1)), I128(Packed(170141183460469231731687303715884105727)), I128(Packed(4097))]`.

crates/datastore/src/locking_tx_datastore/committed_state.rs
```diff
@@ -414,13 +414,24 @@ impl CommittedState {
...
+                    // has system table schemas which do not match our expectations,
+                    // which is almost certainly an unrecoverable error.
+                    return Ok(());
+                } else {
+                    return Err(TableError::Duplicate(e).into());
+                }
             }
-        })?;
+            Err(InsertError::Bflatn(e)) => return Err(TableError::Bflatn(e).into()),
+            Err(InsertError::IndexError(e)) => return Err(IndexError::UniqueConstraintViolation(e).into()),
```
Shubham8287, line 564: I believe we need to handle/ignore `UniqueConstraintViolation` here for system tables to make replay work with duplicate insert in `st_view_arg_id_seq`.


this may work but I also think instead of handling conflicts/duplicates on commitlog replay, we should fix it at layer above, something like -  call `Locking::bootstrap` only for  new databases and enforce replays to have atleast a single snapshot (Which could be snapshot 0 as well) Or basically have some sort of versioning with system tables, using snapshot 0 is one way to do that.

gefjon: `UniqueConstraintViolation` here is unreachable, as indexes don't exist during replay. We build them later, in `Locking::rebuild_state_after_replay`.

gefjon: Moving to a design where we always replay from a snapshot is something we could do, but is a larger undertaking which will require design work. In particular, it will require a higher level of assurance that our snapshots are correct and correspond to the expected state of the commitlog than we currently have, as well as certain atomicity guarantees.

The goal of this PR is to fix a couple of databases which are broken in the wild. We cannot block doing so on a more fundamental rework of our commitlog, snapshot and replay system.

gefjon:
Rebased onto master, as the previous base branch has merged in #3887 .

crates/datastore/src/system_tables.rs
joshua-spacetime: I don't think that's a near term concern.

crates/datastore/src/locking_tx_datastore/committed_state.rs
```diff
@@ -423,19 +511,31 @@ impl CommittedState {
...
-        let (_, row_ref) = table.insert(pool, blob_store, row).map_err(|e| -> DatastoreError {
-            match e {
-                InsertError::Bflatn(e) => TableError::Bflatn(e).into(),
-                InsertError::Duplicate(e) => TableError::Duplicate(e).into(),
-                InsertError::IndexError(e) => IndexError::UniqueConstraintViolation(e).into(),
+
+        let (_, row_ref) = match table.insert(pool, blob_store, row) {
+            Ok(stuff) => stuff,
+            Err(InsertError::Duplicate(e)) => {
+                if is_built_in_meta_row(table_id, row)? {
```
joshua-spacetime, line 518: Is a duplicate insert a problem for any table?

gefjon: It would be odd and would represent an invariant-breaking commitlog, so I wouldn't want to loosen this check unnecessarily. The duplicate insert itself wouldn't cause any problems, but if we saw it in the wild for a case other than `migrate_system_tables`, it would be concerning enough that I'd want to know and to look into it.

jdetter:
I did nothing differently here to get it to merge 😅

PR #3892, opened 2025-12-16
Don't error during replay when deleting a table with no in-memory instance

gefjon:
# Description of Changes

Based on #3891 . Begin reviewing from commit 4370b69cb.

Prior to this commit, we treated it as an error (in fact a panic, prior to an earlier commit of mine today) when deleting a table during replay which didn't have an in-memory instace. This should not have been an error, as it's not problematic, and it can occur in valid commitlogs when a table is initially empty and is never used prior to its deletion.

With this commit, don't error.

N/a

1

# Testing

- [x] Manually replayed a database with such a deleted table. Got an error prior to this commit, succeeded replay with this commit.


gefjon:
Rebased onto updated  #3891 .

PR #3893, opened 2025-12-16
Replay: handle multiple column types in the same table changing

gefjon:
# Description of Changes

Based on #3892 ; begin reviewing at 21b3d1d5f.

I'll expand on this description later.

N/a.

3 - we already rewrote this path once or twice before and thought we'd fixed it.

# Testing

- [x] Manually replayed commitlog which contained an automigration that changed multiple column types from the same table.

### DIFF
--- a/crates/datastore/src/locking_tx_datastore/committed_state.rs
+++ b/crates/datastore/src/locking_tx_datastore/committed_state.rs
@@ -619,10 +619,8 @@ impl CommittedState {
         // `Self::ignore_previous_version_of_column` has marked the old version as ignored,
         // so filter only the non-ignored columns.
         let mut columns = iter_st_column_for_table(self, &table_id.into())?
-            .filter_map(|row_ref| {
-                (!self.replay_columns_to_ignore.contains(&row_ref.pointer()))
-                    .then(|| StColumnRow::try_from(row_ref).map(Into::into))
-            })
+            .filter(|row_ref| self.replay_columns_to_ignore.contains(&row_ref.pointer()))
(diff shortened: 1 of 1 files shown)


crates/datastore/src/locking_tx_datastore/committed_state.rs
```diff
@@ -86,6 +86,22 @@ pub struct CommittedState {
...
+    /// This means that, when multiple columns of a table change type within the same transaction,
+    /// we see all of the newly-inserted `st_column` rows first, and then later, all of the deleted rows.
+    /// We may even see inserts into the altered table before seeing the `st_column` deletes!
+    ///
+    /// In order to maintain a proper view of the schema of tables during replay,
+    /// we must remember the old versions of the `st_column` rows when we insert the new ones,
+    /// so that we can respect only the new versions.
+    ///
+    /// We insert into this set during [`Self::replay_insert`] of `st_column` rows
+    /// and delete from it during [`Self::replay_delete`] of `st_column` rows.
```
kim, line 108: So the set should be empty after each replayed transaction, right?
Wondering if we should assert this in `visit_tx_end`, according to the `ErrorBehavior`.

kim (approved):

crates/datastore/src/locking_tx_datastore/committed_state.rs
gefjon: It should, yes. That does seem like a sensible assertion to add.

gefjon:
Rebased onto updated #3892 .

PR #3884, opened 2025-12-16
Ensure all-ephemeral transactions don't consume a tx offset

kim:
Views are materialized in mutable transactions, but should not increment the transaction offset maintained in the committed state.

This fixes storing completely empty transactions in the commitlog, and maintains that the committed state tx offset is in-sync with the commitlog's tx offset.

# Expected complexity level and risk

2

# Testing

Added a test.

PR #3894, opened 2025-12-16
Make subscribe durable

joshua-spacetime:
# Description of Changes

With the addition of module-defined views, subscriptions are no longer read-only as they may invoke view materialization.

The way this works is that a subscription starts off as a mutable transaction, materializes views if necessary, and then downgrades to a read-only transaction to evaluate the subscription.

Before this patch, we were calling `commit_downgrade` directly on the `MutTxId` in order to downgrade the transaction. This would update the in-memory `CommittedState`, but it wouldn't make the transaction durable.

This would result in us incrementing the transaction offset of the in-memory `CommittedState` without writing anything to the commitlog. This in turn would invalidate snapshots as they would be pointing further ahead into the commitlog than they should, and so when replaying from a snapshot we would potentially skip over commits that were not included in the snapshot.

This patch changes those call sites to use `RelationalDB::commit_tx_downgrade` which both updates the in-memory state **and** makes the transaction durable.

**NOTE:** The fact that views are materialized is purely an implementation detail at this point in time. And technically view tables are ephemeral meaning they are not persisted to the commitlog. So the real bug here was that we were updating the tx offset of the in-memory committed state at all. This is technically fixed by https://github.com/clockworklabs/SpacetimeDB/pull/3884 and so after https://github.com/clockworklabs/SpacetimeDB/pull/3884 lands this change becomes a no-op. However, we still shouldn't be calling `commit` and `commit_downgrade` directly on a `MutTxId` since in most cases it is wrong to bypass the durability layer. And without this change, the bug would still be present were view tables not ephemeral, which they may not be at some point in the future.

None

1. The change itself is trivial, the bug is not.

# Testing

Adding an automated test for this is not so straightforward. First it's view related which means we don't have many options apart from a smoke test, but I don't believe the smoke tests have a mechanism for replaying the commitlog.

If transaction offsets are supposed to be linear, without any gaps, then it would be useful to assert that on each append, in which case we could write a smoke test that would fail as soon as the offsets diverged.

### DIFF
--- a/crates/datastore/src/locking_tx_datastore/mut_tx.rs
+++ b/crates/datastore/src/locking_tx_datastore/mut_tx.rs
@@ -1902,6 +1902,11 @@ impl MutTxId {
     /// Commits this transaction in memory, applying its changes to the committed state.
     /// This doesn't handle the persistence layer at all.
     ///
+    /// IMPORTANT: This method updates the in-memory state of the database but does not make it durable.
+    /// That is, the tx will not be persisted to the commitlog.
+    /// Hence you should be careful when calling this method directly.
+    /// In most cases you'll want to use `RelationalDB::commit_tx` which makes the tx durable.
+    ///
     /// Returns:
     /// - [`TxData`], the set of inserts and deletes performed by this transaction.
     /// - [`TxMetrics`], various measurements of the work performed by this transaction.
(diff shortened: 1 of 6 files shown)


crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -1947,6 +1952,11 @@ impl MutTxId {
     /// The lock on the committed state is converted into a read lock,
     /// and returned as a new read-only transaction.
     ///
+    /// IMPORTANT: This method updates the in-memory state of the database but does not make it durable.
+    /// That is, the tx will not be persisted to the commitlog.
+    /// Hence you should be careful when calling this method directly.
+    /// In most cases you'll want to use `RelationalDB::commit_tx_downgrade` which makes the tx durable.
+    ///
```
Shubham8287, line 1959: Can't mention exact line as its not in diff but just below, can you make `pub fn commit_downgrade`  to be `pub(super)` (like ` fn commit`)? That would have prevented this bug.

kim (approved):

crates/datastore/src/locking_tx_datastore/mut_tx.rs
egormanga: @Shubham8287 FYI: you can use the expand up/down buttons on the diff to comment on lines that go above or below the visible context.

PR #3896, opened 2025-12-17
client-api: Pause time in websocket timeout tests

kim:
Using `#[tokio::test(start_paused = true)]` pauses time, yet tokio will still advance it when encountering `sleep`s while it has no other work to do.

This makes the tests that rely on timeouts deterministic and should prevent those tests from becoming flaky on busy machines.

2

# Testing

This modifies tests.
It does appear to work as described, but it can't hurt if the reviewers convince themselves that it does indeed.


bfops:
I trust Phoebe's review. I wouldn't know what to look for.

PR #3877, opened 2025-12-12
Refactor /docs into Concepts

JasonAtClockwork:
# Description of Changes

Refactoring all sections to match a concept based approach to the docs.

Closes: #3840 

- Modules - Closes: #3841 
- Databases - Closes: #3842 
- Tables - Closes: #3843 
- Reducers - Closes: #3844 
- Procedures - Closes: #3845
- Views - Closes: #3846 
- Codegen - Closes: #3847 
- SDK API - Closes: #3848 
- Subscriptions - Closes: #3849 - No work done here just moved the current order
- Auth - Closes: #3850  - No work done here just moved the current order

# API and ABI breaking changes

N/A

# Expected complexity level and risk

1

# Testing

- [x] Walking through all code references before pushing

PR #3902, opened 2025-12-18
Remove old documentation that was mistakenly left

JasonAtClockwork:
# Description of Changes

With the refactor docs I had left the original module folder and mistakenly didn't remove it as requested as per Tyler's comment: https://github.com/clockworklabs/SpacetimeDB/pull/3877#pullrequestreview-3585291570:
> Also please remove the "old files" before merging.

N/A

1

# Testing

Removal of documents there is no test to run


rekhoff (approved):
All file being removed are marked as old in their slugs and labels, which matches the targeted files being requested for removal. Approved!

PR #3906, opened 2025-12-18
Bump typescript package sizes again

jdetter:
# Description of Changes

- Just bumping typescript size limits again

No

0 - this is just a release change.

# Testing

I have not tested this but I've verified the new limit is greater than the package size.


bfops (approved):
32kb seems fine to me :shrug:

PR #3908, opened 2025-12-18
Removes C# regression test of TimestampCapabilities

rekhoff:
# Description of Changes

Removes the `TimestampCapabilities` test from the C# SDK's regression test suite, due to creating inconsistent failures in automated CI testing.

Not API/ABI breaking

1

# Testing

- [X] Ran locally, no failures in building or running C# regression tests


JasonAtClockwork (approved):
LGTM - simple removal of a test that is giving us trouble.

PR #3910, opened 2025-12-19
feat: update pgwire to 0.37

sunng87:
# Description of Changes

Update pgwire to 0.37

Make `DataRowEncoder` reusable as enabled in this release, make introduce 3x performance on encoding.

N/A

N/A

# Testing

- [ ] 
- [ ]


bfops:

bfops:
I think this is failing tests due to some changes in serialization, which would be a breaking change for us.

sunng87:
@bfops Yes, let me look into the error when I get time.

sunng87:
@bfops this change is related to a correction of how we encode `bytea_array` in pgwire. The are additional bytes required for `bytea_array` than `bytea`. However, when I check the usage of `bytea_array` in spacetimedb, I think we should correct to `bytea` because `bytea` is already byte array, while `bytea_array` is array of byte arrays. Changing the data type will fix the encoding issue.

crates/pg/src/encoder.rs
```diff
@@ -57,7 +57,7 @@ pub(crate) fn type_of(schema: &ProductType, ty: &ProductTypeElement) -> Type {
             _ => Type::ANYARRAY,
         },
         AlgebraicType::Product(_) => match format {
-            PsqlPrintFmt::Hex => Type::BYTEA_ARRAY,
+            PsqlPrintFmt::Hex => Type::BYTEA,
```
bfops, line 60: I want to double-check with @joshua-spacetime his thoughts about this before merging

joshua-spacetime: I agree with @sunng87. `BYTEA` is the correct type.

PR #3912, opened 2025-12-19
Create `db.lock` file only for persistent databases

kim:
This is the first step to make in-memory only databases not touch the disk at all. Pending is an in-memory only sink for module logs.

Responsibility for the lock file is transferred to `Durability`, which means that only persistent databases opened for writing acquire the lock.

As a consequence, the `Durability` trait gains a `close` method that prevents further writes and drains the internal buffers, even when multiple `Arc`-pointers to the `Durability` exist.

2

# Testing

Covered by existing tests.

### DIFF
--- a/tools/xtask-llm-benchmark/src/bench/results_merge.rs
+++ b/tools/xtask-llm-benchmark/src/bench/results_merge.rs
@@ -44,6 +44,9 @@ pub fn ensure_lang<'a>(root: &'a mut Results, lang: &str) -> &'a mut LangEntry {
 
 fn ensure_mode<'a>(lang_v: &'a mut LangEntry, mode: &str, hash: Option<String>) -> &'a mut ModeEntry {
     if let Some(i) = lang_v.modes.iter().position(|m| m.mode == mode) {
+        if let Some(h) = hash {
+            lang_v.modes[i].hash = Some(h);
+        }
         return &mut lang_v.modes[i];
     }
     lang_v.modes.push(ModeEntry {


crates/durability/src/imp/local.rs
```diff
@@ -73,48 +84,55 @@ pub struct Local<T> {
...
         rt: tokio::runtime::Handle,
         opts: Options,
         on_new_segment: Option<Arc<OnNewSegmentFn>>,
-    ) -> io::Result<Self> {
+    ) -> Result<Self, OpenError> {
         info!("open local durability");
 
-        let clog = Arc::new(Commitlog::open(root, opts.commitlog, on_new_segment)?);
-        let (queue, rx) = mpsc::unbounded_channel();
+        // For backwards-compatibility, we keep using the `db.lock` file.
```
gefjon: Could you extend this comment to say what we'd do if backcompat was not a concern?

crates/durability/src/imp/local.rs
```diff
@@ -149,91 +174,134 @@ impl<T: Encode + Send + Sync + 'static> Local<T> {
...
+
+        let mut sync_interval = interval(self.sync_interval);
+        sync_interval.set_missed_tick_behavior(MissedTickBehavior::Delay);
+        // `flush_and_sync` when the loop exits without panicking,
+        // or `flush_and_sync` inside the loop failed.
+        let mut sync_on_exit = true;
+
+        loop {
+            tokio::select! {
+                biased;
```
gefjon, line 223: I think biased select is weird enough that I'd like to see a comment justifying it.

PR #3915, opened 2025-12-19
Reorganize TS sdk

coolreader18:
# Description of Changes

This moves a bunch of stuff from `lib` back into `server` and `sdk`, and removes all but one global variable from the server sdk in preparation for export-based reducer definition.

2 - a pretty big refactor, but it's mostly just code movement.

# Testing

- [x] Refactor, so automated tests are sufficient.


cloutiertyler (approved):
I went over this with Noa and this aligns with our goals. We should do additional testing of the quickstart once this merges.

PR #3918, opened 2025-12-19
[TS] Implement and use point scan ABI

coolreader18:
# Description of Changes

TS equivalent of #3863. I also did optimized single-column indices such that there's a special case that avoids branching.

1 - straightforward and there's Rust precedent to draw from.

# Testing

- [x] Automated testing is sufficient

### DIFF
--- a/crates/bindings-typescript/src/server/runtime.ts
+++ b/crates/bindings-typescript/src/server/runtime.ts
@@ -560,6 +560,7 @@ function makeTableView(
 
     let index: Index<any, any>;
     if (isUnique && serializeSinglePoint) {
+      // numColumns == 1, unique index
       index = {
         find: (colVal: IndexVal<any, any>): RowType<any> | null => {
           const point = serializeSinglePoint(colVal);
@@ -596,11 +597,12 @@ function makeTableView(
         },
       } as UniqueIndex<any, any>;
     } else if (isUnique) {
+      // numColumns != 1, unique index
       index = {
         find: (colVal: IndexVal<any, any>): RowType<any> | null => {
-          if (colVal.length !== numColumns)
+          if (colVal.length !== numColumns) {
             throw new TypeError('wrong number of elements');
-
+          }
           const point = serializePoint(colVal);
           const iter = tableIterator(
             sys.datastore_index_scan_point_bsatn(index_id, point),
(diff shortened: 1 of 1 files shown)


crates/bindings-typescript/src/server/runtime.ts
```diff
@@ -516,33 +532,51 @@ function makeTableView(
...
+          return row;
+        },
+      } as UniqueIndex<any, any>;
+    } else if (isUnique) {
       index = {
         find: (colVal: IndexVal<any, any>): RowType<any> | null => {
-          if (numColumns === 1) colVal = [colVal];
-          const args = serializeBound(colVal);
+          if (colVal.length !== numColumns)
+            throw new TypeError('wrong number of elements');
```
cloutiertyler, line 604: NIT: Can we add {} around here.

crates/bindings-typescript/src/server/runtime.ts
```diff
@@ -499,15 +500,30 @@ function makeTableView(
...
-      if (prefix_elems > numColumns - 1)
-        throw new TypeError('too many elements in prefix');
       for (let i = 0; i < prefix_elems; i++) {
         const elemType = indexType.value.elements[i].algebraicType;
         AlgebraicType.serializeValue(writer, elemType, prefix[i], typespace);
       }
       return writer;
     };
 
+    const serializePoint = (colVal: any[]): Uint8Array => {
```
cloutiertyler, line 537: Can we not constrain the type any further than any in this case?

crates/bindings-typescript/src/server/runtime.ts
```diff
@@ -516,33 +532,51 @@ function makeTableView(
...
+      } as UniqueIndex<any, any>;
+    } else if (isUnique) {
       index = {
         find: (colVal: IndexVal<any, any>): RowType<any> | null => {
-          if (numColumns === 1) colVal = [colVal];
-          const args = serializeBound(colVal);
+          if (colVal.length !== numColumns)
+            throw new TypeError('wrong number of elements');
+
+          const point = serializePoint(colVal);
```
cloutiertyler, line 606: The array conversion from before is missing.

```ts
          if (numColumns === 1) colVal = [colVal];
```

No?

crates/bindings-typescript/src/server/runtime.ts
coolreader18: Not really no; we use `any` all over this function.

crates/bindings-typescript/src/server/runtime.ts
coolreader18: No, because this is the case where `numColumns !== 1`

PR #3921, opened 2025-12-19
CI - Hackily fix V8 linker errors

bfops:
# Description of Changes

Introduce a hacky workaround to our `csharp-testsuite` to address missing `librusty_v8.a`: manually check for that file and manually build the package if it's missing.

CI-only change

1

# Testing

- [x] Locally tested removing the `librusty_v8.a` and then running `cargo clean -p v8 && cargo build -p v8`, and this does seem to repair it.
- [x] The CI has run with a cache that is "broken", but successfully passes `csharp-testsuite`


jdetter (approved):
Alright let's give it a try 🤷

PR #3922, opened 2025-12-19
Bump TS package size limits

bfops:
# Description of Changes

Bump all package size limits to 500 kB after a discussion about repeatedly triggering them.

None.

1

# Testing

None


bfops:
afaik they have already been doubled, so closing this

PR #3930, opened 2025-12-23
Fixed inverted `wasm-opt` condition

egormanga:
# Description of Changes

Reverted the incorrect inversion of `build_debug` condition introduced by https://github.com/clockworklabs/SpacetimeDB/commit/bb432132457b023d98a379beea2a52d76d3bd8d6 that resulted in only debug builds being optimized instead of only release ones, as the comment just below says.

None.

### 1.

# Testing

- [x] Manual testing
- [x] Personal experience (saw `.opt.wasm` files disappear at some point)


egormanga:
@bfops — since I believe you're the one triaging, please notice this as it's probably an important performance fix.

PR #3932, opened 2025-12-27
Use table def accessorName instead of server table name to access table db view

aceysmith:
# Description of Changes

Update the `useTable` hook in the `spacetimedb/react` package to use the client language convention aware table accessor key to lookup the correct table

None

1

# Testing

I built and ran a client using this code change before and after. Previously with multi-word table names in rust module (i.e. `crew_assignments` table), the useTable hook would fail to lookup the table information and hookup the onInsert/onDelete/onUpdate callbacks. With this change, they successfully connect and data returned by the useTable hook now flows.

All of my testing was with module_bindings generated with the 1.11.0 rust module crate. Additional testing for backwards compatibility might be useful. I'm not sure what the clockwork labs target is for that sort of thing.


kistz (commented):
After testing I can confirm that this patch works as expected and my multi word table and views names show the desired output :)

PR #3935, opened 2025-12-30
Implemented `RETURNING` support for SQL

egormanga:
# Description of Changes

SQL `RETURNING` for DML to return affected rows to the caller:
```sql
INSERT INTO test (a, b)
VALUES (1, 2)
RETURNING *;
```
gives:
```
 a | b 
---+---
 1 | 2 
```
(same goes for `UPDATE` & `DELETE`).

Individual columns can be specified, too — separately from the fields the statement operates on:
```sql
UPDATE test
SET a = 1
WHERE b < 2
RETURNING b;
```
allowing to efficiently combine DML with implicit `SELECT`s.

None.

### 2.

# Testing

- [x] Manual testing
- [ ] Unit tests


bfops:
Thank you for filing this! We'll work on getting a reviewer on it.

cloutiertyler:
Going to get someone to review this.

crates/core/src/sql/execute.rs
joshua-spacetime, line 188:
```suggestion
                            rows,
```
This is the other possible return branch.

crates/expr/src/statement.rs
```diff
@@ -153,16 +170,30 @@ pub fn type_insert(insert: SqlInsert, tx: &impl SchemaView) -> TypingResult<Tabl
...
-    let into = schema;
+    let into = schema.clone();
     let rows = rows.into_boxed_slice();
-    Ok(TableInsert { table: into, rows })
+    let mut vars = Relvars::default();
+    vars.insert(table_name.clone(), into.clone());
+    let returning = returning
+        .map(|proj| type_proj(RelExpr::RelVar(Relvar {
+            schema: schema,
+            alias: ST_VAR_NAME.into(),
```
joshua-spacetime, line 180: This alias should be the insert table.

crates/expr/src/statement.rs
```diff
@@ -184,9 +215,17 @@ pub fn type_delete(delete: SqlDelete, tx: &impl SchemaView) -> TypingResult<Tabl
     let expr = filter
         .map(|expr| type_expr(&vars, expr, Some(&AlgebraicType::Bool)))
         .transpose()?;
+    let returning = returning
+        .map(|proj| type_proj(RelExpr::RelVar(Relvar {
+            schema: from.clone(),
+            alias: ST_VAR_NAME.into(),
```
joshua-spacetime, line 221: This should be the delete table.

crates/expr/src/statement.rs
```diff
@@ -249,10 +289,18 @@ pub fn type_update(update: SqlUpdate, tx: &impl SchemaView) -> TypingResult<Tabl
     let filter = filter
         .map(|expr| type_expr(&vars, expr, Some(&AlgebraicType::Bool)))
         .transpose()?;
+    let returning = returning
+        .map(|proj| type_proj(RelExpr::RelVar(Relvar {
+            schema: schema.clone(),
+            alias: ST_VAR_NAME.into(),
```
joshua-spacetime, line 295: Similarly this should be the update table.

crates/sql-parser/src/parser/sql.rs
```diff
@@ -174,12 +174,13 @@ fn parse_statement(stmt: Statement) -> SqlParseResult<SqlAst> {
...
             table: false,
             on: None,
-            returning: None,
+            returning,
             ..
         } if after_columns.is_empty() => Ok(SqlAst::Insert(SqlInsert {
             table: parse_ident(table_name)?,
             fields: columns.into_iter().map(SqlIdent::from).collect(),
             values: parse_values(*source)?,
+            returning: returning.map(parse_projection).transpose()?,
```
joshua-spacetime, line 183: I believe `parse_projection` allows for aggregate expressions like `COUNT`, not valid in `RETURNING`.

joshua-spacetime (commented):
I think this is a good feature to add. I'd just like to see some tests. In particular I think the following tests would either fail or wrongly succeed today:
```
UPDATE a SET ... RETURNING a.*;
UPDATE a SET ... RETURNING count(*) AS n;
```

PR #3936, opened 2025-12-30
Return Options as nullable values over PG Wire

egormanga:
# Description of Changes

Improves usability compared to regular DBs.

* #### Before:
  ```sql
   id | test         
  ----+--------------
    1 | {"some":"x"} 
    2 | {"none":{}}  
  ```
* #### After:
  ```sql
   id | test 
  ----+------
    1 | "x"  
    2 |      
  ```

## Note:
The current implementation returns an empty string (2 bytes: `""`) on the wire. A correct one would return actual 0-byte `NULL`s, but this seemingly requires an elaborate refactoring of the current conversion implementation to support passing `Option<AlgebraicValue>` straight to the encoder, without converting it to a string first.

Breaks the PG Wire protocol to move it closer to standard.

### 3:
Will probably have to hear from `pgwire`'s author if that's correct.

# Testing

- [x] Manual testing
- [ ] Unit tests

### DIFF
--- a/crates/pg/src/encoder.rs
+++ b/crates/pg/src/encoder.rs
@@ -136,8 +136,15 @@ impl TypedWriter for PsqlFormatter<'_> {
         name: Option<&str>,
         value: ValueWithType<AlgebraicValue>,
     ) -> Result<(), Self::Error> {
-        // Is a simple enum?
         if let AlgebraicType::Sum(sum) = &ty.field.algebraic_type {
+            if sum.is_option() {
+                if *value.value() == AlgebraicValue::unit() {
+                    self.write("")?;
+                } else {
+                    self.write(satn::PsqlWrapper { ty, value })?;
+                }
+                return Ok(());
+            }
             if sum.is_simple_enum() {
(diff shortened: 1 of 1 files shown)


joshua-spacetime:
@egormanga this seems reasonable to me, although I'd like to first design how we are going to handle/represent options and more generally sum types in Spacetime SQL before committing to this representation in pgwire.

egormanga:
I believe, an Option should be just a nullable value, which fits nicely with outer join return schemas (e.g. #2892).

General sum types are a different beast. As long as we can deterministically parse them (e.g. `u32 | str`, but not `u32 | i64`), they can probably go untagged, like the usual SQL. If we do need the tags, then the current JSON-like repr seems to be the only possible option.

cloutiertyler:
@egormanga I have a similar question on this one as I had on #3607. Namely, what do you do about `Option<Option<u32>>`. Could you please update the description with a plan for what to do there?

egormanga:
So consequently, the answer here becomes similar to what I answered there.

PR #3937, opened 2025-12-30
Respect updates to `st_table` during replay

gefjon:
# Description of Changes

Prior to this commit, we had special handling for deletes from `st_table` during replay, but we did not pair them with inserts to form updates, instead only treating the delete as a dropped table.

With this commit, we record inserts to `st_table` which will form update pairs during replay, and handle them appropriately, updating the table's schema and not dropping the table.

There is a tricky case where a table exists but is empty, and then within a single transaction:
- The table undergoes a migration s.t. its `table_access` or `primary_key` changes.
- At least one row is inserted into the table.

In this case, the in-memory table structure will not exist at the point of the `st_table` insert, but will be created before the corresponding `st_table` delete, meaning there will be two conflicting `st_table` rows resident. To handle this case, `CommittedState` tracks a side table, `replay_table_updated`, which stores a `RowPointer` to the correct most-recent `st_table` row for the migrating table.

I've also renamed the one previously-extant replay-only side table, `table_dropped`, to include the `replay_` prefix, which IMO improves clarity. And I've made it so `replay_table_dropped` is cleared at the end of each transaction, as the previous behavior of continuing to ignore a table that should be unreachable masked errors which would have been helpful when debugging this issue.

This PR also includes an extended error message when encountering a unique constraint violation while replaying, which I found helpful while debugging.

N/a

2 - replay is complicated and scary, but this PR isn't gonna make things *more* broken than they already were.

# Testing

- [x] Manually replayed a commitlog which included a migration that altered a table's `table_access`, which was previously broken but now replays successfully.

### DIFF
--- a/crates/datastore/src/locking_tx_datastore/committed_state.rs
+++ b/crates/datastore/src/locking_tx_datastore/committed_state.rs
@@ -587,7 +587,6 @@ impl CommittedState {
                 // they are simply ignored in (1).
                 self.replay_table_dropped.insert(dropped_table_id);
             }
-            // TODO: inspect `self.replay_table_updated` and handle appropriately.
         }
 
         if table_id == ST_COLUMN_ID {


crates/datastore/src/locking_tx_datastore/committed_state.rs
```diff
@@ -106,7 +117,26 @@ pub struct CommittedState {
...
+    /// When looking up the `st_table` row for a table, if it has an entry here,
+    /// that means there are two rows resident in `st_table` at this point in replay.
+    /// We return the row recorded here rather than inspecting `st_table`.
+    ///
+    /// We remove from this set when we reach the matching delete,
+    /// and assert this set is empty at the end of each transaction.
+    ///
+    /// [`RowPointer`]s from this set are passed to the `unsafe` [`Table::get_row_ref_unchecked`],
+    /// so it's important to properly maintain only [`RowPointer`]s to valid, extant, non-deleted rows.
+    replay_table_updated: IntMap<TableId, RowPointer>,
```
cloutiertyler, line 139: In the future, a proper update operation could help us simplify quite a bit. Especially if we had row keys (or used primary keys, with a secret primary key, if none is provided).

crates/datastore/src/locking_tx_datastore/committed_state.rs
```diff
@@ -508,18 +559,35 @@ impl CommittedState {
...
+                // This can happen if a table is initially empty at the snapshot or its creation,
+                // and never has any rows inserted into or deleted from it.
+                self.tables.remove(&dropped_table_id);
+
+                // Mark the table as dropped so that when
+                // processing row deletions for that table later,
+                // they are simply ignored in (1).
+                self.replay_table_dropped.insert(dropped_table_id);
+            }
+            // TODO: inspect `self.replay_table_updated` and handle appropriately.
```
cloutiertyler: I don't follow what we would need to do here. Also why here and why not on 575.

gefjon: Oh, I meant to remove this comment. I had added it during development when I stepped away for a bit but didn't want to forget what I was working on. Will delete now.

PR #3938, opened 2025-12-30
fix view deadlock

Shubham8287:
# Description of Changes
Fixes a deadlock in the subscription code and HTTP SQL handler that was caused by calling view methods on the module while holding the transaction lock.

I tried a couple of approaches to make the closures `Send` for all code paths that need to hold the transaction while working with views, but that didn’t work out well. The V8 module communicates with the host through channels, which would require dynamic dispatch.

In the current approach, all existing methods that were calling views from the host are now invoked from inside the module itself. In future, It will be better to move these methods to common place rather than being scattrered.

### DIFF
--- a/crates/core/src/subscription/module_subscription_actor.rs
+++ b/crates/core/src/subscription/module_subscription_actor.rs
@@ -1259,6 +1259,7 @@ impl ModuleSubscriptions {
 
     /// Materialize the views returned by the `view_collector`, if not already materialized,
     /// and subsequently downgrade to a read-only transaction.
+    #[allow(clippy::type_complexity)]
     fn materialize_views_and_downgrade_tx<I: WasmInstance>(
         &self,
         mut tx: MutTxId,


crates/core/src/sql/execute.rs
```diff
@@ -187,31 +191,47 @@ pub struct SqlResult {
...
     head: &mut Vec<(Box<str>, AlgebraicType)>,
 ) -> Result<SqlResult, DBError> {
+    match module {
+        Some(module) => {
+            let info = module.info.clone();
+            module.call_view_sql(info, db, sql_text, auth, subs).await
+        }
+        None => run_from_module::<crate::host::wasmtime::WasmtimeInstance>(None, db, sql_text, auth, subs, head)
+            .map(|x| x.0),
+    }
```
gefjon, line 207: Could you add some commentary about when `module` will be `Some` or `None`, and what each of these cases means for SQL execution? Reading this code, going from "there is no module" to `run_in_module` feels like nonsense.

crates/core/src/sql/execute.rs
```diff
@@ -187,31 +191,47 @@ pub struct SqlResult {
...
+}
+
+pub fn run_from_module<I: WasmInstance>(
```
gefjon, line 210: Would be nice to have some commentary here about how & why this must not be `async`. Also, a comment about where this function is expected to run (on the database core? On the global executor, and it schedules something to the database core internally?) would be appreciated.

crates/core/src/sql/execute.rs
```diff
@@ -187,31 +191,47 @@ pub struct SqlResult {
...
+    instance: Option<(&mut RefInstance<I>, &ModuleDef)>,
+    db: Arc<RelationalDB>,
+    sql_text: String,
+    auth: AuthCtx,
+    subs: Option<ModuleSubscriptions>,
+    head: &mut Vec<(Box<str>, AlgebraicType)>,
+) -> Result<(SqlResult, bool), DBError> {
     // We parse the sql statement in a mutable transaction.
     // If it turns out to be a query, we downgrade the tx.
     let (tx, stmt) = db.with_auto_rollback(db.begin_mut_tx(IsolationLevel::Serializable, Workload::Sql), |tx| {
```
gefjon, line 220: Depending on where this function is running, it may not be correct to `begin_mut_tx` in this function. If we're running on a database core, and we try to run this while another mutable transaction is ongoing (e.g. in a procedure), this will cause a deadlock.

gefjon: It looks like `client_api` reaches this through `using_database`, meaning I think we do run this on the database core. I think that means we have to `asyncify` this call. (Really we should just be using async locks in the datastore, but we're not.)

crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -440,8 +458,6 @@ impl ModuleSubscriptions {
         Ok((update, metrics))
     }
 
-    /// Add a subscription to a single query.
-    #[tracing::instrument(level = "trace", skip_all)]
```
gefjon, line 444: Where did this go?

crates/core/src/sql/execute.rs
Shubham8287: I promise you to raise next PR to add commentary and make code more readable, but will prefer to merge this one right now so I don't want to wait for CIs again :)

PR #3945, opened 2026-01-01
Render arrays correctly in SQL

mamcx:
# Description of Changes

As the title says.

Before arrays incorrectly display and only for the first item.

1

# Testing

- [x] Smoke and unit tests
- [] No inspected in another client outside `psql`

### DIFF
--- a/crates/sats/src/ser.rs
+++ b/crates/sats/src/ser.rs
@@ -6,7 +6,10 @@ mod impls;
 pub mod serde;
 
 use crate::de::DeserializeSeed;
-use crate::{algebraic_value::ser::ValueSerializer, bsatn, buffer::BufWriter, ProductValue, SumValue, ValueWithType};
+use crate::{
+    algebraic_value::ser::ValueSerializer, bsatn, buffer::BufWriter, AlgebraicType, ArrayValue, ProductValue, SumValue,
+    ValueWithType,
+};
 use crate::{AlgebraicValue, WithTypespace};
 use core::marker::PhantomData;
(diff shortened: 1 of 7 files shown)


crates/sats/src/ser.rs
```diff
@@ -150,6 +153,40 @@ pub trait Serializer: Sized {
         value: &T,
     ) -> Result<Self::Ok, Self::Error>;
 
+    fn serialize_array_raw(self, value: &ValueWithType<'_, ArrayValue>) -> Result<Self::Ok, Self::Error> {
+        let mut ty = &*value.ty().elem_ty;
+        loop {
+            // We're doing this because of `Ref`s.
+            break match (value.value(), ty) {
```
egormanga, line 160: Do we have to replicate this code in two to three places? Maybe at least add comments pointing to other places then.

But I think it can be implemented in a common trait.

smoketests/tests/sql.py
```diff
@@ -94,6 +109,17 @@ class SqlFormat(Smoketest):
     };
     ctx.db.t_others().insert(tuple.clone());
     ctx.db.t_others_tuple().insert(TOthersTuple { tuple });
+
```
egormanga, line 112: nit: there's an unstripped indentation left here

egormanga:
> Not inspected in another client outside `psql`

Still doesn't work correctly in `spacetime sql` (no changes in output).

egormanga:
* Fixes #4012

cloutiertyler:
Needs to be reimplemented because of the out of date smoketest.

PR #3948, opened 2026-01-03
Combine table and index attributes in Rust example

kistz:
# Description of Changes

The syntax is not supported like shows and must be included in the `table` macro.

None.

0 No runtime change

# Testing

- [x] Code compiles now.


cloutiertyler:
Is there any more context on this change?

kistz:
@cloutiertyler yes. index being a separate macro invocation is not supported. it must be inside the table macro. i assume this is because you can have multiple tables per struct definition so the separate index would be ambigous. :>

Example that it doesnt work: (error is that index is not found in spacetimedb)
image

egormanga:
It would actually be more reasonable to support this syntax instead.

kistz:
I dont necessarily think so because of the ambiguity mentioned above but thats a separate discussion anyway.
IMO it is worth fixing mistakes in docs immediately since it increases friction for new users and you can always add the syntax back at a later stage.
However i dont feel strongly about this pr so feel free to close. just thought this was faster than creating a issue for it :>

Issue #3914, opened 2025-12-19
View + Nullable type + .Equals(object) issue


xxhollyxx:
`.Equals (object) [0x0001b] in <fdfa2fe etc...`

```
2025-12-19T16:06:56.074166Z  INFO crates\core\src\host\wasmtime\wasmtime_module.rs:32:   Frame #298: <unknown>
2025-12-19T16:06:56.074296Z  INFO crates\core\src\host\wasmtime\wasmtime_module.rs:32:   Frame #297: <unknown>
2025-12-19T16:06:56.074499Z  INFO crates\core\src\host\wasmtime\wasmtime_module.rs:32:   Frame #296: <unknown>
2025-12-19T16:06:56.074699Z  INFO crates\core\src\host\wasmtime\wasmtime_module.rs:32:   Frame #295: <unknown>
2025-12-19T16:06:56.074918Z  INFO crates\core\src\host\wasmtime\wasmtime_module.rs:32:   Frame #294: <unknown>
2025-12-
```

Getting messages like these flooded in the console.
Appears to be related to Views.

```
[SpacetimeDB.Type]
    public partial struct DbVector2
    {
        public int X;
        public int Y;
    }

```

Marking this as nullable in the table is what caused this issue.

bfops:
Thank you for reporting this! We'll work on getting it triaged + fixed.

PR #3949, opened 2026-01-03
Fixed Nullable value-type fields in Views in C# module causing excessive logging

rekhoff:
# Description of Changes

This PR fixes a C# codegen performance/behavior issue triggered by views that include nullable value-type fields (e.g. `DbVector2?`), as reported in #3914.

* Updated the C# BSATN code generator to emit non-boxing equality for `Nullable<T>` members by using `System.Nullable.Equals(a, b)` rather than `a.Equals(b)` (which can box and cause excessive host logging/work in view diff/equality paths).

This causes code generation to change from something like:
``` csharp
// From:
var ___eqNullableIntField = this.NullableIntField.Equals(that.NullableIntField);
// To:
var ___eqNullableIntField = System.Nullable.Equals(
    this.NullableIntField,
    that.NullableIntField
);
```
* Added a regression scenario to the C# regression-test module that exercises the problematic pattern:
  * A table containing a nullable struct field (`DbVector2? Pos`).
  * A public view that returns rows containing that nullable field.
  * A reducer to mutate the nullable field from `some` → `none` → `some` to force view re-evaluation and diffing.
* Updated the regression-test client to:
  * Subscribe to the new view.
  * Validate that the view evaluates successfully and contains the expected rows.
  * Call the reducer and validate view state after updates.
  * Fail the test if view evaluation/diffing produces errors.

None.
* No changes to public SpacetimeDB schema or wire format semantics.
* Changes are limited to generated equality code and regression tests.

2 - Low
* The codegen change is small and localized (special-casing `System.Nullable<T>` equality generation).
* Risk is primarily around subtle behavior differences in equality for nullable value types; however, `System.Nullable.Equals` matches the expected semantics and avoids boxing.
* Regression tests specifically cover the nullable-struct-in-view scenario to guard against regressions.

# Testing

- [X] Ran the C# regression test suite via `run-regression-tests.sh` and confirmed no errors.

### DIFF
--- a/crates/cli/src/tasks/mod.rs
+++ b/crates/cli/src/tasks/mod.rs
@@ -27,7 +27,7 @@ pub fn build(
 
     if lang == ModuleLanguage::Javascript {
         Ok((output_path, "Js"))
-    } else if !build_debug {
+    } else if build_debug {
         Ok((output_path, "Wasm"))
     } else {
         // for release builds, optimize wasm modules with wasm-opt
--- a/crates/core/src/host/module_host.rs
+++ b/crates/core/src/host/module_host.rs
@@ -649,28 +649,24 @@ impl CallReducerParams {
 
 pub enum ViewCommand {
     AddSingleSubscription {
-        info: Arc<ModuleInfo>,
         sender: Arc<ClientConnectionSender>,
         auth: AuthCtx,
         request: SubscribeSingle,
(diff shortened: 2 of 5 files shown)


JasonAtClockwork (approved):
Solid improvement with additional testing! I ran the tests as well with no difficulties.

PR #3953, opened 2026-01-06
Misc docs and small CLI improvements

### LINKED ISSUE #3610: Misleading Cursor instruction for non-windows users
https://github.com/clockworklabs/SpacetimeDB/blob/master/docs/.cursor/rules/spacetimedb.mdc contains the instruction, "You're executing commands in Powershell..." 

That gets copied into my project when I run `spacetimedb init`, and it's confusing my copy of Cursor because I'm not using Windows.

cloutiertyler:
# Description of Changes

This PR does several small things:

1. It removes the explicit `h1` tags on every page, and either uses the side bar title directly, or puts it in the frontmatter
2. It merges what are currently called quickstarts into a single Chat App Tutorial
3. It creates new quickstarts which just use `spacetime dev --template` to get you up and running quickly
4. It adds a "The Zen of SpacetimeDB" page much like the Zen of Python which goes over the 5 key principles of SpacetimeDB
5. It reorders all Tabs groups so that the ordering is `TypeScript`, `C#`, `Rust`, `Unreal`, `C++`, `Blueprints` (order of decreasing popularity).
6. It improves the sidebar navigation by having categories act as overview pages, and also fixes the breadcrumbs
7. It fixes various small typos and issues
8. Closes #3610 and adds cursor rules files generally
9. It fixes general styling on the docs page by bring it inline with the UI design:

Old:
image

New:
image

https://github.com/user-attachments/assets/adc5a78a-ada8-45b5-8078-a45cb81477a3

This PR does NOT change any old links. It does add new pages though.

3 - it's a large change. I manually tested the TypeScript Chat App Tutorial but I have not gone through the Rust and C# quickstarts. However, we have testing on the quickstarts and this is text only so can be carefully reviewed.

# Testing

- [x] Ran through each step of the Chat App TypeScript tutorial to ensure it is working
- [x] Ran and tested the styles and the functionality of the side bar


docs/docs/00100-intro/00100-getting-started/00400-key-architecture.md
```diff
@@ -397,11 +397,11 @@ void AGameManager::OnMakeRequestComplete(const FProcedureEventContext& Context,
 </TabItem>
 <TabItem value="blueprint" label="Unreal Blueprint">
 
-An Unreal C++ [client](#client) can call a procedure defined by a Rust or TypeScript module:
+An Unreal Blueprint [client](#client) can call a procedure defined by a Rust or TypeScript module:
```
cloutiertyler:
```suggestion
An Unreal [client](#client) can call a procedure defined by a Rust or TypeScript module:
```

docs/docs/00100-intro/00100-getting-started/00400-key-architecture.md
```diff
@@ -397,11 +397,11 @@ void AGameManager::OnMakeRequestComplete(const FProcedureEventContext& Context,
...
 
 ![MakeRequest without callback](/images/unreal/intro/ue-blueprint-makerequest-nocallback.png)
 
-A Unreal C++ [client](#client) can also register a callback to run when a procedure call finishes, which will be invoked with that procedure's return value:
+An Unreal Blueprint [client](#client) can also register a callback to run when a procedure call finishes, which will be invoked with that procedure's return value:
```
cloutiertyler:
```suggestion
An Unreal [client](#client) can also register a callback to run when a procedure call finishes, which will be invoked with that procedure's return value:
```

crates/cli/src/subcommands/dev.rs
```diff
@@ -128,11 +135,16 @@ pub async fn exec(mut config: Config, args: &ArgMatches) -> Result<(), anyhow::E
...
-            let init_args = init::cli().get_matches_from(if resolved_server == "local" {
-                vec!["init", "--local"]
-            } else {
-                vec!["init"]
-            });
+            let mut init_argv = vec!["init"];
+            if resolved_server == "local" {
+                init_argv.push("--local");
+            }
+            let template = args.get_one::<String>("template");
```
jdetter, line 142: If the user passes a template via `--template` and the project already exists, should we print an error or warning that `--template` is being ignored?

crates/standalone/src/subcommands/start.rs
```diff
@@ -226,6 +280,54 @@ pub async fn exec(args: &ArgMatches, db_cores: JobCores) -> anyhow::Result<()> {
...
+/// and IPv6 (::1). If SpacetimeDB binds only to IPv4 but another service is using the
+/// same port on IPv6, browsers may connect to the wrong service depending on which
+/// address they try first.
+///
+/// Note: There is a small race condition between this check and the actual bind -
+/// another process could grab the port in between. This is unlikely in practice
+/// and the actual bind will fail with a clear error if it happens.
+fn is_port_available(port: u16) -> bool {
+    let ipv4_addr = SocketAddr::from(([127, 0, 0, 1], port));
+    let ipv6_addr = SocketAddr::V6(SocketAddrV6::new(Ipv6Addr::LOCALHOST, port, 0, 0));
```
jdetter: optional nit: Here you're only checking the port on specifically the 127.0.0.1 address which likely could cause this to check pass when it should fail. Many services bind to a specific interface (eth0, enp0s1, etc.) and do not run on localhost so this check would fail to find those services. I think this should also pass in the requested interface IP as well and we should check that in addition to checking the localhost interface. This is especially important if the user passes `0.0.0.0` as their requested interface.

If we don't do this you would just get whatever error you get when you try to bind to the port on that specific interface, which should also hopefully give a reasonable error.

egormanga: Also, this only scopes on IPv4 not IPv6.

cloutiertyler: @jdetter I believe I've now addressed this issue. Could you take a second look.

crates/cli/src/subcommands/dev.rs
cloutiertyler: Good call out. I now print a warning if they're not initializing a new project and they are specifying a `--template`.

crates/standalone/src/subcommands/start.rs
jdetter: Much better - we discussed one small bug in here that we'll fix in another 👍

bfops:
nit: can we re-title this PR to mention the fact that this doesn't just update docs? I worry that _if_ we are debugging something in the future (e.g. related to templates), we would skip right over this PR since the title sounds like docs-only changes.

cloutiertyler:
> nit: can we re-title this PR to mention the fact that this doesn't just update docs? I worry that _if_ we are debugging something in the future (e.g. related to templates), we would skip right over this PR since the title sounds like docs-only changes.

Done

PR #3956, opened 2026-01-06
Fix rollback metrics

joshua-spacetime:
# Description of Changes

Sets the `committed` label to `false` when we rollback a txn. It was always set to `true` before this change.

None

0

# Testing

- [x] Unit test


drogus (approved):
Awesome, thanks for working on that!

PR #3957, opened 2026-01-06
[TS] Implement de/serialization as a tree of closures

coolreader18:
# Description of Changes

This should optimize better than walking the AlgebraicType tree on every call. I'll do some profiling to confirm, but I am already using it in my performance experiments.

2 - fully rewrites the bsatn encoder/decoder, but it's a fairly simple format.

# Testing

- [x] Automated testing
- [x] Profiling

### DIFF
--- a/crates/bindings-typescript/src/lib/algebraic_type.ts
+++ b/crates/bindings-typescript/src/lib/algebraic_type.ts
@@ -80,10 +80,10 @@ export type Deserializer<T> = (reader: BinaryReader) => T;
 // Caches to prevent `makeSerializer`/`makeDeserializer` from recursing
 // infinitely when called on recursive types.
 //
-// We check for recursion in `{Product,Sum}Type.make{Deser,Ser}ializer` because
-// `ProductType` and `SumType` are the cases where recursive types are possible.
-// Additionally, if the check was in `AlgebraicType`, calling
-// `ProductType.makeSerializer` directly wouldn't cache anything.
+// We check for recursion in `{Product,Sum}Type.make{Deser,Ser}ializer` rather
+// than in `AlgebraciType.make{Deser,Ser}ializer` because we need to store the
+// [de]serializer in the cache before recursing into its fields/variants, and
+// we wouldn't be able to do that in the `AlgebraicType` functions.
 const SERIALIZERS = new Map<ProductType | SumType, Serializer<any>>();
 const DESERIALIZERS = new Map<ProductType | SumType, Deserializer<any>>();
 


coolreader18:
Requested @gefjon's review since they expressed familiarity with the strategy.

crates/bindings-typescript/src/lib/algebraic_type.ts
```diff
@@ -318,78 +436,140 @@ export const AlgebraicType = {
...
+            return `writer.write${element.algebraicType.tag}(value.${element.name!});`;
+          } else {
+            return `serializers.${element.name!}(writer, value.${element.name!});`;
+          }
+        })
+        .join('\n');
+    serializer = Function('serializers', 'writer', 'value', body).bind(
+      undefined,
+      serializers
+    ) as Serializer<any>;
```
gefjon, line 488: You're not really doing tree-of-closures compilation at this point, this is just codegen. I assume the tracing JIT is better able to optimize this (please confirm that you've tested), but it is a different and more complex strategy than the one your PR description claims.

crates/bindings-typescript/src/lib/algebraic_type.ts
```diff
@@ -318,78 +436,140 @@ export const AlgebraicType = {
 export type ProductType = ProductTypeType;
 
 export const ProductType = {
+  makeSerializer(
+    ty: ProductTypeType,
+    typespace?: TypespaceType
+  ): Serializer<any> {
+    let serializer = SERIALIZERS.get(ty);
+    if (serializer != null) return serializer;
```
gefjon, line 449: It seems odd to do this check here rather than in `AlgebraicType.makeSerializer`.

crates/bindings-typescript/src/lib/algebraic_type.ts
```diff
@@ -318,78 +436,140 @@ export const AlgebraicType = {
...
+
+    const primitiveFields = new Set();
+    const serializers: Record<string, Serializer<any>> = {};
+    const body =
+      '"use strict";\n' +
+      ty.elements
+        .map(element => {
+          if (primitives.has(element.algebraicType.tag)) {
+            primitiveFields.add(element.name!);
+            return `writer.write${element.algebraicType.tag}(value.${element.name!});`;
```
gefjon: Has inlining the primitive writers in this way shown improvements in benchmarking? I would expect emitting a call to `serializers.${element.name!}(writer, value.${element.name!})` to be the same, and it would dramatically simplify your codegen if you just used `AlgebraicType.makeSerializer` for all the fields.

crates/bindings-typescript/src/lib/algebraic_type.ts
```diff
@@ -318,78 +436,140 @@ export const AlgebraicType = {
...
+    ({ name, algebraicType: { tag } }) => `\
+view.set${primitiveJSName[tag]}(value.${name!}, writer.offset, ${primitiveSizes[tag] > 1 ? 'true' : ''});
+writer.offset += ${primitiveSizes[tag]};`
+  )
+  .join('\n')}`;
+      serializer = Function('writer', 'value', body) as Serializer<any>;
+    }
+
+    const primitiveFields = new Set();
+    const serializers: Record<string, Serializer<any>> = {};
```
gefjon, line 478:
```suggestion
    // Because V8 forces us to generate our code as a string, rather than a proper syntax tree,
    // we can't have our `body` close over values.
    // Instead, we construct an array of the values we'd otherwise "close over" in `serializers`,
    // include that array as an additional prefix argument in the `Function` constructor,
    // and use `Function.prototype.bind` to pass it as an additional argument.
    //
    // We populate `serializers` after constructing this type's `serializer`
    // so that it can close over itself, in the case that `ty` is recursive.
    const serializers: Record<string, Serializer<any>> = {};
```

gefjon: Would it be better to pass the "closure env" as `this`, rather than giving `this` of `undefined` and an additional argument? I don't know if V8 has special optimizations for method receivers, but it seems plausible that it does.

crates/bindings-typescript/src/lib/algebraic_type.ts
```diff
@@ -318,78 +436,140 @@ export const AlgebraicType = {
...
+            fieldName as keyof SpecialProducts
+          ];
       }
+    }
 
-      if (ty.elements[0].name === '__connection_id__') {
-        return new ConnectionId(reader.readU128());
-      }
+    let deserializer = DESERIALIZERS.get(ty);
+    if (deserializer != null) return deserializer;
```
gefjon, line 528: Same note re: moving this check to `AlgebraicType.makeSerializer`.

crates/bindings-typescript/src/lib/algebraic_type.ts
```diff
@@ -318,78 +436,140 @@ export const AlgebraicType = {
...
+  .join('\n')}
+return result;`;
+      deserializer = Function('reader', body) as Deserializer<any>;
     }
 
-    for (const element of ty.elements) {
-      result[element.name!] = AlgebraicType.deserializeValue(
-        reader,
-        element.algebraicType,
+    const deserializers: Record<string, Deserializer<any>> = {};
```
gefjon, line 557: Would like a similar comment here to the one I suggested on `serializers` above.

crates/bindings-typescript/src/lib/algebraic_type.ts
```diff
@@ -318,78 +436,140 @@ export const AlgebraicType = {
...
+const view = writer.view;
+${ty.elements
+  .map(
+    ({ name, algebraicType: { tag } }) => `\
+view.set${primitiveJSName[tag]}(value.${name!}, writer.offset, ${primitiveSizes[tag] > 1 ? 'true' : ''});
+writer.offset += ${primitiveSizes[tag]};`
+  )
+  .join('\n')}`;
+      serializer = Function('writer', 'value', body) as Serializer<any>;
+    }
```
gefjon, line 469: It looks like this block is dead code, as it gets overwritten by the unconditional write to `serializer` below.

crates/bindings-typescript/src/lib/algebraic_type.ts
```diff
@@ -318,78 +436,140 @@ export const AlgebraicType = {
...
+${ty.elements
+  .map(
+    ({ name, algebraicType: { tag } }) => `\
+result.${name} = view.get${primitiveJSName[tag]}(reader.offset, ${primitiveSizes[tag] > 1 ? 'true' : ''});
+reader.offset += ${primitiveSizes[tag]};`
+  )
+  .join('\n')}
+return result;`;
+      deserializer = Function('reader', body) as Deserializer<any>;
     }
```
gefjon, line 548: This block also looks like dead code.

crates/bindings-typescript/src/lib/algebraic_type.ts
gefjon: If it is, in fact, unused, I'd like us to do some very careful thinking and testing to determine whether it's worth changing to actually use this or to remove it, as it adds significant complexity and testing burden.

crates/bindings-typescript/src/lib/algebraic_type.ts
```diff
@@ -318,78 +436,140 @@ export const AlgebraicType = {
...
+  ): Deserializer<any> {
+    switch (ty.elements.length) {
+      case 0:
+        return unitDeserializer;
+      case 1: {
+        const fieldName = ty.elements[0].name!;
+        if (hasOwn(specialProductDeserializers, fieldName))
+          return specialProductDeserializers[
+            fieldName as keyof SpecialProducts
+          ];
```
gefjon, line 523: This seems like it could ideally be handled by `DESERIALIZERS`, but that would only work if we preserve `===`/pointer-equals of special `ProductType` values. Do we, in fact, preserve that?

crates/bindings-typescript/src/lib/algebraic_type.ts
```diff
@@ -72,6 +73,16 @@ export type AlgebraicType = AlgebraicTypeType;
...
 
+export type Serializer<T> = (writer: BinaryWriter, value: T) => void;
+
+export type Deserializer<T> = (reader: BinaryReader) => T;
+
+// Caches to prevent `makeSerializer`/`makeDeserializer` from recursing
+// infinitely when called on recursive types. We use `WeakMap` because we don't
+// care about iterating these, and it can allow the memory to be freed.
+const SERIALIZERS = new WeakMap<ProductType | SumType, Serializer<any>>();
+const DESERIALIZERS = new WeakMap<ProductType | SumType, Deserializer<any>>();
```
gefjon: I'm not sure using `WeakMap` here is actually profitable. IIRC, we construct all `AlgebraicType` instances at load-time and keep them live in constants for the lifetime of the program, so I wouldn't expect any of its values to ever be freed. Given that, and that weak hash tables have some non-zero bookkeeping overhead (either at GC-time or access-time), it seems better to use a map here.

crates/bindings-typescript/src/lib/algebraic_type.ts
```diff
@@ -426,79 +606,94 @@ export type SumType = SumTypeType;
...
+          serializeOk(writer, value.ok);
+        } else if ('err' in value) {
+          writer.writeU8(1);
+          serializeErr(writer, value.err);
+        } else {
+          throw new TypeError(
+            'could not serialize result: object had neither a `ok` nor an `err` field'
+          );
+        }
+      };
```
gefjon, line 672: Is this necessary because we have a different repr for results compared to other sums? It looks like we use `{ "ok": value }` for result, but `{ "tag": "foo", "value": value }` for other sums. If so, that's unfortunate. Anyways, I'd like to see a comment here justifying why we actually just build a closure here, where for other types we do proper codegen.

crates/bindings-typescript/src/lib/algebraic_type.ts
```diff
@@ -426,79 +606,94 @@ export type SumType = SumTypeType;
...
+    } else {
+      let serializer = SERIALIZERS.get(ty);
+      if (serializer != null) return serializer;
```
gefjon, line 675: Ditto.

crates/bindings-typescript/src/lib/algebraic_type.ts
```diff
@@ -426,79 +606,94 @@ export type SumType = SumTypeType;
...
+          {
+            index,
+            serialize: AlgebraicType.makeSerializer(
+              element.algebraicType,
+              typespace
+            ),
+          },
+        ])
+      );
+      return serializer;
```
gefjon: Why are we building a closure here rather than generating a "native" function? Do you have some reason to believe codegen is unnecessary for sums (this would surprise me, given the lookup in `variants`, which should be very optimizable for a particular sum type but not for any arbitrary sum), or do we just not care that much about the performance of ser/deing sums? Either way, please leave a comment - it would be good to know if codegen here would be profitable in the future, or if you've already ruled it out as a beneficial optimization.

crates/bindings-typescript/src/lib/algebraic_type.ts
```diff
@@ -426,79 +606,94 @@ export type SumType = SumTypeType;
...
+        typespace
+      );
+      return (writer, value) => {
+        if (value !== null && value !== undefined) {
+          writer.writeByte(0);
+          serialize(writer, value);
+        } else {
+          writer.writeByte(1);
+        }
+      };
```
gefjon, line 645: Same questions as for results below.

crates/bindings-typescript/src/lib/algebraic_type.ts
```diff
@@ -426,79 +606,94 @@ export type SumType = SumTypeType;
...
+    writer: BinaryWriter,
     ty: SumTypeType,
+    value: any,
     typespace?: TypespaceType
-  ): any {
-    const tag = reader.readU8();
+  ): void {
+    SumType.makeSerializer(ty, typespace)(writer, value);
+  },
+  makeDeserializer(
```
gefjon, line 724: All the same comments here as in `SumType.makeSerializer`.

crates/bindings-typescript/src/lib/algebraic_type.ts
```diff
@@ -107,148 +118,86 @@ export const AlgebraicType = {
...
-      case 'String':
-        writer.writeString(value);
-        break;
+      default:
+        return primitiveSerializers[ty.tag];
     }
   },
-  deserializeValue: function (
-    reader: BinaryReader,
+  serializeValue(
```
gefjon, line 156: Is this method and friends (`deserializeValue` and the ones on other meta-types) used? It seems like a serious performance hazard and we should remove it if possible, I think.

crates/bindings-typescript/src/lib/algebraic_type.ts
```diff
@@ -426,79 +606,94 @@ export type SumType = SumTypeType;
...
+        }
+        const { index, serialize } = variant;
+        writer.writeU8(index);
+        serialize(writer, value.value);
+      };
+      // In case `ty` is recursive, we cache the function *before* before computing
+      // `variants`, so that a recursive `makeSerializer` with the same `ty` has
+      // an exit condition.
+      SERIALIZERS.set(ty, serializer);
+      const variants = new Map(
```
gefjon: How does this work? It looks like you're initializing `variants` here, so the above closure isn't capturing this value. Is this a hoisting thing?

crates/bindings-typescript/src/lib/algebraic_type.ts
```diff
@@ -507,49 +702,78 @@ export const SumType = {
...
+      deserializer = Function(
+        'deserializers',
+        'reader',
+        `switch (reader.readU8()) {\n${ty.variants
+          .map(
+            ({ name }, i) =>
+              `case ${i}: return { tag: ${JSON.stringify(name!)}, value: deserializers.${name!}(reader) };`
+          )
+          .join('\n')} }`
+      ).bind(undefined, deserializers) as Deserializer<any>;
```
gefjon: So we are doing "native" codegen for deserializing sums, but not for serializing them? And only for generalized sums, not options or results? Huh?

crates/bindings-typescript/src/lib/algebraic_type.ts
coolreader18: Otherwise, calling `ProductType.makeSerializer` directly wouldn't cache the function.

crates/bindings-typescript/src/lib/algebraic_type.ts
coolreader18: I think you're right, I'll revert it.

crates/bindings-typescript/src/lib/algebraic_type.ts
gefjon: Do we call `ProductType.makeSerializer` directly anywhere that we can't easily rewrite to call `AlgebraicType.makeSerializer`?

crates/bindings-typescript/src/lib/algebraic_type.ts
coolreader18: I don't think it's guaranteed, and I think it would be fragile, since anyone can create a `{ __timestamp__: 1n }`.

crates/bindings-typescript/src/lib/algebraic_type.ts
coolreader18: Fair enough.

crates/bindings-typescript/src/lib/algebraic_type.ts
coolreader18: Yeah, I wasn't aware of it til I started working on this. Seems like it was added in #3790.

crates/bindings-typescript/src/lib/algebraic_type.ts
coolreader18: I only didn't do codegen here only because the case I was benchmarking doesn't use sum types. I'll leave a TODO for the future.

crates/bindings-typescript/src/lib/algebraic_type.ts
coolreader18: Also, the codegen optimizes property access. In this case, where it's the same object shapes, it doesn't matter as much.

crates/bindings-typescript/src/lib/algebraic_type.ts
coolreader18: that's just how javascript closures work :) unlike rust closures, they have mutable access to their entire enclosing scope.

crates/bindings-typescript/src/lib/algebraic_type.ts
gefjon: I don't expect our `AlgebraicType` and ser/de machinery to be user-facting. I also don't see how a user constructing an object `{ __timestamp__: 1n }` would ever reach this point, unless they also constructed their own `AlgebriacType` for it and then serialized using that type.

gefjon: That said, if you think relying on `===` here would be overly fragile, then don't make a change, and feel free to mark resolved.

crates/bindings-typescript/src/lib/algebraic_type.ts
coolreader18: I mean, no, but sometimes you have a ProductType and it's nice not to have the extra noise of wrapping it with `AlgebraicType.Product()`. Also, we want to key `SERIALIZERS` on the `ProductType`, not the `AlgebraicType`, since for a `product: ProductType`, `AlgebraicType.Product(product) !== AlgebraicType.Product(product)` due to reference equality.

coolreader18: Oh! I remembered the main reason I did this: we need to insert `serializer` into `SERIALIZERS` *before* we fill out the `serializers`. I originally did it the way you suggested, but it was unbounded recursion.

PR #3958, opened 2026-01-06
Use a class for ReducerCtx

coolreader18:
# Description of Changes

#3538 added a `uuid_counter` field to ReducerCtx, which has no need to be public. Implementing ReducerCtx as a class allows us to encapsulate better, and lets us enumerate exactly the data that it needs to hold so that the runtime could possibly optimize it.

1: straightforward switch

# Testing

- [x] Automated testing is sufficient.


cloutiertyler (approved):
No objection

PR #3961, opened 2026-01-07
In-memory DatabaseLogger

kim:
This is the second step to make in-memory-only databases not touch the disk at all.

While at it, also make it so file-backed module logs are streamed in constant memory where possible.

Depends-on: #3912 

2

# Testing

Added some unit-level tests.

### DIFF
--- a/crates/client-api/src/routes/database.rs
+++ b/crates/client-api/src/routes/database.rs
@@ -1,8 +1,8 @@
 use std::borrow::Cow;
-use std::env;
 use std::num::NonZeroU8;
 use std::str::FromStr;
 use std::time::Duration;
+use std::{env, io};
 
 use crate::auth::{
     anon_auth_middleware, SpacetimeAuth, SpacetimeEnergyUsed, SpacetimeExecutionDurationMicros, SpacetimeIdentity,
@@ -11,8 +11,8 @@ use crate::auth::{
 use crate::routes::subscribe::generate_random_connection_id;
 pub use crate::util::{ByteStringBody, NameOrIdentity};
 use crate::{
-    log_and_500, Action, Authorization, ControlStateDelegate, DatabaseDef, DatabaseResetDef, Host, NodeDelegate,
-    Unauthorized,
+    log_and_500, Action, Authorization, ControlStateDelegate, DatabaseDef, DatabaseResetDef, Host, MaybeMisdirected,
+    NodeDelegate, Unauthorized,
 };
 use axum::body::{Body, Bytes};
 use axum::extract::{Path, Query, State};
@@ -20,9 +20,11 @@ use axum::response::{ErrorResponse, IntoResponse};
 use axum::routing::MethodRouter;
 use axum::Extension;
 use axum_extra::TypedHeader;
+use futures::TryStreamExt;
 use http::StatusCode;
 use log::{info, warn};
 use serde::Deserialize;
+use spacetimedb::database_logger::DatabaseLogger;
 use spacetimedb::host::module_host::ClientConnectedError;
 use spacetimedb::host::{CallResult, UpdateDatabaseResult};
 use spacetimedb::host::{FunctionArgs, MigratePlanResult};
@@ -82,6 +84,7 @@ pub struct CallParams {
 }
 
 pub const NO_SUCH_DATABASE: (StatusCode, &str) = (StatusCode::NOT_FOUND, "No such database.");
+const MISDIRECTED: (StatusCode, &str) = (StatusCode::NOT_FOUND, "Database is not scheduled on this host");
 
 fn map_reducer_error(e: ReducerCallError, reducer: &str) -> (StatusCode, String) {
     let status_code = match e {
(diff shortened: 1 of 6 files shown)


crates/client-api/src/routes/database.rs
```diff
@@ -434,52 +433,28 @@ where
...
+            body,
+        )
+    }
 
-        let stream = futures::stream::once(std::future::ready(lines.into()))
-            .chain(stream)
-            .map(Ok::<_, std::convert::Infallible>);
+    let Some(host) = worker_ctx.leader(database.id).await.map_err(log_and_500)? else {
+        if follow {
+            return Err((StatusCode::NOT_FOUND, "Host not running, cannot follow logs").into());
```
gefjon: I think this is a misleading error message. My reading of the doc comment on `NodeDelegate::leader` is that it will spawn the host if necessary, and will only return `None` for databases whose leader is not assigned to this node. (I am unsure of the interaction with in-memory databases, if that's relevant in this path.)

crates/core/src/database_logger.rs
```diff
@@ -1,24 +1,161 @@
...
+use tokio_stream::wrappers::BroadcastStream;
+use tokio_util::io::ReaderStream;
 
 use spacetimedb_paths::server::{ModuleLogPath, ModuleLogsDir};
 
 use crate::util::asyncify;
 
 pub struct DatabaseLogger {
-    inner: Mutex<DatabaseLoggerInner>,
+    inner: Arc<Mutex<Box<dyn Logger + Send>>>,
```
gefjon: I'm not going to request changes on this PR, but in the future when designing abstractions like this one, I would appreciate it if we could restrict our usage of `dyn` to cases when there is a need for downstream library consumer extensibility, and prefer to use `enum`s for cases where we may have one of a statically bounded and small set of possible implementations.

crates/core/src/database_logger.rs
```diff
@@ -1,24 +1,161 @@
...
+
+/// Storage backend of a [DatabaseLogger].
+trait Logger {
+    /// Append the serialized log `record` at timestamp `ts` to this logger.
+    fn append(&mut self, ts: chrono::DateTime<Utc>, record: Bytes);
+    /// Calculate the size of this logger in bytes.
+    fn size(&self) -> io::Result<u64>;
+    /// Read up to `n` lines from the tail of this logger into memory.
+    fn tail(&self, n: u32) -> io::Result<Bytes>;
+    /// Stream up to `n` lines from the tail of this logger.
```
gefjon, line 54:
```suggestion
    /// Stream up to `n` lines from the tail of this logger.
    /// If `n` is `None`, stream all the contained lines.
```

crates/core/src/database_logger.rs
kim: Is this just your preference or are there deeper reasons?

I had it as an enum initially but got tired of the dispatch boilerplate. Also, I am fairly certain that we will want an external implementation that can gather logs from remote machines soonishly.

crates/client-api/src/routes/database.rs
kim: Hm you're right. This has been incorrect before because it isn't considered that this might be the right host to read disk-backed module logs from, but the database is not running (e.g. because it is suspended, or currently bootstrapping).

For now, I'll revert to return 404 with the message of the `get_leader_replica_by_database` call above, and remove that call as it is redundant.
I'll come up with a follow-up.

crates/core/src/database_logger.rs
```diff
@@ -195,44 +344,93 @@ impl DatabaseLogger {
...
+            // locking and spawning can be avoided.
+            None | Some(0) => self.subscribe().map(Ok).boxed(),
+
+            // Otherwise, we need to hold the lock to prevent writes
+            // while we gather a snapshot of the persistent tail.
+            Some(n) => {
+                // Cap reading the tail into memory at a few hundred KiB.
+                let n = n.min(2500);
+                let (tail, more) = {
+                    let inner = self.inner.lock_arc();
```
coolreader18: This seems questionable to me - won't this block if another thread is in the process of doing the `sync_data()` call below?

coolreader18: If we want to do this we should just make it an async mutex.

kim: We have to do this if we don't want to serve duplicate log lines.

Using an async mutex doesn't work, though, as we are calling `DatabaseLogger:write` from within an async context virtually everywhere. I guess we always had a problem with blocking here.

kim: Made it to use message-passing. PTAL

crates/client-api/src/routes/database.rs
kim: https://github.com/clockworklabs/SpacetimeDB/pull/3984

crates/core/src/database_logger.rs
kim: It's not longer `dyn`

Issue #3962, opened 2026-01-07
C# bug: STDB0024 Error when a View returns a `RefOption` result


rekhoff:
As reported by a Spacetime community user Arc, a bug exists in the C# View code where they were getting an error in 1.11.1 on the code:
``` csharp
[View(Name = nameof(MyAccount), Public = true)]
public static Account? MyAccount(ViewContext ctx)
{
	return ctx.Db.Account.Identity.Find(ctx.Sender) as Account;
}
```
Where `Account?` is a `class`.

The validator for `[SpacetimeDB.View]` return types was only accepting:

* `SpacetimeDB.BSATN.List<...>` (like “Vec<T>”)
* `SpacetimeDB.BSATN.ValueOption<...>` (nullable value types / `T?` where `T : struct`)

But nullable reference types like `Account?` (where `Account : class`) are represented by BSATN as:
* `SpacetimeDB.BSATN.RefOption<...>`

So the analyzer/codegen incorrectly rejected `Account?` with:
> View 'MyAccount' must return Vec or Option. (STDB0024)

The existing [regression test](https://github.com/clockworklabs/SpacetimeDB/blob/e8f9079dc50aff44af71840ef17ee94cf678268a/sdks/csharp/examples~/regression-tests/server/Lib.cs#L111) passed because the row type there is a `struct` (so `Player?` mapped to `ValueOption`) or otherwise didn’t hit the `RefOption` path.

PR #3964, opened 2026-01-07
Enable `RefOption` returns from Views to support Views returning a single `class`

rekhoff:
# Description of Changes

This PR adds a regression test covering nullable reference-type view returns in C# modules (e.g. Account? where Account is a class), as reported in [#3962](https://github.com/clockworklabs/SpacetimeDB/issues/3962).

```csharp
public static Account? MyAccount(ViewContext ctx)
```
* Updated the C# regression-test server module to include a reference-type table and views that exercise the RefOption path:
* A new public reference-type table row:
```csharp
[SpacetimeDB.Table(Name = "account", Public = true)]
public partial class Account { ... }
```
* A public at-most-one view that returns a nullable reference type (Account?) via Find(...):
```csharp
[SpacetimeDB.View(Name = "my_account", Public = true)]
public static Account? MyAccount(ViewContext ctx)
{
    return ctx.Db.account.Identity.Find(ctx.Sender) as Account;
}
```
* A second public view that returns null to ensure the “empty result” case is exercised:
```csharp
[SpacetimeDB.View(Name = "my_account_missing", Public = true)]
public static Account? MyAccountMissing(ViewContext ctx) => null;
```
* Updated ClientConnected to ensure an Account row exists for the connecting identity so the “one row” case is deterministic.
* Updated the C# regression-test client to:
  * Subscribe to the new views:
    * `SELECT * FROM my_account`
    * `SELECT * FROM my_account_missing`
  * Assert correct semantics for nullable reference-type view returns:
    * MyAccount.Count == 1
    * MyAccountMissing.Count == 0
  * Updated the regression-test server project to use local C# runtime/codegen project references so the regression module exercises the in-repo generator/runtime behavior (instead of the published SpacetimeDB.Runtime package).

None.
* No changes to public module schema/wire format semantics beyond adding regression-test-only tables/views.
* No behavior changes outside the C# regression test module + harness.

2 - Low

* Changes are isolated to regression tests and project wiring.
* The scenario specifically guards the nullable reference-type “Option-like view return” path against regressions.

# Testing

- [X] Ran C# regression tests with no failures in new View tests


JasonAtClockwork (approved):
Solid fix + test!

Issue #3960, opened 2026-01-06
Inserting a null string into a table throws an exception


rekhoff:
Steve reported that: Inserting a null string into a table throws an exception:

```
System.ArgumentNullException: ArgumentNull_Generic Arg_ParamName_Name, s
   ...
   at System.Text.UTF8Encoding.UTF8EncodingSealed.GetBytes(String )
   at SpacetimeDB.BSATN.String.Write(BinaryWriter , String )
   at Module.Person.WriteFields(BinaryWriter )
   at SpacetimeDB.BSATN.IStructuralReadWrite.ToBytes[Person](Person )
   at SpacetimeDB.Internal.ITableView`2[[SpacetimeDB.Internal.TableHandles.Person,[Module.Person]].DoInsert(Person )
   at SpacetimeDB.Internal.TableHandles.Person.Insert(Person )
```

It was validated that both of these are happening on 1.11.1. Here's minimal code to trigger both this and #3959:
```csharp
using SpacetimeDB;

public static partial class Module
{
    [Table]
    public partial struct Person
    {
        [PrimaryKey, AutoInc]
        public int Id;
        public string Name;
    }

    [Reducer]
    public static void Test1(ReducerContext ctx) => ctx.Db.Person.Insert(new Person { Name = "" }); //Works fine

    [Reducer]
    public static void Test2(ReducerContext ctx) => ctx.Db.Person.Insert(new Person { Name = null }); //Throws an exception

    [Reducer]
    public static void Test3(ReducerContext ctx) => throw new Exception("Message");
}
```

PR #3967, opened 2026-01-08
Fix C# null string serialization error + add regression coverage for Issue #3960

rekhoff:
# Description of Changes
This PR fixes a C# SDK/BSATN serialization issue where inserting a `null` string into a non-nullable `string` column could throw an `ArgumentNullException` with an unhelpful stack trace, as reported in [#3960](https://github.com/clockworklabs/SpacetimeDB/issues/3960).

* Updated the C# BSATN runtime to reject `null` for non-nullable `string` serialization with a clear, actionable error message (guides users to model nullable strings as `string?` so BSATN encodes an option).
* Added BSATN runtime unit tests to cover:
  * Non-nullable string rejects `null` with an actionable message
  * Nullable string (`RefOption<string, BSATN.String>`) round-trips `null`
* Added end-to-end C# regression-test module + client assertions for #3960:
  * Inserting `""` into non-nullable `string` succeeds
  * Inserting `null` into non-nullable `string` fails with the expected message
  * Inserting `null` into nullable `string?` succeeds and round-trips as `null`
None.
* No schema or wire-format changes.
* Behavior change is limited to improving the error surfaced when attempting to serialize null as a non-nullable BSATN `String`, plus stricter input validation for `ConnectionId` / `Identity` parsing.
2 - Low
* Changes are isolated to regression test coverage.
* No impact on valid serialization paths; only affects invalid/malformed inputs and improves diagnostics.
# Testing
- [X] Ran C# regression-tests client and verified new tests pass.

### DIFF
--- a/crates/bindings-csharp/BSATN.Runtime/Builtins.cs
+++ b/crates/bindings-csharp/BSATN.Runtime/Builtins.cs
@@ -53,13 +53,10 @@ internal static class Util
     public static T Read<T>(ReadOnlySpan<byte> source, bool littleEndian)
         where T : struct
     {
-        var expectedSize = Marshal.SizeOf<T>();
-        if (source.Length != expectedSize)
-        {
-            throw new ArgumentException(
-                $"Error while reading {typeof(T).FullName}: expected source span to be {expectedSize} bytes long, but was {source.Length} bytes."
-            );
-        }
+        Debug.Assert(
+            source.Length == Marshal.SizeOf<T>(),
+            $"Error while reading ${typeof(T).FullName}: expected source span to be {Marshal.SizeOf<T>()} bytes long, but was {source.Length} bytes."
+        );
 
         var result = MemoryMarshal.Read<T>(source);
 
@@ -169,18 +166,8 @@ public readonly record struct ConnectionId
     /// </summary>
     /// <param name="hex"></param>
     /// <returns></returns>
-    public static ConnectionId? FromHexString(string hex)
-    {
-        if (hex.Length != 32)
-        {
-            throw new ArgumentException(
-                $"Expected ConnectionId hex string to be 32 characters long, but was {hex.Length}.",
-                nameof(hex)
-            );
-        }
-
-        return FromBigEndian(Util.StringToByteArray(hex));
-    }
+    public static ConnectionId? FromHexString(string hex) =>
+        FromBigEndian(Util.StringToByteArray(hex));
 
     public static ConnectionId Random()
     {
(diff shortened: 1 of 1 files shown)


crates/bindings-csharp/BSATN.Runtime/BSATN/Runtime.cs
```diff
@@ -542,8 +542,18 @@ public AlgebraicType GetAlgebraicType(ITypeRegistrar registrar) =>
...
         Encoding.UTF8.GetString(ByteArray.Instance.Read(reader));
 
-    public void Write(BinaryWriter writer, string value) =>
+    public void Write(BinaryWriter writer, string value)
+    {
+        if (value is null)
+        {
+            throw new ArgumentNullException(
+                nameof(value),
+                "Cannot serialize a null string as BSATN String. If nulls are allowed, use a nullable string (string?) so it is encoded as an option."
```
jdetter: I think this is a bit more clear:
```suggestion
                "Cannot serialize a null string as BSATN String. To serialize a null string you must use a nullable string (string?) so it is encoded as a BSATN option."
```

crates/bindings-csharp/BSATN.Runtime/Builtins.cs
```diff
@@ -53,10 +53,13 @@ public static string ToHexBigEndian<T>(T val)
...
-            source.Length == Marshal.SizeOf<T>(),
-            $"Error while reading ${typeof(T).FullName}: expected source span to be {Marshal.SizeOf<T>()} bytes long, but was {source.Length} bytes."
-        );
+        var expectedSize = Marshal.SizeOf<T>();
+        if (source.Length != expectedSize)
+        {
+            throw new ArgumentException(
+                $"Error while reading {typeof(T).FullName}: expected source span to be {expectedSize} bytes long, but was {source.Length} bytes."
+            );
+        }
```
jdetter: I know these fixes are related to this point:
> Fixed unrelated Release-only property test flakiness by making ConnectionId / Identity parsing strict about input length (throws on invalid lengths rather than relying on Debug.Assert).

Usually it's better to include these sorts of unrelated fixes as separate PRs. The main benefit to this is if there is a bug in this code and we try to find the PR that introduced the issue it will come back to this PR which is quite confusing.

An optional nit here would be to move these changes out to a new PR. If you have questions let me know 🙂

sdks/csharp/examples~/regression-tests/server/Lib.cs
```diff
@@ -202,6 +222,24 @@ public static void SetNullableVec(ReducerContext ctx, uint id, bool hasPos, int
         }
     }
 
+    [SpacetimeDB.Reducer]
+    public static void InsertEmptyStringIntoNonNullable(ReducerContext ctx)
+    {
+        ctx.Db.null_string_nonnullable.Insert(new NullStringNonNullable { Name = "" });
```
jdetter, line 240: I love to see the regression tests, thank you Ryan 🙏

crates/bindings-csharp/BSATN.Runtime/Builtins.cs
rekhoff: Understood. Moved changes to `crates/bindings-csharp/BSATN.Runtime/Builtins.cs`  from this PR and a separate PR #3988

PR #3968, opened 2026-01-08
deadlock fix: Make procedure tx abis synchronous

Shubham8287:
# Description of Changes
Fixes the deadlock in procedures due to holding tx lock across `async` abis.
More info can be found here - https://discordapp.com/channels/931210784011321394/1011381307965722774/1458432377574658060.

This patch converts following procedure abis from `async` to sync, which may seem problematic but it's not as we already hold tx locks in reducers while blocking the worker thread.
 
```rust
"spacetime_10.3"::procedure_start_mut_tx,
"spacetime_10.3"::procedure_commit_mut_tx,
"spacetime_10.3"::procedure_abort_mut_tx,

```

# API and ABI breaking changes:
I am surprised that this does not turned out to be ABI breaking change.
Maybe because from wasm side, async abis were always treated as synchoronous.

# Expected complexity level and risk
2, diff is straightforward but it can have hidden implications.

# Testing
I want someone else to also do the similar testing as mine. Just in case, I missed something and that appeared it to work.

I did the backward compatible testing as following:

1. I published following module on master branch:
```rust
use spacetimedb::{ProcedureContext, Table};

#[spacetimedb::table(name = my_table)]
struct MyTable {
    a: u32,
    b: u8,
}

#[spacetimedb::procedure]
fn insert_a_value(ctx: &mut ProcedureContext, a: u32, b: u8) {
    ctx.with_tx(|ctx| {
        ctx.db.my_table().insert(MyTable { a, b });
    });
}
```

2. Called a procedure to insert some values: `spacetime call quick insert_a_value 1 2`
3. Switched to my branch, re-compiled and restarted the server.
4. Called a procedure again with different values, verfied both old and new values are present.

PR #3972, opened 2026-01-08
Marks `MutTxId` as `!Send`

Shubham8287:
# Description of Changes
Depends on https://github.com/clockworklabs/SpacetimeDB/pull/3968.

Makes `MutTxId,` `!Send` to avoid being deadlocked by holding it across `await` points
after taking learnings from https://github.com/clockworklabs/SpacetimeDB/pull/3938 and  https://github.com/clockworklabs/SpacetimeDB/pull/3968.

NA

1, It only increases the compile time safety.

# Testing
Existing should be enough.

### DIFF
--- a/crates/core/src/host/instance_env.rs
+++ b/crates/core/src/host/instance_env.rs
@@ -63,6 +63,26 @@ pub struct InstanceEnv {
 /// such as:
 /// https://github.com/clockworklabs/SpacetimeDB/pull/3938 and
 /// https://github.com/clockworklabs/SpacetimeDB/pull/3968.
+/// `InstanceEnv` needs to be `Send` because it is created on the host thread
+/// and moved to module threads for execution (see [`ModuleHost::with_instance`]).
+///
+/// `TxSlot` must be `None` whenever `InstanceEnv` is moved across threads, which is
+/// not enforced at compile time but seems to be upheld in practice.
+///
+/// In the future, we may push to use `InstanceEnv` only within a module thread,
+/// but this still helps prevent a set of bugs that occurred due to `MutTxId` being `Send`,
+/// such as:
+/// https://github.com/clockworklabs/SpacetimeDB/pull/3938 and
+/// https://github.com/clockworklabs/SpacetimeDB/pull/3968.
+///
+/// # Safety
+///
+/// `InstanceEnv` doesn't auto-derive `Send` because it may hold a `MutTxId`,
+/// which we've manually made `!Send` to preserve logical invariants.
+/// As described above, sending an `InstanceEnv` while it holds a `MutTxId` will violate logical invariants,
+/// but this is not a safety concern.
+/// Transferring a `MutTxId` between threads will never cause Undefined Behavior,
+/// though it is likely to lead to deadlocks.
 unsafe impl Send for InstanceEnv {}
 
 #[derive(Clone, Default)]


crates/core/src/host/scheduler.rs
```diff
@@ -394,60 +394,81 @@ pub(super) async fn call_scheduled_function(
...
+                let reschedule = delete_scheduled_function_row(Some(tx), Some((timestamp, instant)));
+                Function::Procedure { params, reschedule }
+            }
         }
-        CallParams::Procedure(params) => {
-            // Delete scheduled row.
-            let reschedule = delete_scheduled_function_row(Some(tx), Some((timestamp, instant)));
+    };
 
+    // Below code is outside of the DB transaction scope because the
```
Shubham8287, line 454: this code has to be refactored only to make compiler believe that it doesn't hold tx across `await`, whcih were already not happening here.

crates/core/src/host/instance_env.rs
```diff
@@ -52,8 +52,22 @@ pub struct InstanceEnv {
...
+///
+/// `TxSlot` must be `None` whenever `InstanceEnv` is moved across threads, which is
+/// not enforced at compile time but seems to be upheld in practice.
+///
+/// In the future, we may push to use `InstanceEnv` only within a module thread,
+/// but this still helps prevent a set of bugs that occurred due to `MutTxId` being `Send`,
+/// such as:
+/// https://github.com/clockworklabs/SpacetimeDB/pull/3938 and
+/// https://github.com/clockworklabs/SpacetimeDB/pull/3968.
+unsafe impl Send for InstanceEnv {}
```
joshua-spacetime, line 86: Very helpful comment. I still have some reservations about having to mark this `unsafe`.

gefjon, line 86:
```suggestion
/// `InstanceEnv` needs to be `Send` because it is created on the host thread
/// and moved to module threads for execution (see [`ModuleHost::with_instance`]).
///
/// `TxSlot` must be `None` whenever `InstanceEnv` is moved across threads, which is
/// not enforced at compile time but seems to be upheld in practice.
///
/// In the future, we may push to use `InstanceEnv` only within a module thread,
/// but this still helps prevent a set of bugs that occurred due to `MutTxId` being `Send`,
/// such as:
/// https://github.com/clockworklabs/SpacetimeDB/pull/3938 and
/// https://github.com/clockworklabs/SpacetimeDB/pull/3968.
///
/// # Safety
///
/// `InstanceEnv` doesn't auto-derive `Send` because it may hold a `MutTxId`,
/// which we've manually made `!Send` to preserve logical invariants.
/// As described above, sending an `InstanceEnv` while it holds a `MutTxId` will violate logical invariants,
/// but this is not a safety concern.
/// Transferring a `MutTxId` between threads will never cause Undefined Behavior,
/// though it is likely to lead to deadlocks.
unsafe impl Send for InstanceEnv {}
```

gefjon: I've suggested an extension with a section in the doc comment justifying the `unsafe`.

crates/core/src/host/instance_env.rs
```diff
@@ -52,8 +52,22 @@ pub struct InstanceEnv {
...
+
 #[derive(Clone, Default)]
 pub struct TxSlot {
+    // Wrapped in Mutex for interior mutability.
```
gefjon, line 90: I don't see this comment as particularly helpful. Obviously the mutex is for interior mutability, that's what mutexes do. A more useful comment would describe why interior mutability is necessary: do we hold multiple `&` shared references to a single `TxSlot`, or do we clone the inner `Arc`? Who holds these pointers, and when do they access them?

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -246,6 +247,8 @@ pub struct MutTxId {
     pub timer: Instant,
     pub ctx: ExecutionContext,
     pub metrics: ExecutionMetrics,
+    // Marks `MutTxId` as `!Send` by embedding a non-`Send` type.
```
gefjon, line 250:
```suggestion
    /// Marks `MutTxId` as `!Send` by embedding a non-`Send` type.
    ///
    /// We want `MutTxId` to be `!Send` to prevent holding it across `await` points,
    /// which would lead to deadlocks.
    /// It is not possible to invoke Undefined Behavior by improperly sending a `MutTxId`,
    /// but it will likely deadlock and cause SpacetimeDB to become unresponsive.
```

PR #3973, opened 2026-01-08
Provide a common `Index` trait & simplify and shrink `TableIndex`

Centril:
# Description of Changes

Provides common traits `Index` and `RangedIndex` that all (current) variants in `TypedIndex` adhere to.
This is then used to simplify `TypedIndex` by merging common code,
with more simplifications to come later.
The responsibility of tracking statistics is also moved into each index type, as some can exploit their properties to provide some statistics for free, rather than storing statistics in fields.

None

2?

# Testing

This is mostly code motion, so it is covered by existing tests.
Further test improvements will come in a follow up PR.

### DIFF
--- a/crates/sats/src/proptest.rs
+++ b/crates/sats/src/proptest.rs
@@ -7,13 +7,14 @@ use crate::{
     AlgebraicType, AlgebraicTypeRef, AlgebraicValue, ArrayValue, ProductType, ProductValue, SumType, SumValue,
     Typespace, F32, F64,
 };
+use core::fmt::Debug;
 use proptest::{
     collection::{vec, SizeRange},
     prelude::*,
     prop_oneof,
 };
 
-const SIZE: usize = 16;
+pub const SIZE: usize = 16;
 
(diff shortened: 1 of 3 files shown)


crates/table/src/table_index/mod.rs
```diff
@@ -1133,19 +1058,13 @@ impl MemoryUsage for TableIndex {
...
-            + key_type.heap_usage()
-            + num_rows.heap_usage()
-            + num_key_bytes.heap_usage()
-            + indexed_columns.heap_usage()
+        idx.heap_usage() + key_type.heap_usage() + indexed_columns.heap_usage()
     }
 }
 
-static_assert_size!(TableIndex, 88);
+static_assert_size!(TableIndex, 80);
```
Centril, line 1067: This space saving is thanks to nixing `num_rows` and `num_key_bytes`. This is good, as adding hash indices will increase the size of `TableIndex` by 8 bytes, so we end up where we started.

PR #3974, opened 2026-01-08
Add error handling for ranged seeks on non-range compat indices

Centril:
# Description of Changes

When doing a ranged seek on a non-ranged index (none such exist yet, but will be added in a follow up), return an (ABI) error.

Also:

- Use point scans in query execution (`IxScan(Delta)Eq`).
- Refactor table index code with macro `same_for_all_types`.

None

2?

# Testing

Testing the error handling will be possible once hash indices are added (follow up PR).

### DIFF
--- a/crates/codegen/src/unrealcpp.rs
+++ b/crates/codegen/src/unrealcpp.rs
@@ -2508,10 +2508,17 @@ fn generate_remote_reducers_class(
         writeln!(output, ");");
         writeln!(output);
 
-        // Generate invoke method
+        // Generate invoke method (UObject version - kept for backwards compatibility)
         write!(
             output,
             "    bool Invoke{reducer_pascal}(const FReducerEventContext& Context, const U{reducer_pascal}Reducer* Args);"
+        );
+        writeln!(output);
+
+        // Generate invoke method (FArgs version - zero allocation, used internally)
+        write!(
+            output,
+            "    bool Invoke{reducer_pascal}WithArgs(const FReducerEventContext& Context, const F{reducer_pascal}Args& Args);"
         );
         writeln!(output);
         writeln!(output);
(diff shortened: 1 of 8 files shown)


crates/core/src/estimation.rs
```diff
@@ -153,10 +157,14 @@ fn row_est(tx: &Tx, src: &SourceExpr, ops: &[Query]) -> u64 {
...
 /// The estimated number of rows that an index probe will return.
-/// Note this method is not applicable to range scans.
+/// Note this method is not applicable to range scans,
+/// but it does work for multi column indices.
 fn index_row_est(tx: &Tx, table_id: TableId, cols: &ColList) -> u64 {
-    tx.num_distinct_values(table_id, cols)
-        .map_or(0, |ndv| tx.table_row_count(table_id).unwrap_or(0) / ndv)
+    let table_rc = || tx.table_row_count(table_id).unwrap_or_default();
+    match tx.num_distinct_values(table_id, cols) {
+        Some(ndv) => ndv.map_or(0, |ndv| table_rc() / ndv),
```
gefjon: `unwrap_or_default` looks like it leads to a divide-by-zero here.

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -1373,27 +1374,36 @@ impl MutTxId {
...
         commit_index: TableAndIndex<'a>,
         bounds: &impl RangeBounds<AlgebraicValue>,
-    ) -> IndexScanRanged<'a> {
+    ) -> IndexSeekRangeResult<IndexScanRanged<'a>> {
         // Get an index seek iterator for the tx and committed state.
-        let tx_iter = tx_index.map(|i| i.seek_range(bounds));
+        let tx_iter = tx_index.map(|i| i.seek_range(bounds)).transpose();
         let commit_iter = commit_index.seek_range(bounds);
 
+        // Deal with not having ranged indices.
```
gefjon:
```suggestion
        // If we don't have a range-capable index, return an error.
```

crates/datastore/src/locking_tx_datastore/tx.rs
```diff
@@ -173,14 +177,16 @@ impl TxId {
     /// Returns `None` if:
     /// - No such table as `table_id` exists.
     /// - The table `table_id` does not have an index on exactly the `cols`.
+    ///
+    /// Returns `Some(None)` if:
```
gefjon: Rather than nesting `Option` here, can we return a custom enum with 3 named variants?

crates/physical-plan/src/plan.rs
```diff
@@ -1185,6 +1185,17 @@ pub struct IxScan {
...
+    /// We do have non-ranged hash indices.
+    /// This means that when we get around to using this variant,
+    /// we must change the rewrite rules such that we do not emit
+    /// [`IxScan`] on a hash index.
+    ///
+    /// Moreover, an equality scan (the variant above)
+    /// `(a0, b0)` on an index `(a, b, c)` is actually a ranged scan.
+    /// We also currently do not emit such `IxScan`s where the number
+    /// of equalities provided are fewer than the number of columns in the index.
+    /// When we do, we must also account for hash indices in the rewrite rules.
```
gefjon, line 1198: Is this note resolved in a later PR in this change? If not, we should make a ticket.

crates/core/src/estimation.rs
Centril: I don't follow, `ndv`, the divisor, is always `NonZero<_>`.

crates/physical-plan/src/plan.rs
Centril: https://github.com/clockworklabs/SpacetimeDB/issues/4026

crates/core/src/estimation.rs
gefjon: D'oh! My mistake.

Issue #1122, opened 2024-04-18
Hash Indexes


Centril:

vovacooper:
would love to see hash indexes in the spacetimedb

egormanga:
Please elaborate on “cases like `entity_id: u32` and friends”, what is the particular case?

I don't see how `entity_id` would be monotonic without `auto_inc`, but that's already only useful on primary keys (thus an implicit index).

Centril:
> I don't see how `entity_id` would be monotonic without `auto_inc`, but that's already only useful on primary keys (thus an implicit index).

You can choose what the implicit index is backed by by doing e.g., `#[primary_key] #[index(direct)]` together.

egormanga:
> #[primary_key] #[index(direct)] together.

Great, thanks! I believe it might be a bit more intuitive to be able to write `#[primary_key, index(direct)]`.

PR #3976, opened 2026-01-08
Add Hash indices

Centril:
# Description of Changes

Fixes https://github.com/clockworklabs/SpacetimeDB/issues/1122.

Adds hash indices and exposes them through `#[index(hash)]` for Rust modules,
with support for typescript and C# to come in follow ups.
On the client/sdk side, for now, any index is backed via a BTree/native index as it is the most general.

A hash index may only be queried through point scan and never ranged scans.
Attempting a ranged scan results in an error, with the mechanism implemented in the previous PR (https://github.com/clockworklabs/SpacetimeDB/pull/3974).

None

2?

# Testing

A test for ensuring that hash indices cannot be used for range scans is added.
Tests exercising hash indices will come in the next PR.

### DIFF
--- a/crates/bindings/src/table.rs
+++ b/crates/bindings/src/table.rs
@@ -566,7 +566,7 @@ impl<Tbl: Table, IndexType, Idx: IndexIsPointed> PointIndex<Tbl, IndexType, Idx>
     where
         P: WithPointArg<K>,
     {
-        filter_point::<Tbl, Idx, P, K>(point)
+        filter_point::<Tbl, Idx, K>(point)
     }
 
     /// Deletes all rows in the database state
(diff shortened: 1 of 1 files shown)


crates/bindings/src/table.rs
```diff
@@ -464,7 +465,246 @@ pub trait Index {
...
+    pub fn filter<P, K>(&self, point: P) -> impl Iterator<Item = Tbl::Row>
+    where
+        P: WithPointArg<K>,
+    {
+        filter_point::<Tbl, Idx, P, K>(point)
+    }
+}
+
+/// Trait used for running point index scans.
+pub trait WithPointArg<K = ()> {
```
gefjon, line 665: What is `K` here?

JasonAtClockwork:
> Looks reasonable. I'd appreciate better generic names in `bindings/src/table.rs`, but otherwise am good with this PR. I haven't really looked at the TypeScript, C# or Unreal codegen; ideally we should get @cloutiertyler or @coolreader18 , @rekhoff and @JasonAtClockwork respectively.

I've double-checked and the impact to the Unreal codegen seems minor here and rebuilt Unreal Blackholio to test locally and it looks good to me.

Centril:
The TS, C#, and C++ changes are very minor, as can be seen if you hide whitespace changes. They amount only to not caring what kind of index it is, treating all indices as a B-Tree, since B-Trees are "the most general". So the change is basically removing panics and using `idx.algorithm.columns()` instead of pattern matching on the index kind. Since this change is so minor, I'll take the liberty of merging the PR in the interest of time.

crates/bindings/src/table.rs
Centril: `K` predates this PR (see `IndexScanRangeBounds`) and is a trick used to deal with the orphan rule.
I couldn't come up with a replacement name that isn't super long, so I added comments in strategic places instead.

PR #3979, opened 2026-01-08
Reduce `Test Suite` test concurrency to reduce test flakes

jdetter:
# Description of Changes

This PR started as an experiment to see if restricting the amount of test job threads would reduce the flakiness in the `Test Suite` CI job. I have tested several values here (`1`, `2` and `10`) and I believe 2 is very stable. I'm also very confused with this change because it seems like somehow the `Test Suite` CI job is now faster than it was before:

Before:

image

After:

image

It's possible that having less test threads means less cpu contention on the github runners, but I'm not completely sure. It could also somehow be less lock contention in SpacetimeDB. I have looked at the output of the `Test Suite` job and it does appear that all tests are running.

None

0 - this only updates our workflow and has no impact on the code.

# Testing

- [x] CI is passing and `Test Suite` is less flakey


tools/ci/src/main.rs
```diff
@@ -127,7 +127,8 @@ fn main() -> Result<()> {
         Some(CiCmd::Test) => {
             // TODO: This doesn't work on at least user Linux machines, because something here apparently uses `sudo`?
 
-            cmd!("cargo", "test", "--all", "--", "--skip", "unreal").run()?;
+            // cmd!("cargo", "test", "--all", "--", "--skip", "unreal").run()?;
+            cmd!("cargo", "test", "--all", "--", "--test-threads=10", "--skip", "unreal").run()?;
```
jdetter:
```suggestion
            cmd!("cargo", "test", "--all", "--", "--test-threads=2", "--skip", "unreal").run()?;
```

Shubham8287 (approved):
It could be due to less contention on github runners CPU, I doubt if it has to do anything with SpacetimeDB internals.

PR #3982, opened 2026-01-09
C# client typed query builder

### LINKED ISSUE #3759: [C#] Client bindings for query builder
See the [proposal](https://github.com/clockworklabs/SpacetimeDBPrivate/pull/2356) for details.

rekhoff:
# Description of Changes
This PR implements the C# client-side typed query builder, as assigned in https://github.com/clockworklabs/SpacetimeDB/issues/3759.

Key pieces:
* Added a small C# runtime query-builder surface in the client SDK (`sdks/csharp/src/QueryBuilder.cs`):
  * `Query` (wraps the generated SQL string)
  * `Table<TRow, TCols, TIxCols>` (entry point for `All()` / `Where(...)`)
  * `Col<TRow, TValue>` and `IxCol<TRow, TValue>` (typed column references)
  * `BoolExpr` (typed boolean expression composition)
  * SQL identifier quoting + literal formatting helpers (`SqlFormat`)
  * `Join(...)` with `WhereLeft(...)` / `WhereRight(...)`
  * `LeftSemijoin(...)` / `RightSemijoin(...)` with `Where(...)` chaining
* Extended C# client bindings codegen (`crates/codegen/src/csharp.rs`) to generate:
  * Per-table/view `*Cols` and `*IxCols` helper classes used by the typed query builder.
  * A generated per-module `QueryBuilder` with a `From` accessor for each table/view, producing `Table<...>` values.
  * A generated `TypedSubscriptionBuilder` which collects `Query<TRow>.Sql` values and calls the existing subscription API.
  * An `AddQuery(Func<QueryBuilder, Query> build)` entry point off `SubscriptionBuilder`, mirroring the proposal’s Rust API.
* Fixed a codegen naming collision found during regression testing:
  * `*Cols`/`*IxCols` helpers are now named after the table/view accessor name (PascalCase) instead of the row type, since multiple tables/views can share the same row type (e.g. alias tables / views returning an existing product type).
* Kept `Cols`/`IxCols` off the public surface:
  * `Table.Cols` and `Table.IxCols` are internal, so consumers only access columns via the `Where(...)`/join predicate lambdas.

C# usage examples (mirroring the proposal’s Rust examples)
1) Typed subscription flow (no raw SQL)
```csharp
void Subscribe(SpacetimeDB.Types.DbConnection conn)
{
conn.SubscriptionBuilder()
    .OnApplied(ctx => { /* ... */ })
    .OnError((ctx, err) => { /* ... */ })
    .AddQuery(qb => qb.From.Users().Build())
    .AddQuery(qb => qb.From.Players().Build())
    .Subscribe();
}
```
2) Typed `WHERE` filters and boolean composition
```csharp
conn.SubscriptionBuilder()
    .OnApplied(ctx => { /* ... */ })
    .OnError((ctx, err) => { /* ... */ })
    .AddQuery(qb => qb.From.Players().Where(p => p.Name.Eq("alice").And(p.IsOnline.Eq(true))).Build())
    .Subscribe();
```
3) “Admin can see all, otherwise only self” (proposal’s “player” view logic, but client-side)
```csharp
Identity self = /* ... */;

conn.SubscriptionBuilder()
    .AddQuery(qb =>
        qb.From.Players().Where(p =>
            p.Identity.Eq(self)
        )
    )
    .Subscribe();
```
4) Index-column access for query construction (IxCols)
```csharp
conn.SubscriptionBuilder()
    .AddQuery(qb =>
        qb.From.Players().Where(
            qb.From.Players().IxCols.Identity.Eq(self)
        )
    )
    .Subscribe();
```
None.
* Additive client SDK runtime types.
* Additive client bindings codegen output.
* No wire-format changes.
2 - Low to moderate
* Mostly additive code + codegen.
* The main risk is correctness/compat of generated SQL strings and name/casing conventions across languages; this is mitigated by targeted unit tests + full C# regression test runs.
# Testing
- [X] Ran run-regression-tests.sh successfully after regenerating C# bindings.
- [X] Ran C# unit tests using `dotnet test sdks/csharp/tests~/tests.csproj -c Release`
- [X] Added a new unit test suite (`sdks/csharp/tests~/QueryBuilderTests.cs`) validating:
  * Identifier quoting / escaping
  * Literal formatting (including `Identity`/`ConnectionId`/`Uuid` hex literals; `U128` integer literal)
  * null + enum unsupported behavior throws
  * Boolean expression parenthesization (`And`/`Or`/`Not`)
  * `Where(...)` overloads including `IxCols`-based predicates
  * left/right semijoin SQL formatting and predicate chaining


sdks/csharp/src/QueryBuilder.cs
```diff
@@ -0,0 +1,186 @@
...
+        {
+            if (value is null)
+            {
+                return IsNull();
+            }
+
+            return new BoolExpr<TRow>($"{RefSql} = {SqlFormat.FormatLiteral(value)}");
+        }
+
+        public BoolExpr<TRow> Neq(TValue value)
```
joshua-spacetime: For the rust builder it's `ne`

sdks/csharp/src/QueryBuilder.cs
```diff
@@ -0,0 +1,186 @@
...
+            return new BoolExpr<TRow>($"{RefSql} <> {SqlFormat.FormatLiteral(value)}");
+        }
+
+        public BoolExpr<TRow> Lt(TValue value) => new($"{RefSql} < {SqlFormat.FormatLiteral(value)}");
+        public BoolExpr<TRow> Lte(TValue value) => new($"{RefSql} <= {SqlFormat.FormatLiteral(value)}");
+        public BoolExpr<TRow> Gt(TValue value) => new($"{RefSql} > {SqlFormat.FormatLiteral(value)}");
+        public BoolExpr<TRow> Gte(TValue value) => new($"{RefSql} >= {SqlFormat.FormatLiteral(value)}");
+
+        public BoolExpr<TRow> IsNull() => new($"{RefSql} IS NULL");
+        public BoolExpr<TRow> IsNotNull() => new($"{RefSql} IS NOT NULL");
```
joshua-spacetime: These aren't supported in SpacetimeSQL

sdks/csharp/src/QueryBuilder.cs
```diff
@@ -0,0 +1,186 @@
...
+        public TIxCols IxCols { get; }
+
+        public Table(string tableName, TCols cols, TIxCols ixCols)
+        {
+            this.tableName = tableName;
+            Cols = cols;
+            IxCols = ixCols;
+        }
+
+        public Query<TRow> All() => new($"SELECT * FROM {SqlFormat.QuoteIdent(tableName)}");
```
joshua-spacetime: I don't think `All` is necessary.

sdks/csharp/src/QueryBuilder.cs
```diff
@@ -0,0 +1,186 @@
...
+            return $"\"{ident.Replace("\"", "\"\"")}\"";
+        }
+
+        private static string EscapeString(string s) => s.Replace("'", "''");
+
+        public static string FormatLiteral(object? value)
+        {
+            if (value is null)
+            {
+                return "NULL";
```
joshua-spacetime: We don't support `NULL`.

sdks/csharp/src/QueryBuilder.cs
```diff
@@ -0,0 +1,186 @@
...
+                return b ? "TRUE" : "FALSE";
+            }
+
+            if (value is char c)
+            {
+                return $"'{EscapeString(c.ToString())}'";
+            }
+
+            var t = value.GetType();
+            if (t.IsEnum)
```
joshua-spacetime: We also don't currently support enum comparisons

sdks/csharp/tests~/QueryBuilderTests.cs
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    private static Table<Row, RowCols, RowIxCols> MakeTable(string tableName) =>
+        new(tableName, new RowCols(tableName), new RowIxCols(tableName));
+
+    [Fact]
+    public void All_QuotesTableName()
+    {
+        var table = MakeTable("My\"Table");
+        Assert.Equal("SELECT * FROM \"My\"\"Table\"", table.All().Sql);
```
joshua-spacetime: Same as earlier, we don't need `All()` here. We can just `.toSql()` the table ref directly.

sdks/csharp/tests~/QueryBuilderTests.cs
```diff
@@ -0,0 +1,182 @@
...
+        var table = MakeTable("T");
+        var sql = table.Where(c => c.Name.Eq("O'Reilly")).Sql;
+        Assert.Equal("SELECT * FROM \"T\" WHERE \"T\".\"Name\" = 'O''Reilly'", sql);
+    }
+
+    [Fact]
+    public void Where_Eq_Null_UsesIsNull()
+    {
+        var table = MakeTable("T");
+        var sql = table.Where(c => c.Name.Eq(null!)).Sql;
```
joshua-spacetime: We should make `Eq`'s arg not nullable.

sdks/csharp/tests~/QueryBuilderTests.cs
```diff
@@ -0,0 +1,182 @@
...
+        Assert.Equal("SELECT * FROM \"T\" WHERE \"T\".\"Name\" IS NULL", sql);
+    }
+
+    [Fact]
+    public void Where_Neq_Null_UsesIsNotNull()
+    {
+        var table = MakeTable("T");
+        var sql = table.Where(c => c.Name.Neq(null!)).Sql;
```
joshua-spacetime: Same for all of the args

joshua-spacetime (commented):
Can we add some end-to-end tests?

rekhoff:
> Can we add some end-to-end tests?

Added end-to-end coverage in `sdks/csharp/examples~/regression-tests` by converging existing SQL queries used, into the typed query builder subscription path (`AddQuery`). Also added a targeted `WHERE` case as that hadn't been covered before.

sdks/csharp/src/QueryBuilder.cs
rekhoff: Updated C# `Neq` to emit `!=` to match the subscription SQL subset / Rust builder behavior.

sdks/csharp/tests~/QueryBuilderTests.cs
rekhoff: `Eq(null)`/`Neq(null)` now throw (`ArgumentNullException`) instead of generating SQL

sdks/csharp/examples~/regression-tests/client/Program.cs
```diff
@@ -60,19 +60,19 @@ void OnConnected(DbConnection conn, Identity identity, string authToken)
...
+        .AddQuery(qb => qb.From.MyAccount().Build())
+        .AddQuery(qb => qb.From.MyAccountMissing().Build())
+        .AddQuery(qb => qb.From.PlayersAtLevelOne().Build())
+        .AddQuery(qb => qb.From.MyTable().Build())
+        .AddQuery(qb => qb.From.NullStringNonnullable().Build())
+        .AddQuery(qb => qb.From.NullStringNullable().Build())
+        .AddQuery(qb => qb.From.MyLog().Build())
+        .AddQuery(qb => qb.From.Admins().Build())
+        .AddQuery(qb => qb.From.NullableVecView().Build())
+        .AddQuery(qb => qb.From.WhereTest().Where(c => c.Value.Gt(10)))
```
joshua-spacetime: Are there any joins that we can convert? Also can you add a similar `WHERE` test but over a non-indexed column?

sdks/csharp/tests~/QueryBuilderTests.cs
```diff
@@ -0,0 +1,143 @@
...
+            "(((\"T\".\"Age\" > 1) AND (\"T\".\"Name\" <> 'x')) OR (\"T\".\"IsAdmin\" = TRUE))",
+            expr.Sql
+        );
+    }
+
+    [Fact]
+    public void QuoteIdent_EscapesDoubleQuotesInColumnName()
+    {
+        var table = MakeTable("T");
+        var sql = table.Where(c => c.Weird.Eq("x")).Sql;
```
joshua-spacetime: Could you add a left and right semijoin test as well?

sdks/csharp/tests~/QueryBuilderTests.cs
```diff
@@ -0,0 +1,143 @@
...
+            "SELECT * FROM \"T\" WHERE \"T\".\"IsAdmin\" = FALSE",
+            table.Where(c => c.IsAdmin.Eq(false)).Sql
+        );
+    }
+
+    [Fact]
+    public void BoolExpr_AndOrNot_AddsParens()
+    {
+        var table = MakeTable("T");
+        var expr = table.Cols.Age.Gt(1)
```
joshua-spacetime: Is `Cols` part of the public api? Could we make it private, so that table handles only expose the query operators?

joshua-spacetime (commented):
Looks great. Only a few small remaining requests/questions.

sdks/csharp/examples~/regression-tests/client/Program.cs
rekhoff: I had missed some items around joins initially, but that's now been added. The C# query builder now supports `Join(...)` plus `WhereLeft(...)` / `WhereRight(...)` and `.Build()`
I also added unit coverage for basic join formatting and combinations of left/right WHERE clauses.
For the `WHERE`, I added a test that exercises `WHERE` over a non-indexed column (see see `Join_Build_FormatsCorrectly`, `Join_WithLeftAndRightWhere_FormatsCorrectly`, `Join_WithWhereLeftChaining_FormatsCorrectly` in `QueryBuilderTests`)

sdks/csharp/tests~/QueryBuilderTests.cs
rekhoff: Added to `LeftSemijoin(...)` and `RightSemijoin(...)` to `QueryBuilderTests.cs`

sdks/csharp/tests~/QueryBuilderTests.cs
rekhoff: Yeah, great catch, I agree. I made `Table.Cols` and `Table.IxCols` internal, so callers can only reference columns through the `Where(...)` / join lambdas

sdks/csharp/src/Table.cs
```diff
@@ -154,6 +154,164 @@ public IEnumerable<Row> Filter(Column value) =>
                 cache.TryGetValue(value, out var rows) ? rows : Enumerable.Empty<Row>();
         }
 
+        public abstract class NullableRefUniqueIndexBase<Column>
```
joshua-spacetime: Is the code in this file new or was it moved? Just wondering why we needed to extend the client cache.

sdks/csharp/tests~/QueryBuilderTests.cs
```diff
@@ -0,0 +1,377 @@
...
+    {
+        var left = MakeLeftTable("users");
+        var right = MakeRightTable("other");
+
+        var sql = left
+            .Where(c => c.Id.Eq(1))
+            .Join(right, (l, r) => l.Id.Eq(r.Uid))
+            .WhereRight(c => c.Uid.Gt(10))
+            .Build()
+            .Sql;
```
joshua-spacetime: The other query builders don't support a generic `Join` yet because subscriptions must return rows from one and only one table. We don't support combining rows from two tables through a join.

sdks/csharp/examples~/regression-tests/client/Program.cs
```diff
@@ -60,19 +60,19 @@ void OnConnected(DbConnection conn, Identity identity, string authToken)
...
+        .AddQuery(qb => qb.From.MyAccountMissing().Build())
+        .AddQuery(qb => qb.From.PlayersAtLevelOne().Build())
+        .AddQuery(qb => qb.From.MyTable().Build())
+        .AddQuery(qb => qb.From.NullStringNonnullable().Build())
+        .AddQuery(qb => qb.From.NullStringNullable().Build())
+        .AddQuery(qb => qb.From.MyLog().Build())
+        .AddQuery(qb => qb.From.Admins().Build())
+        .AddQuery(qb => qb.From.NullableVecView().Build())
+        .AddQuery(qb => qb.From.WhereTest().Where(c => c.Value.Gt(10)).Build())
+        .Subscribe();
```
joshua-spacetime, line 87: We should add some semijoin subscriptions here for some complete end-to-end tests. I can provide examples.

sdks/csharp/examples~/regression-tests/client/Program.cs
joshua-spacetime: More machinery is required to support a generic `Join`, so I think we should hold off on this for now.

sdks/csharp/tests~/QueryBuilderTests.cs
rekhoff: This was a mistake, and has been removed. No more generic joins or cross table joins are included.

sdks/csharp/examples~/regression-tests/client/Program.cs
rekhoff: End to end tests for semi-joins have been added.

sdks/csharp/examples~/regression-tests/client/Program.cs
rekhoff: Generic join code removed

sdks/csharp/src/Table.cs
rekhoff: This was new and part of troubleshooting some issues because C# won’t let you override a method returning `T` with one returning `T?` Ultimately this wasn't actually necessary, and was part of some plumbing fixes I was working through.
Currently both Rust and C++ allow for `Option<T>` to return nullable as a valid value, but C# doesn't in many cases. I was thinking this would be a small fix, but it ended up getting messy and not worth including in this PR. Anyway, it's remove now.

joshua-spacetime (approved):

PR #3983, opened 2026-01-09
`--feature no-core-pinning` disables core pinning

Centril:
# Description of Changes

Adds a feature `no-core-pinning` that disables core pinning.

None

1

# Testing

Covered by existing tests.

### DIFF
--- a/crates/standalone/Cargo.toml
+++ b/crates/standalone/Cargo.toml
@@ -17,9 +17,11 @@ harness = true         # Use libtest harness.
 required-features = [] # Features required to build this target (N/A for lib)
 
 [features]
+unstable = ["spacetimedb-client-api/unstable"]
 # Perfmaps for profiling modules
 perfmap = ["spacetimedb-core/perfmap"]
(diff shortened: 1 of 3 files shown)


crates/standalone/Cargo.toml
```diff
@@ -17,9 +17,11 @@ harness = true         # Use libtest harness.
 required-features = [] # Features required to build this target (N/A for lib)
 
 [features]
+unstable = ["spacetimedb-client-api/unstable"]
 # Perfmaps for profiling modules
 perfmap = ["spacetimedb-core/perfmap"]
-unstable = ["spacetimedb-client-api/unstable"]
+# Disables core pinning
+no-core-pinning = ["spacetimedb-core/no-core-pinning"]
```
kim, line 24: There's more core pinning going on that just `JobCores` (in `main`), so here I'd suggest to name this feature `no-job-core-pinning` or something.

Centril: I changed the implementation instead to actually disable all core pinning.

PR #3984, opened 2026-01-09
Handle database states better when serving module logs

kim:
If a database is not currently running / runnable, we can still serve module logs if, and only if, the routing layer has correctly picked the last known leader.

To determine that this is the case, we make it so a predicate can be evaluated against the error return of `NodeDelegate::leader`.

In the future, we may be able to fetch from the last known leader transparently, or store module logs in distributed storage altogether, at which point the predicate trait can be retired.

In order to not leak information about the server's directory structure, `DatabaseLogger::read_latest_on_disk` will return an empty stream in case no log directory or recent log file exist on disk.

1.5

# Testing

n/a

### DIFF
--- a/crates/client-api/src/lib.rs
+++ b/crates/client-api/src/lib.rs
@@ -34,6 +34,17 @@ pub mod util;
 /// surfaced to the API.
 #[async_trait]
 pub trait NodeDelegate: Send + Sync {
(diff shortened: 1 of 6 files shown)


kim:
Stacked on top of #3961

PR #3986, opened 2026-01-09
Expose a couple things to enable some work in another repo

gefjon:
# Description of Changes

Just adds public accessors for some existing internal functionality, to enable a change in another repo. Review starting from there.

These crates are marked unstable.

0.5

# Testing

See other PR.


crates/datastore/src/locking_tx_datastore/datastore.rs
```diff
@@ -1006,6 +1006,10 @@ impl<F> Replay<F> {
     pub fn next_tx_offset(&self) -> u64 {
         self.committed_state.read_arc().next_tx_offset
     }
+
+    pub fn committed_state(&self) -> &Arc<RwLock<CommittedState>> {
```
kim: I feel uncomfortable leaking the ability to acquire a write lock here.

How about this:

- return `RwLockGuard<'_, CommittedState>` here (or `impl Deref<Target = CommittedState> + '_` if you prefer)
- remove `impl Decoder for Replay` below
- keep only `impl Decoder for &mut Replay`

That should prevent any (accidental) misuse.

kim: Comme ça: https://github.com/clockworklabs/SpacetimeDB/compare/phoebe/cutlery-companion...kim/rdonly-committed-state-from-replay

gefjon: Wanna push that commit to this branch?

kim: Done

PR #3987, opened 2026-01-09
fix(codegen/unreal): eliminate UObject memory leak in reducer event handlers

brougkr:
## Summary

- Fixes critical memory leak in Unreal C++ generated code where `NewObject<UReducer>()` is called for every reducer event but never cleaned up
- At 30Hz server tick rate (common for game servers), this creates ~30 orphaned UObjects per second, causing continuous memory growth
- Solution: Pass stack-allocated `FArgs` struct directly via new `InvokeWithArgs()` method instead of allocating UObjects

## Problem

The generated `ReducerEvent` handler in `SpacetimeDBClient.g.cpp` creates a new UObject for every reducer event:

```cpp
FArgs Args = ReducerEvent.Reducer.GetAs...();
UReducer* Reducer = NewObject<UReducer>();  // LEAK - never cleaned up!
Reducer->Param = Args.Param;
Reducers->Invoke...(Context, Reducer);
```

Unreal's garbage collector cannot keep up at high frequencies. Attempts to fix with `MarkAsGarbage()`, `ConditionalBeginDestroy()`, and transient package flags all failed because GC timing is unpredictable.

## Solution

Eliminate UObject allocation entirely by passing `FArgs` directly:

```cpp
FArgs Args = ReducerEvent.Reducer.GetAs...();
Reducers->Invoke...WithArgs(Context, Args);  // Zero allocation!
```

The original `Invoke()` methods that take `UReducer*` are preserved for backwards compatibility.

## Changes

1. **Header generation** (~line 2515): Add `InvokeWithArgs()` method declaration
2. **ReducerEvent handler** (~line 3194): Call `InvokeWithArgs()` instead of creating UObject
3. **Implementation** (~line 3703): Add `InvokeWithArgs()` that extracts params from `FArgs` struct

## Test Plan

- [x] Verified with 30Hz `TickGameScheduled` reducer - UObject count remains stable
- [x] Memory growth eliminated (was ~1000 UObjects per 14 seconds)
- [x] Backwards compatibility preserved - existing `Invoke()` methods still work

## Before/After

**Before**: `OBJS` count in Unreal stats continuously increasing, memory growing unbounded
**After**: `UObjDelta=0` - no UObject growth from reducer events


JasonAtClockwork:
Thanks again for this PR I think it looks good but running into issues on my device to properly test it, I should have this confirmed on Monday!

PR #3988, opened 2026-01-09
Additional Hex parsing checks to `ConnectionId` and `Identity`

rekhoff:
# Description of Changes
This PR fixes Release-only flakiness in C# runtime/property tests by making `ConnectionId` / `Identity` parsing and related low-level byte decoding strict about input length.

Previously, length validation relied on `Debug.Assert`, which is compiled out in Release builds. That meant malformed inputs (wrong-length spans/strings) could slip through in Release and produce inconsistent behavior.

Changes:

* Updated `ConnectionId.FromHexString(string hex)` to throw when the input is not exactly 32 hex characters.
* Updated `Identity.FromHexString(string hex)` to throw when the input is not exactly 64 hex characters.
* Updated `Util.Read(ReadOnlySpan source, bool littleEndian)` to throw an `ArgumentException` when `source.Length != sizeof(T)` (instead of relying on `Debug.Assert`), preventing Release-mode “partial reads” from incorrectly-sized spans.
None.
* No schema or wire-format changes.
* Behavior change is limited to stricter validation of invalid/malformed inputs for `ConnectionId` / `Identity` parsing and internal runtime decoding. Valid inputs are unaffected.
2 - Low
* Changes are isolated to the C# BSATN runtime input validation paths.
* No impact on valid serialization/normal runtime behavior; only affects invalid inputs and makes behavior consistent between Debug and Release.
# Testing
- [X] Ran C# BSATN runtime tests locally (Debug + Release) and verified they pass.

### DIFF
--- a/crates/core/src/host/scheduler.rs
+++ b/crates/core/src/host/scheduler.rs
@@ -11,7 +11,7 @@ use core::time::Duration;
 use futures::{FutureExt, StreamExt};
 use rustc_hash::FxHashMap;
 use spacetimedb_client_api_messages::energy::EnergyQuanta;
-use spacetimedb_datastore::execution_context::Workload;
+use spacetimedb_datastore::execution_context::{ExecutionContext, ReducerContext, Workload};
 use spacetimedb_datastore::locking_tx_datastore::MutTxId;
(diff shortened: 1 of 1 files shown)


jdetter (approved):
Excellent, thank you Ryan! I haven't tested this myself but I originally reviewed these changes as part of #3967

PR #3991, opened 2026-01-10
Make new `Uuid` type Filterable

kistz:
# Description of Changes

`Uuid` addad in the last update is not filterable even though it is `Copy` and (imo) obviously _very_ suited to filter.
Promote it to a special `FilterableValue` in the same way as `Identity`  and adjust the docs.

None

1. Only copies the same strategy as `Identity` which is also a newtype wrapper around a integer like `Uuid`

# Testing

Tested on my project.

After the change:
image

Before the change:
image

You can also test this yourself when adding:
`spacetimedb = { version = "1.11.2", git = "https://github.com/kistz/SpacetimeDB.git", branch = "uuid-as-filter" }` to your cargo.toml project

- [x] Tested before and after. Before it was a compiler error afterwards not.

### DIFF
--- a/crates/bindings/tests/ui/tables.stderr
+++ b/crates/bindings/tests/ui/tables.stderr
@@ -121,10 +121,10 @@ error[E0277]: `&'a Alpha` cannot appear as an argument to an index filtering ope
   --> tests/ui/tables.rs:32:33
    |
 32 |     ctx.db.delta().compound_a().find(Alpha { beta: 0 });
-   |                                 ^^^^ should be an integer type, `bool`, `String`, `&str`, `Identity`, `ConnectionId`, `Hash` or a no-payload enum which derives `SpacetimeType`, not `&'a Alpha`
+   |                                 ^^^^ should be an integer type, `bool`, `String`, `&str`, `Identity`, `Uuid`, `ConnectionId`, `Hash` or a no-payload enum which derives `SpacetimeType`, not `&'a Alpha`
    |
    = help: the trait `for<'a> FilterableValue` is not implemented for `&'a Alpha`
-   = note: The allowed set of types are limited to integers, bool, strings, `Identity`, `ConnectionId`, `Hash` and no-payload enums which derive `SpacetimeType`,
+   = note: The allowed set of types are limited to integers, bool, strings, `Identity`, `Uuid`, `ConnectionId`, `Hash` and no-payload enums which derive `SpacetimeType`,
    = help: the following other types implement trait `FilterableValue`:
              &ConnectionId
              &Identity
--- a/crates/lib/src/filterable_value.rs
+++ b/crates/lib/src/filterable_value.rs
@@ -38,7 +38,7 @@ use spacetimedb_sats::{hash::Hash, i256, u256, Serialize};
 //   E.g. `&str: FilterableValue<Column = String>` is desirable.
 #[diagnostic::on_unimplemented(
     message = "`{Self}` cannot appear as an argument to an index filtering operation",
-    label = "should be an integer type, `bool`, `String`, `&str`, `Identity` , `Uuid`, `ConnectionId`, `Hash` or a no-payload enum which derives `SpacetimeType`, not `{Self}`",
+    label = "should be an integer type, `bool`, `String`, `&str`, `Identity`, `Uuid`, `ConnectionId`, `Hash` or a no-payload enum which derives `SpacetimeType`, not `{Self}`",
     note = "The allowed set of types are limited to integers, bool, strings, `Identity`, `Uuid`, `ConnectionId`, `Hash` and no-payload enums which derive `SpacetimeType`,"
 )]
 pub trait FilterableValue: Serialize + Private {


cloutiertyler:
Thank you for the submission. Unfortunately this is not sufficient to be able to merge this. We would need to implement this in:

- `spacetime generate` so the client can also filter
- C# (server/client)
- TypeScript (server/client)

It's important that we keep the various languages consistent with the features they support except in rare scenarios.

kistz:
I can look into it today and/or tomorrow and let you know if i can get it to work :)
If not we can view this as an issue and close it/let someone from the team finish it 👍

kistz:
@gefjon Thank you again for the list you provided in #4024 

I believe Uuid was designed for this in the first place and no further changes are necessary in this case. Hence only Rust cant `.filter()` it at the moment cause of the minor trait bound missing.

Please let me know if the following argumentation is lacking and/or there is something missing still:

> Ensure (with automated tests) can equality-compare and ordering-compare values of the column type

1. C# [Ordering](https://github.com/clockworklabs/SpacetimeDB/blob/38ee9e89ba2cb64be2d8edd56cbc237fbee0adbb/crates/bindings-csharp/BSATN.Runtime.Tests/Tests.cs#L936) [Comparable Decl](https://github.com/clockworklabs/SpacetimeDB/blob/38ee9e89ba2cb64be2d8edd56cbc237fbee0adbb/crates/bindings-csharp/BSATN.Runtime/BSATN/Uuid.cs#L13) [Comparable Impl](https://github.com/clockworklabs/SpacetimeDB/blob/38ee9e89ba2cb64be2d8edd56cbc237fbee0adbb/crates/bindings-csharp/BSATN.Runtime/BSATN/Uuid.cs#L294)

2.TS [Ordering](https://github.com/clockworklabs/SpacetimeDB/blob/38ee9e89ba2cb64be2d8edd56cbc237fbee0adbb/crates/bindings-typescript/tests/uuid.test.ts#L66) [Comparable](https://github.com/clockworklabs/SpacetimeDB/blob/38ee9e89ba2cb64be2d8edd56cbc237fbee0adbb/crates/bindings-typescript/tests/uuid.test.ts#L17)

3.RS [Ordering and Equality](https://github.com/clockworklabs/SpacetimeDB/blob/38ee9e89ba2cb64be2d8edd56cbc237fbee0adbb/crates/sats/src/uuid.rs#L24)

> Extend our other supported module languages' bindings libraries.

1. C# already can do this (Thanks @Lethalchip for testing!) 
image
image

2. TS already can do this.
image

3. Rust can also do it after this pr ;)


As i said before please let me know if this is still to little confidence in the pr preferably with concrete todo items :> 
I would be really motivated to get this in ^^

Thanks for reading and considering guys <3

gefjon:
Oh, you're right, everything is done except the impl. Thanks for checking that all out for us.

bfops:
Hey @kistz , it looks like some of the tests are failing - can you investigate?

kistz:
Yes i can investigate (probably have time somewhen on friday) 👍

bfops:
Excellent, thank you :raised_hands:

kistz:
@bfops got to it sooner than expected :>

I think the test failling is because of a expected change in the ui tests.
The error message should change to include Uuid in this case.

Im not sure what action i must take in this case ^^ e.g. if im responsible to update the UI tests or which command to use in that cas 🥴 Some info would be useful or in case where you can do it feel free to do so 👍 

Expected: 
image

Actual:
image


Here is also the git diff when looking at the failed ui diff:
```
diff --git a/./test1.txt b/./test2.txt
index 137fa10f..9e569984 100644
--- a/./test1.txt
+++ b/./test2.txt
@@ -121,10 +121,10 @@ error[E0277]: `&'a Alpha` cannot appear as an argument to an index filtering ope
   --> tests/ui/tables.rs:32:33
    |
 32 |     ctx.db.delta().compound_a().find(Alpha { beta: 0 });
-   |                                 ^^^^ should be an integer type, `bool`, `String`, `&str`, `Identity`, `ConnectionId`, `Hash` or a no-payload enum which derives `SpacetimeType`, not `&'a Alpha`
+   |                                 ^^^^ should be an integer type, `bool`, `String`, `&str`, `Identity` , `Uuid`, `ConnectionId`, `Hash` or a no-payload enum which derives `SpacetimeType`, not `&'a Alpha`
    |
    = help: the trait `for<'a> FilterableValue` is not implemented for `&'a Alpha`
-   = note: The allowed set of types are limited to integers, bool, strings, `Identity`, `ConnectionId`, `Hash` and no-payload enums which derive `SpacetimeType`,
+   = note: The allowed set of types are limited to integers, bool, strings, `Identity`, `Uuid`, `ConnectionId`, `Hash` and no-payload enums which derive `SpacetimeType`,
    = help: the following other types implement trait `FilterableValue`:
              &ConnectionId
              &Identity
```

Also as screenshot with highlithing: 
image

gefjon:
@kistz when running the test locally, somewhere in the tree you should get a new file with the new output. I *think*, but am not confident, that in your case this will be `crates/bindings/tests/ui/tables.stderr.???`, where `???` is some suffix I can't remember. `tables.stderr` without suffix is the "expected" test output. If you rename the new file to `tables.stderr`, overwriting the previous expected output, and commit and push the change, that should make the test pass.

kistz:
@gefjon should be hopefully good now!

Thanks a lot to zeke, tyler and you for guiding me through this 👍 :>

Have a good day ahead!

PR #4000, opened 2026-01-12
Add smoketests for template generation

drogus:
# Description of Changes

Basic tests for templates, the test tries to generate each of the template, build and publish the SpacetimeDB part, and build or type check the client side.

This PR also includes bindings for all of the clients as it's needed for the tests to work and given the choices to either generate during the test run or include the bindings in git we chose the latter. This makes it consistent across the templates as some of them already included bindings.

1


cloutiertyler (approved):
Risk of merging is quite low as this is expanding our testing. Most considerable risk is to the testing infra itself.

bfops (approved):
my only code-owned file just has a whitespace diff, so.. LGTM :)

PR #4002, opened 2026-01-12
Unreal SDK failing to handle fragmented messages

JasonAtClockwork:
# Description of Changes

Updated the handler from OnRawMessage to OnBinaryMessage as OnRawMessage seems to have a bug that won't report the BytesRemaining for incomplete messages. OnBinaryMessage properly returns a bool flag like how the C# implementation works.

N/A

2 - Underlying framework change on how messages are handled

# Testing

Blackholio was no longer working and is back in working order; however Unreal test framework has been broken due to the work on Result<> as the companion PR has not been merged.

- [x] Rebuilt and ran Unreal Blackholio


brougkr:
Found this as well. Check my PR here, I did something different to fix it in my project. Is working reliably:

https://github.com/clockworklabs/SpacetimeDB/pull/4006

rekhoff (approved):
Code looks clean and targets the root problem with minimal changes. Tested Blackholio locally and confirmed no errors found.

PR #4003, opened 2026-01-12
Rust: client query builder

Shubham8287:
# Description of Changes
Client Query builder for rust, as per proposal - https://github.com/clockworklabs/SpacetimeDBPrivate/pull/2356.

1. Pach moves query builder to its separate crate, so that it can be shared between module and sdk.
2. Implements `TypedSubscriptionBuilder` in `sdks/rust` as mentioned in proposal
3. Modify codegen to extend types to support query builder as mentioned in proposal
4. a test

NA, additive changes.

2

# Testing
Added a test.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ concurrency:
 
 jobs:
   docker_smoketests:
-    needs: [lints, llm_ci_check]
+    needs: [lints]
     name: Smoketests
     strategy:
       matrix:
@@ -121,7 +121,7 @@ jobs:
         run: docker compose -f .github/docker-compose.yml down
 
   test:
-    needs: [lints, llm_ci_check]
+    needs: [lints]
     name: Test Suite
     runs-on: spacetimedb-new-runner
     container:
@@ -534,7 +534,7 @@ jobs:
         run: llm_benchmark ci-check
 
   unity-testsuite:
-    needs: [lints, llm_ci_check]
+    needs: [lints]
     # Skip if this is an external contribution.
     # The license secrets will be empty, so the step would fail anyway.
     if: ${{ github.event_name != 'pull_request' || !github.event.pull_request.head.repo.fork }}
@@ -657,7 +657,7 @@ jobs:
           UNITY_SERIAL: ${{ secrets.UNITY_SERIAL }}
 
   csharp-testsuite:
-    needs: [lints, llm_ci_check]
+    needs: [lints]
     runs-on: spacetimedb-new-runner
     container:
       image: localhost:5000/spacetimedb-ci:latest
@@ -767,7 +767,7 @@ jobs:
 
   internal-tests:
     name: Internal Tests
-    needs: [lints, llm_ci_check]
+    needs: [lints]
     # Skip if not a PR or a push to master
     # Skip if this is an external contribution. GitHub secrets will be empty, so the step would fail anyway.
     if: ${{ (github.event_name == 'pull_request' || (github.event_name == 'push' && github.ref == 'refs/heads/master'))


sdks/rust/tests/view-client/src/main.rs
```diff
@@ -131,6 +133,58 @@ fn exec_anonymous_subscribe() {
     test_counter.wait_for_all();
 }
 
+fn exec_anonymous_subscribe_with_query_builder() {
```
Shubham8287, line 137: The test is failing with below error, saying table "player_level" table not exist which is obviously not the case, It works if I modify test to not do semijoins. I don't know enough about how subscription works in sdk, so maybe,, something to do with that?

```
Subscription errored: SubscriptionError { error: "`player_level` is not a valid table, executing: `SELECT \"player\".* FROM \"player_level\" JOIN \"player\" ON \"player_level\".\"entity_id\" = \"player\".\"entity_id\" WHERE (\"player_level\".\"level\" = 0)`" }
note: run with `RUST_BACKTRACE=1` 
```

Shubham8287: It does not look related to query_builder but it seems `Join`s are not working in subscription this way?

joshua-spacetime: Isn't `player_level` private?

Shubham8287: yes :)

crates/query-builder/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "spacetimedb-query-builder"
+version.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+license-file = "LICENSE"
+description = "Rust query builder for SpaceTimeDB"
```
joshua-spacetime:
```suggestion
description = "Rust query builder for SpacetimeDB"
```

sdks/rust/src/subscription.rs
```diff
@@ -256,6 +257,34 @@ impl<M: SpacetimeModule> SubscriptionBuilder<M> {
             })
             .unwrap();
     }
+
+    pub fn add_query<T>(self, build: impl Fn(M::QueryBuilder) -> Query<T>) -> TypedSubscriptionBuilder<M> {
```
joshua-spacetime, line 261:
```suggestion
    /// Add a query to this subscription using the typed query builder.
    pub fn add_query<T>(self, build: impl Fn(M::QueryBuilder) -> Query<T>) -> TypedSubscriptionBuilder<M> {
```

sdks/rust/tests/view-client/src/main.rs
```diff
@@ -131,6 +133,58 @@ fn exec_anonymous_subscribe() {
...
+                ctx.reducers()
+                    .insert_player(Identity::from_byte_array([4; 32]), 0)
+                    .unwrap();
+                ctx.reducers()
+                    .delete_player(Identity::from_byte_array([4; 32]))
+                    .unwrap();
+            })
+            .add_query(|ctx| {
+                ctx.from
+                    .player_level()
```
joshua-spacetime, line 179: Could you also add a test that queries a view?

crates/codegen/src/rust.rs
```diff
@@ -66,6 +66,17 @@ impl __sdk::InModule for {type_name} {{
 ",
         );
 
+        // Do not implement query col types for nested types.
+        // as querying is only supported on top-level table row types.
+        let name = type_ref_name(module, typ.ty);
+        if let Some(table) = module
+            .tables()
```
joshua-spacetime, line 73: We also should do the same for views.

crates/codegen/src/rust.rs
```diff
@@ -608,6 +621,147 @@ impl {func_name} for super::RemoteProcedures {{
     }
 }
 
+fn implement_query_col_types(module: &ModuleDef, out: &mut impl Write, table: &TableDef) -> fmt::Result {
```
joshua-spacetime: Here too. We should do the same for `ViewDef`.

sdks/rust/tests/view-client/src/main.rs
egormanga: * Related: #2830

sdks/rust/tests/view-client/src/main.rs
Shubham8287: added below

crates/query-builder/Cargo.toml
joshua-spacetime:
```suggestion
description = "Rust query builder for SpacetimeDB"
```

joshua-spacetime:
> Can you update the Rust sdk-test clients to use the query builder rather than raw SQL strings?

I don't mind if this is done in a different patch.

cloutiertyler (approved):
Approving the LICENSE file addition

PR #4005, opened 2026-01-12
Fix logic for ipv6 connections in is_port_available

jdetter:
# Description of Changes

- Small fix for checking to see if a port is available on some given interface.

updated:

The original implementation here used `bind` to try to discover if a port is currently in use. This isn't reliable due to platform differences - especially on windows where it's apparently acceptable to have a service running on both `0.0.0.0:3000` and `127.0.0.1:3000`. This would cause bind to return successfully when we wanted it to fail. Also: binding on an ipv6 interface when a machine doesn't have ipv6 enabled caused random errors and it was too unreliable to be useful.

This new implementation uses `get_socket_info` which returns info on all sockets in use on the system. We can then look through this list to find services which conflict with our requested port.

updated 1/14:

This PR now includes a fix for flaky CLI tests. Originally we were using `find_free_port` to pick a random free port, but that was causing a race condition which resulted in test flakes. This PR fixes this issue by using `127.0.0.1:0` as the listen addr so the kernel will automatically pick a free port for us.

None

1 - this is a pretty isolated check, unlikely to introduce larger issues.

# Testing

I tested on macos, windows and linux:
```
ALLOW
docusaurus is already running on   127.0.0.1:3000
SpacetimeDB then tries to start on 192.168.1.10:3000

ALLOW
docusaurus is already running on   ::1:3000
SpacetimeDB then tries to start on 192.168.1.10:3000

DENY
docusaurus is already running on   ::1:3000
SpacetimeDB then tries to start on 127.0.0.1:3000

DENY
docusaurus is already running on   0:0:0:0:0:0:0:0:3000
SpacetimeDB then tries to start on 0.0.0.0:3000

DENY
docusaurus is already running on   0:0:0:0:0:0:0:0:3000
SpacetimeDB then tries to start on 127.0.0.1:3000

DENY
docusaurus is already running on   0:0:0:0:0:0:0:0:3000
SpacetimeDB then tries to start on 192.168.1.10:3000

DENY
docusaurus is already running on   127.0.0.1:3000
SpacetimeDB then tries to start on 0:0:0:0:0:0:0:0:3000

DENY
docusaurus is already running on   192.168.1.10:3000
SpacetimeDB then tries to start on 0:0:0:0:0:0:0:0:3000
```


crates/standalone/src/subcommands/start.rs
```diff
@@ -297,26 +297,42 @@ pub async fn exec(args: &ArgMatches, db_cores: JobCores) -> anyhow::Result<()> {
...
+    }
 
-    ipv4_available && ipv6_available
+    if let Some(ipv4_addr) = ipv4_addr {
+        if !StdTcpListener::bind(ipv4_addr).is_ok() {
+            return false;
+        }
+    }
+    
+    return StdTcpListener::bind(ipv6_addr).is_ok();
```
egormanga: Does this leave any way not to listen on either protocol?

jdetter: No, this is just doing a bind to check if the port is open and then releasing the bind. The final bind happens later.

crates/standalone/src/subcommands/start.rs
```diff
@@ -296,27 +296,68 @@ pub async fn exec(args: &ArgMatches, db_cores: JobCores) -> anyhow::Result<()> {
...
+            Err(_) => {
+                return false;
+            }
+        },
     };
 
-    let ipv4_addr = SocketAddr::from((ipv4, port));
-    let ipv6_addr = SocketAddr::V6(SocketAddrV6::new(ipv6, port, 0, 0));
+    // localhost: require both v4 loopback and v6 loopback free
+    let v4_addr = SocketAddr::from((Ipv4Addr::LOCALHOST, port));
```
cloutiertyler: Why are we checking localhost and not the address that the user specified?

This doesn't make sense to me.

jdetter: We also check the address the user specified below. If it's an ipv6 address we check the ipv6 address, if it's an ipv4 address we check both the ipv4 address and the mapped ipv6 address

jdetter: we discussed this offline, we'll make this check less restrictive.

crates/standalone/src/subcommands/start.rs
```diff
@@ -295,28 +298,127 @@ pub async fn exec(args: &ArgMatches, db_cores: JobCores) -> anyhow::Result<()> {
...
-    } else {
-        // For specific IPs, use the IPv4-mapped IPv6 address
-        ipv4.to_ipv6_mapped()
+pub fn is_port_available(host: &str, port: u16) -> bool {
+    let requested = match parse_host(host) {
+        Some(r) => r,
+        None => return false, // invalid host string => treat as not available
+    };
+
+    let sockets = match get_sockets_info(AddressFamilyFlags::IPV4 | AddressFamilyFlags::IPV6, ProtocolFlags::TCP) {
```
jdetter, line 305: This grabs the socket info from the kernel instead of manually trying to bind to the port on the ipv4/ipv6 interfaces.

cloutiertyler (approved):
This now looks good to me

PR #4006, opened 2026-01-12
fix(unreal): Replace FTickableGameObject with FTSTicker and add compressed message buffering

brougkr:
## Summary
This PR fixes two issues in the Unreal SDK:

1. **FTickableGameObject initialization order bug** - Replaced with FTSTicker for reliable tick registration
2. **Fragmented compressed WebSocket message handling** - Added buffering for messages that span multiple frames

### Issue 1: FTickableGameObject
FTickableGameObject registers itself in its constructor BEFORE UDbConnectionBase's constructor body runs. Even with ETickableTickType::Never, UE's GENERATED_BODY() macro can interfere with base class initialization order, causing the default constructor to be called instead.

### Issue 2: Fragmented Messages
Large compressed messages from SpacetimeDB can arrive fragmented across multiple WebSocket frames. The current implementation assumes complete messages arrive in single frames, causing decompression failures.

### FTSTicker Replacement
- Removed FTickableGameObject inheritance
- Added FTSTicker::FDelegateHandle for manual tick registration
- Added destructor to clean up ticker registration
- Added OnTickerTick() method

### Compressed Message Buffering
- Added CompressedBufferMutex to protect accumulation buffer
- Added IncompleteCompressedBuffer for fragment accumulation
- Added BufferedCompressionType to track compression across fragments
- Enhanced gzip validation to detect incomplete data and trigger buffering

## Changes
- `Public/Connection/DbConnectionBase.h` - Class declaration changes, new members
- `Private/Connection/DbConnectionBase.cpp` - Implementation of buffering and ticker system

## Testing
Tested in production Unreal 5.7 project with SpacetimeDB backend. Fixes previously observed:
- Random tick registration failures on connection
- Decompression errors on large subscription payloads

### DIFF
--- a/sdks/unreal/src/SpacetimeDbSdk/Source/SpacetimeDbSdk/Private/Connection/DbConnectionBase.cpp
+++ b/sdks/unreal/src/SpacetimeDbSdk/Source/SpacetimeDbSdk/Private/Connection/DbConnectionBase.cpp
@@ -1,6 +1,7 @@
 #include "Connection/DbConnectionBase.h"
 #include "Connection/DbConnectionBuilder.h"
 #include "Connection/Credentials.h"
+#include "Containers/Ticker.h"
 #include "ModuleBindings/Types/ClientMessageType.g.h"
 #include "ModuleBindings/Types/SubscribeMultiType.g.h"
 #include "ModuleBindings/Types/UnsubscribeMultiType.g.h"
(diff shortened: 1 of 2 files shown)


JasonAtClockwork:
Thanks again @brougkr! 

Just FYI I've finally come back to this and split out to a new PR with just the changes for the tick as we had the other fix #4002 to handle the fragmentation a little deeper in the SDK. I'm closing this PR and will lead with #4835 to get this merged soon.

PR #4015, opened 2026-01-12
Made the `spacetime dev` `database` argument positional

egormanga:
# Description of Changes

As caught live on Tyler's stream.

Feel free to discuss the design and accept/decline at will :)

# API and ABI breaking changes

CLI argument breaking change (`--database` is no longer accepted).

# Expected complexity level and risk

### 1.

# Testing

- [x] Manual testing

PR #4017, opened 2026-01-13
Fixes basic issues using the basic-react template.

cloutiertyler:
# Description of Changes

- Made spacetime dev <database> a positional argument and deprecated --database
- Fixed double connection in React SDK
- Added a more descriptive error message to unresolved table name.

Deprecates `--database`. Still works, but it prints with a warning.

2

# Testing

- [x] I have tested that the double render fix works in React


egormanga:
> Made spacetime dev a positional argument and deprecated --database

It's not in the PR currently.

crates/cli/src/subcommands/dev.rs
```diff
@@ -37,9 +37,15 @@ pub fn cli() -> Command {
         .about("Start development mode with auto-regenerate client module bindings, auto-rebuild, and auto-publish on file changes.")
         .arg(
             Arg::new("database")
-                .long("database")
```
egormanga, line 40: You should also [change the docs](https://github.com/clockworklabs/SpacetimeDB/pull/4015/changes#diff-29ba1b53a0e6d5b0db4a9be88af9a1839885cc878d7e12cc494e5b28bbfb84a9R218)

bfops (approved):
Well that seems like a headache to track down. Fix LGTM if it fixes something for you (well, once the CLI docs are fixed).

bfops:
re-running failed testsuite

PR #4019, opened 2026-01-13
Time out database shutdown

kim:
It is possible that, under pathological conditions, a database has a
huge transaction backlog, or that there is some bug that prevents
progress on draining this backlog upon shutdown.

In order to avoid piling up `exit_module_host` tasks (which we would not
notice), impose a timeout to be specified after which `exit_module_host`
will drop resources without waiting for the shutdown to complete
gracefully.

### DIFF
--- a/crates/durability/src/imp/local.rs
+++ b/crates/durability/src/imp/local.rs
@@ -12,13 +12,14 @@ use std::{
 use futures::{FutureExt as _, TryFutureExt as _};
 use itertools::Itertools as _;
 use log::{info, trace, warn};
+use scopeguard::ScopeGuard;
 use spacetimedb_commitlog::{error, payload::Txdata, Commit, Commitlog, Decoder, Encode, Transaction};
 use spacetimedb_fs_utils::lockfile::advisory::{LockError, LockedFile};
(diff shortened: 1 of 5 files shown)


crates/durability/src/imp/local.rs
```diff
@@ -357,6 +366,8 @@ impl<T: Send + Sync + 'static> Durability for Local<T> {
                     Ok(())
                 })
                 .await;
+            // Don't abort if we completed normally.
+            let _ = ScopeGuard::into_inner(abort);
```
Shubham8287, line 370: Does ìnto_inner` consume the guard, hence not run it?

Shubham8287: ok, that's what documentations say.

egormanga:
Would this determinism cost potential data loss?

kim:
@egormanga SpacetimeDB allows you to pretend that all transactions will eventually reach the disk, but there is no such guarantee -- some number of recent transactions can be lost if the database crashes for any reason. If you enable confirmed reads (`DBConnectionBuilder::with_confirmed_reads` for the Rust SDK), you will only see transactions that are durable.

PR #3980, opened 2026-01-08
Build all of the typescript templates in CI

jsdt:
This also has a few minor changes to fix build errors for the `test-app`.

# Description of Changes

Updates the typescript-test CI job to build some packages that weren't being built before.

This also updates the root-level `pnpm generate/format/lint/build` commands to also apply to templates.

# Expected complexity level and risk

1

PR #4021, opened 2026-01-13
Add a typed query builder for the typescript client

jsdt:
# Description of Changes

This moves the query builder code out of the `server` package and into `lib` so it can be shared by the client and server.

I put the query builder in the `index.ts` of module bindings as a `query` object that can be imported. The typescript `test-app` has an example of using it with the subscription builder.

This is branched off of `https://github.com/clockworklabs/SpacetimeDB/pull/3980`.

This extends the client subscription builder API to allow `string | RowTypedQuery<any, any>`, so existing client code should be fine.

1.5. This is low risk.

# Testing

I manually tested that the test app still works locally.

### DIFF
--- a/crates/bindings/src/lib.rs
+++ b/crates/bindings/src/lib.rs
@@ -51,8 +51,8 @@ pub use spacetimedb_lib::Uuid;
 pub use spacetimedb_primitives::TableId;
 pub use sys::Errno;
 pub use table::{
-    AutoIncOverflow, RangedIndex, RangedIndexReadOnly, Table, TryInsertError, UniqueColumn, UniqueColumnReadOnly,
-    UniqueConstraintViolation,
+    AutoIncOverflow, PointIndex, PointIndexReadOnly, RangedIndex, RangedIndexReadOnly, Table, TryInsertError,
+    UniqueColumn, UniqueColumnReadOnly, UniqueConstraintViolation,
 };
 
 pub type ReducerResult = core::result::Result<(), Box<str>>;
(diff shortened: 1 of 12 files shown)


crates/bindings-typescript/src/lib/query.ts
```diff
@@ -0,0 +1,778 @@
...
+type From<TableDef extends TypedTableDef> = Readonly<{
+  where(
+    predicate: (row: RowExpr<TableDef>) => BooleanExpr<TableDef>
+  ): From<TableDef>;
+  rightSemijoin<RightTable extends TypedTableDef>(
+    other: TableRef<RightTable>,
+    on: (
+      left: IndexedRowExpr<TableDef>,
+      right: IndexedRowExpr<RightTable>
+    ) => EqExpr<TableDef | RightTable>
```
joshua-spacetime, line 66: This is still too permissive, right? Is it possible to restrict it?

joshua-spacetime (commented):
Do we have tests for the typescript sdk that we could potentially extend to exercise the query builder?

crates/bindings-typescript/src/lib/query.ts
jsdt: What are you concerned about this allowing? The return type of `EqExpr` doesn't enforce that columns are indexed, but the `left` and `right` parameters only have indexed columns, so anyone constructing queries the normal way won't run into issues.

jsdt:
I added another version of the query builder tests that uses the test-app bindings in [client_query.test.ts](https://github.com/clockworklabs/SpacetimeDB/pull/4021/files#diff-d80d8e528809486302b583ca72ae54a864573b147318a626fb6182dbe86a7bdf)

crates/bindings-typescript/src/lib/query.ts
joshua-spacetime: No concern. I missed the `IndexedRowExpr`.

PR #4022, opened 2026-01-13
CI - "Fix" v8 error in TypeScript CI

bfops:
# Description of Changes

See https://github.com/clockworklabs/SpacetimeDB/pull/3921 for a previous version of this PR. Just expanding that hacky fix.

CI-only change.

1

# Testing

- [x] Build fixed in https://github.com/clockworklabs/SpacetimeDB/pull/3980


jdetter (approved):
Seems fine, I haven't tested this myself but we know this fixed this issue in other workflows.

PR #4024, opened 2026-01-13
Expand doc comment on `FilterableValue` with requirements to implement

gefjon:
# Description of Changes

We've received a number of community pull requests which add `impl FilterableValue` for new column types, without implementing the features which `FilterableValue` is intended to describe. This PR extends the doc comments on `FilterableValue` to explicitly write what the requirements are to add a new implementation.

N/a

N/a

# Testing

N/a


bfops:
@gefjon FYI lints failing

PR #4027, opened 2026-01-13
CI - Fix `/update-llm-benchmark` getting cancelled

bfops:
# Description of Changes

Augmented the concurrency group to not have unrelated comments cancel the `/update-llm-benchmark` command.

CI only

1

# Testing

I think this can only be tested once merged? Comments on this PR still seem to use the old workflow.


bfops:
test

.github/workflows/ci.yml.old
```diff
@@ -0,0 +1,725 @@
+on:
```
jdetter: Is this supposed to be here?

PR #4011, opened 2026-01-12
Make `Uuid` Nameable

kistz:
# Description of Changes

Implements the `Nameable` interface for the generated uuid client bindings.

# API and ABI breaking changes
None.

# Expected complexity level and risk

1. Only additive changes pretty much copy pasted from the other types where it was already implemented.

# Testing

- [x] Before it didnt work in my project now it works

PR #4030, opened 2026-01-13
Support `Uuid` in React Hook `where` clause

kistz:
# Description of Changes

Dpends on and includes: #4011 
Will remove commits whenever that merges 👍 

This impacts the `useTable()` functions `where`clause.
We add a new possibility for the `Value` type which is the new `Uuid` type.

Also my formatter is bugging again :( whenever i `pnpm format` it does things differently than the guidelines? Would be cool for that reason if someone can run format for me maybe bfops like the last time :>

None

1. Only lifts the filter for the where clause a bit to further support `Uuid` alongside bool, string, and number.

# Testing
- [x] Getting syntax highlithing for the new availbale row
- [x] Getting row back from subscription when supplying a uuid
- [x] Fix edge case with string

### DIFF
--- a/crates/bindings-typescript/src/react/useTable.ts
+++ b/crates/bindings-typescript/src/react/useTable.ts
@@ -87,11 +87,11 @@ export function evaluate<Column extends string>(
         return v === expr.value;
       }
       if (typeof v === 'object') {
-        // Value of the Column and passed Value are both a Uuid so do an integer comparison. 
+        // Value of the Column and passed Value are both a Uuid so do an integer comparison.
         if (v instanceof Uuid && expr.value instanceof Uuid) {
           return v.asBigInt() === expr.value.asBigInt();
         }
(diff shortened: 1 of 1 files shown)


kistz:
Uhhh woops did the wrong git command ^^

I thought i squashed it but something went wrong now im not sure anymore how to best revert this 🥴 @bfops is this a problem that i should fix or is the squashing while merging going to be enough? ^^

egormanga:
> im not sure anymore how to best revert this

I believe you can just do a `git rebase -i` over whatever branch you're basing it on (most probably, `origin/master`) and choose the commits for squashing, not touching your relevant ones.

bfops:

In the meantime, I rebased and ran `pnpm format`.

kistz:
Thank you very much bfops! :>

cloutiertyler (approved):

PR #4033, opened 2026-01-14
Add `cargo ci dlls` command for building C# DLLs and NuGet packages

rekhoff:
# Description of Changes
* Added a new `cargo ci dlls` subcommand to build/pack the in-repo C# NuGet packages and the C# SDK.
* `cargo ci dlls`  restores `sdks/csharp/SpacetimeDB.ClientSDK.csproj` using the freshly built local package outputs as to populate `sdks/csharp/packages/**`.
* Added a Unity `.meta` skeleton under `sdks/csharp/unity-meta-skeleton~/**` and overlays those `.meta` files onto the latest restored versioned package directory to keep Unity GUIDs stable and import settings consistent.
* Unity-specific import fixes are captured in the skeleton overlay (notably: preventing Unity from importing incompatible TFMs like `net8.0`, and marking analyzer DLLs with the `RoslynAnalyzer` label so Unity can recognize them).
# How to use (local)
```bash
# Build/pack + restore local packages into sdks/csharp/packages/**
cargo ci dlls
```
N/A
2 - Local developer tooling + file overlay into restore output; no runtime/SDK API behavior changes.
# Testing
- [x] `cargo check -p ci`
- [x] Ran `cargo ci dlls` and verified the output under `sdks/csharp/packages/**` and the various NuGet package locations.
- [x] Tested a Unity project importing the SpacetimeDB SDK after generating output and confirmed no errors.


tools/ci/src/main.rs
```diff
@@ -211,6 +339,62 @@ fn main() -> Result<()> {
...
+
+            clear_restored_package_dirs("spacetimedb.bsatn.runtime")?;
+            clear_restored_package_dirs("spacetimedb.runtime")?;
+
+            cmd!(
+                "dotnet",
+                "restore",
+                "SpacetimeDB.ClientSDK.csproj",
+                "--configfile",
+                &nuget_config_path_str,
```
bfops, line 380: wow TIL there's a configfile param

bfops: ah but only on `restore`

tools/ci/src/main.rs
```diff
@@ -30,6 +32,126 @@ struct Cli {
...
+    copy_overlay_dir(&skeleton_root, &versioned_dir)
+}
+
+fn clear_restored_package_dirs(pkg_id: &str) -> Result<()> {
+    let pkg_root = Path::new("sdks/csharp/packages").join(pkg_id);
+    if !pkg_root.exists() {
+        return Ok(());
+    }
+
+    for entry in fs::read_dir(&pkg_root)? {
```
bfops: nit: is there a reason we don't `remove_dir_all(&pkg_root)`?

tools/ci/src/main.rs
```diff
@@ -30,6 +32,126 @@ struct Cli {
...
+        }
+
+        let name = entry.file_name().to_string_lossy().to_string();
+        let Ok(version) = Version::parse(&name) else {
+            continue;
+        };
+
+        match best.as_ref() {
+            None => best = Some((version, entry.path())),
+            Some((best_v, _)) if version > *best_v => best = Some((version, entry.path())),
```
bfops: nit: I think it's fine to just pick the first dir (and maybe assert that there is at most 1), especially since we nuke all subdirs of the package's directory beforehand anyway.

(if you make that change, I would add a comment to that effect though)

tools/ci/src/main.rs
```diff
@@ -30,6 +32,126 @@ struct Cli {
...
+    for entry in fs::read_dir(src)? {
+        let entry = entry?;
+        let src_path = entry.path();
+        let dst_path = dst.join(entry.file_name());
+        if entry.file_type()?.is_dir() {
+            if dst_path.exists() {
+                copy_overlay_dir(&src_path, &dst_path)?;
+            }
+        } else {
+            if let Some(file_name) = dst_path.file_name().and_then(|n| n.to_str()) {
```
bfops: nit: maybe we should just save `entry.file_name()` above (or clone it here) since `dst_path.file_name` is essentially derived from it anyway?

bfops: but more importantly: should this be an `if let Some(...)`? shouldn't we `unwrap` or something? wouldn't we want to know if this failed?

tools/ci/src/main.rs
```diff
@@ -30,6 +32,126 @@ struct Cli {
...
+                if let Some(asset_name) = file_name.strip_suffix(".meta") {
```
bfops: similar here: if the `strip_suffix` failed, wouldn't we want to error out? when would we not expect this to be `Some`?

tools/ci/src/main.rs
```diff
@@ -30,6 +32,126 @@ struct Cli {
...
+                        let _ = fs::remove_file(&dst_path);
+                    }
+                    continue;
+                }
+            }
+
+            if let Some(parent) = dst_path.parent() {
+                fs::create_dir_all(parent)?;
+            }
+            fs::copy(&src_path, &dst_path)?;
```
bfops, line 130: I think this must be related to my confusion above, but I don't understand what case this block is for

tools/ci/src/main.rs
```diff
@@ -211,6 +339,62 @@ fn main() -> Result<()> {
...
+                "restore",
+                "SpacetimeDB.ClientSDK.csproj",
+                "--configfile",
+                &nuget_config_path_str,
+            )
+            .dir("sdks/csharp")
+            .run()?;
+
+            overlay_unity_meta_skeleton("spacetimedb.bsatn.runtime")?;
+            overlay_unity_meta_skeleton("spacetimedb.runtime")?;
```
bfops, line 386: does this directory get created for you? for me I only get `spacetimedb.bsatn.runtime`

sdks/csharp/unity-meta-skeleton~/spacetimedb.bsatn.runtime/lib/net8.0.meta
```diff
@@ -0,0 +1,8 @@
+fileFormatVersion: 2
```
bfops, line 1: on my system the `netstandard2.1` versions get restored, not the `net8.0` versions. Is it maybe platform-dependent? Maybe we should copy the `net8.0` meta files into a netstandard2.1` structure as well, since the code already skips `.meta` files for assets that don't exist at runtime?

tools/ci/src/main.rs
bfops: this seems like by far the cleanest way to do this dance

tools/ci/src/main.rs
rekhoff: Great point, latest update should now do exactly that

tools/ci/src/main.rs
rekhoff: Great point. Implemented, and added an error message instead of a comment.

tools/ci/src/main.rs
rekhoff: Rewrote the .meta detection to avoid string conversion entirely.
So if `src_path.extension() == "meta"`, derive the asset path via `dst_path.file_stem()` (and expect it exists)
Copy `.meta` only if the corresponding asset exists in the restored dir, otherwise delete stale `.meta`

rekhoff: I agree. The rewrite fixes this. Now uses path-based `.extension()`/`.file_stem()` so we don’t rely on UTF-8 or string suffix logic, and we fail with a message if something unexpected happens.

tools/ci/src/main.rs
rekhoff: Yup, the rewrite should cover this

sdks/csharp/unity-meta-skeleton~/spacetimedb.bsatn.runtime/lib/net8.0.meta
rekhoff: Yeah, `SpacetimeDB.ClientSDK.csproj` references `SpacetimeDB.BSATN.Runtime` and that package is multi-targeted (`netstandard2.1` and `net8.0`). I don't recall the specifics on if that was because of platform-dependances, but it's plausible to you have a `netstandard2.1` artifacts locally. Luckily the overlay copies `.meta` only for assets that exist, so we shouldn’t need to duplicate `net8.0` metas into a `netstandard2.1` structure.

tools/ci/src/main.rs
rekhoff: `spacetimedb.runtime` may or may not exist depending on the restore graph; `cargo ci dlls` currently tries to overlay both, but `overlay_unity_meta_skeleton` is already a no-op if the package dir doesn’t exist

rekhoff:
> This looks pretty great to me. I left some nits, and I think there may be some remaining `meta` dance to do.
> 
> also, I see more files getting restored such as `LICENSE` and `logo.png`, so I think we should possibly do what #3648 did and add a gitignore for those files as well.

Updated `.gitignore` to ignore those restore artifacts and keep only `lib/**` + `analyzers/**`

tools/ci/src/main.rs
rekhoff: Good question. that `create_dir_all(parent)` is defensive boilerplate for copying non-`.meta` files, but given the current logic (we only recurse into dirs that already exist in the restored tree, and `dst` must exist), the parent should already exist. I'll remove this block to avoid confusion, since it doesn’t cover a real case right now.

sdks/csharp/unity-meta-skeleton.meta
```diff
@@ -0,0 +1,8 @@
+fileFormatVersion: 2
```
bfops: wait is this file needed? I thought `unity-meta-skeleton~` would make it ignored by unity.

tools/ci/src/main.rs
bfops: oh what I was saying was that I don't think we even need to parse versions, or sort or anything. Like are there going to be other subdirectories hanging around in a package dir?

bfops: it's fine the way it is too, just a nit

tools/ci/src/main.rs
bfops: ah so this case is for hypothetical non-meta files, got it :+1:

tools/ci/src/main.rs
bfops: okay sgtm!

sdks/csharp/unity-meta-skeleton~/spacetimedb.bsatn.runtime/lib/net8.0.meta
bfops: ah it looks like the `net8.0` one gets populated for me during `dotnet pack` but not `dotnet restore` :thinking: interesting. What we had tracked in the repo before this PR didn't include `net8.0`.. I'm not sure what Unity ends up using.. seems fine to include it though.

but what I'm saying is that my `netstandard2.1` gets populated with a dll during this process, so I think it also needs a meta file?

```
git ls-files sdks/csharp/packages/
sdks/csharp/packages/.gitignore
sdks/csharp/packages/spacetimedb.bsatn.runtime.meta
sdks/csharp/packages/spacetimedb.bsatn.runtime/1.11.2/analyzers.meta
sdks/csharp/packages/spacetimedb.bsatn.runtime/1.11.2/analyzers/dotnet.meta
sdks/csharp/packages/spacetimedb.bsatn.runtime/1.11.2/analyzers/dotnet/cs.meta
sdks/csharp/packages/spacetimedb.bsatn.runtime/1.11.2/analyzers/dotnet/cs/SpacetimeDB.BSATN.Codegen.dll
sdks/csharp/packages/spacetimedb.bsatn.runtime/1.11.2/analyzers/dotnet/cs/SpacetimeDB.BSATN.Codegen.dll.meta
sdks/csharp/packages/spacetimedb.bsatn.runtime/1.11.2/lib.meta
sdks/csharp/packages/spacetimedb.bsatn.runtime/1.11.2/lib/net8.0.meta
sdks/csharp/packages/spacetimedb.bsatn.runtime/1.11.2/lib/net8.0/SpacetimeDB.BSATN.Runtime.dll
sdks/csharp/packages/spacetimedb.bsatn.runtime/1.11.2/lib/net8.0/SpacetimeDB.BSATN.Runtime.dll.meta
sdks/csharp/packages/spacetimedb.bsatn.runtime/1.11.2/lib/netstandard2.1.meta
sdks/csharp/packages/spacetimedb.bsatn.runtime/1.11.2/lib/netstandard2.1/SpacetimeDB.BSATN.Runtime.dll
```
^ that last file is missing a `.meta` file.

sdks/csharp/unity-meta-skeleton.meta
rekhoff: You're right, I don't think we do. I tested locally with it removed and it seemed fine. It's been removed now

tools/ci/src/main.rs
```diff
@@ -211,6 +339,62 @@ fn main() -> Result<()> {
...
+            let bsatn_source = repo_root.join("crates/bindings-csharp/BSATN.Runtime/bin/Release");
+            let runtime_source = repo_root.join("crates/bindings-csharp/Runtime/bin/Release");
+
+            let nuget_config_dir = tempfile::tempdir()?;
+            let nuget_config_path = nuget_config_dir.path().join("nuget.config");
+            let nuget_config_contents = format!(
+                "<?xml version=\"1.0\" encoding=\"utf-8\"?>\n<configuration>\n  <packageSources>\n    <clear />\n    <add key=\"Local SpacetimeDB.BSATN.Runtime\" value=\"{}\" />\n    <add key=\"Local SpacetimeDB.Runtime\" value=\"{}\" />\n    <add key=\"nuget.org\" value=\"https://api.nuget.org/v3/index.json\" />\n  </packageSources>\n  <packageSourceMapping>\n    <packageSource key=\"Local SpacetimeDB.BSATN.Runtime\">\n      <package pattern=\"SpacetimeDB.BSATN.Runtime\" />\n    </packageSource>\n    <packageSource key=\"Local SpacetimeDB.Runtime\">\n      <package pattern=\"SpacetimeDB.Runtime\" />\n    </packageSource>\n    <packageSource key=\"nuget.org\">\n      <package pattern=\"*\" />\n    </packageSource>\n  </packageSourceMapping>\n</configuration>\n",
```
jdetter: This is more readable:
```suggestion
            let nuget_config_contents = format!(
                r#"<?xml version="1.0" encoding="utf-8"?>
            <configuration>
              <packageSources>
                <clear />
                <add key="Local SpacetimeDB.BSATN.Runtime" value="{}" />
                <add key="Local SpacetimeDB.Runtime" value="{}" />
                <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
              </packageSources>
              <packageSourceMapping>
                <packageSource key="Local SpacetimeDB.BSATN.Runtime">
                  <package pattern="SpacetimeDB.BSATN.Runtime" />
                </packageSource>
                <packageSource key="Local SpacetimeDB.Runtime">
                  <package pattern="SpacetimeDB.Runtime" />
                </packageSource>
                <packageSource key="nuget.org">
                  <package pattern="*" />
                </packageSource>
              </packageSourceMapping>
            </configuration>
            "#,
                bsatn_runtime_path,
                runtime_path,
            );
```

jdetter: The other option is to just not embed this file here and instead have this file as part of the repo. Then we just copy this file where it needs to go during `cargo build dlls`. I'll leave it up to @bfops.

bfops: I'm fine with either one as well

sdks/csharp/.meta-check-ignore
```diff
@@ -0,0 +1,3 @@
+unity-meta-skeleton.meta
```
bfops: nit: I think this line is no longer needed

tools/ci/src/main.rs
bfops: @rekhoff up to you whether you want to do this. Feel free to resolve.

rekhoff: > @rekhoff up to you whether you want to do this. Feel free to resolve.

You are right, I just removed the version check and renamed some things.

sdks/csharp/.meta-check-ignore
```diff
@@ -0,0 +1,2 @@
+unity-meta-skeleton~
+unity-meta-skeleton~/**
```
bfops, line 2: actually do these files need ignoring? I thought that was the point of the `~` suffix.. or is that only for Unity and not the unity meta check?

not important either way :shrug: it works the way it is. it just occurred to me.

PR #4040, opened 2026-01-14
CI - Fix hint for fixing llm benchmarks

bfops:
# Description of Changes

I believe that local users do not have API tokens for OpenAI, so the existing hint was not helpful. Apparently the correct path is to post `/update-llm-benchmark` on the PR and let the CI take care of it.

None

1

# Testing

None


docs/DEVELOP.md
```diff
@@ -12,21 +12,6 @@ This document explains how to configure the environment, run the LLM benchmark t
...
-What this does:
-1. Runs Rust rustdoc_json pass for GPT-5 only.
-2. Runs C# docs pass for GPT-5 only.
-3. Writes updated results & summary.
-
----
-
-> Model IDs passed to `--models` must match configured routes (see `model_routes.rs`), e.g. `"openai:gpt-5"`.
-
-
```
cloutiertyler, line 29: I would keep this because it is true and consistent with this DEVELOP.md. You can do this, you just need the API keys.

We should perhaps just add a note here that CI will run this for you if you comment /update-llm-benchmark.

Boegie19: It seems to me like outsiders can't run /update-llm-benchmark or it takes 5+ minutes and does not say anyting until it is done.

https://github.com/clockworklabs/SpacetimeDB/pull/4043#issuecomment-3751199467

bfops: Okay, updated!

@Boegie19 that's correct, for security reasons. I've tried to add a comment about that.

cloutiertyler (approved):

PR #4042, opened 2026-01-14
Templates naming standarization

drogus:
# Description of Changes

This PR renames the templates to always use shorthand for the language, specify a framework (or console) if necessary, and shorten the naming in general

1

# Testing

I've tested generating templates manually


eslint.config.js
```diff
@@ -38,9 +38,6 @@ export default tseslint.config(
           './tsconfig.json',
           './crates/bindings-typescript/tsconfig.json',
           './crates/bindings-typescript/test-app/tsconfig.json',
-          './crates/bindings-typescript/examples/basic-react/tsconfig.json',
-          './crates/bindings-typescript/examples/empty/tsconfig.json',
-          './crates/bindings-typescript/examples/quickstart-chat/tsconfig.json',
```
cloutiertyler, line 43: Why are these just removed and not changed?

cloutiertyler (approved):
Quite a beast of a PR, but looks good to me. I left one important question though.

eslint.config.js
drogus: Definitely not intentional, I think I might have deleted too much when resolving conflicts. I'll bring them back

PR #4045, opened 2026-01-15
Use VM-based github runners

jdetter:
# Description of Changes

This workflow change migrates us from docker container runners to the new system which uses virtual machines via qemu. This solves a lot of complexities around docker-in-docker and also allows us to easily run windows VMs in the future.

None

0 - this only updates our workflow and has no impact on the code.

# Testing

- [x] CI is passing


.github/workflows/attach-artifacts.yml
```diff
@@ -9,11 +9,7 @@ on:
 
 jobs:
   upload-assets:
-    runs-on: spacetimedb-new-runner
-    container:
-      image: localhost:5000/spacetimedb-ci:latest
-      options: >-
-        --privileged
+    runs-on: ubuntu-latest
```
jdetter: This is such a small job that we should use our free credits instead

.github/workflows/ci.yml
```diff
@@ -23,13 +23,10 @@ jobs:
     name: Smoketests
     strategy:
       matrix:
-        runner: [spacetimedb-new-runner, windows-latest]
+        runner: [spacetimedb-new-runner-2, windows-latest]
```
bfops: nit: what do you think of removing this line, and instead adding a `name` field to the matrix, and then setting name based on that matrix field? That way the runner name isn't in the CI job name, and we don't have to change the required checks list every time we change the runner name.

bfops (approved):
Looks great to me!

.github/workflows/ci.yml
jdetter: Implemented, thanks for the suggestion 👍

PR #4051, opened 2026-01-15
durability: Notify waiters after lockfile is dropped

kim:
Notifications were sent shortly before dropping the actor (and thus, the lockfile), which could cause the `close` future to return while the lock is still held.

This can lead to a race if the database is re-opened immediately, such as in `TestDB::reopen()`, causing test flakes.

1

# Testing

Should result in "Database is already opened" test flakes to go away.

### DIFF
--- a/crates/durability/src/imp/local.rs
+++ b/crates/durability/src/imp/local.rs
@@ -2,6 +2,7 @@ use std::{
     io,
     num::NonZeroU16,
     panic,
+    path::PathBuf,
     sync::{
         atomic::{AtomicU64, Ordering::Relaxed},
         Arc,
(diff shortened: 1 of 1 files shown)


crates/durability/src/imp/local.rs
```diff
@@ -317,6 +314,32 @@ impl<T: Encode + Send + Sync + 'static> Actor<T> {
...
+
+    pub fn notified(&self) -> OwnedNotified {
+        self.notify_on_drop.clone().notified_owned()
+    }
+}
+
+impl Drop for Lock {
+    fn drop(&mut self) {
+        // Ensure the file lock is dropped before notifying.
+        let _ = self.file.take();
```
bfops: is this definitely definitely an insta-drop? I've never been completely clear on the semantics of this.

jdetter: This is not a definitive source but chatgpt seems to agree that this is an insta drop:
image

bfops: chatgpt told me the opposite:
image

and I got lost while trying to find docs on the behavior, so I figured I'd ask

kim: Well... I'm fairly sure, but can't find formal specification of `let _ =`.
I'll write `map(drop)` instead, and just tell clippy to shut up.

bfops (approved):
this fix makes sense to me! Thank you for finding it :pray: that seems like a pain.

PR #4056, opened 2026-01-16
Fix test failure in fresh checkout of repository

douglance:
# Description of Changes

Use relative import `'../src'` in serde.test.ts to match all other test files.

This is a consistency fix. The `'spacetimedb'` import works in CI (build runs first, enabling package self-reference), but all other tests use `'../src'`.

None

1


bfops:
Hey @douglance, thank you for taking the time to make a contribution!

To minimize noise in the repo, we generally try to avoid small preference-inspired PRs (unless they're specifically bringing us in line with our linter, style guide, etc.)

I hope that's understandable. Please let me know if I've missed something important about the motivation or impact here!

douglance:
@bfops This fixes an issue where the tests fail unless the system has been built. It's not a preference thing. 

I lost some time trying to figure out why this test was failing on a fresh pull down of the repo.

bfops:
Ah got it, thank you! That makes sense.

PR #4058, opened 2026-01-16
fix: docs llm test. make tables that we sql query public

Boegie19:
# Description of Changes

in the llm tests we run to see if docs update break llms.
we are making sql queries to tables that we did not set to public what means the sql query will fail

1

# Testing

Run /update-llm-benchmark and see if it fixes the sql query issues.


cloutiertyler:
@Boegie19 I don't think we should update the answers to expect public be explicit. Public is (or should be) the default. If it is not we should fix that instead.

egormanga:
> Public is (or should be) the default

It isn't, and changing this would be a major security breaking change for existing codebases, which I'm sure should not happen unless bumping to 2.0 **and** providing a source migration tool.

cloutiertyler:
Sorry! I am a fool! They are *private* by default, not public.

Boegie19:
@cloutiertyler Can you run /update-llm-benchmark ? if it works.

PR #4061, opened 2026-01-18
Move unity testsuite onto unity runner

jdetter:
# Description of Changes

- This will eliminate a lot of the test flakes with the unity testsuite. A huge amount of the test flakes here were that we were trying to acquire more than 2 unity license seats at a time which is not allowed. This will have the largest impact when the runners are busy.

The downside here: This will cause jobs to wait which is what we want right now. Jobs may wait to queue for several minutes while other jobs finish but I don't expect this to have a big impact on the total time a PR will spend in the merge queue.

None

1

# Testing

- [x] Unity CI passes even if multiple jobs are running at the same time

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -625,9 +625,6 @@ jobs:
           key: Unity-${{ github.head_ref }}
           restore-keys: Unity-
 
-        # We need this to support "Docker in Docker"
-      - name: Start Docker daemon
-        run: /usr/local/bin/start-docker.sh
       - name: Run Unity tests
         uses: game-ci/unity-test-runner@v4
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -630,9 +625,6 @@ jobs:
           key: Unity-${{ github.head_ref }}
           restore-keys: Unity-
 
-        # We need this to support "Docker in Docker"
-      - name: Start Docker daemon
-        run: /usr/local/bin/start-docker.sh
```
jdetter, line 635: The new runners are very similar to the original github hosted runners - so we don't need docker in docker.

bfops (approved):
Straightforward to me! Thanks @jdetter !

cloutiertyler (approved):

PR #4063, opened 2026-01-19
feat: create Svelte framework integration

clockwork-tien:
# Description of Changes

The PR implements the following updates:

- Create Svelte framework sdk/integration
- Add Svelte Quickstart Docs
- Create `svelte-ts` template

# Screenshots
- `svelte-ts` template

Screenshot1

- Svelte Quickstart Docs

Screenshot2

# Testing

- [ ] 
- [ ]


crates/bindings-typescript/src/svelte/useTable.ts
```diff
@@ -0,0 +1,375 @@
...
+        expr.children.length === 0 || expr.children.every(c => evaluate(c, row))
+      );
+    case 'or':
+      return (
+        expr.children.length !== 0 && expr.children.some(c => evaluate(c, row))
+      );
+  }
+}
+
+function formatValue(v: Value): string {
```
cloutiertyler: Much of this code is duplicated across Vue, React, Svelte. Can we please consolidate it into a shared file?

templates/svelte-ts/src/App.svelte
```diff
@@ -0,0 +1,63 @@
+<script lang="ts">
+import { useSpacetimeDB, useTable, useReducer } from 'spacetimedb/svelte';
+import { tables, reducers } from './module_bindings';
+
+const conn = useSpacetimeDB();
```
cloutiertyler, line 5: Is this idiomatic for Svelte? See: https://discord.com/channels/@me/1414626136121676017/1463374453193834506

cloutiertyler (commented):
This generally looks good to me. Just two questions/comments in the code.

crates/bindings-typescript/src/svelte/useTable.ts
kistz: Im not certain that its worth for now since maybe we could reuse the ts client side query builder for the where clause logic? 🤔 i definetly wanna look into that when that pr lands :> But understandable if its better consolidated now and deleted now later 👍
In case you do end up consolidating it please keep the react version since there i added uuid support for the where clause 😇

clockwork-tien: @cloutiertyler Thank you! I have solidated the code into a shared file 👍

templates/svelte-ts/src/App.svelte
clockwork-tien: Thank you it is idiomatic for Svelte (discussed in more details with Tyler)

crates/bindings-typescript/src/svelte/useTable.ts
cloutiertyler: @kistz I believe your uuid support will not be affected and if it was we would have a conflict because that merged already, IIRC.

cloutiertyler (approved):
This now LGTM

PR #4064, opened 2026-01-19
Add ~/.local/bin to GITHUB_PATH for llm benchmark workflow

drogus:
# Description of Changes

This is a try to fix the Update LLM benchmarks workflow

1


drogus:
I tested it on push and the fix worked here: https://github.com/clockworklabs/SpacetimeDB/actions/runs/21147439404

JulienLavocat (approved):
Seems like an easy enough fix

PR #4065, opened 2026-01-19
Fix log streaming error with spacetime dev

aasoni:
# Description of Changes

Addresses an issue with `spacetime dev` when publishing new databases.

```
Error: Log streaming error: error sending request for url (http://127.0.0.1:3000/v1/database/c2009a1b9add1dcedd94ab67b5027f029e02da013c966e41288aa50035f867a1/logs?num_lines=10&follow=true)
Reconnecting in 10 seconds...
```

The error was caused by the database not being loaded. To fix this we make call to the /schema endpoint to ensure the database is loaded before streaming the logs.

None

1

# Testing

Ran spacetime dev locally and against maincloud with new and existing databases. The log always worked.

### DIFF
--- a/crates/cli/src/subcommands/dev.rs
+++ b/crates/cli/src/subcommands/dev.rs
@@ -284,6 +284,14 @@ pub async fn exec(mut config: Config, args: &ArgMatches) -> Result<(), anyhow::E
     sleep(Duration::from_secs(1)).await;
 
     let db_identity = database_identity(&config, &database_name, Some(resolved_server)).await?;
+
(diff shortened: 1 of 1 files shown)


cloutiertyler (approved):
Approved, but obviously we want to fix the underlying issue.

egormanga:
Sounds logical to me that `/logs` won't load an unloaded database.

aasoni:
Fixed by: https://github.com/clockworklabs/SpacetimeDB/pull/3961
Closing this, as the error will be addressed once we deploy the above change to maincloud.

PR #4072, opened 2026-01-20
Keynote Spacetime simulation comparisons

bradleyshep:
# Description of Changes

Adds a comprehensive benchmarking suite for comparing SpacetimeDB performance against other databases and backends.

**Key features:**
- Load testing framework that runs repeatable benchmarks across multiple connectors
- Supports SpacetimeDB, PostgreSQL, CockroachDB, SQLite, Supabase, Convex, and Bun
- Configurable test parameters: duration, concurrency, Zipf distribution (α) for hot/cold account skew
- Outputs JSON reports with TPS and latency statistics to `./runs/`
- Docker Compose support for containerized benchmarking
- Includes a test SpacetimeDB Rust module (`modules/test-1/server`)

**Connectors included:**
- Direct SQL connectors (Postgres, CockroachDB, SQLite)
- Drizzle ORM wrappers
- RPC-based connectors for network-isolated benchmarking
- SpacetimeDB WebSocket connector
- Supabase and Convex cloud connectors

None. This is a new standalone template with no changes to existing code.

**1** - This is an additive change that introduces a new isolated template folder. It has no interactions with existing SpacetimeDB crates or modules.

# Testing

- [ ] Run `pnpm install` in `templates/keynote-2/`
- [ ] Verify SpacetimeDB module compiles: `cd templates/keynote-2/modules/test-1/server && cargo build`
- [ ] Run a basic benchmark against SpacetimeDB (requires local SpacetimeDB instance)


templates/keynote-2/spacetimedb/Cargo.toml
cloutiertyler, line 1: Could you rename this directory: `templates/keynote-2/modules/test-1/server` -> `templates/keynote-2/spacetimedb`

That's both clearer as to what's in here and also fits our standard.

templates/keynote-2/cockroachdb/docker-compose-crdb-node-1.yml
cloutiertyler, line 1: Can we rename this folder to `cockroachdb`

cloutiertyler (commented):
I would love to clean this up so we have only the values that we show in the keynote, so namely:

SpacetimeDB (TypeScript)
Node.js + Drizzle + SQLite
Bun + Drizzle + Postgres
Node + Drizzle + Postgres
Node + Drizzle + Supabase
Node + Drizzle + CockroachDB
Node + Drizzle + PlanetScale
Convex

I'd like to remove any direct test, or any test that isn't apples to apples, meaning server + ORM + database (unless the platform implements all three a la Convex/SpacetimeDB).

I'd also like to add a README with a summary of the results and a brief discussion of the methodology:

- How many clients are run
- What the transaction is
- Why we use a web server + ORM + database
- The hardware we ran the test on (core count, ram, specifically PhoenixNAP s3.c3.medium machines)
- The hardware the clients ran on (why the clients are run separately)
- The impact of contention on transfer accounts (I would throw in the graph you have in the G drive for this).
- A note about enabling pipelining for clients
- A note about `withConfirmedReads`
- A note about why SpacetimeDB is so much higher performance (this could link back to the keynote actually), but in general should mention the fact that the round trip time from server to database is a huge impediment to good perf.

I think it's important because we might get 1000+ people reading this document. If you want have an LLM read my comments here and take a crack at it as a starting point.

Then I'm good to call this done.

Really excellent job on this!

bradleyshep:
https://github.com/user-attachments/assets/56ab2c86-7a4d-4d2a-b70a-69ae496b6720

(demo command run on my local machine)

templates/keynote-2/README.md
```diff
@@ -0,0 +1,184 @@
...
+## Quick Demo
+
+See SpacetimeDB's performance advantage with one command:
+
+```bash
+npm install
+npm run demo
+```
+
+The demo automatically starts required services, seeds databases, and displays animated results comparing SpacetimeDB vs Convex.
```
cloutiertyler: Starting the services must be done manually, so we just need to adjust this with instructions on how to run both.

templates/keynote-2/README.md
```diff
@@ -0,0 +1,184 @@
...
+Client → Web Server (HTTP) → ORM (Drizzle) → Database
+```
+
+Or for integrated platforms (SpacetimeDB, Convex):
+
+```
+Client → Integrated Platform (compute + storage colocated)
+```
+
+This ensures we're measuring real-world application performance, not raw database throughput.
```
cloutiertyler, line 65: Please add a note about:

1. How we're benchmarking all services out of the box
2. If the service has local deployment and a cloud service (pro tier, etc) we are benchmarking both and taking the higher performing one. (e.g. for planet list the tier of machine used)

cloutiertyler (approved):
I made one tiny tiny change, but this looks really good.

I was able to test it out on my machine as well with no issues.

PR #4073, opened 2026-01-20
 Bump to v8 145 

coolreader18:
# Description of Changes

This release has a couple of patches I've been waiting on for a bit:

* https://github.com/denoland/rusty_v8/pull/1868, which mean we no longer have to use high slot indexes.
* https://github.com/denoland/rusty_v8/pull/1886, which allows us to lazily parse sourcemaps.
  * I also bumped `sourcemap` to get https://github.com/getsentry/rust-sourcemap/pull/137, so that we no longer have to have our hack to work around that.
* https://github.com/denoland/rusty_v8/pull/1892, which allows us to re-enable `run_timeout_and_cb_every` (though I haven't done that in this PR)

2: v8 is a very big and important dependency, but also very stable.

# Testing

- [x] No change to behavior; automated testing is sufficient.


crates/core/src/host/v8/mod.rs
```diff
@@ -94,6 +94,7 @@ impl V8RuntimeInner {
     ///
     /// Should only be called once but it isn't unsound to call it more times.
     fn init() -> Self {
+        v8::V8::set_flags_from_string("--single-threaded");
         // Our current configuration:
         // - will pick a number of worker threads for background jobs based on the num CPUs.
         // - does not allow idle tasks
```
Centril: Oh! I see - the docs say that you have to pass this flag for `new_single_threaded_default_platform` to have effect. So we weren't actually exploiting single-threadedness.

```suggestion
        // We don't want idle tasks nor background worker tasks,
        // as we intend to run on a single core.
        // Per the docs, `new_single_threaded_default_platform` requires
        // that we pass `--single-threaded`.
        v8::V8::set_flags_from_string("--single-threaded");
```

Cargo.toml
```diff
@@ -369,14 +369,6 @@ unexpected_cfgs = { level = "warn", check-cfg = ['cfg(tokio_unstable)'] }
 # FIXME: we should work on this lint incrementally
 result_large_err = "allow"
 
-[patch.crates-io]
-# Note(bfops): We require this override because temporal_rs (pulled via core->v8->ry_temporal_capi)
-# depends on timezone_provider using a `~` which allows a roll-forward to an incompatible version.
-# Maybe this will no longer be an issue if we bump v8 to a version that depends on a post-0.1
-# version of temporal_rs, but I'm not holding my breath.
-timezone_provider = { git = "https://github.com/boa-dev/temporal", tag = "v0.0.11" }
-temporal_rs = { git = "https://github.com/boa-dev/temporal", tag = "v0.0.11" }
```
Centril, line 380: 🎉

crates/core/src/host/v8/error.rs
Centril: Please move these to be functions on `SourceMaps`.

crates/core/src/host/v8/syscall/hooks.rs
```diff
@@ -58,13 +56,11 @@ impl ModuleHookKey {
...
-            ModuleHookKey::CallReducer => 21,
-            ModuleHookKey::CallView => 22,
-            ModuleHookKey::CallAnonymousView => 23,
-            ModuleHookKey::CallProcedure => 24,
+            ModuleHookKey::DescribeModule => 0,
+            ModuleHookKey::CallReducer => 1,
+            ModuleHookKey::CallView => 2,
+            ModuleHookKey::CallAnonymousView => 3,
+            ModuleHookKey::CallProcedure => 4,
         }
```
Centril: Let's do `#[repr(i32)]` and cast `self` to it.

crates/core/src/host/v8/error.rs
coolreader18: I'll do the new code, but I don't think it really makes sense to do the second part.

Centril: that's fair

coolreader18: Also, no. because of no polonius. `cannot borrow `*scope` as mutable because it is also borrowed as immutable
mutable borrow occurs here`

PR #4084, opened 2026-01-21
Upgrade version to 1.12.0

jdetter:
# Description of Changes

Version upgrade to `v1.12.0`.

# API and ABI breaking changes

None

# Expected complexity level and risk

1 - this is just a version upgrade

# Testing

The testsuite failures are fixed by https://github.com/clockworklabs/SpacetimeDB/pull/4120

- [x] License has been properly updated including version number and date
- [x] CI passes

PR #4085, opened 2026-01-21
Fix the upgrade version tool

jdetter:
# Description of Changes

This fixes 2 issues with the upgrade version tool:
1. The typescript bindings need to be updated otherwise the typescript test in CI will fail
2. The snapshots need to be updated

When the version upgrade tool check in CI runs, snapshot changes are accepted automatically via the `--accept-snapshots` cli argument. When you are running this tool locally without `--accept-snapshots` you will be asked to manually review the snapshot changes before doing a final test to make sure the snapshots are correct.

None

1 - this just updates the version upgrade tool

# Testing

- [x] Version bump 1.12.0 worked: https://github.com/clockworklabs/SpacetimeDB/pull/4084
- [x] CI passes


tools/upgrade-version/src/main.rs
```diff
@@ -184,7 +190,49 @@ fn main() -> anyhow::Result<()> {
 
         process_license_file("LICENSE.txt", &full_version);
         process_license_file("licenses/BSL.txt", &full_version);
+        println!("$> cargo check");
         cmd!("cargo", "check").run().expect("Cargo check failed!");
+
+        println!("$> pnpm install");
+        cmd!("pnpm", "install").run().expect("pnpm run build failed!");
```
bfops, line 198: for some reason I still get a `pnpm install` failure when I run this. I'm investigating..

jdetter: What is the failure?

bfops: ah it's because I hadn't pulled

tools/upgrade-version/src/main.rs
```diff
@@ -184,7 +190,49 @@ fn main() -> anyhow::Result<()> {
 
         process_license_file("LICENSE.txt", &full_version);
         process_license_file("licenses/BSL.txt", &full_version);
+        println!("$> cargo check");
```
bfops, line 194:
```suggestion
        // Rebuild `Cargo.lock`
        println!("$> cargo check");
```

tools/upgrade-version/src/main.rs
```diff
@@ -184,7 +190,49 @@ fn main() -> anyhow::Result<()> {
...
+        println!("$> pnpm install");
+        cmd!("pnpm", "install").run().expect("pnpm run build failed!");
+
+        println!("$> pnpm run build");
+        cmd!("pnpm", "run", "build").run().expect("pnpm run build failed!");
+
+        println!("$> pnpm --dir templates/chat-react-ts generate");
+        cmd!("pnpm", "--dir", "templates/chat-react-ts", "generate")
+            .run()
+            .expect("pnpm generate failed!");
```
bfops, line 206: Should the pnpm stuff be in the typescript branch, not the rust branch?

bfops (commented):
This looks good to me, but I have some questions

tools/upgrade-version/src/main.rs
jdetter: So the main thing here is when we update the version number of the CLI we *must* update those bindings. So I think it should be in this branch even though yes it is a pnpm command.

bfops (approved):
This looks great to me!

PR #4086, opened 2026-01-21
Add `Index::merge_from` + `Index::insert` exploits stuff for perf.

Centril:
# Description of Changes

This PR:
- Does some minor readability improvements and organization to indices.
- Adds `Index::merge_from(&mut self, src: Self)`, which merges `src` into `self`. This will be used to speed up merging of transactions.
- Makes `Index::insert` to exploit a property for improved performance in itself and in `Index::merge_from`.
- Adds a trait invariant stating that no `(key, ptr)` occurs twice. This was always the case, but now it is a safety invariant and exploited for performance. This then is bubbled up to `MutTxId` and friends.

This is a precursor to https://github.com/clockworklabs/SpacetimeDB/pull/2744.

None

4 - This touches both `Index::insert`, `Table`, and `MutTxId` in terms of `unsafe`.

# Testing

- Debug assertions are added for the unsafety in `Index::insert`.
- A new proptest `merge_from_is_union` is added that comprehensively exercises `Index::{insert, merge_from, can_merge}` together and for all index kinds and uniqueness.


crates/bench/benches/index.rs
```diff
@@ -184,7 +184,7 @@ impl Index for IBTree {
         Self(<_>::default())
     }
     fn insert(&mut self, key: K, val: RowPointer) -> Result<(), RowPointer> {
-        self.0.insert(key, val)
+        unsafe { self.0.insert(key, val) }
```
coolreader18, line 188: Safety comment?

crates/bench/benches/index.rs
```diff
@@ -250,7 +250,7 @@ impl Index for IDirectIndex {
...
```
coolreader18, line 255: and here

crates/table/src/table_index/hash_index.rs
```diff
@@ -22,55 +22,72 @@ pub struct HashIndex<K: Eq + Hash> {
...
+        // SAFETY: caller promised that `(key, ptr)` does not exist in the index
+        // so it also does not exist in `entry`.
+        unsafe { entry.push(ptr) };
         Ok(())
     }
 
-    /// Deletes `key -> ptr` from this multimap.
-    ///
-    /// Returns whether `key -> ptr` was present.
+    unsafe fn merge_from(&mut self, src_index: Self, mut translate: impl FnMut(RowPointer) -> RowPointer) {
```
coolreader18, line 75: Shoudl have a safety section of docs

crates/table/src/table_index/multimap.rs
```diff
@@ -15,55 +53,72 @@ pub struct MultiMap<K> {
...
-    /// and multimaps do not bind one `key` to the same `ptr`.
-    fn insert(&mut self, key: Self::Key, ptr: RowPointer) -> Result<(), RowPointer> {
-        self.num_rows += 1;
-        self.num_key_bytes.add_to_key_bytes::<Self>(&key);
-        self.map.entry(key).or_default().push(ptr);
+    // =========================================================================
+    // Mutation
+    // =========================================================================
+
+    unsafe fn insert(&mut self, key: Self::Key, ptr: RowPointer) -> Result<(), RowPointer> {
```
coolreader18, line 95: and here, etc etc

crates/table/src/table_index/hash_index.rs
Centril: The safety docs are defined on the trait. It doesn't seem too useful to copy paste that in every implementation.

crates/table/src/table_index/multimap.rs
Centril: See above.

crates/table/src/table_index/hash_index.rs
coolreader18: ahh, gotcha

crates/table/src/table_index/index.rs
```diff
@@ -27,23 +42,62 @@ pub trait Index {
     /// if inserting `key` is not compatible with the index
     /// or if it would be profitable to replace the index with a B-Tree.
     /// The provided default implementation does not return `Despecialize`.
+    ///
+    /// # Safety
+    ///
+    /// The relation `key -> ptr` must not exist in the index.
+    /// Note that inserting `key -> ptr_other`,
+    /// where `ptr != ptr_other`, is legal.
     fn insert_maybe_despecialize(
```
gefjon, line 51: Why is this method not marked `unsafe`, when it has a `Safety` section in its docs and you're calling it with an `unsafe` block?

crates/table/src/table_index/index.rs
```diff
@@ -27,23 +42,62 @@ pub trait Index {
...
+    /// where `ptr != ptr_other`, is legal.
+    unsafe fn insert(&mut self, key: Self::Key, ptr: RowPointer) -> Result<(), RowPointer> {
+        // SAFETY: forward caller requirements.
+        #[allow(unused_unsafe)]
+        unsafe { self.insert_maybe_despecialize(key, ptr) }.unwrap()
     }
 
+    #[cfg(test)]
+    fn insert_for_test(&mut self, key: Self::Key, ptr: RowPointer) -> Result<(), RowPointer> {
+        // SAFETY: proof obligation checked inside the method for debug builds.
```
gefjon, line 80: Does `cfg(test)` necessarily imply `cfg(debug_assertions)`?

crates/table/src/table_index/same_key_entry.rs
```diff
@@ -63,28 +69,104 @@ impl SameKeyEntry {
...
+                // Translate the source.
+                for ptr in &mut src {
+                    *ptr = by(*ptr);
+                }
+
+                // Insert into `self`.
+                match self {
+                    Self::Small(dst) => {
+                        // SAFETY: The intersection is empty, so `dst ++ src` is also a set.
+                        dst.append(&mut src);
```
gefjon, line 137: Do we need to potentially upgrade here to a `Self::Large` if the two sets combined meet some size threshold?

PR #4098, opened 2026-01-22
[1/3] `RawModuleDefV10` definition and validation.

Shubham8287:
# Description of Changes

Patch:
1. `crates/lib/src/db/raw_def/v10.rs` - a definition as per https://github.com/clockworklabs/SpacetimeDBPrivate/issues/2412.
2. Refactors `RawModuleDefV9` validation code to dedup some of the core validation logic, No functionality should change there.
3. Validation logic for new `RawModuleDefV10`.
4. Modify `crates::schema::def::ModuleDef` :
     - To include `raw_module_def_version` 
     -  `visibility` field to existing `RedcuerDef` and `ProcedureDef`
5. It deprecates RLS in favour of Views, If anyoine feel otherwise, let me know.

Lot of code is duplicated from `RawModuleDefV9` includng tests.

NA, `RawModuleDefV10` is not yet exported by modules.

3? close analyses of structure is important to ensure future extensibility.

# Testing
For code motion in `RawModuleDefV9` validation: Existing unit tests seems to cover any functionality change.
For `RawModuleDefV10`: V9 test has been repeated with more `assert`s to check on `visibility`

### DIFF
--- a/crates/schema/src/def/validate/v10.rs
+++ b/crates/schema/src/def/validate/v10.rs
@@ -46,10 +46,7 @@ pub fn validate(def: RawModuleDefV10) -> Result<ModuleDef> {
         .cloned()
         .into_iter()
         .flatten()
-        .map(|reducer| {
-            validator
-                .validate_reducer_def(reducer)
-        })
+        .map(|reducer| validator.validate_reducer_def(reducer))
         // Collect into a `Vec` first to preserve duplicate names.
         // Later on, in `check_function_names_are_unique`, we'll transform this into an `IndexMap`.
         .collect_all_errors::<Vec<_>>();
@@ -407,7 +404,8 @@ impl<'a> ModuleValidatorV10<'a> {
             visibility: visibility.into(),
             ok_return_type,
             err_return_type,
-        }).map(|reducer_def| (name_result, reducer_def))
+        })
+        .map(|reducer_def| (name_result, reducer_def))
     }
 
     fn validate_schedule_def(
@@ -910,22 +908,22 @@ mod tests {
         assert_eq!(def.types[&deliveries_type_name].ty, delivery_def.product_type_ref);
 
         let init_name = expect_identifier("init");
-        assert_eq!(def.reducers[&init_name].name, init_name);
+        assert_eq!(&*def.reducers[&init_name].name, &*init_name);
         assert_eq!(def.reducers[&init_name].lifecycle, Some(Lifecycle::Init));
 
         let on_connect_name = expect_identifier("on_connect");
-        assert_eq!(def.reducers[&on_connect_name].name, on_connect_name);
+        assert_eq!(&*def.reducers[&on_connect_name].name, &*on_connect_name);
         assert_eq!(def.reducers[&on_connect_name].lifecycle, Some(Lifecycle::OnConnect));
 
         let on_disconnect_name = expect_identifier("on_disconnect");
-        assert_eq!(def.reducers[&on_disconnect_name].name, on_disconnect_name);
+        assert_eq!(&*def.reducers[&on_disconnect_name].name, &*on_disconnect_name);
         assert_eq!(
             def.reducers[&on_disconnect_name].lifecycle,
             Some(Lifecycle::OnDisconnect)
         );
 
         let extra_reducer_name = expect_identifier("extra_reducer");
-        assert_eq!(def.reducers[&extra_reducer_name].name, extra_reducer_name);
+        assert_eq!(&*def.reducers[&extra_reducer_name].name, &*extra_reducer_name);
         assert_eq!(def.reducers[&extra_reducer_name].lifecycle, None);
         assert_eq!(
             def.reducers[&extra_reducer_name].params,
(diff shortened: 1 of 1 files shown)


crates/lib/src/db/raw_def/v10.rs
```diff
@@ -0,0 +1,923 @@
...
+    /// Unlike V9 where schedules were embedded in table definitions,
+    /// V10 stores them in a dedicated section.
+    Schedules(Vec<RawScheduleDefV10>),
+
+    /// Lifecycle reducer assignments.
+    ///
+    /// Unlike V9 where lifecycle was a field on reducers,
+    /// V10 stores lifecycle-to-reducer mappings separately.
+    LifeCycleReducers(Vec<RawLifeCycleReducerDefV10>),
+    //TODO: Add section for Event tables, and Case conversion before exposing this from module
```
gefjon, line 86: That's the neat part: we can expose this whenever we want, and adding new sections / variants to this enum is a non-breaking change!

crates/lib/src/db/raw_def/v10.rs
```diff
@@ -0,0 +1,923 @@
...
+/// - Default values are stored inline rather than in `MiscModuleExport`
+/// - Schedules are stored in a separate section rather than embedded here
+#[derive(Debug, Clone, SpacetimeType)]
+#[sats(crate = crate)]
+#[cfg_attr(feature = "test", derive(PartialEq, Eq, PartialOrd, Ord))]
+pub struct RawTableDefV10 {
+    /// The name of the table.
+    /// Unique within a module, acts as the table's identifier.
+    /// Must be a valid `spacetimedb_schema::identifier::Identifier`.
+    pub name: RawIdentifier,
```
gefjon: For added clarity with Tyler's case conversion proposal, could you change this (and other `name` fields) to be called `source_name`?

crates/lib/src/db/raw_def/v10.rs
```diff
@@ -0,0 +1,923 @@
...
+    pub sequences: Vec<RawSequenceDefV10>,
+
+    /// Whether this is a system- or user-created table.
+    pub table_type: TableType,
+
+    /// Whether this table is public or private.
+    pub table_access: TableAccess,
+
+    /// Default values for columns in this table.
+    pub default_values: Vec<RawColumnDefaultValueV10>,
```
gefjon, line 140: Do you think it would make more sense to store this in a separate section? Or any of the other stuff in here? I have no particular opinion, just thought I'd mention it as a possibility.

gefjon (commented):
I haven't looked at your validation logic yet, but I assume it's reasonably straightforward. The raw def format itself looks great, excellent work.

crates/lib/src/db/raw_def/v10.rs
Shubham8287: Default values are coupled to tables, so I think it should go inside Table def.

crates/schema/src/def/validate/v9.rs
```diff
@@ -25,12 +25,14 @@ pub fn validate(def: RawModuleDefV9) -> Result<ModuleDef> {
 
     let known_type_definitions = types.iter().map(|def| def.ty);
 
-    let mut validator = ModuleValidator {
-        typespace: &typespace,
-        stored_in_table_def: Default::default(),
-        type_namespace: Default::default(),
-        lifecycle_reducers: Default::default(),
-        typespace_for_generate: TypespaceForGenerate::builder(&typespace, known_type_definitions),
+    let mut validator = ModuleValidatorV9 {
```
gefjon, line 29: We "validate" `RawModuleDefV8` by translating to V9, then validating the V9 module. Is it possible and potentially profitable to do the same for V9 by translating it to V10? If we can, I would very much like to make it so that we only maintain one version of the actual validation code.

Shubham8287: If you look closely, most of the V10 validation logic is just a thin wrapper around the existing V9 logic but I agree, doing translation might be even better. I will look into it separately.

crates/lib/src/db/raw_def/v10.rs
```diff
@@ -1,13 +1,1101 @@
...
+    /// A BSATN-encoded [`AlgebraicValue`] valid at the column's type.
+    /// (We cannot use `AlgebraicValue` directly as it isn't `SpacetimeType`.)
+    pub value: Box<[u8]>,
+}
+
+/// A reducer definition.
+#[derive(Debug, Clone, SpacetimeType)]
+#[sats(crate = crate)]
+#[cfg_attr(feature = "test", derive(PartialEq, Eq, PartialOrd, Ord))]
+pub struct RawReducerDefV10 {
```
gefjon, line 160: Crap, sorry to forget this for so long and to have to remove from queue. Please extend `RawReducerDefV10` with `ok_return_type: AlgebraicType` and `err_return_type: AlgebraicType`. For now, we'll use unit as the `ok_return_type` and `String` as the `err_return_type`; please assert this in the validate pass.

PR #4102, opened 2026-01-23
Translate smoketests from Python to Rust

cloutiertyler:
# Description of Changes

This PR translates all of our Python smoketests into Rust tests which can be run from `cargo run`

## Motivation

The purpose of this fivefold:

1. All developers on the team are familiar with Rust
2. It simplifies our devops because we can drop Python as a dependency to run the tests
3. You can now run all tests in the repo through the single `cargo test` interface
4. Because we use the `SpacetimeDbGuard` and `cargo test`/`cargo nextest` we can easily parallelize the smoke tests
5. The smoketests can now use machinery imported from SpacetimeDB crates (e.g. `bsatn` etc.)

IMPORTANT NOTE!

There are several ways to implement the smoke tests in Rust (none are great):

1. A separate xtask specifically for the smoke tests
    - This doesn't solve the problem of the CLI tests which also use the `guard` crate
    - Idiosyncratic way to run the smoke tests as opposed to cargo test
    - Does NOT resolve the cargo within cargo problem because we still have to build the test modules with cargo
2. A `build.rs` script in `guard` which first builds the executables as a compile step for compiling guard
    - Deadlocks on a cargo lock file conflict (Outer cargo compiles guard → runs build.rs, inner cargo tries to acquire the build directory lock, outer cargo holds the directory lock, deadlock)
    - If you fix the deadlock by using different target dirs, it still looks stuck on building guard because it's actually compiling all of spacetimedb-standalone and spacetimedb-cli. 
    - Still technically runs cargo inside of cargo.
3. Add `spacetimedb-cli` and `spacetimedb-standalone` as an artifact dependency of the guard crate
    - Has good and clear output but requires +nightly when running the smoketests and CLI tests, otherwise won't do the right thing. See https://github.com/rust-lang/cargo/issues/9096
4. Compile the executables at runtime during the tests themselves where the first test takes a lock while the executables are building using cargo within cargo
    - Makes the tests look like they're taking a long time when they're just waiting for the build to complete
    - Requires relatively complex locking machinery across binaries/tests/processes
5. A two step solution where the developer has to build the binaries before calling the smoke tests
    - Very error prone

None of these are good. `xtask` is not bad, but doesn't enable us to run other integration tests in other crates (e.g. the CLI)

(3) is the correct solution and has the best user experience, but it requires nightly and I don't want to introduce that for all of our tests.

I have chosen to do a combination of (1) and (4). You will now run the smoketests with `cargo smoketest`. If you run `cargo test --all` (or use `guard`) without doing `cargo smoketest` it will fall back to (4) which compiles the executables at runtime. Running `cargo build` is the **only way** to ensure that the executables are not stale because of the internal fingerprint checking. Everything else is fragile not robust.

NOTE! There is no way to avoid cargo within cargo and have the smoke tests be run as cargo tests because the modules under test must be compiled with cargo.

Note that this is a BREAKING CHANGE to `cargo test --all`. The smoketests are now part of `cargo test --all` unless you specifically exclude them.

3, this is partially AI translated. We need to carefully review to ensure the semantics have not regressed. 

# Testing

- [ ] 
- [ ]

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -242,3 +242,6 @@ venv/
 
 # Windows null device artifact
 nul
+
+[Nn][Uu][Gg][Ee][Tt].[Cc][Oo][Nn][Ff][Ii][Gg]
+[Nn][Uu][Gg][Ee][Tt].[Cc][Oo][Nn][Ff][Ii][Gg].meta
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -319,7 +319,14 @@ unicode-normalization = "0.1.23"
 url = "2.3.1"
 urlencoding = "2.1.2"
 uuid = { version = "1.18.1", default-features = false }
-v8 = "145"
+
+# Pinned to a specific version rather than allowing SemVer fuzzy resolution
+# because our flake references the specific version of this library.
+# See librusty_v8.nix for more details.
+# When updating the V8 crate, either update the version and hashes in that file,
+# or ping phoebe @gefjon to do so.
+v8 = "=145.0.0"
+
 walkdir = "2.2.5"
 wasmbin = "0.6"
 webbrowser = "1.0.2"
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@
         <img width="250" src="./images/light/logo-text.svg" alt="SpacetimeDB">
     </a>
     <h3 align="center">
-        Multiplayer at the speed of light.
+        Development at the speed of light.
     </h3>
 </p>
 <p align="center">
--- a/crates/bindings-typescript/package.json
+++ b/crates/bindings-typescript/package.json
@@ -102,7 +102,7 @@
       "name": "esm (gzip)",
       "path": "dist/index.mjs",
       "gzip": true,
-      "limit": "31 kB"
+      "limit": "40 kB"
     },
     {
       "name": "esm (uncompressed)",
--- a/docs/docs/00100-intro/00100-getting-started/00250-zen-of-spacetimedb.md
+++ b/docs/docs/00100-intro/00100-getting-started/00250-zen-of-spacetimedb.md
@@ -26,13 +26,17 @@ Traditional stack:        SpacetimeDB:
 
 ## Everything is Persistent
 
-You will ask, does everything need to be persistent? Won't that be a lot of data? Won't that be slow. The answers are yes, no, and no. SpacetimeDB persists everything by default, even the full history of everything that has every changed. It should be **your choice** to delete data, not the database's. SpacetimeDB gives you that choice.
+SpacetimeDB persists everything by default, including the full history of any rows that have ever changed.
+
+You will ask, does everything need to be persistent? Won't that be a lot of data? Well, you would be surprised! For example, updating 1 million player transforms 10 times per second for a year uses roughly 10 petabytes of data, uncompressed. SpacetimeDB can compress that sort of data by about 5-10x, meaning that keeping every position for every player for a game with a million concurrent players uses only about 1-2 petabytes per year. Storing that much data in Amazon S3 would only cost you between $2,300 and $5,600 per month. A fraction of the cost of a single engineer or data scientist!
+
+You can of course choose to delete the historical data, but it should be **your choice** to delete data, not the database's. SpacetimeDB gives you that choice.
+
+Won't it be slow to persist everything? No. SpacetimeDB is designed so that persistence guarantees only ever increase latency and never decrease throughput! Modern SSDs can write upwards of 15 GB/s of data to disk. DRAM can only do about 4x more. Let's actually use that Samsung-given bandwidth.
 
 SpacetimeDB holds all your data in memory for blazing-fast access, but automatically persists everything to disk. You get the speed of in-memory computing with the durability of a traditional database.
 
 You will be tempted to ask for "ephemeral state". This is a mistake. Persistent everything allows your app to recover to the *exact* state it was in. In principle, you could even debug your production app in the state it was in in the past with a time-traveling debugger.
-
-SpacetimeDB is designed so that persistence guarantees only ever increase latency and never decrease throughput! Modern SSDs can write upwards of 15 GB/s of data to disk. DRAM can only do about 4x more. Let's actually use that Samsung-given bandwidth.
 
 Write your code as if memory were infinite and permanent. Insert rows freely. Query without fear. SpacetimeDB handles the persistence, you handle the logic.
 
--- a/librusty_v8.nix
+++ b/librusty_v8.nix
@@ -22,14 +22,14 @@
     };
 in
   fetch_librusty_v8 {
-    version = "140.2.0";
+    version = "145.0.0";
     shas = {
-      x86_64-linux = "sha256-r3qrYDVaT4Z6udC6YuQG1BKqrsQc7IhuACDCTbr083U=";
+      x86_64-linux = "sha256-chV1PAx40UH3Ute5k3lLrgfhih39Rm3KqE+mTna6ysE=";
       # I (pgoldman 2025-10-17) only use x86_64-linux, so I haven't filled in these hashes.
       # If you use one of these platforms, run the build and wait for it to fail,
       # copy the detected sha256 from the error message in here, then re-run.
       aarch64-linux = "0000000000000000000000000000000000000000000000000000";
       x86_64-darwin = "0000000000000000000000000000000000000000000000000000";
-      aarch64-darwin = "sha256-eZ2l9ovI2divQake+Z4/Ofcl5QwJ+Y/ql2Dymisx1oA=";
+      aarch64-darwin = "0000000000000000000000000000000000000000000000000000";
     };
   }
--- a/sdks/csharp/.gitignore
+++ b/sdks/csharp/.gitignore
@@ -73,8 +73,8 @@ bin~
 obj~
 
 # This is used for local paths to SpacetimeDB packages.
-/nuget.config
-/nuget.config.meta
+[Nn][Uu][Gg][Ee][Tt].config
+[Nn][Uu][Gg][Ee][Tt].config.meta
 .idea/
 
 # Hydrated SDK DLLs (produced by `cargo ci dlls`)
--- a/sdks/csharp/tools~/write-nuget-config.sh
+++ b/sdks/csharp/tools~/write-nuget-config.sh
@@ -1,6 +1,6 @@
 set -ueo pipefail
 
-SPACETIMEDB_REPO_PATH="$1"
+SPACETIMEDB_REPO_PATH="$(readlink -f "$1")"
 
 cd "$(dirname "$(readlink -f "$0")")"
 cd ..


.github/workflows/ci.yml
```diff
@@ -19,7 +19,8 @@ concurrency:
 
 jobs:
   docker_smoketests:
-    needs: [lints]
+    # TODO: before merging re-enable this
+    # needs: [lints]
```
jdetter: Address this before merging

crates/smoketests/tests/smoketests/add_remove_index.rs
```diff
@@ -0,0 +1,89 @@
...
+/// First publish without the indices,
+/// then add the indices, and publish,
+/// and finally remove the indices, and publish again.
+/// There should be no errors
+/// and the unindexed versions should reject subscriptions.
+#[test]
+fn test_add_then_remove_index() {
+    let mut test = Smoketest::builder().module_code(MODULE_CODE).autopublish(false).build();
+
+    let name = format!("test-db-{}", std::process::id());
```
jdetter, line 18: This *would* have been an issue if we were to parallelize the smoketests but now we're running 1 spacetime server per test right?

crates/smoketests/tests/auto_inc.rs
```diff
@@ -0,0 +1,177 @@
...
+
+#[test]
+fn test_autoinc_i32() {
+    do_test_autoinc_basic("i32");
+}
+
+#[test]
+fn test_autoinc_i64() {
+    do_test_autoinc_basic("i64");
+}
```
jdetter: Instead of just testing these 4 here I would have 1 test which iterates over the array of types we used to test before:
```
"u8", "u16", "u32", "u64", "u128", "i8", "i16", "i32", "i64", "i128"
```

crates/smoketests/tests/auto_inc.rs
```diff
@@ -0,0 +1,177 @@
...
+
+#[test]
+fn test_autoinc_unique_u64() {
+    do_test_autoinc_unique("u64");
+}
+
+#[test]
+fn test_autoinc_unique_i64() {
+    do_test_autoinc_unique("i64");
+}
```
jdetter: same here, we're only test 2 of the 10 we used to test before.

crates/smoketests/src/lib.rs
```diff
@@ -0,0 +1,661 @@
...
+
+    /// Internal helper for publishing with options.
+    fn publish_module_opts(&mut self, name: Option<&str>, clear: bool) -> Result<String> {
+        let start = Instant::now();
+        let project_path = self.project_dir.path().to_str().unwrap().to_string();
+
+        // First, run spacetime build to compile the WASM module (separate from publish)
+        let build_start = Instant::now();
+        let cli_path = ensure_binaries_built();
+        let build_output = Command::new(&cli_path)
```
kim: This takes 8-10s on my machine for every test case.

Could we not, instead of a fresh tempdir for each test on every run, put all the module cargo projects under `CARGO_TARGET_TMPDIR`?

crates/smoketests/tests/smoketests/add_remove_index.rs
jdetter: This is actually kind of weird now that we're appending the process ID to the name - why not just use a static name like "test-db"?

crates/smoketests/tests/describe.rs
```diff
@@ -0,0 +1,44 @@
...
+
+/// Check describing a module
+#[test]
+fn test_describe() {
+    let test = Smoketest::builder().module_code(MODULE_CODE).build();
+
+    let identity = test.database_identity.as_ref().unwrap();
+
+    // Describe the whole module
+    test.spacetime(&["describe", "--json", identity]).unwrap();
```
jdetter: I know we weren't doing it before - but we're not doing any output validation here. Possibly worth a TODO.

crates/smoketests/tests/smoketests/new_user_flow.rs
```diff
@@ -0,0 +1,63 @@
+//! Tests translated from smoketests/tests/new_user_flow.py
```
jdetter: This test originally was testing to make sure that our tutorial isn't broken. Since our onboarding has changed we should probably update this test in the future. Probably worth dropping a TODO here.

jdetter: The 80/20 here might just be replacing this with testing the templates to be honest

crates/smoketests/tests/smoketests/add_remove_index.rs
```diff
@@ -0,0 +1,89 @@
+//! Add/remove index tests translated from smoketests/tests/add_remove_index.py
```
kim: I don't love that each test is linked into its own executable.
It's also possible to structure it like this:

```
tests/
  smoketests.rs
  smoketests/
    mod.rs
    add_remove_index.rs
    auto_inc.rs
    ...
```

Which will create a single executable and save a few seconds on link time.
The namespacing will also make it nicer to filter subsets of tests to run, which has been a frequent source of frustration with the python smoketests.

crates/guard/src/lib.rs
```diff
@@ -4,12 +4,79 @@ use std::{
...
 };
 
+/// Lazily-initialized path to the pre-built CLI binary.
+static CLI_BINARY_PATH: OnceLock<PathBuf> = OnceLock::new();
+
+/// Ensures `spacetimedb-cli` and `spacetimedb-standalone` are built once,
+/// returning the path to the CLI binary.
+///
+/// This is useful for tests that need to run CLI commands directly.
+pub fn ensure_binaries_built() -> PathBuf {
```
kim, line 65: Not so great about this approach is that there is no feedback as to what it is going on, all while the test runner tells the user that the tests are running for over 60s (in yellow even, when using nextest).

I'd also expect recursive cargo to create problems on certain systems, despite the env var hack.

Why not just create an [xtask](https://github.com/matklad/cargo-xtask) for running smoketests? Arguably, typing `cargo smoketests` isn't worse than `cargo test -p spacetimedb-smoketests`.
 
 
Invoking cargo from an xtask is fine, plus we can customize the output however we please.

.github/workflows/ci.yml
bfops: @jdetter they're not actually running via `cargo ci smoketests` right now.. also it uses the version of this workflow from the base PR so this didn't make a difference :sweat_smile:

bfops: okay wired up this command to actually run the new smoketests!

cloutiertyler: Nice!

crates/guard/src/lib.rs
cloutiertyler: We can fix up the feedback with `cargo nextest`, so at least that I can resolve.

I'm not opposed to creating an xtask per se, I just wanted it to be so that any schmo could run all of our tests without needing to know about our special special tools (i.e. `cargo test` runs all the tests), but I could be convinced.

In any case this PR is not a regression since I'm largely reusing existing machinery. I would request that we change that up in a separate PR.

crates/smoketests/src/lib.rs
cloutiertyler: I'm investigating this, because indeed building the modules takes the vast majority of the time of the smoke tests.

Not for nothing but also rewriting the smoke tests to run in TypeScript would publish them much faster. Although probably we should have both languages so that we're smoke testing v8.

crates/guard/src/lib.rs
cloutiertyler: Okay, I spent essentially my whole Saturday on this problem. There are the following possible solutions, none of them great:

1. A separate xtask specifically for the smoke tests
    - This doesn't solve the problem of the CLI tests which also use the `guard` crate
    - Idiosyncratic way to run the smoke tests as opposed to cargo test
    - Does NOT resolve the cargo within cargo problem because we still have to build the test modules with cargo
2. A `build.rs` script in `guard` which first builds the executables as a compile step for compiling guard
    - Deadlocks on a cargo lock file conflict (Outer cargo compiles guard → runs build.rs, inner cargo tries to acquire the build directory lock, outer cargo holds the directory lock, deadlock)
    - If you fix the deadlock by using different target dirs, it still looks stuck on building guard because it's actually compiling all of spacetimedb-standalone and spacetimedb-cli. 
    - Still technically runs cargo inside of cargo.
3. Add `spacetimedb-cli` and `spacetimedb-standalone` as an artifact dependency of the guard crate
    - Has good and clear output but requires +nightly when running the smoketests and CLI tests, otherwise won't do the right thing. See https://github.com/rust-lang/cargo/issues/9096
4. Keep what we have which compiles the executables during the tests themselves where the first test takes a lock while the executables are building
    - Makes the tests look like they're taking a long time when they're just waiting for the build to complete
    - Requires relatively complex locking machinery across binaries/tests/processes
5. A two step solution where the developer has to build the binaries before calling the smoke tests
    - Very error prone

None of these are good. `xtask` is not bad, but doesn't enable us to run other integration tests in other crates (e.g. the CLI)

(3) is the correct solution and has the best user experience, but it requires nightly and I don't want to introduce that for all of our tests.

~(1) xtask, is the least bad, so I'm going with that. You will now run the smoketests with `cargo smoketest`. If you run `cargo test --all` (or use `guard`) without doing `cargo smoketest` and the binaries are missing then the smoketests will fail. If the binaries are built, we will detect whether or not the binaries are stale, if the binaries are stale then the smoketests will fail. Note that this is a BREAKING CHANGE, the smoketests are now part of `cargo test --all` unless you specifically exclude them.~

See new PR description.

.cargo/config.toml
```diff
@@ -5,6 +5,7 @@ rustflags = ["--cfg", "tokio_unstable"]
 bump-versions = "run -p upgrade-version --"
 llm = "run --package xtask-llm-benchmark --bin llm_benchmark --"
 ci = "run -p ci --"
+smoketest = "run -p xtask-smoketest -- smoketest"
```
bfops, line 8: I propose that instead of a new toplevel command, we just update `cargo ci smoketests` to run these commands. `cargo ci` is supposed to be the toplevel command that has all the specialty logic for running tests.

.github/workflows/ci.yml
```diff
@@ -18,26 +18,24 @@ concurrency:
...
           - runner: windows-latest
-            smoketest_args: --no-build-cli
             container: null
     runs-on: ${{ matrix.runner }}
     container: ${{ matrix.container }}
     timeout-minutes: 120
-    env:
-      CARGO_TARGET_DIR: ${{ github.workspace }}/target
+    #env:
+    #  CARGO_TARGET_DIR: ${{ github.workspace }}/target
```
bfops: this was added by me for debugging. we should try reverting this to see how it impacts the smoketests. If we leave it commented out, we should add a comment above it about the impact.

.github/workflows/ci.yml
```diff
@@ -84,41 +82,118 @@ jobs:
...
+      # However, we've been unable to fix it so... /shrug
+      - name: Check v8 outputs
+        shell: bash
+        run: |
+          find "${CARGO_TARGET_DIR}"/ -type f | grep '[/_]v8' || true
+          if ! [ -f "${CARGO_TARGET_DIR}"/debug/gn_out/obj/librusty_v8.a ]; then
+            echo "Could not find v8 output file librusty_v8.a; rebuilding manually."
+            cargo clean -p v8 || true
+            cargo build -p v8
+          fi
```
bfops, line 118: I added this because the CI was failing on the now-familiar v8 issues. I don't know why, but "something something caching" is our hypothesis

.github/workflows/ci.yml
```diff
@@ -84,41 +82,118 @@ jobs:
...
+          $ErrorActionPreference = 'Stop'
+          $PSNativeCommandUseErrorActionPreference = $true
+          cd modules
+          dotnet workload config --update-mode manifests
+          dotnet workload update
+
+      - name: Run Python smoketests
+        shell: bash
+        run: |
+          python -m smoketests
```
bfops: I think this workflow probably still needs most of the accommodations that existed before, e.g. skipping several tests, starting the server / docker container, etc.

.github/workflows/ci.yml
```diff
@@ -890,3 +965,34 @@ jobs:
...
+        with:
+          fetch-depth: 0
+
+      - name: Fail if Python smoketests were modified
+        run: |
+          # Get changed files in the PR
+          CHANGED_FILES=$(git diff --name-only origin/${{ github.base_ref }}...HEAD)
+
+          # Check if any files in smoketests/ (Python) directory were changed
+          PYTHON_SMOKETEST_CHANGES=$(echo "$CHANGED_FILES" | grep -E "^smoketests/" || true)
```
bfops: what about:
```suggestion
          PYTHON_SMOKETEST_CHANGES=$(git diff --name-only origin/${{ github.base_ref }} HEAD -- 'smoketests/**.py')
```
(I added `.py` to prevent things like README changes from triggering this)

.github/workflows/ci.yml
bfops: (I'm happy to make these changes, I just don't want to merge conflict with anything you might be doing)

tools/xtask-smoketest/src/main.rs
```diff
@@ -0,0 +1,167 @@
+#![allow(clippy::disallowed_macros)]
```
bfops, line 1: I'm fine with this the way it is to get the PR merged, but this feels like it "should" just be a part of `tools/ci`

bfops: oh wait then it would be running under `cargo` again, right. sigh.

bfops: although I think that's still true if we run via `cargo smoketests`?

.cargo/config.toml
cloutiertyler: Well, I thought about this, but that's not true of cargo test, which is also meant to run all the testing. My thought was that it should *also* live under `cargo ci smoketest` which it currently does.

.github/workflows/ci.yml
cloutiertyler: Should I remove it?

.github/workflows/ci.yml
cloutiertyler: I can make the changes

tools/xtask-smoketest/src/main.rs
cloutiertyler: Apparently it's only an issue to build `spacetimedb-standalone` within Cargo according to Phoebe.

crates/smoketests/tests/smoketests/cli/mod.rs
```diff
@@ -0,0 +1,4 @@
+// CLI integration tests moved from crates/cli/tests/
```
bfops: I don't think this should be in a comment
```suggestion
```

crates/smoketests/tests/smoketests/cli/publish.rs
```diff
@@ -1,30 +1,46 @@
-use assert_cmd::cargo::cargo_bin_cmd;
-use spacetimedb_guard::SpacetimeDbGuard;
+//! CLI publish command tests moved from crates/cli/tests/publish.rs
```
bfops:
```suggestion
//! CLI publish command tests
```

crates/smoketests/tests/smoketests/cli/dev.rs
```diff
@@ -0,0 +1,87 @@
+//! CLI dev command tests moved from crates/cli/tests/dev.rs
```
bfops:
```suggestion
//! CLI dev command tests
```

crates/smoketests/TEST_DURATIONS.md
```diff
@@ -0,0 +1,92 @@
+# Smoketest Duration Report
```
bfops: is this file meant to be committed? is it meant to be updated?

.github/workflows/ci.yml
bfops: I don't think so, just noting for any reviewers

crates/smoketests/src/modules.rs
```diff
@@ -0,0 +1,209 @@
...
+    target.exists() && target.join("smoketest_module_filtering.wasm").exists()
+}
+
+/// Returns the target directory where pre-compiled WASM modules are stored.
+fn modules_target_dir() -> PathBuf {
+    workspace_root().join("crates/smoketests/modules/target/wasm32-unknown-unknown/release")
+}
+
+/// Builds the registry mapping module names to WASM paths.
+fn build_registry() -> HashMap<&'static str, PathBuf> {
```
bfops: It makes me sad that we need to register new modules in `modules/Cargo.toml` as well as here. I don't have a better idea though.

bfops: also AI can do it for us so :shrug:

bfops: perhaps something like we used to do for the docs, where we have a check that ensures that all directories in `modules/` are registered in `build_registry`?

bfops: not super important to do right now, I'd rather get this merged.

crates/smoketests/tests/smoketests/client_connection_errors.rs
```diff
@@ -0,0 +1,50 @@
...
+    let result = test.subscribe(&["SELECT * FROM all_u8s"], 0);
+    assert!(result.is_ok(), "Expected subscribe to succeed");
+
+    let logs = test.logs(100).unwrap();
+    assert!(
+        logs.iter()
+            .any(|l| { l.contains("This should be called, but the `st_client` row should still be deleted") }),
+        "Expected disconnect panic message in logs: {:?}",
+        logs
+    );
```
bfops, line 47: I think we're missing the part that checks the `st_client` results

crates/smoketests/modules/confirmed-reads/src/lib.rs
```diff
@@ -0,0 +1,11 @@
...
+
+#[spacetimedb::table(name = person, public)]
+pub struct Person {
+    name: String,
+}
+
+#[spacetimedb::reducer]
+pub fn add(ctx: &ReducerContext, name: String) {
+    ctx.db.person().insert(Person { name });
+}
```
bfops, line 11: note that this is technically a different module than the python smoketests uses (the original smoketests used the `basic-rs` template), but the change seems entirely benign to me

.cargo/config.toml
bfops: it is true for `cargo test`, isn't it? `cargo ci test`? That's at least what that's _meant_ to be.

bfops: I don't personally see the value in a new toplevel command when `ci smoketests` is right there, but I'm happy to keep it if you think it adds value.

tools/xtask-smoketest/src/main.rs
bfops: +1. Separately, I also think `xtask-smoketest` is just as much "in cargo" as `cargo ci` is, so I do think we could merge that in eventually. Not a priority right now, but perhaps worth a `TODO` or a ticket.

Feel free to resolve when you're satisfied :+1:

crates/smoketests/tests/smoketests/add_remove_index.rs
bfops: I have the same questions. I do feel like the old random-string method was the most trivially "this should be fine" approach.

bfops:
I've confirmed that the list of tests itself looks correctly-migrated:
```
$ diff <(ls smoketests/tests/ | sed 's/.py$//') <(ls crates/smoketests/tests/smoketests/ | sed 's/.rs$//')
5,6c5,6
< clear_database
< client_connected_error_rejects_connection
---
> cli
> client_connection_errors
20c20
< __init__
---
> mod
29c29
< replication
---
> restart
34d33
< teams
37d35
< zz_docker
```
- removed `clear_database` intentionally
- `cli` was never in python
- `client_connected_*` became `client_connection_errors`
- `__init__` becomes `mod`, that's fine
- removed `replication` intentionally
- `zz_docker` became `restart`
- removed `teams` intentionally

crates/smoketests/tests/smoketests/default_module_clippy.rs
```diff
@@ -0,0 +1,24 @@
+//! Tests translated from smoketests/tests/default_module_clippy.py
+
+use spacetimedb_smoketests::Smoketest;
+use std::process::Command;
+
+/// Ensure that the default rust module has no clippy errors or warnings
+#[test]
+fn test_default_module_clippy_check() {
+    // Build a smoketest with the default module code (no custom code)
+    let test = Smoketest::builder().autopublish(false).build();
```
bfops: I believe this is testing completely different module code than the Python version did?

It seems like the spirit of this test would be to run `cargo clippy` on all the Rust templates, or something. I believe the Python version basically tests this for the `basic-rs` template.

crates/smoketests/tests/smoketests/new_user_flow.rs
bfops: I just made somewhat-related comments on `default_module_clippy.rs`: https://github.com/clockworklabs/SpacetimeDB/pull/4102/files#r2729684732

crates/smoketests/tests/smoketests/domains.rs
```diff
@@ -0,0 +1,78 @@
...
+        &test.server_url,
+        "--to",
+        &rename_to,
+        &id_to_rename,
+    ]);
+    assert!(
+        result.is_err(),
+        "Expected rename to fail when target name is already in use"
+    );
+}
```
bfops, line 76: I think we are missing `test_replace_names` from `domains.py`?

crates/smoketests/tests/smoketests/modules.rs
```diff
@@ -0,0 +1,86 @@
...
+    assert!(logs.iter().any(|l| l.contains("Hello, Samantha!")));
+    assert!(logs.iter().any(|l| l.contains("Hello, Julie!")));
+    assert!(logs.iter().any(|l| l.contains("Hello, Robert!")));
+    assert!(logs.iter().any(|l| l.contains("Hello, World!")));
+
+    // Unchanged module is ok
+    test.publish_module_named(&name, false).unwrap();
+
+    // Changing an existing table isn't
+    test.write_module_code(MODULE_CODE_BREAKING).unwrap();
```
bfops: we have a precompiled module for this - is there a reason we're not using it? (should we remove it?)

crates/smoketests/tests/smoketests/modules.rs
```diff
@@ -0,0 +1,86 @@
...
+    test.call("add", &["Julie"]).unwrap();
+    test.call("add", &["Samantha"]).unwrap();
+    test.call("say_hello", &[]).unwrap();
+
+    let logs = test.logs(100).unwrap();
+    assert!(logs.iter().any(|l| l.contains("Hello, Samantha!")));
+    assert!(logs.iter().any(|l| l.contains("Hello, Julie!")));
+    assert!(logs.iter().any(|l| l.contains("Hello, Robert!")));
+    assert!(logs.iter().any(|l| l.contains("Hello, World!")));
+}
```
bfops, line 72: I think we're missing the `UploadModule2` and `HotswapModule` tests from the original python

crates/smoketests/modules/namespaces/src/lib.rs
```diff
@@ -0,0 +1,34 @@
+use spacetimedb::{ReducerContext, Table};
```
bfops, line 1: I don't think this one is actually used?

crates/smoketests/tests/integration.rs
```diff
@@ -0,0 +1,2 @@
+// Single test binary entry point - includes all smoketests
+mod smoketests;
```
bfops, line 2: what's the reasoning behind this extra nested module?

bfops: not a big deal, but curious

crates/smoketests/tests/smoketests/add_remove_index.rs
```diff
@@ -0,0 +1,49 @@
...
+    // Subscription should work now (n=0 just verifies the query is accepted)
+    let result = test.subscribe(&[JOIN_QUERY], 0);
+    assert!(
+        result.is_ok(),
+        "Expected subscription to succeed with indices, got: {:?}",
+        result.err()
+    );
+
+    // Verify call works too
+    test.call("add", &[]).unwrap();
```
bfops: the python version "held" the subscription across this `call` - should this version do that as well?

crates/smoketests/tests/smoketests/permissions.rs
```diff
@@ -0,0 +1,80 @@
...
+    let events = sub.collect().unwrap();
+    assert_eq!(events.len(), 1, "Expected 1 update, got {:?}", events);
+
+    let expected = serde_json::json!({
+        "common_knowledge": {
+            "deletes": [],
+            "inserts": [{"thing": "godmorgon"}]
+        }
+    });
+    assert_eq!(events[0], expected);
```
bfops: the python test also has this at the end:
```
        # Subscribing to both tables returns updates for the public one.
        sub = self.subscribe("SELECT * FROM *", n=1)
        self.call("do_thing", "howdy", anon=True)
        self.assertEqual(sub(), [
            {
                'common_knowledge': {
                    'deletes': [],
                    'inserts': [{'thing': 'howdy'}]
                }
            }
        ])
```

crates/smoketests/tests/smoketests/permissions.rs
```diff
@@ -0,0 +1,80 @@
...
+    for kind in lifecycle_kinds {
+        let reducer_name = format!("lifecycle_{}", kind);
+        let result = test.call(&reducer_name, &[]);
+        assert!(
+            result.is_err(),
+            "Expected call to lifecycle reducer '{}' to fail",
+            reducer_name
+        );
+    }
+}
```
bfops, line 214: we seem to be missing `test_call`, `test_logs`, `test_describe`, `test_publish`, and `test_replace_names` from the original python

crates/smoketests/tests/smoketests/rls.rs
```diff
@@ -0,0 +1,32 @@
...
+    );
+
+    // Create another new identity - should see no users
+    test.new_identity().unwrap();
+    test.assert_sql(
+        "SELECT name FROM users",
+        r#" name
+------"#,
+    );
+}
```
bfops, line 30: I think we're missing `BrokenRls` and `DisconnectRls` from the original python

.cargo/config.toml
cloutiertyler: I mean `cargo test` exists in addition to `cargo ci test`, so my thought was that it would be the same for `cargo smoketest`.

So that's why we have both:

`cargo smoketest`
`cargo ci smoketest`

Where `cargo ci smoketest` just calls `cargo ci`.

crates/guard/src/lib.rs
cloutiertyler: Resolving this comment because we went with the xtask.

crates/smoketests/TEST_DURATIONS.md
cloutiertyler: I was considering it, but I think no actually. I'll just pull it out. It was meant to give people a baseline idea.

tools/xtask-smoketest/src/main.rs
cloutiertyler: We should chat about our naming scheme, but for now I'd like to merge this. We also have `cargo ci smoketest` so we can just use that and remove this later if we decide to do so.

crates/smoketests/TEST_DURATIONS.md
cloutiertyler: Removed.

crates/smoketests/tests/integration.rs
cloutiertyler: It was a suggestion from Kim in order to have a single executable to save linking time. Also, having this mod defined this way is required in tests for some reason. I tried with with

```
smoketests.rs
smoketests/
```

But that does not work in tests.

crates/smoketests/src/modules.rs
cloutiertyler: Yes I agree. I decided to just modify this to derive the `.wasm` file name from the test name, so we can just remove this mapping.

.github/workflows/ci.yml
```diff
@@ -890,3 +965,34 @@
...
+            echo ""
+            echo "Changed files:"
+            echo "$PYTHON_SMOKETEST_CHANGES"
+            echo ""
+            echo "The Python smoketests are being replaced by Rust smoketests."
+            echo "See crates/smoketests/DEVELOP.md for instructions on adding Rust smoketests."
+            exit 1
+          fi
+
+          echo "No Python smoketest changes detected."
```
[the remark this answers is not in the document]
cloutiertyler: Done.

crates/smoketests/tests/smoketests/add_remove_index.rs
cloutiertyler: @jdetter one reason to not use a static name is running the tests against a remote server. Now you can run the tests multiple times without conflicting database names.

crates/smoketests/tests/auto_inc.rs
cloutiertyler: Fixed.

crates/smoketests/tests/auto_inc.rs
cloutiertyler: Fixed.

crates/smoketests/tests/describe.rs
cloutiertyler: Added a comment.

crates/smoketests/tests/smoketests/new_user_flow.rs
cloutiertyler: I think this should be done in a separate PR. I don't think this is related specifically to the translation of the smoketests.

crates/smoketests/src/lib.rs
cloutiertyler: I spent quite a long time on this problem. The way it works now is that we build all the modules simultaneously by putting them in a cargo workspace and running cargo build on that whole workspace.

.github/workflows/ci.yml
cloutiertyler: This has been reverted. I had to do a small fix, but now it's fine.

.github/workflows/ci.yml
cloutiertyler: I have made these changes.

crates/smoketests/modules/confirmed-reads/src/lib.rs
cloutiertyler: Roger. Yeah I think this is because of the move to precompiled modules.

crates/smoketests/tests/smoketests/default_module_clippy.rs
cloutiertyler: I agree. I'm updating this now. I'm going to have this check all the templates and all the modules in /modules

crates/smoketests/tests/smoketests/cli/mod.rs
cloutiertyler: Fixed.

crates/smoketests/tests/smoketests/cli/dev.rs
cloutiertyler: Fixed.

crates/smoketests/tests/smoketests/client_connection_errors.rs
cloutiertyler: Added this in now.

crates/smoketests/tests/smoketests/default_module_clippy.rs
cloutiertyler: This is now done. I also fixed the version number issue in `basic-rs`.

crates/smoketests/tests/smoketests/domains.rs
cloutiertyler: Fixed.

crates/smoketests/tests/smoketests/modules.rs
cloutiertyler: Claude thought it wouldn't compile but it's a runtime breaking thing.

cloutiertyler: Fixed.

crates/smoketests/modules/namespaces/src/lib.rs
cloutiertyler: It is now used.

crates/smoketests/tests/smoketests/permissions.rs
cloutiertyler: Added.

crates/smoketests/tests/smoketests/rls.rs
cloutiertyler: Added.

crates/smoketests/tests/smoketests/permissions.rs
cloutiertyler: Added.

crates/smoketests/tests/smoketests/modules.rs
cloutiertyler: Added.

crates/smoketests/tests/smoketests/add_remove_index.rs
cloutiertyler: I believe I've fixed this. Does this look correct to you?

crates/smoketests/tests/smoketests/modules.rs
```diff
@@ -84,3 +72,86 @@ fn test_upload_module() {
...
+        second.get("person").is_some(),
+        "Expected person table in second update: {:?}",
+        second
+    );
+    let inserts = &second["person"]["inserts"];
+    assert!(
+        inserts.as_array().unwrap().iter().any(|r| r["name"] == "Cindy"),
+        "Expected Cindy in second update: {:?}",
+        second
+    );
```
bfops, line 132: technically, the original tests also check things like the id of the entity, and that `deletes` is empty. unsure if there's an easy way to do a full structural comparison.

crates/smoketests/tests/smoketests/namespaces.rs
```diff
@@ -61,20 +34,24 @@ fn count_matches(dir: &Path, needle: &str) -> usize {
 /// Ensure that the default namespace is working properly
 #[test]
 fn test_spacetimedb_ns_csharp() {
-    let test = Smoketest::builder().module_code(MODULE_CODE).autopublish(false).build();
+    let _test = Smoketest::builder()
```
bfops: out of curiosity, why `_test`?

crates/smoketests/tests/smoketests/views.rs
```diff
@@ -0,0 +1,154 @@
...
+#[spacetimedb::table(name = person, public)]
+pub struct Person {
+    name: String,
+}
+
+#[spacetimedb::view(name = person, public)]
+pub fn person(ctx: &ViewContext) -> Option<Person> {
+    None
+}
+"#;
```
bfops: there's a precompiled module for this

crates/smoketests/tests/smoketests/views.rs
```diff
@@ -0,0 +1,154 @@
...
+    A,
+    B,
+    C,
+}
+
+#[spacetimedb::view(name = person, public)]
+pub fn person(ctx: &ViewContext) -> Option<ABC> {
+    None
+}
+"#;
```
bfops: there's a precompiled module for this

crates/smoketests/tests/smoketests/views.rs
```diff
@@ -0,0 +1,154 @@
...
+ 2  | 2"#,
+    );
+
+    test.assert_sql(
+        "SELECT * FROM player_and_level WHERE id = 2",
+        r#" id | level
+----+-------
+ 2  | 2"#,
+    );
+}
```
bfops, line 229: we're missing several tests from the python `smoketests/tests/views.py`

crates/smoketests/tests/smoketests/servers.rs
```diff
@@ -0,0 +1,91 @@
...
+            "testnet",
+            "--no-fingerprint",
+        ])
+        .unwrap();
+
+    assert!(
+        output.contains("testnet.spacetimedb.com"),
+        "Expected host in output: {}",
+        output
+    );
```
bfops, line 32: nit: the python version checks the host and protocol

crates/smoketests/tests/smoketests/servers.rs
```diff
@@ -0,0 +1,91 @@
...
+
+    // List servers (local-only command)
+    let servers = test.spacetime(&["server", "list"]).unwrap();
+
+    let testnet_re = Regex::new(r"(?m)^\s*testnet\.spacetimedb\.com\s+https\s+testnet\s*$").unwrap();
+    assert!(
+        testnet_re.is_match(&servers),
+        "Expected testnet in server list: {}",
+        servers
+    );
```
bfops, line 42: nit: the python version also tests for localhost

crates/smoketests/tests/smoketests/servers.rs
```diff
@@ -0,0 +1,91 @@
...
+    .unwrap();
+
+    // Check fingerprint commands (local-only command)
+    let output = test.spacetime(&["server", "fingerprint", "test-local", "-y"]).unwrap();
+    // The exact message may vary, just check it doesn't error
+    assert!(
+        output.contains("fingerprint") || output.contains("Fingerprint"),
+        "Expected fingerprint message: {}",
+        output
+    );
```
bfops, line 62: nit: the python version tests fingerprinting of ip address and the presence/absence of `-y` as well. that could maybe be a separate test.

crates/smoketests/tests/smoketests/new_user_flow.rs
bfops:
```suggestion
// TODO: This test originally was testing to make sure that our tutorial isn't broken. Since our onboarding has changed we should probably update this test in the future.
```

crates/smoketests/tests/smoketests/new_user_flow.rs
```diff
@@ -0,0 +1,42 @@
+use spacetimedb_smoketests::Smoketest;
+
+/// Test the entirety of the new user flow.
```
bfops, line 4:
```suggestion
// TODO: This test originally was testing to make sure that our tutorial isn't broken. Since our onboarding has changed we should probably update this test in the future.
/// Test the entirety of the new user flow.
```

crates/smoketests/tests/smoketests/schedule_reducer.rs
```diff
@@ -0,0 +1,85 @@
...
+
+    let logs = test.logs(100).unwrap();
+    let invoked_count = logs.iter().filter(|line| line.contains("Invoked:")).count();
+    assert!(
+        invoked_count > 2,
+        "Expected repeated reducer to run more than twice, but it ran {} times. Logs: {:?}",
+        invoked_count,
+        logs
+    );
+}
```
bfops, line 203: I think we're missing the procedure tests from The Original Python

crates/smoketests/tests/smoketests/schedule_reducer.rs
```diff
@@ -0,0 +1,85 @@
...
+    // Wait for the scheduled reducer to run
+    thread::sleep(Duration::from_secs(2));
+
+    let logs = test.logs(100).unwrap();
+    let invoked_count = logs.iter().filter(|line| line.contains("Invoked:")).count();
+    assert_eq!(
+        invoked_count, 1,
+        "Expected scheduled reducer to run exactly once, but it ran {} times. Logs: {:?}",
+        invoked_count, logs
+    );
```
bfops, line 46: The Original Python also has:
```py

        row_entry = {
            "prev": TIMESTAMP_ZERO,
            "scheduled_id": 2,
            "sched_at": {"Time": TIMESTAMP_ZERO},
        }
        # subscription should have 2 updates, first for row insert in scheduled table and second for row deletion.
        self.assertEqual(
            sub(),
            [
                {"scheduled_table": {"deletes": [], "inserts": [row_entry]}},
                {"scheduled_table": {"deletes": [row_entry], "inserts": []}},
            ],
        )
```

crates/smoketests/tests/smoketests/schedule_reducer.rs
bfops, line 203: The Original Python also has this at the end:
```py
        # scheduling repeated reducer again just to get 2nd subscription update.
        self.call("schedule_reducer")

        repeated_row_entry = {
            "prev": TIMESTAMP_ZERO,
            "scheduled_id": 1,
            "sched_at": {"Interval": {"__time_duration_micros__": 100000}},
        }
        row_entry = {
            "prev": TIMESTAMP_ZERO,
            "scheduled_id": 2,
            "sched_at": {"Time": TIMESTAMP_ZERO},
        }

        # subscription should have 2 updates and should not have any deletes
        self.assertEqual(
            sub(),
            [
                {"scheduled_table": {"deletes": [], "inserts": [repeated_row_entry]}},
                {"scheduled_table": {"deletes": [], "inserts": [row_entry]}},
            ],
        )
```

crates/smoketests/tests/smoketests/schedule_reducer.rs
```diff
@@ -0,0 +1,85 @@
...
+    // Wait a moment for the scheduled insert to complete
+    thread::sleep(Duration::from_millis(500));
+
+    // Query the table to verify both inserts happened
+    let result = test.sql("SELECT * FROM my_table").unwrap();
+    assert!(
+        result.contains("yay!") && result.contains("hello"),
+        "Expected both 'yay!' and 'hello' in table, got: {}",
+        result
+    );
```
bfops, line 225: the original code did a background subscription and then checked it after the calls. I'm not sure whether that was important to what was being tested (i.e. if there was a specific worry about `volatile_nonatomic_schedule_immediate` wasn't updating subscriptions properly)

crates/smoketests/tests/smoketests/rls.rs
bfops: They look good to me!

crates/smoketests/tests/smoketests/auto_migration.rs
```diff
@@ -0,0 +1,264 @@
...
+        logs.iter().any(|l| l.contains("AFTER_PERSON: Husserl - Professor")),
+        "Expected Husserl Professor in AFTER logs: {:?}",
+        logs
+    );
+    assert!(
+        logs.iter().any(|l| l.contains("AFTER_BOOK: 1234567890")),
+        "Expected book ISBN in AFTER logs: {:?}",
+        logs
+    );
+}
```
bfops, line 275: I believe we're missing the `AddTableColumns` class equivalents from The Original Python

crates/smoketests/tests/smoketests/auto_migration.rs
```diff
@@ -0,0 +1,264 @@
...
+        logs.iter().any(|l| l.contains("BEFORE: Julie - Student")),
+        "Expected Julie in logs: {:?}",
+        logs
+    );
+    assert!(
+        logs.iter().any(|l| l.contains("BEFORE: Robert - Student")),
+        "Expected Robert in logs: {:?}",
+        logs
+    );
+
```
bfops, line 228: we don't print the log messages that The Original Python prints, unsure if intentional

crates/smoketests/tests/smoketests/auto_migration.rs
```diff
@@ -0,0 +1,264 @@
...
+    // Update module without clearing database
+    test.write_module_code(MODULE_CODE_UPDATED).unwrap();
+    test.publish_module_clear(false).unwrap();
+
+    // Add new data with updated schema
+    test.call("add_person", &["Husserl", "Student"]).unwrap();
```
bfops, line 234: The Original Python does this after this line:
```py
        # If subscription, we should get 4 rows corresponding to 4 reducer calls (including before and after update)
        sub = sub();
        self.assertEqual(len(sub), 4)
```

crates/smoketests/tests/smoketests/pg_wire.rs
```diff
@@ -0,0 +1,126 @@
...
+  1 | {"Gray": 128}
+(1 row)"#,
+    );
+}
+
+/// Test failure cases
+#[test]
+fn test_failures() {
+    if !have_psql() {
+        eprintln!("Skipping test_failures: psql not available");
```
bfops: should this fail rather than quietly-ish skipping?

crates/smoketests/tests/smoketests/pg_wire.rs
```diff
@@ -0,0 +1,126 @@
...
+    // Empty query returns empty result
+    let output = test.psql("quickstart", "").unwrap();
+    assert!(
+        output.is_empty(),
+        "Expected empty output for empty query, got: {}",
+        output
+    );
+
+    // Connection fails with invalid token - we can't easily test this without
+    // modifying the token, so skip this part
```
bfops: this case seems worth testing

crates/smoketests/tests/smoketests/pg_wire.rs
```diff
@@ -0,0 +1,126 @@
...
+    test.assert_psql(
+        "quickstart",
+        "SELECT * FROM t_enum",
+        r#"id |     color
+----+---------------
+  1 | {"Gray": 128}
+(1 row)"#,
+    );
+}
+
```
bfops, line 76: I think we're missing a `test_sql_conn` equivalent to The Original Python

crates/smoketests/tests/smoketests/csharp_module.rs
```diff
@@ -0,0 +1,113 @@
...
+use std::fs;
+use std::process::Command;
+
+/// Ensure that the CLI is able to create and compile a C# project.
+/// This test does not depend on a running SpacetimeDB instance.
+/// Skips if dotnet 8.0+ is not available.
+#[test]
+fn test_build_csharp_module() {
+    if !have_dotnet() {
+        eprintln!("Skipping test_build_csharp_module: dotnet 8.0+ not available");
```
bfops: I would prefer if this failed instead of skipping, forcing the user to pass a `--skip-dotnet` or similar like The Original Python did. That way, we can tell the difference between the "tests not failing because not run" and "tests not failing because behavior is correct" cases.

crates/smoketests/tests/smoketests/csharp_module.rs
```diff
@@ -0,0 +1,113 @@
...
+        return;
+    }
+
+    let workspace = workspace_root();
+    let bindings = workspace.join("crates/bindings-csharp");
+    // CLI is pre-built by artifact dependencies during compilation
+    let cli_path = ensure_binaries_built();
+
+    // Install wasi-experimental workload
```
bfops, line 26: python did this before this step:
```
            run_cmd("dotnet", "nuget", "locals", "all", "--clear", cwd=bindings, capture_stderr=True)
```
unsure if the change is intentional

crates/smoketests/src/modules.rs
```diff
@@ -0,0 +1,121 @@
...
+            continue;
+        }
+
+        // Extract module name: smoketest_module_foo_bar.wasm -> foo-bar
+        let module_name = filename
+            .strip_prefix("smoketest_module_")
+            .unwrap()
+            .strip_suffix(".wasm")
+            .unwrap()
+            .replace('_', "-");
```
bfops: nit: perhaps we filter out lines 87-98 into a function, and then call that function in the `test`?

crates/smoketests/src/modules.rs
```diff
@@ -0,0 +1,121 @@
...
+            registry.keys().collect::<Vec<_>>()
+        )
+    })
+}
+
+/// Returns true if pre-compiled modules are available.
+///
+/// This checks if the modules workspace target directory exists and contains
+/// at least one WASM file.
+pub fn precompiled_modules_available() -> bool {
```
bfops: is this used?

.github/workflows/ci.yml
```diff
@@ -906,3 +1011,32 @@ jobs:
...
+      contents: read
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Fail if Python smoketests were modified
+        run: |
+          PYTHON_SMOKETEST_CHANGES=$(git diff --name-only origin/${{ github.base_ref }} HEAD -- 'smoketests/**.py')
```
bfops: heads up, I realized that this will start failing in this PR if anyone merges changes into master that touch the smoketests, even if you don't merge master into this PR. I don't know that that's a bad thing, just wanted to mention it before it's confusing.

if we wanted only-changed-in-this-PR logic, it would have to be:
```bash
MERGE_BASE=$(git merge-base origin/${{ github.base_ref }} HEAD)
PYTHON_SMOKETEST_CHANGES="$(git diff --name-only $MERGE_BASE HEAD -- 'smoketests/**.py')"
```

also we probably want to make this check required just before merging - I suggest leaving this comment unresolved until then.

tools/xtask-smoketest/src/main.rs
```diff
@@ -0,0 +1,197 @@
...
+        "-p",
+        "spacetimedb-standalone",
+    ]);
+
+    // Remove cargo/rust env vars that could cause fingerprint mismatches
+    // when the test later runs cargo build from a different environment
+    for (key, _) in env::vars() {
+        let should_remove = (key.starts_with("CARGO") && key != "CARGO_HOME" && key != "CARGO_TARGET_DIR")
+            || key.starts_with("RUST")
+            || key == "__CARGO_FIX_YOLO";
```
bfops, line 94: ?

tools/xtask-smoketest/src/main.rs
```diff
@@ -0,0 +1,197 @@
...
+            .iter()
+            .any(|a| a == "-j" || a.starts_with("-j") || a.starts_with("--jobs"))
+        {
+            cmd.args(["-j", DEFAULT_PARALLELISM]);
+        }
+
+        // Set remote server environment variable if specified
+        if let Some(ref server_url) = server {
+            cmd.env("SPACETIME_REMOTE_SERVER", server_url);
+        }
```
bfops: nit: this could be moved out of the `if` entirely

crates/guard/src/lib.rs
```diff
@@ -4,18 +4,106 @@ use std::{
...
 };
 
+/// Global counter for spawn IDs to correlate log messages across threads.
+static SPAWN_COUNTER: AtomicU64 = AtomicU64::new(0);
+
+fn next_spawn_id() -> u64 {
+    SPAWN_COUNTER.fetch_add(1, Ordering::SeqCst)
+}
+
+/// Returns the workspace root directory.
```
bfops, line 24:
```suggestion
/// Returns the workspace root directory.
// TODO: Should this use something like `git rev-parse --show-toplevel` to avoid being directory-relative?
```

crates/guard/src/lib.rs
```diff
@@ -4,18 +4,107 @@ use std::{
...
+// TODO: Should this use something like `git rev-parse --show-toplevel` to avoid being directory-relative?
```
bfops:
```suggestion
// TODO: Should this use something like `git rev-parse --show-toplevel` to avoid being directory-relative? Or perhaps `CARGO_WORKSPACE_DIR` is set?
```

.github/workflows/ci.yml
```diff
@@ -18,9 +18,107 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  docker_smoketests:
-    needs: [lints]
+  smoketests:
     name: Smoketests
+    strategy:
+      fail-fast: false
```
bfops: undo this fail-fast override before merging
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -18,9 +18,107 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  docker_smoketests:
-    needs: [lints]
```
bfops, line 22: undo this removal before merging

crates/guard/src/lib.rs
```diff
@@ -25,160 +114,361 @@ impl SpacetimeDbGuard {
...
+        eprintln!("[RESTART-{:03}] Starting restart, old pid={}", spawn_id, old_pid);
+
+        self.stop();
+        eprintln!("[RESTART-{:03}] Old process stopped, sleeping 100ms", spawn_id);
+
+        // Brief pause to ensure system resources are fully released
+        sleep(Duration::from_millis(100));
+
+        eprintln!("[RESTART-{:03}] Spawning new server", spawn_id);
+        let (child, logs, host_url, reader_threads) = Self::spawn_server(&self.data_dir, self.pg_port, spawn_id);
```
bfops: note: this assumes we didn't run with `use_installed_cli`

crates/smoketests/src/lib.rs
```diff
@@ -0,0 +1,1299 @@
...
+            }
+            filtered_blocks.push(block);
+        }
+        blocks = filtered_blocks;
+    }
+
+    // Join blocks and replace module name
+    let result = blocks.join("\n").replace("quickstart-chat", module_name);
+    result + &end
+}
```
bfops, line 243: seems weird to put this here instead of in the quickstart test?

crates/smoketests/src/lib.rs
```diff
@@ -0,0 +1,1299 @@
...
+            .unwrap_or(&self.server_url)
+    }
+
+    /// Returns the PostgreSQL wire protocol port, if enabled.
+    ///
+    /// Returns None if running against a remote server or if PostgreSQL
+    /// wire protocol wasn't enabled for the local server.
+    pub fn pg_port(&self) -> Option<u16> {
+        self.guard.as_ref().and_then(|g| g.pg_port)
+    }
```
bfops, line 506: should we be able to run the pg tests against a remote server?

crates/smoketests/src/lib.rs
```diff
@@ -0,0 +1,1299 @@
...
+version = "0.1.0"
+edition = "2021"
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+spacetimedb = {{ path = "{}", features = ["unstable"] }}
+log = "0.4"
+"#,
```
bfops, line 646: duplicated above?

crates/smoketests/src/lib.rs
```diff
@@ -0,0 +1,1299 @@
...
+    () => {
+        if $crate::is_remote_server() {
+            #[allow(clippy::disallowed_macros)]
+            {
+                eprintln!("Skipping test: requires local server");
+            }
+            return;
+        }
+    };
+}
```
bfops, line 106: these should be failures, not skips; we should have a different method of explicitly skipping tests, not silently

tools/xtask-smoketest/src/main.rs
bfops: > The environment variable `__CARGO_FIX_YOLO` is an undocumented, internal-use-only feature for the Rust cargo fix command (and cargo clippy --fix) that forces the application of all available suggestions, including those that are marked as potentially incorrect or dangerous

crates/smoketests/src/modules.rs
bfops: removed

crates/smoketests/tests/smoketests/auto_migration.rs
bfops: I think it's fine; most tests don't do their own explanatory logs

crates/smoketests/tests/smoketests/servers.rs
bfops: I think this test has been manually updated to work better with remote servers. resolving.

crates/smoketests/tests/smoketests/servers.rs
bfops: it's fine.

crates/smoketests/src/lib.rs
bfops: they will currently fail and have to be explicitly skipped, which I think is good

tools/xtask-smoketest/src/main.rs
```diff
@@ -0,0 +1,193 @@
...
+        set_env(&mut cmd, server, dotnet);
+        cmd.args([
+            "nextest",
+            "run",
+            "--release",
+            "-p",
+            "spacetimedb-smoketests",
+            "--no-fail-fast",
+            "-j1",
+            "--test-threads=1",
```
bfops: undo this debug change (it also contained some other arg changes that must be restored)
```
git revert db2a345
```

tools/xtask-smoketest/src/main.rs
```diff
@@ -0,0 +1,196 @@
...
+        .status()?;
+
+    ensure!(status.success(), "Failed to build pre-compiled modules");
+    eprintln!("Pre-compiled modules built.\n");
+    Ok(())
+}
+
+/// Default parallelism for smoketests.
+/// 16 was found to be optimal - higher values cause OS scheduler overhead.
+const DEFAULT_PARALLELISM: &str = "1";
```
bfops: revert this debug change
```suggestion
const DEFAULT_PARALLELISM: &str = "16";
```

crates/smoketests/tests/smoketests/add_remove_index.rs
bfops: I added a TODO to investigate whether we want to include `name` at all. Most other tests just let the DB assign an identity and that's good enough for them.

For now, this seems fine.

crates/smoketests/src/lib.rs
```diff
@@ -191,74 +237,9 @@ pub fn have_psql() -> bool {
...
-    })
-}
-
-/// Parse code blocks from quickstart markdown documentation.
-/// Extracts code blocks with the specified language tag.
-///
-/// - `language`: "rust", "csharp", or "typescript"
-/// - `module_name`: The name to replace "quickstart-chat" with
-/// - `server`: If true, look for server code blocks (e.g. "rust server"), else client blocks
-pub fn parse_quickstart(doc_content: &str, language: &str, module_name: &str, server: bool) -> String {
```
bfops: this got moved into `quickstart.rs`

.github/workflows/ci.yml
```diff
@@ -78,6 +77,9 @@ jobs:
         with:
           run_install: true
 
+      - name: Check for pnpm
+        run: pnpm --version
+
```
cloutiertyler: Do we want to leave this in?

.github/workflows/ci.yml
```diff
@@ -19,20 +19,16 @@ concurrency:
 
 jobs:
   smoketests:
-    needs: [lints]
```
cloutiertyler, line 22: Eh, but fine

.github/workflows/ci.yml
```diff
@@ -114,25 +116,24 @@ jobs:
         uses: taiki-e/install-action@nextest
 
       - name: Run smoketests
-        run: cargo ci smoketests
+        # --test-threads=1 eliminates contention in the C# tests where they fight over bindings
+        # build artifacts.
+        # It also seemed to improve performance a fair amount (11m -> 6m)
+        run: cargo ci smoketests -- --test-threads=1
```
cloutiertyler, line 127: Crazy, but fine

.github/workflows/ci.yml
bfops: oh sorry at one point I had a comment about putting this line back in before merging.. I'm not sure where it went.

.github/workflows/ci.yml
bfops: nope, thank you

crates/smoketests/tests/smoketests/cli/publish.rs
```diff
@@ -1,15 +1,10 @@
...
-
-fn cli_cmd() -> Command {
-    Command::new(ensure_binaries_built())
-}
+use spacetimedb_smoketests::{require_local_server, Smoketest};
 
 #[test]
 fn cli_can_publish_spacetimedb_on_disk() {
-    let spacetime = SpacetimeDbGuard::spawn_in_temp_data_dir();
+    let test = Smoketest::builder().autopublish(false).build();
```
cloutiertyler, line 7: Nice

crates/smoketests/tests/smoketests/add_remove_index.rs
cloutiertyler: Fair point.

jdetter: ack on all of this 👍

PR #4106, opened 2026-01-23
Make all `VmMetrics` for the reducers and views of a module in `InstanceCommon::new`

Centril:
# Description of Changes

With this PR, all `VmMetrics` for all the reducers and views of a module are made in `InstanceCommon::new`
so that they are never made in `call_{reducer/view}_with_tx`.

Here's a flamegraph before this PR, zooming in on the smaller `call_reducer_with_tx`

image

And after, for the same `call_reducer_with_tx`:

image

Here are the performance numbers:
```
wasm, index=hash, branch=master, commit = af4d3f39e485435426b7f9182885175e57aafe44
ran for 10.097515999 seconds
completed 1310720
throughput was 129806.18204812016 TPS

wasm, index=hash, branch=master, commit = bac3d2a5a928af896d315fcfdf709d42e3577b66
ran for 10.842949063 seconds
completed 1474560
throughput was 135992.52301495385 TPS
```

As you can see, this is a gain of about 6k TPS on the phoenix_nap machine.

The second commit also adds the `d3-flamegraph-base.html` and stuff to `.gitignore` to facilitate taking flamegraphs.

None

1, trivial changes that are not load bearing.

### DIFF
--- a/crates/core/src/host/wasm_common/module_host_actor.rs
+++ b/crates/core/src/host/wasm_common/module_host_actor.rs
@@ -1121,7 +1121,9 @@ impl InstanceCommon {
         });
 
         // Report execution metrics on each view call.
-        self.vm_metrics.get_for_view_id(view_id).report(&result.stats);
+        self.vm_metrics
+            .get_for_view_id(view_id, &self.info.database_identity, &view_name)
+            .report(&result.stats);
 
         let trapped = matches!(result.call_result, Err(ExecutionError::Trap(_)));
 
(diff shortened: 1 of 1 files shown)


d3-flamegraph-base.html
```diff
@@ -0,0 +1,30 @@
+<!DOCTYPE html>
```
kim, line 1: Is this supposed to be here?
afaik, recent `perf` has it included.

Centril: No idea; I just copied the file from the notion doc.

PR #4107, opened 2026-01-23
feat: Quickstart and client for TanStack Start

clockwork-tien:
# Description of Changes
- Quickstart template `tanstack-ts` and client for TanStack Start
- Add Vue Quickstart Docs

# Screenshots
- `tanstack-ts` template
image

- TanStack Start Quickstart Docs
image

# Testing

- [x] Tested the templates locally (e.g. able to add people), works well for me


crates/bindings-typescript/src/tanstack/hooks.ts
```diff
@@ -0,0 +1,73 @@
...
+];
+
+export type UseSpacetimeDBSuspenseQueryResult<T> = [
+  T[],
+  false,
+  UseSuspenseQueryResult<T[], Error>,
+];
+
+// returns [data, loading, query] tuple
+export function useSpacetimeDBQuery<TableDef extends UntypedTableDef>(
```
cloutiertyler, line 27: Why is this called `useSpacetimeDBQuery` instead of `useTable` like it is in React?

Also I think a better comment would help here to explain what the options are.

crates/bindings-typescript/src/tanstack/hooks.ts
```diff
@@ -0,0 +1,73 @@
...
+  const query = useQuery({
+    ...queryOptions,
+    ...options,
+  } as UseQueryOptions<RowType<TableDef>[], Error>);
+
+  return [query.data ?? [], query.isPending, query];
+}
+
+// returns [data, false, query] tuple
+export function useSpacetimeDBSuspenseQuery<TableDef extends UntypedTableDef>(
```
cloutiertyler, line 59: What's the difference between this and the above. Would be nice to have a comment here.

crates/bindings-typescript/src/tanstack/hooks.ts
clockwork-tien: Thank you, I have added better comments. This is called `useSpacetimeDBQuery` instead of `useTable` because it wraps TanStack Query `useQuery` instead so it follows TanStack's naming and conventions (with TanStack Query API specific args and returns). Additionally we can also still use the normal React hooks e.g. `useTable` etc. in TanStack Start.

crates/bindings-typescript/src/tanstack/hooks.ts
clockwork-tien: Yes this is the is the Suspense version of the `useSpacetimeDBQuery`, and wraps TanStack Query's `useSuspenseQuery` instead of `useQuery`. I have also added a better comment.

cloutiertyler (approved):

cloutiertyler:
I'm approving this, but can you please confirm that you've tested this and add it to the testing section in the PR description?

clockwork-tien:
> I'm approving this, but can you please confirm that you've tested this and add it to the testing section in the PR description?

Yes I have tested this, works well from my testing 👍

PR #4108, opened 2026-01-23
CI: pack C# NuGet overrides in Release

rekhoff:
# Description of Changes
CI now packs the local `SpacetimeDB.BSATN.Runtime` and `SpacetimeDB.Runtime` NuGet packages using the `Release` configuration in both the `unity-testsuite` and `csharp-testsuite` jobs.

This makes the NuGet override deterministic because `sdks/csharp/tools~/write-nuget-config.sh` points to `crates/bindings-csharp/*/bin/Release`, and `dotnet pack` otherwise defaults to Debug.

This is needed due to a CI failure occurring in #3982 in which the one of our smoketests fails because the resolved version of the package lacks `RefOption<...>.GetListSerializer`, resulting in a `CS0117` error. Because this failure is not actually related of the changes in 3982, it is being pulled out into a separate PR.

None.

1 - Trivial

# Testing
- [ ] Confirm CI pass

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -559,8 +559,8 @@ jobs:
 
       - name: Override NuGet packages
         run: |
-          dotnet pack crates/bindings-csharp/BSATN.Runtime
-          dotnet pack crates/bindings-csharp/Runtime
+          dotnet pack -c Release crates/bindings-csharp/BSATN.Runtime
+          dotnet pack -c Release crates/bindings-csharp/Runtime
 
           # Write out the nuget config file to `nuget.config`. This causes the spacetimedb-csharp-sdk repository
           # to be aware of the local versions of the `bindings-csharp` packages in SpacetimeDB, and use them if
@@ -681,8 +681,8 @@ jobs:
 
       - name: Override NuGet packages
         run: |
-          dotnet pack crates/bindings-csharp/BSATN.Runtime
-          dotnet pack crates/bindings-csharp/Runtime
+          dotnet pack -c Release crates/bindings-csharp/BSATN.Runtime
+          dotnet pack -c Release crates/bindings-csharp/Runtime
 
           # Write out the nuget config file to `nuget.config`. This causes the spacetimedb-csharp-sdk repository
           # to be aware of the local versions of the `bindings-csharp` packages in SpacetimeDB, and use them if


bfops:
> This makes the NuGet override deterministic because `sdks/csharp/tools~/write-nuget-config.sh` points to `crates/bindings-csharp/*/bin/Release`, and `dotnet pack` otherwise defaults to Debug.

I don't think that this is correct? When I run `dotnet pack` in a fresh dir, I only have a `Release` directory. 

Apparently this wasn't always true:
> If you're developing with the .NET 8 SDK or a later version, the command uses the Release configuration by default for projects whose TargetFramework is set to net8.0 or a later version. The default build configuration is Debug for earlier versions of the SDK and for earlier target frameworks. 

(https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-pack#options)

Tests are run in fresh directories, so there shouldn't be any stale directories around. If this were building the wrong dir, I would expect tests to _always_ get it blatantly wrong?

bfops:
(I don't really have an objection to merging this, I'm just not sure that it will actually fix the issue you're experiencing?)

rekhoff:
> > This makes the NuGet override deterministic because `sdks/csharp/tools~/write-nuget-config.sh` points to `crates/bindings-csharp/*/bin/Release`, and `dotnet pack` otherwise defaults to Debug.
> 
> I don't think that this is correct? When I run `dotnet pack` in a fresh dir, I only have a `Release` directory.
> 
> Apparently this wasn't always true:
> 
> > If you're developing with the .NET 8 SDK or a later version, the command uses the Release configuration by default for projects whose TargetFramework is set to net8.0 or a later version. The default build configuration is Debug for earlier versions of the SDK and for earlier target frameworks.
> 
> (https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-pack#options)
> 
> Tests are run in fresh directories, so there shouldn't be any stale directories around. If this were building the wrong dir, I would expect tests to _always_ get it blatantly wrong?

Fair enough, you're right. Well for what it's worth, I re-ran the job on 3982 and it passed this time. ¯\\\_(ツ)\_/¯ Maybe it's not an issue, and this is unnecessary. This was an attempt to prevent flakiness by being explicit.

PR #4111, opened 2026-01-23
Hide LLM benchmark files in git diffs

bfops:
# Description of Changes

Marked the llm benchmark files as autogenerated so they are marked as "autogenerated" in github diffs.

None

1

# Testing

- [x] They are marked correctly as autogenerated if I push a change to those files


.gitattributes
```diff
@@ -1 +1,3 @@
 **/module_bindings/** linguist-generated=true eol=lf
+/docs/llms/** linguist-generated=true
+/docs/llms/*-details.json linguist-generated=false
```
bfops, line 3: @cloutiertyler why did we want to not ignore these? I don't think people are really reviewing the changes to these large json files?

cloutiertyler: This is fine.

PR #4113, opened 2026-01-23
Quickstart remix

bradleyshep:
# Description of Changes

Adds a Remix quickstart template and documentation for SpacetimeDB.

## Template (`templates/remix-ts/`)
- SpacetimeDB module with TypeScript (`spacetimedb/src/index.ts`)
- Remix app with Vite configuration
- SpacetimeDB provider in `app/providers.tsx` with proper SSR/hydration handling
- Example route (`app/routes/_index.tsx`) demonstrating `useTable` and `useReducer` hooks
- Pre-generated module bindings in `src/module_bindings/`

## Documentation (`docs/docs/00100-intro/00200-quickstarts/00175-remix.md`)
- Step-by-step quickstart guide following existing pattern
- Covers project creation with `spacetime dev --template remix-ts`
- Explains project structure, provider pattern, and React hooks usage
- Documents environment variable configuration (`VITE_SPACETIMEDB_HOST`, `VITE_SPACETIMEDB_DB_NAME`)

## Consistency
- Environment variables follow the same `HOST`/`DB_NAME` naming convention as other quickstarts
- Documentation structure mirrors Next.js and React quickstarts
- Dev server runs on port 3001 to avoid conflict with SpacetimeDB on port 3000

None.

1 - Additive change only. New template and documentation, no modifications to existing functionality.

# Testing

- [ ] Run `spacetime dev --template remix-ts my-remix-app` and verify app starts
- [ ] Verify SpacetimeDB connection works at http://localhost:3001
- [ ] Test adding a person via the UI form
- [ ] Verify SSR doesn't crash (SpacetimeDB provider defers to client-side)
- [ ] Check documentation renders correctly in Docusaurus


templates/remix-ts/app/providers.tsx
```diff
@@ -0,0 +1,55 @@
+import { useMemo, useState, useEffect } from 'react';
```
cloutiertyler: NIT: I would just stuff the whole contents of this file in `root.tsx` for simplicity and consistency since our other quickstarts do that.

docs/docs/00100-intro/00200-quickstarts/00175-remix.md
```diff
@@ -0,0 +1,209 @@
...
+spacetime logs my-remix-app
+2025-01-13T12:00:00.000000Z  INFO: Hello, Alice!
+2025-01-13T12:00:00.000000Z  INFO: Hello, World!
+```
+    </StepCode>
+  </Step>
+
+  <Step title="Understand the provider pattern">
+    <StepText>
+      SpacetimeDB is client-side only — it cannot run during server-side rendering. The `app/providers.tsx` file handles this by deferring the SpacetimeDB connection until the component mounts on the client.
```
cloutiertyler: I don't believe this is correct (and if it is correct, I believe that's a bug). I think we should show that you can access SpacetimeDB state in the Loader if we can. We should at least try it out.

You should be able to access SpacetimeDB state both from the server and the client, although they will obviously each have to maintain their own connections.

cloutiertyler (changes requested):
I would like to explore accessing SpacetimeDB state in loaders before approving this. We at the very least need to change the comment, because I don't think it's true (indeed spacetimedb.com uses the TS SDK on the backend).

bradleyshep:
> I would like to explore accessing SpacetimeDB state in loaders before approving this. We at the very least need to change the comment, because I don't think it's true (indeed spacetimedb.com uses the TS SDK on the backend).

Updates:
- Changed default port from 3001 to 5173 (Vite default) - removed from both `package.json` and `vite.config.ts`
- Changed default host from `ws://localhost:3000` to `wss://maincloud.spacetimedb.com`
- Merged `providers.tsx` into `root.tsx` for simplicity
- Added `app/lib/spacetimedb.server.ts` for server-side data fetching
- Added Remix `loader` to `_index.tsx` for SSR data fetching
- Updated quickstart docs to explain SSR pattern and fix incorrect "client-side only" claim

templates/remix-ts/app/lib/spacetimedb.server.ts
```diff
@@ -0,0 +1,49 @@
...
+export type PersonData = Infer<typeof Person>;
+
+/**
+ * Fetches the initial list of people from SpacetimeDB.
+ * This function is designed for use in Remix loaders.
+ *
+ * It establishes a WebSocket connection, subscribes to the person table,
+ * waits for the initial data, and then disconnects.
+ */
+export async function fetchPeople(): Promise<PersonData[]> {
```
cloutiertyler, line 16: I'm not sure we necessarily want to encourage people to connect to SpacetimeDB once per web request, but for a tutorial I suppose this is fine.

I think an alternative like this:

```ts
  import { DbConnection, Person } from '../../src/module_bindings';                                                                                                                                                  
  import type { Infer } from 'spacetimedb';         
                                                                                                                                                                                                                     
  const HOST = process.env.SPACETIMEDB_HOST ?? 'wss://maincloud.spacetimedb.com';                                                                                                                                    
  const DB_NAME = process.env.SPACETIMEDB_DB_NAME ?? 'remix-ts';

  export type PersonData = Infer<typeof Person>;

  let conn: DbConnection | null = null;
  let ready: Promise<void> | null = null;

  function ensureConnection(): Promise<void> {
    if (ready) return ready;

    ready = new Promise((resolve, reject) => {
      conn = DbConnection.builder()
        .withUri(HOST)
        .withModuleName(DB_NAME)
        .onConnect((c) => {
          c.subscriptionBuilder()
            .onApplied(() => resolve())
            .onError((_ctx, error) => reject(error))
            .subscribe('SELECT * FROM person');
        })
        .onConnectError((_ctx, error) => {
          conn = null;
          ready = null;
          reject(error);
        })
        .onDisconnect(() => {
          conn = null;
          ready = null;
        })
        .build();
    });

    return ready;
  }

  export async function fetchPeople(): Promise<PersonData[]> {
    await ensureConnection();
    return Array.from(conn!.db.person.iter());
  }
  ```

Would eventually be better once we have reconnect logic.

cloutiertyler (approved):

PR #4121, opened 2026-01-26
CI: Fix of C# tests failures

jdetter:
# Description of Changes

(jdetter): Github marked Ryan's PR as merged because I rebased it without updating the base branch first - my bad.
Original PR: https://github.com/clockworklabs/SpacetimeDB/pull/4120

This PR fixes C#-related CI failures when testing against `v1.12.0` before the corresponding NuGet packages are available on nuget.org.

* CI: Ensure local C# NuGet packages are used consistently
  * Pack `crates/bindings-csharp/{BSATN.Runtime,Runtime}` with `-c Release` so the configured local package sources (`bin/Release`) actually contain the `.nupkg`s.
  * Run `./sdks/csharp/tools~/write-nuget-config.sh` in CI to generate a `NuGet.Config` that maps `SpacetimeDB.BSATN.Runtime` / `SpacetimeDB.Runtime` to those local sources (with `nuget.org` as fallback for non-SpacetimeDB dependencies).
  * Add an explicit `dotnet restore --configfile NuGet.Config` step before tests, and `run dotnet test --no-restore`, so restore always uses the intended config and does not “accidentally” restore from `nuget.org`.
* `write-nuget-config.sh`: Make NuGet config discovery + mapping deterministic
  * Write `NuGet.Config` (capitalized) and include `<clear />` + an explicit `nuget.org` source to avoid inherited sources and to keep `PackageSourceMapping` functional.
  * Also write a repo-root `NuGet.Config` so `dotnet publish/pack` invoked from templates (outside `sdks/csharp/`) still discovers the same override behavior.
* Smoketests quickstart: avoid `PackageSourceMapping` restore failures
  * Ensure smoketest helper packing uses `Release`, and when repo-root `NuGet.Config` exists, perform `dotnet restore --configfile ...` + `dotnet pack --no-restore` so pack/restore are evaluated with the same sources/mapping.
  * When editing configs, prefer `NuGet.Config` casing for Linux discovery and ensure `nuget.org` exists as a source for the fallback mapping.

No changes

1

# Testing

- [X] Locally tested concept after simulating local repro.
- [x] Confirmed CI passes in this branch


.github/workflows/ci.yml
```diff
@@ -70,6 +70,14 @@ jobs:
         with:
           global-json-file: global.json
 
+      - name: Override NuGet packages
+        shell: bash
+        run: |
+          dotnet pack -c Release crates/bindings-csharp/BSATN.Runtime
+          dotnet pack -c Release crates/bindings-csharp/Runtime
+          cd sdks/csharp
+          ./tools~/write-nuget-config.sh ../..
```
bfops, line 79: I'm surprised that this makes a difference, since I would expect deterministic failures if we were just using the wrong versions of the packages. No harm in adding it though :shrug:

.github/workflows/ci.yml
```diff
@@ -571,6 +579,10 @@ jobs:
           cd sdks/csharp
           ./tools~/write-nuget-config.sh ../..
 
+      - name: Restore .NET solution
+        working-directory: sdks/csharp
+        run: dotnet restore --configfile NuGet.Config SpacetimeDB.ClientSDK.sln
```
bfops, line 584: I think this is redundant? we run `cargo ci dlls` below, which nukes the packages directory and re-restores

.github/workflows/ci.yml
```diff
@@ -693,9 +705,13 @@ jobs:
...
```
bfops, line 710: again I'm surprised that this makes a difference, since we should be running in a fresh dir and so I would expect deterministic failures if we were just failing to restore (and I don't see how we could be non-determinstically restoring the wrong thing).

but again, seems fine to include

smoketests/tests/quickstart.py
```diff
@@ -66,8 +78,13 @@ def add_source(doc: dict, *, key: str, path: str) -> None:
     cfg = doc.setdefault("configuration", {})
     sources = cfg.setdefault("packageSources", {})
     source_entries = sources.setdefault("add", [])
-    source = {"@key": key, "@value": path}
-    source_entries.append(source)
+
+    for source in source_entries:
+        if source.get("@key") == key:
+            source["@value"] = str(path)
+            return
```
bfops: was this needed? at least at one point we were only ever adding completely fresh entries. My concern would just be that if we're overriding something that's already there, it may have been there for a reason?

.github/workflows/ci.yml
bfops: oh nevermind I thought this PR was fixing flaky tests, but it's addressing things that were failing when versions were bumped.

bfops: actually which tests does this affect? because the C# quickstart tests already do this (for this exact reason)

.github/workflows/ci.yml
bfops: (again I thought this PR was to fix flakes, not issues with restoring the wrong thing, although I'm still somewhat surprised that this works differently than the restore introduced by `dotnet test`..)

smoketests/tests/quickstart.py
```diff
@@ -90,12 +107,25 @@ def add_mapping(doc: dict, *, key: str, pattern: str) -> None:
...
+    repo_nuget_config = STDB_DIR / "NuGet.Config"
+    if repo_nuget_config.exists():
+        run_cmd(
+            "dotnet",
+            "restore",
+            "--configfile",
+            str(repo_nuget_config),
+            cwd=source_dir,
+            capture_stderr=True,
+        )
```
bfops, line 114: I'm surprised that this branch is needed. `dotnet pack` should search upward for `nuget.config` and respect them by default.

.github/workflows/ci.yml
bfops: I'm just going to mark this resolved. if it works, it works.

bfops (approved):
LGTM, it seems to work in a test PR for me! (https://github.com/clockworklabs/SpacetimeDB/pull/4124)

smoketests/tests/quickstart.py
rekhoff: Sorry, this was an artifact from an earlier test, before I split out the `NuGet.Config` files. There was an issue where the <add> nodes from pre-existing runs where bumping into new ones. Anyway, looking closer, I don't believe it's needed so I've pushed a change to remove it.

PR #4123, opened 2026-01-26
Added query-with-index C# regression tests

rekhoff:
# Description of Changes

Adds C# regression tests for Query with Indexes.

No. Regression tests only, does not touch backend code.

1

# Testing

- [X] Built and ran regression tests locally, confirming that new regression test asserts pass.


jdetter (approved):
LGTM - seems like tests are passing. Thanks Ryan 👍

PR #4127, opened 2026-01-26
Reorganize types generated for typescript clients

jsdt:
# Description of Changes

This changes generated types in ts client bindings. We currently generate a few different types of types: reducer args, procedure args, rows, and user defined types. To avoid potential conflicts between these types (for example, if a user defined a type called `FooRow`, and also had a tabled named `foo`, we would end up with two types named `FooRow`), this puts each set of types in a different file and namespace. We also stopped exporting the `xxxRow` types, because there is always another type generated for those. We now have a `types` directory, which has an `index.ts` with user defined types, along with `reducers.ts` and `procedures.ts` for the types generated for reducer/procedure parameters. 

```
import type * as Types from './module_bindings/types';

var currentMessages: Types.Message[] = [];
```

or

```
import { type Message } from './module_bindings/types';

var currentMessages: Message[] = [];
```

This has a couple other changes:
 - For procedure and reducer types, this adds a suffix of `Args`, since we may want types for the return values in the future.
 - For all of the types, instead of exposing the schema object, we are now giving the typescript type (e.g. `export type Message = __Infer<typeof MessageRow>;`). I couldn't think of a reason for users to want the schema object, so this should save users from needing to do all of the `Infer` boilerplate.

This is a breaking change for v2.

2. This only changes typescript, and it should generally make thing easier to use.

# Testing

- [ ] 
- [ ]


cloutiertyler:
> For all of the types, instead of exposing the schema object, we are now giving the typescript type (e.g. export type Message = __Infer<typeof MessageRow>;). I couldn't think of a reason for users to want the schema object, so this should save users from needing to do all of the Infer boilerplate.

The reason is that there are helper functions on the schema object. For example:

https://github.com/clockworklabs/SpacetimeDB/blob/3d3c99f8dbf6137457a31311040aa8badca568dd/crates/bindings-typescript/src/lib/type_builders.ts#L1607

```ts
// So you can do stuff like:
MyEnum.create('foo');
MyEnum.create('bar', 32);

// instead of
{ tag: 'foo' }
{ tag: 'bar', value: 32 }
```

As one example. Or get the `AlgebraicType` as another example:

```ts
MyStruct.algebraicType
```
Which I know at least we need because we use it internally. Actually I'm curious to know how this compiles given that we use our own code gen internally.

cloutiertyler:
```ts
import { Rows } from './module_bindings';
var currentMessages: Rows.Message[] = [];
```

Won't this also conflict with a table called `rows`?

cloutiertyler:
> For procedure and reducer types, this adds a suffix of Args, since we may want types for the return values in the future.

This is a much needed change 🙏.

PR #4128, opened 2026-01-26
Rework `JobCores` in order to core-pin v8 instance threads

coolreader18:
# Description of Changes

This got a little bigger than I had hoped, but I think it's still pretty manageable. This PR partially reverts back to before #3263: cores are re-balanced with a `watch::Receiver<CoreId>` that each database thread will listen for updates on in order to repin itself, and multiple OS threads (each matched to a database) can be pinned to one core. As I understand it, that second part is something Phoebe was trying to avoid, but given that there's no way to asyncify a JS module, it's kind of necessary.

JS is single-threaded, and uses cooperative rather than preemptive multitasking (callbacks/async, not green threads). That means that if a JS function has an infinite loop, no other event handlers would be able to run unless that loop were to exit. Coupled with the fact that we can't `Send` a v8 isolate across threads, it makes more sense to keep the module host on one thread and repin that thread as needed. An alternative option, as was brought up, would be to deconstruct and reconstruct the module onto a different thread when needed, since load-balancing won't be happening often anyway. 

3 - reworks the threadpool that databases run on, and so could lead to deadlocks or other concurrency bugs. However, that seems unlikely, since this separates databases each onto their own thread, and as such decreases the likelihood of them interacting poorly with each other.

# Testing

Not sure if there's anything specific I should do, since this doesn't change behavior.

- [ ]


coolreader18:
Can confirm that this improves performance; just this patchset brings v8 from 78.5k TPS to 89k TPS.

crates/core/src/host/v8/mod.rs
```diff
@@ -294,16 +310,45 @@ impl JsInstance {
 
         match extract(reply) {
             Err(err) => unreachable!("should have received {} but got {err:?}", type_name::<T>()),
-            Ok(reply) => (reply, self),
+            Ok(reply) => reply,
+        }
+    }
+
+    pub async fn run_on_thread<F, R>(&mut self, f: F) -> R
```
Centril, line 326: Some documentation would be good here.

crates/core/src/host/v8/mod.rs
```diff
@@ -429,6 +460,10 @@ enum JsWorkerReply {
 // We care about optimizing for `CallReducer` as it happens frequently,
 // so we don't want to box anything in it.
 enum JsWorkerRequest {
+    /// See [`JsInstance::run_on_thread`]
```
```suggestion
    /// See [`JsInstance::run_on_thread`].
```

crates/core/src/host/v8/mod.rs
```diff
@@ -502,11 +534,13 @@ fn new_isolate() -> OwnedIsolate {
...
-/// Otherwise, when [`ModuleCreationContextLimited`] is passed,
+/// Otherwise, when [`ModuleCreationContext`] is passed,
 /// this is the first time both the module and instance are created.
-fn spawn_instance_worker(
+async fn spawn_instance_worker(
     program: Arc<str>,
-    module_or_mcc: Either<ModuleCommon, ModuleCreationContextLimited>,
+    module_or_mcc: Either<ModuleCommon, ModuleCreationContext>,
+    load_balance_guard: Arc<LoadBalanceOnDropGuard>,
+    mut core_pinner: CorePinner,
```
Centril, line 561: These parameters could be incorporated into the function docs.

crates/core/src/util/jobs.rs
```diff
@@ -235,15 +197,56 @@ impl PinnedCoresExecutorManager {
...
+    }
+    pub async fn run(self) {
+        if let Some(mut move_core_rx) = self.move_core_rx {
+            while move_core_rx.changed().await.is_ok() {
+                let core_id = *move_core_rx.borrow_and_update();
+                core_affinity::set_for_current(core_id);
+            }
+        }
+    }
+}
```
Centril, line 272: Docs on these please :)

crates/core/src/util/jobs.rs
```diff
@@ -144,50 +103,52 @@ impl JobCores {
...
-            inner: JobCoresInner::NoPinning(global_runtime),
+            inner: JobCoresInner::NoPinning,
         }
     }
 }
 
 impl PinnedCoresExecutorManager {
-    /// Get a [`runtime::Handle`] for running database operations on,
+    /// Get a core for running database operations on,
     /// and store state in `self` necessary to move that database to a new runtime
```
gefjon: This doc comment needs to be changed; databases no longer move between runtimes.

crates/core/src/host/module_host.rs
```diff
@@ -341,67 +339,78 @@ impl ReducersMap {
...
-        match self {
-            Module::Wasm(module) => module.replica_ctx(),
-            Module::Js(module) => module.replica_ctx(),
-        }
+struct V8ModuleHost {
+    instance_manager: ModuleInstanceManager<super::v8::JsModule>,
+}
+
+/// A module; used as a bound on `InstanceManager`.
+trait GenericModule {
```
gefjon, line 360: Why do we need a trait here? There's only two kinds of module, not an unbounded set of them.

crates/core/src/host/module_host.rs
```diff
@@ -734,9 +743,9 @@ impl CallProcedureParams {
 /// When we introduce procedures, it will be necessary to have multiple instances,
 /// as each procedure invocation will have its own sandboxed instance,
 /// and multiple procedures can run concurrently with up to one reducer.
-struct ModuleInstanceManager {
-    instances: VecDeque<Instance>,
-    module: Arc<Module>,
+struct ModuleInstanceManager<M: GenericModule> {
+    instances: Mutex<VecDeque<M::Instance>>,
+    module: M,
```
gefjon, line 739: Why are we rewriting this to use trait polymorphism rather than tagged enums?

coolreader18: It more accurately models the data it holds. We're never going to have a `ModuleInstanceManager` where the `module` is `Wasm` but the instances are `Js`, or where the elements of `instances` are heterogenous. Now we don't have to do `let Instance::Js(inst) = inst else { unreachable!() };`, and as a nice side benefit the `VecDeque` is more compact (again because of the type system–enforced homogeneity).

gefjon: I don't object to this specifically. My main concern here is that we not regress to the previous version of instance managers which used `dyn Trait` runtime polymorphism.

In general, I urge you to keep an eye on when using more complex types to more accurately model data provides a tangible benefit, versus when it adds needless complexity. In this case, I do not think that `let inst = inst.into_js().unwrap();` would be likely to lead to confusion or bugs, nor do I believe that reducing the `VecDeque`'s size is likely to tangibly affect our performance.

Centril: In this particular case, I agree with @coolreader18. Enforcing the common interface is useful and keeps things in sync.
Type classes are not just about polymorphism, but also about the restrictions they impose and the "laws" that come with them.

crates/core/src/util/jobs.rs
```diff
@@ -235,15 +197,56 @@ impl PinnedCoresExecutorManager {
...
+                let core_id = *move_core_rx.borrow_and_update();
+                core_affinity::set_for_current(core_id);
+            }
+        }
+    }
+    pub async fn run(self) {
+        if let Some(mut move_core_rx) = self.move_core_rx {
+            while move_core_rx.changed().await.is_ok() {
+                let core_id = *move_core_rx.borrow_and_update();
+                core_affinity::set_for_current(core_id);
```
Centril: Also, could you make a function that does this to not duplicate 3x? Make it `#[inline]` also please.

crates/core/src/util/jobs.rs
```diff
@@ -235,15 +197,56 @@ impl PinnedCoresExecutorManager {
...
+}
+
+impl CorePinner {
+    pub fn pin_now(&mut self) {
+        if let Some(move_core_rx) = &mut self.move_core_rx {
+            let core_id = *move_core_rx.borrow_and_update();
+            core_affinity::set_for_current(core_id);
+        }
+    }
+    pub fn pin_if_changed(&mut self) {
```
Centril, line 254: Using this turns out to be critical for performance. The difference is about 6k TPS.

PR #4135, opened 2026-01-27
2.0: [Rust] Remove RLS bindings [Not to merge]

Shubham8287:
# Description of Changes

Remove RLS bindings for 2.0. 

Older modules with RLS should continue to work.

# Testing:
This PR is suppose to remove the interface, hence deletes smoketest and integration tests.
though we may want to keep some unit tests as we still support RLS in older modules. I will check for them before merging the PR.

1


cloutiertyler:
We've decided to close this for now and leave RLS in for the time being. It will be an easier migration path for existing users and it's currently marked as an unstable feature anyway. We can deprecate this in the future once views fully subsume their behavior.

PR #4137, opened 2026-01-27
Define `TableName` and `ReducerName`  backed by `EcoString`

Centril:
# Description of Changes

The first commit defines a type `TableName` that is used in e.g., `TxData` and where determined profitable and necessary to do this change.
`TableName` is backed by [`ecow::EcoString`](https://docs.rs/ecow/0.2.6/ecow/string/struct.EcoString.html) which affords O(1) clones and 15 bytes of inline storage and `mem::size_of::<EcoString>() == 16`.

The second commit does the same for `ReducerName`. This is also used in reducer execution.

Together, these commits increase TPS by around 5-7k TPS.

None

1

# Testing

Covered by existing tests.

### DIFF
--- a/crates/codegen/src/rust.rs
+++ b/crates/codegen/src/rust.rs
@@ -11,6 +11,7 @@ use spacetimedb_lib::sats::layout::PrimitiveType;
 use spacetimedb_lib::sats::AlgebraicTypeRef;
 use spacetimedb_schema::def::{ModuleDef, ProcedureDef, ReducerDef, ScopedTypeName, TableDef, TypeDef};
 use spacetimedb_schema::identifier::Identifier;
+use spacetimedb_schema::reducer_name::ReducerName;
 use spacetimedb_schema::schema::TableSchema;
 use spacetimedb_schema::type_for_generate::{AlgebraicTypeDef, AlgebraicTypeUse};
 use std::collections::BTreeSet;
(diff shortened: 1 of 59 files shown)


crates/client-api-messages/src/websocket.rs
```diff
@@ -151,7 +152,7 @@ impl<Args> ClientMessage<Args> {
 #[sats(crate = spacetimedb_lib)]
 pub struct CallReducer<Args> {
     /// The name of the reducer to call.
-    pub reducer: Box<str>,
+    pub reducer: ReducerName,
```
gefjon: If we want these changes to be permanent, we'll also need to modify https://github.com/clockworklabs/SpacetimeDB/pull/4023 . Given that other team members are actively working on changing the WS format, I'd prefer not to do that right now if we can avoid it. Is WS encoding time included in the keynote benchmarks?

crates/codegen/src/rust.rs
```diff
@@ -343,8 +343,8 @@ pub(super) fn parse_table_update(
         let reducer_name = reducer.name.deref();
         let func_name = reducer_function_name(reducer);
         let set_reducer_flags_trait = reducer_flags_trait_name(reducer);
-        let args_type = function_args_type_name(&reducer.name);
-        let enum_variant_name = reducer_variant_name(&reducer.name);
+        let args_type = function_args_type_name(&reducer.name.into_identifier());
+        let enum_variant_name = reducer_variant_name(&reducer.name.into_identifier());
```
gefjon: Instead of calling `into_identifier`, which looks to me like it conses, you should rewrite `function_args_type_name`, `reducer_variant_name` and friends to take either an `&ReducerName` or an `&str`. I know codegen isn't performance-critical, but constructing `Identifier`s just to deref them into `&str`s is both needlessly, stupidly wasteful and also unclear and needlessly complex.

gefjon: This comment applies to all changes in this file and in the neighboring `typescript.rs`.

crates/commitlog/Cargo.toml
```diff
@@ -21,6 +21,7 @@ async-stream = { workspace = true, optional = true }
 bitflags.workspace = true
 bytes= { workspace = true, optional = true }
 crc32c.workspace = true
+ecow.workspace = true
```
gefjon: Why is this added? There don't appear to be any other changes to the `commitlog` crate.

crates/schema/src/reducer_name.rs
```diff
@@ -0,0 +1,56 @@
...
+
+/// The name of a reducer.
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Default)]
+pub struct ReducerName(
+    // TODO(perf, centril): Use this sort of optimization
+    // in RawIdentifier and `Identifier` and more places.
+    // TODO(perf): Consider `lean_string` instead for `&'static str` optimization.
+    // This could be useful in e.g., `SumType` and friends.
+    pub EcoString,
+);
```
gefjon, line 14: It seems to me that we should also be using `EcoString` names for views and procedures. View names might be best encoded as `TableName`, since views mostly behave like tables in many contexts. Rather than introducing a separate type `ProcedureName`, I would recommend renaming this type to `FunctionName` and using it for both reducers and procedures.

crates/client-api-messages/src/websocket.rs
Centril: Postponing this change.

PR #4140, opened 2026-01-27
Append commit instead of individual transactions to commitlog

kim:
Changes the commitlog (and durability) write API, such that the caller decides how many transactions are in a single commit, and has to supply the transaction offsets.

This simplifies commitlog-side buffering logic to essentially a `BufWriter` (which, of course, we must not forget to flush). This will help throughput, but offers less opportunity to retry failed writes. This is probably a good thing, as disks can fail in erratic ways, and we should rather crash and re-verify the commitlog (suffix) than continue writing.

To that end, this patch liberally raises panics when there is a chance that internal state could be "poisoned" by partial writes, which may be debatable.

# Motivation

The main motivation is to avoid maintaining the transaction offset in two places in such a way that they could diverge. As ordering commits is the responsibility of the datastore, we make it authoritative on this matter -- the commitlog will still check that offsets are contiguous, and refuse to commit if that's not the case.

A secondary, related motivation is the following:

A "commit" is an atomic unit of storage, meaning that a torn (partial) write of a commit will render the entire commit corrupt. There hasn't been a compelling case where we would want this, and have always configured the server to write exactly one transaction per commit.
The code to handle buffering of transactions is, however, rather complex, as it tries hard to allow the caller to retry writes at commit boundaries. An unfortunate consequence of this is that we'd flush to the OS very often, leaving throughput performance on the table.

So, if there is a compelling case for batching multiple transactions in a commit, it should be the datastore's responsibility.

Breaks internal APIs only.

5 - Mostly for the risk

# Testing

Existing tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -280,6 +280,8 @@ jobs:
           ref: ${{ env.GIT_REF }}
 
       - uses: dsherret/rust-toolchain-file@v1
+      - name: Set default rust toolchain
+        run: rustup default $(rustup show active-toolchain | cut -d' ' -f1)
       - name: Cache Rust dependencies
         uses: Swatinem/rust-cache@v2
         with:
@@ -332,6 +334,8 @@ jobs:
         uses: actions/checkout@v3
 
       - uses: dsherret/rust-toolchain-file@v1
+      - name: Set default rust toolchain
+        run: rustup default $(rustup show active-toolchain | cut -d' ' -f1)
       - run: echo ::add-matcher::.github/workflows/rust_matcher.json
 
       - name: Cache Rust dependencies
@@ -359,6 +363,8 @@ jobs:
       - uses: actions/checkout@v3
 
       - uses: dsherret/rust-toolchain-file@v1
+      - name: Set default rust toolchain
+        run: rustup default $(rustup show active-toolchain | cut -d' ' -f1)
       - run: echo ::add-matcher::.github/workflows/rust_matcher.json
 
       - name: Cache Rust dependencies
@@ -379,6 +385,9 @@ jobs:
     permissions: read-all
     steps:
       - uses: actions/checkout@v3
+      - uses: dsherret/rust-toolchain-file@v1
+      - name: Set default rust toolchain
+        run: rustup default $(rustup show active-toolchain | cut -d' ' -f1)
       - name: Set up Python env
         run: |
           test -d venv || python3 -m venv venv
@@ -415,6 +424,8 @@ jobs:
 
       - name: Install Rust
         uses: dsherret/rust-toolchain-file@v1
+      - name: Set default rust toolchain
+        run: rustup default $(rustup show active-toolchain | cut -d' ' -f1)
 
       - name: Install rust target
         run: rustup target add ${{ matrix.target }}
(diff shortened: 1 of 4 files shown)


kim:
Nominating @gefjon and @Centril because they appeared in the reviewer suggestions.
@Centril specifically for hints on how the get rid of `Box<[_]>` allocations.
@Shubham8287 because of previous work on the commitlog.
@joshua-spacetime because of suggesting the change initially.

crates/commitlog/src/tests/partial.rs
```diff
@@ -2,93 +2,101 @@ use std::{
...
+    log.commit([(end, [b'x'; 32])]).unwrap();
+    assert_matches!(
+        log.flush(),
+        Err(e) if e.kind() == io::ErrorKind::StorageFull
+    );
+}
+/// Tests that, when a partial write occurs, we can read all flushed commits
+/// up until the faulty one.
 #[test]
-fn reopen() {
```
kim, line 41: I'm not sure what this was supposed to test originally, so I removed it.

crates/commitlog/src/tests/partial.rs
```diff
@@ -2,93 +2,101 @@ use std::{
...
+fn read_log_up_to_partial_write() {
```
kim, line 52: This is basically the test previously named `reopen`.

crates/commitlog/src/commitlog.rs
```diff
@@ -303,8 +270,63 @@ impl<R: Repo, T> Generic<R, T> {
...
+    /// - [Self::sync] panics (called when rotating segments)
+    pub fn commit<U: Into<Transaction<T>>>(
+        &mut self,
+        transactions: impl IntoIterator<Item = U>,
+    ) -> io::Result<Option<Committed>> {
+        self.panicked = true;
+        let writer = &mut self.head;
+        let committed = writer.commit(transactions)?;
+        if writer.len() >= self.opts.max_segment_size {
+            self.flush().expect("failed to flush segment");
```
kim: This seemed a bit surprising to me at first -- but the `BufWriter` has no way of knowing how many bytes did make it. So if flush fails, the buffer is basically garbage.

Centril:
> @Centril specifically for hints on how the get rid of `Box<[_]>` allocations.

What is the typical length of the slice?

kim:
> > @Centril specifically for hints on how the get rid of `Box<[_]>` allocations.
> 
> What is the typical length of the slice?

Well, 1 :D
One way to solve it is to allow only a single transaction in the `Durability` trait until we actually need more than one. The `impl IntoIterator` in the `commitlog` crate is supposedly going to be optimized away then.

Centril:
> > > @Centril specifically for hints on how the get rid of `Box<[_]>` allocations.
> > 
> > 
> > What is the typical length of the slice?
> 
> Well, 1 :D One way to solve it is to allow only a single transaction in the `Durability` trait until we actually need more than one. The `impl IntoIterator` in the `commitlog` crate is supposedly going to be optimized away then.

Yeah, I think this is the right call until we need something else.

kim:
> I wonder, if we should test replication with this branch? To surface any bug, if exist for `n!=1` case.

We will benefit from `BufWriter` buffering without allowing `n > 1` in the `Durability` trait (just flush after `n` `commit()` calls). So I'm considering to revert the `Durability` change to

```rust
fn append_tx(&self, Transaction<Self::TxData>)
```

which requires the offset to be supplied, but doesn't allocate a `Box<[_]>` that we don't actually need currently.

kim:
Note that Rust does not by default run destructors when the program is terminated by a signal (_any_ signal). This, and the default being unconfirmed reads, are the reason the commitlog before this patch would flush after every write. 

I added a config option to preserve this behavior. Question is if we should make it the default for standalone. (We should probably also make use of the `ctrlc` crate for server processes)

Issue #4131, opened 2026-01-27
Create `spacetime login --interactive`


rafaelGuerreiro:
I have 2 accounts for spacetimedb, one personal, and another for work. I normally use 2 different browsers to separate work from personal projects.

When I do spacetime login, it simply opens the URL on my default browser and that makes it hard to switch between them.
I've created a .env in each project, and then I have a small shell script to login using the token in the .env.

Ideally, if I can do spacetime login --interactive, it can simply show me the url, which I can copy and paste in the right browser window and properly switch to the correct account.

bfops:
Hey @rafaelGuerreiro, thanks for filing an issue. We'll work on triaging this.

In the meantime, I believe the URL should open in the most recently-used browser window, so if you use a "work" window and then run the command, the URL should be opened there. I hope that helps for now.

bfops:
(This should go out in our next release, as `--no-browser`!)

rafaelGuerreiro:
That was quick, I was about to test the last browser version, but this helps a lot already!

PR #4142, opened 2026-01-27
Add `spacetime login --no-browser`

bfops:
# Description of Changes

Addresses https://github.com/clockworklabs/SpacetimeDB/issues/4131.

I did not thread the new option into other commands such as `spacetime init` (they just retain the previous behavior). I think this is a relatively niche use case, and it's easy to work around either way (by just using `spacetime login` first) so I'd wait until someone asks us to thread it through to other places.

None

1

# Testing

- [x] Ran `spacetime login --no-browser` and saw that it didn't open a browser
- [x] Ran `spacetime login` and saw that it still opened a browser


crates/cli/src/util.rs
```diff
@@ -318,10 +318,10 @@ pub async fn get_login_token_or_log_in(
 
     if full_login {
         let host = Url::parse(DEFAULT_AUTH_HOST)?;
-        spacetimedb_login_force(config, &host, false).await
+        spacetimedb_login_force(config, &host, false, true).await
```
jdetter, line 321: I guess if someone is hitting this case and they don't want it to open in their default browser they just have to manually:
```
spacetime login
```
so that they follow the other flow, right?

bfops: Correct:
> I did not thread the new option into other commands such as `spacetime init` (they just retain the previous behavior). I think this is a relatively niche use case, and it's easy to work around either way (by just using `spacetime login` first) so I'd wait until someone asks us to thread it through to other places.

PR #4150, opened 2026-01-28
Shrink `JsWorkerRequest` & use the right HashMap/Set

Centril:
# Description of Changes

This PR shrinks `JsWorkerRequest` so that it is (almost) as small as the call reducer request.
To do that, a bunch of trivial changes had to be done to auth code, that mostly revolves around `String` -> `Box<str>`.
This should help the auth code, but that is incidental.
The main goal was to improve throughput through the request tx/rx channel for V8, which is taking quite a bit of time in flamegraphs.

I also noticed while making this change that the wrong hash map was being used in a bunch of places, so I fixed all of those.

A follow up PR will shrink the reply side to fit within a cache line.
Yet another follow up PR will change the channel to replace flume with `fibre::spsc`.

None

2, fairly trivial changes.

# Testing

Covered by existing tests.


bfops (approved):
my code-owned changes (CLI) look good to me, although I don't know that this distinction really matters in there since it's very not-performance-sensitive.

I don't think I'm the right reviewer for most of the rest of the codebase.

Shubham8287 (approved):

PR #4158, opened 2026-01-29
[2.0 Breaking] `sender` changes for TypeScript modules

joshua-spacetime:
# Description of Changes

Change `sender` to be callable on [Reducer|View|Procedure|Tx]Ctx in TypeScript. Equivalent changes to https://github.com/clockworklabs/SpacetimeDB/pull/4101.

Breaks the module api

1

# Testing

Pure refactor, no additional testing.


coolreader18:
Why not just make it a getter? There's other data on `ctx` that's also a getter, e.g. `identity`

joshua-spacetime:
@coolreader18 I'd be fine with that. The only reason I didn't was to be more in line with the other languages. But happy to change it if you think a getter would be better.

coolreader18:
Yeah, I think we decided when initially designing the TS sdk that having getters makes more sense for values that won't change.

joshua-spacetime:
Closing this since in favor of getters which isn't an api break.

Issue #3750, opened 2025-11-24
[C#] Module bindings for query builder


joshua-spacetime:

rekhoff:
Create the prerequisite PR #4146 which implements the newer View ABI changes in the C# module bindings, which will be needed to support the Typed Query Builder implementation.

PR #4159, opened 2026-01-29
[C#] Module bindings for Typed Query Builder

rekhoff:
# Description of Changes
This is the implementation of the Typed Query Builder for C# modules outlined in #3750. 
This requires the changes from #4146 to be in place before it can properly function.

This aligns the C# typed query builder SQL formatter with the Rust implementation:
* Supports `And/Or` with same grouping styles as Rust.
* All comparison operators (`Eq/Neq/Lt/…`) wrap expressions in parentheses so chained predicates produce byte-for-byte identical SQL.

Not breaking. Existing modules keep producing the same SQL semantics; only redundant parentheses were added to match the Rust formatter.

2 — localized string-format updates plus parity harness tweaks. Low functional risk; largest concern is ensuring every comparison path gained parentheses.

# Testing
- [X] Tested against a local test harness validating output from these changes match current Rust behavior.


joshua-spacetime (commented):
Is it possible to share the core query builder types between the sdk and module bindings? I also think we should have end-to-end tests similar to what we have for the client builder.

crates/bindings-csharp/Codegen.Tests/fixtures/server/Lib.cs
```diff
@@ -280,6 +280,12 @@ public class Module
         return (PublicTable?)ctx.Db.PublicTable.Id.Find(0);
     }
 
+    [SpacetimeDB.View(Name = "public_table_query", Public = true)]
+    public static Query<PublicTable> PublicTableQuery(ViewContext ctx)
+    {
+        return ctx.From.PublicTable().Where(cols => cols.Id.Eq(SpacetimeDB.SqlLit.Int(0))).Build();
```
joshua-spacetime: Does this work too?
```suggestion
        return ctx.From.PublicTable().Where(cols => cols.Id.Eq(0)).Build();
```

crates/bindings-csharp/Codegen.Tests/fixtures/diag/snapshots/ExtraCompilationErrors.verified.txt
```diff
@@ -298,6 +298,144 @@ SpacetimeDB.Internal.Module.RegisterClientVisibilityFilter(global::Module.MY_THI
       ]
     }
   },
+  {/*
+    public global::SpacetimeDB.Table<global::InAnotherNamespace.TestDuplicateTableName, TestDuplicateTableNameCols, TestDuplicateTableNameIxCols> TestDuplicateTableName() =>
+        new("TestDuplicateTableName", new TestDuplicateTableNameCols("TestDuplicateTableName"), new TestDuplicateTableNameIxCols("TestDuplicateTableName"));
+                                          ^^^^^^^^^^^^^^^^^^^^^^^^^^
+}
+*/
+    Message: The call is ambiguous between the following methods or properties: 'TestDuplicateTableNameCols.TestDuplicateTableNameCols(string)' and 'TestDuplicateTableNameCols.TestDuplicateTableNameCols(string)',
```
joshua-spacetime, line 307: Which tests generate these errors?

rekhoff:
> Is it possible to share the core query builder types between the sdk and module bindings? I also think we should have end-to-end tests similar to what we have for the client builder.

Requested changes made, merging both the common Module Bindings and Client SDK sides of Module Builder into the `BSATN.Runtime` portion of the code base.

crates/bindings-csharp/Codegen.Tests/fixtures/server/Lib.cs
joshua-spacetime: Also are there tests that invoke this view?

crates/bindings-csharp/Codegen.Tests/fixtures/diag/snapshots/ExtraCompilationErrors.verified.txt
rekhoff: Running `dotnet test crates/bindings-csharp/Codegen.Tests` executes `TestDiagnostics`, which compiles the `diag` fixture and captures compiler errors for validating the error output.

crates/bindings-csharp/Codegen.Tests/fixtures/server/Lib.cs
rekhoff: There was not, but I've added a python test that checks it specifically.

crates/bindings-csharp/Codegen.Tests/fixtures/diag/snapshots/ExtraCompilationErrors.verified.txt
joshua-spacetime: But do these errors make sense? Are these from new tests you added?

rekhoff: Sorry, not exactly a test I've added, but is a result of the code changes.
`diag` already does a test of duplicate table names (see `TestDuplicateTableName`) in it's `Lib.cs`. With QueryBuilder's additional generated helpers are emitted for both of these tables with a duplicate name, the Roslyn generator sees the duplicate type/constructor symbols and generates the error.
The error's text comes strait from the C# compiler, not something we authored.

joshua-spacetime (commented):
This looks good. My only request would be to add an end-to-end test that subscribes to a query builder view and verify we get an initial result and that the client gets updates when the view is updated. Otherwise it's ready to merge.

sdks/csharp/examples~/regression-tests/server/Lib.cs
```diff
@@ -54,6 +54,36 @@ public partial struct WhereTest
...
+    {
+        public Col<WhereTest, uint> Id { get; }
+
+        public WhereTestCols(string tableName)
+        {
+            Id = new Col<WhereTest, uint>(tableName, nameof(WhereTest.Id));
+        }
+    }
+
+    private sealed class WhereTestIxCols
```
joshua-spacetime: Where are these types used?

sdks/csharp/examples~/regression-tests/client/Program.cs
```diff
@@ -349,6 +352,43 @@ void ValidateWhereSubscription(IRemoteDbContext conn)
     );
 }
 
+void ValidateWhereTestViews(IRemoteDbContext conn)
```
joshua-spacetime: Could you also update the underlying table after subscribing such that the client receives an update for the view?

sdks/csharp/examples~/regression-tests/server/Lib.cs
rekhoff: This is used mainly in the server module side, where `WhereTestIxCols` is used in the `MakeWhereTestTable` function, as the return in the 3rd value following the `Table<TRow, TCols, TIxCols>` formatting defined in `QueryuBuilder.cs`
Then the view `WhereTestQuery` uses it to generate a table that it can call `.Where` on, and performs a `Build()` like we can in the client, allowing the view to return a Query result.

In the client, `WhereTestIxCols` is used implicitly when the generated bindings create `qb.From.WhereTest()`. The test then validates the outcome of that builder query.

sdks/csharp/examples~/regression-tests/client/Program.cs
rekhoff: Added test in https://github.com/clockworklabs/SpacetimeDB/pull/4159/commits/d8f9d32dda4460fe9a16dae798974d0826af998e for this. I changed some of the name values on some of the tables to make it more clear what is being updated and tested for.

joshua-spacetime (approved):

PR #4161, opened 2026-01-29
feat: Quickstart and client for Plain JS Script Tags

clockwork-tien:
# Description of Changes

The PR implements the following updates:

- Create a quickstart guide and client support for using SpacetimeDB directly in browser with script tags
- A template to easily work with/test the implementation and get started with script tags

The bundle is served via `https://unpkg.com/spacetimedb@latest/dist/browser.bundle.js`, which only works after the package is published to npm, therefore to test locally we can use the `browser-ts` template and local bundle:
1. Build local bundle
```
cd crates/bindings-typescript
pnpm install
pnpm build
```
2. Use local bundle built in  template`index.html` (swap the URL from `unpkg`)
```diff
- <script src="https://unpkg.com/spacetimedb@latest/dist/browser.bundle.js"></script>
+ <script src="../../crates/bindings-typescript/dist/browser.bundle.js"></script>
```
3. Open `browser-ts/index.html` directly in browser along with `spacetime start` + publish the module `spacetime publish --project-path spacetimedb browser-ts`

# Screenshots

image

image

# Testing

- [ ] 
- [ ]


crates/bindings-typescript/src/browser/client.ts
```diff
@@ -0,0 +1,552 @@
...
+
+interface Subscription {
+  query: string;
+  callback: (rows: any[]) => void;
+  rows: Map<string, any>;
+}
+
+const TABLE_NAME_REGEX = /FROM\s+(\w+)/i;
+
+export class Client {
```
cloutiertyler: This is odd, it seems like we're rebuilding all of `DbConnectionimpl` here. Also I think we're losing the type information from the `./module_bindings`.

This should be updated to use `DbConnection` from the generated bindings.

cloutiertyler (changes requested):

clockwork-tien:
@cloutiertyler Thank you, I have applied the changes from https://github.com/clockworklabs/SpacetimeDB/pull/4167, merged/resolved conflicts and did a minor template description update. The PR is ready for review.

cloutiertyler:
Removed the SDK-level browser bundle (`src/browser/index.ts`, the IIFE and ESM browser configs in `tsup.config.ts`) since nothing in the template actually uses them. The template bundles its own bindings via Vite instead. The existing `dist/index.browser.mjs` (built from `src/index.ts`) already covers the browser ESM use case.

Long-term, the plan is to have the code generator emit `.js` + `.d.ts` files instead of `.ts`, which would allow the template to drop the Vite build step entirely and use [import maps](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type/importmap) to resolve `spacetimedb` imports natively in the browser. At that point, the generated bindings would just work with a `<script type="importmap">` pointing `spacetimedb` at `dist/index.browser.mjs` on unpkg.

PR #4166, opened 2026-01-29
Make `global.json` global again

bfops:
# Description of Changes

We have had several `global.json` files introduced that aren't symlinks to the root `global.json`.

This PR fixes that drift, and adds a CI check.

None.

2

# Testing

- [x] `cargo ci global-json-policy` succeeds
- [x] `cargo ci global-json-policy` fails if you add a new `global.json` file somewhere
- [x] Rest of CI passes, confirming that templates are still working properly


tools/ci/src/main.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(clippy::disallowed_macros)]
```
jdetter, line 1: Why do we need this?

tools/ci/src/main.rs
```diff
@@ -39,6 +41,61 @@ fn ensure_repo_root() -> Result<()> {
...
+            if ft.is_dir() {
+                out.extend(find_all(&path)?);
+            }
+        }
+        Ok(out)
+    }
+
+    let globals = find_all(Path::new("."))?
+        .into_iter()
+        .filter(|p| p.file_name() == Some(OsStr::new("global.json")))
```
jdetter: optional nit: IMO It would be better to just pass the `"global.json"` into `find_all(...)` so we can filter out just those files and not allocate a ton of strings.

jdetter (approved):
An optional nit, this looks great. Thank Zeke 🙏

tools/ci/src/main.rs
bfops: we use eprintln, and we have a global lint to not do that

tools/ci/src/main.rs
bfops: sure, done

PR #4167, opened 2026-01-30
fix: Use DbConnection instead of custom Client for browser template

cloutiertyler:
## Summary

This PR fixes the browser-ts template to use the standard `DbConnection` pattern instead of a custom `Client` class.

**Changes:**
- Remove the custom `Client` class that used JSON WebSocket protocol and HTTP for reducer calls
- Use `spacetime generate` and the generated `DbConnection` class with BSATN binary protocol
- Bundle the TypeScript bindings with vite into an IIFE that exposes `DbConnection` globally
- Update `index.html` to load the bundle and use inline JavaScript

**Why this matters:**
- Consistency with the rest of the TypeScript SDK
- Uses the standard BSATN binary protocol (more efficient than JSON)
- Uses WebSocket for reducer calls (not HTTP)
- Follows the established code generation pattern

## Test plan

- [ ] Run `spacetime dev --template browser-ts test-app`
- [ ] Run `npm install && npm run build` in the created project
- [ ] Open `index.html` in a browser and verify connection works
- [ ] Add a person and verify real-time updates work


cloutiertyler:
Closing in favor of https://github.com/clockworklabs/SpacetimeDB/pull/4161

PR #4170, opened 2026-01-30
templates/basic-rs properly uses workspace versions

bfops:
# Description of Changes

Replaced the specific versions in `templates/basic-rs` with relative paths into the repo.

This was split out of https://github.com/clockworklabs/SpacetimeDB/pull/4102.

None

1

# Testing

- [x] Upgrade version check still passes


jdetter (approved):
Looks good to me, the upgrade version fix also looks correct. Thanks Zeke! 👍

PR #4171, opened 2026-01-30
CI - more hacky v8 fixes

bfops:
# Description of Changes

This has been popping up in more places, so I've added this in more places.

None. CI only.

1

# Testing
:shrug:


jdetter (approved):
Alright let's give it a try

PR #4177, opened 2026-02-02
Identifiers: Refactor + Improve type-safety & performance

Centril:
# Description of Changes

The goal of this PR is to:
- Make it easy to switch our backing implementation of identifiers, as seen in the commit changing to `LeanString`.
- Improve type safety of our identifiers by having every identifier start with `RawIdentifier` and then `Identifier` is just a wrapper with validation on construction. `TableName` and `ReducerName` are then just wrappers around `Identifier`.
- Reduce allocation in `InstanceEnv` by using the now clone-on-write O(1) + SSO optimized `Identifier`.
- Reduce allocations in the query engine as a consequence of improving `RawIdentifier` and `Identifier`.
- Have `&'static str` strings be further optimized by avoiding to allocate even for long ones. This is supported by `impl From<&'static str> for RawIdentifier` as well as switching to `LeanString`.

The PR results in a roughly 2k TPS improvement over master for WASM + hash indices. This should also help V8, as this is VM-agnostic.

None

3? Mostly just small changes in many places, but in a sea of small changes, mistakes can happen.

# Testing

Covered by existing tests.

### DIFF
--- a/crates/schema/src/def.rs
+++ b/crates/schema/src/def.rs
@@ -65,6 +65,7 @@ pub type StrMap<T> = HashMap<RawIdentifier, T>;
 ///
 /// ```rust
 /// use spacetimedb_lib::RawModuleDef;
+/// use spacetimedb_sats::raw_identifier::RawIdentifier;
 /// use spacetimedb_schema::def::{ModuleDef, TableDef, IndexDef, TypeDef, ModuleDefLookup, ScopedTypeName};
 /// use spacetimedb_schema::identifier::Identifier;
 ///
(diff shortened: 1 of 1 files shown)


Centril:
Next step after this PR and once websocket 2.0 has been merged is to also use the identifier types in the WS format to further reduce allocation.

crates/schema/src/def/validate/v9.rs
```diff
@@ -713,12 +711,12 @@ impl CoreValidator<'_> {
     /// Validates that a type can be used to generate a client type use.
     pub(crate) fn validate_for_type_use(
         &mut self,
-        location: &TypeLocation,
+        mut location: impl FnMut() -> TypeLocation,
```
gefjon, line 714: Why is this change necessary? Also, why are we taking `FnMut`, rather than `FnOnce`?

bfops (approved):
Changes in my code-owned files LGTM

crates/schema/src/def/validate/v9.rs
Centril: This reduces the number of clones of `RawIdentifier` in the happy path. This has to be `FnMut` due to `.into_iter().map(...)`. It's potentially used more than once.

PR #4179, opened 2026-02-02
`RawModuleDefV10` Scheduled functions should not be callable

Shubham8287:
# Description of Changes
- Codegen to skip creating reducer and procedure file if `visibility == FunctionVisiblity::Internal`
-  `UpdateHost` to check for visibility before excuting reducers and procedures.

NA, master is not exposing new raw module version yet.

1. While the current patch is simple, could there be other places where visibility should also be applied?

# Testing
Smoketest can be in `2.0-breaking-changes` branch once we merge this.
Exising test should cover for regression

### DIFF
--- a/crates/bindings-csharp/Runtime/Internal/Autogen/FunctionVisibility.g.cs
+++ b/crates/bindings-csharp/Runtime/Internal/Autogen/FunctionVisibility.g.cs
@@ -10,7 +10,7 @@ namespace SpacetimeDB.Internal
     [SpacetimeDB.Type]
     public enum FunctionVisibility
     {
-        Internal,
+        Private,
         ClientCallable,
     }
 }


cloutiertyler:
Wait, doesn't the module owner still need the option/ability to call private reducers?

Shubham8287:
> Wait, doesn't the module owner still need the option/ability to call private reducers?

Well, if that’s the case, then I misunderstood the requirement. I thought they were only meant to be called by the Host.

gefjon:
Re: making internals callable by the owner, I would suggest that:

1. Until/unless we get a feature like "codegen profiles," internal reducers should not be included in codegen for the SDKs, and it is not useful to make them callable through the WebSocket API.
2. Allowing the owner / admins to call internal reducers via the CLI, and therefore via the HTTP API, is potentially valuable, but I would not consider it a requirement. It's very easy, with a small amount of additional module code, for a user who wants such a thing to make (the code contained in the body of) an internal reducer client-callable.

crates/codegen/src/util.rs
```diff
@@ -93,19 +93,26 @@ pub(super) fn is_reducer_invokable(reducer: &ReducerDef) -> bool {
...
 /// Iterate over all the [`ReducerDef`]s defined by the module, in alphabetical order by name.
 ///
-/// The init reducer is skipped because it should never be visible to the clients.
+/// Skipping the `init` reducer and internal [`FunctionVisibiity::Internal`] reducers because they
+/// should not be directly invokable.
 /// Sorting is not necessary for reducers because they are already stored in an IndexMap.
 pub(super) fn iter_reducers(module: &ModuleDef) -> impl Iterator<Item = &ReducerDef> {
     module
         .reducers()
         .filter(|reducer| reducer.lifecycle != Some(Lifecycle::Init))
```
gefjon, line 105: Is it still necessary to explicitly skip `Init`? Will `Init` now always be internal? Oh, or is this for compatibility with v9 modules, where `Init` will not be marked internal but should still not be generated? If the latter, please add a comment to that effect.

gefjon: Also, skipping internal reducers is not correct in every context: pre-2.0 SDKs will still need codegen for parsing and registering callbacks on reducer events, even for internal reducers. This will stop being an issue once all of our SDKs are updated for 2.0, but it's not clear to me that's going to happen before the initial 2.0 release.

cloutiertyler:
> Well, if that’s the case, then I misunderstood the requirement. I thought they were only meant to be called by the Host.

This would break BitCraft's workflow as they sometimes need to be able to call scheduled reducers manually. I believe that wholly disallowing is a regression.

Note for example that the roadmap item says "Make scheduled reducers internal **by default**" (emphasis mine). It is not that scheduled reducers are *only* callable by the module.

I think this ticket was also just underspecified in general. I would add the following requirements:

1. We call it `private` instead of internal.
2. `private` has the same semantic meaning as a `private` table. (accessible only by the module, the owner, or collaborators)
3. Scheduled reducers are `private` by default
4. You can declare them public with:

```rust
#[reducer(public)]
fn foobar() {
   // ...
}
```

gefjon:
> > Well, if that’s the case, then I misunderstood the requirement. I thought they were only meant to be called by the Host.
> 
> This would break BitCraft's workflow as they sometimes need to be able to call scheduled reducers manually. I believe that wholly disallowing is a regression.
> 
> Note for example that the roadmap item says "Make scheduled reducers internal **by default**" (emphasis mine). It is not that scheduled reducers are _only_ callable by the module.
> 
> I think this ticket was also just underspecified in general. I would add the following requirements:
> 
>     1. We call it `private` instead of internal.
> 
>     2. `private` has the same semantic meaning as a `private` table. (accessible only by the module, the owner, or collaborators)
> 
>     3. Scheduled reducers are `private` by default
> 
>     4. You can declare them public with:
> 
> 
> ```rust
> #[reducer(public)]
> fn foobar() {
>    // ...
> }
> ```

It was my understanding that we would, in fact, do exactly your 4 here, we just didn't need to do so in the MVP.

gefjon:
Ok @Shubham8287 , per discussion with @cloutiertyler out-of-band, please:

- Change terminology for these reducers from "Internal" to "Private," including in the `ModuleDef` and `RawModuleDefV10` definitions.
- Adjust `call_reducer` so that it checks if the sender identity is the owner or a collaborator before invoking, rather than unconditionally rejecting.
- Leave it so that private reducers do not get client codegen, though please do think about my previous comment about codegen for reducer events.

crates/codegen/src/util.rs
gefjon: Actually: for V9 modules, are scheduled reducers `Internal`/`Private` or `Public`? If the latter, then don't worry about this; we intend to change/break reducer events for WS 1.0 clients subscribed to ModuleDef v10 databases anyways.

Shubham8287: In ModuleDef V9, all functions are marked as public only with the intend to not break them.

crates/core/src/host/module_host.rs
```diff
@@ -1755,6 +1765,11 @@ impl ModuleHost {
         Ok(self.call_procedure_with_params(&procedure_def.name, params).await?)
     }
 
+    //TODO(shub) #4195: Also allow for collaborators along with owner
```
Shubham8287, line 1768: For now, I’ve restricted private functions to be callable only by the owner. Adding collaborator checks would require some extra work to expose that mechanism to public code, and we can do that additively later.

gefjon: I've created a ticket to track this.

PR #4182, opened 2026-02-02
Migrate off of `ubuntu-latest` for now

jdetter:
# Description of Changes

# Testing

- [ ] 
- [ ]


bfops (approved):
Works for me 👍

jdetter:
We can't really move anything forward right now with all of the github hosted runners down. I think we just need to wait it out.

image

PR #4089, opened 2026-01-21
Add `wasm32` support to the rust sdk (updated)

kistz:
# Description of Changes

Superscedes and is: #2704 so all credit goes to @thlsrms

This branch is updates to the latest master through cherry-picking leaving out the problematic switch to `anymap3` thats currently holding back the pr.
All commit authorships are kept since i only rebased it :>
Since @thlsrms seems to be unavailbale right now i'd say we push this to finish line @gefjon ?

# Testing

- [x] @jkukatzki built a bevy integration and it works in the browser! Check out his [Discord message](https://discord.com/channels/1037340874172014652/1134597033613869207/1463655684024438918)

PR #4183, opened 2026-02-02
wasm support for Rust SDK

bfops:
# Description of Changes

This PR adapts the Rust SDK test suite to work with the wasm version added in https://github.com/clockworklabs/SpacetimeDB/pull/4089 (which I've closed in favor of this PR).

Most of the changes revolve around wasm's different async semantics - everything runs in one thread, so things that relied on background threads didn't work directly. Several tests would lock up because something in them blocked synchronously, which blocked any background work from progressing.

We moved the test-clients contents into a `test_handlers.rs` so that it could be called from both `main` (for native tests) and `lib` (for wasm tests). To show what actually changed, use:
```bash
git diff --no-index -- <(git show origin/master:sdks/rust/tests/procedure-client/src/main.rs) sdks/rust/tests/procedure-client/src/test_handlers.rs
```
(or similar for other test-clients)

None, I think/hope.

2

# Testing

- [x] I've augmented the CI to also run the test suite with the `web` feature


sdks/rust/tests/connect_disconnect_client/Cargo.toml
```diff
@@ -6,10 +6,36 @@ license-file = "LICENSE"
...
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
+[lib]
+crate-type = ["cdylib", "rlib"]
+
+[features]
+default = ["native"]
+
+# Builds the existing CLI test client.
+native = []
```
bfops: possibly remove this?

sdks/rust/tests/test-client/src/main.rs
```diff
@@ -147,6 +173,16 @@ fn main() {
     }
 }
 
+#[cfg(all(target_arch = "wasm32", feature = "web"))]
+pub(crate) async fn dispatch_async(test: &str) {
+    match test {
+        "row-deduplication-r-join-s-and-r-joint" => {
+            exec_row_deduplication_r_join_s_and_r_join_t_async().await;
```
bfops: ?

.github/workflows/ci.yml
```diff
@@ -228,16 +228,45 @@ jobs:
         with:
           run_install: true
 
-      # Install cmake and emscripten for C++ module compilation tests.
-      - name: Install cmake and emscripten
+      # Install native and wasm toolchains required by SDK tests:
+      # - `emcc` for C++ module compilation tests.
+      # - `wasm32-unknown-unknown` target for Rust web client tests.
+      - name: Install native and wasm test prerequisites
```
bfops:
```suggestion
      # Install cmake and emscripten for C++ module compilation tests.
      - name: Install cmake and emscripten
```

.github/workflows/ci.yml
```diff
@@ -228,16 +228,46 @@ jobs:
...
+      # - `emcc` for C++ module compilation tests.
+      # - `wasm32-unknown-unknown` target for Rust web client tests.
       # Install cmake and emscripten for C++ module compilation tests.
       - name: Install cmake and emscripten
         run: |
           sudo apt-get update
           sudo apt-get install -y cmake
+
+          rustup target add wasm32-unknown-unknown
+
```
bfops:
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -228,6 +228,9 @@ jobs:
         with:
           run_install: true
 
+      # Install native and wasm toolchains required by SDK tests:
+      # - `emcc` for C++ module compilation tests.
+      # - `wasm32-unknown-unknown` target for Rust web client tests.
```
bfops:
```suggestion
```

crates/sql-parser/src/parser/mod.rs
```diff
@@ -207,9 +207,12 @@ pub(crate) fn parse_proj(expr: Expr) -> SqlParseResult<ProjectExpr> {
     }
 }
 
-// These types determine the size of [`parse_expr`]'s stack frame.
+// These types determine the size of [`parse_expr`]'s stack frame on 64-bit targets.
 // Changing their sizes will require updating the recursion limit to avoid stack overflows.
+// wasm32 has different type layouts, so this guard does not apply there.
```
bfops, line 212: I would like a (human) reviewer's input here on whether it is reasonable to omit these checks given the comment about stack overflows.

sdks/rust/src/client_cache.rs
```diff
@@ -373,8 +406,7 @@ impl<M: SpacetimeModule> ClientCache<M> {
         table_name: &'static str,
     ) -> &mut TableCache<Row> {
         self.tables
-            .entry::<HashMap<&'static str, TableCache<Row>>>()
-            .or_insert_with(Default::default)
+            .get_or_insert_with::<HashMap<&'static str, TableCache<Row>>>(Default::default)
```
bfops:
```suggestion
            .entry::<HashMap<&'static str, TableCache<Row>>>()
            .or_insert_with(Default::default)
```

sdks/rust/src/credentials.rs
```diff
@@ -8,144 +8,305 @@
...
-            Ok(bytes) => bytes,
-            Err(e) if matches!(e.kind(), std::io::ErrorKind::NotFound) => return Ok(None),
-            Err(source) => return Err(CredentialFileError::Read { path, source }),
-        };
-
-        let creds = bsatn::from_slice::<Credentials>(&bytes)
-            .map_err(|source| CredentialFileError::Deserialize { path, source })?;
-        Ok(Some(creds.token))
+#[cfg(feature = "web")]
+mod web_mod {
```
bfops, line 157: we should add some tests for this

crates/testing/src/sdk.rs
```diff
@@ -188,7 +198,7 @@ macro_rules! memoized {
 
         MEMOIZED
             .lock()
-            .unwrap()
+            .unwrap_or_else(|e| e.into_inner())
```
bfops: ?

sdks/rust/src/client_cache.rs
```diff
@@ -339,18 +342,48 @@ pub struct ClientCache<M: SpacetimeModule + ?Sized> {
...
     _module: PhantomData<M>,
 }
 
+// We intentionally avoid `anymap` here.
+//
+// In wasm test-client runs (`wasm32-unknown-unknown` under Node),
+// `anymap`'s TypeId hasher path can trigger an alignment-UB check panic:
+// `ptr::copy_nonoverlapping requires aligned pointers`.
+// Using this local `TypeId -> Box<dyn Any + Send + Sync>` map preserves the
+// required functionality without that runtime failure.
```
bfops: this is a little strange to me, I'd like to investigate further.

bfops: a human reviewer who's familiar with this code should take a look at this

sdks/rust/src/client_cache.rs
```diff
@@ -5,7 +5,7 @@
 use crate::callbacks::CallbackId;
 use crate::db_connection::{debug_log, PendingMutation, SharedCell};
 use crate::spacetime_module::{InModule, SpacetimeModule, TableUpdate, WithBsatn};
-use anymap::{any::Any, Map};
+use anymap3::Map;
```
bfops, line 8: Switched ClientCache from anymap to anymap3. The old anymap crate was tripping a wasm runtime panic in its custom TypeId hashing path; anymap3 provides the same type-indexed storage model without that broken implementation, so the SDK cache continues to work the same way on native and web targets.

sdks/rust/tests/connect_disconnect_client/src/main.rs
```diff
@@ -103,7 +133,39 @@ fn main() {
...
+    #[cfg(target_arch = "wasm32")]
+    new_connection.run_background_task();
+
+    reconnect_test_counter.wait_for_all().await;
+
+    // The second connection has no disconnect assertion, but it still needs an explicit
+    // shutdown on wasm so the background task releases its websocket before the test exits.
+    disconnect_connection(&new_connection).await;
+    #[cfg(not(target_arch = "wasm32"))]
+    reconnect_join_handle.join().unwrap();
```
bfops:
```suggestion
```

crates/testing/src/sdk.rs
```diff
@@ -384,24 +395,124 @@ fn compile_client(compile_command: &str, client_project: &str) {
...
+            let bindgen_out_dir = PathBuf::from(bindgen_out_dir);
+            let bindgen_out_dir = if bindgen_out_dir.is_absolute() {
+                bindgen_out_dir
+            } else {
+                Path::new(client_project).join(bindgen_out_dir)
+            };
+
+            create_dir_all(&bindgen_out_dir).expect("Failed to create wasm-bindgen out dir");
+
+            // TODO: Add a browser-faithful wasm runner alongside this Node-based path.
```
bfops:
```suggestion
            // TODO: Make a browser-faithful wasm runner.
```

sdks/rust/tests/test-client/src/main.rs
```diff
@@ -2435,14 +2554,21 @@ fn test_intra_query_bag_semantics_for_join() {
...
         }
-    });
+    })
+    .await;
+
+    // This test is only complete once both registered expectations have reported:
+    // `on_subscription_applied_nothing` confirms the initial state, and `pk_u32_on_delete`
+    // confirms the bag-semantics transition at the end. Without an explicit wait, the harness can
+    // finish before that final delete callback runs, which leaves the test under-synchronized.
+    test_counter.wait_for_all().await;
```
bfops: ?

sdks/rust/tests/test-client/src/main.rs
```diff
@@ -2552,24 +2671,34 @@ fn test_parameterized_subscription() {
...
         3,
         4,
         [ctr_for_test.clone(), ctr_for_subs.clone()],
         [sub_1, insert_1, update_1],
-    );
-    ctr_for_test.wait_for_all();
+    )
+    .await;
+
+    ctr_for_subs.wait_for_all().await;
```
bfops: these had to get moved out of the internal synchronous callback because in wasm everything runs in a single thread. Waiting for this inside the callback caused everything to lock up.

sdks/rust/tests/test-client/src/main.rs
```diff
@@ -2581,61 +2710,77 @@ fn test_parameterized_subscription() {
...
         "client_1",
         "Bob",
         [ctr_for_test.clone(), ctr_for_subs.clone()],
         [sub_1, ins_1],
-    );
-    ctr_for_test.wait_for_all();
+    )
+    .await;
+
+    ctr_for_subs.wait_for_all().await;
```
bfops: these had to get moved out of the internal synchronous callback because in wasm everything runs in a single thread. Waiting for this inside the callback caused everything to lock up.

sdks/rust/tests/test-client/src/main.rs
```diff
@@ -1954,52 +2053,57 @@ fn exec_caller_alice_receives_reducer_callback_but_not_bob() {
...
-                        anyhow::ensure!(
-                            matches!(event.event, Event::Transaction),
-                            "bob's event should be Transaction, but found {:?}",
-                            event.event,
-                        );
-                    }
-                    check_val(row.n, 24)
-                };
-                (one_u16_inserted.take().unwrap())(run_checks());
+        let pre_ins_counter = pre_ins_counter.clone();
```
bfops: small shimmy to deal with the different async semantics in wasm. we have to kick off both things and then await both.

crates/sql-parser/src/parser/mod.rs
gefjon: It is true that 32-bit targets, including wasm32, will have different layouts, and so these assertions will not be accurate. It's odd that they matter at all, 'cause I wouldn't expect this crate to be a part of the dependency tree for `spacetimedb-sdk`, so I wouldn't expect it to ever get built for wasm32. If we were compiling SpacetimeDB (not just the SDK) for 32-bit targets and we cared at all about its performance in those contexts, we'd probably also want to add an assertion about the size of these things on 32-bit targets. But, you know, we aren't and we don't, so let's not.

crates/testing/src/sdk.rs
```diff
@@ -103,11 +107,20 @@ pub struct Test {
     /// - `SPACETIME_SDK_TEST_CLIENT_PROJECT` bound to the `client_project` path.
     /// - `SPACETIME_SDK_TEST_DB_NAME` bound to the database identity or name.
     run_command: String,
+
+    client_runner: ClientRunner,
```
gefjon: What is this? What's it for? It looks like it's selecting between native clients and node clients, but if that's the case, why are we not just making the `run_command` be `node my_wasm_file` or whatever the hell?

crates/testing/src/sdk.rs
```diff
@@ -375,24 +395,124 @@ fn compile_client(compile_command: &str, client_project: &str) {
...
+                .env("RUST_LOG", rust_log)
+                .stderr_to_stdout()
+                .stdout_capture()
+                .unchecked()
+                .run()
+                .expect("Error running wasm client via node");
+
+            status_ok_or_panic(output, run_command, "(running web)");
+        }
+    }
```
gefjon: Yeah, I don't think this belongs here at all. The web target isn't some special magic that deserves special handling by the test framework any more than Unreal is. The test suite should just supply a `run_command` which contains an appropriate invocation of NodeJS.

sdks/rust/src/db_connection.rs
```diff
@@ -1009,17 +971,23 @@ but you must call one of them, or else the connection will never progress.
...
 
     /// Open a WebSocket connection, build an empty client cache, &c,
     /// to construct a [`DbContextImpl`].
     #[cfg(feature = "web")]
     async fn build_impl(self) -> crate::Result<DbContextImpl<M>> {
+        // The wasm/web SDK target runs under `wasm32-unknown-unknown`, where we do not
+        // have the native file APIs that back `with_debug_to_file`. Keeping the
+        // shared `extra_logging` field as `None` lets the rest of the connection and
+        // cache code stay unified without pretending that file logging works on web.
+        let extra_logging = None;
```
gefjon: Why wouldn't we just log to the browser console, if files are unavailable?

kistz:
Hey zeke, cool to see this moving forward :>
Feel free to close my pr if this superceeds it i dont mind at all👍

sdks/rust/tests/connect_disconnect_client/src/lib.rs
```diff
@@ -0,0 +1,20 @@
+#![allow(clippy::disallowed_macros)]
+
+// The shared client implementation lives in `main.rs`, but only the wasm build
+// needs to compile it as a library module. Native uses the binary entrypoint.
+#[cfg(all(target_arch = "wasm32", feature = "web"))]
+#[path = "main.rs"]
+mod cli;
```
gefjon: This is, uh, a weird way to avoid re-organizing this crate. I'd be much happier if we had:
- `main.rs` containing only the actual CLI parts.
- `lib.rs` containing only the WASM entrypoint (and `mod` decls).
- A new file, probably called `test_handlers.rs` or something, that contains the dispatch function and the meat.

Obviously this applies to all of the test clients, not just this one.

sdks/rust/tests/connect_disconnect_client/src/main.rs
```diff
@@ -1,18 +1,46 @@
...
 
 const LOCALHOST: &str = "http://localhost:3000";
 
+#[cfg(all(target_arch = "wasm32", feature = "web"))]
+static WEB_DB_NAME: OnceLock<String> = OnceLock::new();
+
+#[cfg(all(target_arch = "wasm32", feature = "web"))]
+pub(crate) fn set_web_db_name(db_name: String) {
+    WEB_DB_NAME.set(db_name).expect("WASM DB name was already initialized");
+}
```
gefjon: Just make the database name an additional argument to the tests, and have the CLI code read it out of the env var while the WASM entrypoint passes it normal.

sdks/rust/tests/test-client/src/main.rs
```diff
@@ -27,13 +30,41 @@ use unique_test_table::{insert_then_delete_one, UniqueTestTable};
 
 const LOCALHOST: &str = "http://localhost:3000";
 
+fn fixed_test_timestamp() -> Timestamp {
+    // `Timestamp::now()` is stubbed on `wasm32-unknown-unknown`, so client-side tests
+    // that need a timestamp value must use a deterministic literal instead of wall-clock time.
```
gefjon: This should be a doc comment external to the function, not a regular comment interior to it.

sdks/rust/tests/test-client/src/main.rs
```diff
@@ -920,15 +1004,15 @@ fn exec_insert_caller_uuid_v4() {
         }
     });
 
-    test_counter.wait_for_all();*/
+    test_counter.wait_for_all().await;*/
 }
 
 /// This tests that we can serialize and deserialize `Uuid` in various contexts.
-fn exec_insert_caller_uuid_v7() {}
+async fn exec_insert_caller_uuid_v7() {}
```
gefjon: Are the UUID tests stubbed on master? Geez, they are. Don't love that.

sdks/rust/tests/test-client/src/main.rs
```diff
@@ -1798,106 +1840,133 @@ fn exec_reauth_part_2() {
...
-    test_counter.wait_for_all();
+    test_counter.wait_for_all().await;
+}
+
+#[cfg(target_arch = "wasm32")]
+async fn exec_reauth_part_1() {
+    // TODO: Re-enable this once the wasm runner grows the browser-faithful mode
+    // described in `crates/testing/src/sdk.rs`. This test is about persisting web
+    // credentials across separate runs, and the current Node-based wasm harness does
+    // not exercise the browser cookie/storage APIs that web reauth depends on.
```
gefjon: Doc comment!

sdks/rust/tests/test-client/src/main.rs
```diff
@@ -1798,106 +1840,133 @@ fn exec_reauth_part_2() {
...
+}
+
+#[cfg(target_arch = "wasm32")]
+async fn exec_reauth_part_2() {
+    // See the TODO above and in `crates/testing/src/sdk.rs`: this is disabled for the
+    // current Node-based wasm harness for the same reason as `exec_reauth_part_1`.
```
gefjon: Doc comment!

sdks/rust/tests/test-client/src/main.rs
bfops: agreed

crates/testing/src/sdk.rs
bfops: sounds great to me, thank you!

sdks/rust/tests/connect_disconnect_client/src/lib.rs
bfops: yeah that's cleaner. done.

sdks/rust/src/db_connection.rs
bfops: as in, augment `extra_logging` so that it can log to a variety of backends rather than just a file? I can do that, but I would somewhat prefer it to be in a separate PR that this one is based on.

bfops:
> Hey zeke, cool to see this moving forward :> Feel free to close my pr if this superceeds it i dont mind at all👍

I might do that just for ease of changes and review. Thank you!

sdks/rust/tests/connect_disconnect_client/src/main.rs
bfops: yeah that's better, thank you

sdks/rust/tests/test-client/src/main.rs
bfops: done

sdks/rust/tests/test-client/src/main.rs
bfops: done

sdks/rust/tests/test-client/src/main.rs
bfops: done

crates/sql-parser/src/parser/mod.rs
bfops: it looks like sdk -> schema -> sql-parser.

this fails without the `target_pointer_width` constraints:
```
$ cargo check -p spacetimedb-sdk --target wasm32-unknown-unknown --no-default-features --features web
    Checking spacetimedb-sql-parser v2.0.5 (/home/work/SpacetimeDBPrivate/public/crates/sql-parser)
error[E0080]: evaluation panicked: assertion failed: size_of::<Expr>() == 168
   --> crates/sql-parser/src/parser/mod.rs:213:15
    |
213 | const _: () = assert!(size_of::<Expr>() == 168);
    |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ evaluation of `parser::_` failed here

error[E0080]: evaluation panicked: assertion failed: size_of::<SqlParseResult<SqlExpr>>() == 40
   --> crates/sql-parser/src/parser/mod.rs:214:15
    |
214 | const _: () = assert!(size_of::<SqlParseResult<SqlExpr>>() == 40);
    |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ evaluation of `parser::_` failed here

For more information about this error, try `rustc --explain E0080`.
error: could not compile `spacetimedb-sql-parser` (lib) due to 2 previous errors

```

sdks/rust/tests/connect_disconnect_client/src/lib.rs
```diff
@@ -0,0 +1,18 @@
+#![allow(clippy::disallowed_macros)]
+
+#[cfg(all(target_arch = "wasm32", feature = "web"))]
+mod module_bindings;
+#[cfg(all(target_arch = "wasm32", feature = "web"))]
+mod test_handlers;
```
gefjon: Why are we declaring these in both the `lib.rs` and the `main.rs`? Just have the executable crate use the library crate.

bfops: fair point. how's this?

tools/ci/src/main.rs
```diff
@@ -338,6 +338,20 @@ fn main() -> Result<()> {
...
             )
             .run()?;
+            // Run the same SDK suite against wasm+web test clients.
+            cmd!(
+                "cargo",
+                "test",
+                "-p",
+                "spacetimedb-sdk",
+                "--features",
+                "allow_loopback_http_for_tests,web",
```
jdetter: is `--features web` descriptive enough for what we're testing here? Perhaps something like `rust-sdk-wask` or something would be better? I don't feel strongly about it though as there's a good comment above

.github/workflows/ci.yml
```diff
@@ -238,6 +238,30 @@ jobs:
...
+            echo "Failed to determine wasm-bindgen version from Cargo.lock"
+            exit 1
+          fi
+
+          INSTALLED_WASM_BINDGEN_VERSION="$(wasm-bindgen --version 2>/dev/null | awk '{print $2}' || true)"
+          if [ "${INSTALLED_WASM_BINDGEN_VERSION}" != "${REQUIRED_WASM_BINDGEN_VERSION}" ]; then
+            cargo install --locked --force wasm-bindgen-cli --version "${REQUIRED_WASM_BINDGEN_VERSION}"
+          fi
+
+          wasm-bindgen --version
```
jdetter, line 263: This section looks great, very thorough. Thanks Zeke 👍

tools/ci/src/main.rs
bfops: it's on the `spacetimedb-sdk`, so I believe in multi-package cases the name would be `spacetimedb-sdk/web`. I could rename it to `wasm` if that seems better? I agree `web` isn't necessary clear.

jdetter: yeah `wasm` would be a lot better, thanks!

sdks/rust/src/db_connection.rs
bfops: I added a `TODO` for this to log to the browser console

tools/ci/src/main.rs
bfops: renamed to `browser` after out-of-band discussion

PR #4187, opened 2026-02-02
[TS] Module-side performance improvements

coolreader18:
# Description of Changes

Basically, cut down on creating new objects in the hot path as much as possible.

1

# Testing

- [ ] 
- [ ]

### DIFF
--- a/crates/bindings-typescript/test-react-router-app/server/src/lib.rs
+++ b/crates/bindings-typescript/test-react-router-app/server/src/lib.rs
@@ -22,23 +22,23 @@ fn init(ctx: &ReducerContext) {
 
 #[reducer(client_connected)]
 fn client_connected(ctx: &ReducerContext) {
-    let existing_user = ctx.db.offline_user().identity().find(ctx.sender);
+    let existing_user = ctx.db.offline_user().identity().find(ctx.sender());
     if let Some(user) = existing_user {
         ctx.db.user().insert(user);
-        ctx.db.offline_user().identity().delete(ctx.sender);
+        ctx.db.offline_user().identity().delete(ctx.sender());
         return;
     }
     ctx.db.offline_user().insert(User {
-        identity: ctx.sender,
+        identity: ctx.sender(),
         has_incremented_count: 0,
     });
 }
 
 #[reducer(client_disconnected)]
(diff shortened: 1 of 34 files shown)


crates/bindings-typescript/src/server/runtime.ts
```diff
@@ -205,6 +206,19 @@ export const ReducerCtxImpl = class ReducerCtx<
     this.db = getDbView();
   }
 
+  static reset(
+    me: InstanceType<typeof this>,
```
cloutiertyler, line 212: NIT: I think the standard here is `self`

crates/bindings-typescript/tsup.config.ts
```diff
@@ -172,11 +172,11 @@ export default defineConfig([
...
   },
 
-  // Server subpath (SSR / node-friendly): dist/server/index.{mjs,cjs}
+  // Server subpath (SSR / node-friendly): dist/server/index.mjs
   {
     entry: { index: 'src/server/index.ts' },
-    format: ['esm', 'cjs'],
-    target: 'es2022',
+    format: 'esm',
+    target: 'esnext',
```
cloutiertyler, line 179: Hmm can you explain these changes? I think this might impact where the SDK can be used.

coolreader18: It won't, cause this is just the server, not the client-side sdk. That only ever gets bundled by `spacetime build`.

cloutiertyler: Roger

crates/bindings-typescript/src/server/runtime.ts
```diff
@@ -205,6 +206,19 @@ export const ReducerCtxImpl = class ReducerCtx<
     this.db = getDbView();
   }
 
+  static reset(
```
cloutiertyler, line 211: Can we get a comment here?

crates/bindings-typescript/src/server/runtime.ts
```diff
@@ -178,6 +177,8 @@ class AuthCtxImpl implements AuthCtx {
   }
 }
 
+let REDUCER_CTX: InstanceType<typeof ReducerCtxImpl> | undefined;
```
cloutiertyler, line 181: Can we get a comment here explaining this?

PR #4192, opened 2026-02-03
gitignore - properly ignore nuget config files

bfops:
# Description of Changes

Ignoring nuget config files regardless of casing, as well as in the root directory since we generate them there now.

None.

1

# Testing

- [x] Files generated by `write-nuget-config.sh` are now properly ignored


rekhoff (approved):
LGTM, no objections

PR #4193, opened 2026-02-03
`write-nuget-config.sh` resolves paths properly

bfops:
# Description of Changes

The script was running `cd` but using the STDB path as passed, so if it was run from e.g. the root directory, it would generate incorrect paths.

None

1

# Testing

- [x] Running `sdks/csharp/tools~/write-nuget-config.sh .` from the root directory generates a `NuGet.Config` that uses correct paths


rekhoff (approved):
No objections here

PR #4194, opened 2026-02-03
RawModuleDefV10 from V8 modules.

Shubham8287:
# Description of Changes

V8 Modules to expose `RawModuleDefV10` 

It breaks client, hence should be merged in `2.0-breaking-changes`

# Testing

- [ ] Existing smoketests with run on new module def to verify if its correct.

### DIFF
--- a/crates/bindings-typescript/src/lib/table.ts
+++ b/crates/bindings-typescript/src/lib/table.ts
@@ -32,6 +32,7 @@ import type {
   ValidateColumnMetadata,
 } from './type_util';
 import { toPascalCase } from './util';
+import BinaryWriter from './binary_writer';
 
 export type AlgebraicTypeRef = number;
 type ColId = number;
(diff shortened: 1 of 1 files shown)


Shubham8287:
CIs are failing due to base branch.

crates/schema/src/def/validate/v10.rs
```diff
@@ -205,6 +205,50 @@ pub fn validate(def: RawModuleDefV10) -> Result<ModuleDef> {
     })
 }
 
+/// Change the visibility of scheduled functions and lifecycle reducers to Internal.
```
Shubham8287, line 208: This is a code motion from `raw_def/v10.rs` to here (in validation) to re-use logic across Rust, Typescript modules.

crates/bindings-typescript/src/lib/schema.ts
```diff
@@ -115,10 +116,17 @@ type CompoundTypeCache = Map<
   RefBuilder<any, any>
 >;
 
-type ModuleDef = Infer<typeof RawModuleDefV9>;
+export type ModuleDef = {
+  [S in Infer<typeof RawModuleDefV10Section> as Uncapitalize<
+    S['tag']
+  >]: S['value'];
+};
```
coolreader18, line 123: Clever!

PR #4197, opened 2026-02-03
[Moved] Disabled codegen for private tables by default

JasonAtClockwork:
Due to troubles with switching from `2.0-breaking-changes` back to `master` I've set up a new clean PR: [Add --include-private and default private tables to not generate](<https://github.com/clockworklabs/SpacetimeDB/pull/4241>)

# Description of Changes
Updated the codegen table iteration functions to take in a parameter to include private and defaulted it false in all locations for the supported languages. 
- Updated the util.rs functions for iterating tables to check for a boolean flag include_private and defaulted it false in all used places
- Updated the CLI to return a list of all private tables not included (added a TODO to check the --include-private opt):
```bash
Build finished successfully.
Skipping private tables during codegen: secret_note, secret_order, secret_person.
```

Technically API breaking as the private tables will no longer be available. (GitHub labels are not working at the moment)

1 - Simple change the testing took longer

# Testing

Turns out when you remove private tables you invalidate most of the module_bindings across the system!

- [x] Rust test SDK for all languages
- [x] C# SDK tests
- [x] C# dotnet tests
- [x] Updated and checked snap files
- [x] Updated Blackholio (Unreal + Unity) module_bindings and tested
- [x] Ran Unreal SDK tests


cloutiertyler:
Merged in a separate PR.

PR #4199, opened 2026-02-04
`spacetime.json` config implementation

drogus:
# Description of Changes

This PR implements support for the `spacetime.json` configuration file that can be used to set up common `generate` and `publish` targets. An example of `spacetime.json` could look like this:

```
{
  "dev_run": "pnpm dev",
  "generate": [
    { "out-dir": "./foobar", "module-path": "region-module", "language": "c-sharp" },
    { "out-dir": "./global", "module-path": "global-module", "language": "c-sharp" },
  ],
  "publish": {
    "database": "bitcraft",
    "module-path": "spacetimedb",
    "server": "local",
    "children": [
      { "database": "region-1", "module-path": "region-module", server: "local" },
      { "database": "region-2", "module-path": "region-module", server: "local" }
    ]
  }
}
```

With this config, running `spacetime generate` without any arguments would generate bindings for two targets: `region-module` and `global-module`. `spacetime publish` without any arguments would publish three modules, starting from the parent: `bitcraft`, `region-1`, and `region-2`. On top of that, the command `pnpm dev` would be executed when using `spacetime dev`.

It is also possible to pass additional command line arguments when calling the `publish` and `generate` commands, but there are certain limitations. There is a special case when passing either a module path to generate or a module name to publish. Doing that will filter out entries in the config file that do not match. For example, running:

```
spacetime generate --project-path global-module
```

would only generate bindings for the second entry in the `generate` list.

In a similar fashion, running:

```
spacetime publish region-1
```

would only publish the child database with the name `region-1`

Passing other existing arguments is also possible, but not all of the arguments are available for multiple configs. For example, when running `spacetime publish --server maincloud`, the publish command would be applied to all of the modules listed in the config file, but the `server` value from the command line arguments would take precedence. Running with arguments like `--bin-path` would, however, would throw an error as `--bin-path` makes sense only in a context of a specific module, thus this wouldn't work: `spacetime publish --bin-path spacetimedb/target/debug/bitcraft.wasm`. I will throw an error unless there is only one entry to process, thus `spacetime publish --bin-path spacetimedb/target/debug/bitcraft.wasm bitcraft` would work, as it filters the publish targets to one entry.

None

3

The config file in itself is not overly complex, but when coupled with the CLI it is somewhat tricky to get right. There are also some changes that I had to make to how clap arguments are validated - because the values can now come from both the config file and the clap config, we can't use some of the built-in validations like `required`, or at least I haven't found a clean way to do so.

# Testing

I've added some automated tests, but more tests and manual testing is coming.


cloutiertyler:
This one needs rebasing

crates/bindings-typescript/test-app/package.json
```diff
@@ -14,7 +14,7 @@
     "preview": "vite preview",
     "generate": "cargo run -p gen-bindings -- --replacement ../../../src/index && prettier --write src/module_bindings",
     "spacetime:generate": "spacetime generate --lang typescript --out-dir src/module_bindings --project-path server",
-    "spacetime:start": "spacetime start",
+    "spacetime:start": "spacetime start server",
```
cloutiertyler:
```suggestion
    "spacetime:start": "spacetime start",
```

crates/bindings-typescript/test-react-router-app/package.json
```diff
@@ -13,10 +13,10 @@
     "lint": "eslint . && prettier . --check --ignore-path ../../../.prettierignore",
     "preview": "vite preview",
     "generate": "cargo run -p gen-bindings -- --replacement ../../../src/index && prettier --write src/module_bindings",
-    "spacetime:generate": "spacetime generate --lang typescript --out-dir src/module_bindings --project-path server",
+    "spacetime:generate": "spacetime generate --lang typescript --out-dir src/module_bindings --module-path server",
     "spacetime:start": "spacetime start server",
```
cloutiertyler:
```suggestion
    "spacetime:start": "spacetime start",
```

crates/cli/src/spacetime_config.rs
```diff
@@ -0,0 +1,2807 @@
...
+        for (key, value) in overlay_obj {
+            base_obj.insert(key.clone(), value.clone());
+        }
+    }
+}
+
+/// Find and load config with environment layering from the current directory.
+///
+/// Loading order (each overlays the previous via top-level key replacement):
+/// 1. `spacetime.json` (required)
```
cloutiertyler, line 889: Hmm is this really required? If so, should it be?

crates/smoketests/modules/add-remove-index-indexed/src/lib.rs
cloutiertyler, line 1: I'm not sure why there are formatting changes in these. I presume somehow I formatted these and they're not in the above workspace. Should I revert these?

cloutiertyler: I reverted these changes.

docs/docs/00300-resources/00200-reference/00100-cli-reference/00100-cli-reference.md
```diff
@@ -323,23 +327,25 @@ Run `spacetime rename --help` for more detailed information.
 
 Generate client files for a spacetime module.
 
-**Usage:** `spacetime spacetime generate --lang <LANG> --out-dir <DIR> [--project-path <DIR> | --bin-path <PATH> | --module-name <MODULE_NAME> | --uproject-dir <DIR> | --include-private]`
+**Usage:** `spacetime spacetime generate [DATABASE] --lang <LANG> --out-dir <DIR> [--module-path <DIR> | --bin-path <PATH> | --unreal-module-name <MODULE_NAME> | --uproject-dir <DIR> | --include-private]`
```
cloutiertyler:
```suggestion
**Usage:** `spacetime generate [DATABASE] --lang <LANG> --out-dir <DIR> [--module-path <DIR> | --bin-path <PATH> | --unreal-module-name <MODULE_NAME> | --uproject-dir <DIR> | --include-private]`
```

cloutiertyler (approved):

PR #4210, opened 2026-02-05
[TS] Make ProcedureCtx a class

coolreader18:
# Description of Changes

Followup to #3958. Also adds `random` to ProcedureCtx.

1

# Testing

n/a

### DIFF
--- a/crates/bindings/src/lib.rs
+++ b/crates/bindings/src/lib.rs
@@ -938,7 +938,7 @@ pub struct ReducerContext {
     ///
     /// Will be `None` for certain reducers invoked automatically by the host,
     /// including `init` and scheduled reducers.
-    pub connection_id: Option<ConnectionId>,
+    connection_id: Option<ConnectionId>,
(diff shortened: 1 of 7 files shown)


crates/bindings-typescript/src/server/runtime.ts
```diff
@@ -228,7 +228,6 @@ export const ReducerCtxImpl = class ReducerCtx<
    * Create a new random {@link Uuid} `v4` using this `ReducerCtx`'s RNG.
    */
   newUuidV4(): Uuid {
-    // TODO: Use a spacetime RNG when available
```
cloutiertyler, line 243: Isn't this still true?

coolreader18: No, cause it is available now.

coolreader18: We use `this.random` on the next line.

Issue #4207, opened 2026-02-05
Using SpacetimeDB `1.12.0` on Unity `6000.3.2.f1` reports many "does not exist in the namespace 'SpacetimeDB'"


rekhoff:
Community Discord members ezoray and mixin have reported errors when using SpacetimeDB version `1.12.0` on Unity `6000.3.2.f1`

Doing so results in hundreds of errors in the Unity logs about missing data types within SpacetimeDB.
<img width="963" height="302" alt="Image" src="https://github.com/user-attachments/assets/a660b1bc-cda3-4127-b936-4c395aa3cf5c" />

# Steps to Reproduce
1) Start a new Unity project on Unity version `6000.3.2.f1`
2) Open package manager and import Unity using `Import package from git URL...` , entering the URL of:
`https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk.git`
3) Observe errors in Unity's Console window

rekhoff:
Patch released. Following steps to reproduce no longer results in errors.

PR #4211, opened 2026-02-05
Moved C# SDK's `packages` ignore from SDK's `.gitignore` to project root

rekhoff:
# Description of Changes

Moves the `root/sdks/csharp/packages/` directory from the `root/sdks/csharp/.gitignore` to `root/.gitignore`.
This is to resolve Issue #4207 , which was introduced during the DLLs generation rework.

`/sdks/csharp/packages/` should be ignored by git, to prevent locally generated DLLs from accidentally getting added to the repo.

**Root cause:**
 When Unity imports a package by URL, it looks at the `.gitignore` list and filters by it, causing the entire `packages` directory to not be imported. In earlier versions of Unity (like `2022.3` where this was initially tested), this was not a problem because Unity would allow us to do a `dotnet restore`. In modern version of Unity, this is no longer allowed as all imported packages are considered immutable.

Because the `/sdks/csharp/` is cloned into `com.clockworklabs.spacetimedbsdk`, if we move at what level the `root/sdks/csharp/packages/` is ignored at, we can prevent the accidental inclusion of the DLLs from developers in the SpacetimeDB repo, while ensuring `com.clockworklabs.spacetimedbsdk` does not contain a reference to ignore the `packages` directory.

No API or ABI changes

1

If you answered more than a 2, explain what is complex about the PR,
and what other components it interacts with in potentially concerning ways.  -->

# Testing

- [X] Tested removing the `packages` directory from the `.gitignore` list on a custom branch of `com.clockworklabs.spacetimedbsdk` to observe the behavior in both Unity `6000.3.2f1` and `2002.3.62f2`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -245,3 +245,7 @@ nul
 
 [Nn][Uu][Gg][Ee][Tt].[Cc][Oo][Nn][Ff][Ii][Gg]
 [Nn][Uu][Gg][Ee][Tt].[Cc][Oo][Nn][Ff][Ii][Gg].meta
+ 
+# csharp SDK packages
+/sdks/csharp/packages/
+/sdks/csharp/packages.meta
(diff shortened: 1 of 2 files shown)


bfops (approved):
This works for me. The alternative would be to un-ignore the directory during the release flow itself, but I think this works well too.

PR #4213, opened 2026-02-05
Implement server-side support for the v2 websocket protocol

cloutiertyler:
# Description of Changes

This adds the v2 websocket protocol and adds support on the server side. For context on many of the changes/decisions, you can look at the discussion on https://github.com/clockworklabs/SpacetimeDB/pull/4023.

To restate some of the key changes:
 - The reducer event information is no longer sent with transaction updates (because we don't want to broadcast reducer call information anymore).
 - If a client calls a reducer, they are sent a `ReducerResult` which includes the outcome of the reducer call and and related row updates for queries that the client is subscribed to.
 - We no longer dedupe queries that appear in multiple query sets for the same client. This is because we are moving toward per-query storage.
 - Related to that, Unsubscribe requests have an option to send the related rows. We need this for now, since clients don't have per-query storage implemented yet.
 - We don't have the json format in v2.

Notes for reviewers:
 - This moves around the messages in `crates/client-api-messages/src/websocket` (into `common`, `v1`, and `v2`), and this renaming of existing messages adds a lot of noise to the PR.
 - In many places, I chose to duplicate a lot of code to have a v1 version and a v2 version. I went with this to make it easier to remove the v1 version in the future (hopefully we can just fully delete most of the v1 functions).
 - `module_subscription_manager.rs` has probably has the biggest changes, since we now track queries by query_set_id, and we get to remove some complexity of v1's FormatSwitch.

The v1 protocol still works, though we won't send the reducer event info for v10 modules.

4. This touches a lot of places.

# Testing

Unit testing is pretty minimal for the new code paths. I've done some manual e2e testing with the typescript quickstart, and this has been tested with a different branch implementing the v2 rust client.


crates/client-api-messages/src/websocket/v1.rs
```diff
@@ -0,0 +1,765 @@
...
+}
+
+/// The tag recognized by the host and SDKs to mean no compression of a [`ServerMessage`].
+pub const SERVER_MSG_COMPRESSION_TAG_NONE: u8 = 0;
+
+/// The tag recognized by the host and SDKs to mean brotli compression of a [`ServerMessage`].
+pub const SERVER_MSG_COMPRESSION_TAG_BROTLI: u8 = 1;
+
+/// The tag recognized by the host and SDKs to mean gzip compression of a [`ServerMessage`].
+pub const SERVER_MSG_COMPRESSION_TAG_GZIP: u8 = 2;
```
Centril: Is this not common stuff? Surely v2 also does whole-message compression?

crates/client-api-messages/src/websocket/v2.rs
```diff
@@ -0,0 +1,445 @@
...
+    ProcedureResult(ProcedureResult),
+}
+
+#[derive(SpacetimeType, Debug)]
+#[sats(crate = spacetimedb_lib)]
+pub struct InitialConnection {
+    pub identity: Identity,
+    pub connection_id: ConnectionId,
+    pub token: Box<str>,
+}
```
Centril, line 204: This looks like `IdentityToken` in v1 -- move both to common?

crates/client-api-messages/src/websocket/v2.rs
```diff
@@ -0,0 +1,445 @@
...
+/// That [`QuerySetId`] may then be re-used at the client's discretion.
+#[derive(SpacetimeType, Debug)]
+#[sats(crate = spacetimedb_lib)]
+pub struct UnsubscribeApplied {
+    /// Provided by the client via the `Subscribe` message.
+    /// TODO: switch to subscription id?
+    pub request_id: u32,
+    /// The ID included in the `SubscribeApplied` and `Unsubscribe` messages.
+    pub query_set_id: QuerySetId,
+    /// Rows to be removed from the client cache. Only populated if the Unsubscribe message requested it.
```
Centril: mention the specific flag?

crates/client-api-messages/src/websocket/v2.rs
```diff
@@ -0,0 +1,445 @@
...
+///
+/// This message's `request_id` and `query_set_id` will match those the client provided in the [`Unsubscribe`] message.
+///
+/// After receiving this message, the client will no longer receive any [`QuerySetUpdate`]s for the included [`QuerySetId`].
+/// That [`QuerySetId`] may then be re-used at the client's discretion.
+#[derive(SpacetimeType, Debug)]
+#[sats(crate = spacetimedb_lib)]
+pub struct UnsubscribeApplied {
+    /// Provided by the client via the `Subscribe` message.
+    /// TODO: switch to subscription id?
```
Centril: Why?

crates/client-api-messages/src/websocket/v2.rs
```diff
@@ -0,0 +1,445 @@
...
+#[sats(crate = spacetimedb_lib)]
+pub struct PersistentTableRows {
+    pub inserts: BsatnRowList,
+    pub deletes: BsatnRowList,
+}
+
+#[derive(SpacetimeType, Debug, Clone)]
+#[sats(crate = spacetimedb_lib)]
+pub struct EventTableRows {
+    pub events: BsatnRowList,
```
Centril, line 354: The latter could be represented as the former with empty deletes. Why special case?

crates/client-api-messages/src/websocket/v2.rs
```diff
@@ -0,0 +1,445 @@
...
+    /// The reducer returned successfully and its transaction committed.
+    /// The return value and [`TransactionUpdate`] are included here.
+    Ok(ReducerOk),
+    /// The reducer returned successfully and its transaction committed,
+    /// but its return value was zero bytes and its [`TransactionUpdate`] contained zero [`QuerySetUpdate`]s.
+    ///
+    /// This variant is an optimization which saves 8 bytes of wire size,
+    /// due to the BSATN format's using 4 bytes for the length of a variable-length object,
+    /// such as the `ret_value` of [`ReducerOk`] and the `query_sets` of [`TransactionUpdate`].
+    Okmpty,
```
Centril:
```suggestion
    OkEmpty,
```

crates/client-api-messages/src/websocket/v2.rs
```diff
@@ -0,0 +1,445 @@
...
+    /// The payload is an error message, which is intended for diagnostic purposes only,
+    /// and is not intended to have a stable or parseable format.
+    InternalError(Box<str>),
+}
+
+#[derive(SpacetimeType, Debug)]
+#[sats(crate = spacetimedb_lib)]
+pub struct ReducerOk {
+    pub ret_value: Bytes,
+    pub transaction_update: TransactionUpdate,
```
Centril, line 414: docs :)

crates/client-api-messages/src/websocket/v2.rs
Centril: Answered out of band, but a comment with the rationale would be neat.

crates/client-api/src/routes/subscribe.rs
```diff
@@ -716,13 +762,15 @@ async fn ws_recv_task<MessageHandler>(
     while let Some((data, timer)) = recv_handler.next().await {
         let result = message_handler(data, timer).await;
         if let Err(e) = result {
-            if let MessageHandleError::Execution(err) = e {
-                log::error!("{err:#}");
-                // If the send task has exited, also exit this recv task.
-                if unordered_tx.send(err.into()).is_err() {
-                    break;
+            if ws_version == WsVersion::V1 {
```
Centril, line 765: This change I don't understand.

crates/client-api/src/routes/subscribe.rs
```diff
@@ -1258,19 +1310,49 @@ async fn ws_encode_task(
...
-                metrics.report(workload, num_rows, stats);
-                if frames.try_for_each(|frame| outgoing_frames.send(frame)).is_err() {
-                    break;
-                }
+            OutboundWsMessage::Message(message) => {
+                if config.version == WsVersion::V2 {
+                    let (workload, num_rows) = match &message {
+                        OutboundMessage::V2(_) => (message.workload(), message.num_rows()),
+                        OutboundMessage::V1(_) => (None, None),
+                    };
```
Centril: Not sure why this match exists... can it not be integrated in the latter match?

crates/client-api/src/routes/subscribe.rs
```diff
@@ -1249,7 +1297,11 @@ async fn ws_encode_task(
         let buf = buf_pool.pop().unwrap_or_else(|| SerializeBuffer::new(config));
 
         let in_use_buf = match message {
-            OutboundMessage::Error(message) => {
+            OutboundWsMessage::Error(message) => {
+                if config.version == WsVersion::V2 {
+                    log::warn!("dropping v2 error message until v2 serialization is implemented");
+                    continue;
+                }
```
Centril, line 1304: Is there another PR incoming for server side stuff?

crates/client-api/src/routes/subscribe.rs
```diff
@@ -1258,19 +1310,49 @@ async fn ws_encode_task(
...
-                let is_large = num_rows.is_some_and(|n| n > 1024);
-
-                let (stats, in_use, mut frames) =
-                    ws_encode_message(config, buf, message, is_large, &bsatn_rlb_pool).await;
-                metrics.report(workload, num_rows, stats);
-                if frames.try_for_each(|frame| outgoing_frames.send(frame)).is_err() {
-                    break;
-                }
+            OutboundWsMessage::Message(message) => {
+                if config.version == WsVersion::V2 {
```
Centril: `WsVersion::V2` combined with `OutboundMessage::V1` should be impossible, no? In that case we don't need to consider the version at all, but just pattern match on the `message`.

crates/core/src/client/client_connection.rs
```diff
@@ -64,8 +69,10 @@ impl Protocol {
 pub struct ClientConfig {
     /// The client's desired protocol (format) when the host replies.
     pub protocol: Protocol,
+    /// The websocket protocol version negotiated during the handshake.
+    pub version: WsVersion,
     /// The client's desired (conditional) compression algorithm, if any.
-    pub compression: Compression,
+    pub compression: ws_v1::Compression,
```
Centril: Move type to common?

crates/core/src/host/module_host.rs
```diff
@@ -2129,6 +2209,134 @@ impl ModuleHost {
...
+        rlb_pool: impl 'static + Send + RowListBuilderSource<ws_v1::BsatnFormat>,
+    ) -> Result<Option<ExecutionMetrics>, anyhow::Error> {
+        let (tx_offset_sender, tx_offset_receiver) = oneshot::channel();
+        let tx = scopeguard::guard(db.begin_tx(Workload::Sql), |tx| {
+            let (tx_offset, tx_metrics, reducer) = db.release_tx(tx);
+            let _ = tx_offset_sender.send(tx_offset);
+            db.report_read_tx_metrics(reducer, tx_metrics);
+        });
+
+        let result: Result<(ws_v2::SingleTableRows, ExecutionMetrics), anyhow::Error> = (|| {
```
Centril, line 2262: Duplicating code like this makes me real nervous of maintaining both for a year.
I'd like to ask for things like this to be deduped after 2.0 has shipped.

crates/core/src/client/consume_each_list.rs
```diff
@@ -65,14 +62,14 @@ impl ConsumeEachBuffer for CompressableQueryUpdate<BsatnFormat> {
...
         self.deletes.consume_each_list(each);
         self.inserts.consume_each_list(each);
     }
 }
 
-impl ConsumeEachBuffer for BsatnRowList {
+impl ConsumeEachBuffer for ws_v1::BsatnRowList {
     fn consume_each_list(self, each: &mut impl FnMut(Bytes)) {
         let (_, buffer) = self.into_inner();
         each(buffer);
```
Centril, line 133: No v2 types here, why?

crates/core/src/client/messages.rs
```diff
@@ -165,19 +166,45 @@ pub fn serialize(
...
+        bsatn::to_writer(w.into_inner(), &msg).expect("should be able to bsatn encode v2 message");
+    });
+
+    match decide_compression(srv_msg.len(), compression) {
+        ws_v1::Compression::None => buffer.uncompressed(),
+        ws_v1::Compression::Brotli => {
+            buffer.compress_with_tag(ws_v1::SERVER_MSG_COMPRESSION_TAG_BROTLI, brotli_compress)
+        }
+        ws_v1::Compression::Gzip => buffer.compress_with_tag(ws_v1::SERVER_MSG_COMPRESSION_TAG_GZIP, gzip_compress),
+    }
```
Centril, line 206: Given that these diverged already which they shouldn't have, I'd like for these to be fused into a generic function taking `impl Serialize + ConsumeEachBuffer`.

crates/core/src/client/messages.rs
Centril, line 165: This is an important optimization that is lost in v2.

crates/core/src/client/messages.rs
```diff
@@ -230,14 +257,87 @@ impl ToProtocol for SerializableMessage {
...
+            Self::V2(message) => match message {
+                ws_v2::ServerMessage::InitialConnection(_) => None,
+                ws_v2::ServerMessage::SubscribeApplied(_) => Some(WorkloadType::Subscribe),
+                ws_v2::ServerMessage::UnsubscribeApplied(_) => Some(WorkloadType::Unsubscribe),
+                ws_v2::ServerMessage::SubscriptionError(_) => None,
+                ws_v2::ServerMessage::TransactionUpdate(_) => Some(WorkloadType::Update),
+                ws_v2::ServerMessage::OneOffQueryResult(_) => Some(WorkloadType::Sql),
+                ws_v2::ServerMessage::ReducerResult(_) => Some(WorkloadType::Reducer),
+                ws_v2::ServerMessage::ProcedureResult(_) => Some(WorkloadType::Procedure),
+            },
```
Centril, line 291: organization-wise it would've been nice to also just have 1 function for v2

crates/core/src/subscription/module_subscription_manager.rs
```diff
@@ -195,6 +217,7 @@ struct QueryState {
     legacy_subscribers: HashSet<ClientId>,
     // For clients that subscribe to a single query, we track them here.
     subscriptions: HashSet<ClientId>,
+    v2_subscriptions: HashSet<SubscriptionIdV2>,
```
Centril, line 221: doc comment please, in particular explaining the from-v1-structure-difference

crates/core/src/subscription/module_subscription_manager.rs
```diff
@@ -516,7 +540,18 @@ struct ClientUpdate {
...
+    update:
+        ws_v1::FormatSwitch<ws_v1::SingleQueryUpdate<ws_v1::BsatnFormat>, ws_v1::SingleQueryUpdate<ws_v1::JsonFormat>>,
+}
+
+/// A single update for one client and one query.
+#[derive(Debug)]
+struct V2ClientUpdate {
+    id: ClientId,
+    query_set_id: ClientQuerySetId,
+    //table_update: ws_v2::TableUpdate,
```
Centril: remove?

crates/core/src/subscription/module_subscription_manager.rs
```diff
@@ -848,11 +887,134 @@ impl SubscriptionManager {
         Ok(queries_to_return)
     }
 
+    pub fn remove_subscription_v2(
```
Centril, line 890: Looks mostly duplicated from above. After 2.0 this should be deduped. This code is too important and load bearing for allowing accidental differences to creep in over time.

crates/core/src/subscription/module_subscription_manager.rs
```diff
@@ -848,11 +887,134 @@ impl SubscriptionManager {
...
+
+        Ok(queries_to_return)
+    }
+
     /// Adds a single subscription for a client.
     pub fn add_subscription(&mut self, client: Client, query: Query, query_id: ClientQueryId) -> Result<(), DBError> {
         self.add_subscription_multi(client, vec![query], query_id).map(|_| ())
     }
 
+    pub fn add_subscription_v2(
```
Centril, line 954: ditto compared to `add_subscription_multi`.

crates/cli/src/subcommands/subscribe.rs
```diff
@@ -71,16 +71,16 @@ pub fn cli() -> clap::Command {
         .arg(common_args::server().help("The nickname, host name or URL of the server hosting the database"))
 }
 
-fn parse_msg_json(msg: &WsMessage) -> Option<ws::ServerMessage<JsonFormat>> {
+fn parse_msg_json(msg: &WsMessage) -> Option<ws_v1::ServerMessage<ws_v1::JsonFormat>> {
```
cloutiertyler, line 74: Is this to retain the json format in the API?

crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -853,9 +1093,54 @@ impl ModuleSubscriptions {
...
+    ) -> Result<Option<ExecutionMetrics>, DBError> {
+        match host {
+            Some(host) => host.call_view_add_v2_subscription(sender, auth, request, timer).await,
+            None => panic!("v2 subscriptions without a module host are not supported yet"),
+            // self
+            //     .add_multi_subscription_inner::<host::wasmtime::WasmtimeInstance>(
+            //         None, sender, auth, request, timer, _assert,
+            //     )
+            //     .map(|(metrics, _)| metrics),
+        }
```
Centril, line 1133: ?

crates/cli/src/subcommands/subscribe.rs
cloutiertyler: Disregard, I thought I was reading TypeScript.

crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -1196,21 +1581,61 @@ impl ModuleSubscriptions {
...
+                                    request_id,
+                                    timestamp: event.timestamp,
+                                    result,
+                                };
+                                let _ = self.broadcast_queue.send_client_message_v2(
+                                    client,
+                                    None, // This should arguably have a tx_offset, but it shouldn't matter as long as we wait for previous messages to be sent.
+                                    message,
+                                );
+                            }
```
Centril, line 1601: Let's move this to its own function

crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -1231,12 +1656,26 @@ impl ModuleSubscriptions {
...
+                let _wait_guard = subscription_metrics.lock_waiters.inc_scope();
+                let _wait_timer = subscription_metrics.lock_wait_time.start_timer();
+                self.subscriptions.write()
+            };
+            for (client_id, query_set_id) in failed_v2_subscriptions {
+                if let Err(err) = subscriptions.remove_subscription_v2(client_id, query_set_id) {
+                    tracing::warn!(?client_id, ?query_set_id, "failed to remove v2 subscription: {err}");
+                }
+            }
+        }
```
Centril, line 1645: Could use some commentary.

crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -199,8 +207,80 @@ pub(crate) fn commit_and_broadcast_event(
...
+                let rows = if use_deletes {
+                    query_update.deletes
+                } else {
+                    query_update.inserts
+                };
+                inserts.push(rows);
+            }
+            Ok(ws_v2::SingleTableRows {
+                table: table_update.table_name,
+                rows: merge_bsatn_row_lists(inserts),
```
Centril: Why do we do this? The v1 protocol didn't merge these intentionally to avoid having the server do this (expensive) work, which is now being done in v2 while holding the transaction open.

crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -199,8 +207,80 @@ pub(crate) fn commit_and_broadcast_event(
...
+            }
+            ws_v1::RowSizeHint::RowOffsets(offs) => {
+                offsets.extend(offs.iter().map(|o| (base + *o as usize) as u64));
+            }
+        }
+        base += data_len;
+    }
+
+    ws_v1::BsatnRowList::new(ws_v1::RowSizeHint::RowOffsets(offsets.into()), rows_data.freeze())
+}
```
Centril: If we keep this merging logic, it should ideally live in `websocket_building.rs`.

crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -199,8 +207,80 @@ pub(crate) fn commit_and_broadcast_event(
...
+        rows_data.extend_from_slice(&bytes);
+
+        match hint {
+            ws_v1::RowSizeHint::FixedSize(size) => {
+                let size = size as usize;
+                let count = data_len / size;
+                for i in 0..count {
+                    offsets.push((base + i * size) as u64);
+                }
+            }
```
Centril: The function should preserve `FixedSize` if all are equal.

crates/client-api/src/routes/subscribe.rs
jsdt: For v2 we should have sent normal errors earlier (for things like reducers that returned errors), so if any error bubbles up this far, we just send a CloseFrame with the error message.

crates/client-api/src/routes/subscribe.rs
jsdt: Related to the other comment, but we shouldn't send these messages to v2 clients. I'll update the warning message to indicate that we shouldn't hit this case.

crates/client-api-messages/src/websocket/v2.rs
jsdt: We could have, but since we already have branches using `InitialConnection`, it doesn't seem worthwhile to combine.

crates/client-api-messages/src/websocket/v2.rs
jsdt: Oof

crates/core/src/subscription/module_subscription_actor.rs
jsdt: This is a code path that only gets called for unit tests. We need to come up with a way to test the v2 path too, but it won't be in this PR.

crates/core/src/subscription/module_subscription_manager.rs
jsdt: It depends on how long we are keeping v1 support, and how much we intend to make changes to the v1 code paths. If we aren't planning to change the v1 code until we delete it, it can be easier to leave it as is rather than have v1-specific logic in the main code paths.

crates/core/src/client/messages.rs
jsdt: I implemented it for the v2 messages.

crates/core/src/client/messages.rs
jsdt: The optimization of reusing buffers has been added for v2. Unless we are going to add other optimization for the v1 code (which seems unlikely), I don't think we need to make this a generic function.

crates/core/src/subscription/module_subscription_manager.rs
Centril: It's likely that we'll keep v1 around for a year or so. In that timeframe, my concern is that we'll want to add some feature to v2 or do some refactoring which we are forced to do to v1 as well.

crates/client-api/src/routes/subscribe.rs
```diff
@@ -1310,36 +1310,33 @@ async fn ws_encode_task(
...
-                            let (stats, in_use, mut frames) =
-                                ws_encode_message_v2(config, buf, server_message, false, &bsatn_rlb_pool).await;
-                            metrics.report(workload, num_rows, stats);
-                            if frames.try_for_each(|frame| outgoing_frames.send(frame)).is_err() {
-                                break;
-                            }
+            OutboundWsMessage::Message(message) => match message {
+                OutboundMessage::V2(server_message) => {
+                    if config.version != WsVersion::V2 {
+                        log::warn!("dropping v2 message on v1 connection");
```
Centril: Since this is impossible, this should probably be `log::error!(...)`?

crates/client-api/src/routes/subscribe.rs
```diff
@@ -1310,36 +1310,33 @@ async fn ws_encode_task(
...
+                }
+                OutboundMessage::V1(message) => {
+                    if config.version == WsVersion::V2 {
+                        log::warn!(
+                            "dropping v1 message for v2 connection until v2 serialization is implemented: {:?}",
+                            message
+                        );
                         continue;
-                    };
+                    }
```
Centril: Ditto... also impossible... change to `error!`?

PR #4217, opened 2026-02-06
Implement event tables (server, Rust/TS/C# codegen + client SDKs)

cloutiertyler:
## Summary

Implements event tables end-to-end: server datastore, module bindings (Rust/TypeScript/C#), client codegen (Rust/TypeScript/C#), client SDKs (Rust/TypeScript/C#), and integration tests.

Event tables are tables whose rows are ephemeral — they persist to the commitlog and are delivered to V2 subscribers, but are NOT merged into committed state. Rows are only visible within the transaction that inserted them. This is the mechanism that replaces reducer event callbacks in 2.0.

### Server
- `is_event` flag on `RawTableDefV10`, `TableDef`, `TableSchema`
- Event table rows recorded in TxData but skipped during committed state merge
- Commitlog replay treats event table inserts as no-ops
- Migration validation rejects changing `is_event` between module versions
- `SELECT * FROM *` excludes event tables
- V1 WebSocket subscriptions to event tables rejected with upgrade message
- V2 subscription path delivers event table rows correctly
- `CanBeLookupTable` trait — event tables cannot be lookup tables in semijoins
- Runtime view validation rejects event tables

### Module bindings
- **Rust**: `#[spacetimedb::table(name = my_events, public, event)]`
- **TypeScript**: `table({ event: true }, ...)`
- **C#**: `[Table(Event = true)]`

### Client codegen (`crates/codegen/`)
- **Rust**: Generates `EventTable` impl (insert-only) for event tables, `Table` impl for normal tables. `CanBeLookupTable` emitted for non-event tables.
- **TypeScript**: Emits `event: true` in generated table schemas. `ClientTableCore` type excludes `onDelete`/`onUpdate` for event tables via conditional types.
- **C#**: Generates classes inheriting from `RemoteEventTableHandle` (which hides `OnDelete`/`OnBeforeDelete`/`OnUpdate`) for event tables.

### Client SDKs
- **Rust**: `EventTable` trait with insert-only callbacks, client cache bypass, `count()` returns 0, `iter()` returns empty
- **TypeScript**: Event table cache bypass in `table_cache.ts` — fires `onInsert` callbacks but doesn't store rows. Type-level narrowing excludes delete/update methods.
- **C#**: `RemoteEventTableHandle` base class hides delete/update events. Parse/Apply/PostApply handle `EventTableRows` wire format, skip cache storage, fire only `OnInsert`.

### Tests
- 9 datastore unit tests (insert/delete/update semantics, replay, constraints, indexes, auto-inc, cross-tx reset)
- 3 Rust SDK integration tests (basic events, multiple events per reducer, no persistence across transactions)
- Codegen snapshot tests (Rust, TypeScript, C#)
- Trybuild compile tests (event tables rejected as semijoin lookup tables)

## Deferred
- `on_delete` codegen for event tables (server only sends inserts; client synthesis deferred)
- Event tables in subscription joins / views (well-defined but restricted for now)
- C++ SDK support
- RLS integration test

## API and ABI breaking changes

- `is_event: bool` added to `RawTableDefV10` (appended, defaults to `false` — existing modules unaffected)
- `CanBeLookupTable` trait bound on semijoin methods in query builder (all non-event tables implement it, so existing code compiles unchanged)
- `RemoteEventTableHandle` added to C# SDK (new base class for generated event table handles)

3 — Changes touch the schema pipeline end-to-end and all three client SDKs, but each individual change is straightforward. The core risk area is the committed state merge skip in `committed_state.rs`. Client SDK changes are additive (new code paths for event tables, existing paths unchanged).

## Testing

- [x] `cargo clippy --workspace --tests --benches` passes
- [x] `cargo test -p spacetimedb-codegen` (snapshot tests)
- [x] `cargo test -p spacetimedb-datastore --features spacetimedb-schema/test -- event_table` (9 unit tests)
- [x] `pnpm format` passes
- [x] Rust SDK integration tests pass (`event_table_tests` module)

### DIFF
--- a/crates/codegen/src/rust.rs
+++ b/crates/codegen/src/rust.rs
@@ -212,8 +212,7 @@ impl<'ctx> __sdk::EventTable for {table_handle}<'ctx> {{
             // and emit accessors for unique columns.
             write!(
                 out,
-                "
-pub struct {delete_callback_id}(__sdk::CallbackId);
+                "pub struct {delete_callback_id}(__sdk::CallbackId);
 
 impl<'ctx> __sdk::Table for {table_handle}<'ctx> {{
     type Row = {row_type};
--- a/crates/codegen/tests/snapshots/codegen__codegen_rust.snap
+++ b/crates/codegen/tests/snapshots/codegen__codegen_rust.snap
@@ -831,7 +831,6 @@ impl LoggedOutPlayerTableAccess for super::RemoteTables {
 }
 
 pub struct LoggedOutPlayerInsertCallbackId(__sdk::CallbackId);
-
 pub struct LoggedOutPlayerDeleteCallbackId(__sdk::CallbackId);
 
 impl<'ctx> __sdk::Table for LoggedOutPlayerTableHandle<'ctx> {
@@ -2046,7 +2045,6 @@ impl MyPlayerTableAccess for super::RemoteTables {
 }
 
 pub struct MyPlayerInsertCallbackId(__sdk::CallbackId);
-
 pub struct MyPlayerDeleteCallbackId(__sdk::CallbackId);
 
 impl<'ctx> __sdk::Table for MyPlayerTableHandle<'ctx> {
@@ -2224,7 +2222,6 @@ impl PersonTableAccess for super::RemoteTables {
 }
 
 pub struct PersonInsertCallbackId(__sdk::CallbackId);
-
 pub struct PersonDeleteCallbackId(__sdk::CallbackId);
 
 impl<'ctx> __sdk::Table for PersonTableHandle<'ctx> {
@@ -2527,7 +2524,6 @@ impl PlayerTableAccess for super::RemoteTables {
 }
 
 pub struct PlayerInsertCallbackId(__sdk::CallbackId);
-
 pub struct PlayerDeleteCallbackId(__sdk::CallbackId);
 
 impl<'ctx> __sdk::Table for PlayerTableHandle<'ctx> {
@@ -3498,7 +3494,6 @@ impl TestDTableAccess for super::RemoteTables {
 }
 
 pub struct TestDInsertCallbackId(__sdk::CallbackId);
-
 pub struct TestDDeleteCallbackId(__sdk::CallbackId);
 
 impl<'ctx> __sdk::Table for TestDTableHandle<'ctx> {
@@ -3745,7 +3740,6 @@ impl TestFTableAccess for super::RemoteTables {
 }
 
 pub struct TestFInsertCallbackId(__sdk::CallbackId);
-
 pub struct TestFDeleteCallbackId(__sdk::CallbackId);
 
 impl<'ctx> __sdk::Table for TestFTableHandle<'ctx> {


crates/datastore/src/locking_tx_datastore/committed_state.rs
```diff
@@ -1187,6 +1193,22 @@ impl CommittedState {
         //             and the fullness of the page.
 
         for (table_id, tx_table) in insert_tables {
+            // Event tables: record in TxData for commitlog persistence and subscription dispatch,
+            // but do NOT merge into committed state. Their rows are ephemeral in memory.
```
cloutiertyler:
```suggestion
            // but do NOT merge into committed state.
```

sdks/rust/src/table.rs
```diff
@@ -7,7 +7,46 @@
...
+    type Row: 'static;
+
+    /// The `EventContext` type generated for the module which defines this table.
+    type EventContext;
+
+    /// The number of subscribed rows in the client cache (always 0 for event tables).
+    fn count(&self) -> u64;
+
+    /// An iterator over all the subscribed rows in the client cache (always empty for event tables).
+    fn iter(&self) -> impl Iterator<Item = Self::Row> + '_;
```
gefjon: If these always return nothing, why define them?

docs/event-tables-status.md
```diff
@@ -0,0 +1,124 @@
+# Event Tables Implementation Status
+
+Tracking progress against the [event tables proposal](../../SpacetimeDBPrivate/proposals/00XX-event-tables.md).
```
gefjon: This probably shouldn't be in here.

Centril:
A few things:
- It would be good to review and land the datastore changes first.
- When doing so, let's add unit tests (in the datastore crate) asserting that:
   a) an insert/delete pair does nothing to committed state (check this by querying txn +1) and neither to `TxData`.
   b) an update does nothing to committed state and neither to `TxData`.
   c) an insertion to an event table results in no changes to committed state and 1 change to `TxData`.
   d) replaying should also be tested as a unit test by saving the commitlog and then replaying and checking committed state after.
- I didn't see any code handling migrations from/to event tables. Ostensibly, such transitions should not be allowed. If this is already the case, a test confirming so would be good.

docs/event-tables-status.md
cloutiertyler: It's just for me while I'm building so that claude doesn't get confused. It will be removed before I'm done.

sdks/rust/src/table.rs
cloutiertyler: Well the idea was that they are actual tables, so it might be confusing why they are missing. Then again it might be confusing why they never return rows. I'm open to either direction.

sdks/csharp/src/Table.cs
```diff
@@ -597,5 +633,24 @@ public void Invoke(EventContext ctx, Row oldRow, Row newRow)
...
+        where Row : class, IStructuralReadWrite, new()
+    {
+        protected RemoteEventTableHandle(IDbConnection conn) : base(conn, isEventTable: true) { }
+
+        // Hide delete/update events from the public API.
+        // The base class handlers still exist but will never have listeners registered,
+        // so invoking them in PostApply is a harmless no-op.
+        private new event RowEventHandler OnDelete { add { } remove { } }
+        private new event RowEventHandler OnBeforeDelete { add { } remove { } }
+        private new event UpdateEventHandler OnUpdate { add { } remove { } }
```
JasonAtClockwork:
```suggestion
        [Obsolete("OnDelete is not supported for event tables. This callback is never invoked.", error: false)]
        public new event RowEventHandler OnDelete { add { } remove { } }
        [Obsolete("OnBeforeDelete is not supported for event tables. This callback is never invoked.", error: false)]
        public new event RowEventHandler OnBeforeDelete { add { } remove { } }
        [Obsolete("OnUpdate is not supported for event tables. This callback is never invoked.", error: false)]
        public new event UpdateEventHandler OnUpdate { add { } remove { } }
```

Using private new events to hide the base events will not prevent binding in current usage. If we switch these to public new + [Obsolete], we can provide guidance/warnings that they are no-ops. 

Actual enforcement would require a heavier change, e.g. splitting into RemoteTableHandleBase<>, RemotePersistentTableHandle<> (with delete/update events), and RemoteEventTableHandle<> (without those events).

cloutiertyler: Which of those two options do you recommend? For all other similar cases we opted to not generate at all and split things up.

JasonAtClockwork: @cloutiertyler  @rekhoff - I've gone with another approach and split up `RemoteTableHandle` into `RemoteTableHandleBase` which no longer has the `OnDelete/OnBeforeDelete/OnUpdate` and have them in `RemoteTableHandle` based off `RemoteTableHandleBase` allowing for `RemoteEventTableHandle` which doesn't have the events accessible.

I tried to make this as low impact as possible while still removing the events to have parity with the other languages.

rekhoff: Splitting the functionality between a `HandleBase` and separate implementations between the `TableHandle` and the `TableEventHandle` makes a lot of sense to me. I have no objections to this approch

cloutiertyler: That works for me.

crates/physical-plan/src/plan.rs
```diff
@@ -1150,6 +1166,14 @@ impl PhysicalPlan {
             _ => false,
         })
     }
+
+    /// Does this plan use an event table as the lookup (rhs) table in a semi-join?
```
joshua-spacetime, line 1170:
```suggestion
    /// Does this plan use an event table as the lookup (rhs) table in a semi-join?
    ///
    /// Note, we only care about index joins because this method is only relevant for subscriptions,
    /// and index joins are the only type of join allowed in subscriptions.
```

crates/codegen/src/rust.rs
```diff
@@ -169,6 +169,51 @@ impl {accessor_trait} for super::RemoteTables {{
...
+            // so no reason to have them.
+            write!(
+                out,
+                "
+impl<'ctx> __sdk::EventTable for {table_handle}<'ctx> {{
+    type Row = {row_type};
+    type EventContext = super::EventContext;
+
+    fn count(&self) -> u64 {{ self.imp.count() }}
+    fn iter(&self) -> impl Iterator<Item = {row_type}> + '_ {{ self.imp.iter() }}
```
cloutiertyler, line 192: I've elected to keep these at least for now. I would rather remove them but we are out of time, but also we use these in tests to ensure that indeed there are no rows. CC @gefjon

PR #4231, opened 2026-02-09
CI - Use new internal test inputs

bfops:
# Description of Changes

Flipping on some new inputs to `Internal Tests` to get new functionality.

I'll follow-up with a more detailed description in discord.

None. CI only.

2

# Testing
- [x] public PR without private PR just uses master
- [x] public PR with private PR uses that one
  - [x] fails if private PR is not approved
  - [x] fails if private PR does not pass its CI
  - [x] passes if private PR is ready to go


jdetter (approved):
Seems fine, before merging could you write down a short summary of what this is flipping on in the PR description?

PR #4233, opened 2026-02-09
TypeScript modules: Expose hash indices

Centril:
# Description of Changes

This allows `.index('hash')`.

None

1

# Testing

Verified that it works using the keynote 2 benchmark.


crates/bindings-typescript/src/lib/indexes.ts
```diff
@@ -47,14 +48,19 @@ export type Indexes<
 };
 
 /**
- * A type representing a database index, which can be either unique or ranged.
+ * A type representing a database index,
+ * which can be either unique, pointed or ranged.
```
cloutiertyler:
```suggestion
 * which can either be unique or filter for a single value or range of values.
```

cloutiertyler:
Only reason this is marked `release-2.0` is it is used in the keynote benchmark

PR #4241, opened 2026-02-09
[2.0 Breaking] Add --include-private and default private tables to not generate

JasonAtClockwork:
# Description of Changes
Updated the codegen table/function iteration functions to take in a parameter to check visibility in all locations for the supported languages. 
- Updated the util.rs functions for iterating tables/functions to check for a CodegenVisibility enum (IncludePrivate, or OnlyPublic)
- Added a new CodegenOptions struct to pass around the CodegenVisibility and future flags, defaulted visibility to OnlyPublic
- Updated the CLI to return a list of all private tables not included (added a TODO to check the --include-private opt):
```bash
Build finished successfully.
Skipping private tables during codegen: secret_note, secret_order, secret_person.
```

Technically API breaking as the private tables will no longer be available. (GitHub labels are not working at the moment)

1 - Simple change the testing took longer

# Testing

Turns out when you remove private tables you invalidate most of the module_bindings across the system!

- [x] Rust test SDK for all languages
- [x] C# SDK tests
- [x] C# dotnet tests
- [x] Updated and checked snap files
- [x] Updated Blackholio (Unreal + Unity) module_bindings and tested
- [x] Ran Unreal SDK tests


gefjon (approved):

crates/cli/src/subcommands/generate.rs
```diff
@@ -109,6 +112,12 @@ pub fn cli() -> clap::Command {
                 .default_value("")
                 .help("Options to pass to the build command, for example --build-options='--lint-dir='"),
         )
+        .arg(
+            Arg::new("include_private")
+                .long("include-private")
+                .action(SetTrue)
+                .help("Include private tables and functions in generated code (types are always included)."),
```
cloutiertyler, line 120: Is default false implied? Can we make this explicit?

cloutiertyler (approved):

crates/cli/src/subcommands/generate.rs
gefjon: Default false is implied by `action(SetTrue)`.

JasonAtClockwork: Added explicit default value

PR #4247, opened 2026-02-10
Add warning prompt for 1.0 -> 2.0 module upgrade path

joshua-spacetime:
# Description of Changes

Adds a warning prompt for 1.0 -> 2.0 module upgrade path.

None

1

# Testing

This patch checks a wasm module binary compiled with pre-2.0 bindings into source control. A smoketest was added that first publishes the the pre-compiled module and then publishes a new module using the 2.0 bindings in its place.

### DIFF
--- a/docs/docs/00200-core-concepts/00100-databases/00500-migrations/00100-upgrade-notes-2-0.md
+++ b/docs/docs/00200-core-concepts/00100-databases/00500-migrations/00100-upgrade-notes-2-0.md
@@ -1,0 +1,12 @@
+---
+title: 1.x to 2.0 Upgrade Notes
+slug: /upgrade
+---
+
+Upgrade notes for 1.x to 2.0 are coming soon.
+
+This page is intentionally a placeholder and will be filled in with:
+
+- Module API breaking changes
+- Client API breaking changes
+- Migration guidance and rollout checklist
--- a/docs/docs/00200-core-concepts/00100-databases.md
+++ b/docs/docs/00200-core-concepts/00100-databases.md
@@ -95,6 +95,7 @@ See [`spacetime publish`](/databases/building-publishing) for details on the pub
 
 When you republish to an existing database, SpacetimeDB attempts to automatically migrate the schema. For details on what changes are supported and migration strategies:
 
+- [1.x to 2.0 Upgrade Notes](/upgrade) - Required reading before major-version upgrades.
 - [Automatic Migrations](/databases/automatic-migrations) - Learn which schema changes are safe, breaking, or forbidden.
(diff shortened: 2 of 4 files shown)


crates/cli/src/subcommands/publish.rs
```diff
@@ -134,6 +134,27 @@ fn confirm_and_clear(
     Ok(builder)
 }
 
+fn confirm_major_version_upgrade() -> Result<(), anyhow::Error> {
+    println!(
+        "It looks like you're trying to do a major version upgrade from 1.0 to 2.0. We recommend first looking at the upgrade notes before committing to this upgrade: https://spacetimedb.com/docs/upgrade"
```
joshua-spacetime, line 139: This is not a valid URL currently.

crates/smoketests/fixtures/README.md
```diff
@@ -0,0 +1,46 @@
+# Smoketest Fixtures
```
jdetter, line 1: We don't necessarily need to change this, but why include the wasm blob here instead of just including the project source code and building it during the test?

docs/docs/00200-core-concepts/00100-databases/00500-migrations/00100-upgrade-notes-2-0.md
```diff
@@ -0,0 +1,12 @@
...
+slug: /upgrade
+---
+
+Upgrade notes for 1.x to 2.0 are coming soon.
+
+This page is intentionally a placeholder and will be filled in with:
+
+- Module API breaking changes
+- Client API breaking changes
+- Migration guidance and rollout checklist
```
jdetter, line 12: Is there a ticket in the core board to complete this page?

crates/smoketests/fixtures/README.md
jdetter: @bfops might have some thoughts here

joshua-spacetime: This seemed like a lighter weight test. We'd have to build it with a pre-2.0 version of spacetimedb otherwise.

docs/docs/00200-core-concepts/00100-databases/00500-migrations/00100-upgrade-notes-2-0.md
joshua-spacetime: Yes there is.

crates/smoketests/fixtures/README.md
bfops: Yeah this seems good to me since the goal isn't to ever update the wasm

bfops (approved):
This LGTM. The test infrastructure itself generally needs a refactor, since it seems like we have a wall of very similar functions, but that's not the fault or purview of this PR.

PR #4250, opened 2026-02-10
Sort version output of `spacetime version list`

aasoni:
# Description of Changes

The output of `spacetime version list` is sorted in alphabetical order, which is not the correct semver order.
This fixes the order printed, by sorting correctly.

Example before:
```
1.11.0
1.9.0
1.3.0
1.3.1
1.1.2
1.12.0 (current)
1.0.0
```

After:
```
1.12.0 (current)
1.11.0
1.9.0
1.3.1
1.3.0
1.1.2
1.0.0
```

Not an API or ABI breaking change

1

# Testing

Ran `spacetime version list` and `spacetime version list --all` with the new code and confirmed that the result was correctly sorted.
I additionally tested that if the latest available version was not installed that it would show with the upgrade hint command.

### DIFF
--- a/crates/update/src/cli/list.rs
+++ b/crates/update/src/cli/list.rs
@@ -61,7 +61,7 @@ impl List {
             let is_current = Some(ver) == current.as_ref();
 
             if is_current {
-                println!("{} {}", ver, "(current)");
+                println!("{} (current)", ver);
             } else {
                 println!("{ver}");
             }


crates/update/src/cli/list.rs
```diff
@@ -36,19 +36,74 @@ impl List {
...
-                print!(" (current)");
+            match semver::Version::parse(&ver) {
+                Ok(sv) => parsed.push((sv, ver)),
+                Err(_) => unparsed.push(ver),
             }
-            println!();
         }
+        parsed.sort_by(|(a, _), (b, _)| b.cmp(a));
+
+        let sorted_versions: Vec<String> = parsed.into_iter().map(|(_, s)| s).chain(unparsed).collect();
```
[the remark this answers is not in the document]
aasoni: Updated the comment

crates/update/src/cli/list.rs
```diff
@@ -36,19 +36,74 @@ impl List {
...
+            })
+        } else {
+            None
+        };
+
+        // Determine whether the latest version is newer than any installed version.
+        let newest_installed = sorted_versions.first().and_then(|v| semver::Version::parse(v).ok());
+
+        let show_latest = match (&latest, &newest_installed) {
+            (Some(lat), Some(cur)) if lat > cur => Some(lat.to_string()),
```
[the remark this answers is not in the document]
aasoni: The PR description is inaccurate in this case. The hint should only show if there is a newer version than any installed version. This is because all installed versions are listed in `spacetime version list` and therefore discoverable. The main point of showing that there is a new version is that it is otherwise unknown to the user that a new version is available unless they run `spacetime upgrade`.

crates/update/src/cli/install.rs
```diff
@@ -152,6 +152,18 @@ impl ArtifactType {
...
+    let url = format!("{}/latest", releases_url());
+    let release: Release = client
+        .get(url)
+        .send()
+        .await?
+        .error_for_status()?
+        .json()
+        .await?;
+    release.version().context("Could not parse latest version number")
+}
```
bfops: nit: instead of adding this function, I think we could instead factor this out of `download_and_install` instead:
```rust
pub(super) async fn fetch_release(
    client: &reqwest::Client,
    version: Option<semver::Version>,
) -> anyhow::Result<Release> {
    let releases_url = releases_url();
    let url = match &version {
        Some(version) => format!("{releases_url}/tags/v{version}"),
        None => [&*releases_url, "/latest"].concat(),
    };
    let release = client
        .get(url)
        .send()
        .await?
        .error_for_status()
        .map_err(|e| {
            if e.status() == Some(reqwest::StatusCode::NOT_FOUND) {
                if let Some(version) = &version {
                    return anyhow::anyhow!(e).context(format!("No release found for version {version}"));
                }
            }
            anyhow::anyhow!(e).context("Could not fetch release info")
        })?
        .json()
        .await?;
    Ok(release)
}
```

crates/update/src/cli/list.rs
```diff
@@ -36,19 +36,74 @@ impl List {
...
+
         let versions = if self.all {
             let client = super::reqwest_client()?;
             super::tokio_block_on(super::install::available_releases(&client))??
         } else {
             paths.cli_bin_dir.installed_versions()?
         };
+
+        // Sort versions using semver ordering. Versions that fail to parse are
+        // placed at the end in their original listed.
```
bfops: maybe a dumb question, but why would a version fail to parse?

crates/update/src/cli/list.rs
```diff
@@ -36,19 +36,74 @@ impl List {
...
+                Err(_) => unparsed.push(ver),
             }
-            println!();
         }
+        parsed.sort_by(|(a, _), (b, _)| b.cmp(a));
+
+        let sorted_versions: Vec<String> = parsed.into_iter().map(|(_, s)| s).chain(unparsed).collect();
+
+        // Check for a newer version available on GitHub.
+        let latest = if !self.all {
```
bfops: why only in the `!self.all` case? I would expect this hint to happen regardless of whether we run with `--all`?

crates/update/src/cli/list.rs
```diff
@@ -36,19 +36,74 @@ impl List {
...
+            })
+        } else {
+            None
+        };
+
+        // Determine whether the latest version is newer than any installed version.
+        let newest_installed = sorted_versions.first().and_then(|v| semver::Version::parse(v).ok());
+
+        let show_latest = match (&latest, &newest_installed) {
+            (Some(lat), Some(cur)) if lat > cur => Some(lat.to_string()),
```
bfops: I think if the current version doesn't parse, we should probably the hint to upgrade them to the latest version.

crates/update/src/cli/list.rs
bfops: ah because it will be self-evident in that case?

crates/update/src/cli/list.rs
```diff
@@ -36,19 +36,74 @@ impl List {
...
+        for ver in &sorted_versions {
+            let is_current = Some(ver) == current.as_ref();
+
+            // If this is the current version and there's a newer version
+            // available, print the newer version first.
+            if is_current {
+                if let Some(ref new_ver) = show_latest {
+                    println!("{} {}", new_ver, "(available - run `spacetime version upgrade`)");
+                }
+            }
```
bfops: this does seem weird if the current version isn't the latest

bfops: how about instead of trying to show the latest version "inline", we just print a message at the top that says something like "version xyz is available, run `spacetime version upgrade` to update!" and then print the list as normal?

crates/update/src/cli/list.rs
aasoni: They shouldn't, this is just to handle something like a typo in a version in which case we may still want to show it. But maybe we just don't show it if it fails to parse. Open to either

crates/update/src/cli/list.rs
aasoni: Yeah good point, my thought was that --all lists the newer version anyway so we don't need to do anything there to make them discoverable. But the hint would be nice there too

crates/update/src/cli/list.rs
aasoni: true

PR #4251, opened 2026-02-10
Event tables: datastore tests, migration validation, and bootstrap fix

cloutiertyler:
# Description of Changes

Cherry-picks the core event table implementation (`fa83d6d40`, `07d57a552`) and adds:

1. **Bootstrap fix**: Register `st_event_table` in `bootstrap_system_tables()` — it was missing from the cherry-picked commit, causing event table creation to fail at runtime when inserting into `ST_EVENT_TABLE_ID`.

2. **4 datastore unit tests** covering event table behaviors:
   - Insert+delete in the same tx cancels out (no TxData entry, no committed state)
   - Update yields only the final row in TxData (old row replaced, not accumulated)
   - Bare insert records in TxData but not in committed state
   - `replay_insert` is a no-op for event tables (commitlog replay doesn't rebuild state)

3. **Migration validation**: `ChangeTableEventFlag` error in `auto_migrate.rs` prevents changing a table's event flag (event→non-event or vice versa) via auto-migration. Includes tests for both directions and a positive test that identical flags are accepted.

4. **Test fixture updates**: Bootstrap test expectations updated for the new `st_event_table` system table (table row, column, index ID 23, constraint ID 19). Enabled `spacetimedb-schema/test` feature in datastore dev-deps to fix pre-existing test compilation issue with `for_test` methods.

None. The `ChangeTableEventFlag` error variant is added to the internal `AutoMigrateError` enum but this is not a public API.

1 — Straightforward additions. The bootstrap fix is a one-liner that was simply missing from the cherry-pick. The tests exercise existing behavior without changing it. The migration guard follows the exact same pattern as the existing `ChangeTableType` check.

# Testing

- [x] `cargo test -p spacetimedb-datastore` — 77 tests pass (73 existing + 4 new)
- [x] `cargo test -p spacetimedb-schema --features test --lib` — 98 tests pass (96 existing + 2 new)
- [x] `cargo check --workspace --exclude view-client` — compiles clean (`view-client` has a pre-existing issue with `CanBeLookupTable` from the cherry-pick)
- [ ] Reviewer: confirm that the 4 event table test names/assertions match the behaviors described in review feedback

### DIFF
--- a/crates/datastore/src/locking_tx_datastore/state_view.rs
+++ b/crates/datastore/src/locking_tx_datastore/state_view.rs
@@ -160,10 +160,19 @@ pub trait StateView {
             .transpose()?;
 
         // Check if this table is an event table by looking up `st_event_table`.
-        let is_event = self
-            .iter_by_col_eq(ST_EVENT_TABLE_ID, StEventTableFields::TableId, value_eq)?
-            .next()
-            .is_some();
+        //
+        // When restoring a pre-event-tables snapshot, `st_event_table` won't exist yet:
+        //   1. `restore_from_snapshot` restores pages and calls `schema_for_table_raw` (here)
+        //   2. `apply_history` replays the commit log
+        //   3. `migrate_system_tables` creates any missing system tables, including `st_event_table`
+        //
+        // We're in step 1, so `st_event_table` may not exist. Default to `false` in that case.
+        // After step 3 it will be a proper (empty) system table.
+        let is_event = match self.iter_by_col_eq(ST_EVENT_TABLE_ID, StEventTableFields::TableId, value_eq) {
+            Ok(mut iter) => iter.next().is_some(),
+            Err(DatastoreError::Table(TableError::IdNotFound(..))) => false,
+            Err(e) => return Err(e),
+        };
         Ok(TableSchema::new(
             table_id,
             table_name,
--- a/crates/core/src/db/relational_db.rs
+++ b/crates/core/src/db/relational_db.rs
@@ -362,6 +362,16 @@ impl RelationalDB {
         None
     }
 
+    /// Create any system tables that are missing from the datastore.
+    ///
+    /// This runs during database startup, *after* `restore_from_snapshot` and `apply_history`
+    /// have reconstructed the committed state from persisted data. Older snapshots may predate
+    /// the introduction of newer system tables (e.g. `st_event_table`, `st_view`), so those
+    /// tables won't exist in the restored state. This method fills them in.
+    ///
+    /// Because of this ordering, code that runs during snapshot restoration
+    /// (such as `schema_for_table_raw`) must tolerate missing system tables gracefully
+    /// — they will be created here shortly after.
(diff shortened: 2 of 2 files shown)


crates/datastore/Cargo.toml
```diff
@@ -47,6 +47,7 @@ test = ["spacetimedb-commitlog/test"]
 spacetimedb-lib = { path = "../lib", features = ["proptest"] }
 spacetimedb-sats = { path = "../sats", features = ["proptest"] }
 spacetimedb-commitlog = { path = "../commitlog", features = ["test"] }
+spacetimedb-schema = { path = "../schema", features = ["test"] }
```
cloutiertyler: > [dev-dependencies]                                                                                                                                                                                                 
>   spacetimedb-schema = { path = "../schema", features = ["test"] }
>                                                                                                                                                                                                                      
>   This enables the test feature of spacetimedb-schema when running datastore tests. The spacetimedb-schema crate is already a regular dependency of the datastore crate — my change just adds it as a dev-dependency
>   with the test feature flag enabled.
> 
>   The test feature gates for_test() constructors on types like TableName, Identifier, and ColumnSchema. The datastore tests use these constructors extensively (e.g., TableName::for_test("Foo") in TableRow
>   conversions), but the feature wasn't being enabled in test builds — so the tests were already broken on master before our changes.
> 
>   In short: the datastore already depends on schema. I'm just enabling the test feature so the existing test helpers compile.

I am not sure why claude is insistent on this. Could someone with knowledge verify that this is true and necessary?

crates/bindings-macro/src/table.rs
```diff
@@ -841,6 +848,18 @@ pub(crate) fn table_impl(mut args: TableArgs, item: &syn::DeriveInput) -> syn::R
     );
 
     let table_access = args.access.iter().map(|acc| acc.to_value());
+    let is_event = args.event.iter().map(|_| {
```
Centril: It's becoming high time that we break this function up, but in post 2.0

crates/datastore/src/locking_tx_datastore/state_view.rs
```diff
@@ -159,6 +159,12 @@ pub trait StateView {
             .unwrap_or(None)
             .transpose()?;
 
+        // Check if this table is an event table by looking up st_event_table.
+        let is_event = self
+            .iter_by_col_eq(ST_EVENT_TABLE_ID, StEventTableFields::TableId, value_eq)
+            .map(|mut iter| iter.next().is_some())
+            .unwrap_or(false);
+
```
Centril:
```suggestion
        // Check if this table is an event table by looking up `st_event_table`.
        let is_event = self
            .iter_by_col_eq(ST_EVENT_TABLE_ID, StEventTableFields::TableId, value_eq)?
            .next()
            .is_some()
```

crates/datastore/Cargo.toml
Centril:
```suggestion
```

PR #4253, opened 2026-02-10
[TS] Bundle ICU data

### LINKED ISSUE #3926: TypeScript module throws `RangeError: Internal error. Icu error.` procedure when trying to use http
**Bug**: TypeScript module V8 runtime throws `RangeError: Internal error. Icu error.` on Windows (breaks Intl/URL and `ctx.http.fetch`)

**Summary**
On Windows, a SpacetimeDB TypeScript server module throws `RangeError: Internal error. Icu error.` when calling ICU-dependent JS functionality (e.g. `"a".localeCompare("b")`) and also when constructing a URL (`new URL("https://example.com")`). This  breaks `ctx.http.fetch(...)` (HTTP requests), so TS procedures cannot call external apis in some cases (e.g. OpenAI in my use case).

**Environment**
- OS: Windows 11
```(base) PS C:\Users\merob\AppData\Local\SpacetimeDB\bin\current> spacetime --version
spacetime Path: C:\Users\merob\AppData\Local\SpacetimeDB\bin\current\spacetimedb-cli.exe
Commit: bbfa851b45db7096aa2a69d7dbad84f52983559f
spacetimedb tool version 1.11.1; spacetimedb-lib version 1.11.1;
```

**Impact**
- Any ICU/Intl usage appears broken (`localeCompare`, likely `Intl.*`, `toLocaleString`, etc.)
- URL parsing/creation is broken (`new URL(...)`)
- `ctx.http.fetch(...)` fails with the same ICU error, preventing HTTP requests from procedures

**Minimal reproduction**
Create a TS module with the following code and call the procedure from a client:

```ts
// spacetimedb/src/command.ts
import { t } from "spacetimedb/server";

export function registerCommandProcedures(db: { procedure: Function }) {
  db.procedure("icu_smoketest1", {}, t.string(), (_ctx: any) => {
    try {
      "a".localeCompare("b"); // << Triggers the error. This does too: new URL("https://example.com"); //<< Triggers the error
      return "OK";
    } catch (e: any) {
      return `ICU FAIL: ${String(e?.message ?? e)}`;
    }
  });
}
```

**Client:**
```ts
  async function onTestClick() {
    console.log("Sending test command to server");
    let result = await conn.procedures.icuSmoketest1({});
    console.log("Result:", result);
  }
```
```
Sending test command to server
App.svelte:52 Result: ICU FAIL: Internal error. Icu error.
```

coolreader18:
# Description of Changes

Fixes #3926. This PR bundles in icudata (11MB) to the spacetime binary, to allow i18n functions like `String.prototype.localeCompare()` to work. The 2nd commit sets the default locale to [en-001, "International English"](https://www.ctrl.blog/entry/en-001.html), in order to prevent system locale from affecting the execution of reducers.

1

# Testing

- [x] Verify this fixes the issue.

### DIFF
--- a/crates/core/src/host/v8/mod.rs
+++ b/crates/core/src/host/v8/mod.rs
@@ -107,6 +107,9 @@ impl V8RuntimeInner {
         // If the number in the name of this function is changed, update the version
         // of the `deno_core_icudata` dep to match the number in the function name.
         v8::icu::set_common_data_77(deno_core_icudata::ICU_DATA).ok();
+        // Set a default locale for functions like `toLocaleString()`.
+        // en-001 is "International English". <https://www.ctrl.blog/entry/en-001.html>
+        v8::icu::set_default_locale("en-001");
 
         // We don't want idle tasks nor background worker tasks,
         // as we intend to run on a single core.


coolreader18:
Nevermind I am able to reproduce, and this doesn't fix it. Please hold.

PR #4260, opened 2026-02-10
[TS] Throw error objects from syscalls

coolreader18:
# Description of Changes

This lets us avoid the try/catch around every syscall in javascript.

# Testing

- [ ] 
- [ ]


Centril (approved):
Looks good modulo the changes we discussed.

PR #4261, opened 2026-02-10
Standardize query builder syntax across Rust, TypeScript, and C# (Server/Client)

cloutiertyler:
# Description of Changes

Standardizes the query builder API across all three language SDKs (Rust, TypeScript, C#) for consistency.

**Rust:**
- Rename `Query` struct to `RawQuery`, make `Query` a trait with `fn into_sql(self) -> String`
- All builder types (`Table`, `FromWhere`, `LeftSemiJoin`, `RightSemiJoin`) implement `Query<T>` trait
- Views can return `-> impl Query<T>` instead of specifying exact builder types
- The `#[view]` macro auto-detects `impl Query<T>` and rewrites to `RawQuery<T>`
- Add `Not` variant to `BoolExpr` with `.not()` method

**TypeScript:**
- Add `ne()` to `ColumnExpression`
- Refactor `BooleanExpr` to `BoolExpr` class with chainable `.and()`, `.or()`, `.not()` methods
- Make builders valid queries directly (`.build()` deprecated but still works)
- Deprecate `from()` wrapper — use `tables.person.where(...)` directly
- Merge `query` export into `tables` so table refs are also query builders
- Add subscription callback form: `subscribe(ctx => ctx.from.person.where(...))`
- Unify `useTable` with query builder syntax; deprecate `filter.ts`

**C#:**
- Add `Not()` method to `BoolExpr<TRow>`
- Add `IQuery<TRow>` interface implemented by all builder types (`Table`, `FromWhere`, `LeftSemiJoin`, `RightSemiJoin`, `Query`)
- Add `ToSql()` to all builder types so `.Build()` is no longer required
- Update `AddQuery` to accept `IQuery<TRow>` instead of `Query<TRow>`

- Rust: `Query<T>` is now a trait (was a struct). The struct is renamed to `RawQuery<T>`. This is a breaking change for any code that used `Query<T>` as a type directly.
- TypeScript: `BooleanExpr` is now a `BoolExpr` class (was a discriminated union type). The `query` export is deprecated in favor of `tables`.
- C#: `AddQuery` now accepts `Func<QueryBuilder, IQuery<TRow>>` instead of `Func<QueryBuilder, Query<TRow>>`. Existing `.Build()` calls still work since `Query<TRow>` implements `IQuery<TRow>`.

3 — Changes touch multiple language SDKs and codegen, but each individual change is straightforward. The Rust macro rewrite for `impl Query<T>` detection is the most complex piece. All existing `.build()`/`.Build()` calls continue to work.

# Testing

- [x] `cargo test -p spacetimedb-query-builder` — 16/16 tests pass
- [x] `cargo check -p spacetimedb` — clean, no warnings
- [x] `cargo check` on views-query, views-sql, views-basic, views-trapped smoketest modules — all clean
- [x] `cargo test -p spacetimedb-codegen codegen_csharp` — snapshot updated, passes
- [x] `npm test` (TypeScript) — 101/101 tests pass
- [x] C# QueryBuilder tests — new tests for `Not()`, `IQuery<T>` interface
- [ ] CI passes

### DIFF
--- a/smoketests/tests/views.py
+++ b/smoketests/tests/views.py
@@ -724,51 +724,46 @@ fn init(ctx: &ReducerContext) {
 }
 
 #[spacetimedb::view(name = online_users, public)]
-fn online_users(ctx: &ViewContext) -> Query<User> {
-    ctx.from.user().r#where(|c| c.online.eq(true)).build()
+fn online_users(ctx: &ViewContext) -> impl Query<User> {
+    ctx.from.user().r#where(|c| c.online.eq(true))
 }
 
 #[spacetimedb::view(name = online_users_age, public)]
-fn online_users_age(ctx: &ViewContext) -> Query<Person> {
+fn online_users_age(ctx: &ViewContext) -> impl Query<Person> {
     ctx.from
         .user()
         .r#where(|u| u.online.eq(true))
         .right_semijoin(ctx.from.person(), |u, p| u.identity.eq(p.identity))
-        .build()
 }
 
 #[spacetimedb::view(name = offline_user_20_years_old, public)]
-fn offline_user_in_twienties(ctx: &ViewContext) -> Query<User> {
+fn offline_user_in_twienties(ctx: &ViewContext) -> impl Query<User> {
     ctx.from
         .person()
         .filter(|p| p.age.eq(20))
         .right_semijoin(ctx.from.user(), |p, u| p.identity.eq(u.identity))
         .filter(|u| u.online.eq(false))
-        .build()
 }
 
 #[spacetimedb::view(name = users_whos_age_is_known, public)]
-fn users_whos_age_is_known(ctx: &ViewContext) -> Query<User> {
+fn users_whos_age_is_known(ctx: &ViewContext) -> impl Query<User> {
     ctx.from
         .user()
         .left_semijoin(ctx.from.person(), |p, u| p.identity.eq(u.identity))
-        .build()
 }
 
 #[spacetimedb::view(name = users_who_are_above_20_and_below_30, public)]
-fn users_who_are_above_20_and_below_30(ctx: &ViewContext) -> Query<Person> {
+fn users_who_are_above_20_and_below_30(ctx: &ViewContext) -> impl Query<Person> {
     ctx.from
         .person()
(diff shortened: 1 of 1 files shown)


crates/bindings-typescript/src/react/useTable.ts
```diff
@@ -12,131 +12,62 @@ import type { UntypedRemoteModule } from '../sdk/spacetime_module';
...
+ * const [rows, isReady] = useTable(tables.user);
+ * const [onlineUsers, isReady] = useTable(
+ *   tables.user.where(r => r.online.eq(true)),
+ *   { onInsert: (row) => console.log('New user:', row) }
+ * );
  * ```
  */
 export function useTable<TableDef extends UntypedTableDef>(
-  tableDef: TableDef,
+  query: { toSql(): string } & Record<string, any>,
```
jsdt: We should use a `Query<TableDef>` here to get type safety (and tell the ide what type the result is) (https://github.com/clockworklabs/SpacetimeDB/commit/609343e198d43a22c294d437ef056249d7feb2e6).

crates/cli/.templates/quickstart-chat-typescript/src/App.tsx
```diff
@@ -0,0 +1,227 @@
+import React, { useState } from 'react';
```
jsdt: Were these files in `cli/.templates` meant to be added?

cloutiertyler:
@jsdt > Other than that, the way useTable is doing filtering on the client side is going to be wrong for semijoins, so it would be nice to start using per-query storage soon.

Definitely.

crates/cli/.templates/quickstart-chat-typescript/src/App.tsx
cloutiertyler: No, I'll pull those out.

crates/bindings-typescript/src/react/useTable.ts
cloutiertyler: Done

crates/cli/.templates/quickstart-chat-typescript/src/App.tsx
cloutiertyler: Done.

PR #4262, opened 2026-02-11
[C#] Adding `RawModuleDefV10` to C# Module Bindings

### LINKED ISSUE #4255: Export `RawModuleDefV10` from C# modules

rekhoff:
# Description of Changes
This is the implementation of #4255
* Refactored `Module` so all V10 builder state and registration helpers now live on `partial class RawModuleDefV10`. The static `Module` class simply registers components against that builder and serializes via `moduleDef.BuildModuleDefinition()`
* Removed the deprecated `__describe_module__` export. Only `__describe_module_v10__` is emitted now across the runtime, native shim, generated bindings, and snapshots
* Mirrored the Rust TODO that future V10 sections will cover Event tables and Case-conversion policy so we don't lose track of those items

Modules built with these bindings now expose only `__describe_module_v10__`
The legacy `__describe_module__` symbol is gone from both managed and native layers. Hosts expecting the old export must switch to the V10 entry point (which is already the canonical ABI for 2.0).

2 - Low. The refactor keeps the previous builder logic but relocates it, and the export removal matches the already-supported V10 host path.

# Testing
- [X] Successfully built CLI and DLLs without errors.
- [X] Ran `dotnet build crates/bindings-csharp/Runtime/Runtime.csproj` without errors.
- [X] Ran `dotnet test crates/bindings-csharp/Codegen.Tests/Codegen.Tests.csproj` without errors.
- [X] Ran `bash run-regression-tests.sh` without errors.


crates/cli/src/subcommands/dev.rs
```diff
@@ -620,7 +620,13 @@ async fn stream_logs(
...
     let status = res.status();
     if status.is_client_error() || status.is_server_error() {
-        let err = res.text().await?;
+        let mut err = res.text().await?;
+        // The server doesn't always send an error description in the response
+        // body (maybe it should), so default to status code + canonical reason
+        // phrase (e.g. "502 Bad Gateway").
+        if err.is_empty() {
+            err = format!("{status}");
+        }
```
bfops, line 629: I think this is a change from a different PR? I guess maybe `master` was merged in but it's not merged into the base branch?

rekhoff: I think all these changes must have been from the rebasing from `phoebe/rust-sdk-ws-v2` to `tyler/impl-event-tables`. It looks like the rebase made quite a mess of things.

Basically what was done here was:
Before:
`master` <- `jsdt/ws-v2` <- `phoebe/rust-sdk-ws-v2` <- `rekhoff/csharp-raw-definition-v10`
`master` <- `jsdt/ws-v2` <- `phoebe/rust-sdk-ws-v2` <- `tyler/impl-event-tables`

After:
`master` <- `jsdt/ws-v2` <- `phoebe/rust-sdk-ws-v2` <- `tyler/impl-event-tables` <- `rekhoff/csharp-raw-definition-v10`

They should have shared a base branch tree, so I'm not sure why it went sideways just yet.

rekhoff:
I will be cancelling/closing this PR, in favor of #4284 , since this one has it's change list thoroughly messed up from that rebase. 4284 is built on `tyler/impl-event-tables` from the start and should allow for implementation of Event Tables in this C# module bindings PR.

rekhoff:
Closing in favor of #4288, which does the same operation but based on `master`

PR #4263, opened 2026-02-11
Case conversion

Shubham8287:
# Description of Changes

Update the Default casing policy to `snake_case` for `RawModuleDefV10`.

Messy PR contains changes at different places, so that CI can pass:

Here are the main changes as follows:
- `bindings-macro` & `bindings` crate:   `name` macro in Indexes for canonical name and supply it to `RawModuleDefV10` via `ExplicitNames`.
- `bindings-typescript`: 
  - Changes has been reviewed through this PR - https://github.com/clockworklabs/SpacetimeDB/pull/4308.
   
 - `binding-csharp`: a single line change to pass `sourceName` of index instead of null.
- `codegen`:
  
   -  Changes has been  merged from branch - https://github.com/clockworklabs/SpacetimeDB/pull/4337. 
  
   - Except a fix in rust codegen to use canonical name in Query buillder instead of accessor.
  
  - `lib/db/raw_def`:  Extends `RawDefModuleV10` structure to support case conversion.
  
- `schema` crate:
     -  `validate/v9` - Nothing itself should change or changes in v9 validation logic but the file contains a `CoreValidator` which is shared with `validate/v10`. No test have t be updated to `validate/v9` which ensures we aren't regressing it.
     -  `validate/v10`: This is the main meat, look at the new tests added in bottom to understand what it does.
     
   -  Rest of the files are either test updates or module bindings. 
     
    ## Testing:
    1. Extensive unit tests have been added to verify generated `ModuleDef` is correct.
    2.  I have done some e2e testing to verify rust codegen with rust and typescript modules.
    3. I would have like to do more testing for other codegens , I am continue doing .

I have removed `sql.py` smoketest, as that seems to be already migated in new framework and was headache to update.

4, It could have side-effect which aren't easily visible.

 
 
 
 
 
 -  -  -

### DIFF
--- a/crates/bindings-csharp/Runtime/Internal/Autogen/CaseConversionPolicy.g.cs
+++ b/crates/bindings-csharp/Runtime/Internal/Autogen/CaseConversionPolicy.g.cs
@@ -12,7 +12,5 @@ namespace SpacetimeDB.Internal
     {
         None,
         SnakeCase,
-        CamelCase,
-        PascalCase,
     }
 }
--- a/crates/bindings-typescript/src/lib/autogen/types.ts
+++ b/crates/bindings-typescript/src/lib/autogen/types.ts
@@ -48,8 +48,6 @@ export type AlgebraicType = __Infer<typeof AlgebraicType>;
 export const CaseConversionPolicy = __t.enum('CaseConversionPolicy', {
   None: __t.unit(),
   SnakeCase: __t.unit(),
-  CamelCase: __t.unit(),
-  PascalCase: __t.unit(),
 });
 export type CaseConversionPolicy = __Infer<typeof CaseConversionPolicy>;
 
--- a/crates/lib/src/db/raw_def/v10.rs
+++ b/crates/lib/src/db/raw_def/v10.rs
@@ -101,10 +101,6 @@ pub enum CaseConversionPolicy {
     /// Convert to snake_case (SpacetimeDB default)
     #[default]
     SnakeCase,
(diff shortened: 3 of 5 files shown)


crates/codegen/src/typescript.rs
```diff
@@ -861,7 +856,15 @@ fn define_body_for_sum(
...
-    out.with_indent(|out| write_object_type_builder_fields(module, out, variants, None, false, false).unwrap());
+      // Convert variant names to PascalCase
+    let pascal_variants: Vec<(Identifier, AlgebraicTypeUse)> = variants
+        .iter()
+        .map(|(ident, ty)| {
+            let pascal = ident.deref().to_case(Case::Pascal);
+            (Identifier::for_test(pascal), ty.clone())
+        })
+        .collect();
+    out.with_indent(|out| write_object_type_builder_fields(module, out, &pascal_variants,, None, false, false).unwrap());
```
Shubham8287:
```suggestion
    out.with_indent(|out| write_object_type_builder_fields(module, out, &pascal_variants, None, false, false).unwrap());
```

Shubham8287:
This PR have  a known bug related to due to camel casing table identifier in typescript bindings, I have @coolreader18 to look into that.

joshua-spacetime (approved):
Most of this change set has already been reviewed and approved by various individuals including myself in the form of other PRs targeting this branch. Pending the tests succeeding, this should be ready to merge.

PR #4264, opened 2026-02-11
Rust: macro change `name` -> `accessor`

Shubham8287:
## Description of Changes

This PR primarily affects the `bindings-macro` and `schema` crates to review:

### Core changes

1. Replaces the `name` macro with `accessor` for **Tables, Views, Procedures, and Reducers** in Rust modules.
2. Extends `RawModuleDefV10` with a new section for:

   * case conversion policies
   * explicit names
    New sections are not validated in this PR so not functional.
3. Updates index behavior:

   * Index names are now always **system-generated** for clients. Which will be fixed in follow-up PR when we start validating RawModuleDef with explicit names.
   * The `accessor` name for an index is used only inside the module.

## Breaking changes (API/ABI)

1. **Rust modules**

   * The `name` macro must be replaced with `accessor`.

2. **Client bindings (all languages)**

   * Index names are now system-generated instead of using explicitly provided names.

**Complexity:** 3

A follow-up PR will reintroduce explicit names with support for case conversion.

### DIFF
--- a/demo/Blackholio/server-rust/src/lib.rs
+++ b/demo/Blackholio/server-rust/src/lib.rs
@@ -109,7 +109,7 @@ pub struct CircleRecombineTimer {
     player_id: i32,
 }
 
-#[spacetimedb::table(name = consume_entity_event, public, event)]
+#[spacetimedb::table(accessor = consume_entity_event, public, event)]
 pub struct ConsumeEntityEvent {
     consumed_entity_id: i32,
     consumer_entity_id: i32,


tools/xtask-llm-benchmark/src/templates/rust/server/.cargo/config.toml
```diff
@@ -1,4 +1,4 @@
-﻿# .cargo/config.toml
+﻿#[table(name =# .cargo/config.toml
```
Centril: what happened here?

crates/bindings-macro/src/table.rs
```diff
@@ -262,13 +285,27 @@ impl IndexArg {
...
             Ok(())
         })?;
-        let kind = kind
-            .ok_or_else(|| syn::Error::new_spanned(&attr.meta, "must specify kind of index (`btree` or `direct`)"))?;
-        let name = field.clone();
-        Ok(IndexArg::new(name, kind))
+        let kind = kind.ok_or_else(|| {
+            syn::Error::new_spanned(
+                &attr.meta,
+                "must specify kind of index (`btree` , `direct`, `name` or `value`)",
```
Centril: what happened here?

crates/bindings-macro/src/table.rs
```diff
@@ -176,15 +197,15 @@ impl IndexArg {
             });
             Ok(())
         })?;
-        let name = name.ok_or_else(|| meta.error("missing index name, e.g. name = my_index"))?;
+        let accessor = accessor.ok_or_else(|| meta.error("missing index name, e.g. accessor = my_index"))?;
```
Centril:
```suggestion
        let accessor = accessor.ok_or_else(|| meta.error("missing index name, e.g. `accessor = my_index`"))?;
```

Centril (approved):
Aside from these things that caught my attention, this looks good.

PR #4266, opened 2026-02-11
client-api: Resolve organization names via the tld, not the dns table

kim:
As specified in the proposal.


aasoni (approved):
Approving, just wanted to double check that resolve_namespace_owner doesn't appear to used anywhere. Is this meant to be used in the future?

kim:
> resolve_namespace_owner doesn't appear to used anywhere

good point, haha

PR #4269, opened 2026-02-11
CI - rust smoketests lib expansion

bfops:
# Description of Changes

A variety of small expansions/fixes to the smoketests lib.

Also, several of the tests were not working when `--server` was passed, due to name collisions. Each test runs with its own identity, but if it tried to publish over an existing database, it would get an auth error. I have fixed the name collisions. However, the original python smoketests simply used random strings, where we currently use hardcoded names and/or hardcoded names with PID suffixes. This seems less robust and should potentially be changed.

None. Testing only.

2

# Testing

- [x] CI passes


smoketests/tests/replication.py
```diff
@@ -96,11 +96,13 @@ def wait_for_leader_change(self, previous_leader_node, max_attempts=10, delay=2)
             try:
                  current_leader_node = self.get_leader_info()['node_id']
                  if current_leader_node != previous_leader_node:
+                    print(f"Leader changed from {previous_leader_node} to {current_leader_node}")
```
bfops: this was just for debugging, can be removed

crates/smoketests/src/lib.rs
```diff
@@ -401,6 +402,16 @@ impl ApiResponse {
...
 
+#[derive(Clone, Debug, Default)]
+pub struct PublishOptions {
+    pub clear: bool,
+    pub break_clients: bool,
+    pub num_replicas: Option<u32>,
+    pub organization: Option<String>,
+    pub force: bool,
+    pub stdin_input: Option<&str>,
+}
```
bfops, line 426: started refactoring to the builder pattern so we stop having a billion different versions of the `publish` function.

I would like to do this for several other parts of this lib as well, but it's not an imminent priority.

crates/smoketests/tests/smoketests/http_egress.rs
```diff
@@ -66,6 +66,7 @@ fn spawn_redirect_server(location: &str) -> (u16, JoinHandle<std::io::Result<()>
 
 #[test]
 fn test_http_disallowed_ip_is_blocked() {
+    require_local_server!();
```
bfops:
```suggestion
```

crates/smoketests/tests/smoketests/http_egress.rs
```diff
@@ -82,6 +82,7 @@ fn test_http_disallowed_ip_is_blocked() {
 
 #[test]
 fn test_http_redirect_to_disallowed_ip_is_blocked() {
+    require_local_server!();
```
bfops, line 85: I don't really understand how this test works, but it seems like given that we're spawning an extra server, this must only work for local servers

cloutiertyler: It's testing whether you can hit a private network address from within a module.

cloutiertyler (approved):

PR #4271, opened 2026-02-12
Typescript v2 API

jsdt:
# Description of Changes

This update the typescript sdk to use the v2 protocol.

The main user-facing API change is removing the existing reducer callbacks, and instead having reducers return a `Promise`.

The reducer functions `conn.reducers.onX` and `conn.reducers.removeOnX` no longer exist, and reducers now return `Promise<void>`.

2.5. 

# Testing

This has mostly been tested manually. I'm still updating some of the unit tests (I commented out the ones that used reducer callbacks).

### DIFF
--- a/crates/bindings-typescript/package.json
+++ b/crates/bindings-typescript/package.json
@@ -170,7 +170,6 @@
   ],
   "dependencies": {
     "base64-js": "^1.5.1",
-    "fast-text-encoding": "^1.0.0",
     "headers-polyfill": "^4.0.3",
     "prettier": "^3.3.3",
     "pure-rand": "^7.0.1",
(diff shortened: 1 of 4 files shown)


templates/chat-react-ts/src/App.tsx
```diff
@@ -103,7 +103,14 @@ function App() {
   const onSubmitMessage = (e: React.FormEvent<HTMLFormElement>) => {
     e.preventDefault();
     setNewMessage('');
-    sendMessage({ text: newMessage });
+    sendMessage({ text: newMessage }).then(
```
cloutiertyler: NIT given that it's user facing, I would do

```ts
.then().catch()
```

Up to you though.

templates/chat-react-ts/src/module_bindings/procedures.ts
cloutiertyler, line 1: Why are these gone? Should they have already been removed in another PR?

jsdt: These go in `types/procedures.ts` now. There may have been some weird merges that stopped it from being deleted earlier.

PR #4273, opened 2026-02-12
 [TS] schema() takes an object

coolreader18:
# Description of Changes

Implements the rest of the casing proposal.

# Testing

- [ ] 
- [ ]

### DIFF
--- a/crates/bindings-typescript/src/sdk/subscription_builder_impl.ts
+++ b/crates/bindings-typescript/src/sdk/subscription_builder_impl.ts
@@ -7,6 +7,7 @@ import type {
 import { EventEmitter } from './event_emitter';
 import type { UntypedRemoteModule } from './spacetime_module';
 import { isRowTypedQuery, toSql, type RowTypedQuery } from '../lib/query';
+import type { Values } from '../lib/type_util';
 
 export class SubscriptionBuilderImpl<RemoteModule extends UntypedRemoteModule> {
   #onApplied?: (ctx: SubscriptionEventContextInterface<RemoteModule>) => void =
@@ -87,7 +88,7 @@ export class SubscriptionBuilderImpl<RemoteModule extends UntypedRemoteModule> {
   ): SubscriptionHandleImpl<RemoteModule>;
   subscribe(
     queryFn: (
-      tables: RemoteModule['tables'][number]
+      tables: Values<RemoteModule['tables']>
     ) => RowTypedQuery<any, any> | RowTypedQuery<any, any>[]
   ): SubscriptionHandleImpl<RemoteModule>;
   subscribe(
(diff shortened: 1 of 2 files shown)


docs/docs/00200-core-concepts/00200-functions/00400-procedures.md
```diff
@@ -137,7 +137,7 @@ const MyTable = table(
     },
 )
 
-const spacetimedb = schema(MyTable);
```
cloutiertyler, line 140: NIT: it seems like on balance we're going for `camelCase` for table name accessors, we should probably standardize on that.

crates/bindings-typescript/src/sdk/reducers.ts
```diff
@@ -148,10 +148,10 @@ export function reducersToSchema<
  * @returns ColumnBuilder representing the complete database schema
  * @example
  * ```ts
- * const s = schema(
- *   table({ name: 'user' }, userType),
- *   table({ name: 'post' }, postType)
- * );
+ * const s = schema({
+ *   'user': table(, userType),
+ *   'post': table(, postType)
```
cloutiertyler: Looks like this got borked with a find replace:

```suggestion
 *   user: table(userType),
 *   post: table(postType)
```

cloutiertyler: Are TableOpts optional?

crates/cli/src/subcommands/project/typescript/index._ts
```diff
@@ -1,31 +1,32 @@
...
     {
       name: t.string(),
     }
   )
-);
+});
+export default spacetimedb;
 
-spacetimedb.reducer('init', (_ctx) => {
+export const init = spacetimedb.reducer((_ctx) => {
```
cloutiertyler:
```suggestion
export const init = spacetimedb.init((_ctx) => {
```

No?

crates/codegen/src/typescript.rs
```diff
@@ -337,10 +337,7 @@ impl Lang for TypeScript {
 
         writeln!(out);
         writeln!(out, "/** The tables available in this remote SpacetimeDB module. */");
-        writeln!(
-            out,
-            "export const tables = __convertToAccessorMap(tablesSchema.schemaType.tables);"
-        );
+        writeln!(out, "export const tables = tablesSchema.schemaType.tables;");
```
cloutiertyler: This is becoming a query builder in my PR, FYI. We can probably merge this first.

docs/docs/00200-core-concepts/00200-functions/00400-procedures.md
cloutiertyler: We can do that in a follow up PR.

PR #4278, opened 2026-02-12
Windows VM runner test

jdetter:
# Description of Changes

This enables the use of the self hosted windows runners.

|CI Job|Before|After|Change|
|---|---|---|---|
|CI/Smoketests (Windows)|79m|31m|154% faster|

This is *before* we do any work with caching.

None, this is just a CI change.

1 - this is just a CI change.

# Testing

- [x] Ran the windows smoketest several times. Each time it is passing, although several runs I was making performance changes so some runs are faster than others.


cloutiertyler (approved):

cloutiertyler:
I'm into it.

cloutiertyler (approved):

PR #4279, opened 2026-02-12
Limit update() to only work on primary keys

cloutiertyler:
## Summary
- Restricts `update()` across all three server-side SDKs (Rust, C#, TypeScript) to only work on primary key columns, not all unique columns
- Calling `update()` on a non-PK unique column is semantically a delete+insert — clients won't see it as a row update unless the primary key stays the same
- Fresh re-implementation of #1862, extended to cover C# and TypeScript bindings

### Design principle

Primary key is a constraint on columns, not a property of indexes. An index is unique or not unique — whether `update()` is available is derived from the column's attributes at the point of use, not stored on the index.

This differs from #1862 which introduced a `Uniqueness` enum (`No | Unique | PrimaryKey`) on the index itself. Instead, we keep `is_unique: bool` and check the column metadata where needed.

### Rust changes
- Add `PrimaryKey` marker trait and `where Col: PrimaryKey` bound on `UniqueColumn::update()`
- `marker_type()` receives `primary_key_column` and checks the column directly — no `is_pk` field on the index
- `sdk-test` unique tables use `update_non_pk_by` (delete+insert) instead of `update_by`
- Benchmarks that used `update()` on `#[unique]` columns changed to either `#[primary_key]` or delete+insert

### C# changes
- Only emit `Update()` on `UniqueIndex` when the column has `PrimaryKey` attrs
- Update `unique_*` test reducers in `sdk-test-cs` to use `Delete()`+`Insert()` instead

### TypeScript changes
- `UniqueIndex` now has only `find` + `delete` (no `update`)
- `AllColumnsPrimaryKey` type-level helper checks column metadata from the `TableDef`
- `Index` routing conditionally intersects `{ update() }` when columns are PK
- Runtime only attaches `update` method for PK unique indexes
- `sdk-test-ts` unique tables use delete+insert instead of update
- Type-level test in `schema.test-d.ts` verifies `id.update()` compiles (PK) and `name2.update` errors (non-PK unique)

## Test plan
- [x] C# codegen tests pass (`dotnet test crates/bindings-csharp/Codegen.Tests`)
- [x] Rust SDK test module compiles
- [x] Benchmarks compile
- [x] TypeScript type checks pass (`npx tsc --noEmit`)
- [ ] CI passes

Closes #1862

### DIFF
--- a/modules/benchmarks-ts/src/ia_loop.ts
+++ b/modules/benchmarks-ts/src/ia_loop.ts
@@ -226,7 +226,8 @@ function moveAgent(
   ctx.db.gameTargetableState.entity_id.update(targetable);
 
   if (ctx.db.gameLiveTargetableState.entity_id.find(entityId) != null) {
-    ctx.db.gameLiveTargetableState.entity_id.update({
+    ctx.db.gameLiveTargetableState.entity_id.delete(entityId);
+    ctx.db.gameLiveTargetableState.insert({
       entity_id: entityId,
       quad: newHash,
     });


tamaro-skaljic:
I only read through comments and not through code changes and the 2.0 breaking changes / migration guide, but I think this is a breaking change (not prefixed with `[2
0 breaking]` as other PRs)?

PR #4284, opened 2026-02-12
[C#] Adding `RawModuleDefV10` to C# Module Bindings

### LINKED ISSUE #4255: Export `RawModuleDefV10` from C# modules

rekhoff:
# Description of Changes
This is the implementation of #4255
This is a rebuild of #$4255 after a rebase throughly messed up that PR's changelog.
* Refactored `Module` so all V10 builder state and registration helpers now live on `partial class RawModuleDefV10`. The static `Module` class simply registers components against that builder and serializes via `moduleDef.BuildModuleDefinition()`
* Removed the deprecated `__describe_module__` export. Only `__describe_module_v10__` is emitted now across the runtime, native shim, generated bindings, and snapshots
* Mirrored the Rust TODO that future V10 sections will cover Event tables and Case-conversion policy so we don't lose track of those items

Modules built with these bindings now expose only `__describe_module_v10__`
The legacy `__describe_module__` symbol is gone from both managed and native layers. Hosts expecting the old export must switch to the V10 entry point (which is already the canonical ABI for 2.0).

2 - Low. The refactor keeps the previous builder logic but relocates it, and the export removal matches the already-supported V10 host path.

# Testing
- [X] Successfully built CLI and DLLs without errors.
- [X] Ran `dotnet build crates/bindings-csharp/Runtime/Runtime.csproj` without errors.
- [X] Ran `dotnet test crates/bindings-csharp/Codegen.Tests/Codegen.Tests.csproj` without errors.
- [X] Ran `bash run-regression-tests.sh` without errors.


Shubham8287:
Please check this out - https://github.com/clockworklabs/SpacetimeDB/pull/4246, Rust and typescript counterparts are using the `__describe_module__` but returns new variant containing`RawModuleDefV10`

rekhoff:
Closing in favor of #4288, which does the same operation but based on `master`

Issue #3789, opened 2025-11-27
Procedures: enable long-running loops which hold non-`SpacetimeType` state in memory


gefjon:
E.g. a physics world. This will require:
- On-abort handlers for procedures, so the loop can restart after a crash.
- Proper handling of publishing, i.e. killing the procedure after confirmation and a timeout.
- Changing our database-per-core scheduler to re-home running futures on core-repin, rather than just switching new futures to the new core.

onx2:
This will be such a huge win and unlock a lot of possibilities for games in spacetime - definitely my number 1 want/need right now 😄.

innocenzi:
Hey, just curious if this has been planned/prioritized, to get an idea of when we will have this feature? It'll be a huge step forward 👀

PR #4298, opened 2026-02-15
Add `on_abort` handler support for procedures

onx2:
# Description of Changes

Adds support for procedure `on_abort` handlers, including macro parsing, bindings propagation, schema validation, and runtime scheduling of the handler when a procedure traps. Also updates V9 validation tests to use the new `add_procedure(..., on_abort)` signature.

Helps #3789, this allows us to reschedule the procedure when aborted

No breaking changes expected. (Adds optional `on_abort` metadata and preserves existing behavior when omitted.)

Touches multiple layers (macro → bindings → schema → runtime), but the change is additive and limited in scope. Primary risk is mismatched server/client versions or handler scheduling edge cases, mitigated by validation and async scheduling.

# Testing

- [x] Run `cargo test`
- [x] (Optional) Integration check: publish a module with `#[procedure(on_abort = "...")]`, trigger a trap, and verify handler is scheduled

### DIFF
--- a/crates/bindings-macro/src/lib.rs
+++ b/crates/bindings-macro/src/lib.rs
@@ -52,6 +52,7 @@ mod sym {
     symbol!(index);
     symbol!(init);
     symbol!(name);
+    symbol!(on_abort);
     symbol!(primary_key);
     symbol!(private);
     symbol!(public);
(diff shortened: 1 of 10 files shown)


xDovos:
what procedures need is a st_table that holds the process/thread id, if it should stop, maybe an auto restart bool, and what scheduled_id it has in what scheduled table.
that way, it can check if it should stop, the server can check if the thread is still running, and it has the info it needs to restart or remove the procedure from the scheduled table when it crashes or a publish happens

PR #4301, opened 2026-02-15
Trigger view refresh when a WASM procedure commits a transaction

joshua-spacetime:
# Description of Changes

This PR fixes the bug where WASM procedures could commit a transaction without refreshing affected materialized views, which caused view-backed subscriptions to miss updates from procedure writes.

The equivalent changes for V8 will be made in a separate patch.

None

3

This was not a simple translation of the reducer code path, because reducers are run in a single transaction whereas procedures can have multiple transactions via `with_tx` and views must be refreshed with each transaction instead of once at the end of the procedure. This required refresh to be inserted into the syscall itself which required some extra plumbing. Mainly that we had to attach the validated `ModuleDef` to the `WasmInstanceEnv`. This should not affect hotswapping because we instantiate an entirely new `WasmInstanceEnv` in that case.

# Testing

- [x] Added a regression test in the form of a smoketest that subscribes to a view and calls a procedure

### DIFF
--- a/crates/core/src/host/wasm_common/module_host_actor.rs
+++ b/crates/core/src/host/wasm_common/module_host_actor.rs
@@ -82,6 +82,8 @@ pub trait WasmInstance {
 
     fn tx_slot(&self) -> TxSlot;
 
+    fn set_module_def(&mut self, module_def: Arc<ModuleDef>);
+
     fn call_reducer(&mut self, op: ReducerOp<'_>, budget: FunctionBudget) -> ReducerExecuteResult;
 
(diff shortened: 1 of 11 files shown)


crates/core/src/host/wasm_common/module_host_actor.rs
```diff
@@ -143,6 +148,67 @@ fn deserialize_view_rows(
         .collect()
 }
 
+pub(crate) fn run_query_for_view(
```
joshua-spacetime, line 148: No new code was added here. It was just extracted into a helper for reuse in procedure view refresh.

PR #4302, opened 2026-02-16
Trigger view refresh when a V8 procedure commits a transaction

### LINKED ISSUE #4296: Procedures don't trigger view changes
Cross posting here for visibility and tracking. Original post: https://discord.com/channels/1037340874172014652/1472332044460622006

TL;DR; A procedure with the same db update as a reducer does not trigger a change to a view that uses that data.

joshua-spacetime:
# Description of Changes

This PR fixes the bug where V8 procedures could commit a transaction without refreshing affected materialized views, which caused view-backed subscriptions to miss updates from procedure writes.

None

3

Almost the exact same change as https://github.com/clockworklabs/SpacetimeDB/pull/4301. The main difference between the V8 change and the WASM change is that for V8 we persist the hook functions (`__call_view__`, `__call_view_anon__`, etc.) and the hook receiver (`recv` / `this`) into V8 context embedder slots, and reconstruct them later from syscall scope.

# Testing

- [x] Added a regression test in the form of a typescript smoketest that subscribes to a view and calls a procedure

### DIFF
--- a/crates/core/src/host/v8/mod.rs
+++ b/crates/core/src/host/v8/mod.rs
@@ -210,6 +210,7 @@ fn env_on_isolate_unwrap(isolate: &mut Isolate) -> &mut JsInstanceEnv {
 /// The environment of a [`JsInstance`].
 struct JsInstanceEnv {
     instance_env: InstanceEnv,
+    module_def: Option<Arc<ModuleDef>>,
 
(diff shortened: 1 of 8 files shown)


coolreader18 (approved):
LGTM!

PR #4304, opened 2026-02-16
SQL alias

Shubham8287:
# Description of Changes

 - [x] : Store "accessor"s names in ModuleDef (for tables, columns, and indexes for now) and pass them to TableSchema as aliases.

- [x] Create accessor system tables for each of them: `st_table_accessor`,`st_column_accessor`, `st_index_accessor`

- [x] Update the SQL parsing logic to also consider alias names.

Alias name stored will be same as actual names until we merge https://github.com/clockworklabs/SpacetimeDB/pull/4263
NA

3

This patch updates SQL name resolution which isn't a very complex change since table and column name resolution are each handled by their own helper method.

It also adds new system tables which need to be handle correctly during restore/replay. We do have regression tests for this that should catch any incorrect handling or missing steps.

I would say the main risk is that we don't populate cached table and column schemas with the alias/accessor name. Regression tests have been added to catch this, but there are edge cases around migrations and replay that currently aren't covered.

# Testing
 
Added SQL smoketests that use canonical/accessor table/column names. They currently fail because we're currently using the accessor name for both. They should pass once this is updated.

### DIFF
--- a/crates/core/src/db/relational_db.rs
+++ b/crates/core/src/db/relational_db.rs
@@ -10,7 +10,7 @@ use enum_map::EnumMap;
 use log::info;
 use spacetimedb_commitlog::repo::OnNewSegmentFn;
 use spacetimedb_commitlog::{self as commitlog, Commitlog, SizeOnDisk};
-use spacetimedb_data_structures::map::{HashMap, HashSet};
+use spacetimedb_data_structures::map::HashSet;
 use spacetimedb_datastore::db_metrics::DB_METRICS;
 use spacetimedb_datastore::error::{DatastoreError, TableError, ViewError};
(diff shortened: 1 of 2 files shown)


crates/smoketests/tests/sql.rs
```diff
@@ -63,3 +63,72 @@ fn test_sql_format() {
  (bool_opt = (some = true), bool_result = (ok = false), action = (Active = ()))"#,
     );
 }
+
+// #[test]
+fn _test_sql_resolves_accessor_and_canonical_names_for_table() {
```
Shubham8287, line 68: These tests are disable because they expect accessor and canonical names to be different.
They will not pass in this PR but works in https://github.com/clockworklabs/SpacetimeDB/pull/4263 branch.

Issue #4295, opened 2026-02-13
Casing: Canonical name for Csharp.


Shubham8287:
### 2.0 Breaking change:

Convert existing `name` to `accessor` 

### Not breaking(but nice to have in 2.0);

Usee `name` for canonical names.

We want to support canonical name overrides for tables, reducers, views, procedures, and indexes, as described in the proposal.

Modules should provide these canonical names through the `ExplicitNames` section of `RawModuleDefV10`, as outlined in the code commentary in this PR:
[https://github.com/clockworklabs/SpacetimeDB/pull/4294](https://github.com/clockworklabs/SpacetimeDB/pull/4294)

Also provide the way to choose the casing setting and pass it me `RawModuleDefV10` : `CaseConversionPolicy` section

```csharp
public partial class Module
{
    [SpacetimeDB.Settings]
    public const CaseConversionPolicy CASE_CONVERSION_POLICY = CaseConversionPolicy.SnakeCase;
}
```

PR #4306, opened 2026-02-16
Updated C# `Name` attribute to `Accessor`

rekhoff:
# Description of Changes

Implementation of #4295

Convert existing `Name` attribute to `Accessor` to support new Canonical Case conversation of 2.0

Yes, in C# modules, we no longer use the attribute name `Name`, it should now be `Accessor` 

1

# Testing

- [X] Build and tested locally
- [X] Ran regression tests locally


Shubham8287 (approved):
Thanks!

PR #4307, opened 2026-02-16
Add more tests for typescript client and fix some bugs

jsdt:
# Description of Changes

This fixes a couple issues:
 - We are now using the `SendDroppedRows` flag for unsubscribe messages.
 - We parse reducer errors as strings now when throwing errors.
 - `useTable` was not doing client-side filtering correctly for object types (Timestamp, ConnectionId, Identity)
 - `useTable` was not always recomputing snapshots when it needed to, because it couldn't distinguish between two events if they were caused by reducers called by other connections (or two subscription applied events). All events how have a client-generated id attached, so we can tell if there was actually a new event processed. 

Once we have per-query storage, we can purge the client-side filtering code.

This also adds some tests to cover these cases, and includes a bit of refactoring.

Another change added here is to use `SenderError` in the typescript SDK for errors that were returned by reducers (which throw `SenderError` in typescript modules).

1

# Testing

This has some new tests added.

Much of the `useTable` logic was tested manually.

### DIFF
--- a/crates/bindings-macro/src/table.rs
+++ b/crates/bindings-macro/src/table.rs
@@ -80,7 +80,7 @@ impl TableArgs {
         let mut access = None;
         let mut scheduled = None;
         let mut accessor = None;
-        let mut name = None;
+        let mut name: Option<LitStr> = None;
         let mut indices = Vec::new();
         let mut event = None;
         syn::meta::parser(|meta| {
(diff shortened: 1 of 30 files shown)


crates/bindings-typescript/src/sdk/db_connection_impl.ts
```diff
@@ -577,6 +580,24 @@ export class DbConnectionImpl<RemoteModule extends UntypedRemoteModule>
     return pendingCallbacks;
   }
 
+  #applyTransactionUpdates(
+    eventContext: EventContextInterface<RemoteModule>,
+    tu: TransactionUpdate
+  ): PendingCallback[] {
+    const all_updates: CacheTableUpdate<UntypedTableDef>[] = [];
```
cloutiertyler: NIT: use `camelCase`, we really ought to have a linter for this, but alas

crates/bindings-typescript/src/sdk/db_connection_impl.ts
```diff
@@ -409,7 +411,7 @@ export class DbConnectionImpl<RemoteModule extends UntypedRemoteModule>
       ClientMessage.Unsubscribe({
         querySetId: { id: querySetId },
         requestId,
-        flags: UnsubscribeFlags.Default,
+        flags: UnsubscribeFlags.SendDroppedRows,
```
cloutiertyler, line 416: What does this mean? And why are we now enabling it? I'm not so familiar with v2.

I have a vague recollection that it's required because we don't have per query storage, is that right?

crates/bindings-typescript/src/sdk/websocket_test_adapter.ts
```diff
@@ -9,14 +11,21 @@ class WebsocketTestAdapter {
...
     this.messageQueue = [];
+    this.outgoingMessages = [];
     this.closed = false;
   }
 
   send(message: any): void {
+    const parsedMessage = ClientMessageSerde.deserialize(
+      new BinaryReader(message)
+    );
+    this.outgoingMessages.push(parsedMessage);
```
cloutiertyler, line 27: We're just accumulating messages here? What's the purpose of this?

crates/bindings-typescript/src/sdk/db_connection_impl.ts
jsdt: Since we don't have per-query storage implemented, when we Unsubscribe, we need the server to tell us which rows need to be removed from our cache. Once we have it implemented, the client can figure this out, so we can avoid sending it.

crates/bindings-typescript/src/sdk/websocket_test_adapter.ts
jsdt: We are already storing the messages in binary form, I just put the parsing here instead of having it duplicated every time we look at the messages.

The point is to let the tests inspect the outgoing messages. This lets us extract request ids and query set ids, so we can create the right response messages. We could make the adapter take a callback, so each test can decide how to store or throw away messages, rather than store them all forever, but with the current tests it's not a big issue.

crates/bindings-typescript/src/sdk/event.ts
```diff
@@ -1,9 +1,18 @@
...
+type WithId = {
+  /**
+   * A client-generated id to distinguish between different events.
+   */
+  id: string;
+};
+
+export type Event<Reducer extends ReducerEventInfo> = WithId &
+  (
+    | { tag: 'Reducer'; value: ReducerEvent<Reducer> }
```
cloutiertyler, line 13: I believe this should be removed, right?

cloutiertyler (approved):
Left one important comment about removing `ReducerEvent`s

crates/bindings-typescript/src/sdk/event.ts
jsdt: This is still used if the event was triggered by a reducer called by the client.

PR #4308, opened 2026-02-16
js macro change

Shubham8287:
# Description of Changes
Typescript changes:
- rename index `name` to `accessor`
- `name` provided with table & view passed as explicit name to `RawModuleDefV10`, and do not overrride accessor names.

Breaks existing modules.

2

# Testing
CIs


Shubham8287:
This needs case-conversion to pass the CI , I am merging the changes into original PR -https://github.com/clockworklabs/SpacetimeDB/pull/4263.

PR #4310, opened 2026-02-16
Improve `merge_apply_inserts`

Centril:
# Description of Changes

This recovers a bit of lost performance from event tables by:
- avoiding putting the `row: ProductValue` in a vector before merging into the committed state.
- directly constructing `Arc<[ProductValue]>`

This also shaves off 8 bytes from `ReplicaCtx`.

None

2?

# Testing

Covered by existing tests.

### DIFF
--- a/crates/datastore/src/locking_tx_datastore/committed_state.rs
+++ b/crates/datastore/src/locking_tx_datastore/committed_state.rs
@@ -1230,6 +1230,9 @@ impl CommittedState {
 
     /// Collects the inserted rows in `tx_table` into `tx_data`,
     /// and applies `on_row` to each inserted row.
+    ///
+    /// The `on_row` closure will be called with each inserted row.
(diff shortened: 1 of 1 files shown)


crates/datastore/src/locking_tx_datastore/committed_state.rs
```diff
@@ -1210,40 +1210,62 @@ impl CommittedState {
...
+                });
             }
+        }
+    }
 
-            // Add the table to `TxData` if there were insertions.
-            if !inserts.is_empty() {
-                tx_data.set_inserts_for_table(table_id, &schema.table_name, inserts.into());
+    /// Collects the inserted rows in `tx_table` into `tx_data`,
+    /// and applies `on_row` to each inserted row.
```
gefjon, line 1232: Would appreciate expanding this doc comment to explain what the `on_row` is for. Something like, "The `on_row` will be called with each inserted row. `Self::merge_apply_inserts` uses this to add non-event rows to the committed state."

PR #4311, opened 2026-02-16
Bypass `AlgebraicValue` for datastore updates and bsatn based index scans + BytesKey optimization

Centril:
# Description of Changes

Introduces types `TypedIndexKey` and `IndexKey` to free the table index code on its direct dependency on `AlgebraicValue`, allowing index scans by BSATN, `RowRef`s as well without first going through `AlgebraicValue`.
This also has the effect of optimizing string scans by avoiding allocating in `AlgebraicValue::String`.

This also adds a byte-array based future optimization for all-primitive-multi-column indexes.

Also in the future, this will enable optimizing `iter_by_col_eq`, which is used by the frequent connect/disconnect logic.

None

3? Unsafe code and very load bearing code.

# Testing

### DIFF
--- a/crates/table/src/table_index/mod.rs
+++ b/crates/table/src/table_index/mod.rs
@@ -42,7 +42,7 @@ use crate::{read_column::ReadColumn, static_assert_size};
 use core::ops::{Bound, RangeBounds};
 use core::{fmt, iter};
 use spacetimedb_primitives::{ColId, ColList};
-use spacetimedb_sats::bsatn::{decode, from_reader, from_slice};
+use spacetimedb_sats::bsatn::{decode, from_reader};
 use spacetimedb_sats::buffer::{DecodeError, DecodeResult};
 use spacetimedb_sats::memory_usage::MemoryUsage;
 use spacetimedb_sats::product_value::InvalidFieldError;
@@ -1665,9 +1665,14 @@ impl TableIndex {
     ) -> DecodeResult<PointOrRange<'de>> {
         use TypedIndex::*;
 
+        let read_bound = |mut bytes| {
+            let reader = &mut bytes;
+            from_reader::<Bound<()>>(reader).map(|b| b.map(|()| *reader))
+        };
+
         // Decode just whether it's inclusive or other bound forms.
-        let start = from_slice(rstart)?;
-        let end = from_slice(rend)?;
+        let start = read_bound(rstart)?;
+        let end = read_bound(rend)?;
 
         match &self.key_type {
             // Multi-column index case or single-column index on a product field.


crates/core/src/config.rs
```diff
@@ -54,11 +54,12 @@ impl MetadataFile {
         path.write(self.to_string())
     }
 
-    fn check_compatibility(previous: &Self, current: &Self) -> anyhow::Result<()> {
+    fn check_compatibility(previous: &Self, current: &Self, metafile: &Path) -> anyhow::Result<()> {
```
gefjon, line 57: It would have been nice if the changes in this file were a separate PR.

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -280,11 +272,10 @@ impl MutTxId {
         op: &FuncCallType,
         table_id: TableId,
         index_id: IndexId,
-        lower: Bound<AlgebraicValue>,
-        upper: Bound<AlgebraicValue>,
+        point: Option<IndexKey<'_>>,
```
gefjon, line 278: Taking a `point` in a function that's nominally for recording range scans is super weird. I see below that it's because `record_index_scan_range_inner` distinguishes only between point scans and everything else, and has a comment to that effect. But hoo boy does it look like a bug when looking just at this method. Please add some commentary here.

crates/sats/src/de/impls.rs
```diff
@@ -477,6 +477,8 @@ impl<'de, T: Deserialize<'de>, U: Deserialize<'de>> VariantVisitor<'de> for Resu
     }
 }
 
+impl_deserialize!([T: Deserialize<'de>] Bound<T>, de => PhantomData.deserialize(de));
```
gefjon: This impl looks really suspicious. `Bound<T>` is not zero-sized. And we've got a bunch of machinery for deserializing bounds using `WithBound` and `BoundVisitor`, just below, no?

crates/table/src/table_index/btree_index.rs
```diff
@@ -118,6 +104,43 @@ impl<K: Ord + KeySize> Index for BTreeIndex<K> {
     }
 }
 
+impl<K: KeySize + Ord> BTreeIndex<K> {
+    /// See [`Index::delete`].
+    /// This version has relaxed bounds.
```
gefjon: What does "relaxed bounds" mean in this context?

crates/table/src/table_index/btree_index.rs
```diff
@@ -118,6 +104,43 @@ impl<K: Ord + KeySize> Index for BTreeIndex<K> {
...
+        if deleted {
+            self.num_rows -= 1;
+            self.num_key_bytes.sub_from_key_bytes(key);
+        }
+
+        deleted
+    }
+
+    /// See [`Index::seek_point`].
+    /// This version has relaxed bounds.
```
gefjon: Same q re: "relaxed bounds."

crates/table/src/table_index/btree_index.rs
```diff
@@ -127,6 +150,17 @@ impl<K: Ord + KeySize> RangedIndex for BTreeIndex<K> {
     /// Returns an iterator over the multimap that yields all the `V`s
     /// of the `K`s that fall within the specified `range`.
     fn seek_range(&self, range: &impl RangeBounds<Self::Key>) -> Self::RangeIter<'_> {
+        self.seek_range(range)
+    }
+}
+
+impl<K: KeySize + Ord> BTreeIndex<K> {
+    /// See [`RangedIndex::seek_range`].
+    /// This version has relaxed bounds.
```
gefjon: And again.

crates/table/src/table_index/bytes_key.rs
```diff
@@ -0,0 +1,181 @@
...
+        // For sum, we report the greatest possible fixed size.
+        // A key may be of variable size, a long as it fits within an upper bound.
+        Sum(ty) => {
+            let mut max_size = 0;
+            for var in &ty.variants {
+                let variant_size = required_bytes_key_size(&var.algebraic_type, is_range_index)?;
+                max_size = max_size.max(variant_size);
+            }
+            Some(1 + max_size)
+        }
```
gefjon, line 67: Is it correct to store BSATN-ified sums in a range index? I'd appreciate some justification of that.

crates/table/src/table_index/bytes_key.rs
```diff
@@ -0,0 +1,181 @@
...
+
+        // For sum, we report the greatest possible fixed size.
+        // A key may be of variable size, a long as it fits within an upper bound.
+        Sum(ty) => {
+            let mut max_size = 0;
+            for var in &ty.variants {
+                let variant_size = required_bytes_key_size(&var.algebraic_type, is_range_index)?;
+                max_size = max_size.max(variant_size);
+            }
+            Some(1 + max_size)
```
gefjon, line 66: A quick comment like, "Add one byte for the tag" would be nice here.

crates/table/src/table_index/bytes_key.rs
```diff
@@ -0,0 +1,181 @@
...
+        U64 | I64 | F64 => Some(mem::size_of::<u64>()),
+        U128 | I128 => Some(mem::size_of::<u128>()),
+        U256 | I256 => Some(mem::size_of::<u256>()),
+    }
+}
+
+impl<const N: usize> BytesKey<N> {
+    /// Decodes `self` as an [`AlgebraicValue`] at `key_type`.
+    ///
+    /// Panics if the wrong `key_type` is provided, but that should never happen.
```
gefjon: This comment makes it seem like this method will deterministically detect the mismatch and panic, which it won't. Please revise.

crates/table/src/table_index/bytes_key.rs
```diff
@@ -0,0 +1,181 @@
...
+/// serialized to a byte array.
+///
+/// The key can store up to `N` bytes
+/// where `N` is determined by the summed size of each column in the index
+/// when serialized in BSATN format,
+/// which is the same as little-endian encoding of the keys for primitive types.
+///
+/// As we cannot have too many different `N`s,
+/// we have a few `N`s, where each is a power of 2.
+/// A key is then padded to the nearest `N`.
```
gefjon: Padded with what? Zeroes?

crates/table/src/table_index/bytes_key.rs
```diff
@@ -0,0 +1,181 @@
...
+
+    /// Decodes `prefix` and `endpoint` in BSATN to a [`BytesKey<N>`]
+    /// by copying over both if they fit into the key.
+    pub(super) fn from_bsatn_prefix_and_endpoint(
+        prefix: &[u8],
+        prefix_types: &[ProductTypeElement],
+        endpoint: &[u8],
+        range_type: &AlgebraicType,
+    ) -> DecodeResult<Self> {
+        // Validate the BSATN.
```
gefjon, line 119: Why is it necessary to validate the BSATN? I assume because it's provided by untrusted module code, but could use a comment.

crates/table/src/table_index/hash_index.rs
```diff
@@ -72,31 +73,16 @@ impl<K: KeySize + Eq + Hash> Index for HashIndex<K> {
...
-            self.num_rows -= 1;
-            self.num_key_bytes.sub_from_key_bytes::<Self>(key);
-        }
-
-        if is_empty {
-            entry.remove();
-        }
-
-        deleted
+        self.delete(key, ptr)
```
gefjon, line 76: Why is `HashIndex::delete` a separate method from `<HashIndex as Index>::delete`? Why not just inline the impl here?

gefjon: Same question for `seek_point` and for other index impls, but I'm not gonna bother adding all the comments.

crates/table/src/table_index/mod.rs
```diff
@@ -12,19 +12,21 @@
...
-//
-// I (pgoldman 2024-02-05) suspect, but have not measured, that there's no real reason
-// to have a `ProductType` variant, which would apply to multi-column indexes.
-// I believe `ProductValue::cmp` to not be meaningfully faster than `AlgebraicValue::cmp`.
-// Eventually, we will likely want to compile comparison functions and representations
-// for `ProductValue`-keyed indexes which take advantage of type information,
-// since we know when creating the index the number and type of all the indexed columns.
-// This may involve a bytecode compiler, a tree of closures, or a native JIT.
+//!
+//! I (pgoldman 2024-02-05) suspect, but have not measured, that there's no real reason
```
gefjon, line 16: Oh wow, blast from the past.

crates/table/src/table_index/mod.rs
```diff
@@ -110,14 +117,19 @@ enum TypedIndexRangeIter<'a> {
     BTreeI32(BTreeIndexRangeIter<'a, i32>),
     BTreeU64(BTreeIndexRangeIter<'a, u64>),
     BTreeI64(BTreeIndexRangeIter<'a, i64>),
-    BTreeU128(BTreeIndexRangeIter<'a, Packed<u128>>),
-    BTreeI128(BTreeIndexRangeIter<'a, Packed<i128>>),
+    BTreeU128(BTreeIndexRangeIter<'a, u128>),
```
gefjon, line 120: At some point I feel like we should switch to macro-generating a lot of this crap. Seems like a pretty simple X macro with `macro_rules` could save us a lot of LOC, typing, review burden &c.

crates/table/src/table_index/mod.rs
```diff
@@ -215,6 +242,420 @@ impl fmt::Debug for TableIndexRangeIter<'_> {
...
+const BYTES_KEY_SIZE_8_B: usize = 8;
+const BYTES_KEY_SIZE_8_H: usize = size_sub_row_pointer(16);
+const _: () = assert!(BYTES_KEY_SIZE_8_B == BYTES_KEY_SIZE_8_H);
+const BYTES_KEY_SIZE_16_B: usize = 16;
+const BYTES_KEY_SIZE_24_H: usize = size_sub_row_pointer(32);
+const BYTES_KEY_SIZE_32_B: usize = 32;
+const BYTES_KEY_SIZE_56_H: usize = size_sub_row_pointer(64);
+const BYTES_KEY_SIZE_64_B: usize = 64;
+const BYTES_KEY_SIZE_120_H: usize = size_sub_row_pointer(128);
+const BYTES_KEY_SIZE_128_B: usize = 128;
```
gefjon: Can you either rename these or add a comment to make clear that they apply to BTree and Hash indexes?

crates/table/src/table_index/mod.rs
Centril: Yeah maybe, and there is one macro already employed, but there's also a lot of repetition with a few outliers that mess up the uniformity and the ease of writing simple macros.

crates/table/src/table_index/btree_index.rs
Centril: The gist of "relaxed bounds" is that this is essentially the relaxed bounds found on `{Hash,BTree}Map::{get, remove}` that allow you to query with `&str` when you have e.g., `HashMap<String, V>`.

crates/table/src/table_index/hash_index.rs
Centril: I've added comments to all the relevant places.
The main motivation for having relaxed bounds methods is so that we can enable e.g., seeks via borrowed keys, e.g., directly from BSATN-encoded `&[u8]` and avoid some allocations.

crates/table/src/table_index/bytes_key.rs
Centril: Yep, expanded the comment.

crates/sats/src/de/impls.rs
Centril: Woops; this should have been `WithBound(PhantomData)`. Fixed.

crates/core/src/config.rs
Centril: Fair :)

PR #4317, opened 2026-02-17
Fix `spacetime dev` watch filtering and improve quickstart copy-paste experience

cloutiertyler:
## Summary
- Fix `spacetime dev` failing on C# projects by watching the module directory itself instead of hardcoded `spacetimedb_dir/src/` (which doesn't exist for C# templates)
- Add layered file-watch filtering to avoid triggering rebuilds on build artifacts:
  1. Always-ignore dirs (`target/`, `bin/`, `obj/`, `node_modules/`, etc.)
  2. Always-watch exceptions (`.env.local`, `spacetime.*.local.json`)
  3. `.gitignore` rules from global, project, and module levels
- Suppress init's generic "Next steps" message when called from `spacetime dev`, print a dev-appropriate `cd` hint instead when a new project is created in a subdirectory
- Update 12 quickstart docs to tell users to open a new terminal and `cd my-spacetime-app` before running CLI commands

## Test plan
- [x] `cargo build -p spacetimedb-cli` compiles successfully
- [ ] `spacetime dev --template basic-cs` no longer errors with "Input watch path is neither a file nor a directory"
- [ ] Modifying a `.cs` file in `spacetimedb/` triggers a rebuild
- [ ] Build artifacts in `obj/`/`bin/` do not trigger rebuilds
- [ ] Rust projects (`spacetimedb/src/` exists) continue to work as before
- [ ] `spacetime dev --template basic-cs` from a non-project directory prints the "Tip: cd" hint, not the generic "Next steps"
- [ ] `spacetime init` standalone still prints "Next steps" as before

### DIFF
--- a/crates/cli/src/subcommands/dev.rs
+++ b/crates/cli/src/subcommands/dev.rs
@@ -374,17 +374,22 @@ pub async fn exec(mut config: Config, args: &ArgMatches) -> Result<(), anyhow::E
             module_bindings_dir = canonical_created_path.join(module_bindings_path);
             project_dir = canonical_created_path.clone();
 
-            // If the project was created in a subdirectory, hint the user to cd into it.
+            // If the project was created in a subdirectory, hint the user to cd into it
+            // and show useful CLI commands they can run from there.
             let current_dir = std::env::current_dir().context("Failed to get current directory")?;
             if canonical_created_path != current_dir {
                 let rel_path = canonical_created_path
                     .strip_prefix(&current_dir)
                     .unwrap_or(&canonical_created_path);
                 println!(
-                    "\n{} To interact with your database, open a new terminal and run:\n  cd ./{}\n",
+                    "\n{} To interact with your database, open a new terminal and run:",
                     "Tip:".yellow().bold(),
-                    rel_path.display()
                 );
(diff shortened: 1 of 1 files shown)


bfops:
This change seems good to me, two questions:

1) does the test plan still need to be tested?

2) where's the part where we stop watching the `src/` subdir? This seems to only add the ignoring?

cloutiertyler:
> This change seems good to me, two questions:
> 
>     1. does the test plan still need to be tested?
> 
>     2. where's the part where we stop watching the `src/` subdir? This seems to only add the ignoring?

@bfops 

1. as of right now, yes.
2. it actually was already made on master. here's the merge commit: https://github.com/clockworklabs/SpacetimeDB/pull/4317/commits/b87a693677807605cd3c3dd58315136dd135e25e

This spacetime.json PR already [changes the dir](https://github.com/clockworklabs/SpacetimeDB/blame/0cb23814d3e32b4942188a70ecfa45ebbb414776/crates/cli/src/subcommands/dev.rs#L1209) but it does not properly ignore files. That was not the case on master when I first made this PR as evidenced by the merge.

jdetter (approved):
LGTM as well

PR #4323, opened 2026-02-17
Add `accessor_name` field in `ModuleDef`

Shubham8287:
# Description of Changes
Update Schema Defs to add `accessor_name` 

Look at `crates/schema/src/def.rs`.
All other file updates are due to changing existing field names.

NA

1

### DIFF
--- a/crates/schema/src/def.rs
+++ b/crates/schema/src/def.rs
@@ -1010,7 +1010,7 @@ impl From<DirectAlgorithm> for IndexAlgorithm {
 #[derive(Debug, Clone, Eq, PartialEq)]
 #[non_exhaustive]
 pub struct ColumnDef {
-    /// The name of the column as stored in
+    /// The name of the column.
     /// Unique within the containing `TableDef`, but
     /// NOT within the containing `ModuleDef`.
     pub name: Identifier,


crates/schema/src/def.rs
```diff
@@ -603,11 +603,13 @@ pub struct TableDef {
...
     pub name: Identifier,
-
-    /// For V9, it is same as `name`.
-    /// in V10, this is the name of index used inside the module.
+    /// The table identifier as defined in the module source.
+    ///
+    /// All the codegens should use `accessor_name` to derive the table identifier for clients.
+    /// However, the `name` field should still be used when communicating with the
+    /// database over the network, since the database does not necessarily know
+    /// about accessor names.
```
cloutiertyler, line 611: Is this accurate now with the aliases?

crates/schema/src/def.rs
```diff
@@ -1007,13 +1010,14 @@ impl From<DirectAlgorithm> for IndexAlgorithm {
 #[derive(Debug, Clone, Eq, PartialEq)]
 #[non_exhaustive]
 pub struct ColumnDef {
-    /// The name of the column.
+    /// The name of the column as stored in
```
cloutiertyler: Looks like this got cut off.

crates/schema/src/def.rs
Shubham8287: Yes, I deliberately chose not to mention the aliasing of the accessor name here. I felt it could be confusing as websocket format does not work with alias yet.

Def's may later have separate field for alias, what we are doing right now is more like a work-around.

cloutiertyler: Sounds good.

PR #4328, opened 2026-02-17
Pinning C++ and Unreal to 1.12.0

JasonAtClockwork:
# Description of Changes
A few small changes to pin the version:
- Updated the CMakeLists back from 2.0.0 to 1.12.0
- Updated the Unreal Blackholio tutorial documentation to target the 1.12.0 branch and not master
- Updated the Unreal Blackholio tutorial to call out 2.0 is coming soon
- Disabled the upgrade-version tool code path for C++ upgrades, but left logic in place for the near future

N/A

1 - Docs change, config change, and blocking update path for C++

# Testing

- [x] Reviewed the docs
- [x] Tested update command locally
- [x] Tested `spacetime init` for `basic-cpp` and successfully built

### DIFF
--- a/tools/upgrade-version/src/main.rs
+++ b/tools/upgrade-version/src/main.rs
@@ -367,8 +367,16 @@ fn main() -> anyhow::Result<()> {
         )?;
     }
     if matches.get_flag("cpp") || matches.get_flag("all") {
-        rewrite_cmake_version_inplace("crates/bindings-cpp/CMakeLists.txt", &full_version)?;
-        rewrite_cmake_version_inplace("templates/basic-cpp/spacetimedb/CMakeLists.txt", &full_version)?;
+        // TODO(26-02-17 jlarabie): Keep C++ pinned to 1.12.x for the 2.0 release train.
+        // Re-enable this codepath when C++ 2.0 is ready and we're intentionally
+        // moving the C++ SDK/template versions forward.
+        return Ok(());
+
+        #[allow(unreachable_code)]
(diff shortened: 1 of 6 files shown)


tools/upgrade-version/src/main.rs
```diff
@@ -367,8 +367,16 @@ fn main() -> anyhow::Result<()> {
         )?;
     }
     if matches.get_flag("cpp") || matches.get_flag("all") {
-        rewrite_cmake_version_inplace("crates/bindings-cpp/CMakeLists.txt", &full_version)?;
-        rewrite_cmake_version_inplace("templates/basic-cpp/spacetimedb/CMakeLists.txt", &full_version)?;
+        // TODO(26-02-17 jlarabie): Keep C++ pinned to 1.12.x for the 2.0 release train.
+        // Re-enable this codepath when C++ 2.0 is ready and we're intentionally
+        // moving the C++ SDK/template versions forward.
```
bfops, line 372: can you print a warning or something too in this case, so that it's clear to the person running this? Just something like
```
Warning: Not bumping C++ version. See inline comment for details.
```

bfops (approved):
my code-owned file looks good

tools/upgrade-version/src/main.rs
JasonAtClockwork: I should have considered that! It's been added.

jdetter (approved):
LGTM as well 👍

PR #4330, opened 2026-02-17
Improve `spacetime build` for typescript

coolreader18:
# Description of Changes

This fixes the confusing error encountered by @gefjon:

```
$ spacetime publish -s local --module-path spacetimedb --no-config event-table-test-ts
Rolldown warning: Could not resolve 'spacetimedb/server' in spacetimedb/src/index.ts
Error: Your module doesn't import the `spacetimedb/server` package at all - this is likely a mistake, as your module will not be able to interface with the SpacetimeDB host.
```
After:
```
[UNRESOLVED_IMPORT] Error: Could not resolve 'spacetimedb/server' in src/index.ts
   ╭─[ src/index.ts:1:47 ]
   │
 1 │ import { schema, table, t, SenderError } from 'spacetimedb/server';
   │                                               ──────────┬─────────  
   │                                                         ╰─────────── Module not found.
   │ 
   │ Help: You may have forgotten to install dependencies with your package manager.
───╯
```

And also runs `tsc`, so that modules with type errors aren't silently published. Perhaps we want a `--skip-tsc` so that people can run anyway and get runtime errors?

2

# Testing

- [x] Ensured that tsc and rolldown errors are properly emitted.

### DIFF
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -74,3 +74,14 @@ pub async fn exec_subcommand(
     }
     .map(|()| ExitCode::SUCCESS)
 }
+
+/// An error type indicating that the process should exit silently with the
+/// given `ExitCode`.
+#[derive(thiserror::Error, Debug)]
+#[error("exit with {0:?}")]
+pub struct ExitWithCode(pub ExitCode);
+
+impl ExitWithCode {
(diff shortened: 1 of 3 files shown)


cloutiertyler (approved):
Beautiful! This is a great fix.

PR #4332, opened 2026-02-18
`spacetime.json` related fixes and improvements

drogus:
# Description of Changes

This PR fixes a bunch of issues and brings some improvements related to `spacetime.json` config file

-

3

This PR is changing several places that are called during many of the CLI commands.

# Testing

- [x] Added some automated tests
- [ ] Tested locally


crates/cli/src/subcommands/dev.rs
```diff
@@ -487,7 +584,16 @@ pub async fn exec(mut config: Config, args: &ArgMatches) -> Result<(), anyhow::E
 
     // Safety prompt: warn if publishing from spacetime.json (not a dev-specific config)
     if let Some(ref lc) = loaded_config {
-        if !lc.has_dev_file && !force {
+        // Treat local overrides as dev-safe to avoid warning when per-user config is present.
+        // TODO: Should this also accept other env local files (for example: spacetime.staging.local.json)?
```
cloutiertyler, line 615: I think it's probably best as written, but we can see when people actually start using it

crates/cli/src/subcommands/dev.rs
```diff
@@ -870,26 +927,21 @@ async fn generate_build_and_publish(
...
-            "publish".to_string(),
-            db_name.to_string(),
-            "--module-path".to_string(),
-            module_path_str.to_string(),
-            "--yes".to_string(),
-            format!("--delete-data={}", clear_flag),
-        ];
+        let mut publish_entry = HashMap::new();
+        publish_entry.insert("database".to_string(), json!(db_name));
+        publish_entry.insert("module-path".to_string(), json!(module_path_str));
```
cloutiertyler, line 959: Yeah this is better

crates/cli/src/subcommands/dev.rs
```diff
@@ -1255,6 +1312,66 @@ fn detect_and_save_client_command(project_dir: &Path, existing_config: Option<Sp
...
+
+    let mut config = SpacetimeConfig::default();
+    config
+        .additional_fields
+        .insert("server".to_string(), json!("maincloud"));
+
+    if project_dir.join("spacetimedb").is_dir() {
+        config
+            .additional_fields
+            .insert("module-path".to_string(), json!("./spacetimedb"));
```
cloutiertyler, line 1356: I guess we've decided to be explicit about this even though it's already the default? No problem with it, just curious if it's deliberate.

crates/cli/src/subcommands/dev.rs
```diff
@@ -1255,6 +1312,66 @@ fn detect_and_save_client_command(project_dir: &Path, existing_config: Option<Sp
...
+
+    let local_config_path = project_dir.join("spacetime.local.json");
+    if local_config_path.exists() {
+        return Ok(None);
+    }
+
+    let mut local_config = SpacetimeConfig::default();
+    local_config
+        .additional_fields
+        .insert("database".to_string(), json!(database_name));
```
cloutiertyler, line 1388: 👍

crates/cli/src/subcommands/generate.rs
```diff
@@ -277,116 +277,61 @@ pub async fn exec(config: Config, args: &clap::ArgMatches) -> anyhow::Result<()>
...
-                        match lang {
-                            Language::Rust => "rust",
-                            Language::Csharp => "csharp",
-                            Language::TypeScript => "typescript",
-                            _ => unreachable!(),
-                        }
-                    ));
-                }
-            }
+            Language::Rust | Language::Csharp | Language::TypeScript => {}
```
cloutiertyler: This is no longer required because we now have a default location?

crates/cli/src/subcommands/generate.rs
```diff
@@ -396,22 +341,128 @@ pub async fn exec_ex(
...
+        });
+    }
+
+    Ok(runs)
+}
+
+fn detect_default_language(module_path: &Path) -> anyhow::Result<Language> {
+    if !module_path.is_dir() {
+        anyhow::bail!(
+            "Could not auto-detect client language: module path '{}' does not exist. \
```
cloutiertyler:
```suggestion
            "Could not auto-detect module language: module path '{}' does not exist. \
```

right?

crates/cli/src/subcommands/generate.rs
```diff
@@ -396,22 +341,128 @@ pub async fn exec_ex(
...
+             If this is not correct, pass --lang or add a generate target in spacetime.json. \
+             You can also pass --module-path to specify the module location.",
+            module_path.display()
+        );
+    }
+
+    let module_lang = detect_module_language(module_path).map_err(|err| {
+        anyhow::anyhow!(
+            "Could not auto-detect client language from module '{}': {}. \
```
cloutiertyler, line 375:
```suggestion
            "Could not auto-detect module language from module '{}': {}. \
```

crates/cli/src/subcommands/generate.rs
cloutiertyler: Oh, hmm, this is detecting the client language based on the module language? Maybe we should have a bit more in this error explaining what we're trying to do here because it's just a bail. You must provide a `--lang` because we can't determine your client language based on your server language. I assume this is only used in the case where we can't detect the language based on the client directory itself?

crates/cli/src/subcommands/publish.rs
```diff
@@ -502,11 +521,17 @@ pub async fn exec_with_options(
...
-                if let Some(domain) = domain {
+                if let Some(ref domain) = domain {
                     println!("{op} database with name: {domain}, identity: {database_identity}");
                 } else {
                     println!("{op} database with identity: {database_identity}");
                 }
+
+                if is_maincloud_host(&database_host) {
+                    if let Some(domain) = domain.as_ref() {
+                        println!("Dashboard: https://spacetimedb.com/{}", domain.as_ref());
```
cloutiertyler, line 532: 👍

cloutiertyler (approved):
Yeah this looks really good. I left a few comment to be addressed, but overall I think this is a big improvement.

crates/cli/src/subcommands/dev.rs
drogus: I felt like it's more consistent to always save it, but that's a good point, we could skip it just how we skip `generate` configs for standardized locations. I don't really have a strong preference

crates/cli/src/subcommands/generate.rs
drogus: I'm not sure - for Rust/Csharp/Typescript it already does nothing, so this match only matters for Unreal

crates/cli/src/subcommands/generate.rs
drogus: Ah, this message is indeed wrong, this check is not even checking the module language it's just checking if it exists. I'll move to a different place cause this function is for detecting the client language, not the module language, and I'll fix the message

drogus: https://github.com/clockworklabs/SpacetimeDB/pull/4332/changes/1baf3bf3760f5d50e2c9abf59850c8b587a537a4

crates/cli/src/subcommands/dev.rs
cloutiertyler: It's fine how you have it.

PR #4335, opened 2026-02-18
[Docs] Revisions to Chat App doc for 2.0 release

rekhoff:
# Description of Changes

Updates the Chat App tutorial `docs/docs/00100-intro/00300-tutorials/00100-chat-app.md`

* TypeScript (React) code snippet fixes
  * Updated the initial `App.tsx` snippet to use `import type * as Types from './module_bindings/types'` and type `systemMessages` as `Types.Message[]`, matching the current template-generated bindings.
  * Updated the placeholder `onlineUsers` / `offlineUsers` types in the initial snippet to `Types.User[]`.
  * Updated the `main.tsx` connection snippet with the template behavior by sourcing `HOST` / `DB_NAME` from `import.meta.env.VITE_SPACETIMEDB_HOST` / `VITE_SPACETIMEDB_DB_NAME` if available.
  * Removed an TypeScript example that referenced reducer event wiring (`conn.reducers.onSendMessage(...)`) that doesn’t exist in the current TS generated bindings.
* Updated the follow-up instruction text so the “replace `const onlineUsers: ...`” and “replace `const offlineUsers: ...`” guidance matches the updated `Types.User[]` placeholders in the earlier snippet.
* The tutorial section says arguments are JSON; the CLI will auto-quote string args based on reducer parameter types, so the simple form `send_message 'Hello, World!'` remains valid and is kept.
* Added C# Client SDK step to update `.csproj` file with `net8.0` and a filter for code in the `spacetimedb` (server module) subdirectory.

N/A

1

# Testing

- [x] Performed documented operations locally.

### DIFF
--- a/docs/docs/00100-intro/00300-tutorials/00100-chat-app.md
+++ b/docs/docs/00100-intro/00300-tutorials/00100-chat-app.md
@@ -840,9 +840,36 @@ spacetime logs --server local quickstart-chat
 
 SpacetimeDB supports a subset of SQL so you can query your data:
 
+<Tabs groupId="lang">
+<TabItem value="typescript" label="TypeScript">
+
 ```bash
 spacetime sql --server local quickstart-chat "SELECT * FROM message"
 ```
+
+</TabItem>
+<TabItem value="csharp" label="C#">
+
+```bash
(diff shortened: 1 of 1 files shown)


JasonAtClockwork (approved):
Looks good to me

PR #4338, opened 2026-02-18
commitlog: Improve `committed_meta`

kim:
- Extends `commit::Metadata` to include the checksum
- Extends `segment::Metadata` to include `Some(commit::Metadata)`
  containing the last commit in the segment (if there is one)
- Changes `committed_meta` to:
  - ignore empty segments at the end of the log
  - try harder to provide useful metadata, even if only a prefix of the
    latest segment is readable

This is allows to eliminate remaining `Commitlog::open` calls with the
purpose of querying the latest commit (offset). `Commitlog::open`
creates an empty segment if the tail of the log is corrupt, which is a
non-obvious side-effect that can be confusing when debugging.

It also allows to eliminate uses where the `commits_from` iterator is
used to find the latest full commit. The `Commits` iterator requires the
caller to handle the case of a corrupted commit at the end of the log,
by advancing the iterator once more after it has yielded an error in
order to check that it is exhausted, and then deciding whether to ignore
the error. This is easy to forget.

`committed_meta` now just does the right thing, preserving information
about tail corruption for when that's useful.


crates/commitlog/src/repo/mod.rs
```diff
@@ -314,6 +315,32 @@ pub fn open_segment_reader<R: Repo>(
...
+/// [segment::Header::LEN].
+///
+/// Returns the segment offset alongside the raw [Repo::SegmentReader],
+/// or `None` if the repo is empty (i.e. contains no non-empty segment).
+pub(crate) fn open_newest_non_empty_segment<R: Repo>(repo: &R) -> io::Result<Option<(u64, R::SegmentReader)>> {
+    let mut segments = repo.existing_offsets()?;
+
+    let mut newest;
+    let mut reader;
+    loop {
```
Shubham8287: Could there be more than one empty segments in a repo? If no, should this code error out accordingly instead of handling series of empty commitlogs?

crates/commitlog/src/commitlog.rs
```diff
@@ -355,7 +355,77 @@ impl<R: Repo, T> Drop for Generic<R, T> {
...
+        metadata
+    }
+
+    fn extract(repo: impl Repo) -> io::Result<Option<Self>> {
+        let Some((offset, mut storage)) = repo::open_newest_non_empty_segment(&repo)? else {
+            return Ok(None);
+        };
+        let offset_index = repo.get_offset_index(offset).ok();
+        match segment::Metadata::extract(offset, &mut storage, offset_index.as_ref()) {
+            // TODO: Should this be an `assert!` instead?
```
Shubham8287: yes, I will prefet assert or atleast an error here, considering we already requested a non empty segment above.

crates/commitlog/src/repo/mod.rs
kim: I think this really depends on the caller.
I restructured the code to map the case of >1 empty segments to the `CommittedMeta::Partial` variant, so the caller can decide what to do.

PR #4341, opened 2026-02-18
Fix CI error stemming from `cargo update`

coolreader18:
# Description of Changes

Now we only `cargo update` the dependencies of the module, not of `spacetimedb-cli`.

### DIFF
--- a/tools/ci/src/main.rs
+++ b/tools/ci/src/main.rs
@@ -396,6 +396,7 @@ fn main() -> Result<()> {
 
         Some(CiCmd::WasmBindings) => {
             cmd!("cargo", "test", "-p", "spacetimedb-codegen").run()?;
+            // Pre-build the CLI so that it _doesn't_ get `cargo update`d, since that may break the build.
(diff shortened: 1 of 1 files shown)


tools/ci/src/main.rs
```diff
@@ -396,24 +396,21 @@ fn main() -> Result<()> {
 
         Some(CiCmd::WasmBindings) => {
             cmd!("cargo", "test", "-p", "spacetimedb-codegen").run()?;
+            cmd!("cargo", "build", "-p", "spacetimedb-cli").run()?;
```
bfops, line 400:
```suggestion
            // Pre-build the CLI so that it _doesn't_ get `cargo update`d, since that may break the build.
            cmd!("cargo", "build", "-p", "spacetimedb-cli").run()?;
```

kim:
It's not really fixing the issue, as modules also pull in native-tls (why?).
The bad version 0.2.17 got yanked from crates.io a few hours ago, though, so we may just need to re-run the failing jobs.

bfops:
CI does seem to be passing without this now. I'm a fan of still merging this since the intent is correct.

PR #4344, opened 2026-02-18
Rust docs updated for `name` to `accessor`

JasonAtClockwork:
# Description of Changes

- Updated all markdown files that call out `#[spacetimedb::table(name` to `#[spacetimedb::table(accessor`
- Updated the LLM details on the same change

N/A

1 - Small rework in documentation to use the new 

# Testing

- [x] Did a small double check on name vs accessor locally
- [x] Codex tested all the LLM answers

### DIFF
--- a/crates/bindings/README.md
+++ b/crates/bindings/README.md
@@ -115,7 +115,7 @@ The project's `lib.rs` will contain the following skeleton:
 # #[cfg(target_arch = "wasm32")] mod demo {
 use spacetimedb::{ReducerContext, Table};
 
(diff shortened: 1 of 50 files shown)


rekhoff (approved):
All changes look reasonable to me

cloutiertyler (approved):
This changes to the answers look correct to me.

Boegie19:
@cloutiertyler seems like update-llm-benchmark did not run here and I think this change could impact the LLM scores.

PR #4348, opened 2026-02-18
Docs - add shared C++ version notice inside first C++ tab across docs

JasonAtClockwork:
# Description of Changes
- Added a shared component with a message for C++ on the version pinning
- Updated every doc that has a C++ tab to include the new message

N/A

2 - Small document change but used a shared component

# Testing
- [x] Tested locally with pnpm dev
- [x] Hosted on my personal site to double check the shared component

### DIFF
--- a/docs/src/components/CppModuleVersionNotice.tsx
+++ b/docs/src/components/CppModuleVersionNotice.tsx
@@ -5,9 +5,9 @@ export function CppModuleVersionNotice(): JSX.Element {
   return (
     <Admonition type="important" title="C++ Modules and SpacetimeDB 2.0">
       <p>
-        SpacetimeDB <code>2.0</code> is available for other module languages, but C++ server modules
-        are currently pinned to <code>v1.12.0</code>. If you are following the C++ tab in this guide,
-        use the <code>v1.12.0</code> release track for now.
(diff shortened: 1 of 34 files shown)


docs/src/components/CppModuleVersionNotice.tsx
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+import Admonition from "@theme/Admonition";
+
+export function CppModuleVersionNotice(): JSX.Element {
+  return (
+    <Admonition type="important" title="C++ Modules and SpacetimeDB 2.0">
+      <p>
+        SpacetimeDB <code>2.0</code> is available for other module languages, but C++ server modules
+        are currently pinned to <code>v1.12.0</code>. If you are following the C++ tab in this guide,
+        use the <code>v1.12.0</code> release track for now.
```
cloutiertyler: Perhaps a comment about the fact that C++ 2.0 is coming? We may also want to throw in a mention of BETA here so people know that C++ is subject to change.

JasonAtClockwork: Updated to have both beta and 2.0 coming soon

rekhoff (approved):
Looks good to me. It appears to render correctly, and it's great to have the notice in a centralized location so we maintain consistent verbage.

PR #4349, opened 2026-02-19
Update to generated C# template package version

rekhoff:
# Description of Issue

When you run `spacetime dev --template basic-cs` and give it the default inputs, the `.csproj` files it generates use the package versions of `1.*` rather than `2.*`.
The relative path locations are:
* `.\my-spacetime-app\client.csproj`
* `.\my-spacetime-app\spacetimedb\StdbModule.csproj`
This is deceptive because in `.\SpacetimeDB\templates\basic-cs`, each file has a `2.0.*` for the package versions.

This issue results in an error when trying to use a C# template, as specific 2.0 features are not properly detected.

This happens because:
1. When you run `spacetime dev` outside an existing project, the dev subcommand notices the missing spacetimedb/ directory and internally calls `init::exec_with_options` with your requested template (`basic-cs`) to scaffold a project first. That logic lives in `crates/cli/src/subcommands/dev.rs` and triggers whenever `dev` is asked to initialize a template before continuing.
2. The template assets under `templates/basic-cs` indeed still contain `2.0.*` package references, but after the embedded files are copied, the initializer runs `update_csproj_client_to_nuget` and `update_csproj_server_to_nuget`. These functions rewrite the first `.csproj` they find in the client/server directories to depend on NuGet packages rather than local project references.
3. Both helpers call `get_spacetimedb_csharp_runtime_version()` / `get_spacetimedb_csharp_clientsdk_version()`, and those helpers currently hard-code `"1.*"` as the version string.

That’s why your freshly generated `client.csproj` and `spacetimedb/StdbModule.csproj` end up with `1.*` package versions even though the template files started at `2.0.*`

# Description of Changes

Updates the `get_spacetimedb_csharp_runtime_version()` and `get_spacetimedb_csharp_clientsdk_version()` functions to return `2.*`

No API/ABI changes

1 - Low

# Testing

- [X] Ran `spacetime dev --template basic-cs` locally and confirmed `.csproj` files generate with `2.*` package versions.


JasonAtClockwork (approved):
Confirmed locally that now the C# project files receive 2.0 for the runtime and sdk versions with `spacetime dev`

bfops (approved):
Looks like an improvement to me!

bfops:
oh it's already approved. Should we merge this?

PR #4351, opened 2026-02-19
Fix publishing and generating from subdirs if config is present

drogus:
# Description of Changes

When config is present and no explicit `module-path` has been passed, we should be resolving paths from the config and defaults relative to the config file

2

# Testing

- [x] Tested locally


jdetter (approved):
This does fix the publishing issue. `spacetime dev` is still broken but we can fix that in a separate PR perhaps.

PR #4359, opened 2026-02-19
Make Rust test clients listen for reducer errors

gefjon:
# Description of Changes

Prior to this commit, if a reducer call failed during a test in the Rust SDK test suite, the client would hang, as it generally didn't register callbacks when invoking reducers. This could mask some bugs, causing the tests to fail with TIMEOUT rather than a useful message. For example, we relied on consistency between all of the test modules, which was broken by our not having updated the C++ bindings library to use the new case-conversion scheme. This meant that the client was sending reducer calls with incorrect reducer names to at least one database per test run, which would result in a `Panic` status or an `Err(Err(_))` outcome.

With this commit, the SDK test clients are updated to inspect the exit status of reducers they invoke, and (except for intentional reducer failures) to panic the test client when encountering a reducer error or panic.

N/a

1

# Testing

- [x] Automated tests passed locally.
- [x] Intentionally broke reducer calls via the 2.0 WS API, causing them to skip the call and unconditionally respond with `Status::Panic`. Observed "loud" failures in the test suite with non-`TIMEOUT` output that looked more debuggable to me.

### DIFF
--- a/sdks/rust/tests/procedure-client/src/main.rs
+++ b/sdks/rust/tests/procedure-client/src/main.rs
@@ -352,7 +352,13 @@ fn exec_schedule_procedure() {
 
             subscribe_all_then(ctx, move |ctx| {
                 sub_applied_nothing_result(assert_all_tables_empty(ctx));
-                ctx.reducers.schedule_proc().unwrap();
+                ctx.reducers
+                    .schedule_proc_then(|_ctx, outcome| match outcome {
+                        Ok(Ok(())) => (),
+                        Ok(Err(msg)) => panic!("`schedule_proc` reducer returned error: {msg}"),
+                        Err(internal_error) => panic!("`schedule_proc` reducer panicked: {internal_error:?}"),
+                    })
+                    .unwrap();
             });
         }
     });
--- a/sdks/rust/tests/test-client/src/main.rs
+++ b/sdks/rust/tests/test-client/src/main.rs
@@ -9,6 +9,7 @@ use std::sync::{Arc, Barrier, Mutex};
 use module_bindings::*;
 
 use rand::RngCore;
+use spacetimedb_sdk::error::InternalError;
 use spacetimedb_sdk::TableWithPrimaryKey;
 use spacetimedb_sdk::{
     credentials, i256, u256, Compression, ConnectionId, DbConnectionBuilder, DbContext, Event, Identity, ReducerEvent,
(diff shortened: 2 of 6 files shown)


JasonAtClockwork (approved):
Looks good to me and my local test worked.

I merged this branch into Shub's casing branch then ran: 
`cargo test -p spacetimedb-sdk --test test -- --nocapture insert_primitive insert_primitives_as_strings caller_alice_receives_reducer_callback_but_not_bob insert_vec --test-threads=1`

This caused much cleaner errors and failed tests as exampled:
```
thread 'cpp::caller_alice_receives_reducer_callback_but_not_bob' (37316) panicked at crates\testing\src\sdk.rs:148:9:
(running): Error running "cargo run -- caller-alice-receives-reducer-callback-but-not-bob": exited with non-zero exit status exit code: 1. Output:
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.37s
     Running `D:\ClockworkLabs\branches\SpacetimeDB-shub-case\target\debug\test-client.exe caller-alice-receives-reducer-callback-but-not-bob`

thread '<unnamed>' (71856) panicked at sdks\rust\tests\test-client\src\main.rs:433:33:
Subscription errored: SubscriptionError { error: "no such table: `one_u_8`. If the table exists, it may be marked private., executing: `SELECT * FROM one_u_8;`" }
error: process didn't exit successfully: `D:\ClockworkLabs\branches\SpacetimeDB-shub-case\target\debug\test-client.exe caller-alice-receives-reducer-callback-but-not-bob` (exit code: 1)
```

```
thread 'typescript::insert_primitive' (72576) panicked at crates\testing\src\sdk.rs:148:9:
(running): Error running "cargo run -- insert-primitive": exited with non-zero exit status exit code: 1. Output:
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.36s
     Running `D:\ClockworkLabs\branches\SpacetimeDB-shub-case\target\debug\test-client.exe insert-primitive`
[2026-02-19T22:25:30Z DEBUG spacetimedb_sdk::subscription] on_applied called for query QuerySetId { id: 1 }

thread '<unnamed>' (64928) panicked at sdks\rust\tests\test-client\src\simple_test_table.rs:60:1:
Insert reducer panicked: InternalError { message: "no such reducer", cause: None }
error: process didn't exit successfully: `D:\ClockworkLabs\branches\SpacetimeDB-shub-case\target\debug\test-client.exe insert-primitive` (exit code: 1)
```
Ending with:
```
failures:
    cpp::caller_alice_receives_reducer_callback_but_not_bob
    cpp::insert_primitive
    cpp::insert_primitives_as_strings
    cpp::insert_vec
    csharp::caller_alice_receives_reducer_callback_but_not_bob
    csharp::insert_primitive
    csharp::insert_vec
    rust::caller_alice_receives_reducer_callback_but_not_bob
    rust::insert_primitive
    rust::insert_vec
    typescript::caller_alice_receives_reducer_callback_but_not_bob
    typescript::insert_primitive
    typescript::insert_vec
```

PR #4369, opened 2026-02-20
rust sdk: ConnectionId override via builder

kim:
Instead of a global, provide a builder method to override the connection ID.

This should allow using multiple connections per process.

1

# Testing

Trivial change.


kim:
I think we want to use JWT claims for the purpose, but it's unclear to me how the migration path looks like.

PR #4371, opened 2026-02-20
typescript: canonical naming for reducer and procedure.

Shubham8287:
# Description of Changes

Canonical naming for reducer and procedure.
It shouldn't have the `name` field there but if it's there it should be used as canonical name.

Yes

1

### DIFF
--- a/crates/codegen/src/rust.rs
+++ b/crates/codegen/src/rust.rs
@@ -735,7 +735,7 @@ pub struct {cols_ix} {{"
             .iter()
             .find(|col| col.col_id == cols.as_singleton().expect("singleton column"))
             .unwrap();
-        let field_name = column.accessor_name.deref();
+        let field_name = column.accessor_name.deref().to_case(Case::Snake);
         let field_type = type_name(module, &column.ty_for_generate);
 
         writeln!(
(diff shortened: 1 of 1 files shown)


crates/bindings-typescript/src/server/reducers.ts
```diff
@@ -56,27 +54,28 @@ export function makeReducerExport<
...
-  ctx.defineFunction(name);
+  ctx.defineFunction(exportName);
 
   if (!(params instanceof RowBuilder)) {
     params = new RowBuilder(params);
   }
 
   if (params.typeName === undefined) {
-    params.typeName = toPascalCase(name);
+    params.typeName = toPascalCase(exportName);
```
cloutiertyler, line 70: Is this conversion also happening on the host side?

Shubham8287: yes, types are getting converted to PascalCase on Host.
Here, we are assigning a type name only if not provided, I guess for local module access.

cloutiertyler: I see, interesting. Yeah I suppose this is correct. The user doesn't need to provide a name for their args type, so this is fine.

PR #4375, opened 2026-02-20
Don't shadow Math.random() in typescript

coolreader18:
# Description of Changes

This also affects typescript clients, which wasn't the intention.

1

# Testing

- [x] Typescript client code that uses Math.random is no longer broken.


gefjon:
What happens to module code that writes `Math.random()` now?

coolreader18:
Typescript won't give an error, but it will still fail at runtime.

PR #4384, opened 2026-02-20
Update `bump-versions` to support prereleases

bfops:
# Description of Changes

`cargo bump-versions` works properly with prerelease versions. Before it would quietly drop the `-foo` or `+foo` suffix.

None

1

# Testing

- [x] Ran the script and bumped the versions for the 2.0.0 prerelease


bfops:
releasing this

Issue #4093, opened 2026-01-22
C# SDK allocates ~131KB per database read operation regardless of payload size


rekhoff:
Discord user `tfritzy` reported the following:

> I looked into this because my server was stuttering every 10 seconds for 500-1000ms, which I was not able to fix after extensively tracking down garbage created by my code. I eventually did find a solution in manually invoking garbage collection in each tick. I've included a repro project as a zip.
> 
> ⚠️ All below is ai generated
> 
> Every database read operation (`Find()`, `Filter()` iteration, `Iter()` iteration) allocates approximately 131KB of memory regardless of the actual payload size. This fixed overhead makes high-frequency game loops impractical due to excessive GC pressure.
> 
> Test Results
> ```
> Read operations (the problem):
> Find() TinyRecord (8 bytes payload)     → 131,816 bytes allocated
> Find() MediumRecord (~50 bytes payload) → 131,928 bytes allocated
> Find() LargeRecord (1 KB payload)       → 133,872 bytes allocated
> Find() LargeRecord (10 KB payload)      → 152,304 bytes allocated
> Find() LargeRecord (100 KB payload)     → 336,624 bytes allocated
> Iter() 10 rows (80 bytes total)         → 131,632 bytes allocated
> Filter() iterate 20 rows               → 133,728 bytes allocated
> Filter() iterate 100 rows              → 144,688 bytes allocated
> 10x consecutive Find() (80 bytes)       → 1,318,160 bytes allocated
> ```
> 
> Observations
> • ~131KB fixed base overhead on every read, independent of payload
> • Payload size adds on top: total ≈ 131KB + payload_size
> • Filter() and Iter() calls are cheap (72-160 bytes), but iterating results triggers the ~131KB allocation
> • The overhead does not amortize—10 Find() calls = 10× the overhead
> 
> Minimal repro:
> ```csharp
> [Reducer]
> public static void MeasureAllocations(ReducerContext ctx)
> {
>     long before = GC.GetAllocatedBytesForCurrentThread();
>     var tiny = ctx.Db.TinyRecord.Id.Find(1); // 8 byte payload
>     long after = GC.GetAllocatedBytesForCurrentThread();
>     Log.Info($"TinyRecord Find: {after - before} bytes allocated");
>     // Output: TinyRecord Find: 131816 bytes allocated
> }
> ```

[...]

PR #4385, opened 2026-02-20
[C#] Update RawTableIterBase.Enumerator to use `ArrayPool` for buffer

rekhoff:
This is the implementation of a fix for #4093 

# Description of Changes

* Updated `RawTableIterBase.Enumerator` to rent its scratch buffer from `ArrayPool<byte>.Shared` (with dynamic re-rent on `BUFFER_TOO_SMALL`) and return it on dispose, so iterating rows no longer allocates a fresh `byte[]` per step.
* The enumerator still exposes the row bytes via `ArraySegment<byte> Current`, but now the underlying storage is reused across iterations rather than recreated each time.

Testing results from `master`:
```
allocated_bytes=14000
elapsed_ticks=1521829
sum=1249975000
row_count=50000
Find() TinyRecord (8 bytes payload) -> 132208 bytes allocated
Find() MediumRecord (~50 bytes payload) -> 132440 bytes allocated
Find() LargeRecord (1 KB payload) -> 134528 bytes allocated
Find() LargeRecord (10 KB payload) -> 152408 bytes allocated
Find() LargeRecord (100 KB payload) -> 336728 bytes allocated
Iter() 10 rows -> 131896 bytes allocated
Filter() iterate 20 rows -> 133288 bytes allocated
Filter() iterate 100 rows -> 135976 bytes allocated
10x consecutive Find() (TinyRecord) -> 1319120 bytes allocated
```

Testing results with this fix:
```
allocated_bytes=14000
elapsed_ticks=1504949
sum=1249975000
row_count=50000
Find() TinyRecord (8 bytes payload) -> 1096 bytes allocated
Find() MediumRecord (~50 bytes payload) -> 1280 bytes allocated
Find() LargeRecord (1 KB payload) -> 4464 bytes allocated
Find() LargeRecord (10 KB payload) -> 27464 bytes allocated
Find() LargeRecord (100 KB payload) -> 234312 bytes allocated
Iter() 10 rows -> 680 bytes allocated
Filter() iterate 20 rows -> 1872 bytes allocated
Filter() iterate 100 rows -> 3280 bytes allocated
10x consecutive Find() (TinyRecord) -> 8000 bytes allocated
```

No API or ABI changes

2 - low/moderate:
Touches the hot-path iterator that every `Iter`/`Find`/`Filter` call uses

# Testing

- [X] Compiled CLI and ran regression tests locally
- [X] Verified iterator-based harness runs (client + module reducers) on both `master` and this branch, confirming allocations drop from ~131 KB per read to payload-scaled values.
- [X] Ensured no regressions in standard harness sanity checks (`row_count=50000`, `sum=1249975000`).


crates/bindings-csharp/Runtime/Internal/ITable.cs
```diff
@@ -66,6 +68,8 @@ public void Dispose()
                 FFI.row_iter_bsatn_close(handle);
                 handle = FFI.RowIter.INVALID;
             }
+
+            ArrayPool<byte>.Shared.Return(buffer);
```
JasonAtClockwork: This has a slight chance to be a problem if somehow Dispose() was called twice, with using the ArrayPool Shared it might be time to add a flag for internal tracking of disposed?

rekhoff: That a great point. I've updated the buffer to be nullable, and set it to null after the first time it's Disposed.  We could add a flag, if you still want, but this should prevent the issue.

JasonAtClockwork (approved):
Looks solid, we might want to add more testing around this in the future but my small local testing matches up.

PR #4391, opened 2026-02-22
C# smoketest for `IQuery` views

joshua-spacetime:
# Description of Changes

~~Updates Roslyn codegen to handle `IQuery` return type for views.~~

~~`IQuery` wasn't recognized before this change, which meant `.Build()` was still required for query builder views.~~

~~Note, we currently support the old `Query` return type which means we still support `.Build()`. I'm not sure if this was intended when `IQuery` was originally introduced, so I'm maintaining support for it until I can determine otherwise. cc @cloutiertyler.~~

This is now a test only change.

This patch introduces scaffolding for defining and running C# module smoketests. It also adds a new C# smoketest for  an`IQuery` view.

None

2

# Testing

A testing only change.

### DIFF
--- a/crates/smoketests/src/csharp.rs
+++ b/crates/smoketests/src/csharp.rs
@@ -1,0 +1,255 @@
+use anyhow::{anyhow, bail, Context, Result};
+use serde_json::Value;
+use std::fs;
+use std::path::{Path, PathBuf};
+use std::sync::OnceLock;
+
+const PACKAGE_PROJECTS: [(&str, &str); 2] = [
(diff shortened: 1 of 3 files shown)


joshua-spacetime:
`.Build()` is removed entirely in #4393 both in the module and in the client. This patch still is still beneficial though for it's test coverage and smoketest infra.

joshua-spacetime:
Removing the `release-2.0` label given #4393

rekhoff (approved):
Looks good to me. I've not ran the test locally, but given that it's a CI test, if it passes there that is sufficient. The test itself seems extremely strait forward.

PR #4392, opened 2026-02-22
[Docs] [C#] Update docs with `List<T>` returns and `IEnumerable<T>` tests

rekhoff:
# Description of Changes

* Updates `00500-views.md` with:
  * Changes returned value to a nullable value, to match the return type defintion.
  * Wraps sample code in `public partial class Module` so pasting it in to IDE will not return errors.
* Updates `00500-cheat-sheet.md` with:
  * Changes reducer names to not start with `On` to avoid error `STDB0010: Reducer method OnConnect starts with 'On', which is a reserved prefix.`
  * Adds `Accessor` values to views.
  * Updates `Filter()` returned value to use `.ToList()` and the return type to use `List<Player>` rather than `IEnumerable<Player>` to avoid error, due to needing a return type to be `Vec<T>` or `Option<T>`.
* Updated `Coddgen.Test` to include tests verifying current behavior of `IEnumerable<T>`, to aid in future update to allow a return type of `IEnumerable<T>` to be allowed and converted internally to the required type.

None

1 – documentation-only edits with no behavioral impact.

# Testing

- [X] Changed code blocks tested locally to ensure formatting resolved in IDE.


joshua-spacetime:
Is this change still applicable? #4393 removes `Query` entirely, right?

joshua-spacetime:
We should remove any references to `Query` in the docs and replace with `IQuery`. `llms.md` is one such place - see #4394. But I'm not sure if there are others.

docs/docs/00200-core-concepts/00100-databases/00500-cheat-sheet.md
```diff
@@ -624,7 +624,7 @@ public static IEnumerable<Player> TopPlayers(ViewContext ctx)
 // Perform a generic filter using the query builder.
 // Equivalent to `SELECT * FROM player WHERE score < 1000`.
 [SpacetimeDB.View(Public = true)]
-public static IQuery<Player> BottomPlayers(ViewContext ctx)
+public static Query<Player> BottomPlayers(ViewContext ctx)
 {
     return ctx.From.Player.Where(p => p.Score.Lt(1000));
```
JasonAtClockwork:
```suggestion
    return ctx.From.Player().Where(p => p.Score.Lt(1000));
```
The code in the documentation is incorrect for .Where(), and as mentioned by Joshua #4392 means all the IQuery => Query changes will need to be cleaned up.

rekhoff: `Query` has been removed. I will be re-testing this now that #4393 has merged. I suspect the code block you mention changes here will also need the `Accessor = "BottomPlayers"` added. Once I've done a pass, I'll incorporate this change and mode this PR into Ready for Review

JasonAtClockwork (approved):
Looks good to me!

crates/bindings-csharp/Codegen.Tests/fixtures/diag/snapshots/ExtraCompilationErrors.verified.txt
```diff
@@ -137,6 +160,29 @@ SpacetimeDB.Internal.Module.RegisterClientVisibilityFilter(global::Module.MY_FOU
       ]
     }
   },
+  {/*
+    {
+        return ctx.Db.Player.Identity.Filter(ctx.Sender);
+                                      ^^^^^^
+    }
+*/
+    Message: 'ReadOnlyUniqueIndex<PlayerReadOnly, Player, Identity, Identity.BSATN>.Filter(Identity)' is inaccessible due to its protection level,
```
joshua-spacetime: This doesn't look like the intended error message. Perhaps you meant `Find` instead of `Filter`?

crates/bindings-csharp/Codegen.Tests/fixtures/diag/snapshots/Module.verified.txt
```diff
@@ -355,6 +355,90 @@ public partial struct TestScheduleIssues
       }
     },
     {/*
+    [SpacetimeDB.View(Accessor = "view_def_ienumerable_return_from_iter", Public = true)]
+    public static IEnumerable<Player> ViewDefIEnumerableReturnFromIter(ViewContext ctx)
+                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
joshua-spacetime, line 360: This can happen in a follow up, but ideally the location identifier should point to the return type `IEnumerable`.

crates/bindings-csharp/Codegen.Tests/fixtures/diag/snapshots/Module.verified.txt
```diff
@@ -355,6 +355,90 @@ public partial struct TestScheduleIssues
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+    {
+^^^^^
+        return ctx.Db.Player.Iter();
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+    }
+^^^^^
+
+*/
+      Message: View 'ViewDefIEnumerableReturnFromIter' must return Vec<T> or Option<T>.,
```
joshua-spacetime, line 388: Again this can be done in a follow up, but we should probably use `List` and perhaps `Nullable` in place of `Vec` and `Option`? Is it called `Option` in C#?

crates/bindings-csharp/Codegen.Tests/fixtures/diag/snapshots/ExtraCompilationErrors.verified.txt
rekhoff: I split the diagnostics into two explicit cases now: one returning `IEnumerable` from `Iter()` and one from index `Filter(...)`, so the error message matches the actual source.

crates/bindings-csharp/Codegen.Tests/fixtures/diag/snapshots/Module.verified.txt
rekhoff: Created #4425

crates/bindings-csharp/Codegen.Tests/fixtures/diag/snapshots/Module.verified.txt
rekhoff: Created #4425

crates/bindings-csharp/Codegen.Tests/fixtures/diag/snapshots/ExtraCompilationErrors.verified.txt
joshua-spacetime: But why does this error message say that `Filter` is inaccessible.

rekhoff: Ah okay. So back to the `Find` versus `Filter`. `Find` returns a `T?` so wouldn't be useful on checking for `IEnumerable<T>`. `Filter` does return `IEnumerable<T>` but for unique index filters, it's a protected/internal API (because it would be redundant with `Find`). But taking that logic further, I really should just use s a non-unique index to check `Filter` since it can have multiple items returned.
I've pushed a new update that uses a non-unique index in the test, so the resulting error messages are more applicable.

joshua-spacetime: Thanks for the explanation!

joshua-spacetime (approved):

PR #4399, opened 2026-02-23
CI - Fail properly if `psql` failed to install

jdetter:
# Description of Changes

We had an issue where the install failed but did not fail the step:
```
Run choco install psql -y --no-progress
Chocolatey v2.6.0
Installing the following packages:
psql
By installing, you accept licenses for the packages.
Failed to fetch results from V2 feed at 'https://community.chocolatey.org/api/v2/Packages(Id='psql',Version='16.2.0')' with following message : Response status code does not indicate success: 504 (Gateway Time-out).
Need to add specific handling for exception type NuGetResolverInputException
Unable to find package 'psql'. Existing packages must be restored before performing an install or update.
```

None - this is just CI

None - this is just CI

# Testing

- [x] Windows smoketests are able to use `psql` properly

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -108,6 +108,7 @@ jobs:
           $PSNativeCommandUseErrorActionPreference = $true
           choco install psql -y --no-progress
           # Check for existence, since `choco` doesn't seem to fail the step if it fails to install..
+          # See https://github.com/clockworklabs/SpacetimeDB/pull/4399 for more background.
           Get-Command psql
 
       - name: Update dotnet workloads
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,8 +101,10 @@ jobs:
 
       - name: Install psql (Windows)
         if: runner.os == 'Windows'
-        run: choco install psql -y --no-progress
-        shell: powershell
+        shell: pwsh
+        run: |
+          choco install psql -y --no-progress
+          echo "C:\ProgramData\chocolatey\lib\psql\tools" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
 
       - name: Update dotnet workloads
         if: runner.os == 'Windows'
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -101,8 +101,10 @@ jobs:
 
       - name: Install psql (Windows)
         if: runner.os == 'Windows'
-        run: choco install psql -y --no-progress
-        shell: powershell
+        shell: pwsh
+        run: |
+          choco install psql -y --no-progress
+          echo "C:\ProgramData\chocolatey\lib\psql\tools" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
```
bfops: Kim ended up pointing out that the real issue was that the install failed to find the `psql` package.
```suggestion
          # Fail properly if any individual command fails
          $ErrorActionPreference = 'Stop'
          $PSNativeCommandUseErrorActionPreference = $true
          choco install psql -y --no-progress
          # Check for existence, since `choco` doesn't seem to fail the step if it fails to install..
          Get-Command psql
```

.github/workflows/ci.yml
```diff
@@ -101,8 +101,14 @@ jobs:
...
         if: runner.os == 'Windows'
-        run: choco install psql -y --no-progress
-        shell: powershell
+        shell: pwsh
+        run: |
+          # Fail properly if any individual command fails
+          $ErrorActionPreference = 'Stop'
+          $PSNativeCommandUseErrorActionPreference = $true
+          choco install psql -y --no-progress
+          # Check for existence, since `choco` doesn't seem to fail the step if it fails to install..
```
bfops, line 110:
```suggestion
          # Check for existence, since `choco` doesn't seem to fail the step if it fails to install..
          # See https://github.com/clockworklabs/SpacetimeDB/pull/4399 for more background.
```

bfops (approved):
LGTM! This doesn't fix the issue but it at least makes it more obvious. If it happens a lot we can add a retry loop or something..

PR #4401, opened 2026-02-23
fix(vue/svelte): fix useTable type interence

clockwork-tien:
# Description of Changes
- Fix types (no runtime behavior changes) for `useTable` type interence for Vue and Svelte which resulting in `{ [x: string]: any }` instead of properly typed rows
    - Changed to match the pattern React already uses of `Query<TableDef>`

1

# Testing

- [ ] 
- [ ]


clockwork-tien:
Closing as already resolved.

PR #4404, opened 2026-02-23
Append commit instead of individual transactions to commitlog

kim:
Re-open #4140 (reverted in #4292).

The original patch was merged a bit too eagerly.
It should go in _after_ 2.0 is released with some confidence.


Shubham8287:
Is this same patch as reverted one or anything else?

crates/durability/src/imp/local.rs
```diff
@@ -235,21 +238,24 @@ impl<T: Encode + Send + Sync + 'static> Actor<T> {
                     }
                 },
 
-                data = txdata_rx.recv() => {
-                    let Some(txdata) = data else {
+                tx = transactions_rx.recv() => {
```
jsdt: Should we use a `recv_many` here (which will receive up to some number of messages, but always complete as soon as there are any messages)? It seems like that would give us the eager flushing behavior we want without actually flushing for every transaction.

kim: Yeah, sorry for the stacked PRs -- this is done via #4478

kim:
> Is this same patch as reverted one or anything else?

It's the same.

Well, it's no longer the same, because #4478 got merged into this one. 
I intended this one to merge, then #4478, not squash both.

PR #4406, opened 2026-02-23
Remove database names from node-js quickstart

jdetter:
# Description of Changes

- We forgot to remove the database name from the example commands so I've fixed that here.

- None, just a docs change

- None, this is just a docs change

# Testing

- [x] The nodejs quickstart works as written


jdetter:
Cherry picked into tyler/fix-dev-watch-and-env

PR #4408, opened 2026-02-23
[TS] Fix returning a query from a view that returns an option

coolreader18:
# Description of Changes

This was an issue in 1.x as well, so it's not theoretically blocking 2.0.

1

# Testing

- [x] Verified that shub's `PersonAtLevel2` view works correctly now.


joshua-spacetime (commented):
Note, the type signature becomes a little misleading in this case because currently we don't have a way of determining statically whether a query builder view returns at most 1 row.

So I would say we don't want this change but rather would like to restrict the return type to array for query builder returns.

cloutiertyler:
Closing this given the last comment. Feel free to reopen if this is wrong.

PR #4409, opened 2026-02-23
[TS] Add typescript dependency to typescript templates

coolreader18:
# Description of Changes

`spacetime build` now attempts to run tsc for typescript modules, and will emit a warning if it's not present. This PR makes it so our CLI stops emitting a warning for our templates.

1

# Testing

- [x] Verified that typescript gets properly installed.


bfops (approved):
LGTM thank you 🙏

PR #4410, opened 2026-02-23
Llm benchmark updates

cloutiertyler:
# Description of Changes

# Testing

- [ ] 
- [ ]

### DIFF
--- a/crates/smoketests/tests/views.rs
+++ b/crates/smoketests/tests/views.rs
@@ -17,7 +17,7 @@ export default spacetimedb;
 export const my_player = spacetimedb.view(
   { public: true },
   t.option(playerState.rowType),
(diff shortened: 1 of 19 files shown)


joshua-spacetime:
AI made the following corrections:
1. Corrected “1.0 (invalid)” TypeScript examples that were accidentally using 2.0 syntax.
2. Removed incorrect claims that C# accessors must be PascalCase singular or that schedule field names must be PascalCase.
3. Relaxed/clarified C# primary key guidance (no longer claims every table must have one).
4. Fixed migration AI rule examples to use default schema export (`export default ...`) instead of `export const spacetimedb = ...`.
5. Fixed `ScheduleAt` examples in C# AI rules to use constructor syntax consistently.
6. Corrected C# client cache accessor naming description to match generated C# conventions.

PR #4414, opened 2026-02-23
C++ Quickstart - spacetime dev not working

JasonAtClockwork:
# Description of Changes
Updated the /templates/basic-cpp to target Rust 2.0.* and add pre-generated bindings as the updated `spacetime dev` seems to require that.

N/A 

1 - Template only change

# Testing

- [x] Was able to use `spacetimedb-cli.exe dev --template basic-cpp`


bfops (approved):
This seems good to me!

PR #4419, opened 2026-02-24
Confirmed reads default only for v2 connections

bfops:
# Description of Changes

Reducing scope of https://github.com/clockworklabs/SpacetimeDB/pull/4390 to only apply to V2 clients.

I think this is an API change?

1

# Testing

- [ ] 
- [ ]


bfops:
We aligned in a meeting to force-merge this one.

PR #4447, opened 2026-02-25
guard: Attempt to kill server gracefully

kim:
Send it SIGINT / CTRL_BREAK_EVENT, which standalone handles. Wait a bit. Send it SIGKILL.

This gives the server a chance to run finalizers. We aren't well equipped to deal with sudden death currently.


bfops:
What's the impact of this? I kind of assumed that the restart tests were partly trying to test what happens if the server dies un-gracefully.

PR #4459, opened 2026-02-25
Make event tables inaccessible from views

### LINKED ISSUE #4436: Event tables accessible from views
Event tables were not supposed to be accessible from views but we're generating bindings for them in `ViewContext` and `AnonymousViewContext`. https://github.com/clockworklabs/SpacetimeDB/blob/bce97d45accdfd67cf7fdb1e6ff34c11d7a85d4f/docs/docs/00200-core-concepts/00300-tables/00550-event-tables.md?plain=1#L200

joshua-spacetime:
# Description of Changes

According to the docs, event tables should not be accessible from views, but in fact they are. This patch stops generating view bindings for event tables across rust, c#, and typescript.

Technically a breaking change, but they were never supposed to be there in the first place and would not work correctly regardless.

2

# Testing

Added negative compile tests for trying to access event tables from within a view.


crates/bindings-csharp/Codegen.Tests/fixtures/diag/snapshots/ExtraCompilationErrors.verified.txt
```diff
@@ -23,7 +23,53 @@
     }
   },
   {/*
-SpacetimeDB.Internal.Module.RegisterTable<global::TestUniqueNotEquatable, SpacetimeDB.Internal.TableHandles.TestUniqueNotEquatable>();
+    {
+        var _ = ctx.Db.ViewAuditEvent;
+                       ^^^^^^^^^^^^^^
+        return null;
+*/
+    Message: 'LocalReadOnly' does not contain a definition for 'ViewAuditEvent' and no accessible extension method 'ViewAuditEvent' accepting a first argument of type 'LocalReadOnly' could be found (are you missing a using directive or an assembly reference?),
```
joshua-spacetime, line 31: No longer generating bindings for event tables.

crates/bindings-csharp/Codegen.Tests/fixtures/diag/snapshots/ExtraCompilationErrors.verified.txt
```diff
@@ -137,6 +183,52 @@ SpacetimeDB.Internal.Module.RegisterClientVisibilityFilter(global::Module.MY_FOU
       ]
     }
   },
+  {/*
+    {
+        return ctx.From.ViewAuditEvent();
+                        ^^^^^^^^^^^^^^
+    }
+*/
+    Message: 'QueryBuilder' does not contain a definition for 'ViewAuditEvent' and no accessible extension method 'ViewAuditEvent' accepting a first argument of type 'QueryBuilder' could be found (are you missing a using directive or an assembly reference?),
```
joshua-spacetime, line 215: The same is true for the query builder

crates/bindings-csharp/Codegen/Module.cs
```diff
@@ -853,7 +853,7 @@ public IEnumerable<GeneratedReadOnlyAccessor> GenerateReadOnlyAccessors()
             yield break;
         }
 
-        foreach (var accessor in TableAccessors)
+        foreach (var accessor in TableAccessors.Where(accessor => !accessor.IsEvent))
```
joshua-spacetime, line 876: Restricting `Ctx.Db` in C# (procedural views)

crates/bindings-csharp/Codegen/Module.cs
```diff
@@ -888,7 +888,7 @@ public IEnumerable<string> GenerateQueryBuilderMembers()
         var vis = SyntaxFacts.GetText(Visibility);
         var globalRowName = $"global::{FullName}";
 
-        foreach (var accessor in TableAccessors)
+        foreach (var accessor in TableAccessors.Where(accessor => !accessor.IsEvent))
```
joshua-spacetime, line 918: Restricting `Ctx.From` in C# (query builder views)

crates/bindings-macro/src/table.rs
```diff
@@ -1167,18 +1167,22 @@ pub(crate) fn table_impl(mut args: TableArgs, item: &syn::DeriveInput) -> syn::R
         }
     };
 
-    let trait_def_view = quote_spanned! {table_ident.span()=>
-        #[allow(non_camel_case_types, dead_code)]
-        #vis trait #view_trait_ident {
+    let trait_def_view = if args.event.is_none() {
```
joshua-spacetime, line 1170: Restricting `ctx.db` in rust (procedural views)

crates/bindings-macro/src/table.rs
```diff
@@ -1250,14 +1254,18 @@ pub(crate) fn table_impl(mut args: TableArgs, item: &syn::DeriveInput) -> syn::R
 
     };
 
-    let table_query_handle_def = quote! {
-           #[allow(non_camel_case_types, dead_code)]
-           #vis trait #query_trait_ident {
-               fn #table_ident(&self) -> spacetimedb::query_builder::Table<#original_struct_ident> {
-                   spacetimedb::query_builder::Table::new(stringify!(#table_ident))
+    let table_query_handle_def = if args.event.is_none() {
```
joshua-spacetime, line 1257: Restricting `ctx.from` in rust (query builder views)

crates/bindings-typescript/src/server/runtime.ts
```diff
@@ -295,6 +303,25 @@ class ModuleHooksImpl implements ModuleHooks {
     ));
   }
 
+  get #viewSchema() {
+    return (this.#viewSchema_ ??= withoutEventTables(this.#schema.schemaType));
```
joshua-spacetime, line 308: Restrict `ctx.db` and `ctx.from` in typescript

crates/bindings-typescript/src/server/runtime.ts
```diff
@@ -295,6 +303,25 @@ class ModuleHooksImpl implements ModuleHooks {
...
+  }
+
+  get #viewDbView() {
```
coolreader18, line 311: I don't think this is necessary, because with just the typing alone the users won't be able to access the event tables (and if they bypass it, it'll error).

joshua-spacetime: > and if they bypass it, it'll error

Will it? Without these hooks, view contexts get access to the full schema at runtime, so it may not error if you bypass the type system, right?

rekhoff (approved):
After reviewing the C# changes, these look solid. I am approving the C# portion of the changes.

cloutiertyler:
> Technically a breaking change, but they were never supposed to be there in the first place and would not work correctly regardless.

I think they would work fine, no? They just won't have any rows.

PR #4461, opened 2026-02-25
Update C++ module bindings to RawModuleDefV10

JasonAtClockwork:
# Description of Changes
  - Migrated the C++ module-definition assembly path to V10-first internals:
      - Added v10_builder and module_type_registration systems.
      - Switched Module::__describe_module__ to serialize RawModuleDef with V10 payload.
      - Updated macro registration pipeline to register through V10
  - Added explicit naming support across macro surface (*_NAMED variants for reducer/procedure/
        view and field/index macros).
  - Reworked multi-column index macros (FIELD_MultiColumnIndex, FIELD_MultiColumnIndex_NAMED) with
        migration alias.
  - Added SPACETIMEDB_SETTING_CASE_CONVERSION(...) to support case conversion policy
  - Error-path hardening by adding explicit constraint-registration error tracking and preinit validation
  - Codegen updates:
      - Updated C++ moduledef regen to V10 builder types.
      - Adjusted C++ codegen duplicate-variant wrapper generation to emit proper product-type
        wrappers.
  - Test/harness updates:
      - type-isolation-test runner now defaults to focused V10 regression checks; --v9 runs broader
        legacy/full suite.
      - Added focused modules for positive/negative V10 checks:
          - test_multicolumn_index_valid
          - error_multicolumn_missing_field
          - error_default_missing_field
  - Re-enabled C++ paths in sdks/rust/tests/test.rs procedure/view/test suites.

- Refactor of the underlying module definition
- New *_NAMED variant macros for explicit canonical naming
- FIELD_NamedMultiColumnIndex renamed to FIELD_MultiColumnIndex

3 - Large set of changes moving over to V10 with underlying changes to make future updates a little easier

# Testing
- [x] Ran the type isolation test and expanded it
- [x] Ran the spacetimedb-sdk test framework to confirm no more drift between C++ and other module languages
- [x] Ran Unreal test suite though not really applicable
- [x] New app creation with `spacetime init --template basic-cpp`
- [x] Ran describe module tests against Rust + C# matching with C++ on the /modules/sdk-test* modules to find any possible mis-alignment

# Review
- [x] Another look at the new features with C++
- [x] Thoughts on *_NAMED macros, I couldn't come up with a better solution with C++20


crates/bindings-cpp/include/spacetimedb/table_with_constraints.h
rekhoff, line 614: Not dire, but I believe `resolve_index_id()` should be updated to only query the canonical name, rather than attempting multiple patterns. Technically I think this actually makes C++ more robust with the multiple pattern search it does, but is different than Rust and C#.
I've not tested this, but it also may be possible to explicitly craft a name that  resolves to the pattern matched name of another table.

crates/bindings-cpp/include/spacetimedb/table_with_constraints.h
```diff
@@ -101,24 +124,41 @@ void register_table_type_with_constraints(const char* name, bool is_public,
  *   FIELD_PrimaryKeyAutoInc(users, id)
  *   FIELD_Unique(users, email)
  */
-#define SPACETIMEDB_TABLE(type, table_name, access_enum) \
+#define SPACETIMEDB_TABLE_IMPL(type, table_name, access_enum, options_expr) \
```
rekhoff, line 128: While this looks correct when compared to the current C# implementation, we'll likely want to update this PR generally to make Event Tables inaccessible from Views as was done in #4459 to bring C# and Typescript up-to-date with Rust.

rekhoff (commented):
This looks good overall, with a few minor items. In response to the specific Reviewer questions:
* This looks to cover the V10 code paths and lines up with the work I had done in C#.
* As for the `*_NAMED` macros, I can't think of a better solution with C++20.

crates/bindings-cpp/include/spacetimedb/table_with_constraints.h
JasonAtClockwork: This was definitely wrong and I've fixed it up, I'm still leaning heavily into `_btree` only but I'll fix that up soon with adding direct and hash indexes.

It should be more aligned now!

crates/bindings-cpp/include/spacetimedb/table_with_constraints.h
JasonAtClockwork: Oh I missed this one thank you, I've worked it through and we have compile/build time failures: `Event tables are not accessible from views.`

rekhoff (approved):
Changes look good. All reviewer items addressed, marking my approval.

PR #4463, opened 2026-02-26
Remove security warning from 00500-schedule-tables.md

taotien:
Now that scheduled reducers are private, this warning is no longer necessary.

# Description of Changes

- removed Security Considerations section from Schedule Tables documentation

### DIFF
--- a/crates/cli/src/subcommands/build.rs
+++ b/crates/cli/src/subcommands/build.rs
@@ -95,10 +95,8 @@ pub async fn exec_with_argstring(
     project_path: &Path,
     arg_string: &str,
 ) -> Result<(PathBuf, &'static str), anyhow::Error> {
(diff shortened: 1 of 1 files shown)


kistz:
@joshua-spacetime is that really the  case? does hidden from codegen imply uncallable from the http api? Just want to double check it is in fact in no way callable and not just hidden :>

joshua-spacetime:
@kistz that's right, they're not callable unless you're the owner.

taotien:
I'd like to suggest including the hint from the migration doc that you can wrap it in a public reducer to achieve the same goal: https://spacetimedb.com/docs/upgrade#define-wrappers-for-functions-that-are-both-scheduled-and-invoked-by-clients

PR #4465, opened 2026-02-26
feat: Allow adding unique constraints to existing tables

Ludv1gL:
## Summary

Adding `#[unique]` or `#[primary_key]` to an existing column currently triggers `AutoMigrateError::AddUniqueConstraint`, forcing a full database clear to apply the schema change. This PR makes it a non-breaking migration by validating existing data first:

- **If all values are unique**: constraint is added seamlessly (non-breaking migration)
- **If duplicates exist**: migration fails with a detailed error listing up to 10 duplicate groups

## Changes

- `auto_migrate.rs`: Replace hard `AddUniqueConstraint` error with `CheckAddUniqueConstraintValid` precheck + `AddConstraint` migration step
- `update.rs`: Implement precheck (full table scan, project constrained columns, count duplicates) and `AddConstraint` step execution
- `relational_db.rs`: Expose `create_constraint()` (counterpart to existing `drop_constraint()`)
- `traits.rs` / `datastore.rs`: Add `create_constraint_mut_tx` to `MutTxDatastore` trait
- `mut_tx.rs`: Make `create_constraint` public
- `formatter.rs`: Format the new `AddConstraint` step

## Safety

- **Transaction safety**: Precheck and constraint creation run in the same `MutTx` — no window for concurrent duplicate inserts
- **Index creation**: `auto_migrate_indexes()` already handles adding the backing btree index (with `is_unique=true` from the new schema). The constraint step only adds metadata.
- **Rollback**: If the precheck finds duplicates, the entire migration aborts before any changes are applied
- **Error quality**: Duplicate error shows table name, column names, and up to 10 example duplicate values with counts

## Example error output

```
Precheck failed: cannot add unique constraint 'Users_email_key' on table 'Users' column(s) [email]:
3 duplicate group(s) found.
  - String("alice@example.com") appears 2 times
  - String("bob@example.com") appears 3 times
  - String("charlie@example.com") appears 2 times
```

## Test plan

- [x] All 12 `auto_migrate` tests pass
- [x] `cargo check` passes for `spacetimedb-schema` and `spacetimedb-core`
- [x] Verified the previously-expected `AddUniqueConstraint` error test is updated
- [x] Manual test: add `#[unique]` to existing column with clean data → succeeds
- [x] Manual test: add `#[unique]` to existing column with duplicates → fails with detailed error


bfops:
Interesting, thank you for filing this! We'll work on getting it reviewed.

crates/core/src/db/update.rs
```diff
@@ -133,6 +134,53 @@ fn auto_migrate_database(
                     anyhow::bail!("Precheck failed: added sequence {sequence_name} already has values in range",);
                 }
             }
+            spacetimedb_schema::auto_migrate::AutoMigratePrecheck::CheckAddUniqueConstraintValid(
+                constraint_name,
+            ) => {
+                let table_def = plan.new.stored_in_table_def(constraint_name).unwrap();
+                let constraint_def = &table_def.constraints[constraint_name];
+                let cols = constraint_def.data.unique_columns().unwrap();
```
Centril: Replace with an `expect` here please

crates/core/src/db/update.rs
```diff
@@ -133,6 +134,53 @@ fn auto_migrate_database(
...
+                let col_list: ColList = cols.clone().into();
+                let table_id = stdb.table_id_from_name_mut(tx, &table_def.name)?.unwrap();
+
+                // Scan all rows, count occurrences of each projected value
+                let mut value_counts: std::collections::HashMap<AlgebraicValue, usize> =
+                    std::collections::HashMap::new();
+                for row in stdb.iter_mut(tx, table_id)? {
+                    let val = row.project(&col_list)?;
+                    *value_counts.entry(val).or_insert(0) += 1;
+                }
```
Centril: This seems rather inefficient. Every `TableIndex` index already knows what keys have duplicates through `SameKeyEntry`.

I'd like to see the implementation do this instead:
1. Find the `index_id`s affected.
2. Ask each affected index for the list of duplicates via a new method `TableIndex::iter_duplicates(&self) -> impl Iterator<Item = (AlgebraicValue, usize)>` which where `count: usize > 1`. This should forward to `Index::iter_duplicates` for each index impl. Unique indices can return an empty iterator.
3. To `SameKeyEntry` add a method `fn count(&self) -> usize` which is used for `iter_duplicates` above.
4. You then need to merge together the results of different indices by merging together the `(AlgebraicValue, usize)` entries for each index.
5. Proceed with the rest of the implementation as-is.

crates/core/src/db/update.rs
```diff
@@ -133,6 +134,53 @@ fn auto_migrate_database(
...
+                constraint_name,
+            ) => {
+                let table_def = plan.new.stored_in_table_def(constraint_name).unwrap();
+                let constraint_def = &table_def.constraints[constraint_name];
+                let cols = constraint_def.data.unique_columns().unwrap();
+                let col_list: ColList = cols.clone().into();
+                let table_id = stdb.table_id_from_name_mut(tx, &table_def.name)?.unwrap();
+
+                // Scan all rows, count occurrences of each projected value
+                let mut value_counts: std::collections::HashMap<AlgebraicValue, usize> =
```
Centril: Please use `spacetimedb_data_structures::map::HashMap` instead, with an unqualified import.

crates/datastore/src/locking_tx_datastore/mut_tx.rs
Centril, line 2171: This needs to be addressed here. In fact, I don't think rollback is being handled correctly for `ConstraintAdded` in `rollback_pending_schema_change`.

I'd also like to see a transactionality tests (`test_create_constraint_is_transactional`) for `create_constraint` and `drop_constraint` on a single-col and multi-col index.

The `NOTE` here also doesn't seem to be true, as I don't see the index being recreated unless the algorithm changes, which it doesn't if only the uniqueness changes.

crates/schema/src/auto_migrate.rs
```diff
@@ -1701,10 +1704,8 @@ mod tests {
             && type1.0 == prod1_ty && type2.0 == new_prod1_ty
         );
 
-        expect_error_matching!(
-            result,
-            AutoMigrateError::AddUniqueConstraint { constraint } => &constraint[..] == apples_name_unique_constraint
-        );
+        // Note: AddUniqueConstraint is no longer an error - adding unique constraints
```
```suggestion
        // Note: `AddUniqueConstraint` is no longer an error - adding unique constraints
```

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -1816,7 +1816,7 @@ impl MutTxId {
     /// Ensures:
     /// - The constraint metadata is inserted into the system tables (and other data structures reflecting them).
     /// - The returned ID is unique and is not `constraintId::SENTINEL`.
-    fn create_constraint(&mut self, mut constraint: ConstraintSchema) -> Result<ConstraintId> {
+    pub fn create_constraint(&mut self, mut constraint: ConstraintSchema) -> Result<ConstraintId> {
```
Centril: Note that this doesn't actually change the in-memory table to use a unique index instead of a non-unique index, so you are achieving the feature described.
Instead, you are only inserting a row into the system tables.

I would rename this to `create_st_constraint` and the `drop` one as well. The `create_table` will use the renamed method. Then you can add `create_constraint` which will call `create_st_constraint` and also find the commit index and tx index and call a new method `TableIndex::make_unique(&mut self) -> Result<(), ` on both, as well as checking `commit_index.can_merge(&tx_index, is_deleted)`

Centril (changes requested):
This needs some work to actually implement the feature.

crates/core/src/db/update.rs
Ludv1gL: This entire block will be removed along with the HashMap-based duplicate detection. #4542 detects duplicates  directly in `create_constraint()` via `iter_duplicates()`.

crates/core/src/db/update.rs
Ludv1gL: Done in #4542:

1. `SameKeyEntry::count()` added in same_key_entry.rs
2. `iter_duplicates()` added to both `MultiMap `and `HashIndex `— iterates entries where count() > 1
3. `TypedIndex::iter_duplicates()` in mod.rs matches on all non-unique variants, calls the inner `iter_duplicates()`, and converts typed keys to `AlgebraicValue`. Unique variants return an empty vec.
4. `TableIndex::iter_duplicates()` is a thin wrapper delegating to self.idx
5. The `CheckAddUniqueConstraintValid `precheck and its HashMap scanning handler in update.rs are both removed — duplicate detection now happens inside `create_constraint()` using the index's own `iter_duplicates()`.

crates/core/src/db/update.rs
Ludv1gL: All HashMap-based detection is removed so point is moot.

crates/datastore/src/locking_tx_datastore/mut_tx.rs
Ludv1gL: Rollback: `PendingSchemaChange::ConstraintAdded` now carries `IndexId `and `Option<PointerMap>` for proper rollback. On rollback, committed_state.rs calls `make_non_unique()` on the affected index and restores the pointer map if one was taken. Similarly, `ConstraintRemoved `rollback calls `make_unique()` to re-enforce uniqueness.

Transactionality tests: new tests in datastore.rs:
- ` test_create_constraint_is_transactional` (single-col): creates non-unique index -> inserts unique rows -> create_constraint -> verifies index is unique -> rollback -> verifies index is non-unique again
- `test_create_constraint_is_transactional` (multi-col): same with multi-column index
- `test_create_constraint_fails_with_duplicates`: inserts duplicate rows -> verifies create_constraint returns error -> verifies table unchanged
- `test_drop_constraint_is_transactional` (single-col + multi-col): creates unique constraint -> drop_constraint -> rollback -> verifies constraint still enforced

Above comment has been removed, as `make_unique()` just converts the existing non-unique index in-place to a unique one.

crates/schema/src/auto_migrate.rs
Ludv1gL: Fixed formatting

crates/datastore/src/locking_tx_datastore/mut_tx.rs
Ludv1gL: Oooh I assumed adding it to the system table would be enough and that it would hot reload from there on publish like on a module restart. So I guess my previous code worked on restarting the module. But if I had managed to insert duplicate indices before restarting, `horrible_things()` might have happened.
Anyhow, in #4542:

**Renames:**
- `create_constraint()` → `create_st_constraint()` (metadata-only, used by `create_table`)
- `drop_constraint()` → `drop_st_constraint()` (metadata-only)

**New `create_constraint()`:**
1. Calls `create_st_constraint()` for system table metadata
2. Finds the index by constraint columns via `find_index_for_constraint()`
3. Checks for duplicates via `iter_duplicates()` (returns error with details if any found)
4. Calls `make_unique()` on both commit-table and tx-table indices
5. Calls `can_merge()` between commit and tx indices
6. Takes pointer map from table if this becomes the first unique index (pointer map invariant: present iff no unique index exists)
7. Pushes enriched `PendingSchemaChange::ConstraintAdded { table_id, constraint_id, index_id, pointer_map }` for rollback

**New `drop_constraint()`:** Reverse — calls `make_non_unique()`, restores pointer map if no unique indices remain.

**Index conversion infrastructure:**
- `MultiMap::check_and_into_unique()` / `HashIndex::check_and_into_unique()` — validates no duplicates, converts to unique map
- `UniqueMap::from_non_unique()` / `UniqueHashIndex::from_non_unique()` — constructors from non-unique data
- `TypedIndex::make_unique()` — matches all 36 non-unique variants → corresponding unique variant using `mem::replace`
- `TypedIndex::make_non_unique()` — reverse for rollback
- Macro `define_uniqueness_conversions!` generates both directions for all variant pairs

Ludv1gL:
Rebased onto latest master (includes the `MultiMap` → `BTreeIndex` / `UniqueMap` → `UniqueBTreeIndex` rename from #4655). All changes updated to use the new type and file names.

Single squashed commit with all round 1 feedback incorporated. Ready for round 2 review.

Ludv1gL:
Rebased onto latest master (`cdd8ba77a`) which includes #4311 (BytesKey optimization).

**Changes for rebase compatibility:**
- `hash_index.rs` and `unique_btree_index.rs`: Our migration helper methods now coexist with the new relaxed-bound `delete()` and `seek_point()` methods from #4311 in the same `impl` block
- `mod.rs`: `define_uniqueness_conversions!` macro closures updated for `Packed<u128>`/`Packed<i128>` — `iter_duplicates()` returns raw key types, so closures now wrap with `Packed()` when constructing `AlgebraicValue`
- `unique_btree_index.rs`: Trait bound updated to `K: KeySize + Ord` (was `K: Ord + KeySize`) to match upstream ordering
- Added `use spacetimedb_sats::algebraic_value::Packed;` import to `mod.rs`

All 15 constraint tests pass (`cargo test -p spacetimedb-datastore --lib --features test -- constraint`).

crates/schema/src/auto_migrate.rs
```diff
@@ -2399,4 +2403,58 @@ mod tests {
...
+        let plan = ponder_auto_migrate(&old, &new).expect("removing a table should produce a valid plan");
+        assert_eq!(
+            plan.steps,
+            vec![
+                AutoMigrateStep::RemoveTable(&drop_table),
+                AutoMigrateStep::DisconnectAllUsers,
+            ],
+            "plan should only contain RemoveTable + DisconnectAllUsers, no orphan sub-object steps"
+        );
+    }
```
Centril, line 2476: This looks like it is from the other removal PR.

crates/datastore/src/locking_tx_datastore/datastore.rs
```diff
@@ -3983,4 +3991,197 @@ mod tests {
...
+            "Foo_idx_btree",
+            BTreeAlgorithm::from(col_pos),
+        )];
+        basic_table_schema_with_indices(indices, Vec::<ConstraintSchema>::new())
+    }
+
+    /// Helper: create a table with a non-unique btree index on multiple columns but no constraints.
+    fn table_with_non_unique_multi_col_index(cols: impl Into<ColList>) -> TableSchema {
+        let indices = vec![IndexSchema::for_test(
+            "Foo_multi_idx_btree",
```
```suggestion
    /// Creates a table with a non-unique btree index on `cols` but no constraints.
    fn table_with_non_unique_index(cols: impl Into<ColList>) -> TableSchema {
        let indices = vec![IndexSchema::for_test(
            "Foo_idx_btree",
```
The latter function subsumes the former.

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -1725,19 +1725,12 @@ impl MutTxId {
...
-    /// Requires:
-    /// - `constraint.constraint_name` must not be used for any other database entity.
-    /// - `constraint.constraint_id == ConstraintId::SENTINEL`
-    /// - `constraint.table_id != TableId::SENTINEL`
-    /// - `is_unique` must be `true` if and only if a unique constraint will exist on
-    ///   `ColSet::from(&constraint.constraint_algorithm.columns())` after this transaction is committed.
-    ///
-    /// Ensures:
-    /// - The constraint metadata is inserted into the system tables (and other data structures reflecting them).
-    /// - The returned ID is unique and is not `constraintId::SENTINEL`.
```
Centril, line 1739: Why were these comments removed?

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -1725,19 +1725,12 @@ impl MutTxId {
             })
     }
 
-    /// Create a constraint.
+    /// Insert constraint metadata into system tables only.
```
Centril:
```suggestion
    /// Inserts constraint metadata into system tables only.
```

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -1770,15 +1763,15 @@ impl MutTxId {
 
         let ((tx_table, ..), (commit_table, ..)) = self.get_or_create_insert_table_mut(table_id)?;
         constraint.constraint_id = constraint_id;
-        // This won't clone-write when creating a table but likely to otherwise.
```
Centril, line 1773: Why?

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -1791,19 +1784,176 @@ impl MutTxId {
         let table_id = st_constraint_ref.read_col(StConstraintFields::TableId)?;
         self.delete(ST_CONSTRAINT_ID, st_constraint_ref.pointer())?;
 
-        // Remove constraint in transaction's insert table.
         let ((tx_table, ..), (commit_table, ..)) = self.get_or_create_insert_table_mut(table_id)?;
-        // This likely will do a clone-write as over time?
-        // The schema might have found other referents.
```
Centril, line 1797: Why were these comments removed?

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -1791,19 +1784,176 @@ impl MutTxId {
...
+                    .expect("index must exist")
+                    .make_non_unique();
+
+                tx_table
+                    .indexes
+                    .get_mut(&index_id)
+                    .expect("tx index must exist")
+                    .make_non_unique();
+
+                // If no unique indices remain, rebuild the pointer map.
```
Centril: This should be done for the `tx_table` too. In fact, all of this logic can be moved into a new method `Table::make_index_non_unique(&mut self, index_id: IndexId)`.

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -1791,19 +1784,176 @@ impl MutTxId {
...
+                if !commit_table.has_unique_index() {
+                    let pm = commit_table
+                        .scan_rows(commit_blob_store)
+                        .map(|row_ref| (row_ref.row_hash(), row_ref.pointer()))
+                        .collect();
```
Centril: This is `rebuild_pointer_map`.

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -1791,19 +1784,176 @@ impl MutTxId {
...
+
+        // If this was a unique constraint, make the index non-unique.
+        let unique_cols = schema.data.unique_columns().cloned();
+        let made_non_unique_index_id = if let Some(cols) = unique_cols {
+            let col_list: ColList = cols.into();
+
+            let ((tx_table, ..), (commit_table, commit_blob_store, _)) =
+                self.get_or_create_insert_table_mut(table_id)?;
+
+            let index_id = commit_table.get_index_by_cols(&col_list).map(|(id, _)| id);
```
Centril: I believe there could be more than one index here.

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -1791,19 +1784,176 @@ impl MutTxId {
...
+
+        // Step 2: If this is a unique constraint, make the index unique.
+        let (made_unique_index_id, pointer_map) = if let Some(cols) = unique_cols {
+            let col_list: ColList = cols.into();
+
+            let ((tx_table, _, tx_delete_table), (commit_table, commit_blob_store, _)) =
+                self.get_or_create_insert_table_mut(table_id)?;
+
+            // Find the index matching these columns.
+            let index_id = commit_table.get_index_by_cols(&col_list).map(|(id, _)| id);
```
Centril: There could be more than one index.

crates/table/src/table_index/mod.rs
```diff
@@ -1553,6 +1606,81 @@ impl TypedIndex {
     pub fn num_key_bytes(&self) -> u64 {
         same_for_all_types!(self, this => this.num_key_bytes())
     }
+
+    define_uniqueness_conversions! {
```
Centril, line 1669: This is not including the `(RangedCompat)BytesKey` indices. I think the conversions should be negotiated via first converting to `IndexKey` and from there you can convert to `AlgebraicValue`.
I think this code can be simplified, but I can take care of that refactoring later.

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -1791,19 +1784,176 @@ impl MutTxId {
...
+                // Take the pointer map if this is the first unique index.
+                let pointer_map = if !had_unique {
+                    commit_table.take_pointer_map()
+                } else {
+                    None
+                };
+
+                // Also take tx table's pointer map if needed.
+                if !had_unique {
+                    tx_table.take_pointer_map();
```
Centril: This could be moved into the other `if` branch before taking the commit table's pointer map.

crates/datastore/src/locking_tx_datastore/datastore.rs
```diff
@@ -3983,4 +3991,197 @@ mod tests {
...
+
+        // TX3: try to add unique constraint — should fail.
+        let mut tx = begin_mut_tx(&datastore);
+        let mut constraint = ConstraintSchema::unique_for_test("Foo_id_unique", 0u16);
+        constraint.table_id = table_id;
+        let result = datastore.create_constraint_mut_tx(&mut tx, constraint);
+        assert!(result.is_err(), "create_constraint should fail when duplicates exist");
+
+        Ok(())
+    }
```
Centril, line 4109: Would be good to add a test that exercises the case where the duplicate row is in the tx state.

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -1791,19 +1784,176 @@ impl MutTxId {
...
+                let tx_idx = &tx_table.indexes[&index_id];
+                let is_deleted = |ptr: &RowPointer| tx_delete_table.contains(*ptr);
+                commit_idx.can_merge(tx_idx, is_deleted).map_err(|violation| {
+                    let violation = commit_table
+                        .get_row_ref(commit_blob_store, violation)
+                        .expect("row came from scanning the table")
+                        .project(&commit_idx.indexed_columns)
+                        .expect("cols should be valid for this table");
+                    anyhow::anyhow!("Unique constraint violation during merge: {violation:?}")
+                })?;
```
Centril: I think you have a bug here. When rollback happens, because the TX failed, the commit table will still have a unique index. Both indices should be reverted here on error. Please add a test that hits this error, does a rollback and then confirms that the index is still non-unique.

Centril (changes requested):
Getting there but there are still some issues to be worked out.

Ludv1gL:
Rebased onto latest master and addressed all round 2 feedback. Single squashed commit.

**Changes since last review:**

- **`.unwrap()` → `.expect()`** in `AddConstraint` handler (`update.rs`)
- **Backticks** in `AddUniqueConstraint` comment (`auto_migrate.rs`)
- **Restored removed comments**: `Ensures:` doc section, clone-write comments on both `create_st_constraint` and `drop_st_constraint`
- **Doc comment**: `/// Inserts constraint metadata...` (imperative third-person)
- **Multiple indices on same columns**: Added `Table::get_indexes_by_cols()` returning all matches. Both `create_constraint` and `drop_constraint` now iterate over all matching indices. `PendingSchemaChange` uses `Vec<IndexId>` instead of `Option<IndexId>`.
- **Rollback bug fixed**: When `can_merge()` fails, both commit and tx indices are reverted to non-unique *before* returning the error. New test `test_create_constraint_merge_error_reverts_uniqueness` verifies rollback leaves the index non-unique.
- **`Table::make_index_non_unique()`**: Encapsulates index revert + pointer map rebuild via `rebuild_pointer_map()` (no more inlined `scan_rows().map().collect()`). Used by `drop_constraint` for both commit *and* tx tables.
- **`tx_table.take_pointer_map()`** moved into the `if !had_unique` branch alongside `commit_table.take_pointer_map()`
- **BytesKey indices** added to `define_uniqueness_conversions!` (all 4 hash pairs: `HashBytesKey8/24/56/120`)
- **Test helpers merged**: `table_with_non_unique_index(cols: impl Into<ColList>)` subsumes the multi-col variant
- **New tests**: `test_create_constraint_fails_with_tx_state_duplicates`, `test_create_constraint_merge_error_reverts_uniqueness`
- **Removed rebase artifacts**: duplicate `RemoveTable` variant, duplicate import

17 constraint tests pass. Full `spacetimedb-standalone --release` build clean (0 warnings).

crates/core/src/db/update.rs
Ludv1gL: Done — uses .expect() now.

crates/core/src/db/update.rs
Ludv1gL: Done — uses iter_duplicates() via SameKeyEntry::count().

crates/core/src/db/update.rs
Ludv1gL: Moot — no HashMap needed, uses iter_duplicates() directly.

crates/schema/src/auto_migrate.rs
Ludv1gL: Done.

crates/schema/src/auto_migrate.rs
Ludv1gL: Gone — this test is in upstream master now via #4593.

crates/datastore/src/locking_tx_datastore/datastore.rs
Ludv1gL: Done — merged into single table_with_non_unique_index(cols: impl Into<ColList>).

crates/datastore/src/locking_tx_datastore/mut_tx.rs
Ludv1gL: Restored.

crates/datastore/src/locking_tx_datastore/mut_tx.rs
Ludv1gL: Done.

crates/datastore/src/locking_tx_datastore/mut_tx.rs
Ludv1gL: Restored.

crates/datastore/src/locking_tx_datastore/mut_tx.rs
Ludv1gL: Restored.

crates/datastore/src/locking_tx_datastore/mut_tx.rs
Ludv1gL: Done — Table::make_index_non_unique() handles both tables. Uses rebuild_pointer_map().

crates/datastore/src/locking_tx_datastore/mut_tx.rs
Ludv1gL: Done — uses rebuild_pointer_map() inside Table::make_index_non_unique().

crates/datastore/src/locking_tx_datastore/mut_tx.rs
Ludv1gL: Done — uses get_indexes_by_cols() and iterates all.

crates/datastore/src/locking_tx_datastore/mut_tx.rs
Ludv1gL: Done — uses get_indexes_by_cols() and iterates all.

crates/table/src/table_index/mod.rs
Ludv1gL: Done — all 4 HashBytesKey pairs added. Debug formatting for iter_duplicates; you mentioned you'll refactor later.

crates/datastore/src/locking_tx_datastore/mut_tx.rs
Ludv1gL: Done — both calls in the same if !had_unique block.

crates/datastore/src/locking_tx_datastore/datastore.rs
Ludv1gL: Done — test_create_constraint_fails_with_tx_state_duplicates added.

crates/datastore/src/locking_tx_datastore/mut_tx.rs
Ludv1gL: Fixed — both indices reverted before returning error. test_create_constraint_merge_error_reverts_uniqueness verifies rollback leaves index non-unique.

crates/datastore/src/locking_tx_datastore/datastore.rs
```diff
@@ -4004,4 +4012,256 @@ mod tests {
...
+        let result = datastore.create_constraint_mut_tx(&mut tx, constraint);
+        assert!(
+            result.is_err(),
+            "create_constraint should fail when the tx state contains a duplicate of a committed row"
+        );
+        Ok(())
+    }
+
+    #[test]
+    fn test_create_constraint_merge_error_reverts_uniqueness() -> ResultTest<()> {
```
Centril, line 4201: The test above (test_create_constraint_fails_with_tx_state_duplicates) is just a prefix of this test, so you can remove the test above.

crates/datastore/src/locking_tx_datastore/datastore.rs
```diff
@@ -4004,4 +4012,256 @@ mod tests {
...
+        let result = datastore.create_constraint_mut_tx(&mut tx, constraint);
+        assert!(
+            result.is_err(),
+            "create_constraint should fail when tx state conflicts with committed state"
+        );
+
+        // Rollback and verify the index is still non-unique.
+        let _ = datastore.rollback_mut_tx(tx);
+        let mut tx = begin_mut_tx(&datastore);
+        let dup_result = insert(&datastore, &mut tx, table_id, &u32_str_u32(1, "Charlie", 20));
```
Centril, line 4241: Let's do this before the `rollback_mut_tx` as well, and it should succeed.

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -1725,19 +1725,16 @@ impl MutTxId {
...
 
-    /// Create a constraint.
+    /// Inserts constraint metadata into system tables only.
     ///
-    /// Requires:
-    /// - `constraint.constraint_name` must not be used for any other database entity.
-    /// - `constraint.constraint_id == ConstraintId::SENTINEL`
-    /// - `constraint.table_id != TableId::SENTINEL`
-    /// - `is_unique` must be `true` if and only if a unique constraint will exist on
-    ///   `ColSet::from(&constraint.constraint_algorithm.columns())` after this transaction is committed.
```
Centril, line 1735: This comment was not restored.

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -1791,19 +1789,181 @@ impl MutTxId {
         let table_id = st_constraint_ref.read_col(StConstraintFields::TableId)?;
         self.delete(ST_CONSTRAINT_ID, st_constraint_ref.pointer())?;
 
-        // Remove constraint in transaction's insert table.
```
Centril, line 1794: This comment was not restored.

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -1791,19 +1789,181 @@ impl MutTxId {
...
+                self.get_or_create_insert_table_mut(table_id)?;
+
+            // Find all indices matching these columns.
+            let index_ids: Vec<_> = commit_table
+                .get_indexes_by_cols(&col_list)
+                .into_iter()
+                .map(|(id, _)| id)
+                .collect();
+
+            if !index_ids.is_empty() {
```
Centril: A unique constraint must be backed by at least one index. If there isn't one, this should error and early return.
In fact, check this before doing `create_st_constraint`. It would also be best to early return if this is not adding a unique constraint and error in that case, as that is the only constraint that can be added at the moment.

crates/table/src/table_index/mod.rs
```diff
@@ -1612,6 +1665,85 @@ impl TypedIndex {
...
+            HashU256 <=> UniqueHashU256 : |k: &u256| AlgebraicValue::U256(Box::new(*k)),
+            HashI256 <=> UniqueHashI256 : |k: &i256| AlgebraicValue::I256(Box::new(*k)),
+            HashF32 <=> UniqueHashF32 : |k: &F32| AlgebraicValue::F32(*k),
+            HashF64 <=> UniqueHashF64 : |k: &F64| AlgebraicValue::F64(*k),
+            HashString <=> UniqueHashString : |k: &Box<str>| AlgebraicValue::String(k.clone()),
+            HashAV <=> UniqueHashAV : |k: &AlgebraicValue| k.clone(),
+            HashBytesKey8 <=> UniqueHashBytesKey8 : |k: &BytesKey<BYTES_KEY_SIZE_8_H>| AlgebraicValue::String(format!("{k:?}").into()),
+            HashBytesKey24 <=> UniqueHashBytesKey24 : |k: &BytesKey<BYTES_KEY_SIZE_24_H>| AlgebraicValue::String(format!("{k:?}").into()),
+            HashBytesKey56 <=> UniqueHashBytesKey56 : |k: &BytesKey<BYTES_KEY_SIZE_56_H>| AlgebraicValue::String(format!("{k:?}").into()),
+            HashBytesKey120 <=> UniqueHashBytesKey120 : |k: &BytesKey<BYTES_KEY_SIZE_120_H>| AlgebraicValue::String(format!("{k:?}").into()),
```
Centril, line 1717: These are all incorrect, but this PR should go after one of mine which will make fixing this easier.

crates/table/src/table_index/mod.rs
```diff
@@ -1612,6 +1665,85 @@ impl TypedIndex {
...
+            BTreeU64 <=> UniqueBTreeU64 : |k: &u64| AlgebraicValue::U64(*k),
+            BTreeI64 <=> UniqueBTreeI64 : |k: &i64| AlgebraicValue::I64(*k),
+            BTreeU128 <=> UniqueBTreeU128 : |k: &u128| AlgebraicValue::U128(Packed(*k)),
+            BTreeI128 <=> UniqueBTreeI128 : |k: &i128| AlgebraicValue::I128(Packed(*k)),
+            BTreeU256 <=> UniqueBTreeU256 : |k: &u256| AlgebraicValue::U256(Box::new(*k)),
+            BTreeI256 <=> UniqueBTreeI256 : |k: &i256| AlgebraicValue::I256(Box::new(*k)),
+            BTreeF32 <=> UniqueBTreeF32 : |k: &F32| AlgebraicValue::F32(*k),
+            BTreeF64 <=> UniqueBTreeF64 : |k: &F64| AlgebraicValue::F64(*k),
+            BTreeString <=> UniqueBTreeString : |k: &Box<str>| AlgebraicValue::String(k.clone()),
+            BTreeAV <=> UniqueBTreeAV : |k: &AlgebraicValue| k.clone(),
```
Centril, line 1688: What about `BytesKey` for btrees?

Centril (changes requested):

Ludv1gL:
Pushed 7 commits on top of `4d2568d9`. Five of them directly address your 2026-04-20 review — one commit per comment, mapped below. The remaining two (`0e9514d4`, `0e6182e0`) are proactive fixes for bugs surfaced while auditing the constraint-toggle rollback paths; they're on-theme with your round-1 concern about rollback correctness (thread on `mut_tx.rs:1885`). Both carry differential tests — I verified each FAILS on the pre-fix tree and PASSES with the fix.

| Commit | Addresses |
|---|---|
| `00790a3f` | `mut_tx.rs:1735`, `mut_tx.rs:1794` (comment restoration) |
| `9e4cd3c6` | `datastore.rs:4229` (delete redundant test) |
| `ae5f5a65` | `datastore.rs:4260` (pre-rollback assertion) |
| `154d763a` | `table_index/mod.rs:1688` (TODO pointing at #4733) |
| `e6b66b02` | `mut_tx.rs:1828` (validate before `create_st_constraint`) |
| `0e9514d4` | follow-up: `ConstraintRemoved` rollback pointer-map invariant |
| `0e6182e0` | follow-up: Existed-path `last_mut()` clobber |

I've left per-thread replies pointing at the exact SHAs. Happy to rebase or split anything further.

crates/datastore/src/locking_tx_datastore/mut_tx.rs
Ludv1gL: Restored in `00790a3f`. The old "Requires:" list is back; the `is_unique` bullet is rephrased to reflect the split (caller ensures backing-index uniqueness, or uses `Self::create_constraint` which does the conversion).

crates/datastore/src/locking_tx_datastore/mut_tx.rs
Ludv1gL: Restored in `00790a3f`. The `// Remove constraint in transaction's insert table.` comment is back on `drop_st_constraint`, and the `TODO(1.0)/NOTE(centril)` block is back on the public `drop_constraint` after `push_schema_change`.

crates/datastore/src/locking_tx_datastore/datastore.rs
Ludv1gL: Done in `9e4cd3c6`.

crates/datastore/src/locking_tx_datastore/datastore.rs
Ludv1gL: Done in `ae5f5a65`. The new pre-rollback duplicate insert asserts the `make_non_unique` revert actually ran inside the failing tx before returning.

crates/datastore/src/locking_tx_datastore/mut_tx.rs
Ludv1gL: Done in `e6b66b02`. Both checks — non-unique constraint kind and missing backing index — now fire before `create_st_constraint` is called. Added `test_create_constraint_fails_without_backing_index` which also asserts `constraint_id_from_name` returns `None` afterward, proving no orphan `st_constraint` row was written before the check.

crates/table/src/table_index/mod.rs
Ludv1gL: Added a TODO in `154d763a` pointing at #4733 — the btree variants need `RangeCompatBytesKey` to land first. Happy to open the follow-up PR once #4733 merges.

crates/table/src/table_index/mod.rs
Ludv1gL: Acknowledged in the same `154d763a` TODO. I'll revise the `HashBytesKey*` closures in the same follow-up once #4733 provides the right ordering primitive.

crates/datastore/src/locking_tx_datastore/mut_tx.rs
Ludv1gL: Addressed across the PR and extended in two new commits:

- **`ConstraintAdded` rollback**: `PendingSchemaChange::ConstraintAdded` now carries `Vec<IndexId> + Option<PointerMap>` and the rollback in `committed_state.rs` reverses both (see current file).
- **`ConstraintRemoved` rollback**: discovered while re-auditing — forward `drop_constraint` rebuilds the pointer map when no unique index remains, but rollback never undid that. Fixed in `0e9514d4` with a differential test (`test_drop_constraint_rollback_restores_pointer_map_invariant`) that FAILS on pre-fix code and PASSES on post-fix.
- **Transactionality tests**: `test_create_constraint_rollback_restores_non_unique`, `test_drop_constraint_rollback_keeps_unique`, `test_drop_constraint_rollback_restores_pointer_map_invariant`, and `test_create_constraint_merge_error_reverts_uniqueness` now cover single-col, multi-col, duplicates, merge-error revert, and pointer-map invariant.
- The old `NOTE` about recreating the index was removed because `make_unique`/`make_non_unique` now convert in-place (no algorithm change → no recreation).

crates/datastore/src/locking_tx_datastore/mut_tx.rs
Ludv1gL: Done in the prior push: `create_constraint` now calls `TypedIndex::make_unique()` on both commit and tx indices, runs `can_merge` between each pair, and takes the pointer map if this is the first unique index — see current `mut_tx.rs` around line 1870 onward.

crates/datastore/src/locking_tx_datastore/mut_tx.rs
Ludv1gL: Done in the prior push: `Table::make_index_non_unique` lives at `table.rs:1574`, handles both the index flip AND the pointer-map rebuild. `drop_constraint` calls it on both tx and commit tables.

Ludv1gL:
Rebased onto current `upstream/master` (includes #4733 bytes-key btrees, #4782 index housekeeping, #4850 replay refactor, #4666 primary-key migration). Force-pushed — SHAs changed, mapping below.

### Deferred round-3 items resolved (`9d4f4a9bcb`)

Now that #4733 has landed, the two `table_index/mod.rs` items are addressed:

- **`mod.rs:1688`** — 5 new `BTreeBytesKey{8,16,32,64,128} <=> UniqueBTreeBytesKey*` pairs added to `define_uniqueness_conversions!`, using the new `RangeCompatBytesKey<N>`.
- **`mod.rs:1717`** — `HashBytesKey{8,24,56,120}` closures rewritten to go through `TypedIndexKey::*.into_algebraic_value(key_type)` per your "via `IndexKey` → `AlgebraicValue`" hint. The `format!(\"{k:?}\")` placeholder is gone.

To thread the key type through, the macro-generated `iter_duplicates` grew a `key_type: &AlgebraicType` parameter, `TableIndex::iter_duplicates` passes `&self.key_type`, and all ~33 pre-existing scalar closures now accept (and ignore) the extra arg.

### Other rebase work

- **`480377251e`** — adopt accessor `indexed_columns()` for the field privatized in #4782.
- **`78fcb7018d`** — eliminated a double-scan in `create_constraint`. Was running `iter_duplicates` up front, then `make_unique` (which scans internally). Now `make_unique` runs first and `iter_duplicates` only fires on the failure path for the detailed error message. Also catches tx-local duplicates (was previously a latent `.expect(\"tx table should have no duplicates\")`).
- **`7aea551466`** — clippy: `&str` over `&Box<str>` in the two string closures.

### SHA mapping for previously-verified commits

| Old SHA | New SHA | Subject |
|---|---|---|
| `4d2568d9` | `87ec15e0` | feat: Allow adding unique constraints to existing tables |
| `00790a3f` | `4cd9e232` | docs(mut_tx): restore pre-split doc blocks on constraint methods |
| `9e4cd3c6` | `84d0fbaf` | test(datastore): drop redundant test_create_constraint_fails_with_tx_state_duplicates |
| `ae5f5a65` | `070ee5f7` | test(datastore): verify merge-error revert happens inside the failing tx |
| `154d763a` | `c7cffdce` | docs(table_index): note BTreeBytesKey follow-up (TODO now deleted by `9d4f4a9b`) |
| `e6b66b02` | `d0b3b149` | feat(mut_tx): validate constraint data before writing st_constraint |
| `0e9514d4` | `85c6a217` | fix(committed_state): drop rebuilt pointer map on ConstraintRemoved rollback |
| `0e6182e0` | `3479d85e` | fix(mut_tx): short-circuit create_constraint on already-existing st_constraint row |

### Validation

- `cargo test -p spacetimedb-datastore --features test` — 93/93 pass (incl. 10 constraint tests)
- `cargo test -p spacetimedb-schema` — 101/101
- `cargo test -p spacetimedb-core` (update) — 14/14
- `cargo test -p spacetimedb-table` — 103/103
- `cargo build -j32 -p spacetimedb-standalone --release` — clean
- `cargo clippy --tests` on touched crates — clean

crates/table/src/table_index/mod.rs
Ludv1gL: Added in `9d4f4a9bcb`. The 5 `BTreeBytesKey{8,16,32,64,128} <=> UniqueBTreeBytesKey*` pairs use `RangeCompatBytesKey<N>` from #4733. Closures dispatch through `TypedIndexKey::BytesKey{8B,16,32,64,128}(*k).into_algebraic_value(key_type)`.

crates/table/src/table_index/mod.rs
Ludv1gL: Rewritten in `9d4f4a9bcb`. `HashBytesKey{8,24,56,120}` closures now go through `TypedIndexKey::BytesKey{8H,24,56,120}(*k).into_algebraic_value(key_type)`, matching your "via `IndexKey` → `AlgebraicValue`" shape. The `format!("{k:?}")` placeholder is gone. To thread the key type through the macro, `TypedIndex::iter_duplicates` grew a `key_type: &AlgebraicType` parameter and `TableIndex::iter_duplicates` now passes `&self.key_type`.

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -1812,12 +1825,207 @@ impl MutTxId {
...
+                    .take(10)
+                    .map(|(val, count)| format!("  - {val:?} appears {count} times"))
+                    .collect::<Vec<_>>()
+                    .join("\n");
+                return Err(anyhow::anyhow!(
+                    "Cannot add unique constraint on table {table_id} column(s) {col_list:?}:\n\
+                     {total} duplicate group(s) found.\n{examples}{}",
+                    if total > 10 { "\n  ... and more" } else { "" }
+                )
+                .into());
```
Centril: Let's reuse this below for `tx_table`.

Centril (changes requested):

Ludv1gL:
New commit `dc7fee3324` addresses `mut_tx.rs:1939`.

Extracted a `dup_err` closure that runs `iter_duplicates()` and formats the "N duplicate group(s) found" error, then reused it on both the committed-state and tx-state `make_unique` failure paths. The tx-state path used to return only a generic "duplicate values exist in the current transaction" — now both surfaces list up to 10 duplicate groups. A `source` string (`"committed state"` / `"current transaction"`) distinguishes which side fired.

Validated: 19/19 `test_*unique*` + `test_*create_constraint*` + `test_*drop_constraint*` pass; `cargo clippy` clean.

Centril (approved):
Thanks! This looks good now.

Centril:
Looks like you need to run cargo fmt

PR #4473, opened 2026-02-26
Adds code signing to tagged windows builds

rekhoff:
**Note**: This change requires the addition of new entries in the secrets to work properly. These should be added prior to this merging.

# Description of Changes

* Add a tag-only Windows signing job that runs on a self-hosted signing runner.
  * This is an alternative/separate code-path just for the signing job. See **Alternatives Considered** for details.
* Skip the unsigned Windows matrix build on tags so signed artifacts are the only Windows release outputs.
* Sign `spacetimedb-update.exe`, `spacetimedb-cli.exe`, and `spacetimedb-standalone.exe` before packaging, then upload the signed artifacts as usual.

# Alternatives Considered
**Inline signing in the existing Windows packaging step**. This was rejected because it would require all Windows builds (including non-tag builds) to run on the signing-capable runner or to install/signing tooling on GitHub-hosted runners. The chosen approach isolates signing to tag releases, avoids exposing credentials in standard builds, and keeps routine CI behavior unchanged.

None

2 – low risk. CI-only change that adds a new signing job and preserves existing artifact layout.

# Testing

- [X] None (Not running, workflow change only)

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -7,6 +7,8 @@ on:
     branches:
       - master
       - release/*
+permissions:
+  contents: read
 
 jobs:
   build-cli:


.github/workflows/package.yml
```diff
@@ -85,3 +86,89 @@ jobs:
           source_dir: build
           endpoint: https://nyc3.digitaloceanspaces.com
           destination_dir: ${{ steps.extract_branch.outputs.branch }}
+
+  build-cli-windows-signed:
+    if: ${{ startsWith(github.ref, 'refs/tags/') }}
+    name: Build and sign CLI for x86_64 Windows
+    runs-on: [self-hosted, windows, signing]
```
bfops, line 95: is this running on three separate runners?

.github/workflows/package.yml
```diff
@@ -10,6 +10,7 @@ on:
 
 jobs:
   build-cli:
+    if: ${{ !(startsWith(github.ref, 'refs/tags/') && matrix.target == 'x86_64-pc-windows-msvc') }}
```
bfops, line 15: I'm kind of inclined to just drop the windows entry from the matrix and have all the windows builds be signed. This CI workflow is meant for packaging things (presumably for potential shipping), so I think it might be less error-prone to just always do signing when we package a windows build.

(That said, I'd separately be open to not running this workflow on `master` commits.. I'm not sure that we ever really use those).

bfops (commented):
This looks pretty good to me.

I personally lean towards merging the two jobs and adding more `if`s, just to minimize copy-pasta and the risk that we don't update something going forward. I think it's fine the way it is.

.github/workflows/package.yml
```diff
@@ -10,6 +10,7 @@ on:
```
bfops, line 12:
```suggestion
permissions:
  contents: read

```
to make the bot stop complaining

.github/workflows/package.yml
rekhoff: That was intended to have it run on a system that matched all the `runs-on` properties, like is covered https://docs.github.com/en/actions/how-tos/write-workflows/choose-where-workflows-run/choose-the-runner-for-a-job

.github/workflows/package.yml
rekhoff: We only have a limited number of executable signings we can do per year (1000 to be specific), so with 3 executables, we'd be limited to 333 signing operations.
If we had confidence that we aren't going to have a problem if we remove those checks, I'm game.

.github/workflows/package.yml
bfops: oh, makes sense to me!

.github/workflows/package.yml
bfops: ah okay, then this works for me

PR #4482, opened 2026-02-26
fix(cli): read module-path from config when no generate entries exist

### LINKED ISSUE #4475: spacetime generate: "module-path" was not read correctly from the configuration file
My spacetime.json file.
``` json
{
  "module-path": "server-rust"
}
```
And it still using `spacetimedb` as `module-path`
Image

omerarslan0:
When spacetime.json has module-path at the root level but no generate block, spacetime generate ignores it and defaults to spacetimedb/.

The fallback path in exec_ex() created a CommandConfig with an empty HashMap, discarding entity-level fields from the config. Fix by passing the root config's additional_fields instead.

Closes #4475

# Description of Changes

When spacetime.json contains module-path at the root level but no generate block, spacetime generate ignores the configured path and defaults to spacetimedb/. The root cause is in exec_ex(): when get_filtered_generate_configs() returns empty (no generate entries), the fallback CommandConfig was created from an empty HashMap, discarding module-path from the root entity's additional_fields. Fixed by passing loaded.config.additional_fields.clone() instead of HashMap::new().

None. This is a bug fix that makes the existing module-path config option work as documented.

1 - Trivial. Single line change from HashMap::new() to loaded.config.additional_fields.clone() in the fallback path. The existing code paths for configs with generate entries are unaffected.

# Testing

Added regression test test_root_module_path_without_generate_entries_is_forwarded that validates module-path from root config without generate entries is properly forwarded. Existing tests remain unmodified and should continue to pass.


crates/cli/src/subcommands/generate.rs
```diff
@@ -620,7 +620,10 @@ pub async fn exec_ex(
     let (using_config, generate_configs) = if let Some(loaded) = loaded_config_ref {
         let filtered = get_filtered_generate_configs(&loaded.config, &cmd, &schema, args)?;
         if filtered.is_empty() {
-            (false, vec![CommandConfig::new(&schema, HashMap::new(), args)?])
+            // Even without generate entries, the root entity may have module-specific
+            // fields like module-path that should be forwarded.
+            let root_fields = loaded.config.additional_fields.clone();
+            (false, vec![CommandConfig::new(&schema, root_fields, args)?])
```
bfops, line 650: unfortunately I now get a failure in this line:
```
Error: Config key 'database' is not supported in the config file. Available keys: build_options, uproject_dir, namespace, module_path, unreal_module_name, js_file, include_private, wasm_file, out_dir, language
```
(this happens if `spacetime.local.json` has a `database` key)

bfops (changes requested):
Thank you for opening a PR for this! Unfortunately I'm getting a different error in this case now.

omerarslan0 (commented):
Code update

omerarslan0:
@bfops Can you check again?

omerarslan0:
Is there any progress? @bfops

bfops:
> Is there any progress? @bfops

Thank you for updating it. I will try to do another review soon!

omerarslan0:
Is there any progress? @bfops

bfops:
@sehawq I will try to get another review on this soon!

bfops:
Hi @omerarslan0, this PR doesn't build, and the CLA check also seems to be having issues due to commits under a different username.

I'm going to close this in favor of a more comprehensive fix in https://github.com/clockworklabs/SpacetimeDB/pull/5432. I've confirmed that 2/3 of the tests added in that PR don't pass on this one (even after fixing the build).

PR #4484, opened 2026-02-26
[TS] Fix error with accessorless explicit indices

### LINKED ISSUE #4445: TS modules: multiple explicit indexes without `accessor` crash with 'Cannot assign to read only property filter'
## Description

When a TypeScript module defines a table with **two or more explicit btree indexes** and the user does not provide the optional `accessor` field, all reducers PANIC at runtime with:

```
TypeError: Cannot assign to read only property 'filter' of object '#<Object>'
    in makeTableView
```

A single explicit index works fine. The crash only occurs with 2+ explicit indexes on the same table.

## Reproduction

Minimal module that crashes:

```typescript
import { schema, table, t } from 'spacetimedb/server';

const task = table(
  {
    public: true,
    indexes: [
      { name: 'task_assigned_to', algorithm: 'btree' as const, columns: ['assignedTo'] },
      { name: 'task_status', algorithm: 'btree' as const, columns: ['status'] },
    ],
  },
  {
    title: t.string(),
    status: t.string(),
    assignedTo: t.string(),
  }
);

const spacetimedb = schema({ task });
export default spacetimedb;

export const addTask = spacetimedb.reducer(
  { title: t.string() },
  (ctx, { title }) => {
    ctx.db.task.insert({ id: 0n, title, status: 'pending', assignedTo: 'nobody' });
  }
);
```

```bash
spacetime publish my-test --module-path . -s local -y
spacetime call my-test add_task '"hello"' -s local
# → PANIC: Cannot assign to read only property 'filter'
```

Remove either index and the reducer works.

## Root Cause

In `crates/bindings-typescript/src/lib/table.ts` line 428-433, explicit indexes are pushed with:

```typescript
indexes.push({
  sourceName: undefined,
  accessorName: indexOpts.accessor,  // undefined when user omits `accessor`
  algorithm,
  canonicalName: indexOpts.name,
});
```

Since `accessor` is optional in `IndexOpts`, omitting it sets `accessorName: undefined` for **every** explicit index.

Then in `crates/bindings-typescript/src/server/runtime.ts` lines 798-802:

```typescript
if (Object.hasOwn(tableView, indexDef.accessorName!)) {
  freeze(Object.assign(tableView[indexDef.accessorName!], index));
} else {
  tableView[indexDef.accessorName!] = freeze(index) as any;
}
```

1. First index: `Object.hasOwn(tableView, undefined)` → false → `tableView["undefined"] = freeze(index)` — the index object (which has a `filter` property) is frozen
2. Second index: `Object.hasOwn(tableView, undefined)` → true → `Object.assign(tableView["undefined"], index)` — tries to write `filter` onto the frozen object → **crash**

## Workaround

Add `accessor` to every explicit index definition:

```typescript
indexes: [
  { accessor: 'task_assigned_to', name: 'task_assigned_to', algorithm: 'btree', columns: ['assignedTo'] },
  { accessor: 'task_status', name: 'task_status', algorithm: 'btree', columns: ['status'] },
]
```

## Suggested Fix

Either:
- **Derive `accessorName` from `canonicalName` (the `name` field)** when `accessor` is not provided, so each index gets a unique accessor
- **Guard `runtime.ts`** against `undefined` accessor names (skip or throw a clear error)

## Environment

- SpacetimeDB CLI: 2.0.1
- npm `spacetimedb`: 2.0.1
- macOS 15.5 (Apple Silicon)
- Node.js v24.13

coolreader18:
# Description of Changes

Fixes #4445. We shouldn't have been doing `idx.accessorName!`.

1

# Testing

- [x] Verified that provided module no longer errors.

### DIFF
--- a/crates/bindings-typescript/src/lib/schema.ts
+++ b/crates/bindings-typescript/src/lib/schema.ts
@@ -109,7 +109,9 @@ export function tableToSchema<
   // This is intentionally separate from `schema.idxs`, which keeps the original
   // user-declared `IndexOpts` shape for type-level inference.
   const resolvedIndexes: UntypedIndex<AllowedCol>[] = tableDef.indexes
-    .filter(idx => typeof idx.accessorName === 'string' && idx.accessorName.length > 0)
+    .filter(
+      idx => typeof idx.accessorName === 'string' && idx.accessorName.length > 0
+    )
     .map(idx => {
       const accessorName = idx.accessorName!;
 
@@ -138,8 +140,7 @@ export function tableToSchema<
         algorithm,
         columns: columnIds.map(getColName) as AllowedCol[],
       };
-    }
-  );
+    });
 
   return {
     // For client,`schama.tableName` will always be there as canonical name.


coolreader18:
Superseded by #4525

PR #4486, opened 2026-02-27
[C#] Add support for IEnumerable<T> for Views' return types

### LINKED ISSUE #4424: [C#] Add support for `IEnumerable<T>` for Views' return types
Introduced in #4393 during the change from `Query<TRow>` to `IQuery<TRow>`, Views can no longer return `IEnumerable<T>` as a return type. Now users must return a `List<T>`, and take any existing `IEnumerable<T>` and use `.ToList()` to return the supported type.

This resulted in the documentation change seen in #4392 on `00500-cheat-sheet.md`

### Goal:
Add support for `IEnumerable<T>` as a valid return type for Views, eliminating the need for using the return type of `List<T>` and use of `.ToList()` on Query Builder results.

rekhoff:
This is the implementation to resolve #4424 
# Description of Changes
* Add support for `IEnumerable<T>` as a valid C# View return type in codegen, so view implementations can return query results without forcing `ToList()`. The generator now detects `IEnumerable<T>`, serializes it via the list serializer.
* Snapshots have been updated with a test to confirm that `IEnumerable<T>`  rather than checking it is unsupported (which we had added when removing support during the transition away from `Query<T>` usage.
* Refreshes docs/examples to reflect the new supported signature.

No breaking changes. This is additive support for an existing .NET interface.

2 - Low

# Testing
- [X] Rebuilt CLI locally, ran regression tests, and `dotnet tests` in C# module bindings.
- [X] Tested updated docs code snippets to ensure they resolve in an IDE and publish.


crates/bindings-csharp/Codegen/Module.cs
```diff
@@ -1210,11 +1230,12 @@ method.ReturnType is INamedTypeSymbol
             diag.Report(ErrorDescriptor.ViewContextParam, methodSyntax);
         }
 
-        // Validate return type: must be List<T> or T?
+        // Validate return type: must be List<T>, T?, or IEnumerable<T>
         if (
             !ReturnType.BSATNName.Contains("SpacetimeDB.BSATN.ValueOption")
             && !ReturnType.BSATNName.Contains("SpacetimeDB.BSATN.RefOption")
             && !ReturnType.BSATNName.Contains("SpacetimeDB.BSATN.List")
```
joshua-spacetime: Do both of these options implement `IEnumerable`? If so, we could just assert `IEnumerable` or `IQuery`.

rekhoff: No, neither actually implement `IEnumerable` themselves. Technically this is just validating that these are the BSATN representations. `IQuery<T>` also does not implement `IEnumerable` itself.
But if I understand the question, since we already check a bool for `ReturnsQuery` and `ReturnsEnumerable`, are you saying why don't we utilize that information in the test?
We could do that, but we'd miss checking for `T?` (the Option type return).
So I'd need to check something like 
```csharp
bool isOption =
    ReturnType.BSATNName.Contains("SpacetimeDB.BSATN.ValueOption")
    || ReturnType.BSATNName.Contains("SpacetimeDB.BSATN.RefOption");
```
And then could change the check to something like:
```csharp
if (!ReturnsQuery && !ReturnsEnumerable && !ReturnsOption)
{
    diag.Report(ErrorDescriptor.ViewInvalidReturn, methodSyntax);
}
```
If you'd prefer, and find this to be more clear.

joshua-spacetime: Yeah let's do that. I was hoping that `T?` would be `IEnumerable` too, but we can at least have `IEnumerable` subsume `List<T>` to avoid a redundant check.

rekhoff: Requested updates made

crates/bindings-csharp/Codegen/Module.cs
```diff
@@ -1192,9 +1194,40 @@ method.ReturnType is INamedTypeSymbol
...
             ReturnType = TypeUse.Parse(method, method.ReturnType, diag);
+
+            if (
+                method.ReturnType
+                    is INamedTypeSymbol
+                    {
+                        OriginalDefinition: var listDefinition,
+                        TypeArguments.Length: 1,
+                    }
+                && listDefinition.ToString() == "System.Collections.Generic.List<T>"
```
joshua-spacetime, line 1247: Do we still need this check or does `IEnumerable` make this one obsolete?

rekhoff: Yes, we still need it, because the `List<T>` is a valid return type. If we don't have that as a valid return type, any places where `List<T>` is returned, it will report that as an error. Even if there is a conversion between `List<T>` and `IEnumerable`, we accept both at this point.

joshua-spacetime:
Do we have a test that evaluates an `IEnumerable` view?

rekhoff:
> Do we have a test that evaluates an `IEnumerable` view?

We had not, because the `Codegen.Tests` are focused on the code generation side. In `crates/bindings-csharp/Codegen.Tests/fixtures/diag/Lib.cs`, there is `ViewDefIEnumerableReturnFromIter` and `ViewDefIEnumerableReturnFromFilter`, but neither of these actually test the output.

I've gone ahead and added a new test to the C# regression tests that specifically evaluates views that return `IEnumerable<T>`.

joshua-spacetime (approved):

PR #4487, opened 2026-02-27
[rust-sdk] async procedures

kistz:
# Description of Changes

This pr attempts to implement async procedures in a minimal non disruptive way.
By using a oneshot channel we can achieve this pretty easily and simultaneously get a lot of code quality improvements if the application needs it.
In my case this is a procedure which i require to be sucessful and otherwise panic so returning to the original control flow is super helpful in this case.

## Considerations

This is a minimal proof of concept without big changes.
I first wanted to see if such a approach would be feasible because @gefjon said on discord that this is a desirable feature in theory.
Of course the oddities like allowing the `async_fn_in_trait` lint can be circumvented and patched pretty easily on my side i just chose to not bother for now to gather feedback :>

Also this should probabaly done for reducers aswell but for the same reasons i didnt for now but i can add it either in this pr or a followup.

## Before and after
image

None. additive changes.

i would argue 1. very minimal change which reuses the existing code as much as possible only adding one async interaction.

# Testing

- [x] Tested this in my project and it works.

### DIFF
--- a/crates/codegen/src/rust.rs
+++ b/crates/codegen/src/rust.rs
@@ -529,6 +529,7 @@ impl {func_name} for super::RemoteReducers {{
         let procedure_name = procedure.name.deref();
         let func_name = procedure_function_name(procedure);
         let func_name_with_callback = procedure_function_with_callback_name(procedure);
+        let func_name_with_async = procedure_function_with_async_name(procedure);
         let args_type = function_args_type_name(&procedure.accessor_name);
         let res_ty_name = type_name(module, &procedure.return_type_for_generate);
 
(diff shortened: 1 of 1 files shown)


egormanga:
Where (on which executor) does this async function get ran?

kistz:
@egormanga the rust-sdk depends on tokio internally. so i just reused this internal dependency e.g. tokio :>

egormanga:
Oh that's for the client, I see now.

gefjon:
Thanks for the PR. This looks good, and I think we'd like to merge something pretty similar, but I don't know if we can accept a community PR on this feature. Our SDKs require a lot of design work, as we have to carefully balance keeping them consistent between our supported languages with making them idiomatic and ergonomic. Adding async to our SDKs is going to require a fair amount of coordination between the maintainers of our Rust, TypeScript and C# SDKs, and I just don't think it's feasible to do that with a community member in the loop.

kistz:
Hm ok that's a bummer :( but understandable. 
Hope it at least provided some value :>
Thanks for taking the time to consider 👍

egormanga:
@kistz I'll make sure to cherry-pick this PR into my fork, so it won't be lost to time, as long as you don't mind it. This might gain us additional clues on its usability and perhaps provide valuable feedback to the Spacetime design team.

kistz:
@egormanga sure i dont mind do with it what you want :>

egormanga:
Ty! @cloutiertyler, can I assume PRs code is NOT subject to BSL upon being rejected from upstream?

PR #4489, opened 2026-02-27
Version upgrade 2.0.3

jdetter:
# Description of Changes

- Version upgrade 2.0.3

- None, this is a version bump

- 1 - this is a version bump

# Testing

- [x] License has been updated
- [x] Version number is correct (2.0.3)


templates/remix-ts/.gitignore
```diff
@@ -0,0 +1 @@
+build/
```
jdetter, line 1: This is a new addition - for some reason we now get a `build/` directory in the remix directory after running the version upgrade tool.

cloutiertyler (approved):

PR #4504, opened 2026-02-28
`spacetime.json` updates

cloutiertyler:
# Description of Changes

Adds a new reference documentation page for `spacetime.json` and fixes several bugs where the CLI behavior diverged from the proposal.

**Docs:**
- New page at `/cli-reference/spacetime-json` covering config structure, field reference, generate configuration, children/inheritance, `spacetime dev` config, database selection, flag overrides, `--no-config`, `--env`/environments, config file discovery, and editor support

**Bug fixes:**
- `generate` was incorrectly inherited by child databases. A child with a different `module-path` would silently inherit the parent's generate entries, causing bindings to be written to the wrong output directory. Generate is now never inherited, matching the proposal.
- The source conflict rule for `module-path`/`bin-path`/`js-path` was not implemented during inheritance. A child specifying `module-path` could still inherit `bin-path` from the parent. Now, if a child specifies any module source, the others are not inherited.
- `--num-replicas` was not marked as a per-database override, so it could be used with multiple databases selected without error.

None. These are bug fixes aligning the implementation with the intended behavior from the proposal:
- `generate` inheritance was never documented or intended
- The source conflict rule was specified in the proposal but not implemented
- `--num-replicas` as per-database is consistent with the other module-source flags

2 - The changes are small and well-scoped. The generate inheritance fix simplifies the code (removes a parameter). The source conflict rule adds a straightforward check during field inheritance. Tests have been updated to match.

# Testing

- [x] All 136 existing CLI tests pass
- [x] Updated tests for generate non-inheritance behavior
- [x] Docs site builds successfully, page renders in sidebar
- [ ] Manual test: verify a child with a different `module-path` no longer inherits parent's `generate`
- [ ] Manual test: verify `--num-replicas` errors when multiple databases are selected


futile:
These new docs allowed me to quickly understand how the config files work, and the order/hierarchy in which everything is applied. I thought that might be helpful feedback :) Thanks a lot!

crates/cli/src/spacetime_config.rs
```diff
@@ -3084,9 +3089,10 @@ mod tests {
     }
 
     #[test]
-    fn test_generate_dedup_with_inherited_generate() {
-        // Two sibling databases sharing parent's generate + same module path
-        // should deduplicate to a single generate entry
+    fn test_generate_not_inherited_for_children_sharing_module() {
+        // Even when children share the parent's module-path, generate is not inherited.
+        // Deduplication in generate.rs handles the common case; inheritance would be
+        // dangerous when a child overrides module-path.
```
gefjon: Please move test description to outer doc comment.


```suggestion
    /// Even when children share the parent's module-path, generate is not inherited.
    ///
    /// Deduplication in generate.rs handles the common case; inheritance would be
    /// dangerous when a child overrides module-path.
    fn test_generate_not_inherited_for_children_sharing_module() {
```

docs/docs/00300-resources/01000-reference/00100-cli-reference/00300-spacetime-json.md
```diff
@@ -0,0 +1,339 @@
...
+|-------|------|-----------|-------------|
+| `database` | string | No | Database name or identity (required) |
+| `module-path` | string | Yes\* | Path to module source directory |
+| `bin-path` | string | Yes\* | Path to precompiled WASM binary |
+| `js-path` | string | Yes\* | Path to bundled JavaScript file |
+| `server` | string | Yes | Server nickname, domain, or URL |
+| `build-options` | string | Yes | Options passed to the build command |
+| `break-clients` | boolean | Yes | Allow breaking changes |
+| `num-replicas` | number | Yes | Number of database replicas |
+| `anonymous` | boolean | Yes | Use anonymous identity |
```
gefjon, line 41: Does this apply to all CLI operations, or only specific ones? Is this gonna try to anonymously publish my database?

cloutiertyler: Yes it is shared across these operations:

image

So yes it will to publish your database anonymously.

PR #4511, opened 2026-03-01
Upgrade prompt is skipped when `-y` is passed

jdetter:
# Description of Changes

This prompt is now skipped when `-y` is passed:

image

None

1 - this just skips a prompt in the CLI

# Testing

- [x] publish a 1.0 database to maincloud, then upgrade it to 2.0 and pass `-y`. You should no longer get the upgrade prompt.
- [x] publish a 1.0 database to maincloud, then upgrade it to 2.0 without passing `-y`. You should still get the upgrade prompt.

### DIFF
--- a/crates/cli/src/subcommands/publish.rs
+++ b/crates/cli/src/subcommands/publish.rs
@@ -246,13 +246,17 @@ fn confirm_and_clear(
     Ok(builder)
 }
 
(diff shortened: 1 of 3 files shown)


bfops (approved):
LGTM after our discussion. I worry that this may opt users into upgrades that didn't expect it, but that _is_ what `--yes` means!

Issue #4514, opened 2026-03-02
C# module build pipeline depends on deprecated .NET workload; AOT compilation path needed


cloutiertyler:
## Summary

C# modules in SpacetimeDB are compiled to WebAssembly using the .NET 8 Mono interpreter via the `wasi-experimental` workload. Microsoft has deprecated this workload and removed it from .NET 9+. There is no replacement. This means the entire C# module build pipeline has an expiration date: .NET 8 LTS support ends November 2026, and there is no way to build C# modules on any newer version of .NET using the current approach.

Separately, the Mono interpreter is slow, and there is no supported path to ahead-of-time (AOT) compilation for `wasi-wasm` targets that works across all platforms.

This issue documents the full situation, the options we've explored, and the path forward.

## Current state

C# modules are built with `dotnet publish` targeting `wasi-wasm` using the `wasi-experimental` workload on .NET 8. This workload provides the Mono runtime compiled for `wasi-wasm` and the WASI SDK build tooling. Without it, `dotnet publish` with `RuntimeIdentifier=wasi-wasm` does not work at all.

The output is a single `.wasm` file that bundles the Mono runtime, which interprets C# IL bytecode at execution time. We don't actually use WASI. The `wasi-wasm` target is just a means to produce a `.wasm` file, and all WASI imports are stubbed out in `bindings.c`. Our modules use custom host imports/exports defined by the SpacetimeDB ABI.

Key files:
- `crates/bindings-csharp/Runtime/build/SpacetimeDB.Runtime.props` - MSBuild properties
- `crates/bindings-csharp/Runtime/build/SpacetimeDB.Runtime.targets` - MSBuild targets, WASI SDK download, AOT config
- `crates/bindings-csharp/Runtime/bindings.c` - C glue layer for imports/exports and WASI shims
- `crates/bindings-csharp/Runtime/Internal/FFI.cs` - C# FFI declarations
- `crates/cli/src/tasks/csharp.rs` - CLI build orchestration

## The existential problem

The `wasi-experimental` workload is the **only** way to build C# modules today. It was removed from .NET 9 (Feb 2025) and will not come back. Microsoft has no WASI plans for .NET 10 or 11. The WASI tracking issue milestone was moved to "Future" with no committed timeline.

.NET 8 LTS support ends November 2026. After that, there is no supported .NET version that can build C# modules using the current approach.

[...]

PR #713, opened 2024-01-10
C#: add experimental NativeAOT-LLVM support

RReverser:
(PR is marked as a draft for now because it's blocked on #706)

# Description of Changes

Adds alternative experimental mode for building C# modules. For now, it requires user to:

1. Use Windows.
2. Manually install Emscripten SDK (https://github.com/emscripten-core/emsdk).
3. Opt-in to experimental compilation mode via `EXPERIMENTAL_WASM_AOT=1` environment variable.

The development experience will be improved in the future, but for now this should be enough to get some more testing.

Initial numbers suggest it's roughly 15-20x faster in simple cases (where module language overhead dominates over DB operations).

I already published those packages as prerelease versions, so, once this PR is merged and STDB can correctly find Wasm file produced by this mode, users should be able to test via

```xml
<PackageReference Include="SpacetimeDB.Codegen" Version="0.8.1-naot-experimental" />
 <PackageReference Include="SpacetimeDB.Runtime" Version="0.8.2-naot-experimental" />
```

in their `.csproj` projects.

# API and ABI breaking changes

If this is an API or ABI breaking change, please apply the
corresponding GitHub label.

# Expected complexity level and risk

3 - this is somewhat risky because, in order to add conditional compilation, I had to modify paths that affect regular compilation too. It seems to work correctly, but testing by more people would be appreciated.

*How complicated do you think these changes are? Grade on a scale from 1 to 5,
where 1 is a trivial change, and 5 is a deep-reaching and complex change.*

*This complexity rating applies not only to the complexity apparent in the diff,
but also to its interactions with existing and future code.*

*If you answered more than a 2, explain what is complex about the PR,
and what other components it interacts with in potentially concerning ways.*

PR #4515, opened 2026-03-02
Update NativeAOT-LLVM infrastructure to current ABI

cloutiertyler:
## Summary

- Update the experimental NativeAOT-LLVM build path (`EXPERIMENTAL_WASM_AOT=1`) to include all host function imports from ABI versions 10.0 through 10.4
- Fix the compiler package reference to work on both Windows x64 and Linux x64 (was hardcoded to Windows only)
- Add a CI smoketest to verify AOT builds work on Linux x64

## Context

See #4514 for the full writeup on the C# AOT situation. The `wasi-experimental` workload that all C# module builds depend on is deprecated and removed from .NET 9+. NativeAOT-LLVM is the recommended path forward for ahead-of-time compilation of C# to WebAssembly.

The existing NativeAOT-LLVM support (added by RReverser in #713) was stale: missing imports added since then and a Windows-only package reference.

## Changes

**`SpacetimeDB.Runtime.targets`:**
- Add 10 missing `WasmImport` declarations across spacetime_10.0 through 10.4
- Replace `runtime.win-x64.Microsoft.DotNet.ILCompiler.LLVM` with `runtime.$(NETCoreSdkPortableRuntimeIdentifier).Microsoft.DotNet.ILCompiler.LLVM` so it resolves correctly on Linux x64 as well
- Use explicit version strings instead of the `$(SpacetimeNamespace)` variable

**`ci.yml`:**
- Add AOT build smoketest step in the `csharp-testsuite` job

## Test plan

- [x] CI smoketest passes: `EXPERIMENTAL_WASM_AOT=1 dotnet publish -c Release` builds successfully on Linux x64
- [ ] Existing C# tests continue to pass (no changes to the default interpreter path)


rekhoff:
After the revisions made in #4601, testing locally on this branch, using the `keynote-2` template benchmark with an equivalent module authored in C#, I am seeing around nearly a 200% improvement in performance using `NativeAOT-LLVM`.
As part of $4601, we introduced a `NATIVEAOT-LLVM.md‎` explaining how to enable `NativeAOT-LLVM` on an existing C# module.
Additionally, we found that [runtime.linux-x64.microsoft.dotnet.ilcompiler.llvm](https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-experimental/nuget/v3/flat2/runtime.linux-x64.microsoft.dotnet.ilcompiler.llvm/index.json) does not include any .NET 8 versions in it's supported list. This differs from [runtime.win-x64.microsoft.dotnet.ilcompiler.llvm](https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-experimental/nuget/v3/flat2/runtime.win-x64.microsoft.dotnet.ilcompiler.llvm/index.json) which clearly shows Windows has .Net 8 support.
While this does align with past efforts in #713, it means newly added test had to be altered to force these tests against a Windows machine.
This also means that, without creating out own supported Linux distribution for .NET 8, Linux support will either need to wait until migrate to a more-recent major .NET version.

cloutiertyler:
This is being picked up by @rekhoff @JasonAtClockwork and @lisandroct

.github/workflows/ci.yml
```diff
@@ -893,6 +893,42 @@ jobs:
               exit 1
           }
 
+  # NativeAOT-LLVM smoketest runs on Windows because runtime.linux-x64.Microsoft.DotNet.ILCompiler.LLVM
+  # does not have 8.0.0 versions available on the dotnet-experimental NuGet feed.
+  csharp-aot-smoketest:
```
bfops, line 898: this should be a part of `cargo ci`

rekhoff: Commit [84726cf](https://github.com/clockworklabs/SpacetimeDB/pull/4515/commits/84726cf4487e815e22a45aaaaa7e5714738ab365) moves this to the `smoketests` CI workflow. It only runs under Windows, since NativeAOT-LLVM for .Net 8 is limited to Windows support.

sdks/csharp/tools~/write-nuget-config.sh
```diff
@@ -15,6 +15,8 @@ cat >NuGet.Config <<EOF
 <configuration>
   <packageSources>
     <clear />
+    <!-- Experimental NuGet feed for Microsoft.DotNet.ILCompiler.LLVM packages -->
+	<add key="dotnet-experimental" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-experimental/nuget/v3/index.json" />
```
jdetter, line 19: When you're not building for native aot will any of these changes cause issues?

crates/smoketests/tests/smoketests/cli/publish.rs
```diff
@@ -223,9 +223,7 @@ fn cli_publish_with_config_but_no_match_uses_cli_args() {
 
     // Create a config with a different database name
     let config_content = r#"{
-  "publish": {
-    "database": "config-db-name"
-  }
```
jdetter, line 228: why did this test need to change?

crates/smoketests/tests/smoketests/csharp_aot_module.rs
```diff
@@ -0,0 +1,50 @@
...
+    if std::env::consts::OS != "windows" {
+        eprintln!("Skipping AOT test - NativeAOT-LLVM for .NET 8 only available on Windows");
+        return;
+    }
+
+    // Check for emscripten - skip if not available
+    let emscripten_check = Command::new("emcc").arg("--version").output();
+    if emscripten_check.is_err() || !emscripten_check.unwrap().status.success() {
+        eprintln!("Skipping AOT test - emscripten not available");
+        return;
```
jdetter: Maybe this should fail instead of skipping? This dependency is super easy to install and I wouldn't want this test to be getting skipped unintentionally just because we're missing emscripten. What do you think?

jdetter: We could also include the download link in the error if we wanted to be extra helpful for whoever will hit this in the future: https://emscripten.org/docs/getting_started/downloads.html

jdetter (commented):
Few questions on this but no issues so far - I also tested this again and it works

crates/smoketests/tests/smoketests/cli/publish.rs
rekhoff: Because this was actually failing to find the config prior to my changes in #4672

Previously, the config loading is only the current working directory which was the repo root, and not factoring in the `--module-path`. We can see this here:
https://github.com/clockworklabs/SpacetimeDB/blob/af37985a4d634e7db8f7dadb09c7a4a91f5cd907/crates/cli/src/subcommands/publish.rs#L302
And that calls:
https://github.com/clockworklabs/SpacetimeDB/blob/af37985a4d634e7db8f7dadb09c7a4a91f5cd907/crates/cli/src/spacetime_config.rs#L1017-L1019
Which we see is looking for the current directory.

This just meant that the `config_content` was never actually found, and that just meant the value wasn't used.

But in order to resolve an issue in #4672 where I had `--module-path` used, the `--native-aot` flag was getting ignored in the `spacetime.json` file. So as part of the fix, https://github.com/clockworklabs/SpacetimeDB/pull/4672/changes/590a878dc9c19715409f5468618fdb319919c7ab added logic that could also check the module path in the config was not found in the working directory.

This then caused this test to fail because it then turns out that `publish` is not one of the top-level commands allowed in a config schema, as seen here:
https://github.com/clockworklabs/SpacetimeDB/blob/af37985a4d634e7db8f7dadb09c7a4a91f5cd907/crates/cli/src/subcommands/publish.rs#L23-L42

So now with `config_content` being evaluated, the `publish` command that wasn't valid, was causing a CI failure. And that ultimately meant fixing the config used in that test.

sdks/csharp/tools~/write-nuget-config.sh
rekhoff: Correct, I would not anticipate those causing any issues. Currently `csharp-testsuite` and `unity-testsuite` CI jobs both use this config, but don't set `EXPERIMENTAL_WASM_AOT=1` . For good measure, I just double-checked with a non-NativeAOT build of Blackholio with a C# server and Unity client.
Those package source mappings are additive, and with the ItemGroup for these in any `.csproj` being hidden behind the `EXPERIMENTAL_WASM_AOT` environment variable, they should just get ignored.

crates/smoketests/tests/smoketests/csharp_aot_module.rs
rekhoff: Fair enough, I've pushed a changed to cause it to fail rather than just get skipped.

crates/smoketests/tests/smoketests/cli/publish.rs
jdetter: awesome, thanks for the context 👍

jdetter (approved):
Tested working, code looks good, thanks Ryan 👍

PR #4525, opened 2026-03-03
Make `accessor` required for table-level index defs in typescript

### LINKED ISSUE #4520: TypeError in makeTableView due to 'filter' property conflict
## Description

When publishing a SpacetimeDB TypeScript module containing a table with one or more `btree` indexes, the SpacetimeDB JavaScript runtime throws a `TypeError: Cannot assign to read only property 'filter'` during database initialization. This causes the module's `init` reducer to fail, making the database unusable.

## Environment

- **SpacetimeDB CLI Version**: `2.0.2`
- **SpacetimeDB Server Version**: `2.0.2`
- **SpacetimeDB SDK Version**: `^2.0.2` (npm `spacetimedb` package)
- **Runtime Environment**: TypeScript/JavaScript Server Module

## Steps to Reproduce

1. Initialize a new SpacetimeDB TypeScript module.
2. Define a table with at least one `btree` index in the `schema.ts`:

```typescript
import { schema, table, t } from 'spacetimedb/server';

const CelestialCoords = t.object('CelestialCoords', {
    declination: t.f64(),
    rightAscension: t.f64(),
});

export const Reply = table({
    name: 'reply',
    public: true,
    indexes: [
        { name: 'reply_parent_star_id', algorithm: 'btree', columns: ['parentStarId'] },
        { name: 'reply_author_id', algorithm: 'btree', columns: ['authorId'] },
    ],
}, {
    parentStarId: t.u64(),
    authorId: t.identity(),
    message: t.string(),
    coords: CelestialCoords,
    createdAt: t.timestamp(),
});

export default schema({
    reply: Reply,
});
```

3. Build and publish the module locally:
   ```bash
   spacetime publish mydb --server http://localhost:3000
   ```

## Expected Behavior

The module publishes successfully and initializes the database tables and indexes, allowing index-based queries like `ctx.db.reply.reply_parent_star_id.filter(id)`.

## Actual Behavior

The database creation reports success, but the `init` reducer immediately fails with a fatal error in the server logs:

```
2026-03-02T16:42:56.580357Z  INFO crates\core\src\host\v8\error.rs:549: reducer "init" runtime error: Uncaught TypeError: Cannot assign to read only property 'filter' of object '#<Object>'

        at makeTableView (node_modules/spacetimedb/dist/server/index.mjs:6849:21)
        at <anonymous> (node_modules/spacetimedb/dist/server/index.mjs:6423:11)
```

The error originates in the JS runtime inside the `makeTableView` function. When initializing the Table View object, the SDK attempts to attach index accessor properties to it. For `btree` indexes, it tries to define a `.filter()` method. 

However, because the Table View object inherits from the JavaScript prototype chain (or is created with properties configured as `writable: false`), the property name `'filter'` conflicts with the built-in, read-only `Array.prototype.filter` or a mathematically frozen property on the base view object.

The exact code path attempts an assignment like `tableView[indexName].filter = ...` or `tableView.filter = ...` where `filter` is already a protected identifier.

## Workaround

The workaround is to **remove all `btree` indexes** from the table definition in the server schema, and fallback to using full table scans with `iter()` and manual array filtering:

```typescript
// Workaround: Remove indexes array from OPTIONS
export const Reply = table({
    name: 'reply',
    public: true,
}, { ... });

// Workaround query
const existingReplies = [...ctx.db.reply.iter()].filter(r => r.parentStarId === parentStarId);
```

joshua-spacetime:
# Description of Changes

This patch contains two main changes:

1. It makes `accessor` a required argument for table-level index defs in typescript to align with rust.

2. It removes unsound index typing in the TypeScript bindings by splitting index data into two explicit representations:

    - `indexes`: declarative user config (`IndexOpts`) used for type inference
    - `resolvedIndexes`: normalized runtime metadata (`UntypedIndex`) derived from `RawTableDefV10`

    `TableCacheImpl` now builds index accessors from `resolvedIndexes` instead of re-casting `indexes` at runtime.

    This addressed the following comment:
    ```
    // TODO: horrible horrible horrible. we smuggle this `Array<UntypedIndex>`
    // by casting it to an `Array<IndexOpts>` as `TableToSchema` expects.
    // This is then used in `TableCacheImpl.constructor` and who knows where else.
    // We should stop lying about our types.
    ```

    > Why?

    We were conflating two different concepts under `tableDef.indexes`:

      1. declared table-level index options authored by users
      2. resolved runtime index definitions (including field-level inferred indexes)

    This required unsafe casts (`T['idxs']` / `UntypedIndex`) and made the type model unsound.

Note, (2) was largely ai assisted.

Technically breaks the module api, although I believe this is the behavior that is outlined in the spec, and so the current behavior should really be considered a bug.

2

# Testing

Added unit tests covering the following:
1. Table-level explicit index without accessor throws.
2. Table-level duplicate accessor throws.
3. Table-level explicit accessor is accepted and used.
4. Field-level `.index(...)` derives accessor from the field name (`displayName`).

### DIFF
--- a/crates/bindings-typescript/src/lib/indexes.ts
+++ b/crates/bindings-typescript/src/lib/indexes.ts
@@ -9,7 +9,7 @@ import type { ColumnIsUnique } from './constraints';
  * existing column names are referenced.
  */
 export type IndexOpts<AllowedCol extends string> = {
-  accessor?: string;
+  accessor: string;
   name?: string;
 } & (
   | { algorithm: 'btree'; columns: readonly AllowedCol[] }
--- a/crates/bindings-typescript/src/lib/schema.ts
+++ b/crates/bindings-typescript/src/lib/schema.ts
@@ -74,14 +74,14 @@ export function tablesToSchema<
   // `TablesToSchema<T>['tables']` is intentionally readonly in the public type,
   // but we need a mutable builder while materializing it from entries.
   type MutableTableDefs = {
-    -readonly [AccName in keyof TablesToSchema<T>['tables']]: TablesToSchema<
-      T
-    >['tables'][AccName];
+    -readonly [AccName in keyof TablesToSchema<T>['tables']]:
+      TablesToSchema<T>['tables'][AccName];
   };
   const tableDefs = Object.create(null) as MutableTableDefs;
-  for (const [accName, schema] of Object.entries(tables) as Array<
-    [keyof T & string, T[keyof T & string]]
-  >) {
+  for (const [accName, schema] of Object.entries(tables) as [
+    keyof T & string,
+    T[keyof T & string],
+  ][]) {
     tableDefs[accName] = tableToSchema(
       accName,
       schema,
--- a/crates/bindings-typescript/src/lib/schema.ts
+++ b/crates/bindings-typescript/src/lib/schema.ts
@@ -119,7 +119,9 @@ export function tableToSchema<
       }
 
       const columnIds =
-        idx.algorithm.tag === 'Direct' ? [idx.algorithm.value] : idx.algorithm.value;
+        idx.algorithm.tag === 'Direct'
+          ? [idx.algorithm.value]
+          : idx.algorithm.value;
 
       const unique = tableDef.constraints.some(
         c =>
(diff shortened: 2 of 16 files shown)


Shubham8287 (approved):

joshua-spacetime:
@Shubham8287 has approved (1). I've added @coolreader18 for review of (2).

PR #4527, opened 2026-03-03
Implement IEquatable in autogenerated C# enums

GEFNENEN:
# Description of Changes

This PR fixes a compilation error in C# bindings when using Rust enums as primary keys. The issue occurs because generated C# enums don't implement the `IEquatable<T>` interface, which is required by SpacetimeDB's indexing system.

The fix modifies the `autogen_csharp_plain_enum` function in `crates/codegen/src/csharp.rs` to:
- Add `IEquatable<T>` to the enum declaration
- Generate the `Equals()` method implementation

This only affects plain enums (without payload) and does not impact tagged enums which use a different code generation path (`autogen_csharp_sum`).

No API or ABI breaking changes. This is a purely additive change that only affects generated C# code to implement an additional interface.

**Complexity: 1/5** (Trivial change)

This is a minimal change involving only 2-3 lines of code in a single function. The modification is straightforward:
- Adds an interface to the enum declaration
- Adds a simple Equals() method implementation

The change has low risk because:
- It only affects C# code generation for plain enums
- Tagged enums use a separate code path (`autogen_csharp_sum`) and are unaffected
- The IEquatable<T> implementation is a standard pattern in C#
- The existing codebase already has logic to check if enums are equatable (`IsEquatable` property in `ColumnDeclaration`)

# Testing

- [x] Verified that the modified function generates C# enums with IEquatable<T> interface
- [x] Confirmed that generated enums compile without CS0315 errors when used as primary keys
- [ ] Reviewer should test with a Rust module containing enum primary keys to ensure the generated C# SDK compiles and works correctly
- [ ] Reviewer should verify that existing enum functionality remains unaffected (both plain and tagged enums)

The change should be tested by:
1. Creating a Rust module with an enum primary key
2. Generating the C# bindings
3. Verifying the generated enum implements IEquatable<T>
4. Confirming the C# project compiles without errors
5. Testing that the enum works correctly as a primary key in database operations


cloutiertyler (dismissed):
I support this in theory, although I imagine a lot of our CI is going to complain until we regenerate.

bfops:
@GEFNENEN are you able to take a stab at addressing those CI errors?

PR #4541, opened 2026-03-03
Make accessor required for table-level index defs in C#

joshua-spacetime:
# Description of Changes

Make `Accessor` a required argument for table-level index defs in C# to align with rust and typescript. The same change was done for typescript in https://github.com/clockworklabs/SpacetimeDB/pull/4525.

Technically breaks the module api, although I believe this is the behavior that is outlined in the spec, and so the current behavior should really be considered a bug.

1

# Testing

Added negative compile tests


rekhoff (approved):
Looks good to me. I have not tested this locally, but the test cases provided appear to validate correctly.

PR #4543, opened 2026-03-04
Update edition in `.rustfmt.toml` and pre-commit hook

bfops:
# Description of Changes

We've updated to 2024 (https://github.com/clockworklabs/SpacetimeDB/pull/3802).

I'm just adding `edition` to `.rustfmt.toml`, and removing the hardcoded edition from the pre-commit hook. This fixes the pre-commit hook complaining about us using Rust 2024 features.

None

1

# Testing
- [x] `rustfmt crates/auth/src/identity.rs` now succeeds for me


.rustfmt.toml
```diff
@@ -1,2 +1,3 @@
 max_width = 120
 style_edition = "2021"
```
bfops, line 2: changing `style_edition` didn't work for me 🤷

PR #4549, opened 2026-03-04
Persist host type update and honor stored value after restart

kim:
Fixes two issues that would prevent updating a database while also changing the host type:

- We never actually updated the `ModuleKind` in `st_module` (hardcoded to wasm)
- We never actually honored the value from `st_module` when instantiating a module

To do so, the `Program` type from the datastore crate now carries the `ModuleKind`, forcing call sites to make a decision.

Small adjustments to the smoketests / guard crate where made when writing the test for this.

1

# Testing

- [x] Added smoketest


kim:
Hm, we are also not updating the host type in the system table.

joshua-spacetime:
> TODO: Trying to understand how to publish a JS module from the new smoketests framework.

See https://github.com/clockworklabs/SpacetimeDB/blob/cacfd738657896e415bc0958b96e936147d7e92b/crates/smoketests/tests/smoketests/views.rs#L569-L573

PR #4552, opened 2026-03-04
CI - Fix package job

bfops:
# Description of Changes

We were unable to test https://github.com/clockworklabs/SpacetimeDB/pull/4473 because #githubworkflowthings, and it turns out we broke the job.

Here I fixed the job and merge the two jobs into one.

**Signing still doesn't work so I've disabled the signing part for now.**

None. CI only.

1

# Testing

I ran a test job: https://github.com/clockworklabs/SpacetimeDB/actions/runs/22691343426/job/65787327828

- [x] Package job succeeds for non-windows
- [x] Package job succeeds on windows


.github/workflows/package.yml
```diff
@@ -55,70 +54,16 @@ jobs:
         run: |
           cargo build --release --target ${{ matrix.target }} -p spacetimedb-cli -p spacetimedb-standalone -p spacetimedb-update
 
-      - name: Package (unix)
```
bfops, line 58: the diff here is garbage. tl;dr I just moved the two code-signing-related blocks into the main job and added the "if windows" condition

.github/workflows/package.yml
```diff
@@ -11,7 +11,6 @@ permissions:
 
 jobs:
   build-cli:
-    if: ${{ !(startsWith(github.ref, 'refs/tags/') && matrix.target == 'x86_64-pc-windows-msvc') }}
```
bfops, line 14: I merged the windows job into this one

.github/workflows/package.yml
```diff
@@ -51,81 +50,45 @@ jobs:
...
+          [IO.File]::WriteAllBytes("digicert.pfx", [Convert]::FromBase64String($env:DIGICERT_CERT_B64))
 
-      - name: Sign binaries
+      - name: Compile
+        run: |
+          cargo build --release --target ${{ matrix.target }} -p spacetimedb-cli -p spacetimedb-standalone -p spacetimedb-update
+
+      - name: Sign binaries for Windows
+        # Disabled for now since the current flow isn't working.
+        if: false
```
bfops, line 83: note that this step is disabled for now

PR #4564, opened 2026-03-05
Fix stale `--project-path` flag in templates

bfops:
# Description of Changes

We apparently have stale usage of `--project-path` in a ton of our templates. This was renamed to `module-path` a while ago, but it looks like that was only partially fixed in templates.

None

1

# Testing

I don't think it's _more_ broken :shrug:

### DIFF
--- a/templates/svelte-ts/package.json
+++ b/templates/svelte-ts/package.json
@@ -7,7 +7,7 @@
     "dev": "vite",
     "build": "vite build",
     "preview": "vite preview",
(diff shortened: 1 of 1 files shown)


templates/svelte-ts/package.json
```diff
@@ -8,9 +8,9 @@
     "build": "vite build",
     "preview": "vite preview",
     "generate": "pnpm --dir spacetimedb install && cargo run -p gen-bindings -- --out-dir src/module_bindings --project-path spacetimedb && prettier --write src/module_bindings",
```
bfops:
```suggestion
    "generate": "pnpm --dir spacetimedb install && cargo run -p gen-bindings -- --out-dir src/module_bindings --module-path spacetimedb && prettier --write src/module_bindings",
```

PR #4300, opened 2026-02-15
Make cargo ci regen exhaustive across all committed codegen

cloutiertyler:
## Summary

- Restructures `cargo ci regen` into 5 logical categories covering **all** committed codegen in the repo:
  1. **SDK test bindings** — Rust (5 pairs), C# regression tests (3 pairs), Unreal (2 pairs)
  2. **Demo project bindings** — Blackholio Unity C# + Unreal C++
  3. **Starter template bindings** — C#, Rust, TS (auto-discovered via pnpm), deno-ts, Unreal QuickstartChat
  4. **SDK internal codegen** — TS client-api, C# ClientApi, TS test-app, TS test-react-router-app
  5. **Moduledef type bindings** — C#, TS, C++, Unreal SDK
- Adds `generate` scripts to 5 TS templates that were missing them (browser-ts, bun-ts, nodejs-ts, tanstack-ts, keynote-2)
- Adds 5 missing templates to `pnpm-workspace.yaml` (bun-ts, nodejs-ts, nextjs-ts, remix-ts, keynote-2)
- Adds **autogen coverage check** (`--check` mode) that scans the repo for autogen markers and errors if any directory isn't in the `REGEN_DIRS` allowlist
- Adds **template generate script check** that catches templates with committed bindings but no `generate` script
- Adds **version-only change revert** so running `cargo ci regen` doesn't produce noisy diffs when the only change is the commit hash in codegen version comments
- Adds **Windows global.json symlink fix** that temporarily patches broken symlinks (from `core.symlinks=false`) during regen and restores them after
- Also regenerates stale codegen (C++ moduledef, C# procedure-client, C# ClientApi, TS test-react-router-app, chat templates)

## Test plan

- [ ] `cargo build --manifest-path tools/ci/Cargo.toml` compiles
- [ ] `cargo ci regen` runs to completion and generates all bindings
- [ ] `cargo ci regen --check` passes on a clean state (no stale bindings)
- [ ] Version-only changes are correctly reverted (no commit-hash-only diffs)
- [ ] Autogen coverage check catches any uncovered directories
- [ ] Template generate coverage check catches missing `generate` scripts

PR #4565, opened 2026-03-05
Remove broken `generate` command from templates

bfops:
# Description of Changes

We have this `generate` command in our templates that only makes sense in the context of our repo. However, we use these templates to initialize user projects, where the command is just broken.

This PR removes that command.

None

1

# Testing

None


bfops:
Doing this in https://github.com/clockworklabs/SpacetimeDB/pull/4300 instead

PR #4567, opened 2026-03-05
Export Range/Bound from the spacetimedb/server in the Typescript sdk

dusk125:
# Description of Changes

This change exports the Range (and Bound) types so that ranged filtering can be done from Typescript modules. I chose to do export * so that both Range and Bounds are included and I chose to not export as types since the user will need to instantiate a Range for use in a query

None

1
It seems like this export was just missed as the [documentation for range queries](https://spacetimedb.com/docs/tables/indexes/#range-queries) makes reference to them, but I was unable to import them as it recommends (or at all)

# Testing

At the time of writing, I haven't tested the fix yet, I'll see if I can get a local copy of spacetime built and validate I can now imports Ranges.

### DIFF
--- a/.github/workflows/pr_approval_check.yml
+++ b/.github/workflows/pr_approval_check.yml
@@ -1,7 +1,17 @@
 name: Review Checks
 
+# SECURITY: This workflow uses pull_request_target so that it has write access to
+# set commit statuses on external (fork) PRs. pull_request_target runs in the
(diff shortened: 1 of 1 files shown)


cloutiertyler:
@dusk125 Could you make these explicit exports please? If so I can merge it!

dusk125:
@cloutiertyler should be good for review again! Thanks!

bfops:
Thank you for this @dusk125 !

PR #4573, opened 2026-03-06
[TS]: Primary keys for query builder views

joshua-spacetime:
# Description of Changes

Equivalent changes to https://github.com/clockworklabs/SpacetimeDB/pull/4572, but for typescript, and minus any server side changes.

Query builder views can now be annotated with the `.query()` return type, in addition to the pre-existing `.array()` return type. If `.query()` is used, the view will have a primary key if the underlying table that it returns has one. This means client-side `onUpdate` callbacks will be generated.

None. `t.array()` return type views are still supported. 

2

# Testing

The equivalent typescript versions of the sdk tests added in https://github.com/clockworklabs/SpacetimeDB/pull/4572.


joshua-spacetime:
Superseded by #5111

PR #4575, opened 2026-03-06
Replace unwrap() with ? in control_db for proper error propagation

Shiven0504:
## Summary
- Replace 5 `unwrap()` calls in `control_db.rs` with `?` to propagate errors instead of panicking at runtime (`bsatn::to_vec`, `bsatn::from_slice`, and `get_replicas`)
- Change `get_leader_replica_by_database` return type from `Option<Replica>` to `Result<Option<Replica>>` across the `ControlStateDelegate` trait and all implementations/call sites
- Fix help text for `--listen-addr`: duplicate word ("to to") and incorrect default port (said 80, actual default is 3000)

## Details

The `unwrap()` calls in `control_db.rs` could cause panics if serialization/deserialization ever failed. Since all the containing functions already return `Result`, replacing with `?` is straightforward and lets callers handle errors gracefully.

The `get_leader_replica_by_database` method was the only replica-related method on the trait that didn't return `Result`, despite calling `get_replicas()` which can fail. This inconsistency was masked by the `unwrap()`.

## Files changed
| File | Change |
|------|--------|
| `crates/standalone/src/control_db.rs` | Replace 5 `unwrap()` → `?`, change return type |
| `crates/client-api/src/lib.rs` | Update trait definition + blanket impl |
| `crates/standalone/src/lib.rs` | Update trait impl + call site |
| `crates/client-api/src/routes/database.rs` | Handle new `Result` at call site |
| `crates/testing/src/modules.rs` | Handle new `Result` in test helper |
| `crates/standalone/src/subcommands/start.rs` | Fix help text typo |

## Test plan
- [ ] CI compilation and tests should pass — no logic changes, only error propagation
- [ ] The `get_leader_replica_by_database` callers all handle the new `Result` appropriately

### DIFF
--- a/docs/docs/00300-resources/00200-reference/00200-http-api/00100-authorization.md
+++ b/docs/docs/00300-resources/00200-reference/00200-http-api/00100-authorization.md
@@ -16,6 +16,8 @@ Alternately, a new identity and token will be generated during an anonymous conn
 
 Many SpacetimeDB HTTP endpoints either require or optionally accept a token in the `Authorization` header. SpacetimeDB authorization headers are of the form `Authorization: Bearer ${token}`, where `token` is an [OpenID Connect](https://openid.net/developers/how-connect-works/) compliant [JSON Web Token](https://jwt.io/), such as the one returned from [the `POST /v1/identity` HTTP endpoint](./00200-identity.md#post-v1identity).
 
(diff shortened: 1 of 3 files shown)


bfops:
Thank you for opening a PR! To clarify - have any of these `unwrap`s been causing issues for you? Have you observed panics in these places?

Shiven0504:
Hii @bfops, I haven't observed actual panics here. The change is preventative since these are SQLite I/O operations where errors like disk full or corruption are plausible, so I guess propagating them with ? seems safer than panicking.

bfops:
I'm going to close this PR since it's unclear whether there is an actual issue here. Definitely feel free to reopen if you produce a repro for an actual server-takedown here!

PR #4576, opened 2026-03-06
Support persistent default listen-addr for spacetime start

0monish:
# Description of Changes

This change allows `spacetime start` to use a project-configured default listen address from `spacetime.json`.

Behavior precedence is now:

1. Explicit CLI flag, such as `spacetime start --listen-addr 127.0.0.1:5000`
2. `listen-addr` from `spacetime.json`
3. Built-in default `0.0.0.0:3000`

Example:

```json
{
  "listen-addr": "0.0.0.0:4000"
}
```

With that config in place, `spacetime start` will start on `0.0.0.0:4000` unless `--listen-addr` is passed explicitly.

Implementation details:

- `spacetime start` now checks project config before forwarding to the standalone binary.
- Config lookup uses the existing `spacetime.json` discovery behavior, including parent-directory search and `spacetime.local.json` layering.
- The forwarded argument path preserves CLI precedence over config.
- Added unit tests covering detection of explicit `--listen-addr` and `-l` forms so config is only injected when the user did not already provide a listen address.
- Updated standalone config documentation to describe the new `listen-addr` behavior and precedence.

This is backward compatible:

- users without config still get `0.0.0.0:3000`
- existing scripts that pass `--listen-addr` continue to behave the same
- the feature is project-local and works with the existing config-driven workflow

None.

This change does not modify any public API or ABI. It only adds an additional configuration source for an existing CLI option while preserving current defaults and CLI override behavior.

Complexity: 2/5

This is a small, localized CLI behavior change.

The main interaction point is config precedence between forwarded CLI arguments and project configuration. The implementation is low risk because it:

- preserves the existing standalone default when config is absent
- preserves explicit CLI `--listen-addr` and `-l` precedence
- reuses the existing `spacetime.json` discovery and layering behavior rather than introducing a new config system

The main thing reviewers may want to sanity-check is that the new config-backed behavior only applies when the user did not explicitly pass a listen address.

# Testing

Completed:

- [x] `cargo fmt`
- [x] `cargo check -p spacetimedb-cli`
- [x] `cargo test -p spacetimedb-cli subcommands::start::tests`
- [x] `cargo clippy -p spacetimedb-cli --all-targets`

Reviewer checks:

- [ ] Run `spacetime start` without `spacetime.json` and confirm it still binds to `0.0.0.0:3000`
- [ ] Add `"listen-addr": "0.0.0.0:4000"` to `spacetime.json`, run `spacetime start`, and confirm it binds to `0.0.0.0:4000`
- [ ] Run `spacetime start --listen-addr 127.0.0.1:5000` with config present and confirm the CLI flag still wins

### DIFF
--- a/docs/docs/00300-resources/00200-reference/00100-cli-reference/00200-standalone-config.md
+++ b/docs/docs/00300-resources/00200-reference/00100-cli-reference/00200-standalone-config.md
@@ -12,6 +12,28 @@ database running in data directory <b>/home/user/.local/share/spacetime/data</b>
 
 On Linux and macOS, this directory is by default `~/.local/share/spacetime/data`. On Windows, it's `%LOCALAPPDATA%\SpacetimeDB\data`.
 
+## Project-level configuration (`spacetime.json`)
+
+You can set a persistent default listen address for `spacetime start` in your project's `spacetime.json` file. This is useful when port `3000` is already occupied (e.g. by a frontend dev server) or when your team standardizes on a different local port.
+
+### `listen-addr`
+
+```json
+{
+  "listen-addr": "0.0.0.0:4000"
+}
+```
+
+When present, `spacetime start` will use this address instead of the built-in default `0.0.0.0:3000`.
+
+**Precedence (highest to lowest):**
+
+1. Explicit CLI flag: `spacetime start --listen-addr 127.0.0.1:5000`
+2. Value from `spacetime.json`: `"listen-addr": "0.0.0.0:4000"`
+3. Built-in default: `0.0.0.0:3000`
+
+The config is discovered by searching the current directory and its parent directories for `spacetime.json`. If present, `spacetime.local.json` is layered on top of it before `listen-addr` is read.
+
 ## `config.toml`
 
 - [`certificate-authority`](#certificate-authority)
--- a/crates/cli/src/subcommands/start.rs
+++ b/crates/cli/src/subcommands/start.rs
@@ -2,10 +2,11 @@ use std::ffi::OsString;
 use std::io;
 use std::process::{Command, ExitCode};
 
-use anyhow::Context;
+use anyhow::{anyhow, Context};
 use clap::{Arg, ArgMatches};
 use spacetimedb_paths::SpacetimePaths;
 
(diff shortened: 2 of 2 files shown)


0monish:
When I can get this feature merged this helps a lot as I am using the spacetimedb and Learning along with that.
If it's there, anything that I missed or anything wrong? Please let me know.

docs/docs/00300-resources/00200-reference/00100-cli-reference/00200-standalone-config.md
```diff
@@ -12,6 +12,28 @@ database running in data directory <b>/home/user/.local/share/spacetime/data</b>
 
 On Linux and macOS, this directory is by default `~/.local/share/spacetime/data`. On Windows, it's `%LOCALAPPDATA%\SpacetimeDB\data`.
 
+## Project-level configuration (`spacetime.json`)
```
cloutiertyler, line 15: This does not belong here, it should be in this document:

https://github.com/clockworklabs/SpacetimeDB/blob/master/docs/docs/00300-resources/00200-reference/00100-cli-reference/00300-spacetime-json.md

cloutiertyler: Actually come to think of it, this configuration doesn't belong in `spacetime.json` at all. That is meant to be checked in to your project repo. This configuration should be in the `cli.toml` file.

https://github.com/clockworklabs/SpacetimeDB/blob/7143ed9722d18112e2546a47dc3b97516f2d1128/crates/paths/src/lib.rs#L63

cloutiertyler:
@0monish `spacetime.json` is not where this configuration belongs. Instead it should be stored in the `cli.toml` file. The `cli.toml` configuration is for configuring options for the `spacetime` command and other configuration required to run SpacetimeDB Standalone. `spacetime.json` is *project* specific configuration instead. So configuration related to your own project, not the CLI or Standalone. I'm going to close this PR, but feel free to open another one with the configuration in the appropriate file.

https://github.com/clockworklabs/SpacetimeDB/blob/7143ed9722d18112e2546a47dc3b97516f2d1128/crates/paths/src/lib.rs#L63

PR #4578, opened 2026-03-06
CI - Simplify PR approval check

bfops:
# Description of Changes

The previous version ended up incurring two different entries on the PR (one for the `pull_request` event and one for the `pull_request_review` event). Both versions were marked "required", so PRs could end up in an unmergeable state if one check had succeeded but the other had failed (e.g. if you submitted a PR approval, the previous `pull_request` version of the check would still be failed since it didn't refresh).

See the entries at top and bottom here:
image

This PR fixes it by only allowing the `pull_request_review` events. I _think_ this covers all the cases, but I'm not sure.

None.

1

# Testing

I don't know how to test it really :shrug:

### DIFF
--- a/.github/workflows/pr_approval_check.yml
+++ b/.github/workflows/pr_approval_check.yml
@@ -1,4 +1,4 @@
-name: PR Approval Check
+name: Review Checks
 
 on:
   pull_request:
@@ -27,7 +27,7 @@ jobs:
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           script: |
-            const contextName = "bot-pr-approval-check";
+            const contextName = "PR approval check";
 
             let targetSha;
             let state;


.github/workflows/pr_approval_check.yml
```diff
@@ -1,19 +1,15 @@
...
+    types: [submitted, dismissed]
 
 permissions:
   contents: read
 
 jobs:
   check-approvals:
     runs-on: ubuntu-latest
-    # Skip this check if PR is in draft state
-    if: github.event.pull_request.draft == false
```
bfops, line 16: this would end up stale since we no longer run on the `ready_for_review` event. It's simpler to just remove it and have it be possible for this check to pass on a draft PR (if it has two approvals). Merging would still be blocked anyway until it's not a draft anymore, so this check doesn't really need to check that.

.github/workflows/pr_approval_check.yml
```diff
@@ -1,19 +1,16 @@
 name: PR Approval Check
 
 on:
-  pull_request:
-    types: [opened, synchronize, reopened, ready_for_review]
   pull_request_review:
-    types: [submitted]
+    types: [submitted, dismissed]
+  merge_group:
```
bfops, line 8: this doesn't quite work properly

PR #4588, opened 2026-03-08
Fix Rust Chat App Tutorial not showing messages of other users live

OMGeeky:
# Description of Changes

This fixes an Issue in the Rust [Chat App Tutorial](https://spacetimedb.com/docs/tutorials/chat-app#creating-the-client), that is caused by the [Event Type Changes](https://spacetimedb.com/docs/upgrade/?client-language=rust&server-language=rust#event-type-changes) in 2.0

This resulted in other clients not receiving new messages, since they are now Event::Transaction and no longer included within the Event::Reducer

I don't think any? I only changed a tutorial, the real change was in 2.0

1

# Testing

- [x] I ran the Tutorial app with these changes and messages started to appear as expected on the second client when writing something on the first.

There might be more places where this is an Issue, I just noticed this one while following the Tutorial.

### DIFF
--- a/.github/workflows/pr_approval_check.yml
+++ b/.github/workflows/pr_approval_check.yml
@@ -1,7 +1,17 @@
 name: Review Checks
 
+# SECURITY: This workflow uses pull_request_target so that it has write access to
+# set commit statuses on external (fork) PRs. pull_request_target runs in the
+# context of the base branch, which grants the GITHUB_TOKEN write permissions
+# that a regular pull_request event on a fork would not have.
+#
+# IMPORTANT: This workflow must NEVER check out, build, or execute code from the
+# PR branch. Doing so would allow a malicious fork to run arbitrary code with
+# write access to the repository. This workflow only reads PR metadata via the
+# GitHub API, which is safe.
+
 on:
-  pull_request:
+  pull_request_target:
     types: [opened, synchronize, reopened]
   pull_request_review:
     types: [submitted, dismissed]
@@ -21,6 +31,7 @@ jobs:
     name: Set approval status
     runs-on: ubuntu-latest
 
+    # SECURITY: Do not add a checkout step to this job. See comment at the top of this file.
     steps:
       - name: Evaluate and publish approval status
         uses: actions/github-script@v7


OMGeeky:
> D'oh! I fixed this in the template (`templates/chat-console-rs/src/main.rs`), but forgot to update the tutorial.
> 
> In the template I used `!matches!(ctx.event, Event::SubscribeApplied)`. Do you think that's more or less clear than your `matches!(ctx.event, Event::Reducer(_) | Event::Transaction)`? Personally I think excluding `SubscribeApplied` makes more sense than including the other two, but the `!matches!` looks pretty gnarly in Rust syntax.

I'm relatively new to to spacetime-db but in my opinion I think it should be the explicit inclusion of `Reducer` and `Transaction` instead of exclusion of `SubscribeApplied`. The reason for that is that there are other variants like `UnsubscribeApplied`, `Disconnected` or `SubscribeError` and even if all of those are checked, the enum is maked as `non_exhaustive` so new variants might be added at any moment. 
Also considering that we only want something to be printed here, if it is a new message and not other things, this callback should only care about printing new messages, no other info like `SubscribeError` etc. so we should filter for those explicitly in my opinion.

And yeah, `!matches!` looks pretty bad in Rust. I'd rather not invert and have an early return if we were to go that route instead

gefjon:
> I'm relatively new to to spacetime-db but in my opinion I think it should be the explicit inclusion of `Reducer` and `Transaction` instead of exclusion of `SubscribeApplied`. The reason for that is that there are other variants like `UnsubscribeApplied`, `Disconnected` or `SubscribeError` and even if all of those are checked, the enum is maked as `non_exhaustive` so new variants might be added at any moment. Also considering that we only want something to be printed here, if it is a new message and not other things, this callback should only care about printing new messages, no other info like `SubscribeError` etc. so we should filter for those explicitly in my opinion.
> 
> And yeah, `!matches!` looks pretty bad in Rust. I'd rather not invert and have an early return if we were to go that route instead

This all works for me. Could you update the template code in `templates/chat-console-rs/src/main.rs` to do the same thing as you've done in the tutorial?

OMGeeky:
Thanks for the quick approval @gefjon. Anything I need to do about the one failing CI-Check?

gefjon:
> Thanks for the quick approval @gefjon. Anything I need to do about the one failing CI-Check?

No, that's a bug in our CI. We're working on it, and will get this merged as soon as it's fixed.

PR #4593, opened 2026-03-09
feat: Allow dropping empty tables during auto-migration

Ludv1gL:
## Summary

- Removing a table from the schema previously always triggered `AutoMigrateError::RemoveTable`, forcing `--delete-data` which **nukes the entire database**
- Now, empty tables can be dropped seamlessly during `spacetime publish`
- Non-empty tables fail with an actionable error guiding the user to clear rows first

## Changes

- **`auto_migrate.rs`**: Replace hard `RemoveTable` error with `AutoMigrateStep::RemoveTable`. Compute `removed_tables` set (same pattern as `new_tables`) and pass to `auto_migrate_indexes`/`auto_migrate_sequences`/`auto_migrate_constraints` to skip sub-object diffs for removed tables (cascade handled by `drop_table`)
- **`update.rs`**: Execute `RemoveTable` step — O(1) emptiness check via `table_row_count_mut`, then `drop_table`. Fails with clear message if table has data
- **`formatter.rs`** / **`termcolor_formatter.rs`**: Add `format_remove_table` to `MigrationFormatter` trait + implementation
- **`publish.rs` (smoketests)**: Update existing test — removing empty table now succeeds without flags

## Safety

- **Transaction safety**: Emptiness check and drop run in the same `MutTx` — no window for concurrent inserts
- **Cascade**: `drop_table` already handles removing all indexes, constraints, and sequences for the table
- **Sub-object filtering**: Indexes/constraints/sequences belonging to removed tables are filtered from their respective diffs, preventing orphan `RemoveIndex`/`RemoveConstraint`/`RemoveSequence` steps
- **Rollback**: If the emptiness check fails, the entire migration aborts before any changes are applied

## Example error output (non-empty table)

```
Cannot remove table `MyTable`: table contains data. Clear the table's rows (e.g. via a reducer) before removing it from your schema.
```

## Test plan

- [x] `cargo check -p spacetimedb-schema -p spacetimedb-core` passes
- [x] All 14 `auto_migrate` schema tests pass (2 new: `remove_table_produces_step`, `remove_table_does_not_produce_orphan_sub_object_steps`)
- [x] All 3 `update` execution tests pass (2 new: `remove_empty_table_succeeds`, `remove_nonempty_table_fails`)
- [x] Updated existing `auto_migration_errors` test (no longer expects `RemoveTable` error)
- [x] Updated smoke test: `cli_can_publish_remove_empty_table` expects success

### DIFF
--- a/crates/schema/src/auto_migrate.rs
+++ b/crates/schema/src/auto_migrate.rs
@@ -467,8 +467,12 @@ pub fn ponder_auto_migrate<'def>(old: &'def ModuleDef, new: &'def ModuleDef) -> 
             (HashSet::new(), HashSet::new()),
             |(mut added, mut removed), diff| {
                 match diff {
-                    Diff::Add { new } => { added.insert(&new.name); }
-                    Diff::Remove { old } => { removed.insert(&old.name); }
+                    Diff::Add { new } => {
+                        added.insert(&new.name);
+                    }
+                    Diff::Remove { old } => {
+                        removed.insert(&old.name);
+                    }
                     Diff::MaybeChange { .. } => {}
                 }
                 (added, removed)
@@ -933,7 +937,11 @@ fn ensure_old_ty_upgradable_to_new(
     }
 }
 
-fn auto_migrate_indexes(plan: &mut AutoMigratePlan<'_>, new_tables: &HashSet<&Identifier>, removed_tables: &HashSet<&Identifier>) -> Result<()> {
+fn auto_migrate_indexes(
+    plan: &mut AutoMigratePlan<'_>,
+    new_tables: &HashSet<&Identifier>,
+    removed_tables: &HashSet<&Identifier>,
+) -> Result<()> {
     diff(plan.old, plan.new, ModuleDef::indexes)
         .map(|index_diff| -> Result<()> {
             match index_diff {
@@ -970,7 +978,11 @@ fn auto_migrate_indexes(plan: &mut AutoMigratePlan<'_>, new_tables: &HashSet<&Id
         .collect_all_errors()
 }
 
-fn auto_migrate_sequences(plan: &mut AutoMigratePlan, new_tables: &HashSet<&Identifier>, removed_tables: &HashSet<&Identifier>) -> Result<()> {
+fn auto_migrate_sequences(
+    plan: &mut AutoMigratePlan,
+    new_tables: &HashSet<&Identifier>,
+    removed_tables: &HashSet<&Identifier>,
+) -> Result<()> {
     diff(plan.old, plan.new, ModuleDef::sequences)
         .map(|sequence_diff| -> Result<()> {
             match sequence_diff {
@@ -1003,7 +1015,11 @@ fn auto_migrate_sequences(plan: &mut AutoMigratePlan, new_tables: &HashSet<&Iden
         .collect_all_errors()
 }
 
-fn auto_migrate_constraints(plan: &mut AutoMigratePlan, new_tables: &HashSet<&Identifier>, removed_tables: &HashSet<&Identifier>) -> Result<()> {
+fn auto_migrate_constraints(
+    plan: &mut AutoMigratePlan,
+    new_tables: &HashSet<&Identifier>,
+    removed_tables: &HashSet<&Identifier>,
+) -> Result<()> {
     diff(plan.old, plan.new, ModuleDef::constraints)
         .map(|constraint_diff| -> Result<()> {
             match constraint_diff {


crates/core/src/db/update.rs
```diff
@@ -405,4 +418,74 @@ mod test {
...
+
+        let mut tx = begin_mut_tx(&stdb);
+        for def in old.tables() {
+            create_table_from_def(&stdb, &mut tx, &old, def)?;
+        }
+        stdb.commit_tx(tx)?;
+
+        let mut tx = begin_mut_tx(&stdb);
+        let plan = ponder_migrate(&old, &new)?;
+        let _ = update_database(&stdb, &mut tx, auth_ctx, plan, &TestLogger)?;
```
Centril: Check for `UpdateResult::Success` here.

crates/core/src/db/update.rs
```diff
@@ -405,4 +418,74 @@ mod test {
...
+
+        assert!(stdb.table_id_from_name_mut(&tx, "droppable")?.is_none());
```
Centril, line 463: Check `pending_schema_changes` here.

crates/core/src/db/update.rs
```diff
@@ -405,4 +418,74 @@ mod test {
...
+        stdb.commit_tx(tx)?;
+
+        let mut tx = begin_mut_tx(&stdb);
+        let plan = ponder_migrate(&old, &new)?;
+        let result = update_database(&stdb, &mut tx, auth_ctx, plan, &TestLogger);
+        let err = result.err().expect("removing a non-empty table should fail");
+        assert!(
+            err.to_string().contains("table contains data"),
+            "error should mention that the table contains data, got: {err}"
+        );
```
Centril, line 500: check that `pending_schema_changes` is empty here

crates/schema/src/auto_migrate.rs
```diff
@@ -452,17 +453,22 @@ pub fn ponder_auto_migrate<'def>(old: &'def ModuleDef, new: &'def ModuleDef) ->
...
-    let new_tables: HashSet<&Identifier> = diff(plan.old, plan.new, ModuleDef::tables)
-        .filter_map(|diff| match diff {
-            Diff::Add { new } => Some(&new.name),
-            _ => None,
-        })
-        .collect();
-    let indexes_ok = auto_migrate_indexes(&mut plan, &new_tables);
-    let sequences_ok = auto_migrate_sequences(&mut plan, &new_tables);
-    let constraints_ok = auto_migrate_constraints(&mut plan, &new_tables);
+    // Filter out sub-objects of added/removed tables — they're handled by AddTable/RemoveTable.
```
```suggestion
    // Filter out sub-objects of added/removed tables — they're handled by `AddTable`/`RemoveTable`.
```

crates/schema/src/auto_migrate.rs
```diff
@@ -938,6 +943,9 @@ fn auto_migrate_indexes(plan: &mut AutoMigratePlan<'_>, new_tables: &HashSet<&Id
                     Ok(())
                 }
                 Diff::Remove { old } => {
+                    if removed_tables.contains(&plan.old.stored_in_table_def(&old.name).unwrap().name) {
+                        return Ok(()); // handled by RemoveTable
+                    }
                     plan.steps.push(AutoMigrateStep::RemoveIndex(old.key()));
```
Centril: Please make this align with the code for `Diff::Add` and add a closure that takes the hash set as a parameter.

crates/schema/src/auto_migrate.rs
```diff
@@ -630,11 +636,10 @@ fn auto_migrate_tables(plan: &mut AutoMigratePlan<'_>) -> Result<()> {
                     plan.steps.push(AutoMigrateStep::AddTable(new.key()));
                     Ok(())
                 }
-                // TODO: When we remove tables, we should also remove their dependencies, including row-level security.
```
Centril, line 633: Did you consider this?

crates/schema/src/auto_migrate.rs
```diff
@@ -630,11 +636,10 @@ fn auto_migrate_tables(plan: &mut AutoMigratePlan<'_>) -> Result<()> {
...
-                Diff::Remove { old } => Err(AutoMigrateError::RemoveTable {
-                    table: old.name.clone(),
+                Diff::Remove { old } => {
+                    plan.steps.push(AutoMigrateStep::RemoveTable(old.key()));
```
Centril, line 646: We can break clients when removing an empty table, so we should add `AutoMigrateStep::DisconnectAllUsers` if not present.
There's some code that checks if it is present and adds it if not.
I'd like to see this code first refactored into one method: `plan.ensure_disconnect_all_users(); in:
- `auto_migrate_views`
- `auto_migrate_view`
- `auto_migrate_table`
- `auto_migrate_row_level_security`

and then it can be used here too.

crates/schema/src/auto_migrate.rs
```diff
@@ -975,6 +983,9 @@ fn auto_migrate_sequences(plan: &mut AutoMigratePlan, new_tables: &HashSet<&Iden
                     Ok(())
                 }
                 Diff::Remove { old } => {
+                    if removed_tables.contains(&plan.old.stored_in_table_def(&old.name).unwrap().name) {
+                        return Ok(()); // handled by RemoveTable
+                    }
```
Centril: Let's also use a closure here and style without the early return as in the `Add` case.

crates/schema/src/auto_migrate.rs
```diff
@@ -1010,6 +1021,9 @@ fn auto_migrate_constraints(plan: &mut AutoMigratePlan, new_tables: &HashSet<&Id
                     }
                 }
                 Diff::Remove { old } => {
+                    if removed_tables.contains(&plan.old.stored_in_table_def(&old.name).unwrap().name) {
+                        return Ok(()); // handled by RemoveTable
+                    }
                     plan.steps.push(AutoMigrateStep::RemoveConstraint(old.key()));
```
Centril: ditto re closure and early return

crates/schema/src/auto_migrate.rs
```diff
@@ -2399,4 +2411,53 @@ mod tests {
...
+                .build_table_with_new_type("Keep", ProductType::from([("id", AlgebraicType::U64)]), true)
+                .with_access(TableAccess::Public)
+                .finish();
+        });
+
+        let plan = ponder_auto_migrate(&old, &new).expect("removing a table should produce a valid plan");
+        assert!(
+            plan.steps.iter().any(|s| matches!(s, AutoMigrateStep::RemoveTable(name) if &name[..] == "Drop")),
+            "plan should contain a RemoveTable step for 'Drop'"
+        );
```
Centril: I'd prefer to see an explicit list here for this test to also capture the disconnect all clients step.

crates/schema/src/auto_migrate.rs
```diff
@@ -2399,4 +2411,53 @@ mod tests {
...
+
+        let plan = ponder_auto_migrate(&old, &new).expect("removing a table should produce a valid plan");
+        assert!(
+            !plan.steps.iter().any(|s| matches!(s, AutoMigrateStep::RemoveIndex(_))),
+            "plan should not contain orphan RemoveIndex steps for a removed table"
+        );
+        assert!(
+            !plan.steps.iter().any(|s| matches!(s, AutoMigrateStep::RemoveConstraint(_))),
+            "plan should not contain orphan RemoveConstraint steps for a removed table"
+        );
```
Centril: I'd like to see an explicit list here too.

crates/smoketests/tests/smoketests/cli/publish.rs
```diff
@@ -166,11 +166,11 @@ fn cli_can_publish_automigration_change_with_delete_data_always_and_yes_break_cl
...
-fn cli_cannot_publish_breaking_change_without_flag() {
+fn cli_can_publish_remove_empty_table() {
     migration_test(
-        "breaking-change-test",
+        "remove-empty-table-test",
         &["--build-options=--features test-remove-table"],
-        false,
+        true,
     );
 }
```
Centril, line 175: Why was this test changed?

crates/core/src/db/update.rs
Ludv1gL: Done. Now asserting `RequiresClientDisconnect` (since `RemoveTable` now emits `DisconnectAllUsers`).

crates/core/src/db/update.rs
Ludv1gL: Done — asserting exactly 1 pending schema change (`TableRemoved`) from `drop_table`.

crates/core/src/db/update.rs
Ludv1gL: Done — asserting `pending_schema_changes` has len 1 (the `TableRemoved` entry from `drop_table`).

crates/schema/src/auto_migrate.rs
Ludv1gL: Done — backticked both `AddTable` and `RemoveTable`.

crates/schema/src/auto_migrate.rs
Ludv1gL: Done — switched to negated-if style (no early return) to align with the `Diff::Add` pattern. Applied to all three: indexes, sequences, and constraints.

crates/schema/src/auto_migrate.rs
Ludv1gL: Yes — views and RLS referencing a dropped table are caught by schema validation before `ponder_auto_migrate` runs. The new `ModuleDef` won't validate if it references a nonexistent table, so those dependencies are naturally resolved. The TODO about removing dependencies would apply to future work where we want to cascade-remove dependent objects, but the validation layer handles it for now.

crates/schema/src/auto_migrate.rs
Ludv1gL: Great call. Done — added `plan.ensure_disconnect_all_users()` after pushing `RemoveTable`. Also refactored all existing inline `if \!plan.disconnects_all_users() { push(DisconnectAllUsers) }` patterns across the file to use the new `ensure_disconnect_all_users()` method (views, columns_added, RLS).

crates/schema/src/auto_migrate.rs
Ludv1gL: Done.

crates/schema/src/auto_migrate.rs
Ludv1gL: Done.

crates/schema/src/auto_migrate.rs
Ludv1gL: Done — now using `assert_eq\!(plan.steps, vec\![RemoveTable(&drop_table), DisconnectAllUsers])` for explicit step verification.

crates/schema/src/auto_migrate.rs
Ludv1gL: Done — same explicit `assert_eq\!` on the full steps vec.

crates/smoketests/tests/smoketests/cli/publish.rs
Ludv1gL: This test was originally `cli_cannot_publish_breaking_change_without_flag` — it tested that removing a table fails during publish. Now that empty table removal succeeds via auto-migration, the test was updated to `cli_can_publish_remove_empty_table` and expects success. The smoke test module starts with empty tables (fresh publish), so the removal succeeds as expected.

crates/core/src/db/update.rs
```diff
@@ -405,4 +418,85 @@ mod test {
...
+        let mut tx = begin_mut_tx(&stdb);
+        let plan = ponder_migrate(&old, &new)?;
+        let res = update_database(&stdb, &mut tx, auth_ctx, plan, &TestLogger)?;
+
+        assert!(
+            matches!(res, UpdateResult::RequiresClientDisconnect),
+            "removing a table should disconnect clients"
+        );
+        assert!(stdb.table_id_from_name_mut(&tx, "droppable")?.is_none());
+        // drop_table records a TableRemoved schema change for the subscription system.
```
Centril: The Rust convention is that code terms (keywords, functions, binding names, ...) are referenced in commentary using single back-quotes, e.g., `drop_table`.

crates/core/src/db/update.rs
```diff
@@ -405,4 +418,85 @@ mod test {
...
+        assert_eq!(
+            tx.pending_schema_changes().len(),
```
Centril: I'd like this to expect `TableRemoved`, not just that there was one change.

crates/core/src/db/update.rs
Centril: I don't see any new assertion, and the list should be the empty slice, not of length 1.

crates/schema/src/auto_migrate.rs
```diff
@@ -452,17 +461,22 @@ pub fn ponder_auto_migrate<'def>(old: &'def ModuleDef, new: &'def ModuleDef) ->
...
-        .collect();
-    let indexes_ok = auto_migrate_indexes(&mut plan, &new_tables);
-    let sequences_ok = auto_migrate_sequences(&mut plan, &new_tables);
-    let constraints_ok = auto_migrate_constraints(&mut plan, &new_tables);
+    // Filter out sub-objects of added/removed tables — they're handled by `AddTable`/`RemoveTable`.
+    let (new_tables, removed_tables): (HashSet<&Identifier>, HashSet<&Identifier>) =
+        diff(plan.old, plan.new, ModuleDef::tables).fold(
+            (HashSet::new(), HashSet::new()),
+            |(mut added, mut removed), d| {
+                match d {
```
```suggestion
            |(mut added, mut removed), diff| {
                match diff {
```

crates/schema/src/auto_migrate.rs
```diff
@@ -2399,4 +2409,58 @@ mod tests {
...
+        });
+
+        let drop_table = expect_identifier("Drop");
+        let plan = ponder_auto_migrate(&old, &new).expect("removing a table should produce a valid plan");
+        assert_eq!(
+            plan.steps,
+            vec![
+                AutoMigrateStep::RemoveTable(&drop_table),
+                AutoMigrateStep::DisconnectAllUsers,
+            ],
```
```suggestion
            &[
                AutoMigrateStep::RemoveTable(&drop_table),
                AutoMigrateStep::DisconnectAllUsers,
            ],
```
These are tests, but no need to make a heap allocation for an explicit list on the stack.

crates/schema/src/auto_migrate.rs
```diff
@@ -2399,4 +2409,58 @@ mod tests {
...
+                .with_access(TableAccess::Public)
+                .finish();
+        });
+        let new = create_module_def(|_builder| {});
+
+        let drop_table = expect_identifier("Drop");
+        let plan = ponder_auto_migrate(&old, &new).expect("removing a table should produce a valid plan");
+        assert_eq!(
+            plan.steps,
+            vec![
```
Centril: same here

crates/smoketests/tests/smoketests/cli/publish.rs
Centril: The point of the test is not to test that removing a table fails during publish, but that flag is required to acknowledge breaking changes, which removing an empty table still should be, so this test should not need to change, or there is a bug in the PR.

Ludv1gL:
Rebased onto latest master and addressed all round 2 feedback:

1. **Backtick code terms** — fixed in test comments
2. **Assert `TableRemoved` variant** — now checks `matches!(c, PendingSchemaChange::TableRemoved(..))`
3. **Empty pending_schema_changes on failure** — `remove_nonempty_table_fails` now asserts `is_empty()`
4. **Rename `d` → `diff`** — done in fold closure
5. **`&[...]` instead of `vec![...]`** — done in both test assertions
6. **Smoketest reverted** — the original test is unchanged from upstream. Removing an empty table still requires `--break-clients` acknowledgment since it's a client-breaking change.

Ludv1gL:
Rebased onto latest master (`cdd8ba77a`) which includes #4311 (BytesKey optimization).

Minor conflict resolutions during rebase:
- `auto_migrate.rs`: Resolved duplicate `RemoveTable` variant and variable rename (`d` → `diff`)
- `termcolor_formatter.rs`: Removed duplicate `Identifier` import
- `update.rs`: Took the improved assertions per round 2 review feedback

All 14 auto_migrate tests pass (`cargo test -p spacetimedb-schema --lib -- auto_migrate`).

Centril:
This needs a `cargo fmt` it seems.

PR #4595, opened 2026-03-09
CI - `rustfmt` instead of `cargo fmt`

bfops:
# Description of Changes

`cargo ci lint` was running `cargo fmt`, but that didn't pick up all files. As a result, our pre-commit hook (which just runs `rustfmt` on any changed `.rs` files) would change otherwise-untouched files in merge commits.

This PR addresses that discrepancy by having `cargo ci lint` run `rustfmt` on all tracked `.rs` files.

The entire diff is just `rustfmt` changes except for the changes in `tools/ci/src/main.rs`.

2

# Testing

- [x] `cargo ci lint` fails
- [x] `cargo ci lint` passes after running `rustfmt` on everything
- [x] `cargo fmt --all` doesn't cause any diff after doing the above

### DIFF
--- a/crates/bindings/tests/ui/views.stderr
+++ b/crates/bindings/tests/ui/views.stderr
@@ -61,20 +61,20 @@ error[E0425]: cannot find type `ScheduledTable` in this scope
     |                                             ^^^^^^^^^^^^^^ not found in this scope
 
 error[E0425]: cannot find type `T` in this scope
-   --> tests/ui/views.rs:202:60
+   --> tests/ui/views.rs:201:60
     |
-202 | fn view_nonexistent_table(ctx: &ViewContext) -> impl Query<T> {
+201 | fn view_nonexistent_table(ctx: &ViewContext) -> impl Query<T> {
     |                                                            ^ not found in this scope
     |
 help: you might be missing a type parameter
     |
-202 | fn view_nonexistent_table<T>(ctx: &ViewContext) -> impl Query<T> {
+201 | fn view_nonexistent_table<T>(ctx: &ViewContext) -> impl Query<T> {
     |                          +++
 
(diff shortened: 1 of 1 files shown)


tools/ci/src/main.rs
```diff
@@ -361,7 +371,19 @@ fn main() -> Result<()> {
         }
 
         Some(CiCmd::Lint) => {
-            cmd!("cargo", "fmt", "--all", "--", "--check").run()?;
+            ensure_repo_root()?;
+            // `cargo fmt --all` only checks files that Cargo discovers through workspace/package targets.
+            // However, we also keep Rust sources in a locations that are tracked but not part of our workspace,
+            // so this approach properly catches all the files, where `cargo fmt` does not.
+            let mut files = Vec::new();
+            files.extend(tracked_rs_files_under(".")?);
```
jdetter, line 379: `rustfmt` doesn't have to run against the `Cargo.toml` files correct?

bfops: it's at least not how we have it set up in the pre-commit hook, but idk if `cargo fmt` does run against those. That's a good question.

bfops:
> I'm assuming the only real change here was in `tools/ci/src/main.rs` and everything else was a formatting change, if this isn't the case let me know and I can re-review. I also left a question.

That's correct.

The entire diff is just rustfmt changes (*and corresponding snapshot updates) except for the changes in tools/ci/src/main.rs, which are actual code changes.

tools/ci/src/main.rs
bfops: Okay, chatgpt says no  cargo fmt doesn't reformat cargo toml files

jdetter: ok excellent

PR #4601, opened 2026-03-10
Update C# Runtime dependencies to support NativeAOT-LLVM prerequisites

rekhoff:
# Description of Changes
* Add dotnet-experimental feed + package source mapping for LLVM packages in `sdks/csharp/tools~/write-nuget-config.sh`, so generated NuGet.Config files include NativeAOT-LLVM prerequisites.
* Make LLVM toolchain packages explicit dependencies in `SpacetimeDB.Runtime` to ensure restores succeed even when LLVM dependencies are only referenced through the `.nupkg`.
* Import the LLVM targets from the package when `EXPERIMENTAL_WASM_AOT=1` to enable NativeAOT build steps without relying on downstream package reference resolution.
# Context
Changes are required to get `NativeAOT-LLVM` in #4515 to build correct, but moving the packages closer to the build, to ensure they get into the Nuget restore successfully.
Additional changes where needed to `write-nuget-config.sh‎` to allow `Nuget.Config` files generated with required changes during regression testing.
None.
2 (Low–moderate). Changes are scoped to build/restore infrastructure and package configuration.
# Testing
- [X] Built CLI locally
- [X] Ran `run-regression-tests.sh` without errors


.github/workflows/ci.yml
```diff
@@ -869,11 +869,35 @@ jobs:
...
+
+      - name: Setup dotnet
+        uses: actions/setup-dotnet@v3
+        with:
+          global-json-file: global.json
+
+      - name: Setup Emscripten
+        uses: mymindstorm/setup-emsdk@v14
+        with:
+          version: 4.0.23
```
JasonAtClockwork: I avoided using this action on other CI work as it looks like it is no longer maintained but it is cleaner than the manual process used in the Windows smoke-tests.

JasonAtClockwork (approved):
Just one small thing in the CI otherwise this looks good, from our findings we'll still need more work on Linux and .NET 9+.

.github/workflows/ci.yml
rekhoff: Thanks, CI workflow has been updated to avoid use of `mymindstorm/setup-emsdk`

Issue #4603, opened 2026-03-10
Spacetimedb fails to update sender dependent view


gmorenz:
There's some strange concurrency related issue where my view stops receiving updates when it should. I've got a reproduction [here](https://github.com/gmorenz/spacetimedb-issue).

In short: I have a table and a view of that table that filters by identity. I have two connections to the database, both subscribed to the view. When I drop the one connection, the other connection stop updating its version of the view. This reliably happens in the reproduction - though I'm not sure I've captured all relevant portions of the reproduction in my short description of it.

Specifically with that reproduction

I expect to see, and I do see if comment out `drop(_player2)` on line 44

```
Created new database with name: testdb-918057, identity: c200ce9ae79443490f432c41870961263ba21ad9b4430888280be63cdf14341e
Decrementing from GameState { game_id: 0, player_1: Identity(c2006b4c8f707baa7d53c411b3147c34a1025ecaa5a744e01d158d8f72d51cc8), counter: 3 }
[src/main.rs:64:12] game.counter = 3
Update: GameState { game_id: 0, player_1: Identity(c2006b4c8f707baa7d53c411b3147c34a1025ecaa5a744e01d158d8f72d51cc8), counter: 2 }
Insert: GameState { game_id: 0, player_1: Identity(c2006b4c8f707baa7d53c411b3147c34a1025ecaa5a744e01d158d8f72d51cc8), counter: 2 }
Decrementing from GameState { game_id: 0, player_1: Identity(c2006b4c8f707baa7d53c411b3147c34a1025ecaa5a744e01d158d8f72d51cc8), counter: 2 }
[src/main.rs:64:12] game.counter = 2
Update: GameState { game_id: 0, player_1: Identity(c2006b4c8f707baa7d53c411b3147c34a1025ecaa5a744e01d158d8f72d51cc8), counter: 1 }
Insert: GameState { game_id: 0, player_1: Identity(c2006b4c8f707baa7d53c411b3147c34a1025ecaa5a744e01d158d8f72d51cc8), counter: 1 }
Decrementing from GameState { game_id: 0, player_1: Identity(c2006b4c8f707baa7d53c411b3147c34a1025ecaa5a744e01d158d8f72d51cc8), counter: 1 }
[src/main.rs:64:12] game.counter = 1
[...]

PR #4607, opened 2026-03-10
Fix disconnects breaking view updates for other connections

joshua-spacetime:
# Description of Changes

Fixes a bug in client disconnect logic that would mark a client's views as dropped(unsubscribed). However it was marking the identity's views as dropped, not the connection. So if an identity had multiple connections open, each subscribing to different views, and one of them disconnected, the subscriptions for the other connections would break. The observed behavior would be that they would stop receiving subscription updates. This could potentially lead to their client cache getting into a corrupted state.

Now, instead of dropping all of the views for a particular identity on disconnect, we drop only the views for that particular connection. And when I say drop, what I really mean is decrement. A view is not dropped completely unless it no longer had any subscribers.

None

2

# Testing

Regression smoketest was added


gmorenz:
That sounds like a good find, but I don't think it explains the issue I'm seeing in #4603. I have a different identity for each connection because I'm starting them with `.with_token(None::<String>)`. And I just confirmed that

```
println!("{}", player1.conn.identity());
println!("{}", _player2.conn.identity());
```

prints two different things.

aceysmith:
@joshua-spacetime Sorry to say, but I don't think this fix is working. I'm on 2.0.5 standalone and rust module (2.0.4 on TS SDK) and I still see this problem. If I connect with one user, then connect and subsequently disconnect with a second user, almost all of my subscriptions for the first user do not work and I see client errors like:

```
Updating a row that was not present in the cache. Table: mobs, RowId: 279
```

AngelOnFira:
I believe I'm still seeing this as well with the Rust SDK at 2.0.5. Similar behaviour to @aceysmith, the second user connecting then disconnecting triggers a panic:

```
thread '<unnamed>' (1232367) panicked at /Users/forest/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/spacetimedb-sdk-2.0.5/src/client_cache.rs:220:13:
internal error: entered unreachable code: a delete update should correspond to an existing row in the table cache, but the row WithBsatn { bsatn: b">4\x1d#\xdf\r\xfb\xd2\x91\xe7\xe0XQ\xdb|\x8dK`\xecx\x85\xa9\xb7\x078\xd4f\xfaX\x86\0\xc2\n\0\0\0Pilot-343E\x01\0\0\0\0\0\0\0\xe8\x03\0\0\0\0\0\0\x01", row: Player { identity: Identity(c2008658fa66d43807b7a98578ec604b8d7cdb5158e0e791d2fb0ddf231d343e), name: "Pilot-343E", planet_id: 1, credits: 1000, online: true } } was not present

thread '<unnamed>' (1232959) panicked at /Users/forest/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/spacetimedb-sdk-2.0.5/src/client_cache.rs:220:13:
internal error: entered unreachable code: a delete update should correspond to an existing row in the table cache, but the row WithBsatn { bsatn: b">4\x1d#\xdf\r\xfb\xd2\x91\xe7\xe0XQ\xdb|\x8dK`\xecx\x85\xa9\xb7\x078\xd4f\xfaX\x86\0\xc2\n\0\0\0Pilot-343E\x01\0\0\0\0\0\0\0\xe8\x03\0\0\0\0\0\0\x01", row: Player { identity: Identity(c2008658fa66d43807b7a98578ec604b8d7cdb5158e0e791d2fb0ddf231d343e), name: "Pilot-343E", planet_id: 1, credits: 1000, online: true } } was not present

...
```

joshua-spacetime:
Updated fix in #4648

PR #4614, opened 2026-03-11
[C#] Primary keys for query builder views

joshua-spacetime:
# Description of Changes

Equivalent changes to https://github.com/clockworklabs/SpacetimeDB/pull/4573, but for C#.

The return type of a query builder view is now a special SATS product type `{ __query__: T }`. A view with this return type now has a primary key if `T` has a primary key. This means that client codegen will generate an `OnUpdate` callback for such views. It will also generate query builder index bindings for the primary key column.

None. Old modules with query builder views still work, they just don't have primary keys.

2

# Testing

Added equivalent tests to the ones that were added in #4573 and #4572.

### DIFF
--- a/sdks/csharp/examples~/regression-tests/client/Program.cs
+++ b/sdks/csharp/examples~/regression-tests/client/Program.cs
@@ -8,6 +8,7 @@ using System.Diagnostics;
 using System.Linq;
 using System.Runtime.CompilerServices;
 using System.Threading;
+using RegressionTests.Shared;
 using SpacetimeDB;
 using SpacetimeDB.Types;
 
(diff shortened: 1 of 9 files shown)


sdks/csharp/tools~/gen-regression-tests.sh
```diff
@@ -10,3 +10,4 @@ cargo build --manifest-path "$STDB_PATH/crates/standalone/Cargo.toml"
 cargo run --manifest-path "$STDB_PATH/crates/cli/Cargo.toml" -- generate -y -l csharp -o "$SDK_PATH/examples~/regression-tests/client/module_bindings" --module-path "$SDK_PATH/examples~/regression-tests/server"
 cargo run --manifest-path "$STDB_PATH/crates/cli/Cargo.toml" -- generate -y -l csharp -o "$SDK_PATH/examples~/regression-tests/republishing/client/module_bindings" --module-path "$SDK_PATH/examples~/regression-tests/republishing/server-republish"
 cargo run --manifest-path "$STDB_PATH/crates/cli/Cargo.toml" -- generate -y -l csharp -o "$SDK_PATH/examples~/regression-tests/procedure-client/module_bindings" --module-path "$STDB_PATH/modules/sdk-test-procedure"
+cargo run --manifest-path "$STDB_PATH/crates/cli/Cargo.toml" -- generate -y -l csharp -o "$SDK_PATH/examples~/regression-tests/view-pk-client/module_bindings" --module-path "$STDB_PATH/modules/sdk-test-view-pk-cs"
```
JasonAtClockwork, line 13: The line here adds to the functionality to generate module bindings for a new test harness but that harness is never run in the `/tools~/run-regression-tests.sh`. 

If we want these only to be manually run that will work but if the intention is to have the CI run the `/examples~/regression-tests/view-pk-client` the `/tools~/run-regression-tests.sh` will need the `publish` and `dotnet run` added.

PR #4620, opened 2026-03-12
Update docs for primary key views

joshua-spacetime:
# Description of Changes

Updates the reference docs with the implications of primary key inference for query builder views, implemented in #4572, #4573, #4614.

None

1. Docs only change

# Testing

N/A


JasonAtClockwork (approved):
Looks good to me!

PR #4636, opened 2026-03-13
Implement HTTP handlers / webhooks in Rust modules

gefjon:
# Description of Changes

Adds support to Rust modules and the SpacetimeDB host for defining HTTP handlers and registering them to routes.

## User-facing API

In a Rust module, users can annotate functions with the new macro `#[spacetimedb::http::handler]`. A function annotated this way must accept exactly two arguments, of types `&mut spacetimedb::http::HandlerContext` and `spacetimedb::http::Request` (which is a type alias for `http::Request<spacetimedb::http::Body>`. It must also return `spacetimedb::http::Response` (which is a type alias for `http::Response<spacetimedb::http::Body>`).

Once the user has defined an HTTP handler, they can register it to a route by annotating a function with `#[spacetimedb::http::router]`. Such a function must take no arguments and return a `spacetimedb::http::Router`. (The original design put this annotation on a `static` variable rather than a function, but that turned out to be undesirable because it required that constructing a `Router` be `const`.) `Router` exposes various methods for registering handlers to routes.

All of a database's user-defined routes are exposed under `/v1/database/:name_or_identity/route/{*path}`.

## Example

See [the new smoketest](https://github.com/clockworklabs/SpacetimeDB/blob/phoebe/http-handlers-webhooks/crates/smoketests/tests/smoketests/http_routes.rs) for a more exhaustive example.

A simpler example, which stores arbitrary byte data in a table via a `POST` request, returns an ID, and then retrieves that same data via a `GET` request with a query parameter:

```rust
#[spacetimedb::table(accessor = data)]
struct Data {
    #[primary_key]
    #[auto_inc]
    id: u64,
    body: Vec<u8>,
}

#[spacetimedb::http::handler]
fn insert(ctx: &mut HandlerContext, request: Request) -> Response {
    let body: Vec<u8> = request.into_body().into_bytes().into();
    let id = ctx.with_tx(|tx| tx.db.data().insert(Data { id: 0, body: body.clone() }).id);
    Response::new(Body::from_bytes(format!("{id}")))
}

#[spacetimedb::http::handler]
fn retrieve(ctx: &mut HandlerContext, request: Request) -> Response {
    let id = request
        .uri()
        .query()
        .and_then(|query| query.strip_prefix("id="))
        .and_then(|id| u64::from_str(id).ok())
        .unwrap();
    let body = ctx.with_tx(|tx| tx.db.data().id().find(id).map(|data| data.body));
    if let Some(body) = body {
        Response::new(Body::from_bytes(body))
    } else {
        Response::builder().status(404).body(Body::empty()).unwrap()
    }
}

#[spacetimedb::http::router]
fn router() -> Router {
    Router::new().post("/insert", insert).get("/retrieve", retrieve)
}
```

## Design and implementation notes

- As mentioned above, the router is registered via a function, not a `static` or `const` item. This is because `static` or `const` initializers must be `const`, and it turns out to be a pain to make all of the `Router` constructors be `const fn`s.
- The `#[handler]` macro clobbers the original function name with a `const` variable of type `HttpHandler`. This is unfortunate, but AFAICT necessary, 'cause we need to pass the string identifier for the handler to the `Router`, not the function pointer, and Rust allows no (stable and reliable) way to get a unique string identifier out of a function item/value, nor to attach data or implement traits for function items/values. The alternative(s) would involve changing the signature of the `Router` methods to have uglier and more complex callsites, e.g. like `.get("/retrieve", retrieve::handler())`, `.get("/retrieve", handler!(retrieve))` or `.get::<retrieve>("/retrieve")`. I believe that registering handlers will be much more common than calling their functions, so I've chosen to make it so that registering them gets the convenient syntax, even though the inability to call them directly will be somewhat surprising.
- I haven't wired up energy handling or timing metrics for handler execution to anywhere. Procedures are still in the same boat.
- HTTP requests to user-defined routes bypass the usual SpacetimeDB auth middleware, meaning that the host does not validate (or inspect in any way) `Authorization` headers in requests before invoking the user-defined handler. This is required to allow arbitrary user-programmable handling of `Authorization` headers, including those in formations which SpacetimeDB would reject. As a result of this, `HandlerContext` doesn't expose a `sender` or `sender_connection_id`.
- HTTP route paths may consist only of a very restrictive set of characters. I've chosen this set to keep our options open in the future to add additional syntax to routes, like for registering wildcard segments and path parameters:
  - ASCII digits.
  - ASCII letters.
  - `-_~/`.
- The internal data structure that represents a `Router` is currently a `Vec<Route>`, meaning that resolving a request to a route is `O(num_routes)`. Registering a route checks against each previous route for uniqueness, meaning that constructing a router is `O(num_routes ^ 2)`. There are TODO comments to use a trie, but I think this can wait, as I expect most databases to register few routes.
- Commit 999a7c317 contains a fix to a mostly-unrelated bug where a few bindings introduced by the SATS derive macros were unhygienic and not in a reserved namespace, leading to name conflicts. I discovered this 'cause I tried writing an HTTP handler named `index` to serve the index/root of a website and it broke.

## Still TODO

- [x] Resolve various TODO comments in the diff.
- [x] Documentation.
- [x] C# bindings support.
- [x] C++ bindings support.
- [x] V8 host support.
- [x] TypeScript bindings support.

New APIs, currently flagged as `unstable`, which will eventually need stability guarantees. No (intentional) breaking changes, or changes to existing APIs at all.

3? Changes to our HTTP routing to support the user-defined routes.

# Testing

- [x] New smoketest of the behavior!
- [x] I dunno, maybe try hosting a simple webpage and see how it works?
- [x] Build a test app with Stripe integration.
  - @aasoni did this.


bfops (approved):
my code-owned files LGTM:
```
tools/ci/src/main.rs
```

gefjon:
I had Codex vibe-code up a simple static webpage hosted entirely in a SpacetimeDB module. This was easy enough and mostly totally worked, but we ran into a few silly issues which I'll work to fix over the next few days. In particular:

- [x] A name conflict between having an HTTP handler named `index` and a table. This would, I believe, manifest if any module code defined a `const` named `index` in the same file as a `#[table]` macro invocation, but that's unlikely to happen in the wild due to `const`s usually having all-caps names.
- [x] The root route was accessible at `/v1/database/:name/route`, but not at `/v1/database/:name/route/` (note the trailing slash). This was especially unintuitive as the root route is registered as `"/"`.
- [x] The request path visible inside the handler was different from what was expected. The AI said:
  > Externally the page was served at `/v1/database/{db}/route`, but inside the handler the incoming path looked like `/{db}/route...`, [...]
- [x] We get warnings from the `#[handler]` macro output warning about lowercase constants that should be uppercase. This just needs an `#[allow]`.

crates/bindings/src/http.rs
```diff
@@ -5,18 +5,361 @@
...
+    /// Methods for performing HTTP requests.
+    pub http: HttpClient,
+
+    #[cfg(feature = "rand08")]
+    pub(crate) rng: OnceCell<StdbRng>,
+
+    /// A counter used for generating UUIDv7 values.
+    /// **Note:** must be 0..=u32::MAX
+    #[cfg(feature = "rand")]
+    pub(crate) counter_uuid: Cell<u32>,
```
cloutiertyler, line 98: What's the point of the `Cell` out of pure curiosity?

crates/bindings/src/http.rs
```diff
@@ -5,18 +5,361 @@
...
+    }
+
+    /// Causes requests which start with `path` to be processed by `sub_router`.
+    ///
+    /// `sub_router` will be used by stripping the leading `path` from the path of the request.
+    ///
+    /// Panics if `self` already has any handlers registered on paths which start with `path`.
+    ///
+    /// Panics if the `path` does not begin with `/`, or if it contains any characters which are not URL-safe.
+    pub fn nest(self, path: impl Into<String>, sub_router: Self) -> Self {
```
cloutiertyler, line 314: NIT: Why does the path require `/`. Not saying that's wrong, but if I have `/foo/bar/` and I nest `/baz`, does the path end up being `/foo/bar/baz` or `/foo/bar//baz`? Presumably the former.

BIKESHEDDING: I'm fine with `nest`. Presumably we would translate this to all the other languages as well? We might consider if something like `use` would be more familiar, but arguably `use` is less clear/more confusing for newcomers.

crates/bindings/src/http.rs
```diff
@@ -199,6 +542,82 @@ fn convert_response(response: st_http::Response) -> http::Result<http::response:
...
+        version,
+    } = request;
+
+    let method = match method {
+        st_http::Method::Get => http::Method::GET,
+        st_http::Method::Head => http::Method::HEAD,
+        st_http::Method::Post => http::Method::POST,
+        st_http::Method::Put => http::Method::PUT,
+        st_http::Method::Delete => http::Method::DELETE,
+        st_http::Method::Connect => http::Method::CONNECT,
```
cloutiertyler, line 596: What do we do with connect requests?

crates/bindings/tests/ui/http_handlers.rs
```diff
@@ -0,0 +1,68 @@
...
+    todo!()
+}
+
+#[handler]
+fn handler_immutable_ctx(_ctx: &HandlerContext, _req: Request) -> Response {
+    todo!()
+}
+
+#[handler]
+fn handler_wrong_ctx(_ctx: &mut ProcedureContext, _req: Request) -> Response {
```
cloutiertyler, line 15: How does this test work? How do we know these are all failing to compile?

cloutiertyler: JK it's in the next file.

crates/smoketests/tests/smoketests/http_routes.rs
cloutiertyler, line 1: I don't believe these tests access `req.body` anywhere.

tools/ci/src/main.rs
```diff
@@ -464,12 +464,26 @@ fn main() -> Result<()> {
...
                 "spacetimedb-sdk",
+                "--exclude",
+                "spacetimedb",
                 "--",
                 "--test-threads=2",
                 "--skip",
                 "unreal"
             )
             .run()?;
+            // Bindings snapshot tests rely on the unstable feature.
```
cloutiertyler: Porque?

cloutiertyler (commented):
This is a very high quality PR. I have the following comments though:

1. I don't love `/v1/database/:name_or_identity/route/{*path}`, but I understand. What would your thoughts be on mounting it under `/:name_of_identity/{*path}` where we reserve certain names (e.g. v1) and maybe a standardized list.

2. I think we need to think about the request body streaming (and possibly response body streaming). We want to make sure our API allows for it even if we don't support it immediately.

JulienLavocat:
> I don't love /v1/database/:name_or_identity/route/{*path}, but I understand. What would your thoughts be on mounting it under /:name_of_identity/{*path} where we reserve certain names (e.g. v1) and maybe a standardized list.

@cloutiertyler I think that reserving certain name will do more harm than good, **especially** `v1` as it's quite a common pattern to version REST APIs by prefixing the routes with a version like this: `/v1/users/metadata`, `/v2/users/metadata`.

I think having `/v1/database/:name_or_identity/router` or `/v1/database/:name_or_identity/api` is more explicit than `/v1/database/:name_or_identity`

cloutiertyler:
@JulienLavocat I am not suggesting `/v1/database/:name_or_identity`, I am suggesting `/:name_or_identity`

crates/bindings/src/http.rs
gefjon: The methods that generate UUIDs take `&self` shared references, but mutate this field to increment the counter.

crates/bindings/src/http.rs
gefjon: The only way to register a handler that will receive `CONNECT` requests is via the `Router::any` method; we don't expose `Router::connect`. For `any` handlers, we pass the HTTP request method through to the module, where the module can read the fact that the request specified the `CONNECT` method. We don't support streaming or upgrades.

crates/bindings/src/http.rs
gefjon: > if I have /foo/bar/ and I nest /baz, does the path end up being /foo/bar/baz or /foo/bar//baz? Presumably the former.

Correct, when combining path segments we normalize runs of multiple slashes into a single slash.

I don't have a particular opinion about whether we should be treating it as an error to register paths that don't start with `/` - we could instead automatically insert `/`es between path segments when combining.

> We might consider if something like use would be more familiar, but arguably use is less clear/more confusing for newcomers.

`use` is a reserved word in Rust, and would require users to actually write `r#use(PATH, CHILD)` here, which I think we obviously don't want. I default to using the same method name in our other languages as well, unless there's a strong reason not to. IMO `use` is a silly and unhelpful name for this method, whereas `nest` is pretty obvious and intuitive.

crates/smoketests/tests/smoketests/http_routes.rs
gefjon: Oh, you're right. I'll add one that does.

crates/bindings/src/http.rs
cloutiertyler: Roger, I agree.

gefjon:
> 2\. I think we need to think about the request body streaming (and possibly response body streaming). We want to make sure our API allows for it even if we don't support it immediately.

I'm going to make some minor adjustments to the ABI so that the request body is passed as a separate `BytesSource` from the request metadata. That will allow the module to read from it gradually down the line.

I'm not sure how to handle response body streaming. What we want to do on that front will depend on our goals: do we want to allow long-running connections with lazily generated responses? That would likely require adding a different kind of handler which mimics the ExpressJS API where the handler writes to a response sink rather than returning a response. Do we just want to allow returning large rows or files? [Bun supports that](https://bun.com/docs/guides/http/stream-file) using an API very much like our own.

gefjon:
>     1. I don't love `/v1/database/:name_or_identity/route/{*path}`, but I understand. What would your thoughts be on mounting it under `/:name_of_identity/{*path}` where we reserve certain names (e.g. v1) and maybe a standardized list.

In my mind, there are two different uses for HTTP handlers which impose different constraints here:

### Websites or public APIs

A production app tells its end users to send HTTP requests to a particular address. This might be by typing that address into a browser, or by configuring an end user owned bot, tool or application (e.g. a Discord bot, an OBS overlay) to send those requests. The common feature of this category is that end users, not SpacetimeDB customers, are explicitly exposed to the address. In my mind, it's pretty much certain that the production app will want a reverse proxy or redirect s.t. the end users see a domain name specific to that app, owned by our customer. (Perhaps we should offer this as a service?) In that case, it doesn't really matter what the underlying routes SpacetimeDB listens on, so long as it's easy to set up the redirection.

### Webhooks

A customer needs a stable URI which they can plug in to some other service they also use, e.g. Stripe, Github. That URI is not visible to their end users, only to the person or people developing an app with SpacetimeDB. In this case, it also doesn't really matter what the route is SpacetimeDB listens on, so long as it's easy to determine what it is in order to register it with the external service.

### Conclusion

I don't love `/:name_or_identity/{*path}` because of the risk of conflicts and the difficulty of predicting what names we need to reserve, and because I think the payoff of simplifying user-defined routes is relatively low, for the reasons described above. Out of an abundance of caution, I'd prefer that we never have a user-defined route or a wildcard as a sibling to one of our own routes. Other than that, I don't think it matters very much where we nest user-defined routes under our own router, and I don't really care.

crates/bindings-macro/src/http.rs
```diff
@@ -0,0 +1,113 @@
...
+        pub extern "C" fn __register_describer() {
+            spacetimedb::rt::register_http_handler(#handler_name, #internal_ident)
+        }
+    };
+
+    Ok(quote! {
+        #inner_fn
+
+        #[allow(non_upper_case_globals)]
+        #vis const #func_name: spacetimedb::http::Handler = spacetimedb::http::Handler::new(#handler_name);
```
Centril, line 63: Not necessary now, but if we want to in the future, we could re-export the `inner_fn` with `use` with e.g., `foo_original` or something.

tools/ci/src/main.rs
gefjon: I expanded this comment:

```rust
// Bindings snapshot tests rely on the unstable feature,
// as they compile and test APIs which are gated behind that feature,
// e.g. procedures, HTTP handlers.
```

crates/bindings/src/http.rs
```diff
@@ -5,18 +5,442 @@
...
+/// # use spacetimedb::http::{handler, Request, Response, Body, HandlerContext};
+/// #[handler]
+/// fn hello_world(_ctx: &mut HandlerContext, _req: Request) -> Response {
+///     Response::new(Body::from_bytes("Hello, world!"))
+/// }
+/// ```
+///
+/// In order to be reachable, a handler must be registered in the database's [macro@router].
+///
+/// This macro will clobber the original function definition, making it no longer callable by name.
```
Centril:
```suggestion
/// This macro will shadow the original function definition, making it no longer callable by name.
```

crates/bindings/src/http.rs
```diff
@@ -5,18 +5,442 @@
...
+/// # use spacetimedb::http::{handler, Request, Response, Body, HandlerContext};
+/// # #[handler]
+/// # fn hello_world(_ctx: &mut HandlerContext, _req: Request) -> Response {
+/// #     Response::new(Body::from_bytes("Hello, world!"))
+/// # }
+/// # fn foo() {
+/// # let ctx: HandlerContext = todo!();
+/// # let ctx: &mut HandlerContext = &mut ctx;
+/// # let req: Request = todo!();
+/// hello_world(ctx, req); // Won't compile, as our handler `hello_world`'s function was clobbered.
```
Centril:
```suggestion
/// hello_world(ctx, req); // Won't compile, as our handler `hello_world`'s function was shadowed.
```

crates/bindings/src/http.rs
```diff
@@ -5,18 +5,442 @@
...
+    pub fn with_tx<T>(&mut self, body: impl Fn(&TxContext) -> T) -> T {
+        use core::convert::Infallible;
+        match self.try_with_tx::<T, Infallible>(|tx| Ok(body(tx))) {
+            Ok(v) => v,
+            Err(e) => match e {},
+        }
+    }
+
+    /// Acquire a mutable transaction and execute `body` with read-write access.
+    pub fn try_with_tx<T, E>(&mut self, body: impl Fn(&TxContext) -> Result<T, E>) -> Result<T, E> {
```
Centril, line 124: Could this not share some code with `ProcedureContext` by moving both to a common free function? The code seems identical.

crates/bindings/src/http.rs
```diff
@@ -5,18 +5,442 @@
...
+///
+/// The empty string `""` is a valid path, which refers to the root route.
+///
+/// All other paths must start with a slash `/`.
+///
+/// Only characters described by [`spacetimedb_lib::http::ACCEPTABLE_ROUTE_PATH_CHARS_HUMAN_DESCRIPTION`]
+/// are valid for use in paths. This set may be expanded in the future.
+///
+/// SpacetimeDB uses strict routing, meaning that trailing slashes `/` in paths are significant.
+/// `/foo` and `/foo/` are distinct paths.
```
Centril, line 197: Not sure, but this seems a bit surprising? I'm assuming there's precedent...?

crates/bindings/src/http.rs
```diff
@@ -5,18 +5,442 @@
...
+pub(crate) struct RouteSpec {
+    pub method: MethodOrAny,
+    pub path: String,
+    pub handler: Handler,
+}
+
+impl Router {
+    /// Returns a new, empty `Router`.
+    pub fn new() -> Self {
+        Self { routes: Vec::new() }
```
Centril:
```suggestion
        Self::default()
```

crates/lib/src/db/raw_def/v10.rs
```diff
@@ -805,6 +849,46 @@ impl RawModuleDefV10Builder {
...
+            });
+
+        match &mut self.module.sections[idx] {
+            RawModuleDefV10Section::HttpHandlers(handlers) => handlers,
+            _ => unreachable!("Just ensured HttpHandlers section exists"),
+        }
+    }
+
+    /// Get mutable access to the HTTP routes section, creating it if missing.
+    fn http_routes_mut(&mut self) -> &mut Vec<RawHttpRouteDefV10> {
```
Centril, line 873: Feels like this could share some code with the above.

Centril (approved):
Looks good; just have some minor comments.

crates/bindings/src/http.rs
gefjon: Yes, ExpressJS calls this "strict routing." https://expressjs.com/en/5x/api.html#express.router . Axum doesn't have a name for this, it just works this way.

crates/lib/src/db/raw_def/v10.rs
gefjon: What do you mean by "the above?" It's not clear to me what code could usefully be shared here.

N.b. this method and `http_handlers_mut`, match existing methods in this file:

- `explicit_names_mut`
- `row_level_security_mut`
- `types_mut`
- `lifecycle_reducers_mut`
- `schedules_mut`
- `views_mut`
- `procedures_mut`
- `reducers_mut`
- `typespace_mut`

crates/bindings-macro/src/http.rs
gefjon: I wonder if we can impl `Deref` with the target `fn(&mut HandlerContext, Request) -> Response`? I don't know if the function call operator goes through deref coercion.

crates/bindings-typescript/src/server/http_handlers.ts
```diff
@@ -0,0 +1,403 @@
...
+}
+
+function joinPaths(prefix: string, suffix: string) {
+  if (prefix === '/') {
+    return suffix;
+  }
+  if (suffix === '/') {
+    return prefix;
+  }
+  const joinedPrefix = prefix.replace(/\/+$/, '');
```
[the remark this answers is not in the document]
JasonAtClockwork: Updated the joining to a simpler approach

PR #4640, opened 2026-03-14
Small typescript sdk optimizations

joshua-spacetime:
# Description of Changes

Was doing some profiling and found some low hanging optimizations for the ts client. Namely caching table primary keys and reusing the writer for reducer args.

None

1

# Testing

...


coolreader18 (approved):
LGTM!

PR #4647, opened 2026-03-16
Improve benchmark cli, make compatible with deno

coolreader18:
# Description of Changes

Now we get a `--help` for the benchmark, which is nicer. Also now can run under deno, with `deno --sloppy-imports -A src/demo.ts` (might be useful, deno's websocket is implemented in native code while node's is implemented in JS). I removed the [BOM](https://en.wikipedia.org/wiki/Byte_order_mark) because it seems unintentional (only found in `templates/keynote-2`) and was causing a little bit of weirdness.

Also, fix the rust benchmark client as a follow-up to #4616 

1

# Testing

- [x] Works under deno and has usage

### DIFF
--- a/templates/keynote-2/spacetimedb-rust-client/src/main.rs
+++ b/templates/keynote-2/spacetimedb-rust-client/src/main.rs
@@ -130,7 +130,7 @@ fn bench(cli: &Common, bench: &Bench) {
     // Dump some config parameters.
     let alpha = bench.alpha;
     let accounts = cli.accounts;
-    let amount = bench.amount as i64;
+    let amount = bench.amount;
     if !cli.quiet {
         println!("Benchmark parameters:");
         println!("alpha={alpha}, amount = {amount}, accounts = {accounts}");


templates/keynote-2/src/opts.ts
```diff
@@ -0,0 +1,270 @@
...
+  .option(
+    '--stdb-module-path <dir>',
+    'SpacetimeDB module path',
+    str('./spacetimedb', 'STDB_MODULE_PATH'),
+  )
+  .option(
+    '--stdb-module-path <dir>',
+    'SpacetimeDB module path',
+    str('./spacetimedb', 'STDB_MODULE_PATH'),
+  )
```
bradleyshep:
```suggestion
```
Duplicate

templates/keynote-2/src/opts.ts
```diff
@@ -0,0 +1,270 @@
...
+  .option('--stdb-confirmed-reads', 'Enable confirmed reads', {
+    env: 'STDB_CONFIRMED_READS',
+  })
```
bradleyshep: Should this default to true?

PR #4650, opened 2026-03-17
commitlog: Resumption of sealed commitlog

kim:
The commitlog so far assumed that the latest segment is never compressed and can be opened for writing (if it is intact).

However, restoring the entire commitlog from cold storage results in all segments being compressed. Make it so the resumption logic reads the metadata from the potentially compressed last segment, and starts a new segment for writing if the latest one was indeed compressed.

1.5

# Testing

Added a test.

### DIFF
--- a/crates/commitlog/src/repo/mod.rs
+++ b/crates/commitlog/src/repo/mod.rs
@@ -289,6 +289,12 @@ pub fn resume_segment_writer<R: Repo>(
             max_commit: _,
         } = meta;
         let mut writer = repo.open_segment_writer(offset)?;
+        // Ensure we have enough space for this segment.
+        // The segment could have been created without the `fallocate` feature
+        // enabled, so we call this here again to ensure writes can't fail due
+        // to ENOSPC.
+        fallocate(&mut writer, &opts)?;
+        // We use `O_APPEND`, but make the file offset consistent regardless.
         writer.seek(io::SeekFrom::End(0))?;
 
         Ok(Ok(Writer {


crates/commitlog/src/repo/mod.rs
```diff
@@ -243,21 +252,9 @@ pub fn resume_segment_writer<R: Repo>(
     opts: Options,
     offset: u64,
 ) -> io::Result<Result<Writer<R::SegmentWriter>, Metadata>> {
-    let mut storage = repo.open_segment_writer(offset)?;
-    // Ensure we have enough space for this segment.
-    // The segment could have been created without the `fallocate` feature
-    // enabled, so we call this here again to ensure writes can't fail due to
-    // ENOSPC.
-    fallocate(&mut storage, &opts)?;
```
Shubham8287, line 251: Is it intentionally removed?

kim: Oops

crates/commitlog/src/repo/mod.rs
```diff
@@ -266,34 +263,49 @@ pub fn resume_segment_writer<R: Repo>(
...
+        let Metadata {
+            header: _,
+            tx_range,
+            size_in_bytes,
+            max_epoch,
+            max_commit_offset: _,
+            max_commit: _,
+        } = meta;
+        let mut writer = repo.open_segment_writer(offset)?;
+        writer.seek(io::SeekFrom::End(0))?;
```
Shubham8287, line 292: Do we need to `seek` at the end? If yes, how was it working before?

kim: We were reading the metadata, advancing the offset implicitly.
`O_APPEND` is probably enough for files, but I wanted to avoid surprises in case non-file impls have the semantics wrong. I also didn't check if we're reading from the handle somewhere.

PR #4657, opened 2026-03-17
[Rust] Primary keys for procedural views

joshua-spacetime:
# Description of Changes

We already have primary key inference for query builder views. This adds explicit primary key syntax and enforcement for procedural views.

More to come ...

A new `RawViewDefV11` was needed, but `RawViewDefV10` is still supported.

3

# Testing

TODO


joshua-spacetime:
Superseded by #5111

PR #4661, opened 2026-03-18
core, datastore: Work around race in request_durability

kim:
Both `commit` and `commit_downgrade` release the exclusive lock on the committed state before returning control to `RelationalDB`. The latter is then responsible for submitting the `TxData` to the durability worker via `request_durability`.

Perhaps surprisingly -- datastore operations are supposedly running on a single thread -- it is possible that the `RelationalDB` gets preempted after committing the transaction, but before having called `request_durability`. This can result in out-of-order transactions being submitted to the durability layer.

The presented workaround prevents this by forcing the datastore to trigger durability submission before releasing the lock. This fixes the issue at hand, but the author suspects that the long term fix involves revising database scheduling, namely enforcing single-threaded execution.

### DIFF
--- a/crates/datastore/src/locking_tx_datastore/mut_tx.rs
+++ b/crates/datastore/src/locking_tx_datastore/mut_tx.rs
@@ -2032,7 +2032,10 @@ impl MutTxId {
     /// - [`TxData`], the set of inserts and deletes performed by this transaction.
     /// - [`TxMetrics`], various measurements of the work performed by this transaction.
     /// - `String`, the name of the reducer which ran during this transaction.
-    pub(super) fn commit(mut self) -> (TxOffset, TxData, TxMetrics, Option<ReducerName>) {
+    pub(super) fn commit(
+        mut self,
+        before_release: impl FnOnce(&Arc<TxData>),
+    ) -> (TxOffset, Arc<TxData>, TxMetrics, Option<ReducerName>) {
         let tx_offset = self.committed_state_write_lock.next_tx_offset;
         let tx_data = self
             .committed_state_write_lock
@@ -2066,6 +2069,9 @@ impl MutTxId {
             tx_offset
         };
 
+        let tx_data = Arc::new(tx_data);
+        before_release(&tx_data);
+
         (tx_offset, tx_data, tx_metrics, reducer)
     }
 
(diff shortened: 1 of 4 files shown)


Shubham8287:
Hmm, right now we have “single-threaded” access from the module, but the Host parallelly  accessing the db can cause this.

gefjon (commented):
I'd like to see comments throughout this diff describing what the bug was and why we're shaping the code this way. In particular, I'd like you to pick one spot in the code to summarize the issue and to link to this PR. Then, update the doc comments on all of the functions you added `before_release` and `before_downgrade` parameters to have the same comment, which should say something like, "The `before_release` will be run after computing the `TxData` but before releasing the transaction lock. `RelationalDB::commit_tx` passes a closure which does `Durability::request_durability`. See \<reference to where you put the longer description\>."

crates/datastore/src/traits.rs
```diff
@@ -240,6 +240,7 @@ pub struct TxData {
 impl TxData {
     /// Set `tx_offset` as the expected on-disk transaction offset of this transaction.
```
Centril, line 241:
```suggestion
    /// Set `tx_offset` as the expected on-disk transaction offset of this transaction.
    ///
    /// Panics if the offset has already been set for this transaction.
```

crates/datastore/src/locking_tx_datastore/mut_tx.rs
Centril, line 2057: Can you move this to after `before_release` to avoid some of the performance penalties of this PR?

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -2101,6 +2111,9 @@ impl MutTxId {
             &self.committed_state_write_lock,
         );
```
Centril, line 2112: here too

crates/datastore/src/locking_tx_datastore/mut_tx.rs
kim: The metrics? Is it correct, though? Namely the timing stuff.

Centril: It's correct-ish? My take on the metrics is that we don't care about the precision in that level of detail and that we care more about the throughput.

kim:
Closed in favor of #4677

PR #4663, opened 2026-03-18
Replace `JsInstance` pool with single worker and FIFO queue

joshua-spacetime:
# Description of Changes

Before this change, JS reducer requests borrowed a `JsInstance` from a pool. If no idle instance was available, we created another instance, which meant another V8 worker thread. Under load, this meant reducers bouncing across multiple OS threads.

After this change, JS reducers go through a single long-lived `JsInstance` fed by a FIFO queue which results in much better cache locality. More accurately, each module now allocates a single OS thread, on which reducers (and most operations) run. Modules do not share workers/threads. And modules do not create multiple threads for running reducers.

Note, the original instance pool is still used for procedures. It should probably be bounded, but I didn't make any changes to it. It's also used for executing views during initial subscription to avoid a reentrancy deadlock. The latter should be fixed and moved over to the JS worker thread at some point.

N/A

4

# Testing

```
NODE_OPTIONS="--max-old-space-size=8192" \
MAX_INFLIGHT_PER_WORKER=512 \
BENCH_PRECOMPUTED_TRANSFER_PAIRS=1000000 \
pnpm bench test-1 --seconds 10 --concurrency 50 --alpha 1.5 --connectors spacetimedb
```
```
50K TPS -> 85K TPS on m2 mac
```

### DIFF
--- a/crates/core/src/host/module_host.rs
+++ b/crates/core/src/host/module_host.rs
@@ -1269,28 +1269,53 @@ impl ModuleHost {
         .await
     }
 
-    async fn with_js_pooled_instance<R>(
+    /// Run a function for this module using pooled instances.
+    ///
+    /// For WASM, this is identical to [`Self::call`].
+    /// For V8/JS, this uses the pooled procedure instances instead of the
+    /// single instance lane.
+    async fn call_pooled<A, R>(
         &self,
         label: &str,
-        f: impl AsyncFnOnce(&JsInstance) -> R,
-    ) -> Result<R, NoSuchModule> {
+        arg: A,
+        wasm: impl AsyncFnOnce(A, &mut ModuleInstance) -> R + Send + 'static,
+        js: impl AsyncFnOnce(A, &JsInstance) -> R,
+    ) -> Result<R, NoSuchModule>
+    where
+        R: Send + 'static,
+        A: Send + 'static,
+    {
         self.guard_closed()?;
         let timer_guard = self.start_call_timer(label);
 
         scopeguard::defer_on_unwind!({
-            log::warn!("pooled JS instance operation {label} panicked");
+            log::warn!("pooled operation {label} panicked");
             (self.on_panic)();
         });
 
         Ok(match &*self.inner {
-            ModuleHostInner::Wasm(_) => unreachable!("WASM should not use the pooled JS instance path"),
+            ModuleHostInner::Wasm(WasmtimeModuleHost {
+                executor,
+                instance_manager,
+            }) => {
+                instance_manager
+                    .with_instance(async |mut inst| {
+                        executor
+                            .run_job(async move || {
+                                drop(timer_guard);
+                                (wasm(arg, &mut inst).await, inst)
+                            })
+                            .await
+                    })
+                    .await
+            }
             ModuleHostInner::Js(V8ModuleHost {
                 procedure_instances, ..
             }) => {
                 procedure_instances
                     .with_instance(async |inst| {
                         drop(timer_guard);
-                        let res = f(&inst).await;
+                        let res = js(arg, &inst).await;
                         (res, inst)
                     })
                     .await
@@ -1299,21 +1324,13 @@ impl ModuleHost {
     }
 
     async fn call_view_command(&self, label: &str, cmd: ViewCommand) -> Result<ViewCommandResult, ViewCallError> {
-        Ok(match &*self.inner {
-            ModuleHostInner::Wasm(_) => {
-                self.call(
-                    label,
-                    cmd,
-                    async |cmd, inst| Ok::<_, ViewCallError>(inst.call_view(cmd)),
-                    async |_cmd, _inst| unreachable!("WASM should not use the JS call_view path"),
-                )
-                .await??
-            }
-            ModuleHostInner::Js(_) => {
-                self.with_js_pooled_instance(label, async |inst| inst.call_view(cmd).await)
-                    .await?
-            }
-        })
+        self.call_pooled(
+            label,
+            cmd,
+            async |cmd, inst| Ok::<_, ViewCallError>(inst.call_view(cmd)),
+            async |cmd, inst| Ok::<_, ViewCallError>(inst.call_view(cmd).await),
+        )
+        .await?
     }
 
     pub async fn disconnect_client(&self, client_id: ClientActorId) {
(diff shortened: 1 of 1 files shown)


Centril:
On phoenix nap + rust client, this takes performance to about 150k TPS, which is around 15k lower than for the rust module.

crates/core/src/host/v8/mod.rs
```diff
@@ -290,233 +339,512 @@ impl JsInstanceEnv {
...
+/// on the worker thread so the next reducer can start immediately after the
+/// previous one finishes, without waiting for an outer task to hand the
+/// instance back.
+///
+/// When the last handle is dropped, the channels will hang up,
+/// which will cause the worker's loop to terminate and cleanup the isolate
+/// and friends.
+#[derive(Clone)]
 pub struct JsInstance {
+    id: u64,
```
Centril, line 357: Would be good to add a doc comment re. the role of `id`.

crates/core/src/host/v8/mod.rs
```diff
@@ -290,233 +339,512 @@ impl JsInstanceEnv {
...
+///
+/// Most variants carry a `reply_tx` because the worker thread owns the isolate,
+/// executes the request there, and then has to send both the typed result and
+/// the worker's trapped-bit back to the async caller.
 enum JsWorkerRequest {
-    /// See [`JsInstance::run_on_thread`].
-    ///
-    /// This variant does not expect a [`JsWorkerReply`].
     RunFunction(Box<dyn FnOnce() -> LocalBoxFuture<'static, ()> + Send>),
-    /// See [`JsInstance::update_database`].
```
Centril, line 490: Please readd these doc comments as they provide useful link-backs.

crates/core/src/host/v8/mod.rs
```diff
@@ -290,233 +339,512 @@ impl JsInstanceEnv {
...
+    //
+    // This stays safe as long as these invariants hold:
+    // - `replace_lock` is only for trap/disconnect recovery, never the hot path.
+    // - no `parking_lot` guard is held across the `.await` in replacement.
+    // - `create_instance()` must not call back into `JsInstanceLane` or try to
+    //   take `replace_lock`.
+    replace_lock: AsyncMutex<()>,
+}
+
+#[derive(Clone)]
```
Centril, line 617: Doc comment on this, e.g., document what a lane is in this context.

crates/core/src/host/module_host.rs
```suggestion
    fn new(module: M, init_inst: Option<M::Instance>, database_identity: Identity) -> Self {
```

```suggestion
let mut instances = VecDeque::new();
instances.extend(init_inst);
```

crates/core/src/host/module_host.rs
```diff
@@ -800,6 +802,23 @@ impl<M: GenericModule> ModuleInstanceManager<M> {
...
+            database_identity,
+        };
+
+        Self {
+            instances: Mutex::new(VecDeque::new()),
+            module,
+            create_instance_time_metric,
+        }
+    }
+
```
```suggestion
```

crates/core/src/host/module_host.rs
```diff
@@ -1028,8 +1049,13 @@ impl ModuleHost {
             }
             ModuleWithInstance::Js { module, init_inst } => {
                 info = module.info();
-                let instance_manager = ModuleInstanceManager::new(module, init_inst, database_identity);
-                Arc::new(ModuleHostInner::Js(V8ModuleHost { instance_manager }))
+                let instance_lane = super::v8::JsInstanceLane::new(module.clone(), init_inst);
+                let procedure_instances = ModuleInstanceManager::new_empty(module.clone(), database_identity);
```
```suggestion
                let instance_lane = super::v8::JsInstanceLane::new(module.clone(), init_inst);
                let procedure_instances = ModuleInstanceManager::new(module.clone(), None, database_identity);
```

crates/core/src/host/module_host.rs
```diff
@@ -1262,13 +1261,61 @@ impl ModuleHost {
...
+        scopeguard::defer_on_unwind!({
+            log::warn!("pooled JS instance operation {label} panicked");
+            (self.on_panic)();
+        });
+
+        Ok(match &*self.inner {
+            ModuleHostInner::Wasm(_) => unreachable!("WASM should not use the pooled JS instance path"),
+            ModuleHostInner::Js(V8ModuleHost {
+                procedure_instances, ..
+            }) => {
```
coolreader18: Instead of this, why not just have a `call_pooled`, which is just like `call` but uses `procedure_instances` in the JS case? Seems simpler and easier.

crates/core/src/host/module_host.rs
```diff
@@ -1885,26 +1901,63 @@ impl ModuleHost {
...
+            }
+            ModuleHostInner::Js(V8ModuleHost { module, .. }) => {
+                let use_procedure_lane =
+                    match params.uses_procedure_lane(&self.info, module.replica_ctx().relational_db.as_ref()) {
+                        Ok(use_procedure_lane) => use_procedure_lane,
+                        Err(err) => {
+                            log::error!("failed to classify scheduled JS function; routing to procedure lane: {err:#}");
+                            true
+                        }
+                    };
```
coolreader18: And then here you just pull this into
```rust
let use_procedure_lane = match inner {
    Js => params.uses_procedure_lane(...).unwrap_or_else(|err| true),
    Wasm => true,
};
if use_procedure_lane {
    self.call_pooled(foo, bar, baz)
} else {
    self.call(foo, bar, baz)
}
```

crates/core/src/host/module_host.rs
```diff
@@ -1885,26 +1901,63 @@ impl ModuleHost {
...
+                    "unknown scheduled function",
+                    params,
+                    async move |params, inst| Ok(inst.call_scheduled_function(params).await),
+                    async move |_params, _inst| unreachable!("JS scheduled-function lane is not used for WASM modules"),
+                )
+                .await?
+            }
+            ModuleHostInner::Js(V8ModuleHost { module, .. }) => {
+                let use_procedure_lane =
+                    match params.uses_procedure_lane(&self.info, module.replica_ctx().relational_db.as_ref()) {
```
coolreader18: Is this gonna cause issues at all? Won't it deadlock, since it's starting a transaction on a tokio thread?

coolreader18 (changes requested):

crates/core/src/host/module_host.rs
joshua-spacetime: Yes, I meant to refactor this. I've updated it so that all scheduled functions go through the instance pool. We can refactor this later so that at the very least scheduled reducers use the worker instead.

crates/core/src/host/module_host.rs
joshua-spacetime: I took your suggestion, although this is now simplified due to the fact that all scheduled functions are using the pool.

PR #4664, opened 2026-03-18
Add C++ query builder

### LINKED ISSUE #3751: [Unreal] Module bindings for query builder

JasonAtClockwork:
# Description of Changes
- Added a query builder for C++ module bindings
  - Added query-builder table/filter/join types
  - Added semijoin support with compile-time checks for lookup-table and indexed-field usage
  - Added support for returning query-builder queries from C++ views
  - Hooked query-builder metadata into the C++ table/view macros and V10 module-def path
- Added test coverage for the new C++ query-builder behavior
  - Compile tests for pass/fail cases
  - SQL tests for generated query output
  - Added a C++ test module for view primary key coverage
- **Update:** Switched the core to pass the columns and index-columns metadata with the table source for better client-side codegen to have some shared code between server + client.
- No intended API or ABI breaking changes
- Adds a new public query-builder API to the C++ bindings
- C++ views can now return query-builder query types in addition to materialized row results

3 - Mostly contained to C++ bindings, but it touches macros, view registration/serialization, and module-def generation, so there are a few places where the pieces need to stay in sync.

# Testing

I've done end to end testing of I think every type as well as built some tests to confirm the SQL output.

- [x] Run the C++ query-builder SQL tests [crates/bindings-cpp/tests/query-builder-compile/run_query_builder_compile_tests.sh]
- [x] Smoke test a generated C++ module using query-builder views


crates/bindings-cpp/include/spacetimedb/query_builder/expr.h
```diff
@@ -0,0 +1,333 @@
...
+        return BoolExpr(std::make_shared<Node>(Kind::And, root_, other.root_));
+    }
+
+    [[nodiscard]] BoolExpr or_(const BoolExpr& other) const {
+        return BoolExpr(std::make_shared<Node>(Kind::Or, root_, other.root_));
+    }
+
+    [[nodiscard]] BoolExpr not_() const {
+        return BoolExpr(std::make_shared<Node>(Kind::Not, root_, nullptr));
+    }
```
rekhoff, line 214: It might be nice to add a comments here explaining the underscores.
```suggestion
    // Trailing underscore to avoid conflicting with C++ keyword
    [[nodiscard]] BoolExpr and_(const BoolExpr& other) const {
        return BoolExpr(std::make_shared<Node>(Kind::And, root_, other.root_));
    }

    // Trailing underscore to avoid conflicting with C++ keyword
    [[nodiscard]] BoolExpr or_(const BoolExpr& other) const {
        return BoolExpr(std::make_shared<Node>(Kind::Or, root_, other.root_));
    }

    // Trailing underscore to avoid conflicting with C++ keyword
    [[nodiscard]] BoolExpr not_() const {
        return BoolExpr(std::make_shared<Node>(Kind::Not, root_, nullptr));
    }
```

docs/docs/00200-core-concepts/00100-databases/00500-cheat-sheet.md
```diff
@@ -720,6 +720,14 @@ SPACETIMEDB_VIEW(std::vector<Player>, top_players, Public, ViewContext ctx) {
     return ctx.db[player_score].filter(range_from(int32_t(1000))).collect();
 }
 
+// Perform a generic filter using the query builder.
+// Equivalent to `SELECT * FROM player WHERE score < 1000`.
+SPACETIMEDB_VIEW(Query<Player>, bottom_players, Public, ViewContext ctx) {
+    return ctx.from[player].where([](const auto& p) {
+        return p.score.lt(uint64_t(1000));
```
rekhoff: Player score is defined as an `int32_t` so the less-than should check against that data type.
```suggestion
        return p.score.lt(int32_t(1000));
```

rekhoff (commented):
Looks good to me. Some fairly minor nits here.

JasonAtClockwork:
Updated the core query builder to carry generated column and index-column metadata alongside each table source. This was needed so downstream codegen can reason about the source shape directly instead of reconstructing it indirectly. 

In practice, that lets the client generator produce source-aware query authoring APIs, expose the right fields and lookup surfaces, and support a much better Blueprint/C++ developer experience without maintaining a separate query-builder model.

crates/bindings-cpp/include/spacetimedb/query_builder/join.h
```diff
@@ -0,0 +1,358 @@
...
+        return compare(BoolExpr<TRow>::Kind::Gte, rhs);
+    }
+
+    template<typename TRhs>
+    [[nodiscard]] BoolExpr<TRow> Gte(const TRhs& rhs) const { return gte(rhs); }
+
+    template<typename TRhs>
+    [[nodiscard]] BoolExpr<TRow> lte(const TRhs& rhs) const {
+        return compare(BoolExpr<TRow>::Kind::Lte, rhs);
+    }
```
rekhoff, line 82: C# only has 2 literal methods (`Eq(SqlLiteral<T>)` and `Neq(SqlLiteral<T>)`), but we have 6 here, but `gt`/`lt`/`gte`/`lte`. I'm not actually sure any of these are needed in C++, as it might be better to use aa more Rust-like behavior and not use literal overloads on `IxCol`.

Anyway, this only matters because C# doesn't have template overload resolution in the same way. `Eq(SqlLiteral)` and `Eq(IxCol<TOtherRow, T>)` have completely distinct parameter types.

But here, I think we might have a scenario currently where we get some reverse ambiguity of sorts. When the general `eq(const TRhs& rhs)` template competes with the specialized `eq(const IxCol<TOtherRow, TValue>&)` during overload resolution, this can cause wrong-overload selection through generic forwarding where `ix.eq(ix)` in a WHERE context returns `IxJoinEq` instead of `BoolExpr`.

crates/bindings-cpp/include/spacetimedb/query_builder/join.h
```diff
@@ -0,0 +1,358 @@
...
+    template<typename TRhs>
+    [[nodiscard]] BoolExpr<TRow> Eq(const TRhs& rhs) const { return eq(rhs); }
+
+    template<typename TRhs>
+    [[nodiscard]] BoolExpr<TRow> ne(const TRhs& rhs) const {
+        return compare(BoolExpr<TRow>::Kind::Ne, rhs);
+    }
+
+    template<typename TRhs>
+    [[nodiscard]] BoolExpr<TRow> Neq(const TRhs& rhs) const { return ne(rhs); }
```
rekhoff, line 53: I believe the `Neq` comes from what we have in C#, and `ne` is following the snake_case convention. This feels a bit weird. I recommend we just go with `ne` and `Ne` since trying to use `neq` and `Neq` makes the snake_case form look weird.

crates/bindings-cpp/include/spacetimedb/query_builder/table.h
```diff
@@ -0,0 +1,304 @@
...
+    }
+
+    template<typename TFn>
+    [[nodiscard]] auto Where(TFn&& predicate) const {
+        if constexpr (std::is_invocable_v<TFn, const TCols&, const TIxCols&>) {
+            return where_ix(std::forward<TFn>(predicate));
+        } else {
+            return where(std::forward<TFn>(predicate));
+        }
+    }
```
rekhoff, line 140: For most of the templates, the snake_case and PascalCase version have the same behavior, but that's not the case with `Where`. It might be worth adding a comment providing some context or at least noting the distinction here.

rekhoff (commented):
Just a few items

crates/bindings-cpp/include/spacetimedb/query_builder/join.h
JasonAtClockwork: Thanks for calling this out! This is a bleed over from going back from the Unreal work to C++ that wasn't supposed to stick around. I've got it all cleared out now.

crates/bindings-cpp/include/spacetimedb/query_builder/join.h
JasonAtClockwork: Same as above this wasn't supposed to even bet a part of join and I've cleared it out.

crates/bindings-cpp/include/spacetimedb/query_builder/table.h
JasonAtClockwork: As I'm trying to support both the Unreal SDK + core C++ this was a too simple approach. I've expanded it so there is two distinct paths for the where and then both C++ where() and Where() (for Unreal) will is the constexpr to identify the correct call.

crates/bindings-cpp/include/spacetimedb/query_builder/expr.h
```diff
@@ -0,0 +1,361 @@
...
+        : column_(table_name, column_name) {}
+
+    template<typename TRhs>
+    [[nodiscard]] BoolExpr<TRow> eq(const TRhs& rhs) const { return compare(BoolExpr<TRow>::Kind::Eq, rhs); }
+    template<typename TRhs>
+    [[nodiscard]] BoolExpr<TRow> Eq(const TRhs& rhs) const { return eq(rhs); }
+    template<typename TRhs>
+    [[nodiscard]] BoolExpr<TRow> ne(const TRhs& rhs) const { return compare(BoolExpr<TRow>::Kind::Ne, rhs); }
+    template<typename TRhs>
+    [[nodiscard]] BoolExpr<TRow> Neq(const TRhs& rhs) const { return ne(rhs); }
```
rekhoff, line 323:
```suggestion
    [[nodiscard]] BoolExpr<TRow> Ne(const TRhs& rhs) const { return ne(rhs); }
```
Another instances of `Neq` -> `Ne` for consistency.

rekhoff (commented):
Just one minor nit, otherwise looks good

crates/bindings-cpp/tests/query-builder-compile/fail_event_lookup.cpp
```diff
@@ -0,0 +1,31 @@
...
+    Identity identity;
+};
+SPACETIMEDB_STRUCT(AuditEvent, identity)
+SPACETIMEDB_TABLE(AuditEvent, audit_event, Public, true)
+FIELD_PrimaryKey(audit_event, identity)
+
+auto invalid_event_lookup = TableFor<User>("user").right_semijoin(
+    TableFor<AuditEvent>("audit_event"),
+    [](const auto& users, const auto& events) {
+        return users.identity.eq(events.identity);
```
joshua-spacetime, line 30: What about comparing incompatible types? Can you add a negative test for that?

crates/bindings-cpp/include/spacetimedb/query_builder/expr.h
```diff
@@ -0,0 +1,361 @@
...
+    template<typename TRhs>
+    [[nodiscard]] BoolExpr<TRow> Lte(const TRhs& rhs) const { return lte(rhs); }
+
+    [[nodiscard]] constexpr const ColumnRef<TRow>& column_ref() const { return column_; }
+
+private:
+    template<typename TRhs>
+    [[nodiscard]] BoolExpr<TRow> compare(typename BoolExpr<TRow>::Kind kind, const TRhs& rhs) const {
+        return BoolExpr<TRow>::compare(kind, detail::to_operand<TRow>(*this), detail::to_operand<TRow>(rhs));
+    }
```
joshua-spacetime, line 519: Where do we constrain the left and right hand types?

crates/bindings-cpp/include/spacetimedb/query_builder/table.h
```diff
@@ -0,0 +1,318 @@
...
+    [[nodiscard]] auto where_ix(TFn&& predicate) const {
+        auto expr = detail::make_bool_expr<TRow>(std::forward<TFn>(predicate)(cols_, ix_cols_));
+        return FromWhere<TRow, TCols, TIxCols>(*this, std::move(expr));
+    }
+
+    // `where` is the ergonomic entry point: it dispatches to `where_col` or
+    // `where_ix` based on the predicate signature.
+    template<typename TFn>
+    [[nodiscard]] auto where(TFn&& predicate) const {
+        if constexpr (std::is_invocable_v<TFn, const TCols&, const TIxCols&>) {
```
joshua-spacetime: Why does the `where` predicate take both `TCols` and `TIxCols`? Index columns aren't relevant for `where` right?

crates/bindings-cpp/tests/query-builder-compile/fail_event_lookup.cpp
JasonAtClockwork: Added a negative test and doing so showed the path was brittle. I've cleaned up the indexed-column checks and added an explicit message to make it clear during compilation.

crates/bindings-cpp/include/spacetimedb/query_builder/expr.h
JasonAtClockwork: Looks like all my testing was positive only and this was a miss, it would fail after as it tries to run. I've reworked the comparison overloads to force the types to be compatible up front and expanded the overloads to get a clear error message during compilation.

This forced another overload for the column to column type check otherwise you'd fail with something like `user.id.eq(user.id)` so expanded to make sure that works as well.

This will block all narrowing for say user.id.eq(4000) won't work as that would be `int` and not `uint32_t` but adding it could allow something like 4200 into a `uint8_t` at compile time, or to keep that clean I might need to go down a path of a lot more overloads. I think more strict is best for a v1 and if we get comments we look into if we want to expand this later.

crates/bindings-cpp/include/spacetimedb/query_builder/table.h
JasonAtClockwork: Index columns aren’t required for where semantically. This split is there so chained queries don’t lose access to indexed-column predicates after earlier query steps. This isn't necessary for the pure C++ query builder, but I'm reusing this on the Unreal C++ side and want to keep them aligned which means a little extra API.

joshua-spacetime: Is there a way to keep the index cols internal and not part of the user facing api? To maintain consistency with the other query builders?

JasonAtClockwork: > Is there a way to keep the index cols internal and not part of the user facing api? To maintain consistency with the other query builders?

Yes and I should have done that! It's been cleaned up now so the where_ix/where_col are no longer in the public api.

joshua-spacetime: Apologies for the long feedback loop on this one @JasonAtClockwork, but it looks like this still may be part of the public api. Is it currently possible to write an expression like this:
```cpp
users.where([](const auto& cols, const auto& ix) {
    return ix.id.eq(cols.id);
});
```

It really should be:
```cpp
users.where([](const auto& cols) {
    return cols.id.eq(...);
});
```

Because indexed columns are only applicable/necessary for joins.

JasonAtClockwork: This one was a little more awkward as we still need it in the Unreal query builder, I've updated the check to use a new `#define` that is 0 for C++ and is a minor update for Unreal to flag it 1 turning this feature on.

joshua-spacetime (approved):

PR #4671, opened 2026-03-19
Allow obtaining a `AnonymousViewContext` from a `ViewContext`

kistz:
# Description of Changes

Title.
Is useful to more easily reuse code.
I asked @joshua-spacetime on discord if this would be reasonable and he said 👍 .
This mirrors the method available on `ReducerContext` for consistency.

None

1 - minimal api surface addition

# Testing

- [x] Tested in my project and works as expected.


crates/bindings/src/lib.rs
```diff
@@ -950,6 +937,11 @@ impl ViewContext {
     pub fn sender(&self) -> Identity {
         self.sender
     }
+
+    /// Obtain an [`AnonymousViewContext`] for easier code reuse.
+    pub fn as_anonymous_read_only(&self) -> AnonymousViewContext {
```
joshua-spacetime:
```suggestion
    /// Obtain an [`AnonymousViewContext`] by dropping `sender`.
    pub fn as_anonymous(&self) -> AnonymousViewContext {
```

PR #4675, opened 2026-03-19
Add `ConsumeEntityEvent` to Blackholio C++ and C# modules

rekhoff:
This resolves an issue found while testing C# module behavior against the Blackholio demo in an unrelated PR.

# Description of Changes

Updates the C# and C++ server modules for the Blackholio demo, to reflect changes made to the Rust module in #4217
This is needed if testing against Blackholio using either the C# or C++, or else the `ConsumeEntityEvent` will be missing from the generated client bindings.

Not API or ABI breaking.

1

# Testing

- [X] Tested Unity Blackholio demo using the C# module
- [X] Tested Unity Blackholio demo using the C++ module


JasonAtClockwork (approved):
LGTM, I've run both changes against the Unreal client as a test as well.

PR #4677, opened 2026-03-20
core: Keep a reordering window in durability worker

kim:
`RelationalDB::commit_tx` and `RelationalDB::commit_tx_downgrade` release the exclusive transaction lock before calling `DurabilityWorker::request_durability`. This can lead to transactions appearing out-of-order on the worker queue, if transactions on the same database instance are committed from multiple threads.

To mitigate this, the worker now keeps a small min-heap to re-order transactions in that case. We expect the ordering issue to happen very rarely, so the overhead of this should be negligible.

Supersedes #4661, which addressed the issue by holding the transaction lock until durability request submission is completed.

2

# Testing

- [ ] Unit tests for the min-heap implementation
- [ ] Ran the keynote benchmark with various parameters 
      that would trigger the error without this patch

### DIFF
--- a/crates/core/src/db/durability.rs
+++ b/crates/core/src/db/durability.rs
@@ -1,7 +1,8 @@
-use std::{sync::Arc, time::Duration};
+use std::{cmp::Reverse, collections::BinaryHeap, iter, num::NonZeroUsize, sync::Arc, time::Duration};
 
 use futures::TryFutureExt as _;
 use log::{error, info};
+use prometheus::IntGauge;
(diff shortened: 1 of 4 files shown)


bfops:
(this was failed in the merge queue due to flakes in #4183 - re-adding)

bfops:
...and then failed again for an unrelated reason, neat.

bfops:
and then we had the github bug where a completely mergeable PR does not get added to the merge queue for.. reasons..?

PR #4678, opened 2026-03-20
Keynote-2 sqlite fixes

bradleyshep:
# Description of Changes

Fix the SQLite RPC benchmark so transfers actually persist and verification produces useful output.

**`sqlite-rpc-server.ts`:**
- Add `.run()` to both Drizzle `tx.update()` chains in `rpcTransfer`. Without this, the UPDATE statements were never executed — the benchmark was only measuring SELECT + HTTP overhead, not real transactional writes. This inflated SQLite TPS numbers significantly.
- Replace the generic `"internal error"` catch-all in `handleRpc` with `rpcErr()`, which returns the actual error message to the client. Also wrap the outer HTTP handler in a try/catch so errors outside `handleRpc` are surfaced too.

**`sqlite_rpc.ts` (connector):**
- Detect `{ skipped: true }` from the server's verify endpoint and throw a clear error telling the user to set `SEED_INITIAL_BALANCE` on the RPC server process. Previously this was silently treated as success.

**`runner.ts` / `runner_1.ts`:**
- Log "Verification passed" on success and "Verification failed: \<reason\>" (as a string, not a raw Error object) on failure, so the outcome is always visible in bench output.

None.

1 

# Testing

- [ ] Run `npx tsx src/rpc-servers/sqlite-rpc-server.ts` (with `SEED_INITIAL_BALANCE` set), then `npm run test-1 -- --connectors sqlite_rpc --seconds 5` with `VERIFY=1`. Confirm TPS drops significantly vs the old (broken) numbers and "Verification passed" appears.
- [ ] Run the same without `SEED_INITIAL_BALANCE` on the server process. Confirm "Verification failed" with a message about the missing env var (not "internal error").

### DIFF
--- a/templates/keynote-2/src/connectors/rpc/sqlite_rpc.ts
+++ b/templates/keynote-2/src/connectors/rpc/sqlite_rpc.ts
@@ -1,4 +1,4 @@
-﻿import type { RpcConnector } from '../../core/connectors.ts';
+import type { RpcConnector } from '../../core/connectors.ts';
 import { RpcRequest, RpcResponse } from './rpc_common.ts';
 
 export default function sqlite_rpc(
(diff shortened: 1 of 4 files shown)


joshua-spacetime:
Have you checked if there similar errors in the other benchmarks?

joshua-spacetime (approved):

PR #4692, opened 2026-03-23
feat(procedures): add streaming HTTP response support

philtrem:
# Description of Changes

Adds `fetchStreaming()` to the TypeScript procedure HTTP client, returning a `StreamingResponse` that yields body chunks via synchronous iteration. This enables procedures to process large HTTP responses without buffering the entire body in memory.

**Depends on #4691** (headers fix is the first commit in this branch).

Architecture:
- Background tokio task reads chunks from the HTTP response body into a bounded mpsc channel (capacity 8)
- The V8 thread reads synchronously one chunk at a time via `procedure_http_stream_next`
- Extracts shared `prepare_http_request()` to reuse IP filtering, DNS filtering, timeout clamping, and redirect policy between `fetch()` and `fetchStreaming()`
- Transaction guard on `stream_next` prevents blocking inside `withTx`
- Resource cleanup via `FinalizationRegistry` backup + explicit `Symbol.dispose()` + `HttpStreamState::drop` aborts background task
- Follows the established `ResourceSlab` / handle pattern used by row iterators

None. Adds new `fetchStreaming()` method to `HttpClient` and 3 new V8 syscalls (`procedure_http_stream_open`, `procedure_http_stream_next`, `procedure_http_stream_close`). Existing `fetch()` API is unchanged.

3 — Introduces a new streaming code path through the syscall layer, instance environment, and TypeScript bindings. The main interaction concern is blocking the V8 thread per chunk, which stalls all other reducers/procedures for the module instance while waiting. This is documented and is a necessary trade-off until async procedures are supported. The Rust-side changes reuse existing infrastructure (`ResourceSlab`, `prepare_http_request`, metrics) to minimize new surface area.

# Testing

- [x] Unit test: BSATN header round-trip
- [x] Unit test: `http_request` blocked during transaction
- [x] Unit test: `http_stream_open` blocked during transaction
- [x] Unit test: `HttpStreamState::drop` aborts background task
- [x] Smoketest: basic streaming read (chunked encoding, chunks concatenated correctly)
- [x] Smoketest: `stream.next()` inside transaction throws `WouldBlockTransaction`
- [x] Smoketest: response headers preserved through streaming path
- [ ] Reviewer: verify streaming with a slow/large response source if desired

### DIFF
--- a/crates/bindings-typescript/src/server/http_internal.ts
+++ b/crates/bindings-typescript/src/server/http_internal.ts
@@ -286,8 +286,8 @@ function fetchStreaming(url: URL | string, init: RequestOptions = {}): Streaming
     serializeRequest(request),
     body
   );
-  const response = HttpResponse.deserialize(new BinaryReader(responseBuf));
   const stream = new StreamHandle(handle);
+  const response = HttpResponse.deserialize(new BinaryReader(responseBuf));
   const code = response.code;
   const responseHeaders = deserializeHeaders(response.headers);
 
--- a/crates/bindings-typescript/src/server/http_internal.ts
+++ b/crates/bindings-typescript/src/server/http_internal.ts
@@ -138,8 +138,43 @@ export interface RequestOptions {
   // redirect?: RequestRedirect;
 }
 
+/**
+ * A streaming HTTP response that yields body chunks via iteration.
+ *
+ * **Important:** Each iteration blocks the module's V8 thread until the next
+ * chunk arrives. Because there is one V8 thread per module instance, all other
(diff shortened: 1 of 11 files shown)


bfops:
Hey @philtrem, sorry for the delay. A change like this might have wider implications for the design of SpacetimeDB, so we want to circle back with the team and think about how this will fit with other changes as we start to eye bigger  changes for 3.0.

philtrem:
@bfops fair enough. you guys could take a look at if it can be useful as a starting point: https://github.com/clockworklabs/SpacetimeDB/issues/4697

PR #4693, opened 2026-03-23
Make Timestamp a FilterableValue

### LINKED ISSUE #2650: Make `Timestamp` implement `FilterableValue`

rafaelGuerreiro:
# Description of Changes
Make Timestamp a FilterableValue in Rust, C#, and Typescript. I'm not sure this is changing all the places because we have the server and the client in those 3 languages.

It's an additive change.

3. There are some designs decisions, like comparing timestamp to strings/numbers.

# Testing
Added unit tests for the 3 languages.


crates/bindings-csharp/BSATN.Runtime/QueryBuilder.cs
```diff
@@ -744,6 +747,42 @@ public static BoolExpr<TRow> Eq<TRow>(this Col<TRow, Uuid> col, Uuid value) =>
...
+    public static BoolExpr<TRow> Lte<TRow>(this Col<TRow, Timestamp> col, Timestamp value) =>
+        col.Lte(SqlLit.Timestamp(value));
+
+    public static BoolExpr<TRow> Gt<TRow>(this Col<TRow, Timestamp> col, Timestamp value) =>
+        col.Gt(SqlLit.Timestamp(value));
+
+    public static BoolExpr<TRow> Gte<TRow>(this Col<TRow, Timestamp> col, Timestamp value) =>
+        col.Gte(SqlLit.Timestamp(value));
+
+    public static BoolExpr<TRow> Eq<TRow>(this Col<TRow, Timestamp> col, long microsSinceUnixEpoch) =>
```
rafaelGuerreiro: Ensure `long microsSinceUnixEpoch` design is correct here.

gefjon: What is this method (and the similar ones which accept a `long`) for? I would expect users to use the methods which accept a `Timestamp` for typed comparisons. I'm a little uncomfortable with implicit conversions like this. If there's a good reason for it I'm not seeing, though, then I wouldn't want to remove it without knowing.

crates/bindings-csharp/BSATN.Runtime/QueryBuilder.cs
```diff
@@ -873,4 +924,7 @@ public static string FormatHexLiteral(string hex)
 
         return $"0x{s}";
     }
+
+    public static string FormatTimestampLiteral(Timestamp timestamp) =>
+        FormatStringLiteral(timestamp.ToString());
```
gefjon: Have you tested against a real SpacetimeDB host that this formatting is correct?

crates/bindings-typescript/src/lib/filter.ts
```diff
@@ -77,6 +78,24 @@ export function evaluate<Column extends string>(
...
+          if (typeof expr.value === 'string') {
+            return v.toISOString() === expr.value;
+          }
+          // Value of the Column is a Timestamp and passed Value is a Number microsecond count.
+          if (typeof expr.value === 'number') {
+            return (
+              Number.isInteger(expr.value) &&
+              BigInt(expr.value) === v.microsSinceUnixEpoch
+            );
+          }
```
gefjon: Again, I'm somewhat surprised and confused by the implicit conversions here. It's especially surprising to me that in TypeScript you're accepting a different set of implicit conversions than in C# (i.e. accepting strings here, but not in C#).

crates/bindings-typescript/src/lib/filter.ts
```diff
@@ -103,6 +122,9 @@ function formatValue(v: Value): string {
       if (v instanceof Uuid) {
         return `'${v.toString()}'`;
       }
+      if (v instanceof Timestamp) {
+        return `'${v.toISOString()}'`;
```
gefjon, line 115: Same question about having tested this formatting.

gefjon (commented):
This looks great, thanks for all your hard work!

crates/bindings-csharp/BSATN.Runtime/QueryBuilder.cs
rafaelGuerreiro: I wasn't sure I should include these or not. But I do time calculations all the time, and using long is just easier in many languages. In Rust it's easier to use Timestamp, but I'm not sure about C# and TS.

crates/bindings-csharp/BSATN.Runtime/QueryBuilder.cs
rafaelGuerreiro: I haven't. I don't know how to do that. But I remember checking the format. It's been a while, I don't remember if I did a unit test for this case.

crates/bindings-typescript/src/lib/filter.ts
rafaelGuerreiro: Strings are even worse than long in this case, but in js it's a bit harder to control the types, which is why I opted to allow for Strings.

rafaelGuerreiro:
@gefjon If you prefer, I can keep it compatible with Timestamp only, and the user needs to convert before using in the query. On the plus side, it's a lot easier to add these comparisons later than it is to remove them.

crates/bindings-csharp/BSATN.Runtime/QueryBuilder.cs
gefjon: You'd do that either by building a C# client which calls this method to build a subscription, or by manually invoking this method and pasting its output into a query you pass to `spacetime subscribe` or `spacetime sql`.

gefjon:
> @gefjon If you prefer, I can keep it compatible with Timestamp only, and the user needs to convert before using in the query. On the plus side, it's a lot easier to add these comparisons later than it is to remove them.

Yeah, I think that'd be better. Thanks!

crates/bindings-csharp/BSATN.Runtime/QueryBuilder.cs
rafaelGuerreiro:
```
  $ cargo build -p spacetimedb-cli
  Finished `dev` profile [unoptimized + debuginfo] target(s) in 37.00s

  $ target/debug/spacetimedb-cli --version
  spacetime Path: /Users/rguerreiro/workspace/oss/SpacetimeDB/target/debug/spacetimedb-cli
  Commit: 64b8e39430368ab6130dc9e4c0ab96574cb6decf
  spacetimedb tool version 2.1.0; spacetimedb-lib version 2.1.0;

  $ cargo build -p spacetimedb-standalone
  Finished `dev` profile [unoptimized + debuginfo] target(s) in 50.15s

  $ target/debug/spacetimedb-cli start --listen-addr 127.0.0.1:3000 --in-memory --data-dir /tmp/stdb-timestamp-host --non-interactive
  spacetimedb-standalone version: 2.1.0
  spacetimedb-standalone path: /Users/rguerreiro/workspace/oss/SpacetimeDB/target/debug/spacetimedb-standalone
  database running in data directory /tmp/stdb-timestamp-host
  2026-04-28T17:33:32.839079Z  INFO ... Starting SpacetimeDB listening on 127.0.0.1:3000
  2026-04-28T17:33:32.839290Z  WARN ... PostgreSQL wire protocol server disabled

  In another shell:

  $ target/debug/spacetimedb-cli publish timestamp-format-test --server http://127.0.0.1:3000 --module-path modules/sdk-test --delete-data=always --yes --anonymous
  Finished `release` profile [optimized] target(s) in 14.24s
  Build finished successfully.
  Uploading to http://127.0.0.1:3000 => http://127.0.0.1:3000
  This will DESTROY the current timestamp-format-test module, and ALL corresponding data.
  Skipping confirmation due to --yes
  Publishing module...
  Created new database with name: timestamp-format-test, identity: c200f4474f102490343879baca7b59d3a042c8a4b1bc2abd7efd869c3e1b3f94

  $ target/debug/spacetimedb-cli call timestamp-format-test insert_call_timestamp --server http://127.0.0.1:3000 --anonymous
  WARNING: This command is UNSTABLE and subject to breaking changes.

  $ target/debug/spacetimedb-cli sql timestamp-format-test "SELECT * FROM one_timestamp" --server http://127.0.0.1:3000 --anonymous
  WARNING: This command is UNSTABLE and subject to breaking changes.

   t
  ----------------------------------
   2026-04-28T17:34:33.897493+00:00

  $ target/debug/spacetimedb-cli sql timestamp-format-test "SELECT * FROM one_timestamp WHERE t = '2026-04-28T17:34:33.897493Z'" --server http://127.0.0.1:3000 --anonymous
  WARNING: This command is UNSTABLE and subject to breaking changes.

   t
  ----------------------------------
   2026-04-28T17:34:33.897493+00:00

  $ target/debug/spacetimedb-cli sql timestamp-format-test "SELECT * FROM one_timestamp WHERE t = '2026-04-28T17:34:33.897493+00:00'" --server http://127.0.0.1:3000 --anonymous
  WARNING: This command is UNSTABLE and subject to breaking changes.

   t
  ----------------------------------
   2026-04-28T17:34:33.897493+00:00

  Server shutdown:

  ^C
  2026-04-28T17:35:56.061767Z  INFO ... Shutting down server...
  ```

rafaelGuerreiro:
@gefjon I've done all the changes, fixed the review from Copilot, and provided a test run log in the other comment.

gefjon (approved):
Wonderful, thank you!

rafaelGuerreiro:
Sorry, @gefjon, would you approve it once again? I fix the C# formatting.

PR #4704, opened 2026-03-24
Replace JS worker's rendezvous channel with unbounded queue

joshua-spacetime:
# Description of Changes

Previously, a module’s JS worker thread was fed through a zero-capacity channel. That made every request handoff a rendezvous between the async producer task and the single JS worker thread. Under high concurrency, that synchronous handoff showed up as hot `flume`/lock/wakeup stacks on the critical path - the JS worker thread.

This patch brings V8 execution in line with WASM which also uses an unbounded request queue.

Changing that handoff to an unbounded queue decouples request producers from the JS worker. Producers can enqueue work without synchronizing directly with the worker on every request, and the worker can continue draining queued requests without paying the rendezvous cost each time. This shortens the critical path, reduces scheduler/locking overhead and increases throughput.

None

2

# Testing

Manual performance testing

### DIFF
--- a/crates/core/src/host/v8/mod.rs
+++ b/crates/core/src/host/v8/mod.rs
@@ -266,12 +266,7 @@ impl JsModule {
         self.common.info().clone()
     }
 
-    pub async fn create_instance(&self) -> JsInstance {
-        // We use a rendezvous channel for pooled instances, because they are checked
-        // out one request at a time and subsequently returned to the pool, unlike the
-        // long lived instance used for executing reducers which isn't checked out but
-        // fed through a queue.
-        let request_queue = JsWorkerQueue::bounded(0);
+    async fn create_instance_with_queue(&self, request_queue: Arc<JsWorkerQueue>) -> JsInstance {
         let program = self.program.clone();
         let common = self.common.clone();
         let load_balance_guard = self.load_balance_guard.clone();
@@ -292,26 +287,15 @@ impl JsModule {
         instance
     }
 
-    async fn create_lane_instance(&self) -> JsInstance {
-        let program = self.program.clone();
-        let common = self.common.clone();
-        let load_balance_guard = self.load_balance_guard.clone();
-        let core_pinner = self.core_pinner.clone();
-        let heap_policy = self.heap_policy;
-        let request_queue = self.lane_queue.clone();
+    pub async fn create_instance(&self) -> JsInstance {
+        // We use a rendezvous channel for pooled instances, because they are checked
+        // out one request at a time and subsequently returned to the pool, unlike the
+        // long lived instance used for executing reducers.
+        self.create_instance_with_queue(JsWorkerQueue::bounded(0)).await
+    }
 
-        // This has to be done in a blocking context because of `blocking_recv`.
-        let (_, instance) = spawn_instance_worker(
-            program,
-            Either::Left(common),
-            load_balance_guard,
-            core_pinner,
-            heap_policy,
-            request_queue,
-        )
-        .await
-        .expect("`spawn_instance_worker` should succeed when passed `ModuleCommon`");
-        instance
+    async fn create_lane_instance(&self) -> JsInstance {
+        self.create_instance_with_queue(self.lane_queue.clone()).await
     }
 }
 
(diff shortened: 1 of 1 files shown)


crates/core/src/host/v8/mod.rs
```diff
@@ -387,8 +424,112 @@ pub struct JsInstance {
...
+}
+
+/// Lifecycle state for a single JS worker.
+///
+/// The `trapped` bit tells the lane that this worker must not keep serving
+/// future work. The `exited` bit and [`Notify`] are what make the [JsWorkerQueue]
+/// remain single-consumer. Recovery waits on [`Self::wait_exited`] before spawning
+/// the replacement worker, so there is no overlap where two workers can both drain
+/// the queue.
+struct JsWorkerState {
```
Centril, line 499:
```suggestion
#[derive(Default)]
struct JsWorkerState {
```

crates/core/src/host/v8/mod.rs
```diff
@@ -387,8 +424,112 @@ pub struct JsInstance {
...
+}
+
+impl JsWorkerState {
+    fn new() -> Arc<Self> {
+        Arc::new(Self {
+            trapped: AtomicBool::new(false),
+            exited: AtomicBool::new(false),
+            exited_notify: Notify::new(),
+        })
+    }
```
Centril: All of these are `Default`.


```suggestion
```

crates/core/src/host/v8/mod.rs
```diff
@@ -1037,21 +1200,22 @@ async fn spawn_instance_worker(
...
-    // Multiple callers can wait to hand work to the worker, but with
-    // `bounded(0)` there is no buffered backlog inside the channel itself.
-    // The worker still processes requests strictly one at a time.
-    let (request_tx, request_rx) = flume::bounded(0);
-
     // This one-shot channel is used for initial startup error handling within the thread.
     let (result_tx, result_rx) = oneshot::channel();
-    let trapped = Arc::new(AtomicBool::new(false));
-    let worker_trapped = trapped.clone();
+    let worker_state = JsWorkerState::new();
```
```suggestion
    let worker_state: Arc<JsWorkerState> = <_>::default();
```

crates/core/src/worker_metrics/mod.rs
```diff
@@ -312,6 +312,11 @@ metrics_group!(
         #[labels(database_identity: Identity)]
         pub v8_heap_size_limit_bytes: IntGaugeVec,
 
+        #[name = spacetime_worker_v8_instance_lane_queue_length]
+        #[help = "The number of queued requests waiting for a database's JS instance lane worker"]
+        #[labels(database_identity: Identity)]
+        pub v8_instance_lane_queue_length: IntGaugeVec,
```
Centril, line 318: Now that you've discovered the issue and resolved it, I think we can remove these metrics as they mostly add (a little bit of) cost and complexity.

crates/core/src/host/v8/mod.rs
Centril: It would have made reviewing easier and the diff smaller if this code was not duplicated. Please dedup these into a common base method taking `request_queue`.

crates/core/src/host/v8/mod.rs
```diff
@@ -387,8 +424,112 @@ pub struct JsInstance {
...
+    fn unbounded_with_metric(metric: IntGauge) -> Arc<Self> {
+        let (tx, rx) = flume::unbounded();
+        Arc::new(Self {
+            tx,
+            rx,
+            metric: Some(metric),
+        })
+    }
+
+    async fn send_async(&self, request: JsWorkerRequest) -> Result<(), flume::SendError<JsWorkerRequest>> {
```
Centril, line 468: Please add doc comments to these, especially the two above re. how they are used.

crates/core/src/host/v8/mod.rs
```diff
@@ -54,7 +54,7 @@ use std::panic::AssertUnwindSafe;
 use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
```
Centril, line 54: This module is growing to large and dealing with too many concerns. We should split it, but I can do that after this PR.

crates/core/src/host/v8/mod.rs
```diff
@@ -387,8 +424,112 @@ pub struct JsInstance {
...
+///
+/// Note, this queue outlives any single JS worker or V8 isolate.
+/// When the active worker traps or is retired for heap growth,
+/// [`JsInstanceLane::replace_active_if_current`] waits for it to fully exit
+/// and spawns a fresh worker which then starts reading from the queue so that
+/// buffered requests are not dropped.
+#[derive(Clone)]
+struct JsWorkerQueue {
+    tx: flume::Sender<JsWorkerRequest>,
+    rx: flume::Receiver<JsWorkerRequest>,
```
Centril, line 443: Initially, it looks rather strange to see `tx` and `rx` together like this, so it could use some doc commentary.

crates/core/src/host/v8/mod.rs
```diff
@@ -387,8 +424,112 @@ pub struct JsInstance {
...
-    trapped: Arc<AtomicBool>,
+    request_queue: Arc<JsWorkerQueue>,
+    worker_state: Arc<JsWorkerState>,
+}
+
+/// Shared request queue for a JS instance lane.
+///
+/// Async callers enqueue [`JsWorkerRequest`] values here and wait on their
+/// per-request one-shot replies. The dedicated JS worker thread drains this
+/// queue and executes those requests on the isolate.
```
```suggestion
/// Async callers enqueue [`JsWorkerRequest`] values here
/// and wait on their per-request one-shot replies sent back via [`JsReplyTx`].
/// The dedicated JS worker thread, spawned in [`spawn_instance_worker`]
/// drains this queue and executes those requests on the isolate.
```

crates/core/src/host/v8/mod.rs
```diff
@@ -387,8 +424,112 @@ pub struct JsInstance {
...
+    fn mark_trapped(&self) {
+        self.trapped.store(true, Ordering::Relaxed);
+    }
+
+    fn mark_exited(&self) {
+        self.exited.store(true, Ordering::Relaxed);
+        self.exited_notify.notify_waiters();
+    }
+
+    async fn wait_exited(&self) {
```
Centril, line 537: Please add doc comments to these.

crates/core/src/host/v8/mod.rs
```diff
@@ -387,8 +424,112 @@ pub struct JsInstance {
     /// it to tell whether the currently active worker has already been replaced
     /// after a trap or disconnect.
     id: u64,
-    request_tx: flume::Sender<JsWorkerRequest>,
-    trapped: Arc<AtomicBool>,
+    request_queue: Arc<JsWorkerQueue>,
```
Centril, line 421: Why is the `Arc` necessary here? The inner type is `Clone` and consists of cheaply cloneable types where each is already wrapped in `Arc`.

crates/core/src/host/v8/mod.rs
```diff
@@ -791,15 +950,19 @@ impl JsInstanceLane {
...
         work: impl AsyncFnOnce(JsInstance) -> Result<R, WorkerDisconnected>,
     ) -> Result<R, WorkerDisconnected> {
         assert_not_on_js_module_thread(label);
 
-        let active = self.active_instance();
+        let mut active = self.active_instance();
+        if active.needs_recovery() {
+            self.replace_active_if_current(&active).await;
+            active = self.active_instance();
+        }
```
Centril, line 968: Can we enter this method with a bad instance now? Some commentary on this would be good in the code.

crates/core/src/host/v8/mod.rs
```diff
@@ -1147,13 +1314,17 @@ async fn spawn_instance_worker(
                 }
                 JsWorkerRequest::CallReducer { reply_tx, params } => {
                     let (res, trapped) = call_reducer(None, params);
-                    worker_trapped.store(trapped, Ordering::Relaxed);
+                    if trapped {
+                        worker_state_in_thread.mark_trapped();
+                    }
```
Centril, line 1322: One thing that would be good to comment on is why we need to do `worker_state_in_thread.mark_trapped();` when we already do that in `send_request`. Ostensibly, we are setting the flag twice now.

Centril:
On phoenix nap, using the rust client, I get:

- TS module: 156.5k TPS
- Rust module: 168.5k TPS 

The difference with this PR is at 12k TPS 🚀

coolreader18:
I'll review after the changes requested by Mazdak are made

crates/core/src/worker_metrics/mod.rs
joshua-spacetime: These are not debug metrics. They're observability metrics, and the queue is now unbounded so we should definitely keep them since we've lost a mechanism for providing backpressure.

crates/core/src/host/v8/mod.rs
joshua-spacetime: Yes, `run_once` can enter with an already stale active handle if another caller has already observed a trap/disconnect, or heap retirement has marked the worker for replacement, before the swap finishes. It should not execute work on that worker though: we check `active.needs_recovery()` up front and replace before enqueueing any new request. I added a comment.

crates/core/src/host/v8/mod.rs
joshua-spacetime: The `Arc` is intentional here because `JsWorkerQueue` owns the `rx` handle that gets cloned into each worker during replacement, and cloning that wrapper by value would clone the `Receiver` too, which changes flume’s receiver count. It would also run `JsWorkerQueue::drop` once per `JsInstance` clone, which would over-adjust the queue-length metric. So the outer `Arc` is there to keep one queue wrapper identity shared across all `JsInstance` clones, and to make its `Drop` run exactly once.

PR #4707, opened 2026-03-25
[Rust] Add possibility to be generic over `DbContext`

kistz:
# Description of Changes

This patch aims to improve the situation for the need of a generic function over multiple `ctxs`.
The `DbContext` was introduced for this but due to the associated type amibguity when trying to implement it another mmethod is needed.

This pr implements a `db_read_only()` method which always results in a `LocalReadOnly` hence Rust can infer the types of a the returned type (which prior to this was either `__view , __query and a 3rd one i cent remember right now 😓`.

I have chosen to be defensive by implementing it as `unstable` but i can also remove this if that is desired.

It allows the following pattern which has been workig great in my project so its time to contribute it :> 

```rust
pub(crate) trait YourTableNameRead {
   // Your methods which only need read access
  fn test(&self,args);

}

// The read version is a supertrait to give access to the read methods.
pub(crate) trait YourTableNameWrite: YourTableNameRead {
  // Your methods which need read-write access
}

// The read version gets implemented for every DbContext since we can always read.
impl<Db: DbContext> YourTableNameRead for Db {
  fn test(&self,args) {
   self.db_read_only().table_name().whatever(args);
}

// By constraining the associated type to Local we only get this for writeable ctxs.
impl<Db: DbContext<DbView = Local>> YourTableNameWrite for Db {}

```

These allow you to do on the calling site:
```rust
use YourTableNameRead;

#[view|reducer|procedure]
fn my_func(ctx: Reducer|(Anon)View|Tx, args ) {
   ctx.test(args);
}
```

None

1. Minor api and qol adition which is furthermore unstable.

# Testing

- [x] Works in my project

### DIFF
--- a/crates/bindings/src/lib.rs
+++ b/crates/bindings/src/lib.rs
@@ -1457,16 +1457,12 @@ pub trait DbContext {
     ///
     /// This method is provided for times when a programmer wants to be generic over the `DbContext` type.
     /// Concrete-typed code is expected to read the `.db` field off the particular `DbContext` implementor.
-    /// Currently, being this generic is only meaningful in clients,
-    /// as `ReducerContext` is the only implementor of `DbContext` within modules.
     fn db(&self) -> &Self::DbView;
 
     /// Get a read-only view into the tables.
     ///
     /// This method is provided for times when a programmer wants to be generic over the `DbContext` type.
     /// Concrete-typed code is expected to read the `.db` field off the particular `DbContext` implementor.
-    /// Currently, being this generic is only meaningful in clients,
-    /// as `ReducerContext` is the only implementor of `DbContext` within modules.
     #[cfg(feature = "unstable")]
     fn db_read_only(&self) -> &LocalReadOnly;
 }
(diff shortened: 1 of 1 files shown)


crates/bindings/src/lib.rs
```diff
@@ -1460,6 +1460,15 @@ pub trait DbContext {
     /// Currently, being this generic is only meaningful in clients,
     /// as `ReducerContext` is the only implementor of `DbContext` within modules.
     fn db(&self) -> &Self::DbView;
+
+    /// Get a read-only view into the tables.
+    ///
+    /// This method is provided for times when a programmer wants to be generic over the `DbContext` type.
+    /// Concrete-typed code is expected to read the `.db` field off the particular `DbContext` implementor.
+    /// Currently, being this generic is only meaningful in clients,
+    /// as `ReducerContext` is the only implementor of `DbContext` within modules.
```
gefjon: Oh, this isn't true any more. Could you remove this and the similar comment on `fn db`?

gefjon (commented):
I'm happy with this. One quick note about a comment, and then I'll happily approve.

crates/bindings/src/lib.rs
kistz: makes sense did not turn on my brain for that one ^^

kistz:
@gefjon thanks! should be good now :>

gefjon (approved):

PR #4708, opened 2026-03-25
Compression stats and commitlog compression function

kim:
Adds a way to pass a "closure" when compressing a commitlog segment, such that the source and destination are polymorphic. This allows to have them wrapped externally, which is used to apply bandwidth limiting where necessary.

Also return bytes in + bytes out stats from the compressor.

1

# Testing

Elsewhere.


cloutiertyler (approved):
No objection to this.

PR #4716, opened 2026-03-27
IDC: Asynchronous reducer.

Shubham8287:
## Async Reducer Execution (Outbox/Inbox)

Adds a message-driven execution model for reducers with idempotency and transactional guarantees.

Maybe start from looking at [smoketest](https://github.com/clockworklabs/SpacetimeDB/pull/4716/changes#diff-1df05594543ec28cc2554bd6b854c6ae546855871d0785bdb639d99f8381a08f)

### Outbox Macro

`outbox` macros mark tables that emit reducer calls.

- Limited implementation for testing purpose.
- Single `on_result` callback per table (not per row, this difffers from proposal) - [discussion](https://github.com/clockworklabs/SpacetimeDBPrivate/pull/2755#discussion_r3050247528)
- It should merge behing `unstable` feature flag.

#### `ST_OUTBOUND_MSG`

- Maintains an auto-incrementing `msg_id` per sender.
- Stores in-flight outbound reducer calls (outbox pattern).
- Used by `IdcActor` to issue remote calls with strictly increasing `msg_id`.

#### `ST_INBOUND_MSG`

- Tracks latest processed `msg_id` per sender along with reducer result
- For duplicate requests where the `msg_id` is already present, the result is returned without executing the reducer.
- Requests with a `msg_id` older than the latest seen for a sender are dropped

### Actor

`IDCActor` handles execution.

- Drives reducer execution from `ST_OUTBOUND_MSG`
- Calls `on_result` reducer and delete entry from  `ST_OUTBOUND_MSG` in same transaction.
- If reducer fails, delete entry in next transaction, somewhat similar to `SchedulerActror`
- Reducer failure are only considered if reducer actually ran and returned error, all other cases of error will result in retrying.

### API

`call_from_database(database_identity, msg_id, params)`

- Caller should ensure `msg_id` never decrease for same sender `database_identity` else API returns `InvalidRequest`. 
- If reducer ran and failed it returns `StatusCode::UNPROCESSABLE_ENTITY`, which caller shouln't retry. 
- All other errors, including network errors are retried by `IdcActor`.

- New APIs only, but we may not want to make macro implementation public, so it has to be behind `unstable` feature flag.

4/5, Exactly once semantic require closer look and can be tricky.

# Testing

New smoketests.

### DIFF
--- a/crates/client-api/src/routes/database.rs
+++ b/crates/client-api/src/routes/database.rs
@@ -319,6 +319,8 @@ pub async fn call_from_database<S: ControlStateDelegate + NodeDelegate>(
                 ),
                 // 422 = reducer ran but returned Err; the IDC actor uses this to distinguish
                 // reducer failures from other  errors (which it retries).
+                // This is inconsistent with `call` endpoint, which returns 523 status code if
+                // reducer fails
                 ReducerOutcome::Failed(errmsg) => (
                     StatusCode::UNPROCESSABLE_ENTITY,
                     axum::body::Body::from(errmsg.to_string()),


Shubham8287:
Closing in favor of https://github.com/clockworklabs/SpacetimeDB/pull/4780

PR #4721, opened 2026-03-28
Add enabled option to useTable hook

DexterKoelson:
# Description of Changes

This adds an `enabled` option to the `useTable` hook in the React SDK, allowing devs to conditionally disable a subscription without having to wrap it in a component that is conditionally mounted.

With `enabled`, you can control subscription lifecycle as such:

```tsx
const [rows, isReady] = useTable(tables.messages, { enabled: isChatOpen });
```

This is a similar pattern in other data hooks (React Query's `useQuery`, SWR, Apollo's `useSubscription`, etc).

When `enabled` is `false`:
- `computeSnapshot` returns `[[], true]` immediately (no data, ready state)
- The subscription effect skips setup and resets `subscribeApplied`
- The event listener callback returns a no-op cleanup

When `enabled` flips back to `true`, the subscription is re-established automatically via the dependency arrays.

None. The `enabled` field is optional and defaults to `true`, so existing usage is unaffected.

1 — Single file change, additive only. The `enabled` flag gates existing behavior behind early returns and is wired into the existing dependency arrays. No interaction with other components.

# Testing I Did

- [x] Verify `useTable(tables.foo)` works as before (no `enabled` option passed)
- [x] Verify `useTable(tables.foo, { enabled: false })` returns `[[], true]` and does not subscribe
- [x] Verify toggling `enabled` from `false` to `true` establishes the subscription and returns rows
- [x] Verify toggling `enabled` from `true` to `false` clears rows and unsubscribes


JasonAtClockwork (approved):
Thanks @DexterKoelson for this addition! It looks solid and my tests show it working well.

PR #4724, opened 2026-03-29
Pipelined 2PC prototype for cross-database atomicity

cloutiertyler:
## Summary

Implements pipelined two-phase commit for distributed transactions between SpacetimeDB databases. A coordinator database can call a reducer on a participant database via `call_reducer_on_db_2pc`, and the runtime guarantees that either both sides commit or neither does.

- **Participant side**: New HTTP endpoints (`/prepare/:reducer`, `/2pc/commit/:id`, `/2pc/abort/:id`) and `PreparedTransactions` registry on `ModuleHost`
- **Coordinator side**: New `call_reducer_on_db_2pc` host function (ABI `spacetime_10.5`) that calls the participant's prepare endpoint, with automatic post-commit coordination (sends commit to all participants on success, abort on failure)
- **Bindings**: FFI and safe wrapper for `call_reducer_on_db_2pc`
- **Smoketests**: Happy path (atomic transfer between two databases) and abort path (insufficient funds rolls back both sides)

## Protocol flow

1. Coordinator's reducer calls `call_reducer_on_db_2pc(target, "debit", args)`
2. Runtime POSTs to participant's `/prepare/debit` endpoint
3. Participant executes reducer, commits in-memory, returns result + prepare_id
4. Coordinator's reducer continues with the result
5. On coordinator commit: runtime sends `/2pc/commit/{id}` to all participants
6. On coordinator failure: runtime sends `/2pc/abort/{id}` to all participants

## Prototype scope cuts

- No persistence barrier (speculative transactions persist normally)
- COMMIT on participant is a no-op (data already committed during prepare)
- ABORT does not invert in-memory state (logs warning)
- No crash recovery or PREPARE/COMMIT/ABORT commitlog record types
- No commitlog recording of remote return values (replayability deferred)

## Test plan

- [x] `cargo check -p spacetimedb-standalone` compiles
- [x] `test_cross_db_2pc_happy_path` -- transfer 50 between databases, verify both updated
- [x] `test_cross_db_2pc_abort_insufficient_funds` -- transfer 200 with only 100 available, verify both unchanged


crates/core/src/host/module_host.rs
```diff
@@ -1740,6 +1745,100 @@ impl ModuleHost {
...
     }
 
+    /// Execute a reducer in 2PC prepare mode.
+    ///
+    /// Execute a reducer as a 2PC PREPARE.
+    ///
+    /// 1. Executes the reducer and commits in-memory (releasing the write lock).
+    /// 2. Sends the PREPARE to the durability worker.
+    /// 3. Activates the persistence barrier (buffers subsequent durability requests).
+    /// 4. Waits for the PREPARE to become durable.
```
Shubham8287, line 1755: This seems pretty different from what is actually described in algorithm: https://github.com/clockworklabs/SpacetimeDBPrivate/pull/2755/changes#diff-2f8cf48d43b28906812c44f832cf700e998a3321d37a0fb8a48360f437a9e24cR1613.

As per the algorithm:
1. Prepare shouldn't wait for durability.
2. It should keep datastore lock open
3. There should be additional endpoint for `RELEASE_LOCK`, which should dictate transaction commit and inmemory buffering.

crates/core/src/host/module_host.rs
```diff
@@ -1740,6 +1745,100 @@ impl ModuleHost {
...
+            )
+            .await?;
+
+        // Only store prepared tx info and activate barrier if the reducer succeeded.
+        if matches!(result.outcome, ReducerOutcome::Committed) {
+            use std::sync::atomic::{AtomicU64, Ordering};
+            static PREPARE_COUNTER: AtomicU64 = AtomicU64::new(1);
+            let prepare_id = format!("prepare-{}", PREPARE_COUNTER.fetch_add(1, Ordering::Relaxed));
+
+            // Activate the persistence barrier. The PREPARE transaction has already
```
Shubham8287, line 1786: I think `PREPARE` transaction here is the normal commitlog transaction? In that case, it should not be made durable at this step (which I think is happening by calling `call_reducer_with_return` above)

PR #4725, opened 2026-03-29
bindings-macro: add custom index name to prevent index name collisions

### LINKED ISSUE #4719: Index name colliding
I have tables like this:

```
#[spacetimedb::table(accessor = product_category, public)]
pub struct ProductCategory {
    #[primary_key]
    #[auto_inc]
    pub id: u64,
    #[index(btree)]
    pub store_id: u64,
    pub name: String,
}

#[spacetimedb::table(accessor = product, public)]
pub struct Product {
    #[primary_key]
    #[auto_inc]
    pub id: u64,
    #[index(btree)]
    pub store_id: u64,
    #[index(btree)]
    pub category_id: u64,
    pub name: String,
    pub description: Option<String>,
    pub price: u128,
}
```

it compile just fine when using `spacetime build`
but when I `spacetime publish` it return error:

name `product_category_id_idx_btree` is used for multiple entities

From what I see its because of the index is generated by {accessor}_{field}_idx_btree, so it output the same product_category_id_idx_btree for both of the table. It's nice if we can specify the name of index with something like
```
#[index(btree, name="prod_category_id_idx_btree2")]
```
or maybe even better automagically handle that colliding name in macros

oh, after reading the spactimedb code I see the index name will auto generated with this format:
`{table_name}_{cols}_idx_{algo}`
maybe the simple fix is to change the separator to be double __:
`{table_name}__{cols}_idx_{algo}`
or use another character for separator, I don't know, what do you think?

rudcode:
# Description of Changes

SpacetimeDB generates colliding index names like `product_category_id_idx_btree` (`category_id` in `product` and `id` in `product_category` making same `product_category_id` prefix) , causing spacetime publish to fail with `name used for multiple entities`.

This PR allow users to specify custom index names:
```
// Before (error): index name collision
#[index(btree)]
pub category_id: u64,
```
```
// After (fix): specify unique name
#[index(btree, name = "product_category_id_idx_btree2")]
pub category_id: u64,
```

None

Complexity: 1
A simple change to only bindings-macro/src/table.rs. No breaking changes.

# Testing

Manual testing with the reproducer from Issue #4719 - spacetime publish now succeeds when unique index names are provided. And when custom index name is changed, it successfully migrated.

### DIFF
--- a/crates/bindings-macro/src/table.rs
+++ b/crates/bindings-macro/src/table.rs
@@ -243,10 +243,6 @@ impl IndexArg {
                     check_duplicate(&accessor, &meta)?;
                     accessor = Some(meta.value()?.parse()?);
                 }
-                sym::name => {
-                    check_duplicate(&canonical_name, &meta)?;
-                    canonical_name = Some(meta.value()?.parse()?);
-                }
 
                 sym::btree => {
                     check_duplicate_msg(&algo, &meta, "index algorithm specified twice")?;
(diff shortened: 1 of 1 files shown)


rudcode:
I am just thinking, this maybe makes it hard to distinguish between old 1.x and 2.x syntax. Because old 1.x use `name` instead of `accessor`. This will remove all of the error when user wants to migrate from 1.x to 2.x. Do we need to rename the `name` to more explicit names like `index_name`?

rudcode:
Force pushed to add error message if `name` is `Ident` (we assume it's old 1.x code that need migration). `name` should be LitStr

Shubham8287 (commented):
Thanks for the PR.

Right now, Schema validation code on Host expects `index_name` to always be auto-generated and not user supplied.
Which seems like an unneccessary requirement but we need to get host side change before accepting this PR.

I will create separate ticket to do that.

PR #4729, opened 2026-03-30
Add module ABI & API for `clear`ing tables

Centril:
# Description of Changes

Support is added to Rust, C#, C++, and TS modules.

None

1

# Testing

I wasn't able to regen module bindings for sdk tests, hitting an error while doing so.


crates/bindings-cpp/include/spacetimedb/abi/abi.h
```diff
@@ -35,6 +35,10 @@
 #define STDB_IMPORT_10_4(name) \
     __attribute__((import_module("spacetime_10.4"), import_name(#name))) extern
 
+// Clear syscall is in spacetime_10.4
+#define STDB_IMPORT_10_4(name) \
```
JasonAtClockwork:
```suggestion
#define STDB_IMPORT_10_5(name) \
```

Looks like a small typo here to align with the import below

JasonAtClockwork:
Pushed a fix for C++ to allow access to `clear()` from the `TableAccessor` and fully tested end to end for C++.

PR #4733, opened 2026-03-31
Add support for bytes key btree indices

Centril:
# Description of Changes

Add support for btree indices where the keys are encoded byte strings for e.g., multi-column indices of no-unbounded-types (arrays and strings) that aren't floats.

The main interesting stuff in this PR is in `bytes_key.rs` which defines `RangeCompatBytesKey`, a type that is derived from `BytesKey`, by converting little-endian encoded integers to big-endian. Signed integers are now also supported, but floats are not. `table_index/mod.rs` also includes a bunch of interesting stuff.

Technically this fixes pre-existing bugs in the handling of `Excluded` ranges for multi-col indices.

2?

# Testing

- A proptest `order_in_bsatn_is_preserved` is now adjusted and enabled to exercise the ordering of  `RangeCompatBytesKey`.
- A proptest `btree_multi_col_range_scans_work` is added to check the behavior of range scans on multi-col indices.


crates/table/src/table_index/mod.rs
```diff
@@ -1695,6 +1865,51 @@ impl TableIndex {
                 let suffix_len = suffix_types.len();
 
                 match &self.idx {
+                    BTreeBytesKey8(_) => Self::bounds_from_bsatn_bytes_key(
```
joshua-spacetime: Don't we need to handle `UniqueBTreeBytesKey*` here too?

crates/table/src/table_index/bytes_key.rs
```diff
@@ -183,16 +197,195 @@ impl<const N: usize> BytesKey<N> {
...
+        Self::to_bytes_key(*self, key_type).decode_algebraic_value(key_type)
+    }
+
+    /// Decodes `prefix` and `endpoint` in BSATN to a [`RangeCompatBytesKey<N>`]
+    /// by copying over both and massaging if they fit into the key.
+    pub(super) fn from_bsatn_prefix_and_endpoint(
+        prefix: &[u8],
+        prefix_types: &[ProductTypeElement],
+        endpoint: &[u8],
+        range_type: &AlgebraicType,
```
joshua-spacetime, line 380: `prefix` is the fixed (non-range) part of the scan right? But the entire range key itself could still be a prefix of the index? In which case there's an implicit suffix with `MIN` and `MAX` bounds that we're omitting which is not correct.

Currently I believe if you have a 3-column index on `(u64, u64, u64)`, `prefix=5` and `endpoint=1` effectively becomes the range `[(5,1,0), (5,1,0)]`.

joshua-spacetime (changes requested):
I think more testing is needed for this patch.

crates/table/src/table_index/bytes_key.rs
Centril: This should be fixed now in the last commit. There were also some pre-existing bugs on master that I fixed. A proptest is also included for all of this now.

crates/table/src/table_index/mod.rs
```diff
@@ -1753,7 +1912,31 @@ impl TableIndex {
...
+                    BTreeBytesKey16(_) => bounds_from_bsatn_bytes_key!(TypedIndexKey::BytesKey16),
+                    BTreeBytesKey32(_) => bounds_from_bsatn_bytes_key!(TypedIndexKey::BytesKey32),
+                    BTreeBytesKey64(_) => bounds_from_bsatn_bytes_key!(TypedIndexKey::BytesKey64),
+                    BTreeBytesKey128(_) => bounds_from_bsatn_bytes_key!(TypedIndexKey::BytesKey128),
+                    UniqueBTreeBytesKey8(_) => bounds_from_bsatn_bytes_key!(TypedIndexKey::BytesKey8B),
+                    UniqueBTreeBytesKey16(_) => bounds_from_bsatn_bytes_key!(TypedIndexKey::BytesKey16),
+                    UniqueBTreeBytesKey32(_) => bounds_from_bsatn_bytes_key!(TypedIndexKey::BytesKey32),
+                    UniqueBTreeBytesKey64(_) => bounds_from_bsatn_bytes_key!(TypedIndexKey::BytesKey64),
+                    UniqueBTreeBytesKey128(_) => bounds_from_bsatn_bytes_key!(TypedIndexKey::BytesKey128),
                     BTreeAV(_) | HashAV(_) | UniqueBTreeAV(_) | UniqueHashAV(_) => {
```
joshua-spacetime: `bounds_from_bsatn` now handles the `UniqueBTreeBytesKey*` variants in the multi-column Product branch, but it doesn't handle `*HashBytesKey*` variants. The new tests only cover `btree` indexes. Fixed size multi-column hash indexes remain untested.

Centril: I've substantially hardened the tests now, please have another look.

joshua-spacetime (approved):

PR #4746, opened 2026-04-03
Run reducers in their own V8 `HandleScope` for js modules

joshua-spacetime:
# Description of Changes

The JS worker is intentionally long-lived. Before this patch, we essentially had a memory leak where V8 call-local handles and some host-side call state would survive/accumulate across multiple calls on the worker. That created gradual heap growth over time, more GC work, and eventually enough slowdown and heap pressure that the isolate needed to be replaced.

And even though we periodically check heap statistics to determine if/when we need to replace the isolate, execution latencies can (and did) degrade dramatically before this kicks in.

Now each reducer call is given a fresh V8 `HandleScope` in which to execute instead of reusing a single global scope. This scope is then dropped at the end of each run, which avoids retaining/accumulating call-local JS objects over the JS worker's lifetime.

This patch also makes end-of-call host cleanup explicit, lowers the default heap-check cadence, and limits exported heap metrics to the JS worker only. Previously we had poor heap observability for diagnosis as heap metrics from the instance pool (for procedures) could overwrite heap metrics for the worker.

#### 1. Add a fresh V8 `HandleScope` for every invocation

Each reducer, view, and procedure call now opens a nested V8 scope for the duration of that call.

This preserves the existing long-lived isolate and context, but gives every invocation its own temporary handle lifetime. Call-local V8 handles now die when the invocation returns instead of sticking around until the worker exits.

As part of that refactor:

- Hook locals are rebuilt inside the per-call scope instead of being tied to the worker-lifetime scope.
- The reducer args `ArrayBuffer` is stored and re-used as a worker-lifetime local.

#### 2. Make end-of-call cleanup a real boundary

The V8 host now force-clears leftover per-call host state at the end of a function call.

Specifically:

- Any row iterators left behind by guest code are cleared.
- Any unfinished timing spans are cleared.
- We log when that cleanup had to happen so leaked call-local state is visible instead of silently persisting across invocations.

#### 3. Lower the default heap-check cadence

The default V8 heap policy is now more aggressive about checking worker heap usage.

Defaults changed from:

- `heap-check-request-interval = 65536`
- `heap-check-time-interval = 30s`

to:

- `heap-check-request-interval = 4096`
- `heap-check-time-interval = 5s`

These settings remain configurable through the existing `v8-heap-policy` config.

#### 4. Only export heap metrics for the instance-lane worker

Heap metrics now reflect only the long-lived instance lane.

Specifically:

- Exported V8 heap gauges are emitted only for `worker_kind="instance_lane"`.
- Pooled workers no longer publish these heap metrics.
- Per-database metric cleanup was simplified accordingly.

This avoids the last-writer-wins issue from the pooled instances while keeping the metrics focused on the worker that accumulates state over time and is most relevant for long-run slowdown diagnosis.

None

2

# Testing

Manually tested via the keynote-2 benchmark. Will add the keynote benchmark to CI which will serve as a regression test.

### DIFF
--- a/crates/core/src/host/v8/mod.rs
+++ b/crates/core/src/host/v8/mod.rs
@@ -58,7 +58,7 @@ use tokio::sync::{oneshot, Mutex as AsyncMutex, Notify};
 use tracing::Instrument;
 use v8::script_compiler::{compile_module, Source};
 use v8::{
-    scope_with_context, ArrayBuffer, Context, Function, Isolate, Local, MapFnTo, OwnedIsolate, PinScope,
+    scope_with_context, ArrayBuffer, Context, Function, Global, Isolate, Local, MapFnTo, OwnedIsolate, PinScope,
     ResolveModuleCallback, ScriptOrigin, Value,
 };
 
@@ -1379,6 +1379,7 @@ async fn spawn_instance_worker(
         }
 
         // Setup the instance common.
+        let args = Global::new(scope, ArrayBuffer::new(scope, REDUCER_ARGS_BUFFER_SIZE));
         let info = &module_common.info();
         let mut instance_common = InstanceCommon::new(&module_common);
         let replica_ctx: &Arc<ReplicaContext> = module_common.replica_ctx();
@@ -1390,6 +1391,7 @@ async fn spawn_instance_worker(
             scope,
             replica_ctx,
             hooks: &hooks,
+            args: &args,
         };
         if let Some(heap_metrics) = heap_metrics.as_mut() {
             let _initial_heap_stats = sample_heap_stats(inst.scope, heap_metrics);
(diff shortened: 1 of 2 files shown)


crates/core/src/host/v8/mod.rs
```diff
@@ -1542,7 +1648,16 @@ struct V8Instance<'a, 'scope, 'isolate> {
...
+
+macro_rules! with_call_scope {
+    ($scope:expr, |$call_scope:ident, $hooks:ident| $body:block) => {{
+        // Open a fresh HandleScope for this invocation so call-local V8 handles
+        // are released when the reducer/view/procedure returns.
+        v8::scope!(let $call_scope, $scope);
+        let $hooks = get_registered_hooks($call_scope).expect("module hooks should be registered before invoking JS");
+        $body
+    }};
 }
```
coolreader18, line 1664: Couldn't this just be a function?

joshua-spacetime: It could. It would just need a higher-ranked lifetime bound which I usually try to avoid.

PR #4752, opened 2026-04-06
Add useProcedure React hook and export procedures from codegen

### LINKED ISSUE #4751: React SDK missing `useProcedure` hook and procedures accessor map
The React SDK provides `useReducer` and `useTable` hooks but has no equivalent `useProcedure` hook. Additionally, the TypeScript codegen emits `export const reducers = __convertToAccessorMap(...)` but does not do the same for procedures, so there is no way to get individual procedure definitions to pass to a hook.

React users who want to call procedures have to reach into `getConnection().procedures` directly, losing the connection-queuing behavior and ergonomic pattern that `useReducer` provides.

DexterKoelson:
Closes #4751

# Description of Changes
- Add `useProcedure` React hook that mirrors `useReducer` which returns a stable, typed callback that queues calls until the connection is ready
- Add `ProcedureParamsType` and `ProcedureReturnType` utility types to `type_utils.ts`
- Update TypeScript codegen to emit `export const procedures = __convertToAccessorMap(proceduresSchema.procedures)` in generated module bindings, matching the existing pattern for reducers

```ts
import { procedures } from './module_bindings';
import { useProcedure } from 'spacetimedb/react';

const doSomeThing = useProcedure(procedures.doSomeThing);
const result = await doSomeThing({ foo: "..." });
```

None. Additive only — new hook export and new codegen line.

Low. The hook is a near-copy of `useReducer` adapted for procedure signatures. The codegen change adds one line following the identical pattern used for reducers.

# Testing
- [x] TypeScript SDK compiles with no new errors (`tsc --noEmit`)
- [x] All 170 existing tests pass (17 test files)
- [x] Tested end-to-end in a React app calling a procedure


JasonAtClockwork (approved):
Looks solid, mirrors the useReducer I've added a few small changes to get CI to pass.

PR #4763, opened 2026-04-09
Update C# sdk to use v3 websocket api

joshua-spacetime:
# Description of Changes

This PR adds C# SDK support for the `v3.bsatn.spacetimedb` WebSocket transport  which was added in #4761 and keeps `v2` as a fallback.

`v3` is just a transport wrapper around already-encoded v2 logical messages:

- the client offers `v3` first and falls back to `v2` during handshake when needed
- when `v3` is negotiated, outbound client messages are batched into a single transport frame
- v3 batching is capped at `256 KiB` per frame
- after sending one capped v3 frame, the send pump schedules a follow-up pass using the SDK’s existing scheduling primitives rather than draining the entire queue in one uninterrupted pass
- inbound `v3` server frames are unwrapped into their inner v2 messages and then processed by the existing v2 parse/apply pipeline in order

This also includes Unity WebGL transport updates so the C# SDK can:
- offer multiple WebSocket subprotocols
- observe the negotiated protocol
- switch send/receive transport behavior accordingly

None. Existing servers that only support `v2` continue to work because the SDK falls back during handshake.

3

The main risks are:
- preserving message ordering when multiple logical messages share one v3 frame
- preserving fallback behavior for `v2`
- keeping the capped batching behavior from starving other websocket work
- keeping WebGL protocol negotiation and callback marshalling correct

The risk is limited by keeping the existing v2 logical message pipeline intact underneath the new transport layer.

# Testing

Integration tests for v3 and v2 fallback

### DIFF
--- a/crates/client-api-messages/DEVELOP.md
+++ b/crates/client-api-messages/DEVELOP.md
@@ -19,3 +19,12 @@ spacetime generate -p spacetimedb-cli --lang <SDK lang> \
   --out-dir <sdk WebSocket schema bindings dir> \
   --module-def ws_schema_v2.json
 ```
+
+For the v3 WebSocket transport schema:
+
+```sh
+cargo run --example get_ws_schema_v3 > ws_schema_v3.json
+spacetime generate -p spacetimedb-cli --lang <SDK lang> \
+  --out-dir <sdk WebSocket schema bindings dir> \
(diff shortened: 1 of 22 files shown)


joshua-spacetime:
Closing in favor of https://github.com/clockworklabs/SpacetimeDB/pull/5531

PR #4775, opened 2026-04-10
More granular table traits

onx2:
# Description of Changes

Refactored the Rust SDK table trait surface into composable capabilities and updated Rust codegen to match. [Discord thread](https://discord.com/channels/1037340874172014652/1492137984902959164)

### Motivation

The granular traits are a useful type-system cleanup on their own, but the main motivation is to let downstream integration crates like [bevy_stdb](https://github.com/onx2/bevy_stdb) express callback capabilities more directly instead of routing them through broader semantic categories.

For example, `bevy_stdb` may want to support bindings like “forward only inserts from this subscribed view as Bevy messages” for a relation that is not semantically an `EventTable`. Today this can still be made to work with broader abstractions, so this is not primarily about enabling new runtime behavior. The value is mostly at compile time: more precise trait bounds, clearer downstream APIs, and better guarantees that a binding only requires the capabilities it actually uses.

Concretely, this makes it easier for integration crates to model things like insert-only or delete-only bindings for ordinary tables or views without conflating those behaviors with `EventTable` semantics.

Example:

```rust
// Current semantic shape:
plugin.add_event_table::<LogRow>(|reg, db| reg.bind(db.logs()));

// Potential capability-based shape:
plugin.add_insert_source::<CharacterRow>(|reg, db| reg.bind(db.characters_near_me()));
plugin.add_delete_source::<CharacterRow>(|reg, db| reg.bind(db.characters_near_me()));
```

### Summary
- Added a shared `TableLike` trait for common table handle functionality:
  - `Row`
  - `EventContext`
  - `count`
  - `iter`
- Added capability traits:
  - `WithInsert`
  - `WithDelete`
  - `WithUpdate`

None. This is additive only.

2/5

This is additive but potential risks/complexity are that the surface area of the API is larger and there isn't really overlap between original approach and the additive. So maybe in usage we'll need to consider satisfying those new traits too?

# Testing

- [x] Ran `cargo check --manifest-path /Users/onx2/Documents/projects/SpacetimeDB/sdks/rust/Cargo.toml`
- [x] Ran `cargo test --manifest-path /Users/onx2/Documents/projects/SpacetimeDB/Cargo.toml -p spacetimedb-codegen test_codegen_rust`
- [x] Regenerated Rust client bindings using the workspace CLI and verified generated impls match the new trait structure
- [ ] Reviewer: regenerate a Rust client with the workspace CLI and confirm the generated bindings compile cleanly


sdks/rust/src/table.rs
onx2, line 1: This is the main change to functionality. I wanted to have access to specific traits so in [bevy_stdb](https://github.com/onx2/bevy_stdb/blob/main/src/table.rs#L201) I can go from:

**Before/after**
```diff
fn bind_insert<TRow, TTable>(world: &World, table: &TTable)
where
    TRow: Send + Sync + Clone + InModule + 'static,
    RowEvent<TRow>: Send + Sync,
-   TTable: Table<
+   TTable: WithInsert<
            Row = TRow,
            EventContext = <<TRow as InModule>::Module as SpacetimeModule>::EventContext,
        >,
    TTable::EventContext: AbstractEventContext<Event = RowEvent<TRow>>,
{
    let sender = channel_sender::<InsertMessage<TRow>>(world);
    table.on_insert(move |ctx, row| {
        let _ = sender.send(InsertMessage {
            event: ctx.event().clone(),
            row: row.clone(),
        });
    });
}
```

crates/codegen/src/rust.rs
onx2, line 1: Trait changes mean we need updated codegen too (breaking).

crates/codegen/src/rust.rs
```diff
@@ -203,6 +205,8 @@ impl<'ctx> __sdk::EventTable for {table_handle}<'ctx> {{
         self.imp.remove_on_insert(callback.0)
     }}
 }}
+
+impl<'ctx> __sdk::EventTable for {table_handle}<'ctx> {{}}
```
onx2, line 209: Do we need these empty impls? Maybe now Table can just be removed entirely?

sdks/rust/src/table.rs
```diff
@@ -7,12 +8,10 @@
...
 //! Obtain a table handle by calling a method on `ctx.db`, where `ctx` is a `DbConnection` or `EventContext`.
 
-/// Trait implemented by table handles, which mediate access to tables in the client cache.
+/// Common functionality shared by table-like handles.
 ///
 /// Obtain a table handle by calling a method on `ctx.db`, where `ctx` is a `DbConnection` or `EventContext`.
-///
-/// For persistent (non-event) tables only. See [`EventTable`] for transient event tables.
-pub trait Table {
+pub trait TableLike {
```
onx2, line 14: Maybe this name could be better or more in line with the functionality... like WithIterator or WithCountable 🤷🏻

cloutiertyler:
I see the Discord thread, but could you expand on what the motivation for this is in the PR description? An example of what this enables would be helpful.

onx2:
> I see the Discord thread, but could you expand on what the motivation for this is in the PR description? An example of what this enables would be helpful.

Updated 😄

cloutiertyler:
Is there any way to support this functionality without a breaking change? We only do breaking changes at major version boundaries.

onx2:
> Is there any way to support this functionality without a breaking change? We only do breaking changes at major version boundaries.

@cloutiertyler I think the new version should work without breaking changes b/c its only adding new capability traits instead of basing the previous table traits off of them.

onx2:
Is there anything I can do to help get this through the door? Any issues or concerns I can address? I normally wouldn't make a stink of waiting for a review but it has been over 2 months now 😅

gefjon (approved):
This seems reasonable. Sorry for the long delay on review; I've been on vacation for a bit.

PR #4777, opened 2026-04-10
[TS] Implement near-heap-limit termination

coolreader18:
# Description of Changes

Uses `Isolate::add_near_heap_limit_callback` to prevent unbounded heap growth. Upon nearing the heap limit, we will now:
1. Call `Isolate::terminate_execution`.
2. Request that v8 double the heap limit.

Then, upon finishing a function call, we lower the heap limit back down.

This should hopefully fix the issue where v8 hits the heap limit and crashes the whole process.

Also improves the way termination requests are checked for and processed.

2

# Testing

- [ ] Manual testing with memory-leaky modules


crates/core/src/host/v8/mod.rs
```diff
@@ -1731,73 +1757,120 @@ impl WasmInstance for V8Instance<'_, '_, '_> {
...
-                    Ok(None) => {}
-                    Err(exc) => e = ErrorOrException::Exception(exc),
+    let isolate_handle = scope.thread_safe_handle();
+
+    let near_heap_limit_callback = |current_heap_limit: usize, _initial_heap_limit: usize| {
+        heap_limit_hit.update(|x| x + 1);
+        inst.heap_limit_hit_metric.inc();
+        if !isolate_handle.is_execution_terminating() {
+            terminated_because_of_heap_limit.set(true);
+            isolate_handle.terminate_execution();
```
Centril: This should use the `terminate_execution` so that `scope.set_slot(TerminationFlag)` happens and so that attempted syscalls will return immediately.

crates/core/src/host/v8/mod.rs
```diff
@@ -1731,73 +1757,120 @@ impl WasmInstance for V8Instance<'_, '_, '_> {
...
-            // If we're terminating execution, don't try to check `instanceof`.
-            if scope.can_continue()
-                && let Some(exc) = scope.exception()
-            {
-                match process_thrown_exception(scope, hooks, exc) {
-                    Ok(Some(err)) => return err,
-                    Ok(None) => {}
-                    Err(exc) => e = ErrorOrException::Exception(exc),
+    let isolate_handle = scope.thread_safe_handle();
+
```
```suggestion
    // This closure gets called when the configured near heap limit is hit.
    // We use a near-heap limit as V8 aborts the process when the hard limit is hit.
    // We'd like to gracefully terminate execution, and not abort,
    // so we ask V8 to terminate execution as soon as possible
    // and double the heap limit in the hopes
    // that V8 manages to stop execution before hitting the new limit.
    // Note that V8 does not terminate execution immediately when requested
    // so this mechanism is unfortunately not fool-proof
    // but we consider it to be good enough.
    let isolate_handle = scope.thread_safe_handle();
```

crates/core/src/host/v8/mod.rs
```diff
@@ -1731,73 +1757,120 @@ impl WasmInstance for V8Instance<'_, '_, '_> {
...
+                // We can continue.
+                ExecutionError::Recoverable(e.unwrap_or_else(Into::into))
+            } else if scope.has_terminated() {
+                // We can continue if we do `Isolate::cancel_terminate_execution`.
+                scope.cancel_terminate_execution();
+                let e = e.unwrap_or_else(|unknown| {
+                    if terminated_because_of_heap_limit.get() {
+                        anyhow::Error::msg("JavaScript module exceeded memory limit")
+                    } else {
+                        // TODO(noa): check for other causes of termination and give a better error message
```
Centril: This could use some elaboration if you know what the other causes could be.

crates/core/src/host/v8/mod.rs
coolreader18: Hmm. I don't think we can use that, because it takes a scope. But I'll make sure `TerminationFlag` gets set.

crates/core/src/host/v8/mod.rs
```diff
@@ -1770,18 +1777,24 @@ where
...
-    let near_heap_limit_callback = |current_heap_limit: usize, _initial_heap_limit: usize| {
+    // This closure gets called when the configured near heap limit is hit.
+    // We use a near-heap limit as V8 aborts the process when the hard limit is hit.
+    // We'd like to gracefully terminate execution, and not abort,
+    // so we ask V8 to terminate execution as soon as possible
+    // and double the heap limit in the hopes
+    // that V8 manages to stop execution before hitting the new limit.
+    // Note that V8 does not terminate execution immediately when requested
+    // so this mechanism is unfortunately not fool-proof
+    // but we consider it to be good enough.
```
Centril, line 1789: Is this not true anymore? Looks like you made it fool-proof with `handle_interrupts`?

Centril (approved):

crates/core/src/host/v8/mod.rs
coolreader18: No, because it's still theoretically possible for v8 code to go on without checking for interrupts. I don't think that's likely to happen but it'd requiring auditing all of v8 to verify.

PR #4778, opened 2026-04-11
Cache V8 heap stats instead of querying them on every call

joshua-spacetime:
# Description of Changes

Previously we would invoke V8's `GetHeapStatistics` on every reducer call in order to compute the `memory_allocation` part of its `ExecutionStats`. We already sample heap statistics periodically, so now we cache the result after each sample and just read the cached value after each call. Memory allocation is used for energy tracking and so eventual consistency is fine.

None

1

# Testing

Refactor/optimization

### DIFF
--- a/crates/core/src/host/v8/mod.rs
+++ b/crates/core/src/host/v8/mod.rs
@@ -1805,6 +1805,7 @@ where
     let energy = energy_from_elapsed(budget, timings.total_duration);
 
     // Reuse the last periodic heap sample instead of querying V8 on every call.
+    // We use this statistic for energy tracking, so eventual consistency is fine.
     let memory_allocation = env_on_isolate_unwrap(scope).cached_used_heap_size();
 
     let stats = ExecutionStats {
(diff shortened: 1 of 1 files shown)


crates/core/src/host/v8/mod.rs
```diff
@@ -1788,9 +1804,8 @@ where
     // Derive energy stats.
     let energy = energy_from_elapsed(budget, timings.total_duration);
 
-    // Fetch the currently used heap size in V8.
-    // The used size is ostensibly fairer than the total size.
-    let memory_allocation = scope.get_heap_statistics().used_heap_size();
+    // Reuse the last periodic heap sample instead of querying V8 on every call.
```
Centril, line 1807:
```suggestion
    // Reuse the last periodic heap sample instead of querying V8 on every call.
    // We use this statistic for energy tracking, so eventual consistency is fine.
```

Centril (approved):
Looks good, but I'd like to just embed your comment re. eventual consistency before we merge.

coolreader18 (approved):

PR #4779, opened 2026-04-12
move 00300-spacetime-json.md into the right docs folder

xDovos:
# Description of Changes

moves the docs file for spacetime.json into the right folder so that the website shows it.

no

1

# Testing

no testing needed


cloutiertyler (approved):
Thank you!

PR #4783, opened 2026-04-13
Increase timeout for typescript query builder tests

joshua-spacetime:
# Description of Changes

Bumps the timeout for two typescript query builder tests. Both tests were using the default 5 second timeout, and recently they seem to be hitting it pretty consistently. I've increased the timeout to 15 seconds for these two tests.

None

1

# Testing

Typescript `build-and-test` should now pass in CI


bfops (approved):
seems fair to me

PR #4802, opened 2026-04-14
durability: Use `async-channel` to allow blocking send

kim:
The previous approaches would either:

- panic when the queue becomes full, as `append_tx` is run inside the context of a `LocalSet`, which is basically a glorified current thread runtime
- deadlock because the receiver runtime has no way of notifiying the sender of freed capacity in the channel

`async-channel` handles wait queues and notifications internally, so can be used freely from either blocking or async contexts.

This _may_ come at different performance characteristics, but I haven't measured them.


joshua-spacetime (approved):
I ran some benchmarks and it didn't seem to negatively affect performance in any way.

PR #4804, opened 2026-04-14
Extract replay stuff out of `CommittedState`, part 1

Centril:
# Description of Changes

Moves stuff out of `datastore.rs` and `committed_state.rs` that has to do with replay into its own module `replay.rs`.
This also shaves 64 bytes off of `CommittedState` so that we don't continue carrying the cost of replay after it has finished.

This is a more limited version / part 1 of https://github.com/clockworklabs/SpacetimeDB/pull/4055 which focuses on almost pure code motion so that we're more diligent and careful about the transition. Future PRs will continue with other parts of #4055.

None

1, just simple and safe code motion.

# Testing

Covered by existing tests as this is code motion.

### DIFF
--- a/crates/datastore/src/locking_tx_datastore/replay.rs
+++ b/crates/datastore/src/locking_tx_datastore/replay.rs
@@ -1,4 +1,5 @@
 use super::committed_state::CommittedState;
+use super::datastore::Result;
 use crate::db_metrics::DB_METRICS;
 use crate::locking_tx_datastore::datastore::ReplayError;
 use crate::locking_tx_datastore::state_view::StateView;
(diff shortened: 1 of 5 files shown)


crates/datastore/src/locking_tx_datastore/replay.rs
```diff
@@ -332,7 +343,24 @@ impl DerefMut for ReplayCommittedState<'_> {
     }
 }
 
-impl ReplayCommittedState<'_> {
+impl MemoryUsage for ReplayCommittedState<'_> {
+    fn heap_usage(&self) -> usize {
+        let Self {
+            state: _, // We don't attribute usage to `CommittedState`.
```
gefjon: What? That's wacky. I'd like to see a bit more justification here.

gefjon (approved):

crates/datastore/src/locking_tx_datastore/replay.rs
Centril: Well `state` does not own the `CommittedState`, so istm like the right behavior. However, this impl was unused, so I've removed it.

PR #4805, opened 2026-04-14
Keep subscription fanout worker warm with adaptive linger policy

joshua-spacetime:
# Description of Changes

Similar to https://github.com/clockworklabs/SpacetimeDB/pull/4801, after we evaluate subscriptions on the main database thread, we send the results to a worker whose job it is to fan out the updates for the relevant clients. Hence we want to make sure this worker is not constantly parked on `recv()` as each `send` on the main thread will incur overhead waking the task.

To avoid this I've added a utility that wraps an `mpsc` `UnboundedReceiver` with an adaptive "linger" policy. On each message `recv`, the worker will now "linger" for a period of time and wait for any more messages before parking itself on the `recv()` again.

None

1

# Testing

Manual performance testing for now. Automation to follow.

### DIFF
--- a/crates/core/src/util/adaptive_recv.rs
+++ b/crates/core/src/util/adaptive_recv.rs
@@ -1,0 +1,150 @@
+use std::time::Duration;
+
+use tokio::sync::mpsc;
+use tokio::time::sleep;
+
+/// Receives from a Tokio unbounded channel with an adaptive linger policy.
+///
+/// This helper is intended for single-consumer background workers that want
+/// to avoid parking on `recv()` after every message during bursty traffic.
+///
+/// The receiver has two modes - hot and cold. In cold mode it blocks on
+/// `recv()` until the next message arrives. In hot mode it prefers to stay
+/// awake, so after receiving a message, it will drain the channel, sleep for
+/// a short period (linger), and only then poll the channel again. This keeps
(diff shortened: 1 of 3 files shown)


crates/core/src/util/adaptive_recv.rs
```diff
@@ -0,0 +1,150 @@
...
+        loop {
+            if !self.is_hot {
+                let message = self.rx.recv().await?;
+                self.is_hot = true;
+                return Some(message);
+            }
+
+            match self.rx.try_recv() {
+                Ok(message) => return Some(message),
+                Err(mpsc::error::TryRecvError::Disconnected) => return None,
```
Centril: Nit: can you import `TryRecvError`?

crates/core/src/util/adaptive_recv.rs
```diff
@@ -0,0 +1,150 @@
...
+                Err(mpsc::error::TryRecvError::Empty) => {}
+            }
+
+            let linger = self.linger.current();
+            if linger.is_zero() {
```
Centril, line 81:
```suggestion
            // In case the baseline is zero,
            // this will cause the loop to do `self.rx.recv(..)`.
            if linger.is_zero() {
```

Centril (approved):
Looks good, just some small nits.

PR #4807, opened 2026-04-15
Move field `replay_table_updated` to `ReplayCommittedState`

Centril:
# Description of Changes

Shaves another 32 bytes off of `CommittedState` by moving the last replay only map out of the type.

None

2

# Testing

Covered by existing tests.


crates/datastore/src/locking_tx_datastore/committed_state.rs
aasoni, line 61: It sounds like this TODO is addressed now?

Centril: Oh yes, good catch!

aasoni (approved):

PR #4808, opened 2026-04-15
durability: Simplify shutdown

kim:
We switched to using `async-channel` as the backing queue in 2b3aa5a,
which allows to close the channel on the sender side without dropping
the sender. This is unlike tokio channels, which require the sender to
be dropped, or the receiver to request the channel to be closed.

This allows to greatly simplify the async shutdown of a durability
instance by simply awaiting the actor's `JoinHandle` instead of a
complicated chain of notification primitives.

### DIFF
--- a/crates/durability/src/imp/local.rs
+++ b/crates/durability/src/imp/local.rs
@@ -1,14 +1,13 @@
 use std::{
     io,
     num::NonZeroUsize,
-    path::PathBuf,
     sync::{
         atomic::{AtomicU64, Ordering::Relaxed},
-        Arc,
+        Arc, Mutex,
     },
 };
 
-use futures::{FutureExt as _, TryFutureExt as _};
+use futures::FutureExt as _;
 use itertools::Itertools as _;
 use log::{info, trace, warn};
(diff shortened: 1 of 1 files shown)


Shubham8287 (approved):
This is much simplified indeed.

crates/durability/src/imp/local.rs
```diff
@@ -356,26 +303,27 @@ impl<T: Send + Sync + 'static> Durability for Local<T> {
         info!("close local durability");
 
         let durable_offset = self.durable_tx_offset();
-        let shutdown = self.shutdown.clone();
+        let maybe_actor = self.actor.lock().unwrap().take();
```
Shubham8287, line 306: nit: comment just below this line is outdated

kim: I don't think: the scopeguard makes it so that the actor task is aborted if the `Close` future is dropped. We're calling `RelationalDB::shutdown` with a timeout in non-test contexts, so if the timeout elapses before the queue is fully drained, the remaining ones will be dropped.

gefjon (approved):

PR #4816, opened 2026-04-16
Move execution ack off of the critical path

joshua-spacetime:
# Description of Changes

Today, for each client connection, we process requests one at a time, and we wait for each request to finish before processing the next one. The way this works is that each websocket reader waits for a notification (via a one shot channel) from the JS worker that it is done processing the request. This notification is quite costly because it always wakes the receiving task.

To avoid this costly wake, this patch adds an intermediate relay actor. The relay reads from its message queue periodically and forwards the replies on to the correct websocket reader task. This trades latency for increased throughput.

I would classify this as a hack, and so I'm not sure I would recommend merging this. A potentially better solution would be to remove the reply altogether. However if we want to preserve execution order per connection, I'm not sure if this is possible today, because I'm not sure if we have a single serialization point for all requests. But I think it is ultimately the direction we want to go.

None

1

# Testing

Manual

### DIFF
--- a/crates/core/src/host/v8/mod.rs
+++ b/crates/core/src/host/v8/mod.rs
@@ -31,6 +31,7 @@ use crate::host::{ModuleHost, ReducerCallError, ReducerCallResult, Scheduler};
 use crate::module_host_context::ModuleCreationContext;
 use crate::replica_context::ReplicaContext;
 use crate::subscription::module_subscription_manager::TransactionOffset;
+use crate::util::adaptive_recv::AdaptiveUnboundedReceiver;
 use crate::util::jobs::{AllocatedJobCore, CorePinner, LoadBalanceOnDropGuard};
 use crate::worker_metrics::WORKER_METRICS;
 use core::any::type_name;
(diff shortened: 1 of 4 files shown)


crates/core/src/host/v8/mod.rs
```diff
@@ -1437,22 +1502,22 @@ async fn spawn_instance_worker(
                     policy,
                 } => {
                     let res = instance_common.update_database(program, old_module_info, policy, &mut inst);
-                    send_worker_reply("update_database", reply_tx, res);
+                    send_worker_reply(reply_relay.as_ref(), "update_database", reply_tx, res);
```
Centril, line 1505: Can you do this once at the top of the function so that the `.as_ref()` noise is removed?

crates/core/src/host/v8/mod.rs
Centril, line 263:
```suggestion

        let lane_queue = JsWorkerQueue::unbounded_with_metric(
```

The paragraph comment "Convert program to a string." now also includes unrelated code. Let's fix that.

Centril (approved):
Looks OK, just some nits.

I'm fine with merging this in the interest of perf. It's not that bad for now.

PR #4834, opened 2026-04-16
Add some tests and metadata to LockedFile

jsdt:
# Description of Changes

1. This adds some basic tests of `LockedFile` (which would generally have also passed before these changes).
2. `LockedFile` can now have contents, which are added to the errors if we can't aquire the lock. The default metadata has the pid and a timestamp.

I now see that there is a separate `Lockfile` that doesn't clean up locks on a crash, so that should probably also be cleaned up (but in a separate PR).

1.

# Testing

This has unit tests.


kim (approved):
Can't hurt

kim:
We may also consider to log the file path on drop and/or explicitly unlock the file when dropped, so that we could pair close and open attempts.

PR #4835, opened 2026-04-16
@brougkr fix(unreal): Replace FTickableGameObject with FTSTicker

JasonAtClockwork:
# Description of Changes
This is a copy of #4006 with only the updates to the way Unreal handles ticks. As per @brougkr findings:

- **FTickableGameObject initialization order bug** - Replaced with FTSTicker for reliable tick registration:
  - Removed FTickableGameObject inheritance
  - Added FTSTicker::FDelegateHandle for manual tick registration
  - Added destructor to clean up ticker registration
  - Added OnTickerTick() method

### Issue
FTickableGameObject registers itself in its constructor BEFORE UDbConnectionBase's constructor body runs. Even with ETickableTickType::Never, UE's GENERATED_BODY() macro can interfere with base class initialization order, causing the default constructor to be called instead.

- Refactor of an underlying component of the SDK and non-breaking.

2 - This changes the structure of how the database tooling can auto-tick, but is invisible to the developer

# Testing

As this changes a core feature I tested all aspects:
- [x] Reproduced the bug and confirmed against `master` and this branch to see the fix working
- [x] Ran full suite of Unreal tests
- [x] Manually tested Unreal Blackholio


rekhoff (approved):
Looks good to me

PR #4838, opened 2026-04-17
Update the defaultValue check to allow for falsy and undefined values

### LINKED ISSUE #4700: Table field default value 0 or "" isn't interpreted as a default value
When adding a new field to a table that has data and the default value is a number 0 or empty string for the new field, spacetimedb checks if (defaultValue) and refuses to merge because there is no default value.

I found this from the library:
```
if (meta.defaultValue) {
      const writer = new BinaryWriter(16);
      builder.serialize(writer, meta.defaultValue);
      defaultValues.push({
        colId: colIds.get(name)!,
        value: writer.getBuffer(),
      });
    }
```

Should check if (meta.defaultValue !== undefined) or something.

JasonAtClockwork:
# Description of Changes

This fixes a bug in TypeScript table default serialization where falsy defaults like 0, '', false, and explicit undefined were treated as if no default was provided. I did this as a property check in order to allow `undefined` which becomes None() as the API allows `T | undefined`.

Closes: #4700 

N/A 

1 - Simple change to check the default value was added as a property

# Testing

- [x] Reproduced the failing issue in a simple throwaway project


Shubham8287 (approved):
Thanks

PR #4843, opened 2026-04-18
Deprecate ReducerContext::identity in favor of database_identity

### LINKED ISSUE #3201: Deprecate `ReducerContext::identity` in favor of `ReducerContext::database_identity`
We have multiple reports of users (e.g. in [this thread in the public Discord](https://discord.com/channels/1037340874172014652/1409532127279710248/1409532127279710248)) being confused that `ctx.identity()` in module code reads the database identity, and calling this method when they intend to use `ctx.sender`. We should rename `ctx.identity()` to be more clear; `ctx.database_identity()` seems obvious and unambiguous. The path to doing this without a breaking change is to add the new method and to mark the old one as `#[deprecated]`.

Also make the corresponding change in C# modules, and figure out what their equivalent of `#[deprecated]` is.

After doing this, add a new ticket to review the client's `conn.identity()` and consider whether it should be renamed to `conn.client_identity()`.

Mr-Dust0:
## Summary
- add `ReducerContext::database_identity()` in Rust bindings
- deprecate `ReducerContext::identity()` and keep it as a compatibility alias
- update reducer docs example to use `ctx.database_identity()`
- add C# reducer-context equivalent: `DatabaseIdentity` and obsolete `Identity` alias
- update Rust/C# module test callsites to the new API name
- update C# codegen snapshots for generated `ReducerContext` API output

## Why
Issue #3201 reports user confusion between reducer `ctx.identity()` (database/module identity) and `ctx.sender`. This change clarifies naming while preserving compatibility.

## Validation
- `cargo check -p spacetimedb -p module-test` (Passed)
- `dotnet test crates/bindings-csharp/Codegen.Tests/Codegen.Tests.csproj --nologo` (Passed)
- `dotnet test crates/bindings-csharp/Runtime.Tests/Runtime.Tests.csproj --nologo` (Failed) pre-existing unrelated failure:
  - `Runtime.Tests/JwtClaimsTest.cs(10,23): CS1729: 'JwtClaims' does not contain a constructor that takes 2 arguments`

## Compatibility
- Rust: `identity()` still works but is deprecated in favor of `database_identity()`.
- C#: `Identity` still works but is marked `[Obsolete]` in favor of `DatabaseIdentity`.

Closes #3201

### DIFF
--- a/crates/bindings-cpp/include/spacetimedb/procedure_context.h
+++ b/crates/bindings-cpp/include/spacetimedb/procedure_context.h
@@ -97,15 +97,20 @@ public:
      * 
      * Example:
      * @code
-     * auto module_id = ctx.identity();
+     * auto module_id = ctx.database_identity();
      * std::string url = "http://localhost:3000/v1/database/" + 
      *                   module_id.to_hex() + "/schema?version=9";
      * @endcode
      */
-    Identity identity() const {
+    Identity database_identity() const {
         std::array<uint8_t, 32> id_bytes;
         ::identity(id_bytes.data());
         return Identity(id_bytes);
+    }
+
+    [[deprecated("Use database_identity() instead.")]]
+    Identity identity() const {
+        return database_identity();
     }
 
     /**
--- a/crates/bindings-cpp/include/spacetimedb/reducer_context.h
+++ b/crates/bindings-cpp/include/spacetimedb/reducer_context.h
@@ -58,10 +58,15 @@ public:
         return *rng_instance;
     }
 
-    Identity identity() const {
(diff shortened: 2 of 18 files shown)


bfops:
Hi @Mr-Dust0, thank you for opening a PR. We try to keep changes in sync across our many supported languages. Could you make these changes for TypeScript and C++ modules as well?

Mr-Dust0:
Hi @bfops thank you for the review means a lot. I've updated the PR to apply the same changes to the typescript and c++ module APIs as well.

gefjon (approved):

PR #4846, opened 2026-04-20
Wait for database update to become durable

kim:
Confirmed reads applies only to subscription clients, calls to the the
HTTP API publish endpoint return a success response before the operation
is confirmed.

While we await scheduling of a new database, updates require to wait for
the update transaction to be confirmed. To allow this, the
`TransactionOffset` channel and the database's `DurableOffset` need to
be returned all the way up to the request handler.

Note that waiting for confirmation is almost always the right choice, so
can't be opted out of at the time of submission of this patch. Callers
may, however, extend the timeout after which waiting for confirmation is
cancelled.

### DIFF
--- a/crates/client-api/src/routes/database.rs
+++ b/crates/client-api/src/routes/database.rs
@@ -9,6 +9,7 @@ use crate::auth::{
     SpacetimeIdentityToken,
 };
 use crate::routes::subscribe::generate_random_connection_id;
+use crate::util::serde::humantime_duration;
 pub use crate::util::{ByteStringBody, NameOrIdentity};
 use crate::{
     log_and_500, Action, Authorization, ControlStateDelegate, DatabaseDef, DatabaseResetDef, Host, MaybeMisdirected,
@@ -701,17 +702,30 @@ pub struct PublishDatabaseQueryParams {
     organization: Option<NameOrIdentity>,
     /// Duration to wait for a database update to become confirmed (i.e. durable).
     ///
+    /// The value is parsed via the `humantime` crate, e.g. "1m", "23s", "5min".
+    ///
     /// If not given, defaults to [default_update_confirmation_timeout].
+    /// The maximum timeout is capped by [MAX_UPDATE_CONFIRMATION_TIMEOUT].
+    ///
     /// The parameter has no effect when creating a new database.
-    update_confirmation_timeout: Option<Duration>,
+    #[serde(with = "humantime_duration", default = "default_update_confirmation_timeout")]
+    update_confirmation_timeout: Duration,
 }
 
 /// Default timeout for a database update to become confirmed / durable.
 ///
 /// Currently, the value is 5s.
-fn default_update_confirmation_timeout() -> Duration {
+const fn default_update_confirmation_timeout() -> Duration {
     Duration::from_secs(5)
 }
+
+/// Maximum timeout for a database update to become confirmed / durable.
+///
+/// If a replication group doesn't converge within this time span, it is
+/// probably not making progress at all.
+///
+/// Currently, the value is 5min.
+const MAX_UPDATE_CONFIRMATION_TIMEOUT: Duration = Duration::from_secs(5 * 60);
 
 pub async fn publish<S: NodeDelegate + ControlStateDelegate + Authorization>(
     State(ctx): State<S>,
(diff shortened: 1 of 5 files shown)


crates/core/src/host/wasm_common/module_host_actor.rs
```diff
@@ -648,35 +676,26 @@ impl InstanceCommon {
                                 ),
                             };
 
+                            let (_, tx_metrics, reducer) = stdb.rollback_mut_tx(tx);
+                            stdb.report_mut_tx_metrics(reducer, tx_metrics, None);
                             UpdateDatabaseResult::ErrorExecutingMigration(anyhow::anyhow!(msg))
                         } else {
-                            UpdateDatabaseResult::UpdatePerformed
+                            let tx_offset = succeed(self.info.clone(), out.energy_used.into(), out.total_duration, tx);
```
kim, line 682: Moving the code around for borrowchk reasons, I see that `energy_quanta_used` and `host_execution_duration` only have non-zero values in this branch (ie. `EvaluateSubscribedViews`).

Is this intentional and correct?

crates/client-api/src/routes/database.rs
```diff
@@ -695,6 +699,18 @@ pub struct PublishDatabaseQueryParams {
...
+    /// If not given, defaults to [default_update_confirmation_timeout].
+    /// The parameter has no effect when creating a new database.
+    update_confirmation_timeout: Option<Duration>,
+}
+
+/// Default timeout for a database update to become confirmed / durable.
+///
+/// Currently, the value is 5s.
+fn default_update_confirmation_timeout() -> Duration {
+    Duration::from_secs(5)
```
gefjon: `Duration::from_secs` is `const`, so this could be replaced with `const DEFAULT_UPDATE_CONFIRMATION_TIMEOUT: Duration = Duration::from_secs(5);`.

crates/client-api/src/routes/database.rs
```diff
@@ -695,6 +699,18 @@ pub struct PublishDatabaseQueryParams {
     parent: Option<NameOrIdentity>,
     #[serde(alias = "org")]
     organization: Option<NameOrIdentity>,
+    /// Duration to wait for a database update to become confirmed (i.e. durable).
+    ///
+    /// If not given, defaults to [default_update_confirmation_timeout].
+    /// The parameter has no effect when creating a new database.
+    update_confirmation_timeout: Option<Duration>,
```
gefjon: Do we want to impose an upper bound on this?

crates/core/src/host/wasm_common/module_host_actor.rs
```diff
@@ -631,14 +630,43 @@ impl InstanceCommon {
...
+                        caller_connection_id: None,
+                        function_call: ModuleFunctionCall::update(),
+                        status: EventStatus::Committed(DatabaseUpdate::default()),
+                        reducer_return_value: None,
+                        energy_quanta_used,
+                        host_execution_duration,
+                        request_id: None,
+                        timer: None,
+                    };
+                    //TODO: Return back event in `UpdateDatabaseResult`?
```
gefjon: Why?

crates/core/src/host/wasm_common/module_host_actor.rs
gefjon: Hm. It seems like there should be some energy cost to performing an update, but clearly that's separate work from this PR. I've created #4847 to track this.

gefjon (approved):

gefjon:
Oh, something I forgot to include in my review: I agree that this is a reasonable behavior, and I don't see the need to make it more configurable than this.

crates/core/src/host/wasm_common/module_host_actor.rs
kim: It was there before this change. @Shubham8287 is this still something we should consider, or can I remove the TODO?

crates/client-api/src/routes/database.rs
kim: I thought it was pleasant to save the closure for `unwrap_or_else`, but I made the fn const.

crates/core/src/host/wasm_common/module_host_actor.rs
Shubham8287: yep, In a way, we also solved the purpose of this TODO by doing `confirmed_reads` on update. Hence, It can be removed

Shubham8287 (approved):
Looks good

PR #4860, opened 2026-04-21
`cargo ci self-docs` uses doc comments as well as explicit helptext

bfops:
# Description of Changes

Several args for `cargo ci` had empty helptext, because we were only printing the explicitly-annotated helptext. This PR updates it so that inline helptext also shows in the README.

None. CI only.

1

# Testing
- [x] Updated README has more helptext


jdetter (approved):
LGTM, not tested by me

PR #4861, opened 2026-04-21
Update Unreal SDK tests to work on Mac

JasonAtClockwork:
# Description of Changes

- Updated the Unreal SDK test harness to allow Unreal Editor to work with MacOS
- Updated the Unreal SDK test handler to work with Nil as it's a special case

N/A

1 - Small changes to the Unreal SDK tests

# Testing

- [x] Ran full suite of tests on Mac for Unreal SDK
- [x] Ran full suite of tests on Windows + Linux for Unreal SDK to confirm no regression


sdks/unreal/tests/sdk_unreal_harness.rs
```diff
@@ -73,13 +102,16 @@ pub fn make_test_with_suite(suite: &TestSuite, test_name: &str) -> Test {
...
     } else {
         format!(
-            "\"{}\" {}Editor Linux Development \"{}\" -skipbuildengine",
-            build_script, suite.unreal_module, uproject_path
+            "\"{}\" {}Editor {} Development -Project=\"{}\" -skipbuildengine",
+            build_script,
+            suite.unreal_module,
+            host_unreal_platform(),
+            uproject_path
         )
```
rekhoff: Just a nit. Because this is now including `MacOS` as an option, and that `host_unreal_platform()` is already checking the Target OS, this could be simplified to something like
```rust
let compile_command = format!(
    "\"{}\" {}Editor {} Development -Project=\"{}\" {}-skipbuildengine",
    build_script,
    suite.unreal_module,
    host_unreal_platform(),
    uproject_path,
    if cfg!(target_os = "windows") { "-waitmutex " } else { "" }
);
```
I'll leave this as optional though, as this could be revised and included in a future update, so another round of testing wasn't required on an otherwise working PR.

rekhoff (commented):
One minor nit, but otherwise looks good to me

sdks/unreal/tests/sdk_unreal_harness.rs
JasonAtClockwork: I've cleaned this up and the `-waitmutex` should really be on both paths and confirmed it's solid on Win/Linux.

rekhoff (approved):
Perfect, with the nit resolved, looks good to me

PR #4863, opened 2026-04-22
commitlog: Handle empty tail segments upon resumption

kim:
In #4338, the read-only path was made resilient against empty segments at the end of the log, but corresponding logic was not applied to re-opening the commitlog for writing.

This patch rectifies that by ignoring and removing segments from the tail of the log if they contain equal to or less than `segment::Header::LEN` bytes.

Additionally, zero-sized segments are eliminated entirely by ensuring that the header is written before moving the segment into place atomically. The benefit of this is not huge, but could simplify commitlog-consuming code by not having to worry about empty (zero-sized) segments. Happy to revert if that is deemed too less of a benefit.

2

# Testing

Adds a test.


kim:
Well I guess #4886 is justification for the atomic-create-with-header part of this patch.

Shubham8287 (approved):
Yeah, let's keep the atomic header writting with new segment, It indeed help us to get rid  of handling empty files.

PR #4868, opened 2026-04-22
CI - Move the DLL updating code into a function

bfops:
# Description of Changes

Just moving code into functions so it can be reused, no functional changes.

None. CI only.

1

# Testing

None


jdetter (approved):
LGTM - not tested by me

PR #4871, opened 2026-04-22
CI - Move simple jobs into `cargo ci`

bfops:
# Description of Changes

Migrate these checks into `cargo ci`:
- Check that packages are publishable
- Docs test
- TypeScript - Tests

None. CI only.

2

# Testing

- [ ] CI passes


.github/workflows/ci.yml
```diff
@@ -1263,76 +1238,38 @@ jobs:
...
-      #     disown
-
-      # - name: Publish module to SpacetimeDB
-      #   working-directory: SpacetimeDB/templates/quickstart-chat-typescript/spacetimedb
+      # # This step shouldn't be needed, but somehow we end up with caches that are missing librusty_v8.a.
+      # # ChatGPT suspects that this could be due to different build invocations using the same target dir,
+      # # and this makes sense to me because we only see it in this job where we mix `cargo build -p` with
+      # # `cargo build --manifest-path` (which apparently build different dependency trees).
+      # # However, we've been unable to fix it so... /shrug
+      # - name: Check v8 outputs
```
bfops, line 1246: I commented out the blocks that build and install SpacetimeDB, since the parts of this job that _used_ SpacetimeDB were already commented out.

bfops: we should probably look into why they're commented out, since it seems like currently we're not really running any tests besides "the things build"

bfops: created https://github.com/clockworklabs/SpacetimeDB/issues/4872 to track

tools/ci/src/main.rs
```diff
@@ -300,6 +306,65 @@ fn tracked_rs_files_under(path: &str) -> Result<Vec<PathBuf>> {
         .collect())
 }
 
+fn run_publish_checks() -> Result<()> {
```
jdetter, line 387: Not saying we have to do it in this PR, but what do you think about replacing this check with the full "release dry-run" check? I think that check is definitely valuable for typescript and C#/Unity as well (the typescript one being the thing that fails most frequently).

.github/workflows/ci.yml
jdetter: Thanks for making the ticket 👍

jdetter (approved):
Just one question on this, I haven't tested anything here but it looks very straightforward 👍

tools/ci/src/main.rs
bfops: we should separately add that check. However (somewhat ironically), the release dry-run skips the cargo crates because `--dry-run` doesn't work trivially. Perhaps we could merge this into that.

bfops: oh actually the release dry run already runs as part of `Internal Tests`

jdetter: So then should we just remove this check? Seems like we're double testing in that case?

bfops: No, because the release dry run check skips the cargo crates. This checks the cargo crates. It would be redundant if and only if we merged this check into the release dry run check.

jdetter: Yeah I forgot about that, the actual solution here I think is to do the "ephemeral local cargo repository" change that we talked about months ago when I originally worked on `cargo release`. This would then allow us to do a dry-run of the cargo crates against that repository instead of the "main" crates.io repository.

In that case I think we should just merge this as-is for now 👍

PR #4874, opened 2026-04-22
CI - `cargo ci update-flow` runs on Windows

bfops:
# Description of Changes

Removed some "if we're on windows" checks in the CI so that we're always running through `cargo ci update-flow`.

None. CI only.

1.

# Testing

- [x] Upgrade flow tests pass


jdetter (approved):
Excellent, thanks Zeke 👍 Have not tested this myself but this is straightforward

PR #4875, opened 2026-04-22
feat: Allow toggling event flag during auto-migration for empty tables

Ludv1gL:
## Summary

Currently, toggling `#[spacetimedb::table(event)]` on an existing table
fails auto-migration with `AutoMigrateError::ChangeTableEventFlag` and
forces a manual migration. This PR allows the flip in either direction
as a live auto-migration step when the table has **zero committed
rows**. Non-empty tables fail with an actionable error telling the user
to clear the table first. Clients are disconnected on the flip because
the change is observable to subscribers (event tables have no committed
state; v2 subscribers see a different message variant; v1 subscribers
stop seeing updates).

## Changes

- **`crates/datastore/src/locking_tx_datastore/tx_state.rs`**: new
  `PendingSchemaChange::TableAlterEventFlag(TableId, bool)` variant
  storing the old `is_event` value for rollback.
- **`crates/datastore/src/locking_tx_datastore/committed_state.rs`**:
  rollback branch restoring the old value on the live schema.
- **`crates/datastore/src/locking_tx_datastore/mut_tx.rs`**: new
  `alter_table_event_flag(table_id, is_event)` mirroring
  `alter_table_access` — dual-writes to the tx + commit table schemas
  via `with_mut_schema_and_clone`, and to `st_event_table` via
  `insert_via_serialize_bsatn` / `delete_col_eq`. Idempotent no-op
  early-returns before pushing a pending change. New
  `delete_st_event_table_row` helper.
- **`crates/datastore/src/locking_tx_datastore/replay.rs`**: new hooks on
  `ST_EVENT_TABLE_ID` insert/delete that flip `is_event` on the
  referenced user-table's cached schema, mirroring
  `reschema_table_for_st_table_update`. This is load-bearing for cold
  replay across the flip — without it, a snapshot predating the flip
  plus commitlog replay would leave the cached schema stale, and
  post-flip inserts would incorrectly land in committed state.
- **`crates/core/src/db/relational_db.rs`**: thin
  `alter_table_event_flag` wrapper analogous to `alter_table_access`.
- **`crates/schema/src/auto_migrate.rs`**: `event_ok` error branch
  replaced with `AutoMigrateStep::ChangeEventFlag` +
  `plan.ensure_disconnect_all_users()`. Removed the now-unused
  `AutoMigrateError::ChangeTableEventFlag` variant. Replaced
  `test_change_event_flag_rejected` with three new tests asserting the
  plan shape in both directions and confirming no orphan sub-object
  steps.
- **`crates/schema/src/auto_migrate/formatter.rs` /
  `termcolor_formatter.rs`**: new `format_change_event_flag` +
  `EventFlagChangeInfo` mirroring `format_change_access`.
- **`crates/core/src/db/update.rs`**: new `ChangeEventFlag` handler with
  an O(1) row-count precheck via `table_row_count_mut` before any
  mutation. Fails with a clear message if the table has data.

## Safety

- **Transaction safety**: the precheck (row count) and all three writes
  (`st_event_table` row, tx table schema, commit table schema) run in
  the same `MutTx`. No window for concurrent inserts between check and
  flip.
- **Rollback**: `TableAlterEventFlag` stores the pre-flip flag value so
  failed transactions revert `is_event` on the live schema via the
  existing `rollback_pending_schema_change` path. Idempotent flips do
  not push a pending change and thus require no rollback work.
- **Replay correctness**: the new `st_event_table` reschema hook is the
  piece with no prior analog. `st_event_table` was already in the
  commitlog, but the existing reschema path only covered `st_table` and
  `st_column`, so flipping `is_event` mid-life would have been invisible
  to cold replay. Added hooks in both `replay_insert` and
  `replay_delete_by_rel` call the same helper, reusing the existing
  `Self::read_table_id(row)` pattern (no new unsafe code).
- **Client contract**: flipping `event` changes what subscribers see —
  v1 subscribers stop receiving updates (event tables aren't in default
  subscriptions); v2 subscribers receive a different message variant
  (`TableUpdateRows::EventTable`). `ensure_disconnect_all_users` forces
  reconnection so clients observe a consistent state.

## Example error output

```
Cannot change `event` flag on table `my_table`: table contains data.
Clear the table's rows (e.g. via a reducer) before toggling the
`event` annotation.
```

## Test plan

- [x] `cargo test -p spacetimedb-datastore --features test` — 87 pass,
  including four new `alter_table_event_flag_*` tests covering both
  directions, rollback, and the idempotent no-op case
- [x] `cargo test -p spacetimedb-schema` — 103 pass, including three new
  plan-shape tests (both directions + "no orphan sub-object steps")
- [x] `cargo test -p spacetimedb-core` — 192 pass, including two new
  end-to-end update-execution tests (empty succeeds, non-empty fails)
- [x] `cargo clippy -p spacetimedb-datastore -p spacetimedb-schema -p spacetimedb-core --tests` clean
- [x] All 10 pre-existing event-table tests still pass
- [x] Verified live on a development deployment — flipped the
  `event` flag on an empty user table via `spacetime publish`; clients
  disconnected as expected, subsequent inserts correctly routed to the
  commitlog only.

### DIFF
--- a/crates/core/src/db/update.rs
+++ b/crates/core/src/db/update.rs
@@ -360,7 +360,8 @@ mod test {
         db::relational_db::tests_utils::{begin_mut_tx, insert, TestDB},
         host::module_host::create_table_from_def,
     };
-    use spacetimedb_datastore::locking_tx_datastore::state_view::StateView;
+    use pretty_assertions::assert_matches;
+    use spacetimedb_datastore::locking_tx_datastore::test_helpers::assert_is_event_state;
     use spacetimedb_datastore::locking_tx_datastore::PendingSchemaChange;
     use spacetimedb_lib::db::raw_def::v9::{btree, RawModuleDefV9Builder, TableAccess};
     use spacetimedb_sats::{product, AlgebraicType, AlgebraicType::U64};
@@ -609,13 +610,28 @@ mod test {
         use spacetimedb_lib::db::raw_def::v10::RawModuleDefV10Builder;
 
         let mut builder = RawModuleDefV10Builder::new();
-        let t = builder.build_table_with_new_type("events", [("id", U64)], true);
-        let t = if is_event { t.with_event(true) } else { t };
-        t.finish();
+        builder
+            .build_table_with_new_type("events", [("id", U64)], true)
+            .with_event(is_event)
+            .finish();
         builder
             .finish()
             .try_into()
             .expect("should be a valid v10 module definition")
+    }
+
+    /// Create a non-event `events` table from the schema of `single_event_table_module_v10(false)`
+    /// in a fresh tx, commit it, and return the `TableId`. Leaves the table empty.
+    fn setup_events_table(stdb: &TestDB, module: &ModuleDef) -> anyhow::Result<TableId> {
+        let mut tx = begin_mut_tx(stdb);
+        for def in module.tables() {
+            create_table_from_def(stdb, &mut tx, module, def)?;
+        }
+        let table_id = stdb
+            .table_id_from_name_mut(&tx, "events")?
+            .expect("table should exist");
+        stdb.commit_tx(tx)?;
+        Ok(table_id)
     }
 
     #[test]
@@ -625,25 +641,11 @@ mod test {
 
         let old = single_event_table_module_v10(false);
         let new = single_event_table_module_v10(true);
+        let table_id = setup_events_table(&stdb, &old)?;
 
-        // Create the non-event table.
         let mut tx = begin_mut_tx(&stdb);
-        for def in old.tables() {
-            create_table_from_def(&stdb, &mut tx, &old, def)?;
-        }
-        let table_id = stdb
-            .table_id_from_name_mut(&tx, "events")?
-            .expect("table should exist");
-        assert!(
-            !tx.get_schema(table_id)
-                .map(|s| s.is_event)
-                .expect("schema should exist"),
-            "fresh non-event table should have `is_event=false`"
-        );
-        stdb.commit_tx(tx)?;
+        assert_is_event_state(&tx, table_id, false);
 
-        // Apply the auto-migration flipping `is_event=true` on an empty table.
-        let mut tx = begin_mut_tx(&stdb);
         let plan = ponder_migrate(&old, &new)?;
         let res = update_database(&stdb, &mut tx, auth_ctx, plan, &TestLogger)?;
 
@@ -651,18 +653,10 @@ mod test {
             matches!(res, UpdateResult::RequiresClientDisconnect),
             "flipping the `event` flag should disconnect clients"
         );
-        assert!(
-            tx.get_schema(table_id)
-                .map(|s| s.is_event)
-                .expect("schema should exist"),
-            "live schema should reflect the new `is_event=true`"
-        );
-        assert!(
-            tx.pending_schema_changes()
-                .iter()
-                .any(|c| matches!(c, PendingSchemaChange::TableAlterEventFlag(_, false))),
-            "flipping `is_event` should record a TableAlterEventFlag pending change: {:?}",
-            tx.pending_schema_changes()
+        assert_is_event_state(&tx, table_id, true);
+        assert_matches!(
+            tx.pending_schema_changes(),
+            [PendingSchemaChange::TableAlterEventFlag(t, false), ..] if *t == table_id
         );
         Ok(())
     }
@@ -674,40 +668,24 @@ mod test {
 
         let old = single_event_table_module_v10(false);
         let new = single_event_table_module_v10(true);
+        let table_id = setup_events_table(&stdb, &old)?;
 
-        // Create the table, insert a row, then attempt to flip `is_event`.
+        // Insert a row in a separate tx so the pre-flip table state is committed.
         let mut tx = begin_mut_tx(&stdb);
-        for def in old.tables() {
-            create_table_from_def(&stdb, &mut tx, &old, def)?;
-        }
-        let table_id = stdb
-            .table_id_from_name_mut(&tx, "events")?
-            .expect("table should exist");
         insert(&stdb, &mut tx, table_id, &product![42u64])?;
         stdb.commit_tx(tx)?;
 
         let mut tx = begin_mut_tx(&stdb);
         let plan = ponder_migrate(&old, &new)?;
-        let result = update_database(&stdb, &mut tx, auth_ctx, plan, &TestLogger);
-        let err = result
+        let err = update_database(&stdb, &mut tx, auth_ctx, plan, &TestLogger)
             .err()
             .expect("flipping `is_event` on a non-empty table should fail");
         assert!(
             err.to_string().contains("contains data"),
             "error should mention that the table contains data, got: {err}"
         );
-        // The schema should be untouched and there should be no pending changes to roll back.
-        assert!(
-            !tx.get_schema(table_id)
-                .map(|s| s.is_event)
-                .expect("schema should exist"),
-            "failed migration must leave the `is_event` flag unchanged"
-        );
-        assert!(
-            tx.pending_schema_changes().is_empty(),
-            "failed migration should leave no pending schema changes: {:?}",
-            tx.pending_schema_changes()
-        );
+        assert_is_event_state(&tx, table_id, false);
+        assert_eq!(tx.pending_schema_changes(), []);
         Ok(())
     }
 }
--- a/crates/core/src/db/update.rs
+++ b/crates/core/src/db/update.rs
@@ -361,7 +361,9 @@ mod test {
         host::module_host::create_table_from_def,
     };
     use pretty_assertions::assert_matches;
-    use spacetimedb_datastore::locking_tx_datastore::test_helpers::assert_is_event_state;
+    use spacetimedb_datastore::locking_tx_datastore::test_helpers::{
+        assert_is_event_state, check_table_event_flag_altered,
+    };
     use spacetimedb_datastore::locking_tx_datastore::PendingSchemaChange;
     use spacetimedb_lib::db::raw_def::v9::{btree, RawModuleDefV9Builder, TableAccess};
     use spacetimedb_sats::{product, AlgebraicType, AlgebraicType::U64};
@@ -654,10 +656,7 @@ mod test {
             "flipping the `event` flag should disconnect clients"
         );
         assert_is_event_state(&tx, table_id, true);
-        assert_matches!(
-            tx.pending_schema_changes(),
-            [PendingSchemaChange::TableAlterEventFlag(t, false), ..] if *t == table_id
-        );
+        check_table_event_flag_altered(&tx, table_id, true);
         Ok(())
     }
 
--- a/crates/core/src/db/update.rs
+++ b/crates/core/src/db/update.rs
@@ -282,6 +282,27 @@ fn auto_migrate_database(
                 let table_def = plan.new.stored_in_table_def(&table_name.clone().into()).unwrap();
                 stdb.alter_table_access(tx, table_name, table_def.table_access.into())?;
             }
+            spacetimedb_schema::auto_migrate::AutoMigrateStep::ChangeEventFlag(table_name) => {
+                let table_def: &TableDef = plan.new.expect_lookup(table_name);
+                let table_id = stdb
+                    .table_id_from_name_mut(tx, table_name)?
+                    .expect("ChangeEventFlag references a table that should exist");
+
+                // Pre-validate: flipping is only safe when the table has no committed rows.
+                if stdb.table_row_count_mut(tx, table_id).unwrap_or(0) > 0 {
+                    anyhow::bail!(
+                        "Cannot change `event` flag on table `{table_name}`: table contains data. \
+                         Clear the table's rows (e.g. via a reducer) before toggling the `event` annotation."
+                    );
+                }
+
+                log!(
+                    logger,
+                    "Changing `event` flag on table `{table_name}` to `{}`",
+                    table_def.is_event
+                );
+                stdb.alter_table_event_flag(tx, table_name, table_def.is_event)?;
+            }
             spacetimedb_schema::auto_migrate::AutoMigrateStep::ChangePrimaryKey(table_name) => {
                 let table_def = plan.new.stored_in_table_def(&table_name.clone().into()).unwrap();
                 log!(logger, "Changing primary key for table `{table_name}`");
@@ -339,6 +360,7 @@ mod test {
         db::relational_db::tests_utils::{begin_mut_tx, insert, TestDB},
         host::module_host::create_table_from_def,
     };
+    use spacetimedb_datastore::locking_tx_datastore::state_view::StateView;
     use spacetimedb_datastore::locking_tx_datastore::PendingSchemaChange;
     use spacetimedb_lib::db::raw_def::v9::{btree, RawModuleDefV9Builder, TableAccess};
     use spacetimedb_sats::{product, AlgebraicType, AlgebraicType::U64};
@@ -572,6 +594,114 @@ mod test {
         assert!(
             err.to_string().contains("table contains data"),
             "error should mention that the table contains data, got: {err}"
+        );
+        assert!(
+            tx.pending_schema_changes().is_empty(),
+            "failed migration should leave no pending schema changes: {:?}",
+            tx.pending_schema_changes()
+        );
+        Ok(())
+    }
+
+    /// Build a minimal v10 module with a single user table `events` whose
+    /// `is_event` flag matches `is_event`.
+    fn single_event_table_module_v10(is_event: bool) -> ModuleDef {
+        use spacetimedb_lib::db::raw_def::v10::RawModuleDefV10Builder;
+
+        let mut builder = RawModuleDefV10Builder::new();
+        let t = builder.build_table_with_new_type("events", [("id", U64)], true);
+        let t = if is_event { t.with_event(true) } else { t };
+        t.finish();
+        builder
+            .finish()
+            .try_into()
+            .expect("should be a valid v10 module definition")
+    }
+
+    #[test]
+    fn change_event_flag_empty_table_succeeds() -> anyhow::Result<()> {
+        let auth_ctx = AuthCtx::for_testing();
+        let stdb = TestDB::durable()?;
+
+        let old = single_event_table_module_v10(false);
+        let new = single_event_table_module_v10(true);
+
+        // Create the non-event table.
+        let mut tx = begin_mut_tx(&stdb);
+        for def in old.tables() {
+            create_table_from_def(&stdb, &mut tx, &old, def)?;
+        }
+        let table_id = stdb
+            .table_id_from_name_mut(&tx, "events")?
+            .expect("table should exist");
+        assert!(
+            !tx.get_schema(table_id)
+                .map(|s| s.is_event)
+                .expect("schema should exist"),
+            "fresh non-event table should have `is_event=false`"
+        );
+        stdb.commit_tx(tx)?;
+
+        // Apply the auto-migration flipping `is_event=true` on an empty table.
+        let mut tx = begin_mut_tx(&stdb);
+        let plan = ponder_migrate(&old, &new)?;
+        let res = update_database(&stdb, &mut tx, auth_ctx, plan, &TestLogger)?;
+
+        assert!(
+            matches!(res, UpdateResult::RequiresClientDisconnect),
+            "flipping the `event` flag should disconnect clients"
+        );
+        assert!(
+            tx.get_schema(table_id)
+                .map(|s| s.is_event)
+                .expect("schema should exist"),
+            "live schema should reflect the new `is_event=true`"
+        );
+        assert!(
+            tx.pending_schema_changes()
+                .iter()
+                .any(|c| matches!(c, PendingSchemaChange::TableAlterEventFlag(_, false))),
+            "flipping `is_event` should record a TableAlterEventFlag pending change: {:?}",
+            tx.pending_schema_changes()
+        );
+        Ok(())
+    }
+
+    #[test]
+    fn change_event_flag_nonempty_table_fails() -> anyhow::Result<()> {
+        let auth_ctx = AuthCtx::for_testing();
+        let stdb = TestDB::durable()?;
+
+        let old = single_event_table_module_v10(false);
+        let new = single_event_table_module_v10(true);
+
+        // Create the table, insert a row, then attempt to flip `is_event`.
+        let mut tx = begin_mut_tx(&stdb);
+        for def in old.tables() {
+            create_table_from_def(&stdb, &mut tx, &old, def)?;
+        }
+        let table_id = stdb
+            .table_id_from_name_mut(&tx, "events")?
+            .expect("table should exist");
+        insert(&stdb, &mut tx, table_id, &product![42u64])?;
+        stdb.commit_tx(tx)?;
+
+        let mut tx = begin_mut_tx(&stdb);
+        let plan = ponder_migrate(&old, &new)?;
+        let result = update_database(&stdb, &mut tx, auth_ctx, plan, &TestLogger);
+        let err = result
+            .err()
+            .expect("flipping `is_event` on a non-empty table should fail");
+        assert!(
+            err.to_string().contains("contains data"),
+            "error should mention that the table contains data, got: {err}"
+        );
+        // The schema should be untouched and there should be no pending changes to roll back.
+        assert!(
+            !tx.get_schema(table_id)
+                .map(|s| s.is_event)
+                .expect("schema should exist"),
+            "failed migration must leave the `is_event` flag unchanged"
         );
         assert!(
             tx.pending_schema_changes().is_empty(),
--- a/crates/datastore/src/locking_tx_datastore/committed_state.rs
+++ b/crates/datastore/src/locking_tx_datastore/committed_state.rs
@@ -734,6 +734,11 @@ impl CommittedState {
                 let table = self.tables.get_mut(&table_id)?;
                 table.with_mut_schema(|s| s.table_access = access);
             }
+            // A table's `is_event` flag was changed. Change back to the old one.
+            TableAlterEventFlag(table_id, old_is_event) => {
+                let table = self.tables.get_mut(&table_id)?;
+                table.with_mut_schema(|s| s.is_event = old_is_event);
+            }
             // A table's primary key was changed. Change back to the old one.
             TableAlterPrimaryKey(table_id, old_pk) => {
                 let table = self.tables.get_mut(&table_id)?;
--- a/crates/datastore/src/locking_tx_datastore/committed_state.rs
+++ b/crates/datastore/src/locking_tx_datastore/committed_state.rs
@@ -737,6 +737,7 @@ impl CommittedState {
             // A table's `is_event` flag was changed. Change back to the old one.
             TableAlterEventFlag(table_id, old_is_event) => {
                 let table = self.tables.get_mut(&table_id)?;
+                assert_eq!(table.num_rows(), 0);
                 table.with_mut_schema(|s| s.is_event = old_is_event);
             }
             // A table's primary key was changed. Change back to the old one.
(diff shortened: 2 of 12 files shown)


crates/schema/src/auto_migrate.rs
```diff
@@ -2405,17 +2407,75 @@ mod tests {
...
+            builder
+                .build_table_with_new_type("events", ProductType::from([("id", AlgebraicType::U64)]), true)
+                .with_event(true)
+                .finish();
+        });
+        let new = create_v10_module_def(|builder| {
+            builder
+                .build_table_with_new_type("events", ProductType::from([("id", AlgebraicType::U64)]), true)
+                .finish();
+        });
```
Centril: The only difference is the `bool`, so please make a closure out out of this, called twice, that takes a parameter `is_event` where `is_event: bool.

crates/schema/src/auto_migrate.rs
```diff
@@ -2405,17 +2407,75 @@ mod tests {
...
+        let new = create_v10_module_def(|builder| {
+            builder
+                .build_table_with_new_type("events", ProductType::from([("id", AlgebraicType::U64)]), true)
+                .with_unique_constraint(0)
+                .with_index(btree(0), "events_id_idx", "events_id_idx")
+                .with_primary_key(0)
+                .with_event(true)
+                .finish();
+        });
```
Centril: Only difference here is the parameter to `.with_event(is_event)` with an implicit `.with_event(false)` in the `old` case. Let's introduce a closure here too with a single parameter `is_event`.

crates/datastore/src/locking_tx_datastore/replay.rs
```diff
@@ -589,6 +598,19 @@ impl<'cs> ReplayCommittedState<'cs> {
         Ok(())
     }
 
+    /// Update the in-memory table structure's `is_event` flag in response to
+    /// replay of an `st_event_table` mutation.
+    fn reschema_table_for_st_event_table_update(&mut self, table_id: TableId, is_event: bool) {
+        // We only need to update if we've already constructed the in-memory table structure.
+        // If we haven't yet, then `self.get_table_and_blob_store_or_create` will see the correct schema
+        // (via `schema_for_table_raw`'s live `st_event_table` lookup) when it eventually runs.
+        if let Ok((table, ..)) = self.get_table_and_blob_store_mut(table_id) {
```
Centril, line 824:
```suggestion
        if let Ok((table, ..)) = self.get_table_and_blob_store_mut(table_id) {
            assert_eq!(table.num_rows(), 0);
```

crates/core/src/db/update.rs
```diff
@@ -282,6 +282,27 @@ fn auto_migrate_database(
...
+                    .expect("ChangeEventFlag references a table that should exist");
+
+                // Pre-validate: flipping is only safe when the table has no committed rows.
+                if stdb.table_row_count_mut(tx, table_id).unwrap_or(0) > 0 {
+                    anyhow::bail!(
+                        "Cannot change `event` flag on table `{table_name}`: table contains data. \
+                         Clear the table's rows (e.g. via a reducer) before toggling the `event` annotation."
+                    );
+                }
+
```
Centril, line 298: Hmm, feels like this should really be a `AutoMigratePrecheck::CheckTableEmpty(...)` so that we do all validation before we do any mutations.

crates/core/src/db/update.rs
```diff
@@ -580,4 +602,112 @@ mod test {
...
+    /// Build a minimal v10 module with a single user table `events` whose
+    /// `is_event` flag matches `is_event`.
+    fn single_event_table_module_v10(is_event: bool) -> ModuleDef {
+        use spacetimedb_lib::db::raw_def::v10::RawModuleDefV10Builder;
+
+        let mut builder = RawModuleDefV10Builder::new();
+        let t = builder.build_table_with_new_type("events", [("id", U64)], true);
+        let t = if is_event { t.with_event(true) } else { t };
+        t.finish();
```
Centril:
```suggestion
        builder.build_table_with_new_type("events", [("id", U64)], true)
            .with_event(is_event)
            .finish();
```

crates/core/src/db/update.rs
```diff
@@ -580,4 +602,112 @@ mod test {
...
+            "flipping the `event` flag should disconnect clients"
+        );
+        assert!(
+            tx.get_schema(table_id)
+                .map(|s| s.is_event)
+                .expect("schema should exist"),
+            "live schema should reflect the new `is_event=true`"
+        );
+        assert!(
+            tx.pending_schema_changes()
```
Centril: Please assert a list of changes `[a, ..]` rather than using `.any(...)`.

crates/core/src/db/update.rs
```diff
@@ -580,4 +602,112 @@ mod test {
...
+        let new = single_event_table_module_v10(true);
+
+        // Create the table, insert a row, then attempt to flip `is_event`.
+        let mut tx = begin_mut_tx(&stdb);
+        for def in old.tables() {
+            create_table_from_def(&stdb, &mut tx, &old, def)?;
+        }
+        let table_id = stdb
+            .table_id_from_name_mut(&tx, "events")?
+            .expect("table should exist");
```
Centril: This could be shared with the previous test, returning the `TableId`. The `insert` bit can be done in a separate tx.

crates/core/src/db/update.rs
```diff
@@ -580,4 +602,112 @@ mod test {
...
+            !tx.get_schema(table_id)
+                .map(|s| s.is_event)
+                .expect("schema should exist"),
+            "failed migration must leave the `is_event` flag unchanged"
+        );
+        assert!(
+            tx.pending_schema_changes().is_empty(),
+            "failed migration should leave no pending schema changes: {:?}",
+            tx.pending_schema_changes()
+        );
```
Centril:
```suggestion
        assert_eq!(tx.pending_schema_changes(), []);
```
This is achieves the same and is less complex.

crates/core/src/db/update.rs
```diff
@@ -580,4 +602,112 @@ mod test {
...
+            err.to_string().contains("contains data"),
+            "error should mention that the table contains data, got: {err}"
+        );
+        // The schema should be untouched and there should be no pending changes to roll back.
+        assert!(
+            !tx.get_schema(table_id)
+                .map(|s| s.is_event)
+                .expect("schema should exist"),
+            "failed migration must leave the `is_event` flag unchanged"
+        );
```
Centril: Please make a function `assert_is_event_state(&tx, table_id, is_event: bool);` and reuse in all of the newly added tests.

crates/datastore/src/locking_tx_datastore/committed_state.rs
```diff
@@ -921,6 +921,11 @@ impl CommittedState {
                 let table = self.tables.get_mut(&table_id)?;
                 table.with_mut_schema(|s| s.table_access = access);
             }
+            // A table's `is_event` flag was changed. Change back to the old one.
+            TableAlterEventFlag(table_id, old_is_event) => {
+                let table = self.tables.get_mut(&table_id)?;
```
Centril, line 739:
```suggestion
                let table = self.tables.get_mut(&table_id)?;
                assert_eq!(table.num_rows(), 0);
```

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -1082,6 +1082,44 @@ impl MutTxId {
...
+        if old_is_event == is_event {
+            // Idempotent no-op; do not record a pending change or it would confuse rollback.
+            return Ok(());
+        }
+        tx_table.with_mut_schema_and_clone(commit_table, |s| s.is_event = is_event);
+
+        // Update `st_event_table`.
+        if is_event {
+            let row = StEventTableRow { table_id };
+            self.insert_via_serialize_bsatn(ST_EVENT_TABLE_ID, &row)?;
```
Centril: Can you make a method out of this and share it with `fn create_table`?

crates/datastore/src/locking_tx_datastore/datastore.rs
```diff
@@ -3181,6 +3189,166 @@ pub(crate) mod tests {
         Ok(())
     }
 
+    /// Returns whether `st_event_table` contains a row referencing `table_id`.
+    fn st_event_table_has_row(datastore: &Locking, tx: &MutTxId, table_id: TableId) -> bool {
+        datastore
+            .iter_by_col_eq_mut_tx(
+                tx,
+                ST_EVENT_TABLE_ID,
+                ColList::from(StEventTableFields::TableId.col_id()),
```
```suggestion
                StEventTableFields::TableId
```

crates/datastore/src/locking_tx_datastore/datastore.rs
```diff
@@ -3181,6 +3189,166 @@ pub(crate) mod tests {
...
+        commit(&datastore, tx)?;
+
+        // Flip `is_event` from `false` to `true`.
+        let mut tx = begin_mut_tx(&datastore);
+        assert_eq!(
+            tx.get_schema(table_id)
+                .map(|s| s.is_event)
+                .expect("schema should exist"),
+            false
+        );
```
Centril: Here again is the test helper function I asked about for `update.rs`. The function should probably live in `crates/datastore` and be used in `update.rs` too.

crates/datastore/src/locking_tx_datastore/datastore.rs
```diff
@@ -3181,6 +3189,166 @@ pub(crate) mod tests {
...
+        assert!(
+            !st_event_table_has_row(&datastore, &tx, table_id),
+            "fresh non-event table must not have a row in `st_event_table`"
+        );
+
+        tx.alter_table_event_flag(table_id, true)?;
+        assert_matches!(
+            tx.pending_schema_changes(),
+            &[PendingSchemaChange::TableAlterEventFlag(t, false)] if t == table_id
+        );
```
Centril: Repeated 3x; Let's make a function out of this: `check_table_event_flag_altered(&tx, table_id, state: bool)`

crates/datastore/src/locking_tx_datastore/datastore.rs
```diff
@@ -3181,6 +3189,166 @@ pub(crate) mod tests {
...
+            "after flipping within the tx, `st_event_table` should have the row"
+        );
+        let _ = datastore.rollback_mut_tx(tx);
+
+        // After rollback, the schema and `st_event_table` should be back to pre-state.
+        let tx = begin_mut_tx(&datastore);
+        assert!(
+            tx.pending_schema_changes().is_empty(),
+            "rollback should clear pending schema changes"
+        );
```
```suggestion
        assert_eq!(tx.pending_schema_changes(), []);
```

crates/datastore/src/locking_tx_datastore/datastore.rs
```diff
@@ -3181,6 +3189,166 @@ pub(crate) mod tests {
...
+                ColList::from(StEventTableFields::TableId.col_id()),
+                &table_id.into(),
+            )
+            .expect("st_event_table lookup should succeed")
+            .next()
+            .is_some()
+    }
+
+    #[test]
+    fn test_alter_table_event_flag_non_event_to_event() -> ResultTest<()> {
```
Centril, line 3157: In these tests, I'd like to see some assertions about what happens to `TxData` when committing changes that successfully altered the system tables.

crates/datastore/src/locking_tx_datastore/datastore.rs
```diff
@@ -3181,6 +3189,166 @@ pub(crate) mod tests {
...
+    #[test]
+    fn test_alter_table_event_flag_idempotent_no_pending_change() -> ResultTest<()> {
+        // Create a non-event table.
+        let (datastore, tx, table_id) = setup_table()?;
+        commit(&datastore, tx)?;
+
+        // Flipping to the same value must be a no-op with no pending change.
+        let mut tx = begin_mut_tx(&datastore);
+        tx.alter_table_event_flag(table_id, false)?;
+        assert_matches!(tx.pending_schema_changes(), &[]);
```
```suggestion
        assert_eq!(tx.pending_schema_changes(), []);
```

Centril (changes requested):

Ludv1gL:
Thanks for the review! Addressed all 16 actionable items — the deferred `AutoMigratePrecheck::CheckTableEmpty` refactor (your `update.rs:298` comment) is sitting out of scope for this PR per our offline agreement and will follow in a separate PR.

Rebased onto current `upstream/master` (includes #4850 replay extraction, #4666 primary-key migration, etc.). The branch is now two commits:

| SHA | Subject |
|---|---|
| `667dfd5401` | `feat: Allow toggling event flag during auto-migration for empty tables` (the original, re-based) |
| `ead511c3f4` | `refactor(event-flag): apply Centril 2026-04-23 review feedback` (this round) |

### Summary of changes in `ead511c3f4`

- **`schema/auto_migrate.rs`** — merged the two `test_change_event_flag_produces_step_*` tests into one that calls a `build = |is_event: bool| ...` closure twice. `test_change_event_flag_does_not_produce_orphan_sub_object_steps` uses the same closure pattern with an explicit `.with_event(is_event)` in both branches (no more implicit-`false` asymmetry).
- **`datastore/replay.rs`** — `assert_eq!(table.num_rows(), 0)` inside the `reschema_table_for_st_event_table_update` hook. Enforces the feature's empty-table invariant on cold replay.
- **`datastore/committed_state.rs`** — same assertion on the `TableAlterEventFlag` rollback arm.
- **`datastore/mut_tx.rs`** — extracted `insert_st_event_table_row(table_id)` shared by `create_table` and `alter_table_event_flag`.
- **`datastore/locking_tx_datastore/test_helpers.rs`** (new) — `assert_is_event_state`, `st_event_table_has_row`, `check_table_event_flag_altered`. Gated by `#[cfg(any(test, feature = "test"))]`, re-exported from `locking_tx_datastore::mod` so `spacetimedb-core`'s test module can import it.
- **`datastore/datastore.rs`** — the 4 `test_alter_table_event_flag_*` tests now use `check_table_event_flag_altered` (was the 3× repeated inline pattern). Successful commits now assert `TxData` reflects exactly the `st_event_table` insert/delete and leaves the user-table row data untouched. `is_empty()` / `assert_matches!(&[])` → `assert_eq!(tx.pending_schema_changes(), [])`.
- **`core/update.rs`** — cleaner v10 builder (`.with_event(is_event)` chain); extracted `setup_events_table()` returning `TableId`; insert moved to a separate tx in the non-empty-fails test; `.any(|c| matches!(...))` → `assert_matches!([pat, ..])`; `.is_empty()` → `assert_eq!([])`; adopt the shared `assert_is_event_state` helper.

### Validation

- `cargo test -p spacetimedb-datastore --features test --lib` — 87 pass (incl. all 4 `test_alter_table_event_flag_*`)
- `cargo test -p spacetimedb-schema` — 102 pass (incl. 3 refactored event-flag tests)
- `cargo test -p spacetimedb-core --lib -- update` — 16 pass (incl. both `change_event_flag_*`)
- `cargo clippy --tests` on touched crates — clean
- `cargo build -j32 -p spacetimedb-standalone --release` — clean

Per-thread replies incoming with the specific commit SHA for each item.

crates/schema/src/auto_migrate.rs
Ludv1gL: Done in `ead511c3f4`. Merged the two `test_change_event_flag_produces_step_*` tests into one `test_change_event_flag_produces_step` that defines a `build = |is_event: bool| ...` closure and an `assert_flip = |old_is_event, new_is_event| ...` closure, then calls `assert_flip(false, true)` and `assert_flip(true, false)`.

crates/schema/src/auto_migrate.rs
Ludv1gL: Done in `ead511c3f4`. `test_change_event_flag_does_not_produce_orphan_sub_object_steps` now uses the same `build = |is_event: bool| ...` closure pattern, with an explicit `.with_event(is_event)` in both the old and new branches — no more implicit-`false` asymmetry.

crates/datastore/src/locking_tx_datastore/replay.rs
Ludv1gL: Applied in `ead511c3f4`.

crates/core/src/db/update.rs
Ludv1gL: Acknowledged — per our offline discussion, deferring this to a follow-up PR that adds `AutoMigratePrecheck::CheckTableEmpty` and moves the row-count check into the precheck pass alongside the existing `CheckAddSequenceRangeValid` etc.

crates/core/src/db/update.rs
Ludv1gL: Done in `ead511c3f4`: `single_event_table_module_v10` now chains `.build_table_with_new_type(...).with_event(is_event).finish()` directly, no more intermediate `if is_event { ... } else { ... }` rebind.

crates/core/src/db/update.rs
Ludv1gL: Done in `ead511c3f4`: replaced with `assert_matches!(tx.pending_schema_changes(), [PendingSchemaChange::TableAlterEventFlag(t, false), ..] if *t == table_id)`.

crates/core/src/db/update.rs
Ludv1gL: Done in `ead511c3f4`: extracted `setup_events_table(&stdb, &module) -> anyhow::Result<TableId>` that creates the table in a fresh tx and commits. Both `change_event_flag_empty_table_succeeds` and `change_event_flag_nonempty_table_fails` call it. The non-empty test then opens a separate tx to insert the row.

crates/core/src/db/update.rs
Ludv1gL: Done in `ead511c3f4`.

crates/core/src/db/update.rs
Ludv1gL: Done in `ead511c3f4`. The helper lives in `crates/datastore/src/locking_tx_datastore/test_helpers.rs` (new module, gated by the existing `test` feature) and is re-exported from `locking_tx_datastore::mod`. Both `update.rs` tests and the `datastore.rs` tests import it via `use spacetimedb_datastore::locking_tx_datastore::test_helpers::assert_is_event_state;`.

crates/datastore/src/locking_tx_datastore/committed_state.rs
Ludv1gL: Applied in `ead511c3f4`.

crates/datastore/src/locking_tx_datastore/mut_tx.rs
Ludv1gL: Done in `ead511c3f4`: new private `MutTxId::insert_st_event_table_row(table_id)` — paired with the existing `delete_st_event_table_row` — called by both `create_table` and `alter_table_event_flag`.

crates/datastore/src/locking_tx_datastore/datastore.rs
Ludv1gL: Done in `ead511c3f4` (moved into the test helper where this lookup now lives — `crates/datastore/src/locking_tx_datastore/test_helpers.rs:st_event_table_has_row`).

crates/datastore/src/locking_tx_datastore/datastore.rs
Ludv1gL: Done in `ead511c3f4`. The helpers live in the new `crates/datastore/src/locking_tx_datastore/test_helpers.rs` module (pub, gated by `#[cfg(any(test, feature = "test"))]`, re-exported from `locking_tx_datastore::mod`). `spacetimedb-core`'s `update.rs` tests import them via the existing `features = ["test"]` dev-dep already in Cargo.toml.

crates/datastore/src/locking_tx_datastore/datastore.rs
Ludv1gL: Done in `ead511c3f4`. The new `check_table_event_flag_altered(datastore, tx, table_id, expected_is_event)` helper in `test_helpers.rs` bundles the schema `is_event` check with the `st_event_table` row-presence check. The 4 `test_alter_table_event_flag_*` tests each call it in lieu of the inline pattern.

crates/datastore/src/locking_tx_datastore/datastore.rs
Ludv1gL: Done in `ead511c3f4`.

crates/datastore/src/locking_tx_datastore/datastore.rs
Ludv1gL: Done in `ead511c3f4`. `test_alter_table_event_flag_non_event_to_event` now captures `TxData` from `commit(&datastore, tx)?` and asserts `tx_data.inserts_for_table(ST_EVENT_TABLE_ID).map(<[_]>::len) == Some(1)` plus the user-table has `None` inserts and `None` deletes. The event→non test does the symmetric `deletes_for_table` assertion.

crates/datastore/src/locking_tx_datastore/datastore.rs
Ludv1gL: Done in `ead511c3f4`.

crates/core/src/db/update.rs
```diff
@@ -609,60 +610,53 @@ mod test {
...
-        assert!(
-            tx.pending_schema_changes()
-                .iter()
-                .any(|c| matches!(c, PendingSchemaChange::TableAlterEventFlag(_, false))),
-            "flipping `is_event` should record a TableAlterEventFlag pending change: {:?}",
-            tx.pending_schema_changes()
+        assert_is_event_state(&tx, table_id, true);
+        assert_matches!(
+            tx.pending_schema_changes(),
+            [PendingSchemaChange::TableAlterEventFlag(t, false), ..] if *t == table_id
```
Centril: We can use `assert_eq!` here instead, it's simpler.

crates/datastore/src/locking_tx_datastore/datastore.rs
Centril: This doesn't seem like it was actually done.

crates/datastore/src/locking_tx_datastore/datastore.rs
```diff
@@ -3149,163 +3149,102 @@ pub(crate) mod tests {
...
-                &table_id.into(),
-            )
-            .expect("st_event_table lookup should succeed")
-            .next()
-            .is_some()
-    }
-
     #[test]
     fn test_alter_table_event_flag_non_event_to_event() -> ResultTest<()> {
-        // Create a non-event table.
```
Centril: Comment was useful, please bring it back.

crates/datastore/src/locking_tx_datastore/datastore.rs
```diff
@@ -3149,163 +3149,102 @@ pub(crate) mod tests {
...
+        use crate::locking_tx_datastore::test_helpers::check_table_event_flag_altered;
+
         let (datastore, tx, table_id) = setup_table()?;
         commit(&datastore, tx)?;
 
-        // Flip `is_event` from `false` to `true`.
```
Centril: Comment was useful, please bring it back.

crates/datastore/src/locking_tx_datastore/datastore.rs
Centril: By this, I meant a function that checks `pending_schema_changes`. This function should be named `check_table_event_flag_altered`. The one you made in `test_helpers` doesn't do what its name suggests.

crates/datastore/src/locking_tx_datastore/datastore.rs
```diff
@@ -3149,163 +3149,102 @@ pub(crate) mod tests {
...
-            tx.get_schema(table_id)
-                .map(|s| s.is_event)
-                .expect("schema should exist"),
-            true
-        );
-        assert!(
-            st_event_table_has_row(&datastore, &tx, table_id),
-            "after flipping to event, `st_event_table` should have the row"
+            tx_data.inserts_for_table(ST_EVENT_TABLE_ID).map(<[_]>::len),
+            Some(1),
```
Centril: Please assert the contents of the row too.

crates/datastore/src/locking_tx_datastore/datastore.rs
```diff
@@ -3149,163 +3149,102 @@ pub(crate) mod tests {
...
-                .expect("schema should exist"),
-            true
-        );
-        assert!(
-            st_event_table_has_row(&datastore, &tx, table_id),
-            "event table should have a row in `st_event_table`"
-        );
+        check_table_event_flag_altered(&datastore, &tx, table_id, true);
 
-        // Flip `is_event` from `true` to `false`.
```
Centril: Useful, please keep it.

crates/datastore/src/locking_tx_datastore/datastore.rs
```diff
@@ -3149,163 +3149,102 @@ pub(crate) mod tests {
...
-            tx.get_schema(table_id)
-                .map(|s| s.is_event)
-                .expect("schema should exist"),
-            false
-        );
-        assert!(
-            !st_event_table_has_row(&datastore, &tx, table_id),
-            "after flipping to non-event, `st_event_table` should not have the row"
+            tx_data.deletes_for_table(ST_EVENT_TABLE_ID).map(<[_]>::len),
+            Some(1),
```
Centril: Please assert the row contents here too.

crates/datastore/src/locking_tx_datastore/datastore.rs
```diff
@@ -3149,163 +3149,102 @@ pub(crate) mod tests {
...
-            )
-            .expect("st_event_table lookup should succeed")
-            .next()
-            .is_some()
-    }
-
     #[test]
     fn test_alter_table_event_flag_non_event_to_event() -> ResultTest<()> {
-        // Create a non-event table.
+        use crate::locking_tx_datastore::test_helpers::check_table_event_flag_altered;
```
Centril: Could be imported once for the test module instead of in each function.

Centril (changes requested):
Getting there but still some things I'd like to change.

Ludv1gL:
New commit `fb321ac85b` addresses all 9 items from this round — no other changes, scoped tightly per your feedback.

| Thread | Fix |
|---|---|
| `update.rs:659` (`assert_eq!` simpler) | Replaced `assert_matches!` with a call to the new `check_table_event_flag_altered` helper (uses `assert_eq!` internally), which also addresses your earlier "this function should live in `crates/datastore` and be used in `update.rs` too" from round 2. |
| `datastore.rs:3198` (Outdated — "doesn't seem like it was actually done") | Applied in `test_helpers.rs`: `ColList::from(StEventTableFields::TableId.col_id())` → `StEventTableFields::TableId` (the `cols: impl Into<ColList>` plus blanket `From<C: Into<ColId>>` handle the conversion directly; matches `system_tables.rs:623-625`). |
| `datastore.rs:3168` (useful comment) | `// Create a non-event table.` restored. |
| `datastore.rs:3172` (useful comment) | `// Flip is_event from false to true.` restored. |
| `datastore.rs:3166` (Outdated — helper misnamed) | Replaced: `check_table_event_flag_altered(tx, table_id, state: bool)` now asserts exactly `tx.pending_schema_changes() == [PendingSchemaChange::TableAlterEventFlag(table_id, !state)]` via `assert_eq!`. The previous bundled schema+row helper is gone; sites that need those use `assert_is_event_state` + `st_event_table_has_row` directly. |
| `datastore.rs:3174` (assert row contents) | `inserts_for_table(ST_EVENT_TABLE_ID)` now compared against `Some(&[ProductValue::from(StEventTableRow { table_id })][..])` instead of only `.map(<[_]>::len) == Some(1)`. |
| `datastore.rs:3227` (useful message) | "event table should have a row in `st_event_table`" assertion message restored in the event→non-event test. |
| `datastore.rs:3206` (assert row contents) | Same row-content fix as `:3174` for `deletes_for_table`. |
| `datastore.rs:3154` (import once) | `use …::{assert_is_event_state, check_table_event_flag_altered, st_event_table_has_row};` hoisted to the top of `pub(crate) mod tests` — removed the per-function imports. |

Validated: datastore event-flag tests 4/4 pass, core `change_event_flag_*` tests 2/2 pass, `cargo clippy` clean.

Ludv1gL:
Superseded by #5694, which now generally allows any schema mutation of an empty table. The `is_event` toggle from this PR is folded in there, reworked onto the `AutoMigratePrecheck::CheckTableEmpty` refactor requested in this PR's review (validation before any mutations, instead of the mid-step check here). Closing.

Issue #3784, opened 2025-11-26
2.0: make `spacetime publish` `--yes` argument an enum


jdetter:
Right now the options are a bit confusing for publish:

`--break-clients` can be specified if you want to skip the "are you sure you want to break clients" prompt, but it won't break clients if there isn't a breaking change.

`--yes` responds yes to all prompts except the login prompt

## TODO

We'd like `--yes` to instead be an enum which can accept the following things:

```
--yes all|migrate|break-clients|skip-login
```

`migrate` - responds yes (if applicable) to the prompt: "Are you sure you want to migrate?"
`break-clients` - responds yes (if applicable)  to the prompt: "Are you sure you want to break clients?"
`skip-login` - responds no (if applicable) to the login prompt
`--all` is the same as `migrate|break-clients|skip-login`

multiple options can be specified, like:

```
# skip prompts for migrations and login but not for breaking clients
spacetime publish --yes "migrate|skip-login"
```

PR #4885, opened 2026-04-23
`spacetime publish --yes` now takes optional arguments

jdetter:
# Description of Changes

This was prompted from another request from the discord. Conversation context: https://discord.com/channels/1037340874172014652/1138987509834059867/1496964407278698656

We've had the ticket to make `--yes` take an enum value: https://github.com/clockworklabs/spacetimedb/issues/3784 . Since we can do this in a non-API/ABI breaking way we're just implementing this ticket.

*Disclaimer: I used claude to write ~90% of this PR.*

None. `--yes` now just takes an optional argument string that we parse, existing CLI commands that use `--yes` will be unaffected.

1 - This is just modifying the `--yes` argument and I've included tests in the PR.

# Testing

- [x] `spacetime publish --yes` still works as expected
- [x] `spacetime publish --yes=all` does the same thing as `spacetime publish --yes`
- [x]  `spacetime publish --yes=remote` only skips asking if publishing to a remote server is ok.
- [x]  `spacetime publish --yes=remote` only skips asking if publishing to a remote server is ok.
- [x] `spacetime publish --yes=migrate|remote` skips both the migrate prompt and the remote server prompt.
- [x] new tests are passing


crates/cli/src/subcommands/publish.rs
```diff
@@ -19,6 +19,67 @@ use crate::util::{add_auth_header_opt, get_auth_header, strip_verbatim_prefix, A
...
+/// Decoded `--yes` selections, used at every prompt site.
+#[derive(Clone, Copy, Debug, Default)]
+pub struct YesFlags {
+    pub remote: bool,
+    pub migrate: bool,
+    pub break_clients: bool,
+    pub skip_login: bool,
+    /// True iff `--yes` was passed with no value or `--yes=all`. Used for destructive prompts
+    /// not represented by an explicit option (e.g. `confirm_and_clear`) to preserve back-compat.
+    pub all: bool,
```
bfops: it seems like this is only used for `confirm_and_clear`, I think this should just be replaced by a `delete_data` field. Otherwise it's a bit confusing what values mean when `all` is true.

crates/cli/src/subcommands/publish.rs
```diff
@@ -19,6 +19,67 @@ use crate::util::{add_auth_header_opt, get_auth_header, strip_verbatim_prefix, A
...
+    /// Skip the "this will BREAK existing clients" confirmation.
+    BreakClients,
+    /// Don't prompt the user to log in; act non-interactively for auth.
+    SkipLogin,
+}
+
+/// Decoded `--yes` selections, used at every prompt site.
+#[derive(Clone, Copy, Debug, Default)]
+pub struct YesFlags {
+    pub remote: bool,
```
bfops: nit:
```suggestion
    pub publish_to_remote: bool,
```

crates/cli/src/subcommands/publish.rs
```diff
@@ -19,6 +19,67 @@ use crate::util::{add_auth_header_opt, get_auth_header, strip_verbatim_prefix, A
...
+    pub migrate: bool,
```
bfops: nit:
```suggestion
    pub migrate_major_version: bool,
```

crates/cli/src/subcommands/publish.rs
```diff
@@ -208,7 +269,21 @@ i.e. only lowercase ASCII letters and numbers, separated by dashes."),
                 .help("The nickname, domain name or URL of the server to host the database."),
         )
         .arg(
-            common_args::yes()
+            Arg::new("yes")
+                .long("yes")
+                .short('y')
+                .num_args(0..=4)
```
bfops: nit: I know I originally proposed the `--yes=foo|bar` syntax, but perhaps we should just support `--yes=foo --yes=bar ...` as well? It seems more canonical, and has the benefit of being able to always append more `--yes` values (e.g. if you're constructing the arg string across multiple callers).

bfops: chatgpt recommends:
```
        .action(ArgAction::Append)
        .num_args(0..=1)
```
and then we can just have an arbitrary number of them accumulating into the vec

crates/cli/src/subcommands/publish.rs
```diff
@@ -208,7 +269,21 @@ i.e. only lowercase ASCII letters and numbers, separated by dashes."),
...
+                .require_equals(true)
+                .value_delimiter('|')
```
bfops: maybe `,` is more familiar than `|` now that I tihnk about it?

crates/cli/src/subcommands/publish.rs
jdetter: true - updated 👍

crates/cli/src/subcommands/publish.rs
jdetter: updated 👍

crates/cli/src/subcommands/publish.rs
jdetter: updated 👍

bfops (approved):
LGTM!

PR #4887, opened 2026-04-23
feat: Multi-module React provider (SpacetimeDBMultiProvider)

Ludv1gL:
# Description of Changes

Adds `<SpacetimeDBMultiProvider>` to the React bindings: mount N labeled module connections under one provider (with nest-to-merge composition for mixing long-lived and short-lived connections), plus a `useSpacetimeDBStatus` hook and label-aware `useTable` / `useReducer` / `useSpacetimeDB`.

**Scope note (2026-07-08):** this PR originally also introduced `ConnectionManager.rebuild()`; that primitive landed separately via #5375 (thanks @krisajenkins for lifting and adapting it). The branch is now merged up to master and no longer touches `sdk/connection_manager.ts` at all — it is provider-only, built on the landed `ConnectionManager` API (`getKey` / `retain` / `release` / `subscribe` / `getSnapshot` / `getConnection`). The extra cases in `connection_manager.test.ts` are regression coverage for multi-connection pooling running against the landed implementation.

None — additive only. Existing single-connection `SpacetimeDBProvider` is untouched.

2 — React bindings only; no server or protocol changes.

# Testing

- [x] 13 multi-provider tests (`tests/multi_provider.test.tsx`): mount/unmount ref-counting, nested merge composition, per-label status, hook re-binding
- [x] Extended `connection_manager.test.ts` multi-connection pooling coverage (41 tests total on the branch)
- [x] Full bindings-typescript suite green after merging master (256 tests)
- [x] Exercised in production by a multi-module web app (10+ labeled module connections per page)


Ludv1gL:
Added a follow-up commit (c9eee82) that makes `<SpacetimeDBMultiProvider>`s **compose by merging** when nested. Motivating use case: apps with a mix of long-lived and short-lived modules (e.g. org-scoped connections that outlive a single project, plus project-scoped connections that swap on project change).

Behavior:
- Outer labels remain visible from an inner subtree (`useSpacetimeDB('label')` resolves through the parent provider even when called inside a child `<SpacetimeDBMultiProvider>`).
- Inner entries shadow outer entries only on label collision.
- Unmounting the inner provider only releases the inner's pool retains; outer retains are untouched — no reliance on the release-then-retain deferred-cleanup trick for the persistent half of the tree.

Covered by four new tests:
- `nested provider: outer labels remain visible from inner subtree`
- `nested provider: inner provider shadows outer label on collision`
- `nested provider: inner unmount only releases inner entries`
- `subset swap: changing one label releases that entry, keeps siblings alive`

Total: 199 tests pass. Lint + build clean.

PR #4889, opened 2026-04-24
[REOPEN]: Wait for database update to become durable

kim:
Re-submit of #4846, which got reverted due to a flying spaghetti monster attack.


gefjon (approved):

kim:
Closing again as it is not picking up the right companion

PR #4897, opened 2026-04-24
Rekhoff/reapply update nativeaot llvm infrastructure

rekhoff:
# Description of Changes

Reapply changes from #4515 after reversion

No API or ABI changes

2 - This PR change itself is trivial, as it just reimplements #4515, however as #4515 had broken the `quickstart` smoketest, this should be considered when reviewing this PR.

# Testing

- [X] Tested against `python3 -m smoketests quickstart` locally


JasonAtClockwork (approved):
LGTM, I ran tests locally and confirmed --native-aot working as intended.

.github/workflows/ci.yml
```diff
@@ -102,6 +102,10 @@ jobs:
           cd $env:USERPROFILE\emsdk
           .\emsdk install 4.0.21
           .\emsdk activate 4.0.21
+          # Add emscripten to PATH for subsequent steps and subprocesses
+          $emsdkPath = "$env:USERPROFILE\emsdk\upstream\emscripten"
+          Add-Content -Path $env:GITHUB_PATH -Value $emsdkPath
+          Write-Host "Added $emsdkPath to PATH"
```
bfops: huh, was this necessary? we at least attempt to do something similar on line 178, and I'm not quite sure why this would only be newly-required on this PR

bfops (approved):
sounds like my code-owned files are the same as they were in the originally-approved PR, so approving.

the tests seem to all pass on this one, so that works for me 👍

.github/workflows/ci.yml
rekhoff: Great point, I agree this is no longer necessary. This was a hold over from some previous troubleshooting with emscripten that never got purged. I missed this one on my clean up pass. It's now been removed.

PR #4900, opened 2026-04-26
Use cli.toml for default start listen address

0monish:
# Description of Changes

This change addresses reviewer feedback on #4576.

As requested by `@cloutiertyler`, here are the changes:

- moved the persistent default listen address for `spacetime start` from project `spacetime.json` to CLI `cli.toml`
- threaded the already-loaded CLI `Config` into `spacetime start` instead of introducing a separate project-config lookup in that subcommand
- preserved precedence so explicit `--listen-addr` / `-l` still wins over config, and config still wins over the built-in standalone default
- documented the setting in the `spacetime start` CLI reference instead of the standalone runtime-config page

Behavior precedence is now:

1. Explicit CLI flag, such as `spacetime start --listen-addr 127.0.0.1:5000`
2. `listen_addr` from `cli.toml`
3. Built-in standalone default `0.0.0.0:3000`

Example:

```toml
listen_addr = "0.0.0.0:4000"
```

With that config in place, `spacetime start` will bind to `0.0.0.0:4000` unless the user passes `--listen-addr` explicitly.

Implementation details:

- added a top-level `listen_addr` field to the existing CLI config parser and persistence path
- updated `spacetime start` to use the loaded CLI config rather than reading project config directly
- added/kept focused tests for `--listen-addr` and `-l` detection so config is only injected when the user did not provide an explicit listen address
- updated the checked-in CLI reference text for `spacetime start`

Related:

- Follow-up to #4576

None.

This change does not modify any public API or ABI. It only changes where the CLI reads an optional default value for an existing `spacetime start` flag.

Complexity: 2/5

This is a small, localized CLI behavior change.

The main interaction point is precedence between forwarded CLI args and persisted CLI config. Risk is low because this change:

- preserves the existing standalone default when `listen_addr` is absent
- preserves explicit CLI `--listen-addr` and `-l` precedence
- reuses the existing `cli.toml` loading path instead of adding another config system

# Testing

Completed:

- [x] `cargo fmt --all --check`
- [x] `cargo check -p spacetimedb-cli`
- [x] `cargo clippy -p spacetimedb-cli --all-targets`
- [x] `cargo test -p spacetimedb-cli config::tests --lib`
- [x] `cargo test -p spacetimedb-cli subcommands::start::tests --lib`

Reviewer checks:

- [ ] Run `spacetime start` without `listen_addr` in `cli.toml` and confirm it still binds to `0.0.0.0:3000`
- [ ] Add `listen_addr = "0.0.0.0:4000"` to `cli.toml`, run `spacetime start`, and confirm it binds to `0.0.0.0:4000`
- [ ] Run `spacetime start --listen-addr 127.0.0.1:5000` with config present and confirm the CLI flag still wins

### DIFF
--- a/crates/cli/src/subcommands/start.rs
+++ b/crates/cli/src/subcommands/start.rs
@@ -46,30 +46,6 @@ enum Edition {
     Cloud,
 }
 
-/// Check whether the forwarded args already contain `--listen-addr` or `-l`.
-///
-/// Handles all common forms:
-/// - `--listen-addr <value>` (two separate tokens)
-/// - `--listen-addr=<value>`
-/// - `-l <value>` (two separate tokens)
-/// - `-l<value>` (short flag with attached value, e.g. `-l0.0.0.0:4000`)
-fn has_listen_addr_arg(args: impl Iterator<Item = impl AsRef<std::ffi::OsStr>>) -> bool {
-    for arg in args {
-        let s = arg.as_ref().to_string_lossy();
-        if s == "--listen-addr" || s.starts_with("--listen-addr=") {
-            return true;
-        }
-        if s == "-l"
-            || (s.starts_with("-l")
-                && !s.starts_with("--")
-                && s.as_bytes().get(2).is_some_and(|b| !b.is_ascii_alphabetic()))
-        {
-            return true;
-        }
-    }
-    false
-}
-
 pub async fn exec(config: Config, paths: &SpacetimePaths, args: &ArgMatches) -> anyhow::Result<ExitCode> {
     let edition = args.get_one::<Edition>("edition").unwrap();
     let forwarded_args: Vec<OsString> = args.get_many::<OsString>("args").unwrap_or_default().cloned().collect();
@@ -85,9 +61,9 @@ pub async fn exec(config: Config, paths: &SpacetimePaths, args: &ArgMatches) -> 
         .arg("--jwt-key-dir")
         .arg(&paths.cli_config_dir);
 
-    if !has_listen_addr_arg(forwarded_args.iter())
-        && let Some(config_addr) = config.start_listen_addr()
-    {
+    if let Some(config_addr) = config.start_listen_addr() {
+        // Prepend the config default; standalone takes the last --listen-addr value,
+        // so an explicit flag in forwarded_args wins.
         cmd.arg("--listen-addr").arg(config_addr);
     }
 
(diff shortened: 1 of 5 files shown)


crates/cli/src/subcommands/start.rs
```diff
@@ -40,9 +46,33 @@ enum Edition {
...
+fn has_listen_addr_arg(args: impl Iterator<Item = impl AsRef<std::ffi::OsStr>>) -> bool {
+    for arg in args {
+        let s = arg.as_ref().to_string_lossy();
+        if s == "--listen-addr" || s.starts_with("--listen-addr=") {
+            return true;
+        }
+        if s == "-l"
+            || (s.starts_with("-l")
+                && !s.starts_with("--")
+                && s.as_bytes().get(2).is_some_and(|b| !b.is_ascii_alphabetic()))
```
bfops: can you explain this block? at first glance it seems like if `s.starts_with("-l")` passes, these other checks would as well.

crates/cli/src/subcommands/start.rs
```diff
@@ -40,9 +46,33 @@ enum Edition {
...
 
-pub async fn exec(paths: &SpacetimePaths, args: &ArgMatches) -> anyhow::Result<ExitCode> {
+/// Check whether the forwarded args already contain `--listen-addr` or `-l`.
+///
+/// Handles all common forms:
+/// - `--listen-addr <value>` (two separate tokens)
+/// - `--listen-addr=<value>`
+/// - `-l <value>` (two separate tokens)
+/// - `-l<value>` (short flag with attached value, e.g. `-l0.0.0.0:4000`)
+fn has_listen_addr_arg(args: impl Iterator<Item = impl AsRef<std::ffi::OsStr>>) -> bool {
```
bfops: this function can be removed entirely. if `--listen-addr` is passed multiple times to `spacetimedb-standalone`, the last one will be respected.

I would add a comment to this effect when passing `--listen-addr`.

crates/cli/src/subcommands/start.rs
0monish: the s == "-l" arm was needed because the compound branch's get(2).is_some_and(...) guard excludes bare -l. That said, the whole function is going away per your next comment, so moot either way.

crates/cli/src/subcommands/start.rs
0monish: Removed has_listen_addr_arg and switched to prepend + last-value-wins behavior. Explicit --listen-addr from CLI still overrides config.

bfops (approved):
This LGTM! Thank you!

bfops:
@0monish what is the reason for re-requesting review?

0monish:
> @0monish what is the reason for re-requesting review?

Apologies, please bear with me, this re-review request was submitted by mistake.

0monish:
> This LGTM! Thank you!

Welcome 🤗

PR #4902, opened 2026-04-27
fix: update seq table when migrating table.

Shubham8287:
# Description of Changes
`MutTxId::add_columns_to_table` creates new table but only copies sequences to in-memory state, which causes `autoinc` columns to reset on restart.

The existing implementation relies on `create_table_and_update_seq` helper, which only updates the sequence state in memory. This change ensures that the `allocation` is also persisted to the system table, keeping it consistent across restarts.

NA

2

# Testing
Added a test, which migrate table and checks for `autoinc` column value without and with restart.

### DIFF
--- a/crates/core/src/db/update.rs
+++ b/crates/core/src/db/update.rs
@@ -586,6 +586,7 @@ mod test {
         let auth_ctx = AuthCtx::for_testing();
         let stdb = TestDB::durable()?;
 
+        // Define the old module that was before.
         let module_v1: ModuleDef = {
             let mut b = RawModuleDefV9Builder::new();
             b.build_table_with_new_type("seq_t", [("id", AlgebraicType::I64)], true)
@@ -596,6 +597,7 @@ mod test {
             b.finish().try_into().expect("valid module v1")
         };
 
+        // Define the module that we're migrating to.
         let module_v2: ModuleDef = {
             let mut b = RawModuleDefV9Builder::new();
             b.build_table_with_new_type(
@@ -604,13 +606,15 @@ mod test {
                 true,
             )
             .with_auto_inc_primary_key(0)
-            .with_index_no_accessor_name(RawIndexAlgorithm::BTree { columns: 0.into() })
+            .with_index_no_accessor_name(btree(0))
             .with_access(TableAccess::Public)
             .with_default_column_value(1, product![0u64].into())
             .finish();
             b.finish().try_into().expect("valid module v2")
         };
 
+        // Create the old tables and insert two rows
+        // that use the auto-inc sequence.
         {
             let mut tx = begin_mut_tx(&stdb);
 
@@ -626,6 +630,7 @@ mod test {
             stdb.commit_tx(tx)?;
         }
 
+        // Successfully update the database to the new module. 
         {
             let mut tx = begin_mut_tx(&stdb);
 
@@ -640,6 +645,8 @@ mod test {
             stdb.commit_tx(tx)?;
         }
 
+        // Check that the new table has reused the sequence
+        // from the old table such that the last row has the value 3.
         {
             let mut tx = begin_mut_tx(&stdb);
 
@@ -648,7 +655,7 @@ mod test {
             insert(&stdb, &mut tx, table_id, &product![0i64, 99u64])?;
 
             let mut ids = stdb
-                .iter_by_col_range_mut(&tx, table_id, 0u16, AlgebraicValue::I64(0)..)?
+                .iter_mut(&tx, table_id)?
                 .map(|r| r.read_col::<i64>(0))
                 .collect::<Result<Vec<_>, _>>()?;
 
@@ -661,8 +668,10 @@ mod test {
             );
         }
 
+        // Check that we can replay.
         let stdb = stdb.reopen()?;
 
+        // After replay, the allocation cursor should be preserved.
         {
             let mut tx = begin_mut_tx(&stdb);
 
--- a/crates/core/src/db/update.rs
+++ b/crates/core/src/db/update.rs
@@ -340,7 +340,7 @@ mod test {
         host::module_host::create_table_from_def,
     };
     use spacetimedb_datastore::locking_tx_datastore::PendingSchemaChange;
-    use spacetimedb_lib::db::raw_def::v9::{btree, RawModuleDefV9Builder, TableAccess};
+    use spacetimedb_lib::db::raw_def::v9::{btree, RawIndexAlgorithm, RawModuleDefV9Builder, TableAccess};
     use spacetimedb_sats::{product, AlgebraicType, AlgebraicType::U64};
     use spacetimedb_schema::{auto_migrate::ponder_migrate, def::ModuleDef};
 
@@ -432,7 +432,7 @@ mod test {
         let stdb = TestDB::durable()?;
 
         // Step 1: Table with a primary key (requires unique constraint + index).
-        let module_v1 = {
+        let module_v1: ModuleDef = {
             let mut builder = RawModuleDefV9Builder::new();
             builder
                 .build_table_with_new_type("person", [("name", AlgebraicType::String)], true)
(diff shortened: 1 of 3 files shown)


crates/core/src/db/update.rs
```diff
@@ -580,4 +580,110 @@ mod test {
...
+
+        let module_v2: ModuleDef = {
+            let mut b = RawModuleDefV9Builder::new();
+            b.build_table_with_new_type(
+                "seq_t",
+                [("id", AlgebraicType::I64), ("payload", AlgebraicType::U64)],
+                true,
+            )
+            .with_auto_inc_primary_key(0)
+            .with_index_no_accessor_name(RawIndexAlgorithm::BTree { columns: 0.into() })
```
Centril:
```suggestion
            .with_index_no_accessor_name(btree(0))
```

crates/core/src/db/update.rs
```diff
@@ -580,4 +580,110 @@ mod test {
         );
         Ok(())
     }
+
+    #[test]
+    fn auto_inc_sequence_survives_add_column_migration() -> anyhow::Result<()> {
```
Centril, line 593: Can you add some "paragraph" commentary to this test?

crates/core/src/db/update.rs
```diff
@@ -580,4 +580,110 @@ mod test {
...
+        let auth_ctx = AuthCtx::for_testing();
+        let stdb = TestDB::durable()?;
+
+        let module_v1: ModuleDef = {
```
Centril, line 598:
```suggestion
        // Define the old module that was before.
        let module_v1: ModuleDef = {
```

crates/core/src/db/update.rs
```diff
@@ -580,4 +580,110 @@ mod test {
...
+            let mut b = RawModuleDefV9Builder::new();
+            b.build_table_with_new_type("seq_t", [("id", AlgebraicType::I64)], true)
+                .with_auto_inc_primary_key(0)
+                .with_index_no_accessor_name(RawIndexAlgorithm::BTree { columns: 0.into() })
+                .with_access(TableAccess::Public)
+                .finish();
+            b.finish().try_into().expect("valid module v1")
+        };
+
+        let module_v2: ModuleDef = {
```
Centril, line 609:
```suggestion
        // Define the module that we're migrating to.
        let module_v2: ModuleDef = {
```

crates/core/src/db/update.rs
```diff
@@ -580,4 +580,110 @@ mod test {
...
+            )
+            .with_auto_inc_primary_key(0)
+            .with_index_no_accessor_name(RawIndexAlgorithm::BTree { columns: 0.into() })
+            .with_access(TableAccess::Public)
+            .with_default_column_value(1, product![0u64].into())
+            .finish();
+            b.finish().try_into().expect("valid module v2")
+        };
+
+        {
```
Centril, line 643:
```suggestion
        // Create the old tables and insert two rows
        // that use the auto-inc sequence.
        {
```

crates/core/src/db/update.rs
```diff
@@ -580,4 +580,110 @@ mod test {
...
+
+            let table_id = stdb.table_id_from_name_mut(&tx, "seq_t")?.expect("seq_t should exist");
+
+            insert(&stdb, &mut tx, table_id, &product![0i64])?;
+            insert(&stdb, &mut tx, table_id, &product![0i64])?;
+
+            stdb.commit_tx(tx)?;
+        }
+
+        {
```
Centril, line 659:
```suggestion
        // Successfully update the database to the new module. 
        {
```

crates/core/src/db/update.rs
```diff
@@ -580,4 +580,110 @@ mod test {
...
+            let mut tx = begin_mut_tx(&stdb);
+
+            let table_id = stdb.table_id_from_name_mut(&tx, "seq_t")?.expect("seq_t should exist");
+
+            insert(&stdb, &mut tx, table_id, &product![0i64, 99u64])?;
+
+            let mut ids = stdb
+                .iter_by_col_range_mut(&tx, table_id, 0u16, AlgebraicValue::I64(0)..)?
```
Centril:
```suggestion
                .iter_mut(&tx, table_id)?
```
Being specific about the range/column seems unnecessary here.

crates/core/src/db/update.rs
```diff
@@ -580,4 +580,110 @@ mod test {
...
+
+            assert!(matches!(
+                res,
+                UpdateResult::Success | UpdateResult::RequiresClientDisconnect
+            ));
+
+            stdb.commit_tx(tx)?;
+        }
+
+        {
```
Centril, line 675:
```suggestion
        // Check that the new table has reused the sequence
        // from the old table such that the last row has the value 3.
        {
```

crates/core/src/db/update.rs
```diff
@@ -580,4 +580,110 @@ mod test {
...
+            let table_id = stdb.table_id_from_name_mut(&tx, "seq_t")?.expect("seq_t should exist");
+
+            insert(&stdb, &mut tx, table_id, &product![0i64, 99u64])?;
+
+            let mut ids = stdb
+                .iter_by_col_range_mut(&tx, table_id, 0u16, AlgebraicValue::I64(0)..)?
+                .map(|r| r.read_col::<i64>(0))
+                .collect::<Result<Vec<_>, _>>()?;
+
+            ids.sort();
```
Centril: This is duplicated above; please make a closure out of it.

crates/datastore/src/locking_tx_datastore/mut_tx.rs
```diff
@@ -1015,6 +1015,25 @@ impl MutTxId {
...
+    ) -> Result<R> {
+        let st_sequence_ref = self
+            .iter_by_col_eq(ST_SEQUENCE_ID, StSequenceFields::SequenceId, &sequence_id.into())?
+            .last()
+            .ok_or(SequenceError::NotFound(sequence_id))?;
+        let ptr = st_sequence_ref.pointer();
+        let mut row = StSequenceRow::try_from(st_sequence_ref)?;
+
+        let ret = updater(&mut row);
+        self.delete(ST_SEQUENCE_ID, ptr)?;
```
Centril: Why is the order flipped compared to the method above?

crates/datastore/src/locking_tx_datastore/sequence.rs
```diff
@@ -82,6 +82,17 @@ impl Sequence {
         self.value = new_value;
     }
 
+    /// Update the persisted allocation cursor for the sequence.
+    pub(super) fn update_allocation(&mut self, new_allocated: i128) {
+        if new_allocated < self.schema.min_value || new_allocated > self.schema.max_value {
```
Centril:
```suggestion
        if !(self.schema.min_value..=self.schema.max_value).contains(&new_allocated) {
```

crates/core/src/db/update.rs
```diff
@@ -580,4 +580,110 @@ mod test {
...
+            stdb.commit_tx(tx)?;
+
+            assert!(
+                ids.iter().last().unwrap() == &3,
+                "expected id 3 after migration, got {ids:?}"
+            );
+        }
+
+        let stdb = stdb.reopen()?;
```
Centril, line 684:
```suggestion
        // Check that we can replay.
        let stdb = stdb.reopen()?;
```

crates/core/src/db/update.rs
```diff
@@ -580,4 +580,110 @@ mod test {
...
+
+        {
```
Centril, line 687:
```suggestion
        // After replay, the allocation cursor should be preserved.
        {
```

Centril (changes requested):

crates/datastore/src/locking_tx_datastore/mut_tx.rs
Shubham8287: Doesn't seems to matter but I will make them same.

Centril (approved):

PR #4906, opened 2026-04-27
Windows Binary Signing through DigiCert

rekhoff:
# Description of Changes

Add EV code signing for Windows CLI binaries using DigiCert KeyLocker. The workflow now signs `spacetimedb-update.exe`, `spacetimedb-cli.exe`, and `spacetimedb-standalone.exe` on tag pushes using `smctl sign` with a cloud HSM-backed certificate.

These changes reflect the updated DigiCert guidance for code signing through GitHub found here: https://github.com/marketplace/actions/digicert-binary-signing

No API or ABI changes. This change only affects the CI/CD packaging workflow.

1 - This PR only adds code signing to existing CI packaging. Risk is limited to the Windows packaging step failing on tags; Linux and macOS builds are unaffected.

# Testing

- [X] Tested via workflow dispatch on tag `test-signing-v0.0.1`
- [X] All three executables signed and verified successfully
- [X] Signature verification confirms certificate chain
- [X] Signed artifacts uploaded successfully


.github/workflows/package.yml
```diff
@@ -66,41 +66,55 @@ jobs:
...
-        if: ${{ runner.os == 'Windows' }}
-        shell: powershell
-        env:
-          DIGICERT_CERT_B64: ${{ secrets.DIGICERT_CERT_B64 }}
+      - name: Decode DigiCert client auth certificate
+        if: ${{ runner.os == 'Windows' && startsWith(github.ref, 'refs/tags/') }}
+        shell: bash
         run: |
-          [IO.File]::WriteAllBytes("digicert.pfx", [Convert]::FromBase64String($env:DIGICERT_CERT_B64))
+          echo "${{ secrets.SM_CLIENT_CERT_FILE_B64 }}" | base64 --decode > /d/Certificate_pkcs12.p12
```
jdetter: Don't hardcode `/d/` or `D:\` here, use `$RUNNER_TEMP` instead. Eventually I think we should move this job over to our custom runner and our custom runner does not have a `D:\` drive.

e.g.:
```
  - shell: bash
    env:
      SM_CLIENT_CERT_FILE_B64: ${{ secrets.SM_CLIENT_CERT_FILE_B64 }}
    run: echo "$SM_CLIENT_CERT_FILE_B64" | base64 --decode > "$RUNNER_TEMP/Certificate_pkcs12.p12"
```

jdetter (changes requested):
Only one small issue with this, ping me again when you want me to take another look 👍

.github/workflows/package.yml
jdetter: I actually think this would be a good time to change this from `windows-latest` to `spacetimedb-windows-runner` and make sure that it works there

.github/workflows/package.yml
rekhoff: Updated in [c6b9908](https://github.com/clockworklabs/SpacetimeDB/pull/4906/commits/c6b9908b722053f644401e1c6b7acb8d6eec7ed8)

.github/workflows/package.yml
rekhoff: Updated in [c6b9908](https://github.com/clockworklabs/SpacetimeDB/pull/4906/commits/c6b9908b722053f644401e1c6b7acb8d6eec7ed8)

jdetter (approved):
This is excellent, thank you Ryan 🙏

PR #4915, opened 2026-04-29
Add .NET 10 support via NativeAOT-LLVM along side .NET 8 support

rekhoff:
# Description of changes

Builds on #4741 to clean up and refine the NativeAOT-LLVM integration. The major changes:

- **Standardized the repo on .NET 10 SDK** (`global.json` → `10.0.100`). The team is migrating to .NET 10 ahead of .NET 8 EOL. End users are unaffected as their SDK version is controlled by their own `global.json` emitted by `spacetime init`.

- **Multi-target NuGet packages** instead of conditional single-target:
  - `SpacetimeDB.Runtime` now unconditionally targets `net8.0;net10.0`
  - `SpacetimeDB.BSATN.Runtime` now unconditionally targets `netstandard2.1;net8.0;net10.0`
  - NuGet packages ship both TFMs in a single `.nupkg`. No `EXPERIMENTAL_WASM_AOT` env var needed at pack time
  - `EXPERIMENTAL_WASM_AOT` define constant is set automatically when `TargetFramework == net10.0`
  - `ILCompiler.LLVM` package references conditioned on `TargetFramework == net10.0` (not env var)

- **Centralized NativeAOT-LLVM build logic** in `SpacetimeDB.Runtime.props` and `SpacetimeDB.Runtime.targets`:
  - Moved `PublishTrimmed`, `SelfContained`, `WasmEnableThreads` into `.props` (previously duplicated across 4+ `.csproj` files)
  - Moved the `.wit`-stripping `UseWasiRuntimeOverlayWithoutComponentWit` target into `.targets` (was duplicated in every server `.csproj`)
  - Added `IlcLlvmTarget=wasm32-unknown-wasip1` override after ILCompiler.LLVM.targets import (previously passed as `/p:` arg)
  - Added `_InitializeWasiSdk` to `ObtainWasiSdk` BeforeTargets for .NET 10 compatibility
  - Consumer `.csproj` files no longer need any AOT boilerplate

- **Refactored `csharp.rs` build logic** to support three build paths via `CsharpBuildPath` enum:
  - `Net8Jit` : stable .NET 8 path using `wasi-experimental` workload (Mono WASM)
  - `Net8Aot` : .NET 8 NativeAOT-LLVM (opt-in via `--native-aot` or `EXPERIMENTAL_WASM_AOT=1`)
  - `Net10Aot` : .NET 10 NativeAOT-LLVM (auto-detected when .NET 10 SDK is active)
  - Auto-detects .NET SDK version from `dotnet --version` (respects `global.json`)
  - Automatically sets `EXPERIMENTAL_WASM_AOT=1` for AOT paths so MSBuild conditionals activate
  - Unified `dotnet publish` command for all paths. Build-specific config handled by props/targets
  - Removed redundant `/p:IlcLlvmTarget` and `/p:WasmEnableThreads` CLI args (now in `.targets`)

- **Refactored `init.rs`** for .NET version-aware scaffolding:
  - Added `--dotnet-version` CLI arg to explicitly select .NET 8 or 10
  - Added `resolve_dotnet_major()` which auto-detects or prompts interactively when multiple SDKs are installed
  - .NET 10 auto-enables NativeAOT-LLVM (`--native-aot` not required)
  - .NET 10 template emits `<TargetFramework>net10.0</TargetFramework>` directly (no conditional)
  - .NET 8 AOT template adds `ILCompiler.LLVM 8.0.0-*` package refs gated on `EXPERIMENTAL_WASM_AOT=1`
  - Emits correct `global.json` per SDK version (`8.0.100` or `10.0.100`)

- **Added `_initialize` call in `wasmtime_module.rs`**: NativeAOT-LLVM modules are WASI reactors that export `_initialize` to bootstrap the native runtime. This is called before preinit functions. Traditional .NET 8 WASI modules export `_start` instead and are unaffected.

- **Updated `FFI.cs` `WasmImportLinkageAttribute` guard** to `#if EXPERIMENTAL_WASM_AOT && NET10_0_OR_GREATER`, ensuring the real attribute is only used when both the AOT flag and .NET 10+ TFM are active. The dummy shim is used for all other builds.

- **Updated `csharp_aot_module.rs` smoketest**:
  - Auto-detects .NET SDK version and adjusts TFM accordingly
  - .NET 8 AOT: Windows-only (ILCompiler.LLVM 8.0.0-* not published for Linux)
  - .NET 10 AOT: Windows and Linux
  - Removed emscripten dependency (NativeAOT-LLVM uses WASI SDK, not emscripten)

- **Updated CLI reference docs** to reflect `--native-aot` behavior change (not needed for .NET 10) and new `--dotnet-version` arg.

This PR addresses several issues from #4741 review:

1. **NativeAOT-LLVM build logic was duplicated** in every server `.csproj` (4+ files had identical `PropertyGroup` and `Target` blocks). This is now centralized in the shipped `.props`/`.targets`.

2. **NuGet packages only shipped a single TFM**. The env var controlled which TFM was built, meaning you needed `EXPERIMENTAL_WASM_AOT=1` at pack time to get the `net10.0` DLL. Now both TFMs are always included.

3. **`spacetime init` emitted unnecessary conditional wrappers** in the generated `.csproj` for .NET 10 projects. Since the project is definitively targeting .NET 10, the TFM should be unconditional.

4. **`spacetime publish` passed MSBuild properties via `/p:` args** that are now handled by centralized `.targets`, simplifying the CLI code.

- `global.json` now requires .NET 10 SDK for repo development. All developers and CI agents need .NET SDK 10.0+ installed.
- End user module projects are **not affected**. They continue to use whichever SDK their local `global.json` specifies.
- .NET 8 JIT and .NET 8 AOT paths remain functional for end users until .NET 8 EOL.

2 - The core risk is the same as #4741: future changes in the upstream NativeAOT-LLVM/WASI pipeline. Centralizing the build logic in `.props`/`.targets` reduces the surface area for breaking changes (one place to update vs. N consumer projects). The `.wit`-stripping workaround remains necessary until [dotnet/runtimelab#3144](https://github.com/dotnet/runtimelab/issues/3144) is resolved.

# Testing

- Ran C# module publish + client connection tests for all three build paths:
  - .NET 8 JIT (`spacetime publish` without `--native-aot`)
  - .NET 8 AOT (`spacetime publish` with `EXPERIMENTAL_WASM_AOT=1` and .NET 8 SDK)
  - .NET 10 AOT (`spacetime publish` with .NET 10 SDK, auto-detected)
- Verified `spacetime init --lang csharp` generates correct project structure for both .NET 8 and .NET 10
- Verified `spacetime init --lang csharp --dotnet-version 10` emits `net10.0` TFM and `10.0.100` in `global.json`
- Verified C# regression tests complete successfully when ran under .NET 8 JIT, .NET 8 AOT, .NET 10 AOT
  - This required making local customizations to the repo versions of the regression tests, in order to conform to the configuration output by `spacetime init`, but the tests themselves remained identical.
- Verified NuGet packages contain both `lib/net8.0/` and `lib/net10.0/` DLLs
- Updated smoketest `test_build_csharp_module_aot` to work with both .NET 8 and .NET 10 SDK


docs/docs/00100-intro/00200-quickstarts/00600-c-sharp.md
JasonAtClockwork:
```suggestion
spacetime dev --template basic-cs --dotnet-version 8
```
With the updated documentation I think `--dotnet-version` is required here.

crates/cli/src/subcommands/init.rs
```diff
@@ -150,6 +152,9 @@ impl InitOptions {
             non_interactive: args.get_flag("non-interactive"),
             skip_next_steps: false,
             native_aot: args.get_flag("native-aot"),
+            dotnet_version: args
+                .get_one::<String>("dotnet-version")
+                .and_then(|s| s.parse::<u8>().ok()),
```
JasonAtClockwork: nit: Should we protect here for only 8 or 10 and return an error if not? Or maybe we output with `init` what version is in use so if you use `--dotnet-version 12` it outputs that it used .NET 10?

JasonAtClockwork (approved):
A couple small things but it looks great.

crates/cli/src/subcommands/init.rs
lisandroct: Yeah, we probably should. It's a good point.

docs/docs/00100-intro/00200-quickstarts/00600-c-sharp.md
lisandroct: Yeap, you're right.

bfops:
merging in master to get some new internal tests behavior (https://github.com/clockworklabs/SpacetimeDB/pull/5520)

.github/workflows/ci.yml
```diff
@@ -153,6 +153,26 @@ jobs:
           ./emsdk install 4.0.21
           ./emsdk activate 4.0.21
 
+      # Install WASI SDK for .NET 10 NativeAOT-LLVM compilation.
+      - name: Install WASI SDK (Linux)
```
bfops: optional nit: it would be to factor this (and the windows version) into a separate CI action so that we can just include the "install wasi" action in all the places that need it? I imagine AI can do it for us.

.github/workflows/ci.yml
```diff
@@ -182,9 +225,14 @@ jobs:
           $PSNativeCommandUseErrorActionPreference = $true
 
           cd modules
-          # the sdk-manifests on windows-latest are messed up, so we need to update them
```
bfops, line 185: why was this comment removed?

.github/workflows/ci.yml
```diff
@@ -1075,7 +1142,54 @@ jobs:
...
+            tar -xzf "wasi-sdk-${WASI_SDK_VERSION}.0-x86_64-linux.tar.gz"
+            sudo mv "wasi-sdk-${WASI_SDK_VERSION}.0-x86_64-linux" "$WASI_SDK_PATH"
+            rm -f "wasi-sdk-${WASI_SDK_VERSION}.0-x86_64-linux.tar.gz"
+            echo "WASI SDK installed successfully"
+          fi
+          echo "WASI_SDK_PATH=$WASI_SDK_PATH" >> "$GITHUB_ENV"
+          echo "Using WASI SDK at: $WASI_SDK_PATH"
+
+      # Remove broken global.json symlinks before regression scripts temporarily pin
+      # each C# module directory to the SDK version under test.
```
bfops: wait why do these symlinks get broken?

.github/workflows/ci.yml
```diff
@@ -1170,6 +1286,21 @@ jobs:
       - name: Run regression tests
         run: |
           bash sdks/csharp/tools~/run-regression-tests.sh
+          # Restore global.json symlinks (we replaced them with files to work around .NET 10 SDK bug)
+          # server is 5 levels deep from root, republishing dirs are 6 levels deep
+          if [ -f sdks/csharp/examples~/regression-tests/server/global.json ] && [ ! -L sdks/csharp/examples~/regression-tests/server/global.json ]; then
+            echo "Restoring symlink at server/global.json"
+            rm -f sdks/csharp/examples~/regression-tests/server/global.json
+            ln -s ../../../../../global.json sdks/csharp/examples~/regression-tests/server/global.json
```
bfops, line 1196: optional nit: I wonder if a `git checkout HEAD -- <file>` would be cleaner?

bfops: and/or `git restore` which I think is the modern version

.github/workflows/ci.yml
```diff
@@ -1075,7 +1142,54 @@ jobs:
...
+                     sdks/csharp/examples~/regression-tests/republishing/server-republish; do
+            if [ -L "$dir/global.json" ] && [ ! -f "$dir/global.json" ]; then
+              echo "Fixing broken symlink at $dir/global.json"
+              rm -f "$dir/global.json"
+            fi
+          done
+
       - name: Override NuGet packages
+        env:
+          EXPERIMENTAL_WASM_AOT: 1
```
bfops: why is this the only one that needs this env var set? (we have other `override nuget packages` steps in this file that don't get this override)

crates/cli/src/subcommands/build.rs
```diff
@@ -39,6 +40,12 @@ pub fn cli() -> clap::Command {
                 .action(SetTrue)
                 .help("Builds the module using debug instead of release (intended to speed up local iteration, not recommended for CI)"),
         )
+        .arg(
+            Arg::new("dotnet_version")
+                .long("dotnet-version")
+                .value_name("VERSION")
+                .help("Target .NET SDK major version for C# projects (e.g. 8 or 10). Auto-detected when omitted.")
```
bfops: optional nit: we could probably move this into `common_args.rs` since it's used in several places

crates/cli/src/subcommands/build.rs
```diff
@@ -60,15 +67,18 @@ pub async fn exec(_config: Config, args: &ArgMatches) -> Result<(PathBuf, &'stat
     };
     let build_debug = args.get_flag("debug");
     let features = features.cloned();
+    let dotnet_version = parse_dotnet_version(args.get_one::<String>("dotnet_version").map(String::as_str))?;
```
bfops: nit: perhaps we want to add `.value_parser(parse_dotnet_version)` on the arg itself?

tools/ci/src/main.rs
```diff
@@ -79,7 +79,10 @@ fn check_global_json_policy() -> Result<()> {
         }
 
         let contents = fs::read_to_string(&p)?;
-        if contents != root_contents {
+        // Templates are exempt from content matching to preserve the .NET 8 JIT path for
+        // module developers importing templates, while the main codebase uses .NET 10 AOT.
+        // TODO: Remove this exemption once .NET 10 is the default and templates should use it.
+        if contents != root_contents && !is_template_global_json {
```
bfops: I'm open to pushback, but I would prefer if maybe we had something like `templates/template-global.json` or something and then updated this logic to be that `global.json`s under `templates/` must match that file and everything else must match the root file. that way we still prevent drift between the `global.json` files.

crates/cli/src/subcommands/mod.rs
```diff
@@ -5,6 +5,7 @@ pub mod delete;
 pub mod describe;
 pub mod dev;
 pub mod dns;
+pub(crate) mod dotnet;
```
bfops: nit: why is this specific one `pub(crate)`? I don't think anything really imports the cli crate so the distinction seems odd

crates/cli/src/subcommands/init.rs
```diff
@@ -1622,6 +1630,82 @@ fn check_for_cargo() -> bool {
...
+/// Resolution order:
+///   1. Explicit `--dotnet-version` flag
+///   2. .NET 8 on macOS, where NativeAOT-LLVM is not supported
+///   3. .NET 8, if it is the only installed .NET SDK major version
+///   4. .NET 10 default
+fn resolve_dotnet_major(options: &InitOptions) -> anyhow::Result<u8> {
+    if let Some(v) = options.dotnet_version {
+        match v {
+            8 | 10 => return Ok(v),
+            _ => anyhow::bail!("Unsupported --dotnet-version {v}. Supported values: 8, 10."),
```
bfops: should this validation be in `parse_dotnet_version`? unless we intend to support other dotnet versions in other places but limit to 8 or 10 specifically here?

bfops: then this entire function is `options.dotnet_version.unwrap_or(resolve_default_dotnet_major)`, at which point the function could be inlined

crates/cli/src/subcommands/build.rs
```diff
@@ -110,8 +128,9 @@ pub async fn exec_with_argstring(
         Some(PathBuf::from(lint_dir))
     };
     let build_debug = arg_matches.get_flag("debug");
+    let dotnet_version = parse_dotnet_version(arg_matches.get_one::<String>("dotnet_version").map(String::as_str))?;
```
bfops: it seems like in every call site of `exec_with_argstring` we use `build_options_with_dotnet_version` to add the dotnet version param to the arg string.. the only purpose of which is to parse it back out here?

we're passing `native_aot` into this function anyway, why not just pass `dotnet_version` in?

crates/cli/src/subcommands/dotnet.rs
```diff
@@ -0,0 +1,55 @@
...
+    let Some(version) = parse_dotnet_version(dotnet_version)? else {
+        return Ok(build_options.to_string());
+    };
+
+    Ok(if build_options.is_empty() {
+        format!("--dotnet-version {version}")
+    } else {
+        format!("{build_options} --dotnet-version {version}")
+    })
+}
```
bfops: I left comments in other files but tl;dr I think this file should be removed.

`build_options_with_dotnet_version` seems entirely unnecessary in the 2 places it's used, and `parse_dotnet_version` could be bundled into a common arg definition in `common_args`.

crates/cli/src/tasks/csharp.rs
```diff
@@ -1,14 +1,229 @@
...
+
+fn dotnet_global_json(major: u8) -> anyhow::Result<String> {
+    match major {
+        8 => Ok(r#"{"sdk":{"version":"8.0.100","rollForward":"latestFeature"}}"#.to_string()),
+        10 => Ok(r#"{"sdk":{"version":"10.0.100","rollForward":"latestMinor"}}"#.to_string()),
+        _ => anyhow::bail!("Unsupported .NET SDK version: {major}. SpacetimeDB requires .NET SDK 8.0 or 10.0."),
+    }
+}
+
+fn temporarily_pin_project_sdk(project_path: &Path, major: u8) -> anyhow::Result<TemporaryGlobalJson> {
```
bfops: I'm surprised to see this. Why are we overriding users' `global.json`s?

crates/cli/src/tasks/csharp.rs
```diff
@@ -17,49 +232,166 @@ pub(crate) fn build_csharp(project_path: &Path, build_debug: bool) -> anyhow::Re
...
-                ));
+    let native_aot_flag = native_aot;
+
+    let project_global_json = find_global_json(project_path);
+    let cwd = std::env::current_dir()?;
+    let cwd_global_json = find_global_json(&cwd);
+    let dotnet_version_result = if project_global_json.is_some() {
+        Some(dotnet!("--version").read())
+    } else if cwd_global_json.is_some() {
+        Some(duct::cmd!("dotnet", "--version").read())
```
bfops, line 201: nit: why do we not use the `dotnet!` macro on this one?

.github/workflows/ci.yml
lisandroct: To be fair, I don't know if we even need these. The .csproj auto downloads `wasi-sdk` on demand. But both (these steps and the auto download) were added before I started working on this feature, so I don't know if the steps are required because the auto download fails on CI and we need them or they're outdated. I'll try removing them and see if the checks still pass.

crates/cli/src/tasks/csharp.rs
bfops: oh nvm that would use the project dir as current dir

crates/cli/src/tasks/csharp.rs
```diff
@@ -17,49 +232,166 @@ pub(crate) fn build_csharp(project_path: &Path, build_debug: bool) -> anyhow::Re
...
+        None => None,
+    };
+
+    // Resolution order:
+    //   1. --dotnet-version CLI flag (explicit user override)
+    //   2. global.json-selected SDK, if one applies to the project or current directory
+    //   3. Single <TargetFramework> in the project's .csproj
+    //   4. .NET 10 default for missing or multi-target project context
+    let csproj_tfms = read_tfms_from_csproj(project_path);
+    eprintln!(
```
bfops: is this intended to be left in?

crates/cli/src/subcommands/build.rs
lisandroct: We're gonna remove it quite soon (.NET 8 loses support in November, that's less than 4 months in the future). Probably not a necessary clean up.

bfops (commented):
I left comments on my code-owned files.

The `init`, `dev`, and `tasks::csharp` subcommand changes are large enough that I find them hard to review, so I can't say that I've been super thorough with them.

.github/workflows/ci.yml
lisandroct: @rekhoff This was removed in [this commit](https://github.com/clockworklabs/SpacetimeDB/commit/c2ac4285dadc6fb2905d70aa9ba4a3393b1b481f).

crates/cli/src/subcommands/mod.rs
lisandroct: This one I don't know. I just moved common functions that were being duplicated in multiple files to `dotnet.rs` and Codex added this here. I'm not that proficient with Rust so I can't know why AI decided to do it this way.

.github/workflows/ci.yml
lisandroct: This got added in [044d516b3053133698592a5eb0af2381fc9ab215](https://github.com/clockworklabs/SpacetimeDB/commit/044d516b3053133698592a5eb0af2381fc9ab215). I don't know if @rekhoff has any recollection.

I understand that the symlinks get broken due to the "Install .NET workloads" step above.

.github/workflows/ci.yml
lisandroct: I think here's the only place where we're targeting both .NET 8 and .NET 10. This variable is only needed in .NET 8 AOT. .NET 10 presupposes that it's AOT.
It might be outdated and not needed anymore. I'll check it.

lisandroct: For my own personal tracking of all of these changes: This was added in [2d06a3d833f319253bee772886aa28d616867a15](https://github.com/clockworklabs/SpacetimeDB/commit/2d06a3d833f319253bee772886aa28d616867a15).

crates/cli/src/subcommands/build.rs
lisandroct: This makes sense. I'll do it.

crates/cli/src/subcommands/mod.rs
lisandroct: And these are going to go away as soon as we deprecate .NET 8.

crates/cli/src/tasks/csharp.rs
lisandroct: No, we can delete this. Thank you.

crates/cli/src/subcommands/build.rs
lisandroct: Done. 👍🏻

.github/workflows/ci.yml
lisandroct: It seems like they were indeed not needed. I removed them and the CI checks are still passing. 👍🏻

crates/cli/src/tasks/csharp.rs
lisandroct: Done. 👍🏻

crates/cli/src/subcommands/init.rs
lisandroct: I added the validation in `parse_dotnet_version` in a previous commit.
I'm inlining `options.dotnet_version.unwrap_or(resolve_default_dotnet_major)` now. 👍🏻

crates/cli/src/subcommands/build.rs
lisandroct: Yes, this makes sense. I just pushed a commit making `dotnet_version` act like `native_aot`.

crates/cli/src/subcommands/dotnet.rs
lisandroct: With all the other cleaning that I've done, yes, we can safely remove this file now. Just did. Thank you.

crates/cli/src/subcommands/mod.rs
lisandroct: I just removed `dotnet.rs`, so this is gone. 👍🏻

crates/cli/src/subcommands/build.rs
lisandroct: I ended up doing it since I was already cleaning up stuff. Much cleaner now. 👍🏻

tools/ci/src/main.rs
lisandroct: This was outdated. I added this when I was moving to .NET 10 and the templates had different version than the root `global.json` but now they match. So I reverted it to what it was before (except that now we'll check only for the tracked files because for some tests we generate some templates targeting .NET 8).

crates/cli/src/tasks/csharp.rs
lisandroct: I changed this to just throw an error if the build .NET target doesn't match the project's sdk version. We're using this internally for tests. I don't expect users to run `generate` or `publish` with `--dotnet-version` because the module project will already have a version. It wouldn't make sense. In testing we have some multi version projects and it's needed.

.github/workflows/ci.yml
lisandroct: Effectively, this wasn't needed. I removed it and all CI checks still passed. I imagined it was required at one point and was simply outdated.

.github/workflows/ci.yml
rekhoff: I suspect this was just an oversight on my part. I see no reason not to re-add the comment

lisandroct: Ryan just confirmed that this was probably an oversight and shouldn't have been deleted. I re-added it.

.github/workflows/ci.yml
rekhoff: When that got added, it was a LLM suggestion I was attempting. I know there was a number of times when the specific .NET install sequence was being extremely finicky. I suspect this was part of trying to fix that.
It might be of value to test with removing this bit and see if it is still necessary.

lisandroct: Okay, I deleted it and all the checks are still passing. 👍🏻

.github/workflows/ci.yml
bfops: even better!

.github/workflows/ci.yml
bfops: excellent

.github/workflows/ci.yml
bfops: nice

tools/ci/src/main.rs
bfops: okay that sounds great

crates/cli/src/subcommands/build.rs
bfops: nice thank you

crates/cli/src/tasks/csharp.rs
bfops: that sounds much safer, thank you!

bfops (approved):
all my comments on my code-owned files have been addressed.

my previous comment still stands:
> The `init`, `dev`, and `tasks::csharp` subcommand changes are large enough that I find them hard to review, so I can't say that I've been super thorough with them.

but from the point of view of this being okay CLI code, I'm fine to merge. I can't speak to the dotnet-validity / edge cases around that stuff, but that's not what my codeowner review is for.

PR #4919, opened 2026-04-30
Update the C++ smoketest to use `latest`

JasonAtClockwork:
# Description of Changes

The update version tool sets the /templates/basic-cpp/spacetimedb/CMakeLists.txt but as we're in the middle of release this can block the smoketest. This PR changes it to dynamically change to the `latest` release to allow release night smoketests to work correctly.

N/A

1 - Small smoketest change

# Testing

- [x] Ran the C++ quickstart smoketest locally
- [x] Updated local template and tested with / without the change


jdetter (approved):
This fixed my smoketest issue in the Release v2.2.0 PR (internal repo)

PR #4920, opened 2026-04-30
Godot SDK and Blackholio tutorial

lisandroct:
# Description of Changes
Very small additions to the C# SDK specific for Godot.
I wrote the Godot Blackholio tutorial and updated the Unity one.
I added the image assets necessary for the tutorial.
I added the files for the Godot demo.

No breaking changes.

1. There's really no risk for current systems or projects, it's all new additions to support nicely(-ish) Godot, new Blackholio demo for Godot and updates to the tutorials.

# Testing
- [X] Follow the tutorial and verify everything works.
- [X] Build for Windows and verify it works.
- [x] Build for Linux and verify it works.
- [x] Build SDK
- [x] Publish to Nuget


docs/docs/00100-intro/00300-tutorials/00500-godot-tutorial/00200-part-1.md
```diff
@@ -0,0 +1,84 @@
+---
+title: 1 - Setup
+slug: /tutorials/godot/part-1
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+
+![Unity Tutorial Hero Image](/images/unity/part-1-hero-image.png)
```
rekhoff, line 10: Update this with a screenshot of the Godot implementation, or removed it from the tutorial for now.

docs/docs/00100-intro/00300-tutorials/00500-godot-tutorial/00200-part-1.md
```diff
@@ -0,0 +1,84 @@
...
+>
+> [https://github.com/clockworklabs/SpacetimeDB/tree/master/demo/Blackholio](https://github.com/clockworklabs/SpacetimeDB/tree/master/demo/Blackholio)
+
+## Setting up the Tutorial Godot Project
+
+In this section, we will guide you through the process of setting up a Godot Project that will serve as the starting point for our tutorial. By the end of this section, you will have a basic Godot project and be ready to implement the server functionality.
+
+### Step 1: Create a Blank Godot Project
+
+SpacetimeDB supports Godot version `4.6.2` or later. See [the overview](.) for more information on specific supported versions.
```
rekhoff, line 24: I think this link is broken. Also, when going to download [Godot 4.6.2](https://godotengine.org/download/archive/4.6.2-stable/), for Windows, macOS, and Linux, there is a `Standard` and a `.NET` install. I don't see all call out to use the `.NET` (on the download page it does say ".NET builds offer support for C# as a scripting language.", but it might be good to call out the .NET version, unless either will work.

docs/docs/00100-intro/00300-tutorials/00500-godot-tutorial/00200-part-1.md
```diff
@@ -0,0 +1,84 @@
...
+
+### Create a new Scene
+
+1. **Create a 2D Scene**:
+   - In the **Scene** Dock (usually on the top left), click on **Create Root Node: -> 2D Scene**.
+   - Alternatively, you can click the **+** button to add a child node and select a **Node 2D**.
+
+![Create 2D Scene](/images/godot/part-1-godot-create-2d-scene.jpg)
+
+2. **Rename the GameObject**:
```
rekhoff:
```suggestion
2. **Rename the root Node**:
```

sdks/csharp/src/AuthToken.cs
```diff
@@ -13,7 +13,74 @@
...
+            var key = "spacetimedb.identity_token";
+#if UNITY_EDITOR
+            // Different editors need different keys
+            key += $" - {Application.dataPath}";
+#endif
+            return key;
+        }
+    }
+}
+#elif GODOT
```
rekhoff, line 43: I don't believe this will work to allow Godot to have code only relevant to it within the NuGet packages. Unity is able to get away with unique build flags because we have Unity developers [Import the SpacetimeDB SDK using the C# SDK repo](https://spacetimedb.com/docs/tutorials/unity/part-1#import-the-spacetimedb-unity-sdk). This means for Unity developers, the original code is intact and constants like `UNITY_5_3_OR_NEWER` and `UNITY_EDITOR` are evaluated locally. Only the [SpacetimeDB.BSATN.Codegen.dll](https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/blob/release/latest/packages/spacetimedb.bsatn.runtime/2.1.0/analyzers/dotnet/cs/SpacetimeDB.BSATN.Codegen.dll) and [SpacetimeDB.BSATN.Runtime.dll](https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/blob/release/latest/packages/spacetimedb.bsatn.runtime/2.1.0/lib/netstandard2.1/SpacetimeDB.BSATN.Runtime.dll) are pre-built head of time.

When we publish the NuGet packages for the general C# developers, that will strip out the `GODOT` build flag because it's not included as a constant during compilation. If we did include it, that would make it available to all SDK users (which of course would break things since it's specifically imports `using Godot`.

We'll likely need another approch than using `GODOT` build flags, or would need to make a custom target framework specifically for this.

For additional context, if we follow the tutorial, but point at a local NuGet package build from the `lisandro/godot-blackholio` branch, we'd get the errors like:
```csharp
GameManager.cs(43,23): Error CS0117 : 'AuthToken' does not contain a definition for 'TryGetToken'
GameManager.cs(49,9): Error CS0103 : The name 'STDBUpdateManager' does not exist in the current context
```
And again, this is because the code is getting stripped out.

sdks/csharp/src/GodotDebugLogger.cs
```diff
@@ -0,0 +1,32 @@
+﻿/*  SpacetimeDB logging for Godot
+ *  *  This class is only used in Godot projects.
+ *  *
+ */
+#if GODOT
```
rekhoff, line 5: See my other note about this getting stripped out from the NuGet package during compilation.

sdks/csharp/src/ISpacetimeDBLogger.cs
```diff
@@ -19,6 +19,8 @@ public static class Log
 
 #if UNITY_5_3_OR_NEWER
             new UnityDebugLogger();
+#elif GODOT
+            new GodotDebugLogger();
```
rekhoff, line 23: Again, see my other note about this getting stripped out from the NuGet package during compilation.

sdks/csharp/src/STDBUpdateManager.cs
```diff
@@ -0,0 +1,110 @@
+#if GODOT
```
rekhoff, line 1: And one more: see my other note about this getting stripped out from the NuGet package during compilation.

rekhoff (changes requested):
I was only able to get part-way through `docs/docs/00100-intro/00300-tutorials/00500-godot-tutorial/00300-part-2.md` before hitting a blocker with the implementation.
The tutorial so far it looks good with only some minor nits, but the NuGet package issue will need to get addressed before I can finish reviewing.

docs/docs/00100-intro/00300-tutorials/00500-godot-tutorial/00200-part-1.md
lisandroct: Ah, yeah, I missed that one.

docs/docs/00100-intro/00300-tutorials/00500-godot-tutorial/00200-part-1.md
lisandroct: If it's broken here, I imagine it's broken in the Unity tutorial. I'll check.

docs/docs/00100-intro/00300-tutorials/00500-godot-tutorial/00200-part-1.md
lisandroct: Yeap, my bad.

sdks/csharp/src/AuthToken.cs
lisandroct: I was wondering if this was going to work. I couldn't get my local NuGet working because I was missing what you dm me yesterday.
I guess the ~~only~~best solution we have is to compile with the GODOT preprocessor flag and publish a different NuGet package.

docs/docs/00100-intro/00300-tutorials/00500-godot-tutorial/00200-part-1.md
lisandroct: Confirmed it's the same way in the Unity tutorial and it works.

docs/docs/00100-intro/00300-tutorials/00500-godot-tutorial/00400-part-3.md
```diff
@@ -0,0 +1,1645 @@
...
+    
+    public override void _Draw() => DrawCircle(Vector2.Zero, Radius, Color);
+}
+```
+
+#### EntityController
+
+Let's also create an `EntityController` script which will serve as a base class for both our `CircleController` and `FoodController` classes since both `Circle`s and `Food` are entities.
+
+Create a new C# script called `EntityController.cs` and replace its contents with:
```
rekhoff, line 1012: When creating this EntityController.cs script using Godot's Create Script dialog, you can't select `Circle2D` as what it Inherits from.
Since we aren't really asking them to define inheritance elsewhere and the code we copy in puts the correct inheritance in place anyway, this is just an optional thing.

If actually wanted them to be able to select our C# classes in this manner though, by having the Circle2D class defined with the [GlobalClass] attribute. Documentation around this is at https://docs.godotengine.org/en/4.4/tutorials/scripting/c_sharp/c_sharp_global_classes.html
With that in place, if we then press play once, so the project rebuilds, Circle2D shows up under Node2D as something that can be inherited from.

Since we don't actually ask them to do this, it would purely be for teaching users how to interact between the Godot Editor side of the house, and C# backend code that works with SpacetimeDB.

docs/docs/00100-intro/00300-tutorials/00500-godot-tutorial/00500-part-4.md
```diff
@@ -0,0 +1,898 @@
...
+</Tabs>
+
+Finally, let's schedule a reducer to run every 50 milliseconds to move the player's circles around based on the most recently set player input.
+
+<Tabs groupId="server-language" defaultValue="rust">
+<TabItem value="csharp" label="C#" >
+
+```csharp
+public partial struct MoveAllPlayersTimer
+{
```
rekhoff, line 339:
```suggestion
```csharp
[Table(Accessor = "move_all_players_timer", Scheduled = nameof(MoveAllPlayers), ScheduledAt = nameof(scheduled_at))]
public partial struct MoveAllPlayersTimer
{
```
Without this, we'll get an error on the C# step that reads
> Add the following to your `Init` reducer to schedule the `MoveAllPlayers` reducer to run every 50 milliseconds.

rekhoff (commented):
I have tested the C# Server implementation instructions of the Godot tutorial and left my comments below.
I have also tested the `client-godot` version in the demo folder and confirmed it works as intended.

sdks/csharp/src/AuthToken.cs
rekhoff: In my most-recent testing, I was able to rebuild the NuGet packages with the `SpacetimeDB.ClientSDK.Godot.2.1.0.nupkg` package that is now getting generated.

sdks/csharp/SpacetimeDB.ClientSDK.Godot.csproj
```diff
@@ -0,0 +1,50 @@
...
+    <Company>Clockwork Labs</Company>
+    <Product>SpacetimeDB</Product>
+    <Description>The SpacetimeDB Godot SDK is a software development kit (SDK) designed to simplify the interaction with SpacetimeDB server modules from Godot projects.</Description>
+    <Copyright>2024</Copyright>
+    <PackageProjectUrl>https://spacetimedb.com</PackageProjectUrl>
+    <PackageIcon>logo.png</PackageIcon>
+    <PackageReadmeFile>README.md</PackageReadmeFile>
+    <RepositoryUrl>https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk</RepositoryUrl>
+    <AssemblyVersion>2.1.0</AssemblyVersion>
+    <Version>2.1.0</Version>
```
rekhoff:
```suggestion
    <AssemblyVersion>2.2.0</AssemblyVersion>
    <Version>2.2.0</Version>
```
As a note, we'll need to make sure the process we are using to update version numbers includes `SpacetimeDB.ClientSDK.Godot.csproj` going forward. Since that last update was pulled in from the `master` branch, the other NuGet packages are all on `2.2.0` while the Godot was still on `2.1.0`

docs/docs/00100-intro/00300-tutorials/00500-godot-tutorial/00400-part-3.md
lisandroct: I thought it was easier to follow this way. I do have a note mentioning that our script inherits from `Circle2D` later. But we can add it.

docs/docs/00100-intro/00300-tutorials/00500-godot-tutorial/00500-part-4.md
lisandroct: I'm confused about this one. What you're saying makes sense, but I followed the tutorial and everything worked just fine. Also, this is a copy/paste from Unity, so we will have to update it there too.

sdks/csharp/SpacetimeDB.ClientSDK.Godot.csproj
lisandroct: I was planning on syncing with Zeke after this PR was merged to make sure all of these processes account for the Godot SDK.

docs/docs/00100-intro/00300-tutorials/00500-godot-tutorial/00500-part-4.md
lisandroct: I just pushed a commit with this. But I'm still confused about it.

sdks/csharp/SpacetimeDB.ClientSDK.Godot.csproj
lisandroct: Pushed a commit with the updated version.

rekhoff (approved):
Requested changes have been made. Marking my approval. Great work on this!

tools/regen/src/csharp.rs
```diff
@@ -219,6 +219,19 @@ pub fn regen_dlls() -> Result<()> {
     .dir(&sdk)
     .run()?;
 
+    cmd!(
+        "dotnet",
+        "restore",
+        "SpacetimeDB.ClientSDK.Godot.csproj",
+        "--configfile",
+        path_arg(&nuget_config_path),
+        "-p:BaseOutputPath=bin~/",
```
bfops, line 229: were these params needed? it seems like the csproj already specifies this

lisandroct: They were at one point. Maybe they're not anymore. I'll check.

lisandroct: Okay, I tested without the params and I get the following error:
```
/usr/lib64/dotnet/sdk/8.0.125/Sdks/Microsoft.NET.Sdk/targets/Microsoft.PackageDependencyResolution.targets(266,5): error NETSDK1004: Assets file '/var/home/clockworklabs/Projects/SpacetimeDB/sdks/csharp/obj~/godot/project.assets.json' not found. Run a NuGet package restore to generate this file. [/var/home/clockworklabs/Projects/SpacetimeDB/sdks/csharp/SpacetimeDB.ClientSDK.Godot.csproj]
```

bfops: okay interesting. Thanks for checking!

bfops: I had codex make a change to do this via `Directory.Build.props` like we do for the main C# SDK, so then we don't have to add these params everywhere we use the csproj.

bfops (approved):
my code-owned file LGTM

Issue #4859, opened 2026-04-21
Move all workflow CI under `cargo ci`


bfops:

bfops:
moving this to blocked since I am prioritizing test flakiness above this

bfops:
https://github.com/clockworklabs/SpacetimeDB/pull/4950
https://github.com/clockworklabs/SpacetimeDB/pull/4921
https://github.com/clockworklabs/SpacetimeDB/pull/4948

PR #4921, opened 2026-04-30
CI - `cargo test` runs C# and TypeScript tests

### LINKED ISSUE #4923: Run simple language tests under `cargo test`
- [x] TypeScript
- [x] C#

bfops:
# Description of Changes

C# and TypeScript tests can now be run via `cargo test`.

This also involved updating some test code to take a server url override, so they could use the existing `SpacetimeDBGuard` (which starts the server on an arbitrary port rather than a hardcoded 3000).

- `master`
  - [#4972 `cargo ci dlls` -> `cargo regen csharp dlls`](https://github.com/clockworklabs/SpacetimeDB/pull/4972)
    - [#4971 Translate some bash scripts to Rust](https://github.com/clockworklabs/SpacetimeDB/pull/4971)
      - **[This PR] CI - `cargo test` runs C# and TypeScript tests**
        - [#4948 CI - Run unity tests through `cargo test`](https://github.com/clockworklabs/SpacetimeDB/pull/4948)

None. CI only.

2

# Testing
Failure reporting was checked with synthetic failing tests after the standard harness checks.

TypeScript harness:
- Run the wrapped TypeScript suite with `cargo test -p spacetimedb-typescript-tests`
- This runs `pnpm build`, `pnpm vitest run` with a JUnit report, then `pnpm vitest typecheck --run`. Reports are written under `target/language-tests/typescript/`.

C# harness:
- Pre-build the local server binaries before running the full C# harness:
  - `cargo build -p spacetimedb-cli -p spacetimedb-standalone --features spacetimedb-standalone/allow_loopback_http_for_tests`
- Run the wrapped C# suite with `cargo test -p spacetimedb-csharp-tests`
- This runs the bindings tests, packs/restores the SDK solution against local NuGet packages, runs the SDK tests, and then runs the C# regression script against a temporary SpacetimeDB server started through `spacetimedb-guard`. TRX reports are written under `target/language-tests/csharp/`.

## Synthetic failure check
- Added a temporary failing Vitest test and ran:
  - `cargo test -p spacetimedb-typescript-tests --test typescript -- --filter SyntheticHarnessFailure`
- Verified the command exited with status 1 and reported the failure like this:

```text
FAIL  tests/synthetic_harness_failure.test.ts > SyntheticHarnessFailure reports TypeScript assertion failures
AssertionError: expected 'actual TypeScript failure' to be 'expected TypeScript success'

Expected: "expected TypeScript success"
Received: "actual TypeScript failure"

command failed in .../crates/bindings-typescript:
  pnpm test -- --reporter=default --reporter=junit --outputFile=... -t SyntheticHarnessFailure
status: exit status: 1
```

- Added a temporary failing xUnit test and ran:
  - `cargo test -p spacetimedb-csharp-tests --test csharp -- --filter SyntheticHarnessFailure`
- Verified the command exited with status 1 and reported the failure like this:

```text
SyntheticHarnessFailureTests.SyntheticHarnessFailureReportsCsharpAssertionFailures [FAIL]
Error Message:
 Assert.Equal() Failure: Strings differ
Expected: "expected C# success"
Actual:   "actual C# failure"

command failed in .../sdks/csharp:
  dotnet test -warnaserror --results-directory ... --logger trx;LogFileName=sdk.trx --no-restore --filter SyntheticHarnessFailure
status: exit status: 1
```


tools/ci/src/main.rs
```diff
@@ -464,6 +464,12 @@ fn main() -> Result<()> {
                 "spacetimedb-smoketests",
                 "--exclude",
                 "spacetimedb-sdk",
+                "--exclude",
+                "spacetimedb-language-test-support",
```
bfops:
```suggestion
```

tools/run-all-tests.sh
```diff
@@ -11,7 +11,10 @@ cd "$stdb_root"
 
 tools/clippy.sh
 
-cargo test --all
+cargo test --all \
+    --exclude spacetimedb-language-test-support \
```
bfops:
```suggestion
```

crates/language-test-support/src/lib.rs
```diff
@@ -0,0 +1,308 @@
...
+
+#[derive(Clone, Debug, Default)]
+pub struct HarnessArgs {
+    pub filter: Option<String>,
+    pub list: bool,
+    pub passthrough: Vec<String>,
+}
+
+impl HarnessArgs {
+    pub fn parse() -> Self {
```
bfops: this should really just use `clap` or something, no?

crates/typescript-tests/tests/typescript.rs
```diff
@@ -0,0 +1,53 @@
...
+        }
+        run_command_forward("pnpm", &cmd, &cwd)?;
+        return Ok(());
+    }
+
+    run_command("pnpm", &["build".to_string()], &cwd)?;
+
+    let mut test_args = vec![
+        "vitest".to_string(),
+        "run".to_string(),
```
bfops: this is now `vitest run` instead of the original `pnpm test` - are these equivalent in terms of the tests they run?

bfops: how do I confirm that the list of tests is the same?

crates/typescript-tests/tests/typescript.rs
```diff
@@ -0,0 +1,53 @@
...
+    }
+}
+
+fn run() -> Result<()> {
+    let args = HarnessArgs::parse();
+    require_tool("pnpm")?;
+
+    let workspace = spacetimedb_language_test_support::workspace_root();
+    let cwd = workspace.join("crates/bindings-typescript");
+    let out_dir = artifact_dir("typescript")?;
```
bfops: why is the `vitest.junit.xml` being output here? this doesn't seem like a default location

bfops: I don't think we need a shared `artifact_dir` helper. maybe a helper to get target dir.

crates/language-test-support/src/lib.rs
```diff
@@ -0,0 +1,308 @@
...
+
+pub fn require_tool(tool: &str) -> Result<()> {
+    if find_on_path(tool).is_some() {
+        Ok(())
+    } else {
+        bail!("required tool `{tool}` was not found on PATH")
+    }
+}
+
+fn find_on_path(tool: &str) -> Option<PathBuf> {
```
bfops: I think there are libraries that can do a better job than manually rolling this function. I believe we already use some kind of `which` replacement library elsewhere

crates/language-test-support/src/lib.rs
```diff
@@ -0,0 +1,308 @@
...
+}
+
+pub fn run_command_forward(program: &str, args: &[String], cwd: &Path) -> Result<()> {
+    let output = run_command(program, args, cwd)?;
+    print!("{}", String::from_utf8_lossy(&output.stdout));
+    eprint!("{}", String::from_utf8_lossy(&output.stderr));
+    Ok(())
+}
+
+pub fn run_command_env(program: &str, args: &[String], cwd: &Path, envs: &[(&str, String)]) -> Result<Output> {
```
bfops: okay this is basically a duplicate of the above function, please do a better job of factoring them together.

bfops: perhaps just a helper to do the output success assertion and bail

crates/language-test-support/src/lib.rs
```diff
@@ -0,0 +1,308 @@
...
+}
+
+pub fn parse_junit(path: &Path) -> Result<Vec<TestCaseResult>> {
+    parse_xml_results(path, XmlKind::Junit)
+}
+
+enum XmlKind {
+    Trx,
+    Junit,
+}
```
bfops: seems like this should be in the specific libraries that use them, instead of having a central enum here that everything needs to go through

bfops: `parse_xml_results` can just be public

bfops: is there perhaps a library we can use to do this parsing, or output the results in a different format that's easier to parse? they do not need to be a shared format between languages.

crates/guard/src/lib.rs
```diff
@@ -252,6 +266,7 @@ impl SpacetimeDbGuard {
         pg_port: Option<u16>,
         spawn_id: u64,
         use_installed_cli: bool,
+        listen_addr: &str,
```
bfops: why was this needed? why can't we keep just using `:0` and then use whatever port it gives us?

tools/run-all-tests.sh
bfops: Addressed in the current branch. tools/run-all-tests.sh is back to plain cargo test --all.

crates/typescript-tests/tests/typescript.rs
bfops: Partly changed. The wrapper still writes a JUnit report to target/language-tests/typescript so it has a stable file to parse, but the shared artifact_dir helper is gone; each wrapper now derives its own path from target_dir().

crates/language-test-support/src/lib.rs
bfops: Addressed. HarnessArgs now uses clap for the supported wrapper flags and only manually splits the post--- passthrough.

crates/typescript-tests/tests/typescript.rs
bfops: Addressed. The wrapper now invokes the package script as pnpm test -- ... so the base test selection is exactly the existing script. I also verified the wrapper lists/runs 182 Vitest tests.

tools/ci/src/main.rs
bfops: Addressed. The support crate is no longer excluded from the broad cargo test path; only the two harness crates are excluded because they wrap native test suites.

crates/typescript-tests/tests/typescript.rs
bfops: Addressed. artifact_dir is removed; the shared helper is just target_dir(), and each wrapper creates its own report directory.

crates/guard/src/lib.rs
bfops: Kept intentionally. The client-side regression script can now target an arbitrary server URL, but two server modules still make HTTP egress calls to http://localhost:3000 internally. Until those are parameterized too, this wrapper has to start the local server on 127.0.0.1:3000 for this suite.

crates/language-test-support/src/lib.rs
bfops: Partly addressed. parse_xml_results and XmlKind are now public, so callers are not forced through per-format wrappers. I kept the tiny quick-xml parser shared because both native runners already emit XML formats and we only need name/outcome/message summaries.

crates/language-test-support/src/lib.rs
bfops: Addressed. run_command and run_command_env now share run_command_inner plus a single ensure_success helper. run_command_forward streams directly to the terminal and uses the same success assertion.

crates/language-test-support/src/lib.rs
bfops: Addressed. require_tool now uses the which crate instead of manually searching PATH.

crates/language-test-support/src/lib.rs
```diff
@@ -0,0 +1,279 @@
...
+            .map(|index| args.split_off(index + 1))
+            .unwrap_or_default();
+        if args.last().is_some_and(|arg| arg == "--") {
+            args.pop();
+        }
+
+        let mut parsed = <HarnessArgs as Parser>::parse_from(args);
+        parsed.passthrough = passthrough;
+        if parsed.filter.is_none()
+            && let Some(filter) = parsed.positional.first()
```
bfops: why do we need this?

is it necessary to have this arg parsing be shared between C# and typescript? perhaps they should each construct their own args and just share some definitions where needed? having this custom parsing is a smell.

crates/language-test-support/src/lib.rs
bfops: Addressed by factoring the status check into ensure_success, shared by all command wrappers.

crates/language-test-support/src/lib.rs
bfops: Kept the existing native XML outputs: Vitest JUnit and dotnet TRX. I did not add a heavier parser dependency because we only need a small summary for cargo output, and the quick-xml parser is limited to that.

bfops: Addressed. XmlKind and parse_xml_results are public now, while parse_junit and parse_trx remain as small convenience wrappers.

crates/language-test-support/src/lib.rs
```diff
@@ -0,0 +1,279 @@
...
+        .to_path_buf()
+}
+
+pub fn target_dir() -> PathBuf {
+    env::var_os("CARGO_TARGET_DIR")
+        .map(PathBuf::from)
+        .unwrap_or_else(|| workspace_root().join("target"))
+}
+
+pub fn require_tool(tool: &str) -> Result<()> {
```
bfops: I don't think we need this function at all. it will fail one way or another if the tool isn't installed

crates/typescript-tests/tests/typescript.rs
```diff
@@ -0,0 +1,55 @@
...
+    }
+}
+
+fn run() -> Result<()> {
+    let args = HarnessArgs::parse();
+    require_tool("pnpm")?;
+
+    let workspace = spacetimedb_language_test_support::workspace_root();
+    let cwd = workspace.join("crates/bindings-typescript");
+    let out_dir = target_dir().join("language-tests").join("typescript");
```
bfops: I do not think we need to `.join("language-tests")`. perhaps we should just do `.join("typescript-tests")`.

same for C# section

crates/language-test-support/src/lib.rs
```diff
@@ -0,0 +1,279 @@
...
+    if which::which(tool).is_ok() {
+        Ok(())
+    } else {
+        bail!("required tool `{tool}` was not found on PATH")
+    }
+}
+
+pub fn run_command(program: &str, args: &[String], cwd: &Path) -> Result<Output> {
```
bfops: remove this function

crates/language-test-support/src/lib.rs
```diff
@@ -0,0 +1,279 @@
...
+        },
+    )?;
+    Ok(())
+}
+
+pub fn run_command_env(program: &str, args: &[String], cwd: &Path, envs: &[(&str, String)]) -> Result<Output> {
+    run_command_inner(program, args, cwd, envs)
+}
+
+fn run_command_inner(program: &str, args: &[String], cwd: &Path, envs: &[(&str, String)]) -> Result<Output> {
```
bfops: remove this function. callers can do this themselves.

crates/language-test-support/src/lib.rs
```diff
@@ -0,0 +1,279 @@
...
+    } else {
+        bail!("required tool `{tool}` was not found on PATH")
+    }
+}
+
+pub fn run_command(program: &str, args: &[String], cwd: &Path) -> Result<Output> {
+    run_command_inner(program, args, cwd, &[])
+}
+
+pub fn run_command_forward(program: &str, args: &[String], cwd: &Path) -> Result<()> {
```
bfops: inline this function into callers

crates/guard/src/lib.rs
bfops: what is "this suite"? can the relevant suite be instead adapted to target the port that the server starts on, like the smoketests do?

crates/language-test-support/src/lib.rs
bfops: remove this enum. move the xml parsing into each crate that actually needs it, since there's a big `match` on the kind in the middle anyway. this is not a good use case for factoring.

if factoring is still necessary, do so with the actual common code, rather than something that is tightly coupled to the caller's context anyway.

helpers may still be in here.

crates/language-test-support/src/lib.rs
bfops: Addressed. Removed require_tool and the which dependency; the command wrappers now just report the spawn failure if the tool is missing.

crates/typescript-tests/tests/typescript.rs
bfops: Addressed. TypeScript now writes under target/typescript-tests and C# under target/csharp-tests.

crates/language-test-support/src/lib.rs
bfops: Addressed in 107ebc62a. XML parsing now lives in the individual TS/C# harness crates; language-test-support no longer has XmlKind, parse_xml_results, parse_junit, or parse_trx.

crates/language-test-support/src/lib.rs
bfops: Addressed in 107ebc62a. Removed this shared command helper from language-test-support; the command execution is local to the harness that needs it.

crates/language-test-support/src/lib.rs
bfops: Addressed in 107ebc62a. Argument parsing is no longer shared; the TS and C# harnesses each define their own local Args type and construct their own native command arguments.

crates/guard/src/lib.rs
bfops: Addressed in 107ebc62a. The guard API is back to dynamic 127.0.0.1:0, and the C# regression schema procedure now accepts the server URL so the suite uses the port selected by SpacetimeDbGuard.

crates/language-test-support/src/lib.rs
bfops: Addressed in 107ebc62a. Removed this shared helper; the C# harness now owns the env-aware command invocation directly.

crates/language-test-support/src/lib.rs
bfops: Addressed in 107ebc62a. Inlined this into the callers; language-test-support no longer exposes command runner helpers.

crates/typescript-tests/tests/typescript.rs
```diff
@@ -0,0 +1,215 @@
...
+            .map(|index| args.split_off(index + 1))
+            .unwrap_or_default();
+        if args.last().is_some_and(|arg| arg == "--") {
+            args.pop();
+        }
+
+        let mut parsed = <Args as Parser>::parse_from(args);
+        parsed.passthrough = passthrough;
+        if parsed.filter.is_none()
+            && let Some(filter) = parsed.positional.first()
```
bfops: I previously asked why this custom first-positional-argument parsing is necessary. please explain it or do something more conventional (which is preferable).

crates/csharp-tests/tests/csharp.rs
```diff
@@ -0,0 +1,290 @@
...
+    run_command_env(
+        "bash",
+        &["tools~/run-regression-tests.sh".to_string()],
+        &workspace.join("sdks/csharp"),
+        &[("SPACETIMEDB_SERVER_URL", guard.host_url.clone())],
+    )?;
+    Ok(())
+}
+
+fn run_command(program: &str, args: &[String], cwd: &Path) -> Result<Output> {
```
bfops: you were asked to inline several such functions

crates/typescript-tests/tests/typescript.rs
bfops: Addressed in 746a3ab3d. Removed the first-positional-argument filter shortcut; filtering is now only through the explicit --filter wrapper option, with post--- args passed through to Vitest.

crates/csharp-tests/tests/csharp.rs
bfops: Addressed in 746a3ab3d. Removed the C# local command runner wrappers and inlined those command invocations at their call sites; only the shared success/error formatting helper remains.

bfops: what about the same functions in typescript? the disagreement was with the utility and generalizability of such functions, not with those specific instances

bfops: Addressed in cc6d732f5. Removed the TypeScript run_command/run_command_forward helpers as well and inlined the pnpm invocations at the call sites, matching the C# harness treatment.

crates/typescript-tests/tests/typescript.rs
```diff
@@ -0,0 +1,215 @@
...
+        let mut cmd = vec!["vitest".to_string(), "list".to_string()];
+        if let Some(filter) = args.filter {
+            cmd.push(filter);
+        }
+        let status = Command::new("pnpm")
+            .args(&cmd)
+            .current_dir(&cwd)
+            .stdin(Stdio::inherit())
+            .stdout(Stdio::inherit())
+            .stderr(Stdio::inherit())
```
bfops: isn't `inherit` the default?

crates/typescript-tests/tests/typescript.rs
```diff
@@ -0,0 +1,215 @@
...
+            .status()
+            .with_context(|| format!("failed to spawn `{}` in {}", shell_line("pnpm", &cmd), cwd.display()))?;
+        ensure_success(
+            &cwd,
+            "pnpm",
+            &cmd,
+            &Output {
+                status,
+                stdout: Vec::new(),
+                stderr: Vec::new(),
```
bfops: why do we not capture and pass these..?

crates/typescript-tests/tests/typescript.rs
```diff
@@ -0,0 +1,215 @@
...
+use clap::Parser;
+use quick_xml::events::Event;
+use quick_xml::Reader;
+use spacetimedb_language_test_support::{print_results, target_dir, Outcome, TestCaseResult};
+use std::fs;
+use std::path::Path;
+use std::process::{Command, Output, Stdio};
+
+#[derive(Clone, Debug, Default, Parser)]
+#[command(disable_help_flag = true)]
```
bfops: why disable help flag? (here and elsewhere)

crates/typescript-tests/tests/typescript.rs
```diff
@@ -0,0 +1,215 @@
...
+
+    #[arg(long, alias = "list-tests")]
+    list: bool,
+
+    #[arg(skip)]
+    passthrough: Vec<String>,
+}
+
+impl Args {
+    fn parse() -> Self {
```
bfops: why this custom parsing? can't we just use clap "directly"?

bfops: here and elsewhere

crates/typescript-tests/tests/typescript.rs
```diff
@@ -0,0 +1,215 @@
...
+fn run() -> Result<()> {
+    let args = Args::parse();
+
+    let workspace = spacetimedb_language_test_support::workspace_root();
+    let cwd = workspace.join("crates/bindings-typescript");
+    let out_dir = target_dir().join("typescript-tests");
+    fs::create_dir_all(&out_dir).with_context(|| format!("failed to create {}", out_dir.display()))?;
+    let report = out_dir.join("vitest.junit.xml");
+
+    if args.list {
```
bfops, line 40: factor this into a separate function

crates/typescript-tests/tests/typescript.rs
```diff
@@ -0,0 +1,215 @@
...
+            &Output {
+                status,
+                stdout: Vec::new(),
+                stderr: Vec::new(),
+            },
+        )?;
+        return Ok(());
+    }
+
+    let build_args = ["build".to_string()];
```
bfops, line 65: factor the rest of this function into a separate function (e.g. `run_tests`)

crates/typescript-tests/tests/typescript.rs
bfops: Addressed in 0fbec0329. Removed the explicit stdio inheritance; the streamed list commands now use Command::status() defaults. I applied the same cleanup to the C# list path.

crates/typescript-tests/tests/typescript.rs
bfops: Addressed in 0fbec0329. The streamed list path no longer fabricates an Output with empty stdout/stderr; both TS and C# now use a status-specific success check for Command::status(), while captured commands still pass the real Output.

crates/typescript-tests/tests/typescript.rs
bfops: Addressed in 0fbec0329. Removed disable_help_flag in both TS and C# harnesses and verified cargo test ... -- --help works for both.

crates/typescript-tests/tests/typescript.rs
bfops: Applied here and elsewhere in 0fbec0329: both TypeScript and C# now use direct clap parsing with a trailing passthrough arg.

crates/typescript-tests/tests/typescript.rs
bfops: Addressed in 0fbec0329. Moved the TypeScript build/test/report flow into run_tests(), leaving run() to parse args and dispatch.

crates/typescript-tests/tests/typescript.rs
bfops: Addressed in 0fbec0329. Moved TypeScript list mode into list_tests().

crates/typescript-tests/tests/typescript.rs
bfops: Addressed in 0fbec0329. Removed the custom parse implementation in both harnesses; they now use clap directly with #[arg(last = true)] for post--- passthrough.

crates/csharp-tests/tests/csharp.rs
```diff
@@ -0,0 +1,283 @@
...
+            format!(
+                "failed to spawn `{}` in {}",
+                shell_line("dotnet", &args),
+                workspace.display()
+            )
+        })?;
+    ensure_success(workspace, "dotnet", &args, &output)?;
+
+    let cwd = workspace.join("sdks/csharp");
+    let args = ["./tools~/write-nuget-config.sh".to_string(), "../..".to_string()];
```
bfops: there used to be a large comment on this in the ci.yml. please restore it here.

crates/csharp-tests/tests/csharp.rs
```diff
@@ -0,0 +1,283 @@
...
+        .with_context(|| {
+            format!(
+                "failed to spawn `{}` in {}",
+                shell_line("dotnet", &args),
+                workspace.display()
+            )
+        })?;
+    ensure_success(workspace, "dotnet", &args, &output)?;
+
+    let args = ["pack".to_string(), "crates/bindings-csharp/Runtime".to_string()];
```
bfops: inline this

crates/csharp-tests/tests/csharp.rs
```diff
@@ -0,0 +1,283 @@
...
+        let path = entry.path().join(name);
+        if path.exists() {
+            return Ok(path);
+        }
+    }
+    bail!("TRX report {} not found", preferred.display())
+}
+
+fn prepare_csharp_sdk_solution(workspace: &Path) -> Result<()> {
+    let args = ["pack".to_string(), "crates/bindings-csharp/BSATN.Runtime".to_string()];
```
bfops: inline this

crates/csharp-tests/tests/csharp.rs
bfops: also inline this `args`

crates/csharp-tests/tests/csharp.rs
```diff
@@ -0,0 +1,283 @@
...
+    let cwd = workspace.join("sdks/csharp");
+    let args = ["./tools~/write-nuget-config.sh".to_string(), "../..".to_string()];
+    let output = Command::new("bash")
+        .args(&args)
+        .current_dir(&cwd)
+        .output()
+        .with_context(|| format!("failed to spawn `{}` in {}", shell_line("bash", &args), cwd.display()))?;
+    ensure_success(&cwd, "bash", &args, &output)?;
+
+    let args = [
```
bfops: inline this

crates/csharp-tests/tests/csharp.rs
```diff
@@ -0,0 +1,283 @@
...
+    Ok(())
+}
+
+fn run_dotnet_test(
+    suite: &str,
+    cwd: &Path,
+    out_dir: &Path,
+    report_name: &str,
+    args: &Args,
+    extra_args: &[String],
```
bfops: just inline this function into `rust_sdk_tests`.

tools/ci/src/main.rs
```diff
@@ -437,6 +446,41 @@ fn run_typescript_tests() -> Result<()> {
...
+        "--release",
+        "-p",
+        "spacetimedb-cli",
+        "-p",
+        "spacetimedb-standalone",
+        "--features",
+        "spacetimedb-standalone/allow_loopback_http_for_tests"
+    )
+    .run()?;
+    cmd!("cargo", "test", "-p", "spacetimedb-csharp-tests", "--test", "csharp").run()?;
```
bfops: remove `--test csharp`. it's redundant.

crates/csharp-tests/tests/csharp.rs
```diff
@@ -0,0 +1,283 @@
...
+        .args(&args)
+        .current_dir(&cwd)
+        .env("SPACETIMEDB_SERVER_URL", &guard.host_url)
+        .output()
+        .with_context(|| format!("failed to spawn `{}` in {}", shell_line("bash", &args), cwd.display()))?;
+    ensure_success(&cwd, "bash", &args, &output)?;
+    Ok(())
+}
+
+fn ensure_status_success(cwd: &Path, program: &str, args: &[String], status: ExitStatus) -> Result<()> {
```
bfops: remove this function. update the call site to use an `output` so that it can just use `ensure_success`. if typescript has the same issue, do this there as well.

crates/csharp-tests/tests/csharp.rs
bfops: Addressed in aabc71a00. Restored the NuGet config rationale comment from ci.yml in the C# harness.

bfops: Addressed in aabc71a00. Inlined this one-off args binding, and did the same for the other static command arg bindings in the C# harness.

crates/csharp-tests/tests/csharp.rs
bfops: Addressed in aabc71a00. Inlined this static args binding into the Command::args call and passed a literal command line to ensure_success.

crates/csharp-tests/tests/csharp.rs
bfops: Addressed in aabc71a00. Inlined this static args binding as well.

crates/csharp-tests/tests/csharp.rs
bfops: Addressed in aabc71a00. Inlined the dotnet restore args into the command call and passed a literal command line to ensure_success.

crates/csharp-tests/tests/csharp.rs
bfops: Addressed in aabc71a00. Removed ensure_status_success; the list paths now capture Output and use ensure_success. Applied the same change to TypeScript.

tools/ci/src/main.rs
bfops: Addressed in aabc71a00. Removed the redundant --test csharp from the CI cargo test command and verified package-level list mode with cargo test -p spacetimedb-csharp-tests -- --list.

crates/csharp-tests/tests/csharp.rs
bfops: Addressed in aabc71a00. Removed run_sdk_tests and inlined its body into run().

crates/csharp-tests/tests/csharp.rs
```diff
@@ -0,0 +1,269 @@
...
+            format!(
+                "failed to spawn `bash tools~/run-regression-tests.sh` in {}",
+                cwd.display()
+            )
+        })?;
+    ensure_success(&cwd, "bash tools~/run-regression-tests.sh", &output)?;
+    Ok(())
+}
+
+fn ensure_success(cwd: &Path, command_line: &str, output: &Output) -> Result<()> {
```
bfops: change this to just take the `ExitStatus` as a param. Also, update all the call sites so that they are printing their output directly anyway.

do the same in C#.

bfops: Addressed in 452f7eb7f. Both C# and TypeScript ensure_success helpers now take ExitStatus only, and each command call site prints captured stdout/stderr directly before checking the status.

bfops: Follow-up in 6f2877eba: converted the command call sites that only streamed output from .output() plus manual printing to .status(), in both C# and TypeScript harnesses.

crates/csharp-tests/tests/csharp.rs
```diff
@@ -0,0 +1,265 @@
...
+    suite: &str,
+    cwd: &Path,
+    out_dir: &Path,
+    report_name: &str,
+    args: &Args,
+    extra_args: &[String],
+) -> Result<()> {
+    let report = out_dir.join(report_name);
+
+    if args.list {
```
bfops: I think this args.list blob should be in the caller, not in here

crates/csharp-tests/tests/csharp.rs
```diff
@@ -0,0 +1,270 @@
...
+        &["-warnaserror".to_string(), "--no-restore".to_string()],
+    )?;
+    run_regression_tests(&workspace)?;
+
+    Ok(())
+}
+
+fn list_dotnet_tests(cwd: &Path, extra_args: &[String]) -> Result<()> {
+    let mut list_args = vec!["test".to_string(), "--list-tests".to_string()];
+    list_args.extend(extra_args.iter().cloned());
```
bfops: just inline extra_args. this is only called in one place.

crates/csharp-tests/tests/csharp.rs
```diff
@@ -0,0 +1,270 @@
...
+    ensure_success(cwd, &command_line, status)
+}
+
+fn run_dotnet_test(
+    suite: &str,
+    cwd: &Path,
+    out_dir: &Path,
+    report_name: &str,
+    args: &Args,
+    extra_args: &[String],
```
bfops: just inline `extra_args`. this is only called in one place.

crates/csharp-tests/tests/csharp.rs
bfops: Addressed in b11eda9e0. Moved the args.list dispatch up into run(), matching the TypeScript harness shape, and left run_dotnet_test focused on the actual test run path.

crates/csharp-tests/tests/csharp.rs
```diff
@@ -0,0 +1,270 @@
...
+        .with_context(|| format!("failed to spawn `{command_line}` in {}", cwd.display()))?;
+    ensure_success(cwd, &command_line, status)
+}
+
+fn run_dotnet_test(
+    suite: &str,
+    cwd: &Path,
+    out_dir: &Path,
+    report_name: &str,
+    args: &Args,
```
bfops: do not pass all of `args` to this function. instead just pass the required parts.

crates/csharp-tests/tests/csharp.rs
bfops: Addressed in 8b7c00a08. Removed the extra_args parameter from list_dotnet_tests and inlined the fixed SDK list args there.

crates/csharp-tests/tests/csharp.rs
bfops: Addressed in 8b7c00a08. Removed the extra_args parameter from run_dotnet_test and inlined the fixed SDK test args into the test_args construction.

crates/csharp-tests/tests/csharp.rs
```diff
@@ -86,8 +73,8 @@ fn run_dotnet_test(
         out_dir.display().to_string(),
         "--logger".to_string(),
         format!("trx;LogFileName={report_name}"),
+        "--no-restore".to_string(),
     ];
-    test_args.extend(extra_args.iter().filter(|arg| arg.as_str() != "-warnaserror").cloned());
```
bfops: what happened to `-warnaserror`?

crates/csharp-tests/tests/csharp.rs
bfops: Addressed in 0150d6a7c. run_dotnet_test now takes only the filter and passthrough values it needs. I applied the same cleanup to the TypeScript run_tests helper.

crates/csharp-tests/tests/csharp.rs
bfops: Checked in 0150d6a7c: -warnaserror is still present in the normal C# dotnet test argument list, immediately after test. The list path also still includes it.

tools/ci/src/main.rs
```diff
@@ -525,9 +576,6 @@ fn main() -> Result<()> {
             )
             .run()?;
             cmd!("bash", "tools/check-diff.sh", "crates/bindings-csharp").run()?;
-            cmd!("dotnet", "test", "-warnaserror")
-                .dir("crates/bindings-csharp")
-                .run()?;
```
bfops, line 551: this is now a part of the `CsharpTests` variant (see https://github.com/clockworklabs/SpacetimeDB/issues/4859).

crates/csharp-tests/tests/csharp.rs
```diff
@@ -0,0 +1,271 @@
...
+
+    let cwd = workspace.join("sdks/csharp");
+    // Write out the NuGet config file to `nuget.config`. This causes the spacetimedb-csharp-sdk repository
+    // to be aware of the local versions of the `bindings-csharp` packages in SpacetimeDB, and use them if
+    // available. Otherwise, `spacetimedb-csharp-sdk` will use the NuGet versions of the packages.
+    // This means that (if version numbers match) we will test the local versions of the C# packages, even
+    // if they're not pushed to NuGet.
+    // See https://learn.microsoft.com/en-us/nuget/reference/nuget-config-file for more info on the config file.
+    let status = Command::new("bash")
+        .args(["./tools~/write-nuget-config.sh", "../.."])
```
jdetter: In order for this to work on Windows as well it might be best to translate this script into rust and then inline it here, what do you think? Either that or we need an equivalent powershell script so that this will also work on windows and we need to invoke the powershell script instead if we're running on windows.

crates/csharp-tests/tests/csharp.rs
```diff
@@ -0,0 +1,271 @@
...
+        status,
+    )?;
+    Ok(())
+}
+
+fn run_regression_tests(workspace: &Path) -> Result<()> {
+    let guard = SpacetimeDbGuard::spawn_in_temp_data_dir();
+    let cwd = workspace.join("sdks/csharp");
+    let status = Command::new("bash")
+        .args(["tools~/run-regression-tests.sh"])
```
jdetter: same here - what do you think about translating this script into rust and inlining it here?

jdetter:
Very unlikely to be related to this PR but the typescript tests don't pass on Windows:

```
src/sdk/logger.ts:1:42 - error TS2307: Cannot find module 'safe-stable-stringify' or its corresponding type declarations.

1 import { stringify as ssStringify } from 'safe-stable-stringify';
                                           ~~~~~~~~~~~~~~~~~~~~~~~

src/sdk/logger.ts:76:23 - error TS7006: Parameter 'key' implicitly has an 'any' type.

76   ssStringify(value, (key, current) => {
                         ~~~

src/sdk/logger.ts:76:28 - error TS7006: Parameter 'current' implicitly has an 'any' type.

76   ssStringify(value, (key, current) => {
                              ~~~~~~~


Found 3 errors in the same file, starting at: src/sdk/logger.ts:1

 ELIFECYCLE  Command failed with exit code 1.
command failed in C:\Users\boppy\clockwork\SpacetimeDB\crates/bindings-typescript:
  pnpm build
status: exit code: 1
error: test failed, to rerun pass `-p spacetimedb-typescript-tests --test typescript`

Caused by:
  process didn't exit successfully: `C:\Users\boppy\clockwork\SpacetimeDB\target\debug\deps\typescript-f5bc0a56063fb3d6.exe` (exit code: 1)
```

jdetter:
csharp tests don't work on windows due to executing a bash script:
```
Workload updates are available. Run `dotnet workload list` for more information.
: invalid option nameconfig.sh: line 1: set: pipefail
command failed in C:\Users\boppy\clockwork\SpacetimeDB\sdks/csharp:
  bash ./tools~/write-nuget-config.sh ../..
status: exit code: 2
error: test failed, to rerun pass `-p spacetimedb-csharp-tests --test csharp`

Caused by:
  process didn't exit successfully: `C:\Users\boppy\clockwork\SpacetimeDB\target\debug\deps\csharp-2603e4bb2c95094a.exe` (exit code: 1)
```

jdetter (commented):
I think ideally this would work on both Unix and Windows based systems. There seem to be a few problems preventing this from working on Windows and I think that should maybe be a requirement for this PR. What do you think?

bfops:
> Very unlikely to be related to this PR but the typescript tests don't pass on Windows:

@jdetter Interesting. Created https://github.com/clockworklabs/SpacetimeDB/issues/4969 to track.

crates/csharp-tests/tests/csharp.rs
bfops: Yeah, that's fair.

I've created https://github.com/clockworklabs/SpacetimeDB/issues/4970 to do this more generally for all of our bash scripts.

bfops: translating these ones in https://github.com/clockworklabs/SpacetimeDB/pull/4971

jdetter (commented):
Other than the script failures on windows I don't have any objections left to this PR, LGTM. I've tested in places where I've left PR feedback and whatnot. I would like to test one more time once the PR with the script changes merges into this.

PR #4929, opened 2026-05-01
CI - C# test scripts accept `SPACETIMEDB_SERVER_URL` override

bfops:
# Description of Changes

The regression tests in `sdks/csharp` accept a `SPACETIMEDB_SERVER_URL` environment variable override. This makes it easier to run the tests through the same infrastructure as our other tests (see https://github.com/clockworklabs/SpacetimeDB/pull/4921).

None

1

# Testing

actually none. I could be convinced to test. It will be exercised in https://github.com/clockworklabs/SpacetimeDB/pull/4921.


jdetter (approved):
Seems fine, not tested by me 👍

PR #4927, opened 2026-05-01
Revert "Properly handle execution time<->energy conversion in v8 host (#4884)"

jdetter:
# Description of Changes

This reverts commit 458eac8c85845c959dac430bfbafb5ef95e8d164.

This rolls back a pricing change that we made to V8 (TypeScript) based modules. This will be applied as a hotfix to 2.2.0.

# API and ABI breaking changes

None

# Expected complexity level and risk

1 - Just a pricing change.

# Testing

- [x] Tested on staging - I didn't verify that the energy used went down because I don't think I can do that on staging but at least I know this doesn't appear to break anything.

PR #4930, opened 2026-05-01
Stop conflating EnergyQuanta and FunctionBudget

coolreader18:
# Description of Changes

A sort of followup to/unrevert of #4884, an alternative to #4927.

In #3832, we made FunctionBudget a different unit than EnergyQuanta, but there were still many places where we treated them the same and directly converted between them. I got confused by that in #4884, and now this PR properly separates them and corrects the v8 energy calculation.

Also, since we now benchmark the same module in rust and typescript in the form of the keynote benchmark, this patch adds a new assertion to that test that the fuel usage of both modules is within 2X of each other.

3 - touches energy calculation without reverting the problematic #4884, but I'm confident it's correct this time.

# Testing

- [x] Verified that the conversion factors make sense for wasmtime and for v8.
- [x] Assert that typescript and rust keynote bench runs have similar cpu usage


crates/core/src/client/messages.rs
```diff
@@ -408,7 +409,12 @@ impl ToProtocol for TransactionUpdateMessage {
                     args,
                     request_id,
                 },
-                energy_quanta_used: event.energy_quanta_used,
+                // This conversion is lying. We used to tell the client how much eV a transaction
+                // used, but now the database just tracks cpu usage, and it's converted to energy
+                // elsewhere. So, we just pretend that this is `EnergyQuanta` when it's actually
+                // a different unit, and it doesn't really matter to the client anyway.
+                // TODO(noa): maybe we could just have this be zero, unconditionally?
+                energy_quanta_used: EnergyQuanta::new(event.execution_energy_used.get().into()),
```
cloutiertyler: We could take this opportunity to rename `energy_quanta_used` to avoid accidental errors in the future. That seems like it might be worth it.

crates/core/src/host/module_host.rs
```diff
@@ -208,7 +207,7 @@ pub struct ModuleEvent {
     pub function_call: ModuleFunctionCall,
     pub status: EventStatus,
     pub reducer_return_value: Option<Bytes>,
-    pub energy_quanta_used: EnergyQuanta,
+    pub execution_energy_used: FunctionBudget,
```
cloutiertyler: I would also not call this energy if it's not energy. CPU instruction estimate seems more apt.

cloutiertyler (commented):
Left some comments.

crates/core/src/client/messages.rs
coolreader18: This is the ws_v1 client api struct.

joshua-spacetime:
> Keynote transfer fuel comparison: TypeScript=46444571114, Rust=48181790881, relative difference 1.04x

bfops (approved):
I defer my codeowner review to @joshua-spacetime who says this is good.

joshua-spacetime (approved):

PR #4935, opened 2026-05-03
fix: connection lifecycle callbacks

onx2:
# Description of Changes

This pull request improves the handling of connection lifecycle events in the Rust client SDK for SpacetimeDB, particularly distinguishing between connection failures and disconnections. It introduces a new `ConnectionLifecycle` state machine to track connection progress, ensures that the correct callback (`on_connect_error` or `on_disconnect`) is invoked based on the connection state.

**Changes**

* `ConnectionLifecycle` enum to track the connection state (`Connecting`, `Connected`, `Ended`)
* Refactored error handling so that if a connection fails before establishment, the `on_connect_error` callback is invoked; if the connection fails after establishment, the `on_disconnect` callback is invoked. See `end_connection`.
* Updated where disconnections are handled (`advance_one_message_blocking`, `advance_one_message_async`, and message processing) to use `finish_connection`
* Improved handling of user-initiated disconnects during the connection process to avoid reporting them as connection errors and to ensure proper cleanup.

I guess maybe if people relied on the `on_connect_error` to actually fire the `on_disconnect` then this changes that behavior.

Maybe a 2? Seems pretty low risk but I'm still new to the codebase, please double check.

This doesn't fix the websocket issues, that'll be for another day. I noticed websocket.rs has some places it just drops and the error isn't handled properly. We could technically surface that information and run our callbacks with more specific error messages.

# Testing

I had an agent build and run loads of tests for this but didn't commit those since it would have made the PR massive. I was planning on testing locally though to see if I could trigger a connection failure at some point, maybe via an invalid access token.

### DIFF
--- a/sdks/rust/src/db_connection.rs
+++ b/sdks/rust/src/db_connection.rs
@@ -1132,9 +1132,10 @@ but you must call one of them, or else the connection will never progress.
         self
     }
 
-    /// Register a callback to run when the connection is successfully initiated.
+    /// Register a callback to run when the connection is successfully established.
     ///
-    /// The callback will receive three arguments:
+    /// The connection is established after the initial connection message is
+    /// received from the host. The callback will receive three arguments:
     /// - The `DbConnection` which has successfully connected.
     /// - The `Identity` of the successful connection.
     /// - The private access token which can be used to later re-authenticate as the same `Identity`.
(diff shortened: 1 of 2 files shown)


sdks/rust/src/db_connection.rs
```diff
@@ -137,18 +137,25 @@ impl<M: SpacetimeModule> DbContextImpl<M> {
...
-            // then invoke the on_connect callback.
+            // store them, then invoke the on_connect callback.
             ParsedMessage::IdentityToken(identity, token, conn_id) => {
+                let on_connect = {
+                    let mut inner = self.inner.lock().unwrap();
+                    match inner.connection_lifecycle {
+                        ConnectionLifecycle::Connecting => {
+                            inner.connection_lifecycle = ConnectionLifecycle::Connected;
+                            inner.on_connect.take()
+                        }
```
onx2, line 154: `IdentityToken` is actually `InitialConnection` just renamed so we can set the connection lifecycle to connected when this message comes in.

gefjon (approved):
This looks great, thanks! It's a weight off my shoulders to finally have those TODOs resolved.

PR #4942, opened 2026-05-04
CI - Try to fix Internal Tests on `master`

bfops:
# Description of Changes

https://github.com/clockworklabs/SpacetimeDB/pull/4231 changed our CI to always pass a parameter corresponding to the PR number, which.. broke on `master` commits since they don't have a PR number.

None. CI only.

1.

# Testing

I think I don't know how to test this. But it's basically the old behavior on `master` commits, so it should work fine? One hopes?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -889,6 +889,10 @@ jobs:
             const publicRef = (context.eventName === 'pull_request') ? context.payload.pull_request.head.ref : context.sha;
             const publicPrNumber = context.payload.pull_request?.number ?? context.payload.inputs?.pr_number;
             const preDispatch = new Date().toISOString();
+            const inputs = { public_ref: publicRef };
+            if (publicPrNumber) {
+              inputs.public_pr_number = String(publicPrNumber);
+            }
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -889,14 +889,18 @@ jobs:
...
+              inputs.public_pr_number = String(publicPrNumber);
+            }
 
             // Dispatch the workflow in the target repository
             await github.rest.actions.createWorkflowDispatch({
               owner: targetOwner,
               repo: targetRepo,
               workflow_id: workflowId,
               ref: targetRef,
-              inputs: { public_ref: publicRef, public_pr_number: String(publicPrNumber) }
```
jdetter, line 899: If we don't pass a public pr number here I'm assuming then the default value of "" is used (e.g. the value we have specified in the workflow)?

jdetter (approved):
LGTM, just a question on this

.github/workflows/ci.yml
bfops: yes that's correct

PR #4948, opened 2026-05-04
CI - Run unity tests through `cargo test`

### LINKED ISSUE #4924: Run Unity tests under `cargo test`

bfops:
# Description of Changes

The unity testsuite can now be run via `cargo test` using a Rust crate. This has also been threaded into `cargo ci`.

We also updated the unity tests to accept an arbitrary server url instead of hardcoding one, so we can use the existing `SpacetimeDBGuard`.

- `master`
  - [#4972 `cargo ci dlls` -> `cargo regen csharp dlls`](https://github.com/clockworklabs/SpacetimeDB/pull/4972)
    - [#4971 Translate some bash scripts to Rust](https://github.com/clockworklabs/SpacetimeDB/pull/4971)
      - [#4921 CI - `cargo test` runs C# and TypeScript tests](https://github.com/clockworklabs/SpacetimeDB/pull/4921)
        - **[This PR] CI - Run unity tests through `cargo test`**

2

# Testing

- [x] Tests pass, and it lists which tests passed


.github/workflows/ci.yml
```diff
@@ -668,25 +641,11 @@ jobs:
...
-              exit 1
-          }
+      - uses: actions/cache@v3
+        with:
+          path: demo/Blackholio/client-unity/Library
+          key: Unity-${{ github.head_ref }}
+          restore-keys: Unity-
 
       - name: Hydrate Unity SDK DLLs
         run: cargo ci dlls
```
bfops: does this step rely on the nuget.config overrides we used to do earlier?

crates/unity-tests/tests/unity.rs
```diff
@@ -0,0 +1,539 @@
...
+#[cfg(not(unix))]
+fn link_or_copy(src: &Path, dst: &Path) -> io::Result<()> {
+    fs::copy(src, dst).map(|_| ())
+}
+
+struct SpacetimeServer {
+    child: std::process::Child,
+}
+
+impl SpacetimeServer {
```
bfops: it seems like we reimplemented something similar to `SpacetimeDBGuard`? perhaps we should re-use the guard crate like other tests do?

.github/workflows/ci.yml
bfops: Yes, it does rely on the local NuGet override/restore behavior. I kept that in the workflow as `cargo ci dlls` before the Unity meta check, and then run `cargo ci unity-tests --skip-dlls` so the harness does not repeat the DLL hydration. `cargo ci dlls` now owns the same local package pack/restore behavior that was previously spelled out inline in this job.

crates/unity-tests/tests/unity.rs
bfops: Agreed. I removed the custom server wrapper and switched the Unity harness to `SpacetimeDbGuard::spawn_in_temp_data_dir()`. To make that work without forcing port 3000, I updated the Unity client/tests and Blackholio publish script to accept `SPACETIMEDB_SERVER_URL`, defaulting to the old `local` / `http://127.0.0.1:3000` behavior when the env var is not provided.

.github/workflows/ci.yml
```diff
@@ -699,45 +658,10 @@ jobs:
...
       - name: Run Unity tests
-        uses: game-ci/unity-test-runner@v4
-        with:
-          unityVersion: 2022.3.32f1 # Adjust Unity version to a valid tag
-          projectPath: demo/Blackholio/client-unity # Path to the Unity project subdirectory
-          githubToken: ${{ secrets.GITHUB_TOKEN }}
-          testMode: playmode
-          useHostNetwork: true
-          artifactsPath: ""
+        run: cargo ci unity-tests --skip-dlls
```
bfops: please add a comment to why `--skip-dlls` is okay here

bfops: Added. The workflow now calls out that `cargo ci dlls` has already hydrated the Unity SDK DLLs for the meta check, so `cargo ci unity-tests --skip-dlls` avoids repeating the same setup before entering the Rust Unity harness.

bfops: I don't see the comment. Please add it above this line.

bfops: Moved it to the command site. The `Run Unity tests` step now uses a multi-line `run:` block with the explanation immediately above `cargo ci unity-tests --skip-dlls`.

.github/workflows/ci.yml
```diff
@@ -668,25 +641,11 @@ jobs:
...
-        run: |
-          tools/check-diff.sh demo/Blackholio/client-unity/Assets/Scripts/autogen || {
-              echo 'Error: Bindings are dirty. Please run `demo/Blackholio/server-rust/generate.sh`.'
-              exit 1
-          }
+      - uses: actions/cache@v3
+        with:
+          path: demo/Blackholio/client-unity/Library
+          key: Unity-${{ github.head_ref }}
+          restore-keys: Unity-
```
bfops, line 621: was this added?

.github/workflows/ci.yml
```diff
@@ -608,38 +608,11 @@ jobs:
...
-
-      - name: Restore .NET solution
-        working-directory: sdks/csharp
-        run: dotnet restore --configfile NuGet.Config SpacetimeDB.ClientSDK.sln
-
-      # Now, setup the Unity tests.
-      - name: Patch spacetimedb dependency in Cargo.toml
-        working-directory: demo/Blackholio/server-rust
-        run: |
-          sed -i "s|spacetimedb *=.*|spacetimedb = \{ path = \"../../../crates/bindings\" \}|" Cargo.toml
```
bfops, line 593: is this part maintained in the new version?

tools/ci/src/main.rs
```diff
@@ -496,6 +502,26 @@ fn run_csharp_tests() -> Result<()> {
...
+        "-p",
+        "spacetimedb-standalone",
+        "--features",
+        "spacetimedb-standalone/allow_loopback_http_for_tests"
+    )
+    .run()?;
+    if !skip_dlls {
+        run_dlls()?;
+    }
+    cmd!("cargo", "test", "-p", "spacetimedb-unity-tests", "--test", "unity").run()?;
```
bfops: remove `--test unity`

.github/workflows/ci.yml
```diff
@@ -608,38 +608,11 @@ jobs:
...
         uses: actions/checkout@v4
 
-      # Run cheap .NET tests first. If those fail, no need to run expensive Unity tests.
-
       - name: Setup dotnet
         uses: actions/setup-dotnet@v3
         with:
           global-json-file: global.json
 
-      - name: Override NuGet packages
```
jdetter: Did we not need this? I can't seem to find these changes in the `unity.rs` file

jdetter: We at least need to `dotnet pack` and move the binaries into the proper place within the Unity repo, no?

jdetter (commented):
This looks great so far, just one question from me.

I think it's a bit bad dev UX that a unity path is required:
```
could not find Unity. Pass --unity-path, set UNITY_PATH or UNITY_EXECUTABLE, install Unity 2022.3.32f1 in a standard GitHub runner path, or set UNITY_USE_DOCKER=1
error: test failed, to rerun pass `-p spacetimedb-unity-tests --test unity`

Caused by:
  process didn't exit successfully: `C:\Users\boppy\clockwork\SpacetimeDB\target\debug\deps\unity-5022c0874751acf7.exe` (exit code: 1)
Error: command ["cargo", "test", "-p", "spacetimedb-unity-tests", "--test", "unity"] exited with code 1
error: process didn't exit successfully: `target\debug\ci.exe unity-tests --skip-dlls` (exit code: 1)
```

I think the error message here is helpful enough here where it wouldn't take a typical developer long to get the right parameter so I think it's fine. I think a better dev UX would be that if you could specify a unity version then we could try to find that unity version at the typical install paths and then if we can't find it we just give up. We definitely shouldn't just use whatever Unity version is available, I think you should at least have to specify that. What do you think?

jdetter:
One more thing on this, I actually tried to pass `--unity-path` and it didn't work:
```
PS C:\Users\boppy\clockwork\SpacetimeDB> cargo ci unity-tests --skip-dlls --unity-path 'C:\Program Files\Unity\Hub\Editor\2022.3.62f3\Editor'
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.46s
     Running `target\debug\ci.exe unity-tests --skip-dlls --unity-path "C:\Program Files\Unity\Hub\Editor\2022.3.62f3\Editor"`
error: unexpected argument '--unity-path' found

Usage: ci.exe unity-tests --skip-dlls

For more information, try '--help'.
error: process didn't exit successfully: `target\debug\ci.exe unity-tests --skip-dlls --unity-path "C:\Program Files\Unity\Hub\Editor\2022.3.62f3\Editor"` (exit code: 2)
```

jdetter:
The C# tests didn't work immediately for me in WSL so I asked Codex to try to fix it and it came up with these 2 changes that it thought we should make:
```
  1. crates/csharp-tests/tests/csharp.rs

  The test harness was packing C# packages into their normal project output dirs:

  crates/bindings-csharp/*/bin/Release

  Those dirs already contained stale packages, including old 1.11.2 packages. Then the generated NuGet.Config pointed NuGet at those dirty dirs, so restore could resolve
  old local packages instead of the packages just built by the test.

  I changed the harness to:

  - delete/recreate target/csharp-tests/local-feed
  - pack SpacetimeDB.BSATN.Runtime and SpacetimeDB.Runtime into that clean feed
  - write NuGet.Config to consume SpacetimeDB packages from that feed
  - use -p:BuildInParallel=false / --disable-parallel for more reliable MSBuild invocation

  Why needed: the test should be hermetic with respect to local package output. It should test packages built from the current checkout, not whatever old .nupkg happens to
  be lying around.

  2. sdks/csharp/examples~/regression-tests/republishing/*/StdbModule.csproj

  Both republishing module fixtures were pinned to:

  <PackageReference Include="SpacetimeDB.Runtime" Version="1.5.0" />

  But their C# source uses newer runtime/codegen surface, for example table Accessor and [Default(...)]. Those symbols are not available in 1.5.0, so once the stale local
  package behavior was removed, the fixture failed honestly with missing-symbol errors.

  I changed both to:

  <PackageReference Include="SpacetimeDB.Runtime" Version="2.2.*" />

  Why needed: these fixtures are part of the current SDK regression suite and need to compile against the current runtime API. The test is about republishing/migration
  behavior, not compatibility with a 1.5.0 C# runtime.
```

I'm not sure how important it is, but we probably should be bumping these versions in the upgrade version script:
```
diff --git a/sdks/csharp/examples~/regression-tests/republishing/server-initial/StdbModule.csproj b/sdks/csharp/examples~/regression-tests/republishing/server-initial/StdbModule.csproj
index 3d6a76999..6fe4266c4 100644
--- a/sdks/csharp/examples~/regression-tests/republishing/server-initial/StdbModule.csproj
+++ b/sdks/csharp/examples~/regression-tests/republishing/server-initial/StdbModule.csproj
@@ -8,7 +8,7 @@
   </PropertyGroup>

   <ItemGroup>
-    <PackageReference Include="SpacetimeDB.Runtime" Version="1.5.0" />
+    <PackageReference Include="SpacetimeDB.Runtime" Version="2.2.*" />
   </ItemGroup>

 </Project>
diff --git a/sdks/csharp/examples~/regression-tests/republishing/server-republish/StdbModule.csproj b/sdks/csharp/examples~/regression-tests/republishing/server-republish/StdbModule.csproj
index 3d6a76999..6fe4266c4 100644
--- a/sdks/csharp/examples~/regression-tests/republishing/server-republish/StdbModule.csproj
+++ b/sdks/csharp/examples~/regression-tests/republishing/server-republish/StdbModule.csproj
@@ -8,7 +8,7 @@
   </PropertyGroup>

   <ItemGroup>
-    <PackageReference Include="SpacetimeDB.Runtime" Version="1.5.0" />
+    <PackageReference Include="SpacetimeDB.Runtime" Version="2.2.*" />
   </ItemGroup>

 </Project
```

.github/workflows/ci.yml
bfops: this is done by `cargo regen csharp dlls`

bfops:
> The C# tests didn't work immediately for me in WSL so I asked Codex to try to fix it and it came up with these 2 changes that it thought we should make:
> 
> ```
>   1. crates/csharp-tests/tests/csharp.rs
> 
>   The test harness was packing C# packages into their normal project output dirs:
> 
>   crates/bindings-csharp/*/bin/Release
> 
>   Those dirs already contained stale packages, including old 1.11.2 packages. Then the generated NuGet.Config pointed NuGet at those dirty dirs, so restore could resolve
>   old local packages instead of the packages just built by the test.
> 
>   I changed the harness to:
> 
>   - delete/recreate target/csharp-tests/local-feed
>   - pack SpacetimeDB.BSATN.Runtime and SpacetimeDB.Runtime into that clean feed
>   - write NuGet.Config to consume SpacetimeDB packages from that feed
>   - use -p:BuildInParallel=false / --disable-parallel for more reliable MSBuild invocation
> 
>   Why needed: the test should be hermetic with respect to local package output. It should test packages built from the current checkout, not whatever old .nupkg happens to
>   be lying around.
> 
>   2. sdks/csharp/examples~/regression-tests/republishing/*/StdbModule.csproj
> 
>   Both republishing module fixtures were pinned to:
> 
>   <PackageReference Include="SpacetimeDB.Runtime" Version="1.5.0" />
> 
>   But their C# source uses newer runtime/codegen surface, for example table Accessor and [Default(...)]. Those symbols are not available in 1.5.0, so once the stale local
>   package behavior was removed, the fixture failed honestly with missing-symbol errors.
> 
>   I changed both to:
> 
>   <PackageReference Include="SpacetimeDB.Runtime" Version="2.2.*" />
> 
>   Why needed: these fixtures are part of the current SDK regression suite and need to compile against the current runtime API. The test is about republishing/migration
>   behavior, not compatibility with a 1.5.0 C# runtime.
> ```

Is this a problem with the `csharp-tests` (in which case we should fix this in https://github.com/clockworklabs/SpacetimeDB/pull/4921) or was this trying to address issues in the unity tests? What were the issues?

bfops:
> I'm not sure how important it is, but we probably should be bumping these versions in the upgrade version script:
> 
> ```
> diff --git a/sdks/csharp/examples~/regression-tests/republishing/server-initial/StdbModule.csproj b/sdks/csharp/examples~/regression-tests/republishing/server-initial/StdbModule.csproj
> index 3d6a76999..6fe4266c4 100644
> --- a/sdks/csharp/examples~/regression-tests/republishing/server-initial/StdbModule.csproj
> +++ b/sdks/csharp/examples~/regression-tests/republishing/server-initial/StdbModule.csproj
> @@ -8,7 +8,7 @@
>    </PropertyGroup>
> 
>    <ItemGroup>
> -    <PackageReference Include="SpacetimeDB.Runtime" Version="1.5.0" />
> +    <PackageReference Include="SpacetimeDB.Runtime" Version="2.2.*" />
>    </ItemGroup>
> 
>  </Project>
> diff --git a/sdks/csharp/examples~/regression-tests/republishing/server-republish/StdbModule.csproj b/sdks/csharp/examples~/regression-tests/republishing/server-republish/StdbModule.csproj
> index 3d6a76999..6fe4266c4 100644
> --- a/sdks/csharp/examples~/regression-tests/republishing/server-republish/StdbModule.csproj
> +++ b/sdks/csharp/examples~/regression-tests/republishing/server-republish/StdbModule.csproj
> @@ -8,7 +8,7 @@
>    </PropertyGroup>
> 
>    <ItemGroup>
> -    <PackageReference Include="SpacetimeDB.Runtime" Version="1.5.0" />
> +    <PackageReference Include="SpacetimeDB.Runtime" Version="2.2.*" />
>    </ItemGroup>
> 
>  </Project
> ```

Added this to https://github.com/clockworklabs/SpacetimeDB/issues/5121

PR #4950, opened 2026-05-05
CI - version upgrade check happens in `ci.yml`

bfops:
# Description of Changes

Merged the `upgrade-version-check.yml` into `ci.yml`, and moved the business logic under `cargo ci`.

I would also be very open to just removing this test until we choose to define a better suite of tests for `cargo bump-version`.

None. CI only.

1

# Testing

- [x] Ran it locally. It made a diff


jdetter (approved):
Looks fine, not tested by me but this one looks straightforward 👍

Issue #4954, opened 2026-05-05
Relax strict ordering of client messages processing re: procedures


gefjon:
As of writing, messages from a single WS client are always processed and responded to strictly in-order. This is almost correct, but should be loosened slightly for procedures, as they may be long running. `CallProcedure` requests should not block the processing of subsequent messages by the same client.

joshua-spacetime:
Fixed by #4962 and #4973.

PR #4955, opened 2026-05-05
Add some tests of procedure concurrency

gefjon:
# Description of Changes

Add several new tests of concurrency behavior re: procedures.

The new tests are in the SDK tests, 'cause I thought the easiest way to observe this behavior was by connecting a client to a database and calling some functions in it. This is yet another mild misuse of the SDK test suite, as the behavior in question is not in the SDK, it's in the host. The tests also have a new module/client pair added, as we don't (yet?) expose `procedure_sleep_until` to any module languages other than Rust, so we can't implement the same test in any other languages.

### `procedure_reducer_interleaving`

Verifies that a procedure and a reducer can run concurrently, with the procedure cooperatively yielding using `ctx.sleep_until`. Uses two separate connections due to #4954 .

### `procedure_reducer_same_client_interleaved`

Same as previous, but with only a single connection. Now that #4954 is closed, this has the same semantics as previous.

### `procedure_concurrent_with_scheduled_reducer`

Verifies that a non-scheduled procedure can schedule a reducer and then sleep, and the reducer will execute before the procedure wakes back up.

### `scheduled_procedure_scheduled_reducer_not_interleaved`

Schedules a procedure and a reducer, which you might expect to execute concurrently, but don't in a way similar to `procedure_reducer_same_client_not_interleaved`.

This is the behavior that kicked off this whole thing. The scheduler subsystem behaves like a single client, in the sense that it waits for a single function to terminate before scheduling the next function. 

N/a

1 - tests

# Testing

- [x] Ran the tests!

I didn't bother intentionally breaking the host to verify that the tests would fail. It feels pretty apparent to me based on just the test code that we won't see false negatives.


modules/sdk-test-procedure-concurrency/src/lib.rs
```diff
@@ -0,0 +1,89 @@
...
+    ctx.db().procedure_concurrency_row().insert(ProcedureConcurrencyRow {
+        insertion_order: 0,
+        insertion_context: "reducer".into(),
+    });
+}
+
+#[procedure]
+fn procedure_sleep_between_inserts(ctx: &mut ProcedureContext) {
+    ctx.with_tx(|ctx| insert_procedure_concurrency_row(ctx, "procedure_before"));
+    ctx.sleep_until(ctx.timestamp + Duration::from_secs(10));
```
joshua-spacetime: Optional: my only concern here is that these sleeps make this test run for > 30s. Perhaps we could add a polling loop that sleeps for shorter intervals until we observe an insert event?

joshua-spacetime (approved):

modules/sdk-test-procedure-concurrency/src/lib.rs
gefjon: Fixed; where possible we now have a polling loop with a short sleep and a longer timeout.

PR #4962, opened 2026-05-06
Pipeline js module operations

joshua-spacetime:
# Description of Changes

The core motivation for this change is simple: avoid cross-thread handoffs and synchronization on the main execution path.

Before this change, the ingress task for each websocket connection would wait for a completion response on each request before submitting the next request to the database. This was mainly used to guarantee that we delivered message responses in receive-order per connection. However it also meant that for every request, we notified a waiting Tokio task, which potentially incurred kernel-assisted wakeup and scheduler overhead.

Note this design existed mainly for historical reasons. Before the database had a dedicated job thread, requests were not serialized through a single queue. The module instance was gated behind a semaphore which guaranteed mutual exclusion, but it did not guarantee FIFO ordering. Awaiting the completion of each request in `ws_recv_task` was therefore the mechanism that enforced per-connection receive-order semantics. However it now serves primarily as a source of overhead.

Procedures are the important exception. They are not serialized through the main worker queue. Instead they use their own instance pool so as to be able to run concurrently with other requests. However procedures may be composed of multiple transactions and they may effectively yield between transactions. This means that before this change, if a procedure were to yield, it would effectively block all subsequent requests from that client until it returned which is quite undesirable.

So with this change, procedures may execute out of order with other operations received on the same WebSocket. Hence if this is not a desirable property, clients must enforce ordering themselves by waiting for a response before submitting the next request.

### 1. Different instance managers for procedures and everything else

Procedures use a bounded instance pool where each instance is backed by an isolate running in a thread. Reducers and all other operations are serialized through an mpsc queue that feeds a single isolate running in a thread.

Trapped isolates are replaced inline. Only a fatal error within one of the instance threads results in the `ModuleHost` and all its connections being dropped. The host controller will recreate a new `ModuleHost` lazily on the next request.

### 2. New enqueue-only `ModuleHost` interface

`ClientConnection` now calls enqueue-only methods on `ModuleHost` which return immediately after enqueuing on the main instance lane or in the case of a procedure, checking out an available instance and starting the operation.

### 3. Separate `ModuleHost` interfaces for scheduled reducers and scheduled procedures

Scheduled reducers now target the main js instance/worker, while scheduled procedures go through the pool. The scheduler now distinguishes between reducers and procedures and calls the appropriate method.

Note, the scheduler does not pipeline its operations. It waits for each one to complete before scheduling the next operation. This means that a long running procedure will block all other operations from being scheduled. This will need to be fixed at some point, but this patch doesn't change the current behavior.

### 4. Misc

This patch also names the main js worker thread for better diagnostics. It also disables core pinning by default and makes it an explicit opt-in.

This last one is pretty important. The current architecture reduces thread and context switching significantly such that naive core pinning may perform worse than just deferring to the OS scheduler on certain platforms. As it stands, the main motivation which led us to our original core pinning strategy no longer exists, so we should probably just defer to the OS until we've designed a proper scheduler that suits our needs.

As mentioned above, with this change, procedures may execute out of order with other operations received on the same WebSocket. Hence if this is not a desirable property, clients must enforce ordering themselves by waiting for a response before submitting the next request.

4

# Testing

This is mainly a performance oriented refactor, so no additional correctness tests were added. However this patch does touch a lot of code that could probably use more coverage in general. Benchmarks were run to verify expected performance characteristics.

### DIFF
--- a/crates/client-api/src/lib.rs
+++ b/crates/client-api/src/lib.rs
@@ -137,7 +137,7 @@ impl Host {
     pub async fn exec_sql(
         &self,
         auth: AuthCtx,
-        database: Database,
+        _database: Database,
         confirmed_read: bool,
         body: String,
     ) -> axum::response::Result<Vec<SqlStmtResult<ProductValue>>> {
@@ -146,51 +146,44 @@ impl Host {
             .await
             .map_err(|_| (StatusCode::NOT_FOUND, "module not found".to_string()))?;
 
-        let (tx_offset, durable_offset, json) = self
-            .host_controller
-            .using_database(database, self.replica_id, move |db| async move {
-                tracing::info!(sql = body);
-                let mut header = vec![];
-                let sql_start = std::time::Instant::now();
-                let sql_span = tracing::trace_span!("execute_sql", total_duration = tracing::field::Empty,);
-                let _guard = sql_span.enter();
+        tracing::info!(sql = body);
+        let mut header = vec![];
+        let sql_start = std::time::Instant::now();
+        let sql_span = tracing::trace_span!("execute_sql", total_duration = tracing::field::Empty,);
+        let _guard = sql_span.enter();
+        let db = module_host.relational_db().clone();
+        let durable_offset = db.durable_tx_offset();
 
-                let result = sql::execute::run(
-                    db.clone(),
-                    body,
-                    auth,
-                    Some(module_host.info.subscriptions.clone()),
-                    Some(module_host),
-                    &mut header,
-                )
-                .await
-                .map_err(|e| {
-                    log::warn!("{e}");
-                    (StatusCode::BAD_REQUEST, e.to_string())
-                })?;
+        let result = sql::execute::run(
+            db,
+            body,
+            auth,
+            Some(module_host.info.subscriptions.clone()),
+            Some(module_host),
+            &mut header,
+        )
+        .await
+        .map_err(|e| {
+            log::warn!("{e}");
+            (StatusCode::BAD_REQUEST, e.to_string())
+        })?;
 
-                let total_duration = sql_start.elapsed();
-                drop(_guard);
-                sql_span.record("total_duration", tracing::field::debug(total_duration));
+        let total_duration = sql_start.elapsed();
+        drop(_guard);
+        sql_span.record("total_duration", tracing::field::debug(total_duration));
 
-                let schema = header
-                    .into_iter()
-                    .map(|(col_name, col_type)| ProductTypeElement::new(col_type, Some(col_name)))
-                    .collect();
+        let schema = header
+            .into_iter()
+            .map(|(col_name, col_type)| ProductTypeElement::new(col_type, Some(col_name)))
+            .collect();
 
-                Ok::<_, (StatusCode, String)>((
-                    result.tx_offset,
-                    db.durable_tx_offset(),
-                    vec![SqlStmtResult {
-                        schema,
-                        rows: result.rows,
-                        total_duration_micros: total_duration.as_micros() as u64,
-                        stats: SqlStmtStats::from_metrics(&result.metrics),
-                    }],
-                ))
-            })
-            .await
-            .map_err(log_and_500)??;
+        let tx_offset = result.tx_offset;
+        let json = vec![SqlStmtResult {
+            schema,
+            rows: result.rows,
+            total_duration_micros: total_duration.as_micros() as u64,
+            stats: SqlStmtStats::from_metrics(&result.metrics),
+        }];
 
         if confirmed_read && let Some(mut durable_offset) = durable_offset {
             let tx_offset = tx_offset.await.map_err(|_| log_and_500("transaction aborted"))?;
--- a/crates/core/src/client/client_connection.rs
+++ b/crates/core/src/client/client_connection.rs
@@ -7,7 +7,7 @@ use std::sync::Arc;
 use std::task::{Context, Poll};
 use std::time::{Instant, SystemTime};
 
-use super::messages::{OneOffQueryResponseMessage, ProcedureResultMessage};
+use super::messages::ProcedureResultMessage;
 use super::{message_handlers, ClientActorId, MessageHandleError, OutboundMessage};
 use crate::db::relational_db::RelationalDB;
 use crate::error::DBError;
@@ -17,8 +17,6 @@ use crate::host::{
     ReducerCallResult,
 };
 use crate::subscription::module_subscription_manager::BroadcastError;
-use crate::subscription::row_list_builder_pool::JsonRowListBuilderFakePool;
-use crate::util::asyncify;
 use crate::util::prometheus_handle::IntGaugeExt;
 use crate::worker_metrics::WORKER_METRICS;
 use bytes::Bytes;
@@ -965,15 +963,9 @@ impl ClientConnection {
         subscription: ws_v1::SubscribeSingle,
         timer: Instant,
     ) -> Result<Option<ExecutionMetrics>, DBError> {
-        let me = self.clone();
         self.module()
-            .on_module_thread_async("subscribe_single", async move || {
-                let host = me.module();
-                host.subscriptions()
-                    .add_single_subscription(Some(&host), me.sender, me.auth.clone(), subscription, timer, None)
-                    .await
-            })
-            .await?
+            .call_view_add_single_subscription(self.sender(), self.auth.clone(), subscription, timer)
+            .await
     }
 
     pub async fn unsubscribe(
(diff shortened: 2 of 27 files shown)


crates/client-api/src/lib.rs
```diff
@@ -146,51 +146,44 @@ impl Host {
             .await
             .map_err(|_| (StatusCode::NOT_FOUND, "module not found".to_string()))?;
 
-        let (tx_offset, durable_offset, json) = self
```
joshua-spacetime, line 149: The below is just an indentation change due to the removal of `using_database`. `sql::execute::run` is pushed to the module thread instead.

crates/core/src/host/wasmtime/mod.rs
```diff
@@ -129,11 +138,12 @@ impl WasmtimeRuntime {
             .map_err(InitializationError::Instantiation)?;
 
         let module = WasmtimeModule::new(module);
+        let executor_thread_name = wasm_executor_thread_name(&mcc.replica_ctx.database_identity);
 
         let (module, init_inst) = WasmModuleHostActor::new(mcc, module)?;
         Ok(super::module_host::ModuleWithInstance::Wasm {
             module,
-            executor: core.spawn_async_executor(),
+            executor: core.spawn_named_async_executor(executor_thread_name),
```
joshua-spacetime, line 146: Naming the main execution thread for diagnostic purposes.

crates/core/src/host/scheduler.rs
```diff
@@ -53,6 +53,7 @@ enum MsgOrExit<T> {
 enum SchedulerMessage {
     Schedule {
         id: ScheduledFunctionId,
+        function_name: Box<str>,
```
joshua-spacetime: I added the function name to scheduler messages so that the scheduler can determine if a function is a reducer or a procedure and target the appropriate runtime. All the changes in this file are related to that.

crates/core/src/client/message_handlers_v1.rs
```diff
@@ -38,18 +36,6 @@ pub async fn handle(client: &ClientConnection, message: DataMessage, timer: Inst
 
     let module = client.module();
     let mod_info = module.info();
-    let mod_metrics = &mod_info.metrics;
-    let database_identity = mod_info.database_identity;
-    let db = module.relational_db();
-    let record_metrics = |wl| {
-        move |metrics| {
-            if let Some(metrics) = metrics {
-                db.exec_counters_for(wl).record(&metrics);
```
joshua-spacetime, line 47: These metrics are now merged with the metrics of the open transaction in module_subscription_manager.

crates/core/src/client/message_handlers_v2.rs
```diff
@@ -28,44 +26,21 @@ pub(super) async fn handle_decoded_message(
...
 ) -> Result<(), MessageHandleError> {
-    let module = client.module();
-    let mod_info = module.info();
-    let mod_metrics = &mod_info.metrics;
-    let database_identity = mod_info.database_identity;
-    let db = module.relational_db();
-    let record_metrics = |wl| {
-        move |metrics| {
-            if let Some(metrics) = metrics {
-                db.exec_counters_for(wl).record(&metrics);
```
joshua-spacetime, line 39: These metrics are now merged with the metrics of the open transaction in module_subscription_actor.

crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -666,14 +666,15 @@ impl ModuleSubscriptions {
...
-        let (tx, tx_offset, trapped) =
+        let (mut tx, tx_offset, trapped) =
             self.materialize_views_and_downgrade_tx(mut_tx, instance, &query, auth.caller())?;
 
         let (table_rows, metrics) = return_on_err_with_sql_bool!(
             self.evaluate_initial_subscription(sender.clone(), query.clone(), &tx, &auth, TableUpdateType::Subscribe),
             query.sql(),
             send_err_msg
         );
+        tx.metrics.merge(metrics);
```
joshua-spacetime, line 677: We merge subscription execution metrics with the open transaction metrics.

crates/core/src/subscription/module_subscription_actor.rs
```diff
@@ -1597,6 +1612,16 @@ impl ModuleSubscriptions {
         tx: MutTx,
     ) -> Result<CommitAndBroadcastEventResult, DBError> {
         let subscription_metrics = &self.metrics.update;
+        if let (Some(timer), Some(reducer)) = (event.timer, event.function_call.reducer.as_ref()) {
+            WORKER_METRICS
+                .request_round_trip
```
joshua-spacetime: Note, this is really not the right place to be recording round trip time for the request. It should be recorded right before we send the response on the wire. Previously it was recorded by the module host caller after the call returned which was also not right.

crates/core/src/config.rs
```diff
@@ -179,6 +206,46 @@ pub struct LogConfig {
     pub directives: Vec<String>,
 }
 
+#[derive(Clone, Copy, Debug)]
+pub struct V8Config {
+    pub procedure_instance_pool_size: NonZeroUsize,
```
joshua-spacetime, line 211: The procedure instance pool is configurable.

joshua-spacetime:
I'm labeling this as breaking due to the change in execution order for procedures.

gefjon:
I'm not planning on reviewing this PR, but I wanted to weigh in that this breaking change is a good thing, and that this PR will close #4954 because of it. We should make sure to document the change in our release notes, but it's a good thing.

crates/core/src/db/relational_db.rs
```diff
@@ -1717,21 +1718,44 @@ pub async fn snapshot_watching_commitlog_compressor(
...
+            }
+
+            let durability = durability.clone();
+            if let Err(err) = asyncify(move || durability.compress_segments(&[segment_offset])).await {
+                tracing::warn!("failed to compress commitlog segment {segment_offset}: {err}");
+                compressed_all_segments = false;
+                break;
+            }
+            last_compressed_segment = Some(segment_offset);
+            tokio::task::yield_now().await;
```
coolreader18: Is this necessary? If so it could probably use a comment.

crates/core/src/host/v8/mod.rs
```diff
@@ -1339,288 +1300,546 @@ where
...
+                let (hooks, module_common) = match startup_result {
+                    Err(_) => {
+                        if let Some(result_tx) = startup_result_tx.take() {
+                            let _ = result_tx
+                                .send(Err(anyhow::anyhow!("JS worker panicked during startup")))
+                                .inspect_err(|_| {
+                                    // This should never happen as we immediately `.recv` on the
+                                    // other end of the channel, but sometimes it gets cancelled.
+                                    log::error!("startup result receiver disconnected");
+                                });
```
coolreader18: This and the similar one below could just be `if tx.send().is_err() { log::error!() }`

crates/core/src/client/client_connection.rs
```diff
@@ -653,6 +647,110 @@ impl<T> MeteredSender<T> {
...
+    }
+
+    pub fn poll_recv(&mut self, cx: &mut Context<'_>) -> Poll<Option<T>> {
+        self.inner.poll_recv(cx).map(|maybe_item| {
+            maybe_item.inspect(|_| {
+                if let Some(gauge) = &self.gauge {
+                    gauge.dec()
+                }
+            })
+        })
```
coolreader18: Fewer combinators might be clearer here.
```suggestion
        let poll = self.inner.poll_recv(cx);
        if let Poll::Ready(Some(_)) = poll && let Some(gauge) = &self.gauge {
            gauge.dec()
        }
        poll
```

crates/core/src/host/scheduler.rs
coolreader18: It'd be nice if this could be an `Arc<str>`, though it's probably not a huge deal.

crates/core/src/util/jobs.rs
```diff
@@ -86,7 +86,7 @@ impl JobCores {
     /// and runs all databases in the `global_runtime`.
     pub fn from_pinned_cores(cores: impl IntoIterator<Item = CoreId>) -> Self {
         let cores: IndexMap<_, _> = cores.into_iter().map(|id| (id, CoreInfo::default())).collect();
-        let inner = if cfg!(feature = "no-job-core-pinning") || cores.is_empty() {
+        let inner = if cfg!(not(feature = "core-pinning")) || cores.is_empty() {
```
coolreader18, line 89: Not that this doesn't work, but I feel like this could just be an environment variable or config value or something.

coolreader18 (approved):
I have some ultimately inconsequential nits, but otherwise LGTM!

sephirith:
Does Maincloud need to be updated to support this?

joshua-spacetime:
@sephirith this has already been deployed to Maincloud.

PR #4963, opened 2026-05-06
Minor docs additions based on a discussion in the public Discord

gefjon:
# Description of Changes

A user in the public Discord was confused about the intended way(s) to interact with SpacetimeDB, and was attempting to (mis)use the HTTP API's SQL endpoint in production rather than or in addition to the WebSocket API.
After some discussion, we determined these additions to the documentation would be helpful.

N/a

1

# Testing

N/a


JulienLavocat (approved):
LGTM

PR #4973, opened 2026-05-07
Pipeline wasm module operations

joshua-spacetime:
# Description of Changes

The equivalent changes as https://github.com/clockworklabs/SpacetimeDB/pull/4962 but for wasm.

See https://github.com/clockworklabs/SpacetimeDB/pull/4962

4

# Testing

See https://github.com/clockworklabs/SpacetimeDB/pull/4962

### DIFF
--- a/crates/cli/src/subcommands/subscribe.rs
+++ b/crates/cli/src/subcommands/subscribe.rs
@@ -240,6 +240,8 @@ enum Error {
     },
     #[error("encountered failed transaction: {reason}")]
     TransactionFailure { reason: Box<str> },
+    #[error("encountered error in initial subscribe: {reason}")]
+    SubscribeFailure { reason: Box<str> },
     #[error("error formatting response: {source:#}")]
     Reformat {
         #[source]
@@ -295,6 +297,9 @@ where
                 }
                 break;
             }
+            ws_v1::ServerMessage::SubscriptionError(error) => {
+                return Err(Error::SubscribeFailure { reason: error.error });
+            }
             ws_v1::ServerMessage::TransactionUpdate(ws_v1::TransactionUpdate { status, .. }) => {
                 return Err(match status {
                     ws_v1::UpdateStatus::Failed(msg) => Error::TransactionFailure { reason: msg },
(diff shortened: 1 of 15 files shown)


Centril (commented):
Looks good, although I'll leave it to Noa to also review and approve.

crates/cli/src/subcommands/subscribe.rs
```diff
@@ -240,6 +240,8 @@ enum Error {
     },
     #[error("encountered failed transaction: {reason}")]
     TransactionFailure { reason: Box<str> },
+    #[error("encountered error in initial subscribe: {reason}")]
+    SubscribeFailure { reason: Box<str> },
```
jdetter, line 244: So when we roll this change to subscribe do we also need to roll this change to maincloud in order for this error to actually be used or no?

jdetter (commented):
Quick q on this

crates/cli/src/subcommands/subscribe.rs
joshua-spacetime: No

joshua-spacetime: We just weren't handling subscription errors before. So an old cli will just continue to ignore them. A new cli will handle them regardless of server change.

crates/core/src/host/module_host.rs
```diff
@@ -1411,45 +1629,14 @@ impl ModuleHost {
...
-            }
-        })
+    fn wasm_instance_manager(
+        host: &WasmtimeModuleHost,
+        kind: InstanceKind,
+    ) -> Arc<ModuleInstanceManager<Arc<super::wasmtime::Module>>> {
+        match kind {
+            InstanceKind::Main => host.main_instance.clone(),
+            InstanceKind::Procedure => host.procedure_instances.clone(),
+        }
```
coolreader18: Seems like this should be a method on `WasmtimeModuleHost`

crates/core/src/host/module_host.rs
```diff
@@ -1541,50 +1738,170 @@ impl ModuleHost {
...
+                        log::warn!("{kind} {label} panicked");
+                        on_panic();
+                    });
+
+                    drop(timer_guard);
+                    f().await;
+                });
+                Ok(())
+            }
+            ModuleHostInner::Js(_) => unreachable!("enqueue_wasm_job should only be used for wasm"),
```
coolreader18: Don't love this, it seems to me like this should method should be on `WasmtimeModuleHost` and the wasm `AsyncFnOnce` parameter for `enqueue_main_operation` should be more like the `js` one, but it's fine for now.

coolreader18 (commented):
LGTM!

jdetter (approved):
Cli changes look fine to me, not tested by me

PR #4977, opened 2026-05-07
CI - Fix `gen-quickstart` check

bfops:
# Description of Changes

CI was running `gen-quickstart.sh` and then checking for a diff.. but it was checking in the wrong directory.

I have also regenerated the files because the fixed check was failing.

None.

1

# Testing

- [x] CI passes
- [x] updated CI failed without the changes to the other files


sdks/csharp/tests~/SnapshotTests.cs
```diff
@@ -257,8 +257,8 @@ private static List<byte> EncodeBsatnString(string value)
         return [.. o.ToArray()];
     }
 
-    private static readonly string USER_TABLE_NAME = "User";
-    private static readonly string MESSAGE_TABLE_NAME = "Message";
+    private static readonly string USER_TABLE_NAME = "user";
+    private static readonly string MESSAGE_TABLE_NAME = "message";
```
bfops, line 261: I don't know why, but these changes seemed to be needed after a regen. Perhaps these bindings had not been regenerated since The Great Case Unification.

jdetter (approved):
Not tested by me but this change looks good, thanks Zeke 👍

.github/workflows/ci.yml
```diff
@@ -830,10 +830,9 @@ jobs:
           ln -sf $CARGO_HOME/bin/spacetimedb-cli $CARGO_HOME/bin/spacetime
 
       - name: Check quickstart-chat bindings are up to date
-        working-directory: sdks/csharp
         run: |
-          bash tools~/gen-quickstart.sh
```
jdetter, line 835: kinda sucks that this wasn't failing, should we be doing `set -e` on all of these bash scripts?

bfops: I believe GitHub does that already. It wasn't failing because we were checking the wrong directory, so there was indeed no diff. I don't know that it's a detectable failure.

bfops: unless gen-quickstart were to print out / return the location that it wrote to or something

jdetter: misunderstood where the script was located, thanks for the clarification 👍

PR #4982, opened 2026-05-08
Abstract `SnapshotWorker` and `durability::Local`

Shubham8287:
# Description of Changes

  - Introduce `SnapshotRepo` as the snapshot backend trait and decouple `SnapshotWorker` from the
  filesystem-backed implementation.
    - Dynamic dispatch-ing is not needed but done to avoid propagating generics all over the place.
    - Snapshotting is also not on the hot path, and durability also follows the same `dyn` pattern.

  - Make commitlog `durability::Local` generic over `Repo`, so the same durability implementation can be
  reused with in-memory and fault-injected repositories.

  - Add a few small API exposure changes under `cfg(test)` for DST integration.

NA.

2

# Testing

Existing tests should be enough.


crates/commitlog/src/lib.rs
```diff
@@ -374,15 +403,12 @@ impl<T> Commitlog<T> {
...
-    pub fn size_on_disk(&self) -> io::Result<SizeOnDisk> {
-        let inner = self.inner.read().unwrap();
-        inner.repo.size_on_disk()
-    }
 }
 
-impl<T: Encode> Commitlog<T> {
+impl<T: Encode, R> Commitlog<T, R>
+where
+    R: Repo,
```
kim, line 438: nit: Should we put all trait bounds in the `where` clause or all of them in the `impl` type params?

crates/core/src/db/relational_db.rs
```diff
@@ -771,6 +771,19 @@ impl RelationalDB {
         r
     }
 
+    #[cfg(any(feature = "test", test))]
+    #[tracing::instrument(level = "trace", skip_all)]
+    pub fn try_begin_mut_tx(&self, isolation_level: IsolationLevel, workload: Workload) -> Option<MutTx> {
```
kim, line 776: Seems unrelated

crates/core/src/db/relational_db.rs
```diff
@@ -1009,7 +1022,7 @@ impl RelationalDB {
         Ok(self.inner.alter_table_row_type_mut_tx(tx, table_id, column_schemas)?)
     }
 
-    pub(crate) fn add_columns_to_table(
+    pub(crate) fn add_columns_to_table_mut_tx(
```
kim, line 1025: Also unrelated?

crates/core/src/db/relational_db.rs
```diff
@@ -1021,6 +1034,17 @@ impl RelationalDB {
             .add_columns_to_table_mut_tx(tx, table_id, column_schemas, default_values)?)
     }
 
+    #[cfg(any(feature = "test", test))]
+    pub fn add_columns_to_table(
```
kim, line 1038: Here too

crates/datastore/src/locking_tx_datastore/datastore.rs
```diff
@@ -924,6 +923,29 @@ impl MutTx for Locking {
 }
 
 impl Locking {
+    #[cfg(any(feature = "test", test))]
+    pub fn try_begin_mut_tx(&self, _isolation_level: IsolationLevel, workload: Workload) -> Option<MutTxId> {
```
kim, line 927: Here too

crates/commitlog/src/repo/mod.rs
```diff
@@ -144,6 +144,18 @@ pub trait Repo: Clone + fmt::Display {
     }
 }
 
+/// Marker for repos that do not require an external lock file.
+///
+/// Durability implementations can use this to expose repo-backed opening
+/// only for storage backends where skipping the filesystem `db.lock` cannot
+/// violate single-writer safety.
+pub trait RepoWithoutLockFile: Repo {}
```
kim, line 160: We are actually locking the repo in private, using a wrapper. Maybe we should do this here, too?

crates/core/src/db/relational_db.rs
Shubham8287: yeah, as I have put in description. They are small, trivial and under `test` flag.
> Add a few small API exposure changes under cfg(test) for DST integration.

kim (approved):

crates/snapshot/src/lib.rs
```diff
@@ -261,6 +266,28 @@ impl UnflushedSnapshotInner {
     }
 }
 
+pub trait PendingSnapshot: Send {
+    /// Sync all snapshot state and return the finalized transaction offset.
+    fn sync_all(self: Box<Self>) -> Result<TxOffset, SnapshotError>;
+}
+
+pub type BoxedPendingSnapshot = Box<dyn PendingSnapshot>;
+pub type DynSnapshotRepo = dyn SnapshotRepo<Pending = BoxedPendingSnapshot>;
```
gefjon, line 271:
```suggestion
// We pass snapshot repos as `dyn` rather than compile-time generics for convenience,
// as threading generics through a deep callstack would be a hassle.
pub type BoxedPendingSnapshot = Box<dyn PendingSnapshot>;
pub type DynSnapshotRepo = dyn SnapshotRepo<Pending = BoxedPendingSnapshot>;
```

I'm wary of writing code like this without comments, 'cause it can be difficult to tell after the fact what was intentional choice based on design, versus what was convenience or shortcutting. I think a reader's assumption is that most of the code is intentional, so I'm trying to move towards leaving comments on the stuff that isn't, so that a reader knows it's safe to remove.

gefjon (approved):

crates/commitlog/src/repo/mod.rs
Shubham8287: FYI, I reverted the file-lock ownership change in the `Repo` implementation commit.

The reason is that, in the public implementation, the Durability `Actor` is not the sole owner of `Repo`. `Repo` is also `Arc`’d with the `Local` handle, which can sometimes lead to flaky `db.lock` cleanup behavior.

I think this is related to `durability::spawn_close` spawning a new task while still holding `durability`, but there may be other  similar places too.

PR #4986, opened 2026-05-08
Fix segfault in v8

coolreader18:
# Description of Changes

This was originally introduced in #4302; essentially, we stopped unconditionally setting `HookFunctions.recv` to undefined and started setting it to the value stored in `ctx.get_embedder_data(RECV_SLOT_INDEX)`. However, in the code path for v1 js modules, we never actually set that embedder data slot, and so recv was a garbage value.

1: concentrated fix

# Testing

- [x] Repro no longer segfaults.


cloutiertyler (approved):

cloutiertyler:
I approved this because I saw with my own two eyes that it resolved the problem.

PR #4990, opened 2026-05-11
Add recursive mounts for submodules [Namespaces Part 1]

### LINKED ISSUE #5086: Namespaces
Parent tracker for parts [1](#4990) thru [8](#5081).

aasoni:
# Description of Changes

Add a recursive `mounts` fields to ModuleDef that references submodules.

This should not be a breaking change. However new V10 modules that are built with this change will not be publishable to older hosts as the new field won't be recognized.

2

# Testing

Mostly tested using ci tests. Will be tested more thoroughly in the typescript PR that will use the new mounts field.


coolreader18 (approved):
LGTM!

PR #4995, opened 2026-05-11
CI: skip Internal Tests dispatch for docs-only changes

cloutiertyler:
## Summary

- Add a `dorny/paths-filter` step to the `internal-tests` job that checks for any changed file outside `docs/`.
- Gate the private-repo dispatch and the wait-for-completion steps on that filter, so a docs-only PR no longer fires off the private CI workflow.
- The job itself still runs and completes successfully, so any branch protection requiring the `Internal Tests` check continues to be satisfied.

## Test plan

- [ ] Open a docs-only follow-up PR and confirm `Internal Tests` reports success without dispatching a private run.
- [ ] Open a PR touching `crates/` (or anything non-docs) and confirm the private dispatch still fires as before.


bfops (approved):
seems like a small fix for an edge case to me, but no issue with the functionality

cloutiertyler:
This may be a small fix, but I expect this type of thing to grow considerably. The nature of a monorepo is that things are in there which are unrelated to the correctness of other things. If it reduces friction to getting docs out there, I think it's worth it.

PR #5016, opened 2026-05-13
Deterministic runtime crate

Shubham8287:
# Description of Changes.

Introduces deterministic runtime crate.
Integrate it with RelationalDB.

I think best steps to review:
- Read the [README](https://github.com/clockworklabs/SpacetimeDB/blob/shub/sim/crates/runtime/README.md) of runtime crate.
- Look at the integration with existing crates - `durability`, `core`, `snapshot`, etc.
- Read runtime crate's code.

Draft branch to Test code - https://github.com/clockworklabs/SpacetimeDB/pull/5019

NA

Does not intend to change any production functionality, but it's big code.

# Testing

- new crate contains unit and integration tests.
- Existing tests should work for production.


crates/runtime/src/lib.rs
```diff
@@ -0,0 +1,338 @@
...
+pub struct AbortHandle {
+    inner: AbortHandleInner,
+}
+
+enum JoinHandleInner<T> {
+    #[cfg(feature = "tokio")]
+    Tokio(tokio::task::JoinHandle<T>),
+    #[cfg(feature = "simulation")]
+    Simulation(sim::JoinHandle<T>),
+    Detached(PhantomData<T>),
```
kim, line 54: This one is interesting and could use some commentary!

crates/runtime/src/sim_std.rs
```diff
@@ -0,0 +1,207 @@
...
+// small means all entropy decisions stay centralized in `getrandom`.
+#[unsafe(no_mangle)]
+#[inline(never)]
+unsafe extern "C" fn getentropy(buf: *mut u8, buflen: usize) -> i32 {
+    if buflen > 256 {
+        return -1;
+    }
+    match unsafe { getrandom(buf, buflen, 0) } {
+        -1 => -1,
+        _ => 0,
```
kim, line 180: Is it a good idea to map all non-zero returns to zero?

crates/runtime/DETERMINISM_COVERAGE.md
```diff
@@ -0,0 +1,50 @@
...
+| `HashMap` randomized iteration | Audited | Runtime and caller code | Runtime does not force deterministic hash seeding; correctness must not depend on iteration order | Hidden ordering dependencies cause flaky replay | Prefer ordered maps or explicit sorting where observable order matters |
+| `tokio::sync` primitives | Constrained | Core crates above runtime | These can be replay-compatible only when all participating tasks remain simulator-owned and progress stays on simulator-controlled async paths | Wake ordering or blocking semantics diverge once code depends on a real runtime or host-driven progress | Audit per primitive and push deep-core paths toward runtime-owned or single-threaded structures |
+| `parking_lot::{}` and `std::sync::{}` | Constrained | Core crates, especially datastore | Safe only where access stays single-threaded or non-contended under DST | Host synchronization leaks nondeterministic acquisition order | Keep out of deep-core execution paths; prefer runtime-owned or single-threaded structures |
```
kim, line 38: This seems to be quite a limitation. Are we planning to provide custom synchronization primitives?

crates/runtime/src/sim/time/mod.rs
```diff
@@ -0,0 +1,308 @@
...
+};
+use sleep::wake_all;
+use spin::Mutex;
+
+pub use sleep::Sleep;
+
+/// Shared virtual clock and timer registry for one simulation runtime.
+///
+/// Virtual clock that only advances when explicitly driven — no wall-clock
+/// progression, like Tokio's time-pause mode.
```
kim: Can you explain in more detail (in the comment) what the difference to tokio paused-time are? Or is this comment saying that it is equivalent to tokio?

crates/snapshot/tests/remote.rs
```diff
@@ -227,14 +227,14 @@ impl SourceSnapshot {
 
 async fn create_snapshot(repo: Arc<SnapshotRepository>) -> anyhow::Result<TxOffset> {
     let start = Instant::now();
-    let rt = tokio::runtime::Handle::current();
+    let rt = spacetimedb_runtime::Handle::tokio_current();
```
kim, line 230: Does this mean that we should eventually thread through runtime handles from `main`?

crates/runtime/src/sim_std.rs
```diff
@@ -0,0 +1,207 @@
...
+        eprintln!("note: use simulator tasks instead.");
+        return -1;
+    }
+
+    type PthreadAttrInit = unsafe extern "C" fn(*mut libc::pthread_attr_t) -> libc::c_int;
+    static PTHREAD_ATTR_INIT: OnceLock<PthreadAttrInit> = OnceLock::new();
+    let original = PTHREAD_ATTR_INIT.get_or_init(|| unsafe {
+        // `RTLD_NEXT` skips this interposed function and finds the libc
+        // implementation that would have been called without the simulator.
+        let ptr = libc::dlsym(libc::RTLD_NEXT, c"pthread_attr_init".as_ptr().cast());
```
kim, line 128: Wild.

crates/runtime/src/sim_std.rs
Shubham8287: `getentropy` is suppose to return only -1 or 0 as per the spec - https://man7.org/linux/man-pages/man3/getentropy.3.html. `getrandom` returns the `buf` size  which should be discarded here.

crates/runtime/DETERMINISM_COVERAGE.md
Shubham8287: I think we may eventually need some custom synchronization primitives, mainly for places that rely on blocking behavior and can't be refactored to use async primitives (moving them to async path should be the first choice)

  For async paths,, `tokio::sync` should take us pretty far. Tokio documents it as [runtime agnostic](https://docs.rs/tokio/latest/tokio/sync/index.html#runtime-compatibility), so the non-blocking APIs should work with our runtime as long as execution stays inside the normal wake/poll path. Note, `blocking_send` and brothers are the problem here, since they hand control back to the OS by parking the thread or doing similar stuff.

  `datastore::Locking` uses `parking_lot`. Which is only deterministic until not contended. As module runs already single-threaded, contention doesn't happen from there.  I could be wrong, but I think it could only happen due to subscriptions so we are fine until we do testing without subscriptions and then we either have to switch this lock to async one or implement custom.

crates/snapshot/tests/remote.rs
Shubham8287: umm yes, as of now, but we can also implement something like Tokio's context if needed in future: https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#method.enter. I deffered to do it, until we have to do it.

crates/runtime/src/sim/time/mod.rs
Shubham8287: Yeah, It is conceptually similar to tokio paused-time written to work with our custom executor with handful of APIs. I have updated the comment.

crates/runtime/src/sim_std.rs
kim: Uh, errno

crates/runtime/DETERMINISM_COVERAGE.md
kim: Ok, it is not entirely clear to me which async codepaths we intend to simulate, so fine to defer until it becomes an issue.

kim (approved):

crates/runtime/src/sim_std.rs
Shubham8287: overriding `libc` methods turning out to be limited for detecting source of nondeterminism. I am experimenting with eBPF, which looks more powerful.

crates/runtime/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+- **Single-threaded runtime.** The simulator exposes interleaving and timeout bugs, but not bugs that require true parallel execution. The direction is to keep deep-core code single-threaded or close to thread-per-core; simulating real parallelism is out of scope.
+
+- **No built-in network, storage, or I/O simulation.** This crate provides deterministic execution primitives only. Higher-level harnesses should model message delivery, disk behavior, and failures.
+
+- **Not a Tokio replacement.** This crate does not aim to simulate APIs like `tokio::net` or `tokio::fs`. Code that depends on them needs a higher-level abstraction boundary.
+
+- **Zero dependency.** The simulation core in `sim/` is already `no_std + alloc`. The `sim_std` module is a thin OS-facing wrapper — the std dependency lives there, not in the simulation core itself. It stays until the application logic above this crate also moves to `no_std`.
+
+## Current Limitations
```
jsdt, line 54: You touched on this in the point about it being a single-threaded runtime, but this is also limited in how granularly it can interleave executions of different tasks, which is going to make it impossible to produce some race conditions that would be possible in a multi-threaded runtime.  If I'm reading this correctly, the sim executor picks one of the tasks, then it polls that task (which runs the task until it finishes or an `await` returns `Pending`), then it picks a task again. This means we can only mess with scheduling when a task yields.

For a simple example of a case of an execution that we can't simulate:
```
async fn task_a(mut rx: Receiver<Msg>) {
    log("A: waiting");
    rx.recv().await;      // Pending until Task B sends.
    log("A: received a message");
    do_a_work();          // Cannot run until Task B yields or finishes.
}

async fn task_b(mut tx: Sender<Msg>) {
    log("B: running");
    tx.send(Msg).await;   // Ready immediately; wakes Task A but does not yield.
    do_b_work();      // Always runs before Task A resumes.
    log("B: finished working");
}
```

We could never produce:
```
A: waiting
B: running
A: received a message
B: finished working
```

So if there is a bug when `do_a_work` runs before `do_b_work`, we would never be able to catch it.

I think the fix for this specific case would be to have channels be part of the runtime crate, so that when `send` is called, we can insert a `yield_now` call in the sim version (if `send` is async).

More generally, if the goals is to be able to simulate any possible race condition, we need to be able to suspend/interleave execution at any point in the code that either affects other tasks/threads or is affected by other tasks/threads.

For things with an `async` api, having a `sim` version that inserts a `yield_now` will make it possible to interleave, but addressing this for sync apis that affect other threads will require changing the executor. Even channels have some sync functions that we would want to be able to suspend (like `try_recv` or `send` on an `UnboundedSender`).

If we want this to work for sync functions that interact with other tasks, then the sim executor will need more than one thread.

Shubham8287: As we also discussed this out of band, but putting it here too:

> More generally, if the goals is to be able to simulate any possible race condition, we need to be able to suspend/interleave execution at any point in the code that either affects other tasks/threads or is affected by other tasks/threads.

Agreed. With the current DB architecture, a single-threaded executor cannot simulate every possible race condition.

That said, I think the broader goal should be to move toward an architecture where data races are impossible by design. In practice, that means moving toward a single-threaded runtime model, which also aligns with our thread-per-core principle.

Practically, I think this would mean allowing only the database thread to touch the `datastore`, while potentially keeping separate threads for things like `ws` message handoff. If the “deep core” part of our DB runs single-threaded, we should be able to simulate all of it. The parts that must remain multi-threaded should live outside the “core,” and have to be tackled later.


> If we want this to work for sync functions that interact with other tasks, then the sim executor will need more than one thread.

I think putting more than one thread in executor make the tests not replayable by seed.

So, maybe I think current scope should be to keep executor single-threaded only, even if does not let us explore all race conditions, it is still very helpful for other classes of bugs - Schema bugs, replayability bugs, fault-tolerance related, etc from day 1.

Shubham8287: Btw, I’m also checking out https://docs.rs/loom/latest/loom/ to understand how it detects multithread-specific bugs while still being deterministic, and whether we can leverage something from it.

jsdt: > I think putting more than one thread in executor make the tests not replayable by seed.

I'm sure that is not the case. We can have multiple threads while staying replayable/deterministic as long as we ensure that only one thread is unparked at any given time. I'm working on a PoC of that today.

> I think current scope should be to keep executor single-threaded only, even if does not let us explore all race conditions, it is still very helpful for other classes of bugs - Schema bugs, replayability bugs, fault-tolerance related, etc from day 1.

I think that is reasonable. Maybe if the multi-threaded version is easy to implement, it would make sense to add it sooner, but it can probably be added later without breaking what we have now.

> That said, I think the broader goal should be to move toward an architecture where data races are impossible by design. In practice, that means moving toward a single-threaded runtime model, which also aligns with our thread-per-core principle.

Making races impossible sounds great, but even with a thread-per-core design, we will almost certainly still have multiple threads interacting with each other.

Shubham8287: > I'm sure that is not the case. We can have multiple threads while staying replayable/deterministic as long as we ensure that only one thread is unparked at any given time. I'm working on a PoC of that today.

Right, that's a good approach which you already explain but missed from my mind :)

crates/runtime/src/lib.rs
```diff
@@ -0,0 +1,350 @@
...
+#[cfg(feature = "tokio")]
+pub type TokioHandle = tokio::runtime::Handle;
+
+#[derive(Clone)]
+pub enum Handle {
+    #[cfg(feature = "tokio")]
+    Tokio(TokioHandle),
+    #[cfg(feature = "simulation")]
+    Simulation(sim::Handle),
+}
```
cloutiertyler, line 27: I don't want this comment to block merging this PR, so feel free to close it, but I'm curious about the chose of enums vs traits to implement this. Seems reasonable, but could you comment on that choice briefly?

crates/runtime/src/sim/executor/mod.rs
```diff
@@ -0,0 +1,793 @@
...
+            panic!("no runnable tasks; all simulated tasks are blocked");
+        }
+    }
+
+    /// Drain the runnable queue, selecting tasks in deterministic RNG order.
+    ///
+    /// Paused-node tasks are diverted into that node's paused buffer instead of
+    /// being polled immediately.
+    fn run_all_ready(&self) {
+        while let Some(runnable) = self.queue.try_recv_random(&self.rng) {
```
cloutiertyler, line 511: As we discussed, I also think that round robin order is correct and possibly preferrable (within a single node).

crates/runtime/src/sim/buggify.rs
```diff
@@ -0,0 +1,51 @@
+use crate::sim::Runtime;
```
cloutiertyler, line 1: NIT: You might consider having `sim` be a separate crate, because eventually you're going to want to mock out things like the disk/network/etc. which isn't really part of the runtime.

crates/runtime/src/sim/time/mod.rs
cloutiertyler, line 1: This is all very similar to what I have in my experimental project.

crates/runtime/src/lib.rs
Shubham8287: No strong reasons, but this felt more natural due to small set of variants. Using trait would have require to wrap `TokioHandle` in some newtype to be able to implement custom trait.

crates/runtime/src/sim_std.rs
cloutiertyler: TIL about eBPF. Interesting.

crates/runtime/src/sim_std.rs
cloutiertyler, line 1: This is all reasonable, but I do want to flag that we should work towards just not using these things inside the deep database at all.

cloutiertyler (approved):
This is quite good overall. I think it's a good start.

crates/runtime/src/sim_std.rs
Shubham8287: It can be used to do so many things, https://github.com/clockworklabs/SpacetimeDB/pull/5177/changes#diff-8c7a3b299146fed405fed8a131643bb6aaac19152857cb129826ef17457d1a89. This is some experimental low level code to detect `futex` call (under simulation, this means determinism leakage).

crates/runtime/src/sim_std.rs
Shubham8287: Yeah, that's how this crate is structured. everything inside `sim` module is `no_std`. `sim_std` is helper wrapper until dependent code also uses `std`.

crates/runtime/src/sim/executor/mod.rs
jsdt: Why would round-robin order be preferrable? That seems like it would limit the set race conditions that can be simulated significantly.

crates/runtime/src/sim/executor/mod.rs
```diff
@@ -0,0 +1,793 @@
...
+            self.run_all_ready();
+            if task.is_finished() {
+                let waker = Waker::noop();
+                return match Pin::new(&mut task).poll(&mut Context::from_waker(waker)) {
+                    Poll::Ready(output) => output,
+                    Poll::Pending => unreachable!("task.is_finished() was true"),
+                };
+            }
+
+            if self.time.wake_next_timer() {
```
jsdt, line 498: If we only advance timers when none of the other tasks are runnable, are we going to be unable to simulate some timing scenarios? We could randomly decide to wake a timer instead of running a task sometimes (in `run_all_ready`), though that might just cause a lot of time outs.

crates/runtime/DETERMINISM_COVERAGE.md
```diff
@@ -0,0 +1,50 @@
...
+This document tracks which sources of nondeterminism are under control in `spacetimedb-runtime`, which ones are only constrained by current architecture, and which ones still escape the simulator boundary.
+
+It is meant to serve two purposes:
+
+1. Make the current determinism boundary explicit for runtime code, core crates, and DST harnesses.
+2. Provide a place to record and review assumptions when a PR changes that boundary.
+
+## Status Definitions
+
+- `Controlled`
```
jsdt, line 12: I think we need one more status type to distinguish between a source of nondeterminism that we can make deterministic (which is what you are calling controlled), vs a source of non-determinism which we can deterministically simulate all possible outcomes.

Some examples that are deterministic with this framework, but can result in race conditions that we can't create are channels and atomic variables.

jsdt (approved):
This looks like a solid start

crates/runtime/src/sim/executor/mod.rs
Shubham8287: Oh, right! I think more subtle way to do this would be to occasionally advance time inside `run_all_ready` by few microseconds. This can be done by calling `self.time.advace()` api. Which will also take care to wake the registered sleepers, if needed.
This will advance the clock more naturally, and let timer tasks to be waken up along with other tasks?

`time.wake_next_timer()` thing could still be there when no tasks presents, to accelerate the time.

jsdt: I think `time.wake_next_timer()` does need to be there so avoid getting stuck (for cases where we are actually hitting a time out).

The idea of randomly/periodically advancing the clock makes sense to me. We could have a parameter like `polls_per_second` or `avg_time_per_poll` to control how quickly the clock advances as tasks are executed. With a `avg_time_per_poll: Duration` parameter, we could just advance time by that amount after every piece of work, or we could make it random with something like `avg_time_per_poll.mul_f64(-random::<f64>().ln())`.

Issue #4943, opened 2026-05-04
client reducer call args missmatch log spam


xDovos:
during some testing i found that the server logs each time a client sends a bad reducer call with a log like this

<img width="1185" height="487" alt="Image" src="https://github.com/user-attachments/assets/548d6b36-3b4a-4b84-a074-13a4d9b73291" />

this is one client calling it every 100ms with bad parameters.
1. one log would be enough per client
2. it should disconnect the client as the client is clearly broken or someone is trying to make a custom client and got the parameters wrong.

right now this is a DOS attack point as it fills the logs and as reducers aren't rate limited, it can spam the logs with a call every microsecond or faster. (1000+ bad requests per second but theoretically 10k+ bad requests per second).

PR #5017, opened 2026-05-13
More aggressively detect and disconnect ill-behaving or broken clients

### LINKED ISSUE #4945: server side client message args validation is wrong sometimes
during testing of #4943 i also found that the server doesn't check if the size of numbers matches the expected bit size. i had 2 u32 and 1 u64 and changed it on the client to 1 u32 and 2 u64 and besides it reading the wrong value, it didn't notice that it had 32 bits left over and just ran the reducer with the wrong values. i could even change one of them to a string and send "hello this is an invalid arg" as string and the server would just ignore that the arg size is way too big.
from my testing the error message only appears if the args length is too small.

i made this issue separate so that it is easier to track.

### LINKED ISSUE #5002: Disconnect clients who send ill-typed or unparseable function args
We generally rely on clients having a correct and up-to-date schema. A clear indication they do not have such a schema in the form of an ill-typed function call request means that we cannot correctly interoperate with that client. Even if some of the client's other messages may incidentally succeed processing, we should not assume them to be processed correctly. As such, the SpacetimeDB server should disconnect any client that sends an ill-typed function call along a WebSocket connection.

Note that our SDKs already behave according to this theory; they will error and disconnect when receiving row data they cannot parse from the server.

gefjon:
# Description of Changes

Contains two separate but related fixes:

## Reject reducer and procedure args which parse but leave behind trailing bytes

Fixes #4945 .

BSATN parsing generally accepts the case where there are unused trailing bytes in a buffer after parsing a type. This allows both building up compound-typed objects by repeatedly parsing their members, and packing multiple values into the same buffer sequentially.

However, it has an unfortunate consequence when parsing untrusted inputs: if a client submits an input e.g. for a reducer call which is not of the expected type, but has a prefix that parses at the expected type, a direct use of the BSATN parser will accept it and silently ignore the suffix. One simple example is a client attempting to submit an i64 when SpacetimeDB expects an i32, resulting in the high 4 bytes of the client submission being ignored, potentially resulting in a different number being parsed than the one submitted.

In this commit, we check when parsing user-submitted function arguments that not only did the parse succeed, but that it also consumed the entire input. If the entire input was not consumed, we treat it as an error in the class described above.

## Disconnect clients which issue invalid function calls

Fixes #5002 , #4943 . Prior to this change, clients who sent invalid or ill-typed function calls would receive error responses, but would not be disconnected. Clients which didn't handle those errors would continue sending invalid function calls continuously, consuming server time without performing useful work.

With this change, errors encountered when calling reducers or procedures are divided into three categories:

1. Those that send an error response and do not terminate the client, as previously.
2. Those that close the connection with the close code `Invalid`, meaning the client is broken and should not attempt to reconnect. Invalid and ill-typed function call arguments, as described above, are in this category.
3. Those that close the connection with the close code `Again`, meaning the client should attempt to reconnect. These are things that (may) indicate that a leader failover has occurred and the leader replica is now housed on a different SpacetimeDB node.

Notably, this PR does not include any changes to any of the client SDKs. We may recognize the `Again` close code and attempt to reconnect in the future, but do not do so currently.

As part of this change, I introduced a new mechanism by which the `core` crate can disconnect a `ClientConnectionSender`, and changed the existing disconnect when the client exceeds its queue limit to use that new mechanism.

I'm not sure "ABI break" is technically the correct label, but this is from a certain perspective a breaking change to the raw WebSocket API, in the sense that we now reject some requests that previously we would accept, and close some connections that we would previously have left open.

## For the user-facing changes

2: possible some user somewhere has been relying on this behavior and sending unused bytes in the arguments part of a `CallReducer` or `CallProcedure` message somehow. Most likely this would be a user of the raw WebSocket format, but it could also be someone with outdated generated bindings in a benign way.

## For the internal change to `ClientConnectionSender` and the WebSocket loop

3: this code is tricky. In particular, there's some concern that this change breaks ping/ponging, disconnecting clients which fill their queues, or some other aspect of the connection lifecycle.

# Testing

- [x] Added some unit tests.
- [x] Manually tested that clients which exceed their queue length are still disconnected in a timely fashion.
- [ ] Manually build a client with incorrect bindings for a reducer's arguments in the way described above and encounter the new error and disconnection.


gefjon:
This PR has merged pretty significantly from master. I'll close this, and we can take another pass at the same functionality some other time.

PR #5018, opened 2026-05-14
Do not `msync` the entire offset index file on every transaction

joshua-spacetime:
# Description of Changes

`msync` the modified pages instead of the entire file.

None

1

# Testing

Refactor

### DIFF
--- a/crates/commitlog/src/lib.rs
+++ b/crates/commitlog/src/lib.rs
@@ -109,7 +109,7 @@ impl Default for Options {
 impl Options {
     pub const DEFAULT_MAX_SEGMENT_SIZE: u64 = 1024 * 1024 * 1024;
     pub const DEFAULT_OFFSET_INDEX_INTERVAL_BYTES: NonZeroU64 = NonZeroU64::new(4096).expect("4096 > 0, qed");
-    pub const DEFAULT_OFFSET_INDEX_REQUIRE_SEGMENT_FSYNC: bool = false;
+    pub const DEFAULT_OFFSET_INDEX_REQUIRE_SEGMENT_FSYNC: bool = true;
     pub const DEFAULT_PREALLOCATE_SEGMENTS: bool = false;
     pub const DEFAULT_WRITE_BUFFER_SIZE: usize = 8 * 1024;
 
--- a/crates/commitlog/src/segment.rs
+++ b/crates/commitlog/src/segment.rs
@@ -356,9 +356,10 @@ impl OffsetIndexWriter {
             return Ok(());
         }
 
-        self.head
+        let entry_offset = self
+            .head
             .append(self.candidate_min_tx_offset, self.candidate_byte_offset)?;
-        self.head.async_flush()?;
+        self.head.async_flush_entry(entry_offset)?;
         self.reset();
 
         Ok(())
@@ -371,10 +372,6 @@ impl FileLike for OffsetIndexWriter {
         let _ = self.append_internal().map_err(|e| {
             warn!("failed to append to offset index: {e:?}");
         });
-        let _ = self
-            .head
-            .async_flush()
-            .map_err(|e| warn!("failed to flush offset index: {e:?}"));
         Ok(())
     }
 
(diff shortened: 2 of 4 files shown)


crates/commitlog/src/segment.rs
```diff
@@ -371,10 +372,6 @@ impl FileLike for OffsetIndexWriter {
         let _ = self.append_internal().map_err(|e| {
             warn!("failed to append to offset index: {e:?}");
         });
-        let _ = self
-            .head
-            .async_flush()
-            .map_err(|e| warn!("failed to flush offset index: {e:?}"));
```
joshua-spacetime, line 377: This flush isn't needed since `append_internal` already flushes.

crates/durability/src/imp/local.rs
```diff
@@ -282,9 +282,7 @@ where
             let ready_len = tx_buf.len();
             self.queue_depth.fetch_sub(ready_len as u64, Relaxed);
             tx_buf = spawn_blocking(move || -> io::Result<Vec<PreparedTx<Txdata<T>>>> {
-                for tx in tx_buf.drain(..) {
-                    clog.commit([tx.into_transaction()])?;
-                }
+                clog.commit(tx_buf.drain(..).map(|tx| tx.into_transaction()))?;
```
joshua-spacetime: This could include quite a lot of transactions, up to the size of the durability queue size. Not sure if this should be bounded further or if it matters at all.

kim (commented):
I've thought about it many times, but at this point I'm not on board moving away from the 1-tx-per-commit restriction.

A torn write in the middle of a commit is _guaranteed_ to destroy `commit.n` transactions. A smaller number of transactions per commit results in a higher number of commits per write, and increases the chance that at least _some_ transactions are recoverable.

The trouble is that we are rather prone to torn writes, not least because they (the writes) are unaligned. Just advising to use confirmed reads is not enough, I'd argue, because users have no way of even knowing how many transactions could potentially be lost -- outside of benchmarking scenarios, I at least would want to design my application such that it doesn't write too much ahead of the "uncertainty window" of the durability layer.

We can certainly do better by improving our I/O model and recovery mechanisms, but at this point I think we'd basically weaken durability guarantees, and I don't think this is a good idea.


The offset index changes look fine to me. I would suggest to increase `Options::offset_index_interval_bytes` if there is data that suggests that we're updating the index too often.

joshua-spacetime:
Understood @kim. Thank you for the detailed argument. I'll keep the offset index changes only.

crates/commitlog/src/lib.rs
```diff
@@ -109,7 +109,7 @@ impl Default for Options {
 impl Options {
     pub const DEFAULT_MAX_SEGMENT_SIZE: u64 = 1024 * 1024 * 1024;
     pub const DEFAULT_OFFSET_INDEX_INTERVAL_BYTES: NonZeroU64 = NonZeroU64::new(4096).expect("4096 > 0, qed");
-    pub const DEFAULT_OFFSET_INDEX_REQUIRE_SEGMENT_FSYNC: bool = false;
+    pub const DEFAULT_OFFSET_INDEX_REQUIRE_SEGMENT_FSYNC: bool = true;
```
joshua-spacetime, line 112: > I would suggest to increase `Options::offset_index_interval_bytes` if there is data that suggests that we're updating the index too often.

@kim do you think this is a reasonable default given the frequency at which we fsync?

kim: Hm interesting. Coupling this with the fsync interval, we'd write an index entry every 4-16KiB, which is the max amount of data that needs to be read from the commitlog in case the requested offset does not fall on an entry.

It does not matter much for local durability, as the index is only used when replaying from a snapshot offset. With replication, the tail of the log will be examined much more often.

I think 4-16KiB is probably acceptable, but I think we need to increase the batch size in private (it's smaller than `offset_index_interval_bytes`).

kim (approved):

PR #5023, opened 2026-05-14
C++ HTTP handlers - Module Bindings

### LINKED ISSUE #4880: HTTP handlers: C++ library
Extend the C++ module bindings library to allow defining HTTP handlers and registering them to routes. See #4636  for the Rust version. Also add documentation.

JasonAtClockwork:
# Description of Changes

Adding C++ HTTP handlers based on #4636

- adds the C++ handler/router API `SPACETIMEDB_HTTP_HANDLER()`, `SPACETIMEDB_HTTP_ROUTER()`
- wires C++ HTTP handlers into module definition/build/runtime registration
- mirrors the Rust/TypeScript HTTP smoketests and docs coverage
- update to the the HTTP handlers docs examples

New APIs for http handler and router behind `SPACETIMEDB_UNSTABLE_FEATURES` and should not be breaking.

3 - this touches the C++ binding surface, runtime/module registration, and the smoketest/docs paths, so there is more churn than I’d like, but it is still fairly contained to the C++ HTTP work.

# Testing

- [x] Expanded the C++ compile-surface tests
- [x] Added C++ unit coverage for HTTP conversion helpers
- [x] Expanded `crates/smoketests/tests/smoketests/http_routes.rs` with C++ mirrors of the Rust HTTP
route tests
- [x] Added a C++ docs-example smoketest for `docs/docs/00200-core-concepts/00200-functions/00600-HTTP-
handlers.md`

I also did some manual testing with a throw away project


crates/bindings-cpp/include/spacetimedb/http_client_impl.h
rekhoff: Are these 3 `LOG_INFO` lines still needed?
The `error_bytes` data doesn't seem like it' d be useful to the module author. 
The other two, if the are needed, should they be `LOG_WARN` or `LOG_ERROR`?

crates/bindings-cpp/include/spacetimedb/http_handler_macros.h
```diff
@@ -0,0 +1,61 @@
...
+    using ReturnType = typename traits::result_type;
+    static_assert(std::is_same_v<ContextType, HandlerContext>, "First parameter of HTTP handler must be HandlerContext");
+    static_assert(std::is_same_v<RequestType, HttpRequest>, "Second parameter of HTTP handler must be HttpRequest");
+    static_assert(std::is_same_v<ReturnType, HttpResponse>, "HTTP handlers must return HttpResponse");
+
+    std::function<HttpResponse(HandlerContext&, HttpRequest)> handler =
+        [func](HandlerContext& ctx, HttpRequest request) -> HttpResponse {
+            return func(ctx, std::move(request));
+        };
+    RegisterHttpHandlerHandler(handler_name, reinterpret_cast<const void*>(func), std::move(handler));
```
rekhoff: Casting between data pointers and function pointers is technically undefined behavior, though I don't think this will cause problems in practice on all current WASM targets. A comment might be good acknowledging this though, just so the intent is explicit.

crates/bindings-cpp/include/spacetimedb/http_client_impl.h
rekhoff:
```suggestion
    static const uint8_t empty_sentinel = 0;
    const uint8_t* body_ptr = body_bytes.empty() ? &empty_sentinel : body_bytes.data();
```
I believe casting the `""`, a `const char*`, to `const uint8_t*` via `reinterpret_cast` is technically undefined behavior because C++ doesn't allow accessing an object through a pointer to an incompatible type. Though this is a non-issue because it works in WASM, it's just not following C++ standards. Here is a change that should achieve the same goal.

rekhoff (approved):
I've tested these changes using some local throw-away test projects, and everything seems to work as intended, and function like the Rust-side changes do.
I have some minor nits, but they are optional items, not critical changes.
Marking my approval, looks good to me.

crates/bindings-cpp/include/spacetimedb/http_client_impl.h
JasonAtClockwork: Very no it was during testing while I was having troubles I've cleared these up.

crates/bindings-cpp/include/spacetimedb/http_handler_macros.h
JasonAtClockwork: Yeah this was a mistake it should have been a type to make this clear, I've updated it across the places necessary.

crates/bindings-cpp/include/spacetimedb/http_client_impl.h
JasonAtClockwork: This one was a miss on my part that I meant to clean up, I've updated this as well with a comment to make it clear.

crates/bindings-cpp/include/spacetimedb/http_handler_macros.h
rekhoff: Yup, perfect, that fix looks good to me

PR #5028, opened 2026-05-15
  chore: replace lazy_static with std LazyLock

onx2:
# Description of Changes

Replaces direct `lazy_static` usage with standard library primitives.

Most usages were replaced with `std::sync::LazyLock`. One simple static list of required JWT claims was replaced with a plain const array, and an unused lazy regex in the benchmark summarizer was removed.

This also removes direct `lazy_static` dependencies from the workspace crates that no longer use it.

Various other deps still use it but the general ecosystem vibe is to use `std` instead:
- `colored v2.2.0` -> https://github.com/colored-rs/colored/pull/231
- `prometheus v0.14.0` -> https://github.com/tikv/rust-prometheus/pull/556
- `sharded-slab v0.1.7` -> (https://github.com/hawkw/sharded-slab/pull/112, https://github.com/hawkw/sharded-slab/pull/111)
- `tracing-flame v0.2.0` -> transitive through `sharded-slab` which has PRs already.

This is primarily dependency hygiene rather than a performance optimization. Since this codebase targets a modern Rust version, direct `lazy_static` usage can be replaced with standard library primitives, reducing maintenance burden and one direct dependency.

None

1 - trivial, matching behavior in std

# Testing

Ran:

`cargo fmt --all --manifest-path SpacetimeDB/Cargo.toml`

`cargo check --manifest-path SpacetimeDB/Cargo.toml -p spacetimedb-bench --benches`

`cargo check --manifest-path SpacetimeDB/Cargo.toml -p spacetimedb-standalone`

Also verified there are no remaining direct `lazy_static` usages in Rust source files.


crates/core/src/auth/token_validation.rs
```diff
@@ -137,10 +137,8 @@ struct BasicTokenValidator {
     pub issuer: Option<Box<str>>,
 }
 
-lazy_static! {
-    // Eventually we will want to add more required claims.
-    static ref REQUIRED_CLAIMS: Vec<&'static str> = vec!["sub", "iss"];
-}
+// Eventually we will want to add more required claims.
+const REQUIRED_CLAIMS: [&str; 2] = ["sub", "iss"];
```
onx2, line 141: This seemed like it should have been `const`, can change to static though if I'm missing something.

crates/bench/src/bin/summarize.rs
```diff
@@ -102,10 +100,6 @@ mod criterion {
         crit_dir.join(format!("{name}.json"))
     }
 
-    lazy_static::lazy_static! {
-        static ref EMOJI: Regex = Regex::new(r"(on_disk|mem)").unwrap();
-    }
-
```
onx2, line 108: Unused 🤔

bfops:
Hi @onx2, we don't really have a preference for `LazyLock` over `lazy_static`, so I'm going to close this PR.

onx2:
> Hi @onx2, we don't really have a preference for `LazyLock` over `lazy_static`, so I'm going to close this PR.

Fair enough! I get wanting to avoid the code churn. I just wanted to offer the migration since `std::sync::LazyLock` became the official standard-library replacement back in Rust 1.80 to phase out `lazy_static`. Thanks for taking a look anyway!

PR #5029, opened 2026-05-15
Use owo-colors instead of colored

onx2:
# Description of Changes

Replaces direct usage of `colored` with `owo-colors` in the CLI and update crates.

`owo-colors` provides similar styling ergonomics while avoiding `colored`'s `lazy_static` dependency. It is also zero-allocation for common styling operations, so slightly faster, and has a smaller implementation footprint.

None.

1

This is a small dependency and import replacement for terminal styling in CLI-facing output.

# Testing

- [x] `cargo check -p spacetimedb-cli -p spacetimedb-update`
- [x] Check output in terminal respect color for a command
- [ ] Reviewer: verify work

image


crates/cli/src/subcommands/repl.rs
```diff
@@ -61,7 +61,7 @@ pub async fn exec(con: Connection) -> Result<(), anyhow::Error> {
     let api = ClientApi::new(con);
 
     loop {
-        let readline = rl.readline(&format!("🪐{}>", &database).green());
+        let readline = rl.readline(&format!("🪐{}>", &database).green().to_string());
```
onx2, line 64: Macros like format and println are ok with anything that impl Display but this readline needs a `&str` so we need to first call `to_string` to convert unlike before which was automatically alloc a string for us.

bfops:
See the comment here: https://github.com/clockworklabs/SpacetimeDB/pull/5028#issuecomment-4500591827

onx2:
> See the comment here: [#5028 (comment)](https://github.com/clockworklabs/SpacetimeDB/pull/5028#issuecomment-4500591827)

Understood, I appreciate the review. Just to clarify for future reference, is the policy generally to avoid swapping out working deps for std lib or more optimized alternatives (e.g., zero-allocation, fewer dependencies, etc) to avoid code churn? I respect that, just want to make sure I know what to focus on next time. Thanks!

PR #4471, opened 2026-02-26
Add Kotlin Multiplatform SDK (JVM + iOS)

AndroidPoet:
## What

Kotlin Multiplatform client SDK for SpacetimeDB, targeting JVM and iOS (arm64, simulator-arm64, x64). This brings first-class Kotlin/Java/Swift (via KMP interop) support to the SpacetimeDB ecosystem.

## Why

There's currently no way to build native Android/JVM or iOS clients without going through the C# or Rust SDKs. The Kotlin SDK fills this gap and opens up the entire JVM and Apple platform ecosystem — Android apps, backend services, iOS apps, and any Kotlin target.

## Architecture

- **Wire protocol**: Full `v2.bsatn.spacetimedb` support with BSATN binary serialization
- **Transport**: Ktor 3.0.3 WebSocket client (OkHttp engine on JVM, Darwin engine on iOS)
- **Concurrency**: kotlinx-coroutines with `Mutex`-serialized message handling, `atomicfu` for transport flags
- **Compression**: Configurable (None/Gzip/Brotli) via `withCompression()`. Gzip on both platforms via `expect`/`actual`, Brotli on JVM via `org.brotli:dec`
- **Cache**: Client-side row cache with ref-counted rows for overlapping subscription support

## Features

- Connect/disconnect with callbacks
- SQL subscriptions (`subscribe`, `unsubscribe`, `subscribeToAllTables`)
- Reducer invocation with result callbacks
- One-off queries (both `suspend` and callback variants)
- Automatic reconnection with configurable exponential backoff (`ReconnectPolicy`)
- Ping/pong keep-alive (30s idle timeout, matching the Rust SDK pattern)
- Configurable server message compression (None, Gzip, Brotli)
- Builder pattern for connection setup (`DbConnectionBuilder`)

## File layout

```
sdks/kotlin/
├── build.gradle.kts / settings.gradle.kts / gradle.properties
├── README.md / DEVELOP.md
├── src/
│   ├── commonMain/          # Shared code (all targets)
│   │   └── com/clockworklabs/spacetimedb/
│   │       ├── SpacetimeDBClient.kt      # DbConnection + builder + CompressionMode
│   │       ├── Identity.kt               # Identity, ConnectionId, Address, Timestamp
│   │       ├── ClientCache.kt            # Row cache + TableCache
│   │       ├── TableHandle.kt            # Per-table insert/delete/update callbacks
│   │       ├── SubscriptionHandle.kt     # Subscription lifecycle
[...]

PR #5033, opened 2026-05-15
Added client sdk for Kotlin Multiplatform (Jvm + iOS)

Jessewb786:
# Description of Changes

Building on the original work of androidPoet, this provides a Kotlin Multiplatform client sdk.
It also adds the kotlin code generator to the CLI.
This is should be enough to close issue Add Kotlin Multiplatform SDK support #4821.

This is a 5, but excepting the minor changes to the CLI, almost all changes are orthogonal to existing code, so there's very little chance of any form of regressions. 

# Testing

There is a suite of tests, all of which pass, and I'm using this implementation of what is intended to become a production ready Kotlin Multiplatform application.
I don't know if I've covered all corner cases, but it's so far covering everything I've needed from it.
That means, working on Spring backend services, Android and iOS.

# Side note
I would love that this be extended so that the spacetime module itself can be written in Kotlin.


bfops:
Hi @Jessewb786, thank you for opening this PR. At the moment we are refraining from expanding language support until we have done a bit more legwork. See Tyler's comment here: https://github.com/clockworklabs/SpacetimeDB/pull/4471#issuecomment-4247615903.

We will keep this as a reference implementation for once we have done that legwork!

PR #5051, opened 2026-05-16
Pipeline the websocket send path

joshua-spacetime:
# Description of Changes

In keeping with the "pipeline everything" ethos, I've replaced `recv` with `recv_many` on `ClientConnectionReceiver` so that a client connection's websocket send path works on a batch of messages at a time, if possible.

None

1.5

# Testing

Refactor

### DIFF
--- a/crates/core/src/client/client_connection.rs
+++ b/crates/core/src/client/client_connection.rs
@@ -19,6 +19,7 @@ use bytes::Bytes;
 use bytestring::ByteString;
 use derive_more::From;
 use futures::prelude::*;
+use log::warn;
 use prometheus::{Histogram, IntCounter, IntGauge};
 use spacetimedb_auth::identity::{ConnectionAuthCtx, SpacetimeIdentityClaims};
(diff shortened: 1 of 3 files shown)


crates/core/src/client/client_connection.rs
```diff
@@ -167,80 +170,115 @@ impl ClientConnectionReceiver {
...
+            .fold((0, 0), |(count, max_offset), tx_offset| {
+                (count + 1, max_offset.max(tx_offset))
+            })
+    }
+
+    /// Drain the pending [`ClientUpdate`]s, up to `max, into `buf`.
+    fn drain_pending(&mut self, buf: &mut Vec<OutboundMessage>, max: usize) -> usize {
+        let n = self.pending.len().min(max);
+        buf.reserve(n);
+        for update in self.pending.drain(..n) {
```
kim: nit: I think `extend()` might be slightly more efficient

kim (approved):
Nice!

PR #5056, opened 2026-05-16
CI - Close some gaps with GitHub discord notifications

bfops:
# Description of Changes

We currently have discord notifications when PRs merge into master. We have two problems:
1. These notifications don't run properly for external PRs (due to missing secrets)
2. We don't get notifications for other pushes to master

This PR fixes those issues by changing the job to run on any push to master rather than when a PR is closed.

None. CI only.

1

# Testing

I don't think we can test this without merging into master.


.github/workflows/discord-posts.yml
```diff
@@ -1,14 +1,64 @@
...
+            const prs = await github.paginate(github.rest.repos.listPullRequestsAssociatedWithCommit, {
+              owner,
+              repo,
+              commit_sha: sha,
+            });
+
+            const mergedPrs = prs
+              .filter((pr) => pr.merged_at && pr.base?.ref === 'master')
+              .sort((a, b) => new Date(b.merged_at) - new Date(a.merged_at));
+            const pr = mergedPrs[0];
```
bfops: we should error if there is not exactly one PR in this list. below we should also have an on-failure hook for this job that posts in discord a general "job failed" message (with a link to the job run) if this job itself fails

jdetter (approved):
Seems fine to me, let's try it 👍

PR #5061, opened 2026-05-17
Batch websocket responses using the v3 protocol

joshua-spacetime:
# Description of Changes

In keeping with the "pipeline everything" approach of SpacetimeDB, this patch serializes multiple client updates in a single websocket message using the v3 protocol.

None. `spacetime subscribe` was updated to use the v3 websocket api, but it falls back to v1 if protocol negotiation fails.

2

# Testing

This patch updates `spacetime subscribe` to use the v3 websocket protocol by default in order to get adequate coverage via the smoketests.

### DIFF
--- a/crates/client-api/src/routes/subscribe.rs
+++ b/crates/client-api/src/routes/subscribe.rs
@@ -1313,16 +1313,15 @@ struct V2OutboundMessage {
 
 /// Convert an outbound message into the binary websocket schema.
 ///
-/// Binary websocket connections should only receive v2 server messages. Error
-/// messages are still represented by the v1 type in the internal send path, so
-/// they are dropped here instead of being sent to a binary client.
+/// v2 connections should only receive v2 server messages.
+/// v1 messages are dropped.
+///
+/// TODO: For better type safety, [`ClientConnectionReceiver`] should be made
+/// generic over the protocol version.
 fn v2_outbound_message(message: OutboundWsMessage) -> Option<V2OutboundMessage> {
     let message = match message {
         OutboundWsMessage::Error(message) => {
-            log::error!(
-                "dropping v1 error message sent to a binary websocket client: {:?}",
-                message
-            );
+            log::error!("dropping v1 error message on v2 connection: {:?}", message);
             return None;
         }
         OutboundWsMessage::Message(message) => message,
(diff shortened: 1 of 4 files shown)


crates/core/src/worker_metrics/mod.rs
```diff
@@ -204,8 +204,8 @@ metrics_group!(
         pub tokio_mean_polls_per_park: GaugeVec,
 
         #[name = spacetime_websocket_sent_msg_size_bytes]
-        #[help = "The size of messages sent to connected sessions"]
-        #[labels(db: Identity, workload: WorkloadType)]
+        #[help = "The size of websocket payloads sent to connected sessions"]
+        #[labels(db: Identity)]
```
joshua-spacetime, line 208: It doesn't make sense to label these metrics by workload if we are sending multiple messages from potentially multiple workloads in a single payload. I'm calling this change out in case we're using this metric for billing.

bfops (approved):
My code-owned changes LGTM (in that they're mostly Spacetime-meat code that I don't have any strong opinions about).

I would like to have at least a manual test of the old-server-compatibility flow.

crates/core/src/worker_metrics/mod.rs
joshua-spacetime: We are **not** using these metrics for billing per @JulienLavocat 🙏

joshua-spacetime:
@gefjon feel free to ignore the `spacetime subscribe` changes if you'd like.

crates/client-api/src/routes/subscribe.rs
```diff
@@ -1290,105 +1290,409 @@ enum OutboundWsMessage {
...
+struct V2OutboundMessage {
+    message: ws_v2::ServerMessage,
+    num_rows: Option<usize>,
+}
+
+/// Convert an outbound message into the binary websocket schema.
+///
+/// Binary websocket connections should only receive v2 server messages. Error
+/// messages are still represented by the v1 type in the internal send path, so
+/// they are dropped here instead of being sent to a binary client.
```
gefjon: Is this something that should be changed in the future? Is this a TODO?

gefjon (approved):

gefjon:
I did, indeed, ignore the `spacetime subscribe` CLI changes.

crates/client-api/src/routes/subscribe.rs
joshua-spacetime: No this is saying that v2 connections should not receive v1 message types, but in the event that they do, we just drop them. This maintains the current behavior.

Note, this is just an artifact of `ClientConnectionReceiver` not being generic over the ws protocol version. But that's a refactor for another time.

joshua-spacetime: I've clarified this in the comment.

PR #5062, opened 2026-05-18
Use IndexMap for mounts and validate duplicates [Namespaces Part 2]

### LINKED ISSUE #5086: Namespaces

aasoni:
# Description of Changes

- Replace ModuleDef `mounts` type with IndexMap rather than Vec to explicitly require name uniqueness
- Validate mount name uniqueness

No

2 - Fairly simple change to new (as of yet unused) module def type

# Testing

- [ ] Added a test to check that the duplicate name check is working correctly


gefjon (approved):
Easy enough. Good stuff.

PR #5071, opened 2026-05-19
Store non-full pages in a `BTreeSet`, not a `Vec`

gefjon:
# Description of Changes

Reviving a previous patch I wrote during our (internal) TPCC experimentation. This has become important because, in addition to its performance implications, it makes row insertion locations deterministic regardless of datastore restarts, which previously they were not.

Previously, restarting the datastore would re-order the `non_full_pages` list (i.e. sort it by increasing `PageIndex`, where normally it was not sorted), meaning that which page a new row would be inserted into depended on when the datastore was last restarted.

With this patch, that is not the case: the `non_full_pages` are always kept in a deterministic order, so which page a new row goes into is also deterministic.

Original commit message follows:

And sort them by number of available var-len granules. This prevents an accidentally quadratic behavior where, for a table where the average row contains many var-len granules, after inserting a large number of rows, there would be a large number of pages in `non_full_pages` each of which had enough space for at least one fixed-len row part, but insufficient space for an actual row in practice due to insufficient var-len granules. Each insertion would then do a linear scan over `non_full_pages` before either inserting into the last page or allocating a new page which went to the end.

Now, non-full pages are stored in a `BTreeSet` sorted by the number of free var-len granules, and the search for a useable page is done with a `BTreeSet::range` iterator for only the pages with enough granules. I think there may still be an off-by-one-ish bug here, where a page may have enough bytes in the gap that it could either store the fixed-len part or the var-len granules, but not both, but this fix hopefully will suffice for now.

N/a

2? Table code is a bit fiddly, and this path is performance-sensitive when inserting rows.

# Testing

- [x] Passes table crate tests.
- [x] Was included in our internal TPCC experimentation, where it significantly improved performance (due to that benchmark exercising the accidentally-quadradic behavior this patch is designed to protect).
- [x] Joshua ran the keynote-2 benchmarks with this patch and did not observe a decrease in throughput.


crates/table/src/pages.rs
```diff
@@ -151,14 +167,13 @@ impl Pages {
...
+        if let Some((page_num_free_granules, page_idx)) = self
             .non_full_pages
-            .iter()
+            .range((num_var_len_granules, PageIndex(0))..)
             .copied()
-            .enumerate()
             .find(|(_, page_idx)| self[*page_idx].has_space_for_row(fixed_row_size, num_var_len_granules))
         {
-            self.non_full_pages.swap_remove(page_idx_idx);
+            self.non_full_pages.remove(&(page_num_free_granules, page_idx));
```
joshua-spacetime, line 210: Is this the only place where we remove entries from `non_full_pages`?

joshua-spacetime (commented):
I tested this on the keynote-2 benchmark, and I didn't observe a decrease in overall throughput fwiw.

crates/table/src/pages.rs
gefjon: It must be - the type of `non_full_pages.remove` has changed. It's also the only call to a method named `remove` in this file, and `non_full_pages` is private.

joshua-spacetime: I think this means that `non_full_pages` is only updated on insert and not delete.

gefjon: Ah, good catch: `Table::delete_internal_skip_pointer_map` goes to `Pages::delete_row`, which only updates `non_full_pages` if the page was previously full.

gefjon: I fixed this and added a test.

crates/table/src/pages.rs
```diff
@@ -247,23 +280,69 @@ impl Pages {
...
+    /// Record the number of available var-len granules in the page at `self[page_index]` into [`Self::non_full_pages`].
+    ///
+    /// Prior to calling this function, there must not be an entry for `page_index` in [`Self::non_full_pages`].
+    fn record_page_available_granules(&mut self, page_index: PageIndex) {
+        debug_assert!(!self.non_full_pages.iter().any(|(_, idx)| *idx == page_index));
+
+        let available_granules = self[page_index].available_var_len_granules();
+
+        if available_granules != 0 {
+            self.non_full_pages.insert((available_granules, page_index));
```
joshua-spacetime: What about deletes that don't include any `var_len_granules`? If I'm reading this correctly, it won't record the page as `non_full`, right?

crates/memory-usage/src/lib.rs
```diff
@@ -126,6 +126,13 @@ impl<K: MemoryUsage, V: MemoryUsage> MemoryUsage for std::collections::BTreeMap<
     }
 }
 
+impl<T: MemoryUsage> MemoryUsage for std::collections::BTreeSet<T> {
+    fn heap_usage(&self) -> usize {
+        // NB: this is best-effort, since we don't have a `capacity()` method on `BTreeMap`.
```
kim, line 131: Isn't this how B-trees work? Like, there isn't any excess, preallocated capacity 🤔

crates/table/src/pages.rs
```diff
@@ -78,7 +133,14 @@ impl Pages {
...
         }
         // Mark every page non-full.
-        self.non_full_pages = (0..self.pages.len()).map(|idx| PageIndex(idx as u64)).collect();
+        self.non_full_pages = (0..self.pages.len())
+            // We could probably compute the number of available granules once and use it for all pages,
+            // rather than calling the method on each page,
+            // but we'd have to do some amount of reasoning to demonstrate it was correct
+            // based on the definition of `Page::clear`,
+            // and why bother?
+            .map(|idx| (self.pages[idx].available_var_len_granules(), PageIndex(idx as u64)))
```
kim, line 143: For my own education: why is it that the number of granules isn't known statically at this point (we just cleared all pages)? Does the size of a granule depend on the table schema or something?

kim (approved):
I am unsure about Josh's remaining comment, but otherwise this looks good to me.

crates/memory-usage/src/lib.rs
gefjon: IIRC, individual nodes in a B-tree (at least in the Rust impl) are vec-like, with a fixed capacity but a variable length, so that you don't have to realloc on every mutation.

crates/table/src/pages.rs
gefjon: My recollection is that we always reserve space for at least one fixed-len part, and the length of that fixed-len part depends on the row layout, so the number of var-len granules available in an empty table is not statically known at Rust compile time.

crates/table/src/pages.rs
gefjon: Oh, yeah. I suppose checking for `Page::is_full` was load-bearing, and I'll have to restore that behavior.

gefjon: Ok, it's back. Now we check `Page::is_full` instead of `available_granules == 0` here and elsewhere. I also added a test, `prefer_earlier_non_full_page`, which allocates a fully fixed-len row into a page with 0 var-len granules available.

crates/table/src/pages.rs
```diff
@@ -135,30 +185,29 @@ impl Pages {
...
+        self.record_page_non_full(page_index, fixed_row_size);
         Ok((page_index, res))
     }
 
     /// Find a page with sufficient available space to store a row of size `fixed_row_size`
     /// containing `num_var_len_granules` granules of var-len data.
     ///
     /// Retrieving a page in this way will remove it from the non-full set.
-    /// After performing an insertion, the caller should use [`Pages::maybe_mark_page_non_full`]
+    /// After performing an insertion, the caller should use [`Pages::record_page_available_granules`]
```
joshua-spacetime:
```suggestion
    /// After performing an insertion, the caller should use [`Pages::record_page_non_full`]
```

joshua-spacetime (approved):

PR #5074, opened 2026-05-19
Add commitlog knobs to server config

joshua-spacetime:
# Description of Changes

Adds the commitlog knobs `max_segment_size`, `write_buffer_size`, and `preallocate_segments` to the server's `config.toml`. There are other seemingly more advanced knobs that I did not add at this time.

This patch also increases the `DEFAULT_WRITE_BUFFER_SIZE` from `8KiB` to `128KiB` to optimize high throughput workloads like the keynote-2 benchmark.

None

1

# Testing

Manual


kim (commented):
The change is fine by me, but I'll note that commitlog config is exposed in the other edition, too, under a different key. It is probably not a good idea to let the `config.toml` diverge unless we have a good reason.

Also note that there is `docs/docs/00300-resources/00200-reference/00100-cli-reference/00200-standalone-config.md`, which I think should reflect these settings, even if they're deemed unstable.

kim (approved):

PR #5075, opened 2026-05-19
Allow to use `Uuid` as literal value in ts Query Builder

### LINKED ISSUE #5073: Filtering after a uuid field on a react client doesn't work
I added a uuid field to the person column from the react template project.
Then I tried to filter after the id field.

module
```ts
const spacetimedb = schema({
  person: table(
    { public: true },
    {
      id: t.uuid(),
      name: t.string(),
    },
  ),
});
```
client
```ts
 const sampleId = Uuid.parse("019e4174-c1bf-7000-8000-000062f0678a");
 const [people] = useTable(tables.person.where((row) => row.id.eq(sampleId)));
```

This resulted in the following type error

```txt
No overload matches this call. Overload 1 of 2,
(literal: LiteralValue & Uuid): BooleanExpr<TableToSchema<"person", TableSchema<CoerceRow<{ id: UuidBuilder; name: StringBuilder; }>, []>>>
, gave the following error. Argument of type Uuid is not assignable to parameter of type
LiteralValue & Uuid
. Type Uuid is not assignable to type
Timestamp & Uuid
. Type Uuid is missing the following properties from type Timestamp: timestamp_micros_since_unix_epoch, microsSinceUnixEpoch, toMillis, toDate, and 2 more. Overload 2 of 2,
(value: never): BooleanExpr<any>
, gave the following error. Argument of type Uuid is not assignable to parameter of type never. (ts 2769)
```

as well as the following runtime error

```txt
Uncaught TypeError: BigInt value can't be serialized in JSON
    literalValueToSql query.ts:871
    valueExprToSql query.ts:844
    booleanExprToSql query.ts:809
    renderSelectSqlWithJoins query.ts:367
    toSql query.ts:211
    toSql query.ts:58
    useTable useTable.ts:88
    App App.tsx:15
    React 11
    workLoop scheduler.development.js:266
    flushWork scheduler.development.js:239
    performWorkUntilDeadline scheduler.development.js:533
```

kistz:
# Description of Changes

Was forgotten while migrating to the QueryBuilder i suppose.
Closes #5073

Maybe @clockwork-tien could lend a hand again in reviewing? ^^

None

1

# Testing
Works in my project :>

- [x] Type Error goes away when using .eq(uuid)


crates/bindings-typescript/src/lib/query.ts
ClemensWon, line 867: here also needs to be a a special case for the uuid type, when it runs into the default case, the JSON serializer throws an error, since it can't handle the bigint.

I tried adding
```ts

  if (value instanceof Uuid) {
    return value.toString();
  }
```

which solved the serialization issue, but then didn't return anything for my subscription.

kistz: This could be a separate issue (that subscriptions  dont return anything) because i am seeing the same thing happen to me in thr vue bindings (and not tied to Uuids)
Will add what you suggested and then i will investigate the where clause issue separately :)
Thanks for checking!

kistz:
@joshua-spacetime do you think you can give this a quick review sometime? :>

crates/bindings-typescript/src/lib/query.ts
```diff
@@ -858,6 +860,9 @@ function literalValueToSql(value: unknown): string {
   if (value instanceof Timestamp) {
     return `'${value.toISOString()}'`;
   }
+  if (value instanceof Uuid) {
+    return value.toString();
```
joshua-spacetime: In C#, UUID literals are formatted as hex values
```
0x00000000000000000000000000000000
```

kistz: is this a concern?
because as long as the where clause works server side it should be fine no?

i can change this ofc if desired :>

joshua-spacetime: I'd prefer if they were consistent across the different languages personally. But also notably, I'm not sure the string version works as currently written without wrapping the value in quotes `''`.

joshua-spacetime (approved):

PR #5077, opened 2026-05-19
Client binaries from DigitalOcean -> AWS

bfops:
# Description of Changes

Use an AWS bucket for client binaries instead of DigitalOcean.

Note that this will effectively hamstring the DigitalOcean mirror for any old clients (i.e. ones before this change is released). But it's only a mirror, and they can "fix" it by upgrading to a version with this change.

None

1

# Testing

- [x] Manually ran the package job on this PR and confirmed that some of the resulting URLs are indeed downloadable from the AWS urls

### DIFF
--- a/.github/workflows/attach-artifacts.yml
+++ b/.github/workflows/attach-artifacts.yml
@@ -60,3 +60,4 @@ jobs:
           gh release upload "$RELEASE_TAG" ./* \
             --repo "$GITHUB_REPOSITORY" \
             --clobber
+
--- a/.github/workflows/attach-artifacts.yml
+++ b/.github/workflows/attach-artifacts.yml
@@ -20,7 +20,7 @@ jobs:
       - name: Download artifacts from AWS base URL
         env:
           RELEASE_TAG: ${{ github.event.inputs.release_tag }}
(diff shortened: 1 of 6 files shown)


.github/workflows/attach-artifacts.yml
```diff
@@ -17,10 +17,10 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
-      - name: Download artifacts from private base URL
+      - name: Download artifacts from AWS base URL
```
bfops, line 20:
```suggestion
      - name: Download artifacts
```

.github/workflows/benchmarks.yml
```diff
@@ -24,8 +23,10 @@ jobs:
   benchmark:
     name: run criterion benchmarks
     runs-on: benchmarks-runner
+    # disable until we fix the benchmarks
+    if: false
```
bfops, line 27: I disabled these jobs. They used the same GH secret as the upload-client-binaries job, which I think we will want to change and give each job its own narrowly-permissioned secret.

However, these jobs are broken anyway, so I figured just disable them for now and we can fix them up when we want them fixed up.

cloutiertyler (approved):
This will only break for old clients if GitHub is down, right?

PR #5078, opened 2026-05-19
Add required ci check for `keynote-2` benchmark

joshua-spacetime:
# Description of Changes

Adds a new required ci check for keynote-2 benchmark regressions. The test runs for 60s and fails if throughput < 300K TPS.

Note, this check will be flaky as long as it's running concurrently with other CI jobs. It may need a dedicated runner/host machine. Although it may be sufficient to only schedule one runner/VM to a single host machine at a time. I'll need to sync with @jdetter to determine the best way forward here.

UPDATE: We're using a dedicated runner. See the **Testing** section.

N/A

2

Mainly copy-paste from the other CI workflows.

# Testing

This job now uses `spacetimedb-benchmark-runner` which is entirely dedicated to this one CI job. I've tested this at different times of the day when the CI runners are under load and not. The performance is consistent and the test isn't flaky. It has passed every time.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -302,6 +302,9 @@ jobs:
           save-if: false
           prefix-key: v1
 
+      - name: Build keynote benchmark binaries
+        run: cargo build --release -p spacetimedb-cli -p spacetimedb-standalone
+
       # Node 24 is the current Active LTS line.
       - name: Set up Node.js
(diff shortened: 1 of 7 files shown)


bfops (approved):
This LGTM in terms of CI. I can't speak to the correctness of how the benchmark is being run, but it seems to be passing.

PR #5084, opened 2026-05-20
Undo `.npmrc` in templates

### LINKED ISSUE #5083: undo .npmrc in templates
We added `.npmrc` in templates in https://github.com/clockworklabs/SpacetimeDB/pull/5032, but this has the side effect of adding it to every user project. This is not something we should do, as it means that users will not be able to pull the latest versions of SpacetimeDB, which will likely cause confusion.

bfops:
# Description of Changes

See https://github.com/clockworklabs/SpacetimeDB/issues/5083 for motivation.

None

2

# Testing

- [x] `cargo run -pspcaetimedb-cli -- init --template nodejs-ts` no longer creates a directory with `.npmrc`


cloutiertyler (approved):
Seems reasonable to me.

PR #5095, opened 2026-05-21
Use synchronous runtime for the main wasm execution lane

joshua-spacetime:
# Description of Changes

Before this change, we used a single async-enabled wasm runtime for all requests, even though procedures are the only operation that can yield. Now each module gets two separate runtimes. We continue to use the same async runtime for procedures, but now reducers are executed against a synchronous wasm runtime, backed by a single OS-thread instead of a Tokio runtime.

The purpose of this change is to remove from the critical path the overhead associated with async calls that really aren't async at all.

Also includes the following fix from #5135:

> After #4973, WASM procedures can execute concurrently with later operations on the same WebSocket.

> Before this change, the C# regression testsuite queued several procedures, then immediately queued `UnsubscribeThen`. After #4973, the unsubscribe could be applied before the `SubscriptionEventOffset` procedure callback ran, clearing `MyTable` from the local subscribed cache. The callback then failed while asserting that the `offset-test:` row was present.

> This change treats unsubscribe as a separate phase. It is scheduled after the main work is queued, but only starts once `waiting == 0`, so all callbacks that inspect subscribed state run before the cache is cleared.

None

2.5

# Testing

Pure refactor. Relies on current test coverage. #5078 will ensure the performance is on par with V8.

### DIFF
--- a/crates/core/src/host/wasmtime/mod.rs
+++ b/crates/core/src/host/wasmtime/mod.rs
@@ -11,7 +11,7 @@ use spacetimedb_paths::server::ServerDataDir;
 use std::borrow::Cow;
 use std::time::Duration;
 use wasmtime::{self, Engine, Linker, StoreContext, StoreContextMut};
-pub use wasmtime_module::{WasmtimeInstance, WasmtimeModule};
+pub use wasmtime_module::{WasmtimeAsyncModule, WasmtimeInstance, WasmtimeModule};
 
 #[cfg(unix)]
 mod pooling_stack_creator;
@@ -19,8 +19,10 @@ mod wasm_instance_env;
 mod wasmtime_module;
 
 pub struct WasmtimeRuntime {
-    engine: Engine,
-    linker: Box<Linker<WasmInstanceEnv>>,
+    sync_engine: Engine,
+    sync_linker: Box<Linker<WasmInstanceEnv>>,
+    async_engine: Engine,
+    async_linker: Box<Linker<WasmInstanceEnv>>,
     config: WasmConfig,
 }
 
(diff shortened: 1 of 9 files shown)


crates/core/src/host/wasmtime/mod.rs
```diff
@@ -46,76 +48,106 @@ pub(crate) fn epoch_ticker(mut on_tick: impl 'static + Send + FnMut() -> Option<
...
+    format!("wasm-main-{suffix}")
+}
+
+fn wasm_procedure_executor_thread_name(database_identity: &spacetimedb_lib::Identity) -> String {
     let hex = database_identity.to_hex();
     // We use the tail of the identity to avoid the common structured prefix.
     let suffix = &hex.as_str()[hex.as_str().len() - THREAD_NAME_DATABASE_ID_SUFFIX_LEN..];
-    format!("wasm-{suffix}")
+    format!("wasm-proc-{suffix}")
 }
```
coolreader18, line 153: Max length of a linux thread name is 15, so this will only show `wasm-proc-XXXXX`, even though `suffix` is `XXXXXXXX`

crates/core/src/host/wasmtime/wasmtime_module.rs
```diff
@@ -570,6 +710,14 @@ impl module_host_actor::WasmInstance for WasmtimeInstance {
         budget: FunctionBudget,
     ) -> (module_host_actor::ProcedureExecuteResult, Option<TransactionOffset>) {
         let store = &mut self.store;
+        if !self.supports_async {
+            let res = module_host_actor::ProcedureExecuteResult {
+                stats: zero_execution_stats(store),
+                call_result: Err(anyhow::anyhow!("procedures require an async Wasmtime instance")),
+            };
+            return (res, None);
+        }
```
coolreader18: Might it make sense for this to be an `assert!()`? I think this would be a programmer error, not something a user could run into.

coolreader18 (approved):

crates/core/src/util/jobs.rs
```diff
@@ -405,6 +421,111 @@ impl SingleCoreExecutor {
     }
 }
 
+/// A handle to a plain OS-thread executor for synchronous database work.
+///
+/// Unlike [`SingleCoreExecutor`], it is intended for synchronous runtimes.
+/// This executor never enters Tokio and never polls futures on its worker
```
Shubham8287, line 427: I think we should always be running `SingleThreadedExecutor` and `SingleCoreExecutor` on same`CorePinner, esle it could lead to avoid unneccessary lock contention which may lead to perfomance regression. If this is true, should we put it in comment?

Shubham8287 (approved):

crates/core/src/util/jobs.rs
joshua-spacetime: This is true, assuming core pinning is enabled. I added a comment in both places where we pass the `CorePinner`. Note however, this is not necessarily true by default, because we don't core pin by default.

PR #5110, opened 2026-05-22
Add batching Test Suite tests to avoid an OOM SIGKILL

rekhoff:
DO NOT COMMIT

This is to do a CI job test where we batch the tests, in order to prevent a SIGKILL that appears to be OOM related.


rekhoff:
No longer needed. OOM issue has been resolved, and parent branch has fixed to address issues already.

PR #5113, opened 2026-05-25
Move `RelationalDB` to `spacetimedb-engine` crate.

Shubham8287:
# Description of Changes

 Moves `RelationalDB` and related database code into a new `spacetimedb-engine` crate.
The main motivation is to tighten dependency control around the engine layer and isolate `RelationalDB`
 behind a crate boundary.
  - Majority of this PR is code-motion.
  - Removes direct production dependence on `tokio` from `spacetimedb-engine`.
  - Keeps `tokio` only as a dev-dependency for test-only code in `spacetimedb-engine`.
  - This is intended to be a structural refactor only and should not result in any functional change in
  production.
  - Adds a CI check to ensure `spacetimedb-engine` continues to compile in simulation mode

NA

1.5.

# Testing
Existing tests should be enough.


cloutiertyler:
Why is "Expected complexity level and risk" unspecified? Could you please update. I assume you're saying it's low risk because it's just mechanical?

Shubham8287:
> Why is "Expected complexity level and risk" unspecified? Could you please update. I assume you're saying it's low risk because it's just mechanical?.


No, I would have come to back it, when asked someone to review it. It is draft rn. (I should have marked).

crates/durability/Cargo.toml
```diff
@@ -8,6 +8,9 @@ license-file = "LICENSE"
 description = "Traits and single-node implementation of durability for SpacetimeDB."
 
 [features]
+default = ["tokio"]
```
Shubham8287, line 11: All the crates, which are directly or in-directly depend on "runtime" crate, now have to do this "feature forwarding".

To be able to compiled with simulated runtime when "simulated" feature is on.

Not a fan of doing it this way, but not sure about better ways.

crates/engine/src/relational_db.rs
```diff
@@ -1826,19 +1830,21 @@ fn default_row_count_fn(db: Identity) -> RowCountFn {
 
 #[cfg(any(test, feature = "test"))]
```
joshua-spacetime, line 1812: The `core` crate needs to have `spacetimedb-engine` as a dev-dependency with the `test` feature enabled, for `cargo test -p spacetimedb-core` to work.

Shubham8287: Right, It got mask in CI.

joshua-spacetime (approved):

cloutiertyler (approved):

PR #5115, opened 2026-05-26
Fix TypeScript Random.fill typed array handling

### LINKED ISSUE #5091: Random::fill not implemented in runtime (typescript)
the `fill` method is defined here,

```ts
// src/server/rng.ts
export interface Random {
// ...
  /**
   * Like `crypto.getRandomValues()`. Fills a `TypedArray` with random integers
   * in a uniform distribution, mutating and returning it.
   */
  fill<T extends IntArray>(array: T): T;
// ...
}
```

but at runtime, we see `Uncaught TypeError: ctx.random.fill is not a function`

Shiven0504:
## Summary
- Make TypeScript Random.fill detect BigInt typed arrays by array type instead of array.at(0), so empty typed arrays are still handled correctly.
- Use Math.pow for numeric typed-array bounds so 32-bit arrays use the full 32-bit range instead of JS bit-shift wrapping.
- Add focused coverage for Random.fill availability, all integer typed arrays, and empty arrays.

Fixes #5091.

## Tests
- corepack pnpm --filter spacetimedb exec vitest run tests/rng.test.ts
- corepack pnpm --filter spacetimedb build:types
- corepack pnpm --filter spacetimedb exec prettier src/server/rng.ts tests/rng.test.ts --check --ignore-path ../../.prettierignore


JasonAtClockwork (approved):
LGTM, it fixes two bugs for Random.fill

PR #5116, opened 2026-05-26
commitlog: Truncate segment before resuming

kim:
`read_exact` doesn't distinguish between EOF and not enough bytes to fill the given buffer. We may thus consider a segment valid to be resumed, but leave in trailing bytes (less than the commit header length). That can cause silent data loss, because appending more data will render the segment corrupt: a restart will then start a new segment, leaving anything after the trailing bytes unreachable.

To fix this, truncate the segment to the size determined by `Metadata::extract` before resuming writes.

2

# Testing

Added a test


Shubham8287:
Don't know if it's a github bug or what but It is looking like a empty commit.

crates/commitlog/src/repo/mod.rs
```diff
@@ -384,7 +384,15 @@ pub fn resume_segment_writer<R: Repo>(
             max_commit: _,
         } = meta;
         let mut writer = repo.open_segment_writer(offset)?;
+        // Ensure that the segment's size is exactly what we determined.
+        //
+        // When `Metadata` encounters EOF, it could be that there actually are
+        // trailing bytes in the segment, but less than the commit header
+        // length. This is difficult to detect due to the use of `read_exact`,
+        // so ensure we remove any trailing bytes.
+        writer.ftruncate(tx_range.end, size_in_bytes)?;
```
Shubham8287: I would have preferred to treat this case similar to corrupted commits i.e to return `ResumedSegment::Corrupted` here and let caller create new segment. . Which looks more consistent and safer behavior.

Maybe, we can check in `Metadata::with_header` , if reader is at EOF. wdyt?

kim: We'd need to look at all ways in which `read_exact` is used. Not sure if that gains us that much.

Shubham8287 (approved):

crates/commitlog/src/repo/mod.rs
kim: I added an assert that we're not truncating more than `commit::Header::LEN - 1` trailing bytes.

PR #5122, opened 2026-05-26
V8 heap metrics: Track instance memory usage for procedure workers too

gefjon:
# Description of Changes

Prior to this PR, the `V8HeapMetrics` were tracked only for the "main" instance of a database, i.e. the reducer worker. This meant that we had little to no visibility into memory usage by procedures.

In this PR, we also track values for the procedure workers. We considered tracking each instance's usage separately with a unique integer `instance_id` label, but were concerned about cardinality (see discussion), so decided instead to track only two sets of label values per database: `JsWorkerKind::Main` and `JsWorkerKind::Procedure`. The entries for `JsWorkerKind::Procedure` store the sum of the values for all procedure workers for that database.

I also moved the logic for calling `remove_label_values` into an associated function on `V8HeapMetrics`, rather than listing them all in `remove_database_gauges`. This hides the fact that we have label values for both `JsWorkerKind` variants.

We don't use any of these metrics for billing, and otherwise do not consider our metrics a stable API.

2: it would be unfortunate if we reported incorrect values for these metrics, though (as mentioned above) they are not used for billing, only diagnostics.

# Testing

I do not know how to test metrics.


crates/core/src/host/v8/mod.rs
```diff
@@ -955,118 +975,107 @@ struct V8HeapMetrics {
     external_memory_bytes: IntGauge,
     native_contexts: IntGauge,
     detached_contexts: IntGauge,
-    last_observed: V8HeapSnapshot,
```
joshua-spacetime, line 958: Something to bear in mind, right after a module update it's possible for both the old and the new module to be alive at the same time. The old module could still be processing inflight requests while the new module begins accepting new requests, which means one module can still overwrite labels from the other.

gefjon: Whuh? That sounds like a bug. Don't we rely on the uniqueness of the main worker for ordering? If the old module remains and continues processing outstanding requests while the new module begins accepting new requests, don't we interleave them?

Anyway, from a metrics perspective, it sounds like the solution is just to not special-case the main instance as having `instance_id: 0`, and instead to draw main instances' IDs from the `instance_id_source`. Also to either lift the `instance_id_source` to live outside of the `JsModule`, or make it global.

gefjon: Probably just making the `instance_id_source` into a global counter is easiest.

joshua-spacetime: > Also to either lift the `instance_id_source` to live outside of the `JsModule`, or make it global.

This would introduce unbounded cardinality which I don't think we want to do.

gefjon: WDYM by that? Are integer Prometheus label values stored in an array, such that clustering them around 0 is beneficial? I was under the impression that label values were backed by a hash map, so that the distribution of label values didn't matter, only the count.

joshua-spacetime: My concern is the cardinality within prometheus. If we make `instance_id_source` global, each new module instance gets a new ID, which corresponds to a unique series in prometheus, which can't handle high cardinality all that well.

gefjon: Out-of-band discussion here revealed that, while `remove_label_values` clears the rows locally, they remain in the remote Prometheus database forever, as far as we can tell. As such, I'll revise this PR to eliminate `instance_id`, and instead to sum all procedure instances' usage into a single row, and to restore incremental updating.

gefjon:
Updated PR description and title: I'll make the changes for Wasmtime modules in a separate PR.

joshua-spacetime (approved):

PR #5131, opened 2026-05-27
Improve accuracy and change semantics of metric `wasm_memory_bytes`

gefjon:
# Description of Changes

Prior to this commit, the metric `wasm_memory_bytes` had several problems:

1. Despite its name, it was used for both Wasmtime and V8 modules. For V8 modules, it was the same value as `v8_used_heap_size_bytes`.
2. It stored only the value for a single instance at any given time, so it under-reported a database's memory usage.
3. The same row (set of label values) was written concurrently by all instances of a particular database, with each one clobbering the previously written value.

In this commit, we change the metric so that:

1. It is recorded only for Wasmtime instances, not V8 instances. For V8 instances, instead directly check `v8_used_heap_size_bytes`, or one of the other V8 heap metrics. This change involved moving the recording of this metric from `module_host_actor.rs` to `wasmtime/wasm_instance_env.rs`
2. Similar to the V8 heap metrics, all the instances cooperatively share the metric entry, updating it by incrementing and decrementing rather than `set`ting.

Note that this metric is used for billing, and so we will need to update our billing code (elsewhere) to account for the change. In particular, our billing code should now charge for the sum of `wasm_memory_bytes` and `v8_used_heap_size_bytes`. We also should expect with this change for each database's recorded usage to increase, as we are now accurately recording the usage for all instances, not just one.

Billing metric semantics changed.

3: billing metric semantics changed.

# Testing

I do not know how to test metrics.


crates/core/src/host/wasmtime/wasm_instance_env.rs
```diff
@@ -133,6 +135,53 @@ pub(super) struct WasmInstanceEnv {
...
+    }
+
+    fn observe(&mut self, memory_usage: usize) {
+        let delta = memory_usage as i64 - self.last_observed;
+
+        if delta > 0 {
+            self.wasm_memory_bytes.add(delta);
+        } else {
+            self.wasm_memory_bytes.sub(-delta);
+        }
```
joshua-spacetime: Need to update `last_observed`

crates/core/src/host/v8/mod.rs
```diff
@@ -1117,7 +1104,6 @@ fn sample_heap_stats(scope: &mut PinScope<'_, '_>, metrics: &mut V8HeapMetrics)
     // Whenever we sample heap statistics, we cache them on the isolate so that
     // the per-call execution stats can avoid querying them on each invocation.
```
joshua-spacetime:
```suggestion
```
Since we're no longer caching heap stats.

gefjon: I'm also renaming the function from `sample_heap_stats` to `record_heap_stats`.

crates/core/src/host/wasmtime/wasm_instance_env.rs
gefjon: :facepalm: I sure do need to do that.

joshua-spacetime (approved):

PR #5133, opened 2026-05-27
fix `unity-testsuite`: clear intermediate Godot build state

joshua-spacetime:
# Description of Changes

Harden `cargo regen csharp dlls` against flaky Godot restore/build state in the Unity test suite.

I suspect this flake is runner specific. Some Unity runners appear to have incomplete Godot package state, where `sdks/csharp/packages/godotsharp` exists but is missing `GodotSharp.dll`. In that state, NuGet can still report restore success and generate a `project.assets.json` that contains `GodotSharp` metadata but no usable compile asset.

The regen flow now removes the Godot MSBuild intermediate directory (`sdks/csharp/obj~/godot`) before restoring `SpacetimeDB.ClientSDK.Godot.csproj`. It also clears the targeted `GodotSharp` package cache (`sdks/csharp/packages/godotsharp`) so NuGet cannot reuse a partially restored package missing `GodotSharp.dll`.

After restore, `regen` parses `project.assets.json` and verifies that `GodotSharp` is present and that the referenced `GodotSharp.dll` exists before running `dotnet pack --no-restore`.

This should either prevent the flaky missing-`Godot` namespace failure, or fail earlier with a clearer restore/package-state error.

NOTE/WARNING: This code was entirely AI generated.

N/A

~

# Testing

- [x] `unity-testsuite` passes (I ran it at least 5 times to try to get it scheduled to different runners)

### DIFF
--- a/tools/regen/Cargo.toml
+++ b/tools/regen/Cargo.toml
@@ -7,4 +7,5 @@ edition.workspace = true
 anyhow.workspace = true
 clap.workspace = true
 duct.workspace = true
+serde_json.workspace = true
 tempfile.workspace = true
--- a/tools/regen/Cargo.toml
+++ b/tools/regen/Cargo.toml
@@ -7,5 +7,4 @@ edition.workspace = true
 anyhow.workspace = true
 clap.workspace = true
 duct.workspace = true
-serde_json.workspace = true
 tempfile.workspace = true
--- a/tools/regen/src/csharp.rs
+++ b/tools/regen/src/csharp.rs
@@ -1,12 +1,15 @@
 #![allow(clippy::disallowed_macros)]
 
-use anyhow::{bail, Result};
+use anyhow::{bail, Context, Result};
 use duct::cmd;
 use std::env;
 use std::fs;
 use std::path::{Path, PathBuf};
 
 const BSATN_PACKAGE_ID: &str = "spacetimedb.bsatn.runtime";
+const GODOTSHARP_ASSET_ID: &str = "GodotSharp";
+const GODOTSHARP_DLL_FILE_NAME: &str = "GodotSharp.dll";
+const GODOTSHARP_PACKAGE_ID: &str = "godotsharp";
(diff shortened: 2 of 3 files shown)


tools/regen/src/csharp.rs
```diff
@@ -116,6 +120,73 @@ fn clear_restored_package_dirs(pkg_id: &str) -> Result<()> {
...
+
+fn restored_lib_package_file<'a>(package: &'a Value, file_name: &str) -> Option<&'a str> {
+    package.get("files")?.as_array()?.iter().find_map(|file| {
+        let path = file.as_str()?;
+        let actual_file_name = path.rsplit('/').next()?;
+        (path.starts_with("lib/") && actual_file_name == file_name).then_some(path)
+    })
+}
+
+fn verify_godotsharp_restore() -> Result<()> {
```
bfops: this function seems unnecessary to me. it's a lot of code just to change one failure into another. I would personally remove this.

tools/regen/Cargo.toml
```diff
@@ -7,4 +7,5 @@ edition.workspace = true
 anyhow.workspace = true
 clap.workspace = true
 duct.workspace = true
+serde_json.workspace = true
```
bfops: if we remove the `verify_*` function, I would revert this as well

tools/regen/src/csharp.rs
```diff
@@ -207,7 +278,9 @@ pub fn regen_dlls() -> Result<()> {
     )?;
 
     clear_restored_package_dirs(BSATN_PACKAGE_ID)?;
+    clear_restored_package_dirs(GODOTSHARP_PACKAGE_ID)?;
     clear_restored_package_dirs(RUNTIME_PACKAGE_ID)?;
+    clear_godot_intermediate_outputs()?;
```
bfops, line 225: please add a comment to this line briefly explaining why we added this (maybe link to this PR since the description is good)

bfops (approved):
Ship it

PR #5134, opened 2026-05-27
Add new money exchange TS template

JasonAtClockwork:
# Description of Changes

Add a simple account based exchange demo, each account is created with $100.00 and create a nickname. The data is tracked with a simple double entry for both accounts, you select an account and send money.
- TypeScript module 
- Simple React front-end

N/A

1 - Added a small template

# Testing

- [x] Ran through `spacetime init` against local


gefjon (approved):
I haven't actually reviewed this, but it's a demo, so pretty low-risk, and we'd really like to get it live quite soon for the hackathon. We can fix any issues in follow-ups.

JulienLavocat (approved):
LGTM

PR #5135, opened 2026-05-28
Wait to unsubscribe in csharp-testsuite until required ops have finished

joshua-spacetime:
# Description of Changes

After #4973, WASM procedures can execute concurrently with later operations on the same WebSocket.

Before this change, the C# regression testsuite queued several procedures, then immediately queued `UnsubscribeThen`. After #4973, the unsubscribe could be applied before the `SubscriptionEventOffset` procedure callback ran, clearing `MyTable` from the local subscribed cache. The callback then failed while asserting that the `offset-test:` row was present.

This change treats unsubscribe as a separate phase. It is scheduled after the main work is queued, but only starts once `waiting == 0`, so all callbacks that inspect subscribed state run before the cache is cleared.

N/A

1

# Testing

- [x] `csharp-testsuite` no longer flakes

### DIFF
--- a/sdks/csharp/examples~/regression-tests/client/Program.cs
+++ b/sdks/csharp/examples~/regression-tests/client/Program.cs
@@ -23,6 +23,8 @@ const ulong EXPECTED_TEST_EVENT_VALUE = 42;
 DbConnection db = null!;
 uint waiting = 0;
 var runComplete = false;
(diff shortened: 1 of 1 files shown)


joshua-spacetime:
Included in #5095 since #5095 made the test flake more consistent.

PR #5136, opened 2026-05-28
Isolate Godot NuGet restore from Unity package cache

joshua-spacetime:
# Description of Changes

A follow-on to #5133 that fixes another Godot flake that was observed after #5133 merged. Specifically a C# SDK regen flake where the Godot package restore path reused `sdks/csharp/packages`, which is also the Unity-visible package cache.

During `cargo regen csharp dlls`, the Godot flow previously did this:

- cleared `sdks/csharp/packages/godotsharp`
- cleared `sdks/csharp/obj~/godot`
- restored the Godot project back into `sdks/csharp/packages`
- packed the Godot project using that same Unity-visible package area

That meant a Godot-only cleanup could delete package content under the Unity SDK package cache. If a previous GodotSharp restore was partial or stale, regen tried to fix it by deleting `packages/godotsharp`, but that path is visible to Unity and can cause dirty/missing package content during regen.

With this change, Godot restore state is now isolated under `sdks/csharp/obj~/godot/packages`, and the regen flow becomes:

- clear only `sdks/csharp/obj~/godot`
- restore the Godot project into `sdks/csharp/obj~/godot/packages`
- pack the Godot project under the same path

Clearing `obj~/godot` now removes both Godot intermediates and Godot-only NuGet packages, while leaving the normal Unity SDK package cache under `sdks/csharp/packages` untouched.

N/A

1

# Testing

- [x] `unity-testsuite` passes

### DIFF
--- a/sdks/csharp/SpacetimeDB.ClientSDK.Godot.csproj
+++ b/sdks/csharp/SpacetimeDB.ClientSDK.Godot.csproj
@@ -19,7 +19,7 @@
     <AssemblyVersion>2.3.0</AssemblyVersion>
     <Version>2.3.0</Version>
     <DefaultItemExcludes>$(DefaultItemExcludes);*~/**</DefaultItemExcludes>
-    <RestorePackagesPath>packages</RestorePackagesPath>
+    <RestorePackagesPath>obj~/godot/packages</RestorePackagesPath>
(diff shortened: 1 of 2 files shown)


bfops (approved):
SGTM thank you!

PR #5140, opened 2026-05-28
Add TypeScript Blackholio server + client

JasonAtClockwork:
# Description of Changes

Adding the TypeScript version of Blackholio, both the server and client side.
- Mirrored the server-rust building server-ts targetting same name/accessors to have no client binding differences
- Added new TypeScript client using Phaser 4 - designed to be as close to others for future tutorial
- No doc updates yet

N/A

1 - Additive only for the /demo/Blackholio

# Testing

- [x] Tested client against server-rust and new server-ts
- [x] Server tested against new client and new bindings generation


gefjon (approved):
I haven't actually reviewed this, but it's a demo, so pretty low-risk, and we'd really like to get it live quite soon for the hackathon. We can fix any issues in follow-ups.

PR #5141, opened 2026-05-28
Make `RelationalDB::commit_tx_downgrade` fallible

joshua-spacetime:
# Description of Changes

`RelationalDB::commit_tx` is already fallible, so `commit_tx_downgrade` probably always should have been too.

This is mainly for the purpose of enforcing memory/resource limits. We could enforce such limits one level higher at the various call sites for reducers, procedures, sql etc. But we run the risk of missing a write path in that case. So we should probably have a failsafe mechanism in place which is to catch such errors at commit-time.

N/A

1

# Testing

Refactor

### DIFF
--- a/crates/core/src/db/relational_db.rs
+++ b/crates/core/src/db/relational_db.rs
@@ -831,8 +831,11 @@ impl RelationalDB {
         Ok(Some((tx_offset, tx_data, tx_metrics, reducer)))
     }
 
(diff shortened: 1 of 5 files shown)


gefjon (approved):
This doesn't actually add an `Err` return anywhere, yeah?

joshua-spacetime:
I'm going to close this for now, as we have decided that exceeding memory limits will not fail transactions for the time being.

PR #5145, opened 2026-05-29
Fix index schema from st tables.

### LINKED ISSUE #4701: JS runtime 2.0.4 derives internal index names from accessor keys instead of canonical schema names, causing "No such index" panics after publish
We hit a runtime bug in the JS `spacetimedb` package (`2.0.4`) that breaks published databases with existing snake_case schemas.

  ## Symptom
  - `spacetime publish` succeeds
  - the new instance then fatally crashes on first reducer call
  - subsequent CLI calls fail with `403` / `The instance encountered a fatal error`
  - logs show panics like `Uncaught No such index`

  ## What we observed
  It looks like internal runtime index/source names are being derived from JS accessor keys instead of canonical table/column names.

  Example:
  - table accessor: `cardDetails`
  - field/accessor: `nameKey`
  - runtime expected:
    - `cardDetails_nameKey_idx_btree`
  - actual published DB index name:
    - `card_details_name_key_idx_btree`

  Another example:
  - runtime expected:
    - `runtimeIslandStates_ownerId_idx_btree`
  - actual DB index name:
    - `runtime_island_states_owner_id_idx_btree`

  So the runtime looks up a non-existent index and panics.

  ## Why we think this is the cause
  We patched the installed runtime in `node_modules/spacetimedb/dist/server/index.mjs` with these two changes:

  1. Use canonical column names when building the index column list
  ```js
  colNameList.push(row.row[elem.name]?.columnMetadata?.name ?? elem.name);

  instead of:

  colNameList.push(elem.name);

  2. Use canonical table name instead of accessor name when building index source names

  const sourceName = index.sourceName = `${tableName}_${colS}_idx_${index.algorithm.tag.toLowerCase()}`;

  instead of:

  const sourceName = index.sourceName = `${accName}_${colS}_idx_${index.algorithm.tag.toLowerCase()}`;

  After that patch:

  - reducer calls stopped panicking
  - config reducers like set_storage_config worked again
  - runtime smoke tests passed on both staging and live published databases

  ## Environment

  - spacetimedb CLI/runtime version 2.0.4

  ## Expected behavior

  Runtime index/source-name resolution should be based on canonical schema names, not JS accessor identifiers.

Shubham8287:
# Description of Changes

fixes https://github.com/clockworklabs/SpacetimeDB/issues/4701.

`IndexSchema` when build from system tables, wasn't considering `st_index_accessor` table to set alias.

NA 

2.5

# Testing

Added a smoketest for typescript, this is more apparent in typescript due to `snake_case` name casing results in different canonical and accessor name.


crates/datastore/src/locking_tx_datastore/state_view.rs
```diff
@@ -187,7 +203,14 @@ pub trait StateView {
         // Look up the indexes for the table in question.
         let indexes = self
             .iter_by_col_eq(ST_INDEX_ID, StIndexFields::TableId, value_eq)?
-            .map(|row| StIndexRow::try_from(row).map(Into::into))
+            .map(|row| {
+                let row = StIndexRow::try_from(row)?;
+                let mut index_schema = IndexSchema::from(row);
+                index_schema.alias = self
+                    .find_st_index_accessor_row_by_index_name(index_schema.index_name.as_ref())?
+                    .map(|row| row.accessor_name);
```
joshua-spacetime, line 211: This restores the persisted `source_name` alias when rebuilding `IndexSchema`, which fixes the restart/recreate path, but I think auto-migration still has the same issue. `auto_migrate_indexes` only compares `accessor_name` and `algorithm`, not `IndexDef.source_name`. If a TS module changes the generated source name while keeping the same canonical index name/accessor/algorithm, the migration can succeed without updating `st_index_accessor`, and the runtime will still call `index_id_from_name` with the new source name.

Should we reject `source_name` changes as requiring manual migration, or add a step to update `st_index_accessor`?

crates/smoketests/tests/smoketests/typescript_index_source_name.rs
```diff
@@ -0,0 +1,105 @@
...
+      emailAddress,
+    });
+  },
+);
+"#;
+
+const TYPESCRIPT_MODULE_WITH_NEW_COLUMNS: &str = r#"import { schema, table, t } from "spacetimedb/server";
+
+const users = table(
+  { name: "users", public: false },
```
joshua-spacetime, line 37: Can we make the table accessor differ from the canonical table name? The reported issue involved runtime/source index names being built from both table and column accessor names while the stored index used canonical snake_case names. This test covers column accessor divergence but not table accessor divergence because both names are currently `users`.

crates/datastore/src/locking_tx_datastore/state_view.rs
Shubham8287: I have added a migration step.

crates/smoketests/tests/smoketests/typescript_index_source_name.rs
```diff
@@ -0,0 +1,154 @@
...
+  {
+    id: t.u64().primaryKey().autoInc(),
+    name: t.string(),
+    emailAddress: t.string().index("btree"),
+  },
+);
+
+const spacetimedb = schema({
+  renamedUsers,
```
joshua-spacetime: This renames the table accessor too. We're migrating `st_index_accessor`, but what about `st_table_accessor`?

Shubham8287: I think I should fix `accessor` name changes in separate PR, as that change requires more thorough considerations. 

I am keeping the current PR to fix building correct `IndexSchema`  from sys tables, Which will unblock the table migration path. As people will run into table migration more frequently than changing accessor name especially during Hackothon

joshua-spacetime: That sounds reasonable to me.

joshua-spacetime (approved):

PR #5148, opened 2026-05-29
Update rust `test-client` module bindings to `2.3.0`

rekhoff:
# Description of Changes

This regenerates the Rust test's `test-client` module bindings to the `2.3.0` version of SpacetimeDB.

N/A

1 - Only regen of bindings

# Testing

None, if CI jobs pass that should be enough given the nature of the change.


rekhoff:
Outdated and no longer needed

PR #5149, opened 2026-05-29
Remove physical module layout info from read sets

### LINKED ISSUE #4947: [CRITICAL] Views recomputation can roll back transactions.
I experienced a bug where the view could not be recomputed and it causing it to propagate the error to the reducer which called it.

This panic is also not in the logs of the databse so it just siltently rolls back the data. (Internal Server Error is returned for reducers i dont know if it times out a computation limit or something )

Sadly i lost quite a bit of mission critical stuff in that incident because this reducer is built to always archive it and never panic :(

Read only data should not be able to impact a successfully commiting transaction!

I will ping @cloutiertyler and @joshua-spacetime because i think this is a _very_ severe bug hope yall understand :>

If you have further questions pls reply or ping (also possible on discord) but i will go to sleep for now because it has been stressful figuring out the bug

joshua-spacetime:
# Description of Changes

I believe this to fix #4947.

Previously, committed read sets stored module-local indexes used for view dispatch. These indexes are not stable across module updates, but they were not removed from the committed state, so that later when a write triggered a view refresh, the runtime could dispatch the wrong view function or attempt to materialize into the wrong backing table which would result in a fatal error.

This patch removes these unstable indexes from read sets. The core behavior change is that committed read sets no longer depend on stale per-module layout details. They identify the logical view call, and refresh resolves the current module metadata when needed.

N/A

2

# Testing

- [x] Auto-migrate smoketests

### DIFF
--- a/crates/core/src/db/relational_db.rs
+++ b/crates/core/src/db/relational_db.rs
@@ -16,7 +16,9 @@ use spacetimedb_datastore::locking_tx_datastore::datastore::TxMetrics;
 use spacetimedb_datastore::locking_tx_datastore::state_view::{
     IterByColEqMutTx, IterByColRangeMutTx, IterMutTx, StateView,
 };
-use spacetimedb_datastore::locking_tx_datastore::{ApplyHistoryCounters, IndexScanPointOrRange, MutTxId, TxId};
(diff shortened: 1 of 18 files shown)


Shubham8287 (approved):
hmm, looks good.

Issue #4970, opened 2026-05-06
Translate bash/Python scripts into Rust


bfops:
We have many bash scripts floating around in e.g. `tools` and `sdks/csharp/tools~`. We would like these translated into (platform-independent) Rust.

Note that some scripts like `sdks/csharp/tools~/gen-*` should probably be a part of #4823 instead.

bfops:
ports the doctests.sh script https://github.com/clockworklabs/SpacetimeDB/pull/5155

PR #5155, opened 2026-05-31
Port `bindings-doctests.sh` into CI

bfops:
# Description of Changes

Part of https://github.com/clockworklabs/SpacetimeDB/issues/4970.

Doing this part in a separate PR because it turns out that this test was failing, so this PR also includes the fixes.

None

1

# Testing

CI passes (but only with the "code" changes).


crates/bindings/src/lib.rs
```diff
@@ -365,9 +365,9 @@ pub use spacetimedb_bindings_macro::settings;
 ///     // The following line would panic, since we use `insert` rather than `try_insert`.
 ///     // let result = ctx.db.country().insert(Country { code: "CN".into(), national_bird: "Blue Magpie".into() });
 ///
-///     // If we wanted to *update* the row for Australia, we can use the `update` method of `UniqueIndex`.
-///     // The following line will succeed:
-///     ctx.db.country().code().update(Country {
+///     // If we wanted to replace the row for Australia, we can delete it and insert the new row.
+///     assert!(ctx.db.country().code().delete("AU".to_string()));
+///     ctx.db.country().insert(Country {
```
bfops, line 370: AI says:
>   UniqueColumn::update is now constrained to primary-key columns only:
>
> ```
>   pub fn update(&self, new_row: Tbl::Row) -> Tbl::Row
>   where
>       Col: PrimaryKey,
> ```
> The docs in table.rs explicitly say this is intentional: for a non-primary unique column, use .delete(key) > followed by .insert(row). So the old text saying “update the row for Australia” with ctx.db.country().code().update(...) failed doctests because code is unique, not primary key.

tools/ci/src/main.rs
```diff
@@ -616,6 +616,10 @@ fn main() -> Result<()> {
                 .dir("crates/bindings-csharp")
                 .run()?;
             cmd!("pnpm", "lint").run()?;
+            cmd!("cargo", "test", "--doc", "--target", "wasm32-unknown-unknown")
+                .dir("crates/bindings")
+                .run()?;
+            cmd!("cargo", "test", "--doc").dir("crates/bindings").run()?;
             // `bindings` is the only crate we care strongly about documenting,
```
jdetter, line 623: Do we also need to do this 3rd step or is it already covered?
```
# And look for broken links
RUSTDOCFLAGS="-D warnings" cargo doc
```

jdetter (commented):
Question on this one

crates/bindings/src/lib.rs
cloutiertyler: This is correct. Were we not testing these docs before at all?

cloutiertyler (approved):
Looks good, but I don't see where this is actually adding these tests to CI.

bfops:
> Looks good, but I don't see where this is actually adding these tests to CI.

They're added here: https://github.com/clockworklabs/SpacetimeDB/pull/5155/changes#diff-8eeef336da059361b3b8f02cf2bae974d463bf491e998acef11397dd68038eb4R619-R622

tools/ci/src/main.rs
bfops: it's done here: https://github.com/clockworklabs/SpacetimeDB/pull/5155/changes/BASE..b09bbc14ec03b4aa21b09bec2efe098b1a7a5010#diff-8eeef336da059361b3b8f02cf2bae974d463bf491e998acef11397dd68038eb4R628-R632

jdetter (approved):
Seems fine, not manually tested by me.

cloutiertyler:
Ahh sorry, I thought that was adding a new command, not adding it to an existing command. Apologies.

crates/bindings/src/lib.rs
bfops: Unfortunately it seems that way. The original script had a comment about why it wasn't integrated into CI, but the reasoning is now outdated. We're now running this in CI!

PR #5159, opened 2026-06-01
Drop tps threshold for rust keynote-2 ci check

joshua-spacetime:
# Description of Changes

The keynote benchmark check has had a couple spurious failures in CI recently, namely for the rust module. I set a higher TPS threshold for rust than I did for typescript since rust is a bit faster, but it looks like the threshold I set is within the margin of error, so this change drops it to 275K - exactly what it is for typescript.

A failure now almost certainly represents a real performance regression and requires deeper investigation.

N/A

0 (one liner)

# Testing

Should have no more spurious failures for this CI job.


bfops (approved):
works for me!

PR #5163, opened 2026-06-01
Capture cpu profile for keynote bench in nightly job

joshua-spacetime:
# Description of Changes

Also capture a cpu profile on a keynote bench job failure.

N/A

2

# Testing

- [ ] Profile artifacts are generated and stored by the CI job

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -348,7 +348,8 @@ jobs:
           exit "$status"
 
       - name: Upload keynote benchmark profile
-        if: failure() && steps.keynote_bench.outcome == 'failure'
+        if: always()
+        # if: failure() && steps.keynote_bench.outcome == 'failure'
         uses: actions/upload-artifact@v4
         with:
           name: keynote-bench-profile
@@ -357,7 +358,8 @@ jobs:
           retention-days: 14
 
       - name: Report keynote benchmark profile instructions
-        if: failure() && steps.keynote_bench.outcome == 'failure'
+        if: always()
+        # if: failure() && steps.keynote_bench.outcome == 'failure'
         run: |
           {
             echo "## Keynote benchmark profile"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -273,8 +273,11 @@ jobs:
     runs-on: spacetimedb-benchmark-runner
     timeout-minutes: 60
     env:
-      CARGO_TARGET_DIR: ${{ github.workspace }}/target
       RUST_BACKTRACE: full
+      CARGO_TARGET_DIR: ${{ github.workspace }}/target
+      CARGO_PROFILE_RELEASE_DEBUG: line-tables-only
+      KEYNOTE_PROFILE_DIR: ${{ github.workspace }}/keynote-profile
+      STDB_V8_FLAGS: "--perf-basic-prof --perf-basic-prof-only-functions --interpreted-frames-native-stack"
     steps:
       - name: Find Git ref
         env:
@@ -306,7 +309,14 @@ jobs:
           prefix-key: v1
 
       - name: Build keynote benchmark binaries
-        run: cargo build --release -p spacetimedb-cli -p spacetimedb-standalone
+        run: >
+          cargo
+          --config 'build.rustflags = ["-C", "force-frame-pointers=yes"]'
+          build
+          --release
+          -p spacetimedb-cli
+          -p spacetimedb-standalone
+          --features spacetimedb-standalone/perfmap
 
       # Node 24 is the current Active LTS line.
       - name: Set up Node.js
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -319,8 +329,54 @@ jobs:
         run: pnpm build
         working-directory: crates/bindings-typescript
 
+      - name: Install samply
```
joshua-spacetime: TODO: This should be installed at the infra level - not by the test every time.

.github/workflows/ci.yml
```diff
@@ -319,8 +329,54 @@ jobs:
...
+          fi
+          samply --version
+
       - name: Run keynote-2 benchmark regression check
-        run: cargo ci keynote-bench
+        id: keynote_bench
+        run: |
+          mkdir -p "$KEYNOTE_PROFILE_DIR"
+          set +e
+          samply record \
```
joshua-spacetime, line 340: TODO: Need to set `/proc/sys/kernel/perf_event_paranoid` to `1` in the runner in order to give samply access to kernel-level perf events.

.github/workflows/ci.yml
cloutiertyler: You mean baked into the image?

.github/workflows/ci.yml
cloutiertyler: Is that a requirement for this to work?

cloutiertyler:
I'd like @jdetter to review this one.

.github/workflows/ci.yml
joshua-spacetime: To capture a cpu profile, yes.

cloutiertyler: Perhaps this TODO should be a code comment. Otherwise we might forget and perhaps the DevOps team won't realize.

joshua-spacetime: Sorry for the confusion. These TODOs are for me. I'll get them resolved before merging.

.github/workflows/keynote-bench-nightly.yml
```diff
@@ -0,0 +1,108 @@
...
+          ls -lh "$KEYNOTE_PROFILE_DIR" || true
+
+      - name: Upload keynote benchmark profile
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: keynote-bench-nightly-profile
+          path: ${{ env.KEYNOTE_PROFILE_DIR }}
+          if-no-files-found: warn
+          retention-days: 90
```
joshua-spacetime, line 92: Note for @jdetter: I'm not sure if or how much of an impact this will have on our github actions spend. Happy to reconfigure this. I would just like some level of cpu profile history for diagnostic purposes.

jdetter: This should be fine for now, if the spend here becomes too large then we can always move these to an S3 bucket or something

jdetter: How large are these artifacts?

.github/workflows/ci.yml
```diff
@@ -273,8 +273,11 @@ jobs:
     runs-on: spacetimedb-benchmark-runner
     timeout-minutes: 60
     env:
-      CARGO_TARGET_DIR: ${{ github.workspace }}/target
       RUST_BACKTRACE: full
+      CARGO_TARGET_DIR: ${{ github.workspace }}/target
+      CARGO_PROFILE_RELEASE_DEBUG: line-tables-only
```
jdetter, line 278: This may cause cargo cache misses with the new vm caching change but that's more of a me problem for later

.github/workflows/ci.yml
```diff
@@ -306,7 +309,14 @@ jobs:
...
       - name: Build keynote benchmark binaries
-        run: cargo build --release -p spacetimedb-cli -p spacetimedb-standalone
+        run: >
+          cargo
+          --config 'build.rustflags = ["-C", "force-frame-pointers=yes"]'
+          build
+          --release
+          -p spacetimedb-cli
+          -p spacetimedb-standalone
+          --features spacetimedb-standalone/perfmap
```
jdetter, line 319: This will also probably cause a cache miss - if this workflow ends up being too slow I can bake a new cache with all of these options enabled so we're more likely to get a compilation cache hit.

.github/workflows/keynote-bench-nightly.yml
```diff
@@ -0,0 +1,108 @@
+name: Keynote Bench Nightly
+
+on:
+  schedule:
+    # Nightly at 3 AM UTC.
+    - cron: '0 3 * * *'
```
jdetter, line 6: I would prefer this to run more around 8am UTC - when we do maintenance on the runners we typically like to do it in the middle of the night when nobody is working to reduce the impact on productivity. If you keep this as 3AM UTC you will just run the risk of it getting cancelled sometimes.

jdetter (approved):
Nothing blocking this from merging, just a few comments and a question

.github/workflows/keynote-bench-nightly.yml
joshua-spacetime: ~4MB each.

.github/workflows/ci.yml
joshua-spacetime: FWIW, I think `--features spacetimedb-standalone/perfmap` should be the default for all of our release builds.

